PR #86, opened 2022-07-01
Report error events to Monorail

gonzaloriestra:

Related to https://github.com/Shopify/shopify-cli-planning/issues/274

- Saves the current command and args from the prerun hook to have it when it fails (postrun is not called then)
- Sends a report to Monorail when the command fails, including the error message
- Fixes the `create-hydrogen` command, which was sending `create-app` by mistake

### How to test your changes?

- Comment out [this line](https://github.com/Shopify/shopify-cli-next/blob/8d24d8311e8c4a07161f932ffef1e5762b589363/packages/cli-main/bin/dev.js#L5) to disable debug mode
- `yarn shopify app info` (from a folder without an app)


packages/cli-kit/src/analytics.ts
```diff
@@ -55,7 +59,7 @@ const buildHeaders = (currentTime: number) => {
   }
 }
 
-const buildPayload = async (command: string, args: string[] = [], currentTime: number) => {
+const buildPayload = async (errorMessage: string | undefined, currentTime: number) => {
```
pepicrft, line 62: What about using an options type here to make it more obvious from the caller what arguments are? Looking at the following code it's hard to tell that the first string is an error message:

```js
analytics.reportEvent("foo")
```

pepicrft (approved):

PR #87, opened 2022-07-01
Load version from cli-kit with relative path

isaacroldan:
We weren't loading the package.json from the correct path to obtain the package version

- Specify a relative URL to load the cli-kit package.json

### How to test your changes?
These commands work and show the correct version: 
`create-app` , `app info`, `version`

### DIFF
--- a/packages/cli-kit/src/constants.ts
+++ b/packages/cli-kit/src/constants.ts
@@ -53,7 +53,8 @@ const constants = {
    */
   versions: {
     cliKit: async () => {
-      const packageJsonPath = (await findUp('package.json', {type: 'file'})) as string
+      const relativePath = pathJoin('node_modules', '@shopify', 'cli-kit', 'package.json')
+      const packageJsonPath = (await findUp(relativePath, {type: 'file'})) as string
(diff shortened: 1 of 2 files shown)


packages/cli-kit/src/constants.ts
```diff
@@ -53,7 +53,8 @@ const constants = {
    */
   versions: {
     cliKit: async () => {
-      const packageJsonPath = (await findUp('package.json', {type: 'file'})) as string
+      const relativePath = pathJoin('node_modules', '@shopify', 'cli-kit', 'package.json')
```
pepicrft, line 56: Correct me if I'm wrong, but `join` by default makes the path relative to the working directory. I'd change that to be relative to the directory that contains the module, `.../dist/constant.js`:
```js
pathJoin(__dirname, 'node_modules', '@shopify', 'cli-kit', 'package.json')
```

You can use [this function](https://github.com/Shopify/cli/blob/main/packages/cli-kit/src/path.ts#L37) to obtain the directory of the module: `moduleDirectory(import.meta.url)`

pepicrft:
Ready to merge?

pepicrft (approved):

PR #90, opened 2022-07-01
Fix bugsnag notify

isaacroldan:
We weren't correctly reporting some of the bugs

- Refactor how we call Bugsnag notify if we receive a non-error object
- Add await to the sourcemap uploader


packages/cli-main/src/index.ts
```diff
@@ -39,21 +39,24 @@ async function runCLI() {
...
-    }
-    await new Promise((resolve, reject) => {
-      Bugsnag.notify(mappedError, undefined, resolve)
-    })
+  let mappedError: Error
+
+  // eslint-disable-next-line no-prototype-builtins
+  if (Error.prototype.isPrototypeOf(errorToReport)) {
+    mappedError = new Error(errorToReport.message)
+    if (errorToReport.stack) mappedError.stack = errorToReport.stack.replace(new RegExp('file:///', 'g'), '/')
```
pepicrft, line 49: Do Windows stacktraces have `file:///` as prefix? [Here's](https://app.bugsnag.com/shopify/cli/errors/62bf091dacaaaf00089a6652?event_id=62bf091d009557501f540000&i=sk&m=nw) an example of a Windows error. What about using `fileURLToPath` from the `node:url` module to standardize the path?

isaacroldan: good catch, i'll try that

isaacroldan: So after some investigation, node does add the `file:///` prefix even in windows:
IMG_0555AD2A135D-1

I've update the regex to take that pattern into account for windows.

packages/cli-kit/src/node/cli.ts
```diff
@@ -78,21 +78,28 @@ export async function runCreateCLI(options: RunCLIOptions) {
...
-      mappedError = errorToReport
+  let mappedError: Error
+
+  // eslint-disable-next-line no-prototype-builtins
+  if (Error.prototype.isPrototypeOf(errorToReport)) {
+    mappedError = new Error(errorToReport.message)
+    if (errorToReport.stack) {
+      // For mac/linux, remove `file:///` from stacktrace
+      // For windows, remove `file:///C:/` from stacktrace
+      mappedError.stack = errorToReport.stack.replace(new RegExp('file:///([a-zA-Z]:/)?', 'g'), '/')
```
gonzaloriestra, line 91: What about removing everything before `node_modules`?

- `file:///C:/Users/isaac/test-bugsnag/node_modules/@shopify/cli/src/index.ts:52` => `@shopify/cli/src/index.ts:52`
- `file:///home/gonzalo/projects/my-hydrogen-app/node_modules/swiper/swiper.cjs.js:14` => `swiper/swiper.cjs.js:14`

isaacroldan: Umm, we could do it, bugsnag works as it is right now, would it be useful to apply this? 🤔

gonzaloriestra: That part can contain sensitive information and we don't need it...

isaacroldan: I can definitely remove it from the stack, but bugsnag will still report the full path... (don't know how they do it)

isaacroldan: Updated the regular expression, you can see how it behaves here: https://regex101.com/r/1lvUqX/1

gonzaloriestra:
Awesome, thanks!

But you added two backslashes at the beginning in the code:

const regex = ***@***.***/)?'

It should be just one, shouldn't it?

PR #96, opened 2022-07-04
Fix path outputs for the working directory

gonzaloriestra:

When using `output.token.path` to print the current directory, we are showing and empty string instead.

This is breaking some messages like this one:

04-14-url0w-uys9g

Show the full path when it is the current one:

04-16-oc4jf-w4tec

### How to test your changes?

`yarn shopify app info`(without a configuration file there)

### DIFF
--- a/.changeset/sweet-phones-lie.md
+++ b/.changeset/sweet-phones-lie.md
@@ -1,0 +1,6 @@
+---
+'@shopify/app': patch
+'@shopify/cli-kit': patch
+---
+
+Fix path outputs for the working directory


packages/cli-kit/src/path.ts
```diff
@@ -22,7 +22,7 @@ export function relativize(path: string): string {
   const result = commondir([path, process.cwd()])
   const relativePath = relative(process.cwd(), path)
   const relativeComponents = relativePath.split('/').filter((component) => component === '..').length
-  if (result === '/' || relativeComponents > 2) {
+  if (result === '/' || relativePath === '' || relativeComponents > 2) {
```
isaacroldan, line 25: How does this behave in windows? the root path there would be `C:/` (or some other lette) i guess?

pepicrft (approved):

PR #99, opened 2022-07-04
Split hydrogen template question

### LINKED ISSUE #49: [Feature]: S

Hydrogen custom storefront

Substantial change to existing feature

### Overview

Split hydrogen "choose your template?" question into 2 questions: 

1) for choosing the template 
Screenshot 2022-06-24 at 06 36 59

2) for choosing the language

Screenshot 2022-06-24 at 06 37 06

### Motivation

This pattern follows conventions set by other CLIs, as well as the rest of the Shopify CLI.

SherRao:
Fixes #49 - Now the CLI will ask which template to use, as well as which language to go along with it as two separate questions.

#### Before
https://user-images.githubusercontent.com/11851551/177179090-0d4bf45a-9982-4aa5-856d-bd99fe996111.mov

#### After
https://user-images.githubusercontent.com/11851551/177179031-ac39cd88-3ea4-495f-82c7-1d6471b3565e.mov

### How to test your changes?
Run the `dev create-hydrogen` command.


gfscott:
👋🏻 Also: cc @mcvinci re Hydrogen CLI docs: https://shopify.dev/custom-storefronts/tools/cli-hydrogen

packages/create-hydrogen/src/prompts/init.ts
```diff
@@ -1,40 +1,34 @@
...
-const TEMPLATE_NAMES = Object.keys(TEMPLATE_DATA)
-
-function toHydrogenTemplateName(key: string) {
-  const normalized = string.hyphenize(key).toLocaleLowerCase()
-  const withExtension = normalized.endsWith('-ts') || normalized.endsWith('-js') ? normalized : `${normalized}-js`
-
-  return TEMPLATE_NAMES.includes(withExtension) ? withExtension : false
+const LANGUAGE_NAME_DATA = {
+  js: {
+    description: 'JavaScript',
```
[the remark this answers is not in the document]
SherRao: This sounds good - since I'll be rewriting some stuff to fix passing explicit template's through the CLI, might also look into throwing this in there too :)

packages/create-hydrogen/src/prompts/init.ts
```diff
@@ -48,42 +46,73 @@ const init = async (options: InitOptions, prompt = ui.prompt): Promise<Required<
...
+    parsedTemplate.user === 'Shopify' &&
+    parsedTemplate.subDirectory.startsWith('templates/')
+  const template = looksLikeHydrogenTemplate && missingBranch ? `${parsedTemplate.full}#${BRANCH}` : parsedTemplate.full
+  output.info(template)
+  return {name, template, language} as Required<InitOptions>
+}
+
+/**
+ * Checks if the provided name is a Shopify template, or if it is a custom URL to a user-provided template.
+ *
```
lordofthecactus, line 105: this is great! we should use more docs like this

packages/create-hydrogen/src/prompts/init.ts
```diff
@@ -1,40 +1,38 @@
...
+const LANGUAGE_NAME_DATA = {
+  js: {
+    name: 'JavaScript',
+    description: 'This is the default language, regular old JavaScript',
   },
-  'hello-world-ts': {
-    description: 'Hello World (TypeScript)',
+  ts: {
+    name: 'TypeScript',
+    description: 'A typed superset of JavaScript created by Microsoft',
```
lordofthecactus, line 24: I think I would use different descriptions for these. @gfscott what do you think? 

Or not use any.

lordofthecactus (commented):
Great work!!! We are close to it.
It is unfortunate that we don't have much tests here, could you add them?

I tophatted the code manually and noticed something:
* Before the changes, the code checked for the template and failed before even asking for any input. I think this is good dev experience.
* After the change,  I could get asked the language and the name and at the end fail. 

Could you check the templates exist before asking for input?

Another thing that I mentioned in slack and maybe it was missed: Rather than using --language typescript|javascript
I think it would be better dev experience to shorten it to:
`--ts` if we need typescript, and no need to add anything if we want javascript

Happy to pair tomorrow on this if you'd like!

benjaminsehl:
Very happy to see this!! Was on a wish list to get done, so thanks @SherRao!!

packages/create-hydrogen/src/prompts/init.ts
```diff
@@ -48,42 +46,73 @@ const init = async (options: InitOptions, prompt = ui.prompt): Promise<Required<
...
+  let language
+  const promptResults = await prompt(questions)
+  if (options.name) name = options.name
+  else name = promptResults.name
+
+  if (options.template) templateName = options.template
+  else templateName = promptResults.templateName
+
+  if (options.language) language = options.language
+  else language = promptResults.language
```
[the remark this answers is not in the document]
isaacroldan: You should validate that the options are not only non-empty but that they have an accepted value

packages/create-hydrogen/src/prompts/init.ts
```diff
@@ -48,42 +46,73 @@ const init = async (options: InitOptions, prompt = ui.prompt): Promise<Required<
...
+    parsedTemplate.name === 'hydrogen' &&
+    parsedTemplate.user === 'Shopify' &&
+    parsedTemplate.subDirectory.startsWith('templates/')
+  const template = looksLikeHydrogenTemplate && missingBranch ? `${parsedTemplate.full}#${BRANCH}` : parsedTemplate.full
+  output.info(template)
+  return {name, template, language} as Required<InitOptions>
+}
+
+/**
+ * Checks if the provided name is a Shopify template, or if it is a custom URL to a user-provided template.
```
rennyG, line 104:
```suggestion
 * Checks if the provided name is a Shopify template, or if it's a custom URL to a user-provided template.
```

packages/create-hydrogen/src/prompts/init.ts
```diff
@@ -1,89 +1,143 @@
...
+ * @param templateName The given name of a Shopify template.
+ * @returns The URL of the template.
+ */
+const convertTemplateNameToUrl = (templateName: string): string => `${TEMPLATE_BASE}${templateName}#${BRANCH}`
+
+/**
+ * Checks to see if the provided URL can be parsed by Github. For Shopify-specific templates, adds any
+ * missing information (missing branch for example) to the repo information in the URL.
+ *
+ * @param templateUrl The URL of the template to parse.
```
rennyG, line 129:
```suggestion
 * @param templateUrl: The URL of the template to parse.
```

cartogram (changes requested):
Just did a quick tophat and looks like there are still a few issues.

1. I am seeing output displayed to the user that should not be there. Such as:
```
!options.language: true
!isAShopifyTemplateName: false
```

2. If I run with no flags, I am never asked what language I want.
3. The flow of questions has been inverted.
4. I also wasn't able to successfully complete the command.

Screenshot 2022-07-27 at 17 22 24

lordofthecactus (commented):
@SherRao ✅ I went thought the code and tophatted all the cases we discussed and find it works great!!!

This is what I got in the tophatting. Is this correct? @cartogram I'm keen on approving this, if you are ok with it. @SherRao is doing early failing which I find great 💯 
| command | ask for template | asks for language | asks for name |
|---|---|---|---|
| `dev create-hydrogen`  | true   | true | true | 
|  `dev create-hydrogen --ts` | true   | false (sets as typescript)   | true   |  
| `dev create-hydrogen --template demo-store`  | false (sets demo-store) | true  | true  |
| `dev create-hydrogen --template hello-world` |  false (sets hello-world) | true | true |
| `dev create-hydrogen --template github:lordofthecactus/hydrogen --ts` | false (github repo) | false (gets ignored) | true |

I have one more request @SherRao before we ask to update the docs:
I did not find tests for all of the cases above. Should we add tests for the following cases?
* if language is not provided, it should ask for it
* if --ts is provided typescript should be selected
* if github repo is given, language should not be asked

Mainly I would cover the cases above to make sure nobody breaks the flows we created. wdyt?

lordofthecactus (approved):
Great! @SherRao, could you wait for documentation changes? cc. @gfscott do you know what should we update for this? I describe the behavior in this comment: https://github.com/Shopify/cli/pull/99#pullrequestreview-1061909803

.changeset/small-rabbits-fail.md
```diff
@@ -0,0 +1,5 @@
+---
+'@shopify/create-hydrogen': patch
+---
+
+Separate create-hydrogen template question into template and language questions.
```
lordofthecactus, line 5: @cartogram looks like a possible major?
If someone coded commands to be `dev create-hydrogen --path ../cli-tests --template hello-world-ts` it won't work anymore, they will have to use:
```
dev create-hydrogen --path ../cli-tests --template hello-world --ts
```

lordofthecactus: thoughts?

SherRao (commented):
Changes added

PR #101, opened 2022-07-04
Export `ruby`, `dot-env`, and `checksum` modules from `@shopify/cli-kit`

pepicrft:
These changes are part of the effort of improve the ergonomics consuming and mocking the modules from `@shopify/cli-kit`.

I moved `checksum`, `dot-env` and `ruby` to the `src/node` directory of `@shopify/cli-kit` and took the opportunity to add missing documentation and renamed some exports to have a more explicit name. 

### How to test your changes?

Run any command with the app in `fixtures/app`, for example `app dev`.


gonzaloriestra (approved):
Why the `node` folder? Couldn't we have them directly in the `cli-kit` folder?

pepicrft:
> Why the node folder? Couldn't we have them directly in the cli-kit folder?

To differentiate the code that's dependent on a runtime (e.g. Node), or runtime-agnostic. In the near future we might provide runtime APIs that run in other environments (e.g. browser), so it's hygienic that our module hierarchy reflects that. We can also leverage Node [conditional exports](https://nodejs.org/api/packages.html#conditional-exports) and build tools to detect if we are trying to load Node code in a Browser runtime.

PR #105, opened 2022-07-04
Increase yarn timeout to handle slower CI installs on Windows/Mac

amcaplan:

We've had many CI builds failing due to yarn install taking too long.

Extends the timeout from the 30-second default to 5 minutes. According to [this](https://github.com/yarnpkg/yarn/issues/8242#issuecomment-661881292) it should be a good enough workaround, as the problems are likely caused by the filesystem of Windows/Mac being significantly slower than Linux, so it needs some extra time.

### How to test your changes?

CI


amcaplan:
Cowboy-merging as this only affects CI

PR #111, opened 2022-07-05
Rename the `dependency` module to `node-package-manager` and move it to `src/node`

pepicrft:
This is the continuation work of [this PR](https://github.com/Shopify/cli/pull/101) in the effort of improving the ergonomics of contributing to the project.

- [x] Rename `dependency` to `node-package-manager`
- [x] Export it so that it can be imported as `import { ... } from "@shopify/cli-kit/node/node-package-manager"`
- [x] Rename `dependencyManager` to `packageManager` throughout the codebase to align with the ecosystem's naming convention
- [x] Change imports and tests to accommodate to the changes

### How to test your changes?
Create a project and try to `dev` it.


packages/cli-kit/src/analytics.test.ts
```diff
@@ -1,214 +1,284 @@
...
+          uname: 'darwin arm64',
+          cli_version: version,
+          ruby_version: '3.1.1',
+          node_version: process.version.replace('v', ''),
+          is_employee: false,
+          api_key: undefined,
+          partner_id: undefined,
+        },
+      }
+      expect(http.fetch).toHaveBeenCalled()
```
gonzaloriestra, line 90:
```suggestion
```

packages/cli-kit/src/analytics.ts
```diff
@@ -90,9 +90,9 @@ const buildPayload = async (errorMessage: string | undefined, currentTime: numbe
   return {
     schema_id: 'app_cli3_command/1.0',
     payload: {
-      project_type: await getProjectType(join(directory, 'web')),
+      project_type: await getProjectType(joinPath(directory, 'web')),
```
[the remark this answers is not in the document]
gonzaloriestra: Oops, I changed it to receive a string recently but I forgot to update that part. Thanks!

packages/cli-kit/src/node/node-package-manager.ts
```diff
@@ -204,8 +256,18 @@ interface AddNPMDependenciesIfNeededOptions {
...
+ */
 export interface DependencyVersion {
+  /**
+   * Dependency name
+   */
   name: string
+
+  /**
+   * Dependency version
+   */
```
gonzaloriestra, line 270: This is something we should discuss as a team, but I think this kind of comments don't add much value... I'm talking about this specific example, I find it useful for other cases.

packages/create-app/src/commands/init.ts
```diff
@@ -37,9 +37,9 @@ export default class Init extends Command {
       env: 'SHOPIFY_FLAG_TEMPLATE',
     }),
     // eslint-disable-next-line @typescript-eslint/naming-convention
-    'dependency-manager': Flags.string({
+    'package-manager': Flags.string({
```
gonzaloriestra, line 40: Did we have this flag documented somewhere?

packages/cli-kit/src/analytics.ts
```diff
@@ -108,3 +108,20 @@ const buildPayload = async (errorMessage: string | undefined, currentTime: numbe
     },
   }
 }
+
+export type ProjectType = 'node' | 'php' | 'ruby' | undefined
+
+export async function getProjectType(directory: string): Promise<ProjectType> {
```
gonzaloriestra, line 114: If this is only for internal use, I wouldn't export it and change the tests to check the `project_type` reported for every case.

Otherwise, I would move it to another file, it feels weird to import `getProjectType` from an analytics module, don't you think?

gonzaloriestra (approved):
Great! 👏 

But next time, it will be easier to review if you split it into several PRs when there are different things 🙏

packages/app/package.json
```diff
@@ -21,7 +21,7 @@
     "prepack": "cross-env NODE_ENV=production yarn run build && cp ../../README.md README.md",
     "lint": "prettier -c src/** && eslint \"src/**/*.ts\"",
     "lint:fix": "prettier -w src/** && eslint 'src/**/*.ts' --fix",
-    "test": "vitest run",
+    "test": "vitest run --isolate",
```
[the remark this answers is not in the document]
pepicrft: I was debugging something and forgot to remove it. I'll revert it.

packages/create-app/src/commands/init.ts
pepicrft: This is a hidden flag that's only used by acceptance tests to test workflows with different package managers.

packages/cli-kit/src/analytics.ts
pepicrft: Good idea. I've [made ](https://github.com/Shopify/cli/pull/111/commits/5457538054e7d0a420362231b471502df28f4b4a#diff-04dd7360e7e77135b59ccb6c6dc9bf5405b8003d0e93785fb2188c007b2ea186R60)the function internal and updated the `reportEvent` tests to test all the project type scenarios.

Issue #112, opened 2022-07-05
[Bug]: Please update ThemeCheckVersion to 1.10.3


awd:
### Please confirm that you have:

- [X] Searched [existing issues](.) to see if your issue is a duplicate. (If you’ve found a duplicate issue, feel free to add additional information in a comment on it.)
- [X] Reproduced the issue in [the latest CLI version](https://www.npmjs.com/package/@shopify/cli).

### In which of these areas are you experiencing a problem?

Extension

### Expected behavior

CLI V3 has an outdated version of theme-check, `1.10.2`. There is a subtle but important change in `1.10.3` which allows theme app extensions to `yarn deploy` without complaining about the use of the `app` liquid drop.

This is a big blocker from working on and testing out checkout-ui-extensions as the deploy fails on the theme check task.

### Actual behavior

Theme check should not fail when encountering the `app` liquid drop, as per `theme-check v1.10.3` 

### Reproduction steps

1. Bump the version of theme check found in: https://github.com/Shopify/cli/blob/402cf14649b6ec3501d9738992f21a22c67e984f/packages/cli-kit/src/node/ruby.ts#L15

### Operating System

Mac OS Monterey

### Shopify CLI version (check your project's `package.json` if you're not sure)

3.0.27

### Shell

iTerm2

### Node version (run `node -v` if you're not sure)

v18.3.0

### What language and version are you using in your application?

3.1.2

awd:
When `theme-check` is running on the current version, this is the error received:

`blocks/app-embed.liquid:14: error: UndefinedObject: Undefined object `app`.`

The theme-check change log shows it is fixed in `v1.10.3`: 
https://github.com/Shopify/theme-check/blob/main/CHANGELOG.md

PR #113, opened 2022-07-05
Bumping theme-check version to 1.10.3

awd:
Fixes https://github.com/Shopify/cli/issues/112

Bumping `theme-check` version to unblock `yarn deploy` (needed for checkout-ui-extension development)

This PR bumps the version of `theme-check` to the latest version, which does not fail when checking `theme-app-extensions` using `app` liquid drops.


amcaplan:
Gonna close and reopen to see if that'll pick up the workflow changes which should allow this to run.

Issue #4, opened 2022-06-21
Support for cloning private repositories


pepicrft:
[Reference](https://github.com/Shopify/shopify-cli-next/issues/233)

Some users might need to scaffold and create apps using templates that live in private repositories. When that happens and the `git` CLI prompts the user to authenticate the CLI hangs indefinitively.

PR #114, opened 2022-07-05
Fix support for using private template repo when no git credentials setup

craigmichaelmartin:

Fixes https://github.com/Shopify/cli/issues/4

Attempting to use a private github repo as a template for `create-app` would result in infinite "hanging" if git credentials were not already configured.

This would happen because git (if no credentials configured for the remote) would prompt the user for credentials (username and password) but from within a `listr` controlled prompt, where `listr` is continuously re-writing the terminal, so the terminal is re-written before the user could interact (or even have seen) the git prompt, and the command thus "hangs" forever.

This PR moves the code to clone the template repo outside of the `listr` prompt, so the user can interact with the underlying git prompt for credentials. If, however, the terminal is is not interactive (eg if running in a CI) it passes a config option so that if credentials are not set up the command will fail rather than hang waiting for credentials.

This PR also fixes a mixed usage of both callback and promise-based API for `clone`. With the current mixed approach usage an error immediately crashes the CLI.

Screen Shot 2022-07-05 at 11 03 24 AM

Using strictly the promise-based API allows the error to be caught and handled within the CLI.

Screen Shot 2022-07-05 at 11 03 44 AM

### How to test your changes?

**Setup**
1. Create a private template repo

**Steps**
1. Create an app using your private template repo: `yarn create-app --template=https://github.com/[USERNAME]/[PRIVATE_TEMPLATE_REPO]`
1. Observe it succeeding (using your already configured credentials)
1. Copy your git credentials configuration to a temporary file: `cp ~/.config/dev/gitconfig ~/.config/dev/gitconfigGOOD
1. Remove your configured git credentials: `echo "[credential]\n\thelper = " > ~/.config/dev/gitconfig`
1. Create an app using your private template repo: `yarn create-app --template=https://github.com/[USERNAME]/[PRIVATE_TEMPLATE_REPO]`
1. Observe being prompted for the github credentials. Do one with wrong information to see error path. Setup an access token in github to use when prompted for your password to see the successful path.

**Cleanup**
1. Copy back your git credentials configuration from the temporary file: `cp ~/.config/dev/gitconfigGOOD ~/.config/dev/gitconfig`
1. Remove the temporary file: `rm ~/.config/dev/gitconfigGOOD`

### DIFF
--- a/packages/cli-kit/src/output.ts
+++ b/packages/cli-kit/src/output.ts
@@ -23,6 +23,8 @@ import cjs from 'color-json'
 import stripAnsi from 'strip-ansi'
 import {Writable} from 'node:stream'
 
+export {default as logUpdate} from 'log-update'
+
 let logFile: string
 
 export function initiateLogging({
@@ -71,6 +73,8 @@ enum ContentTokenType {
 interface ContentMetadata {
   link?: string
 }
+
+export type Logger = (message: string) => void
 
 class ContentToken {
   type: ContentTokenType
@@ -303,26 +307,42 @@ export const clearCollectedLogs = () => {
 }
 
 /**
- * Ouputs information to the user. This is akin to "console.log"
+ * Ouputs information to the user.
  * Info messages don't get additional formatting.
  * Note: Info messages are sent through the standard output.
  * @param content {string} The content to be output to the user.
+ * @param logFunc {Function} The logging function to use to output to the user.
  */
-export const info = (content: Message) => {
+export const info = (content: Message, logFunc: Logger = consoleLog) => {
+  const message = stringifyMessage(content)
   if (isUnitTest()) collectLog('info', content)
-  message(content, 'info')
+  outputWhereAppropriate('info', logFunc, message)
 }
 
 /**
  * Outputs a success message to the user.
- * Success message receive a special formatting to make them stand out in the console.
+ * Success messages receive a special formatting to make them stand out in the console.
  * Note: Success messages are sent through the standard output.
  * @param content {string} The content to be output to the user.
+ * @param logFunc {Function} The logging function to use to output to the user.
  */
-export const success = (content: Message) => {
+export const success = (content: Message, logFunc: Logger = consoleLog) => {
   const message = colors.bold(`✅ Success! ${stringifyMessage(content)}.`)
   if (isUnitTest()) collectLog('success', content)
-  outputWhereAppropriate('info', consoleLog, message)
+  outputWhereAppropriate('info', logFunc, message)
+}
+
+/**
+ * Outputs a started message to the user.
+ * Started messages receive a special formatting to make them stand out in the console.
+ * Note: Started messages are sent through the standard output.
+ * @param content {string} The content to be output to the user.
+ * @param logFunc {Function} The logging function to use to output to the user.
+ */
+export const started = (content: Message, logFunc: Logger = consoleLog) => {
+  const message = `${colors.yellow('❯')} ${stringifyMessage(content)}`
+  if (isUnitTest()) collectLog('started', content)
+  outputWhereAppropriate('info', logFunc, message)
 }
 
 /**
@@ -330,11 +350,25 @@ export const success = (content: Message) => {
  * Completed message receive a special formatting to make them stand out in the console.
  * Note: Completed messages are sent through the standard output.
  * @param content {string} The content to be output to the user.
+ * @param logFunc {Function} The logging function to use to output to the user.
  */
-export const completed = (content: Message) => {
+export const completed = (content: Message, logFunc: Logger = consoleLog) => {
   const message = `${colors.green('✔')} ${stringifyMessage(content)}`
   if (isUnitTest()) collectLog('completed', content)
-  outputWhereAppropriate('info', consoleLog, message)
+  outputWhereAppropriate('info', logFunc, message)
+}
+
+/**
+ * Outputs a failed message to the user.
+ * Failed messages receive a special formatting to make them stand out in the console.
+ * Note: Failed messages are sent through the standard output.
+ * @param content {string} The content to be output to the user.
+ * @param logFunc {Function} The logging function to use to output to the user.
+ */
+export const failed = (content: Message, logFunc: Logger = consoleLog) => {
+  const message = `${colors.red('✖')} ${stringifyMessage(content)}`
+  if (isUnitTest()) collectLog('failed', content)
+  outputWhereAppropriate('info', logFunc, message)
 }
 
 /**
@@ -342,10 +376,12 @@ export const completed = (content: Message) => {
  * Debug messages don't get additional formatting.
  * Note: Debug messages are sent through the standard output.
  * @param content {string} The content to be output to the user.
+ * @param logFunc {Function} The logging function to use to output to the user.
  */
-export const debug = (content: Message) => {
+export const debug = (content: Message, logFunc: Logger = consoleLog) => {
   if (isUnitTest()) collectLog('debug', content)
-  message(colors.gray(stringifyMessage(content)), 'debug')
+  const message = colors.gray(stringifyMessage(content))
+  outputWhereAppropriate('debug', logFunc, message)
 }
 
 /**
@@ -353,10 +389,12 @@ export const debug = (content: Message) => {
  * Warning messages receive a special formatting to make them stand out in the console.
  * Note: Warning messages are sent through the standard output.
  * @param content {string} The content to be output to the user.
+ * @param logFunc {Function} The logging function to use to output to the user.
  */
-export const warn = (content: Message) => {
+export const warn = (content: Message, logFunc: Logger = consoleWarn) => {
   if (isUnitTest()) collectLog('warn', content)
-  consoleWarn(colors.yellow(stringifyMessage(content)))
+  const message = colors.yellow(stringifyMessage(content))
+  outputWhereAppropriate('warn', logFunc, message)
 }
 
 /**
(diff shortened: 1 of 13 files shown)


packages/create-app/src/services/init.ts
```diff
@@ -47,23 +47,16 @@ async function init(options: InitOptions) {
...
     let tasks: ConstructorParameters<typeof ui.Listr>[0] = []
 
+    output.info(output.content`❯ Downloading template from ${repoUrl}`)
+    await git.downloadRepository({
+      repoUrl,
+      destination: templateDownloadDir,
+      shallow: true,
+    })
+    process.stdout.moveCursor(0, -1)
+    process.stdout.clearLine(1)
```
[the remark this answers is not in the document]
craigmichaelmartin: I ended up removing these. My thought in including them was to "clean up" the downloading template action after it finished, which I perceived to be more consistent with the CLIs UX (eg, how listr appears to collapse sub-step actions like this). But I am new to the CLI and unsure if this is in fact more consistent, or if it's better to leave the log of cloning the repo. Open to any feedback!

pepicrft (approved):
@craigmichaelmartin I'd also update the other modules from where the `downloadRepository` is being called. In particular, `packages/app/src/cli/services/scaffold/extension.ts` and `packages/create-hydrogen/src/services/init.ts`.

packages/create-app/src/services/init.ts
```diff
@@ -47,23 +47,14 @@ async function init(options: InitOptions) {
     await file.mkdir(templateDownloadDir)
     let tasks: ConstructorParameters<typeof ui.Listr>[0] = []
 
+    output.info(output.content`❯ Downloading template from ${repoUrl}`)
```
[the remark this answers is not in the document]
amcaplan: The progressUpdater is pretty valuable to the UX, it's unfortunate to see it go. Is there any way to have a self-replacing output similar to what Listr provides, so we can display progress?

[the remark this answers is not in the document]
amcaplan: Actually, log-update would allow us to start with `❯ Downloading...` and then replace it with `✓ Downloaded...` at the end. And that would look nice with Listr too!

craigmichaelmartin:
Hey @pepicrft and @amcaplan! Thanks for the reviews! Would you mind taking another look, specifically at how I integrated `log-update`?

The new code is in [this commit](https://github.com/Shopify/cli/pull/114/commits/a23893a8191e0345f1ecaf268ab10c5f5d18d89d) if that makes it easier to review.

My approach was to configure the low level `output` logging utilities (`info`, `success`, `completed`, `warn`, `debug`) to accept a logger to write with. It defaults to their current one, but now allows them to accept a `logUpdate` logger instead. This means any of these logging utilities now support writing/rewriting messages (if a logUpdate logger is provided), while keeping the current logging flow (eg of checking log level and writing all logging regardless to logfile). I added two new "primitives" to the these logging utilities (`started` and `failed`, to go allowing with the existing `completed` utility).

These logging utilities are used by the newly added `simpleTask` in `ui` to write/rewrite the start and results of a task (while leaving stdout untouched while the task runs, which was the motivation behind this PR).

This `simpleTask` is used in the original location, and the two others I neglected that Pedro pointed out.

Here are the results.

When cloning the repository is started (same icon/color as listr in progress):
Screen Shot 2022-07-06 at 4 48 15 PM

When cloning the repository is complete (same icon/color as listr complete):
Screen Shot 2022-07-06 at 4 48 52 PM

If cloning the repository were to have failed instead (same icon/color as listr failure):
Screen Shot 2022-07-06 at 4 50 33 PM

But, now here is the current limitation. In the event the stdout _is_ used, logUpdate is unaware of this (it relies on [counting lines of output](https://github.com/sindresorhus/log-update/blob/main/index.js#L59) to know how much to erase, and so in both failure and success cases when stdout _is_ used by the task, the re-write isn't perfect: logUpdate rewrites one line, but that line is now the line asking for the password instead of the original title:

Screen Shot 2022-07-06 at 4 53 12 PM

Screen Shot 2022-07-06 at 4 54 49 PM

This is certainly an improvement over the current state without this PR (the CLI hanging because listr is continuously rewriting stdout), and even with this limitation I think certainly an improvement over not using logUpdate: with the "happy" paths working as expected, and the limitation being a close fallback to non-interactive. If, however, someone has an idea on how to better accommodate (or resolve) the limitation, I am happy to make the changes!

packages/cli-kit/src/ui.ts
```diff
@@ -48,6 +48,31 @@ interface BaseQuestion<TName extends string> {
...
+  title: string
+  success?: string
+  task: () => Promise<void | string>
+}
+
+/**
+ * Performs a task with the title kept up to date and stdout available to the
+ * task while it runs (there is no re-writing stdout while the task runs).
+ */
+export const simpleTask = async ({title, success: _success, task}: SimpleTaskOptions) => {
```
pepicrft, line 61: Nitpick: What about `task` _(assuming that by simple you refer to one task)_

packages/cli-kit/src/ui.ts
```diff
@@ -48,6 +48,31 @@ interface BaseQuestion<TName extends string> {
...
+  let success
+  started(title, logUpdate)
```
pepicrft, line 63: Is this the only place from where we are using `started`? In that case I'd scope the formatting logic to this `ui` module, either in this same function or in a different function. We can extract it in the near future if the reusing need emerges.

packages/cli-kit/src/ui.ts
```diff
@@ -48,6 +48,31 @@ interface BaseQuestion<TName extends string> {
   result?: (value: string) => string | boolean
 }
 
+export interface SimpleTaskOptions {
+  title: string
+  success?: string
+  task: () => Promise<void | string>
```
pepicrft, line 54: It's hard to tell what `string` means without reading at the docs or the implementation. What about using a custom type:

```ts
Promise<void | { successMessage: string }>
```

pepicrft: Also, being able to pass a success message through the return value of the `task` function and the `success?` attribute seems redundant. Wouldn't it be sufficient with the return value of the task?

packages/cli-kit/src/ui.test.ts
```diff
@@ -0,0 +1,53 @@
...
+    const title = 'Starting the task'
+    const task = vi.fn(() => Promise.resolve('Finished the task (created X and Y)'))
+    const outputMock = mockAndCaptureOutput()
+    outputMock.clear()
+
+    // When
+    await simpleTask({title, task})
+
+    // Then
+    expect(outputMock.started()).toMatch('Starting the task')
```
pepicrft, line 33: I'd recommend using `toMatchInlineSnapshot` to make it easier to re-record the value we check against in the future if the content/format changes.

pepicrft: This comment applies to all the string `toMatch` in this test module.

packages/cli-kit/src/output.ts
```diff
@@ -307,60 +311,94 @@ export const clearCollectedLogs = () => {
...
+}
+
+/**
+ * Outputs a failed message to the user.
+ * Failed messages receive a special formatting to make them stand out in the console.
+ * Note: Failed messages are sent through the standard output.
+ * @param content {string} The content to be output to the user.
+ * @param logFunc {Function} The logging function to use to output to the user.
+ */
+export const failed = (content: Message, logFunc: Logger = consoleLog) => {
```
pepicrft, line 372: Nitpick: I'd rename `logFunc` to `logger` in all the functions to align with the interface name `Logger`.

packages/cli-kit/src/ui.ts
```diff
@@ -48,6 +48,31 @@ interface BaseQuestion<TName extends string> {
...
+  started(title, logUpdate)
+  try {
+    const result = await task()
+    success = _success || result || title
+  } catch (err) {
+    failed(title, logUpdate)
+    logUpdate.done()
+    throw err
+  }
+  completed(success, logUpdate)
```
pepicrft, line 72: `logUpdate` has a `logUpdateStderr` interface to log through `stderr` instead of `stdout`. Should we use that one here?

packages/create-hydrogen/src/services/init.ts
```diff
@@ -62,29 +62,23 @@ async function init(options: InitOptions) {
...
+      task: async () => {
+        await git.downloadRepository({
+          repoUrl,
+          destination: templateDownloadDir,
+          shallow: true,
+        })
+        if (!(await file.exists(path.join(templatePath, 'package.json')))) {
+          throw new error.Abort(`The template ${templatePath} was not found.`, suggestHydrogenSupport())
+        }
       },
```
pepicrft, line 78: If you adjust the API as suggested in another comment, you can do here:
```js
return { successMessage: `Downloaded template from ${repoUrl}` }
```

pepicrft (approved):
@craigmichaelmartin great iteration 👏🏼. I left some comments mostly around developer ergonomics using the new APIs.

packages/cli-kit/src/ui.ts
craigmichaelmartin: We want to re-write the task's original `title` line indicating that the task succeeded or failed. The original title line is written by a logUpdate using stdout, so we want to use that same created logUpdate to re-write it. If we use a new logUpdate created from the logUpdateStderr, it would not rewrite the title line, since this state is [kept in a closure for each logUpdate](https://github.com/sindresorhus/log-update/blob/main/index.js#L36-L38). Throwing the error allows the CLI to write the details of the error to stderr after re-writing the title to indicate the failure.

pepicrft:
I'd wait for @amcaplan's final review before merging. Thanks a lot for addressing the comments @craigmichaelmartin.

pepicrft:
> There is a lot of really nice code here, though I'm not 100% sure what's the end result here. We have lost the git progress tracker, which I think is unfortunate though not the end of the world, given that clones are much quicker these days with a fetch-depth of 1. But once we've dropped listr, are we going to all this trouble with log-update just to rewrite 1 line of output after the git prompt? And to have that not work in all cases?

@amcaplan what makes you think maintaining this would be a trouble for us? Between having this scenario working, or having the CLI hanging I lean more towards going ahead with the changes but I might be missing some maintenance costs based on your comment.

PR #125, opened 2022-07-06
Only run acceptance tests on earliest and latest nodes

amcaplan:

No need to spend CPU cycles on running acceptance tests in all cases. It's unlikely anything works on 14, got broken on 16, and got fixed by 18.

Drops acceptance test runs on Node 16 and 17.

### How to test your changes?

CI


amcaplan:
What if we drop Node 17 instead, which is [already EOLed](https://github.com/nodejs/Release#end-of-life-releases)?

pepicrft (approved):

PR #127, opened 2022-07-07
Split app.ts in different files

isaacroldan:
- App.ts grew a lot recently

Extract some parts for app.ts to their own files
- `extensions.ts` with all the extensions schemas
- `identifiers.ts` with all the identifiers schemas and the functions to load them
- `app-loader.ts` with the AppLoader class 

This is a first iteration separating code with different purposes, a future refactor will come with more improvements


pepicrft (approved):
Tophatted and works as expected 🚀. Some follow-up work that we could do later on is using classes and exposing the utility functions as functions of the class. I think that'll improve the developer ergonomics working with those APIs

PR #129, opened 2022-07-07
Don't auto-report unhandled errors and fix stack traces on Bugsnag

pepicrft:
There are errors like [this one](https://app.bugsnag.com/shopify/cli/errors/62630ceebd38f1000994a723?filters[release.seen_in]=3.0.27) that make their way straight to the Node runtime and the Bugsnag instance captures those and reports them to Busnag as bugs. Most of them are not bugs but errors in users' projects so they shouldn't be reported as bugs. As a consequence, there's a lot of noise on Bugsnag that makes it hard to triage real bugs.

Moreover, we are [still receiving](https://app.bugsnag.com/shopify/cli/errors/62c6c91e7a9b040008c8176c?filters[release.seen_in]=3.0.27) stacktraces whose path is not normalized preventing Bugsnag from showing up the code on Bugsnag's UI

- [x] I changed Bugsnag configuration to not auto-report unhandled errors. Unfortunately we can't group them in bugs and not bugs so eliminating the noise comes at the cost of some real bugs not showing up on Bugsnag. Users most likely report them through GitHub issues.
- [x] Updated the error reporting logic to use Stacktracey to:
  - Clean the stacktraces removing code that's internal to other packages and duplicated lines.
  - Normalize the paths using `path.normalize` instead of regular expressions.

Here's an example of [an error reported after the changes](https://app.bugsnag.com/shopify/cli/errors/62c6e4bc19047d0008fd13da?filters[release.seen_in]=3.0.27&filters[search]=Pedro5). Note that the directory structure varies between development and production (i.e. all packages under `node_modules`), we might see a different behaviour. I'll see if I can replicate the production setup without having to release.

### How to test your changes?
1. Add a `throw new Abort("xx")` in one of the commands, for example `app info`.
2. Update `error.ts` to always initialize Bugsnag and report errors.

### DIFF
--- a/.changeset/many-meals-fly.md
+++ b/.changeset/many-meals-fly.md
@@ -1,0 +1,5 @@
+---
+'@shopify/cli-kit': minor
+---
+
+Massage error stacktraces to be properly formatted on Bugsnag
--- a/.changeset/quick-pandas-boil.md
+++ b/.changeset/quick-pandas-boil.md
@@ -1,0 +1,5 @@
+---
+'@shopify/cli-kit': minor
+---
+
+Not report unhandled errors that go straight to the Node runtime


isaacroldan (approved):

packages/cli-kit/src/node/cli.ts
```diff
@@ -34,10 +35,12 @@ export async function runCLI(options: RunCLIOptions) {
     settings.debug = true
   } else {
     Bugsnag.start({
+      appType: 'node',
       apiKey: bugsnagApiKey,
       logger: null,
       appVersion: await constants.versions.cliKit(),
       autoTrackSessions: false,
+      autoDetectErrors: false,
```
gonzaloriestra, line 43: Sorry because I'm probably missing some context, but why do we need this? Where are the uncaught errors coming from? The user code is only run in `dev`, shouldn't it be enough to catch them there?

pepicrft: There are errors that go straight to the Node runtime skipping any `try/catch` in our code. [This](https://app.bugsnag.com/shopify/cli/errors/62630ceebd38f1000994a723?filters[release.seen_in]=3.0.27) is an example of one of those errors coming from `esbuild`, which is used internally by `Vite`. I think the reason being is that those tools use [Node thread workers](https://github.com/evanw/esbuild/blob/79a35122062c8e99bf9ab62f03fbafa641139e15/lib/npm/node.ts#L21), which run in their own isolated thread, so Bugsnag is subscribed to those errors and is sending them automatically.

packages/cli-kit/src/node/cli.ts
```diff
@@ -86,12 +89,14 @@ const reportError = async (errorToReport: Error): Promise<Error> => {
...
-    if (errorToReport.stack) {
-      // For mac/linux, remove `file:///` from stacktrace
-      // For windows, remove `file:///C:/` from stacktrace
-      const regex = '\\((.*node_modules.)(@shopify.)?'
-      mappedError.stack = errorToReport.stack.replace(new RegExp(regex, 'g'), '(')
-    }
+    const mappedStacktrace = new StackTracey(errorToReport)
+      .clean()
+      .items.map((item) => {
+        const filePath = normalize(item.file).replace('file:/', '/').replace('C:/', '')
```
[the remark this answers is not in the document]
pepicrft: I replaced part of the regex expression with the [normalize](https://nodejs.org/api/path.html#pathnormalizepath) function because that's intended to be used for fixing these normalizations issue. Then I debated between doing the replacements with regex or string replacement and I found the replaces would be easier to follow when reading the code. 
I'm in contact with Bugsnag because they are only able to relativize the paths of macOS stacktraces but not Windows'. I think tried with and without the `C:/` prefix but it doesn't work.

packages/cli-kit/src/node/cli.ts
gonzaloriestra: Ahh, thanks for the explanation! 👍

PR #131, opened 2022-07-07
Cache node_modules pre-install in CI for faster builds

amcaplan:

Make CI faster! Cache allthethings!

Caches node_modules, as it's faster to reconstitute from a cache than to install via yarn, which involves lengthy resolution, linking, and build steps. We then run `yarn install anyway` because:

1. Maybe it's the first time
2. `yarn install` should be just a second or two when it's not the first time, as it'll find that everything is already there and built!

### How to test your changes?

CI

### DIFF
--- a/.github/workflows/shopify-cli.yml
+++ b/.github/workflows/shopify-cli.yml
@@ -91,7 +91,7 @@ jobs:
           path: |
             ./**/*
             ./**/.*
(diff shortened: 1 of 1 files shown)


pepicrft (approved):

amcaplan:
Apparently the cache is so slow on Windows and Mac as to make caching node_modules worse than just running the yarn install in the first place. Go figure.

PR #132, opened 2022-07-08
Exit without error message on deliberate user actions

### LINKED ISSUE #11: Create an `AbortGraceful` function to exit more cleanly on deliberate user actions
[Reference](https://github.com/Shopify/shopify-cli-next/issues/383)

[Screenshot to come later, having yarn issues]

Currently there are a couple cases in, for instance, the `create-hydrogen` flow where users can deliberately decline to continue the process, but we exit as if it is an error.

Example: if you try to create a project that would require overwriting existing files, the CLI prompts you to ask whether you want to overwrite or not. If you choose "no", the [console throws an error with `AbortSilent`](https://github.com/Shopify/shopify-cli-next/blob/78671cb9b48f829aa4ce9d2395011932f26692e2/packages/cli-kit/src/ui.ts#L57), exiting with code 1.

It's odd for a clearly deliberate user action like this to result in an error. I'd prefer to create an alternate function like `AbortGracefully` (just a working name) to use in cases where we're exiting in response to user choice, not an error.

Instead of throwing an error, it should simply output a brief message that the process is exiting, and return the user safely to their terminal prompt.

alvaro-shopify:

Fixes #11

- Reusing existing ```CancelExecution``` exception to be used by developers when the app needs to be closed in the middle of its execution due to any kind of deliberate user action.
- The exception is now handled inside the global exception handler piece of code
- In case a message is added to the exception, this will be displayed to the user. Otherwise a fixed message will be showed

### How to test your changes?
- ```dev create-hydrogen``` using a name that is already used previously. Answer ```no``` when you are prompted to overwrite

### DIFF
--- a/packages/cli-kit/src/node/base-command.ts
+++ b/packages/cli-kit/src/node/base-command.ts
@@ -1,9 +1,9 @@
-import {globalCommandErrorHandler} from './global-error-handler.js'
+import {errorHandler} from './error-handler.js'
 import {Command} from '@oclif/core'
 
 // eslint-disable-next-line import/no-anonymous-default-export
 export default abstract class extends Command {
   async catch(error: Error & {exitCode?: number | undefined}) {
-    globalCommandErrorHandler(error)
+    errorHandler(error)
   }
 }
--- a/packages/cli-kit/src/node/cli.ts
+++ b/packages/cli-kit/src/node/cli.ts
@@ -1,6 +1,6 @@
 // CLI
 import {findUpAndReadPackageJson} from './node-package-manager.js'
-import {globalCommandErrorHandler} from './global-error-handler.js'
+import {errorHandler} from './error-handler.js'
 import {initializeCliKitStore} from '../store.js'
 import {initiateLogging} from '../output.js'
 import {isDebug} from '../environment/local.js'
@@ -37,7 +37,7 @@ export async function runCLI(options: RunCLIOptions) {
     })
   }
 
-  run(undefined, options.moduleURL).then(flush).catch(globalCommandErrorHandler)
+  run(undefined, options.moduleURL).then(flush).catch(errorHandler)
 }
 
 /**
--- a/packages/create-app/src/commands/init.test.ts
+++ b/packages/create-app/src/commands/init.test.ts
@@ -2,12 +2,12 @@ import Init, {InvalidGithubRepository, UnsupportedTemplateAlias} from './init.js
 import initPrompt from '../prompts/init.js'
 import initService from '../services/init.js'
 import {describe, it, expect, vi, beforeEach} from 'vitest'
-import {globalCommandErrorHandler} from '@shopify/cli-kit/node/global-error-handler'
+import {errorHandler} from '@shopify/cli-kit/node/error-handler'
 
 beforeEach(() => {
   vi.mock('../prompts/init')
   vi.mock('../services/init')
-  vi.mock('@shopify/cli-kit/node/global-error-handler')
+  vi.mock('@shopify/cli-kit/node/error-handler')
 
   vi.mocked(initPrompt).mockResolvedValue({name: 'name', template: 'http://test.es'})
 })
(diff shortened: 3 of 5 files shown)


pepicrft (approved):

packages/cli-kit/src/node/cli.ts
```diff
@@ -46,7 +47,18 @@ export async function runCLI(options: RunCLIOptions) {
...
+    .catch(async (error: Error): Promise<void | Error> => {
+      if (error instanceof CancelExecution) {
+        info(
+          `✨  ${
+            error.message && error.message !== ''
+              ? error.message
+              : `Executed using CLI ${content`${token.yellow(await constants.versions.cliKit())}`.value}`
+          }`,
+        )
+        process.exit(0)
```
gonzaloriestra, line 59: Returning here instead of letting the command to finish as we did before will skip the analytics reporting when the command is canceled. But I guess it's not a big deal.

packages/cli-kit/src/node/cli.ts
```diff
@@ -46,7 +47,18 @@ export async function runCLI(options: RunCLIOptions) {
 
   run(undefined, options.moduleURL)
     .then(flush)
-    .catch((error: Error): Promise<void | Error> => {
+    .catch(async (error: Error): Promise<void | Error> => {
+      if (error instanceof CancelExecution) {
+        info(
+          `✨  ${
```
gonzaloriestra, line 53: I don't think we need to print anything else here. Also, we are already showing the `Done in X seconds` message, so that's enough IMO:

image

packages/cli-kit/src/node/cli.ts
alvaro-shopify: Maybe we could let finishing the execution instead of exit with success code.

packages/cli-kit/src/node/cli.ts
alvaro-shopify: I think you are right. We could leave the option to print the exception message only in case the developer add it to it when it is thrown

gonzaloriestra: Nope:

11-10-ufvyp-vczsf

gonzaloriestra: Ah, right... I thought it was oclif. Anyway, I think there's no need to show anything else in this case.

packages/create-app/src/commands/init.test.ts
```diff
@@ -39,22 +41,19 @@ describe('create app command', () => {
...
 
-    // Then
-    await expect(result).rejects.toThrow(
-      'Only \u001b[33mnode\u001b[39m, \u001b[33mphp\u001b[39m, \u001b[33mruby\u001b[39m template aliases are supported',
-    )
+    expect(globalCommandErrorHandler).toHaveBeenCalledWith(UnsupportedTemplateAlias())
   })
 
-  it('throw an error when using a non github url repo', async () => {
+  it.skip('throw an error when using a non github url repo', async () => {
```
pepicrft, line 52: Should we revert this `.skip`?

packages/cli-kit/src/node/base-command.ts
```diff
@@ -0,0 +1,9 @@
+import {globalCommandErrorHandler} from './global-error-handler.js'
+import {Command} from '@oclif/core'
+
+// eslint-disable-next-line import/no-anonymous-default-export
+export default abstract class extends Command {
+  async catch(error: Error & {exitCode?: number | undefined}) {
```
pepicrft, line 6: TIL about the `catch` function in commands.

packages/cli-kit/src/node/cli.ts
```diff
@@ -1,18 +1,11 @@
 // CLI
 import {findUpAndReadPackageJson} from './node-package-manager.js'
+import {globalCommandErrorHandler} from './global-error-handler.js'
```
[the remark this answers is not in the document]
alvaro-shopify: All the exceptions are handled there so you are right it is the only error handler. Renaming in progress

PR #138, opened 2022-07-10
Faster acceptance tests in CI

amcaplan:

Faster CI

We always build before running feature tests, which makes sense locally. But in CI, there's no need to build again, as we just built in an earlier step!

### How to test your changes?

CI

### DIFF
--- a/.github/workflows/shopify-cli.yml
+++ b/.github/workflows/shopify-cli.yml
@@ -96,7 +96,7 @@ jobs:
       - name: Unit tests
         run: npx nx ${{ github.event.inputs.nxLevel || 'affected' }} --target=test --exclude=features --skip-nx-cache
       - name: Acceptance tests
-        run: npx nx run features:test --skip-nx-cache
+        run: yarn run --cwd=packages/features test
   pr:
     name: Node ${{ matrix.node }} in ${{ matrix.os }}
     runs-on: ${{ matrix.os }}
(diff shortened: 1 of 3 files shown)


amcaplan:
This seems to speed up the ubuntu builds (where we run acceptance tests) by 1-2 minutes, as we were building TWICE (one rebuild thanks to nx, one rebuild because [we rebuild explicitly](https://github.com/Shopify/cli/blob/58c170ff9dada2910c3b1e467bbdcf4c19dbc098/packages/features/steps/environment.steps.ts#L51) before running tests!) after we already built the first time.

amcaplan:
I discovered we can run cucumber tests in parallel, which makes sense in our case as they're mostly IO-bound.

I also had the ubuntu tests go first and the [more expensive](https://docs.github.com/en/billing/managing-billing-for-github-actions/about-billing-for-github-actions#minute-multipliers) Windows/MacOS runs only run after when we're 99% sure they'll pass.

packages/features/cucumber.js
```diff
@@ -10,6 +10,7 @@ const common = [
   '--require steps/**/*.ts',
   '--require lib/**/*.ts',
   '--format @cucumber/pretty-formatter',
+  '--parallel 3',
```
isaacroldan, line 13: We only have 2 acceptance tests (and 1 is trivial), but it might be useful for the future :)

.github/workflows/shopify-cli.yml
```diff
@@ -97,16 +97,17 @@ jobs:
         run: npx nx ${{ github.event.inputs.nxLevel || 'affected' }} --target=test --exclude=features --skip-nx-cache
       - name: Acceptance tests
         run: npx nx run features:test --skip-nx-cache
-  pr:
+        env:
+          SKIP_BUILD_BEFORE_ACCEPTANCE_TESTS: 1
```
isaacroldan, line 101: 👏

PR #145, opened 2022-07-12
Skip acceptance tests if no meaningful pr changes

amcaplan:

Save money in CI!

In the rare case of PRs without code changes, we run an hour of CI (actually several times that when you take into account Windows x 2 and Mac x 10).

We add a quick check before running acceptance tests. If nx says there were changes, we'll run as usual. If nx says no changes, we move on to the next step.

[Before](https://github.com/Shopify/cli/actions/runs/2655166972/usage):

Screen Shot 2022-07-12 at 16 28 54

[After](https://github.com/Shopify/cli/actions/runs/2656405258/usage):

Screen Shot 2022-07-12 at 16 29 04

Note that we could also use a more complex strategy of skipping all CI on PRs without substantive changes. However, that would require adding an extra pre-step which takes a runner, loads the app, installs dependencies, and runs nx to check if anything is affected. This will make CI slower and cost computation time. Since 90%+ of PRs are actual code changes, I don't think it's worth it.

### How to test your changes?

CI

### DIFF
--- a/.github/workflows/shopify-cli.yml
+++ b/.github/workflows/shopify-cli.yml
@@ -98,7 +98,7 @@ jobs:
       - name: Acceptance tests
         # If nx says there are no changes, consider the step completed. Otherwise run acceptance tests.
         run: |
-          ruby -e "p \`npx -s nx print-affected --target=test --select=projects\`.strip.empty? || exit(1)" ||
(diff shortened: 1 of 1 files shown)


isaacroldan (approved):
tested the updated line locally and seems to be working as expected!

PR #152, opened 2022-07-13
Report MODULE_NOT_FOUND errors coming from @oclif

pepicrft:
@oclif eager-loads all plugins and swallows module loading errors to present a more user-friendly error, `command not found`. This is not great because some module loading errors might be actual bugs in our code, like it was the case with `@shopify/cli-kit` introducing breaking changes in a minor version, so we need to report those too.

Since `@oclif/core` [sends those errors](https://github.com/oclif/core/blob/main/src/config/config.ts#L244) through the `process.emitWarning` API, we can subscribe to those events and make sure we report the ones that might indicate that there's a bug in the CLI. Note that it'll take some iterations until we find the right level of granularity in the error handling logic. For instance, once we have the notion of 3P plugins, we need to filter out module-loading errors that don't come from Shopify's plugins or CLI.

### How to test your changes?

Add an invalid import to the info command and invoke it:

```ts
import { invalid } from "non-existing-module"

invalid()
```

### DIFF
--- a/.changeset/short-pumpkins-rest.md
+++ b/.changeset/short-pumpkins-rest.md
@@ -1,0 +1,5 @@
+---
+'@shopify/cli-kit': patch
+---
+
+Report errors coming from @oclif
--- a/packages/cli-kit/src/node/cli.ts
+++ b/packages/cli-kit/src/node/cli.ts
@@ -1,6 +1,6 @@
 // CLI
 import {findUpAndReadPackageJson} from './node-package-manager.js'
-import {errorHandler} from './error-handler.js'
+import {errorHandler, subscribeToProcessEmittedErrors} from './error-handler.js'
 import {initializeCliKitStore} from '../store.js'
 import {initiateLogging} from '../output.js'
 import {isDebug} from '../environment/local.js'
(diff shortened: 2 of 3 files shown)


packages/cli-kit/src/node/error-handler.test.ts
```diff
@@ -44,3 +46,58 @@ describe('errorHandler', () => {
...
+  it('error output uses same input error instance when the error type is bug', async () => {
+    // Given
+    const bugError = new Bug('error message', 'try message')
+    const outputMock = outputMocker.mockAndCaptureOutput()
+
+    // When
+    await outputError(bugError)
+
+    // Then
+    expect(unstyled(outputMock.error())).not.toBe('')
```
pepicrft, line 87: I tried few approaches here and ultimately landed on this but I'm not quite happy. I initially used `toMatchInlineSnapshot` but because it includes the stacktraces it makes the test dependent on the environment. Then I tried to remove the environment-dependent lines from the string to match the rest of the content but due to the indentations and the spaces it failed to match too. I'm open to suggestions.

Issue #157, opened 2022-07-14
[Bug]: Extension preview not working with Node < 17


gonzaloriestra:
### Please confirm that you have:

- [X] Searched [existing issues](.) to see if your issue is a duplicate. (If you’ve found a duplicate issue, feel free to add additional information in a comment on it.)
- [X] Reproduced the issue in [the latest CLI version](https://www.npmjs.com/package/@shopify/cli).

### In which of these areas are you experiencing a problem?

Extension

### Expected behavior

`shopify app dev` should allow previewing both apps and extensions with any compatible Node version (14-18)

### Actual behavior

The preview for extensions doesn't work for Node < 17

### Reproduction steps

1. Create an app and a Checkout UI extension
2. `yarn dev`
3. Open the dev-console link
4. No extensions found

### Operating System

Mac OS Monterey

### Shopify CLI version (check your project's `package.json` if you're not sure)

3.2.0

### Shell

iTerm2

### Node version (run `node -v` if you're not sure)

16.16.0

PR #160, opened 2022-07-15
Fix websockets not working on node <17

isaacroldan:
We detected that our reverse proxy was not working with websockets if the node version was 16 or lower.

Our reverse-proxy solution was based on fastify http proxy, after long debugging sessions I couldn't find the issue. Websocket requests were not properly redirected.

Because of this issue, dev-console or extensions live-reload weren't working properly.

Fixes https://github.com/Shopify/cli/issues/157 

- Replace `fastify/http-proxy` with [node-http-proxy](https://github.com/http-party/node-http-proxy)

### How to test your changes?
Update `.nvmrc` and `dev.yml` to use node version 16.15.1
`dev up`
`dev shopify app dev --path fixtures/app`
Open the dev-console and it should load all extensions.


yarn.lock
```diff
@@ -5046,32 +5054,32 @@ fast-safe-stringify@^1.0.8, fast-safe-stringify@^1.2.3:
...
 fastify-plugin@^3.0.0:
   version "3.0.1"
   resolved "https://registry.yarnpkg.com/fastify-plugin/-/fastify-plugin-3.0.1.tgz#79e84c29f401020f38b524f59f2402103fd21ed2"
   integrity sha512-qKcDXmuZadJqdTm6vlCqioEbyewF60b/0LOFCcYN1B6BIZGlYJumWWOYs70SFYLDAH4YqdE1cxH/RKMG7rFxgA==
 
-fastify@^4.0.0-rc.4:
-  version "4.0.0-rc.5"
-  resolved "https://registry.yarnpkg.com/fastify/-/fastify-4.0.0-rc.5.tgz#6e52feb5a54f9a301ffcce763ed1127779a490b6"
-  integrity sha512-Ki2GXFsHU6vLNODqEpAQOF1Hd8ibzMwPXOH+alpnYTBtjmGHwzUODVxRj50BYB03D7CUkuPYymShrjpXAudraA==
+fastify@^4.2.1:
```
shauns, line 5072: wonder why fastify is still here?

packages/app/src/cli/utilities/app/http-reverse-proxy.ts
```diff
@@ -67,14 +53,38 @@ export async function runConcurrentHTTPProcessesAndPathForwardTraffic(
 
   const availablePort = portNumber ?? (await port.getRandomPort())
 
+  const proxy = httpProxy.createProxy()
+  const server = http.createServer(function (req, res) {
+    const target = match(rules, req)
+    if (target) return proxy.web(req, res, {target})
+
+    res.statusCode = 500
```
shauns, line 61: I think we can provide a better experience here in this case -- `debug` statements for verbose mode probably useful.

packages/app/src/cli/utilities/app/http-reverse-proxy.ts
```diff
@@ -67,14 +53,38 @@ export async function runConcurrentHTTPProcessesAndPathForwardTraffic(
...
+    res.end('The request url and path did not match any of the listed rules!')
+  })
+
+  // Capture websocket requests and forward them to the proxy
+  server.on('upgrade', function (req, socket, head) {
```
shauns, line 66: Is there any corresponding event we need to look at to stop the proxy? Or this is enough?

packages/app/src/cli/utilities/app/http-reverse-proxy.ts
```diff
@@ -67,14 +53,38 @@ export async function runConcurrentHTTPProcessesAndPathForwardTraffic(
 
   const availablePort = portNumber ?? (await port.getRandomPort())
```
shauns, line 54: Probably worth a verbose mode output here with the `rules` mapping and the `availablePort`

shauns (changes requested):

yarn.lock
isaacroldan: fastify is still used for the authentication redirect server. (`redirect-listener.ts`)

packages/app/src/cli/utilities/app/http-reverse-proxy.ts
isaacroldan: I can improve this, but it's impossible to reach this point because we are capturing every possible path (anything from `/`)

packages/app/src/cli/utilities/app/http-reverse-proxy.ts
isaacroldan: nop, the `upgrade` is an http request to upgrade from http to ws. From there everything is handled by the websocket server.

PR #174, opened 2022-07-20
Add no tunnel flag to dev

alvaro-shopify:

Fixes https://github.com/Shopify/shopify-cli-planning/issues/319

An Ngrok tunnel is always created when using `dev` command. A new flag is needed to run the command `dev` without a tunnel exposing directly a localhost port.

Listening to localhost is now available after adding that option to web project Content Security Policy in this PR https://github.com/Shopify/web/pull/68866

- Added a new flag `--no-tunnel` to `dev` command. By default is disabled so an ngrok tunnel is automatically created if the flag is not used 
- In case using `--no-tunnel` extensions testing is failing because a validation is made by extensions binary` that checks the using of `localhost` or `:<local_port>``
<p align="center">image</p>
- No reverse proxy is used when `--no-tunnel` is used and there is no extensions

### How to test your changes?
- `yarn shopify app dev --no-tunnel --path <app_path>`


packages/app/src/cli/commands/app/dev.ts
```diff
@@ -60,9 +60,17 @@ export default class Dev extends Command {
     // eslint-disable-next-line @typescript-eslint/naming-convention
     'tunnel-url': Flags.string({
       hidden: false,
-      description: 'Override the ngrok tunnel URL. Format: "https://my-tunnel-url:port"',
+      description:
+        'Override the ngrok tunnel URL. Format: "https://my-tunnel-url:port". Has no effect in case of using --no-tunnel',
```
shauns, line 64: Can you revert the doc change. Let's keep `no-tunnel` quiet for now.

packages/app/src/cli/commands/app/dev.ts
```diff
@@ -60,9 +60,17 @@ export default class Dev extends Command {
...
       env: 'SHOPIFY_FLAG_TUNNEL_URL',
     }),
+    // eslint-disable-next-line @typescript-eslint/naming-convention
+    'no-tunnel': Flags.boolean({
+      hidden: false,
```
shauns, line 69: hidden should be true I think

shauns (changes requested):

pepicrft (commented):
Correct me if I'm wrong, but assuming the following combinatory of flags is right:

| `--no-tunnel` | `--tunnel-url` | Valid | Behaviour |
| --- | --- | --- | -- |
| Passed | Not passed | Yes | Localhost |
| Passed | Passed | Yes | Localhost |
| Not passed | Not passed | Yes | Ngrok |
| Not passed | Passed | Yes | Passed URL |

I think there's an opportunity here to simplify things for the user. Here's a proposal:
```
shopify dev --public-url localhost
shopify dev --public-url tunnel
shopify dev --public-url https://your-own-url....
```

Since the above is a behaviour breaking change because users would have to go from `shopify dev` to `shopify dev --public-url tunnel`, we can keep the current behaviour but include a deprecation message that tells the user about the new workflow.

pepicrft (approved):

PR #178, opened 2022-07-21
Theme improvements (mainly theme init)

amcaplan:

Makes theme init work properly

• Update Ruby CLI to latest
• Fix: Await Ruby subprocess
• Add `--path` flag to init for better development flow
• Add `--verbose` flag to see more output from the Node CLI

### How to test your changes?

Run `shopify theme init` locally, optionally with specifying name (positional arg), `--path`, and `-v/--verbose`

### DIFF
--- a/packages/cli-kit/src/node/ruby.ts
+++ b/packages/cli-kit/src/node/ruby.ts
@@ -8,7 +8,7 @@ import {coerce} from '../semver.js'
 import {AdminSession} from '../session.js'
 import {content, token} from '../output.js'
 // eslint-disable-next-line no-restricted-imports
-import {spawn} from 'child_process'
+import {spawnSync} from 'child_process'
 import {Writable} from 'node:stream'
 
 const RubyCLIVersion = '2.16.0'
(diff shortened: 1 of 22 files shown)


karreiro (commented):
Thank you, @amcaplan!

Should the `path`/`verbose` flags appear here?

image

Thanks again for this PR :)

packages/cli-kit/src/node/ruby.ts
```diff
@@ -33,10 +31,9 @@ export async function execCLI2(args: string[], adminSession?: AdminSession) {
     SHOPIFY_CLI_STORE: adminSession?.storeFqdn,
   }
 
-  spawn('bundle', ['exec', 'shopify'].concat(args), {
+  await system.exec('bundle', ['exec', 'shopify'].concat(args), {
```
pepicrft, line 34: Have you encountered any issues with `system.exec`? I did a similar change but reverted it because @isaacroldan ran into issues when using `system.exec`.

pepicrft (approved):

packages/cli-kit/src/node/ruby.ts
amcaplan: I didn't have any such issues, @pepicrft

packages/theme/package.json
```diff
@@ -30,6 +30,9 @@
     "vitest": "^0.17.1"
   },
   "oclif": {
-    "commands": "dist/cli/commands"
+    "commands": "dist/cli/commands",
+    "hooks": {
+      "prerun": "dist/hooks/prerun"
```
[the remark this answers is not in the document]
amcaplan: That won't work with the logging to file as designed currently. `cli-kit` is agnostic of where it's being run, so it can't specify a logfile.

If we get rid of multiple logfiles (which is under discussion), it'll be possible to push this into `cli-kit` as suggested.

PR #185, opened 2022-07-22
Clean error traces from OCLIF plugins

shauns:

Fixes Shopify/shopify-cli-planning/issues/309

Stack traces for Bugsnag vary depending on the user's folder structure, OS, etc. This adds an [on-error callback to Bugsnag](https://docs.bugsnag.com/platforms/javascript/customizing-error-reports/#updating-events-using-callbacks) to make file paths for OCLIF plug-ins consistent.

### How to test your changes?

Run a command that breaks. I tested by tweaking the info command to have an unexpected null:

image


packages/cli-kit/src/bugsnag.ts
```diff
@@ -0,0 +1,52 @@
...
+import Bugsnag from '@bugsnag/js'
+import {realpath} from 'fs/promises'
+
+/**
+ * If the given file path comes from within a plugin, return the relative path, plus the plugin name.
+ *
+ * This gives us very consistent paths for errors thrown from plugin code.
+ *
+ */
+export function cleanStackFrameFilePath(
```
pepicrft, line 12: Nitpick: What about merging this module into [error-handler.ts](https://github.com/Shopify/cli/blob/main/packages/cli-kit/src/node/error-handler.ts) to consolidate or the error handling logic in the same module?

packages/cli-kit/src/bugsnag.ts
```diff
@@ -0,0 +1,52 @@
...
+  pluginLocations: {name: string; pluginPath: string}[],
+): string {
+  const fullLocation = path.isAbsolute(currentFilePath) ? currentFilePath : path.join(projectRoot, currentFilePath)
+
+  const matchingPluginPath = pluginLocations.filter(({pluginPath}) => fullLocation.indexOf(pluginPath) === 0)[0]
+
+  if (matchingPluginPath !== undefined) {
+    return path.join(matchingPluginPath.name, path.relative(matchingPluginPath.pluginPath, fullLocation))
+  }
+  return currentFilePath
```
pepicrft, line 24: It's hard to tell what path manipulations are happening here. What about adding an example of an input and an output?

packages/cli-kit/src/bugsnag.ts
```diff
@@ -0,0 +1,52 @@
...
+}
+
+/**
+ * Register a Bugsnag error listener to clean up stack traces for errors within plugin code.
+ *
+ */
+export async function registerCleanBugsnagErrorsFromWithinPlugins(plugins: Interfaces.Plugin[]) {
+  // Bugsnag have their own plug-ins that use this private field
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  const bugsnagConfigProjectRoot: string | undefined = (Bugsnag as any)?._client?._config?.projectRoot
```
pepicrft, line 34: This line is a good candidate candidate to break considering we are accessing private properties and we can't rely on Typescript to warn us if Bugsnag has changed the internals of the library. According to the [docs](https://docs.bugsnag.com/platforms/javascript/configuration-options/#projectroot) the default value is `process.cwd()` so we can read that value directly from here.

packages/cli-kit/src/bugsnag.ts
```diff
@@ -0,0 +1,52 @@
...
+  const projectRoot = path.normalize(bugsnagConfigProjectRoot)
+  const pluginLocations = await Promise.all(
+    plugins.map(async (plugin) => {
+      const followSymlinks = await realpath(plugin.root)
+      return {name: plugin.name, pluginPath: path.normalize(followSymlinks)}
+    }),
+  )
+  Bugsnag.addOnError((event) => {
+    event.errors.forEach((error) => {
+      error.stacktrace.forEach((stackFrame) => {
```
pepicrft, line 47: It's neat that Bugsnag has given errors some structured.

packages/cli-kit/src/bugsnag.test.ts
```diff
@@ -0,0 +1,33 @@
...
+    ['some relative path', 'users/own/code.ts', 'users/own/code.ts'],
+    ['some absolute path', '/global/code.ts', '/global/code.ts'],
+    [
+      'plugin in a complex location',
+      'node_modules/.something/@plugin+complex/@plugin/complex-path/foo/bar.ts',
+      '@plugin/complex-path/foo/bar.ts',
+    ],
+  ])('%s', (_, path, result) => {
+    expect(
+      cleanStackFrameFilePath(path, '/my/project', [
```
pepicrft, line 17: Nitpick: It's hard to tell what the arguments represent by looking at the call. What about:
```js
cleanStackFrameFilePath({
  stacktraceFilepath: "...",
  projectRootPath: "...",
  plugins: []
})
```

pepicrft (approved):

Issue #159, opened 2022-07-15
[Bug]: Unmet peer dependency when creating Hydrogen storefront with pnpm


gonzaloriestra:
### Please confirm that you have:

- [X] Searched [existing issues](.) to see if your issue is a duplicate. (If you’ve found a duplicate issue, feel free to add additional information in a comment on it.)
- [X] Reproduced the issue in [the latest CLI version](https://www.npmjs.com/package/@shopify/cli).

### In which of these areas are you experiencing a problem?

Hydrogen custom storefront

### Expected behavior

Hydrogen storefront creation should work with pnpm.

### Actual behavior

It throws an error: `ERR_PNPM_PEER_DEP_ISSUES  Unmet peer dependencies`

### Stack trace

```shell
❯ pnpm create @shopify/create-hydrogen
Library/pnpm/store/v3/tmp/dlx-29404      | +479 ++++++++++++++++++++++++++++++++++++++++++++++++
Packages are hard linked from the content-addressable store to the virtual store.
  Content-addressable store is at: /Users/gonzalo/Library/pnpm/store/v3
  Virtual store is at:             Library/pnpm/store/v3/tmp/dlx-29404/node_modules/.pnpm
Library/pnpm/store/v3/tmp/dlx-29404      | Progress: resolved 478, reused 294, downloaded 184, added 479, done
 WARN  The integrity of 2732 files was checked. This might have caused installation to take longer.
✔ Name your new Hydrogen storefront · july15-hydrogen-pnpm
✔ Choose a template · Demo Store
✔ Template downloaded
✔ App initialized
  ✔ Template files parsed
  ✔ Package.json updated
✔ [Shopifolks-only] Project's NPM registry configured.
✖ Command failed with exit code 1: pnpm install
  Progress: resolved 1, reused 0, downloaded 0, added 0

     ╭─────────────────────────────────────────────────────────────────╮
  │                                                                 │
  │                Update available! 7.1.5 → 7.5.1.                 │
  │   Changelog: https://github.com/pnpm/pnpm/releases/tag/v7.5.1   │
  │                 Run pnpm add -g pnpm to update.                 │
  │                                                                 │
  │     Follow @pnpmjs for updates: https://twitter.com/pnpmjs      │
  │                                                                 │
     ╰─────────────────────────────────────────────────────────────────╯

  Progress: resolved 18, reused 3, downloaded 12, added 0
  Progress: resolved 43, reused 3, downloaded 24, added 0
[...]

PR #186, opened 2022-07-22
Auto install peer dependencies when using pnpm in Hydrogen projects

pepicrft:
The creation of Hydrogen storefronts [fails](https://github.com/Shopify/cli/issues/159) when using pnpm because transitive dependencies have peer dependencies that are not automatically installed by pnpm by default.
 
I'm extending the initialization logic to add `auto-install-peers = true` to the `.npmrc` file in the root of the project to instruct pnpm to install unresolved peer dependencies.

### How to test your changes?
Run `dev create-hydrogen --package-manager pnpm`

### DIFF
--- a/packages/cli-kit/src/file.test.ts
+++ b/packages/cli-kit/src/file.test.ts
@@ -12,9 +12,11 @@ import {
   remove,
   stripUp,
   format,
+  touch,
+  appendFile,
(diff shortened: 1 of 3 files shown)


blittle (approved):
LGTM! Thank you!

PR #187, opened 2022-07-22
Report 5xx responses from the Partners' API as Aborts

pepicrft:
I noticed that we [are reporting](https://app.bugsnag.com/shopify/cli/errors/62ac4e8ed3273c0008d0b78a?filters[event.since]=30d&filters[error.status]=open&filters[release.seen_in]=3.3.2) 5xx responses from the Partners' API as bugs to Bugsnag. That's not necessary because 5xx errors coming from the API are already tracked server-side.

I adjusted the logic to use an `Abort` error when the response status code is `5xx`.

### DIFF
--- a/packages/cli-kit/src/api/admin.ts
+++ b/packages/cli-kit/src/api/admin.ts
@@ -1,8 +1,8 @@
-import {buildHeaders, sanitizedHeadersOutput} from './common.js'
+import {buildHeaders, debugLogRequest, handlingErrors} from './common.js'
 import {AdminSession} from '../session.js'
 import {debug, content, token as outputToken} from '../output.js'
 import {Bug, Abort} from '../error.js'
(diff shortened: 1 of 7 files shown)


packages/cli-kit/src/api/admin.ts
```diff
@@ -20,36 +20,15 @@ const UnknownError = () => {
...
-
-With variables:
-${variables ? JSON.stringify(variables, null, 2) : ''}
-
-And headers:
-${sanitizedHeadersOutput(headers)}
-`)
-  try {
+  const api = 'Admin'
+  return handlingErrors(api, async () => {
```
isaacroldan, line 24: Any reason not to do `handlingErrors('Admin', ....` directly?

isaacroldan (approved):

packages/cli-kit/src/api/admin.ts
pepicrft: It's used further down in line 28

PR #189, opened 2022-07-24
Add flags for theme check

amcaplan:

`theme-check` run via Node doesn't currently support params. We need to allow that.

1. Adds the relevant params, parses and passes through to the Ruby CLI.
2. Adds a `--path` param allowing you to specify where the command should be run. This was non-trivial as we were previously relying on the command's cwd to be in the same place as the Ruby CLI2 executable. We're now using `BUNDLE_GEMFILE` to tell bundler where it can find the Ruby dependencies, then running the command in a specified directory or defaulting to the current working directory.

Change 2 will be very big in improving our development workflow, as we can run `dev shopify theme check --path=/path/to/theme/project in development, or we can write feature tests depending on this behavior, as we've done for `app`.

### How to test your changes?

```sh
$ dev shopify theme init --path=/path/to/some/writable/directory --name=my-theme-for-testing
$ dev shopify theme check --path=/path/to/some/writable/directory/my-theme-for-testing
```

### DIFF
--- a/packages/cli-kit/src/node/ruby.ts
+++ b/packages/cli-kit/src/node/ruby.ts
@@ -29,11 +29,12 @@ export async function execCLI2(args: string[], adminSession?: AdminSession) {
     ...process.env,
     SHOPIFY_CLI_ADMIN_AUTH_TOKEN: adminSession?.token,
     SHOPIFY_CLI_STORE: adminSession?.storeFqdn,
+    BUNDLE_GEMFILE: join(shopifyCLIDirectory(), 'Gemfile'),
   }
 
   await system.exec('bundle', ['exec', 'shopify'].concat(args), {
     stdio: 'inherit',
-    cwd: shopifyCLIDirectory(),
+    cwd: process.cwd(),
     env,
   })
 }
--- a/packages/cli-kit/src/node/ruby.ts
+++ b/packages/cli-kit/src/node/ruby.ts
@@ -15,15 +15,21 @@ const MinBundlerVersion = '2.3.8'
 const MinRubyVersion = '2.3.0'
 const MinRubyGemVersion = '2.5.0'
 
+interface ExecCLI2Options {
+  // Contains token and store to pass to CLI 2.0, which will be set as environment variables
+  adminSession?: AdminSession
+  // Directory in which to execute the command. Otherwise the current directory will be used.
+  directory?: string
+}
 /**
  * Execute CLI 2.0 commands.
  * Installs a version of RubyCLI as a vendor dependency in a hidden folder in the system.
  * User must have a valid ruby+bundler environment to run any command.
  *
  * @param args {string[]} List of argumets to execute. (ex: ['theme', 'pull'])
- * @param adminSession {AdminSession} Contains token and store to pass to CLI 2.0, which will be set as environment variables
+ * @param options {ExecCLI2Options}
  */
-export async function execCLI2(args: string[], adminSession?: AdminSession) {
+export async function execCLI2(args: string[], {adminSession, directory}: ExecCLI2Options = {}) {
   await installCLIDependencies()
   const env = {
     ...process.env,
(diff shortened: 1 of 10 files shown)


packages/cli-kit/src/node/ruby.ts
```diff
@@ -15,25 +15,32 @@ const MinBundlerVersion = '2.3.8'
...
  * Execute CLI 2.0 commands.
  * Installs a version of RubyCLI as a vendor dependency in a hidden folder in the system.
  * User must have a valid ruby+bundler environment to run any command.
  *
  * @param args {string[]} List of argumets to execute. (ex: ['theme', 'pull'])
- * @param adminSession {AdminSession} Contains token and store to pass to CLI 2.0, which will be set as environment variables
+ * @param options {ExecCLI2Options}
  */
-export async function execCLI2(args: string[], adminSession?: AdminSession) {
+export async function execCLI2(args: string[], {adminSession, directory}: ExecCLI2Options = {}) {
```
amcaplan, line 32: Note that we're going from a positional argument to keyword arguments. Thankfully TS has our back here! All the changes should be present below.

packages/cli-kit/src/node/ruby.ts
```diff
@@ -15,25 +15,32 @@ const MinBundlerVersion = '2.3.8'
...
   await installCLIDependencies()
   const env = {
     ...process.env,
     SHOPIFY_CLI_ADMIN_AUTH_TOKEN: adminSession?.token,
     SHOPIFY_CLI_STORE: adminSession?.storeFqdn,
+    BUNDLE_GEMFILE: join(shopifyCLIDirectory(), 'Gemfile'),
```
amcaplan, line 38: `BUNDLE_GEMFILE` is [documented here](https://bundler.io/v1.15/bundle_config.html#LIST-OF-AVAILABLE-KEYS)

packages/theme/src/cli/commands/theme/check.ts
```diff
@@ -1,15 +1,94 @@
...
-import {cli} from '@shopify/cli-kit'
+import {cli, path} from '@shopify/cli-kit'
+import {Flags} from '@oclif/core'
 
 export default class Check extends Command {
   static description = 'Validate the theme'
 
   static flags = {
     ...cli.globalFlags,
+    ...themeFlags,
```
amcaplan, line 12: The above 2 lines add `--path` and `--verbose` flags.

packages/theme/src/cli/commands/theme/check.ts
```diff
@@ -1,15 +1,94 @@
...
+    // eslint-disable-next-line @typescript-eslint/naming-convention
+    'auto-correct': Flags.boolean({
```
amcaplan, line 14: This list was taken from the help of CLI 2.0.

packages/theme/src/cli/commands/theme/check.ts
```diff
@@ -1,15 +1,94 @@
...
+      description: 'Print Theme Check version',
+      env: 'SHOPIFY_FLAG_VERSION',
+    }),
   }
 
   async run(): Promise<void> {
-    await execCLI2(['theme', 'check'])
+    const {flags} = await this.parse(Check)
+    const passThroughFlags: string[] = []
+    for (const [label, value] of Object.entries(flags)) {
```
amcaplan, line 83: Given the long list of flags, I elected to try a new pattern here. Rather than explicitly declaring each flag (as we do [here](https://github.com/Shopify/cli/blob/main/packages/theme/src/cli/commands/theme/open.ts#L39-L49) and could break each time we add a new flag), I created a generalizable system of choosing which flags to pass through and how. What do we think?

isaacroldan: I like it, as long as the flags have the same name this is a cool generic way of doing it.
Does this work if the user uses chars instead of the full flag? or at this point we always receive the full flag?

PR #193, opened 2022-07-25
Add payment_customization support

jianghong:

This allows us to push payment_customization functions as we are building it out.

### DIFF
--- a/packages/app/src/cli/constants.ts
+++ b/packages/app/src/cli/constants.ts
@@ -60,7 +60,7 @@ export const publicFunctionExtensions = {
   types: ['product_discounts', 'order_discounts', 'shipping_discounts'],
 } as const
 export const functionExtensions = {
-  types: [...publicFunctionExtensions.types, 'payment_methods', 'shipping_rate_presenter'],
+  types: [...publicFunctionExtensions.types, 'payment_methods', 'shipping_rate_presenter', 'payment_customization'],
 } as const
 
 export const functionExtensionTemplates = [
@@ -140,6 +140,8 @@ export const getFunctionExtensionPointName = (type: FunctionExtensionTypes) => {
       return 'SHIPPING_DISCOUNTS'
     case 'payment_methods':
       return 'PAYMENT_METHODS'
+    case 'payment_customization':
+      return 'PAYMENT_CUSTOMIZATION'
     case 'shipping_rate_presenter':
       return 'SHIPPING_METHODS'
   }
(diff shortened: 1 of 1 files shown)


packages/app/src/cli/constants.ts
```diff
@@ -140,6 +140,8 @@ export const getFunctionExtensionPointName = (type: FunctionExtensionTypes) => {
       return 'SHIPPING_DISCOUNTS'
     case 'payment_methods':
       return 'PAYMENT_METHODS'
+    case 'payment_customization':
```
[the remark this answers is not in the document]
jianghong: Yup that's right

DuncanUszkay1:
Is there intent to merge this or are we just using the branch?

If we're merging, we'll need a strategy to ensure this doesn't appear on any option lists and confuse devs 🤔

DuncanUszkay1:
Oh NVM, looks like `PAYMENT_METHODS` is already there... I suppose we have a whitelist of options that we display?

packages/app/src/cli/constants.ts
```diff
@@ -266,6 +268,7 @@ export const extensionGraphqlId = (type: ExtensionTypes) => {
     case 'order_discounts':
     case 'shipping_discounts':
     case 'payment_methods':
+    case 'payment_customization':
     case 'shipping_rate_presenter':
       // As we add new extensions, this bug will force us to add a new case here.
```
DuncanUszkay1, line 273: Super confusing comment 🤔  What bug?

DuncanUszkay1 (approved):
No concerns with the code, just want to validate that we aren't exposing this as a workable option for current API devs. We should also make note that our current process requires us to publicly reveal the name of our API to use it.

packages/app/src/cli/constants.ts
isaacroldan: This is an old comment, no longer applies

DuncanUszkay1:
>payment_methods and shipping_rate_presenter are displayed.

😬 I don't mind if folks find out by snooping through the code, but I wouldn't want someone to think that it was an option only to find out that it doesn't work in production.

@isaacroldan is there any way for us to limit visibility beyond deletion? We used to have a beta flag system in CLI 2.0

jianghong:
>  will functions have eventually something similar?  

Probably won't check for user permission necessarily but we may eventually consolidate this list of constants to some API and hit that API as the single source of truth instead

DuncanUszkay1:
>The non-public types won't be visible in the help menu

This is the context I was missing, as long as we're not surfacing them this is fine for now.

isaacroldan (approved):

jianghong:
Hey @isaacroldan, I see the merge button is blocked for me. How do I merge and ship this?

packages/app/src/cli/constants.ts
```diff
@@ -60,7 +60,7 @@ export const publicFunctionExtensions = {
   types: ['product_discounts', 'order_discounts', 'shipping_discounts'],
```
DuncanUszkay1, line 60: **Context for Guide Readers:** This is the only file that you need to change to prototype your API.

PR #196, opened 2022-07-25
Allow WS connections to the app and not only extensions

isaacroldan:
Initially we only supported 1 websocket through our reverse proxy (because of how it worked), this was the extensions websocket needed for updates and dev-console.

With the proxy refactor we can now easily support multiple websockets, basically 1 per route. So one for `/` and one for `/extensions`

Allow websockets in `/`, fixing the vite HMR issue

### DIFF
--- a/.changeset/light-apples-jump.md
+++ b/.changeset/light-apples-jump.md
@@ -1,0 +1,5 @@
+---
+'@shopify/app': patch
+---
+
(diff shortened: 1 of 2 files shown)


pepicrft (approved):
I'd add a changeset entry

paulomarg (approved):
Nice!

PR #217, opened 2022-07-29
Fix the deployment of extensions from Windows environments

pepicrft:
The deployments are failing on Windows because the paths inside the generated `zip` file follow the Windows' backslash convention for paths. 

I adjusted the `zip` function in `@shopify/cli-kit`'s archive module to [normalize](https://github.com/unjs/pathe/blob/eba7d7cecf2ced83c7e518996d4500a63d54d2ee/src/path.ts#L26) the Zip entries' paths. I also added an integration test to run this on CI Windows environments and ensure the normalization works as expected. 

### How to test your changes?
- Run the integration tests form a Windows environment.
- Try to deploy an app with extensions from a Windows environment.


packages/cli-kit/src/node/archiver.integration.test.ts
```diff
@@ -0,0 +1,36 @@
...
+import StreamZip from 'node-stream-zip'
+
+describe('zip', () => {
+  test('zips a directory', async () => {
+    await inTemporaryDirectory(async (tmpDir) => {
+      // Given
+      const zipPath = joinPath(tmpDir, 'output.zip')
+      const outputDirectoryName = 'output'
+      const outputDirectoryPath = joinPath(tmpDir, outputDirectoryName)
+      const structure = [`extensions/first/main.js`, `test.json`]
```
isaacroldan, line 14: Should this test inlcude the `\` for the windows case to test that is is normalized?

pepicrft: Note that that's what happens in the implementation does `glob`. We'll get paths that are `C:\\...` that need to be normalized.

.changeset/khaki-bottles-flow.md
```diff
@@ -0,0 +1,5 @@
+---
+'@shopify/app': minor
+---
+
+Fix deployment of extensions on Windows environments
```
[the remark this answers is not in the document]
pepicrft: Yep

PR #220, opened 2022-07-29
Make inquirer default for inputs

alvaro-shopify:

Fixes https://github.com/Shopify/shopify-cli-planning/issues/298

- Refactor rendering prompts library selection and execution trying to add a higher abstraction level
- Inquirer prompts are used by default. 
- Enquirer integration remains available using the env variable SHOPIFY_USE_ENQUIRER
- There are no differences in the behaviour of `select` component
- There are no differences in the behaviour of `password` component
- There are slight differences in the behaviour `input` and `autocomplete`. Please click on the sections below to review them
<details>
  <summary>Old Input vs. New Input</summary>
<br/>
Old Input
<br/>

<br/>
<br/>
<br/>
New Input
<br/>

<br/>
<br/>
<br/>
Diferences:
<br>
<ul>
  <li>Cursor color</li>
  <li>Underline style, thinner and separated from the input characters</li>
</ul>
</details>

<details>
  <summary>Old autocomplete vs. New autocomplete</summary>
<br/>
Old autocomplete
<br/>

<br/>
<br/>
<br/>
New autocomplete
<br/>

<br/>
<br/>
<br/>
Diferences:
<br>
<ul>
  <li>Paginated choices (fixed to 10)</li>
</ul>
</details>

### How to test your changes?
- Run the command `SHOPIFY_USE_INQUIRER=1 yarn shopify app dev --reset`
  - First prompt is using new `select` component
    - Select `Yes, create it as a new app` to check the `input` component
    - Select `No, connect it to an existing app` to check the `autocomplete` component. Some app should be created first


gonzaloriestra (changes requested):
As we talked, there are a couple of issues to fix on Windows:

1. The cursor appears after the placeholder
win-cursor

2. The errors appear in the same line, instead of in the next one in red:
win-error

Regarding the ugly underline, it would be nice if you can find a quick workaround, but otherwise we can take care as part of https://github.com/Shopify/shopify-cli-planning/issues/251

alvaro-shopify:
> As we talked, there are a couple of issues to fix on Windows:
> 
> 1. The cursor appears after the placeholder
>    win-cursor
> 2. The errors appear in the same line, instead of in the next one in red:
>    win-error
> 
> Regarding the ugly underline, it would be nice if you can find a quick workaround, but otherwise we can take care as part of [Shopify/shopify-cli-planning#251](https://github.com/Shopify/shopify-cli-planning/issues/251)

1. Regards first point, this problems is shared in every operative systems. Inquirer screen manager that handles cursor position and so on should be overwritten. In that case we should wonder if hacking the library so much is recommended.

2. I will try to overwrite that behaviour in the render method.

alvaro-shopify:
1. Finally the cursor is set at the first position of the input
2. Errors now is shown at the bottom of the input
image
3. Underline style changed to make it compatible with every ascii terminal. Slight differences with the older one, thinner and more separation between input characters and the underline
image
---
image
4. Inquirer prompt used by default. Enquirer still works using SHOPIFY_USE_ENQUIRER. The code is refactored so prompts managing has a little abstraction. It could be interesting to leave the `enquirer` integration in case we need a component not implemented in `inquirer`

packages/cli-kit/package.json
```diff
@@ -86,11 +86,13 @@
     "find-versions": "^5.0.0",
     "form-data": "^4.0.0",
     "fs-extra": "^10.0.0",
+    "fuzzy": "^0.1.3",
```
gonzaloriestra, line 89: I'm missing changes in the yarn.lock, could you push it?

packages/cli-kit/src/ui/executor.ts
```diff
@@ -0,0 +1,88 @@
...
+    return (await inquirer.prompt(question as QuestionCollection<Answers>, {...(question as Question).choices}))[
+      questionName
+    ]
+  }
+}
+
+function inquirerMapper(question: Question): unknown {
+  switch (question.type) {
+    case 'input':
+    case 'password':
```
gonzaloriestra, line 35: The password type is not hiding the characters. You can check it with `dev shopify ngrok auth`.

gonzaloriestra (commented):
Almost there! Once all the components work in a similar way, I'd completely remove the Enquirer library.

packages/cli-kit/src/ui/executor.ts
alvaro-shopify: Ouch!. I didn't realize about it. Hope to have it ready soon

gonzaloriestra (approved):
Working like a charm now! 👏 

My only concern now is the Enquirer library. What's the point of keeping it? I'd remove it for several reasons:
- It's not a light library (0.2MB)
- It won't be used, as the flag is not documented
- In case we miss anything, it will probably be a visual detail

alvaro-shopify:
Added highlighting feature to the filtered result list when autocompleting. Updated comment content in the PR about differences between autocomplete versions

packages/cli-kit/src/ui/inquirer/input.ts
```diff
@@ -0,0 +1,55 @@
...
+    const isFinal = this.status === 'answered'
+
+    let prompt = this.getQuestion()
+
+    let bottomContent = ''
+    if (isFinal) {
+      prompt += `${colors.dim('·')} ${color(this.formatContent(this.answer))}`
+    } else {
+      prompt += `\n${color('>')} ${this.rl.line ? this.formatContent(this.rl.line) : color.dim(this.opt.default)}`
+      bottomContent = '─'.repeat(Math.max(this.rl.line.length, 30))
```
isaacroldan, line 26: This is the bottom line? have you tried using `'▔'` as we do in `enquirer`?

yarn.lock
```diff
@@ -4160,7 +4160,7 @@ end-of-stream@^1.0.0, end-of-stream@^1.1.0, end-of-stream@^1.4.1, end-of-stream@
   dependencies:
     once "^1.4.0"
 
-enquirer@^2.3.0, enquirer@^2.3.6, enquirer@~2.3.6:
+enquirer@^2.3.0, enquirer@~2.3.6:
```
gonzaloriestra, line 4163: Ouch, there are still references here... It looks like is a nested dependency of Nx

packages/cli-kit/src/ui/executor.ts
```diff
@@ -1,33 +1,23 @@
...
 import {CustomPassword} from './inquirer/password.js'
-import {isTruthy} from '../environment/utilities.js'
 import {Question} from '../ui.js'
 import inquirer, {Answers, QuestionCollection} from 'inquirer'
-import enquirer from 'enquirer'
 import fuzzy from 'fuzzy'
 
 export function getMapper(): (question: Question) => unknown {
-  return isEnquirer() ? enquirerMapper : inquirerMapper
+  return inquirerMapper
```
gonzaloriestra, line 10: What about moving the code of inquirerMapper directly here?

yarn.lock
alvaro-shopify: Yes, it seems that cli will have to download this dependency anyway.

packages/cli-kit/src/ui/executor.ts
alvaro-shopify: This `executor` was originally planned as an abstraction layer for the cli ui module. The `executor` implements integration with different external libraries, so adding a new one doesn't require to modify cli core code base. Initially enquirer and inquirer were supported. We could remove completely the `executor` implementation for now

PR #225, opened 2022-08-01
Fix flaky checksum hash validation test

matteodepalo:

The test was calling an async function (without waiting for it) and setting up an expectation on it, but that meant that sometimes the function would throw before having the chance being caught by vitest.

Instead of calling the function before passing it to `expect` we wrap it into another function and then pass it to `expect`.

### DIFF
--- a/packages/cli-kit/src/node/checksum.test.ts
+++ b/packages/cli-kit/src/node/checksum.test.ts
@@ -41,7 +41,7 @@ describe('validate', () => {
       }
       vi.mocked(fetch).mockResolvedValue(response)
 
-      const got = validateMD5({file, md5FileURL: 'https://test.shopify.com/md5-file'})
+      const got = () => validateMD5({file, md5FileURL: 'https://test.shopify.com/md5-file'})
 
       // When
       await expect(got).rejects.toThrowError(InvalidChecksumError({file, expected: remoteHash, got: hash}))


isaacroldan (approved):
If this is the intended way of catching errors by vitest, we might need to review other tests too. (no need to do it know, just if we see that other flaky tests appear)

shauns:
Going to offer a different view than @isaacroldan and say we should update the other tests 😆  -- If you search for `rejects.toThrowError` there's not really that many, and _most_ of them already do things the preferred way.

Reason -- generally lots of tests may start with copy/pasting or looking at how some other test case does something: let's make sure every test is safe to copy from

matteodepalo:
Makes sense, I'll check the other tests and also double check that this fix actually does something, because I haven't been able to reproduce this locally, this is just a guess based on an intuitive understanding of promises, but I might be off.

isaacroldan:
I haven't been able to reproduce it either, so we think this should fix it but maybe is worth to first merge and see what happens with a few PRs? I can create an issue to come back to this later this week and update all the pending tests if we don't see any more flakyness from this one.

I see there are a bunch (23) of tests doing `rejects.toThrowError` but none of them encapsulate the execution in a function, @shauns what do you mean when you say _most_ of them already do things the preferred way?

matteodepalo:
I've managed to reproduce this.

This fails all the time

```
const got = validateMD5({file, md5FileURL: 'https://test.shopify.com/md5-file'})

      // wait 2 seconds
      await new Promise((resolve) => setTimeout(resolve, 2000))

      // When
      await expect(got).rejects.toThrowError(InvalidChecksumError({file, expected: remoteHash, got: hash}))
```

My fix doesn't.

```
const got = () => validateMD5({file, md5FileURL: 'https://test.shopify.com/md5-file'})

      // wait 2 seconds
      await new Promise((resolve) => setTimeout(resolve, 2000))

      // When
      await expect(got).rejects.toThrowError(InvalidChecksumError({file, expected: remoteHash, got: hash}))
```

That said, I find it odd that vitest complains about exeptions thrown by an async function since they should just result in a rejected promise 🤔 Maybe vitest just wraps everything in a `try` `catch` and warns you if you have some uncaught rejections? So my initial explanation is not really correct, async functions shouldn't blow up the main thread.

isaacroldan:
Great that you could reproduce it! Then it makes sense to update all the `reject` tests with this approach :)

matteodepalo:
Fixes https://github.com/Shopify/shopify-cli-planning/issues/333

PR #232, opened 2022-08-03
Fix targetPort is always random. Before generating new randomPort ch…

### LINKED ISSUE #136: [Bug]: Static Ngrok URLs no longer work

App

### Expected behavior

Navigate to the static ngrok urls I've been using for dev in the past and everything tunnel correctly

### Actual behavior

Overrides tunneling. Even though by default it uses ngrok, it suggests cloudflared as the approved method. Conflicting documentation and implementation. 

### Stack trace

_No response_

### Reproduction steps

1. ngrok start configname
2. npm run dev
3. or npm run dev -- --tunnel-url ngrokURL
4. overrides and fails

### Operating System

MAC OS

3

### Shell

_No response_

_No response_

_No response_

wisniewski94:

Fixes #136 

Possible issue: same port value for each target inside .map()

This will try to assign a port from the function parameter.

### How to test your changes?

This should make --tunnel-url parameter work

### DIFF
--- a/packages/app/src/cli/utilities/app/http-reverse-proxy.ts
+++ b/packages/app/src/cli/utilities/app/http-reverse-proxy.ts
@@ -40,7 +40,7 @@ export async function runConcurrentHTTPProcessesAndPathForwardTraffic(
 
   const processes = await Promise.all(
     proxyTargets.map(async (target): Promise<output.OutputProcess> => {
-      const targetPort = await port.getRandomPort()
+      const targetPort = portNumber ?? (await port.getRandomPort())
       rules[target.pathPrefix ?? '/'] = `http://localhost:${targetPort}`
       return {
         prefix: target.logPrefix,


isaacroldan:
Hi @wisniewski94 , thank you for your PR :)

I'm afraid that this change is not valid:

- We have a reverse proxy that is listening to a specific port (either `portNumber` if provided or a random one). If you provide a custom tunnel with `--tunnel-url=https://mytunnelprovider.com:8000` we will use the 8000 port for the reverse proxy. (Because that's were the tunnel provider will redirect the traffic). The proxy will then be in charge of redirecting the traffic to the correct local server.

- The port you are modifying here is the one used for the local servers (web, extensions, backend, etc...) that are served during `dev`. This needs to be a random port because each server needs a different one, you can't use the same as the proxy.

Hope everything is clear, but let me know if it's not!

wisniewski94:
Thanks for the quick response @isaacroldan 

The problem I am trying to solve is in fact related to dev mode. The randomness here is quite problematic and I have the feeling that having a random port at each dev server run is not a good idea. I expected that provided tunnel will work in dev mode.

So just to confirm:

If I enter --tunnel-url 'https://mytunnel.test:22222' in dev mode the tunnel is not going to have 22222 port but 8000?

wisniewski94:
I have removed my previous comment because I didn't read your response right. I have checked the code again and you are right. I was focusing on targetPort too much and missed the obvious line ```server.listen(availablePort)```

Everything works as expected :)

PR #233, opened 2022-08-03
Use our list of words when generating app and extension names

matteodepalo:

We want to make sure that the words used by our random name generator are part of a list of safe words. We also want to remove the token at the end of the generated name since it doesn't look very clean.

Fixes https://github.com/Shopify/shopify-cli-planning/issues/286

This PR makes sure that:
- The list of word used is hard coded and safe
- We re-roll the generated app name in case of duplicates

It also removes the haikunator dependency.


packages/cli-kit/src/haiku.ts
```diff
@@ -1,9 +1,65 @@
...
+    tokenLength: 0,
+  },
+})
 
 export function generate(suffix: string): string {
   const generated = haikunator.haikunate()
-  const [adjective, noun, token] = generated.split('-')
-  return [adjective, noun, suffix, token].join('-')
+  const [adjective, noun] = generated.split('-')
+  return [adjective, noun, suffix].join('-')
```
isaacroldan, line 64: Can this be just:

```ts
const generated = haikunator.haikunate()
return [generated, suffix].join('-')
```
?

packages/cli-kit/src/haiku.test.ts
```diff
@@ -0,0 +1,29 @@
+import {inTemporaryDirectory, write} from './file.js'
+import {join} from './path.js'
+import {generate} from './haiku.js'
+import {describe, expect, test} from 'vitest'
+
+describe('generate', () => {
+  test('rerolls the name if a directory exists with the same name', async () => {
+    // make randonmess predictable and pick from the array of adjectives and nouns
```
matteodepalo, line 8: Actually I discovered there is a better way to do this with function spies, I'll refactor it with that.

packages/app/src/cli/commands/app/scaffold/extension.ts
```diff
@@ -80,6 +80,7 @@ export default class AppScaffoldExtension extends Command {
       extensionTypesAlreadyAtQuota: this.limitedExtensionsAlreadyScaffolded(app),
       name: flags.name,
       extensionFlavor,
+      directory: `${directory}/extensions`,
```
isaacroldan, line 83: always use path.join, as the `/` won't be compatible with Windows :)

packages/cli-kit/src/haiku.ts
```diff
@@ -1,9 +1,77 @@
...
+  'deal',
+  'merchandise',
+  'transaction',
+  'sale',
+]
+
+export async function generate({
+  suffix,
+  directory,
+  randomPick = (array: string[]) => array[Math.floor(Math.random() * array.length)],
```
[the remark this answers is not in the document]
matteodepalo: I just wanted to make the function injectable but I can definitely add a constant so that this becomes shorter.

packages/create-app/src/prompts/init.test.ts
```diff
@@ -23,7 +24,9 @@ describe('init', () => {
         name: 'name',
         preface: '\nWelcome. Let’s get started by naming your app. You can change it later.',
         message: "Your app's name?",
-        default: expect.stringMatching(/^\w+-\w+-app-\d+$/),
+        default: expect.stringMatching(
+          new RegExp(`^${haiku.SAFE_ADJECTIVES.join('|')}-${haiku.SAFE_NOUNS.join('|')}-app$`),
+        ),
```
isaacroldan, line 29: since you are adding specific tests to `haiku.ts`, there is no need to test again here that the generated values are "valid". 
I'd revert this test, and maybe make `SAFE_ADJECTIVES` and `SAFE_NOUNS` private if possible

packages/cli-kit/src/haiku.test.ts
```diff
@@ -0,0 +1,27 @@
+import {inTemporaryDirectory, write} from './file.js'
+import {join} from './path.js'
+import {generate} from './haiku.js'
+import {describe, expect, test, vi} from 'vitest'
+
+describe('generate', () => {
+  test('rerolls the name if a directory exists with the same name', async () => {
```
isaacroldan, line 7: We could add a test in this file to check that the generated value follows the expected pattern.

matteodepalo: Makes sense, I'll move the check from the prompt test.

packages/cli-kit/src/haiku.ts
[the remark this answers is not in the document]
isaacroldan: If this is only for testing, it can be it's own function and it can be mocked.

A `randomPick` function might be needed in another places, I think we should make it reusable (and you can mock it if needed for tests!)

packages/cli-kit/src/haiku.ts
```diff
@@ -1,9 +1,73 @@
...
+  return `${randomPick(SAFE_ADJECTIVES)}-${randomPick(SAFE_NOUNS)}`
+}
+
+export async function generate({suffix, directory}: {suffix: string; directory: string}): Promise<string> {
+  const generated = `${generateRandomName()}-${suffix}`
+  const isAppDirectoryTaken = await exists(join(directory, generated))
+
+  if (isAppDirectoryTaken) {
+    // eslint-disable-next-line no-return-await
+    return await generate({suffix, directory})
```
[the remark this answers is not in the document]
matteodepalo: Good point

PR #234, opened 2022-08-03
Allow running `shopify` instead of `npm/yarn/pnpm shopify`

amcaplan:

Allows someone who's installed the shopify CLI package globally to run `shopify` instead of `npm/yarn/pnpm shopify`. This should help avoid users getting confused as we will shortly be supporting a globally installed CLI for certain use cases.

Since `run.js` is the entrypoint to the CLI, we hook in there and check if the running script is the same we'd expect based on `npm list` in the current working directory.

If it's the same script, we can just move on as usual.

If `npm list` fails, it's likely because you're not inside a directory with a `package.json` or [your dependencies are messed up](https://github.com/npm/npm/issues/17624#issuecomment-313294516), and we just go on using the current CLI.

If you're running a different script, we run the correct `run.js`, handing over `argv` as is (excluding the first 2, which based on my testing should be `node` and the current script).

### How to test your changes?

```bash
$ npm install -g /path/to/your/local/repo/copy
$ cd ~
$ shopify version # should be the same as your development version
$ cd /path/to/some/app
$ shopify version # should be the same as your app's installed @shopify/cli version
$ shopify app info --verbose # tests that flags are passed correctly, and also shows the correct version

### DIFF
--- a/packages/cli-kit/src/constants.ts
+++ b/packages/cli-kit/src/constants.ts
@@ -31,6 +31,8 @@ const constants = {
     codespaces: 'CODESPACES',
     codespaceName: 'CODESPACE_NAME',
     gitpod: 'GITPOD_WORKSPACE_URL',
+    enableCliRedirect: 'SHOPIFY_ENABLE_CLI_REDIRECT',
+    skipCliRedirect: 'SHOPIFY_SKIP_CLI_REDIRECT',
   },
   paths: {
     executables: {
--- a/packages/cli-kit/src/constants.ts
+++ b/packages/cli-kit/src/constants.ts
@@ -24,6 +24,8 @@ const constants = {
     noAnalytics: 'SHOPIFY_CLI_NO_ANALYTICS',
     alwaysLogAnalytics: 'SHOPIFY_CLI_ALWAYS_LOG_ANALYTICS',
     firstPartyDev: 'SHOPIFY_CLI_1P_DEV',
+    enableCliRedirect: 'SHOPIFY_CLI_ENABLE_CLI_REDIRECT',
+    skipCliRedirect: 'SHOPIFY_CLI_SKIP_CLI_REDIRECT',
     debugGoBinary: 'SHOPIFY_DEBUG_GO_BINARY',
     deviceAuth: 'SHOPIFY_CLI_DEVICE_AUTH',
     // Variables to detect if the CLI is running in a cloud environment
@@ -31,8 +33,6 @@ const constants = {
     codespaces: 'CODESPACES',
     codespaceName: 'CODESPACE_NAME',
     gitpod: 'GITPOD_WORKSPACE_URL',
-    enableCliRedirect: 'SHOPIFY_ENABLE_CLI_REDIRECT',
-    skipCliRedirect: 'SHOPIFY_SKIP_CLI_REDIRECT',
   },
   paths: {
     executables: {
--- a/packages/cli-kit/src/node/cli.ts
+++ b/packages/cli-kit/src/node/cli.ts
@@ -43,6 +43,7 @@ export async function runCreateCLI(options: RunCLIOptions) {
 }
 
 export async function replaceGlobalCLIWithLocal(filepath: string): Promise<boolean> {
+  if (process.env.SHOPIFY_SKIP_CLI_REDIRECT) return false
   if (process.env.npm_config_user_agent) return false
 
   const cliPackage = await localCliPackage()
@@ -51,7 +52,10 @@ export async function replaceGlobalCLIWithLocal(filepath: string): Promise<boole
   const correctExecutablePath = join(cliPackage.path, cliPackage.bin.shopify)
   if (correctExecutablePath === filepath) return false
   try {
-    await exec(correctExecutablePath, process.argv.slice(2, process.argv.length), {stdio: 'inherit'})
+    await exec(correctExecutablePath, process.argv.slice(2, process.argv.length), {
+      stdio: 'inherit',
+      env: {SHOPIFY_SKIP_CLI_REDIRECT: '1'},
+    })
     // eslint-disable-next-line no-catch-all/no-catch-all, @typescript-eslint/no-explicit-any
   } catch (processError: any) {
     process.exit(processError.exitCode)
--- a/packages/cli-kit/src/node/cli.ts
+++ b/packages/cli-kit/src/node/cli.ts
@@ -48,7 +48,12 @@ export async function replaceGlobalCLIWithLocal(filepath: string): Promise<boole
 
   const correctExecutablePath = join(cliPackage.path, cliPackage.bin.shopify)
   if (correctExecutablePath === filepath) return false
-  await exec(correctExecutablePath, process.argv.slice(2, process.argv.length), {stdio: 'inherit'})
+  try {
+    await exec(correctExecutablePath, process.argv.slice(2, process.argv.length), {stdio: 'inherit'})
+    // eslint-disable-next-line no-catch-all/no-catch-all, @typescript-eslint/no-explicit-any
+  } catch (processError: any) {
+    process.exit(processError.exitCode)
+  }
   return true
 }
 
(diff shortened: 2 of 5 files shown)

### CODE REVIEW

File: packages/cli-main/bin/run.js [hunk 1]
```diff
@@ -2,6 +2,8 @@
 
 process.removeAllListeners('warning');
 
-import runCLI from "@shopify/cli";
+import runCLI, {replaceGlobalCLIWithLocal} from "@shopify/cli";
 
-runCLI();
+// If we run a local CLI instead, don't run the global one again after!
```
shauns, line 7: I think we should hit the same path in `dev.js` (for testing, if possible)

packages/cli-kit/src/node/cli.ts
```diff
@@ -52,4 +53,36 @@ export async function runCreateCLI(options: RunCLIOptions) {
...
 }
 
+export async function replaceGlobalCLIWithLocal(filepath: string): Promise<boolean> {
+  const cliPackage = await localCliPackage()
+  if (!cliPackage) return false
+
+  const correctExecutablePath = join(cliPackage.path, cliPackage.bin.shopify)
+  if (correctExecutablePath === filepath) return false
+  try {
+    await exec(correctExecutablePath, process.argv.slice(2, process.argv.length), {stdio: 'inherit'})
```
shauns, line 63: We should also `debug` if we hit this path, probably including printing out the args and the path.

packages/cli-main/bin/run.js
```diff
@@ -2,6 +2,8 @@
...
+const ranLocalInstead = await replaceGlobalCLIWithLocal(import.meta.url);
+if (!ranLocalInstead) runCLI();
```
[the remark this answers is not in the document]
shauns: we wouldn't want to pass this env variable to the spawned process though.

packages/cli-kit/src/node/cli.ts
```diff
@@ -52,4 +53,44 @@ export async function runCreateCLI(options: RunCLIOptions) {
...
+  let npmListOutput = ''
+  try {
+    npmListOutput = await captureOutput('npm', ['list', '@shopify/cli', '--json', '-l'])
+    // eslint-disable-next-line no-catch-all/no-catch-all
+  } catch (err) {
+    return
+  }
+  const localShopifyCLI = JSON.parse(npmListOutput)
+  const dependenciesList =
+    localShopifyCLI.dependencies ?? localShopifyCLI.devDependencies ?? localShopifyCLI.peerDependencies
```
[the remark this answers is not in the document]
amcaplan: Very true! Fixed.

shauns:
Leaving this unapproved whilst we decide on action based on the above. But happy with the changes 👍

packages/cli-kit/src/constants.ts
```diff
@@ -27,6 +27,8 @@ const constants = {
     noAnalytics: 'SHOPIFY_CLI_NO_ANALYTICS',
     alwaysLogAnalytics: 'SHOPIFY_CLI_ALWAYS_LOG_ANALYTICS',
     firstPartyDev: 'SHOPIFY_CLI_1P_DEV',
+    enableCliRedirect: 'SHOPIFY_ENABLE_CLI_REDIRECT',
+    skipCliRedirect: 'SHOPIFY_SKIP_CLI_REDIRECT',
```
pepicrft, line 31: Nitpick. Since these variables are scoped to the CLI, I'd use the `SHOPIFY_CLI` prefix.

packages/cli-kit/src/node/cli.ts
```diff
@@ -52,4 +54,50 @@ export async function runCreateCLI(options: RunCLIOptions) {
   await runCLI(options)
 }
 
+export async function replaceGlobalCLIWithLocal(filepath: string): Promise<boolean> {
+  // Temporary flag while we test out this feature and ensure it won't break anything!
+  if (!isTruthy(process.env[constants.environmentVariables.enableCliRedirect])) return false
+  if (isTruthy(process.env[constants.environmentVariables.skipCliRedirect])) return false
+  if (process.env.npm_config_user_agent) return false
```
pepicrft, line 61: The connection between the `npm_config_user_agent` environment variable and the returned value is not obvious. What about adding an explanatory comment?

packages/cli-kit/src/node/cli.ts
```diff
@@ -52,4 +54,50 @@ export async function runCreateCLI(options: RunCLIOptions) {
...
+  bin: {shopify: string}
+}
+
+async function localCliPackage(): Promise<CliPackageInfo | undefined> {
+  let npmListOutput = ''
+  try {
+    npmListOutput = await captureOutput('npm', ['list', '@shopify/cli', '--json', '-l'])
+    // eslint-disable-next-line no-catch-all/no-catch-all
+  } catch (err) {
+    return
```
pepicrft, line 92: I'd silently report this to Bugsnag. It's very unlikely that `npm list` breaks its contract, but it's better to be aware when it happens.

packages/cli-kit/src/node/cli.ts
```diff
@@ -52,4 +54,50 @@ export async function runCreateCLI(options: RunCLIOptions) {
...
+  }
+  const localShopifyCLI = JSON.parse(npmListOutput)
```
pepicrft, line 94: I'd include this one in the `try {`. Just in case the output contains additional characters that make the JSON decoding fail.

packages/cli-kit/src/node/cli.ts
```diff
@@ -52,4 +54,50 @@ export async function runCreateCLI(options: RunCLIOptions) {
...
+      env: {[constants.environmentVariables.skipCliRedirect]: '1'},
+    })
+    // eslint-disable-next-line no-catch-all/no-catch-all, @typescript-eslint/no-explicit-any
+  } catch (processError: any) {
+    process.exit(processError.exitCode)
+  }
+  return true
+}
+
+interface CliPackageInfo {
```
pepicrft, line 81: TIL about this schema. Sweet!

packages/cli-kit/src/node/cli.ts
```diff
@@ -52,4 +54,50 @@ export async function runCreateCLI(options: RunCLIOptions) {
...
+export async function replaceGlobalCLIWithLocal(filepath: string): Promise<boolean> {
+  // Temporary flag while we test out this feature and ensure it won't break anything!
+  if (!isTruthy(process.env[constants.environmentVariables.enableCliRedirect])) return false
+  if (isTruthy(process.env[constants.environmentVariables.skipCliRedirect])) return false
+  if (process.env.npm_config_user_agent) return false
+
+  const cliPackage = await localCliPackage()
+  if (!cliPackage) return false
+
+  const correctExecutablePath = join(cliPackage.path, cliPackage.bin.shopify)
```
pepicrft, line 66: Is `cliPackage.bin.shopify` a path relative to the package's directory (e.g. bin/dev.js)?

packages/cli-kit/src/node/cli.ts
```diff
@@ -52,4 +54,50 @@ export async function runCreateCLI(options: RunCLIOptions) {
   await runCLI(options)
 }
 
+export async function replaceGlobalCLIWithLocal(filepath: string): Promise<boolean> {
```
pepicrft, line 57: Nitpick: What about renaming this one to: `runLocalShopifyIfPresent`

packages/cli-kit/src/node/cli.ts
amcaplan: The issue is more that `npm list` can break if the user's `package.json` is broken, as mentioned in the PR description. So we can't rely on `npm list` executing successfully.

Since this is ultimately a convenience feature, I'm fine with defaulting to just assuming we should run without leveraging this feature.

I'm a bit concerned that reporting to bugsnag will just add a lot of noise, what do you think @pepicrft?

packages/cli-kit/src/node/cli.ts
amcaplan: Agreed the method name needed improvement. I went with `useLocalCLIIfDetected`

PR #246, opened 2022-08-05
Pass env var to CLI2 to hide upgrade warning

matteodepalo:

Fixes https://github.com/Shopify/shopify-cli-planning/issues/343

We need to tell CLI2 that upgrade messages can be ingnored when we execute it from CLI3

Pass an env var to CLI2 which it will use to hide the message.

### DIFF
--- a/packages/cli-kit/src/node/ruby.ts
+++ b/packages/cli-kit/src/node/ruby.ts
@@ -9,7 +9,7 @@ import {AdminSession} from '../session.js'
 import {content, token} from '../output.js'
 import {Writable} from 'node:stream'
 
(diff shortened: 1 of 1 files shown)


matteodepalo:
We could still merge this but it wouldn't have any effect until the release happens in shopify-cli 🤔 It's theoretically not blocking but merging it later might make it easier to report the change.

isaacroldan (approved):

matteodepalo:
Let's wait then so we can merge this and close the main issue.

matteodepalo:
Thank you!

PR #247, opened 2022-08-05
Expand command analytics

shauns:

Fixes https://github.com/Shopify/shopify-cli-planning/issues/274

- [x] Update the monorail topic name
- [x] Track `cmd_all`, `cmd_extensions`, `cmd_scaffold`, `app`, `env` topics
- [x] Change metadata tracking to be behind a catch-all by default

### How to test your changes?

Run with `--verbose`, take a look at the analytics payload. Particularly scaffold extension.


shauns:
Keeping draft until Monorail changes land.

packages/cli-kit/src/node/node-package-manager.test.ts
```diff
@@ -43,12 +43,23 @@ describe('packageManagerUsedForCreating', () => {
...
+    const env = {npm_config_user_agent: 'npm'}
+
+    // When
+    const got = packageManagerUsedForCreating(env)
+
+    // Then
+    expect(got).toBe('npm')
+  })
+
   it('returns npm when the package manager cannot be detected', () => {
```
isaacroldan, line 57:
```suggestion
  it('returns unknown when the package manager cannot be detected', () => {
```

packages/cli-kit/src/analytics.test.ts
```diff
@@ -61,11 +64,12 @@ describe('event tracking', () => {
         storeFqdn: 'domain1',
         directory: '/cached',
       })
-      start({command, args, currentTime: currentDate.getTime() - 100})
+      await start({command, args, currentTime: currentDate.getTime() - 100})
 
       // When
       const config = {
         runHook: vi.fn().mockResolvedValue({successes: [], failures: []}),
+        plugins: [],
```
isaacroldan, line 72: should we add some plugins to the test to validate that they are added to the payload?

packages/app/src/cli/commands/app/scaffold/extension.ts
```diff
@@ -64,27 +67,43 @@ export default class AppScaffoldExtension extends Command {
...
-      extensionFlavor,
+      extensionFlavor: flags.template,
       directory: path.join(directory, 'extensions'),
     })
 
+    const {extensionType, extensionFlavor} = promptAnswers
+    await metadata.addPublic(() => ({
+      cmd_scaffold_template_flavor: extensionFlavor,
+      cmd_scaffold_type: extensionType,
+      cmd_scaffold_type_family: extensionTypeCategory(extensionType),
```
isaacroldan, line 98: family vs category, should we use the same naming?

packages/app/src/cli/models/app/loader.ts
```diff
@@ -66,14 +67,20 @@ class AppLoader {
     const configuration = await this.parseConfigurationFile(AppConfigurationSchema, configurationPath)
     const extensionsPath = path.join(this.appDirectory, `${blocks.extensions.directoryName}`)
     const dotenv = await this.loadDotEnv()
-    const functions = await this.loadFunctions(extensionsPath)
-    const uiExtensions = await this.loadUIExtensions(extensionsPath)
-    const themeExtensions = await this.loadThemeExtensions(extensionsPath)
+    const {functions, usedCustomLayout: usedCustomLayoutForFunctionExtensions} = await this.loadFunctions(
+      extensionsPath,
+    )
```
isaacroldan, line 72: what is the custom layout? isn't always false?

packages/cli-kit/src/environment/local.ts
```diff
@@ -90,3 +95,37 @@ export async function hasGit(): Promise<boolean> {
     return false
   }
 }
+
+/**
+ * Gets info on the CI platform the CLI is running on, if applicable
+ */
+export function ciPlatform(env = process.env): {isCI: true; name: string} | {isCI: false; name?: undefined} {
+  if (isTruthy(env.CI)) {
```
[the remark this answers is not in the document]
shauns: It's a standard

packages/app/src/cli/commands/app/scaffold/extension.ts
```diff
@@ -64,27 +67,43 @@ export default class AppScaffoldExtension extends Command {
 
   public async run(): Promise<void> {
     const {flags} = await this.parse(AppScaffoldExtension)
+
+    await metadata.addPublic(() => ({
+      cmd_scaffold_required_auth: false,
+      cmd_scaffold_template_custom: flags['clone-url'] !== undefined,
+      cmd_scaffold_type_owner: '@shopify/app',
```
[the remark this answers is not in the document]
shauns: yes -- there's more notes in the instrumentation spreadsheet, but this one is for "who owned the extension type". they're all owned by `app` for now.

packages/app/src/cli/models/app/loader.ts
shauns: Right now it is, but it's due a change pretty soon -- we're likely to want to allow overriding the globs for the various extensions. It's better to have this field added and populated earlier, even if the value is consistent right now. There's a few bits like this where some fields are added because we have something due in the next couple of weeks.

packages/app/src/cli/commands/app/scaffold/extension.ts
shauns: good point!

isaacroldan (approved):

PR #252, opened 2022-08-08
Bug-fix: Eject early if unable to find backend

shauns:

Fixes https://github.com/Shopify/shopify-cli-planning/issues/350 

One-liner

### How to test your changes?

Dev an app with no backend

### DIFF
--- a/packages/app/src/cli/models/app/loader.ts
+++ b/packages/app/src/cli/models/app/loader.ts
@@ -328,6 +328,7 @@ async function getProjectType(webs: Web[]): Promise<'node' | 'php' | 'ruby' | un
     return
   } else if (backendWebs.length === 0) {
     output.debug('Unable to decide project type as no web backend')
+    return
   }
   const {directory} = backendWebs[0]
 


isaacroldan (approved):

packages/app/src/cli/models/app/loader.ts
```diff
@@ -346,6 +347,11 @@ async function getProjectType(webs: Web[]): Promise<'node' | 'php' | 'ruby' | un
...
 
 async function logMetadataForLoadedApp(app: App) {
-  const projectType = await getProjectType(app.webs)
-  metadata.addPublic({project_type: projectType})
+  try {
+    const projectType = await getProjectType(app.webs)
+    metadata.addPublic({project_type: projectType})
+    // eslint-disable-next-line no-catch-all/no-catch-all
+  } catch (error) {
+    output.debug(`Unable to logMetadataForLoadedApp ${error}`)
```
pepicrft, line 355: This would swallow legit bugs that we want to see reported to Bugsnag.

packages/app/src/cli/models/app/loader.ts
```diff
@@ -328,6 +328,7 @@ async function getProjectType(webs: Web[]): Promise<'node' | 'php' | 'ruby' | un
     return
   } else if (backendWebs.length === 0) {
     output.debug('Unable to decide project type as no web backend')
+    return
```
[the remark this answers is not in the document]
shauns: I'll put that on the other analytics ticket -- the change behind this line was moving some code around, not a new feature

packages/app/src/cli/models/app/loader.ts
shauns: It would, though not bugs we want to justify crashing for. Mentioned to Isaac, but will add something for doing soft-reporting to Bugsnag for these scenarios, where we can still proceed with execution. But not in this branch - got to get the fix out soon.

packages/app/src/cli/models/app/loader.ts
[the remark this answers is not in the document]
pepicrft: Note it's not a new feature but fixing an implementation that's not accurate.

packages/app/src/cli/models/app/loader.ts
pepicrft: It's not a big deal if we go ahead with merging this as it is, but I've seen projects taking this stance too far and turning many  the bugs into "not worth the crash". I'd lean on the side of letting errors happen and having the right systems in place (i.e. tests) to catch those early.

pepicrft (approved):

PR #254, opened 2022-08-08
Move & modify shopify-cli-extensions GitHub workflow actions

dmhenry:

Fixes [335](https://github.com/Shopify/shopify-cli-planning/issues/335) 

Move `shopify-cli-extensions` GitHub workflow actions & modify them to make them work under `cli`. This is necessary as part of the merge of `shopify-cli-extensions` under `cli`.

- Moving `shopify-cli-extensions` GitHub workflow files to the corresponding top-level directory in `cli/.github/workflow`
- Adjusting the environment, e.g., paths, to make the checks functional

### How to test your changes?

- Verify the GitHub Checks pass


packages/extensions/shopify-cli-extensions/packages/shopify-cli-extensions/tsconfig.json
```diff
@@ -7,6 +7,7 @@
     "outDir": "./",
     "declaration": true,
     "strict": true,
+    "skipLibCheck": true,
```
vividviolet, line 10: Why did we need to add this? If it's due to a typescript version conflict can we bump the TS version for this package?

.github/workflows/shopify-cli-extensions-testing.yml
```diff
@@ -1,21 +1,29 @@
-name: Testing
+name: shopify-cli-extensions-testing
```
vividviolet, line 1: Let's use the same naming convention throughout - currently the extentions ones are all sentences

.github/workflows/shopify-cli-extensions-release.yml
```diff
@@ -6,7 +6,13 @@ on:
   pull_request:
     branches: [main]
 
+env:
+  ARGO_EXTENSIONS_DIR: packages/extensions/shopify-cli-extensions
+
 jobs:
+  defaults:
+    run:
+      working-directory: "$env.ARGO_EXTENSIONS_DIR"
```
vividviolet, line 15: I'm curious why we need to set this to an env variable?

alfonso-noriega: Probably this variable can also be reused within the makefile to avoid hardcoding the path the different steps.

alfonso-noriega (approved):

packages/extensions/shopify-cli-extensions/packages/shopify-cli-extensions/tsconfig.json
pepicrft: Correct. There're are two versions of the same nested dependency through different branches and the Typescript validation ends up failing because of that.

PR #266, opened 2022-08-10
Add option to choose typescript templates for ui extensions

matteodepalo:

Fixes https://github.com/Shopify/shopify-cli-planning/issues/329

The go binary already supported passing an extension flavor that includes the word `typescript`, so we want to leverage that option to allow users to select typescript for ui extension templates.

I've added two new options to the extension flavor question:
- `Typescript React`
- `Typescript`

### How to test your changes?

Run `dev fixture scaffold extension` and choose any of the new options. The file names will be either `ts` or `tsx`.

### DIFF
--- a/packages/app/src/cli/constants.ts
+++ b/packages/app/src/cli/constants.ts
@@ -89,8 +89,8 @@ export type UIExtensionTypes = typeof uiExtensions.types[number]
 export const uiExtensionTemplates = [
   {name: 'TypeScript', value: 'typescript'},
   {name: 'Vanilla JavaScript', value: 'vanilla-js'},
+  {name: 'JavaScript React', value: 'react'},
   {name: 'TypeScript React', value: 'typescript-react'},
-  {name: 'Vanilla JavaScript React', value: 'vanilla-js-react'},
 ]
 
 export function isUiExtensionType(extensionType: string) {
--- a/packages/app/src/cli/constants.ts
+++ b/packages/app/src/cli/constants.ts
@@ -87,9 +87,9 @@ export const activeUIExtensions = {
 export type UIExtensionTypes = typeof uiExtensions.types[number]
 
 export const uiExtensionTemplates = [
-  {name: 'Typescript', value: 'typescript'},
+  {name: 'TypeScript', value: 'typescript'},
   {name: 'Vanilla JavaScript', value: 'vanilla-js'},
-  {name: 'Typescript React', value: 'typescript-react'},
+  {name: 'TypeScript React', value: 'typescript-react'},
   {name: 'Vanilla JavaScript React', value: 'vanilla-js-react'},
 ]
 
--- a/packages/app/src/cli/constants.ts
+++ b/packages/app/src/cli/constants.ts
@@ -88,9 +88,9 @@ export type UIExtensionTypes = typeof uiExtensions.types[number]
 
 export const uiExtensionTemplates = [
   {name: 'TypeScript', value: 'typescript'},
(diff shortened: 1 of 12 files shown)


packages/app/src/cli/constants.ts
```diff
@@ -87,8 +87,10 @@ export const activeUIExtensions = {
 export type UIExtensionTypes = typeof uiExtensions.types[number]
 
 export const uiExtensionTemplates = [
-  {name: 'React', value: 'react'},
-  {name: 'vanilla JavaScript', value: 'vanilla-js'},
+  {name: 'TypeScript', value: 'typescript'},
+  {name: 'Vanilla JavaScript', value: 'vanilla-js'},
```
matteodepalo, line 91: I'm starting to question wether we need `Vanilla` at this point. We could just go with `Javascript` now that we have `Typescript`

isaacroldan: Probably yes, maybe @MeredithCastile should confirm the specific names for this 🤔

isaacroldan:
Don't forget to update the `--template` flag in scaffold command, it should accept the new templates...

Another detail, Hydrogen team recently added a `--ts` flag in their create command and split the template/language options in two questions. Not saying we should do that, but just mentioning it here in case we want to keep consistency somehow (although is not exactly the same)

matteodepalo:
I was thinking about having this as a separate question. I don't feel strongly about either option, but it made sense to put all in the same question.

One thing I'm not sure of is, do we need to keep the `react` flag as `react` or can we rename it to `vanilla-js-react`? In theory we could be breaking user scripts that use our commands with certain flags. If that's the case I can just rename it back to `react`.

packages/app/src/cli/constants.ts
isaacroldan: umm, I think it's ok if we remove the Vanilla everywhere, it's easy to understand:
- TypeScript
- JavaScript
- TypeScript React
- JavaScript React

What do you think?

pepicrft (approved):

matteodepalo:
@pepicrft @isaacroldan may I have a review of the last commit? I've added some acceptance tests, just want to make sure that they make sense.

packages/features/features/app.feature
```diff
@@ -1,10 +1,13 @@
 Feature: Extension creation
 
-Scenario: I scaffold theme, ui, and function extensions
+Background:
```
pepicrft, line 3: TIL

packages/app/src/cli/constants.ts
MeredithCastile: Let's update this to be a framework question. Like so:

> What framework would you like to use?
> - JavaScript, no additional framework
> - TypeScript
> - TypeScript React
> - JavaScript React

matteodepalo: I like the idea of reframing this question to be a framework question, but should we repeat the "no additional framework" like for `TypeScript` as well since that's also just a language choice?

MeredithCastile: You've got me deep into [internet research](https://dev.to/bytebodger/is-typescript-really-a-language-ep1#:~:text=So%20apparently%20their%20survey%20endorses,lists%20TypeScript%20as%20a%20language.) on whether typescript is a language or a framework, @matteodepalo! I always understood TypeScript to be a validation "framework" on top of JavaScript, but I guess it's a "language" to most people.

What about getting rid of the noun altogether?

> What would you like to work in?
> - JavaScript
> - TypeScript
> - TypeScript React
> - JavaScript React

PR #268, opened 2022-08-10
Use extensions binary source code in local environment

alvaro-shopify:
In case the CLI is launched in a local environment a local built extensions binary is used


packages/app/src/cli/utilities/extensions/cli.test.ts
```diff
@@ -58,7 +86,25 @@ describe('runGoExtensionsCLI', () => {
...
+    vi.mocked(path.findUp).mockResolvedValue(extensionsGoCliExecutable)
+
+    // When
+    const got = await nodeExtensionsCLIPath()
+
+    // Then
+    expect(got).not.toBeUndefined()
+  })
+
+  test('returns the path when running not running it locally', async () => {
```
alfonso-noriega, line 102: I guess one 'running' should be out ;)

packages/app/src/cli/utilities/extensions/cli.ts
```diff
@@ -19,35 +15,18 @@ const NodeExtensionsCLINotFoundError = () => {
...
 export async function runGoExtensionsCLI(args: string[], options: system.WritableExecOptions = {}) {
   const stdout = options.stdout || {write: () => {}}
-  if (useExtensionsCLISources()) {
-    const projectDirectory = path.join(
-      environment.local.homeDirectory(),
-      'src/github.com/shopify/shopify-cli-extensions',
-    )
-    stdout.write(`Using extensions CLI from ${projectDirectory}`)
+  if (environment.local.isDebug()) {
+    const extensionsGoCliDirectory = (await path.findUp('packages/ui-extensions-go-cli/', {
```
alfonso-noriega, line 19: would it make sense to store this kind of paths to packages in a utility, something like a directory resolver for packages?

alfonso-noriega (approved):

packages/app/src/cli/utilities/extensions/cli.ts
alvaro-shopify: Good point. I will create an issue to add that feature.

PR #270, opened 2022-08-10
Use Node 16.13.1 in development

pepicrft:
One of the extensions' package requires Node 16.13.1 in development and therefore we need to lower the Node version used in the repository. Note that this is in development. The resulting CLI should run in versions above that and that's something we are currently checking through our acceptance tests in CI.

Lowering the Node version used in development to 16.13.1

### How to test your changes?
You can try to run the CLI locally against the app in `fixtures/app`


isaacroldan:
Hey, I thought this wasn't necessary anymore 🤔 
I removed node-sass from extensions-cli and that was the only dependency that required node 18 right? is there another one?

PR #276, opened 2022-08-11
Rename SHOPIFY_CONFIG to SHOPIFY_ENV

pepicrft:
While working with @alvaro-shopify on merging the `shopify-cli-extensions` projects we realized that the `SHOPIFY_CONFIG=debug` environment variable is misleading to represent a development environment. 

I'm renaming the variable to `SHOPIFY_ENV=development`, which aligns with the convention many projects follow.

### How to test your changes?
Tests should pass on CI and you should be able to interact with the fixture project.

### DIFF
--- a/packages/cli-main/bin/dev.js
+++ b/packages/cli-main/bin/dev.js
@@ -2,7 +2,7 @@
 
 process.removeAllListeners('warning');
 
-process.env.SHOPIFY_CONFIG = "debug"
+process.env.SHOPIFY_ENV = process.env.SHOPIFY_ENV ?? "development"
 
 import runCLI from "../dist/index.js";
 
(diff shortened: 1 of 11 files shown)


packages/cli-main/bin/dev.js
```diff
@@ -2,7 +2,7 @@
 
 process.removeAllListeners('warning');
 
-process.env.SHOPIFY_CONFIG = "debug"
+process.env.SHOPIFY_ENV = process.env.SHOPIFY_ENV ?? "development"
```
pepicrft, line 5: Changes here provide allow overriding the value when invoking the CLIs. In the case of shopifolks the override is done by `dev` by exposing the variable declared in the `dev.yml`.

alvaro-shopify (approved):

amcaplan (approved):

PR #277, opened 2022-08-11
Define new hooks for tunnel plugins

isaacroldan:
In the process of migrating the ngrok tunnel to use hooks, we first need to define those hooks in cli-kit so that they can be the imported in any tunnel plugin.

Contributes to https://github.com/Shopify/shopify-cli-planning/issues/305

Add two new typed hooks:
- `tunnel_provider` for discoverability. Any plugin implementing this hooks means it can handle tunnels.
- `tunnel_start_${provider}` for execution. After discovering the plugins, each one will report what hook to use to start a tunnel, this hook is unique per plugin.

This approach allow us to have multiple tunnel plugins installed and decide at runtime which one we want to use for tunnelling depending on user input/selection without having incompatibilities.

### How to test your changes?
Nothing to test yet

### DIFF
--- a/packages/cli-kit/src/plugins.ts
+++ b/packages/cli-kit/src/plugins.ts
@@ -59,7 +59,7 @@ interface HookReturnsPerPlugin {
   tunnel_provider: {
     options: {[key: string]: never}
     pluginReturns: {
-      [pluginName: string]: Required<TunnelProvider>
+      [pluginName: string]: {name: string}
     }
   }
   public_command_metadata: {
@@ -89,3 +89,5 @@ export type FanoutHookFunction<
   this: Interfaces.Hook.Context,
   options: TPluginMap[TEvent]['options'] & {config: Interfaces.Config},
 ) => Promise<PluginReturnsForHook<TEvent, TPluginName, TPluginMap>>
+
+export const defineProvider = (input: {name: string}) => input
--- a/packages/cli-kit/src/plugins.ts
+++ b/packages/cli-kit/src/plugins.ts
@@ -3,6 +3,7 @@ import {debug, content} from './output.js'
 import {JsonMap} from './json.js'
 import {PickByPrefix} from './typing/pick-by-prefix.js'
 import {MonorailEventPublic} from './monorail.js'
+import {HookReturnPerTunnelPlugin} from './plugins/tunnel.js'
 import {Interfaces} from '@oclif/core'
 
 const TUNNEL_PLUGINS = ['@shopify/plugin-ngrok']
@@ -57,7 +58,7 @@ interface HookReturnsPerPlugin extends HookReturnPerTunnelPlugin {
   }
 }
 
-type PluginReturnsForHook<
+export type PluginReturnsForHook<
   TEvent extends keyof TPluginMap,
   TPluginName extends keyof TPluginMap[TEvent]['pluginReturns'],
   TPluginMap extends HookReturnsPerPlugin = HookReturnsPerPlugin,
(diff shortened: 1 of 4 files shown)


pepicrft (approved):

packages/cli-kit/src/plugins.ts
```diff
@@ -70,4 +88,4 @@ export type FanoutHookFunction<
 > = (
   this: Interfaces.Hook.Context,
   options: TPluginMap[TEvent]['options'] & {config: Interfaces.Config},
-) => Promise<Partial<PluginReturnsForHook<TEvent, TPluginName, TPluginMap>>>
+) => Promise<PluginReturnsForHook<TEvent, TPluginName, TPluginMap>>
```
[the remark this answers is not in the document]
shauns: I think that change is fine 👍  -- the partials in this area are mainly meant to be about not relying on a plug-in being installed.

packages/cli-kit/src/plugins.ts
```diff
@@ -70,4 +88,14 @@ export type FanoutHookFunction<
...
+
+export const tunnel = {
+  defineProvider: (input: {name: string}) => () => input,
+  startTunnel: (options: {provider: string; action: (port: number) => Promise<string | undefined>}) => {
+    return (inputs: {port: number; provider: string}) => {
+      if (inputs.provider !== options.provider) return undefined
+      return options.action(inputs.port)
+    }
+  },
```
isaacroldan, line 100: By having this middleware we can control which plugins are actually called.

Each plugin defines a provider and an action to be executed, then when the hook is called the action will be run only if the defined provider is the same as the intended provider (sent via runHook).

The types between this functions and `HookReturnsPerPlugin` are not the same, I'd need to find a way to tie them.

isaacroldan:
@pepicrft @shauns After @pepicrft feedback on the ngrok plugin I added specific tunneling interfaces that makes it easier to create a tunnel plugin.

The changes are:
- Both hooks are now generic, `tunnel_start_{provider}` is no `tunnel_start`, all will be called, but only the ones that match the intended provider will be actually executed.
- Each plugin will need to implement these two hooks like this:

```ts
// provider.ts
export const TUNNEL_PROVIDER = 'ngrok'
export default plugins.tunnel.defineProvider({name: TUNNEL_PROVIDER})
```

```ts
// tunnel.ts
export default plugins.tunnel.startTunnel({ 
  provider: TUNNEL_PROVIDER, 
  action: async (port: number) => start({port})
})

async function start(.... // this handles the tunnel creating and returning the URL.
```

You can see examples in the updated [PR for the ngrok plugin](https://github.com/Shopify/awesome-shopify-cli-plugins/pull/31/files)


What do you think @shauns @pepicrft ? Do you like this approach more than the previous proposal? for reference:
```ts
const hook: plugins.FanoutHookFunction<'tunnel_provider', 'ngrok'> = async function () {
  return {hook: 'tunnel_start_ngrok', provider: 'ngrok'}
}
```

pepicrft:
> What do you think @shauns @pepicrft ? Do you like this approach more than the previous proposal? for reference:

I like how idiomatic the new API is.

pepicrft (approved):

packages/cli-kit/src/plugins.ts
```diff
@@ -5,12 +5,18 @@ import {PickByPrefix} from './typing/pick-by-prefix.js'
 import {MonorailEventPublic} from './monorail.js'
 import {Interfaces} from '@oclif/core'
 
+export type TunnelHook = `tunnel_start_${string}`
```
pepicrft, line 8: This `plugins.ts` is a good candidate to grow into a huge module. As soon as we see this one growing too much we should consider breaking it down into smaller and more manageable modules.

pepicrft (approved):

matteodepalo (approved):

PR #278, opened 2022-08-11
[Feature] adding blockProgress capability for checkout ui extensions

cpeddecord:

Resolves https://github.com/Shopify/checkout-web/issues/12795, we've added a new [`block_progress`](https://vault.shopify.io/gsd/projects/29745/decisions/484) flag to our checkout ui extension configuration, this allows developers to state that their extension can block checkout progress and will allow merchants to grant or deny this capability from within the checkout editor.

Like `network_access`, when running an extension locally, this capability needs to be detected from the TOML file and passed as part of the extension data.

This capability is documented in this [PR](https://github.com/Shopify/shopify-dev/pull/25455).

### How to test your changes?

* create a new Checkout UI Extension or use an existing one
* add a `block_progress` flag under `capabilities` within the configuration TOML file
* booleans should persist to the JSON being served

### DIFF
--- a/packages/ui-extensions-go-cli/api/api_test.go
+++ b/packages/ui-extensions-go-cli/api/api_test.go
@@ -806,7 +806,8 @@ func TestWebsocketClientDispatchEventWithoutMutatingData(t *testing.T) {
 			"localization": null,
 			"surface": "checkout",
 			"capabilities": {
-				"networkAccess": false
+				"networkAccess": false,
+				"blockProgress": false
 			}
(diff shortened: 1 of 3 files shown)


jamesvidler (approved):
Nice work 👍 

I had trouble tophatting this though. Couldn't get a scaffolded extension to use this patch. I'm not worried about this, since its a straight-forward change though.

alfonso-noriega (approved):

PR #279, opened 2022-08-11
Add settings example in checkout ui extension template

alxclark:

1:1 port from https://github.com/Shopify/shopify-cli-extensions/pull/381


ginmrt (approved):
🚀

PR #287, opened 2022-08-12
Refactor `shopify hydrogen deploy` into separate folders

pepicrft:
I merged [this PR](https://github.com/Shopify/cli/pull/285) by mistake so I'm re-opening the PR


packages/cli-hydrogen/src/cli/commands/hydrogen/deploy.ts
```diff
@@ -0,0 +1,43 @@
...
+
+export default class Deploy extends Command {
+  static description = 'Deploy your Hydrogen app to Oxygen'
+
+  // note: we may want to check for the environment variable OXYGEN_DEPLOYMENT_TOKEN as a fallback when
+  // we are in a workflow
+  static flags = {
+    ...cli.globalFlags,
+    ...hydrogenFlags,
+    deploymentToken: Flags.string({
```
pepicrft, line 15: @grahamscott you might want to review the content here.

packages/cli-hydrogen/src/cli/commands/hydrogen/deploy.ts
```diff
@@ -0,0 +1,43 @@
+import {hydrogenFlags} from '../../flags.js'
+import {deployToOxygen} from '../../services/deploy.js'
+import {cli} from '@shopify/cli-kit'
+import Command from '@shopify/cli-kit/node/base-command'
+import {Flags} from '@oclif/core'
+
+export default class Deploy extends Command {
```
pepicrft, line 7: cc @MeredithCastile for alignment with the app side of things

packages/cli-hydrogen/src/cli/services/deploy/config.ts
```diff
@@ -0,0 +1,59 @@
+import {DeployConfig, ReqDeployConfig} from './types.js'
```
pepicrft, line 1: This module is lacking tests. I'd add an integration test that ensures we are getting the right configuration or errors given the state of the repository:
- Fails if git is not found
- Fails if it's not inside a git directory
- Fails if it can't get the latest commit, or repository, or head ref.

Moreover, note that `git.ensureInsideGitDirectory()` uses the current working directory, which might be different from the Hydrogen's directory if you invoke the command using the `--path`  flag. I'd update it to be something along the lines of:

```ts
const directory = config.path ? path.resolve(config.path) : process.cwd()
await git.ensureInsideGitDirectory({workingDirectory: directory})
```

packages/cli-hydrogen/src/cli/services/deploy/config.ts
```diff
@@ -0,0 +1,59 @@
...
+    commitAuthor: config.commitAuthor ?? latestCommit.author_name,
+    commitSha: latestCommit.hash,
+    commitRef,
+    timestamp: latestCommit.date,
+    repository,
+    path: directory,
+  }
+}
+
+const getLatestCommit = async () => {
```
pepicrft, line 27: All these internal utility functions are useful beyond the Hydrogen domains o I recommend moving it down to the `git` module inside `@shopify/cli-kit`.

packages/cli-kit/src/git.ts
```diff
@@ -69,3 +73,13 @@ export async function ensurePresentOrAbort() {
     throw GitNotPresentError()
   }
 }
+
+/**
+ * If command run from outside a .git directory tree
+ * it throws an abort error.
+ */
+export async function ensureInsideGitDirectory() {
```
pepicrft, line 81: I'd extend this API to support passing a working directory:

```ts
type EnsureInsideGitDirectoryOptions = {
  workingDirectory: string
}
export async function ensureInsideGitDirectory(options: EnsureInsideGitDirectoryOptions) {

} 
```

packages/cli-kit/src/git.ts
```diff
@@ -12,6 +12,10 @@ export const GitNotPresentError = () => {
   )
 }
 
+export const OutsideGitDirectoryError = () => {
```
pepicrft, line 15: I'd include in the error the directory tied to this error. @MeredithCastile what do you think about the error message?

packages/cli-kit/src/git.ts
pepicrft, line 81: I'd add an integration test here

packages/cli-kit/package.json
```diff
@@ -124,6 +124,6 @@
     "@types/js-yaml": "^4.0.5",
     "@types/semver": "^7.3.9",
     "node-stream-zip": "^1.15.0",
-    "vitest": "^0.17.1"
+    "vitest": "^0.21.1"
```
pepicrft, line 127: I'd do the version bump in a different PR and make sure it's done across all the packages in the repository. Otherwise we'll end up with inconsistencies.

packages/cli-kit/src/api.ts
```diff
@@ -2,5 +2,6 @@ import * as admin from './api/admin.js'
 import * as partners from './api/partners.js'
 import * as graphql from './api/graphql/index.js'
 import * as identity from './api/identity.js'
+import * as common from './api/common.js'
```
pepicrft, line 5: I'd not export the entire module because it was designed to be internal. What about only exporting the function that you are using, `buildHeaders`?

packages/cli-kit/src/git.test.ts
```diff
@@ -73,7 +74,7 @@ describe('ensurePresentOrAbort', () => {
     vi.mocked(hasGit).mockResolvedValue(false)
 
     // Then
-    await expect(() => ensurePresentOrAbort()).rejects.toThrowError(GitNotPresentError())
```
pepicrft, line 76: @matteodepalo shouldn't this one be `() => ...`?

packages/cli-kit/src/git.test.ts
```diff
@@ -84,3 +85,21 @@ describe('ensurePresentOrAbort', () => {
     await expect(ensurePresentOrAbort()).resolves.toBeUndefined()
   })
 })
+
+describe('ensureInsideGitDirectory', () => {
```
pepicrft, line 89: This one is a great candidate to be an integration test in a temporary directory. I'd recommend creating a `git.integration.test.ts` module and move this piece there as an integration test that does:

1. I create a temporary directory
2. I call `ensureInsideGitDirectory`
3. I expect an error

packages/cli-hydrogen/src/cli/services/deploy.ts
```diff
@@ -0,0 +1,20 @@
...
+export async function deployToOxygen(_config: DeployConfig) {
+  const config = await getDeployConfig(_config)
+  // eslint-disable-next-line no-console
+  console.log('Deployment Config: ', config)
+
+  const {deploymentID, assetBaseURL, error} = await createDeploymentStep(config)
+
+  output.info(`Deployment ID: ${deploymentID}`)
+  output.info(`Base Asset URL: ${assetBaseURL}`)
+  output.info(`Error Message: ${error?.debugInfo}`)
```
pepicrft, line 15: If the deployment failed I recommend mapping thee error to an `Abort` error and throwing it. That way the error will be handled and formatted consistently with the rest of teh codebase.

packages/cli-hydrogen/src/cli/services/deploy.ts
```diff
@@ -0,0 +1,20 @@
+import {DeployConfig} from './deploy/types.js'
+import {getDeployConfig} from './deploy/config.js'
+import {createDeploymentStep, runBuildCommandStep} from './deploy/deployer.js'
+import {output} from '@shopify/cli-kit'
+
+export async function deployToOxygen(_config: DeployConfig) {
+  const config = await getDeployConfig(_config)
+  // eslint-disable-next-line no-console
```
pepicrft, line 8: `console.log`'s are discouraged and also disabling the rule. If you want to output something for the user use `output.info` or `output.debug` if you want to make it opt-in via the `--verbose` flag.

graygilmore (commented):
Should this PR be targeting `add-hydrogen-deploy-command` instead of `main` like the other PR?

packages/cli-hydrogen/src/cli/services/deploy/deployer.ts
```diff
@@ -0,0 +1,59 @@
...
+  // need to make workflowID optional on DMS so we dont need to generate a random one
+  const variables = {
+    input: {
+      repository: config.repository,
+      branch: config.commitRef,
+      commitHash: config.commitSha,
+      commitAuthor: config.commitAuthor,
+      commitMessage: config.commitMessage,
+      commitTimestamp: config.timestamp,
+      workflowID: `${Math.floor(Math.random() * 100000)}`,
```
jgodson, line 31: This is no longer required as of this morning, so should be able to remove it.

jgodson (commented):
This is looking good to be me in general. I left a couple comments, but I'll mostly defer to the shopify-cli experts for a more thorough review.

tempor1s (changes requested):
Yeah - This refactor was meant to be merged into `add-hydrogen-deploy-command` as its still very much a work in progress (missing tests, doesn't have the full happy path yet, using console.logs, extra debug logging, etc etc) but we were just doing separate PR's for that feature branch so that we were able to review as we went. Should having a working path that we could look into merging into main next week. Thanks for the feedback though, we will take it all into account for building out the feature branch!

tempor1s:
I re-added the changes of https://github.com/Shopify/cli/pull/285 into our `add-hydrogen-deploy-command` feature branch and we will continue to work off of that. Going to close this PR since we do not want to merge into main, and once we have a happy path of this command we will open a separate PR that we will merge into main.

PR #290, opened 2022-08-12
Use ngrok plugin with hooks

isaacroldan:
As part of the process to start using the plugin system, we migrated the ngrok plugin to use the new tunnel hooks.
This PR finishes that work.

- Update CLI <> Ngrok plugin communication to use hooks

### How to test your changes?
- `dev` with a tunnel and ngrok should still work :)


packages/app/src/cli/services/dev/urls.ts
```diff
@@ -1,18 +1,21 @@
...
-  const tunnelPlugin = await plugins.lookupTunnelPlugin(pluginList)
-  if (!tunnelPlugin) throw new error.Bug('The tunnel could not be found')
-  const url = await tunnelPlugin?.start({port: frontendPort})
+export async function generateURL(config: Config, frontendPort: number): Promise<string> {
+  // For the moment we assume to always have ngrok, this will change in a future PR
+  // and will need to use "getListOfTunnelPlugins" to find the available tunnel plugins
+  const result = await plugins.runTunnelPlugin(config, frontendPort, 'ngrok')
+
+  if (result.error === 'multiple-urls') throw new error.Bug('Multiple tunnel plugins for ngrok found')
+  if (result.error === 'no-urls' || !result.url) throw new error.Bug('Ngrok failed to start the tunnel')
```
pepicrft, line 16: Are those bugs?

isaacroldan: Right now they are, because in the current release they should never happen.

Once we add more tunnel plugins and ngrok is not added by default, these won't be bugs

PR #299, opened 2022-08-16
Fix building and testing the Go CLI from a clean environment

pepicrft:
@gonzaloriestra found out that `build` and `test` of the Go CLI fails when the environment is clean. It turns out the Go CLI embeds files that are output by [the dev console](https://github.com/Shopify/cli/blob/main/packages/ui-extensions-dev-console/vite.config.ts#L9) project when building it. Because of it, it's important that we build dev-console before running `build` or `test` in the Go CLI.

I'm adjusting the `test` and `build` scripts in the Go CLI package to run through Nx. Since Nx is aware of the dependency graph, it'll make sure the dependent projects are built first.

### How to test your changes?
1. `yarn clean` at the root.
2. `cd packages/ui-extensions-go-cli`
3. `yarn build` should work 🎉


.github/workflows/shopify-cli.yml
```diff
@@ -60,7 +60,7 @@ jobs:
       - name: Install dependencies
         run: yarn install --ignore-engines
       - name: Build the fixture app
-        run: yarn shopify app build --path ./fixtures/app --skip-dependencies-installation
+        run: yarn shopify app build --path ./fixtures/app
```
pepicrft, line 63: The fixtures are part of the workspace in this repository and therefore we don't need to install dependencies at runtime. I've created an empty `yarn.lock` file in the fixture's directory for the CLI to detect this project as a Yarn project. That results to a noop `yarn install` operation.

packages/ui-extensions-go-cli/project.json
```diff
@@ -6,6 +6,13 @@
     "implicitDependencies": ["ui-extensions-cli"],
     "tags": ["scope:feature", "scope:ui-extensions"],
     "targets": {
+      "build": {
+        "executor": "nx:run-script",
+        "dependsOn": ["^build"],
```
pepicrft, line 11: Building the Go CLI requires building the Dev console. The `^build` value for `dependsOn` means "run build in all my direct and transitive dependencies".

packages/ui-extensions-go-cli/package.json
```diff
@@ -3,15 +3,13 @@
...
-    "build": "node bin/build.js",
-    "clean": "cd .",
+    "build": "yarn clean && yarn --cwd ../.. nx run ui-extensions-go-cli:build",
+    "build:nx": "node bin/build.js",
+    "clean": "rimraf api/dev-console/",
     "start": "cd packages/dev-console-app; yarn start",
-    "test": "yarn --cwd ../.. nx run ui-extensions-go-cli:test",
-    "test:nx": "node bin/test.js",
+    "test": "yarn clean && yarn --cwd ../.. nx run ui-extensions-go-cli:test",
+    "test:nx": "go test ./...",
```
gonzaloriestra, line 12: ❤️

gonzaloriestra (approved):
Working like a charm ✨

PR #307, opened 2022-08-17
Reuse cache when running `info --web-env`

matteodepalo:

While working on a task related to `app info --web-env` I noticed that I was asked to select the app every time.

I propose that we can reuse the cached information from `dev` if it's present when choosing the app.

### How to test your changes?

- Run `dev fixture dev` to completion
- Run `dev fixture shopify app info --web-env`
- You shouldn't be asked to choose an app

### DIFF
--- a/packages/app/src/cli/services/info.test.ts
+++ b/packages/app/src/cli/services/info.test.ts
@@ -1,5 +1,5 @@
 import {info} from './info.js'
-import {fetchOrgAndApps, fetchOrganizations} from './dev/fetch.js'
+import {fetchAppFromApiKey, fetchOrgAndApps, fetchOrganizations} from './dev/fetch.js'
 import {selectOrCreateApp} from './dev/select-app.js'
 import {AppInterface} from '../models/app/app.js'
 import {selectOrganizationPrompt} from '../prompts/dev.js'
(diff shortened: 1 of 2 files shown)


matteodepalo:
> Shorter term: is this "automagical"? Any concerns around selecting the app for them? (My ability to parse code is still limited.)

I was also thinking about what's the path to reset this. Perhaps it's worth adding a `reset` flag to escape the cache if needed?

gonzaloriestra (approved):
LGTM!

My only concern is that this command is used to get the variables for the deployment, but we are reusing the cache of the dev command. If some people deploy to a different app, it could be annoying to require a reset here... But I'm not sure about the common user flows 🤷 

Also, I guess this is a command you typically use just once to set up your CI/CD env, so having to select an app is not a big deal actually...

Btw, @shainaraskas I've realized we are not suggesting to use this command (`shopify app info --web-env`) in the step to retrieve the API key and secret: https://shopify.dev/apps/tools/cli/cli-2/ci-cd#step-2-retrieve-your-api-key-and-secret. What about adding it?

matteodepalo:
Thinking more about this problem, this command might be too magical to the user and, as you pointed out @gonzaloriestra, the `dev` command is not exactly connected with the use case of this command. I'm going to close this PR for now, but I might come back to it once I review the `deploy` code better.

shauns:
> realized we are not suggesting to use this command

We do in the instructions for things like deploying to fly.io

> the dev command is not exactly connected with the use case of this command

Yup - it's for deployment

PR #308, opened 2022-08-17
Call develop.sh using bash directly

dmhenry:

Fixes #0000


alfonso-noriega (commented):
I see that we are still skipping some tests [here](https://github.com/Shopify/cli/blob/windows_tests_breakage_fixes/packages/ui-extensions-server-kit/src/context/ExtensionServerProvider.test.tsx#L141) and [here](https://github.com/Shopify/cli/blob/windows_tests_breakage_fixes/packages/ui-extensions-server-kit/src/state/reducers/extensionServerReducer.test.ts#L71). Should we remove this before merging to ensure everything is passing?

dmhenry:
@alfonso-noriega I tried re-enabling those tests and they failed under Node 14. Since this PR already covers re-enabling them, I decided to just skip it in the interest of fixing the few issues we'd identified in this PR.

PR #309, opened 2022-08-17
WIP: web-env command

matteodepalo:

Fixes #0000


packages/app/src/cli/services/web-env.ts
```diff
@@ -0,0 +1,101 @@
...
+
+  const newEnvFileContent = patchEnvFile(envFileContent, updatedValues)
+  await file.write(envFile, newEnvFileContent)
+
+  return output.content`Updated ${envFile} to be:
+${newEnvFileContent}
+`
+}
+
+export async function outputEnv(app: AppInterface): Promise<output.Message> {
```
matteodepalo, line 51: I might refactor this to accept a `format` param so that I can reuse the same logic in the `info --web-dev` command.

pepicrft (commented):
👏🏼 Amazing work on this one. I'd add a changesets update.

packages/app/src/cli/services/dev/select-app.ts
```diff
@@ -34,6 +40,15 @@ export async function selectOrCreateApp(
   return app
 }
 
+export async function selectApp(): Promise<OrganizationApp> {
```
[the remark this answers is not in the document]
matteodepalo: Good point, it's not really just for dev anymore.

packages/app/src/cli/services/web-env.ts
```diff
@@ -0,0 +1,101 @@
...
+  const selectedApp = await selectApp()
+
+  return output.content`
+  ${output.token.green('SHOPIFY_API_KEY')}=${selectedApp.apiKey}
+  ${output.token.green('SHOPIFY_API_SECRET')}=${selectedApp.apiSecretKeys[0].secret}
+  ${output.token.green('SCOPES')}=${app.configuration.scopes}
+`
+}
+
+function patchEnvFile(envFileContent: string | null, updatedValues: {[key: string]: string}): string {
```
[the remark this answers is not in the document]
matteodepalo: That makes sense. I was a bit hesitant at first because in that module we use an external library to parse the env file and in this function we use a custom approach. But the two use cases should be clear from the naming.

packages/app/src/cli/services/web-env.ts
matteodepalo: Yes, it would essentially be a redirect to the new command. Not sure what's our policy here but I can imagine people having already used the `--web-dev` command to produce their own `.env` files with a custom script.

packages/app/src/cli/services/web-env.ts
```diff
@@ -0,0 +1,101 @@
...
+        if (value === newValue) {
+          outputLines.push(line)
+        } else {
+          const newLine = toLine(key, newValue)
+          outputLines.push(newLine)
+        }
+      }
+    } else {
+      outputLines.push(line)
+    }
```
gonzaloriestra, line 91: I find this piece of code a bit hard to follow. What about this alternative? I think it should be equivalent:
```suggestion
    let lineToWrite = line
    if (match) {
      const key = match[1].trim()
      const newValue = updatedValues[key]
      if (newValue) {
        alreadyPresentKeys.push(key)
        lineToWrite = toLine(key, newValue)
      }
    }
    outputLines.push(lineToWrite)
```

gonzaloriestra (commented):
Looks good! We will need to add this new command to the [documentation](https://shopify.dev/apps/tools/cli/commands) as well. cc: @shainaraskas

packages/app/src/cli/services/web-env.ts
matteodepalo: Nice, not caring about the fact that `value` has changed or not is much better 👍

packages/cli-kit/src/output.ts
```diff
@@ -141,6 +143,10 @@ export const token = {
   failIcon: () => {
     return new ContentToken('✖', {}, ContentTokenType.ErrorText)
   },
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  linesDiff: (value: any) => {
```
matteodepalo, line 147: I'm not very happy with using `any` here, but I've tried to make a generic type work without much success. I might take another stab at this later.

packages/cli-kit/src/output.ts
```diff
@@ -228,6 +234,17 @@ export function content(strings: TemplateStringsArray, ...keys: (ContentToken |
         case ContentTokenType.Green:
           output += colors.green(stringifyMessage(enumToken.value))
           break
+        case ContentTokenType.LinesDiff:
+          ;(enumToken.value as unknown as Change[]).forEach((part) => {
```
pepicrft, line 238: I think `prettier` got confused here.

matteodepalo: Yeah, but to be honest this casting shouldn't be necessary, but given the way `...keys` is typed it's hard to make this generic. Maybe we should disable prettier for this line.

packages/app/src/cli/commands/app/web-env.ts
```diff
@@ -0,0 +1,37 @@
+import {appFlags} from '../../flags.js'
+import {AppInterface} from '../../models/app/app.js'
+import {load as loadApp} from '../../models/app/loader.js'
+import {webEnv} from '../../services/web-env.js'
+import {Flags} from '@oclif/core'
+import {output, path, cli} from '@shopify/cli-kit'
+import Command from '@shopify/cli-kit/node/base-command'
+
+export default class WebEnv extends Command {
```
shauns, line 9: This command should be hidden. I think `static hidden = true` will do the job. If not hidden then we have to get the naming etc. right!

packages/app/src/cli/services/web-dev.test.ts
```diff
@@ -0,0 +1,235 @@
+import {fetchOrgAndApps, fetchOrganizations} from './dev/fetch.js'
```
shauns: This file is meant to be `web-env.test.ts`

packages/app/src/cli/services/web-env.ts
```diff
@@ -0,0 +1,77 @@
...
+    return outputEnv(app, 'text')
+  }
+}
+
+export async function updateEnvFile(app: AppInterface, envFile: WebEnvOptions['envFile']): Promise<output.Message> {
+  const selectedApp = await selectApp()
+
+  const updatedValues = {
+    SHOPIFY_API_KEY: selectedApp.apiKey,
+    SHOPIFY_API_SECRET: selectedApp.apiSecretKeys[0].secret,
```
shauns, line 26: I think it's possible for this to be empty -- if you look in `ensureDevEnvironment` you'll see similar logic that checks this

packages/app/src/cli/services/web-env.ts
```diff
@@ -0,0 +1,77 @@
...
+    SCOPES: app.configuration.scopes,
+  }
+
+  if (await file.exists(envFile)) {
```
shauns, line 30: I think all this file manipulation would need to be relative to the App's root folder (it can be passed in via `--path` or the command could be run from a sub-folder)

packages/cli-kit/src/node/dot-env.ts
```diff
@@ -50,3 +50,42 @@ export async function readAndParseDotEnv(path: string): Promise<DotEnvFile> {
...
   await writeFile(file.path, stringify(file.variables))
 }
+
+/**
+ * Given an .env file content, generates a new one with new values
+ * without removing already existing lines.
+ * @param envFileContent {string | null} .env file contents.
+ * @param updatedValues {[key: string]: string}} object containing new env variables values.
+ */
+export function patchEnvFile(envFileContent: string | null, updatedValues: {[key: string]: string}): string {
```
shauns, line 60: Probably need to test all this with Windows `\r\n` style line breaks too (We should cope with them, and we should preserve them)

packages/cli-kit/src/output.ts
shauns: Yeah the whole `class ContentToken` needs some attention and would probably be better served as a discriminated union

shauns (changes requested):

packages/app/src/cli/services/web-env.ts
matteodepalo: Good catch, I'll change it.

packages/cli-kit/src/node/dot-env.ts
matteodepalo: @shauns does the last commit test what you were pointing out here correctly?

matteodepalo:
@shauns I've refactored content tokens completely to inherit from a common abstract class. Let me know if that makes sense.

PR #311, opened 2022-08-17
Fixing a few things...

lynchv:
Doing a few things in this PR:

- Adding an integration test for the config service testing with real git commands.
- Added a new function to the cli-kit git called `commitAll` and tested it. Needed this for the integration test and will reuse later when initializing the repository!
- Adding an early return to the `deploy` command if user is not inside Shopify env.
- Took the decision to entirely remove `temporarelyChangeCWD` and instead have the option to pass a directory to each git function. This seemed to be the proper way to move forward when looking at existing code.
- Making sure we only export `buildHeaders` from the api cli-kit and not the entire `common.ts` file.
- removing the usage of both `repository` and `workflowId`

Note: Sorry for 1 commit, got to get use to not `--amend` my commit...


tempor1s (approved):
🔥 Looks good to me outside of Gray's comments. We are probably going to need to rebase this once https://github.com/Shopify/cli/pull/310 merges due to the changes in the `api.buildHeaders` export.

lynchv:
### Force Push notes

- Rebase with feature branch to resolve conflict
- Removed unused code

PR #315, opened 2022-08-18
Added simple health check with incremental rollback

lynchv:
Fix https://github.com/Shopify/oxygen-platform/issues/494

Simply added a health check step with a new flag called `skipHealthCheck` which defaults to false.

### DIFF
--- a/packages/cli-hydrogen/src/cli/commands/hydrogen/deploy.ts
+++ b/packages/cli-hydrogen/src/cli/commands/hydrogen/deploy.ts
@@ -34,6 +34,11 @@ export default class Deploy extends Command {
       env: 'SHOPIFY_HYDROGEN_FLAG_DMS_ADDRESS',
       default: 'https://oxygen-dms.shopifycloud.com',
     }),
+    healthCheck: Flags.boolean({
+      env: 'SHOPIFY_HYDROGEN_FLAG_HEALTH_CHECK',
+      default: true,
+      description: 'Require a health check before the deployment succeeds.',
+    }),
   }
 
   public async run(): Promise<void> {
--- a/packages/cli-hydrogen/src/cli/services/deploy.ts
+++ b/packages/cli-hydrogen/src/cli/services/deploy.ts
@@ -1,6 +1,6 @@
 import {DeployConfig} from './deploy/types.js'
 import {getDeployConfig} from './deploy/config.js'
-import {createDeploymentStep, runBuildCommandStep, uploadDeploymentStep} from './deploy/deployer.js'
+import {createDeploymentStep, runBuildCommandStep, uploadDeploymentStep, healthCheck} from './deploy/deployer.js'
 import {output} from '@shopify/cli-kit'
 
 export async function deployToOxygen(_config: DeployConfig) {
@@ -16,9 +16,12 @@ export async function deployToOxygen(_config: DeployConfig) {
 
   // note: need to handle this error
   const buildResponse = await runBuildCommandStep(config, assetBaseURL)
-  const deployResponse = await uploadDeploymentStep(config, deploymentID)
+  const previewURL = await uploadDeploymentStep(config, deploymentID)
 
-  output.success('Deployed and healthy!')
-  output.info(`• Preview URL: ${deployResponse}`)
-  // note: should try output URL after the "success" message for scripting purposes like oxygenctl does right now
+  output.info(`Preview URL: ${previewURL}`)
+
+  await healthCheck(previewURL)
+
+  // For scripting purposes, last line is preview URL.
+  output.info(previewURL)
 }
--- a/packages/cli-hydrogen/src/cli/services/deploy/deployer.ts
+++ b/packages/cli-hydrogen/src/cli/services/deploy/deployer.ts
@@ -4,6 +4,7 @@ import {
   CreateDeploymentQuerySchema,
   CreateDeploymentQuery,
 } from './graphql/create_deployment.js'
+import {retryOnError, WebPageNotAvailable} from './error.js'
 import buildService from '../build.js'
 import {output, api, http} from '@shopify/cli-kit'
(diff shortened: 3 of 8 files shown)


packages/cli-hydrogen/src/cli/services/deploy.ts
```diff
@@ -16,9 +16,19 @@ export async function deployToOxygen(_config: DeployConfig) {
...
+    const isHealthy = await healthCheck(previewURL)
+    if (!isHealthy) {
+      output.info("🤕 The deployment was uploaded but can't be reached yet.")
+      return
+    }
+    output.success('✅ Deployed and healthy!')
+  }
+
+  // For scripting purposes, last line is preview URL.
+  output.info(previewURL)
```
tempor1s, line 33: Does this show up after the final `✨  Done in 2.20s.` message? When I was testing previously it did not IIRC

packages/cli-hydrogen/src/cli/commands/hydrogen/deploy.ts
```diff
@@ -34,6 +34,11 @@ export default class Deploy extends Command {
       env: 'SHOPIFY_HYDROGEN_FLAG_DMS_ADDRESS',
       default: 'https://oxygen-dms.shopifycloud.com',
     }),
+    skipHealthCheck: Flags.boolean({
```
tempor1s, line 37: In `oxygenctl` I think we just call this flag `healthCheck` and then [default it to false](https://github.com/Shopify/oxygenctl/blob/5d5d30668fbc00de9faebe672600ce6458141a62/cmd/oxygenctl/deploy.go#L121) so I think we should follow that behavior just for now.

packages/cli-hydrogen/src/cli/commands/hydrogen/deploy.ts
```diff
@@ -34,6 +34,11 @@ export default class Deploy extends Command {
...
+      env: 'SHOPIFY_HYDROGEN_FLAG_SKIP_HEALTH_CHECK',
+      default: false,
+      description: 'Exit directly after upload without waiting for webpage to be available',
```
tempor1s, line 40: If you go with my above change, we can just change the copy of this description to also match what is currently on `oxygenctl`. Obviously Graham will be looking over all of this so not a big deal.

`Require a health check before the deployment succeeds.`

packages/cli-hydrogen/src/cli/services/deploy/deployer.ts
```diff
@@ -92,6 +92,22 @@ export const uploadDeploymentStep = async (config: ReqDeployConfig, deploymentID
...
+    /* eslint-disable no-await-in-loop */
+    const result = await http.fetch(pingUrl, {method: 'GET'})
+    if (result.status === 200) return true
+    await system.sleep(seconds)
+    /* eslint-enable no-await-in-loop */
+  }
+
+  return false
+}
+
```
tempor1s, line 110: Do we want to extract this "retry" logic into a separate method that we can for our other methods? As part of the "refactor" I think all of our fetch / graphql calls are going to need to be able to retry so might be worth making it generalized.

cartogram (approved):

packages/cli-hydrogen/src/cli/services/deploy/error.ts
```diff
@@ -0,0 +1,24 @@
...
+      return await execute()
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    } catch (error: any) {
+      // eslint-disable-next-line no-await-in-loop
+      await system.sleep(typeof waitTime === 'number' ? waitTime : waitTime(iteration))
+
+      if (iteration + 1 === maxIteration) throw new error.Abort(error.message)
+    }
+  }
+}
```
lynchv, line 24: @tempor1s Created a new file called `deploy/error.ts` where we can put all our error related stuff there. Made sense to me for the work still ahead.

packages/cli-hydrogen/src/cli/services/deploy.ts
```diff
@@ -16,9 +16,12 @@ export async function deployToOxygen(_config: DeployConfig) {
...
+  const previewURL = await uploadDeploymentStep(config, deploymentID)
 
-  output.success('Deployed and healthy!')
-  output.info(`• Preview URL: ${deployResponse}`)
-  // note: should try output URL after the "success" message for scripting purposes like oxygenctl does right now
+  output.info(`Preview URL: ${previewURL}`)
+
+  await healthCheck(previewURL)
+
+  // For scripting purposes, last line is preview URL.
```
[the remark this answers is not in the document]
lynchv: We would like to have the last line of our command be a specific output for scripting purposes but we seem to run into this line being printed last no matter what:

image

Just a reminder for now to address the situation

packages/cli-hydrogen/src/cli/services/deploy/deployer.ts
```diff
@@ -92,6 +93,31 @@ export const uploadDeploymentStep = async (config: ReqDeployConfig, deploymentID
...
+  const pingUrl = `${url}/__health`
+  const rollBack = [5, 10, 15, 30, 60]
+
+  try {
+    await retryOnError(
+      () => ping(pingUrl),
+      rollBack.length,
+      (iteration) => rollBack[iteration],
+    )
+    // eslint-disable-next-line no-catch-all/no-catch-all
```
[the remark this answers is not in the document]
graygilmore: 👍🏻 I forgot we were addressing error handling later anyways

packages/cli-hydrogen/src/cli/services/deploy.ts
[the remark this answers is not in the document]
gonzaloriestra: @lynchv That message is coming from Yarn. You can get rid of it by using npm or pnpm instead, or by using the `--silent` flag: `yarn --silent shopify version`

PR #320, opened 2022-08-19
Replace Yarn with PNPM

pepicrft:
[Yarn flakines](https://github.com/Shopify/cli/runs/7913938724?check_suite_focus=true)s on Windows environment while accessing its global cache led me to explore the alternative [pnpm](https://pnpm.io/) after reading that [the solution proposed by many users ](https://github.com/yarnpkg/yarn/issues/7212) comes down to limit the concurrency. Long-story-short, Yarn seems to have a rare-condition that's rooted in how it manages its cache, and by limiting the concurrency we are reducing the likelihood for it to happen. Unfortunately, limiting the concurrency has an impact on the installation time, which is undesirable considering the size of the workspace.

### PNPM

PNPM takes a different approach for managing its cache using a global store and sym and hard links. Because there are no copies, not only is faster but it creates the graph under `node_modules` more deterministically. I can't tell with 100% certainty that the above issue is gone, but considering the approach they take for managing the store, it'll most likely the case.
PNPM is also more strict which as it's well captured [in this blog post](https://www.kochan.io/nodejs/pnpms-strictness-helps-to-avoid-silly-bugs.html) leads to fewer bugs. As part of introducing the tool I could spot some dependencies that were missing. Despite their absence in `package.json`s and due to Yarn hoisting approach, tasks that depended on the dependency's presence worked accidentally.

Note that widely-used tools like [Vite](https://github.com/vitejs/vite), [Vitest](https://github.com/vitest-dev/vitest), and [Vue](https://github.com/vuejs/vue), have adopted it as their default package manager.


frandiox:
We'd like to do the same in Hydrogen https://github.com/Shopify/hydrogen/issues/848

pepicrft:
We've mitigated the issues that we came across so I'll go ahead and close this one. We can get back to it if we see there's a significant value in switching package managers.

PR #322, opened 2022-08-19
Fix the installation of dependencies on app creation in Windows environments

pepicrft:
Related: https://github.com/Shopify/cli/pull/320

Acceptance tests [uncovered](https://github.com/Shopify/cli/runs/7913938724?check_suite_focus=true) Yarn installation errors in Windows environments when it tries to concurrently access the global cache. 

I'm adjusting the app creation logic to limit the concurrency to one as recommended [here](https://github.com/yarnpkg/yarn/issues/7212) by other Yarn users. Some are experiencing it with newer versions of Yarn too so it seems that the issue is rooted in some core business logic of Yarn.
 
### How to test your changes?

Acceptance tests should pass on Windows

### DIFF
--- a/.changeset/red-seas-impress.md
+++ b/.changeset/red-seas-impress.md
@@ -1,0 +1,5 @@
+---
+'@shopify/create-app': minor
+---
+
+Limit the concurrency installing dependencies with Yarn on Windows to prevent race conditions from happening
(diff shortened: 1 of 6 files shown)


matteodepalo (approved):
👍 Nice

PR #323, opened 2022-08-19
Expand metadata sent to Bugsnag

shauns:

https://github.com/Shopify/shopify-cli-planning/issues/362

Captures metadata from installed plugins; uses the same environment data regular analytics uses

### How to test your changes?

Shared an internal video of how this appears. Can be recreated by triggering an error.

### DIFF
--- a/packages/cli-kit/src/analytics.ts
+++ b/packages/cli-kit/src/analytics.ts
@@ -1,6 +1,5 @@
 import * as environment from './environment.js'
 import {platformAndArch} from './os.js'
-import {resolve} from './path.js'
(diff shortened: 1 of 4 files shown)


packages/cli-kit/src/analytics.ts
```diff
@@ -77,16 +76,10 @@ const buildPayload = async ({config, errorMessage}: ReportEventOptions) => {
   const {startCommand, startArgs, startTime} = commandStartOptions
   const currentTime = new Date().getTime()
 
-  let directory = process.cwd()
```
shauns, line 80: we don't need this any more

PR #325, opened 2022-08-19
[BETA] Enable Theme Kit Access passwords

gonzaloriestra:

Second step for https://github.com/Shopify/shopify-cli-planning/issues/332
Fixes https://github.com/Shopify/shopify-cli-planning/issues/366

⚠️ We are still testing this functionality, it will be ready to use for everyone when we properly announce it and publish the documentation.

Enable Theme Kit Access passwords through env variable (`SHOPIFY_CLI_THEME_TOKEN`) or flag (`--password`) for theme commands, so that the login is not required. The store is also required (`SHOPIFY_FLAG_STORE` or `--store`).

### How to test your changes?

Check that theme commands work the same way when logged in:

- `shopify theme list`

Check that they also work with a Theme Kit Access password:

- Install [Theme Kit Access](https://apps.shopify.com/theme-kit-access) app and generate a new password
- `shopify auth logout`
- `shopify theme list --store your-store --password your-password`

The last command also allows env variables:
- `SHOPIFY_FLAG_STORE=your-store SHOPIFY_CLI_THEME_TOKEN=your-password shopify theme list`

### DIFF
--- a/packages/cli-kit/src/session.ts
+++ b/packages/cli-kit/src/session.ts
@@ -153,10 +153,12 @@ ${token.json(scopes)}
 
 /**
  * Ensure that we have a valid session to access the Theme API.
- * If SHOPIFY_CLI_THEME_TOKEN exists, that token will be used against Theme Access API.
+ * If a password is provided, that token will be used against Theme Access API.
  * Otherwise, it will ensure that the user is authenticated with the Admin API.
- * @param scopes {string[]} Optional array of extra scopes to authenticate with.
- * @returns {Promise<AdminSession>} The access token and store
+ * @param store - Store fqdn to request auth for
+ * @param password - Password generated from Theme Access app
+ * @param scopes - Optional array of extra scopes to authenticate with.
+ * @returns The access token and store
  */
 export async function ensureAuthenticatedThemes(
   store: string,
--- a/yarn.lock
+++ b/yarn.lock
@@ -12803,10 +12803,10 @@ vite-tsconfig-paths@^3.3.14:
     recrawl-sync "^2.0.3"
     tsconfig-paths "^4.0.0"
 
-vite@2.9.9, "vite@^2.9.12 || ^3.0.0-0", vite@^2.9.13:
-  version "2.9.9"
-  resolved "https://registry.yarnpkg.com/vite/-/vite-2.9.9.tgz#8b558987db5e60fedec2f4b003b73164cb081c5e"
-  integrity sha512-ffaam+NgHfbEmfw/Vuh6BHKKlI/XIAhxE5QSS7gFLIngxg171mg1P3a4LSRME0z2ZU1ScxoKzphkipcYwSD5Ew==
+vite@2.9.12, "vite@^2.9.12 || ^3.0.0-0", vite@^2.9.13:
+  version "2.9.12"
+  resolved "https://registry.yarnpkg.com/vite/-/vite-2.9.12.tgz#b1d636b0a8ac636afe9d83e3792d4895509a941b"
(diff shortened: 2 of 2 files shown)


packages/cli-kit/src/node/ruby.ts
```diff
@@ -9,7 +9,7 @@ import {AdminSession} from '../session.js'
 import {content, token} from '../output.js'
 import {Writable} from 'node:stream'
 
-const RubyCLIVersion = '2.20.1'
+const RubyCLIVersion = '2.23.0'
```
gonzaloriestra, line 12: I'm assuming this will be the version containing [this CLI2 PR](https://github.com/Shopify/shopify-cli/pull/2558), which is needed for this one.

packages/cli-kit/src/session.ts
```diff
@@ -150,6 +150,25 @@ ${token.json(scopes)}
...
+ * Ensure that we have a valid session to access the Theme API.
+ * If SHOPIFY_CLI_THEME_TOKEN exists, that token will be used against Theme Access API.
+ * Otherwise, it will ensure that the user is authenticated with the Admin API.
+ * @param scopes {string[]} Optional array of extra scopes to authenticate with.
+ * @returns {Promise<AdminSession>} The access token and store
+ */
+export async function ensureAuthenticatedThemes(
+  store: string,
+  password: string | undefined,
+  scopes: string[] = [],
```
pepicrft, line 163: Is there a set of default scopes that we might want to include here for themes?

packages/theme/src/cli/flags.ts
```diff
@@ -13,4 +13,15 @@ export const themeFlags = {
...
     default: '.',
   }),
+  password: Flags.string({
+    hidden: false,
+    description: 'Password generated from the Theme Access app.',
+    env: 'SHOPIFY_CLI_THEME_TOKEN',
+  }),
+  store: Flags.string({
+    char: 's',
+    description: 'Store URL.',
```
pepicrft, line 23: Is this the URL or the domain? What about including an example to prevent some confusion for the user?

packages/theme/src/cli/utilities/theme-store.ts
```diff
@@ -3,7 +3,10 @@ import {store as conf, error} from '@shopify/cli-kit'
 export function getTheme(flags: {store: string | undefined}): string {
   const store = flags.store || conf.cliKitStore().getTheme()
   if (!store) {
-    throw new error.Abort('A store is required', 'Specify the store using --store={your_store}')
+    throw new error.Abort(
+      'A store is required',
+      'Specify the store passing --store={your_store_url} or set the SHOPIFY_FLAG_STORE environment variable.',
```
pepicrft, line 8: Since we have the flags extracted in a module, what about reading the flag name and the environment variable from the flag?

pepicrft (approved):

packages/theme/src/cli/utilities/theme-store.ts
gonzaloriestra: Done!

packages/theme/src/cli/flags.ts
gonzaloriestra: Done!

packages/theme/src/cli/flags.ts
```diff
@@ -13,4 +13,17 @@ export const themeFlags = {
     env: 'SHOPIFY_FLAG_PATH',
     default: '.',
   }),
+  password: Flags.string({
+    hidden: false,
+    description: 'Password generated from the Theme Access app.',
+    env: 'SHOPIFY_CLI_THEME_TOKEN',
```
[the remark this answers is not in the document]
pepicrft: I'd be consistent with the flags convention because with presets flags will become the primitive for configuring workflows. `SHOPIFY_CLI_PARTNERS_TOKEN` should indeed have been a flag with its counterpart environment variable.

PR #326, opened 2022-08-19
Fix using the wrong package manager when installing dependencies

pepicrft:
The Go binary is still installing dependencies when creating new extensions even though that's a responsibility that the CLI has already. This led to dependency creation using a package manager other than the one used by the app.

I'm removing the installation of dependencies from the Go binary because it's not needed.

### How to test your changes?
Try to scaffold a new extension in the fixture app.


shauns (approved):

vividviolet (approved):
👍

PR #328, opened 2022-08-19
Add initial services and command file for init

### LINKED ISSUE #54: [Feature][Hydrogen]: Add `init` command

Hydrogen custom storefront

New feature

### Overview

This command would write a new hydrogen.config file to an existing project. It would prompt for items needed in the config, storefront token, storedomain and write the value to a new hydrogen.config file.

I think the [Form module](https://github.com/enquirer/enquirer#form-prompt) from enquirer is a good candidate for this.

form-prompt

### Motivation

I see this command as being table stakes. Allow users to easily write a hydrogen config without opening a code editor.

SherRao:
Fixes #54 - still a WIP, just wanted to get a draft PR up to see how it's progressing.

This PR will add a new subcommand to the hydrogen command in the CLI. Adds command-line args to the command, as well as a prompt for any un-passed args.

### How to test your changes?
Run the `hydrogen init` command

### DIFF
--- a/packages/cli-hydrogen/src/cli/commands/hydrogen/init.ts
+++ b/packages/cli-hydrogen/src/cli/commands/hydrogen/init.ts
@@ -1,0 +1,44 @@
+import {init} from '../../services/init.js'
+import {hydrogenFlags} from '../../flags.js'
+import {Flags} from '@oclif/core'
+import {cli, path} from '@shopify/cli-kit'
+import Command from '@shopify/cli-kit/node/base-command'
+
+export default class Build extends Command {
+  static description = 'Allows the quick initialisation of a Hydrogen storefront config file.'
+  static flags = {
+    ...cli.globalFlags,
+    ...hydrogenFlags,
+    country: Flags.string({
+      description: 'The default country code.',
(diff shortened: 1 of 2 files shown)


packages/cli-hydrogen/src/cli/commands/hydrogen/init.ts
```diff
@@ -0,0 +1,44 @@
...
+import {Flags} from '@oclif/core'
+import {cli, path} from '@shopify/cli-kit'
+import Command from '@shopify/cli-kit/node/base-command'
+
+export default class Build extends Command {
+  static description = 'Allows the quick initialisation of a Hydrogen storefront config file.'
+  static flags = {
+    ...cli.globalFlags,
+    ...hydrogenFlags,
+    country: Flags.string({
```
cartogram, line 12: Can we remove this flag as well as the `lang` one? let's scaffold the most minimal config possible.

https://github.com/Shopify/hydrogen/blob/v1.x-2022-07/templates/hello-world/hydrogen.config.ts#L3

SherRao: Done!

PR #331, opened 2022-08-22
Pin Node versions used on CI

pepicrft:
Our CI pipelines are configured to run with the highest minor version of Node 14, 16, and 18. This approach introduces a source of non-determinism in our CI builds which might lead to sudden failing builds (e.g. if a new Node version breaks something).

I'm pinning the main CI pipeline to the Node versions `14.20.0`, `16.17.0`, `18.7.0`. I'm also updating the `dev.yml` and `.nvmrc` files to use the last one from the list, `18.7.0`.

### How to test your changes?
CI should pass and you should be able to interact with the `fixtures/app` project from the local environment.


pepicrft:
> Do we prefer deterministic CI runs instead of quickly detecting issues with newer versions (that should work as well)? 🤔

I lean towards the former like we are doing with dependencies through lockfiles. Otherwise failing builds become a bit cumbersome to debug. If we update a Node or dependency version and that leads to a failing PR it's faster to pinpoint what the issue might be. What we can do to not compromise "quickly detecting" is having a GitHub action that opens a PR updating the Node versions if there are new versions available. How does it sound?

PR #333, opened 2022-08-22
New Result type for handling errors

alvaro-shopify:
Fixes https://github.com/Shopify/shopify-cli-planning/issues/312
Follow-up: https://github.com/Shopify/cli/pull/319

- New `Result` type added. Used as a return type to indicate the operation has suceded or should return an error
- Three utility methods for creating an ok and error response, and to get the value of an ok response or throw a fatal exception (Bug, Abort, AbortSilent) otherwise
- Used new type in `automaticMatchmaking`

TBD: It also should be added to `manualMatchIds` in case the new type is approved

### How to test your changes?
- Run `yarn shopify app deploy` with a remote app that matches local extensions and then with another that does not match to see the abort error message

### DIFF
--- a/packages/app/src/cli/services/environment/id-matching.ts
+++ b/packages/app/src/cli/services/environment/id-matching.ts
@@ -15,10 +15,8 @@ export async function automaticMatchmaking(
   remoteRegistrations: ExtensionRegistration[],
   identifiers: {[localIdentifier: string]: string},
 ): ResultAsync<MatchResult> {
-  const invalidEnvironmentError = err<MatchResult>(new Error('invalid-environment'))
-
   if (remoteRegistrations.length > localExtensions.length) {
-    return invalidEnvironmentError
+    return err()
   }
 
   const validIdentifiers = identifiers
@@ -65,13 +63,13 @@ export async function automaticMatchmaking(
   // The user must solve the issue in their environment or deploy to a different app
   const impossible = newRemotePending.filter((reg) => !newLocalPending.map((ext) => ext.graphQLType).includes(reg.type))
   if (impossible.length > 0 || newRemotePending.length > newLocalPending.length) {
-    return invalidEnvironmentError
+    return err()
   }
 
   // If there are more remote pending than local in total, then we can't automatically match
   // The user must solve the issue in their environment or deploy to a different app
   if (newRemotePending.length > newLocalPending.length) {
-    return invalidEnvironmentError
+    return err()
   }
 
   const extensionsToCreate: Extension[] = []
--- a/packages/app/src/cli/services/environment/id-matching.ts
+++ b/packages/app/src/cli/services/environment/id-matching.ts
@@ -1,7 +1,7 @@
 import {ExtensionRegistration} from '../dev/create-extension.js'
 import {IdentifiersExtensions} from '../../models/app/identifiers.js'
 import {Extension} from '../../models/app/extensions.js'
-import {Result, err, ok} from '@shopify/cli-kit/common/result'
+import {err, ok, ResultAsync} from '@shopify/cli-kit/common/result'
 
 export interface MatchResult {
   identifiers: IdentifiersExtensions
@@ -14,7 +14,7 @@ export async function automaticMatchmaking(
   localExtensions: Extension[],
   remoteRegistrations: ExtensionRegistration[],
   identifiers: {[localIdentifier: string]: string},
-): Promise<Result<MatchResult, Error>> {
+): ResultAsync<MatchResult> {
   const invalidEnvironmentError = err<MatchResult>(new Error('invalid-environment'))
 
   if (remoteRegistrations.length > localExtensions.length) {
--- a/packages/app/src/cli/services/environment/id-matching.ts
+++ b/packages/app/src/cli/services/environment/id-matching.ts
@@ -1,26 +1,24 @@
 import {ExtensionRegistration} from '../dev/create-extension.js'
 import {IdentifiersExtensions} from '../../models/app/identifiers.js'
 import {Extension} from '../../models/app/extensions.js'
+import {Result, err, ok} from '@shopify/cli-kit/common/result'
 
-export type MatchResult =
-  | {
-      result: 'ok'
-      identifiers: IdentifiersExtensions
-      pendingConfirmation: {extension: Extension; registration: ExtensionRegistration}[]
-      toCreate: Extension[]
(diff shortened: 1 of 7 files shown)


packages/app/src/cli/services/environment/id-matching.ts
```diff
@@ -1,26 +1,24 @@
...
+  toCreate: Extension[]
+  toManualMatch: {local: Extension[]; remote: ExtensionRegistration[]}
+}
 
 export async function automaticMatchmaking(
   localExtensions: Extension[],
   remoteRegistrations: ExtensionRegistration[],
   identifiers: {[localIdentifier: string]: string},
-): Promise<MatchResult> {
+): Promise<Result<MatchResult, Error>> {
```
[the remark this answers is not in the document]
alvaro-shopify: Regarding your comments:
* We could add also a ResultAsync<T, E> with T being a Promise as a wrapper of Promise<Result<T, E>>
* I don't think we should migrate all the code for using this new type. I think it is suitable for methods like this that makes you to design en structure that could return a correct response or an error. I guess that any other person that has to design a solution will go with another different design
* In case of using third party libraries like here, we should wrap the code to catch the error and return a custom `Result` 
https://github.com/Shopify/cli/blob/9f461c0adeb5b47564ff4ec9dd6b1ee7bfc732b4/packages/cli-kit/src/api/common.ts#L71:L94

alvaro-shopify:
Regarding your comments:
* We could add also a ResultAsync<T, E> with T being a Promise as a wrapper of Promise<Result<T, E>>
* I don't think we should migrate all the code for using this new type. I think it is suitable for methods like this that makes you to design en structure that could return a correct response or an error. I guess that any other person that has to design a solution will go with another different design
* In case of using third party libraries like here, we should wrap the code to catch the error and return a custom `Result` 
https://github.com/Shopify/cli/blob/9f461c0adeb5b47564ff4ec9dd6b1ee7bfc732b4/packages/cli-kit/src/api/common.ts#L71:L94

packages/cli-kit/src/common/result.ts
```diff
@@ -0,0 +1,37 @@
+export type Result<T, TError = Error> = {ok: true; value: T} | {ok: false; error: TError}
+
+export type ResultAsync<T, TError = Error> = Promise<Result<T, TError>>
+
+export const ok = <T, TError = Error>(value: T): Result<T, TError> => {
```
pepicrft, line 5: No strong preference here, but have you considered naming these `Ok` and `Err` instead of `ok` and `err`?

packages/cli-kit/src/common/result.ts
```diff
@@ -0,0 +1,37 @@
...
+  return {ok: true, value}
+}
+
+const UnknownError = new Error('Unknown error')
+
+export const err = <T = never>(error?: unknown): Result<T, Error> => {
```
pepicrft, line 11: Why are we making "an error that don't wrap errors" a valid state?

packages/cli-kit/src/common/result.ts
```diff
@@ -0,0 +1,37 @@
...
+    errorToUse = error
+  } else if (typeof error === 'string') {
+    errorToUse = new Error(error)
+  } else {
+    errorToUse = UnknownError
+  }
+  return {ok: false, error: errorToUse}
+}
+
+export const valueOrThrow = <T, TError = Error>(result: Result<T, TError>, error?: Error): T => {
```
pepicrft, line 25: This API opens the door for ignoring the error that's wrapped in the `result` parameter. What if we remove the second `error? parameter and have the following API instead:

```
mapError(result, (internalError) => { ... })
```

pepicrft (approved):

pepicrft:
> A LOT of pending work to modify the current code

I think this can be something gradual. Also, I don't think this means all our APIs will have result as a return type. I'd suggest that we do it in those APIs that return errors that we expect the consumer to manage. If in a method we come across scenarios that will make the execution enter some unpredictable state I think we should continue doing throws of aborts that cause the execution to terminate.

gonzaloriestra:
> I think this can be something gradual.

From my experience, unless you are very strict, that strategy usually ends up with unfinished stuff forever. So we should be happy living with a mix.

packages/cli-kit/src/common/result.ts
alvaro-shopify: But these are methods, does naming convention in Typescript recommends recommend capitalizing?

packages/cli-kit/src/common/result.ts
alvaro-shopify: To make the using of the method more flexible and easier. 
- You can pass your Error,
- You can pass a message to construct an Error
- You can use the method without parameters you a generic error is constructed

packages/cli-kit/src/common/result.ts
alvaro-shopify: But the `mapError` transform Result<T,E> to Result<T,U>. We could leave the API method `valueOrThrow` without the `error` parameter
At this point, why not using classes to define the result types and these API methods to that class. We will avoid having to pass the result as parameter and allows to concat several methods callings over the same result

packages/cli-kit/src/common/result.ts
```diff
@@ -0,0 +1,38 @@
+/* eslint-disable node/handle-callback-err */
+export type Result<T, TError> = Ok<T, TError> | Err<T, TError>
```
pepicrft, line 2: Beautiful APIs and simple implementation. Love it!

pepicrft (approved):

packages/cli-kit/src/common/result.test.ts
```diff
@@ -0,0 +1,59 @@
...
+  it('create ok with value', () => {
+    // When
+    const result = ok(123)
+
+    // Then
+    if (result.isErr()) {
+      fail('Should return an Ok result')
+    } else {
+      expect(result.value).toEqual(123)
+    }
```
isaacroldan, line 15: Why not use expects for the irErr too?
```suggestion
      expect(result.isErr()).toBeFalsy()
      expect(result.value).toEqual(123)
```

alvaro-shopify: Using the method `isErr` infers the type of the `result` (Ok or Err) so either `value` or `error` could be accessed. `expect` could be used but evaluation of the method `isErr` is mandatory to access one of those values directly

PR #337, opened 2022-08-23
Fix post merge debug flows

dmhenry:

The CLI and Go binary use different tech stacks, and hence, different debuggers. Previously, I set up a debug "pipeline" wherein the CLI Node process would kick off a remote Go debugger via shell script. This needs re-configuring as part of the merge of `shopify-cli-extensions` into `cli`.

Separating out launch configurations for the CLI and Go binaries. This is necessary because a given VS Code instance can only launch one debug config at a time. To debug the entire flow, a CLI config is necessary. The CLI then kicks a Go debugger and then transfers control to the Go binary. A second VS Code instance is then used to attach to the remote debugger. Note that, while this is technically a remote debug scenario, it is all performed on localhost.

### How to test your changes?

1. Open two VS Code instances: One from the root of the `cli` project and another from the `packages/ui-extensions-go-cli` directory.
2. In the latter instance, set a breakpoint on the first line of `func main()` in `packages/ui-extensions-go-cli/main.go`.
3. Optional: Control is transferred from the CLI to the Go binary in `packages/app/src/cli/utilities/extensions/cli.ts` on the line that invokes the `init-debug-session` shell script. Set a breakpoint here in the `cli` VS Code instance if you would like to inspect the CLI state before transferring control to the Go binary.
4. Using an existing app, launch one of the CLI debug launch configs: `shopify app build`, `shopify app dev`, or `shopify app scaffold extension`, responding to the prompts. If you set a breakpoint in the previous step, wait until the breakpoint is hit and then continue execution when you're satisfied. Note: The next step will not work if you did not set a breakpoint in the Go binary or if debugging is paused in the CLI.
5. From the second (Go binary) VS Code instance, launch the "Debug Go Binary" config. You will see two prompts, but assuming you did not change how the Go binary is launched in `cli.ts` or the `init-debug-session` shell script, you should use the pre-populated default values.
6. Observe the Go binary breakpoint pausing execution in `main.go`.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -7,20 +7,42 @@
       {
         "type": "node",
         "request": "launch",
+        "name": "shopify app build",
+        "cwd": "${workspaceFolder}",
+        "runtimeExecutable": "pnpm",
+        "runtimeArgs": [
+          "shopify",
+          "app",
+          "build",
+          "--path=${input:appDir}",
+        ],
+        "env": {
+          "DEBUG": "\"*\"",
+          "SHOPIFY_DEBUG_GO_BINARY": "1",       // Switch to 1 to pause execution of the Go binary until the debgger is attached.
+        },
+        "skipFiles": [
+          "<node_internals>/**",
+        ],
+      },
+      {
+        "type": "node",
+        "request": "launch",
         "name": "shopify app dev",
-        "runtimeExecutable": "yarn",
+        "cwd": "${workspaceFolder}",
+        "runtimeExecutable": "pnpm",
         "runtimeArgs": [
           "shopify",
           "app",
           "dev",
-          "--path=${input:appDir}"
+          "--no-update",
+          "--path=${input:appDir}",
         ],
         "env": {
           "DEBUG": "\"*\"",
-          "SHOPIFY_DEBUG_GO_BINARY": "0",       // Switch to 1 to pause execution of the Go binary until the debgger is attached.
+          "SHOPIFY_DEBUG_GO_BINARY": "1",       // Switch to 1 to pause execution of the Go binary until the debgger is attached.
         },
         "skipFiles": [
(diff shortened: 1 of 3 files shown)


.vscode/launch.json
```diff
@@ -4,23 +4,45 @@
...
+          "<node_internals>/**",
+        ],
+      },
       {
         "type": "node",
         "request": "launch",
         "name": "shopify app dev",
-        "runtimeExecutable": "yarn",
+        "cwd": "${workspaceFolder}",
+        "runtimeExecutable": "pnpm",
```
pepicrft, line 32: Is there a reason why we replaced the `yarn` executable? We did some experimentation with `pnpm` but we haven't made a decision yet on whether we should revisit our default option.

.vscode/launch.json
```diff
@@ -40,51 +62,19 @@
...
-        "host": "${input:remoteIpAddress}",
-        "port": "${input:remotePort}",
-      },
     ],
     "inputs": [
       {
         "id": "appDir",
         "type": "promptString",
         "description": "App Directory: ",
+        "default": ".",
```
pepicrft, line 77: What about setting as default the app under `fixtures/app`?

packages/ui-extensions-go-cli/.vscode/launch.json
```diff
@@ -0,0 +1,31 @@
+{
+    // Use IntelliSense to learn about possible attributes.
+    // Hover to view descriptions of existing attributes.
+    // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
+    "version": "0.2.0",
+    "configurations": [
```
pepicrft, line 6: Does VSCode load `.vscode/launch.json` files located in nested directories?

pepicrft (approved):

PR #341, opened 2022-08-23
Merge the docs into the repository

pepicrft:
Despite the value for the project, writing up documentation for contributors and keeping it up to date is gets often disregarded. This is unfortunate, and more so in a project like Shopify's CLI that's open-source and whose core package, `@shopify/cli-kit` and soon `@shopify/cli-ui` will be the foundation upon which 3P and 1P devs will build plugins to extend the CLI. We need to do better. We should aim at:
1. Removing unnecessary friction.
2. Making code the source of truth for API documentation.
3. Making writing and updating docs an integral part of contributing to the CLI.

This PR addresses (1). I noticed having the documentation in the repository's Wiki brings UI niceties with some cost:
- Changes can't be reviewed through PRs
- Documentation updates can't be shipped atomically with the code changes they are associated to.
- Updating the docs locally requires cloning the CLI repo with a flag to pull its submodules too.

To mitigate the above costs, I'm merging the docs into the repository in a `/docs` subdirectory. I'm also removing the outdated documentation of `@shopify/cli-kit` modules with the plan of automating its generation from JSDoc comments in those modules.


matteodepalo (approved):
Awesome change, having everything in one place makes a lot of sense 👍

docs/architecture.md
```diff
@@ -0,0 +1,33 @@
...
+It's important to understand the responsibility of each of the packages to ensure new code is placed in the right package,
+and that you can find the component you are looking for more easily. The image below shows an overview of the stack:
+
+<div style="textAlign: 'center'">
+  <img
+    src="assets/assets-stack.png"
+    alt="Example banner"
+    width="500"
+  />
+</div>
```
gonzaloriestra, line 21: The image is quite outdated, I think we should update it or remove it.

```suggestion
and that you can find the component you are looking for more easily.
```

gonzaloriestra (approved):
Good idea!

PR #343, opened 2022-08-23
Adding automation around initializing a git repository

lynchv:

Fixes https://github.com/orgs/Shopify/projects/4617/views/3

- When a user runs `shopify hydrogen deploy`, the current repository might not be git initialize. In that case we prompt the user if he desires the setup to be done as part of the command.
- When a user runs `yarn create @shopify/hydrogen` we add a step to git initialize the newly created project so the following deploy commands have less friction.


packages/cli-hydrogen/src/cli/commands/hydrogen/deploy.ts
```diff
@@ -29,16 +29,16 @@ export default class Deploy extends Command {
...
-    }),
     healthCheck: Flags.boolean({
       env: 'SHOPIFY_HYDROGEN_FLAG_HEALTH_CHECK',
       default: true,
       description: 'Require a health check before the deployment succeeds.',
     }),
+    dmsAddress: Flags.string({
+      hidden: true,
+      env: 'SHOPIFY_HYDROGEN_FLAG_DMS_ADDRESS',
+      default: 'oxygen-dms.shopifycloud.com',
```
lynchv, line 40: Small bug fix here where `https://` was added twice.

packages/cli-hydrogen/src/cli/services/deploy/config.integration.test.ts
```diff
@@ -1,44 +1,46 @@
...
+import {gitInit} from '../../prompts/git-init.js'
+import {describe, beforeEach, it, expect, vi} from 'vitest'
+import {error, file, git} from '@shopify/cli-kit'
 
 const defaultConfig = {deploymentToken: 'abcdefg', dmsAddress: 'https://integration.test', healthCheck: true}
+const mockedGitInitPrompt = vi.fn()
+
+beforeEach(() => {
+  vi.mock('../../prompts/git-init.js')
+  vi.mocked(gitInit).mockImplementation(mockedGitInitPrompt)
```
lynchv, line 11: Even though this is an integration test we still need to mock at least the user prompt interaction.

packages/cli-hydrogen/src/cli/services/deploy/config.integration.test.ts
```diff
@@ -1,44 +1,46 @@
...
+        const config = await getDeployConfig({...defaultConfig, path: tmpDir})
+
+        expect(config.commitMessage).toBe('Initial Commit')
+        expect(config.commitRef).toBe(`refs/heads/${defaultBranchConfig.value ?? 'master'}`)
+        expect(config.deploymentToken).toBe(defaultConfig.deploymentToken)
+        expect(config.dmsAddress).toBe(defaultConfig.dmsAddress)
+        expect(config.healthCheck).toBe(defaultConfig.healthCheck)
+        expect(config.path).toBe(tmpDir)
+        expect(new Date(config.timestamp).getTime()).toBeLessThan(new Date().getTime())
+        // can't test config.commitAuthor & config.commitSha
```
lynchv, line 42: Since this integration test can be run anywhere, there's certain things we can't test for due to flakiness. The commit author could or couldn't be set and there is no clear fallback like for default branch.

packages/create-hydrogen/src/services/init.ts
```diff
@@ -164,6 +164,24 @@ async function init(options: InitOptions) {
           task.title = 'Completed clean up'
         },
       },
+      {
+        title: 'Initializing a Git repository...',
```
lynchv, line 168: @cartogram If you wouldn't mind taking a look at the extra initialization step I added here. In addition, how where you testing the init command locally?

packages/create-hydrogen/src/services/init.ts
```diff
@@ -164,6 +164,24 @@ async function init(options: InitOptions) {
...
+        task: async (_, task) => {
+          await git.initializeRepository(templateScaffoldDir)
+          try {
+            await git.commitAll('Initial Commit', {directory: templateScaffoldDir})
```
graygilmore, line 172: I don't think we should be adding our own commit like this. I know this is common when setting up a repository for someone but that's because we have to. In my experience of running these kinds of commands it's just doing `git init` and letting you handle the rest. 

This is particularly important if there are some things being committed that shouldn't be (e.g. `node_modules`).

packages/create-hydrogen/src/services/init.ts
graygilmore: Is there also a default `.gitignore` file we should provide?

packages/cli-hydrogen/src/cli/services/deploy/config.ts
```diff
@@ -24,3 +25,18 @@ export const getDeployConfig = async (config: DeployConfig): Promise<ReqDeployCo
...
+    await git.ensureInsideGitDirectory(directory)
+  } catch (err: unknown) {
+    if ((err as error.Abort)?.message !== git.OutsideGitDirectoryError().message) throw error
+
+    const shouldGitInit = await gitInit()
+
+    if (!shouldGitInit) throw new error.AbortSilent()
+
+    await git.initializeRepository(directory)
+    await git.commitAll('Initial Commit', {directory})
```
graygilmore, line 40: Similar concern here (double commenting the same issue so I don't forget)

packages/cli-hydrogen/src/cli/prompts/git-init.ts
```diff
@@ -0,0 +1,18 @@
+import {ui} from '@shopify/cli-kit'
+
+export const gitInit = async (): Promise<boolean> => {
```
graygilmore, line 3: This is cool! Love interactions like this to help solve problems.

packages/create-hydrogen/src/services/init.ts
tempor1s: I really like the idea of a default `.gitignore` if we don't provide one already.

packages/create-hydrogen/src/services/init.ts
graygilmore: 👍🏻 sounds good, if that's the pattern that things are going these days then let's follow suit! I just did a `create-react-app` and it came with a `.gitignore` of this:

```
# See https://help.github.com/articles/ignoring-files/ for more about ignoring files.

# dependencies
/node_modules
/.pnp
.pnp.js

# testing
/coverage

# production
/build

# misc
.DS_Store
.env.local
.env.development.local
.env.test.local
.env.production.local

npm-debug.log*
yarn-debug.log*
yarn-error.log*
```

lynchv:
@tempor1s @graygilmore I added a function called `createGitIngore` in `packages/cli-kit/src/git.ts`, before I write some test let me know if that is what you had in mind.

I decided to use `appendSync` instead of `writeSync` since I don't want to squash someone's existing .gitignore just in case...?

packages/cli-hydrogen/src/cli/services/deploy/config.ts
```diff
@@ -24,3 +25,19 @@ export const getDeployConfig = async (config: DeployConfig): Promise<ReqDeployCo
...
+    await git.ensureInsideGitDirectory(directory)
+  } catch (err: unknown) {
+    if ((err as error.Abort)?.message !== git.OutsideGitDirectoryError().message) throw error
+
+    const shouldGitInit = await gitInit()
+
+    if (!shouldGitInit) throw new error.AbortSilent()
+
+    await git.initializeRepository(directory)
+    await git.createGitIngore(directory)
```
graygilmore, line 40:
```suggestion
    await git.createGitIgnore(directory)
```
😊

packages/cli-kit/src/git.ts
```diff
@@ -30,12 +31,38 @@ export const DetachedHeadError = () => {
...
+}
+
 export async function initializeRepository(directory: string) {
   debug(content`Initializing git repository at ${token.path(directory)}...`)
   await ensurePresentOrAbort()
   await git(directory).init()
 }
 
+export async function createGitIngore(directory: string) {
+  debug(content`Creating .gitignore at ${token.path(directory)}...`)
```
graygilmore, line 45: Is this for our own purposes?

packages/cli-kit/src/git.ts
```diff
@@ -30,12 +31,38 @@ export const DetachedHeadError = () => {
...
+  return new Abort('Nothing to commit')
+}
+
 export async function initializeRepository(directory: string) {
   debug(content`Initializing git repository at ${token.path(directory)}...`)
   await ensurePresentOrAbort()
   await git(directory).init()
 }
 
+export async function createGitIngore(directory: string) {
```
graygilmore, line 44: I'm not sure this is the appropriate place for this function in this package since it's specific to Hydrogen. Could this be a part of `create-hydrogen` instead? Or this function could be more generic where it takes in both a directory and content but by that point it wouldn't be doing much.

packages/cli-hydrogen/src/cli/services/deploy/config.integration.test.ts
graygilmore: 👍🏻 adding the context of why we can't test those things to the comment will be helpful

packages/cli-hydrogen/src/cli/services/deploy/config.ts
```diff
@@ -24,3 +25,19 @@ export const getDeployConfig = async (config: DeployConfig): Promise<ReqDeployCo
...
+    await git.commitAll('Initial Commit', {directory})
```
graygilmore, line 41:
```suggestion
    await git.commitAll('Initial commit generate by create-hydrogen', {directory})
```

(or something like that, maybe check in with Graham on some of these?)

packages/cli-kit/src/git.ts
lynchv: Just copying what was already there in other functions like `initializeRepository`.

lynchv:
### Force push notes:

- Added function to create a .gitignore file. Function takes in a template argument for future generalization.
- Made initial commit message more specific (will ask Graham for review at a latter stage as well)
- Bunch of small miscellaneous (typo, comments, etc.)

packages/cli-kit/src/git.test.ts
```diff
@@ -1,19 +1,8 @@
...
-  commitAll,
-  getHeadSymbolicRef,
-  GitNotPresentError,
-  OutsideGitDirectoryError,
-  ensurePresentOrAbort,
-  ensureInsideGitDirectory,
-  NoCommitError,
-  DetachedHeadError,
-} from './git.js'
+import * as git from './git.js'
```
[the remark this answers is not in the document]
lynchv: no but we were essentially doing the same by importing every single thing explicitly, just thought the import statement was getting out of hand.

tempor1s (approved):

graygilmore (approved):

Issue #312, opened 2022-08-17
[Bug]: `shopify hydrogen build` fails to find `tailwind.config.ts` when supplying `--path` flag


tempor1s:
### Please confirm that you have:

- [X] Searched [existing issues](.) to see if your issue is a duplicate. (If you’ve found a duplicate issue, feel free to add additional information in a comment on it.)
- [X] Reproduced the issue in [the latest CLI version](https://www.npmjs.com/package/@shopify/cli).

### In which of these areas are you experiencing a problem?

Hydrogen custom storefront

### Expected behavior

When I supply the `--path` flag when using `dev shopify hydrogen build` the build will succeed and the Hydrogen project would be built.

Note: I also talked with @cartogram about this internally who recommended I submit a bug report, so he has more context of the slack conversation.

### Actual behavior

When I supply the `--path` flag when using `dev shopify hydrogen build` the build will fail due to not being able to find the vite config file (seemingly because its using the cli directory to look for that file which you can see in the stack trace).

### Stack trace

```shell
dev shopify hydrogen build --path ../hydrogen/templates/demo-store
👩‍💻  Running yarn shopify from dev.yml
yarn run v1.22.18
$ nx build cli-main && node packages/cli-main/bin/dev.js hydrogen build --path ../hydrogen/templates/demo-store

   ✔    9/9 dependent project tasks succeeded [4 read from cache]

   Hint: you can run the command with --verbose to see the full dependent project outputs

 ———————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————


> nx run cli-main:build

$ tsc -b ./tsconfig.build.json

 ———————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————

 >  NX   Successfully ran target build for project cli-main and 9 task(s) it depends on (9s)

   Nx read the output from the cache instead of running the command for 4 out of 10 tasks.

⠙ Building client code
⠼ Building client code
◼ Building worker code
✓ 39 modules transformed.
✖ Cannot read properties of undefined (reading 'config')
◼ Building worker code
transforming (40) ../../node_modules/react-dom/node_modules/scheduler/cjs/scheduler.production.min.js
━━━━━━ Error ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
[...]

PR #346, opened 2022-08-24
Add tailwind error message for building with a different path

cartogram:

Fixes https://github.com/Shopify/cli/issues/312

This PR adds an error message for projects that build using a different path and also use tailwind.css. Currently, this combination doesn't work and the error message is confusing. 

Provides a better error under when those conditions are met to help inform the user of what the problem is.

### How to test your changes?

Pull down this branch and run `dev shopify hydrogen dev --path ../hydrogen/templates/demo-store`

### DIFF
--- a/packages/cli-hydrogen/src/cli/services/build.ts
+++ b/packages/cli-hydrogen/src/cli/services/build.ts
@@ -1,5 +1,5 @@
 import {build as viteBuild} from 'vite'
-import {ui, environment, error as kitError} from '@shopify/cli-kit'
+import {ui, environment, error as kitError, file, path, output} from '@shopify/cli-kit'
 
 type Target = 'node' | 'client' | 'worker'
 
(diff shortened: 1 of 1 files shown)


packages/cli-hydrogen/src/cli/services/build.ts
```diff
@@ -32,6 +32,26 @@ async function build({directory, targets, base}: DevOptions) {
...
+            ) {
+              const tailwindError = new kitError.Abort(
+                output.content`Running ${output.token.genericShellCommand(
+                  'shopify hydrogen build',
+                )} using a ${output.token.genericShellCommand(
+                  `--path`,
+                )} flag is not supported for projects using Tailwind.css.`,
+              )
+              throw tailwindError
+            }
```
pepicrft, line 53: I recommend turning this into models and logic that feels less scripty. Here's an example:

```ts
// Loading
const subject: HydrogenStorefront = try loadHydrogenStorefront()

// Validation
validate(subject)

// Execution
build(subject)
```

The storefront model can have an attribute whether or not a Tailwind configuration file exists and that's the information the validator uses to fail. Think of loading the model as taking a snapshot of what's in the system. Once the snapshot has been loaded, you pass it onto the next action in the pipeline to work with that object that you've loaded.

pepicrft: Needless to say, unit testing is easier because validators are pure functions that take an in-memory representation as input.

Issue #345, opened 2022-08-24
[Bug]: Live reload no longer works for extensions


vividviolet:
### Please confirm that you have:

- [X] Searched [existing issues](.) to see if your issue is a duplicate. (If you’ve found a duplicate issue, feel free to add additional information in a comment on it.)
- [X] Reproduced the issue in [the latest CLI version](https://www.npmjs.com/package/@shopify/cli).

### In which of these areas are you experiencing a problem?

Extension

### Expected behavior

When developing a local extension I expect the following to work:
- when I make changes to the extension src/index.tsx the script is rebuilt and reloaded automatically
- when I make changes to the localization json files for a Checkout UI extension the extension should reload automatically

### Actual behavior

- Making changes to the extension src/index.tsx does not trigger a rebuild of the extension script
- Making changes to the localization sometimes triggers a rebuild of the localization config but does not trigger a reload consistently

**Checkout UI**

https://user-images.githubusercontent.com/29458473/186295414-b596351b-6f03-4e79-8bf7-34720523bf2e.mov

### Reproduction steps

1. Scaffold a Checkout UI extension
2. Run `yarn dev`
3. Click on the Checkout UI extension preview link
4. Make changes to the src/index.tsx
5. Observe that the local extension doesn't reload
6. Make changes to the locales/en.default.json
7. Observe that the local extension doesn't reload

### Operating System

Mac OS Monterey

### Shopify CLI version (check your project's `package.json` if you're not sure)

3.8.0

### Shell

VS Code terminal and iTerm2

### Node version (run `node -v` if you're not sure)

v18.0.0

PR #351, opened 2022-08-24
Fixed Live reload no longer works for extensions

alvaro-shopify:

Fixes https://github.com/Shopify/cli/issues/345
Fixes https://github.com/Shopify/ui-extensions/issues/449

TBD: 
- Problem with reloading the locale content automatically could not be reproduced. To be decided if another issue should be created for that problem

- Fixes the configuration of the esBuild watch handler. Now the go extensions binary receives the messages when the extension is rebuild

### How to test your changes?
1. Scaffold a Checkout UI extension
2. Run yarn dev
3. Click on the Checkout UI extension preview link
4. Make changes to the src/index.tsx
5. Observe that the local extension is reloaded


pixelchutes:
Thank you @alvaro-shopify! Looks like this will fix https://github.com/Shopify/ui-extensions/issues/449

PR #353, opened 2022-08-24
Make Nx the interface to interact with the repository projects

pepicrft:
As explained in the decision-record document added in this PR, the implicit dependencies between actions codified in `package.json` `scripts` is undesirable for Nx's incremental build and caching capabilities. Nx provides an interface to declare each project's tasks and the dependencies between them across projects. 

I'm removing `scripts` from packages' `scripts` section with the exception of `prepack` which is executed when publishing new packages. Contributors to this repository will interact directly with `nx` by running `yarn nx build/test/lint` instead of `yarn run build/test/lint`. Nx will run the tasks in the right order and skip the ones whose output is already in the cache.

### How to test your changes?
You can try to run `yarn nx build/lint/test/tsc` with every project of the repository or with the entire repository.

### DIFF
--- a/.github/workflows/shopify-cli.yml
+++ b/.github/workflows/shopify-cli.yml
@@ -131,6 +131,7 @@ jobs:
       matrix:
         os: ['ubuntu-latest']
         node: ["18.7.0"]
+        target: ['build', 'tsc', 'lint']
     steps:
       - uses: actions/checkout@v2
         with:
@@ -168,12 +169,8 @@ jobs:
         run: yarn config set network-timeout 300000
       - name: Install dependencies
         run: yarn install --ignore-engines
-      - name: Build
-        run: yarn nx ${{ github.event.inputs.nxLevel || 'affected' }} --target=build
-      - name: Type-check
-        run: yarn nx ${{ github.event.inputs.nxLevel || 'affected' }} --target=tsc
-      - name: Lint
-        run: yarn nx ${{ github.event.inputs.nxLevel || 'affected' }} --target=lint
+      - name: ${{ matrix.target }}
+        run: yarn nx ${{ github.event.inputs.nxLevel || '--all --skip-nx-cache' }} --target=${{ matrix.target }}
 
   pr-platform-dependent:
     name: Testing with Node ${{ matrix.node }} in ${{ matrix.os }}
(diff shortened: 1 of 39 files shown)


.github/workflows/shopify-cli.yml
```diff
@@ -123,14 +123,15 @@ jobs:
...
-    name: Building, linting, and type-checking with Node ${{ matrix.node }} in ${{ matrix.os }}
+    name: ${{ matrix.target }} with Node ${{ matrix.node }} in ${{ matrix.os }}
     runs-on: ${{ matrix.os }}
     if: ${{ github.event_name == 'pull_request' }}
     timeout-minutes: 30
     strategy:
       matrix:
         os: ['ubuntu-latest']
         node: ["18.7.0"]
+        target: ['build', 'type-check', 'lint']
```
shauns, line 134: FWIW, build also type-checks

packages/app/project.json
```diff
@@ -6,11 +6,47 @@
     "implicitDependencies": ["ui-extensions-go-cli"],
     "tags": ["scope:feature"],
     "targets": {
+      "clean": {
+        "executor": "nx:run-commands",
```
shauns, line 10: would these be better as `run-script` executors? https://nx.dev/packages/nx/executors/run-script

(or the specific executor the command if one exists, I _suppose_ that's best?)

packages/cli-hydrogen/project.json
```diff
@@ -3,5 +3,49 @@
     "sourceRoot": "packages/cli-hydrogen/src",
     "projectType": "library",
     "implicitDependencies": [],
-    "tags": ["scope:feature"]
+    "tags": ["scope:feature"],
+    "targets": {
+      "clean": {
```
shauns, line 8: Given how similar these are between packages, it's a pity we have to restate them so often. Is there no way some base config could be extended?

shauns (changes requested):

packages/app/project.json
pepicrft: That executor runs a script but not an executable distributed by an NPM package. I'd write our own executor to simplify this, but I'd do it in a follow-up PR>

pepicrft:
@shauns I addressed your comments here. Few notes:
- I managed to get linting working without having to build the declaration files. We can pass the `EXPERIMENTAL_useSourceOfProjectReferenceRedirect: true` option to the Typescript ESLint plugin and that makes project references work.
- I agree there's a lot of duplication throughout the project.json's, but I'd suggest that we tackle it in a follow up PR by building a custom executor.

As follow-up items are: setup caching so that we have incremental builds and extract shared tasks into executors. Nx has also the concept of "generators" that we might want to use in the future for contributors to create new plugin packages in the repo.

shauns (approved):

PR #354, opened 2022-08-24
Update broken link

alxclark:

There's a broken link in our settings example. Initially the docs were named configuration and we rename `extension configuration` to `extension settings` but we ended up still keeping `/configuration` for the url of the shopify.dev page.

This fixes the broken url.


jmp-shopify (approved):
Verified URL goes to the correct place, LGTM!

PR #358, opened 2022-08-24
Add `@shopify/cli-polaris` to run Polaris tooling on Shopify apps

samrose3:

Related to https://github.com/Shopify/polaris/pull/6852

This allows Shopify app developers to run migrations for Polaris on their codebase, helping them update and stay on the latest version of Polaris.

Available commands:
- `shopify polaris migrate <migration> <path>` – performs a codemod migration related to Polaris components, styles, or other assets

This PR creates a new plugin for the Shopify CLI so Polaris-related tooling can be leveraged from within the same CLI.

### How to test your changes?

I tested using yalc in order to get access to the `@shopify/polaris-migrator` package, but this will be published soon on the npm registry.  _coming soon ✨_

Once published, you can test by running the following commands:

```sh
dev up
dev build
dev shopify polaris migrate <migration> <path>

# example
dev shopify polaris migrate replace-sass-spacing "**/*.scss"
```

### DIFF
--- a/packages/cli-polaris/package.json
+++ b/packages/cli-polaris/package.json
@@ -32,8 +32,7 @@
   },
   "dependencies": {
     "@oclif/core": "1.9.2",
-    "@shopify/cli-kit": "3.8.0",
-    "@shopify/polaris-migrator": "^0.0.0"
+    "@shopify/cli-kit": "3.8.0"
   },
   "devDependencies": {
     "vitest": "^0.22.1",
--- a/packages/cli-polaris/package.json
+++ b/packages/cli-polaris/package.json
@@ -1,0 +1,58 @@
+{
+  "name": "@shopify/cli-polaris",
+  "version": "3.8.0",
+  "private": false,
+  "description": "Commands for managing Polaris in Shopify apps",
+  "bugs": {
+    "url": "https://github.com/Shopify/cli/issues"
+  },
+  "homepage": "https://github.com/shopify/cli#readme",
+  "type": "module",
+  "files": [
+    "/dist"
+  ],
+  "publishConfig": {
+    "access": "public",
+    "@shopify:registry": "https://registry.npmjs.org"
+  },
+  "scripts": {
+    "prepack": "cross-env NODE_ENV=production yarn nx build && cp ../../README.md README.md",
+    "build": "nx build",
+    "clean": "nx clean",
+    "lint": "nx lint",
+    "lint:fix": "nx lint:fix",
+    "test": "nx test",
+    "test:watch": "nx test:watch",
+    "type-check": "nx type-check"
+  },
+  "eslintConfig": {
+    "extends": [
+      "../../.eslintrc.cjs"
+    ]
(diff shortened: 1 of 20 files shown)


gonzaloriestra:
I know it's still a draft, but one initial comment: this is not adding the Polaris commands as an optional plugin, but it will be included by default. If you want to make it optional, you should better move it to https://github.com/Shopify/awesome-shopify-cli-plugins and then enable [@oclif/plugin-plugins](https://github.com/oclif/plugin-plugins) here to allow installing them. Happy to help if you need!

packages/cli-polaris/project.json
```diff
@@ -0,0 +1,7 @@
+{
```
pepicrft, line 1: Same here, once we [merge this PR](https://github.com/Shopify/cli/pull/353) we'll have to add a `targets` section here with all the tasks.

packages/cli-polaris/package.json
```diff
@@ -0,0 +1,58 @@
...
+  },
+  "eslintConfig": {
+    "extends": [
+      "../../.eslintrc.cjs"
+    ]
+  },
+  "dependencies": {
+    "@oclif/core": "1.9.2",
+    "@shopify/cli-kit": "3.8.0",
+    "@shopify/polaris-migrator": "^0.0.0",
```
pepicrft, line 36: Does the tool already live in another package?

samrose3: Yes, the tool was build and then suggested that it be usable from the Shopify CLI. We could move all of the source code over to the CLI but thought this might be a good first-attempt strategy.

packages/cli-polaris/package.json
```diff
@@ -0,0 +1,58 @@
...
+    "@shopify/prettier-config": "^1.1.2",
+    "@types/prettier": "^2.6.3",
+    "prettier": "^2.6.1",
+    "typescript": "^4.7.4",
+    "vite": "^2.9.9"
```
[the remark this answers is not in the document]
samrose3: No, good point. These should probably be `devDependencies` since they are only required for the npm scripts and monorepo tooling 👍

samrose3:
> this is not adding the Polaris commands as an optional plugin, but it will be included by default.

Thanks @gonzaloriestra, great call out! The goal of adding this here is to make it more accessible to Shopify app developers. Making it a separate plugin would create a similar dev experience that currently exists using `npx @shopify/polaris-migrator`. I think it would be best to keep it part of the CLI by default to lower the barrier to usage and make it more easily discovered.

For context, I created the `@shopfiy/polaris-migrator` package not considering using the Shopify CLI. Then, it was suggested to integrate it and bring it into the current Shopify app developer experience.  Hope that makes sense!

If we still want to make it a separate plugin, then I would think it might be best kept out of the CLI since the usage and discoverability aren't improved much. I could see going either way, but the footprint is pretty small and could be a starting point for more robust Polaris tooling functionality in the future.

PR #359, opened 2022-08-25
CI improvements and fixes

pepicrft:
- [x] Adjust the build, tsc, and lint steps in the PR CI job to run concurrently (faster feedback in PRs).
- [x] Disable selective execution through the `affected` flag until [this work is done](https://github.com/Shopify/cli/pull/353) and each project accurately declares what are their input and output files. Otherwise we'll continue to see green PRs that might turn `main` red.
- [x] Use `--skip-nx-cache` when invoking Nx for the same reasons as the point above.
- [x] Remove `tsc` because type-checking is happening as part of the `build` action.


pepicrft:
I'm continuing the work [in this PR](https://github.com/Shopify/cli/pull/353)

PR #366, opened 2022-08-25
Using listr2 to run our deploy command

lynchv:

fix https://github.com/Shopify/oxygen-platform/issues/495
fix https://github.com/Shopify/oxygen-platform/issues/496
fix https://github.com/Shopify/oxygen-platform/issues/497

- Changed our command to use listr2 which automatically provides us with basic error retry logic. 
- Added missing unit test for create deployment, upload deployment and main deploy function.
- Re-structured config parsing step since we cannot prompt user inside a listr2 ([context](https://shopify.slack.com/archives/C030LHFCA5T/p1661451203648899))

Here is an example of what the output looks like :)

image


packages/cli-hydrogen/package.json
```diff
@@ -48,7 +48,8 @@
     "graphql-request": "^4.3.0",
     "prettier": "^2.6.1",
     "typescript": "^4.7.4",
-    "vite": "^2.9.9"
+    "vite": "^2.9.9",
+    "enquirer": "^2.3.6"
```
pepicrft, line 52: I'd advise not to introduce `enquirer`. It was our default in the past but [we replaced it with inquirer](https://github.com/Shopify/shopify-cli-planning/issues/298) because of issues in Linux environments.

tempor1s (commented):
Did a tophat and ran into a few things. Pushing a commit that we discussed to return the build step tasks to be executed & actually output the preview URL again. We also need to figure out a long term solution for issue of not being able to find new dependencies in `cli-kit`.

packages/cli-hydrogen/src/cli/services/deploy.test.ts
```diff
@@ -0,0 +1,105 @@
...
+  commitAuthor: 'commitAuthor',
+  commitSha: 'commitSha',
+  timestamp: 'timestamp',
+  commitRef: 'commitRef',
+}
+
+const mockedValidateProject = vi.fn()
+const mockedFillDeployConfig = vi.fn()
+const mockedCreateDeployment = vi.fn()
+const mockedBuildervice = vi.fn()
```
cartogram, line 27:
```suggestion
const mockedBuildService = vi.fn()
```

packages/cli-hydrogen/package.json
```diff
@@ -38,21 +38,21 @@
...
+    "prettier": "^2.6.1",
+    "vite": "^2.9.9",
+    "@shopify/cli-kit": "3.8.0",
     "fast-glob": "^3.2.11",
     "fs-extra": "^10.0.0",
     "graphql-request": "^4.3.0",
-    "prettier": "^2.6.1",
     "typescript": "^4.7.4",
-    "vite": "^2.9.9"
+    "@shopify/prettier-config": "^1.1.2"
```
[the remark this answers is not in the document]
lynchv: Had some issue with versioning so checkout the package.json from main for this branch. It's different then our current feature branch but should be similar to main.

Anyways, will need to review all package.json when opening the PR with main.

packages/cli-hydrogen/src/cli/services/build.ts
```diff
@@ -10,46 +10,46 @@ interface DevOptions {
...
+          await viteBuild({
+            ...commonConfig,
+            build: {
+              outDir: `dist/${key}`,
+              ssr: typeof value === 'string' ? value : undefined,
+              manifest: key === 'client',
+            },
+          })
+          // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        } catch (error: any) {
```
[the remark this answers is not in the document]
lynchv: Fully agree with that and the article is very useful will keep in mind for the future. This code was already written so preferred to make my changes as minimal as possible. Everything is green just because I split the code.

packages/cli-hydrogen/src/cli/services/deploy/deployer.test.ts
```diff
@@ -1,66 +1,123 @@
...
-  vi.mock('./error.js', async () => {
-    const module: any = await vi.importActual('./error.js')
-    return {
-      ...module,
-      retryOnError: vi.fn(),
+  vi.mock('@shopify/cli-kit/node/archiver')
+  vi.mock('node:fs')
+})
+
+describe('createDeploymentStep()', () => {
```
tempor1s, line 26: It could be nice if these were a bit more descriptive rather than just the method name (e.g. creating a deployment on DMS) or something like that.

tempor1s (approved):
Outside of the small nit, changes look good to me! 🎉

Issue #280, opened 2022-08-11
[Bug]: ui-extensions-server-kit should clear out unreachable extensions


NathanJolly:
### Please confirm that you have:

- [X] Searched [existing issues](.) to see if your issue is a duplicate. (If you’ve found a duplicate issue, feel free to add additional information in a comment on it.)
- [X] Reproduced the issue in [the latest CLI version](https://www.npmjs.com/package/@shopify/cli).

### In which of these areas are you experiencing a problem?

Extension

### Expected behavior

The ui-extensions-server-kit should clear unreachable extensions from its state, or it should provide some flag in the state that denotes this, or it should provide some function to clear state.

### Actual behavior

There is no indication that any extension is unreachable.

Example payload where the first extension is unreachable, and the second extension is reachable:
`[
  {
    "assets": {
      "main": {
        "name": "main",
        "url": "https://6cf1-2601-196-8780-d620-3942-12e8-331c-9bed.ngrok.io/extensions/dev-a8a22e16-0bf5-4239-8f57-3e3a82154ee2/assets/main.js?lastUpdated=1660160016",
        "lastUpdated": 1660160016
      }
    },
    "capabilities": {
      "networkAccess": false
    },
    "development": {
      "resource": {
        "url": "invalid_url"
      },
      "root": {
        "url": "https://6cf1-2601-196-8780-d620-3942-12e8-331c-9bed.ngrok.io/extensions/dev-a8a22e16-0bf5-4239-8f57-3e3a82154ee2"
      },
      "hidden": false,
      "status": "success",
      "localizationStatus": ""
    },
    "extensionPoints": [
      "Retail::SmartGrid::Tile",
      "Retail::SmartGrid::Modal"
    ],
    "localization": null,
    "metafields": null,
    "type": "pos_ui_extension",
    "uuid": "dev-a8a22e16-0bf5-4239-8f57-3e3a82154ee2",
    "version": "0.1.0",
    "surface": "pos",
    "title": "dev-workflow-tester"
  },
  {
    "assets": {
      "main": {
        "name": "main",
        "url": "https://0006-2601-196-8780-d620-11d2-7c4e-b1d3-c0a2.ngrok.io/extensions/dev-fdd5ccca-4d57-4a8f-a8b9-146c23185d3d/assets/main.js?lastUpdated=1660226306",
        "lastUpdated": 1660226306
      }
    },
    "capabilities": {
      "networkAccess": false
    },
    "development": {
      "resource": {
[...]

PR #367, opened 2022-08-25
Fix extensions being remembered from previous connections

dmhenry:
Extensions from previous connects were attempting to be replaced by UUID though the UUID had changed. The effect being that extensions were being remembered from previous connections. This PR fixes that issue.

Co-authored-by: Richard Powell <richard.powell@shopify.com>

Fixes #[280](https://github.com/Shopify/cli/issues/280) 

Resets the extensions per connect rather than trying to remove them from a map by UUID.'

### Measuring impact

The unit test labeled `replaces existing extensions when connecting twice` tests extension state using multiple connections

### How to test your changes?

The unit test labeled `replaces existing extensions when connecting twice` tests extension state using multiple connections


packages/ui-extensions-server-kit/src/state/reducers/extensionServerReducer.test.ts
```diff
@@ -15,16 +15,40 @@ describe('extensionServerReducer()', () => {
   it('connects to server', () => {
     const app = mockApp()
     const extension = mockExtension()
-    const action = createConnectedAction({app, extensions: [extension], store: 'test-store.com'})
-    const state = extensionServerReducer(INITIAL_STATE, action)
+    const action1 = createConnectedAction({app, extensions: [extension], store: 'test-store.com'})
```
vividviolet, line 18: accidental change?

packages/ui-extensions-server-kit/src/state/reducers/extensionServerReducer.test.ts
```diff
@@ -15,16 +15,40 @@ describe('extensionServerReducer()', () => {
...
+    const firstState = extensionServerReducer(INITIAL_STATE, action1)
```
vividviolet, line 19: same here - accidental change?

packages/ui-extensions-server-kit/src/state/reducers/extensionServerReducer.ts
```diff
@@ -17,7 +17,7 @@ export function extensionServerReducer(state: ExtensionServerState, action: Exte
         ...state,
         store: action.payload.store,
         app: {...(state.app ?? {}), ...(action.payload.app ?? {})},
```
vividviolet, line 19: Can you fix this as well? Everything should be replaced on a "connected" message

dmhenry: Are you saying line 19 should just be `app`?

vividviolet:
Yes, when a connected message comes in everything in the state should be
replaced with the new payload - the app should not be persisted from a
previously run

isaacroldan (approved):

vividviolet (approved):

PR #370, opened 2022-08-26
Introduce theme app extensions support in the `app dev` command

karreiro:

Fixes https://github.com/Shopify/shopify-cli/issues/2540.

This PR introduces the hot reload feature for theme app extension developers. It reuses the same development server that theme developers use, in the context of theme app extension development.

### How to test your changes?

- Create an app
- Generate a theme app extension with `yarn shopify app generate extension --path <app_directory>`
- Run `yarn shopify app dev --path <theme_app_extension_directory>`
- Notice the CLI 3.x is running the `extension serve` command

demo

### Post-release steps

Merge the documentation PR https://github.com/Shopify/shopify-dev/pull/26125.

### Update checklist

- [x] I've added a CHANGELOG entry for this PR (if the change is public-facing)
- [x] I've considered possible cross-platform impacts (Mac, Linux, Windows).
- [x] I've left the version number as is (we'll handle incrementing this when releasing).
- [x] I've included any post-release steps in the section above (if needed).

### DIFF
--- a/packages/app/src/cli/commands/app/dev.ts
+++ b/packages/app/src/cli/commands/app/dev.ts
@@ -78,11 +78,10 @@ export default class Dev extends Command {
       description: 'Theme ID or name of the theme app extension host theme.',
       env: 'SHOPIFY_FLAG_THEME',
     }),
-    port: Flags.integer({
+    'theme-app-extension-port': Flags.integer({
       hidden: false,
-      char: 'p',
       description: 'Local port of the theme app extension development server.',
-      env: 'SHOPIFY_FLAG_PORT',
+      env: 'SHOPIFY_FLAG_THEME_APP_EXTENSION_PORT',
     }),
   }
 
@@ -106,7 +105,7 @@ export default class Dev extends Command {
       tunnel: flags.tunnel,
       noTunnel: flags['no-tunnel'],
       theme: flags.theme,
-      port: flags.port,
+      themeExtensionPort: flags['theme-app-extension-port'],
     })
   }
 }
--- a/packages/app/src/cli/commands/app/dev.ts
+++ b/packages/app/src/cli/commands/app/dev.ts
@@ -72,6 +72,18 @@ export default class Dev extends Command {
       default: false,
       exclusive: ['tunnel-url', 'no-tunnel'],
     }),
+    theme: Flags.string({
+      hidden: false,
+      char: 't',
+      description: 'Theme ID or name of the theme app extension host theme.',
+      env: 'SHOPIFY_FLAG_THEME',
+    }),
+    port: Flags.integer({
+      hidden: false,
+      char: 'p',
+      description: 'Local port of the theme app extension development server.',
+      env: 'SHOPIFY_FLAG_PORT',
+    }),
   }
 
   public async run(): Promise<void> {
@@ -93,6 +105,8 @@ export default class Dev extends Command {
       tunnelUrl: flags['tunnel-url'],
       tunnel: flags.tunnel,
       noTunnel: flags['no-tunnel'],
+      theme: flags.theme,
+      port: flags.port,
     })
   }
 }
--- a/packages/app/src/cli/services/dev.ts
+++ b/packages/app/src/cli/services/dev.ts
@@ -3,6 +3,7 @@ import {generateFrontendURL, generatePartnersURLs, getURLs, shouldOrPromptUpdate
 import {installAppDependencies} from './dependencies.js'
 import {devExtensions} from './dev/extension.js'
 import {outputAppURL, outputExtensionsMessages, outputUpdateURLsResult} from './dev/output.js'
+import {themeExtensionArgs} from './dev/theme-extension-args.js'
 import {
   ReverseHTTPProxyTarget,
   runConcurrentHTTPProcessesAndPathForwardTraffic,
@@ -13,6 +14,7 @@ import {fetchProductVariant} from '../utilities/extensions/fetch-product-variant
 import {analytics, output, port, system, session, abort} from '@shopify/cli-kit'
 import {Config} from '@oclif/core'
 import {OutputProcess} from '@shopify/cli-kit/src/output.js'
+import {execCLI2} from '@shopify/cli-kit/node/ruby'
 import {Writable} from 'node:stream'
 
 export interface DevOptions {
@@ -28,6 +30,8 @@ export interface DevOptions {
   tunnelUrl?: string
   tunnel: boolean
   noTunnel: boolean
+  theme?: string
+  port?: number
 }
 
 interface DevWebOptions {
@@ -96,8 +100,9 @@ async function dev(options: DevOptions) {
   const proxyTargets: ReverseHTTPProxyTarget[] = []
   const proxyPort = usingLocalhost ? await port.getRandomPort() : frontendPort
   const proxyUrl = usingLocalhost ? `${frontendUrl}:${proxyPort}` : frontendUrl
+
   if (options.app.extensions.ui.length > 0) {
-    const devExt = await devExtensionsTarget(
+    const devExt = await devUIExtensionsTarget(
       options.app,
       apiKey,
       proxyUrl,
@@ -106,6 +111,10 @@ async function dev(options: DevOptions) {
       options.subscriptionProductUrl,
       options.checkoutCartUrl,
     )
+    proxyTargets.push(devExt)
+  }
+  if (options.app.extensions.theme.length > 0) {
+    const devExt = await devThemeExtensionTarget(options, apiKey, storeFqdn, token)
     proxyTargets.push(devExt)
   }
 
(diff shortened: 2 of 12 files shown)


karreiro:
(the branch was rebased)

mgmanzella (approved):

packages/app/src/cli/commands/app/dev.ts
```diff
@@ -63,6 +63,18 @@ export default class Dev extends Command {
...
+      char: 't',
+      description: 'Theme ID or name of the theme app extension host theme.',
+      env: 'SHOPIFY_FLAG_THEME',
+    }),
+    port: Flags.integer({
+      hidden: false,
+      char: 'p',
+      description: 'Local port of the theme app extension development server.',
+      env: 'SHOPIFY_FLAG_PORT',
+    }),
```
isaacroldan, line 77: We wanted to add a `port` property for future use when serving the extensions via localhost or custom tunnel 🤔 
Maybe we could have a more generic message here? or leave it like this and update it in the future? // @pepicrft

packages/app/src/cli/services/dev.ts
```diff
@@ -163,6 +173,24 @@ interface DevFrontendTargetOptions extends DevWebOptions {
...
+  options: DevOptions,
+  apiKey: string,
+  store: string,
+  token: string,
+): ReverseHTTPProxyTarget {
+  return {
+    logPrefix: 'extensions',
+    action: async (_stdout: Writable, _stderr: Writable, _signal: abort.Signal, _port: number) => {
+      const adminSession = await session.ensureAuthenticatedAdmin(store)
+      const storefrontToken = await session.ensureAuthenticatedStorefront()
```
isaacroldan, line 186: Not sure about doing this here 🤔 
I think these tokens should be obtained at the start of `dev` (same as the partners token) if needed.

Admin tokens are a bit tricky and prone to errors because of the identity issue with development stores, better to throw that earlier than here where everything else is already running.

packages/app/src/cli/services/dev/theme-extension-args.test.ts
```diff
@@ -0,0 +1,96 @@
...
+      // Optional properties
+      '--theme',
+      'theme ID',
+      '--port',
+      '8282',
+    ])
+  })
+})
+
+export function testApp(): AppInterface {
```
isaacroldan, line 63: there is already a `testApp` method in file `app.test-data.ts`, better to reuse that one :)

packages/app/src/cli/services/dev/theme-extension-args.test.ts
```diff
@@ -0,0 +1,96 @@
...
+      ui: [],
+      theme: [testThemeExtensions()],
+      function: [],
+    },
+    hasExtensions: () => true,
+    updateDependencies: async () => {},
+  }
+}
+
+export function testThemeExtensions(): ThemeExtension {
```
isaacroldan, line 83: This can be moved to the `app.test-data.ts` file as well

packages/app/src/cli/commands/app/dev.ts
karreiro: @isaacroldan @amcaplan I believe both approaches work, we might use a more generic description or a different flag name.

Personally, as  developer will have different ports to manage, I believe we maybe could rename it to `--extension-port`. What do you think?

amcaplan: Isn't there already a port for the UI extension dev console? And another port where UI extensions are actually served? This is a third kind of extension port. I think something like `--theme-app-extension-port`, while lengthy, might be best.

isaacroldan: Agree with @amcaplan. Also, couldn't the CLI just select a random port?

packages/app/src/cli/services/dev/theme-extension-args.ts
```diff
@@ -0,0 +1,38 @@
+import {AppInterface} from '../../models/app/app.js'
+import {ensureDeployEnvironment} from '../environment.js'
+
+export async function themeExtensionArgs(
+  apiKey: string,
+  options: {app: AppInterface; theme?: string; port?: number; reset: boolean},
+) {
+  const {identifiers} = await ensureDeployEnvironment(options)
```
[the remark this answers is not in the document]
isaacroldan: Problem is, we can't have this during the `dev` flow, it will essentially break it (because of the validations we do in `ensureDeployEnvironment`)
We will need to find a way to obtain those IDs without calling `ensureDeployEnvironment`. I can take a look tomorrow on alternatives and maybe we can pair and discuss the solution :)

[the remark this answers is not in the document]
karreiro: Awesome, let's do that :)

packages/app/src/cli/services/dev/theme-extension-args.test.ts
karreiro: Good catch! Fixed :)

packages/app/src/cli/services/dev.ts
karreiro: That makes sense, it's better fail earlier indeed! Fixed :)

packages/app/src/cli/services/dev/theme-extension-args.test.ts
karreiro: Good catch! Fixed :)

packages/app/src/cli/commands/app/dev.ts
karreiro: I agree with @amcaplan as well, it's more flexible to keep the port attached to theme extensions, then we can keep the port flag available for future usage.

Regarding the usage of a random port, I'd prefer to keep this following the theme development server standard. Still, we might revisit it in a future PR applying the random port approach for both cases.

packages/app/src/cli/services/environment.ts
```diff
@@ -225,28 +228,31 @@ async function fetchDevAppAndPrompt(app: AppInterface, token: string): Promise<O
...
-    envIdentifiers = {app: undefined, extensions: {}}
-  } else if (envIdentifiers.app) {
-    const apiKey = options.apiKey ?? envIdentifiers.app
-    partnersApp = await fetchAppFromApiKey(apiKey, token)
-    if (!partnersApp) throw DeployAppNotFound(apiKey, options.app.packageManager)
-  } else {
-    partnersApp = await fetchDevAppAndPrompt(options.app, token)
+  if (remoteRegistrations.length > 0) {
+    return remoteRegistrations[0]!
   }
```
isaacroldan, line 244: this is not correct, `remoteRegistrations` includes all extensions for the app, of all types. The first one is not necessarily the theme app one. You'd need to filter the list to check if there is a theme app extension and return that.
All this code assumes we won't never increase the allowed number of extensions, but i guess that's an issue to solve in the future :)

packages/app/src/cli/services/dev.ts
```diff
@@ -107,6 +114,11 @@ async function dev(options: DevOptions) {
     )
     proxyTargets.push(devExt)
   }
+  if (options.app.extensions.theme.length > 0) {
+    const extension = options.app.extensions.theme[0]!
+    const devExt = await devThemeExtensionTarget(options, extension, apiKey, adminSession!, storefrontToken!, token)
+    proxyTargets.push(devExt)
```
isaacroldan, line 120: The theme extension doesn't need to be accessed through the proxy, right?
(i.e., doesn't need to be access publicly via ngrok for instance)

Then instead of adding it to `proxyTargets`, you should add it to `additionalProcesses`, like we do with the backend (that is not accessible directly through the tunnel)

⚠️  A note for the future: we could have a public URL so you can share it, something like:
`https://{tunnel_url}/theme-app-ext`. In that case then it should be added to `proxyTargets` and we'd need to update the proxy

packages/app/src/cli/services/environment.ts
karreiro: Great catch! I've applied the fix in a new commit :) - Thank you, @isaacroldan! :)

packages/app/src/cli/services/dev.ts
```diff
@@ -55,6 +60,7 @@ async function dev(options: DevOptions) {
     tunnelPlugin,
   } = await ensureDevEnvironment(options, token)
   const apiKey = identifiers.app
+  const [adminSession, storefrontToken] = await ensureThemeExtensionTokens(options, storeFqdn)
```
isaacroldan, line 63: Not sure about this being optional here 🤔 
If it's going to be after `ensureDevEnvironment` then is better for this to be on line 118 where we are sure we need the tokens, instead of them being optional.

The other option is to call this on line 54, but i guess that's not possible because we need the store

packages/app/src/cli/services/environment.ts
```diff
@@ -288,6 +294,32 @@ async function fetchOrganizationAndFetchOrCreateApp(
   return {orgId, partnersApp}
 }
 
+async function fetchAppAndIdentifiers(
+  options: {app: AppInterface; reset: boolean; packageManager?: PackageManager; apiKey?: string},
+  token: string,
+): Promise<[OrganizationApp, Partial<UuidOnlyIdentifiers>]> {
```
isaacroldan, line 302: 👏

packages/app/src/cli/services/dev.ts
karreiro: Precisely, the theme extension doesn't need to be accessed through the proxy indeed. I've applied the fix - Thank you! :)

karreiro:
Thank you, @isaacroldan! 🚀 

@amcaplan I've included the fixes you've suggested on new commits. Please, let me know if you have any other concerns, then I may apply on the following PR :)

packages/app/src/cli/services/environment.ts
karreiro: 🚀

PR #383, opened 2022-08-29
Fix product subscription react template

henrytao-me:

This PR fixes product subscription react template. Generally, it works right now. However, as soon as we add hook to App component, it breaks. The reason is because the `render` function expects a function returning a React element.

Fix the render call and use `useExtensionApi` to query for `extensionPoint`.

### How to test your changes?

N/A

### Measuring impact

### Checklist


henrytao-me:
Note that: I double check other react template. Both `checkout_post_purchase` and `pos_ui_extension` are also wrong.

PR #387, opened 2022-08-30
Rename web-env to env pull

matteodepalo:

Fixes https://github.com/Shopify/shopify-cli-planning/issues/370

We want to rename the `web-env` command to `env pull` so that this new namespace can also be used by other `env` related commands in other areas of the CLI.

I've renamed `web-env` into `env pull` and `env show`.
- `env pull` always updates the `.env` file
- `env shows` just displays the env vars

### How to test your changes?

Run `dev fixture env pull` and `dev fixture env show`

### Measuring impact

### Checklist

### DIFF
--- a/.changeset/pink-lies-buy.md
+++ b/.changeset/pink-lies-buy.md
@@ -2,4 +2,4 @@
 '@shopify/app': patch
 ---
 
-Rename `web-env` command to `env populate` and `env show`
+Rename `web-env` command to `env pull` and `env show`
--- a/packages/app/src/cli/commands/env/pull.ts
+++ b/packages/app/src/cli/commands/env/pull.ts
@@ -1,13 +1,13 @@
 import {appFlags} from '../../flags.js'
 import {AppInterface} from '../../models/app/app.js'
 import {load as loadApp} from '../../models/app/loader.js'
-import {populateEnv} from '../../services/env/populate.js'
+import {pullEnv} from '../../services/env/pull.js'
 import {Flags} from '@oclif/core'
 import {output, path, cli} from '@shopify/cli-kit'
 import Command from '@shopify/cli-kit/node/base-command'
 
-export default class EnvPopulate extends Command {
-  static description = 'Populates the environment variables for your app and extensions'
+export default class EnvPull extends Command {
+  static description = 'Pulls the environment variables for your app and extensions'
 
   static hidden = true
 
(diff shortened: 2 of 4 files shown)


packages/app/src/cli/commands/env/show.ts
```diff
@@ -0,0 +1,27 @@
...
+  }
+
+  public async run(): Promise<void> {
+    const commandConfig = this.config
+    const {flags} = await this.parse(EnvShow)
+    const directory = flags.path ? path.resolve(flags.path) : process.cwd()
+    const envFile = path.resolve(directory)
+    const app: AppInterface = await loadApp(directory, 'report')
+    output.info(await showEnv(app, {config: commandConfig}))
+    if (app.errors) process.exit(2)
```
gonzaloriestra, line 25: Will the previous output work if there are errors? Also, instead of exiting, we usually call abort with the error message.

packages/app/src/cli/services/env/show.ts
```diff
@@ -0,0 +1,34 @@
...
+
+type Format = 'json' | 'text'
+
+interface ShowEnvOptions {
+  config: Config
+}
+
+export async function showEnv(app: AppInterface, {config}: ShowEnvOptions): Promise<output.Message> {
+  const output = await outputEnv(app, 'text')
+  await analytics.reportEvent({config})
```
gonzaloriestra, line 14: I think this is not needed, as it's already done in the postrun hook, no?

gonzaloriestra (commented):
At some point we will probably add a populate command to generate content for a store, as [we have in CLI2](https://shopify.dev/apps/tools/cli/cli-2#populate). So maybe this `env populate` could be a bit misleading. What about `env save` instead?

matteodepalo:
I think `save` could work, but I'd like to hear from the Hydrogen team as well if it would fit the way they'd want to pull env vars from different environments.

pepicrft (approved):

packages/app/src/cli/commands/env/show.ts
matteodepalo: @gonzaloriestra I investigated this more. I took inspiration from this commit https://github.com/Shopify/cli/commit/a0434ac2b68f00061dc32a24c52b14d001f691ae and I think we want to exit when showing `env` because the command might output stuff even if the app info is invalid. I'm happy to remove this line though if you think it's not worth it.

matteodepalo: I think it's fine in this case to do nothing if the app has errors, I feel like this is more suited for the `info` task. But it's good to know that we can throw `AbortSilent` for these cases.

isaacroldan (approved):

PR #389, opened 2022-08-30
`tunnel` flag and default localhost

isaacroldan:
As part of the new `dev` flow, we want `localhost` to be the default option but with tunnels being easy to opt-in.
This is the first iteration. Tunnel selection and multiple plugins will come later.

Fixes https://github.com/Shopify/shopify-cli-planning/issues/326

- Add to `tunnel` flag (false by default)
- Make `tunnel`, `no-tunnel` and `tunnel-url` incompatible
- Make `no-tunnel` public (and false by default)
- Refactor some `dev` code for better readability
- Extract frontend URL generation and test it

### How to test your changes?
- Create an app without extensions
  - `dev` that app, you should see `localhost` links
  - use `dev --tunnel`, you should see ngrok links
  - use both `--tunnel` and `--tunnel-url` you should see an error
- Add an extension to the app:
  - `dev` without `tunnel`, you should see ngrok links
  - `dev` with `no-tunnel` you should see localhost links

### Post-release steps

Release this documentation PR: https://github.com/Shopify/shopify-dev/pull/25786

### Measuring impact

### Checklist


packages/app/src/cli/commands/app/dev.ts
```diff
@@ -57,11 +57,13 @@ export default class Dev extends Command {
...
-      hidden: true,
-      description: 'Automatic creation of a tunnel is disabled. Service entry point will listen to localhost instead',
-      env: 'SHOPIFY_FLAG_NO_TUNNEL',
+    tunnel: Flags.boolean({
+      hidden: false,
+      description: 'Use ngrok to create a tunnel. Will be enabled automatically if you have extensions',
+      env: 'SHOPIFY_FLAG_TUNNEL',
       default: false,
+      allowNo: true,
+      exclusive: ['tunnel-url'],
```
pepicrft, line 66: TIL about this attribute :P

packages/app/src/cli/commands/app/dev.ts
```diff
@@ -82,7 +85,7 @@ export default class Dev extends Command {
       subscriptionProductUrl: flags['subscription-product-url'],
       checkoutCartUrl: flags['checkout-cart-url'],
       tunnelUrl: flags['tunnel-url'],
-      noTunnel: flags['no-tunnel'],
+      tunnel: flags.tunnel || hasUIExtensions,
```
pepicrft, line 88: Do we want to move this into the service and test that the right `tunnel` value is used based on the number of extensions?

packages/app/src/cli/services/dev.ts
```diff
@@ -151,12 +131,44 @@ async function dev(options: DevOptions) {
   }
 }
 
+async function generateFrontendURL(options: DevOptions): Promise<{frontendUrl: string; frontendPort: number}> {
```
pepicrft, line 134: What about extracting this into its own service that we can unit-test in isolation? For example `services/dev/generate-url.ts`

packages/app/src/cli/services/dev.ts
```diff
@@ -151,12 +131,44 @@ async function dev(options: DevOptions) {
...
+  let frontendPort: number
+  let frontendUrl: string
+
+  if (options.tunnel === false) {
+    frontendPort = await port.getRandomPort()
+    frontendUrl = 'http://localhost'
+  } else if (options.tunnelUrl) {
```
pepicrft, line 141: In recent testing of 1P apps from a local environment I realized there's a need for passing the port without a tunnel URL (i.e. `--port`). This is the case locally because `dev` runs the processes in a virtualized environment and opens a port to access it from the host. That being said, I think we should break the convention here and require users to use two flags: `--tunnel-url https://my-tunnel-url.com --port 12345`

pepicrft (approved):

packages/app/src/cli/services/dev/urls.test.ts
```diff
@@ -253,3 +257,120 @@ describe('shouldOrPromptUpdateURLs', () => {
     expect(outputMock.output()).toMatch(/example.com\/auth\/callback/)
   })
 })
+
+describe('generateFrontendURL', () => {
```
gonzaloriestra, line 261: ❤️

gonzaloriestra (approved):
🎩 All the tunnel flags are working as expected now, thanks!

The only pending thing would be that `Shareable app URL` message, but it's a minor detail.

packages/ui-extensions-go-cli/api/root/templates/checkout_post_purchase/tunnel-error.html.tpl
```diff
@@ -35,7 +35,7 @@
   <body>
   <div class="content">
       <p>This page is served by your local UI Extension development server. Instead of visiting this page directly, you will need to connect your local development environment to a real checkout environment.</p>
-      <p>Make sure you have a secure URL for your local development server by running <code>shopify extension tunnel start --port={{ .Port }}</code>, create a checkout, and append <code>?dev=https://TUNNEL_URL{{.ApiRoot}}/{{.UUID}}</code> to the URL, where <code>TUNNEL_URL</code> is replaced with your own ngrok URL.</p>
+      <p>Make sure you have a secure URL for your local development server by using the <code>--tunnel</code> flag.</p>
```
isaacroldan, line 38: Since now you can potentially execute extensions without a tunnel, this error might appear so I took the opportunity to update the messaging.

isaacroldan: Captura de Pantalla 2022-08-31 a las 11 36 47
to
Captura de Pantalla 2022-08-31 a las 11 36 55

isaacroldan:
@gonzaloriestra I'll discuss the Shareable copy with Meredith and update it if necessary in the next PR 👌

PR #405, opened 2022-08-31
Add hydrogen deploy command

lynchv:

Fixes #0000 

### Measuring impact

### Checklist


packages/cli-hydrogen/src/cli/services/deploy/upload.ts
```diff
@@ -0,0 +1,79 @@
...
+  return response.createDeployment
+}
+
+export const uploadDeployment = async (config: ReqDeployConfig, deploymentID: string): Promise<string> => {
+  let headers = await api.buildHeaders(config.deploymentToken)
+
+  // note: may need validation for invalid deploymentID? oxygenctl does it
+  // note: we may want to remove the zip that we create in in this step
+  const distPath = `${config.path}/dist`
+  const distZipPath = `${distPath}/dist.zip`
```
tempor1s, line 39: Going to be moving this to be created in a temporary directory so that we do not pollute the `dist` directory that is created by Vite.

lynchv: This will be part of another PR later down the road.

lynchv:
Squashed commits into a single one.

.changeset/lovely-foxes-bow.md
```diff
@@ -0,0 +1,14 @@
+---
+'@shopify/cli-kit': minor
+'@shopify/create-hydrogen': minor
+'@shopify/cli-hydrogen': patch
+---
+
+- cli-hydrogen:
```
tempor1s, line 7: We are not mentioning the `shopify hydrogen deploy` command in the release notes since its hidden & restricted to Shopify staff only at the moment. Once we remove those conditions we will add it as part of the release notes.

packages/cli-kit/src/git.ts
```diff
@@ -12,12 +13,61 @@ export const GitNotPresentError = () => {
   )
 }
 
+export const OutsideGitDirectoryError = () => {
```
pepicrft, line 16: I'd include the directory in which we are trying to run a Git command. That'll make it easier to debug issues that arise.

packages/cli-kit/src/git.ts
```diff
@@ -12,12 +13,61 @@ export const GitNotPresentError = () => {
...
+    [GitIgnoreTemplate.Hydrogen]: {
+      system: ['.DS_Store'],
+      logs: ['logs', '*.log', 'npm-debug.log*', 'yarn-debug.log*', 'yarn-error.log*'],
+      testing: ['/coverage', '*.lcov'],
+      dependencies: ['/node_modules', '.npm', '.yarn-integrity', '/.pnp', '.pnp.js'],
+      typescript: ['*.tsbuildinfo'],
+      environment: ['.env', '.env.test', '.env.local'],
+      production: ['/dist'],
+    },
+  }
```
pepicrft, line 60: `cli-kit` should have no knowledge on the dependent packages (e.g. Hydrogen). I recommend moving this logic to the Hydrogen framework and expose a function from here with with the base content that all NPM projects should have.

```ts
export async function getNPMProjectGitignore(): string
```

packages/cli-kit/src/git.ts
```diff
@@ -60,6 +110,42 @@ export async function downloadRepository({
...
 
+export async function getLatestCommit(directory?: string) {
+  try {
+    const logs = await git({baseDir: directory}).log({
+      maxCount: 1,
+    })
+    if (!logs.latest) throw NoCommitError()
+    return logs.latest
+  } catch {
+    throw NoCommitError()
```
pepicrft, line 121: Is this true? I'd let those errors bubble up and iterate on the catching logic here once we have more data.

packages/create-hydrogen/src/services/init.ts
```diff
@@ -164,6 +164,25 @@ async function init(options: InitOptions) {
...
+            // eslint-disable-next-line no-catch-all/no-catch-all
+          } catch (error) {
+            // We couldn't commit in already initialized git repo, maybe the commit author config is not set.
+            // In the future, we might supply our own committer like Ember CLI does, but for now, let's just
+            // remove the Git files to avoid a half-done state.
+            await cleanup(templateScaffoldDir)
+            task.title = 'Failed to initialize repository'
+            return
+          }
+          task.title = 'Git repository initialized'
```
pepicrft, line 183: What about embracing the "let it fail" mindset here and adding an acceptance test that runs on Windows, Linux, and Mac and gives us the confidence we need?

packages/cli-kit/src/git.test.ts
```diff
@@ -50,37 +64,165 @@ describe('downloadRepository()', () => {
...
     vi.mocked(hasGit).mockResolvedValue(true)
 
     // Then
-    await expect(ensurePresentOrAbort()).resolves.toBeUndefined()
+    await expect(git.ensurePresentOrAbort()).resolves.toBeUndefined()
+  })
+})
+
+describe('ensureInsideGitDirectory()', () => {
+  it('throws an error if not inside a git directory', async () => {
```
pepicrft, line 201: I'd recommend doing `expect(async () => { await subjectFunction() })` so that the asynchronous function logic is executed by Vitest when the test runs and not when the test is declared. This comment applies to all the changes in this file.

packages/cli-hydrogen/src/cli/services/deploy/config.ts
```diff
@@ -0,0 +1,39 @@
...
+}
+
+export const initializeGit = async (config: DeployConfig) => {
+  const shouldGitInit = await gitInit()
+
+  if (!shouldGitInit) throw new error.AbortSilent()
+
+  await git.initializeRepository(config.path)
+  git.createGitIgnore(config.path, git.GitIgnoreTemplate.Hydrogen)
+  await git.commitAll('Initial commit generated by deploy-hydrogen', {directory: config.path})
```
gfscott, line 22: I don't think we need to be so specific about which script did it, I think "Hydrogen", the entity as a whole, is the active party here.

```suggestion
  await git.commitAll('Initial commit generated by Hydrogen', {directory: config.path})
```

packages/create-hydrogen/src/services/init.ts
```diff
@@ -164,6 +164,25 @@ async function init(options: InitOptions) {
           task.title = 'Completed clean up'
         },
       },
+      {
+        title: 'Initializing a Git repository...',
+        task: async (_, task) => {
+          await git.initializeRepository(templateScaffoldDir)
+          git.createGitIgnore(templateScaffoldDir, git.GitIgnoreTemplate.Hydrogen)
+          try {
+            await git.commitAll('Initial commit generated by create-hydrogen', {directory: templateScaffoldDir})
```
gfscott, line 173:
```suggestion
            await git.commitAll('Initial commit generated by Hydrogen', {directory: templateScaffoldDir})
```

packages/cli-kit/src/git.ts
```diff
@@ -12,12 +13,61 @@ export const GitNotPresentError = () => {
   )
 }
 
+export const OutsideGitDirectoryError = () => {
+  return new Abort('Must be inside a Git directory to continue')
+}
+
+export const NoCommitError = () => {
+  return new Abort(
+    'Must have at least one commit to run command',
```
gfscott, line 22: Is this Git helper generic? Hence just `run command`? If it's specific to the deploy command, then let's use the actual command:

```suggestion
    'Git repo needs at least one commit to deploy.',
```

packages/cli-kit/src/git.ts
```diff
@@ -12,12 +13,61 @@ export const GitNotPresentError = () => {
...
+    content`Run ${token.genericShellCommand("git commit -m 'Initial commit'")} to create your first commit.`,
```
gfscott, line 23: Should we tell them to `git add` first?

packages/cli-kit/src/git.ts
```diff
@@ -12,12 +13,61 @@ export const GitNotPresentError = () => {
...
+  )
+}
+
+export const DetachedHeadError = () => {
+  return new Abort(
+    "Git HEAD can't be detached to run command",
```
gfscott, line 29:
```suggestion
    "Git HEAD can't be detached to deploy.",
```

gfscott (commented):
Looking good!

A few questions, about Emoji usage, the retry backoff threshold, and `git add`.

Also, I'd ask you to double-check against the [reserved flags list](https://shopify-cli-next.docs.shopify.io/kit/naming-conventions) just to make sure none of the commands or flag short forms conflict. I don't think they do, but single-letter flag short-forms are valuable real estate, changing them is a breaking change, and I just want to make sure we're choosing them well.

packages/cli-kit/src/constants.ts
```diff
@@ -17,6 +17,7 @@ const constants = {
     partnersEnv: 'SHOPIFY_PARTNERS_ENV',
     shopifyEnv: 'SHOPIFY_SHOPIFY_ENV',
     identityEnv: 'SHOPIFY_IDENTITY_ENV',
+    dmsEnv: 'SHOPIFY_DMS_ENV',
```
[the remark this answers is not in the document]
lynchv: deployment management system

packages/cli-hydrogen/src/cli/services/deploy.ts
```diff
@@ -0,0 +1,96 @@
...
+  const tasks: ui.ListrTask<TaskContext>[] = [
+    {
+      title: '📝 Getting deployment config',
+      task: async (ctx, task) => {
+        ctx.config = await fillDeployConfig(_config)
+        task.title = '📝 Deployment config parsed'
+      },
+    },
+    {
+      title: '✨ Creating a deployment',
```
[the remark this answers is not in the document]
lynchv: After some discussion, we agreed to let emoji as they are now. May revisit in the future.

packages/cli-hydrogen/src/cli/services/deploy.ts
```diff
@@ -0,0 +1,96 @@
...
+      },
+      retry: 2,
+    },
+    {
+      title: '📡 Checking deployment health',
+      task: async (ctx, task) => {
+        const retryCount = task.isRetrying()?.count
+
+        if (retryCount === backoffPolicy.length) {
+          task.title = "🤕 The deployment was uploaded but can't be reached yet."
```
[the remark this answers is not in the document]
gfscott: So after talking this over, I have two opinions:
1. We should be explicit about our timeout window — if it's two minutes but the deployment succeeds at 2m01s that's relevant information to the user
2. The message needs to propose some sort of action.

So something like:
```suggestion
          task.title = "The deployment uploaded but hasn’t become reachable within 2 minutes. Check the preview URL to see if deployment succeeded. If it didn’t, then check {what can they check? GitHub actions?}"
```

packages/cli-kit/src/git.ts
```diff
@@ -69,3 +155,13 @@ export async function ensurePresentOrAbort() {
     throw GitNotPresentError()
   }
 }
+
+/**
+ * If command run from outside a .git directory tree
+ * it throws an abort error.
+ */
+export async function ensureInsideGitDirectory(directory?: string) {
```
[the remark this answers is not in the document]
lynchv: For our use case it seems repetitive for nothing, seems like invoking `ensureInsideGitDirectory` once and then firing all our other git commands is preferable. On the other hand, we would be re-calling `ensureInsideGitDirectory` multiple times for no good reason.

packages/cli-kit/src/git.ts
amcaplan: @gfscott it's in CLI kit so should stay generic. The caller may, of course, raise a new error if desired.

packages/cli-kit/src/git.ts
```diff
@@ -60,6 +110,42 @@ export async function downloadRepository({
...
+  }
+}
+
+export async function commitAll(message: string, options?: {directory?: string; author?: string}) {
+  const simpleGit = git({baseDir: options?.directory})
+  const status = await simpleGit.status()
+
+  if (!status.files.length) throw NothingToCommitError()
+
+  await simpleGit.add(status.files.map((file) => file.path))
```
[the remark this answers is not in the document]
amcaplan: Given there is a difference, I think `commitAll` is a misnomer, as I'd expect it to do the equivalent of `git commit --all`.

Instead, maybe make 2 functions:

* `addAll` which runs `git add --all`
* `commit` which runs - you guessed it! - `git commit`

As an aside, I think that would leave each function with a clear single responsibility, and it would also save us from running in Node logic that should really just be a flag to a git command.

packages/cli-hydrogen/src/cli/services/deploy.ts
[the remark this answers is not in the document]
lynchv: I would simply add `...If it didn’t, then try again later.` @tempor1s not much else they can do for now?

packages/cli-hydrogen/src/cli/prompts/git-init.ts
```diff
@@ -0,0 +1,18 @@
+import {ui} from '@shopify/cli-kit'
+
+export const gitInit = async (): Promise<boolean> => {
+  const question: ui.Question = {
+    type: 'select',
```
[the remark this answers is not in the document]
lynchv: After quickly looking into it, it seems to be a bit more work than I thought. Notably needing to add a custom inquirer renderer, will open a ticket on our side to do this at a latter stage.

packages/create-hydrogen/src/services/init.ts
lynchv: Agree on the "let it fail" mindset. On the other hand, we already have `cli-hydrogen/src/cli/services/deploy/config.integration.test.ts` which essentially test the same logic as the step above. I'm leaning on the side that this is sufficient but I could also add another similar integration test for the create-hydrogen package?

packages/cli-kit/src/api.ts
```diff
@@ -2,5 +2,6 @@ import * as admin from './api/admin.js'
 import * as partners from './api/partners.js'
 import * as graphql from './api/graphql/index.js'
 import * as identity from './api/identity.js'
+import {buildHeaders} from './api/common.js'
 
-export {admin, partners, graphql, identity}
+export {admin, partners, graphql, identity, buildHeaders}
```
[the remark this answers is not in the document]
tempor1s: Thanks for bringing this up, and I think you are right. I am going to make a PR that will be a followup to this one that extracts these DMS calls into a seperate service that makes the calls more generic (and removes the need to export `buildHeaders`).

packages/cli-kit/src/git.ts
lynchv: I did not do so since git itself will tell you if it has nothing to commit and provide solutions.

packages/cli-hydrogen/src/cli/services/build.ts
```diff
@@ -3,50 +3,55 @@ import {ui, environment, error as kitError} from '@shopify/cli-kit'
...
+      task: async (_, task) => {
+        if (key === 'worker') {
+          process.env.WORKER = 'true'
+        }
+        if (assetBaseURL) {
+          process.env.HYDROGEN_ASSET_BASE_URL = assetBaseURL
+        }
+
+        try {
+          await viteBuild({
```
jgodson, line 30: Mentioned in slack too, but thought I'd bring it up for discussion here. Thoughts on accepting a `buildCommand` similar to `oxygenctl` or maybe just running `yarn build` in general? Opens up a bit of flexibility for possibly projects that didn't use vite (which I think is a [possibility](https://github.com/Shopify/remix-oxygen/blob/main/package.json))

packages/cli-hydrogen/src/cli/services/deploy/config.ts
```diff
@@ -0,0 +1,50 @@
...
+
+const MINIMAL_GIT_IGNORE: git.GitIgnoreTemplate = {
+  system: ['.DS_Store'],
+  logs: ['logs', '*.log', 'npm-debug.log*', 'yarn-debug.log*', 'yarn-error.log*'],
+  testing: ['/coverage', '*.lcov'],
+  dependencies: ['/node_modules', '.npm', '.yarn-integrity', '/.pnp', '.pnp.js'],
+  typescript: ['*.tsbuildinfo'],
+  environment: ['.env', '.env.test', '.env.local'],
+  production: ['/dist'],
+}
```
lordofthecactus, line 13: Reminder note: We mentioned in our meeting yesterday that our hydrogen templates already bring a `.gitignore`. Custom templates by the community might not. cc @blittle @wizardlyhel

.changeset/lovely-foxes-bow.md
```diff
@@ -0,0 +1,14 @@
+---
+'@shopify/cli-kit': minor
+'@shopify/create-hydrogen': minor
```
lordofthecactus, line 3: Not that aware of the CLI versioning but since there is a git requirement now. should this be a `major`? if `git` was already required by the CLI then should be fine a `minor`

lordofthecactus (commented):
👋  I did a bit of 🎩 top hatting trying to create hydrogen repo and got some errors. I ran:
```
dev create-hydrogen --path ../cli-tests
```
Errors:
https://screenshot.click/09-43-teca7-2ubyv.png
https://screenshot.click/09-43-dn8lo-lyrk0.png

Without the path flag, it works.
cc @blittle @cartogram

.changeset/lovely-foxes-bow.md
lynchv: Git was already a requirement.

packages/cli-hydrogen/src/cli/services/deploy/config.ts
lynchv: Just to clarify, for this particular PR we are generating the `.gitignore` file in 2 different way.

1. During `npm init @shopify/hydrogen`: We simply rename the `_gitignore` file to `.gitignore`
2. During `shopify hydrogen deploy ...`: We generate a generic strip down version of `.gitignore`. As seen above.

We built an [issue](https://github.com/Shopify/oxygen-platform/issues/570) to go over a better implementation in the future.

packages/cli-hydrogen/src/cli/services/build.ts
lynchv: I think it's a good idea to essentially let developers deploy "anything" with the command. This can be achieved in a couple different ways that will be documented in this [issue](https://github.com/Shopify/oxygen-platform/issues/571). For this particular PR we will move forward with the assumption that only Hydrogen is being used.

packages/cli-hydrogen/src/cli/services/deploy.ts
```diff
@@ -0,0 +1,97 @@
...
+        task.title = '📝 Deployment config parsed'
+      },
+    },
+    {
+      title: '✨ Creating a deployment',
+      task: async (ctx, task) => {
+        const {deploymentID, assetBaseURL} = await createDeployment(ctx.config)
+        ctx.assetBaseURL = assetBaseURL
+        ctx.deploymentID = deploymentID
+        task.title = '✨ Deployment created'
```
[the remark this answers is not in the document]
tempor1s: Yeah, its just a bit tricky to convey this since we have to "create" the deployment on DMS before we upload code to it, but want to provide that feedback to the user incase it fails (or needs to retry, etc)

packages/cli-hydrogen/src/cli/services/deploy/graphql/upload_deployment.ts
```diff
@@ -0,0 +1,11 @@
+import {gql} from 'graphql-request'
+
+export const UploadDeploymentQuery = gql`
+  mutation uploadDeployment($file: Upload!, $deploymentID: ID!) {
+    uploadDeployment(file: $file, deploymentID: $deploymentID) {
+      deployment {
```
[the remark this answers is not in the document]
tempor1s: I request it in [this](https://github.com/Shopify/cli/pull/443) subsequent PR that does error management. We wanted to get this PR out into main first, but I guess we could also just merge #443 and get re-reviews.

packages/cli-hydrogen/src/cli/services/deploy/upload.ts
```diff
@@ -0,0 +1,79 @@
...
+
+  const response: CreateDeploymentQuerySchema = await client.request(CreateDeploymentQuery, variables)
+  return response.createDeployment
+}
+
+export const uploadDeployment = async (config: ReqDeployConfig, deploymentID: string): Promise<string> => {
+  let headers = await api.buildHeaders(config.deploymentToken)
+
+  // note: may need validation for invalid deploymentID? oxygenctl does it
+  // note: we may want to remove the zip that we create in in this step
```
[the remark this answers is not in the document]
tempor1s: Cleaned up in [this PR](https://github.com/Shopify/cli/pull/443)

packages/cli-hydrogen/src/cli/commands/hydrogen/deploy.ts
```diff
@@ -0,0 +1,51 @@
...
+    commitAuthor: Flags.string({
+      env: 'SHOPIFY_HYDROGEN_FLAG_COMMIT_AUTHOR',
+      description: 'Override the default Git commit author.',
+    }),
+    healthCheck: Flags.boolean({
+      env: 'SHOPIFY_HYDROGEN_FLAG_HEALTH_CHECK',
+      default: true,
+      description: 'Require a health check before the deployment succeeds.',
+    }),
+    dmsAddress: Flags.string({
```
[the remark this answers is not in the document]
tempor1s: In #443 where we create a separate "api service" to call DMS I changed it to be "Oxygen" as well as I think it made more sense (and we also don't need to change the name when we switch over to targeting `oxygen-hub`). I think keeping this flag for now is fine since it's hidden and descriptive of what it is actually doing.

packages/cli-hydrogen/src/cli/services/deploy/graphql/upload_deployment.ts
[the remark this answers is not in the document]
graygilmore: No problem if it's coming in the follow-up!

packages/cli-hydrogen/src/cli/services/deploy.ts
```diff
@@ -0,0 +1,97 @@
...
+  const tasks: ui.ListrTask<TaskContext>[] = [
+    {
+      title: '📝 Getting deployment config',
+      task: async (ctx, task) => {
+        ctx.config = await fillDeployConfig(_config)
+        task.title = '📝 Deployment config parsed'
+      },
+    },
+    {
+      title: '✨ Creating a deployment',
```
gfscott, line 31:
```suggestion
      title: '💡 Initializing deployment',
```
I'm going with [initialize](https://www.merriam-webster.com/dictionary/initialize) over [initiate](https://www.merriam-webster.com/dictionary/initiate), I think it's the more specific meaning. Also we should drop the article here.

Alternate emoji ideas: 🌱 🧩 ⚡

packages/cli-hydrogen/src/cli/services/deploy.ts
```diff
@@ -0,0 +1,97 @@
...
+
+        if (retryCount === backoffPolicy.length) {
+          task.title =
+            "The deployment uploaded but hasn't become reachable within 2 minutes. Check the preview URL to see if deployment succeeded. If it didn't, then try again later."
+          return
+        }
+        if (retryCount && !isUnitTest) await system.sleep(backoffPolicy[retryCount - 1]!)
+
+        await healthCheck(ctx.previewURL)
+        task.title = '✅ Deployed and healthy!'
```
gfscott, line 82: Let's drop the exclamation mark here and make the success message a little more generic:
```suggestion
        task.title = '✅ Deployed successfully'
```

packages/cli-hydrogen/src/cli/commands/hydrogen/deploy.ts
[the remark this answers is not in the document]
tempor1s: After some discussion over Slack I decided that it makes sense to use "Oxygen" language here as it will make our transition over to using `oxygen-hub` in the future easier and it also matches the new service name in #443

.changeset/lovely-foxes-bow.md
```diff
@@ -0,0 +1,14 @@
...
+'@shopify/cli-hydrogen': patch
+---
+
+- cli-hydrogen:
+  - add support for asset URL rewriting when doing `shopify hydrogen build`
+- cli-kit:
+  - expose additional git functionality for commits, creating .gitignore, and insuring command is run in a git directory
+  - add additional 'dms' service type
+  - add additional exposed api & file functionality
```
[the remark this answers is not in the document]
tempor1s: Updated the content of this entire file to use full sentences & better describe the public changes that are being made.

.changeset/lovely-foxes-bow.md
```diff
@@ -0,0 +1,14 @@
...
+- create-hydrogen
+  - add support for initializing a new local git repository when you initialize a new project
```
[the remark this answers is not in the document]
tempor1s: Updated the content of this entire file to use full sentences & better describe the public changes that are being made.

graygilmore (approved):
Thanks for addressing all of that feedback. Looking forward to the iterations on this!

lynchv:
### Force Push Notes

Rebasing with main, fix merge conflict with `packages/cli-kit/src/ui.ts`

packages/create-hydrogen/src/services/init.ts
```diff
@@ -164,6 +164,16 @@ async function init(options: InitOptions) {
           task.title = 'Completed clean up'
         },
       },
+      {
+        title: 'Initializing a Git repository...',
+        task: async (_, task) => {
+          await file.move(`${templateScaffoldDir}/_gitignore`, `${templateScaffoldDir}/.gitignore`)
```
cartogram, line 170: I was under the impression that we were using a consistent template here? Is there a reason we can't create the gitignore? If not, we should remove this line because it is duplicating functionality from the hydrogen repo.

https://github.com/Shopify/hydrogen/blob/v1.x-2022-07/scripts/compile-template.mjs#L82

lynchv:
### Force push notes:

- Fixed PR check on config.integration.ts by adding new flag `assumeYes`.

PR #408, opened 2022-09-01
Fix store and Bugsnag initialization for Homebrew

gonzaloriestra:

The init hook for the theme package is not working correctly when installed through Homebrew, causing the CLIKitStore to fail to initialize:

Screen Shot 2022-08-31 at 6 11 49 PM

The Homebrew installation is installing two Shopify libraries under node_modules: cli and theme, both of them containing a copy of cli-kit. And for some reason, the current init hook (`@shopify/cli-kit/node/hooks/init`) is calling the file from the cli package instead of theme.

01-40-pn8c0-qvqz8

- Remove the explicit initialization at the start for CLIKitStore and Bugsnag, and initialize them when used (only the first time). Both errors from the screenshot above should be gone.
- Remove `--store` from the flags passed to CLI2 for the pull command, as it's not required and throws an error now (`The option --store is not supported.`)
- Remove the RubyGems validation, as I was getting an error with the Homebrew installation, but we are not using it directly and [I think there's no need to validate it](https://github.com/Shopify/shopify-cli-next/pull/498#discussion_r894610664).

### How to test your changes?

Development:
- `yarn shopify theme pull --store wadus`
- Tweak the code to throw an error, with debug mode disabled, and check if it was sent to Bugsnag ([already tested](https://app.bugsnag.com/shopify/cli/errors/6311dd43e1a0a20008e8b9c0?event_id=6311dd430096490fb8ff0000&i=sk&m=nw))

Hombrew:
- `brew update`
- `brew install shopify-cli@3`
- Go to the CLI repo and run yarn build
- Update the changed dist files from this PR in `/opt/homebrew/Cellar/shopify-cli@3/3.9.0/libexec/lib/node_modules/@shopify`
`shopify3 theme pull --store wadus`

### Measuring impact

### Checklist

### DIFF
--- a/packages/cli-kit/src/node/error-handler.ts
+++ b/packages/cli-kit/src/node/error-handler.ts
@@ -11,6 +11,7 @@ import {getEnvironmentData, reportEvent} from '../analytics.js'
 import * as path from '../path.js'
 import * as metadata from '../metadata.js'
 import {fanoutHooks} from '../plugins.js'
+import constants, {bugsnagApiKey} from '../constants.js'
 import {settings, Interfaces} from '@oclif/core'
 import StackTracey from 'stacktracey'
 import Bugsnag, {Event} from '@bugsnag/js'
@@ -88,6 +89,7 @@ export async function sendErrorToBugsnag(
   reportableError.stack = `Error: ${reportableError.message}\n${formattedStacktrace}`
 
   if (report) {
+    await initializeBugsnag()
     await new Promise((resolve, reject) => {
       Bugsnag.notify(reportableError, undefined, (error, event) => {
         if (error) {
@@ -142,6 +144,7 @@ export async function registerCleanBugsnagErrorsFromWithinPlugins(config: Interf
       return {name: plugin.name, pluginPath: path.normalize(followSymlinks)}
     }),
   )
+  await initializeBugsnag()
   Bugsnag.addOnError(async (event) => {
     event.errors.forEach((error) => {
       error.stacktrace.forEach((stackFrame) => {
@@ -203,5 +206,19 @@ export async function addBugsnagMetadata(event: Event, config: Interfaces.Config
   }
   Object.entries(bugsnagMetadata).forEach(([section, values]) => {
     event.addMetadata(section, values)
+  })
+}
+
+async function initializeBugsnag() {
+  if (Bugsnag.isStarted()) {
+    return
+  }
+  Bugsnag.start({
+    appType: 'node',
+    apiKey: bugsnagApiKey,
+    logger: null,
+    appVersion: await constants.versions.cliKit(),
+    autoTrackSessions: false,
+    autoDetectErrors: false,
   })
 }
--- a/packages/cli-kit/src/store.ts
+++ b/packages/cli-kit/src/store.ts
@@ -1,6 +1,4 @@
 import {content, token, debug} from './output.js'
-import constants from './constants.js'
-import {Abort} from './error.js'
 import Conf, {Schema} from 'conf'
 
 const migrations = {}
@@ -44,21 +42,13 @@ let _instance: CLIKitStore | undefined
 
 export function cliKitStore() {
   if (!_instance) {
-    throw new Abort("The CLIKitStore instance hasn't been initialized")
-  }
-  return _instance
-}
-
-export async function initializeCliKitStore() {
-  if (!_instance) {
-    // eslint-disable-next-line require-atomic-updates
     _instance = new CLIKitStore({
       schema,
       migrations,
       projectName: 'shopify-cli-kit',
-      projectVersion: await constants.versions.cliKit(),
     })
   }
+  return _instance
 }
 
 export class CLIKitStore extends Conf<ConfSchema> {
--- a/packages/theme/package.json
+++ b/packages/theme/package.json
@@ -50,7 +50,7 @@
   "oclif": {
     "commands": "dist/cli/commands",
     "hooks": {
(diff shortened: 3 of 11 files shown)


packages/theme/src/cli/commands/theme/pull.ts
```diff
@@ -59,7 +59,7 @@ export default class Pull extends ThemeCommand {
       validPath = path.resolve(flags.path)
     }
 
-    const flagsToPass = this.passThroughFlags(flags, {exclude: ['path', 'verbose']})
+    const flagsToPass = this.passThroughFlags(flags, {exclude: ['path', 'verbose', 'store']})
```
amcaplan, line 62: Excellent catch, thanks!

amcaplan (approved):

packages/theme/package.json
```diff
@@ -50,7 +50,7 @@
   "oclif": {
     "commands": "dist/cli/commands",
     "hooks": {
-      "init": "@shopify/cli-kit/node/hooks/init"
+      "init": "@shopify/theme/node_modules/@shopify/cli-kit/dist/node/hooks/init.js"
```
pepicrft, line 53: This assumes a certain structure and might break if you install the CLI globally through a package manager. For example Yarn might end up organizing the packages like:

```
@shopify/
  theme
  cli-main
  theme
```

pepicrft (changes requested):
Let me think of a solution that it's not coupled to some structure of modules.

packages/theme/package.json
gonzaloriestra: Yeah, I didn't like this solution, but I couldn't find a better one...

pepicrft: What about moving this to the `cli-main` package? That way we ensure there's only one call.

gonzaloriestra: You mean moving the hooks from cli-kit, right? That could work, let me try...

pepicrft: The implementation of the hooks remain in the `cli-kit` module, but we remove [this](https://github.com/Shopify/cli/blob/main/packages/theme/package.json#L53) across all the packages and place it in the `package.json` of `cli-main`.

pepicrft: I noticed we have some inconsistencies already, for example [cli-hydrogen](https://github.com/Shopify/cli/blob/main/packages/cli-hydrogen/package.json) doesn't have the hook define. So I think `cli-main` is a more suitable place for it.

gonzaloriestra: Ah, that won't work... The hook from cli-main it's already being called (as you can see in [this log](https://user-images.githubusercontent.com/14979109/187916269-64ddb089-2323-41be-9e26-0064ff07fc76.png)). The problem is that the cli-kit is duplicated across both packages (theme and cli), so it's initializing the store in the wrong package. I can show you this in a pair session if you want.

And the hooks should stay in `cli-kit` because they are also used from `create-app` or `create-hydrogen`, which don't depend on `cli-main`.

packages/cli-kit/src/node/error-handler.ts
```diff
@@ -205,3 +208,17 @@ export async function addBugsnagMetadata(event: Event, config: Interfaces.Config
     event.addMetadata(section, values)
   })
 }
+
+async function initializeBugsnag() {
+  if (Bugsnag.isStarted()) {
```
gonzaloriestra, line 213: Source: https://github.com/bugsnag/bugsnag-js/blob/v7.16.7/packages/node/src/notifier.js#L75

packages/cli-kit/src/store.ts
```diff
@@ -44,21 +42,13 @@ let _instance: CLIKitStore | undefined
...
-}
-
-export async function initializeCliKitStore() {
-  if (!_instance) {
-    // eslint-disable-next-line require-atomic-updates
     _instance = new CLIKitStore({
       schema,
       migrations,
       projectName: 'shopify-cli-kit',
-      projectVersion: await constants.versions.cliKit(),
```
gonzaloriestra, line 59: I removed the projectVersion as the [default value](https://github.com/sindresorhus/conf/tree/v10.1.2#projectversion) works for us (the version from the closest package.json). This way, there's no need to make the method async.

gonzaloriestra (commented):
After talking to @pepicrft, I've changed the approach to remove the explicit initializations of the store and Bugsnag instead of trying to fix the modules/hooks mess.

packages/cli-kit/src/node/error-handler.ts
```diff
@@ -205,3 +208,17 @@ export async function addBugsnagMetadata(event: Event, config: Interfaces.Config
     event.addMetadata(section, values)
   })
 }
+
+async function initializeBugsnag() {
```
amcaplan, line 212: Does this need to be async?

packages/cli-kit/src/store.ts
```diff
@@ -44,21 +42,13 @@ let _instance: CLIKitStore | undefined
...
   if (!_instance) {
-    throw new Abort("The CLIKitStore instance hasn't been initialized")
-  }
-  return _instance
-}
-
-export async function initializeCliKitStore() {
-  if (!_instance) {
-    // eslint-disable-next-line require-atomic-updates
     _instance = new CLIKitStore({
```
amcaplan, line 45: I assume there's no issue if a race conditions causes this to be called twice, right?

amcaplan (approved):
Love this implementation! Very elegant.

Left a couple of minor comments, nothing really game-changing.

pepicrft (approved):

Issue #317, opened 2022-08-18
When using the CLI flow for an existing application, you can only select from the first five pages of apps in your partner dashboard


greenlemur:
Problem:

When using the CLI flow for an existing application, you can only select from the first five pages of apps in your partner dashboard

<img width="690" alt="Screen Shot 2022-08-18 at 1 26 46 PM" src="https://user-images.githubusercontent.com/60752324/185460308-7412c5fb-5301-41d7-aab5-4cbfd31ab221.png">

^ You can only go as far down as the first five pages of apps in your partner dashboard. Anything beyond that will not show up in the terminal as an option to choose from. We have 11 pages of apps in our partner account (partner ID: 1362879).

amcaplan:
@alvaro-shopify I think this is a legit issue. [We only retrieve the first 100 apps](https://github.com/Shopify/cli/blob/c54f708c59665ad4d1769a2422c2d47bdaa41997/packages/cli-kit/src/api/graphql/find_org.ts#L11) for the organization, there's no pagination beyond that.

amcaplan:
@greenlemur thanks for bringing this case to our attention. Until it's solved, I believe there is a workaround, at least for `dev`. If you open your app's page in the admin, you'll see the API key. Copy that key, and pass it to the command using the flag `--api-key=YOURAPIKEY`.

There is no such flag for `deploy` but we've considered adding once, here's another reason to push in that direction. /cc @isaacroldan

isaacroldan:
Yes, using the `--api-key` flag is the best (or only) workaround for this. @amcaplan not sure why we don't have this flag for deploy but I think we should! 
Maybe create a ticket for the next cycle?

amcaplan:
@greenlemur I know it isn't the best solution, but for now note that support for the `--api-key` flag in `deploy` has been [merged](https://github.com/Shopify/cli/pull/412) and released in the latest version, 3.10.0. So you have a solid workaround for now.

PR #412, opened 2022-09-01
Add `--api-key` flag to `shopify app deploy`

amcaplan:

We have it for app dev. It makes sense to add it here, too.

Also serves as a critical workaround in [1 known case](https://github.com/Shopify/cli/issues/317#issuecomment-1232143625).

And eventually will allow users to automate deploys to multiple Partner admin apps from the same codebase.

Adds a flag `--api-key` to `shopify app deploy`. The implementation is nearly identical to the same flag for `shopify app dev`.

### How to test your changes?

`shopify app deploy` without this flag. Then, `shopify app deploy --api-key=WHATEVER` to deploy to another app using its API key. Then `shopify app deploy` and note that your change sticks.

### Measuring impact

Would be good to check this is used, but I think down the road we'll see tons of use as we roll out more flag automation.

### Checklist

### DIFF
--- a/packages/app/src/cli/commands/app/deploy.ts
+++ b/packages/app/src/cli/commands/app/deploy.ts
@@ -12,6 +12,11 @@ export default class Deploy extends Command {
   static flags = {
     ...cli.globalFlags,
     ...appFlags,
+    'api-key': Flags.string({
+      hidden: false,
+      description: 'The API key of your app.',
+      env: 'SHOPIFY_FLAG_APP_API_KEY',
+    }),
     reset: Flags.boolean({
       hidden: false,
       description: 'Reset all your settings.',
@@ -24,6 +29,6 @@ export default class Deploy extends Command {
     const {args, flags} = await this.parse(Deploy)
     const directory = flags.path ? path.resolve(flags.path) : process.cwd()
     const app: AppInterface = await loadApp(directory)
-    await deploy({app, reset: flags.reset})
+    await deploy({app, apiKey: flags['api-key'], reset: flags.reset})
   }
 }
--- a/.changeset/polite-carpets-roll.md
+++ b/.changeset/polite-carpets-roll.md
@@ -1,0 +1,5 @@
+---
+'@shopify/app': minor
+---
+
+Add --api-key flag for app deploy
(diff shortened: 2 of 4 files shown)


packages/app/src/cli/commands/app/deploy.ts
```diff
@@ -12,6 +12,11 @@ export default class Deploy extends Command {
   static flags = {
     ...cli.globalFlags,
     ...appFlags,
+    'api-key': Flags.string({
+      hidden: false,
+      description: 'The API key of your app.',
+      env: 'SHOPIFY_FLAG_APP_API_KEY',
```
pepicrft, line 18: I'd make an exception here for the flag name and align with the name that we are using in `.env`, `SHOPIFY_API_KEY`. What do you think?

amcaplan: No strong feelings. But note this was copied from `app dev` so we should be consistent for both.

amcaplan: So here's an even weirder thing. I'm not even sure where we're using the value from dotenv. Because we seem to fetch the app info in this line: 

https://github.com/Shopify/cli/blob/1f1d9786b6bbab0ed6c56637cd04921be103f823/packages/app/src/cli/services/environment.ts#L132

And that's using cached info in CLI kit store, not the value from .env.

You can prove it by devving with 1 app, then changing the API key in the `.env` file to the key from somewhere else. The value in CLI kit store is used, and the value in `.env` is ignored.

@isaacroldan might be able to provide more context.

Regardless, I think one of the goals of the flag is to allow overriding ENV on a temporary basis. And on those grounds, I think it make sense to leave the flag name with a different ENV variable here, as is done in dev. And if we shorten to `SHOPIFY_API_KEY` in the future, we should update both.

isaacroldan: `dev` always uses the value from the CLI Kit store.
The value from the `.env` file is used only for `deploy` (we load it [here](https://github.com/Shopify/cli/blob/1f1d9786b6bbab0ed6c56637cd04921be103f823/packages/app/src/cli/services/environment.ts#L226))

Is confusing, but at the time it was the best solution to keep config for `dev` and `deploy` separated. With the presets this will definitely improve

PR #413, opened 2022-09-01
Add env_package_manager_workspaces to public metadata

matteodepalo:

Related to https://github.com/Shopify/shopify-cli-planning/issues/346

We want to be able to track wether an app uses workspaces or not.

This PR adds the `env_package_manager_workspaces` field to the public metadata.

A PR in the monorail repo that adds this field will need to be merged before this PR is merged.

I was debating naming this `app_package_manager_workspaces` instead of having the `env_` prefix, but then was convinced by @stephszeto that this feels like an "environment" property, even if we can know about it only when loading the app and not from the environment where commands are run itself. For pure environment variables we've been using the oclif `Config` object, but we can't really use that here.

### How to test your changes?

Not sure if we can test before deploying.

### Measuring impact

### Checklist


matteodepalo:
@gonzaloriestra let me know if this makes sense.

packages/app/src/cli/models/app/loader.ts
```diff
@@ -81,6 +87,8 @@ class AppLoader {
     const nodeDependencies = await getDependencies(packageJSONPath)
     const packageManager = await getPackageManager(this.appDirectory)
     const {webs, usedCustomLayout: usedCustomLayoutForWeb} = await this.loadWebs()
+    const pnpmWorkspacesPath = path.join(this.appDirectory, pnpmWorkspaceFile)
+    const usesWorkspaces = await appUsesWorkspaces(packageJSONPath, pnpmWorkspacesPath)
```
gonzaloriestra, line 91: What about passing just the app directory and generating `packageJSONPath` and `pnpmWorkspacesPath` inside? It would simplify a bit the usage and it would be consistent with other functions like `getPackageManager`.

packages/cli-kit/src/analytics.ts
```diff
@@ -124,10 +124,10 @@ export function getEnvironmentData(config: Interfaces.Config) {
   return {
     uname: `${platform} ${arch}`,
     env_ci: ciPlatform.isCI,
-    env_ci_platform: ciPlatform.name,
+    env_ci_platform: ciPlatform.name ?? '',
```
gonzaloriestra, line 127: If a field is undefined/null, I think Monorail ignores it. But this way, I guess they will be saved... Have you checked how it works? Probably not the case, but these changes may break a dashboard if we have queries like `where env_ci_platform IS NOT NULL`.

gonzaloriestra (changes requested):
It looks good! However, I think you forgot to add the new field to `analytics.buildPayload()`. You can easily test this by running `yarn shopify version --verbose`, which is not sending anything to Monorail in development, but it shows the payload.

packages/app/src/cli/models/app/loader.ts
matteodepalo: Good idea, I'll change it.

matteodepalo:
@gonzaloriestra I thought that because this new field is sent every time the app is loaded we don't need to add it there. The field should be sent in the `buildPayload` thanks to `...metadata.getAllPublic(),`.
`version` doesn't load the app so it doesn't show there.

The reason why I built it this way is that `environmentData` doesn't use the `appDirectory` in any way. I suppose we could also pass it inside, but I thought it would be more convenient to reuse the already existing `App` structure.

packages/cli-kit/src/analytics.ts
matteodepalo: I'll revert to what it was, not worth changing it.

matteodepalo:
This makes me think though that we might want to rename this flag to `app_package_manager_workspaces`. Workspaces is not really something that belongs to the environment developers are running commands in (e.g. shell, package manager), but depends directly on something that is saved inside the app.

packages/cli-kit/src/monorail.ts
```diff
@@ -76,6 +76,7 @@ export interface Schemas {
       // Environment
       env_ci?: Optional<boolean>
       env_ci_platform?: Optional<string>
+      env_package_manager_workspaces?: Optional<boolean>
```
pepicrft, line 79: I think we have to create a [new version of the schema](https://github.com/Shopify/monorail/blob/42884ff91db940665b4263277d30cccf04f1daa8/schemas/app_cli3_command/1.1.yml)

packages/cli-kit/src/node/node-package-manager.ts
```diff
@@ -183,6 +186,19 @@ export async function getDependencies(packageJsonPath: string): Promise<{[key: s
...
 }
 
+/**
+ * Returns true if the app uses workspaces, false otherwise.
+ * @param packageJsonPath {string} Path to the package.json file
+ * @param pnpmWorkspacePath {string} Path to the pnpm-workspace.yaml file
+ * @returns A promise that resolves with true if the app uses workspaces, false otherwise.
+ */
+
+export async function usesWorkspaces(appDirectory: string): Promise<boolean> {
```
pepicrft, line 196: What about adding a test to cover the two scenarios here?

packages/cli-kit/src/analytics.ts
```diff
@@ -98,7 +98,7 @@ const buildPayload = async ({config, errorMessage}: ReportEventOptions) => {
     },
     sensitive: {
       args: startArgs.join(' '),
-      error_message: errorMessage,
+      error_message: errorMessage!,
```
pepicrft, line 101: Is this safe?

pepicrft (approved):

packages/cli-kit/src/analytics.ts
matteodepalo: I'll remove this as well 👍 not needed.

packages/cli-kit/src/node/node-package-manager.ts
matteodepalo: I've added integration tests that cover it, but I'll also add unit tests as you suggested.

packages/cli-kit/src/monorail.ts
matteodepalo: Do we? I was under the impression that if we add new optional fields we don't need to bump the schema 🤔

pepicrft: Wouldn't that statement hold true for other attributes like [env_ci](https://github.com/Shopify/monorail/blob/42884ff91db940665b4263277d30cccf04f1daa8/schemas/app_cli3_command/1.1.yml#L259). Do you know @gonzaloriestra ?

gonzaloriestra: We don't need to bump a major version of the schema, but I think a minor one is required: https://production-platform.docs.shopify.io/understanding/monorail/schemas/intro#versioning-rules

matteodepalo:
I've created a PR for the new schema https://github.com/Shopify/monorail/pull/13589

alvaro-shopify (commented):
I think workspaces is not a string[] but { packages: string[]}

alvaro-shopify:
> @alvaro-shopify interesting, I've never seen it used like that. [In the yarn docs](https://classic.yarnpkg.com/lang/en/docs/workspaces/) it's just a plain array, but I suppose it can be used like that too?

Anyway, it seems that 

> @alvaro-shopify interesting, I've never seen it used like that. [In the yarn docs](https://classic.yarnpkg.com/lang/en/docs/workspaces/) it's just a plain array, but I suppose it can be used like that too?

Anyway, as your PR is only verifying that `workspaces` is present in the `package.json` then it works for both of the scenarios

alvaro-shopify (approved):

packages/app/src/cli/models/app/loader.ts
```diff
@@ -14,7 +14,12 @@ import {mapUIExternalExtensionTypeToUIExtensionType} from '../../utilities/exten
 import metadata from '../../metadata.js'
 import {error, file, id, path, schema, string, toml, output} from '@shopify/cli-kit'
 import {readAndParseDotEnv, DotEnvFile} from '@shopify/cli-kit/node/dot-env'
-import {getDependencies, getPackageManager, getPackageName} from '@shopify/cli-kit/node/node-package-manager'
+import {
+  getDependencies,
+  getPackageManager,
+  getPackageName,
+  usesWorkspaces as appUsesWorkspaces,
```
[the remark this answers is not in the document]
matteodepalo: It sounded a bit weird to say `getIs..`, I felt like for booleans we can change the convention. Not a big deal, happy to change it if you feel strongly about it.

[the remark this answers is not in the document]
alvaro-shopify: I agree with Matteo, the convention for boolean should be different. In this case I prefer `uses` instead of `is`

PR #364, opened 2022-08-25
add structured logging

alfonso-noriega:
### WHY are these changes introduced?

Fixes [#371](https://github.com/Shopify/shopify-cli-extensions/issues/371)
Enhance the extension server logging replacing the human readable messages by structured Json logs.
The problem is describe further in this [document](https://docs.google.com/document/d/1EaxKuiDOHuukkqayNr_Ot-7JcAwomjhDYnLcwfNuLko/edit#)

### WHAT is this pull request doing?
- Created a log structure with semantically descriptive context
- disable standard logging to use the new structured logs

### How to test your changes?
- create a new go binary
- run any of the supported flows using the binary (scaffold or dev)
- check the output logs

PR #415, opened 2022-09-01
Feature/support parsing structured logs

alfonso-noriega:

Fixes [#364](https://github.com/Shopify/cli/pull/364)
The Extension server is now logging structured logs and the CLI needs to be able to parse them in to the desired format.
The problem is describe further in this [document](https://docs.google.com/document/d/1EaxKuiDOHuukkqayNr_Ot-7JcAwomjhDYnLcwfNuLko/edit#)

Added Structured logging to the Extension server and parsing of the structured logging in the CLI

### How to test your changes?

- Create an App through the CLI `yarn create app --local`
- Go the the app folder and create an extension in the app `yarn scaffold extension`
- Serve the app&extension `yarn dev`
- check the output logs
- Introduce a breaking change in the extension template like changing an HTML tag and see the error log

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -36,7 +36,6 @@
   },
   "eslint.workingDirectories": [{ "pattern": "packages/*" }],
   "vitest.enable": true,
-  "testing.automaticallyOpenPeekView": "never",
   "javascript.preferences.importModuleSpecifierEnding": "js",
   "typescript.preferences.importModuleSpecifierEnding": "js",
 }
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -36,6 +36,7 @@
   },
   "eslint.workingDirectories": [{ "pattern": "packages/*" }],
   "vitest.enable": true,
+  "testing.automaticallyOpenPeekView": "never",
   "javascript.preferences.importModuleSpecifierEnding": "js",
   "typescript.preferences.importModuleSpecifierEnding": "js",
 }
--- a/packages/app/src/cli/utilities/extensions/cli.ts
+++ b/packages/app/src/cli/utilities/extensions/cli.ts
@@ -1,8 +1,9 @@
 import {getBinaryPathOrDownload} from './binary.js'
 import metadata from '../../metadata.js'
-import {environment, error, path, system} from '@shopify/cli-kit'
+import {environment, error, path, system, output as outputKit} from '@shopify/cli-kit'
 import {fileURLToPath} from 'url'
 import {platform} from 'node:os'
+import {Writable} from 'stream'
 
 const NodeExtensionsCLINotFoundError = () => {
   return new error.Bug(`Couldn't find the shopify-cli-extensions Node binary`)
@@ -16,7 +17,7 @@ const NodeExtensionsCLINotFoundError = () => {
  * @param options {system.ExecOptions} Options to configure the process execution.
  */
 export async function runGoExtensionsCLI(args: string[], options: system.WritableExecOptions = {}) {
-  const stdout = options.stdout || {write: () => {}}
+  const stdout: Writable = options.stdout || new Writable({write: () => {}})
   if (environment.local.isDevelopment()) {
     await metadata.addPublic(() => ({cmd_extensions_binary_from_source: true}))
     const extensionsGoCliDirectory = (await path.findUp('packages/ui-extensions-go-cli/', {
@@ -25,6 +26,11 @@ export async function runGoExtensionsCLI(args: string[], options: system.Writabl
     })) as string
 
     stdout.write(`Using extensions CLI from ${extensionsGoCliDirectory}`)
+
+    // eslint-disable-next-line no-warning-comments
+    // TODO: how to solve thiseslint problem
+    // eslint-disable-next-line require-atomic-updates
+    options.stdout = parseGoLogs(stdout)
     try {
       if (environment.local.isDebugGoBinary()) {
         await system.exec('sh', [path.join(extensionsGoCliDirectory, 'init-debug-session')].concat(args), options)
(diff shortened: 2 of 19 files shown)


alfonso-noriega:
This PR contains both, the go structured log implementation and the CLI integration

.vscode/settings.json
```diff
@@ -36,7 +36,6 @@
   },
   "eslint.workingDirectories": [{ "pattern": "packages/*" }],
   "vitest.enable": true,
-  "testing.automaticallyOpenPeekView": "never",
```
isaacroldan, line 39: what is this setting?

packages/app/src/cli/utilities/extensions/cli.ts
```diff
@@ -68,3 +84,56 @@ export async function nodeExtensionsCLIPath(): Promise<string> {
     return executablePath
   }
 }
+
+/**
+ * This method provides a Writable wraper which parses the incoming chunks as GoLog from Json and writes them back to
+ * the Writable given as parameter. If Parsing to Json fails it wirtes the chunk as it is and thows an error.
```
isaacroldan, line 90: writes

packages/app/src/cli/utilities/extensions/cli.ts
```diff
@@ -68,3 +84,56 @@ export async function nodeExtensionsCLIPath(): Promise<string> {
...
+      const lines = outputKit.stripAnsiEraseCursorEscapeCharacters(chunk.toString('ascii')).split(/###LOG_END###/)
+      for (const line of lines) {
+        if (line) {
+          try {
+            const log = JSON.parse(line) as GoLog
+            output.write(parseGoLogMessage(log))
+            // eslint-disable-next-line no-catch-all/no-catch-all
+          } catch (err) {
+            output.write(line)
+          }
```
isaacroldan, line 106: Does this work? your are parsing every line as a JSON, shouldn't you parse the whole "lines" object?

packages/app/src/cli/utilities/extensions/cli.ts
```diff
@@ -68,3 +84,56 @@ export async function nodeExtensionsCLIPath(): Promise<string> {
...
+  if (!log.level || !log.status || !log.payload.message) {
+    throw new Error(`Invalid log: ${log}`)
+  }
+  if (log.extensionName) {
+    return `${log.extensionName} ${styleWorkflowStep(log.workflowStep)} ${log.status}: ${styleContent(
+      log.payload.message,
+      log.status,
+    )}`
+  }
+  return `\x1b[94m[${log.workflowStep}]\x1b[0m ${log.status}: ${styleContent(log.payload.message, log.status)}`
```
isaacroldan, line 124: what are these characters? If you want to add color/text styles we have methods for that (search for `output.content` and `output.token`)

packages/app/src/cli/utilities/extensions/cli.ts
```diff
@@ -68,3 +84,56 @@ export async function nodeExtensionsCLIPath(): Promise<string> {
...
+/**
+ * This method provides a Writable wraper which parses the incoming chunks as GoLog from Json and writes them back to
+ * the Writable given as parameter. If Parsing to Json fails it wirtes the chunk as it is and thows an error.
+ * @param output {Writable} A Writable which will be wrapped and in which the parsed chunks will be written.
+ * @returns {Writable} A Writable wrapping the parameter.
+ */
+export function goLogWritable(output: Writable): Writable {
+  return new Writable({
+    write(chunk, _encoding, next) {
+      const lines = outputKit.stripAnsiEraseCursorEscapeCharacters(chunk.toString('ascii')).split(/###LOG_END###/)
```
isaacroldan, line 97: have you tried with different logs? do you get chunks that include only part of a log or more than 1 json?
Wondering if the LOG_END is necessary here

isaacroldan: Nevermind, 🎩 'ed and added some logs and saw that yes you can receive multiple jsons in a chunk

packages/ui-extensions-go-cli/api/api.go
```diff
@@ -127,14 +130,15 @@ func getJSONRawMessage(data interface{}) (result json.RawMessage, err error) {
 	return
 }
 
-func interfaceToMap(data interface{}) (result map[string]interface{}, err error) {
+func interfaceToMap(data interface{}, logBuilder logging.LogEntryBuilder) (result map[string]interface{}, err error) {
+	logBuilder.SetStatus(logging.Failure)
```
alvaro-shopify, line 134: Does it make sense to setStatus to `Failure` in case no error is produced?

packages/ui-extensions-go-cli/api/api.go
```diff
@@ -127,14 +130,15 @@ func getJSONRawMessage(data interface{}) (result json.RawMessage, err error) {
 	return
 }
 
-func interfaceToMap(data interface{}) (result map[string]interface{}, err error) {
+func interfaceToMap(data interface{}, logBuilder logging.LogEntryBuilder) (result map[string]interface{}, err error) {
```
alvaro-shopify, line 133: Do this kind of utility methods knows about logEntryBuilder?, maybe different error types could be returned so the caller can logged properly. This suggestion could avoid adding new parameters to the methods

packages/ui-extensions-go-cli/main.go
```diff
@@ -87,36 +107,57 @@ func (cli *CLI) build(args ...string) {
...
 
 		go build.WatchLocalization(ctx, extension, func(result build.Result) {
+			logWatchLocalizationBuilder := logBuilder.AddWorkflowSteps(logging.Build, logging.Watch, logging.Localization)
+			logWatchLocalizationBuilder.SetExtensionId(result.Extension.UUID)
+			logWatchLocalizationBuilder.SetExtensionName(result.Extension.Title)
 			if result.Success {
-				api.Notify([]core.Extension{result.Extension})
-				fmt.Println(result)
+				logWatchLocalizationBuilder.SetStatus(logging.Success)
+				logCompleted := logWatchLocalizationBuilder.Build(result.Message)
```
alvaro-shopify, line 154: Why the new variable `logCompleted` is needed?

MeredithCastile:
This is valuable. Thanks for taking on this work, @alfonso-noriega.

### Copy updates

One small optimization would be to consistently capitalize the first line of the success/error messages and to eliminate the word "successfully" because it's already implied in the string.

The formula could be:
- "Configuration loaded"
- "Localization built"
- "Build complete"
- <noun> <past participle>

image

### Error question

When I look at the error string printed inline, I'm confused by the beginning of the messaging:

image

 Would a developer get meaning from this?
 
___

 cc: @grahamscott, @mynameisadamf, @matteodepalo, since this is a pattern that we may want to tokenize or write guidelines around.

PR #416, opened 2022-09-01
Load the `@oclif/plugin-plugins` plugin in `cli-main`

pepicrft:
I noticed we are not currently loading the `@oclif/plugin-plugins` plugin in `cli-main` and therefore CLI users don't have access to the `shopify plugins` API for managing their plugins. Is there are a reason for it?

I'm adding the plugin to the list of plugins that get loaded with the CLI.

### How to test your changes?

`yarn shopify plugins` should show the help.

### Measuring impact

### Checklist

### DIFF
--- a/packages/cli-main/package.json
+++ b/packages/cli-main/package.json
@@ -77,7 +77,8 @@
       "@shopify/theme",
       "@shopify/cli-hydrogen",
       "@oclif/plugin-help",
-      "@shopify/plugin-ngrok"
+      "@shopify/plugin-ngrok",
+      "@oclif/plugin-plugins"
     ],
     "scope": "shopify",
     "topicSeparator": " ",


isaacroldan:
I think we didn't want to expose this until we "released" plugins 🤔 
Was going to add it as part of the tunneling plugins work, but i guess we can go with it already

isaacroldan:
I think we didn't want to expose this until we "released" plugins 🤔 
Was going to add it as part of the tunneling plugins work, but i guess we can go with it already

pepicrft:
I see. It's fine if we defer adding it too. Is the idea that users don't see `plugins` when they do `--help`?

isaacroldan:
Yep, since it wasn't clear how plugins were going to work

isaacroldan (approved):

PR #430, opened 2022-09-06
Add theme info command

isaacroldan:
CLI3 is missing some commands like `store` or `whoami` that provide some useful info to the user.
This new command adds `theme info` to display some of this information. The plan would be to keep improving and adding data to this command.

Fixes https://github.com/Shopify/shopify-cli-planning/issues/361

- Add `theme info` command showing basic system data and the currently cached store
- Extract some of the code used in `app info` to `cli-kit` so it can be reused from `theme`
- Hide the upgrade command info from the `theme info` output.

### How to test your changes?
`dev shopify theme info`

### Measuring impact

### Checklist

Captura de Pantalla 2022-09-05 a las 14 11 33

### DIFF
--- a/packages/theme/src/cli/commands/theme/info.ts
+++ b/packages/theme/src/cli/commands/theme/info.ts
@@ -1,8 +1,6 @@
 import ThemeCommand from '../../utilities/theme-command.js'
-import {getTheme} from '../../utilities/theme-store.js'
-import {cli, os, output, string} from '@shopify/cli-kit'
-import {version as rubyVersion} from '@shopify/cli-kit/node/ruby'
-import {checkForNewVersion} from '@shopify/cli-kit/node/node-package-manager'
+import {themeInfo} from '../../services/info.js'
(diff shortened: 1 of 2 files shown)


packages/theme/src/cli/commands/theme/info.ts
```diff
@@ -0,0 +1,48 @@
...
+  static description = 'Print basic information about your theme environment'
+
+  static flags = {
+    ...cli.globalFlags,
+  }
+
+  public async run(): Promise<void> {
+    const sections: [string, string][] = [await this.themeConfigSection(), await this.systemInfoSection()]
+    const message = sections.map((sectionContents) => output.section(...sectionContents)).join('\n\n')
+    output.info(message)
```
pepicrft, line 17: What about keeping the command's responsibility small (parsing and validating arguments) and delegating the logic to a `services/theme/info.ts` service. It's not a lot of logic, but we'd remain consistent with other commands and we can unit-test that logic outside of the context of the CLI interface.

PR #434, opened 2022-09-06
Fix store initialization for Homebrew (project version error)

gonzaloriestra:

When you run any command using the store through the Homebrew installation (like `shopify3 theme pull`) outside the CLI project folder, it throws an error:

error-project-version

The explicit project version was removed [here](https://github.com/Shopify/cli/pull/408#discussion_r961580718) in favor of the automatic inference by Conf, but it's not working as expected. Maybe related to https://github.com/sindresorhus/conf/issues/150

- Pass projectVersion to CLIKitStore initialization
- Add helper functions to the store to simplify the usage

### How to test your changes?

`yarn shopify theme pull`

### Measuring impact

### Checklist

### DIFF
--- a/packages/cli-kit/src/store.ts
+++ b/packages/cli-kit/src/store.ts
@@ -76,14 +76,14 @@ export async function clearAppInfo(directory: string): Promise<void> {
   store.clearAppInfo(directory)
 }
 
-export async function getTheme(): Promise<string | undefined> {
+export async function getThemeStore(): Promise<string | undefined> {
   const store = await cliKitStore()
-  return store.getTheme()
+  return store.getThemeStore()
 }
 
-export async function setTheme(theme: string): Promise<void> {
+export async function setThemeStore(themeStore: string): Promise<void> {
   const store = await cliKitStore()
-  store.setTheme(theme)
+  store.setThemeStore(themeStore)
 }
 
 export async function getSession(): Promise<string | undefined> {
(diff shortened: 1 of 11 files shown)


packages/cli-kit/src/store.ts
```diff
@@ -40,17 +41,66 @@ const schema = {
 
 let _instance: CLIKitStore | undefined
 
-export function cliKitStore() {
+export async function cliKitStore() {
```
amcaplan, line 44: I'm going to go ahead and assume you've vetted carefully that we're updating every place we're calling this function to ensure we're awaiting as appropriate.

packages/cli-kit/src/store.ts
```diff
@@ -100,19 +150,19 @@ export class CLIKitStore extends Conf<ConfSchema> {
     return this.get('themeStore')
   }
 
-  setTheme(store: string): void {
+  setTheme(theme: string): void {
```
amcaplan, line 153: The method is poorly named, but it is setting a store, not a theme. So this variable rename doesn't really make sense.

packages/cli-kit/src/store.ts
```diff
@@ -40,17 +41,66 @@ const schema = {
...
       schema,
       migrations,
       projectName: 'shopify-cli-kit',
+      projectVersion: await constants.versions.cliKit(),
     })
   }
   return _instance
 }
 
+export async function getAppInfo(directory: string): Promise<CachedAppInfo | undefined> {
```
pepicrft, line 57: This made me think that we should consider breaking store down into smaller stores (i.e. app, theme, hydrogen, cli-kit). Otherwise the cli-kit store will have knowledge on its dependent packages.

packages/cli-kit/src/store.ts
gonzaloriestra: Yep 👍

packages/cli-kit/src/store.ts
gonzaloriestra: Yes, I thought about that too, but I didn't want to go any further with the refactor. I'll create a new issue.

packages/cli-kit/src/store.ts
gonzaloriestra: Oops, I just changed the parameter by looking at the method name. I've renamed the methods as well.

PR #437, opened 2022-09-07
Move the UI extensions' build logic from Go to Node

pepicrft:
Continuation of: https://github.com/Shopify/cli/pull/423

In the aim of removing the unnecessary indirection of having some UI extensions' business logic in Go, I'm moving the `build` workflow from the extensions' CLIs into the `@shopify/app` package.

- [x] Move the logic to a new module that does the bundling using ESBuild.
- [x] Adjust the `build` service to consume the new APIs.
- [x] Unit test the new module for bundling.  

### How to test your changes?
You can try to build the fixture or any other app.

### Measuring impact

### Checklist

### DIFF
--- a/packages/app/src/cli/services/extensions/bundle.test.ts
+++ b/packages/app/src/cli/services/extensions/bundle.test.ts
@@ -1,0 +1,213 @@
+import {bundleExtension} from './bundle.js'
+import {testApp, testUIExtension} from '../../models/app/app.test-data.js'
+import {describe, expect, test, vi} from 'vitest'
+import {build as esBuild, BuildOptions, WatchMode} from 'esbuild'
+import {abort, path} from '@shopify/cli-kit'
+
+vi.mock('esbuild', async () => {
+  const esbuild: any = await vi.importActual('esbuild')
+  return {
+    ...esbuild,
+    build: vi.fn(),
+  }
+})
+
+describe('buildExtension', () => {
+  test('invokes ESBuild with the right options and forwards the logs', async () => {
+    // Given
+    const extension = testUIExtension()
+    const stdout: any = {
+      write: vi.fn(),
+    }
+    const stderr: any = {
+      write: vi.fn(),
+    }
+    const app = testApp({
+      directory: '/project',
+      dotenv: {
+        path: '/project/.env',
+        variables: {
+          FOO: 'BAR',
+        },
+      },
+      extensions: {
+        ui: [extension],
+        theme: [],
+        function: [],
+      },
+    })
+    vi.mocked(esBuild).mockResolvedValue(esbuildResultFixture())
+
+    // When
+    await bundleExtension({
+      env: app.dotenv?.variables ?? {},
+      outputBundlePath: extension.outputBundlePath,
+      minify: true,
+      environment: 'production',
+      sourceFilePath: extension.entrySourceFilePath,
+      stdout,
+      stderr,
+    })
+
+    // Then
+    const call = vi.mocked(esBuild).mock.calls[0] as any
+    expect(call).not.toBeUndefined()
+    const options: BuildOptions = call[0]
+
+    expect(options.bundle).toBeTruthy()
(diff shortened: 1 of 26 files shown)


packages/app/src/cli/services/extensions/bundle.test.ts
```diff
@@ -0,0 +1,213 @@
...
+vi.mock('esbuild', async () => {
+  const esbuild: any = await vi.importActual('esbuild')
+  return {
+    ...esbuild,
+    build: vi.fn(),
+  }
+})
+
+describe('buildExtension', () => {
+  test('invokes ESBuild with the right options and forwards the logs', async () => {
```
amcaplan, line 16: Little nitpick, I think we usually use `it` not `test`.

packages/app/src/cli/services/extensions/bundle.ts
```diff
@@ -0,0 +1,143 @@
...
+}
+
+/**
+ * Returns true if the "graphql" and "graphql-tag" packages can be
+ * resolved. This information is sued to determine whether we should
+ * or not include the esbuild-graphql-loader plugin when invoking
+ * ESBuild
+ * @returns {boolean} Returns true if the "graphql" and "graph"
+ */
+function isGraphqlPackageAvailable() {
```
amcaplan, line 134:
```suggestion
function isGraphqlPackageAvailable(): boolean {
```

amcaplan (commented):
I definitely didn't understand nearly 100%, and I'm sure there will be some bugs emerging from this, but it's totally worth it to remove an unnecessary double layer of indirection that doesn't add anything to the final product.

I wonder if it would be wise to lower the risk of interruption to our partners by starting with opt-in-by-env-var, and then pulling the plug once we've done some broad internal testing.

packages/app/src/cli/services/extensions/bundle.ts
```diff
@@ -0,0 +1,143 @@
...
+
+  return plugins
+}
+
+/**
+ * Returns true if the "graphql" and "graphql-tag" packages can be
+ * resolved. This information is sued to determine whether we should
+ * or not include the esbuild-graphql-loader plugin when invoking
+ * ESBuild
+ * @returns {boolean} Returns true if the "graphql" and "graph"
```
isaacroldan, line 132:
```suggestion
 * resolved. This information is used to determine whether we should
 * or not include the esbuild-graphql-loader plugin when invoking ESBuild
 * @returns {boolean} Returns true if the "graphql" and "graphql-tag" can be resolved.
```

packages/app/src/cli/services/extensions/bundle.test.ts
```diff
@@ -0,0 +1,213 @@
...
+      env: app.dotenv?.variables ?? {},
+      outputBundlePath: extension.outputBundlePath,
+      minify: true,
+      environment: 'production',
+      sourceFilePath: extension.entrySourceFilePath,
+      stdout,
+      stderr,
+      watchSignal: abortController.signal,
+    })
+    abortController.abort()
```
isaacroldan, line 128: Don't you need to provide a `watch` function for esBuild to actually keep watching so you can abort it?
In this case `esbuildStop` is being called, but is not stopping anything, right?

isaacroldan (approved):
🎩 'ed and works as expected 👌

packages/app/src/cli/services/extensions/bundle.test.ts
pepicrft: I agree we should be more consistent here. I'll place a linting rule and do some renames in a follow-up PR.

packages/app/src/cli/services/extensions/bundle.test.ts
pepicrft: > Don't you need to provide a watch function for esBuild to actually keep watching so you can abort it?

That's correct. Otherwise the call to `esbuild` blocks until the bundling completes. In this case I'm testing that the `abort` call propagates to `esbuild`, which is mocked.

PR #443, opened 2022-09-07
add hydrogen deploy error handling & oxygen api service

tempor1s:

Closes https://github.com/Shopify/oxygen-platform/issues/528

This PR introduces error handling and retry logic for the various calls that we make to DMS so that we can be more resilient / provide context to the user on why their deploy might be failing, etc. I also greatly increased the scope of what the unit tests cover.

I also created a new service within the CLI kit that is used for interacting with DMS as it seems this is the pattern and convention that is being used to interact with other APIs (identity, partners, shopify, etc) and removes the need for the CLI kit to export internal functionality (namely `buildHeaders`)

This PR also places the zip file of built assets into a temporary directory to follow the convention that `oxygenctl` is currently using instead of just placing the zip in the dist.

### Checklist

### DIFF
--- a/packages/cli-hydrogen/src/cli/services/deploy.test.ts
+++ b/packages/cli-hydrogen/src/cli/services/deploy.test.ts
@@ -78,20 +78,75 @@ describe('deployToOxygen()', () => {
     expect(mockedHealthCheck).toHaveBeenCalledWith('previewURL')
   })
 
-  it('retries on error for healthCheck, uploadDeployment and createDeployment', async () => {
-    mockedFillDeployConfig.mockResolvedValue(reqDeployConfig)
-    mockedCreateDeployment
-      .mockRejectedValueOnce(new Error())
-      .mockResolvedValue({deploymentID: 'deploymentID', assetBaseURL: 'assetBaseURL'})
-    mockedBuildTaskList.mockReturnValue([])
-    mockedUploadDeployment.mockRejectedValueOnce(new Error()).mockResolvedValue('previewURL')
-    mockedHealthCheck.mockRejectedValueOnce(new Error()).mockRejectedValueOnce(new Error())
+  describe('when error is recoverable', async () => {
+    it('retries on error for healthCheck, uploadDeployment and createDeployment', async () => {
+      mockedFillDeployConfig.mockResolvedValue(reqDeployConfig)
+      mockedCreateDeployment
+        .mockRejectedValueOnce(new Error())
+        .mockResolvedValue({deploymentID: 'deploymentID', assetBaseURL: 'assetBaseURL'})
+      mockedBuildTaskList.mockReturnValue([])
+      mockedUploadDeployment.mockRejectedValueOnce(new Error()).mockResolvedValue('previewURL')
+      mockedHealthCheck.mockRejectedValueOnce(new Error()).mockRejectedValueOnce(new Error())
 
-    await deployToOxygen(deployConfig)
+      await deployToOxygen(deployConfig)
 
-    expect(mockedCreateDeployment).toHaveBeenCalledTimes(2)
-    expect(mockedUploadDeployment).toHaveBeenCalledTimes(2)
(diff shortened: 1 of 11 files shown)


packages/cli-hydrogen/src/cli/services/deploy.test.ts
```diff
@@ -93,6 +93,54 @@ describe('deployToOxygen()', () => {
     expect(mockedHealthCheck).toHaveBeenCalledTimes(3)
   })
 
+  it('does retry for createDeployment if the error is recoverable', async () => {
```
lynchv, line 96: Could throw these in describe blocks
```suggestion
describe('when error is recoverable')
  it('retry createDeployment')
  
  it('retry uploadDeployment')
```

packages/cli-hydrogen/src/cli/services/deploy/upload.ts
```diff
@@ -8,14 +8,17 @@ import {UploadDeploymentQuery} from './graphql/upload_deployment.js'
...
 import {zip} from '@shopify/cli-kit/node/archiver'
+import {ClientError} from 'graphql-request'
 
-export const createDeployment = async (config: ReqDeployConfig): Promise<CreateDeploymentResponse> => {
-  const headers = await api.buildHeaders(config.deploymentToken)
-  const client = await http.graphqlClient({
-    headers,
-    url: getOxygenAddress(config.oxygenAddress),
-  })
+export const TooManyRequestsError = () => {
```
lynchv, line 13: Should put those in `packages/cli-hydrogen/src/cli/services/deploy/error.ts`

packages/cli-hydrogen/src/cli/services/deploy/upload.ts
```diff
@@ -26,36 +29,73 @@ export const createDeployment = async (config: ReqDeployConfig): Promise<CreateD
...
+      }
+    }
+
+    throw error
+  }
 }
 
 export const uploadDeployment = async (config: ReqDeployConfig, deploymentID: string): Promise<string> => {
-  let headers = await api.buildHeaders(config.deploymentToken)
+  let deploymentData: UploadDeploymentResponse | undefined
```
lynchv, line 61: I don't think this is needed, could simply be defined as a `const` below.

tempor1s:
## Force Push Notes
- Extracted errors into separate file
- Changed `unrecoverable` errors to throw instead of just returning to stop the tasks from continuing.
- Extracted retry logic into separate method
- Improve tests to cover new exception functionality

packages/cli-hydrogen/src/cli/services/deploy.ts
```diff
@@ -95,3 +99,24 @@ export async function deployToOxygen(_config: DeployConfig) {
 
   return list.run()
 }
+
+async function shouldRetryOxygenCall(
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  task: ui.ListrTaskWrapper<TaskContext, any>,
+  backoffPolicy: number[],
```
lynchv, line 106: I think we could make the `backoffPolicy` a global var and not pass it down?

lynchv (approved):
lgtm

tempor1s:
Hey @amcaplan / @pepicrft would you mind taking a look at this PR before I merge it since I am adding a new service to the CLI kit? Thanks!

packages/cli-hydrogen/src/cli/services/deploy.test.ts
```diff
@@ -77,20 +77,75 @@ describe('deployToOxygen()', () => {
...
+
+      await deployToOxygen(deployConfig)
+
+      expect(mockedCreateDeployment).toHaveBeenCalledTimes(1)
+      expect(mockedUploadDeployment).toHaveBeenCalledTimes(3)
+    })
+  })
+
+  describe('when error is unrecoverable', async () => {
+    it('does retry for createDeployment', async () => {
```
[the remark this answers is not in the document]
tempor1s: Good catch!

PR #446, opened 2022-09-08
Add DeviceAuth flow for Spin

isaacroldan:
The use of cloud environments for development has increased recently (both in and out of shopify). 

Running the Shopify CLI inside one of those environments is a bit tricky as the CLI usually depends on localhost being a real localhost, opening a tunnel or creating a local server for the auth process or to serve apps.

This PR fixes the first blocker users are going to encounter in cloud environments: **Authentication**.
Instead of relying on a local server that listens to identity callbacks when the user completes the authentication, we are adding a new (complementary) auth flow that allows you to login in local environment (your local browser) but getting a token in your cloud environment.

This new auth method will only be used if the CLI detects that is being executed in a cloud environment.
By default the classic method will be used.

Fixes https://github.com/Shopify/shopify-cli-planning/issues/330

- Add new Device Auth flow to obtain an identity token, once obtained the process of exchanging that token for application tokens or refreshing an expired token is the same.
- Improve how spin FQDN is obtained, specially when the CLI is being executed inside spin.
- Add new env var to force using the new device auth flow
- Detect common cloud environments using pre-defined env vars

### How to test your changes?
- Try the new auth flow using `SHOPIFY_CLI_DEVICE_AUTH=1` (remember to logout first)
- Try this branch in a Spin instance, it should automatically activate the device auth flow.

### Measuring impact

### Checklist

https://user-images.githubusercontent.com/3757193/189137666-d0ff97a1-be4e-45da-8f3b-206545c3449b.mov


packages/cli-kit/src/session/device-authorization.ts
```diff
@@ -0,0 +1,106 @@
...
+  })
+
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  const jsonResult: any = await response.json()
+
+  debug(content`Received device authorization code: ${token.json(jsonResult)}`)
+
+  info('\nTo run this command, log in to Shopify Partners.')
+  info(content`User verification code: ${jsonResult.user_code}`)
+  info(content`👉 Open ${token.link('this Link', jsonResult.verification_uri_complete)} to start the auth process`)
```
gonzaloriestra, line 46: What about showing the full URL instead of the hyperlink here? Users will have to copy&paste it sometimes instead of just clicking, right?

packages/cli-kit/src/session/device-authorization.ts
```diff
@@ -0,0 +1,106 @@
...
+
+  return {
+    deviceCode: jsonResult.device_code,
+    userCode: jsonResult.user_code,
+    verificationUri: jsonResult.verification_uri,
+    expiresIn: jsonResult.expires_in,
+    verificationUriComplete: jsonResult.verification_uri_complete,
+    interval: jsonResult.interval,
```
gonzaloriestra, line 54: So Identity is suggesting an interval for the polling? What's the default value? 5 seconds?

gonzaloriestra (approved):
This is awesome! 👏 

It works great and the code LGTM. Actually, I wonder if we should make this the default authentication method. What's the advantage of keeping the old one? We can change that later, of course.

packages/cli-kit/src/session/device-authorization.ts
isaacroldan: Yep! 5 seconds

isaacroldan:
Can't make it default because it doesn't work with the employee scopes to access internal orgs (I asked if this is intended or is going to change, but no response yet). But that doesn't affect spin because from there you'll access only spin stuff.

Also, if you are already logged in a browser, this method doesn't use the same session and will force you to login again (not sure why), so that's a bit worse experience.

packages/cli-kit/src/environment/spin-cache.ts
```diff
@@ -0,0 +1,9 @@
+let cachedSpinFQDN: string
+
+export function getCachedSpinFqdn(): string | undefined {
+  return cachedSpinFQDN
+}
+
+export function setCachedSpinFqdn(fqdn: string) {
+  cachedSpinFQDN = fqdn
+}
```
pepicrft, line 9: Adding state to modules in `@shopify/cli-kit` caused issues in the past because we can't prevent multiple copies of `@shopify/cli-kit` from being in the graph and with them duplicated states. I'd consider loading this information once at the service level and passing it down. Alternatively, we shell out to get the information. Have you measured how much time it takes that operation?
cc @isaacroldan

packages/cli-kit/src/environment/local.test.ts
```diff
@@ -128,3 +128,60 @@ describe('analitycsDisabled', () => {
...
+    const env = {GITPOD_WORKSPACE_ID: '1'}
+
+    // When
+    const got = useDeviceAuth(env)
+
+    // Then
+    expect(got).toBe(true)
+  })
+
+  it('returns false otherwise', () => {
```
pepicrft, line 177: This `otherwise` makes assumption regarding the position of the test inside the `describe` block. I'd consider changing it to: `returns false when GITPOD_WORKSPACE_ID if falsy`.

packages/cli-kit/src/session/exchange.ts
```diff
@@ -127,7 +157,7 @@ async function requestAppToken(
 }
 
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
-async function tokenRequest(params: {[key: string]: string}): Promise<any> {
+async function tokenRequest(params: {[key: string]: string}, throws = true): Promise<any> {
```
pepicrft, line 160: Having two throwing behaviours within the same function controlled via function argument is a bit uncommon. Also, the API is not very idiomatic: 

```js
await tokenRequest(params, false)
```

I'd consider removing that argument and use the [new Result type](https://github.com/Shopify/cli/pull/333) to let the caller decide what to do with the error.

packages/cli-kit/src/session/device-authorization.ts
gonzaloriestra: I'm worried about other environments where a browser can't be opened automatically, but we can change it later.

PR #447, opened 2022-09-08
Don't allow the log file to grow too much

alvaro-shopify:

Fixes https://github.com/Shopify/shopify-cli-planning/issues/344

- Switching from loading in memory the whole logfile content to truncate the first lines to using streams so only little chunks are handled at a time
- This way each time a command is run the log is reduced to a max size, even if the current log file size is huge
- Not added the truncating process each time a message is written to the log to avoid overload.

### How to test your changes?
- Overwrite the existing log in ` ~/Library/Caches/shopify-cli-nodejs/shopify.cli.log` with a downloaded [sample huge csv file](https://eforexcel.com/wp/wp-content/uploads/2017/08/2m-Sales-Records.7z) and run any command. The log file should be truncated to a maximum (fixed) size similar to 5MB 

### Measuring impact

### Checklist


packages/theme/package.json
```diff
@@ -48,13 +48,10 @@
     "win32"
   ],
   "oclif": {
-    "commands": "dist/cli/commands",
-    "hooks": {
-      "init": "@shopify/cli-kit/node/hooks/init"
-    }
```
isaacroldan, line 54: Is removing this intended?

packages/cli-kit/src/output.ts
```diff
@@ -31,39 +30,82 @@ import {
...
+      await pipeline(createReadStream(tmpLogFile), createWriteStream(logFile))
+      unlinkSync(tmpLogFile)
+      truncating = false
+    })
+  }
+}
+
+function calculateNumLinesToRetain(): number {
+  return Math.floor(maxLogFileSize / sizePerLogLine)
 }
```
isaacroldan, line 109: If `maxLogFileSize` and `sizePerLogLine` are constants, why not make have:
`const filesToRetain = Math.floor(maxLogFileSize / sizePerLogLine)` at the top? no need for a function that will recalculate the same value everytime 🤔

packages/cli-kit/src/output.ts
```diff
@@ -31,39 +30,82 @@ import {
...
+  if (fileSizeSync(logFile) > maxLogFileSize) {
+    truncating = true
+    const truncateLines = new LinesTruncatorTransformer(calculateNumLinesToRetain())
+    await pipeline(createReadStream(logFile), truncateLines, createWriteStream(tmpLogFile)).then(async () => {
+      await pipeline(createReadStream(tmpLogFile), createWriteStream(logFile))
+      unlinkSync(tmpLogFile)
+      truncating = false
+    })
+  }
+}
```
isaacroldan, line 105: Can this be simplified like this?

```ts
async function truncateLogs(logFile: string) {
  const tmpLogFile = logFile.concat('.tmp')
  if (truncating || fileSizeSync(logFile) < maxLogFileSize) return

  truncating = true
  const truncateLines = new LinesTruncatorTransformer(linesToRetain)
  await pipeline(createReadStream(logFile), truncateLines, createWriteStream(tmpLogFile))
  await pipeline(createReadStream(tmpLogFile), createWriteStream(logFile))
  unlinkSync(tmpLogFile)
  // eslint-disable-next-line require-atomic-updates
  truncating = false
}
```

1. Adding an early return if conditions are not met
2. Use 2 awaits instead of a `.then`

packages/cli-kit/src/output.ts
alvaro-shopify: Yeah. Initially, this was a more complex calculation, that't why I didn't remove it

packages/theme/package.json
amcaplan: I don't think this is right. It needs to be run twice because there are 2 copies of cli-kit, one for `@shopify/cli-main` and one for `@shopify/theme`. This is an artifact of the CLI being installed globally.

packages/cli-kit/src/log.ts
```diff
@@ -0,0 +1,175 @@
...
+    return
+  }
+  const tmpLogFile = logFile.concat('.tmp')
+  truncating = true
+  const truncateLines = new LinesTruncatorTransformer(maxLogFileSize)
+  await pipeline(createReadStream(logFile), truncateLines, createWriteStream(tmpLogFile))
+  await pipeline(createReadStream(tmpLogFile), createWriteStream(logFile)).then(() => {
+    unlinkSync(tmpLogFile)
+    truncating = false
+  })
```
isaacroldan, line 129: We try to avoid mixing `await` and `then`, could this be?:
```ts
await pipeline(createReadStream(tmpLogFile), createWriteStream(logFile))
await unlinkSync(tmpLogFile)
truncating = false
```

packages/cli-kit/src/log.test.ts
```diff
@@ -0,0 +1,156 @@
...
+      vi.mocked(generateRandomUUID).mockReturnValue('random-uuid')
+      const logPath = join(tmpDir, 'shopify.cli.log')
+      createLogFile(logPath, KB_256, MB_7)
+      const fileSize = fileSizeSync(logPath)
+
+      // When
+      await initiateLogging({logDir: tmpDir, override: true})
+
+      // Then
+      expect(fileSize).toBeGreaterThan(fileSizeSync(logPath))
```
isaacroldan, line 68: Could you add an expect to check that `fileSizeSync(logPath)` is less than the maximum allowed size?

isaacroldan:
Great work with the refactor and tests 👏

packages/cli-kit/src/log.ts
alvaro-shopify: I have little lint problem with the using of `truncating` global variable. I will add an odd workaround to avoid disable the rule

packages/cli-kit/src/log.test.ts
```diff
@@ -0,0 +1,156 @@
...
+import {beforeEach, describe, expect, it, vi} from 'vitest'
+import {appendFileSync} from 'fs'
+import {EOL} from 'os'
+import {randomBytes} from 'crypto'
+
+const KB_256 = 0.256 * 1024 * 1024
+const MB_2 = 2 * 1024 * 1024
+const MB_7 = 7 * 1024 * 1024
+
+beforeEach(() => {
```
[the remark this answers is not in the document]
alvaro-shopify: But these are constants I don't think I could added them to `beforeAll` right?

isaacroldan (approved):

PR #448, opened 2022-09-08
Refactor structure of package updates functions

amcaplan:

The TODO says it all. We should have a straightforward "upgrade without thinking" function which can be called by an "upgrade only certain things" function, not the other way around.

I meant to do this some time ago, especially given it broke `shopify upgrade` right at the Summer Edition launch, but I never got back to it. UNTIL NOW.

Refactors the update code to have functions' dependencies point in the right direction.

### How to test your changes?

Run `shopify upgrade` or generate an extension on an existing project.

### Measuring impact

### Checklist

### DIFF
--- a/packages/cli-kit/src/node/node-package-manager.ts
+++ b/packages/cli-kit/src/node/node-package-manager.ts
@@ -319,7 +319,10 @@ ${token.json(options)}
   await addNPMDependencies(dependenciesToAdd, options)
 }
 
-export async function addNPMDependencies(dependencies: DependencyVersion[], options: AddNPMDependenciesIfNeededOptions) {
+export async function addNPMDependencies(
+  dependencies: DependencyVersion[],
+  options: AddNPMDependenciesIfNeededOptions,
+) {
   let args: string[]
   const depedenciesWithVersion = dependencies.map((dep) => {
     return dep.version ? `${dep.name}@${dep.version}` : dep.name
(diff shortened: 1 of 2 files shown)


packages/cli-kit/src/node/node-package-manager.ts
```diff
@@ -334,7 +335,7 @@ ${token.json(options)}
       args = argumentsToAddDependenciesWithPNPM(depedenciesWithVersion, options.type)
       break
   }
-  options.stdout?.write(`Executing...${args.join(' ')}`)
+  options.stdout?.write(`Executing... ${[options.packageManager, ...args].join(' ')}`)
```
amcaplan, line 338: This line is not strictly related to the PR topic, but the output seems much more helpful.

packages/cli-main/src/cli/commands/upgrade.ts
```diff
@@ -86,10 +85,10 @@ export default class Upgrade extends Command {
     const packagesToUpdate = packages.filter((pkg: string): boolean => {
       const pkgRequirement: string | undefined = deps[pkg]
       return Boolean(pkgRequirement)
-    })
+    }).map((pkg) => { return {name: pkg, version: 'latest'} })
```
amcaplan, line 88: This needs to be updated because the method accepts an array of `DependencyVersion`s, not strings.

packages/cli-kit/src/node/node-package-manager.ts
```diff
@@ -298,7 +298,6 @@ export interface DependencyVersion {
 export async function addNPMDependenciesIfNeeded(
   dependencies: DependencyVersion[],
   options: AddNPMDependenciesIfNeededOptions,
-  force = false,
```
amcaplan, line 301: This `force` param is the gross hack that I wanted to get around. `IfNeeded` means this is a more specialized method, and we shouldn't have a `force` param which contradicts the `IfNeeded`.

isaacroldan (approved):

PR #454, opened 2022-09-12
Add `extension_directories` attribute to `shopify.app.toml`

pepicrft:
Resolves https://github.com/Shopify/shopify-cli-planning/issues/310
Documentation PR: https://github.com/Shopify/shopify-dev/pull/26029

Some partners [reported](https://shopify.slack.com/archives/C03NT4BP8N9/p1661967429402539) that the opinion around where extensions should be located within an app is too strong and they can't adopt it easily due to how they are organizing extensions in multiple repositories.

I'm adding support for a new attribute in the `shopify.app.toml`, `extension_directories`, that's an array of glob expressions pointing to directories containing extensions:

```toml
extension_directories = [
  "my_other_extension",
  "other_directory/extensions/*"
]
``` 
The default value of the attribute is `extensions/*`.

### How to test your changes?
1. Rename the `extensions` directory in the `fixtures/app` directory to `my-extensions`.
2. Edit the app configuration file and add `extension_directories = ["my-extensions"]`
3. Run `yarn shopify app info --path ./fixtures/app`

### Measuring impact

### Checklist

### DIFF
--- a/packages/app/src/cli/models/app/app.ts
+++ b/packages/app/src/cli/models/app/app.ts
@@ -7,6 +7,7 @@ import {getDependencies, PackageManager, readAndParsePackageJson} from '@shopify
 
 export const AppConfigurationSchema = schema.define.object({
   scopes: schema.define.string().default(''),
+  extension_paths: schema.define.string().default('extensions/*'),
 })
 
 export enum WebType {
--- a/packages/app/src/cli/models/app/app.ts
+++ b/packages/app/src/cli/models/app/app.ts
@@ -7,7 +7,7 @@ import {getDependencies, PackageManager, readAndParsePackageJson} from '@shopify
 
 export const AppConfigurationSchema = schema.define.object({
   scopes: schema.define.string().default(''),
-  extensionDirectories: schema.define.array(schema.define.string()).default(['extensions/*']),
+  extensionDirectories: schema.define.array(schema.define.string()).optional(),
 })
 
 export enum WebType {
--- a/packages/app/src/cli/models/app/app.ts
+++ b/packages/app/src/cli/models/app/app.ts
@@ -7,7 +7,7 @@ import {getDependencies, PackageManager, readAndParsePackageJson} from '@shopify
 
 export const AppConfigurationSchema = schema.define.object({
   scopes: schema.define.string().default(''),
-  extension_paths: schema.define.array(schema.define.string()).default(['extensions/*']),
+  extensionDirectories: schema.define.array(schema.define.string()).default(['extensions/*']),
 })
 
 export enum WebType {
--- a/packages/app/src/cli/models/app/app.ts
+++ b/packages/app/src/cli/models/app/app.ts
@@ -7,7 +7,7 @@ import {getDependencies, PackageManager, readAndParsePackageJson} from '@shopify
 
 export const AppConfigurationSchema = schema.define.object({
   scopes: schema.define.string().default(''),
-  additionalExtensionDirectories: schema.define.array(schema.define.string()).default([]),
+  extensionDirectories: schema.define.array(schema.define.string()).default(['extensions/*']),
 })
 
 export enum WebType {
--- a/packages/app/src/cli/models/app/app.ts
+++ b/packages/app/src/cli/models/app/app.ts
@@ -7,7 +7,7 @@ import {getDependencies, PackageManager, readAndParsePackageJson} from '@shopify
 
 export const AppConfigurationSchema = schema.define.object({
   scopes: schema.define.string().default(''),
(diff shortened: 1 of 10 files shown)


pepicrft:
@stephszeto would you like to have metrics around this?

packages/app/src/cli/models/app/loader.ts
```diff
@@ -65,16 +65,18 @@ class AppLoader {
     this.appDirectory = await this.findAppDirectory()
     const configurationPath = await this.getConfigurationPath()
     const configuration = await this.parseConfigurationFile(AppConfigurationSchema, configurationPath)
-    const extensionsPath = path.join(this.appDirectory, `${blocks.extensions.directoryName}`)
+    const extensionDirectories = configuration.extensionDirectories.map((extensionDirectory) => {
+      return path.join(this.appDirectory, extensionDirectory)
+    })
     const dotenv = await this.loadDotEnv()
     const {functions, usedCustomLayout: usedCustomLayoutForFunctionExtensions} = await this.loadFunctions(
```
shauns, line 72: I think we'd expect to see `usedCustomLayout` changing if this config is used. If we add that behaviour then analytics are already in place to measure usage of this feature.

stephszeto:
> @stephszeto would you like to have metrics around this?

Yes please!

> I think we'd expect to see usedCustomLayout changing if this config is used. If we add that behaviour then analytics are already in place to measure usage of this feature.

Ideally would be great if we had the ability to distinguish between someone customizing their web location vs customizing their extension locations.

shauns:
> This is information that we can capture too. Note that right now we don't have a similar attribute for web.

We do track this for web. We check for your web TOMLs everywhere, and then we also check only in `web`. If the counts vary, we log that you have a non-standard project set-up.

For the extensions (UI, theme, function) the same fields (x3) exist, they're just always false. With the addition of this option would expect it to be true when those TOMLs are found outside of `extensions/*/`

packages/app/src/cli/models/app/app.ts
```diff
@@ -7,6 +7,7 @@ import {getDependencies, PackageManager, readAndParsePackageJson} from '@shopify
 
 export const AppConfigurationSchema = schema.define.object({
   scopes: schema.define.string().default(''),
+  additionalExtensionDirectories: schema.define.array(schema.define.string()).default([]),
```
lemonmade, line 10: Purely additive configuration is a very unusual concept, and IMO misses the point for developers who want control over their projects. Having configuration that is only partially overridable by developers is bound to get in the way. It also forces us to choose complex and ugly configuration options; `extensions = ["extensions/*"]` is so much nicer, and mirrors what they would see in defining Yarn/NPM/PNPM workspace configurations.

amcaplan:
I don't love the name `additional_extension_directories`, it feels a bit clunky, but I don't have a better suggestion at the moment.

More importantly, though, I think that if it doesn't overwrite but instead is additive, we definitely don't want to have the key be simply `extension_directories` which gets in the way of allowing an overwrite in the future. So the instinct is very good, I'm glad you thought to rename the setting, I just wish I had a better name to suggest!

packages/app/src/cli/models/app/app.ts
lemonmade: I am not sure the link you posted is really relevant — the conventions for configuration as code vs static configuration are naturally going to be a little different. That said, even in the example you posted, presumably setting an array does not merge your options in with Rails' defaults.

I think we need to do a lot less assuming what developers *should* do. We have almost no apps that use extensions in production at Shopify; developers know **much** better than us what they need to be doing to make use of the tools we give them. I imagine you have seen a lot of the same feedback I have in checkout, but from my perspective it has been very clear: developers need flexibility. They were stuck because of us sticking to `web/*` for too long, and many of them are stuck because they have a good reason for separating extensions into multiple places/ repos.

lemonmade: I am honestly struggling to see how we could prefer a developer write this configuration:

```toml
# Merges with ["extensions/*"], which you can't override
additional_extension_directories = ["my-extensions/*"]
```

over this one:

```toml
# Only the extensions listed will count
extensions = ["my-extensions/*"]
```

They both allow you to omit the field entirely to get the default `extensions = ["extensions/*"]` option; the second one is significantly easier to read, understand, and document, and leaves the developer in full control when they are taking the wheel.

pepicrft:
@lemonmade / @lemonmade I reverted the change and made it an "override" attribute `extension_directories`. If users disagree with our opinions they should take full control instead of building upon our defaults. If they want it to be additive, our convention will be documented so they can maintain it.

shauns (approved):

PR #455, opened 2022-09-12
UI extensions serve in Node

pepicrft:
This PR is part of the effort of moving the logic for creating, building, and serving UI extensions from Node to Go.

This PR Duplicates the `serve` command from `packages/ui-extensions-go-cli` in `packages/app`.  Shopifolk are then opted into the new version unless they specify the `SHOPIFY_RUN_AS_USER=true` env var.

This code has a few concerns:

1. Creating a **bundler** to build extensions
2. Creating a **Payload Store** to maintain the state of all extensions
3. Creating a web server to respond to **fetch requests** to `/extensions/...`
4. Creating a **web socket server** at `/extensions/` that can be used to update the Payload Store, mediate messages, or inform clients of changes to the Payload Store
   
I'll explain these concerns in depth now:

#### Bundler
First, for each extension the bundler uses esbuild to watch for changes to the extension entry file.  E.g: `/extensions/[extension]/src/index.jsx`.  When ESBuild detects a change:

1. EsBuild builds the file to `/extensions/[extension]/dist/main.jsx`
2. EsBuild informs the Payload Store that of the build status.
3. The Payload Store then recalculates the data for that extension. 
4. The WebSockets then pass an update message to every client with the updated extension data

Second, for each extension the bundler uses [chokidar](https://www.npmjs.com/package/chokidar?activeTab=readme) to watch the `/extensions/[extension]/locales/` folder for changes to json files. When a file is added, edited or deleted:

1. Chokidar informs the PayloadStore that extension has changed
2. The Payload Store recalculates the data for that extension, including the localization data.

#### Payload Store
The Payload Store is responsible for keeping the state of extensions up to date.  At a high level it is an event emitter with methods to mutate it's internal state.  Whenever it updates it's internal state it emits events to allow listeners to do what they need to do.  In more detail it is responsible for:

1. Updating extension data: The websocket server could ask the Payload Store to update an extension.  An example of this would be when a user hovers, or hides an extension in the Dev Console.  Or the Bundler could ask to update the extension.
2. Updating app data:  Extension points can save arbitrary data to the app data that the Payload Store is responsible for.  In this situation an extension point would send a websocket message with new App Data, and the Payload Store would merge that data with existing data.
3. Informing Event Listeners on updates: The websocket server listens for when the Payload Store  updates data and informs clients of those updates.
4. Calculating data: One example of this is calculating the localization for any extension with localization files.  This involves parsing the JSON files on disk and converting those files into the Payload stores internal state.  Arguably this could be the responsibility of the bundler, but the current architecture is quite simple to reason about.

#### WebServer  for fetch Requests
The HTTP WebServer for fetch requests has the following end points:

1. `/extensions/dev-console` returns the UI and assets for the dev console 
2. `/extensions/[id]` returns the redirect URL to view the extension point for that extension.  That extension point will then connect to the server `extensions/[id]/assets/some/asset.js` to get the assets for that extension.  Right now this is just the `dist/main.js`
3. `/extensions/` returns the JSON data for all the extensions the developer has setup

In addition, the following applies to all endpoints

1. We allow Cross origin requests
2. We disable caching
3. We log requests for debugging

#### Websocket server
The web socket server is responsible for:

1. Sending the app & extensions state to clients when a client first creates a websocket connection.
2. Informing clients when there are changes to an extension.  An example of this is when a user edits a localization the end result is that the new extension data is sent via an `UPDATE` message to every client. Another example of this is when it responds to an `UPDATE` Message sent by a client.  It does this by asking the Payload Store to update the app or extension data with the data the client passed in the message. Once the Payload Store completes this operation it is the responsibility of the Websocket server to pass the new Payload Store data to all clients via an `UPDATE` message.
3. Mediating messages from clients.  Clients can send `DISPATCH` message.  The websocket server then passes this message on to every client.

### How to test your changes?

@dmhenry @pepicrft and @alfonso-noriega have tested each extension manually.  So I'll only add instructions for 2 extensions (Thanks Trish @vividviolet for these)

#### Checkout

1. Ensure your dev shop has the following beta force_checkout_one, checkout_one_argo_extensions
2. Run `yarn shopify app dev --path ./fixtures/app`.
3. Click on the extension preview link
4. Verify that you see an extension rendered inside the order summary
5. Make some changes to localization files in `/fixtures/app/extensions/checkout_ui`
6. Verify that the extension is reloaded and reflect the changes
7. Make some changes to the extension script in `/fixtures/app/extensions/checkout_ui`
8. Verify that the extension is reloaded and reflect the changes

#### Admin Product Subscription (published and unpublished extension)

1. Ensure your dev shop has the following beta argo_dev_portal, argo_dev_portal_mobile and the app is installed on your shop
2. Run `yarn shopify app dev --path ./fixtures/app`.
3. Click on the extension preview link
4. Verify that Web loads with the Dev Console appearing as an iframe at the bottom of the screen
5. Verify that you are redirect to a product details page
6. Navigate down to the Purchase Options card and click on Add purchase options
7. Verify that you can select your app and render the "Create new option" extension point
8. Make some changes to the extension script
9. Verify that the extension is reloaded and reflect the changes
10. Verify the following all work with the Product Subscription extension:
  - hovering on the Product Subscription extension should highlight that extension
  - clicking on the hide icon for the Product Subscription extension should make the extension unavailable. If you have 
  - multiple apps installed that powers the Subscription extension then clicking on "hide" should remove your test app from the list
  - clicking on the link for the Product Subscription extension should navigate to the product details page for the first product if you're not already on the page
  - clicking on the "refresh" button should force reload the extension
11. Testing a deployed the extension
  - Deploy your extension
  - Make some changes to the extension script
  - Click on the preview link again to load up the extension
  - Select your app and render the "Create new option" extension point
  - Make some changes to the extension script
  - Verify that the extension is reloaded and reflect the changes
  - Click the "hide" button on the Dev Console
  - Verify that you now see the published extension script - local changes should have no effect
  - Click the "show" button on the Dev Console
  - Verify that you see your local script rendered

### Measuring impact

### Checklist


configurations/vite.config.ts
```diff
@@ -22,7 +22,17 @@ export default function config(packagePath: string) {
 
 export const aliases = (packagePath: string) => {
   return [
-    {find: /@shopify\/cli-kit\/(.+)/, replacement: path.join(packagePath, '../cli-kit/src/$1.ts')},
+    {
+      find: /@shopify\/cli-kit\/(.+)/,
```
shauns, line 26: Don't _love_ this change riding along with this work

packages/app/src/cli/services/dev/extension/websocket/handlers.ts
```diff
@@ -0,0 +1,110 @@
...
+      data: options.payloadStore.getConnectedPayload(),
+      version: '3',
+    }
+    output.debug(output.content`Sending connected payload: ${output.token.json(connectedPayload)}`)
+    ws.send(JSON.stringify(connectedPayload))
+    ws.on('message', onMessageHandler(wss, options))
+  }
+}
+
+export function onMessageHandler(wss: WebSocketServer, options: SetupWebSocketConnectionOptions) {
```
shauns, line 39: Name is a bit misleading as its not the handler, its a factory that _gives_ you the handler.

packages/app/src/cli/services/dev/extension/websocket/handlers.ts
```diff
@@ -0,0 +1,110 @@
...
+        options.payloadStore.updateExtensions(eventData.extensions)
+      }
+    } else if (eventType === 'dispatch') {
+      const outGoingMessage = getOutgoingDispatchMessage(jsonData, options)
+      output.debug(output.content`Sending websocket dispatch with event type ${outGoingMessage.data.type} and data:
+${output.token.json(outGoingMessage.data)}
+        `)
+
+      notifyClients(wss, outGoingMessage)
+    }
```
shauns, line 67: Should we log other `eventType`?

packages/app/src/cli/services/dev/extension/utilities.ts
```diff
@@ -0,0 +1,29 @@
...
+}
+
+/**
+ * Returns true if an extension needs a cart URL.
+ */
+export function getExtensionNeedsCartURL(extension: UIExtension) {
+  return extension.configuration.type === 'checkout_ui_extension'
+}
+
+export async function getLastUpdatedTimestamp(path: string) {
```
shauns, line 25: I don't think this really belongs in `services/dev` -- feels much more generic?

packages/app/src/cli/services/dev/extension/websocket/handlers.ts
```diff
@@ -0,0 +1,110 @@
...
+    const {event: eventType, data: eventData} = jsonData
+
+    output.debug(output.content`Received websocket message with event type ${eventType} and data:
+${output.token.json(eventData)}
+          `)
+
+    if (eventType === 'update') {
+      /**
+       * App updates must take precedence over extensions. Otherwise the websocket server
+       * will send an updte to the client with missing app data and will cause the loading
```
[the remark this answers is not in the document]
byrichardpowell: Thanks, fixed 👍

packages/app/src/cli/services/dev/extension/server/middlewares.ts
```diff
@@ -0,0 +1,229 @@
...
+          url: new URL('/extensions/dev-console', devOptions.url).toString(),
+        },
+        store: devOptions.storeFqdn,
+        extension: await getUIExtensionPayload(extension, devOptions),
+      }),
+    )
+  }
+}
+
+function getWebscoketUrl(devOptions: ExtensionsPayloadMiddlewareOptions['devOptions']) {
```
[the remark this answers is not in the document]
byrichardpowell: Oops, fixed :+1:

packages/app/src/cli/services/dev/extension/utilities.ts
byrichardpowell: Yeah, fair point.  I've moved it to  `cli-kit` and renamed it to `lastUpdatedTimestamp` be more consistent with the pre-existing `lastUpdated()` utility.

packages/app/src/cli/services/dev/extension/websocket/handlers.ts
byrichardpowell: Yeah, fair.  

I've reviewed all our handlers.  We now use `getSomeHander()` if the function returns a handler, or `someHandler()` if the function is the handler

byrichardpowell:
@pepicrft  I've updated the PR Description with some details to help testers.  We could also use these as a starting points for any docs we want to add.

TobiasBales (approved):
Seems to work fine with customer accounts

packages/app/src/cli/services/dev/extension.ts
```diff
@@ -1,59 +1,25 @@
...
-   */
   stdout: Writable
-  /**
-   * Standard error stream to send the error output through.
-   */
   stderr: Writable
-
-  /**
-   * Signal to abort the build process.
-   */
```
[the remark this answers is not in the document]
dmhenry: Restored

packages/app/src/cli/services/dev/extension/localization.ts
```diff
@@ -0,0 +1,93 @@
...
+    await output.warn(message)
+    throw new Error(message)
+  }
+}
+
+async function calculateLastUpdatedTimestamp(path: string, localization: Localization) {
+  const lastUpdatedDateTime = await file.lastUpdatedTimestamp(path)
+
+  if (lastUpdatedDateTime > localization.lastUpdated) {
+    localization.lastUpdated = lastUpdatedDateTime
```
vividviolet, line 91: This would mutate the `lastUpdated` timestamp even if the localization fails to build which might break Checkout. Currently they are only using the `lastUpdated` timestamp to know when to update the localization data  - they expect the data to be valid instead of checking the build status

packages/app/src/cli/services/dev/extension/payload.ts
```diff
@@ -0,0 +1,71 @@
...
+): Promise<UIExtensionPayload> {
+  const url = `${options.url}/extensions/${extension.devUUID}`
+  const configurationCapabilities = extension.configuration.capabilities
+    ? Object.entries(extension.configuration.capabilities).reduce((acc, [key, value]) => {
+        acc[string.camelize(key)] = value as boolean
+        return acc
+      }, {} as Capabilities)
+    : {}
+  const capabilities = {
+    blockProgress: false,
```
vividviolet, line 26: Why do we need to hardcode these values here?

packages/app/src/cli/services/dev/extension/websocket/handlers.ts
```diff
@@ -0,0 +1,111 @@
...
+${output.token.json(eventData)}
+          `)
+
+    if (eventType === 'update') {
+      /**
+       * App updates must take precedence over extensions. Otherwise the websocket server
+       * will send an update to the client with missing app data and will cause the loading
+       * of extensions to fail.
+       */
+      if (eventData.app) {
```
vividviolet, line 54: This is missing a check to verify that the app.apiKey matches the one in memory - otherwise the update should be ignored

packages/app/src/cli/services/dev/extension/localization.test.ts
```diff
@@ -0,0 +1,122 @@
...
+        en: {greeting: 'Hi!'},
+        fr: {greeting: 'Bonjour!'},
+      })
+    })
+  })
+
+  it('returns the lastUpdated timestamp of the most recently updated locale', async () => {
+    await file.inTemporaryDirectory(async (tmpDir) => {
+      const timestamps = [457, 789, 132]
+      vi.spyOn(file, 'lastUpdatedTimestamp').mockImplementation(async () => {
```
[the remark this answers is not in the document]
byrichardpowell: Chatted with Trish.  Some notes:

1. The actual number doesn't matter as long as it increments...  But the using the timestamp when `getLocalization()` completed is simpler, so we should simplify.
2. If localization changes and there is an error we should send the last valid localization object for `.localization`, but  `development.localizationStatus` should be `errror`.

packages/cli-kit/src/session/redirect-listener.ts
```diff
@@ -12,8 +12,9 @@ import {
 } from './post-auth.js'
 import {Abort, Bug} from '../error.js'
 import {content, info, token} from '../output.js'
-import Fastify from 'fastify'
+import {createApp, IncomingMessage, ServerResponse} from 'h3'
```
isaacroldan, line 15: Why are we replacing `fastify` for `h3`?

packages/cli-kit/src/http.ts
```diff
@@ -1,11 +1,25 @@
...
+  createApp,
+  createRouter,
+  IncomingMessage,
+  ServerResponse,
+  CompatibilityEvent,
+  createError,
+  send,
+  sendError,
+  sendRedirect,
+} from 'h3'
```
isaacroldan, line 21: I guess you can add `H3Error` to the same list as the rest of exported items

packages/cli-kit/src/common/object.ts
```diff
@@ -0,0 +1,13 @@
...
+
+/**
+ * Deep merges the two objects and returns a new object
+ * with the merge result.
+ * @param lhs {any} One of the objects to be merged.
+ * @param rhs {any} Another object to be merged.
+ * @returns {any} A Javascrip tobject with th emerged objects.
+ */
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export function deepMergeObjects(lhs: any, rhs: any): any {
```
isaacroldan, line 11: can `typescript` do its magic and infer the result?
@shauns you are the expert in types, would something like this be possible?
```ts
deepMergeObjects<TLhs, TRhs>(lhs: TLhs, rhs: TRhs): TLhs & TRhs
```

isaacroldan: Actually, deepmerge is already typed... [link](https://github.com/TehShrike/deepmerge/blob/master/index.d.ts#L2)
Can we avoid using `any` here?

packages/cli-kit/src/session/redirect-listener.ts
pepicrft: To reduce the number of dependencies (4 vs 14). Since the HTTP servers that we need are for development, we think it's a good idea to stay lean there and not bring the sophistication of production-oriented servers like Express and Fastify.

packages/app/src/cli/services/dev.ts
```diff
@@ -282,26 +289,13 @@ async function devUIExtensionsTarget(
         storeFqdn,
         apiKey,
         grantedScopes,
-        cartUrl,
+        checkoutCartUrl,
```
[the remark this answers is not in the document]
alfonso-noriega: Right, that generation of the cart Url was missing.

packages/app/src/cli/services/dev/extension/websocket/handlers.ts
```diff
@@ -0,0 +1,113 @@
...
+
+    if (eventType === 'update') {
+      /**
+       * App updates must take precedence over extensions. Otherwise the websocket server
+       * will send an update to the client with missing app data and will cause the loading
+       * of extensions to fail.
+       */
+      const payloadStoreApiKey = options.payloadStore.getRawPayload().app.apiKey
+      const eventAppApiKey = eventData.app?.apiKey
+      if (eventData.app && payloadStoreApiKey === eventAppApiKey) {
```
byrichardpowell, line 56: @pepicrft @alfonso-noriega @vividviolet If these don't match should we return early?

Right now if they don't match we might still update the extensions.  My guess is if the API Key doesn't match it's a signal we shouldn't trust anything about this message and we should bail early.

packages/app/src/cli/services/dev/extension/bundler.ts
```diff
@@ -0,0 +1,84 @@
...
+            error ? 'an error' : 'changed'
+          }`,
+        )
+
+        options.payloadStore
+          .updateExtension(extension, {
+            status: error ? 'error' : 'success',
+          })
+          .then((_) => {})
+          .catch((_) => {})
```
matteodepalo, line 46: Any specific reason why we call `then`, `catch` instead of using `try`/`catch`? Not sure if we follow this convention everywhere in the codebase, but I think we use `try`/`catch` blocks more. In this PR there are other occurrences of this if you search by `.catch`.

matteodepalo: @pepicrft answered this offline, `esbuild` doesn't allow us to pass `async` functions. It might still be worth adding some logs here and also when calling `then`.

packages/app/src/cli/services/dev/extension/server/utilities.ts
```diff
@@ -0,0 +1,33 @@
...
+    rawUrl.searchParams.append('dev', `${options.url}/extensions`)
+
+    return rawUrl.toString()
+  } else {
+    const rawUrl = new URL(`https://${options.storeFqdn}/`)
+    rawUrl.pathname = 'admin/extensions-dev'
+    rawUrl.searchParams.append('url', getExtensionUrl(extension, options))
+
+    return rawUrl.toString()
+  }
```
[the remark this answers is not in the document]
vividviolet: By the way, you've tagged the wrong Trish 😅

packages/app/src/cli/services/dev/extension/websocket/handlers.ts
vividviolet: It should return early and not continue processing the message

byrichardpowell: David and I updated this.

packages/app/src/cli/services/dev/extension/websocket/handlers.ts
dmhenry: Done

packages/app/src/cli/services/dev/extension/server/utilities.ts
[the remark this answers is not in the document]
dmhenry: Leaving this for now.

PR #456, opened 2022-09-12
Make using the Node logic for building UI extensions in `deploy` opt-in

pepicrft:
As part of [migrating the build of UI extensions from Go to Node](https://github.com/Shopify/cli/pull/437) I made it opt-in via an environment variable, `SHOPIFY_CLI_UI_EXTENSIONS_USE_NODE=1`. However, I forgot to update the `deploy` logic to make the new logic opt-in there as well.

I'm updating the `deploy` workflow to use the Node logic for building UI extensions only if the `SHOPIFY_CLI_UI_EXTENSIONS_USE_NODE` environment variable is truthy.

### How to test your changes?
Try to deploy a Shopify app with and without the `SHOPIFY_CLI_UI_EXTENSIONS_USE_NODE=1` environment variable.

### Measuring impact

### Checklist

### DIFF
--- a/packages/app/src/cli/services/build/extension.ts
+++ b/packages/app/src/cli/services/build/extension.ts
@@ -1,7 +1,9 @@
 import {AppInterface} from '../../models/app/app.js'
 import {UIExtension, FunctionExtension, ThemeExtension} from '../../models/app/extensions.js'
 import {bundleExtension} from '../extensions/bundle.js'
-import {error, system, abort} from '@shopify/cli-kit'
+import {extensionConfig} from '../../utilities/extensions/configuration.js'
+import {runGoExtensionsCLI} from '../../utilities/extensions/cli.js'
+import {error, system, abort, environment, output, yaml} from '@shopify/cli-kit'
 import {execThemeCheckCLI} from '@shopify/cli-kit/node/ruby'
 import {Writable} from 'node:stream'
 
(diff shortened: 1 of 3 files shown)


packages/app/src/cli/services/build/extension.ts
```diff
@@ -54,6 +56,47 @@ export async function buildThemeExtensions(options: ThemeExtensionBuildOptions):
...
 
+interface BuildUIExtensionsOptions {
+  app: AppInterface
+}
+
+export function buildUIExtensions(options: BuildUIExtensionsOptions): output.OutputProcess[] {
+  if (options.app.extensions.ui.length === 0) {
+    return []
+  }
+  if (environment.utilities.isTruthy(process.env.SHOPIFY_CLI_UI_EXTENSIONS_USE_NODE)) {
```
isaacroldan, line 67: should we have this kind of checks abstracted?
We have `local.ts` with a bunch of them but i guess only for cli-kit, maybe makes sense to make that file public, or have a similar one for `app`.
What do you think?

isaacroldan (approved):

packages/app/src/cli/services/build/extension.ts
pepicrft: Note that this is very temporary. It'll go away once we have everything migrated and well tested.

PR #461, opened 2022-09-13
Device auth improvements & fixes

isaacroldan:
Got some feedback from @pepicrft on a merged PR so this one mostly addresses his comments.
Also enables gitpod and codespaces to work without tunnels.

- Address issues from the last PR:
- Improve tests
- Make `tokenRequest` use the new `Result` 
- Enable gitpod and codespaces to work without tunnels.

### How to test your changes?
Device auth still works (using `SHOPIFY_CLI_DEVICE_AUTH`)

### Measuring impact

### Checklist

### DIFF
--- a/packages/cli-kit/src/environment/local.ts
+++ b/packages/cli-kit/src/environment/local.ts
@@ -50,7 +50,7 @@ export async function isShopify(env = process.env): Promise<boolean> {
     return !isTruthy(env[constants.environmentVariables.runAsUser])
   }
   const devInstalled = await fileExists(constants.paths.executables.dev)
-  return devInstalled || isSpin()
+  return devInstalled || isSpin(env)
 }
 
 /**
@@ -110,10 +110,6 @@ function isCodespaces(env = process.env): boolean {
 
 function isGitpod(env = process.env): boolean {
   return isSet(env[constants.environmentVariables.gitpod])
-}
-
-function isSpin(env = process.env): boolean {
-  return isTruthy(env[constants.environmentVariables.spin])
 }
 
 /**
--- a/packages/cli-kit/src/environment/local.test.ts
+++ b/packages/cli-kit/src/environment/local.test.ts
@@ -1,4 +1,3 @@
-import {isSpin} from './spin.js'
 import {hasGit, isDevelopment, isShopify, isUnitTest, analyticsDisabled, useDeviceAuth} from './local.js'
 import {exists as fileExists} from '../file.js'
 import {exec} from '../system.js'
@@ -6,9 +5,6 @@ import {expect, it, describe, vi, test} from 'vitest'
 
 vi.mock('../file')
 vi.mock('../system')
-vi.mock('./spin', () => ({
-  isSpin: vi.fn(),
-}))
 
 describe('isUnitTest', () => {
   it('returns true when SHOPIFY_UNIT_TEST is truthy', () => {
@@ -63,10 +59,10 @@ describe('isShopify', () => {
 
   it('returns true when it is a spin environment', async () => {
     // Given
-    vi.mocked(isSpin).mockReturnValue(true)
+    const env = {SPIN: '1'}
 
     // When
-    await expect(isShopify()).resolves.toBe(true)
+    await expect(isShopify(env)).resolves.toBe(true)
   })
 })
 


packages/app/src/cli/services/dev.ts
```diff
@@ -51,7 +51,14 @@ async function dev(options: DevOptions) {
...
 
-  if (options.noTunnel === true) {
+  if (environment.local.codespaceURL()) {
+    frontendPort = 4040
+    frontendUrl = `https://${environment.local.codespaceURL()}-${frontendPort}.githubpreview.dev`
+  } else if (environment.local.gitpodURL()) {
+    const defaultUrl = environment.local.gitpodURL()?.replace('https://', '')
+    frontendPort = 4040
+    frontendUrl = `https://${frontendPort}-${defaultUrl}`
+  } else if (options.noTunnel === true) {
```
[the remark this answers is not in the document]
isaacroldan: The good thing about this is that the URL will always be the same 👌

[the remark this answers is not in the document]
pepicrft: Nice! I agree we should use the port that they open and minimize the actions that users need to take.

packages/cli-kit/src/environment/local.ts
```diff
@@ -90,6 +90,14 @@ export function useDeviceAuth(env = process.env): boolean {
   return isTruthy(env[constants.environmentVariables.deviceAuth]) || isCloudEnvironment(env)
 }
 
+export function gitpodURL(env = process.env): string | undefined {
+  return env[constants.environmentVariables.gitpod]
+}
+
+export function codespaceURL(env = process.env): string | undefined {
```
pepicrft, line 97: Similar comment here regarding GitHub Codespaces

packages/cli-kit/src/session/device-authorization.ts
```diff
@@ -14,6 +16,10 @@ export interface DeviceAuthorizationResponse {
   interval?: number
 }
 
+const DeviceAuthError = () => {
+  return new Abort('Failed to start authorization process')
```
pepicrft, line 20: This is not very actionable. Should we make it a `Bug` to have some metrics on Bugsnag around how often and when it happens?

isaacroldan: Will do, there is nothing for the user to do... just try again.

packages/cli-kit/src/session/exchange.ts
```diff
@@ -152,42 +159,47 @@ async function requestAppToken(
...
+function tokenRequestErrorHandler(error: string) {
+  if (error === 'invalid_grant') {
+    // There's an scenario when Identity returns "invalid_grant" when trying to refresh the token
+    // using a valid refresh token. When that happens, we take the user through the authentication flow.
+    throw new InvalidGrantError()
+  }
+  if (error === 'invalid_request') {
+    // There's an scenario when Identity returns "invalid_request" when exchanging an identity token.
+    // This means the token is invalid. We clear the session and throw an error to let the caller know.
+    throw InvalidIdentityError
```
pepicrft, line 184: I'd change this one from:

```js
const InvalidIdentityError = new Abort(
  '\nError validating auth session',
  "We've cleared the current session, please try again",
)
```

To: 

```js
const InvalidIdentityError = () => new Abort(
  '\nError validating auth session',
  "We've cleared the current session, please try again",
)
```
Otherwise the stacktrace won't be accurate because it's captured when the error is instantiated. Then this piece of code should be:

```js
throw InvalidIdentityError()
```

packages/cli-kit/src/session/exchange.ts
```diff
@@ -152,42 +159,47 @@ async function requestAppToken(
...
-      throw InvalidIdentityError
-    } else {
-      if (!throws) return payload
-      throw new Abort(payload.error_description)
-    }
-  }
-  return payload
+
+  if (res.ok) return ok(payload)
+  if (payload.error === 'invalid_request') await secureStore.remove()
```
pepicrft, line 198: The `await secureStore.remove()` side effect feels very deep in the logic. What about letting the logic that coordinates everything to deal with it?

packages/cli-kit/src/session/exchange.ts
isaacroldan: good catch 👌

pepicrft (approved):

packages/cli-kit/src/session/exchange.ts
isaacroldan: I thought of moving it to the errorHandler, but didn't want to make that function async since it's just a mapper... I'll take another look 👌

packages/cli-kit/src/constants.ts
```diff
@@ -29,7 +29,8 @@ const constants = {
     // Variables to detect if the CLI is running in a cloud environment
     spin: 'SPIN',
     codespaces: 'CODESPACES',
-    gitpod: 'GITPOD_WORKSPACE_ID',
```
shauns, line 32: there's a `webIDEPlatform` function in environment.local that we use for analytics. Would be nice to use these there so that we can track where the CLI is running. currently it only knows about CODESPACES.

isaacroldan: good catch, I thought we didn't have this tracked. Added gitpod 👌

PR #468, opened 2022-09-14
Send the command generate as scaffold in analytics events

pepicrft:
Addresses: https://github.com/Shopify/cli/pull/376#discussion_r961440781

This PR addresses @shauns's point in [this comment](https://github.com/Shopify/cli/pull/376#discussion_r961440781). With the recent rename of `scaffold` to `generate` we'd send events with `command: generate` resulting in two types of events that represent the same thing but with different names. This is prone to confusion in the future because any queries and dashboards will have to take this into account.

I'm adjusting the analytics logic to allow commands to override the command name. 

### How to test your changes?
Run `yarn shopify app generate extension --path ./fixtures/app --verbose` and you should see in the analytics event that was sent that `command_name` has `scaffold` as value.

### Measuring impact

### Checklist


packages/cli-kit/src/node/hooks/prerun.ts
```diff
@@ -8,5 +9,5 @@ export const hook: Hook.Prerun = async (options) => {
   const command = cmd.replace(/:/g, ' ')
   const args = options.argv
   debug(`Running command ${command}`)
-  await start({command, args, commandClass: options.Command})
+  await start({command, args, commandClass: options.Command as unknown as typeof Command})
```
pepicrft, line 12: @shauns I'm assuming here that all plugins use the base command class exported by `cli-kit`. I think it's a safe assumption but want to know your thoughts about this.

isaacroldan (approved):

PR #471, opened 2022-09-14
Propagate --verbose to oclif's logging

pepicrft:
The `DEBUG=*` environment variable has been quite handy in numerous time to debug issues coming from oclif. They were most of the times related to modules that failed to load. The problem with the variable is that it's not obvious that it exists. Moreover, there's already another flag that developers are already familiar with `--verbose` and serves a similar purpose.

I'm adjusting the function that triggers `oclif` to set the environment variable `DEBUG=*` when the `--verbose` flag is passed. Note that I had to make some changes in how the modules are loaded because static imports led to `debug`, the library that oclif uses for logging, loading and caching the value of the variable before we set it at runtime.

### How to test your changes?
Easy, run any Shopify command with `--verbose ` and you should see all the oclif logs.

### Measuring impact

### Checklist


packages/cli-kit/src/node/cli.ts
```diff
@@ -16,6 +26,17 @@ interface RunCLIOptions {
  * @param module {RunCLIOptions} Options.
  */
 export async function runCLI(options: RunCLIOptions) {
+  setupEnvironmentVariables(options)
+
+  /**
+   * These imports need to be dynamic because if they are static
+   * they are loaded before se set the DEBUG=* environment variable
```
cartogram, line 33:
```suggestion
   * they are loaded before we set the DEBUG=* environment variable
```

packages/cli-kit/src/node/cli.ts
```diff
@@ -28,6 +49,15 @@ export async function runCLI(options: RunCLIOptions) {
  * @param options
  */
 export async function runCreateCLI(options: RunCLIOptions) {
+  /**
+   * We need to call this method before we do any imports because they
+   * migth transitively initialize debug and DEBUG=* has no effect then.
```
cartogram, line 54:
```suggestion
   * might transitively initialize debug and DEBUG=* has no effect then.
```

packages/cli-main/src/index.ts
```diff
@@ -1,8 +1,9 @@
 import {runCLI} from '@shopify/cli-kit/node/cli'
 
-async function runShopifyCLI() {
+async function runShopifyCLI(development: boolean) {
```
gonzaloriestra, line 3: What about using an object for the parameter instead so that you can see what the parameter means? That way we could use `runCLI({development: true})` instead of `runCLI(true)`. And maybe a default value of false?

gonzaloriestra (approved):
Great idea!!

packages/cli-main/src/index.ts
pepicrft: Much better. I'll update it.

pepicrft:
>  Aren't oclif logs way too verbose?
> This also affects the frontend/backend servers no? I just tried and vite prints a hundred lines at least 😬

Oclif uses [debug](https://www.npmjs.com/package/debug) internally, which supports a lot of customization. I'd say that we merge it as it is, and consider adding verbosity levels in the future throughout the CLI if we see there's a need for more granularity.

PR #489, opened 2022-09-19
Add additional baseline instrumentantion

alvaro-shopify:
Fixes https://github.com/Shopify/shopify-cli-planning/issues/346
Fixes https://github.com/Shopify/shopify-cli-planning/issues/347

- [x] Update the monorail topic name
- [x] Track `cmd_all`, `cmd_extensions`, `cmd_scaffold`, `app`, `env` topics
- [x] Change metadata tracking to be behind a catch-all by default

### How to test your changes?
Run with `--verbose`, take a look at the analytics payload. Particularly dev command.

### Measuring impact

### Checklist


packages/cli-kit/src/node/node-package-manager.ts
```diff
@@ -444,3 +446,53 @@ export async function findUpAndReadPackageJson(fromDirectory: string): Promise<{
     throw FindUpAndReadPackageJsonNotFoundError(fromDirectory)
   }
 }
+
+export async function resolveFramework(frontRootDirectory: string) {
```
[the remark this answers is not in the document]
pepicrft: Indeed, I see there's already `frameworks.ts`. Why not moving it there?

yarn.lock
```diff
@@ -8757,6 +8757,11 @@ lru-cache@^6.0.0:
   dependencies:
     yallist "^4.0.0"
 
+macaddress@^0.5.3:
```
pepicrft, line 8760: Have we considered generating a unique id in the environment and persists across runs over using PII information?

packages/cli-main/src/cli/commands/upgrade.ts
```diff
@@ -7,6 +7,7 @@ import {
   getPackageManager,
 } from '@shopify/cli-kit/node/node-package-manager'
 import Command from '@shopify/cli-kit/node/base-command'
+import {getCliProjectDir} from '@shopify/cli-kit/src/cli.js'
```
pepicrft, line 10: This import is invalid and will break in production. You'll have to use the exported modules instead:

```ts
import { cli } from "@shopify/cli-kit"
```

packages/cli-kit/src/node/hooks/prerun.ts
```diff
@@ -1,13 +1,54 @@
...
-import {Hook} from '@oclif/core'
+import {Hook, Interfaces} from '@oclif/core'
 
+export declare interface CommandContent {
+  command: string
+  topic?: string
+  alias?: string
+}
 // This hook is called before each command run. More info: https://oclif.io/docs/hooks
 export const hook: Hook.Prerun = async (options) => {
```
pepicrft, line 12: This one seems a good candidate for a unit test.

packages/cli-kit/src/cli.ts
```diff
@@ -11,3 +13,24 @@ export const globalFlags = {
     env: 'SHOPIFY_FLAG_VERBOSE',
   }),
 }
+
+export async function isCliProject(directory: string) {
```
pepicrft, line 17: I'd consider defining an internal plugin contract around this and get the `@shopify/app`, `@shopify/cli-hydrogen`, and `@shopify/theme` packages to implement it. Otherwise `@shopify/cli-kit` has knowledge on the packages that are built on it, and it shouldn't.  https://en.wikipedia.org/wiki/Dependency_inversion_principle

Untitled Diagram-Page-3 drawio

packages/cli-kit/src/environment/local.ts
```diff
@@ -154,7 +160,12 @@ export function ciPlatform(env = process.env): {isCI: true; name: string} | {isC
  * Gets info on the Web IDE platform the CLI is running on, if applicable
  */
 export function webIDEPlatform(env = process.env) {
```
pepicrft, line 162: `webIDEPlatform` and `cloudEnvironment` seem somewhat related. The former is a subset of the data returned by the latter. What about having a single function and adjusting the return type to allow filtering:

```js
{ platform: 'codespaces', editor: false },
{ platform: 'gitpod', editor: true }
{ platform: 'spin', editor: false }
```

packages/cli-kit/src/cli.ts
alvaro-shopify: Finally I have reused the existing method `localCliPackage` to detect the using of a local CLI instance

yarn.lock
alvaro-shopify: @shauns in the spreadsheet the note about this field indicates `MAC address (in theory unique) of the machine executing the command`, do we only need the `macaddress` as a unique identifier of the CLI instance?, if that is the case, I agree with @pepicrft about generating a unique id. The problem would be that in case the environment variables are cleared, the relationship with previous execution will be lost

alvaro-shopify: Finally the mac address fields has been removed and added a new one `env_device_id` instead. The value of this field will be the mac address hash to avoid the creation of a new unique id that should be stored across executions

PR #493, opened 2022-09-19
Extract the business logic of the version command into a service

pepicrft:
While working on the migration from Go to Node I noticed the tests of the `version` command started failing even though there's no connection between that command and the extensions' serve logic.

Placing business logic in commands is a discouraged pattern because it leads to integration tests that test the integration of the business logic and oclif's setup. Therefore I created a service, `versionService` that allows unit-testing the logic in isolation.

### How to test your changes?
Tests should pass and `version` should yield the right content.

### Measuring impact

### Checklist

### DIFF
--- a/packages/cli-main/src/cli/services/commands/version.test.ts
+++ b/packages/cli-main/src/cli/services/commands/version.test.ts
@@ -1,6 +1,6 @@
-import Version from './version'
+import {versionService} from './version'
 import {afterEach, beforeEach, describe, expect, it, vi} from 'vitest'
-import {outputMocker, output} from '@shopify/cli-kit'
+import {outputMocker, output, error} from '@shopify/cli-kit'
 import {
   checkForNewVersion,
   PackageManager,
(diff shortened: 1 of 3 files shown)


packages/cli-main/src/cli/services/commands/version.test.ts
```diff
@@ -40,7 +39,9 @@ describe('check CLI version', () => {
       const outputReminder = vi.mocked(output.getOutputUpdateCLIReminder).mockReturnValue('CLI reminder')
 
       // When
-      await Version.run()
+      await expect(async () => {
+        await versionService({currentVersion})
```
gonzaloriestra, line 43: Is the inner await needed? 🤔

gonzaloriestra (approved):

packages/cli-main/src/cli/services/commands/version.test.ts
pepicrft: Alternatively we can return the promise:
```
return versionService({currentVersion})
```

PR #494, opened 2022-09-19
Add missing import that causes the production version of the CLI to fail

pepicrft:
A recent refactoring in the executables led to a missing import that will cause the production version of the CLI not to work.

I'm adding the missing import.

### How to test your changes?
1. `yarn build`.
2. `node packages/cli-main/bin/run.js` should run.

### Measuring impact

### Checklist

### DIFF
--- a/.changeset/large-dodos-unite.md
+++ b/.changeset/large-dodos-unite.md
@@ -1,0 +1,5 @@
+---
+'@shopify/cli': patch
+---
+
+Add missing import that causes the CLI execution to fail in production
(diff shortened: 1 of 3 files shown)


gonzaloriestra (approved):
Good catch!! To prevent this problem again, I'm changing the acceptance tests to use the production executables

pepicrft:
> Good catch!! To prevent this problem again, I'm changing the acceptance tests to use the production executables

Great idea :+1:

PR #496, opened 2022-09-19
Make upgrade work on global installs

amcaplan:

Fixes https://github.com/Shopify/shopify-cli-planning/issues/378

We have an upgrade command that works on per-project installs, but not on system-wide installs.

I realized we will nearly always do `npm install -g` to upgrade, even if we end up supporting Linux with explicit packages, and the only real exception is Homebrew. So may as well just have this command work in global installs.

* Checks whether we're running the CLI inside a package manager. If yes, proceed as before.
* If not, apparently we're running the CLI globally and we should upgrade globally.
* Check if it's been installed via Homebrew, and install the appropriate formula.
* If it's not installed via Homebrew, install via `npm install -g`.

I also added significant testing around the whole flow.

Note that I could have opted for a service object, but for now I wanted to avoid introducing that concept to cli-main, especially since this is a specialized workflow that isn't likely to change.

### How to test your changes?

```sh
$ bin/run.js upgrade --path=/Users/YOURUSER
```

Try with/without Homebrew-installed CLI, with an installed older version, etc.

### Measuring impact

### Checklist

### DIFF
--- a/packages/cli-kit/src/version.ts
+++ b/packages/cli-kit/src/version.ts
@@ -1,5 +1,5 @@
 import {content, token, debug} from './output.js'
-import {moduleDirectory} from './path.js'
+import {findUp, moduleDirectory} from './path.js'
 import {Bug} from './error.js'
 import {findUpAndReadPackageJson} from './node/node-package-manager.js'
 import latestVersion from 'latest-version'
--- a/packages/cli-kit/src/version.ts
+++ b/packages/cli-kit/src/version.ts
@@ -1,5 +1,5 @@
 import {content, token, debug} from './output.js'
-import {findUp, moduleDirectory} from './path.js'
+import {moduleDirectory} from './path.js'
 import {Bug} from './error.js'
 import {findUpAndReadPackageJson} from './node/node-package-manager.js'
 import latestVersion from 'latest-version'
--- a/packages/cli-main/src/cli/commands/upgrade.test.ts
+++ b/packages/cli-main/src/cli/commands/upgrade.test.ts
@@ -1,6 +1,6 @@
 import Upgrade from './upgrade'
 import {afterEach, beforeEach, describe, expect, it, vi} from 'vitest'
-import {constants, file, outputMocker} from '@shopify/cli-kit'
+import {constants, file, outputMocker, path} from '@shopify/cli-kit'
 // import {
   // checkForNewVersion,
 // } from '@shopify/cli-kit/node/node-package-manager'
@@ -34,6 +34,31 @@ describe('upgrade global CLI', () => {
     const currentCliVersion = await constants.versions.cliKit()
     await file.inTemporaryDirectory(async (tmpDir) => {
       // Given
+      const outputMock = outputMocker.mockAndCaptureOutput()
+      vi.spyOn(Upgrade.prototype as any, 'parse').mockResolvedValue({flags: {path: tmpDir}})
+      vi.spyOn(Upgrade.prototype as any, 'getCurrentVersion').mockReturnValue(currentCliVersion)
+      vi.spyOn(Upgrade.prototype as any, 'usingPackageManager').mockReturnValue(false)
+
+      // When
+      await Upgrade.run()
+
+      // Then
+      expect(outputMock.info()).toMatchInlineSnapshot(`
+        "You're on the latest version, ${currentCliVersion}, no need to upgrade!"
+      `)
+    })
+  })
+})
+
+describe('upgrade local CLI', () => {
+  it('does not upgrade locally if the latest version is found', async () => {
+    const currentCliVersion = await constants.versions.cliKit()
+    await file.inTemporaryDirectory(async (tmpDir) => {
+      // Given
+      await file.write(
+        path.join(tmpDir, 'package.json'),
+        JSON.stringify({dependencies: {'@shopify/cli': currentCliVersion, '@shopify/app': currentCliVersion}}),
+      )
       const outputMock = outputMocker.mockAndCaptureOutput()
       vi.spyOn(Upgrade.prototype as any, 'parse').mockResolvedValue({flags: {path: tmpDir}})
       vi.spyOn(Upgrade.prototype as any, 'getCurrentVersion').mockReturnValue(currentCliVersion)
--- a/packages/cli-main/src/cli/commands/upgrade.test.ts
+++ b/packages/cli-main/src/cli/commands/upgrade.test.ts
@@ -1,6 +1,6 @@
 import Upgrade from './upgrade'
 import {afterEach, beforeEach, describe, expect, it, vi} from 'vitest'
-import {constants, file, outputMocker, path, system} from '@shopify/cli-kit'
+import {constants, file, os, outputMocker, path, system} from '@shopify/cli-kit'
 // import {
   // checkForNewVersion,
 // } from '@shopify/cli-kit/node/node-package-manager'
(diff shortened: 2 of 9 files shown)


packages/cli-main/src/cli/commands/upgrade.ts
```diff
@@ -21,43 +22,27 @@ export default class Upgrade extends Command {
   }
 
   async run(): Promise<void> {
```
pepicrft, line 24: I'd advise extracting all the business logic into a service to allow unit-testing instead of integration-testing through oclif's primitives.

packages/cli-main/src/cli/commands/upgrade.ts
```diff
@@ -1,12 +1,13 @@
...
-import {error, file, output, path} from '@shopify/cli-kit'
+import {error, file, os, output, path, system} from '@shopify/cli-kit'
 import {
   addNPMDependencies,
   checkForNewVersion,
   DependencyType,
   getPackageManager,
 } from '@shopify/cli-kit/node/node-package-manager'
 import Command from '@shopify/cli-kit/node/base-command'
+const cliDependency = '@shopify/cli'
```
pepicrft, line 10: Considering this information is in the `package.json` of the package containing this module, I'd keep that as the source of truth and read the information at runtime.

packages/cli-kit/src/version.ts
```diff
@@ -1,5 +1,5 @@
 import {content, token, debug} from './output.js'
-import {moduleDirectory} from './path.js'
+import {findUp, moduleDirectory} from './path.js'
```
pepicrft, line 2: Is this change needed?

packages/cli-main/src/cli/commands/upgrade.test.ts
```diff
@@ -0,0 +1,180 @@
...
+  })
+  vi.mocked(os.platformAndArch).mockReturnValue({platform: 'win32', arch: 'amd64'})
+})
+afterEach(() => {
+  outputMocker.mockAndCaptureOutput().clear()
+})
+
+describe('upgrade global CLI', () => {
+  it('does not upgrade globally if the latest version is found', async () => {
+    const currentCliVersion = await constants.versions.cliKit()
```
pepicrft, line 36: Why not moving this one inside the `inTemporaryDirectory`?

packages/cli-main/src/cli/services/upgrade.ts
```diff
@@ -0,0 +1,174 @@
...
+  findUpAndReadPackageJson,
+  checkForNewVersion,
+  DependencyType,
+  getPackageManager,
+} from '@shopify/cli-kit/node/node-package-manager'
+
+// Canonical list of oclif plugins that should be installed globally
+const globalPlugins = ['@shopify/theme']
+
+interface PackageJsonContents {
```
pepicrft, line 13: What about reusing [this model](https://github.com/Shopify/cli/blob/main/packages/cli-kit/src/node/node-package-manager.ts#L210) from `@shopify/cli-kit`?

packages/cli-main/src/cli/services/upgrade.ts
```diff
@@ -0,0 +1,174 @@
...
+  }
+}
+
+export async function upgrade(directory: string, currentVersion: string): Promise<void> {
+  let newestVersion: string | void
+
+  const projectDir = await getProjectDir(directory)
+  if (projectDir) {
+    newestVersion = await upgradeLocalShopify(projectDir, currentVersion)
+  } else if (usingPackageManager()) {
```
pepicrft, line 28: I find this clause a bit confusing. What does it mean that the `npm_config_user_agent` environment variable is not present? Who is setting that variable and when?

packages/cli-main/src/cli/services/upgrade.ts
```diff
@@ -0,0 +1,174 @@
...
+async function upgradeLocalShopify(projectDir: string, currentVersion: string): Promise<string | void> {
+  const packageJson = (await findUpAndReadPackageJson(projectDir)).content as PackageJsonContents
+  const packageJsonDependencies = packageJson.dependencies || {}
+  const packageJsonDevDependencies = packageJson.devDependencies || {}
+
+  let resolvedVersion: string = {...packageJsonDependencies, ...packageJsonDevDependencies}[await cliDependency()]!
+  if (resolvedVersion.slice(0, 1).match(/[\^~]/)) resolvedVersion = currentVersion
+  const newestVersion = await checkForNewVersion(await cliDependency(), resolvedVersion)
+
+  if (!newestVersion) return wontInstall(resolvedVersion)
```
pepicrft, line 69: `wontInstall` doesn't convey what the method does. What about:

```js
if (!newestVersion) {
  outputWontInstallMessage(resolvedVersion)
  return;
}

File: packages/cli-main/src/cli/services/upgrade.ts [hunk 8]
```diff
@@ -0,0 +1,174 @@
...
+
+  outputUpgradeMessage(resolvedVersion, newestVersion)
+
+  await installJsonDependencies('prod', packageJsonDependencies, projectDir)
```
pepicrft, line 73: If we pass `packageJsonDependencies` aren't we updating all the dependencies existing in the project?

packages/cli-main/src/cli/services/upgrade.ts
```diff
@@ -0,0 +1,174 @@
...
+      // eslint-disable-next-line no-catch-all/no-catch-all
+    } catch (err) {
+      output.warn('Homebrew upgrade failed. Falling back to standard npm install.')
+    }
+  }
+  const command = 'npm'
+  const args = [
+    'install',
+    '-g',
+    `${await cliDependency()}@latest`,
```
pepicrft, line 107: I see more than one `await cliDependency()` scattered in this function. What about calling it once and passing it down?

packages/cli-main/src/cli/services/upgrade.ts
amcaplan: It's a [standard variable](https://docs.npmjs.com/cli/v8/using-npm/config#user-agent) set by the package managers. If it's there, we assume this is in a project and therefore the upgrade is local. Otherwise we're assuming a global upgrade.

packages/cli-main/src/cli/services/upgrade.ts
amcaplan: No, since `installJsonDependencies filters to just `@shopify/cli` and canonical oclif plugins.

packages/cli-main/src/cli/services/upgrade.ts
```diff
@@ -0,0 +1,174 @@
...
+        )
+        await system.exec('brew', ['upgrade', 'shopify-cli@3'], {stdio: 'inherit'})
+        return newestVersion
+      }
+      // eslint-disable-next-line no-catch-all/no-catch-all
+    } catch (err) {
+      output.warn('Homebrew upgrade failed. Falling back to standard npm install.')
+    }
+  }
+  const command = 'npm'
```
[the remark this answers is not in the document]
amcaplan: Good point. Also if it's MacOS but it was installed by `npm`, this will have weird output.

packages/cli-main/src/cli/services/upgrade.ts
```diff
@@ -0,0 +1,174 @@
...
+          return file.exists(path.join(dir, configFile))
+        }),
+      )
+      if (configFilesExist.some((bool) => bool)) return dir
+    },
+    {
+      cwd: directory,
+      type: 'directory',
+    },
+  )
```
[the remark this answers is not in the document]
amcaplan: This was a hack because findUp doesn't document supporting an array of paths (as far as I could tell). However, you're right that [it does](https://github.com/sindresorhus/find-up/blob/a3b2f1e788e3c4023eab8f6282b085b57131ae15/index.js#L14). Way better!

pepicrft (approved):

PR #497, opened 2022-09-19
Sort apps by name when presented with prompt

stevemar:
This PR proposes to sort the list of apps when presented with the app select prompt. This is useful when you have a lot of apps and you want to find a specific one quickly. It was one of the pieces of feedback that we heard from folks at Unite.

Co-Authored-By: David Witt <david.witt@shopify.com>
Signed-off-by: Steve Martinelli <steve.martinelli@shopify.com>

Fixes https://github.com/Shopify/shopify-cli-planning/issues/380

This PR adds a sort call to the list of apps when presented with the app select prompt. It also updates the wording a bit to make it more obvious that the user can start typing when prompted to filter the list of apps.

#### Before

Screen Shot 2022-09-19 at 2 30 53 PM

#### After

Screen Shot 2022-09-19 at 2 32 08 PM

### How to test your changes?

Locally, see screenshots, and updated existing test + added a new one.

### Measuring impact

### Checklist


gonzaloriestra (commented):
Thanks for the PR! I have 2 concerns:
- I think ordering by most recent would be more useful. Stephanie agrees [here](https://github.com/Shopify/shopify-cli-planning/issues/380#issuecomment-1251705006). If the user has many apps and they just want to use the latest one, this doesn't help. And if they want to use one with a specific name, they can use the filter to avoid having to scroll too much.
- Instead of ordering locally, I'd do it in the graphql query to avoid extra work in the CLI.

isaacroldan:
Agree with @gonzaloriestra 👌

pepicrft:
> @gonzaloriestra that seems fair enough. Before I close out this PR would it be worth stripping out the part that includes to message to start typing at select app prompt?

That's a question for @MeredithCastile & @grahamscott

PR #503, opened 2022-09-20
React dependency only added for react flavour extensions

alvaro-shopify:

Fixes https://github.com/Shopify/shopify-cli-planning/issues/381

- When adding the additional dependencies needed for an UI extension the react one is only react flavours select by the user should add react dependency

### How to test your changes?
- Run `yarn shopify app generate extension` from a `php` or `ruby` app
- Select `checkout UI` extension
- Select either `typescript-react` or `react` flavour
- Open the `package.json` at the root path of the app and there should be no `react` dependency

Retry the same steps again but selecting either `vanilla-js` or `typescript` and the `react` dependency should appears in the `package.json` 

### Measuring impact

### Checklist

### DIFF
--- a/packages/app/src/cli/services/generate/extension.ts
+++ b/packages/app/src/cli/services/generate/extension.ts
@@ -149,7 +149,7 @@ export function getRuntimeDependencies({
     case 'customer_accounts_ui_extension':
     case 'checkout_post_purchase': {
       const dependencies: DependencyVersion[] = []
-      if (extensionFlavor === 'react' || extensionFlavor === 'typescript-react') {
+      if (extensionFlavor?.includes('react')) {
         dependencies.push({name: 'react', version: versions.react})
       }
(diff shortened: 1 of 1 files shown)


packages/app/src/cli/services/generate/extension.ts
```diff
@@ -139,15 +139,19 @@ async function uiExtensionInit({
...
   switch (extensionType) {
     case 'product_subscription':
     case 'checkout_ui_extension':
     case 'pos_ui_extension':
     case 'web_pixel_extension':
     case 'customer_accounts_ui_extension':
     case 'checkout_post_purchase': {
-      const dependencies: DependencyVersion[] = [{name: 'react', version: versions.react}]
+      const dependencies: DependencyVersion[] = []
+      if (extensionFlavor === 'react' || extensionFlavor === 'typescript-react') {
```
isaacroldan, line 152: maybe... `extensionFlavor.includes('react')`?

PR #505, opened 2022-09-20
Fix extension react templates

isaacroldan:
Extensions are being generated always with the non-react flavor, even when selecting the react version.

- If the flavor is `react` or `typescript-react`, convert it to `react` as this is the only flavor accepted by the liquid templates.
- Add tests to be sure to always call the template function with the right flavor.

### How to test your changes?
generate different extensions with different flavors and validate that all of them are correct

### Post-release steps

- [ ] I've made sure that any changes to `dev` or `deploy` have been reflected in the [the documentation](https://www.figma.com/file/7vqUp50u6dm48Zfb4JRRn8/CLI3-Internals).

### Measuring impact

### Checklist

### DIFF
--- a/packages/app/src/cli/services/generate/extension.test.ts
+++ b/packages/app/src/cli/services/generate/extension.test.ts
@@ -9,7 +9,7 @@ import {
 } from '../../constants.js'
 import {load as loadApp} from '../../models/app/loader.js'
 import {describe, it, expect, vi, test, beforeEach} from 'vitest'
-import {file, output, path} from '@shopify/cli-kit'
+import {file, output, path, template} from '@shopify/cli-kit'
 import {addNPMDependenciesIfNeeded} from '@shopify/cli-kit/node/node-package-manager'
 import type {ExtensionFlavor} from './extension.js'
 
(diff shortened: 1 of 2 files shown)


packages/app/src/cli/services/generate/extension.test.ts
```diff
@@ -99,34 +99,63 @@ describe('initialize a extension', () => {
...
+
+        expect(recursiveDirectoryCopySpy).toHaveBeenCalledWith(expect.any(String), expect.any(String), {
           flavor: liquidFlavor,
           type: extensionType,
           name,
         })
-        expect(srcIndexFile.trim()).not.toBe('')
+
+        recursiveDirectoryCopySpy.mockRestore()
+        fileMoveSpy.mockRestore()
```
pepicrft, line 158: What happens if any of the above methods throws? Isn't the the original function restored automatically when the test completes?

packages/app/src/cli/services/generate/extension.test.ts
```diff
@@ -99,34 +99,63 @@ describe('initialize a extension', () => {
...
         const name = 'extension-name'
-        const spy = vi.spyOn(template, 'recursiveDirectoryCopy')
 
         await createFromTemplate({name, extensionType, extensionFlavor, appDirectory: tmpDir})
 
         const srcIndexFile = await file.read(path.join(tmpDir, 'extensions', name, 'src', `index.${fileExtension}`))
+        expect(srcIndexFile.trim()).not.toBe('')
+      })
+    },
+    30 * 1000,
```
pepicrft, line 127: I'd extract this one into `extension.integration.test.ts` to make it clear that the test is integration and therefore future failures might be caused by some flakiness.

PR #514, opened 2022-09-21
Add a lint rule to discourage the the usage of factory functions

pepicrft:
This PR is continuation of [this work](https://github.com/Shopify/cli/pull/504) to introduce some conventions into our codebase to ensure the codebase evolves healthily and that developers have a great experience contributing to the repository. In particular, I'm adding an ESLint rule that makes `lint` fail if new modules use the following pattern:

```ts
import { error } from "@shopify/cli-kit"

const SomeError = () => {
  return new error.Abort("some custom message")
}

function someService() {
   // Before
  throw SomeError()

  // Now
  throw new error.Abort("some custom message")
}
```

I'm adding the ESLint rule with a [shitlist](https://sirupsen.com/shitlists) of modules that haven't yet adopted the new convention. This ensures that we adopt the new convention in new modules and that there's some accountability to motivate the migration of the modules in the list.

### How to test your changes?

`yarn lint` must succeed.

### Measuring impact

### Checklist

### DIFF
--- a/packages/app/src/cli/commands/app/generate/extension.ts
+++ b/packages/app/src/cli/commands/app/generate/extension.ts
@@ -161,6 +161,7 @@ export default class AppScaffoldExtension extends Command {
     const functionExtensionTemplateNames = functionExtensionTemplates.map((template) => template.value)
 
     const invalidTemplateError = (templates: string[]) => {
+      // eslint-disable-next-line rulesdir/no-error-factory-functions
       return new error.Abort(
         'Specified extension template on invalid extension type',
         `You can only specify a template for these extension types: ${templates.join(', ')}.`,
--- a/packages/cli-kit/src/session.ts
+++ b/packages/cli-kit/src/session.ts
@@ -87,12 +87,6 @@ export interface OAuthSession {
   partners?: string
   storefront?: string
 }
-export const PartnerOrganizationNotFoundError = () => {
-  return new Abort(
-    `Couldn't find your Shopify Partners organization`,
-    `Have you confirmed your accounts from the emails you received?`,
-  )
-}
 
 /**
  * Ensure that we have a valid session to access the Partners API.
@@ -280,7 +274,10 @@ export async function ensureUserHasPartnerAccount(partnersToken: string) {
     output.warn(output.content`Make sure you've confirmed your Shopify and the Partner organization from the email`)
     await keypress()
     if (!(await hasPartnerAccount(partnersToken))) {
-      throw PartnerOrganizationNotFoundError()
+      throw new Abort(
+        `Couldn't find your Shopify Partners organization`,
+        `Have you confirmed your accounts from the emails you received?`,
+      )
     }
   }
 }


pepicrft:
Next here is adding some documentation around:
- Error and result types.
- When and how to throw errors
- How to test logic that throws errors.

bin/git-files.sh
```diff
@@ -0,0 +1,3 @@
+#!/usr/bin/env bash
+
+git ls-tree -r --name-only HEAD -z | TZ=UTC xargs -0n1 -I_ git --no-pager log -1 --date=iso-local --format="%ad _" -- _
```
pepicrft, line 3: @Shopify/cli-foundations this one is a bit slow and blocks the execution of `lint` for seconds. Any idea of how to make this one more performant?

eslint-rules/no-error-factory-functions.cjs
```diff
@@ -0,0 +1,69 @@
...
+
+        if (isCLIKitError) {
+          const filePath = context.getFilename()
+          const relativePath = path.relative(path.resolve(__dirname, '..'), filePath)
+          const fileUpdatedAt = context.settings.gitFilesLastModified[relativePath]
+          const shouldFail = !shitlist[relativePath] || shitlist[relativePath] < fileUpdatedAt
+          if (shouldFail) {
+            context.report(
+              node,
+              `Factory functions returning instances of abort and bug errors are discouraged. This error arouse because either you are not following the pattern in a new module or you touched an existing one that needs migration.`,
```
shauns, line 34:
```suggestion
              `Factory functions returning instances of abort and bug errors are discouraged. This error arouse because either you are not following the pattern in a new module or you touched an existing one that needs migration.`,
              `Factory functions returning instances of abort and bug errors are discouraged. This error arose because either you are not following the pattern in a new module or you touched an existing one that needs migration.`,,
```

bin/git-files.sh
Arkham: what do you think if we instead checked the last modified time of the offending file when the error arises? right now we are paying the price of reading the last modified times of EVERY file while we only care about a small subset of them

docs/README.md
```diff
@@ -2,24 +2,43 @@
...
+* [ESLint rules](./cli/eslint-rules.md)
+* [Release process](./cli/release.md)
+* [Testing strategy](./cli/testing-strategy.md)
+* [Cross-OS compatibility](./cli/cross-os-compatibility.md)
+* [Troubleshooting](./cli/troubleshooting.md)
+* [FAQ](./cli/faq.md)
+
+## CLI Kit
+
+The [`@shopify/cli-kit`](https://www.npmjs.com/package/@shopify/cli-kit) NPM package provides utilities to abstract away interactions with the Shopify platform (e.g. authentication, API requests) and ensure experiences are consistent across the board. If you are creating a new plugin or contributing to an existing one, we recommend checking out the following resources:
```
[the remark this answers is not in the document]
pepicrft: I agree. We need to tell contributors about the existence of the components, the role they play in the CLI, and guidelines around how to use them. Feel free to go ahead with a separate PR for it.

bin/git-files.sh
pepicrft: Let me do some benchmarking. Getting the last modified date for each file will require a spawning a process which is also expensive to do for every file. Perhaps less than doing all the files together.

pepicrft: It's orders of magnitude faster. Let me know what you think @Arkham

PR #515, opened 2022-09-21
Check response of wasm blob upload request to inform users when upload fails

jacobsteves:

Closes https://github.com/Shopify/script-service/issues/5604

This change will display a relevant error message when the wasm upload fails. Without this step, the user gets an unhelpful server error and does not know why they cannot deploy successfully.

When we make the request to upload the Wasm module, check the response. CLI 2.0 had [logic](https://github.com/Shopify/shopify-cli/blob/0e5eccffe34c245d0e27420bc82c1fff9708b70b/lib/project_types/script/layers/infrastructure/script_uploader.rb) in place to do this, so the code I added just migrates those checks over.

### How to test your changes?

- Try to push a module that is within limits. Everything should succeed.
- Try to push a module that exceeds the 256kb limit. You should get an error:
image

### Checklist

- [X] ~~I've considered possible [documentation](https://shopify.dev) changes~~
- [X] ~~I've made sure that any changes to `dev` or `deploy` have been reflected in the [internal flowchart](https://www.figma.com/file/7vqUp50u6dm48Zfb4JRRn8/CLI3-Internals)~~.

### DIFF
--- a/packages/app/src/cli/services/deploy/upload.test.ts
+++ b/packages/app/src/cli/services/deploy/upload.test.ts
@@ -3,7 +3,7 @@ import {blocks} from '../../constants.js'
 import {Identifiers} from '../../models/app/identifiers.js'
 import {FunctionExtension} from '../../models/app/extensions.js'
 import {afterEach, beforeEach, describe, expect, test, vi} from 'vitest'
-import {path, file, api, http} from '@shopify/cli-kit'
+import {error, path, file, api, http} from '@shopify/cli-kit'
 
 function mockFunctionCompilation(jobId = 'job-id'): void {
   const status = 'completed'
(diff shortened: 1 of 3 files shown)


isaacroldan:
Is this limit likely to change? should we validate the size before trying to deploy the function?

jacobsteves:
Yeah I see this limit changing at some point. It is configurable on the server-side so I think it makes sense to keep the CLI generic and simply check the response. That way, we can roll out limit changes without depending on CLI versions.

packages/app/src/cli/services/deploy/upload.ts
```diff
@@ -240,6 +238,24 @@ ${output.token.json(userErrors)}
...
+  const res = await http.fetch(url, {body: functionContent, headers, method: 'PUT'})
+
+  if (res.status === 200) {
+    return url
+  } else if (res.status === 400 && res.body?.read().includes('EntityTooLarge')) {
+    const errorMessage = output.content`The size of the Wasm binary file for Function ${extension.localIdentifier} is too large. It must be less than ${maxSize}.`
+    throw new error.Abort(errorMessage)
+  } else {
+    const errorMessage = output.content`Something went wrong and the Function ${extension.localIdentifier} couldn't be deployed. Try again.`
+    throw new error.Abort(errorMessage)
```
andrewhassan, line 255: Do you think we should actually report this case as a bug so we can add better error messages? IIUC, we can throw `error.Bug` and it'll get Bugsnagged.

andrewhassan (approved):

packages/app/src/cli/services/deploy/upload.ts
isaacroldan: I don't think so, we use `error.Bug` for CLI bugs that are actionable within the CLI, this wouldn't be a CLI bug but an API one. I guess the API already reports its own errors somewhere?

andrewhassan: I think it would be a CLI bug, since it's a user error that the backend is returning (e.g. your function description is too long). Based on the user error, the CLI should show a helpful error message. Ideally, we'd handle all user error cases so that there are properly formatted messages for all of them.

packages/app/src/cli/services/deploy/upload.ts
```diff
@@ -240,6 +237,27 @@ ${output.token.json(userErrors)}
...
+  const {url, headers, maxSize} = await getFunctionExtensionUploadURL({apiKey, token})
+  headers['Content-Type'] = 'application/wasm'
+
+  const functionContent = fs.readFileSync(extension.buildWasmPath())
+  const res = await http.fetch(url, {body: functionContent, headers, method: 'PUT'})
+  const resBody = res.body?.read()?.toString() || ''
+
+  if (res.status === 200) {
+    return url
+  } else if (res.status === 400 && resBody.includes('EntityTooLarge')) {
```
pepicrft, line 250: `resBody.includes('EntityTooLarge')` seems very brittle. Should we define a more structured contract?

packages/app/src/cli/services/deploy/upload.ts
```diff
@@ -240,6 +237,27 @@ ${output.token.json(userErrors)}
...
+    const errorMessage = output.content`The size of the Wasm binary file for Function ${extension.localIdentifier} is too large. It must be less than ${maxSize}.`
+    throw new error.Abort(errorMessage)
+  } else {
+    const errorMessage = output.content`Something went wrong uploading the Function ${
+      extension.localIdentifier
+    }. The server responded with status ${res.status.toString()} and body: ${resBody}`
+    throw new error.Bug(errorMessage)
+  }
```
pepicrft, line 258: I'd be more granular here. Considering 5xx errors are already tracked server-side, we don't want them to show up in the CLI Bugsnag project. I'd only report as bugs those that are 4xx and that are not caused by the binary size exceeded.

packages/app/src/cli/services/deploy/upload.ts
```diff
@@ -240,6 +237,27 @@ ${output.token.json(userErrors)}
   return res.data.functionSet.function?.id as string
 }
 
+async function uploadWasmBlob(extension: FunctionExtension, apiKey: string, token: string): Promise<string> {
+  const {url, headers, maxSize} = await getFunctionExtensionUploadURL({apiKey, token})
+  headers['Content-Type'] = 'application/wasm'
+
+  const functionContent = fs.readFileSync(extension.buildWasmPath())
```
pepicrft, line 244: I'd import `file` from `@shopify/cli-kit` and use the promise version of this API: `file.readFile`

packages/app/src/cli/services/deploy/upload.test.ts
```diff
@@ -148,6 +148,95 @@ describe('uploadFunctionExtensions', () => {
...
+            return 'EntityTooLarge'
+          },
+        },
+      })
+      vi.mocked(http.fetch).mockImplementation(fetch)
+
+      // When
+      await expect(() => uploadFunctionExtensions([extension], {token, identifiers})).rejects.toThrowError(
+        /The size of the Wasm binary file for Function my-function is too large. It must be less than 200 kb./,
+      )
```
pepicrft, line 184: I'd expect that we are throwing the right type of error.

packages/app/src/cli/services/deploy/upload.test.ts
```diff
@@ -148,6 +148,95 @@ describe('uploadFunctionExtensions', () => {
...
+            return 'error body'
+          },
+        },
+      })
+      vi.mocked(http.fetch).mockImplementation(fetch)
+
+      // When
+      await expect(() => uploadFunctionExtensions([extension], {token, identifiers})).rejects.toThrowError(
+        /Something went wrong uploading the Function my-function. The server responded with status 401 and body: error body/,
+      )
```
pepicrft, line 228: Similar comment here regarding the error type.

PR #521, opened 2022-09-22
Added app id to retail mobile qr code

js-goupil:

This PR adds the appId to the mobile QR code generated for POS. The appId is required for certain APIs that POS makes available to UI extensions. In a dev environment, this is the only way for POS to access the appId for a dev app.

It checks to make sure the app exists, and if so adds the appId in the URL QR code.

### Checklist

### DIFF
--- a/packages/ui-extensions-dev-console/src/QRCodeModal/QRCodeModal.test.tsx
+++ b/packages/ui-extensions-dev-console/src/QRCodeModal/QRCodeModal.test.tsx
@@ -72,7 +72,7 @@ describe('QRCodeModal', () => {
     );
 
     expect(container?.find(QRCode)?.prop('value')).toStrictEqual(
-      `com.shopify.pos://pos-ui-extensions?url=${extension.development.root.url}`,
+      `com.shopify.pos://pos-ui-extensions?url=${extension.development.root.url}&appId=${app.id}`,
     );
   });
 });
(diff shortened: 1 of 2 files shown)


heltisace (approved):
LGTM

packages/ui-extensions-dev-console/src/QRCodeModal/QRCodeModal.test.tsx
```diff
@@ -72,7 +72,7 @@ describe('QRCodeModal', () => {
     );
 
     expect(container?.find(QRCode)?.prop('value')).toStrictEqual(
-      `com.shopify.pos://pos-ui-extensions?url=${extension.development.root.url}`,
+      `com.shopify.pos://pos-ui-extensions?url=${extension.development.root.url}&appId=${app.id}`,
```
vividviolet, line 75: I don't think this is safe to use - the only field that's guaranteed to be populated currently is the `app.apiKey`

vividviolet (changes requested):

js-goupil:
@vividviolet interesting. @henrytao-me do you know how Shopify mobile gets the appId? The reason why we need this is because the adminGenerateSession endpoint requires the app ID as a parameter. So as we're writing out the API for the app extension to be able to call that, we're trying to figure out a way to get that app ID in the dev environment.

PR #522, opened 2022-09-22
Add generate schema command

andrewhassan:

To make working with function input queries easier, we want to introduce a command that will fetch the latest GraphQL schema for a function's API and version.

Related issue: https://github.com/Shopify/script-service/issues/5190

This PR adds a new command to fetch the GraphQL schema for a function. The schema is outputted to STDOUT so that it can be written to a file or consumed by a code generation tool.

Since fetching a schema requires a valid API key to work, this command resolves the app in the following order:
1. If the `api-key` flag is passed in, that key is used
2. If there is an API key stored in the `.env`, that is used
3. Otherwise, the user is prompted to choose an app. In this case, the prompt is displayed using STDOUT, so it may conflict with any tools consuming STDOUT.

### How to test your changes?

You will need to use the branches in the following PRs:

1. https://github.com/Shopify/shopify/pull/374232
2. https://github.com/Shopify/partners/pull/42045

Then in this repo, run `yarn shopify app generate schema product-discount --path ./fixtures/app` (preface this with `bin/spin` if running in Spin).

### Measuring impact

The existing command usage analytics should be suitable for now.

### Checklist

- [ ] ~I've made sure that any changes to `dev` or `deploy` have been reflected in the [internal flowchart](https://www.figma.com/file/7vqUp50u6dm48Zfb4JRRn8/CLI3-Internals).~


packages/app/src/cli/services/generate-schema.ts
```diff
@@ -0,0 +1,30 @@
...
+}
+
+export async function generateSchemaService(options: GenerateSchemaOptions) {
+  const {extension, app} = options
+  const token = await session.ensureAuthenticatedPartners()
+  const {apiVersion: version, type} = extension.configuration
+  const apiKey =
+    options.apiKey ||
+    getAppIdentifiers({app}).app ||
+    (await fetchOrganizationAndFetchOrCreateApp(app, token)).partnersApp.apiKey
```
andrewhassan, line 20: @pepicrft I'm not sure if having this fallback to prompt makes sense if the user is going to consume the output (e.g. pipe to code generation tool). What do you think?

andrewhassan (commented):
@pepicrft I'd appreciate your 👀 on my progress so far for the `generate schema` command. I'm unsure about how I'm retrieving the API key. I tried to make a best guess on doing this based on the existing code. Please let me know what's out of place 😅

I'll add tests once the code looks good.

packages/app/src/cli/commands/app/generate/schema.ts
```diff
@@ -0,0 +1,37 @@
...
+
+  public async run(): Promise<void> {
+    const {flags, args} = await this.parse(GenerateSchema)
+    const apiKey = flags['api-key']
+    const directory = flags.path ? path.resolve(flags.path) : process.cwd()
+    const app: AppInterface = await loadApp(directory)
+    const extension = app.extensions.function.find((f) => f.localIdentifier === args.function)
+
+    if (!extension) {
+      throw new error.Abort(`Function ${args.function} not found`)
```
pepicrft, line 32: What about making this error more actionable and including the list of function IDs that are available? cc @MeredithCastile

packages/app/src/cli/services/generate-schema.ts
```diff
@@ -0,0 +1,30 @@
...
+
+  const query = api.graphql.ApiSchemaDefinitionQuery
+  const variables: api.graphql.ApiSchemaDefinitionQueryVariables = {
+    apiKey,
+    version,
+    type,
+  }
+  const response: api.graphql.ApiSchemaDefinitionQuerySchema = await api.partners.request(query, token, variables)
+  output.info(response.definition ?? '')
```
pepicrft, line 29: What does `""` mean? When can that happen? I'd recommend to output something more meaningful to the user. Otherwise they'll end up opening an issue reporting that they get no output.

pepicrft (approved):

packages/app/src/cli/commands/app/generate/schema.ts
```diff
@@ -0,0 +1,42 @@
...
+    const app: AppInterface = await loadApp(directory)
+    const extension = app.extensions.function.find((extension) => extension.localIdentifier === args.function)
+
+    if (!extension) {
+      const functions = app.extensions.function.map((extension) => extension.localIdentifier).join(', ')
+
+      throw new error.Abort(
+        output.content`No function named ${args.function} found in this app.`,
+        output.content`Use one of the available functions: ${functions}`,
+      )
```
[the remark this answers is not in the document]
MeredithCastile: @andrewhassan Looks fine.

Question, though — is there a way that we can prevent this error in the first place?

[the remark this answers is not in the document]
andrewhassan: Thanks!

>Question, though — is there a way that we can prevent this error in the first place?

I think the only way we can guarantee this value is valid is by having the user select the function from an interactive list. However, this makes it impossible for the command to be used in a non-interactive environment (e.g. CI), which is something we want to support. Since this means the value can be provided by the user, there's no way to validate it beforehand.

shauns (approved):

PR #530, opened 2022-09-27
A few hydrogen deploy enhancements

lynchv:

Update initialize repository with initial branch
Fix https://github.com/Shopify/oxygen-platform/issues/599

Allow deployment of anything
fix https://github.com/Shopify/oxygen-platform/issues/571

Unless specified otherwise in the code any git repo initialization will create the default branch as "main" regardless of ones local git config. This matches our Shopify default including the ones in the Hydrogen/Oxygen Admin.

Until we can think of a more user friendly method of deploying any folder to Oxygen without the limitation of the asset URL, I think we should keep the new option `pathToBuild` hidden. If this option is used, build step is skipped and will simply zip and upload the path provided.

### Checklist


packages/cli-hydrogen/src/cli/commands/hydrogen/deploy.ts
```diff
@@ -24,6 +24,11 @@ export default class Deploy extends Command {
       env: 'SHOPIFY_HYDROGEN_FLAG_COMMIT_AUTHOR',
       description: 'Override the default Git commit author.',
     }),
+    pathToBuild: Flags.string({
+      hidden: true,
+      env: 'SHOPIFY_HYDROGEN_FLAG_PATH_TO_BUILD',
+      description: 'Skip build process and use provided value as build',
```
lynchv, line 30: @gfscott Could use some 👀 on wording here.

gfscott: Cool! Questions:
1. What's the default behavior if they _don't_ pass this flag?
2. The path they specify: is it intended to point to a fully built project that was created by some other build process? Or is it a different build command? Not clear to me from the current wording!

lynchv: > What's the default behavior if they don't pass this flag?

We build the project ourself either in the current working directory or in the one provided by the `path` argument.

> is it intended to point to a fully built project that was created by some other build process?

Yes

packages/cli-hydrogen/src/cli/services/deploy/config.ts
```diff
@@ -42,6 +42,7 @@ export const fillDeployConfig = async (config: DeployConfig): Promise<ReqDeployC
 
   return {
     ...config,
+    pathToBuild: config.pathToBuild ? path.resolve(config.pathToBuild) : '',
```
graygilmore, line 45: Is `''` the actual value that we want or should it be `undefined`?

graygilmore (approved):
One small question but otherwise changes look good to me

packages/cli-hydrogen/src/cli/services/deploy/config.ts
lynchv: It seems like every time I have an internal debate as to wether or not I should do something you have a question on it 🤣 , haha!

So we have two config types, `DeployConfig` & `ReqDeployConfig`. Both have the same properties but the second requiring all properties. First type is used for the CLI argument parsing and second type is used for `fillDeployConfig`. I agree that `undefined` would be the best solution but it just doesn't fit in our pattern/structure. I thought of using `string | null` instead of only `string` but both `''` & `null` evaluate to false so I chose to use empty string as the "no value" value.

packages/cli-hydrogen/src/cli/commands/hydrogen/deploy.ts
gfscott: Does this work?

```suggestion
      description: 'Skip the build process and deploy the built project at the specified path. You’ll need to manage your own project build step.',
```

tempor1s (approved):
Looks good to me!

lynchv:
### Force Push Notes:

- Adding changeset to PR

PR #537, opened 2022-09-28
Add a schema to the capabilities attribute of extensions

pepicrft:
Related: https://github.com/Shopify/cli/pull/455#discussion_r980469922

The `capabilities` section of the UI extensions' configuration file has no schema in the CLI. Because of it, the serving logic ended up having [some logic](https://github.com/Shopify/cli/pull/455/files/cacc3f232d5d88608d31a64c65c418fa25a43ec4#diff-e5f8a674cef6431ad1097c19b9112388b3267bcf2f8db6939b38e1d01effeb10R25) that assumes the schema and its defaults. This is information that shouldn't be part of the serving business logic so I'm codifying it in the schema.

I'm adding a schema and defaults to the `capabilities` attribute of the UI extension configuration file.

### How to test your changes?
The `product-subscription` UI extension of the `fixtures/app` directory has the `network_access` capability set to true. Run `dev` with that project passing `--verbose` and confirm that `networkAccess` is returned as `true` from the extensions endpoint.

### Measuring impact

### Checklist


packages/app/src/cli/models/app/extensions.ts
```diff
@@ -28,8 +28,19 @@ export const UIExtensionConfigurationSchema = schema.define.object({
...
+      network_access: schema.define.boolean().default(false),
+    })
+    .default({
+      block_progress: false,
+      network_access: false,
+    })
+    .transform((capabilities) => ({
+      blockProgress: capabilities.block_progress,
+      networkAccess: capabilities.network_access,
+    })),
```
byrichardpowell, line 43: With these changes we get different JSON between go and Node for the product subscriptions JSON.

Node on the left, go on the right (I think Node is correct):

Screen Shot 2022-09-28 at 11 12 20 AM

This is when going to `/extensions`.

byrichardpowell (changes requested):
Really supportive of these changes.  The one question about Go vs Node is why I'm requesting changes.

packages/app/src/cli/models/app/extensions.ts
pepicrft: That makes me think the attribute was not being passed correctly before. @vividviolet is the Node one accurate?

packages/app/src/cli/models/app/extensions.ts
```diff
@@ -28,8 +28,19 @@ export const UIExtensionConfigurationSchema = schema.define.object({
     )
     .default([]),
   extensionPoints: schema.define.array(schema.define.string()).optional(),
-  capabilities: schema.define.any().optional(),
-
+  capabilities: schema.define
```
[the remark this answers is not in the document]
pepicrft: That makes sense. We can iterate as needed. Thanks for the context @vividviolet

packages/app/src/cli/models/app/extensions.ts
pepicrft: #### `localizationStatus`
@vividviolet I aligned the behaviour with Go's

#### Empty `resource`
I aligned this one with Go too.

#### networkAccess
The changes in the PR introduced a regression in the Go flow. I fixed it too.

packages/app/src/cli/models/app/extensions.ts
```diff
@@ -28,8 +28,19 @@ export const UIExtensionConfigurationSchema = schema.define.object({
...
+    .object({
+      block_progress: schema.define.boolean().default(false),
+      network_access: schema.define.boolean().default(false),
+    })
+    .default({
+      block_progress: false,
+      network_access: false,
```
vividviolet, line 38: Can we actually remove these defaults? I think having these values in the config will actually break extensions validation during deploy for all except Checkout UI extensions. We haven't standardized support for these yet

PR #541, opened 2022-09-29
Fix wasm file binary read failure

jacobsteves:
Released in https://github.com/Shopify/cli/pull/515, we changed the way we read the wasm file. This method seems to be unable to read binary files correctly so any new deploy will fail due to a compilation failure.

That PR has not been released yet, so no need for a new changelog. 

### How to test your changes?

Try to deploy a function.

### Checklist

### DIFF
--- a/packages/app/src/cli/services/deploy/upload.ts
+++ b/packages/app/src/cli/services/deploy/upload.ts
@@ -241,7 +241,7 @@ async function uploadWasmBlob(extension: FunctionExtension, apiKey: string, toke
   const {url, headers, maxSize} = await getFunctionExtensionUploadURL({apiKey, token})
   headers['Content-Type'] = 'application/wasm'
 
-  const functionContent = await file.read(extension.buildWasmPath())
+  const functionContent = await file.read(extension.buildWasmPath(), {})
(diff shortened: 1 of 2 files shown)


mkcny (approved):
🎩ed locally and confirmed this fixed deploys for me

packages/app/src/cli/services/deploy/upload.ts
```diff
@@ -241,7 +241,7 @@ async function uploadWasmBlob(extension: FunctionExtension, apiKey: string, toke
   const {url, headers, maxSize} = await getFunctionExtensionUploadURL({apiKey, token})
   headers['Content-Type'] = 'application/wasm'
 
-  const functionContent = await file.read(extension.buildWasmPath())
+  const functionContent = fs.readFileSync(extension.buildWasmPath())
```
KevinRizzoTO, line 244: It looks like you can pass an encoding object here https://github.com/Shopify/cli/blob/main/packages/cli-kit/src/file.ts#L35 which has a default that might be creating the bug. If you pass an empty object here that should take away the encoding passed to `fs.readFile`, which is the same fix I had to do [here](https://github.com/Shopify/cli/commit/f79392179e615a92b8da3f2183b3714575a960bb)

PR #553, opened 2022-10-03
Enable spin store domains

alvaro-shopify:
As part of the effort in [streamlining the first-party experience when creating apps](https://vault.shopify.io/gsd/projects/25765), we would like to make it as easy as possible for 1p developers to create apps on Spin.

Right now, all templates are only allowing using either `myshopify.com` or `myshopify.io` as store domains. When running the cli inside a spin instance, the stores created have a randomly generated `spin domain`, so a way to send that domain to the apps and to validate it from them must be added. 

- Add a new environment variable `SHOPIFY_CUSTOM_DOMAIN` with the specific `spin domain` where the `dev` command is run

### How to test your changes?
- The [proposal](https://docs.google.com/document/d/1V1lMGR5uqi2Lsy8Y1e2zj_-8cGUFjUbVOZ-2z4CuQA4/edit) should be approved first thus the PRs for the templates are merged.
- Create a new app. It is also possible to modify existing apps by following the steps to support the new environment variable added in their README
- Run `yarn dev`

### Measuring impact

### Checklist


packages/app/src/cli/services/dev.ts
```diff
@@ -238,6 +238,9 @@ function devBackendTarget(web: Web, options: DevWebOptions): output.OutputProces
     BACKEND_PORT: `${options.backendPort}`,
     SCOPES: options.scopes,
     NODE_ENV: `development`,
+    ...(environment.service.serviceEnvironment() === environment.network.Environment.Spin && {
```
isaacroldan, line 241: If you extract this from the other PR, use it here too :)

PR #567, opened 2022-10-04
Render fatal errors with `Banner`

matteodepalo:

We want to render all our errors consistently with the new `Banner` component.
This PR is stacked on top of https://github.com/Shopify/cli/pull/512 so we need to wait for that to be merged first.
You can look at the last few commits related to errors to see changes specific to this PR. 

Render all `Fatal` errors by using the new `Banner` component. For example:

Before:
Screenshot 2022-10-04 at 16 50 29

After:
Screenshot 2022-10-04 at 16 47 55

### How to test your changes?

- Run `yarn shopify kitchen-sink`

### Measuring impact

### Checklist

### DIFF
--- a/packages/cli-kit/src/error.ts
+++ b/packages/cli-kit/src/error.ts
@@ -1,5 +1,6 @@
-import {Message, stringifyMessage, error as outputError} from './output.js'
+import {Message, stringifyMessage} from './output.js'
 import {normalize} from './path.js'
+import {renderFatalError} from './public/node/ui.js'
 import {Errors} from '@oclif/core'
 
 export {ExtendableError} from 'ts-error'
@@ -75,7 +76,7 @@ export async function handler(error: unknown): Promise<unknown> {
     }
   }
 
-  await outputError(fatal)
+  renderFatalError(fatal)
   return Promise.resolve(error)
 }
 
--- a/.changeset/angry-socks-develop.md
+++ b/.changeset/angry-socks-develop.md
@@ -1,0 +1,7 @@
+---
+'@shopify/cli-kit': patch
+'@shopify/cli': patch
+'@shopify/plugin-ngrok': patch
+---
+
+Improve error outputs by using the new Banner component
--- a/packages/cli-kit/src/error.test.ts
+++ b/packages/cli-kit/src/error.test.ts
@@ -1,46 +1,46 @@
 import {Abort, Bug, handler, cleanSingleStackTracePath} from './error.js'
-import {error} from './output.js'
 import {describe, expect, test, vi, beforeEach, it} from 'vitest'
+import {renderFatalError} from '@shopify/cli-kit/node/ui'
 
 beforeEach(() => {
-  vi.mock('./output')
+  vi.mock('@shopify/cli-kit/node/ui')
 })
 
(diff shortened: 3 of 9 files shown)


MeredithCastile:
I'm experiencing visceral relief at the sudden up-leveling of design quality this change brings. Bravo!

packages/cli-kit/src/error.ts
```diff
@@ -75,7 +76,7 @@ export async function handler(error: unknown): Promise<unknown> {
     }
   }
 
-  await outputError(fatal)
+  renderFatalError(fatal)
```
gonzaloriestra, line 79: Why having both renderError and renderFatalError? Couldn't we have just one and show the stacktrace only when required?

packages/cli-kit/src/private/node/components/Banner.tsx
```diff
@@ -0,0 +1,34 @@
...
+    success: 'green',
+    error: 'red',
+    warning: 'yellow',
+    info: 'white',
+  }[type]
+}
+
+const Banner: React.FC<Props> = ({type, fullWidth = false, children}) => {
+  const {stdout} = useStdout()
+  const width = fullWidth ? stdout!.columns : Math.min(60, stdout!.columns)
```
gonzaloriestra, line 22: Instead of using the full width for fatal errors, what about adjusting it to the longest line from the content? It looks a bit ugly in full screen. We could do the same for all the boxes to avoid too much empty space.

06-36-y33w0-zmkk0

gonzaloriestra (approved):
LGTM! Just a couple of comments.

Btw, you can edit the PR to change the destination branch, so if you use `ink-select-test` instead of `main`, it will be much easier to review 🙂

packages/cli-kit/src/private/node/components/Banner.tsx
matteodepalo: I've thought about this and couldn't find a good way to do it with `flex-box`. I'll need to dig deeper, but I agree the outcome you describe would be ideal. The only problem I can see is having a bunch of boxes on top of the other of slightly different widths. Not sure if that would look pleasing to the eye.

matteodepalo:
> Btw, you can edit the PR to change the destination branch, so if you use ink-select-test instead of main, it will be much easier to review 🙂

In that case I'd need to remember to change it before merging, though. I guess it's better than the current state :D

packages/cli-kit/src/error.ts
gonzaloriestra: I see, thanks! You can forget about this 👍

packages/cli-kit/src/private/node/components/Banner.tsx
gonzaloriestra: Don't worry, I think it's ok for now, it was just a nitpicking suggestion.

You are right that having boxes with different sizes would be weird, but maybe we can define a minimum and just make it bigger when the content is too long. What do you think, @MeredithCastile?

matteodepalo: This has been reworked to be two thirds of the width of the terminal with a minimum width.

PR #572, opened 2022-10-04
Output messages with partners urls fixed when running against an spin…

alvaro-shopify:
- It's is not possible to convert a dev store to a test one (transfer disabled, this is needed at least in production environment for the app to work) for `1Party users` in a spin instance. Next error is returned from partners
image
- After testing `dev` with the conversion logic removed and verifying that everything was working, the conclusion is that there is no needing for converting the stores to a test one for `1party users` in spin

- In case `1party user` in spin, the conversion of the dev store to a test one is not executed.

### How to test your changes?
- Create a spin instance
- Run `SHOPIFY_CLI_1P_DEV=1 SHOPIFY_SERVICE_ENV="spin" SPIN_INSTANCE="<your_spin_instance_here>" yarn shopify app dev` 

### Measuring impact

### Checklist

### DIFF
--- a/packages/app/src/cli/services/dev/select-store.ts
+++ b/packages/app/src/cli/services/dev/select-store.ts
@@ -55,7 +55,7 @@ export async function selectStore(
     return store
   }
 
-  output.info(`\n${CreateStoreLink(org.id)}`)
+  output.info(`\n${await CreateStoreLink(org.id)}`)
   await system.sleep(5)
 
   const reload = await reloadStoreListPrompt(org)
--- a/packages/app/src/cli/services/dev/select-store.ts
+++ b/packages/app/src/cli/services/dev/select-store.ts
@@ -17,8 +17,8 @@ const InvalidStore = (storeName: string) => {
   )
 }
 
-const CreateStoreLink = (orgId: string) => {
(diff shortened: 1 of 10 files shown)


packages/app/src/cli/services/dev/select-store.ts
```diff
@@ -118,6 +118,15 @@ export async function convertToTestStoreIfNeeded(
...
 ): Promise<void> {
+  /**
+   * Is not possible to convert stores to dev ones in spin environmets. Should be created directly as development.
+   */
+  if (
+    environment.service.serviceEnvironment() === environment.network.Environment.Spin &&
+    environment.local.firstPartyDev()
+  ) {
+    return
+  }
```
isaacroldan, line 129: All stores used in the CLI are `development stores`, what this piece of code is convert them to "test" stores, which is just development stores with transfer disabled.
The comment should reflect that.

I'll try to abstract  `environment.service.serviceEnvironment() === environment.network.Environment.Spin` into a function (maybe inside environment?), this is something that can be reused to know if we are connecting to a spin instance (independently of where the CLI is)

matteodepalo (approved):

PR #575, opened 2022-10-04
Fix dev breakage when missing capabilities fields

dmhenry:

Fixes #573 

When `shopify.ui.extensions.toml` doesn't include a `capabilities` clause, the browser reports the following error:

```
vendors~app.latest.en.e5376343c5580b97aa7a.js:2 TypeError: Cannot read properties of undefined (reading 'networkAccess')
    at app.latest.en.280faa0996ccd0490853.js:1:183003
    at Array.map (<anonymous>)
    at app.latest.en.280faa0996ccd0490853.js:1:182862
````

This PR sets default values for `capabilities` and `capabilities.network_access` and `capabilities.blocks_progress`. Because omitting any of the defaulted fields in tests is regarded as a syntax error, there is no way I can see to write unit tests.

### How to test your changes?

Create a new checkout extension and `dev` it, ensuring the `shopify.ui.extensions.toml` file has no `capabilities` clause. Click the checkout extension preview link to ensure you are taken to the checkout screen.

Ensure you can see the checkout extension surface area and NOT this:

image

### Measuring impact

### Checklist

### DIFF
--- a/packages/app/src/cli/models/app/extensions.ts
+++ b/packages/app/src/cli/models/app/extensions.ts
@@ -30,10 +30,14 @@ export const UIExtensionConfigurationSchema = schema.define.object({
   extensionPoints: schema.define.array(schema.define.string()).optional(),
   capabilities: schema.define
     .object({
-      block_progress: schema.define.boolean().optional(),
-      network_access: schema.define.boolean().optional(),
+      block_progress: schema.define.boolean().optional().default(false),
+      network_access: schema.define.boolean().optional().default(false),
     })
-    .optional(),
+    .optional()
+    .default({
+      block_progress: false,
+      network_access: false,
+    }),
 
   // Only for CheckoutUiExtension
   settings: schema.define.any().optional(),
--- a/packages/app/src/cli/models/app/extensions.ts
+++ b/packages/app/src/cli/models/app/extensions.ts
@@ -30,14 +30,10 @@ export const UIExtensionConfigurationSchema = schema.define.object({
   extensionPoints: schema.define.array(schema.define.string()).optional(),
   capabilities: schema.define
     .object({
-      block_progress: schema.define.boolean().optional().default(false),
-      network_access: schema.define.boolean().optional().default(false),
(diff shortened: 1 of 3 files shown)


isaacroldan (approved):

vividviolet (changes requested):
Can we verify that setting these defaults don't break deploys? Is the issue only occurring in local development? Is this a new issue or has it always been a problem? Does Checkout return a default config for capabilities when it's missing in the extension through GraphQL? If so, we should make the Dev Server follow suit but not mutate the actual config that gets sent to Core.

I just want to get down to the root cause. I thought capabilities is optional.

isaacroldan:
> Can we verify that setting these defaults don't break deploys? 

Tried adding a subscription-ui and deploy worked

> Is the issue only occurring in local development? Is this a new issue or has it always been a problem? 

Only happens with the new Node flow, Go works fine because [it adds some fields](https://github.com/Shopify/cli/blob/main/packages/ui-extensions-go-cli/core/core.go#L73) to the config by default
Should this be the default behavior for Node too instead of this fix to the schema?

> Does Checkout return a default config for capabilities when it's missing in the extension through GraphQL? If so, we should make the Dev Server follow suit but not mutate the actual config that gets sent to Core.

Yes, it does. I think it makes sense that the dev-server automatically sets these capabilities if they are not present. But not modify what we read from the toml.

isaacroldan:
Since this only affects the node flow, i'd say is not urgent. @dmhenry Could we update this so that the Node server does the same as the Go one?

isaacroldan:
I think we shouldn't add defaults to the schema, because that schema will be used both for dev and deploy. If we want the `dev` server to include defaults that should be done in the server itself, what do you think?

vividviolet:
> I think we shouldn't add defaults to the schema, because that schema will be used both for dev and deploy. If we want the `dev` server to include defaults that should be done in the server itself, what do you think?

I agree that it's safest to include defaults in the dev server itself instead of modifying the real configs

PR #583, opened 2022-10-05
Add jsdoc support for eslint

Arkham:
We want to remove `@param` and `@returns` tags that don't add information.

We also make all warnings into errors so that CI fails.

I recently noticed that we have some duplicate information between jsdoc comments and typescript type signatures. So in this PR i've added an eslint plugin to detect `@param` and `@returns` annotations that don't add information.

- add jsdoc support for eslint
- remove offending lines to get the CI back to 🍏 
- turn all eslint warnings into errors so they can fail the CI

### How to test your changes?

- run `yarn lint`

### Measuring impact

### Checklist


Arkham:
> > turn all eslint warnings into errors so they can fail the CI
> 
> Where is that? 🤔

I added that `eslint-plugin-only-error` but then realized it's not necessary :)

Issue #532, opened 2022-09-28
[Bug]: Theme List after Auth never completes


sb18281:
### Please confirm that you have:

- [X] Searched [existing issues](.) to see if your issue is a duplicate. (If you’ve found a duplicate issue, feel free to add additional information in a comment on it.)
- [X] Reproduced the issue in [the latest CLI version](https://www.npmjs.com/package/@shopify/cli).

### In which of these areas are you experiencing a problem?

Theme

### Expected behavior

CLI command should complete, and return back to the console

### Actual behavior

After running `shopify theme list --store=xx` it  proceeded through auth successfully. Returned and displayed the theme list, but then seems to hang and never complete. 

Input seems to work (typed test), but it seems to have fully frozen at that point.

![image](https://user-images.githubusercontent.com/2426010/192665790-8e31397e-c9ed-42c7-990a-b7af40acfefc.png)

### Reproduction steps

1. Run `shopify auth logout`
2. Run `shopify theme list --store=xx` with the store flag
3. Login and complete auth to continue the process listing the theme

### Operating System

Ubuntu 20.04.5

### Shopify CLI version (check your project's `package.json` if you're not sure)

3.15.0

### Shell

zsh

### Node version (run `node -v` if you're not sure)

v16.17.0

PR #598, opened 2022-10-06
Fix cli process not finishing in some cases

isaacroldan:
- This https://github.com/Shopify/cli/issues/532 issue reported that the CLI won't finish the execution if the flow included auth.
- Turns out we had this problem for a while, and any command that required auth won't finish on its own, you'd need to finish it yourself with ctrl+c. Now for dev this is OK because you have to close it manually anyway, so that's why it might have gone undetected until now.
- The problem was in the `keypress()` method used to wait for user keypress, it wasn't closing the file descriptor opened to wait for a key press.

Fixes https://github.com/Shopify/cli/issues/532

- Refactor `keypress` to close the listener after a key press.
- Also took the opportunity to improve it and not accept `CTRL+C` as a valid key press, instead it will finish the execution as intended.
- Fixed an unrelated issue where the server was requesting `favicon.ico` but we were only expecting a `favicon.svg`

### Measuring impact

### Checklist

### DIFF
--- a/packages/cli-kit/src/ui.ts
+++ b/packages/cli-kit/src/ui.ts
@@ -1,6 +1,6 @@
-import {CancelExecution, Abort} from './error.js'
+import {CancelExecution, Abort, AbortSilent} from './error.js'
 import {remove, exists} from './file.js'
-import {info, completed, content, token, logUpdate, Message, Logger, stringifyMessage} from './output.js'
+import {info, completed, content, token, logUpdate, Message, Logger, stringifyMessage, debug} from './output.js'
 import {colors} from './node/colors.js'
 import {relative} from './path.js'
 import {isTerminalInteractive} from './environment/local.js'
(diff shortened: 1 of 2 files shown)


packages/cli-kit/src/ui.ts
```diff
@@ -172,12 +172,23 @@ export async function terminateBlockingPortProcessPrompt(port: number, stepDescr
 }
 
 export const keypress = async () => {
-  process.stdin.setRawMode(true)
-  process.stdin.resume()
-  return new Promise<void>((resolve) =>
-    process.stdin.once('data', () => {
+  return new Promise((resolve, reject) => {
+    const handler = (buffer: Buffer) => {
+      process.stdin.removeListener('data', handler)
```
Arkham, line 177: ❓ is this necessary? reading the code the listener is installed only once, so in theory if we are inside the callback there is no need to remove the listener, or am I reading this wrong?

packages/cli-kit/src/ui.ts
```diff
@@ -172,12 +172,23 @@ export async function terminateBlockingPortProcessPrompt(port: number, stepDescr
...
       process.stdin.setRawMode(false)
-      resolve()
-    }),
-  )
+      process.stdin.pause()
+
+      const bytes = Array.from(buffer)
+
+      if (bytes.length && bytes[0] === 3) {
```
Arkham, line 183: fun fact:

```
ctrl+a: 1
ctrl+b: 2
ctrl+c: 3
```

Arkham (approved):
Nice catch, LGTM! 👍

PR #606, opened 2022-10-10
Document how to ensure cross-OS support

pepicrft:
We (people working at Shopify) develop the CLI from a **macOS environments** so if we are not careful enough we might end up building workflows that don't work in other OSs (e.g. Windows, Linux) or introducing regressions. Breaking workflows leads to developers' frustrations and erodes their trust in their platform.

This PR documents:
- How to implement workflows that work across OSs using the modules exported by `@shopify/cli-kit`, and provides.
- How to manually test workflows on Windows and Linux using virtualized environments.
- When we should leverage unit tests to catch broken contracts with the OS in CI

### How to test your changes?
Read the docs and feel free to make suggestions.

### Measuring impact

### Checklist

### DIFF
--- a/docs/cross-os-compatibility.md
+++ b/docs/cross-os-compatibility.md
@@ -22,23 +22,18 @@ Please don't assume that a successful working workflow in the OS in which it was
 
 #### Linux ([Podman](https://podman.io/))
 
-Run the following command from the CLI directory to create an temporary virtual Linux environment with the directory mounted in it:
+Run the following command from the CLI directory to create an temporary virtual Linux environment:
 
 ```bash
-podman run \
-  --rm --interactive --tty \
-  --volume "$(pwd):/src" \
-  --workdir "/src" \
-  node:18 \
-  /bin/bash
+podman run --rm --interactive --tty node:18 /bin/bash
 ```
 
-Then run `yarn install` to resolve and pull all the project dependencies into the environment.
+Then clone the [CLI repository](https://github.com/Shopify/cli) and install the dependencies with `yarn install`.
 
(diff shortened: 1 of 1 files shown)


docs/cross-os-compatibility.md
```diff
@@ -0,0 +1,48 @@
...
+  node:18 \
+  /bin/bash
+```
+
+Then run `yarn install` to resolve and pull all the project dependencies into the environment.
+
+
+#### Windows ([Parallels](https://www.parallels.com/pd/general/))
+
+After you've installed Parallels and virtualized the Windows environment, go to `Actions > Configure > Options (Tab) > Sharing > Share Custom Mac folders with Windows` and share the folder containing the CLI repository. Then open the coherence mode through `View > Enter Coherence`. The coherence mode will show Windows' windows as if they were part of the host. You might need to run `yarn install` from the Windows shell to ensure dependencies with native extensions pull the binaries for the Windows architecture. You might also need to run the following command if Yarn throws errors related to unsigned scripts:
```
matteodepalo, line 41: I thought that reusing the same folder was a good idea, but then realized that the dependencies in `node_modules` need to be owned by either OS. If you switch back to mac after running `yarn install` on windows you will get permission denied errors. I think it's better to encourage an approach that uses git to pull the repo and install the dependencies in a separate folder.

pepicrft: That's a good point. You don't want to be messing with the `node_modules` directory. I'll update it.

gonzaloriestra (approved):
👌

PR #608, opened 2022-10-10
Fixed method to check in which cloud environment the cli is running

alvaro-shopify:
- Fixes the method that checks if the CLI is run inside a cloud environment. Among other things, it is used by the CLI to select the type of authentication it must be used. Right now, in case the CLI is run inside a cloud environment, the device authentication method must be used. The problem now is that the normal authentication flow is being used in cloud environments so it is not working.

- Environment variables variables used to verify the cloud environments are not booleans, so checking their existence is should be just enough as a validation

### How to test your changes?

1. Fork the CLI in your own repo
2. Add CLI repo as a remote to checkout the branch `git remote add coworker git://github.com/Shopify/cli.git`
3. Checkout the branch and push it your own repo
   - `git fetch cli bugfix/fix_checking_cloud_environments`
   - `git checkout bugfix/fix_checking_cloud_environments`
4. `git push origin bugfix/fix_checking_cloud_environments`
5. Open this link in the browser (modify it with your repo data) `https://gitpod.io/#https://github.com/<oyu_repo>/<forked_porject>`
6. Checkout the branch ` git checkout bugfix/fix_checking_cloud_environments`
7. Run `yarn shopify app dev --path ./fixtures/app`

NOTE: Skip first five step and go directly to 6th using my repo `https://gitpod.io/#https://github.com/alvaro-shopify/cli`

### Measuring impact

### Checklist

### DIFF
--- a/packages/cli-kit/src/environment/local.ts
+++ b/packages/cli-kit/src/environment/local.ts
@@ -124,7 +124,7 @@ export function cloudEnvironment(env: NodeJS.ProcessEnv = process.env): {
   platform: 'spin' | 'codespaces' | 'gitpod' | 'localhost'
   editor: boolean
 } {
-  if (isTruthy(env[constants.environmentVariables.codespaces])) {
+  if (isSet(env[constants.environmentVariables.codespaces])) {
     return {platform: 'codespaces', editor: true}
   }
   if (isSet(env[constants.environmentVariables.gitpod])) {


packages/cli-kit/src/environment/local.test.ts
```diff
@@ -178,7 +178,7 @@ describe('cloudEnvironment', () => {
     ['GITPOD_WORKSPACE_URL', 'gitpod'],
   ])('returns correct cloud platform when %s is truthy', (envVar, platform) => {
```
gonzaloriestra, line 179:
```suggestion
  ])('returns correct cloud platform when %s exists', (envVar, platform) => {
```

Arkham (approved):
LGTM, thanks :)

packages/cli-kit/src/environment/local.test.ts
alvaro-shopify: `isTruthy` validates possible values that matches a boolean value. In this case, the values are homogeneous, so checking their existence is enough to consider them as a cloud environment

isaacroldan: Actually, do you think this works for all systems? `GITPOD_WORKSPACE_URL` will provide a URL, not a truthy value...

alvaro-shopify: It is working for the three cloud environments cli are currently supporting:
* gitpod
* codespaces
* spin
It's only checking that the environment variable exists. Maybe we could add a different validation depending on the value type

isaacroldan: ah yeah, sorry i misread

packages/cli-kit/src/environment/local.ts
```diff
@@ -127,10 +127,10 @@ export function cloudEnvironment(env: NodeJS.ProcessEnv = process.env): {
   if (isTruthy(env[constants.environmentVariables.codespaces])) {
     return {platform: 'codespaces', editor: true}
   }
-  if (isTruthy(env[constants.environmentVariables.gitpod])) {
+  if (isSet(env[constants.environmentVariables.gitpod])) {
```
isaacroldan, line 130: I'd use `isSet` for codespaces too

PR #615, opened 2022-10-11
[Fix] localization typing inside ExtensionPayload (ui-extensions-server-kit)

Safi1012:
Fixes a false typing for the ExtensionPayload interface (inside of `ui-extensions-server-kit`). 

The received localizations inside the DEV-CLI have the following format:

```
{
  "defaultLocale": "en",
  "translations": {
    "fr": {
      "welcome": "Bienvenue dans l'extension customer-accounts-ui!",
      "apples.one": "Tu as {{count}} pomme",
      "apples.other": "Tu as {{count}} pommes"
    },
    "en": {
      "welcome": "Welcome to the customer-accounts-ui extension!",
      "apples.one": "You have {{count}} apple",
      "apples.other": "You have {{count}} apples"
    }
  },
  "lastUpdated": 1665411936350
}
```

As a result the old (broken) typing doesn't match the received format:

OLD:
```
translations: {[locale: string]: string}
```

NEW:
```
translations: {[locale: string]: ExtensionTranslationMap}
```

I need this fix in order to have the correct typing inside of the customer-accounts-web project, which makes uses of the `ui-extensions-server-kit`. The following PR requires the typing fix: https://github.com/Shopify/customer-account-web/pull/1023

It just updates a single falsely typed property.

### Measuring impact

### Checklist

### DIFF
--- a/packages/ui-extensions-server-kit/src/types.ts
+++ b/packages/ui-extensions-server-kit/src/types.ts
@@ -78,7 +78,7 @@ export interface ExtensionPayload {
   localization?: {
     defaultLocale: string
     lastUpdated: number
-    translations: {[locale: string]: string}
+    translations: {[locale: string]: ExtensionTranslationMap}
   }
 }
 
@@ -102,4 +102,8 @@ export interface App {
   }
   supportEmail?: string
   supportLocales?: string[]
+}
+
+interface ExtensionTranslationMap {
+  [key: string]: string
 }


packages/ui-extensions-server-kit/src/types.ts
```diff
@@ -103,3 +103,7 @@ export interface App {
   supportEmail?: string
   supportLocales?: string[]
 }
+
+interface ExtensionTranslationMap {
+  [key: string]: string
```
DiagoSmith, line 108: Is it possible to have nested objects or are translations always a flat structure ?

TobiasBales (approved):
LGTM if no one else is relying on the old format - did we add it - I think so but just want to be sure

Safi1012:
@TobiasBales > LGTM if no one else is relying on the old format - did we add it - I think so but just want to be sure

Yep, I added this tying 1-2 weeks ago and messed it up 😅

packages/ui-extensions-server-kit/src/types.ts
Safi1012: Ah, the `app/foundation/i18n/en.json` file is just being used for the Hub UI translations and not for the extensions. For  `customer-account-ui-extensions` we won't have nested objects translations for now. 

Thanks for your comment :)

alfonso-noriega (approved):

pepicrft (approved):

PR #625, opened 2022-10-12
Fix initializeRepository for git versions before 2.28

Arkham:

Fixes https://github.com/Shopify/shopify-cli-planning/issues/423

We were using `git init --initial-branch`, but this is only supported by git 2.28+.

Instead of using `git init --initial-branch`, we do `git init` followed by `git checkout -b`. As this [SO answer](https://stackoverflow.com/questions/42871542/how-can-i-create-a-git-repository-with-the-default-branch-name-other-than-maste) points out, there will be leftover `master` branch since there are no commits yet.

### Measuring impact

### Checklist

### DIFF
--- a/.changeset/calm-islands-pay.md
+++ b/.changeset/calm-islands-pay.md
@@ -1,0 +1,5 @@
+---
+'@shopify/cli-kit': patch
+---
+
+Support git 2.28 or earlier when initializing a repository


lynchv (approved):
LGTM! Thank you for taking care of this 😄

matteodepalo (approved):

PR #654, opened 2022-10-20
Introduce ExternalError and output external errors with a new component

matteodepalo:

@isaacroldan reported that he saw an error that broke the borders of the banner component.

Captura de Pantalla 2022-10-20 a las 12 44 31

I've introduced a new type of error called `ExternalError`. This is used when rescuing errors from `execa` and allows us to print these kind of errors differently, like so:

Screenshot 2022-11-01 at 13 49 19

### Measuring impact

### Checklist

### DIFF
--- a/packages/cli-kit/src/private/node/ui/components/FatalError.tsx
+++ b/packages/cli-kit/src/private/node/ui/components/FatalError.tsx
@@ -4,6 +4,7 @@ import {Bug, cleanSingleStackTracePath, Fatal} from '../../../../error.js'
 import {Box, Text} from 'ink'
 import React from 'react'
 import StackTracey from 'stacktracey'
+import stripAnsi from 'strip-ansi'
 
 export interface FatalErrorProps {
   error: Fatal
@@ -34,7 +35,7 @@ const FatalError: React.FC<FatalErrorProps> = ({error}) => {
   return (
     <Banner type="error">
       <Box>
-        <Text>{error.message}</Text>
+        <Text>{stripAnsi(error.message)}</Text>
       </Box>
 
       {error.tryMessage && (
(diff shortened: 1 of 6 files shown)


isaacroldan:
Fast! ⚡ 
Were you able to reproduce a similar error without the box bug?

matteodepalo:
Unfortunately no :/ I couldn't find an error that would have that behavior. Did you run some specific command?

isaacroldan:
Yeah, this was on a specific project that didn't support node18, so the `yarn install` we execute inside `dev` fails.
I'll try to tophat this today 🙏

isaacroldan:
Captura de Pantalla 2022-10-21 a las 13 54 32
Works 👌

isaacroldan (approved):

packages/cli-kit/src/private/node/ui/components/FatalError.tsx
```diff
@@ -4,6 +4,7 @@ import {Bug, cleanSingleStackTracePath, Fatal} from '../../../../error.js'
 import {Box, Text} from 'ink'
 import React from 'react'
 import StackTracey from 'stacktracey'
+import stripAnsi from 'strip-ansi'
```
pepicrft, line 7: I'd add a changeset update to mention the fix.

matteodepalo:
@amcaplan wow that link is really interesting! Definitely something to consider in the future. If we want to respect 3P output properly we need something like that in place. For now though I'd rather merge this PR, but I'll write a note for that. I also need to make a spike about 3P errors in general so it will come useful soon.

matteodepalo:
I don't think we can be that surgical and remove only the characters to preserve indentations. I would go with a flat line at the top and everything that comes from external tools aligned to the left as it's the approach that has less risk.

matteodepalo:
Screenshot 2022-11-01 at 13 49 19
@MeredithCastile I've just made the above change in this PR. How does that look? We can say "error coming from _command that was run_".

matteodepalo:
@isaacroldan @pepicrft could I get another review? The code has changed quite a lot.

isaacroldan (approved):

Issue #546, opened 2022-09-30
[Bug]: shopify theme dev session "timeout" 


ceri-waterscreative:
### Please confirm that you have:

- [X] Searched [existing issues](.) to see if your issue is a duplicate. (If you’ve found a duplicate issue, feel free to add additional information in a comment on it.)
- [X] Reproduced the issue in [the latest CLI version](https://www.npmjs.com/package/@shopify/cli).

### In which of these areas are you experiencing a problem?

Theme

### Expected behavior

Should be prompted to log back in

### Actual behavior

Served an error message 

"[API] Service is not valid for authentication"

### Reproduction steps

1. Run shopify theme dev
2. Leave it running until login session expires
3. Error Appears

### Operating System

Windows 11

### Shopify CLI version (check your project's `package.json` if you're not sure)

3.15

### Shell

Git Bash

### Node version (run `node -v` if you're not sure)

16.15.1

### What language and version are you using in your application?

N/A

notrealdev:
Same here with notice `"[API] Service is not valid for authentication"`
Is there an option to increase the sesion timeout limit?

isaacroldan:
Just to check, are you using the latest CLI version? (3.22.1)

notrealdev:
Hi @isaacroldan , if I do nothing after some time, it disconnects itself without giving a warning.
How can I stop this, every time like this I have to run `shopify theme dev` again, it's inconvenient

notrealdev:
Just updated to v3.28.0 and it's fixed! After 2 hrs I'm still logged in. Thank you @isaacroldan!

PR #660, opened 2022-10-21
Automatically refresh Theme session

isaacroldan:
Up until now our commands execution only needed a valid token during the first seconds (maybe minutes). 
The only command with a long live period is `app dev` and it doesn't need a token once the servers are up.

But with themes this is different, during `theme dev` a valid token is always required to push changes to the remote store.

Fixes https://github.com/Shopify/cli/issues/546

- Before running `theme dev`, always refresh the tokens to be sure they are valid for at least 90min. 
- After running `theme dev` and every 90 min: close the ruby process, refresh tokens and execute it again with updated tokens.

❓  Tokens expire at 120min, should we refresh the tokens every 100, 110 or 115 min maybe?

### How to test your changes?
- To test that the process finishes, refreshes and starts again, you can modify `ThemeRefreshTimeouInMinutes` to something smaller (maybe 0.5)
- To test that the process works always, just run `theme dev` for hours...

### Measuring impact

### Checklist

### DIFF
--- a/.changeset/hip-dogs-kiss.md
+++ b/.changeset/hip-dogs-kiss.md
@@ -1,0 +1,6 @@
+---
+'@shopify/cli-kit': patch
+'@shopify/theme': patch
+---
+
+Refresh theme dev session every 90 minutes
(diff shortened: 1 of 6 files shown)


matteodepalo (approved):
💯

karreiro (approved):
Thank you, @isaacroldan!

PR #555, opened 2022-10-03
Upgrade TS 4.84, OCLIF 1.16.4

shauns:
### WHY are these changes introduced?

Fixes https://github.com/Shopify/shopify-cli-planning/issues/394

### WHAT is this pull request doing?

- Upgrades Typescript to 4.8.4 (current latest)
- Upgrades OCLIF to 1.16.4 (also the current latest)
- Fixes some very small incompatibilities

### How to test your changes?

We'll want to do a full run through really. Main thing to watch out for are changes in how OCLIF behaves, e.g. when it flushes output.

### Measuring impact

How do we know this change was effective? Please choose one:

- [x] n/a - this doesn't need measurement, e.g. a linting rule or a bug-fix
- [ ] Existing analytics will cater for this addition
- [ ] PR includes analytics changes to measure impact

### Checklist

- [x] I've considered possible cross-platform impacts (Mac, Linux, Windows)
- [x] I've considered possible [documentation](https://shopify.dev) changes
- [x] I've made sure that any changes to `dev` or `deploy` have been reflected in the [internal flowchart](https://www.figma.com/file/7vqUp50u6dm48Zfb4JRRn8/CLI3-Internals).

PR #662, opened 2022-10-24
Fix: Temporarily disable env var arguments for boolean default-true/exclusive flags

amcaplan:

Apparently these never worked anyway for boolean flags before oclif 1.16.1. Then they started working, but only to specify true; otherwise, they essentially force a false as though it were specified.

More details: https://github.com/oclif/core/pull/536

Comments out the env-sourced way to set these boolean flags.

### How to test your changes?

```
$ shopify app dev --path=fixtures/app --tunnel-url=https://my-tunnel-url:port
```

On `main` you get this error:

```
│  The following errors occurred:                                              │
│    --tunnel-url=https://my-tunnel-url:port cannot also be provided when      │
│  using --no-tunnel                                                           │
│    --tunnel-url=https://my-tunnel-url:port cannot also be provided when      │
│  using --tunnel                                                              │
│  See more help with --help                                                   │
```

On this branch, the command proceeds successfully.

### Measuring impact

### Checklist

### DIFF
--- a/packages/app/src/cli/commands/app/dev.ts
+++ b/packages/app/src/cli/commands/app/dev.ts
@@ -58,17 +58,19 @@ export default class Dev extends Command {
       env: 'SHOPIFY_FLAG_TUNNEL_URL',
       exclusive: ['no-tunnel', 'tunnel'],
     }),
+    // eslint-disable-next-line rulesdir/command-flags-with-env
     'no-tunnel': Flags.boolean({
       hidden: true,
       description: 'Automatic creation of a tunnel is disabled. Service entry point will listen to localhost instead',
-      env: 'SHOPIFY_FLAG_NO_TUNNEL',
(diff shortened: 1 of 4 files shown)


alvaro-shopify (approved):
* In case we revert [Upgrade TS 4.84, OCLIF 1.16.4](https://github.com/Shopify/cli/pull/555) will also do the trick?. 
* is it better to revert and wait for the [oclif fix](https://github.com/oclif/core/pull/536) to be solved?

isaacroldan (approved):

amcaplan:
@alvaro-shopify there's little benefit to reverting that upgrade, as previously none of the env settings for boolean flags were working. (Support for that was introduced in 1.16.1... it just also happened to break these flags.) So the functionality is better with the sum total change than without it.

PR #666, opened 2022-10-24
Downgrade oclif and typescript

amcaplan:

Upgrading oclif introduced bugs with boolean flags. Need to downgrade again until https://github.com/oclif/core/pull/536 is merged to oclif.

This is an alternative approach vs https://github.com/Shopify/cli/pull/662

Downgrades oclif and typescript, essentially reverting https://github.com/Shopify/cli/pull/555 (but with some extra work because the `base-command` code has changes)

### How to test your changes?

Run the CLI...

### Measuring impact

### Checklist

### DIFF
--- a/packages/cli-main/package.json
+++ b/packages/cli-main/package.json
@@ -51,7 +51,7 @@
     ]
   },
   "dependencies": {
-    "@oclif/core": "1.16.4",
+    "@oclif/core": "1.9.2",
     "@oclif/plugin-commands": "^2.2.0",
     "@oclif/plugin-help": "^5.1.14",
     "@oclif/plugin-plugins": "^2.1.1",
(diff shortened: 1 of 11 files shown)


packages/cli-main/package.json
```diff
@@ -51,7 +51,7 @@
     ]
   },
   "dependencies": {
-    "@oclif/core": "1.16.4",
+    "@oclif/core": "1.9.2",
     "@oclif/plugin-commands": "^2.2.0",
     "@oclif/plugin-help": "^5.1.14",
```
shauns, line 56: these two should also be nudged back:
image

amcaplan: @shauns excellent point, I've updated those as well.

PR #677, opened 2022-10-26
Don't report handled tunnel plugin errors to Bugsnag

alvaro-shopify:
Right now every error returned from the Ngrok plugin is reported to Bugsnag. Most of that errors are handled directly by the plugin showing a suitable error output message to the users. These errors include:
* Run two instances of the ngrok plugin
* Incorrect ngrok auth token
These are users errors not bugs, so they not should be reported.

* Moving the `Result` type to the `public` folder following the convention (sorry because no diff possible with previous version :( )
* Adding new methods to the `Result` type to make the using easier:
  * `map` - Change the type of the result value when the response is ok
  * `doOnOk` - Let you execute an action when the result is ok without modifying it.
* `tunnel_start` hooks now return a `Result` type. It supports these errors `tunnel-already-running | wrong-credentials | unknown`
* Only `unknown` errors are reported to Bugsnag, the rest abort silently the execution of the command as the error output is handled by the plugin

### How to test your changes?
- Run `yarn shopify app dev` and the tunnel should be created 

### Measuring impact

- [x ] n/a - this doesn't need measurement, e.g. a linting rule or a bug-fix

### Checklist


packages/plugin-ngrok/src/tunnel.test.ts
```diff
@@ -63,12 +62,13 @@ describe('start', () => {
...
 
     // Then
-    expect(error.Abort).toHaveBeenCalledWith(
+    expect(AbortError).toHaveBeenCalledWith(
       'The ngrok tunnel could not be started.\n\nYour account has been suspended',
       undefined,
     )
-    expect(renderFatalError).toHaveBeenCalledWith(expect.any(error.Abort))
-    expect(got).toEqual({url: undefined})
+    expect(renderFatalError).toHaveBeenCalledWith(expect.any(AbortError))
```
pepicrft, line 69: Can't we merge this into?:
```js
expect(renderFatalError)
  .toHaveBeenCalledWith(new AbortError('The ngrok tunnel could not be started.\n\nYour account has been suspended', undefined))
```

packages/plugin-ngrok/src/tunnel.test.ts
```diff
@@ -93,16 +95,18 @@ describe('start', () => {
...
+    expect(AbortError).toHaveBeenCalledWith(
       'The ngrok tunnel could not be started.\n\nerror message contains code err_ngrok_108',
       expect.stringContaining(
         'Kill all the ngrok processes with \u001b[1m\u001b[33mtaskkill /f /im ngrok.exe\u001b[39m\u001b[22m',
       ),
     )
-    expect(renderFatalError).toHaveBeenCalledWith(expect.any(error.Abort))
+    expect(renderFatalError).toHaveBeenCalledWith(expect.any(AbortError))
+    expect(got.isErr() && got.error.type).toEqual('tunnel-already-running')
+    expect(got.isErr() && got.error.message).toEqual('error message contains code err_ngrok_108')
```
pepicrft, line 109: Nitpick:

```js
 expect(got.error?.type).toEqual('tunnel-already-running')
 expect(got.error?.message).toEqual('error message contains code err_ngrok_108')
```

The presence of an error in `got.error` implicitly tells that the result represents an error so there's no need for `got.isErr()`

packages/cli-kit/src/plugins/tunnel.ts
```diff
@@ -1,5 +1,15 @@
+import {err, Result} from '../public/common/result.js'
 import {FanoutHookFunction, PluginReturnsForHook} from '../plugins.js'
 
+export type TunnelErrorType = 'invalid-provider' | 'tunnel-already-running' | 'wrong-credentials' | 'unknown'
+export class TunnelError extends Error {
```
pepicrft, line 5: I'd inherit from [`ExtendableError`](https://github.com/Shopify/cli/blob/main/packages/cli-kit/src/error.ts#L7). Extendable error is a custom subclass of `Error` that allows doing comparisons.

packages/app/src/cli/services/dev/urls.ts
```diff
@@ -151,3 +150,16 @@ export async function shouldOrPromptUpdateURLs(options: ShouldOrPromptUpdateURLs
   }
   return shouldUpdate
 }
+
+function mapRunTunnelPluginError(tunnelPluginError: plugins.TunnelPluginError) {
```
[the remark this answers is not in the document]
alvaro-shopify: * I'm going to change hardcoded `ngrok` for the value of the requested provider
* x should be a url to a list of available plugins right?, what should we use right now?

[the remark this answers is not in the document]
pepicrft: I think it's fine for now since we don't have a published list of tunnel plugins yet. Once we have it, we can adjust the message to add an hyperlink to it.

packages/plugin-ngrok/src/tunnel.test.ts
pepicrft: Oh I see! Today I learned about [type predicates](https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates). It makes sense to keep it then.

PR #678, opened 2022-10-26
Add new checkout UI template

### LINKED ISSUE #671: CLI: Add Feature flag to opt into new functionality
- [ ] Add a new flag to the CLI.  We will use this flag to hide new CLI behavior from developers.  If you run commands with the new flag you’ll get the new behavior
- [ ] Once the plugin architecture lands we’ll delete this and use a plugin

https://github.com/Shopify/cli/issues/671#tasklist-block-11ce4c9d-02b1-417e-8af8-d4e308116961

### LINKED ISSUE #673: CLI: yarn generate extension creates the new checkout UI extension
- [ ] Add the checkout UI extension to the list (hard coded for now)
- [ ] Selecting the new item creates a new template
- [ ] Use existing template, but add admin settings extension point and use the new toml format
- [ ] Set some one from checkout extensibility as a reviewer

byrichardpowell:

Fixes #671
Fixes #673

Adds a new option to `yarn generate extension`:

Screen Shot 2022-10-26 at 3 04 34 PM

This option should only be available to shopifolk.  Selecting this option creates a new template:

Screen Shot 2022-10-26 at 3 06 26 PM

This template is based upon the existing checkout UI extension with the following changes:

1. `shopify.ui.extension.toml` uses the new schema.
2. Two extension points are created (`Checkout::Dynamic::Render` and `Admin::Checkout::Editor`).
3. The readme has updated instructions to reflect that both of these extension points are created.

### How to test your changes?

1. Delete `fixtures/app/extensions`
2. Inside `fixtures/app` run `yarn generate extension` and select a pre-existing UI extension
3. Repeat step 2 for every UI extension EXCEPT the new one.
4. Run `yarn dev` to check that nothing broke the generation of existing extensions.
5. Run `yarn generate extension` and this time select the new UI Extension ( `Checkout UI Beta (shopifolk only)` ).

Creating a new extension should work. Note that if you run any command after creating the new extension it will fail because the CLI doesn't understand the new `toml` schema.

### Post-release steps

None

### Measuring impact

### Checklist

### DIFF
--- a/packages/app/templates/ui-extensions/projects/checkout_ui_beta/shopify.ui.extension.toml.liquid
+++ b/packages/app/templates/ui-extensions/projects/checkout_ui_beta/shopify.ui.extension.toml.liquid
@@ -6,7 +6,7 @@ block_progress = true
 
 [[extension_points]]
 target = "Checkout::Dynamic::Render"
-module = "./src/checkoutDynamicRender.{{ srcFileExtension }}"
+module = "./src/CheckoutDynamicRender.{{ srcFileExtension }}"
 
 # [[extension_points.metafields]]
 # namespace = "my-namespace"
@@ -14,7 +14,7 @@ module = "./src/checkoutDynamicRender.{{ srcFileExtension }}"
 
 [[extension_points]]
 target = "Admin::CheckoutEditor::RenderSettings"
-module = "./src/adminCheckoutEditorRenderSettings.{{ srcFileExtension }}"
+module = "./src/AdminCheckoutEditorRenderSettings.{{ srcFileExtension }}"
 
 # Read more on extension settings at https://shopify.dev/api/checkout-extensions/checkout/configuration
 # [settings]
--- a/packages/app/src/cli/models/app/extensions.ts
+++ b/packages/app/src/cli/models/app/extensions.ts
@@ -16,18 +16,29 @@ export interface Extension {
   graphQLType: string
 }
 
+const metafieldsSchema = schema.define
+  .array(
+    schema.define.object({
+      namespace: schema.define.string(),
+      key: schema.define.string(),
+    }),
+  )
+  .default([])
+
+export const OldExtensionPointsSchema = schema.define.array(schema.define.string()).default([])
+export const NewExtensionPointsSchema = schema.define.array(
+  schema.define.object({
+    target: schema.define.string(),
+    module: schema.define.string(),
+    metafields: metafieldsSchema,
+  }),
+)
+
 export const UIExtensionConfigurationSchema = schema.define.object({
   name: schema.define.string(),
-  type: schema.define.enum(uiExtensions.types),
-  metafields: schema.define
-    .array(
-      schema.define.object({
-        namespace: schema.define.string(),
-        key: schema.define.string(),
-      }),
-    )
-    .default([]),
-  extensionPoints: schema.define.array(schema.define.string()).optional(),
+  type: schema.define.enum(uiExtensions.types).default('ui_extension'),
+  metafields: metafieldsSchema,
+  extensionPoints: schema.define.union([OldExtensionPointsSchema, NewExtensionPointsSchema]).optional(),
   capabilities: schema.define
     .object({
       block_progress: schema.define.boolean().optional(),
@@ -49,7 +60,7 @@ export const UIExtensionConfigurationSchema = schema.define.object({
 })
 
 export const UIExtensionConfigurationSupportedSchema = UIExtensionConfigurationSchema.extend({
-  type: schema.define.enum([...uiExtensions.types, ...uiExternalExtensionTypes.types]),
+  type: schema.define.enum([...uiExtensions.types, ...uiExternalExtensionTypes.types]).default('ui_extension'),
 })
 
 export const FunctionExtensionConfigurationSchema = schema.define.object({
@@ -101,7 +112,7 @@ export type ThemeExtension = Extension & {
 
 export type UIExtension = Extension & {
   configuration: UIExtensionConfiguration
-  entrySourceFilePath: string
+  entrySourceFilePath?: string
   outputBundlePath: string
   devUUID: string
 }
@@ -110,3 +121,4 @@ type UIExtensionConfiguration = schema.define.infer<typeof UIExtensionConfigurat
 type FunctionExtensionConfiguration = schema.define.infer<typeof FunctionExtensionConfigurationSchema>
 type ThemeExtensionConfiguration = schema.define.infer<typeof ThemeExtensionConfigurationSchema>
 type FunctionExtensionMetadata = schema.define.infer<typeof FunctionExtensionMetadataSchema>
+export type NewExtensionPointType = schema.define.infer<typeof NewExtensionPointsSchema>
--- a/packages/app/src/cli/models/app/loader.ts
+++ b/packages/app/src/cli/models/app/loader.ts
@@ -7,6 +7,7 @@ import {
   ThemeExtensionConfigurationSchema,
   UIExtensionConfigurationSupportedSchema,
   Extension,
+  OldExtensionPointsSchema,
 } from './extensions.js'
 import {AppConfigurationSchema, Web, WebConfigurationSchema, App, AppInterface, WebType} from './app.js'
 import {configurationFileNames, dotEnvFileNames, extensionGraphqlId} from '../../constants.js'
@@ -21,6 +22,7 @@ import {
   usesWorkspaces as appUsesWorkspaces,
 } from '@shopify/cli-kit/node/node-package-manager'
 import {resolveFramework} from '@shopify/cli-kit/node/framework'
+import type {NewExtensionPointType} from './extensions.js'
 
 const defaultExtensionDirectory = 'extensions/*'
 
@@ -233,6 +235,68 @@ class AppLoader {
     return parseResult.data
   }
 
+  async getLegacyEntrySourceFilePath(directory: string) {
+    const entrySourceFilePath = (
+      await Promise.all(
+        ['index']
+          .flatMap((name) => [`${name}.js`, `${name}.jsx`, `${name}.ts`, `${name}.tsx`])
+          .flatMap((fileName) => [`src/${fileName}`, `${fileName}`])
+          .map((relativePath) => path.join(directory, relativePath))
+          .map(async (sourcePath) => ((await file.exists(sourcePath)) ? sourcePath : undefined)),
+      )
+    ).find((sourcePath) => sourcePath !== undefined)
+
+    if (!entrySourceFilePath) {
+      this.abortOrReport(
+        output.content`Couldn't find an index.{js,jsx,ts,tsx} file in the directories ${output.token.path(
+          directory,
+        )} or ${output.token.path(path.join(directory, 'src'))}`,
+        undefined,
+        directory,
+      )
+    }
+
+    return entrySourceFilePath
+  }
+
+  async validateUIExtensionPointConfig(directory: string, extensionPoints: NewExtensionPointType) {
+    const errors: string[] = []
+    const uniqueTargets: string[] = []
+    const duplicateTargets: string[] = []
+
+    for await (const {module, target} of extensionPoints) {
+      const fullPath = path.join(directory, module)
+      const fileExists = await file.exists(fullPath)
+
+      if (!fileExists) {
+        const notFoundPath = output.token.path(path.join(directory, module))
+
+        errors.push(
+          output.content`Couldn't find ${notFoundPath}
+Please check the module path for ${target}`.value,
+        )
+      }
+
+      if (uniqueTargets.indexOf(target) === -1) {
+        uniqueTargets.push(target)
+      } else {
+        duplicateTargets.push(target)
+      }
+    }
+
+    if (duplicateTargets.length) {
+      errors.push(`Duplicate targets found: ${duplicateTargets.join(', ')}\nExtension point targets must be unique`)
+    }
+
+    if (errors.length) {
+      const tomlPath = path.join(directory, configurationFileNames.extension.ui)
+
+      errors.push(`Please check the configuration in ${tomlPath}`)
+
+      this.abortOrReport(`${errors.join('\n\n')}`, undefined, directory)
+    }
+  }
+
   async loadUIExtensions(
     extensionDirectories?: string[],
   ): Promise<{uiExtensions: UIExtension[]; usedCustomLayout: boolean}> {
@@ -247,22 +311,18 @@ class AppLoader {
         UIExtensionConfigurationSupportedSchema,
         configurationPath,
       )
+
       const configuration = {
         ...configurationSupported,
         type: mapUIExternalExtensionTypeToUIExtensionType(configurationSupported.type),
       }
 
-      const entrySourceFilePaths = await path.glob(path.join(directory, 'src', '*.+(ts|js|tsx|jsx)'))
+      let entrySourceFilePath: string | undefined
 
-      if (!entrySourceFilePaths[0]) {
-        // TODO: Previosuly this was abortOrReport, but aborting guarantees types safety
-        // In what circumstance would we want this process to continue if there are no src files?
-        // There being no src files seems like a terminal problem.
-        throw new error.Abort(
-          output.content`Couldn't find any js, jsx, ts or tsx files in the directories ${output.token.path(
-            directory,
-          )} or ${output.token.path(path.join(directory, 'src'))}`,
-        )
+      if (await isLegacyUIExtension(configuration)) {
+        entrySourceFilePath = await this.getLegacyEntrySourceFilePath(directory)
+      } else {
+        await this.validateUIExtensionPointConfig(directory, configuration.extensionPoints as NewExtensionPointType)
       }
 
       return {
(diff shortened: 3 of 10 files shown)


packages/app/src/cli/constants.ts
```diff
@@ -144,6 +149,8 @@ export function getUIExtensionRendererDependency(extensionType: UIExtensionTypes
   switch (extensionType) {
     case 'product_subscription':
       return {name: '@shopify/admin-ui-extensions-react', version: '^1.0.1'}
+    case 'checkout_ui_extension_beta':
+      return {name: '@shopify/checkout-ui-extensions-react', version: '^0.20.0'}
```
byrichardpowell, line 153: Note that the new checkout extension differs from existing extension in that it has two rendering packages:

1. "@shopify/admin-ui-extensions"
2. "@shopify/checkout-ui-extensions"

To support this we'd need to update some code in the CLI to be able to return two render dependencies, e.g:

```
export function getUIExtensionRendererDependency(extensionType: UIExtensionTypes): DependencyVersion | undefined {
  switch (extensionType) {
    case 'checkout_ui_extension_beta':
      return {name: '@shopify/checkout-ui-extensions-react', version: '^0.20.0'}
    // ...etc
  }
}
```

Becomes
```
export function getUIExtensionRendererDependency(extensionType: UIExtensionTypes): DependencyVersion | undefined {
  switch (extensionType) {
    case 'checkout_ui_extension_beta':
      return [
        {name: '@shopify/checkout-ui-extensions-react', version: '^0.20.0'},
        {name: '@shopify/admin-ui-extensions-react', version: '^0.20.0'}
      ]
    // ...etc
  }
}
```

This becomes the dependencies we add to the apps package.json.

But! I _think_ both of these packages will become the same package... so can we just skip this for now?

byrichardpowell: > Umm, if the idea is that no extension should have more than 1 dependency (I guess a common @shopify/ui-extension is the goal) I think we shouldn't change this: You'll be forced to update the code in more files to support an array here and is affecting the CLI codebase for internal work 

Yeah, agreed.  We'll leave this as is for now.

> Would it make more sense to do this in a private fork? That way you could hardcode any change you need without worrying about the CLI codebase for the moment.

My instinct is I'd prefer to merge into main so we don't have to deal with a bunch of merge conflicts later. WDYT?

isaacroldan: In theory we are going to start refactoring this next week, so instead of merging back and having conflicts we could just add support for the new extension type after the refactor. (while you work in your fork with this version)

I'm saying this in case you need to do more changes here, because then we will have conflicts during the refactor, but if you don't plan to do any other changes to this, then is ok merge it

packages/app/src/cli/constants.ts
```diff
@@ -80,7 +80,12 @@ export const publicUIExtensions = {
 } as const
 
 export const uiExtensions = {
-  types: [...publicUIExtensions.types, 'pos_ui_extension', 'customer_accounts_ui_extension'],
+  types: [
+    ...publicUIExtensions.types,
+    'pos_ui_extension',
+    'customer_accounts_ui_extension',
+    'checkout_ui_extension_beta',
```
vividviolet, line 87: Why are we calling this `checkout_ui_extension_beta`? Should this not be `ui_extension`?

isaacroldan: Agree with this 👍

packages/app/src/cli/constants.ts
byrichardpowell: Ok, I'll use this as a feature branch then.  Mostly I'll use the feature branch to learn, and to have some code that we can copy to the new architecture.

This means that this PR will remain open and I'll merge into this branch.  Not ideal, but it's just unfortunate timing with the refactor... which is well worth it!

byrichardpowell:
> I haven't tested this but from your screenshot it looks like the file extension is missing in the generated template?

Yeah, this is now fixed.

packages/app/src/cli/constants.ts
```diff
@@ -158,7 +160,15 @@ export function getUIExtensionRendererDependency(extensionType: UIExtensionTypes
 }
 
 export const uiExternalExtensionTypes = {
-  types: ['web_pixel', 'post_purchase_ui', 'checkout_ui', 'pos_ui', 'subscription_ui', 'customer_accounts_ui'],
+  types: [
+    'web_pixel',
+    'post_purchase_ui',
+    'checkout_ui',
+    'checkout_ui_beta',
```
vividviolet, line 167: Should this be removed?

packages/app/src/cli/constants.ts
```diff
@@ -225,6 +236,7 @@ export const externalExtensionTypeNames = {
     'Post-purchase UI',
     'Theme app extension',
     'Checkout UI',
+    'Checkout UI Beta',
```
vividviolet, line 239: This should be UI Extension no?

packages/app/src/cli/services/deploy.ts
```diff
@@ -184,9 +177,23 @@ async function configFor(extension: UIExtension, app: AppInterface) {
     case 'pos_ui_extension':
     case 'product_subscription': {
       const result = await getUIExtensionRendererVersion(type, app)
-      if (result === 'not_found') throw RendererNotFoundBug(type)
+      if (result === 'not_found') {
```
vividviolet, line 180: Why do we need to change this existing code?

packages/app/src/cli/utilities/extensions/configuration.ts
```diff
@@ -106,6 +106,8 @@ export function getUIExtensionSurface(uiExtensionType: UIExtensionTypes) {
   switch (uiExtensionType) {
     case 'checkout_ui_extension':
       return 'checkout'
+    case 'ui_extension':
+      return 'checkout_beta'
```
vividviolet, line 110: Why does this get mapped to `checkout_beta`?

packages/app/templates/ui-extensions/projects/checkout_ui_beta/README.md.liquid
```diff
@@ -0,0 +1,89 @@
...
+
+## Getting Started
+Initially, your extension will have the following files:
+
+```
+└── my-app
+  └── extensions
+    └── my-checkout-ui-extension
+        ├── src
+        │   └── AdminCheckoutEditor.{{ srcFileExtension }} The source code for the ADMIN::CHECKOUT::EDITOR extension point
```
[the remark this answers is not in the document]
byrichardpowell: I admit I haven't given this much thought, happy to bike shed.  Fortunately we can use whatever we want here, but since this will form the basis of future UI extensions it would be good to establish some conventions.

Objectives:

* It should be very easy to see which file is responsible for which extension point.
* It shouldn't be too unconventional.
* It should scale well for the most common amount of extension pints (approximately)

I won't cover subtle differences like uppercase vs lowercase first letter since we can iterate on that once we have the more macro level direction.

### Option 1: Use Extension point name without `::`
E.g: 

* `src/AdminCheckoutEditorRenderSettings.ts`
* `src/CheckoutDynamicRender.ts`

Pros:

* Matches files to extension points quite explicitly
* Flat structure is simple enough for most extensions
* Has a (weak) parallel with React [component conventions](https://github.com/Shopify/hydrogen/tree/v1.x-2022-07/packages/hydrogen/src/components)

Cons:

* Does loosing `::` in the filename make the connection with the extension point less obvious?

### Option 2: Use exactly the extension point name
E.g: 

* `src/Admin::CheckoutEditor::RenderSettings.ts`
* `src/Checkout::Dynamic::Render.ts`

Pros:

* Very explicitly matches files to extension points
* Flat structure is simple enough for most extensions

Cons:

* Quite unconventional
* Would the `::` cause us problems? (I don't think so, but I'm nervous)
* If an extension has MANY extension points would the flat structure benefit from some nesting?

### Option 3: Use nested folders

* `src/Admin/CheckoutEditor/RenderSettings.ts`
* `src/Checkout/Dynamic/Render.ts`

Pros:

* If extensions scale to MANY extension points the nesting might be useful
* Matches the rails way of nesting (I think this was the original inspiration)

Cons:

* Would 3rd party developers make the connection between `::` and the folder structure? Maybe not?
* if most extensions are only a few extension points ins this overkill? I suspect so.

----------

Thoughts?

packages/app/src/cli/constants.ts
byrichardpowell: No I don't think so.

This is the name we present to users.  If we use UI Extension, how might we distinguish between options for other types of UI extensions? E.g: Post Purchase, Subscription or Customer accounts?  Yes, eventually these will all be a ui_extension under the hood, but we still need a way to group extension points into pre defined templates that roughly map to concepts like post purchase, subscription or customer accounts.  Which is what we are doing here.

Does that make sense? What do you think?

packages/app/src/cli/utilities/extensions/configuration.ts
byrichardpowell: Yeah, this is wrong.  I've updated this to `checkout`

In a follow on PR we'll have to reconsider this entirely since the new extension has multiple surfaces:

This affects:

1. Constructing the JSON store which is returned by the extensions dev server.
2. Constructing the redirect URL for an extension.
3. The extensionConfig we pass to go (But since this is going 💥 we don't need to worry about this).

It looks like all these points are handled in pre-exisitng tickets, so I'll resolve this.  Please re-open if I've missed anything.

packages/app/src/cli/services/deploy.ts
byrichardpowell: This was caught by a lint error that forbids factory functions for errors ([code here](https://github.com/Shopify/cli/blob/main/eslint-rules/no-error-factory-functions.cjs))

Because I edited the file eslint complained and I had to update the code, even though I hadn't touched it.

packages/app/src/cli/constants.ts
byrichardpowell: `checkout_ui_beta`? No I don't think we can remove this, since it's used in a few places in this file: https://github.com/Shopify/cli/blob/add-new-checkout-ui-template/packages/app/src/cli/utilities/extensions/name-mapper.ts

That assume the approach I've taken in this PR is generally sound though:

* `Checkout UI (Beta)` for the human friendly name.
* `checkout_ui_beta` to map the human friendly name to a template in `packages/app/templates/ui-extensions/projects/checkout_ui_beta`
* `ui_extension` for the extension specification type (used to default the type value for the toml schema).

packages/app/src/cli/utilities/extensions/name-mapper.ts
```diff
@@ -10,6 +10,8 @@ export function mapExtensionTypeToExternalExtensionType(extensionType: Extension
       return 'post_purchase_ui'
     case 'checkout_ui_extension':
       return 'checkout_ui'
+    case 'ui_extension':
+      return 'checkout_ui_beta'
```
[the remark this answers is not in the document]
byrichardpowell: I've removed some duplication by calling `mapUIExternalExtensionTypeToUIExtensionType` from within `mapExternalExtensionTypeToExtensionType`, e.g:

```
export function mapExternalExtensionTypeToExtensionType(
  externalExtensionType: ExternalExtensionTypes | ExtensionTypes,
): ExtensionTypes {
  switch (externalExtensionType) {
    case 'post_purchase_ui':
    case 'checkout_ui':
    case 'checkout_ui_beta':
    case 'subscription_ui':
    case 'web_pixel':
    case 'pos_ui':
    case 'customer_accounts_ui':
      return mapUIExternalExtensionTypeToUIExtensionType(externalExtensionType)
    // etc
```

I did try to update `mapExternalExtensionTypeToExtensionType` so that typescript understood when it would return a UI extension vs an extension. My thinking here was that then you wouldn't even need `mapUIExternalExtensionTypeToUIExtensionType`, you could just use `mapExternalExtensionTypeToExtensionType`, but I didn't have any luck getting typescript to understand.  Would love any pointers if you have some.

These name-mapper functions are used to map the `--type` flag for the `yarn generate extension` command to the specification type that we use. E.g:

Screen Shot 2022-11-02 at 10 16 26 AM

I suppose that is because the options we output when you don't specify a flag aren't really ideal to be passed as a flag:

Screen Shot 2022-11-02 at 10 22 10 AM

packages/app/src/cli/constants.ts
```diff
@@ -144,6 +144,8 @@ export function getUIExtensionRendererDependency(extensionType: UIExtensionTypes
   switch (extensionType) {
     case 'product_subscription':
       return {name: '@shopify/admin-ui-extensions-react', version: '^1.0.1'}
+    case 'ui_extension':
+      return {name: '@shopify/checkout-ui-extensions-react', version: '^0.20.0'}
```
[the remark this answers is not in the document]
byrichardpowell: I really don't have any context on our extension UI packages, but I would guess:

* The UI Components for admin that the checkout editor needs.
* The UI Components for checkout. 
* render/extend functions
* Any internationalization api's we need.

But I don't know anything about what's happening in that regard.  @vividviolet do you have any context?

packages/app/src/cli/constants.ts
vividviolet: I'm still confused as to why this needs to be exposed in `uiExternalExtensionTypes`. We only need to care about what template they want to use during the generate step. Can we just simplify for now by renaming `checkout_ui_beta` to `ui_extension`? Once the plugins system is built we can improve this to display an additional prompt for selecting template instead of hardcoding it to the checkout template.

packages/app/src/cli/constants.ts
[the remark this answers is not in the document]
lemonmade: Yeah, we still need a React package, good point. I will try and get us to a place relatively soon where we could move over to the generic packages, no worries about it for this PR!

packages/app/src/cli/utilities/extensions/name-mapper.ts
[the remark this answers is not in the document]
lemonmade: Thanks for the context!

packages/app/templates/ui-extensions/projects/checkout_ui_beta/README.md.liquid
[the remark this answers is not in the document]
lemonmade: I think I like option 1 the best, personally. The names can get kind of long, but it is explicit and we can automate it, which is nice for moving fast. What do you think?

packages/app/templates/ui-extensions/projects/checkout_ui_beta/shopify.ui.extension.toml.liquid
```diff
@@ -0,0 +1,29 @@
...
+
+# [[extension_points.metafields]]
+# namespace = "my-namespace"
+# key = "my-key"
+
+[[extension_points]]
+target = "Admin::Checkout::Editor"
+module = "./src/AdminCheckoutEditor.{{ srcFileExtension }}"
+
+# [[extension_points.metafields]]
```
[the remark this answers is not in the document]
lemonmade: Yes, it should stay for the checkout extension point. Thank you!

packages/app/templates/ui-extensions/projects/checkout_ui_beta/README.md.liquid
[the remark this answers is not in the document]
byrichardpowell: Yeah, option 1 is my favourite too.

One question:  Is it `src/AdminCheckoutEditorRenderSettings.ts` (which I prefer) or `src/adminCheckoutEditorRenderSettings.ts` (lowercase a).

packages/app/templates/ui-extensions/projects/checkout_ui_beta/shopify.ui.extension.toml.liquid
```diff
@@ -0,0 +1,25 @@
+name = "{{ name }}"
+
+[capabilities]
+network_access = true
+block_progress = true
+
+[[extension_points]]
+target = "Checkout::Dynamic::Render"
+module = "./src/checkoutDynamicRender.{{ srcFileExtension }}"
```
lemonmade, line 9: On the "final nitpicking" front: assuming we like the "file name is extension point without `::`s", I prefer pascal case. It matches the extension point, and for React consumers I think the pascal case name is pretty conventional.

packages/app/templates/ui-extensions/projects/checkout_ui_beta/src/adminCheckoutEditorRenderSettings.liquid
```diff
@@ -0,0 +1,34 @@
...
+
+extend("Admin::CheckoutEditor::RenderSettings", (root, { extensionPoint }) => {
+  root.appendChild(
+    root.createComponent(
+      Text,
+      {},
+      `Welcome to the ${extensionPoint} extension`
+    )
+  );
+  root.mount();
```
lemonmade, line 32: `root.mount()` can be removed in both of these examples

packages/app/src/cli/constants.ts
vividviolet: Discussed offline - we can leave this for now since we made the effort to get this to work and no good reason to undo it as the underlying type is still `ui_extension`

byrichardpowell:
Closing as we've been shipping this code ontop of Isaac's refactor.

PR #681, opened 2022-10-27
Fix dependency versions

pepicrft:
Releasing breaking changes in minor updates can easily happen because flagging a change as breaking or not is a human decision that's hard to automate. This is not an issue in development or when deploying the app, because the presence of the lockfile ensures that the same version of dependencies are resolved across environments, but it can be a problem in a CLI because dependencies are resolved at installation time with no lockfile. If that happens users end up with a CLI that blows up at runtime.

[Here](https://github.com/Shopify/cli/pull/666) is an example of a minor update introducing breaking changes.

I'm fixing the version of the dependencies to reduce the chances of this happening. Note that this doesn't eliminate the non-determinism of resolving the dependencies completely because the change only affects the direct dependencies and not the transitive ones. Still, it's better than having looser version requirements. 

I've added a check that runs on CI for every commit and ensures we don't introduce regressions here:
image

**Note** that one caveat of being more strict is that the `node_modules` resolved on the user side might contain duplicated copies of the same dependency (different versions), but that's better than having a non-working CLI.

### How to test your changes?
CI should pass.

### Measuring impact

### Checklist


pepicrft:
> @pepicrft can we dedupe dependencies post install?

I don't think I follow 😬. Do you mind expanding a bit?

PR #682, opened 2022-10-27
Split up the `haiku` module and remove the export from `@shopify/cli-kit`

pepicrft:
Having a public module `index.ts` in `@shopify/cli-kit` that centralizes all the exports proved to be a bad idea for performance since all the modules need to be loaded even if only one is being used. We agreed that we should use sub-modules instead, so this is continuation of that work. I'm taking the opportunity to revisit the APIs to make them more idiomatic, add documentation, and ensure the functions are placed in the right modules. I'm doing it because a great API `@shopify/cli-kit` is tightly connected with a great experience building plugins. 

- Remove the `haiku` export form the `index.ts` of `@shopify/cli-kit`.
- Create a module `@shopify/cli-kit/node/fs` moving the utility function from `haiku` to generate a random directory name.
- Create a module `@shopify/cli-kit/common/string` moving the utility to generate a random name.

### How to test your changes?
You should be able to create a new app and get a random name suggested in the process.

### Measuring impact

### Checklist


configurations/tsconfig.json
```diff
@@ -23,9 +23,11 @@
         "@shopify/cli-kit": ["../packages/cli-kit/src/index.ts"],
         "@shopify/cli-kit/common/array": ["../packages/cli-kit/src/public/common/array.ts"],
         "@shopify/cli-kit/common/result": ["../packages/cli-kit/src/public/common/result.ts"],
-        "@shopify/cli-kit/node/ui": ["../packages/cli-kit/src/public/node/ui.tsx"],
-        "@shopify/cli-kit/node/presets": ["../packages/cli-kit/src/public/node/presets.ts"],
+        "@shopify/cli-kit/common/string": ["../packages/cli-kit/src/public/common/string.ts"],
```
Arkham, line 26: hey! the other day we noticed that if you add lines to this file but not to `package.json`, vscode will work fine, but the moment you try to import `@shopify/cli-kit/common/string` from another subpackage oclif will silently fail to load the module. is there a way to keep this file syncronized with the `package.json` exports? or at least add a test so we can catch this in CI?

pepicrft: Opening a PR shortly

pepicrft: https://github.com/Shopify/cli/pull/683/files

matteodepalo (approved):

PR #692, opened 2022-10-31
always use extension dev uuid for dev

faouzioudouh:
Closes https://github.com/Shopify/checkout-web/issues/15057

Extensions should be using a `dev-uuid` formatted id during development. This allows them to be ignored from reports. Currently after a local extension is published and activated it is using the registration id instead of this `dev-uuid`.

Use dev UUID for the dev extensions

### How to test your changes?
- Checkout branch
- Create app + extension
- Persist the extension 
- (Observe the extension id in monorail events - is should not be prefixed with `dev-`)
- Open the extension id dev mode from CLI
- (Observe the extension id in monorail events - should be prefixed with `dev-`)

### Measuring impact

### Checklist

### DIFF
--- a/packages/app/src/cli/services/dev.ts
+++ b/packages/app/src/cli/services/dev.ts
@@ -89,9 +89,6 @@ async function dev(options: DevOptions) {
     outputAppURL(storeFqdn, exposedUrl)
   }
 
-  // If we have a real UUID for an extension, use that instead of a random one
-  options.app.extensions.ui.forEach((ext) => (ext.devUUID = identifiers.extensions[ext.localIdentifier] ?? ext.devUUID))
-
   const backendOptions = {
     apiKey,
     backendPort,


vividviolet:
@isaacroldan yes, this is still required and would be a great addition. The use case hasn't changed - we still need a way to match the local extension with the published one in development so we can toggle between the live vs. local version.

Is there a way to use the current published version's uuid instead? The registration uuid is a good start but there's no way currently for us to match that up to a published extension on the front end

faouzioudouh:
@vividviolet so I'm going to decline this PR then, since we are planning to toggle extension between persisted and dev.

Regarding the version id/registration, [I just merged a PR](https://github.com/Shopify/checkout-web/issues/15056) to add registration id to extension data, so we should be able to use registration id to find the extension that needs to be replaced.

isaacroldan:
that's awesome, @vividviolet does that solve the issue with ids??

faouzioudouh:
@vividviolet  Indeed in this PR it's more about instrumentation events, but what I'm trying to highlight is that we load the registration id from graphql and it is now part of persisted extension data, so it can be used to do the match up.

PR #699, opened 2022-11-01
[WIP] Add the ability to parse the new schema format

### LINKED ISSUE #685: CLI: Parse the new toml file schema
- [x] CLI can parse the new toml format into memory
- [x] CLI passes the new extension object to the extensions dev server.
- [x] Avoid combining new schema with old schema, so that we don’t end up with a whole bunch of optional fields

byrichardpowell:

Fixes #685

We want to be able to parse the new toml format (See [this doc](https://docs.google.com/document/d/1rlv3DcgGZi4_Hf_Ekd14DnYWJ1rjsfZlMBVC80gFWjc/edit#bookmark=id.ri74g3odxziy))

Note, this implementation kinda bad. Because of all the different schema formats, bascially every field becomes optional, which reduces type safety and validation posibilities a lot.  But, I don't want to spend much time on this right now because:

1. This will be merged into a feature branch and then refactored with Isaac's new approach to initializing extensions
2. After this project we will migrate all ui extensions to this new format, so this we can simplify these schemas a lot and reclaim some type safety.

**1) Updates the zod schema** 
This is what the CLI uses to parse the toml file into memory.  This then becomes part of the apps configuration.

Note, this implementation kinda bad. Because of all the different schema formats, basically every field becomes optional, which reduces type safety and validation possibilities a lot.  But, I don't want to spend much time on this right now because:

1. This will be merged into a feature branch and then refactored with Isaac's new approach to initializing extensions
2. After this project we will migrate all ui extensions to this new format, so this we can simplify these schemas a lot and reclaim some type safety.

**2) Updates the App Loader to load the new format**

| Previously  |  Now |
|---|---|
|  A UI extension could only have one `/src/**` file | A UI extension can have multiple `src/**` files    |
|  The file in `/src/**` would always be called `index` | Files in `/src/**` can be called anything   |

To support these changes we need to update how we do validation:

| Previously  |  Now |
|---|---|
|  Throw or report an error if there is no `src/index.js|ts|jsx|tsx` file | Throw an error if there is no `src/**.js|ts|jsx|tsx` file    |

### How to test your changes?
Note that this is incomplete functionality.  The CLI can load the new format, but it will not be able to build it.  That will come in another PR.

To test this:

1. `cd fictures/app`
2. `yarn generate extension`

You will see an error that is related to esbuild, but the toml file should be loaded.

### Post-release steps
None, this will be refactored after Isaac's changes.

### Measuring impact

### Checklist

### DIFF
--- a/packages/app/src/cli/services/dev/extension/payload/models.ts
+++ b/packages/app/src/cli/services/dev/extension/payload/models.ts
@@ -43,7 +43,7 @@ export interface UIExtensionPayload {
     status: ExtensionAssetBuildStatus
     localizationStatus: ExtensionAssetBuildStatus
   }
-  extensionPoints: string[] | null
+  extensionPoints: string[] | null | {metafields?: {namespace: string; key: string}[]; target: string; module: string}[]
   localization: Localization | null
   categories: string[] | null
   metafields?: {namespace: string; key: string}[] | null
--- a/packages/app/src/cli/models/app/loader.ts
+++ b/packages/app/src/cli/models/app/loader.ts
@@ -252,22 +252,16 @@ class AppLoader {
         type: mapUIExternalExtensionTypeToUIExtensionType(configurationSupported.type),
       }
 
-      const entrySourceFilePath = (
-        await Promise.all(
-          ['index']
-            .flatMap((name) => [`${name}.js`, `${name}.jsx`, `${name}.ts`, `${name}.tsx`])
-            .flatMap((fileName) => [`src/${fileName}`, `${fileName}`])
-            .map((relativePath) => path.join(directory, relativePath))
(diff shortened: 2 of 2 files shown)


packages/app/src/cli/services/dev/extension/payload/models.ts
```diff
@@ -43,7 +43,7 @@ export interface UIExtensionPayload {
     status: ExtensionAssetBuildStatus
     localizationStatus: ExtensionAssetBuildStatus
   }
-  extensionPoints: string[] | null
+  extensionPoints: string[] | null | {metafields: {namespace: string; key: string}[]; target: string; module: string}[]
```
vividviolet, line 46: I think `metafields` needs to be optional

isaacroldan (approved):
Changes make sense, in the future each extension type will have its own schema, so optionality won't be a problem.

PR #704, opened 2022-11-01
Bundle multiple extension point files

byrichardpowell:

Fixes [#1899](https://github.com/Shopify/ui-extensions-private/issues/1899)

TLDR; Previously extensions had one `src/index.*` file.  Now there are multiple `src/*.*` files.  Here we update the dev command to build multiple `src/*.*` files into one `dist/main.js` file.

The commit messages go into a lot of detail on the approach I've taken, so please read those.

To summarise:

1. Update the dev command to bundle multiple `src/*.*` files into one `dist/main.js` file.
2. Output a message when we build an extension.
3. Update how the app loads the paths to the src files.

But! The commit messages have a lot of detail, so please read those.

### How to test your changes?

Testing Dev:

1. `cd fixtures/app` 
2. `yarn generate extension` (Select the new **Checkout UI Beta** extension under **Shopify private**)
3. `yarn dev`
4. Make a valid change to `/extensions/[name]/src/AdminCheckoutEditor.*`.
  - See a message in the console 
  - Check `/extensions/[name]/dist/` to see your changes
5. Make a valid change to `/extensions/[name]/src/CheckoutDynamicRender.*`.
  - See a message in the console 
  - Check `/extensions/[name]/dist/` to see your changes
6. Make an invalid change to `/extensions/[name]/src/*.*`.
  - See a message in the console 
7. Make a change to an existing extension
  - Check that your changes are built
  - Open the preview link for an existing extension point.  Check that it renders.

Testing Build:

1. `cd fixtures/app` 
2. `yarn generate extension` (Select the new **Checkout UI Beta** extension under **Shopify private**)
3. `yarn build`

The new extension should be built: 

Screenshot 2022-11-16 at 10 36 37 AM

Testing validation with `yarn dev` && `yarn build`:

1. `cd fixtures/app` 
2. `yarn generate extension` (Select the new **Checkout UI Beta** extension under **Shopify private**)
3. Add a new extension point for the newly created toml file whose src file does not exist.  Run `yarn dev` or `yarn build`, you should see an error
4. In the toml file duplicate an extension point that already exists.  Run `yarn dev` or `yarn build`, you should see a different message.
5. Try different combinations of errors (module path is incorrect & duplicate extension points).  The error message should always make sense

Screenshot 2022-11-16 at 10 32 33 AM

**Note this isn't working right now, I'm investigating**

1. `cd fixtures/app` 
2. `yarn generate extension` (Select the new **Checkout UI Beta** extension under **Shopify private**)
3. `yarn shopify app info`.  You should see error messages for the newly created extension

### Post-release steps
None

### Measuring impact

### Checklist

### DIFF
--- a/packages/app/src/cli/models/app/loader.ts
+++ b/packages/app/src/cli/models/app/loader.ts
@@ -8,7 +8,6 @@ import {
   UIExtensionConfigurationSupportedSchema,
   Extension,
   OldExtensionPointsSchema,
-  NewExtensionPointsSchema,
 } from './extensions.js'
 import {AppConfigurationSchema, Web, WebConfigurationSchema, App, AppInterface, WebType} from './app.js'
 import {configurationFileNames, dotEnvFileNames, extensionGraphqlId} from '../../constants.js'
@@ -23,6 +22,7 @@ import {
   usesWorkspaces as appUsesWorkspaces,
 } from '@shopify/cli-kit/node/node-package-manager'
 import {resolveFramework} from '@shopify/cli-kit/node/framework'
+import type {NewExtensionPointType} from './extensions.js'
 
 const defaultExtensionDirectory = 'extensions/*'
 
@@ -235,6 +235,68 @@ class AppLoader {
     return parseResult.data
   }
 
+  async getLegacyEntrySourceFilePath(directory: string) {
+    const entrySourceFilePath = (
+      await Promise.all(
+        ['index']
+          .flatMap((name) => [`${name}.js`, `${name}.jsx`, `${name}.ts`, `${name}.tsx`])
+          .flatMap((fileName) => [`src/${fileName}`, `${fileName}`])
+          .map((relativePath) => path.join(directory, relativePath))
+          .map(async (sourcePath) => ((await file.exists(sourcePath)) ? sourcePath : undefined)),
+      )
+    ).find((sourcePath) => sourcePath !== undefined)
+
+    if (!entrySourceFilePath) {
+      this.abortOrReport(
+        output.content`Couldn't find an index.{js,jsx,ts,tsx} file in the directories ${output.token.path(
+          directory,
+        )} or ${output.token.path(path.join(directory, 'src'))}`,
+        undefined,
+        directory,
+      )
+    }
+
+    return entrySourceFilePath
+  }
+
+  async validateUIExtensionPointConfig(directory: string, extensionPoints: NewExtensionPointType) {
+    const errors: string[] = []
+    const uniqueTargets: string[] = []
+    const duplicateTargets: string[] = []
+
+    for await (const {module, target} of extensionPoints) {
+      const fullPath = path.join(directory, module)
+      const fileExists = await file.exists(fullPath)
+
+      if (!fileExists) {
+        const notFoundPath = output.token.path(path.join(directory, module))
+
+        errors.push(
+          output.content`Couldn't find ${notFoundPath}
+Please check the module path for ${target}`.value,
+        )
+      }
+
+      if (uniqueTargets.indexOf(target) === -1) {
+        uniqueTargets.push(target)
+      } else {
+        duplicateTargets.push(target)
+      }
+    }
+
+    if (duplicateTargets.length) {
+      errors.push(`Duplicate targets found: ${duplicateTargets.join(', ')}\nExtension point targets must be unique`)
+    }
+
+    if (errors.length) {
+      const tomlPath = path.join(directory, configurationFileNames.extension.ui)
+
+      errors.push(`Please check the configuration in ${tomlPath}`)
+
+      this.abortOrReport(`${errors.join('\n\n')}`, undefined, directory)
+    }
+  }
+
   async loadUIExtensions(
     extensionDirectories?: string[],
   ): Promise<{uiExtensions: UIExtension[]; usedCustomLayout: boolean}> {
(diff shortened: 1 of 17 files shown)


packages/app/src/cli/services/dev/extension/bundler.ts
```diff
@@ -24,11 +24,19 @@ export async function setupBundlerAndFileWatcher(options: FileWatcherOptions) {
   const bundlers: Promise<void>[] = []
 
   options.devOptions.extensions.forEach((extension) => {
+    const stdinContent = extension.entrySourceFilePaths
```
vividviolet, line 27: I think we might have a problem making this work with the automatically calling `shopify.extend` for UI extension as we won't know when we can simply call `import './${path.relative(extension.directory, filePath)}';` vs. needing to do 
```ts
import * as ExtensionModule1 from './${path.relative(extension.directory, filePath)}';`;

if ('default' in ExtensionModule1) {
  // Inject the code to register the extension point
  shopify.extend(
    extension.target, 
    ExtensionModule1,
} else {
  // With no default export, we assume they used the global API directly
}
```

vividviolet: I think we just a have a conditional here to generate the stinConten only if it's a `ui_extension`

```ts
options.devOptions.extensions.forEach((extension) => {
    const stdin = extension.type === 'ui_extension' ? getUIExtensionIndex(extension) : undefined;

    bundlers.push(
      bundleExtension({
        minify: false,
        outputBundlePath: extension.outputBundlePath,
        sourceFilePaths: stdinContent ? undefined : extension.entrySourceFilePath,,
        stdin,
        ...
      })
   )
 });
```

vividviolet (commented):
Is there a way we can support ui_extension but minimize impact on existing extensions? The current approach changes the `entrySourceFilePaths` to an array but I think it's necessary. 

We can encapsulate the logic custom logic when setting the stdin instead and only construct the stdin content if it's a `ui_extension`. I think this will be more inline with the future plugins system as they custom behaviours of extensions based on the type.

packages/app/src/cli/services/dev/extension/bundler.ts
byrichardpowell: In a previous version I did something similar to this, but I thought it better to avoid the conditional logic since all extension points can be built using stdin.

But, I think you're right.  Will revert.

packages/app/src/cli/services/build/extension.ts
```diff
@@ -83,7 +83,7 @@ export async function buildUIExtension(extension: UIExtension, options: Extensio
   await bundleExtension({
     minify: true,
     outputBundlePath: extension.outputBundlePath,
-    sourceFilePath: extension.entrySourceFilePath,
+    sourceFilePath: extension.entrySourceFilePaths[0] as string,
```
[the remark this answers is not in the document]
byrichardpowell: Resolving as this is no longer the case.  See https://github.com/Shopify/cli/pull/704/commits/b7162cc103fb13b3e104e427c24fd659b6306c75

packages/app/src/cli/models/app/loader.ts
```diff
@@ -247,30 +311,26 @@ class AppLoader {
...
-        // There being no src files seems like a terminal problem.
-        throw new error.Abort(
-          output.content`Couldn't find any js, jsx, ts or tsx files in the directories ${output.token.path(
-            directory,
-          )} or ${output.token.path(path.join(directory, 'src'))}`,
-        )
+      if (await isLegacyUIExtension(configuration)) {
+        entrySourceFilePath = await this.getLegacyEntrySourceFilePath(directory)
+      } else {
+        await this.validateUIExtensionPointConfig(directory, configuration.extensionPoints as NewExtensionPointType)
```
byrichardpowell, line 325: I spent a bit of time trying to avoid as here. In the end I decided it was not worth the effort since:

The plan is this code will never get merged into master.
After Isaac's refactor, and migrating existing extensions to the ui_extension type, this should be much simpler

packages/app/src/cli/services/extensions/bundle.ts
```diff
@@ -142,3 +144,16 @@ function isGraphqlPackageAvailable(): boolean {
...
+
+// This will probably be on the extension instance after Isaac's refactor
+export function getBundleExtensionStdIn({configuration, directory, entrySourceFilePath}: UIExtension) {
+  if (configuration.type === 'ui_extension') {
+    const extensionPoints = configuration.extensionPoints as NewExtensionPointType
+
+    return extensionPoints.map(({module}) => `import '${module}';`).join('\n')
+  }
+
+  const relativeImportPath = (entrySourceFilePath as string).replace(directory, '')
```
byrichardpowell, line 156: Again, I didn't invest in type safety here since:

1. The plan is this code will never get merged into master.
2. After Isaac's refactor, and migrating existing extensions to the `ui_extension` type, this should be much simpler

byrichardpowell:
@isaacroldan in this PR I add validation logic for the new UI extension type.  When the extension config is invalid there are no error messages for it, but this looks like it's a pre existing issue.

In this screenshot there is no src file (which is an error), but the app info message didn't show an error:

Screenshot 2022-11-16 at 11 07 54 AM

vividviolet (approved):
Tophat'd - looks great! 👍

isaacroldan:
@byrichardpowell Yes, i think the current implementation only shows some errors, and only if the extension was loaded. This is something to improve with the loader refactor once we tackle that

PR #712, opened 2022-11-02
Cache latest version

amcaplan:

We make a call to npm each CLI run when checking for versions. This is wasteful and time-consuming.

Makes it faster! We still pay full price on the initial run, but afterwards will save time and frustration by caching the result.

This is a generalizable system for any similarly expensive operation that doesn't need to be repeated more than once daily (or another amount of time of our choice).

### How to test your changes?

`dev shopify version`

It should still work, and get faster!

### Measuring impact

### Checklist

### DIFF
--- a/packages/cli-kit/src/store.ts
+++ b/packages/cli-kit/src/store.ts
@@ -18,6 +18,7 @@ interface ConfSchema {
   appInfo: CachedAppInfo[]
   themeStore: string
   session: string
+  cache: {[key: string]: {value: unknown; updatedAt: string}}
 }
 
 const schema = {
@@ -108,6 +109,15 @@ export async function clearAllAppInfo(): Promise<void> {
   store.clearAllAppInfo()
 }
 
+export async function cache<TValue>(
+  key: string,
+  source: () => Promise<TValue>,
+  expirySeconds = 24 * 3600,
+): Promise<TValue> {
+  const store = await cliKitStore()
+  return store.cache(key, source, expirySeconds)
+}
+
 export class CLIKitStore extends Conf<ConfSchema> {
   getAppInfo(directory: string): CachedAppInfo | undefined {
     debug(content`Reading cached app information for directory ${token.path(directory)}...`)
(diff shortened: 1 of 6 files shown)


matteodepalo (approved):
Nice! 💯

packages/cli-kit/src/version.ts
```diff
@@ -9,9 +10,10 @@ import latestVersion from 'latest-version'
  * @param name - The name of the NPM package.
  * @returns A promise to get the latest available version of a package.
  */
-export async function latestNpmPackageVersion(name: string) {
+export async function latestNpmPackageVersion(name: string, options = {force: false}) {
   debug(content`Getting the latest version of NPM package: ${token.raw(name)}`)
-  return latestVersion(name)
+  if (options.force) return latestVersion(name)
```
shauns, line 15: If I force load this, shouldn't the cache be updated?

packages/cli-kit/src/store.ts
```diff
@@ -18,6 +18,7 @@ interface ConfSchema {
   appInfo: CachedAppInfo[]
   themeStore: string
   session: string
+  cache: {[key: string]: {value: unknown; updatedAt: string}}
```
shauns, line 21: Shouldn't `cache` be in the `schema` below?

packages/cli-kit/src/version.ts
```diff
@@ -9,9 +10,10 @@ import latestVersion from 'latest-version'
...
+  return cache(`latest-npm-package-version-${name}`, async () => latestVersion(name))
```
shauns, line 16: I think it we're doing this it needs to be clearer that `cache` is a system-wide thing, not local to a project or in-memory on a process.

packages/cli-kit/src/store.ts
```diff
@@ -183,4 +193,18 @@ export class CLIKitStore extends Conf<ConfSchema> {
     debug(content`Removing session store...`)
     this.set('sessionStore', '')
   }
+
+  async cache<TValue>(key: string, source: () => Promise<TValue>, expirySeconds = 24 * 3600): Promise<TValue> {
+    const expiryMilliseconds = expirySeconds * 1000
+    const cache = this.get('cache') || {}
+    if (cache[key]) {
```
shauns, line 200: if you pull `cache[key]` into a variable before hitting the `if`, TS's type narrowing should mean you need a lot fewer `!`.

amcaplan:
I think this PR is hopelessly out of date with how CLI-kit is organized. Closing, hoping to reopen a better version.

PR #723, opened 2022-11-04
Polish log truncation

alvaro-shopify:

Fixes https://github.com/Shopify/shopify-cli-planning/issues/447

- Removed the `init` hook that initiates the `logger` and moved that initiation to the `BaseCommand`. This way that process is run only once a command is executed instead of running it for each plugin that includes the `cli-kit`
- Added some logic to apply the log file truncation process only for the last 30MB (fixed) to truncate the file to a maximum of 5MB (fixed). Truncating huge files (5GB) takes now few seconds instead of several minutes
- Applied union strategy for array fields when using the method `deep merging` with two objects
  - Old strategy: ```deepMerge({foo: ['bar']}, {foo: ['bar', 'none']}) ``` => ```{foo: ['bar', 'bar', 'none']}```
  - New strategy:  ```deepMerge({foo: ['bar']}, {foo: ['bar', 'none']}) ``` => ```{foo: ['bar', 'none']}```    
- Removed older go extension module path from the changeset configuration

### How to test your changes?
- Unzip this file [shopify.cli.log.zip](https://github.com/Shopify/cli/files/9940999/shopify.cli.log.zip) in ` ~/Library/Caches/shopify-cli-nodejs` and run any command. It should take no more than 10 seconds to truncate the huge log file and continue with the command
- Run `yarn shopify app dev --verbose` using an app with extensions and each time  a websocket event is received `supportedLanguage` content should not grow

### Measuring impact

### Checklist

### DIFF
--- a/packages/cli-kit/src/log.ts
+++ b/packages/cli-kit/src/log.ts
@@ -17,18 +17,42 @@ import {EOL} from 'node:os'
 
 const logFileName = 'shopify.cli.log'
 const maxLogFileSize = 5 * 1024 * 1024
+const maxLogFileSizeToTruncate = 30 * 1024 * 1024
 let logFileStream: WriteStream
 let commandUuid: string
 let logFilePath: string
 export class LinesTruncatorTransformer extends Transform {
   linesToRetain: string[] = []
   lastLineCompleted = true
+  contentSize = 0
 
-  constructor(readonly maxFileSize: number, opts?: TransformOptions) {
+  constructor(
+    readonly fileSize: number,
+    readonly maxFileSize: number = maxLogFileSize,
+    readonly maxFileSizeToTruncate: number = maxLogFileSizeToTruncate,
+    opts?: TransformOptions,
+  ) {
     super(opts)
   }
 
   _transform(chunk: unknown, encoding: BufferEncoding, callback: TransformCallback): void {
+    this.contentSize += (chunk as string).toString().length
+    if (this.shouldTruncate()) {
+      this.truncate(chunk)
+    }
+    callback()
+  }
+
+  _flush(callback: TransformCallback): void {
+    this.push(this.linesToRetain.join(EOL))
+    callback()
+  }
+
+  shouldTruncate(): boolean {
+    return this.fileSize - this.contentSize < this.maxFileSizeToTruncate
+  }
+
+  truncate(chunk: unknown) {
     const tokens = (chunk as string).toString().split(EOL)
 
     this.completeLastLine(tokens)
@@ -43,12 +67,6 @@ export class LinesTruncatorTransformer extends Transform {
     if (this.linesToRetain.length > numLinesToRetain) {
       this.linesToRetain = this.linesToRetain.splice(this.linesToRetain.length - numLinesToRetain)
     }
-    callback()
-  }
-
-  _flush(callback: TransformCallback): void {
-    this.push(this.linesToRetain.join(EOL))
-    callback()
   }
 
   // Lines retained length average is used so the number of lines depends on the length of them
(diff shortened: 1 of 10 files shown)


packages/cli-kit/src/log.ts
```diff
@@ -17,18 +17,42 @@ import {EOL} from 'node:os'
...
 let logFilePath: string
 export class LinesTruncatorTransformer extends Transform {
   linesToRetain: string[] = []
   lastLineCompleted = true
-
-  constructor(readonly maxFileSize: number, opts?: TransformOptions) {
+  contentSize = 0
+
+  constructor(
+    readonly fileSize: number,
```
amcaplan, line 30: Given there are several arguments, might be better to make these named parameters.

packages/cli-kit/src/log.ts
```diff
@@ -120,11 +138,12 @@ function getLogFilePath(options: {logDir?: string; override?: boolean} = {}) {
...
+  const fileSize = fileSizeSync(logFile)
+  if (fileSize < maxLogFileSize) {
     return
   }
   const tmpLogFile = logFile.concat('.tmp')
-  const truncateLines = new LinesTruncatorTransformer(maxLogFileSize)
+  const truncateLines = new LinesTruncatorTransformer(fileSize)
   const pipeline = promisify(Stream.pipeline)
   await pipeline(createReadStream(logFile), truncateLines, createWriteStream(tmpLogFile))
   await pipeline(createReadStream(tmpLogFile), createWriteStream(logFile))
```
amcaplan, line 149: In the interest of further performance improvement, it feels like this should be a copy/move instead of reading a file just to write it elsewhere.

If this was `await moveFile(tmpLogFile, logFile)` we could also skip the `unlinkSync` below.

packages/cli-kit/src/log.ts
```diff
@@ -120,11 +138,12 @@ function getLogFilePath(options: {logDir?: string; override?: boolean} = {}) {
 
 // Shaves off older log lines if logs are over maxLogFileSize long.
 async function truncateLogs(logFile: string) {
-  if (fileSizeSync(logFile) < maxLogFileSize) {
+  const fileSize = fileSizeSync(logFile)
```
amcaplan, line 141: If this is an async function, I believe it would be preferable to await an async filesize function.

In the past, this wasn't enclosed in an async function, so it made sense to do sync.

amcaplan (approved):
A few nitpicks, but mostly I'm OK with the direction, and the fix is very important even if not perfect!

I do have to say that as we invest more and more in logging, it feels like we should adopt a standard npm package to handle the complexity for us...

Also, 10 seconds seems like a lot, even for a big file. I wonder if as a follow-up we could figure out what's taking so long. I wonder if constantly checking the average line length is the culprit, or if it just takes this long to read the file and there's nothing we can do about it.

We could also consider reading the file backwards in some way, there are npm packages that claim to do just that.

But all this is for another iteration!

packages/cli-kit/src/log.ts
alvaro-shopify: The `tmpLogFile` is the result of the truncation process so its size should be 5MB at maximum. With that size I don't think that we have a better performance switching from this `pipeline` to a `move` operation, what do you think?

alvaro-shopify:
> I do have to say that as we invest more and more in logging, it feels like we should adopt a standard npm package to handle the complexity for us...

I'm agree with that. I have tested this [pacakge](https://www.npmjs.com/package/rotating-file-stream) in my first try and it works like a charm. The problem was that the rotation strategies always empties the current log file instead of retain the last 5MB of content.

> Also, 10 seconds seems like a lot, even for a big file. I wonder if as a follow-up we could figure out what's taking so long. I wonder if constantly checking the average line length is the culprit, or if it just takes this long to read the file and there's nothing we can do about it.
10 seconds is the time needed for truncation a 5GB log file. Maybe we can rise the `chunk` size used by the `pipeline` process so less iterations should be needed.

> We could also consider reading the file backwards in some way, there are npm packages that claim to do just that.
Good point!. Anyway, maybe a new issue could be created to achieve all of these suggestions in the future

pepicrft (approved):

PR #727, opened 2022-11-07
Extensions refactor: Register all supported extensions & functions

isaacroldan:
As part of the extensions refactor, this PR handles all extensions/functions registrations, defining all supported extensions in a cleaner way.

Fixes https://github.com/Shopify/shopify-cli-planning/issues/438

⚠️ To keep this PR from being too big, tests will be added in a new PR. We'll define service tests that handle the whole extension registration and loading.

- Define every extension/function using the new types, each one in a single file. That'd be the local source of truth for each extension.
- Dynamically import all extension/function definitions: After adding a new definition, no need to manually import it anywhere, we'll take care of it automatically.
- Some updates to the interfaces as new necessities are discovered (specially for functions where there isn't a remote spec)

### How to test your changes?
No need to test anything yet, this code is not used anywhere.

### Measuring impact

### Checklist

### DIFF
--- a/packages/app/src/cli/models/extensions/function-specifications/delivery_customization.ts
+++ b/packages/app/src/cli/models/extensions/function-specifications/delivery_customization.ts
@@ -1,0 +1,11 @@
+import {createFunctionSpec} from '../functions.js'
+
+const spec = createFunctionSpec({
+  identifier: 'delivery_customization',
+  externalType: 'delivery_customization',
+  externalName: 'Delivery customization',
+  public: false,
+  templatePath: (lang) => `checkout/${lang}/delivery-customization/defaultt`,
+})
+
+export default spec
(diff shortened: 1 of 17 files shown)


matteodepalo (approved):
LGTM!

packages/app/src/cli/models/extensions/function-specifications/delivery_customization.ts
```diff
@@ -0,0 +1,11 @@
+import {createFunctionSpec} from '../functions.js'
+
+const spec = createFunctionSpec({
+  identifier: 'delivery_customization',
+  externalType: 'delivery_customization',
+  externalName: 'Delivery customization',
+  public: false,
+  templatePath: (lang) => `checkout/${lang}/delivery-customization/defaultt`,
```
dphm, line 8: This typo is causing an empty folder to be generated for Delivery Customization function extensions

gonzaloriestra: Thanks! It will be fixed in https://github.com/Shopify/cli/pull/922

PR #734, opened 2022-11-08
Reduce the changesets concurrency

pepicrft:
Releases [fail](https://shipit.shopify.io/shopify/cli/production/deploys/1888745) because changesets attempts to publish packages that have already been published. The issue seems to be rooted in [flakiness](https://github.com/changesets/changesets/issues/427#issuecomment-668515080) that arises when changesets concurrently sends requests to the NPM API to obtain the packages information. There's a [PR](https://github.com/changesets/changesets/pull/853/files) open to reduce the concurrency but it hasn't been merged.

I'm "patching" the fix at dependency installation time by using [patch-package](https://www.npmjs.com/package/patch-package)

### How to test your changes?
It's tricky to test from a branch. I'd suggest merging and running an actual release.

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,7 +27,7 @@
   "devDependencies": {
     "@apollo/client": "^3.4.8",
     "@bugsnag/source-maps": "^2.3.0",
-    "@changesets/cli": "^2.22.0",
+    "@changesets/cli": "2.25.2",
     "@nrwl/eslint-plugin-nx": "^14.0.5",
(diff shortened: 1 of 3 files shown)


gonzaloriestra (approved):
LGTM as a workaround while changesets get fixed.

Because the patch will only be applied to @changesets/cli@2.25.2 and if we upgrade it, the patch will be ignored, right?

Issue #672, opened 2022-10-25
[Bug]: Preset does not work for themes


bakura10:
### Please confirm that you have:

- [X] Searched [existing issues](.) to see if your issue is a duplicate. (If you’ve found a duplicate issue, feel free to add additional information in a comment on it.)
- [X] Reproduced the issue in [the latest CLI version](https://www.npmjs.com/package/@shopify/cli).

### In which of these areas are you experiencing a problem?

Theme

### Expected behavior

* Created a "shopify.presets.toml" file at the root of my folder.
* Run the command `shopify theme dev --preset=dev`
* CLI properly extracts the variables, but fail to run the command:

<img width="477" alt="Capture d’écran 2022-10-25 à 21 39 36" src="https://user-images.githubusercontent.com/1198915/197866554-d0b86e6f-8735-47fb-a016-f0a6baf89a85.png">

### Actual behavior

* CLI should not show any error

### Reproduction steps

1.
2.
3.

### Operating System

OS X

### Shopify CLI version (check your project's `package.json` if you're not sure)

3.20

blanklob:
@bakura10 Hey I made an example that uses env variables in a preset file similar to the toml file you mentioned above, until the full implementation hit production and gets announced, you can use this: https://github.com/blanklob/shopify-cli-3-themes-demo

bakura10:
@blanklob , I tried yesterday and the preset approach now works, so we can get rid of the extra dependency of your demo ;)

PR #743, opened 2022-11-08
Allowlist relevant flags for CLI2 passthrough

amcaplan:

Fixes https://github.com/Shopify/cli/issues/672 and prevents further similar issues

We are passing through all CLI3 flags to CLI2, which means any new global flags need to be mentioned in theme commands.

Better to take an allowlist approach since the per-command flags won't change that much and will only need updates in 1 place.

### How to test your changes?

All the commands should still work!

### Measuring impact

### Checklist


packages/theme/src/cli/utilities/theme-command.ts
```diff
@@ -4,15 +4,15 @@ interface FlagValues {
   [key: string]: boolean | string | string[] | number | undefined
 }
 interface PassThroughFlagsOptions {
-  // Exclude flags that are only for CLI3 but will cause errors if passed to CLI2
-  exclude?: string[]
+  // Only pass on flags that are relevant to CLI2
+  relevantFlags?: string[]
```
gonzaloriestra, line 8: I think `allowedFlags` would be more accurate, but it's just a suggestion.

packages/theme/src/cli/utilities/theme-command.ts
```diff
@@ -4,15 +4,15 @@ interface FlagValues {
...
 }
 
 export default abstract class ThemeCommand extends Command {
-  passThroughFlags(flags: FlagValues, {exclude}: PassThroughFlagsOptions): string[] {
+  passThroughFlags(flags: FlagValues, {relevantFlags}: PassThroughFlagsOptions): string[] {
```
gonzaloriestra, line 12: What about adding the default value here?

```suggestion
  passThroughFlags(flags: FlagValues, {relevantFlags: Array = []}: PassThroughFlagsOptions): string[] {
```

packages/theme/src/cli/utilities/theme-command.ts
```diff
@@ -4,15 +4,15 @@ interface FlagValues {
...
     const passThroughFlags: string[] = []
     for (const [label, value] of Object.entries(flags)) {
-      if ((exclude ?? []).includes(label)) {
+      if (!(relevantFlags ?? []).includes(label)) {
```
gonzaloriestra, line 15:
```suggestion
      if (!relevantFlags.includes(label)) {
```

gonzaloriestra (approved):
Thanks for taking care of this, 👌

karreiro (approved):
Thank you, @amcaplan! 🚀

packages/theme/src/cli/commands/theme/dev.ts
```diff
@@ -64,13 +64,15 @@ export default class Dev extends ThemeCommand {
     }),
   }
 
+  static cli2Flags = ['host', 'live-reload', 'poll', 'theme-editor-sync', 'port', 'theme', 'only', 'ignore', 'stable']
```
Poitrin, line 67: As I merged https://github.com/Shopify/cli/pull/751 _before_, the `--force` flag got introduced for `theme dev/pull/push/share`.

I therefore added the necessary changes and created a follow-up PR: https://github.com/Shopify/cli/pull/764

You can either cherry-pick my changes onto this PR's branch, or review https://github.com/Shopify/cli/pull/764

Thanks!

packages/theme/src/cli/utilities/theme-command.ts
amcaplan: This doesn't work as cleanly with TypeScript, I believe we'd have to do:

```suggestion
  passThroughFlags(flags: FlagValues, {relevantFlags}: PassThroughFlagsOptions = {relevantFlags: []}): string[] {
```

Also note it's a default value for the whole object, so if `{}` is passed then `relevantFlags` would still be `undefined`.

packages/theme/src/cli/commands/theme/dev.ts
amcaplan: Thanks for the note, @Poitrin! I finally got around to 🎩 and added a few `force`s, so maybe just rebase/merge against main and make sure I haven't missed anything.

PR #762, opened 2022-11-10
Search for apps by title

### LINKED ISSUE #317: When using the CLI flow for an existing application, you can only select from the first five pages of apps in your partner dashboard

amcaplan:

Fixes #317

Currently Partners can't select an app that's too far down in their org's list of apps. They need to work around using the `--api-key` flag. This PR allows Partners to search by app title if their app count exceeds the number we fetch.

[Videobin](https://videobin.shopify.io/v/Pv11DW)

I think this is best explained by seeing the code comments, which describe the algorithm at length.

TL;DR if they have >100 apps in the organization, typing to search triggers a call to the Partners API searching for the apps by title.

### How to test your changes?

Acting as a 1P developer, connect to the Shopify partners organization and search for an app far down the list.

### Measuring impact

- [x] We stop getting issues and Slack messages complaining (quite reasonably!) about this issue

### Checklist

### DIFF
--- a/packages/app/src/cli/prompts/dev.ts
+++ b/packages/app/src/cli/prompts/dev.ts
@@ -19,7 +19,7 @@ export async function selectOrganizationPrompt(organizations: Organization[]): P
 }
 
 export async function selectAppPrompt(apps: MinimalOrganizationApp[], orgId: string, token: string): Promise<MinimalOrganizationApp> {
-  const appsByApiKey: {[apiKey: string]: MinimalOrganizationApp} = Object.fromEntries(apps.map((app) => [app.apiKey, app]))
+  const appsByApiKey: {[apiKey: string]: MinimalOrganizationApp} = {}
   const addToApiKeyCache = (apps: MinimalOrganizationApp[]) => apps.forEach((app) => appsByApiKey[app.apiKey] = app)
   addToApiKeyCache(apps)
   const toAnswer = (app: MinimalOrganizationApp) => ({name: app.title, value: app.apiKey})
@@ -40,6 +40,7 @@ export async function selectAppPrompt(apps: MinimalOrganizationApp[], orgId: str
    * action.
    */
   let cachedResults: {[input: string]: MinimalOrganizationApp[]} = {'': apps}
+  // eslint-disable-next-line @typescript-eslint/no-misused-promises
   const fetchInterval = setInterval(async () => {
     let input: string | undefined
     do {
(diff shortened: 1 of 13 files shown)


packages/app/src/cli/prompts/dev.ts
```diff
@@ -17,17 +18,86 @@ export async function selectOrganizationPrompt(organizations: Organization[]): P
...
+   * per second to avoid hitting rate limits.
+   * If we don't have any previously unseen input to search for, return without
+   * action.
+   */
+  const cachedResults: {[input: string]: OrganizationApp[]} = {'': apps}
+  // eslint-disable-next-line @typescript-eslint/no-misused-promises
+  const fetchInterval = setInterval(async () => {
+    let input: string | undefined
+    do {
+      input = allInputs.pop()
```
matteodepalo, line 48: Do we need to keep track of all inputs? I get that going back is going to be faster this way, but we're also going to hit the API many times, when most likely only the results returned by the last input will be used.
What if instead we performed a debounced search only when the user has stopped typing for like 500ms. This should cover the most common scenario and only perform one query.

I still like the idea of caching results so that if you search for something, add some characters and go back we don't fetch again, but I feel like doing extra queries just in case might be a bit wasteful here.

packages/app/src/cli/prompts/dev.ts
```diff
@@ -17,17 +18,86 @@ export async function selectOrganizationPrompt(organizations: Organization[]): P
...
+            cachedFiltered[input] = await filterFunction(cachedResults[input]!.map(toAnswer), input)
+          }
+
+          /* If the user types "hello" we will have this function running 5
+           * times, for each substring ("h", "he", "hel", "hell", "hello").
+           * If the latest input "hello" is available, display that result.
+           * Otherwise, display the result of whatever intermediate step is
+           * available, so the user at least sees something is happening.
+           */
+          return cachedFiltered[latestRequest] || cachedFiltered[input]!
```
matteodepalo, line 94: I wonder wether this behavior might be a bit surprising if compared with typical web autocomplete. Could we instead display `\`no results for ${input}\``, or something similar?

amcaplan: So, imagine that I type "hello" and we are loading results as we go.

If we just received results for `hell` we would show them until the results from `hello` return, and then even if the `hello` results are empty, we'd replace the `hell` results with a statement that no results were found.

So the pecking order is:

Results from the latest input, even if empty > The most recently returned results, if not empty > "Searching..."

This may matter much less anyway if we're using the debouncing strategy mentioned above.

matteodepalo: Got it, that makes sense 👍 And yes this will probably matter less if we go with a debouncing strategy.

amcaplan:
Sorry coverage report, this kinda thing is just going to make you sad. 😢

PR #765, opened 2022-11-11
Fetch extension specifications

matteodepalo:

Fixes [#0000 ](https://github.com/Shopify/shopify-cli-planning/issues/443)

This PR uses the extension specifications query from the partners API to fetch extension specifications instead of hard-cording them. Tests are still WIP. I've tried testing it manually but get this error

```
"message": "Field 'extensionSpecifications' doesn't exist on type 'QueryRoot'; Variable $api_key is declared by fetchSpecifications  │
│   but not used"
```

### Measuring impact

### Checklist


packages/app/src/cli/services/environment.ts
```diff
@@ -69,6 +59,58 @@ interface DevEnvironmentOutput {
...
+ */
+export async function ensureGenerateEnvironment(options: {
+  apiKey?: string
+  directory: string
+  reset: boolean
+  token: string
+}): Promise<string> {
+  if (options.apiKey) {
+    const app = await fetchAppFromApiKey(options.apiKey, options.token)
+    if (!app) throw InvalidApiKeyError(options.apiKey)
```
pepicrft, line 80: I'd have expected ESLint to fail this one. Let's abort from here and move away from the factory function pattern for creating errors.

packages/app/src/cli/services/environment.ts
```diff
@@ -69,6 +59,58 @@ interface DevEnvironmentOutput {
...
+    const selectedApp = await selectOrCreateApp(localAppName, apps, organization, options.token, cachedInfo?.appId)
+    await store.setAppInfo({
+      appId: selectedApp.apiKey,
+      title: selectedApp.title,
+      directory: options.directory,
+      orgId,
+    })
+    return selectedApp.apiKey
+  }
+}
```
pepicrft, line 112: This workflow seems very similar to the one that we follow for `dev`. Have you considered reusing it to prevent potential inconsistencies down the road?

packages/app/src/cli/commands/app/generate/extension.ts
```diff
@@ -17,16 +15,17 @@ import {AppInterface} from '../../../models/app/app.js'
...
+import {ensureGenerateEnvironment} from '../../../services/environment.js'
+import {fetchExtensionSpecifications} from '../../../utilities/extensions/fetch-extension-specifications.js'
+import {allFunctionSpecifications} from '../../../models/extensions/specifications.js'
+import {FunctionSpec} from '../../../models/extensions/functions.js'
+import {output, path, cli, error, environment, session} from '@shopify/cli-kit'
 import {Flags} from '@oclif/core'
 import {PackageManager} from '@shopify/cli-kit/node/node-package-manager'
+import {RemoteSpecification} from '@shopify/cli-kit/src/api/graphql/extension_specifications.js'
 
 export default class AppScaffoldExtension extends Command {
```
pepicrft, line 30: I'd recommend looking at commands as views in the MVC and having them as lean as possible. All they should do is declare themselves (flags, arguments, description), and delegate the business logic to a service function. Looking at the implementation of the function here, it's starting to have too much logic and it might be a good time to extract the functionality. Besides being able to write tests that don't have to instantiate an Oclif command class, we can also reuse the logic in other contexts.

packages/app/src/cli/commands/app/generate/extension.ts
```diff
@@ -74,26 +84,59 @@ export default class AppScaffoldExtension extends Command {
     const {flags} = await this.parse(AppScaffoldExtension)
 
     await metadata.addPublic(() => ({
-      cmd_scaffold_required_auth: false,
+      cmd_scaffold_required_auth: true,
       cmd_scaffold_template_custom: flags['clone-url'] !== undefined,
       cmd_scaffold_type_owner: '@shopify/app',
```
shauns, line 89: This is intended to be
> For now always @shopify / app -- all types registered by one plugin. But in the future custom plugins might register pre-release types

So I think I'd expect this to change -- if the scaffolded extension was registered by a different plug-in.

packages/app/src/cli/models/extensions/functions.ts
```diff
@@ -156,6 +160,9 @@ export function createFunctionSpec<
     configSchema: BaseFunctionConfigurationSchema,
     metadataSchema: BaseFunctionMetadataSchema,
     public: true,
+    options: {
+      registrationLimit: spec.registrationLimit ?? 10,
```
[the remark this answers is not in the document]
isaacroldan: The 10 was the limit proposed by the functions team as the default limit. I'll move it to the constants file 👌

packages/app/src/cli/services/environment.ts
isaacroldan: Yep, will change it

isaacroldan: I think the rule for non factory errors doesn't work in all cases 🤔

packages/app/src/cli/commands/app/generate/extension.ts
isaacroldan: You are totally right. In this case, all the logic was already in the command file, and to avoid a bigger PR and refactor, I just updated what I needed here.
But I have a list of "things to clean up" after the extensions work is done, including this file and the `loader.ts` for instance.

If it's ok with you, i'll do this refactor in a different PR with proper tests for the service (I feel it will be a BIG PR)

packages/app/src/cli/commands/app/generate/extension.ts
isaacroldan: Got it, we are still not loading extensions from plugins, but i'll take this into account when we do

PR #796, opened 2022-11-21
Remove end of line new lines from chunks

matteodepalo:

Fixes https://github.com/Shopify/shopify-cli-planning/issues/464

We used to have a function that cleaned just the ansi characters that moved the cursor before displaying concurrent output. I removed this function replacing it with `stripAnsi` which just strips every ansi character. This is so that we can have total control of the formatting, including color.

Part of this function was also a replacement of the end of line `\n` characters which I inadvertedly removed.

This PR puts that replacement back.

### How to test your changes?

- Run `yarn shopify app dev --path fixtures/app`
- Change something in the backend `index.js`
- There should be no line breaks between the nodemon logs

### Measuring impact

### Checklist

### DIFF
--- a/packages/cli-kit/src/output.ts
+++ b/packages/cli-kit/src/output.ts
@@ -303,20 +303,6 @@ export interface OutputProcess {
   action: (stdout: Writable, stderr: Writable, signal: AbortSignal) => Promise<void>
 }
 
-/**
- * This regex can be used to find the erase cursor Ansii characters
- * to strip them from the string.
- * https://gist.github.com/fnky/458719343aabd01cfb17a3a4f7296797#erase-functions
- */
-const eraseCursorAnsiRegex = [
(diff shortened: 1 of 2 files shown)


packages/cli-kit/src/output.ts
```diff
@@ -303,20 +303,6 @@ export interface OutputProcess {
   action: (stdout: Writable, stderr: Writable, signal: AbortSignal) => Promise<void>
 }
 
-/**
- * This regex can be used to find the erase cursor Ansii characters
- * to strip them from the string.
- * https://gist.github.com/fnky/458719343aabd01cfb17a3a4f7296797#erase-functions
- */
-const eraseCursorAnsiRegex = [
```
matteodepalo, line 311: I forgot to remove this in the last cleanup.

PR #798, opened 2022-11-22
Add single select and confirmation prompts

matteodepalo:

We want to rework our prompts designs which means that we need to move away from inquirer as it's not very customizable in terms of styles.

This PR implements the newly designed single select prompt which includes headers and a information table.
Every choice has a customizable hotkey that selects it (numbers by default).
You can use this new component by calling `renderPrompt`.
Screenshot 2022-12-02 at 10 48 46

It also adds a new `renderConfirmation` method which reuses the single select component, but is meant to be used to ask yes/no questions.
Screenshot 2022-12-02 at 10 49 05

### To-do

- [x] Need to figure out a way to test this
- [x] What happens when there are 10+ items? Pressing 1, 0 doesn't go to the 10th item.

### How to test your changes?

- Run `kitchen-sink` and check out the new components!

### Measuring impact

### Checklist


packages/cli-main/src/cli/services/kitchen-sink.ts
```diff
@@ -174,5 +182,27 @@ export async function kitchenSink() {
...
+      {label: 'third', value: 'third'},
+      {label: 'fourth', value: 'fourth'},
+      {label: 'fifth', value: 'fifth', group: 'Automations'},
+      {label: 'sixth', value: 'sixth', group: 'Automations'},
+      {label: 'seventh', value: 'seventh'},
+      {label: 'eighth', value: 'eighth', group: 'Merchant Admin'},
+      {label: 'ninth', value: 'ninth', group: 'Merchant Admin'},
+      {label: 'tenth', value: 'tenth'},
+    ],
+    infoTable: {Add: ['new-ext'], Remove: ['integrated-demand-ext', 'order-discount']},
```
pepicrft, line 199: What about `remove` instead of `Remove` to align with the JS conventions?

packages/cli-main/src/cli/services/kitchen-sink.ts
```diff
@@ -174,5 +182,27 @@ export async function kitchenSink() {
...
+    onChoose: () => {},
```
pepicrft, line 200: How is this intended to be used? Is it to subscribe to the selection of an element? If so, what about making it a return element?

```js
const selection = await renderPrompt(...)

// instead of
let selection
await renderPrompt({
  onChoose: (_selection) => _selection = selection
})
```

packages/cli-main/src/cli/services/kitchen-sink.ts
```diff
@@ -174,5 +182,27 @@ export async function kitchenSink() {
     },
   }
 
+  await renderPrompt({
```
pepicrft, line 185: Very idiomatic API 👏🏼

packages/cli-kit/src/private/node/ui/components/ConcurrentOutput.test.tsx
```diff
@@ -0,0 +1,60 @@
...
+      },
+    }
+    // When
+
+    const {lastFrame} = render(
+      <ConcurrentOutput processes={[backendProcess, frontendProcess]} abortController={new AbortController()} />,
+    )
+
+    // wait for all output to be rendered
+    await new Promise((resolve) => setTimeout(resolve, 100))
```
pepicrft, line 47: This one can become a source of flakiness fairly quickly. Doesn't Ink provide a callback to know when the render has happened?

packages/cli-kit/src/private/node/ui/components/Prompt.test.tsx
```diff
@@ -0,0 +1,97 @@
...
+    const renderInstance = render(
+      <Prompt
+        message="Associate your project with the org Castile Ventures?"
+        choices={items}
+        infoTable={infoTable}
+        onChoose={() => {}}
+      />,
+    )
+
+    expect(renderInstance.lastFrame()).toMatchInlineSnapshot(`
```
pepicrft, line 72: Why do we have a `sleep` in the above method (using a Promise) and not here?

packages/cli-kit/src/testing/ui.ts
```diff
@@ -1,22 +1,18 @@
-import * as path from '../path.js'
-import {ExecaChildProcess, execaNode} from 'execa'
+export function waitForInputsToBeReady() {
```
pepicrft, line 1: 😬 Is there are way to subscribe to the "readiness"? I'm afraid this will result in flakiness (e.g. when running the tests in environments where the system needs more time for the input to be ready).

packages/cli-kit/src/private/node/ui/components/SelectInput.tsx
```diff
@@ -0,0 +1,142 @@
...
+  const [inputTimeout, setInputTimeout] = useState<NodeJS.Timeout | null>(null)
+  const [selectedIndex, setSelectedIndex] = useState(0)
+  const keys = useRef(new Set(items.map((item) => item.key)))
+  const {exit: unmountInk} = useApp()
+  const groupedItems = groupItems(items)
+  const groupTitles = Object.keys(groupedItems)
+
+  const previousItems = useRef<Item<T>[]>(items)
+
+  useEffect(() => {
```
pepicrft, line 40: Have you considered increasing the readability of this component by extracting the hooks invocations into hooks with descriptive names?

```js
useSelectFirstItemHook(items)
useHandleInputCallback(items, (selectedIndex) => { })
// ...
```

pepicrft (approved):
It's amazing seeing all these components taking shape. Besides the comments that I left, I noticed component tests have a similar structure so we might consider providing some utilities to improve the testing ergonomics:

- Mount the component
- Wait for stdin to be ready
- Interact with the component
- Check the last frame

They other thing that I wonder while looking at the tests is whether we are using the process's stdin and stdout for the tests or we are providing test-scoped streams. If it's the former, I'd recommend revisiting the approach to isolate the tests as much as possible.

packages/cli-kit/src/private/node/ui/components/ConcurrentOutput.tsx
```diff
@@ -96,7 +96,7 @@ const ConcurrentOutput: FunctionComponent<Props> = ({processes, abortController,
       }),
     )
 
-    unmountInk()
+    if (process.env.CI !== 'true') unmountInk()
```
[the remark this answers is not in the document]
matteodepalo: There is a comment above the other line that does this that explains it. In a follow up PR I've refactored this into a single method and put the comment there. In essence, what happens is that on `unmount` Ink only outputs the last frame in CI environments and often that's empty. I've [commented on a relevant issue](https://github.com/vadimdemedes/ink/pull/266#issuecomment-1350768870) but unless we get a change in master I didn't want to go with a fork of Ink so for now I put the workaround in the code.

packages/cli-kit/src/private/node/ui/components/SelectInput.tsx
matteodepalo: Unless we're going to be reusing these hooks somewhere else I prefer to put comments above this if we need to explain what they do. It seems to me it's more in react's style to not create too many custom hooks and instead go for simple ones when possible.

PR #827, opened 2022-11-25
[Bug] [WIP] Fix importing graphql files

byrichardpowell:
The CLI is not adding the GraphQL plugin to esbuild.  In this PR I fix some code to ensure that the Graphql plugin gets added.

This bug was happening because:

1. require will not be available when `"type": "module"` is in the package.json
2. We wrap `require.resolve('graphql')` in a  try catch`.  This was supposed to catch when `graphql` and `graphql-tag` dependencies are not available, but because `require` is not available `isGraphqlPackageAvailable()` always returns false.
3. Therefore we never add the graphql plugin from `@luckycatfactory/esbuild-graphql-loader`

There are 2 commits:

1. First we introduce a graphql file in an extension fixture.  This is our reproduceable test case.
2. Then we add some code to ensure that require is available in the module format.  [This stackoverflow issue](https://stackoverflow.com/questions/31931614/require-is-not-defined-node-js) has more context.

### How to test your changes?

Checkout the first commit and run `yarn dev`, `yarn build` or `yarn deploy`.  The command should fail with output like this:

Screenshot 2022-11-25 at 4 55 27 PM

Now checkout the second commit and run `yarn dev`, `yarn build` or `yarn deploy`. The command should succeed.

### Post-release steps

Should this be added to the changelog? I'm not sure.

### Measuring impact

### Checklist


byrichardpowell:
@pepicrft @vividviolet @isaacroldan  I'm not 100% sure this is the correct solution.  Once I get a 👍  from the CLI team I'll add a unit test to prevent regressions.

JulienHe (approved):
We worked on it together saw that it was working.

dmhenry:
🎩

PR #833, opened 2022-11-28
Get types from packages/ui-extensions-server-kit instead of duplicating them

byrichardpowell:

Fixes #0000 

1. `UIExtension` in `packages/app` is swapped for `ExtensionPayload` from `/packages/ui-extensions-server-kit`
2. Update `ExtensionPayload` to include data that `packages/app` returns, but are not included in the type.  I assume the types fell out of sync a while back and we never caught it, which is a good indicator of the problem this commit is trying to solve

Note, we could also take the types for Server Events (dispatch, connected, update) from packages/ui-extensions-server-kit, but this can be a follow on commit

### How to test your changes?

1. Are tests, types & linting green?
2. `cd fixtures/app`
3. `yarn dev`
4. Open the extensions dev console
5. Using the extensions dev console UI, do preview links work?
6. Using the extensions dev console UI, does disabling an extension work?
7. Does `yarn build` work?
8. Does `yarn deploy work?

### Measuring impact

### Checklist


packages/app/src/cli/models/extensions/extension-specifications/post_purchase_extension.ts
```diff
@@ -8,7 +8,7 @@ const spec = createExtensionSpec({
   identifier: 'checkout_post_purchase',
   externalIdentifier: 'post_purchase_ui',
   externalName: 'Post-purchase UI',
-  surface: 'post_purchase',
+  surface: 'post-checkout',
```
byrichardpowell, line 11: `ui-extensions-server-kit` expects this value.

packages/app/src/cli/models/extensions/extension-specifications/customer_accounts_ui_extension.ts
```diff
@@ -14,7 +14,7 @@ const spec = createExtensionSpec({
   identifier: 'customer_accounts_ui_extension',
   externalIdentifier: 'customer_accounts_ui',
   externalName: 'Customer accounts UI',
-  surface: 'customer_accounts',
+  surface: 'customer-accounts',
```
byrichardpowell, line 17: `ui-extensions-server-kit` expects this value.

packages/app/src/cli/utilities/extensions/configuration.ts
```diff
@@ -8,13 +8,14 @@ interface GetUIExensionResourceURLOptions {
...
+): {url: string} {
+  if (uiExtensionType === 'checkout_ui_extension' && options.checkoutCartUrl) {
+    return {url: options.checkoutCartUrl}
   }
+
+  if (uiExtensionType === 'product_subscription') {
+    return {url: options.subscriptionProductUrl ?? ''}
+  }
+
+  return {url: ''}
```
byrichardpowell, line 20: I switched to if/else because the new types understand that `options.checkoutCartUrl` can be undefined, and that the url should never be undefined.

packages/ui-extensions-server-kit/src/ExtensionServerClient/types.ts
```diff
@@ -230,7 +230,7 @@ declare global {
   }
 }
 
-export const AVAILABLE_SURFACES = ['admin', 'checkout', 'post-checkout', 'pos', 'customer-accounts'] as const
+export const AVAILABLE_SURFACES = ['admin', 'checkout', 'post-checkout', 'pos', 'customer-accounts', null] as const
```
byrichardpowell, line 233: This change makes me wonder if we are overloading the concept of `ui extensions`.  `WebPixel` and `Theme` extensions end up in the dev console, but are very different beasts to everything else in this list.  They have no surface, hence the additional null `value`.

pepicrft: Can they be interacted with through the console? Of is it only to show that they are being served?

packages/app/package.json
```diff
@@ -77,15 +77,15 @@
     "@luckycatfactory/esbuild-graphql-loader": "3.7.0",
     "@oclif/core": "1.9.2",
     "@shopify/cli-kit": "3.25.0",
-    "@shopify/shopify-cli-extensions": "3.21.0",
+    "@shopify/ui-extensions-server-kit": "3.25.0",
```
pepicrft, line 80: Since we are only using the types in development, what about making this one a `devDependency`?

packages/ui-extensions-server-kit/src/ExtensionServerClient/types.ts
byrichardpowell: Thinking about this some more, with the Dev Index project the extensions dev console is about to become more than "just" extensions, so I think this is ok.

Theres a bigger change on the horizon though, which is to remove the "extensions" namespace from the dev console and supporting code.

packages/cli-kit/src/api/graphql/extension_specifications.ts
```diff
@@ -37,13 +37,15 @@ export interface RemoteSpecification {
   }
   features?: {
     argo?: {
-      surface: string
+      surface: Surface
     }
   }
 }
 
+type Surface = 'admin' | 'checkout' | 'post-checkout' | 'pos' | 'customer-accounts' | 'all' | null | undefined
```
isaacroldan, line 45: I think these values are not aligned with what comes from the Specifications API 🤔 
Just checked and it returns `customer_accounts` and `point_of_sale` 
Also for post_purchase, the returned surface is `checkout`

Where do these came from?

packages/app/src/cli/models/extensions/ui-specifications/post_purchase_extension.ts
```diff
@@ -8,7 +8,7 @@ const spec = createUIExtensionSpec({
   identifier: 'checkout_post_purchase',
   externalIdentifier: 'post_purchase_ui',
   externalName: 'Post-purchase UI',
-  surface: 'post_purchase',
+  surface: 'post-checkout',
```
isaacroldan, line 11: i'm removing the `surface` value from the specs in a WIP PR. 
We will only use the surface provided by the Specifications API, so that there are no confusions.

packages/cli-kit/src/api/graphql/extension_specifications.ts
isaacroldan: saying this because we defaulted to use the values from the API, not the hardcoded values anymore.

isaacroldan: yep, right now we are not fetching the values from Core during `dev`, but i'm working on it, so maybe is worth changing the values and merging this PR first

packages/app/src/cli/services/dev/extension/payload/models.ts
```diff
@@ -21,42 +20,3 @@ export interface ExtensionsEndpointPayload extends ExtensionsPayloadInterface {
     url: string
   }
 }
```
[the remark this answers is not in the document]
byrichardpowell: You mean should `@shopify/ui-extensions-server-kit` depend on `@shopify/app` ? No, I don't think so, but I do think it could depend on `@shopify/cli-kit` and that would be better than the approach in this PR.  

I'll work on that in a seperate PR and if we prefer that approach we can close this PR.

PR #834, opened 2022-11-28
Add urls per extension point

byrichardpowell:
Previously there was only one preview link per extension.  That's because each extension would only render on one surface.  With the new `ui_extension` type an extension can render multiple extensions across multiple extensionPoints.

As such, the Dev Console needs to:

1. EITHER render one preview link per extensions (for every existing extension type).
2. OR render one preview link per extension point (for the new `ui_extension` extension type).

In this PR we update the JSON data that the dev console recieves from the CLI so that it it can handle either situation. In a future PR you can imagine that a `ui_extension` will have multiple rows of icons (one per extension point), but older extension types will only have one row of icons.

Fixes #1954

1. Update some types in `packages/ui-extensions-server-kit` to respect the fact that the JSON for extensionPoints has changed.
2. Add the logic to construct the URL for each `extensionPoint`

The end result is a JSON structure where the following is possible

```
extensions: [{
   extensionsPoints: null
   // ...etc
}, {
   extensionsPoints: ["Extension::Point::A"]
   // ...etc
}, {
   extensionsPoints: [{
      target: "Extension::Point::A",
      module: "./src/file.js"
      root: {
        url: "http://tunnel-url/extensions/[uuid]/Extension::Point::A"
      }
   }]
   // ...etc
}]
```

### How to test your changes?

1. `cd fixtures/app`
2. `yarn generate extension` (Choose the new UI extension)
3. `yarn dev`
4. Go to the dev console URL
5. Inspect the JSON that in the connected websocket message.  You should see something like the image below.  Make sure going to that URL redirects:

Screenshot 2022-11-28 at 3 50 26 PM

### Post-release steps

none

### Measuring impact

### Checklist

### DIFF
--- a/packages/app/src/cli/services/dev/extension/payload.test.ts
+++ b/packages/app/src/cli/services/dev/extension/payload.test.ts
@@ -128,4 +128,59 @@ describe('getUIExtensionPayload', () => {
       })
     })
   })
+
+  test('adds root.url to extensionPoints[n] when extensionPoints[n] is an object', async () => {
+    await file.inTemporaryDirectory(async (tmpDir) => {
+      // Given
+      const uiExtension = await testUIExtension({
+        devUUID: 'devUUID',
+        configuration: {
+          name: 'UI Extension',
+          type: 'ui_extension',
+          metafields: [],
+          capabilities: {
+            block_progress: false,
+            network_access: false,
+          },
+          extensionPoints: [
+            {
+              target: 'Extension::Point::A',
+              module: './src/ExtensionPointA.js',
+            },
+            {
+              target: 'Extension::Point::B',
+              module: './src/ExtensionPointB.js',
+            },
+          ],
+        },
+      })
+      const options: ExtensionDevOptions = {} as ExtensionDevOptions
+      const development: Partial<UIExtensionPayload['development']> = {}
+
+      // When
+      const got = await getUIExtensionPayload(uiExtension, {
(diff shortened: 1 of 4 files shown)


packages/ui-extensions-server-kit/src/types.ts
```diff
@@ -73,7 +73,7 @@ export interface ExtensionPayload {
   version: string
   surface: Surface
   title: string
-  extensionPoints?: string[]
+  extensionPoints?: string[] | {target: string; module: string; main: {url: string}}
```
byrichardpowell, line 76: These types are duplicated with `packages/ui-extensions-server-kit/src/ExtensionServerClient/types.ts`.

I opened a WIP PR here to address the duplication and to unify the types: https://github.com/Shopify/cli/pull/833

packages/app/src/cli/services/dev/extension/payload.test.ts
```diff
@@ -128,4 +128,59 @@ describe('getUIExtensionPayload', () => {
...
+              module: './src/ExtensionPointA.js',
+            },
+            {
+              target: 'Extension::Point::B',
+              module: './src/ExtensionPointB.js',
+            },
+          ],
+        },
+      })
+      const options: ExtensionDevOptions = {} as ExtensionDevOptions
```
vividviolet, line 157: nit: add a blank line

PR #836, opened 2022-11-29
[Feature] Add theme-app-extension support to bundle.zip

jfisher727:

When deploying an app to Shopify, the CLI currently generates a bundle.zip of all the UI extensions that are part of the app. We want to include theme-app-extensions as part of this process to bring it inline with the other CLI managed extensions. 

Fixes #0000 

[App-Deploys 83](https://github.com/Shopify/app-deploys/issues/83)

We're adding a new environment flag to determine if theme-app-extensions should be included in the bundle, or if they should be uploaded through the current graphql mutation. If the environment flag is enabled, we will copy the theme-app-extension files from their current directories over to the temp directory that will eventually be zipped up. 

### How to test your changes?

1. Generate a test app and associate it with your dev store within Partners. 
2. Generate a theme-app-extension to be associated with your app.

### Measuring impact

### Checklist


packages/app/src/cli/services/deploy.ts
```diff
@@ -52,6 +52,18 @@ export const deploy = async (options: DeployOptions) => {
       }
     }),
   )
+  if (environment.local.useThemeBundling()) {
+    const themeExtensions = await Promise.all(
+      options.app.extensions.theme.map(async (extension) => {
+        return {
+          uuid: identifiers.extensions[extension.localIdentifier]!,
+          config: '{"theme_extension": {"files": {}}}',
```
isaacroldan, line 60: is this hardcoded config correct?

gonzaloriestra (approved):
It works as expected and the code LGTM! 👏 

Just two comments:
- Could you add some tests?
- Do we really need to hide it under a flag? Is there any use case you are not sure about or why not making this the default behavior now?

packages/app/src/cli/services/deploy.ts
jfisher727: Code within core will populate the `files` field.

isaacroldan:
Do you want to merge this to test it internally with the flag or wait for core to be ready?

packages/app/src/cli/services/build/extension.ts
```diff
@@ -52,6 +52,25 @@ export async function buildThemeExtensions(options: ThemeExtensionBuildOptions):
...
+        options.stdout.write(`Bundling theme extension ${extension.localIdentifier}...`)
+        const files = await path.glob(path.join(extension.directory, '/**/*'))
+
+        await Promise.all(
+          files.map(function (filepath) {
+            if (!(filepath.includes('.gitkeep') || filepath.includes('.toml'))) {
+              const relativePath = path.relative(extension.directory, filepath)
+              const outputFile = path.join(extension.outputBundlePath, relativePath)
+              return file.copy(filepath, outputFile)
+            }
```
isaacroldan, line 68: @Arkham I guess you meant this change? We shouldn't be bundling on every `build`, this method should just execute the `execThemeCheckCLI`.

We should have a different `bundleThemeExtension` method for this that's called only for deploy at `packages/app/src/cli/services/deploy/bundle.ts`

Issue #818, opened 2022-11-24
CLI Follow on: Discuss if we still need instantiate ExtensionInstance with ExtensionPoints


byrichardpowell:
See: Interface https://github.com/Shopify/cli/pull/808#discussion_r1031782890

PR #855, opened 2022-12-01
Cleanup ExtensionPointSpec

byrichardpowell:
In the original design of extensions the CLI would have an extension point spec for each extension point. This would  allow each extensions Point to express:

 1. The redirectURL for this extension point.
 2. The previewMessage for this extension point.

We've since used an architecture where:

1. The redirectURL is logic that the extensions dev server controls and is consistent across all extension points
2. The previewMessage is constructed by the extension spec by looping through the extension points

As such, I don't see a need to keep the ExtensionPointSpec around.  This PR removes it, but the PR is intended to drive discussion.  I'm not sure this is the right move, so I'm proposing this so we can all get on the same page.

Fixes: https://github.com/Shopify/cli/issues/818

The commit messages have the details.

Not that this changes how plugins are registered.  I'm not sure of the consequences here, but the commit message have my thoughts.

### How to test your changes?

1. `yarn && yarn build`
2. `cd fixtures/app`
3. `yarn dev`
4. Does the app run?
5. How does this affect plugins 

### Post-release steps

None

### Measuring impact

### Checklist


isaacroldan:
The original design assumed that every extension point could have a different configuration and we weren't inferring anything from the point target. It also allowed us to define new points in private plugins with specific custom configurations. (thinking on points that show different instructions or URLs even if they are for the same surface)

So, is that a real use case that could happen in the future or is it safe for now to assume that all extension points from the same surface will behave the same way? Maybe we can remove this until there is a need for it, just wanted to be sure

byrichardpowell:
> So, is that a real use case that could happen in the future or is it safe for now to assume that all extension points from the same surface will behave the same way? Maybe we can remove this until there is a need for it, just wanted to be sure

Yeah, that's the important question.  What are your thoughts @trisha-chill ?

lemonmade:
Hmm. There are definitely different "behaviors" per extension point — post-purchase and the soon-to-be-added order status page extension points are each deep in the flow, and we ideally want to land developers directly on them. So something needs to differentiate how to handle these extension points. But maybe that could be a responsibility of the host, and not the CLI? Like, we could make all the hosts handle some sort of protocol-based URL structure that the CLI uses for every extension point, and leave the hosts to be responsible for the logic of where to redirect to...

vividviolet:
> But maybe that could be a responsibility of the host, and not the CLI? Like, we could make all the hosts handle some sort of protocol-based URL structure that the CLI uses for every extension point, and leave the hosts to be responsible for the logic of where to redirect to...

This is already the case for Admin extensions but not through any protocol but by the extension point/type itself.  Ex. Admin Home always redirects to the app home page after its loaded.

byrichardpowell:
@vividviolet and I chatted about this on tuple and we are aligned with this change.

@isaacroldan  please can you review and approve the change if you are aligned? I'll rebase now.

PR #856, opened 2022-12-01
[WIP] Propose seperating UI and theme extensions

byrichardpowell:
In main we have 2 extension specifications:

* Function specification
* Extension specification, which serves both Theme and UI extensions.

Whilst working on extension grouping the extension specification has gained more functionality that is only relevant to UI extensions.  For example:

* `deployConfig()`
* `validate()`
* `preDeployValidation()`
* The ability to customize `previewMessage()` through the `specification` you pass
* `getBundleExtensionStdinContent`
* `hasExtensionPointTarget`

The purpose of this PR is show the effects of hard separating `UI` from `theme`.  Right now there is some separation via the `createExtensionSpec` function, but there is still a coupling because an Extension has to satisfy the requirements of every type of UI extension and the theme extension.

I'm hoping to use this PR to discuss the pros and cons of separating the extensions.  Before we do that, I want to acknowledge the limits of my context:

* I do not have deep context of how theme extensions are developed or deployed.
* I do not have any context on the theme extensions roadmap.  For example, are more theme extensions incoming?
* I do not have context on if these changes complicate the things with plugins
* I do not have broader context of CLI concerns.

So I'm hoping you all can help fill in the gaps.

Fixes https://github.com/Shopify/cli/issues/818

1. Create a new UI centric Extension Class. Use this when loading a new UI extension.
2. Create a new Theme centric extension class.  Use this when loading a new Theme extension
3. Assume that we do not need to customize a theme extension.  Given this assumption we then simplify 2 things:
  1. How we load a theme extensions
  2. The implementation of the theme extension specification.

### How to test your changes?

I'll fill these steps in if we get alignment on the approach.

### Post-release steps

I'll fill these steps in if we get alignment on the approach.

### Measuring impact

### Checklist


isaacroldan:
I think the proposal makes sense, but this approach breaks `generate extension` for theme app extensions. For generate we load all `specs`, then check with the Specifications API to see which ones the user has access to and then we show those. So `theme` needs to be a spec (either a generic one like we had or a `ThemeExtensionSpec`) that can be used in `src/cli/commands/app/generate/extension.ts`

The idea of having everything together was that the Specifications API has all the extensions together, so it seemed like a good idea to continue that pattern.

I see two options:
- Keep the current implementation for now and see how it evolves (splitting in later won't shouldn't be difficult)
- Approve the proposed change, but define a `ThemeExtensionSpec` that behaves exactly the same as the UI Extension spec and Function spec, for consistency. There will be more code for sure.

packages/app/src/cli/models/extensions/theme.ts
```diff
@@ -0,0 +1,101 @@
...
+ * the config toml file for the theme extension follow the ThemeExtensionSchema
+ *
+ * This class holds the public interface to interact with theme extensions
+ */
+export class ThemeExtensionInstance<TConfiguration extends ThemeConfigContents = ThemeConfigContents>
+  implements ThemeExtension<TConfiguration>
+{
+  outputBundlePath: string
+  devUUID: string
+  localIdentifier: string
```
vividviolet, line 32: Same here

vividviolet (commented):
I think separating out UI and Theme extension is a great idea. I think there are a few unused props that can now be deleted from Theme extensions - can you double check?

packages/app/src/cli/models/extensions/theme.ts
```diff
@@ -0,0 +1,101 @@
...
+ * Before creating this class we've validated that
+ * the config toml file for the theme extension follow the ThemeExtensionSchema
+ *
+ * This class holds the public interface to interact with theme extensions
+ */
+export class ThemeExtensionInstance<TConfiguration extends ThemeConfigContents = ThemeConfigContents>
+  implements ThemeExtension<TConfiguration>
+{
+  outputBundlePath: string
+  devUUID: string
```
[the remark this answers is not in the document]
byrichardpowell: Good catch, removed

packages/app/src/cli/models/extensions/theme.ts
```diff
@@ -0,0 +1,101 @@
...
+ * Class that represents an instance of a local theme extension
+ * Before creating this class we've validated that
+ * the config toml file for the theme extension follow the ThemeExtensionSchema
+ *
+ * This class holds the public interface to interact with theme extensions
+ */
+export class ThemeExtensionInstance<TConfiguration extends ThemeConfigContents = ThemeConfigContents>
+  implements ThemeExtension<TConfiguration>
+{
+  outputBundlePath: string
```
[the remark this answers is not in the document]
byrichardpowell: Thanks Trish, removed!

packages/app/src/cli/validators/extensions/ui.ts
```diff
@@ -1,5 +1,5 @@
 import {UIExtension} from '../../models/app/extensions.js'
 
 export async function validateUIExtensions(extensions: UIExtension[]) {
-  return Promise.all(extensions.map((ext) => ext.preDeployValidation()))
+  return Promise.all(extensions.map((ext) => ext.validate()))
```
isaacroldan, line 4: Why this change? This function is called during deploy so it makes sense to use the `preDeployValidation` no?

byrichardpowell: Yeah, you're right.  This was a mistake, reverted! Thanks :+1:

packages/app/src/cli/models/extensions/functions.ts
```diff
@@ -143,9 +144,8 @@ export function createFunctionSpec<TConfiguration extends FunctionConfigType = F
     configSchema: BaseFunctionConfigurationSchema,
     gated: false,
     registrationLimit: spec.registrationLimit ?? blocks.functions.defaultRegistrationLimit,
-
     category: (): ExtensionCategory => 'function',
   }
 
-  return {...defaults, ...spec}
+  return {...defaults, ...spec, surface: 'none'}
```
isaacroldan, line 150: why is this needed? functions won't ever need surface, is this because of some typings issue?

byrichardpowell: Yeah, good catch.  

it looks like I added this to makes types pass, but we've since made changes that mean this is not needed.  So i've removed the concept of surface from this file entirely.

isaacroldan (approved):

PR #869, opened 2022-12-06
Run node with the --experimental-vm-modules flag which is needed by @shopify/mini-oxygen

pepicrft:
`@shopify/cli-hydrogen`'s `@shopify/mini-oxygen` dependencies requires the Node runtime to be started with the `--experimental-vm-modules` flag. 

I updated the shebang of the `cli-main` and `cli-hydrogen` executables to include the flag. 

### How to test your changes?
It's important that we test this across the supported OSs (Windows, macOS, and Linux) and the lowest (14.17) and highest (most recent) supported Node versions.
I'd recommend checking out [these steps](https://github.com/Shopify/cli/blob/main/docs/cli/cross-os-compatibility.md#manual-testing) for testing across OSs.

### Measuring impact

### Checklist


packages/cli-main/bin/run.js
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env node
+#!/usr/bin/env node --experimental-vm-modules
```
frandiox, line 1: This is [the only one we need in the Hydrogen plugin](https://github.com/Shopify/h2/blob/v0.x-2022-10/packages/cli/bin/postinstall.mjs#L21) since it's only required in `dev` and `preview` commands.

However, we should pass `-S` parameter to `/usr/bin/env` to accept other flags in Linux. It works in Mac without `-S` but it won't work on CI or Linux machines due to the way `env` parses its parameters. If you want a longer explanation, search for "23.2.2 -S/--split-string usage in scripts" in the [GNU docs page](https://www.gnu.org/software/coreutils/manual/html_node/env-invocation.html).

```suggestion
#!/usr/bin/env -S node --experimental-vm-modules
```

PR #871, opened 2022-12-06
Testing for CLI main commands; Automated coverage

shauns:
Fixes https://github.com/Shopify/internal-cli-foundations/issues/432

- Adds tests for the commands in CLI main
- Adds an automated coverage message for packages supporting test coverage (main + CLI kit)

### DIFF
--- a/.github/actions/download-and-publish-test-coverage/action.yml
+++ b/.github/actions/download-and-publish-test-coverage/action.yml
@@ -3,7 +3,9 @@ description: "Downloads test coverage stats from the base branch, publishes a co
 runs:
   using: "composite"
   steps:
+    # Strips forward slashes and dots as not supported by artifact commands
     - name: Clean base ref name
+      shell: bash
       env:
           SAFE_BASE_REF_NAME: "${{ github.base_ref }}"
       run: |
--- a/.github/actions/run-and-save-test-coverage/action.yml
+++ b/.github/actions/run-and-save-test-coverage/action.yml
@@ -5,13 +5,19 @@ runs:
   steps:
     - name: Unit tests with coverage
       run: pnpm nx run-many --all --skip-nx-cache --target=test:coverage --exclude=features
+      shell: bash
     - name: Combine Jest coverages
       run: npx istanbul-merge --out="coverage.raw.json" "packages/*/coverage/coverage-final.json"
+      shell: bash
     - name: Convert coverage to Jest
       run: ./bin/save-coverage-file.js
+      shell: bash
     - name: Take copy of report as new baseline for branch
       run: cp ./report.json ./baseline-report.json
+      shell: bash
+    # Strips forward slashes and dots as not supported by artifact commands
     - name: Clean ref name
+      shell: bash
       env:
           SAFE_REF_NAME: "${{ github.head_ref }}"
       run: |


.github/workflows/shopify-cli.yml
```diff
@@ -101,6 +101,31 @@ jobs:
...
+            SAFE_REF_NAME: "${{ github.head_ref }}"
+        run: |
+          SAFE_REF_NAME=${{ env.SAFE_REF_NAME }}
+          SAFE_REF_NAME=${SAFE_REF_NAME//[\/.]/}
+          echo SAFE_REF_NAME=${SAFE_REF_NAME} >> $GITHUB_ENV
+      - uses: actions/upload-artifact@v3
+        if: ${{ matrix.node == '18.7.0' && matrix.os == 'ubuntu-latest' }}
+        with:
+          name: ${{ env.SAFE_REF_NAME }}--coverage-report
+          path: ./baseline-report.json
```
Arkham, line 128: could we extract the shared bits in a shared action inside `.github/actions` to reduce duplication? in this way we can also have a single `if: https://shopify.slack.com/archives/C0BUY1F5L/p1670411369252219`

.github/actions/download-and-publish-test-coverage/action.yml
```diff
@@ -0,0 +1,26 @@
...
+runs:
+  using: "composite"
+  steps:
+    - name: Clean base ref name
+      env:
+          SAFE_BASE_REF_NAME: "${{ github.base_ref }}"
+      run: |
+        SAFE_BASE_REF_NAME=${{ env.SAFE_BASE_REF_NAME }}
+        SAFE_BASE_REF_NAME=${SAFE_BASE_REF_NAME//[\/.]/}
+        echo SAFE_BASE_REF_NAME=${SAFE_BASE_REF_NAME} >> $GITHUB_ENV
```
pepicrft, line 12: Since this is in two places. What about having a separate action for this?

.github/actions/run-and-save-test-coverage/action.yml
```diff
@@ -0,0 +1,24 @@
...
+  using: "composite"
+  steps:
+    - name: Unit tests with coverage
+      run: pnpm nx run-many --all --skip-nx-cache --target=test:coverage --exclude=features
+    - name: Combine Jest coverages
+      run: npx istanbul-merge --out="coverage.raw.json" "packages/*/coverage/coverage-final.json"
+    - name: Convert coverage to Jest
+      run: ./bin/save-coverage-file.js
+    - name: Take copy of report as new baseline for branch
+      run: cp ./report.json ./baseline-report.json
```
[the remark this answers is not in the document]
shauns: Not quite, in a PR it works like:
1. product big report.json
2. copy it to baseline-report.json
3. upload it as most recent report for this branch
4. download most recent report for this branch -- replaces baseline-report.json. this step is allowed to fail.
5. use report.json (this branch) and baseline-report.json (base branch) to generate the summary

On main we do the first three steps only. The copying is important so this doesn't fail even if the base branch doesn't have a report available yet.

shauns:
> What about adding a section to [this page](https://github.com/Shopify/cli/blob/main/docs/cli/testing-strategy.md) where we explain how we expect the reported data to be used?

Once we've got a baseline from main in place, we should have something we can screenshot and decide what to do with.

PR #825, opened 2022-11-25
Clean up (almost)all hardcoded extension types

isaacroldan:
### WHY are these changes introduced?
As part of the extensions project, since all the data for a single extension is in a single file, we can remove all hardcoded stuff....

### WHAT is this pull request doing?
- Remove almost everything that has a hardcoded extension type.
- Wherever we were using `UIExtensionType` we are now using an ExtensionSpec that contains all the needed info.
- LOOK HOW CLEAN `constants.ts` IS NOW!

### How to test your changes?
Just that everything keeps working...

### Measuring impact

How do we know this change was effective? Please choose one:

- [ ] n/a - this doesn't need measurement, e.g. a linting rule or a bug-fix
- [ ] Existing analytics will cater for this addition
- [ ] PR includes analytics changes to measure impact

### Checklist

- [ ] I've considered possible cross-platform impacts (Mac, Linux, Windows)
- [ ] I've considered possible [documentation](https://shopify.dev) changes
- [ ] I've made sure that any changes to `dev` or `deploy` have been reflected in the [internal flowchart](https://www.figma.com/file/7vqUp50u6dm48Zfb4JRRn8/CLI3-Internals).

PR #875, opened 2022-12-07
Add internal order_routing_location_rule extension type

Keaton11:

This PR adds the ``order_routing_location_rule`` extension type.

Related Issue: https://github.com/Shopify/shopify/issues/393386

This is one of [two changes](https://vault.shopify.io/page/Function-API-prototyping-guide~Dyet.md#registering-your-api-name) needed to register the ``order_routing_location_rule`` API name so that related functions can be deployed. 

### WHAT should reviewers focus on?

I based the template paths on those that were originally written for ``order_routing_ranker`` but these might need to be changed.

### Checklist

### DIFF
--- a/packages/app/src/cli/models/extensions/function-specifications/order_routing_location_rule.ts
+++ b/packages/app/src/cli/models/extensions/function-specifications/order_routing_location_rule.ts
@@ -1,0 +1,11 @@
+import {createFunctionSpec} from '../functions.js'
+
+const spec = createFunctionSpec({
+  identifier: 'order_routing_location_rule',
+  externalIdentifier: 'order_routing_location_rule',
+  externalName: 'Order routing location rule',
+  gated: true,
+  templatePath: (lang) => `order-routing/${lang}/rankers/default`,
+})
+
+export default spec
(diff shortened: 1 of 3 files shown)


DuncanUszkay1:
>For the template path, does there need to be changes in https://github.com/Shopify/function-templates (adding the order routing directory and sample extensions) before this is merged?

I don't see much reason to add templates as nobody is allowed to upload these but us right? Seems like something we could circle back to post editions.

You might want to add the template first just so that you can tophat

DuncanUszkay1 (approved):
LGTM

luanpy (approved):
LGTM!

packages/app/src/cli/models/extensions/function-specifications/order_routing_location_rule.ts
```diff
@@ -0,0 +1,11 @@
+import {createFunctionSpec} from '../functions.js'
+
+const spec = createFunctionSpec({
+  identifier: 'order_routing_location_rule',
+  externalIdentifier: 'order_routing_location_rule',
+  externalName: 'Order routing location rule',
+  gated: true,
```
luanpy, line 7: The [PR](https://github.com/Shopify/cli/pull/825/files#diff-b0e909dd71b245fc4d00358a78abb504fddc6bee0d9eb2068c927280ce14971b) where `public` was changed to `gated`

PR #879, opened 2022-12-08
Update @oclif/core to 1.21.0 to fix the dependency installation issue in the fixture app

pepicrft:
`dev` and `build` install dependencies by default. When they do, they use the lockfile in the project directory to determine the package manager to use. Because the `fixture` project is part of the repository workspace, there's no PNPM lockfile and therefore tries to install dependencies there with `npm` leading to the following error reported by @amcaplan:

```
  82773 ── external error ─────────────────────────────────────────────────────────────────
  82774 
  82775 Error coming from `npm install`
  82776 
  82777 Command failed with exit code 1: npm install
  82778 npm ERR! Cannot read properties of null (reading 'matches')
  82779 
  82780 npm ERR! A complete log of this run can be found in:
  82781 npm ERR!     /Users/amcaplan/.npm/_logs/2022-12-08T12_00_24_674Z-debug-0.log
  82782 
  82783 ───────────────────────────────────────────────────────────────────────────────────
```

Until we can have a discussion around the responsibilities of the CLI and the future of the `--skip-dependencies-install` flag, I'm configuring `dev.yml` to set the environment `SHOPIFY_FLAG_SKIP_DEPENDENCIES_INSTALLATION: 1` and disable that functionality while working in the monorepo because it's not necessary. 
Setting the variable is not sufficient because the version of `@oclif/core` that we are using only uses the environment variables for flags that represent options. Therefore I'm updating the version of `@oclif/core` to have @amcaplan's fix https://github.com/oclif/core/pull/536.

### How to test your changes?
Try to create a new app and interact with it. Also try the commands against the main fixture app.

### Measuring impact

### Checklist


amcaplan:
Note that in the past we've explicitly [reverted a version upgrade](https://github.com/Shopify/cli/pull/666) due to concerns about breaking changes in oclif (which prompted my PR to oclif that you mentioned). I think we'd discussed adding some high-level tests (likely some kind of golden master) to gain more confidence that the breaking changes haven't broken anything.

Is it a good idea to change that approach now?

PR #882, opened 2022-12-08
WIP add appId to extension endpoint

misha-reyzlin-shopify:
we need the real appId in order to obtain storefront API tokens to enable storefront query API in extensions

Fixes #0000 

### Measuring impact

### Checklist


pepicrft:
I'd get a review from @isaacroldan

isaacroldan:
I think I'm missing something here...
You defined the ID in the interface, but is not being set anywhere right? I 🎩 and can't see the ID value on my extensions.

The ID is a different thing from the API_key of the app right?

packages/app/src/cli/services/dev.ts
```diff
@@ -108,6 +109,7 @@ async function dev(options: DevOptions) {
   if (options.app.extensions.ui.length > 0) {
     const devExt = await devUIExtensionsTarget({
       app: options.app,
+      id: app.id,
```
misha-reyzlin-shopify, line 112: @isaacroldan we are passing it off the app object here, there was a typo before where it was `options.id` but we have caught that now

balaclark (approved):
LGTM from my point of view, I've tophatted locally and the id is coming through. 

app id

PR #885, opened 2022-12-09
[Bug CI] Use action-download-artifact

shauns:
This is needed to access artifacts from separate runs.


Arkham (commented):
Is this needed so we can access the artifact from `main` inside the PR runs? If so, could we just add a comment before the section that explains why we're not using the default `action/download-artifact`? 🙏

Arkham (approved):
TY!!

PR #890, opened 2022-12-12
Only request granted scopes on selected app

amcaplan:

Fixes the request for an org's apps taking way too long (>10 seconds in some cases if there are many apps!)

It turns out most of the time is spent fetching granted scopes for each app. Given how expensive this is, so we should only request it once on the selected app, not on each returned app.

### How to test your changes?

Any operation (simplest is `shopify app env show --verbose`) you should see the request time for the app list decrease dramatically.

Worth tophatting various affected flows (`dev` and `deploy`) as well.

### Measuring impact

### Checklist

### DIFF
--- a/.changeset/strong-dolls-pay.md
+++ b/.changeset/strong-dolls-pay.md
@@ -1,0 +1,5 @@
+---
+'@shopify/app': patch
+---
+
+Only request necessary information to speed up fetching apps list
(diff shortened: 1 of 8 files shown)


gonzaloriestra (approved):
Awesome work! 👏 

I've tested the `env show` command with my partners account, and it went from 40 to 6 seconds 🚀

amcaplan:
I guess I have fewer apps, so it's 13s to 5s... but still a very nice outcome!

I've tophatted sufficiently to build my own confidence in this work, so this is ready to merge (pending CI)!

Issue #809, opened 2022-11-23
Add a new `api_access` capability to extensions


balaclark:
We are adding support for the Storefront API to Checkout extensions. Developers should enable this via a `api_access` capability. Eventually this capability will be used to grant access to other APIs.

This will require updating:
* [x] CLI
* [x] [Capabilities schema](https://github.com/Shopify/cli/blob/main/packages/app/src/cli/models/extensions/schemas.ts#L12)
* [x] [Extension payload](https://github.com/Shopify/cli/blob/17e1f5f4af1fd20f11c564525ebf5af3ff7bbdf9/packages/app/src/cli/services/dev/extension/payload.ts#L33-L36)
* [x] [Checkout editor](https://github.com/Shopify/web/blob/85bd581ad85f9e783a4dfa61eb6301cb8d041ceb/app/sections/Checkout/CheckoutEditor/foundation/Extensions/types.ts#L60-L63)

PR #893, opened 2022-12-12
[Feature] adds "api_access" capability to ui extensions #809

balaclark:

Checkout UI Extensions are getting Storefront API access soon, this PR adds in a capability so that extension developers can opt-in to this feature. Checkout-web will auto generate auth tokens for these extensions, so SFAPI usage needs to be stated ahead of time.

Closes https://github.com/Shopify/cli/issues/809 

Adds in the new `api_access` capability and sends it with the extension payload.

### How to test your changes?

The unit test coverage should be enough here.

### Post-release steps

Once the SFAPI is released we'll update the docs, leaving this undocumented for now.

### Measuring impact

### Checklist


packages/app/src/cli/models/app/app.test-data.ts
```diff
@@ -39,6 +39,7 @@ export async function testUIExtension(uiExtension: Partial<UIExtension> = {}): P
     capabilities: {
       block_progress: false,
       network_access: false,
+      storefront_api: false,
```
lemonmade, line 42: I think we had a thread going about making this be `api_access` instead of `storefront_api` — where did we land on that?

balaclark: Updated now

packages/app/templates/ui-extensions/projects/ui_extension/shopify.ui.extension.toml.liquid
```diff
@@ -3,6 +3,7 @@ name = "{{ name }}"
 [capabilities]
 network_access = true
 block_progress = true
+api_access = true
```
lemonmade, line 6: Super nitpicky thing: I think this is going to be the most widely-used capability, so can we put it above the other two? It is also a little shorter and my brain feels better when lines go from short to long 😆

lemonmade (approved):
❤️ Thanks for the quick naming update!

gonzaloriestra (approved):

faouzioudouh (approved):
🔥

PR #896, opened 2022-12-12
Add Tasks component

matteodepalo:

We want to be able to convey to the user that a some tasks are running. So far we've been using `listr2`, but we want to follow our own design approach.

~~We'll need to merge [this other PR](https://github.com/Shopify/cli/pull/798) first before we can review this properly.~~

This PR introduces a new `renderTasks` method which takes an array of items in the form of: `{title: string, task: Promise}` and displays an animated bar while these tasks are running. At the end it displays a green bar in case of success or a red bar in case any of the tasks has failed to complete.

### How to test your changes?

Run `kitchen-sink` and see the bar for yourself at the end! It should look like this:

### Measuring impact

### Checklist


packages/cli-kit/src/private/node/ui/components/TextAnimation.tsx
```diff
@@ -1,65 +1,47 @@
-import {renderString} from '../../ui.js'
-import chalkAnimation from 'chalk-animation'
+/* eslint-disable id-length */
 import {Text} from 'ink'
-import React, {useEffect, useState} from 'react'
-
-type AnimationName = 'rainbow' | 'pulse' | 'glitch' | 'radar' | 'neon' | 'karaoke'
+import React, {useEffect, useRef, useState} from 'react'
+import gradient from 'gradient-string'
```
pepicrft, line 4: Not blocking, but these imports, especially React, [might have a negative impact](https://github.com/Shopify/cli/pull/970) on the startup time of the CLI. I'm working on a tool that will help us see if this is a bottleneck, and if it is we'll have to move them to dynamic imports.

packages/cli-kit/src/private/node/ui/components/Prompt.test.tsx
```diff
@@ -34,7 +34,8 @@ describe('Prompt', async () => {
 
     expect(renderInstance.lastFrame()).toMatchInlineSnapshot(`
       "?  Associate your project with the org Castile Ventures?
-      [36m✔[39m  [36msecond[39m"
```
pepicrft, line 37: When I see these tests I always wonder if we should do matches without ASCI escape characters. As a code reviewer, `[36m` says nothing to me. I'd have to go and search what the escape code means. Nothing actionable in for this PR though, just a random thought that I had.

packages/cli-kit/src/private/node/ui/hooks/use-async.ts
```diff
@@ -0,0 +1,26 @@
...
+  asyncFunction: () => Promise<unknown>,
+  {onResolve = () => {}, onReject = () => {}}: Options,
+) {
+  const {exit: unmountInk} = useApp()
+
+  useEffect(() => {
+    asyncFunction()
+      .then(() => {
+        onResolve()
+        unmountInk()
```
pepicrft, line 19: What does `unmounting` mean in the context of a component?

pepicrft: Also, that side effect doesn't seem well captured in the name of the function, `useAsync`.

packages/cli-kit/src/private/node/ui/hooks/use-layout.ts
```diff
@@ -0,0 +1,23 @@
+import {useStdout} from 'ink'
+
+const MIN_WIDTH = 80
+
+export default function useLayout() {
```
pepicrft, line 5: What about adding a unit test for this one?

packages/cli-kit/src/private/node/ui/components/Tasks.tsx
```diff
@@ -0,0 +1,57 @@
...
+  const {width} = useLayout()
+  const loadingBar = new Array(width).fill(loadingBarChar).join('')
+  const [currentTask, setCurrentTask] = useState<Task>(tasks[0]!)
+  const [state, setState] = useState<'success' | 'failure' | 'loading'>('loading')
+
+  const runTasks = async () => {
+    for (const task of tasks) {
+      setCurrentTask(task)
+      // eslint-disable-next-line no-await-in-loop
+      await task.task()
```
pepicrft, line 28: Are there plans to implement a version of this component with tasks that run concurrently?

packages/cli-kit/src/private/node/ui/hooks/use-async.ts
```diff
@@ -0,0 +1,26 @@
...
+import {useEffect} from 'react'
+
+interface Options {
+  onResolve?: () => unknown
+  onReject?: (error: Error) => void
+}
+
+export default function useAsync(
+  asyncFunction: () => Promise<unknown>,
+  {onResolve = () => {}, onReject = () => {}}: Options,
```
pepicrft, line 11: Nitpick: What about aligning the [naming](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/then#parameters) of the `then` function: `onFulfilled` and `onRejected`?

packages/cli-kit/src/public/node/ui.tsx
```diff
@@ -235,3 +236,10 @@ export async function renderConfirmation({question, infoTable}: ConfirmationProp
 
   return prompt({message: question, choices, infoTable})
 }
+
+/**
+ * Runs async tasks and displays their progress to the console.
+ */
+export function renderTasks(tasks: Task[]) {
```
pepicrft, line 243: Task is a `public` interface so I recommend moving it to the `public/node` folder as well. Otherwise we might introduce breaking changes there in the future without realising it.

pepicrft (approved):

packages/cli-kit/src/private/node/ui/hooks/use-async.ts
matteodepalo: Basically, this hook allows you to wait inside react components for async behavior. Once this behavior resolves we want to unmount the rendering `ink` instance explicitly so that the promise we're awaiting from the outside (`waitForExit`) resolves.

Regarding the naming I'm open to suggestions. Would something like `awaitAndUnmount` be better?

packages/cli-kit/src/private/node/ui/hooks/use-layout.ts
matteodepalo: Will do 👍

packages/cli-kit/src/private/node/ui/components/Tasks.tsx
matteodepalo: I looked at listr2 and it seems like tasks are not being processed concurrently there as well so for now I think we can keep this sequential. Parallel tasks would also need a different interface to report progress.

packages/cli-kit/src/public/node/ui.tsx
matteodepalo: `renderPrompt` also uses `PromptProps` which is an interface defined privately. We could remove the dependency on these private interfaces by creating new public ones that mirror the private ones, but this might introduce some differences in the future that typescript will not catch for us (e.g. we add a field to the private ones and forget to update the public one).

packages/cli-kit/src/private/node/ui/hooks/use-async.ts
matteodepalo: @pepicrft how about `useEffectAsync`?

packages/cli-kit/src/private/node/ui/components/Prompt.test.tsx
pepicrft: That's a good idea. In a future iteration we could provide a custom matcher that outputs in the terminal the two outputs so that you can see the styles applied.

pepicrft (approved):

packages/cli-kit/src/private/node/ui/components/Prompt.test.tsx
matteodepalo: Actually if you get an error right now it already works, you will see the colored output in the terminal since the escape sequences are interpreted.

PR #897, opened 2022-12-12
Discourage accessing `process.env` via ESLint

pepicrft:
I started to notice tests that mutate the global `process.env` state like:

```js
const oldEnv = process.env
process.env.FOO = "bar"

doSomething()

process.env = oldEnv
```
This is a bad pattern that might lead to flakiness due to state leaking across tests and prevent us from running tests concurrently. 

- [x] Add a ESLint rule `no-process-env` to disable the usage of `process.env`
- [x] Expose the function `getEnvironmentVariables` from `@shopify/cli-kit/node/environment` to read the environment variables.
- [x] Update all the usages of `process.env` to use the new function and fix the tests that broke as a result. 

### How to test your changes?
`pnpm run lint` and `pnpm run test` should pass.

### Measuring impact

### Checklist

### DIFF
--- a/.changeset/lucky-teachers-help.md
+++ b/.changeset/lucky-teachers-help.md
@@ -1,0 +1,10 @@
+---
+'@shopify/app': minor
+'@shopify/cli-hydrogen': minor
+'@shopify/cli-kit': minor
+'@shopify/cli': minor
+'@shopify/features': minor
+'@shopify/theme': minor
+---
+
+Not use the process.env global variable and read the values from the function exported by the @shopify/cli-kit/node/environment module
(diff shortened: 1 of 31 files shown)


pepicrft:
> Could we achieve the same result ("not letting people touch process.env directly") by replacing the old occurrences of env = process.env with env = getEnvironmentVariables() in the function params?

I reverted some of the changes to keep dependency injection but as I progressed I noticed that we end up with with all the function signatures having `env`. I think having logic to mock a dependency of the module is expected and it's not as much as to justify investing in improving the ergonomics. Down the road we could consider something like:

```js
withEnvironment({ FOO: "BAR"}, () => {

})
```

Issue #864, opened 2022-12-05
[Bug]: Error when running dev in app or extension


joshuaiz:
### Please confirm that you have:

- [X] Searched [existing issues](.) to see if your issue is a duplicate. (If you’ve found a duplicate issue, feel free to add additional information in a comment on it.)
- [X] Reproduced the issue in [the latest CLI version](https://www.npmjs.com/package/@shopify/cli).

### In which of these areas are you experiencing a problem?

Extension

### Expected behavior

Extension should run
(also true for app)

### Actual behavior

Extension or app should run

### Stack trace

```shell
error ──────────────────────────────────────────────────────────────────────╮
│                                                                              │
│  (intermediate value).doOnOk(...).mapError(...).valueOrThrow is not a        │
│  function                                                                    │
│                                                                              │
│  To investigate the issue, examine this stack trace:                         │
│  at generateURL (.pnpm/@shopify+app@3.23.0_graphql-tag@2.12.6/node_modules/  │
│  @shopify/app/src/cli/services/dev/urls.ts:97)                               │
│   .valueOrThrow()                                                            │
│  at processTicksAndRejections (node:internal/process/task_queues:95)         │
│  at async generateFrontendURL (.pnpm/@shopify+app@3.23.0_graphql-tag@2.12.6  │
│  /node_modules/@shopify/app/src/cli/services/dev/urls.ts:80)                 │
│   frontendUrl = await generateURL(options.commandConfig, frontendPort)       │
│  at async dev (.pnpm/@shopify+app@3.23.0_graphql-tag@2.12.6/node_modules/@s  │
│  hopify/app/src/cli/services/dev.ts:66)                                      │
│   const {frontendUrl, frontendPort, usingLocalhost} = await                  │
│   generateFrontendURL({                                                      │
│  at run (.pnpm/@shopify+app@3.23.0_graphql-tag@2.12.6/node_modules/@shopify  │
│  /app/src/cli/commands/app/dev.ts:100)                                       │
│   await dev({                                                                │
│                                                                              │
╰──────────────────────────────────────────────────────────────────────────────╯
 ELIFECYCLE  Command failed with exit code 1.
```

### Reproduction steps

1. Created app extension
2. cs into extension directory
3. run `pnp run dev`
4. same error if trying to run app

The app was running as of yesterday but stopped working after running the extension creation script.

Not sure what to do now. Thanks!

### Operating System

macOS Ventura 13.0.1

[...]

PR #940, opened 2022-12-16
Fix generation of extensions

pepicrft:
Fixes https://github.com/Shopify/cli/issues/864#issuecomment-1354088698

The globbing to get the list of specification modules ignores the `ignore` value passed. 

Instead of using `ignore` I'm passing a list of patterns, the second which uses the `!` prefix to exclude the `.d.ts` modules from the list.

### How to test your changes?
Try to generate a new extension in a project.

### Measuring impact

### Checklist

### DIFF
--- a/packages/app/src/cli/models/extensions/specifications.ts
+++ b/packages/app/src/cli/models/extensions/specifications.ts
@@ -26,7 +26,7 @@ async function loadSpecs(directoryName: string) {
    * in the list of files.
    */
   const url = path.join(path.dirname(fileURLToPath(import.meta.url)), path.join(directoryName, '*.{js,ts}'))
-  let files = await path.glob(url, {ignore: ['**.d.ts', '**.test.ts'], dot: true})
+  let files = await path.glob(url, {ignore: ['**.d.ts', '**.test.ts']})
 
   // From Node 18, all windows paths must start with file://
   const {platform} = os.platformAndArch()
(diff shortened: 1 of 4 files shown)


packages/app/src/cli/models/extensions/specifications.ts
```diff
@@ -20,8 +20,19 @@ export async function allThemeSpecifications(): Promise<ThemeExtensionSpec[]> {
...
+   * transform the TS module into a JS one before loading it. Hence the inclusion of .ts
+   * in the list of files.
+   */
+  let files = (await path.glob(path.join(url, '*'))).filter((filePath) => {
+    const isTest = filePath.endsWith('.test.ts')
+    const isTypescript = filePath.endsWith('.ts')
+    const isDeclaration = filePath.endsWith('.d.ts')
+    const isJS = filePath.endsWith('.js')
+    return isJS || (isTypescript && !isDeclaration && !isTest)
+  })
```
isaacroldan, line 35:
```suggestion
  let files = (await path.glob(path.join(url, , '*.{js,ts}'))).filter((filePath) => {
    return !filePath.endsWith('.test.ts') && !filePath.endsWith('.d.ts'))
  })
```
Feel free to ignore the suggestion 😅

isaacroldan (approved):

matteodepalo (approved):

packages/app/src/cli/models/extensions/specifications.ts
gonzaloriestra, line 35: Another alternative, using the [dot option](https://github.com/mrmlnc/fast-glob#dot):

```suggestion
  const url = path.join(path.dirname(fileURLToPath(import.meta.url)), path.join(directoryName, '*.{js,ts}'))
  /**
   * When running tests, "await import('.../spec..ts')" is handled by Vitest which does
   * transform the TS module into a JS one before loading it. Hence the inclusion of .ts
   * in the list of files.
   */
  let files = await path.glob(url, {ignore: ['**.d.ts', '**.test.ts'], dot: true})
```

PR #950, opened 2022-12-18
Fix session refresh

amcaplan:

Fixes recurring auth prompt. We are treating expired auth as revoked auth due to a faulty check.

Assume the best: If it's possible to refresh, do that. We always fall back to full auth anyway, so this should be safe.

While touching this code, I took the opportunity to parallelize 2 calls for access tokens that had been executed serially; each took about half a second.

### How to test your changes?

Wait for your tokens to expire, then run `dev` or any other command requiring auth. It'll hang a bit then run, instead of prompting the user to log in again.

### Measuring impact

### Checklist

### DIFF
--- a/packages/cli-kit/src/api/partners.test.ts
+++ b/packages/cli-kit/src/api/partners.test.ts
@@ -3,7 +3,7 @@ import {buildHeaders} from './common.js'
 import {partners} from '../environment/fqdn.js'
 import {graphqlClient} from '../http/graphql.js'
 import {test, vi, expect, describe, beforeEach} from 'vitest'
-import {GraphQLClient, ClientError} from 'graphql-request'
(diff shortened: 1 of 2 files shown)


packages/cli-kit/src/session/validate.ts
```diff
@@ -77,11 +73,6 @@ function isTokenExpired(token: ApplicationToken): boolean {
   return token.expiresAt < expireThreshold()
 }
 
-async function isPartnersTokenRevoked(token: ApplicationToken) {
-  if (!token) return false
-  return partners.checkIfTokenIsRevoked(token.accessToken)
```
isaacroldan, line 82: can you remove also the `checkIfTokenIsRevoked` method?

isaacroldan (approved):
Minor comment and 👌

PR #953, opened 2022-12-19
Enable password flag for theme dev

gonzaloriestra:

Related to https://github.com/Shopify/internal-cli-foundations/issues/332

Since https://github.com/Shopify/cli/pull/325, most theme commands were already working with Theme Access passwords, but the `dev theme` command required additional work. Now it's ready and we can enable the password flag for that command it as well.

- Allows password flag/env variable for `theme dev` command
- Upgrades theme-check to get intelligent code completion

### How to test your changes?

- `pnpm shopify theme dev --store your-store --password your-password --path /your/theme/path`
- `SHOPIFY_FLAG_STORE=your-store SHOPIFY_CLI_THEME_TOKEN=your-password pnpm shopify theme dev`

### Measuring impact

### Checklist


packages/cli-kit/src/output.ts
```diff
@@ -357,7 +357,7 @@ export function getOutputUpdateCLIReminder(
   const versionMessage = `💡 Version ${version} available!`
   if (!packageManager || packageManager === 'unknown') return versionMessage
 
-  const updateCommand = token.packagejsonScript(packageManager, 'shopify', 'upgrade')
+  const updateCommand = token.packagejsonScript(packageManager, 'shopify upgrade')
```
gonzaloriestra, line 360: This was showing something like `Run npm run shopify -- upgrade` for npm.

pepicrft: If I'm not wrong I think that's necessary with `npm` because otherwise it treats those flags and arguments as `npm`'s

gonzaloriestra: Not sure what you mean. In this case, `upgrade` is not a flag, but part of the command, so we don't need the `--` separator. The third argument is used for flags, which do need the separator for npm.

We are doing the same in other places: https://github.com/Shopify/cli/blob/upgrade-cli2/packages/cli-hydrogen/src/cli/services/preview.ts#L23

PR #959, opened 2022-12-20
Force the registry in the deploy environment

pepicrft:
@gonzaloriestra ran an issue deploying a new version of the CLI where changesets fails to push a version of the theme package that already exists in the registry. It happens because the [info](https://github.com/changesets/changesets/blob/e0459f7b2dbb8ebbb58f8e4d859d0cfb3e0eda1e/packages/cli/src/commands/publish/npm-utils.ts#L92) command returns a 404 for that version and package indicating that the version doesn't exist.

It's possible that the deployment environment is configured to point to the Yarn registry, `https://registry.yarnpkg.com`, which might be returning the invalid state due to some caching there. I couldn't find the rationale behind having another registry but it's possible that it's a proxy to the NPM registry to speed up dependency installs.

### Measuring impact

### Checklist

### DIFF
--- a/shipit.production.yml
+++ b/shipit.production.yml
@@ -5,7 +5,7 @@ dependencies:
   override:
     - curl -fsSL https://get.pnpm.io/install.sh | SHELL=`which bash` bash -
     - bash -i -c "pnpm install"
-
+    - echo 'registry "https://registry.npmjs.org/"' | tee .npmrc
 deploy:
   override:
     - bash -i -c "npm_config_loglevel=verbose pnpm changeset publish"


shipit.production.yml
```diff
@@ -5,7 +5,7 @@ dependencies:
   override:
     - curl -fsSL https://get.pnpm.io/install.sh | SHELL=`which bash` bash -
     - bash -i -c "pnpm install"
-
+    - echo 'registry "https://registry.npmjs.org/"' | tee .npmrc
```
gonzaloriestra, line 8: I'm very confused. So we are using pnpm to call changesets, then they are running npm internally, and finally the yarn registry is used? 😄 

According to [this comment](https://github.com/yarnpkg/yarn/issues/2935#issuecomment-355292633), we should update the registry in .yarnrc instead.

pepicrft: The registry configuration is independent from the package manager you use and it can be configured globally or locally. And then there are package managers like Yarn that decide to come up with their own conventions, like having their own `.yarnrc` configuration file 🤦🏼. I'll update that file too just in case.

PR #970, opened 2022-12-21
Optimize the graph load time by importing lodash dynamically

pepicrft:
I noticed we statically imported the whole `lodash-es` even if only we were using some of its utilities. This is not ideal in a CLI because it might end up impacting the startup time.

I refactored the code to import Lodash dynamically. To do so without ending up with all the utilities being `async`, I replaced `lodash-es` with its CommonJS counterpart and exposed the functions from `@shopify/cli-kit`. I also took the opportunity to document the connection between ESM and the startup time, and added a lint that fails CI if the startup time goes above 1s.

**Note** that there might be other third-party dependencies with a module approach similar to `lodash-es` that would benefit from some work. However, I'll leave those for other PRs to not increase the size of this one.

### How to test your changes?
You should be able to interact with the fixture app (build and dev) and tests should pass.

### Measuring impact

### Checklist


packages/cli-kit/src/public/common/array.ts
```diff
@@ -1,3 +1,8 @@
+import {List, ValueIteratee} from 'lodash'
```
isaacroldan, line 1: shouldn't we avoid these kind of imports?

workspace/src/lint.js
```diff
@@ -1,8 +1,11 @@
+#!/usr/bin/env node
+
 import * as path from 'pathe'
 import fg from 'fast-glob'
 import * as url from 'node:url'
 import {promises as fs} from 'node:fs'
 import {createRequire} from 'node:module'
+import {execa} from 'execa'
```
isaacroldan, line 8: is this import necessary? you are not using it right?

workspace/src/generate-benchmark-report.js
```diff
@@ -0,0 +1,68 @@
...
+const rows = []
+for (const command of Object.keys(benchmark).sort()) {
+  if (baselineBenchmark[command] === undefined) {
+    rows.push(['⚪️', `\`${command}\``, 'N/A', `${benchmark[command]} ms`, 'N/A'])
+  } else {
+    const diff = (benchmark[command] / baselineBenchmark[command] - 1) * 100
+    let icon = '⚪️'
+    if (diff <= 0) {
+      icon = '🟢'
+    } else if (diff < 2) {
```
isaacroldan, line 49: is this `2` a % difference?

workspace/src/benchmark.js
```diff
@@ -0,0 +1,48 @@
...
+const require = createRequire(import.meta.url)
+const colors = require('ansi-colors')
+
+const rootDirectory = path.join(url.fileURLToPath(new URL('.', import.meta.url)), '../..')
+const workspaceDirectory = path.join(rootDirectory, 'workspace')
+
+/**
+ * LINT 2 - Startup time
+ * ----
+ * This lint runs all the CLI commands and fails if the startup time for any is above 1s
```
isaacroldan, line 24: I guess this comment is not true anymore right?

isaacroldan (approved):

PR #979, opened 2022-12-23
Fix error handling on session refresh

gonzaloriestra:

When the session refresh fails, we are showing this ugly error instead of running the full authentication flow again:

```
╭─ error ───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│  To investigate the issue, examine this stack trace:                                                                                                      │
│  at tokenRequestErrorHandler (cli-kit/src/session/exchange.ts:175)                                                                                        │
│   return new InvalidGrantError()                                                                                                                          │
│  at mapError (cli-kit/src/public/node/result.ts:143)                                                                                                      │
│   return err(mapper(this.error))                                                                                                                          │
│  at refreshAccessToken (cli-kit/src/session/exchange.ts:85)                                                                                               │
│   const value = tokenResult.mapError(tokenRequestErrorHandler).valueOrBug()                                                                               │
│  at processTicksAndRejections (node:internal/process/task_queues:95)                                                                                      │
│  at async refreshTokens (cli-kit/src/session.ts:336)                                                                                                      │
│   const identityToken = await refreshAccessToken(token)                                                                                                   │
│  at async ensureAuthenticated (cli-kit/src/session.ts:214)                                                                                                │
│   newSession = await refreshTokens(fqdnSession.identity, applications, fqdn)                                                                              │
│  at ensureAuthenticatedPartners (cli-kit/src/session.ts:106)                                                                                              │
│   const tokens = await ensureAuthenticated({partnersApi: {scopes}})                                                                                       │
│  at async dev (app/src/cli/services/dev.ts:52)                                                                                                            │
│   const token = await session.ensureAuthenticatedPartners()                                                                                               │
│  at run (app/src/cli/commands/app/dev.ts:97)                                                                                                              │
│   await dev({                                                                                                                                             │

``` 

Since https://github.com/Shopify/cli/pull/950 we are trying to refresh the tokens in more situations. But apparently we were not handling the errors properly, and [there's a peak of these errors since the release](https://app.bugsnag.com/shopify/cli/errors/6332f63be4c6b00008465919?filters[event.since]=30d&filters[error.status]=open).

The problem is that the [tokenRequestErrorHandler](https://github.com/Shopify/cli/blob/48f8494e188c85d7b4c8f732a430793b2329bd9f/packages/cli-kit/src/session/exchange.ts#L170) function is async, but we are not waiting for the result [here](https://github.com/Shopify/cli/blob/48f8494e188c85d7b4c8f732a430793b2329bd9f/packages/cli-kit/src/session/exchange.ts#L85), so the `ensureAuthenticated` method is finally getting a Promise instead of a `InvalidGrantError` [here](https://github.com/Shopify/cli/blob/48f8494e188c85d7b4c8f732a430793b2329bd9f/packages/cli-kit/src/session.ts#L215).

### Measuring impact

### Checklist

### DIFF
--- a/packages/cli-kit/src/session.ts
+++ b/packages/cli-kit/src/session.ts
@@ -11,6 +11,7 @@ import {
   ExchangeScopes,
   refreshAccessToken,
   InvalidGrantError,
+  InvalidRequestError,
 } from './session/exchange.js'
 
 import {content, token, debug} from './output.js'
@@ -26,6 +27,7 @@ import {partners} from './api.js'
 import {RequestClientError} from './api/common.js'
 import {firstPartyDev, useDeviceAuth} from './environment/local.js'
 import {pollForDeviceAuthorization, requestDeviceAuthorization} from './session/device-authorization.js'
+import {AbortError} from './public/node/error.js'
 import {gql} from 'graphql-request'
 
 const NoSessionError = new Bug('No session found after ensuring authenticated')
@@ -215,6 +217,9 @@ ${token.json(applications)}
     } catch (error) {
       if (error instanceof InvalidGrantError) {
         newSession = await executeCompleteFlow(applications, fqdn)
+      } else if (error instanceof InvalidRequestError) {
+        await secureStore.remove()
+        throw new AbortError('\nError validating auth session', "We've cleared the current session, please try again")
       } else {
         throw error
       }
--- a/packages/cli-kit/src/session/exchange.ts
+++ b/packages/cli-kit/src/session/exchange.ts
@@ -1,8 +1,6 @@
 import {ApplicationToken, IdentityToken} from './schema.js'
 import {applicationId, clientId as getIdentityClientId} from './identity.js'
 import {CodeAuthResult} from './authorize.js'
-import * as secureStore from './store.js'
-import {Abort} from '../error.js'
 import {API} from '../network/api.js'
 import {identity as identityFqdn} from '../environment/fqdn.js'
 import {shopifyFetch} from '../http.js'
@@ -10,9 +8,7 @@ import {err, ok, Result} from '../public/node/result.js'
 import {AbortError} from '../public/node/error.js'
 
 export class InvalidGrantError extends Error {}
-
-const InvalidIdentityError = () =>
-  new Abort('\nError validating auth session', "We've cleared the current session, please try again")
+export class InvalidRequestError extends Error {}
 
 export interface ExchangeScopes {
   admin: string[]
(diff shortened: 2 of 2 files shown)


packages/cli-kit/src/session.ts
```diff
@@ -215,6 +217,9 @@ ${token.json(applications)}
     } catch (error) {
       if (error instanceof InvalidGrantError) {
         newSession = await executeCompleteFlow(applications, fqdn)
+      } else if (error instanceof InvalidRequestError) {
+        await secureStore.remove()
+        throw new AbortError('\nError validating auth session', "We've cleared the current session, please try again")
```
gonzaloriestra, line 222: I've tried to touch as little as possible, but shouldn't we just execute the complete flow for these two cases? Why removing the session and asking the user to try again instead of just doing it?

packages/cli-kit/src/session/exchange.ts
```diff
@@ -74,7 +70,7 @@ export async function exchangeAccessForApplicationTokens(
  * Given an expired access token, refresh it to get a new one.
  */
 export async function refreshAccessToken(currentToken: IdentityToken): Promise<IdentityToken> {
-  const clientId = await getIdentityClientId()
+  const clientId = getIdentityClientId()
```
gonzaloriestra, line 73: This function is not asnyc 🤷

packages/cli-kit/src/session/exchange.ts
```diff
@@ -1,18 +1,14 @@
...
 import {API} from '../network/api.js'
 import {identity as identityFqdn} from '../environment/fqdn.js'
 import {shopifyFetch} from '../http.js'
 import {err, ok, Result} from '../public/node/result.js'
 import {AbortError} from '../public/node/error.js'
 
 export class InvalidGrantError extends Error {}
-
-const InvalidIdentityError = () =>
-  new Abort('\nError validating auth session', "We've cleared the current session, please try again")
```
gonzaloriestra, line 15: Unused

packages/cli-kit/src/session.ts
isaacroldan: yeah, probably it makes sense to remove the session and execute the full flow in both cases.
But for now we can leave it like this and improve in a couple of weeks :)

pepicrft, line 222: What about retrying on behalf of users? As a user of a tool, I'd expect to see no information about the internals of the tool (e.g., not telling them that some internal validation failed), and ideally self-healing such that I don't have to do something that the tool could have done for me.

pepicrft (approved):

packages/cli-kit/src/session.ts
gonzaloriestra: This is what we were talking here: https://github.com/Shopify/cli/pull/979#discussion_r1056262144

PR #997, opened 2023-01-03
Update dev console actions

byrichardpowell:

We want to match the designs [here](https://www.figma.com/proto/wqkgzilsd4RqqGvpt61BzL/Dev-console-%2F-full-page-index%2C-admin-index?node-id=780%3A29096&scaling=min-zoom&page-id=607%3A25444&starting-point-node-id=780%3A29294&show-proto-sidebar=1). This is the first step:

Screenshot 2023-01-03 at 4 01 35 PM

First we updates styles to match the designs a little better.  There is still work to do here, but I'm not 100% sure what is "right" as the designs might be WIP still.  I'll further style improvements in future PR's, once I've chatted with Meredith.

Then we change the columns to match the designs:

1. Add the preview link.
2. Add the view mobile button.
3. Add the Show Live/Local column.  
4. Remove the Actions column since every action except for refresh has moved to other columns (Refresh is removed entirely)

As part of this we remove Polaris Button and ButtonGroup in favor of a custom component.  We do this because Polaris is designed for admin, it's styles don't match our needs and theres only hacky ways of updated those styles.  I chatted with the Polaris team and they agreed it would be best to move away from Polaris.  Once this project has finished I'd like to have removed Polaris entirely. This is the first step.

### How to test your changes?

1. `pnpm install && pnpm build`
2. `cd fixtures/app`
3. `pnpm run dev --skip-dependencies-installation`
4. Open the dev console URL, e.g: https://ngrok.url/extensions/dev-console
5. Test that preview links work
6. Test that the QR Code works
7.  Test that local vs live development works.

### Post-release steps
None

### Measuring impact

### Checklist


packages/ui-extensions-dev-console/src/theme.module.css
```diff
@@ -1,16 +1,30 @@
 .Theme {
-  /* colors */
```
byrichardpowell, line 2: I plan to review all these colors with @MeredithCastile once we have all the functionality in place

PR #998, opened 2023-01-04
Use app host as frontend url in spin

alvaro-shopify:
When a new Shopify app (either 1P or not) spin instance is created  it automatically creates an `nginx` rule to redirect every host `<app-repo-name>.<spin-fqdn>` request to a reserved port in which the app is going to be listening to.
Despite a specific fixed rule was created  for the `cli` in the `partners constellation`: 
https://github.com/Shopify/partners/blob/f414cb03d558b301f1a2b563ed493974526ccc36/.spin/nginx.conf.erb#L49:L72
using it has some disadvantages:
- `partners` constellation should be included, even if the app does not need it to be tested
- 1P apps are automatically configured to use the `<app-repo-name>.<spin-fqdn>` so in case they want to use the CLI they should update that configuration
- Only one app can be run inside a spin instance using the CLI

Anyway, the `partners` rule should not be removed because it can be helpful in the following scenario:
- Create an spin instance directly from `partners` constellation
- Create a temporary app using the [CLI](https://shopify.dev/apps/getting-started/create#step-1-create-a-new-app)
- Run the `dev` command for the newly created app
In that case the `partners` nginx rule redirects the traffic to the service run by the CLI `dev` command

- In case are set the env variables reserved by spin for apps with constellations, they will be used as host and port of the service
- If not, the `partners nginx rule` will be used 

### How to test your changes?
Only the second scenario from the ones defined in the previous section can be tested:
- Create a spin instance using partners with agencies (creates a complete usable development store in spin): `spin up partners:agencies`
- Clone `cli` git repo and checkout the branch associated to this PR `feature/use-app-host-in-spin`
- Create an app and run `dev` command from the `cli` folder using the flag `--path`
- The app should be able to be accesible using a url that matches `cli.<spin-fdn>` host

### Measuring impact

### Checklist

### DIFF
--- a/packages/app/src/cli/services/dev/urls.ts
+++ b/packages/app/src/cli/services/dev/urls.ts
@@ -68,8 +68,11 @@ export async function generateFrontendURL(options: FrontendURLOptions): Promise<
   }
 
   if (isSpin() && !options.tunnelUrl) {
-    frontendPort = appPort() ?? frontendPort
-    frontendUrl = appHost() ? `https://${appHost()}` : `https://cli.${await spinFqdn()}`
+    frontendUrl = `https://cli.${await spinFqdn()}`
+    if (appPort() !== undefined) {
+      frontendPort = appPort() ?? frontendPort
+      frontendUrl = `https://${appHost()}`
+    }
     return {frontendUrl, frontendPort, usingLocalhost}
   }
 
--- a/packages/app/src/cli/services/dev/urls.test.ts
+++ b/packages/app/src/cli/services/dev/urls.test.ts
@@ -524,7 +524,7 @@ describe('generateFrontendURL', () => {
       usingLocalhost: false,
     })
     expect(setAppInfo).not.toBeCalled()
-    expect(prompt).not.toBeCalled()
+    expect(renderSelectPrompt).not.toBeCalled()
   })
 
   it('Returns a custom tunnel url if we are in a spin environment but a custom tunnel option is active', async () => {
--- a/packages/cli-kit/src/public/node/context/spin.test.ts
+++ b/packages/cli-kit/src/public/node/context/spin.test.ts
@@ -1,4 +1,4 @@
-import {show, isSpin, spinFqdn, instance, isSpinEnvironment} from './spin.js'
+import {show, isSpin, spinFqdn, instance, isSpinEnvironment, appPort, appHost} from './spin.js'
 import {getCachedSpinFqdn, setCachedSpinFqdn} from '../../../private/node/context/spin-cache.js'
 import {captureOutput} from '../system.js'
 import {describe, test, expect, vi, it} from 'vitest'


packages/app/src/cli/services/dev/urls.ts
```diff
@@ -61,7 +61,10 @@ export async function generateFrontendURL(options: FrontendURLOptions): Promise<
   }
 
   if (environment.spin.isSpin() && !options.tunnelUrl) {
-    frontendUrl = `https://cli.${await environment.spin.fqdn()}`
+    frontendPort = environment.spin.appPort() ?? frontendPort
+    frontendUrl = environment.spin.appHost()
+      ? `https://${environment.spin.appHost()}`
+      : `https://cli.${await environment.spin.fqdn()}`
     return {frontendUrl, frontendPort, usingLocalhost}
   }
```
pepicrft, line 69: Have we considered providing an interface that can be used not only by the Spin setup but also by 3P developers?

```
--tunnel-url  <-- We mark it as deprecated
--url ${SPIN_APP_HOST} --port ${SPIN_APP_PORT}
```

packages/cli-kit/src/environment/spin.ts
```diff
@@ -106,3 +102,22 @@ export function namespace(env = process.env): string | undefined {
...
 }
+
+/**
+ * Returns the value of the SERVER_PORT environment variable.
+ * @param env - Environment variables
+ * @returns The value of the SERVER_PORT environment variable.
+ */
+export function appPort(env = process.env): number | undefined {
+  const port = Number(env[constants.environmentVariables.spinAppPort])
+  return (!isNaN(port) && port) || undefined
```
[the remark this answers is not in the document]
alvaro-shopify: In case `spinAppPort` is malformed `Number` will return `Nan` which is a number itself. We want to return `undefined` in that case

packages/cli-kit/src/environment/spin.ts
```diff
@@ -106,3 +102,22 @@ export function namespace(env = process.env): string | undefined {
...
+}
+
+/**
+ * Returns the value of the SPIN_APP_HOST environment variable.
```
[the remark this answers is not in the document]
alvaro-shopify: It's an existing `spin convention`

packages/app/src/cli/services/dev/urls.ts
alvaro-shopify: But in case we are inside a `spin environment` we want to apply this rule automatically, right?

packages/cli-kit/src/environment/spin.ts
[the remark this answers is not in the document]
alvaro-shopify: There is a test for this requirement
https://github.com/Shopify/cli/blob/6c00acb8a132f9cf81ddf7c87f5a417f1955911c/packages/cli-kit/src/environment/spin.test.ts#L237:L247

packages/app/src/cli/services/dev/urls.ts
pepicrft: Applying things automatically is what we've been doing so your work aligns with it. The problem that I see with continuing with that approach is that 1P would benefit from an interface for customizing these things, but they can't because we are coupling some internals to our Spin use case. In other words, if we think more broadly, we'll end up with a CLI that's more flexible and can work in more scenarios.

MeredithCastile: Hmmm, what scenarios would this unlock, @pepicrft? I'm with you in theory, but I need help understanding the impact/effort here. cc: @stephszeto

alvaro-shopify:
> @alvaro-shopify is this still relevant?

sorry for the delay. As we eventually removed the `non-interactive` mode for the `dev` command, now the users should run manually in `spin` the command `yarn dev` to run the 1P app. Besides the `server` init command should be removed from the `Procfile` of the constellation config.
As a result of the latest, the `spin` system does not assign the opened port in `nginx` for 1P apps so there is no way to access the CLI using that 1P app host. 
This PR , in case it detects that the cli is running automatically (the SERVER_PORT is assigned) it uses the 1P app host otherwise it will use the fixed `cli` host created in `partners` nginx config

pepicrft (approved):

PR #1000, opened 2023-01-04
Add support in spin to set the flag store using only the store name

alvaro-shopify:
The flag `--store` available when running the command `dev` allows the user to set the store used to install the app. The final requirement to install the app in the store is setting the complete `store hostname`.
The CLI supports either setting the  hostname or only the store name. In case the second option is used the CLI complete it with the domain to build the hostname. However this feature is not supported when the CLI is run against a spin instance.

- When the user only sets the store name for the flag the CLI checks the current environment and complete the store hostname properly

### How to test your changes?
- Run the `dev` command against a spin instance and use the flag `--store <only-store-name>`

### Measuring impact

### Checklist


matteodepalo (approved):

packages/cli-kit/src/string.ts
```diff
@@ -39,15 +40,18 @@ export function capitalize(string: string) {
 }
 
 /**
- * Given a store, returns a valid store fqdn removing protocol and adding .myshopify.com domain
+ * Given a store, returns a valid store fqdn removing protocol and adding the proper domain in case is missing
  * @param store - Original store name provided by the user
  * @returns a valid store fqdn
  */
 export function normalizeStoreName(store: string) {
```
amcaplan, line 47: It feels like `normalizeStoreName` is moving beyond a `string` operation. Is there a better place for this to live?

packages/cli-kit/src/session.ts
```diff
@@ -190,8 +190,12 @@ export async function ensureAuthenticated(
...
   const fqdn = await identityFqdn()
 
-  if (applications.adminApi?.storeFqdn) {
-    applications.adminApi.storeFqdn = normalizeStoreName(applications.adminApi.storeFqdn)
+  const previousStoreFqdn = applications.adminApi?.storeFqdn
+  if (previousStoreFqdn) {
+    const normalizedStoreName = await normalizeStoreName(previousStoreFqdn)
+    if (previousStoreFqdn === applications.adminApi?.storeFqdn) {
+      applications.adminApi.storeFqdn = normalizedStoreName
+    }
```
isaacroldan, line 198: do you need this `if`? it should be ok to always update the storeFqdn no?

packages/cli-kit/src/string.ts
isaacroldan: agree

isaacroldan (approved):

packages/cli-kit/src/session.test.ts
```diff
@@ -111,7 +110,7 @@ beforeAll(() => {
 })
 
 beforeEach(() => {
-  vi.mocked(identity).mockResolvedValue(fqdn)
+  vi.spyOn(fqdnModule, 'identity').mockResolvedValue(fqdn)
```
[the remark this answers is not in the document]
alvaro-shopify: I have moved the the method `normalizeStoreName` to the `fqdn` module. Previously in these tests the complet `fqdn` module was mocked so some tests were failing. 
I think that when you only need to mock one method from a module, it's better to do it in this ways so no collateral effects arise. 
In other tests we are using a different alternative:
```js
vi.mock('./environment/fqdn', async () => {
  const fqdn: any = await vi.importActual('./environment/fqdn')
  return {
    ...fqdn,
    identity: vi.fn(),
    },
  }
})
vi.mocked(identity).mockResolvedValue(fqdn)
```
but I think using the `spyOn` is less cumbersome.
@isaacroldan  what do you think about this?

PR #1025, opened 2023-01-05
[Feature] Add `api_version` to UI extension type

lemonmade:

We are working on adding [API versioning](https://github.com/Shopify/ui-extensions-private/discussions/1931) to checkout UI extensions, and are planning to do it as part of migrating to the new UI extension type. This PR adds the code to the CLI to have it pass the API version from a developer's extension configuration file, to the dev server and deploy commands.

In this PR, I've included the four versions we are likely to care about for the initial release: `2023-01` (the first real version), `2022-10` (a backfilled version that gets developers on the new UI extension, without the breaking API changes introduced in `2023-01`), `unstable` (the most recent public version of the API), and `internal` (the most recent private version of the API). There is more detail about the proposed versioning flow in [this discussion](https://github.com/Shopify/ui-extensions-private/discussions/1806).

### How to test your changes?

Add an `api_version` field to a `ui_extension`. Run the `app dev` command locally to see the extension dev server passing this information to the dev command:

```
pnpm shopify app dev --path /path/to/my/extension
```

Screenshot 2023-01-05 at 3 51 25 PM

If you try to deploy this extension, it seems to work, but there is a validation error because I have not yet added handling of this new field on the backend (working on it now):

Screenshot 2023-01-05 at 4 22 11 PM

### Measuring impact

### Checklist

### DIFF
--- a/packages/app/src/cli/models/extensions/schemas.ts
+++ b/packages/app/src/cli/models/extensions/schemas.ts
@@ -27,11 +27,15 @@ export const NewExtensionPointSchema = schema.object({
 export const OldExtensionPointsSchema = schema.array(schema.string()).default([])
 export const NewExtensionPointsSchema = schema.array(NewExtensionPointSchema)
 export const ExtensionPointSchema = schema.union([OldExtensionPointsSchema, NewExtensionPointsSchema])
+export const ApiVersionSchema = schema.string()
+
+export type ApiVersionSchemaType = schema.infer<typeof ApiVersionSchema>
 
 export const BaseUIExtensionSchema = schema.object({
   name: schema.string(),
   description: schema.string().optional(),
   type: schema.string().default('ui_extension'),
+  apiVersion: ApiVersionSchema.optional(),
   extensionPoints: schema.any().optional(),
   capabilities: CapabilitiesSchema.optional(),
   metafields: schema.array(MetafieldSchema).optional().default([]),
--- a/.changeset/fast-ants-invite.md
+++ b/.changeset/fast-ants-invite.md
@@ -1,0 +1,5 @@
+---
+'@shopify/cli': minor
+---
+
+Enable configuring api_version for ui_extension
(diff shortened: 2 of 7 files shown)


packages/app/src/cli/models/extensions/schemas.ts
```diff
@@ -27,10 +27,19 @@ export const NewExtensionPointSchema = schema.define.object({
...
+  schema.define.literal('unstable'),
+  schema.define.literal('internal'),
+])
+
+export type ApiVersionSchemaType = schema.define.infer<typeof ApiVersionSchema>
 
 export const BaseUIExtensionSchema = schema.define.object({
   name: schema.define.string(),
   type: schema.define.string().default('ui_extension'),
+  apiVersion: ApiVersionSchema.optional(),
```
byrichardpowell, line 42: Adding this here adds is as an optional field to all UI extensions.   I assume not all UI extensions would support this, so is this what we want? Instead, should we be adding it to just the `ui_extension` here? https://github.com/Shopify/cli/blob/add-app-home/packages/app/src/cli/models/extensions/ui-specifications/ui_extension.ts/#L11-L14

Long term it might not matter since all extensions will be moving to the `ui_extension`

byrichardpowell (commented):
Just the one question

packages/app/src/cli/models/extensions/schemas.ts
lemonmade: I did add the field to that schema, too, since it is used to parse the config during publishing. However, during development, in order to add that field on the dev server payload (https://github.com/Shopify/cli/blob/df63e49dd9a9c8763277b1467b6fdac65da65df8/packages/app/src/cli/services/dev/extension/payload.ts#L64), I had to add it here, too — in that context, the extension's specification aren't currently available to narrow the type better.

I found this whole thing to be pretty confusing to follow tbh, but I just went with it since a number of other fields on `BaseUIExtensionSchema` are also only relevant for a subset of extension types (`extensionPoints`, `capabilities`, `metafields`, `categories`).

isaacroldan (approved):
Minor comment and 👍

marvinhagemeister (approved):
Just tried out this PR again locally and can see the version in the network tab. Was also able to publish an app with that so it seems like some work has been already done on that side too 👍

PR #1030, opened 2023-01-06
Align `@types/node` with the minimum supported Node version and remove `abort-controller`.

pepicrft:
Related: https://github.com/Shopify/internal-cli-foundations/issues/491

I noticed that the version of the `@nodejs/types` package doesn't align with the minimum NodeJS version that we support. This might lead to code that succeeds type-checking and running in our environments (NodeJS 18/19) and that fails to run in older versions of Node.

I'm aligning the version of `@nodejs/types` with the minimum supported version of NodeJS such that if we use an interface that's only available in newer versions of Node Typescript will catch the issue. I've also taken the opportunity to remove `abort-controller` since [`AbortController`](https://nodejs.org/api/globals.html#class-abortcontroller) is present in the Node runtime since Node 14.

### How to test your changes?
Try to run the CLI against an app and theme project.

### Measuring impact

### Checklist


packages/cli-kit/src/path.ts
```diff
@@ -3,6 +3,8 @@ import commondir from 'commondir'
 import {relative, dirname, join, normalize, resolve, basename, extname, isAbsolute, parse} from 'pathe'
 import {findUp as internalFindUp} from 'find-up'
 import {fileURLToPath} from 'url'
+// eslint-disable-next-line node/prefer-global/url
+import type {URL} from 'url'
```
amcaplan, line 7: Why not just delete this line?

packages/cli-kit/src/private/node/ui/components/TextAnimation.tsx
```diff
@@ -37,7 +37,7 @@ const TextAnimation: React.FC<Props> = ({text}): JSX.Element => {
     renderAnimation()
 
     return () => {
-      clearTimeout(timeout.current)
+      clearTimeout(timeout.current as unknown as number)
```
amcaplan, line 40: Curious what changed

packages/cli-kit/src/public/node/node-package-manager.ts
```diff
@@ -524,7 +523,7 @@ async function getLatestNPMPackageVersion(name: string) {
 }
 
 interface FindPackageVersionUpOptions {
-  fromModuleURL: URL | string
+  fromModuleURL: import('url').URL | string
```
amcaplan, line 526: Curious why this is necessary

packages/cli-kit/src/path.ts
pepicrft: Because the older version of `@types/node` didn't expose `URL` type globally and therefore you are required to import it.

packages/cli-kit/src/public/node/node-package-manager.ts
pepicrft: Same as in another comment the version of `@types/node` that we are using now doesn't expose the URL type globally so we need to import it. Note that we are importing the type so it doesn't have any impact at runtime.

packages/cli-kit/src/private/node/ui/components/TextAnimation.tsx
pepicrft: And same here,  the version `@types/node` that we are using now has the types for timeouts wrongly defined so I'm doing type erasing.

PR #1032, opened 2023-01-06
Fix issue with prompt not being cleared correctly

matteodepalo:

I accidentally found a bug with prompts not being cleared properly after submission if the height of the terminal is lower than the height of the prompt itself.

Current implementation doesn't clean the prompt after submission and keeps the last 2 renders in the terminal output:

New implementation:

### How to test your changes?

Use kitchen sink to test prompts at different terminal heights.

### Measuring impact

### Checklist

### DIFF
--- a/packages/cli-kit/src/private/node/ui/components/FullScreen.tsx
+++ b/packages/cli-kit/src/private/node/ui/components/FullScreen.tsx
@@ -8,31 +8,28 @@ import React, {useEffect, useState} from 'react'
  */
 const FullScreen: React.FC = ({children}): JSX.Element => {
   const {stdout} = useStdout()
-
-  if (!stdout) {
-    throw new Error('Could not find stdout')
-  }
+  const standardOutput = stdout!
 
   const [size, setSize] = useState({
-    columns: stdout.columns,
-    rows: stdout.rows,
+    columns: standardOutput.columns,
+    rows: standardOutput.rows,
   })
 
   useEffect(() => {
     function onResize() {
       setSize({
-        columns: stdout!.columns,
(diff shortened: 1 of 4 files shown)


packages/cli-kit/src/private/node/ui/components/FullScreen.tsx
```diff
@@ -7,26 +7,32 @@ import React, {useEffect, useState} from 'react'
  * - You want to respond to the resize event of the terminal. Whenever the user resizes their terminal window the output's height and width will be recalculated and re-rendered properly.
  */
 const FullScreen: React.FC = ({children}): JSX.Element => {
+  const {stdout} = useStdout()
+
+  if (!stdout) {
+    throw new Error('Could not find stdout')
+  }
```
pepicrft, line 14: When does `!stdout` happen? If we don't know when it happens, what do you think about assuming `stdout` has a value?

pepicrft (approved):

packages/cli-kit/src/private/node/ui/components/FullScreen.tsx
matteodepalo: Yeah I don't foresee any cases in which it doesn't exist. This `if` is just there so TS understand that all calls after this are done to something that is not `undefined`. The alternative would be to do `stdout!.yyy` all the time. Would you prefer another way to accomplish the same thing? Maybe reassigning?

pepicrft: I'd go the reassigning path to eliminate some verbosity.

PR #1038, opened 2023-01-08
Add nightly snapshots

amcaplan:

Provide nightly snapshots to avoid anyone being held up in development waiting for a release.

Based on docs and local testing, this should create a package of the form `0.0.0-nightly-20230108195058` where `20230108...` is a continuously incrementing datetime representing the moment of release.

Since the release is tagged, users can point to e.g. `@shopify/app@nightly`

Adds a new shipit stack that will deploy continuously once in 24 hours.

### How to test your changes?

Enable it, hope it works!

### Post-release steps

Maybe announce it, maybe not...

### Measuring impact

We will see in our analytics which version was used; nightly snapshots use very distinctive versions.

### Checklist

### DIFF
--- a/shipit.nightly.yml
+++ b/shipit.nightly.yml
@@ -1,0 +1,18 @@
+ci:
+  require: []
+
+dependencies:
+  override:
+    # We are making sure the deploys point to the NPM registry to prevent
+    # 404 issues like the one described in this comment:
+    # https://github.com/yarnpkg/yarn/issues/2935#issuecomment-355292633
+    - echo 'registry "https://registry.npmjs.org/"' | tee .npmrc .yarnrc
+    - curl -fsSL https://get.pnpm.io/install.sh | SHELL=`which bash` bash -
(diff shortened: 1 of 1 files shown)


gonzaloriestra (approved):
Cool! This can be useful for testing changes that may behave differently on development, and also asking users with a problem to test a fix.

But taking into account that we release at least once per week, I'd wouldn't announce it publicly for now.

pepicrft:
@amcaplan what do you think about extending the `create-x` CLIs to support a `--nightly` flag?

amcaplan:
@gonzaloriestra I agree, this is a quiet feature and we are just trying to surface issues and catch some kinds of bugs that only arise when releases happen, as well as unblocking development processes (e.g. new extension types) by making it easy to point a project to nightly instead of having to point to a local copy or wait for a release to happen.

@pepicrft what would that flag do? I can imagine a couple of meanings, so maybe you can explain what you had in mind?

PR #1040, opened 2023-01-09
Adopt the CLI UI kit in the `shopify theme list` command

Poitrin:

Fixes:
* https://github.com/Shopify/internal-cli-foundations/issues/498

- Displays a table with all themes, following the CLI UI Kit convention
- Calls Admin API’s `/themes` endpoint with query parameter `fields`, to only retrieve fields `id`, `name`, `role`, for both `theme open` and `theme list` command
- Introduces `--role`, `--id` and `--name` parameters for `theme list` command, to filter for themes with the specific role/name/ID

### How to test your changes?

- Run `pnpm run shopify theme list`…
  - in dark mode: Verify that text is white/grey on dark background
  - in bright mode: Verify that text is black/grey on bright background
- Run `pnpm run shopify theme list --role=live` and verify that only `live` themes are listed
  - Run `pnpm run shopify theme list --role=ottawa` and verify that error `Expected --role=ottawa to be one of: live, unpublished, development` is displayed
- Given a theme named "My theme 12345", run `pnpm run shopify theme list --name="eMe 123"` and verify that your theme is listed
  - Run `pnpm run shopify theme list --name=(random string)` and verify that error `The … store doesn't have a theme with the "…" ID or name` is displayed
- Run `pnpm run shopify theme list --id=(ID of existing theme)` and verify that your theme is listed
  - Run command with non-existing ID and verify that error `The … store doesn't have a theme with the "…" ID or name` is displayed

Note: Multi-filter is not supported.
E.g.: `role` would rule over `id`, and `id` would rule over `name`.
```
shopify theme list --id=123 --name=a --role=unpublished

name    role           id
Studio  [unpublished]  #456
```

### Post-release steps

Update `shopify theme list` [docs](https://shopify.dev/themes/tools/cli/commands#list) with the new flags.

### Measuring impact

### Checklist


packages/cli-kit/src/private/node/ui/components/Table/Row.tsx
```diff
@@ -0,0 +1,55 @@
+import ScalarDict from './ScalarDict.js'
+import {Column} from './Column.js'
+import {Box, Text} from 'ink'
+import React from 'react'
+
+interface RowProps<T extends ScalarDict> {
+  line: string
```
matteodepalo, line 7: I was a bit confused by the naming of this prop at first. What do you think about renaming it to something like `fillerCharacter`? Or `fillerChar`? Because in case you pass `" "` it's not really a line.

packages/cli-kit/src/private/node/ui/components/Table/Row.tsx
```diff
@@ -0,0 +1,55 @@
...
+        (index) => {
+          const key = `${props.rowKey}-horizontal-separator-${index}`
+          return <Text key={key}>{'  '}</Text>
+        },
+
+        props.columns.map((column) => {
+          const content = props.data[column.name]
+
+          if (content === undefined || content === null) {
+            const key = `${props.rowKey}-empty-${column.name.toString()}`
```
matteodepalo, line 37: Do we need a separate key? I'm thinking that maybe we could remove this early return and apply the same logic to all rows. In this case because `String(content).length` is `0`, then `const marginRight = column.width - String(content).length` would be equal to `column.width` which is what this branch covers. If we removed this branch everything should still be working.

If it doesn't let you use `undefined` below you can always do `{content ?? ''}`

packages/cli-kit/src/private/node/ui/components/Table/Table.tsx
```diff
@@ -0,0 +1,55 @@
...
+    (headings, [column, {header}]) => ({...headings, [column]: header || column}),
+    {},
+  )
+
+  return (
+    <Box flexDirection="column">
+      <Row
+        rowKey="heading"
+        line=" "
+        columns={columns.map((column) => ({...column, color: undefined}))}
```
matteodepalo, line 40: I get that you want to make it explicit that these rows have no color, but perhaps we can save a `map` here by passing `columns` as they are. I think it should be self evident that by default rows have no color. What do you think?

matteodepalo (commented):
Nice! Excited to get this component in the codebase!

packages/cli-kit/src/public/node/ui.tsx
```diff
@@ -237,6 +239,10 @@ export async function renderConfirmation({question, infoTable}: ConfirmationProp
   return prompt({message: question, choices, infoTable})
 }
 
+export function renderTable<T extends ScalarDict>(rows: TableProps<T>['rows'], columns: TableProps<T>['columns']) {
+  return render(<Table rows={rows} columns={columns} />)
```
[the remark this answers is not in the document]
Poitrin: That's a great tip, I will change it :)
(`list` function will still return a promise though, due to `fetchStoreThemes` :)

packages/cli-kit/src/private/node/ui/components/Table/Row.tsx
Poitrin: Agreed, I will change it :)

packages/cli-kit/src/private/node/ui/components/Table/Table.tsx
Poitrin: If I pass `columns` as they are, headings and separators would be colored too, even though they should keep their standard color:
10-02-obz4w-fo51t

With the changes in https://github.com/Shopify/cli/pull/1040#discussion_r1064741876, I decided to introduce the flag `ignoreColumnColor`, instead of `.map`ping twice (for headings and separators)

packages/cli-kit/src/private/node/ui/components/Table/Row.tsx
```diff
@@ -0,0 +1,55 @@
...
+import React from 'react'
+
+interface RowProps<T extends ScalarDict> {
+  line: string
+  rowKey: string
+  data: Partial<T>
+  columns: Column<T>[]
+}
+
+function intersperse<T, TI>(intersperser: (index: number) => TI, elements: T[]): (T | TI)[] {
```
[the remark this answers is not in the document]
Poitrin: It's the pseudo equivalent of `columns.join(separator)`, which isn't feasible in React.
I renamed the function and switched the params order – maybe this might make it clearer?

packages/cli-kit/src/private/node/ui/components/Prompt/Prompt.test.tsx
```diff
@@ -1,5 +1,5 @@
 import Prompt from './Prompt.js'
-import {sendInput, waitForInputsToBeReady} from '../../../../testing/ui.js'
```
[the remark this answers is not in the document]
Poitrin: I noticed that, since our last discussion about "your" `Table` component, "my" `Table` is now in a separate folder, so, if you don't mind, I keep "your" `Table`’s name unchanged, and leave the remaining work to you :)

packages/cli-kit/src/private/node/ui/components/Table/Row.tsx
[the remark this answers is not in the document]
mgmanzella: yeh this change helped a lot -- thanks!

mgmanzella (approved):
🔥 🚀

Poitrin:
_Note:_ I forgot to implement code for the requirement "Introduce the filter by `name`, `role`, and `id`" (see issue).
Therefore, PR is still _work in progress_ 😊

PR #1047, opened 2023-01-09
Fix QRCode link when Surface is Point of Sale

samiulmonir06:

Fixes https://github.com/Shopify/pos-next-react-native/issues/19682

For POS UI Extensions, scanning QRCode from developer console opens Shopify mobile instead of POS.

Fixes the wrong surface name and updating the `retail-ui-extensions` to point to the most recent version,

### How to test your changes?

1. Created a new Extension.
2. Updated Tests

### Post-release steps

N/A

### Measuring impact

### Checklist

### DIFF
--- a/fixtures/app/extensions/pos-ui-extension/src/index.jsx
+++ b/fixtures/app/extensions/pos-ui-extension/src/index.jsx
@@ -1,0 +1,28 @@
+import { extend, Text } from "@shopify/retail-ui-extensions";
+
+extend('Retail::SmartGrid::Tile', (root, api) => {
+  const tileProps = {
+    title: 'My app',
+    subtitle: 'A test app',
+    enabled: true,
+    onPress: () => {
+      api.navigation.navigateToFullScreenModal();
+    }
+  }
+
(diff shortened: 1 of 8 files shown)


fixtures/app/extensions/pos-ui-extension/src/index.jsx
```diff
@@ -0,0 +1,28 @@
+import { extend, Text } from "@shopify/retail-ui-extensions";
+
+extend('Retail::SmartGrid::Tile', (root, api) => {
```
isaacroldan, line 3: did you intend to modify the pos template? this file is the fixtures/app that we use for testing. Is good to keep it updated, just wondering if you also wanted to update the template

samiulmonir06: @isaacroldan: Initially It was not intended to add `pos_ui_extension` into fixtures. However, adding `pos_ui_extension` in fixtures helped me to quickly start an extension and investigate the root cause of the issue we were having. Thats why I kept it here so it can be used for future testing/investigations.

PR #1058, opened 2023-01-11
[cli-kit] Refactor string and crypto

gonzaloriestra:

- Extract string helpers to public/common
- Extract crypto helpers to public/node

### How to test your changes?

Check auth and some random commands

### Measuring impact

### Checklist

### DIFF
--- a/packages/cli-kit/src/public/common/string.ts
+++ b/packages/cli-kit/src/public/common/string.ts
@@ -203,7 +203,7 @@ export {camelCase as camelize} from 'change-case'
 /**
  * Transform a string to param-case.
  */
-export {paramCase as hyphenize} from 'change-case'
+export {paramCase as hyphenate} from 'change-case'
 
 /**
  * Transform a string to snake_case.
--- a/packages/app/src/cli/services/generate/extension.ts
+++ b/packages/app/src/cli/services/generate/extension.ts
@@ -10,7 +10,7 @@ import {
   addResolutionOrOverride,
   DependencyVersion,
 } from '@shopify/cli-kit/node/node-package-manager'
-import {hyphenize} from '@shopify/cli-kit/common/string'
+import {hyphenate} from '@shopify/cli-kit/common/string'
 import {recursiveLiquidTemplateCopy} from '@shopify/cli-kit/node/liquid'
 import {fileURLToPath} from 'url'
 import stream from 'stream'
@@ -217,7 +217,7 @@ async function functionExtensionInit(options: FunctionExtensionInitOptions) {
 }
 
 async function ensureExtensionDirectoryExists({name, app}: {name: string; app: AppInterface}): Promise<string> {
-  const hyphenizedName = hyphenize(name)
+  const hyphenizedName = hyphenate(name)
   const extensionDirectory = path.join(app.directory, blocks.extensions.directoryName, hyphenizedName)
   if (await file.exists(extensionDirectory)) {
     throw new error.Abort(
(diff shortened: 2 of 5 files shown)


packages/cli-kit/src/public/common/string.ts
```diff
@@ -113,3 +114,87 @@ export function getRandomName(family: RandomNameFamily = 'business'): string {
...
+
+/**
+ * Transform a string to camelCase
+ */
+export {camelCase as camelize} from 'change-case'
+
+/**
+ * Transform a string to param-case
+ */
+export {paramCase as hyphenize} from 'change-case'
```
amcaplan, line 190: While `hyphenize` is technically a valid term, the more common word would be `hyphenate`. (For perspective, `hyphenate` generates 48M google results, while `hyphenize` generates just 4k).

I'd prefer we use `hyphenate` as the name, or at the very least alias.

packages/cli-kit/src/public/node/crypto.ts
```diff
@@ -0,0 +1,43 @@
+import crypto from 'crypto'
```
amcaplan, line 1: Most of these methods could be moved to a `private` module, as they're only in use within CLI kit.

The smaller the surface area, the less we are committing to support.

packages/cli-kit/src/public/common/string.ts
```diff
@@ -113,3 +114,87 @@ export function getRandomName(family: RandomNameFamily = 'business'): string {
...
+    .replace(/[^\w\s-]/g, '')
+    .replace(/[\s_-]+/g, '-')
+    .replace(/^-+|-+$/g, '')
+}
+
+/**
+ * Given a string, it returns it with the special regex characters escaped
+ * More info: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions#escaping
+ */
+export function escapeRegExp(str: string): string {
```
pepicrft, line 178: Nitpick: Even though it's common to see functions in Javascript that are not very idiomatic. I think we can use this refactoring opportunity to make ours more idiomatic. Ruby does a better job at this with conventions to indicate whether you are modifying the input argument or returning a new copy using the `!` suffix. What about naming the functions here:

```
import {
  slugifiedString,
  camelCasedString,
  paramCasedString,
  snakeCasedString,
  constantCasedString,
  capitalizedString,
  decodeIntegerFromString
} from "@shopify/cli-kit/common/string"
```

packages/cli-kit/src/public/common/string.ts
```diff
@@ -113,3 +114,87 @@ export function getRandomName(family: RandomNameFamily = 'business'): string {
...
+    }
+  }
+  return asInt
+}
+
+/**
+ * Given a series of rows inside an array, where each row is an array of strings (representing columns)
+ * Parse it into a single string with the columns aligned
+ */
+export function linesToColumns(lines: string[][]): string {
```
pepicrft, line 143: What about having a module, `@shopify/cli-kit/node/terminal`, that exports this function as `buildTableString`.

packages/cli-kit/src/public/node/crypto.ts
gonzaloriestra: So do you suggest to have this?
- public/node/crypto: randomUUID, hashString
- private/node/crypto: randomHex, randomBytes, base64URLEncode, sha256

While I agree on making public as little as possible, all of the functions here are very similar, so I don't see the point of separating them (why should sha1 be public, but sha256 private?). And I think we won't have to support much because it's very simple and unlikely to change.

Do you still think it's worth separating them?

packages/cli-kit/src/public/common/string.ts
gonzaloriestra: Hmm, I'm not sure to see the benefit. In JS the parameters are always passed by value (well, the simple ones at least), so why do we need to make a distinction?

For example, I think `camelize(myString)` is quite clear and simple. I know it will just generate a new string because it can't modify my variable. Also, we are following this style (infinitive) in many other places.

PR #1060, opened 2023-01-11
[cli-kit] Refactor api modules

isaacroldan:
- Create a `graphql.ts` file in the private folder that takes care of creating graphql clients and executing queries
- Create a `headers.ts` file in the private folder that builds headers for rest and graphql queries
- Moved `admin, partners and oxygen` api files to a public folder under `node/api/`, refactored them to simplify them and use the generic client from `graphql.ts`
- Moved `identity.ts` from `api` to `session` and renamed it to `identity-token-validation.ts` as it made more sense
- Fixed all imports and tests 

### How to test your changes?
Everything keeps working 😬 

### Measuring impact

### Checklist


packages/cli-kit/src/public/node/api/admin.ts
```diff
@@ -0,0 +1,62 @@
+import {AdminSession} from '../../../session.js'
+import {content, token as outputToken} from '../../../output.js'
+import {Bug, Abort} from '../../../error.js'
+import {graphqlRequest} from '../../../private/common/api/graphql.js'
+import {ClientError, gql, RequestDocument, Variables} from 'graphql-request'
+
+export async function adminRequest<T>(
+  query: RequestDocument,
```
pepicrft, line 8: What do you think about making `graphql-request` an internal dependency and change this function signature to:

```js
export async function adminGraphQLRequest<T>(
  query: string,
  session: AdminSession,
  variables: object
)
```
Note that I also included `GraphQL` in the name to make it clear that we are sending a GraphQL request.

packages/cli-kit/src/public/node/api/admin.ts
```diff
@@ -0,0 +1,62 @@
+import {AdminSession} from '../../../session.js'
+import {content, token as outputToken} from '../../../output.js'
+import {Bug, Abort} from '../../../error.js'
+import {graphqlRequest} from '../../../private/common/api/graphql.js'
+import {ClientError, gql, RequestDocument, Variables} from 'graphql-request'
+
+export async function adminRequest<T>(
```
pepicrft, line 7: This function is missing documentation.

packages/cli-kit/src/public/node/api/admin.ts
```diff
@@ -0,0 +1,62 @@
...
+  session: AdminSession,
+  variables?: Variables,
+): Promise<T> {
+  const api = 'Admin'
+  const version = await fetchApiVersion(session)
+  const url = adminUrl(session.storeFqdn, version)
+  return graphqlRequest(query, api, url, session.token, variables)
+}
+
+async function fetchApiVersion(session: AdminSession): Promise<string> {
```
pepicrft, line 18: I'd recommend moving all the non-exported elements to `private/node/api/admin.ts`.

packages/cli-kit/src/public/node/api/oxygen.ts
```diff
@@ -1,25 +1,17 @@
...
-import {shopifyFetch} from '../http.js'
+import {shopifyFetch} from '../../../http.js'
+import {graphqlRequest} from '../../../private/common/api/graphql.js'
+import {buildHeaders} from '../../../private/common/api/headers.js'
 import {Variables, RequestDocument} from 'graphql-request'
 import FormData from 'form-data'
 import {Response} from 'node-fetch'
 
-export async function request<T>(
+export async function oxygenRequest<T>(
```
pepicrft, line 8: Similar comment here regarding keeping `graphql-request` as an implementation detail, adding documentation to the function, and moving the non-exported functions to `private/node/api/oxygen`.

packages/cli-kit/src/public/node/api/partners.ts
```diff
@@ -0,0 +1,43 @@
+import {partners as partnersFqdn} from '../../../environment/fqdn.js'
+import {graphqlRequest} from '../../../private/common/api/graphql.js'
+import {ScriptServiceProxyQuery} from '../../../api/graphql/index.js'
+import {Variables, RequestDocument} from 'graphql-request'
+
+export async function partnersRequest<T>(query: RequestDocument, token: string, variables?: Variables): Promise<T> {
```
pepicrft, line 6: And same here.

packages/cli-kit/src/private/common/api/graphql.ts
```diff
@@ -0,0 +1,76 @@
+import {buildHeaders, RequestClientError, sanitizedHeadersOutput} from './headers.js'
```
pepicrft, line 1: All the modules under `private/common/api`  in this PR have transitive modules that depend on Node. Therefore I'd move `grahql` and `headers` under `private/node/api`. In case of doubt, I'd lean on the `node` side unless the module's content is an obvious runtime-agnostic piece of code. For example the lodash-powered functions.

packages/cli-kit/src/public/node/api/oxygen.ts
```diff
@@ -1,25 +1,17 @@
...
-    headers,
-    url: getOxygenAddress(oxygenAddress),
-  })
-
-  const response = await client.request<T>(query, variables)
-  return response
+  return graphqlRequest(query, 'Oxygen', getOxygenAddress(oxygenAddress), token, variables)
 }
 
 export async function uploadDeploymentFile(oxygenAddress: string, token: string, data: FormData): Promise<Response> {
```
pepicrft, line 17: I'd reflect in the name of the function that it belongs to the `oxygen` domain. For example renaming it to `uploadOxygenDeploymentFile`:

```js
import { uploadDeploymentFile } from "@shopify/cli-it/node/api/oxygen"


/** further down **/
uploadDeploymentFile(...) // To where?

// vs
uploadOxygenDeploymentFile(...)
```

pepicrft (approved):
Great job on this one 👏🏼

packages/cli-kit/src/public/node/api/admin.ts
isaacroldan: Yeah, that makes sense, no need to expose graphql-types here 👌

packages/cli-kit/src/public/node/api/admin.ts
isaacroldan: this function is only used by the function above, I think it makes sense to have it here but not exported since it won't be used in any other places, what do you think?

PR #1061, opened 2023-01-11
Compare against a baseline benchmark generated in the same environment

pepicrft:
Comparing benchmarks generated from different CI builds is unreliably because we don't have dedicated hardware. Even with dedicated hardware we might find out that CPU throttling, thermal regulation and other issues can introduce noise in measurements.

Adjust our approach to generate the baseline benchmark in the same environment by cloning `main` in a temporary directory. Moreover, I run each benchmarking 5 times and calculate the average.

### How to test your changes?

We should see the new report in the PR.

### Measuring impact

### Checklist


workspace/src/generate-benchmark-report.js
```diff
@@ -7,13 +7,22 @@ import * as url from 'url'
...
 import {execa} from 'execa'
+import {osMetadata} from './os-metadata.js'
 
 const require = createRequire(import.meta.url)
 const rootDirectory = path.join(url.fileURLToPath(new URL('.', import.meta.url)), '../..')
 
 const baselineBenchmarkPath = path.join(path.join(rootDirectory, 'baseline-benchmark.json'))
 const benchmarkPath = path.join(rootDirectory, 'benchmark.json')
 
+function getOSMarkdownReport(osMetadata) {
```
gonzaloriestra, line 18: What about showing the current CPU/memory usage as well? The load while benchmarking could be different and affect as well, no?

Maybe this helps: https://labs.quansight.org/blog/2021/08/github-actions-benchmarks

gonzaloriestra (approved):

workspace/src/generate-benchmark-report.js
pepicrft: Yeah. I minimized it a bit by running it a few times and obtaining the average. It's much better than what we had before.

PR #1068, opened 2023-01-11
Update Checkout UI Extensions

igor10k:
Bumping up the version


isaacroldan:
You'd need to run pnpm install, the CI is failing because of an outdated lock file :)

PR #1071, opened 2023-01-11
Removed

joshua-arts:
Removed


KevenLYF (approved):
Great to see we can deploy constraint function to spin instance!

isaacroldan (approved):
Removed

joshua-arts:
> This function will only be available for shopifolk for now (because gated: true)

Yep!

> but you are sure you want this code to be public already right?

Hmm... @KevenLYF is there an issue with revealing fulfillment constraints to the public at this point? If so, we could keep this branch un-merged for now and just make sure we have it checked out to deploy constraint functions.

PR #1089, opened 2023-01-13
[cli-kit] Refactor version

gonzaloriestra:

We need to read the CLI version many times per command (up to 10 for dev), because it's used for API calls, local storage, analytics... And every time we need it, we are currently looking for the closest package.json in the file system, reading it and extracting the version.

This is probably not affecting the performance too much, but: 
- It makes the code more complex because it requires many operations to be async
- It adds a lot of noise to the verbose output (`Reading the content of file at packages/cli-kit/package.json...`)

- Extract the version from the package.json at release time and save it as a constant in cli-kit
- Make a lot of functions sync
- Use the new module from everywhere

### How to test your changes?

CI, `pnpm prerelease`

### Measuring impact

### Checklist


packages/create-hydrogen/src/services/init.ts
```diff
@@ -20,7 +18,6 @@ interface InitOptions {
   directory: string
   packageManager?: string
   shopifyCliVersion?: string
-  cliHydrogenPackageVersion?: string
```
gonzaloriestra, line 23: Unused

gonzaloriestra:
@Arkham done! You can now try with `pnpm prerelease`. And it will be fired only as part of the release process (the next Version Packages PR should contain a version.ts update).

Arkham (approved):
Sweet, LGTM!

PR #1106, opened 2023-01-16
Use the ESM version of `pathe`

pepicrft:
Using ESM dependencies improves the load time of the CLI because the loading of the dependencies' modules can be parallelized with the loading of the CLI modules.

I noticed there's a version of [pathe](https://github.com/unjs/pathe) that supports ESM so I'm updating `pathe` to that version.

### How to test your changes?
A lot of business and tests logic goes through the `path` module so we should be confident that CI will raise any incompatibilities at runtime.

### Measuring impact

### Checklist


packages/cli-kit/src/path.ts
```diff
@@ -37,9 +37,11 @@ export async function findUp(
...
  */
-export function relativize(path: string): string {
-  const result = commondir([path, process.cwd()])
-  const relativePath = relative(process.cwd(), path)
+export function relativize(path: string, cwd: string = process.cwd()): string {
+  const result = commondir([path, cwd])
+  const relativePath = relative(cwd, path)
+  // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+  // @ts-ignore
   const relativeComponents = relativePath.split('/').filter((component) => component === '..').length
```
pepicrft, line 45: `pathe` imports Node modules using the `node:....` module ID, which is not recognized by the version 
 of `@types/node` that we are using and therefore Typescript recognizes some of the types as `any`. This mitigates the issue until we can drop support for Node 14 and update `@types/node`.

packages/cli-kit/src/path.test.ts
```diff
@@ -25,3 +25,17 @@ describe('path-glob', () => {
     expect(FastGlob).toBeCalledWith('pattern', {dot: false})
   })
 })
+
+describe('relativize', () => {
+  test('relativizes the path', () => {
```
pepicrft, line 30: I added a test to ensure `relativize` does return what we expect.

packages/app/src/cli/services/info.test.ts
```diff
@@ -238,9 +238,8 @@ describe('info', () => {
 
     // Then
     expect(result).toContain('Extensions with errors')
-    expect(result).toContain('📂 ui_extension    tmp/project/extensions/test-ui-extension')
```
pepicrft, line 241: The fixing of the logic to let `output` relativize the path to the working directory leads to the output content being dependent on the temporary directory created for this test case. Since the `relativize` function is already tested with unit test, I refrained from stubbing `cwd`.

isaacroldan (approved):

PR #1112, opened 2023-01-16
Split kitchen sink

matteodepalo:

The `kitchen-sink` command was starting to become too big.

I've split the command into three subcomands:
- `prompts` displays just prompts such as selects and text inputs
- `banners` displays only the banner components
- `async` displays the two components that deal with async tasks: `Tasks` and `ConcurrentOutput`.

### How to test your changes?

- Run `shopify kitchen-sink all` to see the previous behavior
- Run `shopify kitchen-sink prompts` to see just prompts
- Run `shopify kitchen-sink banners` to see just banners
- Run `shopify kitchen-sink async` to see just async stuff

### Measuring impact

### Checklist

### DIFF
--- a/packages/cli/src/cli/services/kitchen-sink/prompts.ts
+++ b/packages/cli/src/cli/services/kitchen-sink/prompts.ts
@@ -23,6 +23,9 @@ export async function prompts() {
   await renderTextPrompt({
     message: 'App project name (can be changed later)',
     defaultValue: 'expansive commerce app',
+    validate: (value) => {
+      if (value.includes('shopify')) return 'Can\'t include "shopify" in the name'
+    },
   })
 
   const database = [


isaacroldan (approved):

matteodepalo:
Actually about the first point, I found a way to hide the topic but not the subcommands so you can run `pnpm shopify kitchen-sink --help` and get a full list of commands.

matteodepalo:
>Good idea! Is there any quick way to do this with a topic? I tried looking in the oclif docs but I didn't find anything related to this. Maybe a index.ts command?

I found that I can do `index.ts` and use `aliases = ['kitchen-sink all']` to achieve this.

gonzaloriestra:
Regarding the default option, I don't know if there is an easier way, but we could achieve it by using flags instead of subcommands, like `kitchen-sink`, `kitchen-sink --prompts`...

About the name, it's probably my problem because I know is a common term, but I always have problems to remember it 😅

gonzaloriestra (approved):
Thanks for the changes!

PR #1125, opened 2023-01-17
Themes cli migration

alvaro-shopify:
CLI2 is going to be deprecated in some weeks. Although the team is working on the migration of the `themes` commands from Ruby to Typescript, a fallback is needed in case those task don't finish before the deadline. 

Another disadvantage of having both CLI version on different repositories is that each time a new feature of fix is needed, two PRs must be created and the deployment process should be synced between them.

So with this PR the CLI2 is going to be embed inside the CLI3. As a first step the complete CLI2 will be copied inside but in a next step only the code related with `theme` commands should be remain

- Move the Ruby CLI2 codebase to the `theme` module
- Integrate CLI3 `themes` commands with the embed version of the CLI2
- Add a feature flag to switch between the using of the `embed` version and the `bundled` one
- Integrate lint tasks in the current github workflow

### MISSING points
- `theme-cli` unit tests are not working on github. Removed from the workflow
- `theme-cli` includes the complete CLI codebase. Retain only `theme commands` related code.

### How to test your changes?
- `embed` version of the CLI2will be used in the development environment as this env variable `SHOPIFY_CLI_THEME_CLI=embed`  has been added to the `dev.yml`, so please run `dev up ` before the first running 
- Run any theme command using `--verbose` flag and check that the message `Using embed Theme CLI` is being displayed
- In the release version the `bundled` version of the CLI2 is going to be used so far. To simulate this in the local environment run any `theme` command prepending this `SHOPIFY_CLI_THEME_CLI=bundled`

### Measuring impact

### Checklist


packages/theme/README.md
```diff
@@ -0,0 +1,63 @@
+<img src="https://github.com/Shopify/cli/blob/main/assets/logo.png?raw=true" width="150"/>
+
+# Shopify CLI
```
gonzaloriestra, line 3: Have you copied this file here by mistake? 🤔

packages/theme/src/cli/utilities/ruby.ts
```diff
@@ -0,0 +1,84 @@
...
+
+/**
+ * Execute Theme CLI commands which is an embed simplified version of the CLI2.
+ * Installs a version of RubyCLI as a vendor dependency in a hidden folder in the system.
+ * User must have a valid ruby+bundler environment to run any command.
+ *
+ * @param args - List of argumets to execute. (ex: ['theme', 'pull']).
+ * @param options - Options to customize the execution of cli2.
+ */
+export async function execCLI(args: string[], options: ExecCLI2Options = {}): Promise<void> {
```
gonzaloriestra, line 21: Instead of creating this new utility and function, that requires changing all the theme commands, wouldn't it be easier to just add the new logic inside execCLI2 (from cli-kit/src/public/node/ruby) and run the required CLI from there (RubyGems or embedded)?

packages/theme/src/cli/utilities/ruby.ts
```diff
@@ -0,0 +1,84 @@
...
+  await list.run()
+}
+
+/**
+ * It returns the directory where the Ruby CLI is located.
+ *
+ * @returns The absolute path to the directory.
+ */
+async function shopifyCLIDirectory(): Promise<string> {
+  return (await path.findUp('theme-cli', {
```
gonzaloriestra, line 80: Why finding the path if it's a fixed one?

packages/theme/package.json
```diff
@@ -15,7 +15,8 @@
   "type": "module",
   "files": [
     "/dist",
-    "/oclif.manifest.json"
+    "/oclif.manifest.json",
+    "/theme-cli"
```
gonzaloriestra, line 19: This is including all the files there in the package, right? What's the difference in size?

What about being a bit more explicit to avoid the test folders for example? Another alternative is using `.npmignore` (more info [here](https://blog.npmjs.org/post/165769683050/publishing-what-you-mean-to-publish)).

packages/theme/src/cli/utilities/ruby.ts
gonzaloriestra: Also, it seems you missed to replace execCLI2 in the dev command, which is required to run Theme App extensions. Another reason to move this to cli-kit.

alvaro-shopify:
After some team discussion, we have concluded that it's better to embed the Ruby code in its own npm module with a Typescript API that executes it.
Besides we should analyze the option of creating this new module as a plugin that could be installed dinamically, for example, only in case you create a theme o generate a theme_app_extension

.github/workflows/changelog-reminder.yml
```diff
@@ -1,7 +1,7 @@
 on:
   pull_request:
     types: [opened, synchronize, reopened, ready_for_review]
-    paths: ["packages/*/src/**"]
+    paths: ["packages/*/src/**", "packages/*/theme-cli"]
```
gonzaloriestra, line 4: Do we need this? Isn't the theme-cli folder matched by the first expression?

gonzaloriestra (changes requested):
Great work! It's almost ready 👏 

I left some comments, but the main one is to use the local code also [here](https://github.com/Shopify/cli/blob/85cfa339918e4740f12923ed24e4195a872f2c2b/packages/app/src/cli/services/dev.ts#L207) to run Theme app extensions.

PR #1128, opened 2023-01-17
Fix dev console hover issue

byrichardpowell:
Previously when you hovered over a table row for an extension the entire table would re-render.  This meant that when you hovered over a table row you would loose your scroll position.

This PR Fixes that issues

This PR splits `useExtensionsInternal` into multiple hooks:

* `useApp` provides app level data like the app title and url
* `useExtensions` provides a list of extension ID's and types.  This is key to the fix.  I'll explain this below.
* `useExtension` provides a single extension and methods to mutate that extensions given an extensions uuid.
*  `useExtensionServerOptions` provides the options the extensions server was initialized with.
*  `useNavigate` provides a websocket message to inform the parent window to navigate to a different resource URL.  It's what get's called when you click an extension preview link and the Dev console is embedded.

The keys to the fix are:

1. The Extensions component only consumes extensions ID's and types.  These never change, and because we memoize this data, consuming it never forces a re-render
2. Child components then use the more granular hooks provides above to get the data they care about.  For example, instead of receiving an `Extension` the `ExtensionRow` receives an Extension UUID which it then passed to `useExtension` to get the data it cares about.  

Combined these two strategies mean that each Row should only re-render when it's data changes.

One side effect of the change is we no longer need to pass Modal state from the rows to the parent.  We did this (I assume) to get around the fact that the entire table used to re-render. Now each Row can render it's own modal.  This introduces a little more code duplication (in tests, and in Modal state), but the overall paradigm is simpler so it's well worth it.

### How to test your changes?

1. `pnpm clean && pnpm install && pnpm build`
2. `cd fixtures/app`
3. `pnpm dev`
4.  Open the dev console link the CLI outputs
5. Click each Preview link.  Does it open in a new window and go to the appropriate extension?
6. Rename `fixtures/app/extensions/checkout-ui` to `fixtures/app/extensions/_checkout-ui` Does the build status in the dev console update?
7. Hover over each extension when looking at the surface that renders the extension.  Does the surface highlight? E.g: Click `product-subscription`
8. Do the buttons in the view column update their state?

### Post-release steps

None

### Measuring impact

### Checklist

### DIFF
--- a/packages/ui-extensions-dev-console/src/sections/Extensions/Extensions.test.tsx
+++ b/packages/ui-extensions-dev-console/src/sections/Extensions/Extensions.test.tsx
@@ -1,6 +1,6 @@
 import {Extensions} from './Extensions.js'
 
-import {AppHomeRow, QRCodeModal, ExtensionRow} from './components'
+import {AppHomeRow, ExtensionRow, PostPurchaseRow} from './components'
 import React from 'react'
 import {ExtensionServerClient} from '@shopify/ui-extensions-server-kit'
 import {mockExtension} from '@shopify/ui-extensions-server-kit/testing'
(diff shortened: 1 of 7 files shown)


matteodepalo (approved):
Great refactor! :+1:

PR #1135, opened 2023-01-18
[cli-kit] Refactor typings module

isaacroldan:
- Moved all typings files under the `public/common` folder. Not all of them are used outside of cli-kit, but it makes sense to be public in case we want to use it in the future. Since these are only types, makes sense to have them in `common`

### Measuring impact

### Checklist


packages/app/src/cli/metadata.ts
```diff
@@ -1,5 +1,5 @@
 import {metadata as metadataLib, monorail} from '@shopify/cli-kit'
-import type {PickByPrefix} from '@shopify/cli-kit/typing/pick-by-prefix'
+import type {PickByPrefix} from '@shopify/cli-kit/common/typing/pick-by-prefix'
```
pepicrft, line 2: `typing` or `types` is usually used to to group declaration files for modules that don't have types. What about naming this one `@shopify/cli-kit/common/ts` to reflect that are Typescript utilities?

pepicrft (approved):

PR #1145, opened 2023-01-18
Adopt the CLI UI kit in the `shopify theme package` command 

mgmanzella:

Fixes https://github.com/Shopify/internal-cli-foundations/issues/506

- Remove the usage of the execCLI2 from the package command
- Zip file is generated that can be unzipped on both Windows and Mac
- Adopt the UI elements following the prototype (Note: Have an open comment "CLI UI Kit adoption – Prototypes" still and the UI won't match the issue)

### How to test your changes?
On both Windows and Mac:

#### Without path
- Checkout branch locally 
- `pnpm shopify theme init --latest` and name your theme
-  `cd  /path/to/theme`
- `pnpm shopify theme package`
- Confirm zip file is created for theme
- Unzip the file and confirm theme files present

#### With path
- cd outside of your theme directory
- `pnpm shopify theme package --path /path/to/theme`
- Confirm zip file is created for theme
- Unzip the file and confirm theme files present

### Post-release steps
Update [the docs](https://shopify.dev/themes/tools/cli/commands#package) and remove "For Windows, you need to [install 7-Zip](https://www.7-zip.org/)" 

### Measuring impact

### Checklist


packages/cli-kit/src/public/node/archiver.ts
```diff
@@ -27,10 +49,30 @@ export async function zip(inputDirectory: string, outputZipPath: string): Promis
...
+ * @param filePath - A file path.
+ * @returns A file path compatible with the current platform architecture.
+ */
+function platformCompatiblePath(filePath: string) {
+  if (platform === 'windows') {
+    return filePath.replaceAll('/', '\\')
+  }
+
+  return filePath
+}
```
isaacroldan, line 78: Maybe this function makes more sense to move it to `path.ts`, what do you think?

packages/cli-kit/src/public/node/archiver.ts
```diff
@@ -1,22 +1,49 @@
+/* eslint-disable tsdoc/syntax */
```
mgmanzella, line 1: es lint was complaining about the doc string for `zip`, would love advice on how to make a valid doc string for `zip`'s use case

packages/theme/src/cli/services/package.test.ts
```diff
@@ -0,0 +1,170 @@
...
+
+async function readArchiveFiles(zipPath: string) {
+  await expect(fileExists(zipPath)).resolves.toBeTruthy()
+  // eslint-disable-next-line @babel/new-cap
+  const archive = new StreamZip.async({file: zipPath})
+  const archiveEntries = Object.keys(await archive.entries())
+  await archive.close()
+
+  return archiveEntries
+}
```
mgmanzella, line 165: these functions are also repeated in `archiver.integration.test.ts`, not sure if there's a concept of test helper functions in the cli

packages/cli-kit/src/public/node/path.ts
```diff
@@ -62,3 +63,20 @@ export function cwd(): string {
...
+
+/**
+ * Takes forward slash file path as a input and returns backslash
+ * file path if platform is Windows.
+ *
+ * @param filePath - A file path.
+ * @returns A file path compatible with the current platform architecture.
+ */
+export function platformCompatiblePath(filePath: string): string {
+  const {platform} = platformAndArch()
```
mgmanzella, line 75: is `platformAndArch` an expensive call? because this will be called for every theme file when a theme is being packaged

mgmanzella (commented):
Dropping some questions, also:
- What's the rule of thumb for creating a `integration.test.ts` file vs `test.ts` file?
- When generating a changeset, should all changed packages be included?

packages/cli-kit/src/public/node/archiver.ts
matteodepalo: What was the syntax issue? Was the error specific?

packages/theme/src/cli/services/package.ts
```diff
@@ -0,0 +1,53 @@
...
+  'config',
+  'layout',
+  'locales',
+  'sections',
+  'snippets',
+  'templates',
+  'release-notes.md',
+].join('/**|')
+
+export async function packageTheme(inputDirectory: string) {
```
Poitrin, line 18: I know, we don't have a real convention, but in e.g. `open.ts` or `publish.ts`, exported function has the same name as the file.
Therefore, was there a reason why `packageTheme` was not simply called `package`?
Import path `packages/theme/src/cli/services/package.ts` should contain enough context to make clear that it's about _themes_ :)

mgmanzella: image

package is a reserved word, i can add that in a comment for context. i figured it was better to only change the function name and not the file name so it matches the rest of the commands. this makes it easy to find files

packages/cli-kit/src/public/node/archiver.ts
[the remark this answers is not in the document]
mgmanzella: very true ty

packages/theme/src/cli/services/package.ts
Poitrin: This is of course a valid argument to keep `packageTheme` and `package.ts` :)

packages/cli-kit/src/public/node/archiver.ts
```diff
@@ -1,22 +1,49 @@
...
+ * @param options.outputZipPath - The absolute path to the output zip file.
+ * @param options.matchFilePattern - Pattern to match when adding files to zip, uses glob expressions.
+ * @param options.usePlatformCompatiblePath - Use file path slashes associated with the platform where the files are being zipped.
  */
-export async function zip(inputDirectory: string, outputZipPath: string): Promise<void> {
+export async function zip({
+  inputDirectory,
+  outputZipPath,
+  matchFilePattern = '**/*',
+  usePlatformCompatiblePath = false,
```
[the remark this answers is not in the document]
Poitrin: > can the zip file be opened and the contents examined?

yes, I was able to open the zip file and examine its contents 🤔

packages/cli-kit/src/public/node/archiver.ts
mgmanzella: ty @karreiro that definitely looks cleaner ! with this change im still getting es lint errors 😭 

@matteodepalo this is the error im getting:
image
 and then when i fix that error i get this one:
 
image

packages/cli-kit/src/public/node/archiver.ts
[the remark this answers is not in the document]
Poitrin: Awesome, thanks for checking/confirming :)

Poitrin (approved):
LGTM, from my side :)

krysclick (commented):
Approve

PR #1153, opened 2023-01-19
Fix regression introduced by file refactor

Arkham:

Fixes issue while deploying

In the case of deploys we use `readSync` to read a zip file, so we don't want to read it as a string nor interpret it as utf8.

### How to test your changes?

- Create a new app
- Add an extension
- Deploy it

### Measuring impact

### Checklist

### DIFF
--- a/packages/cli-kit/src/public/node/fs.ts
+++ b/packages/cli-kit/src/public/node/fs.ts
@@ -100,16 +100,14 @@ export async function readFile(path: string, options: ReadOptions = {encoding: '
 }
 
 /**
- * Synchronously reads a file and returns its content as a string.
+ * Synchronously reads a file and returns its content as a buffer.
  *
  * @param path - Path to the file to read.
- * @param options - Options to read the file with (defaults to utf-8 encoding).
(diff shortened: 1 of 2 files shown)


packages/cli-kit/src/public/node/fs.ts
```diff
@@ -100,16 +100,14 @@ export async function readFile(path: string, options: ReadOptions = {encoding: '
...
 /**
- * Synchronously reads a file and returns its content as a string.
+ * Synchronously reads a file and returns its content as a buffer.
  *
  * @param path - Path to the file to read.
- * @param options - Options to read the file with (defaults to utf-8 encoding).
  * @returns The content of the file.
  */
-export function readSync(path: string, options: object = {encoding: 'utf-8'}): string {
+export function readSync(path: string): Buffer {
```
isaacroldan, line 108: could you take the opportunity to rename it to `readFileSync`? I forgot to do it and is not following the same name pattern as other functions

isaacroldan (approved):

PR #1162, opened 2023-01-19
Import performance from perf_hooks

amcaplan:
It's https://github.com/Shopify/cli/pull/1094 all over again


Arkham (approved):
Not again 🗡️

pepicrft (approved):
Can we automate catching these issues?

amcaplan:
@pepicrft I'm not sure. `eslint` should be catching this stuff based on the `engines` in our `package.json`

PR #1179, opened 2023-01-20
[WIP] Migrate checkout UI extension

byrichardpowell:
WIP PR to confirm this is the experience we want.

Partially implements https://github.com/Shopify/shopify/issues/398990 

Deploy now checks for extensions that changed type from 'checkout_ui` to `ui_extension`.  If it finds any, it prompts the user to migrate the extension.  If they cancel deploy is cancelled.  If they accept the CLI performs a graphQL mutation.  Right now the GraphQL mutation is not in prod, but can be seen here: https://github.com/Shopify/partners/pull/44865

### How to test your changes?

1. `pnpm clean && pnpm install && pnpm build`
2. `pnpm shopify app deploy --path ./fixtures/app --verbose` to deploy the existing extensions.
3. Edit `shopify.ui.extension.toml` to migrate the checkout_ui extension. It should be like this:

```
name = "checkout-ui"

[capabilities]
api_access = true
network_access = true
block_progress = true

[[extension_points]]
target = "Checkout::Dynamic::Render"
module = "./src/index.jsx"
```

4. `pnpm shopify app deploy --path ./fixtures/app --verbose` to deploy your app again.
5. It should prompt you to migrate the extension.  Select no.  The deploy should stop
6. Now try step 5 & 6 again but accept the migration.  It will fail due to a GraphQL error (mutation is not deployed yet)

### Post-release steps

None

### Measuring impact

### Checklist

### DIFF
--- a/packages/app/src/cli/services/environment/id-matching.ts
+++ b/packages/app/src/cli/services/environment/id-matching.ts
@@ -9,6 +9,7 @@ export interface MatchResult {
   identifiers: IdentifiersExtensions
   toConfirm: {local: LocalSource; remote: RemoteSource}[]
   toCreate: LocalSource[]
+  toMigrate: {local: LocalSource; remote: RemoteSource}[]
   toManualMatch: {local: LocalSource[]; remote: RemoteSource[]}
 }
 
@@ -115,6 +116,7 @@ export async function automaticMatchmaking(
   const localSourcesIds = localSources.map((source) => source.localIdentifier)
   const ids = pickBy(identifiers, (_, id) => localSourcesIds.includes(id))
   const localUUIDs = Object.values(ids)
+
   const existsRemotely = (local: LocalSource) => {
     return Boolean(
       remoteSources.find(
@@ -143,5 +145,33 @@ export async function automaticMatchmaking(
     toConfirm,
     toCreate,
     toManualMatch: pending,
+    toMigrate: getExtensionsToMigrate(localSources, remoteSources, ids),
   }
+}
+
+function getExtensionsToMigrate(
+  localSources: LocalSource[],
+  remoteSources: RemoteSource[],
+  ids: {[key: string]: string},
+) {
+  const remoteExtensionTypesToMigrate = ['CHECKOUT_UI_EXTENSION']
+
+  return localSources.reduce<MatchResult['toMigrate']>((accumulator, localSource) => {
+    if (localSource.type === 'ui_extension') {
+      const remoteSource = remoteSources.find((source) => source.uuid === ids[localSource.configuration.name])
+
+      if (!remoteSource) {
+        return accumulator
+      }
+
+      const typeMimatches = remoteSource.type !== localSource.type
+      const typeIsToMigrate = remoteExtensionTypesToMigrate.includes(remoteSource.type)
+
+      if (typeMimatches && typeIsToMigrate) {
+        accumulator.push({local: localSource, remote: remoteSource})
+      }
+    }
+
+    return accumulator
+  }, [])
 }
--- a/packages/app/src/cli/services/environment/id-matching.ts
+++ b/packages/app/src/cli/services/environment/id-matching.ts
@@ -5,11 +5,15 @@ import {uniqBy, difference} from '@shopify/cli-kit/common/array'
 import {pickBy} from '@shopify/cli-kit/common/object'
 import {slugify} from '@shopify/cli-kit/common/string'
 
+export interface LocalRemoteSource {
+  local: LocalSource
+  remote: RemoteSource
+}
+
 export interface MatchResult {
   identifiers: IdentifiersExtensions
-  toConfirm: {local: LocalSource; remote: RemoteSource}[]
+  toConfirm: LocalRemoteSource[]
   toCreate: LocalSource[]
-  toMigrate: {local: LocalSource; remote: RemoteSource}[]
   toManualMatch: {local: LocalSource[]; remote: RemoteSource[]}
 }
 
@@ -113,17 +117,13 @@ export async function automaticMatchmaking(
   identifiers: IdentifiersExtensions,
   remoteIdField: 'id' | 'uuid',
 ): Promise<MatchResult> {
-  const localSourcesIds = localSources.map((source) => source.localIdentifier)
-  const ids = pickBy(identifiers, (_, id) => localSourcesIds.includes(id))
+  const ids = getExtensionIds(localSources, identifiers)
   const localUUIDs = Object.values(ids)
 
-  const existsRemotely = (local: LocalSource) => {
-    return Boolean(
-      remoteSources.find(
-        (remote) => remote[remoteIdField] === ids[local.localIdentifier] && remote.type === local.graphQLType,
-      ),
+  const existsRemotely = (local: LocalSource) =>
+    remoteSources.some(
+      (remote) => remote[remoteIdField] === ids[local.localIdentifier] && remote.type === local.graphQLType,
     )
-  }
 
   // We try to automatically match sources if they have the same name and type,
   // by considering local sources which are missing on the remote side and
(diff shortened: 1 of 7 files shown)


packages/app/src/cli/services/environment/prompts.ts
```diff
@@ -75,3 +75,19 @@ export async function deployConfirmationPrompt(summary: SourceSummary): Promise<
...
   })
 }
+
+export async function migratePrompt(local: LocalSource, remote: RemoteSource): Promise<boolean> {
+  const remoteType = remote.type.toLowerCase()
+  const localType = local.type
+  const localName = local.configuration.name
+
+  return renderConfirmationPrompt({
+    message: `You've changed ${localName} from ${remoteType} to ${localType}. Would you like to migrate it in Shopify Partners? This will all previous versions of the extension`,
```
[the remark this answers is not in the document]
isaacroldan: If there's going to be a page in the docs about this, it'd be nice to also include the link here as a `Learn more:`

packages/app/src/cli/services/environment/identifiers-extensions.ts
```diff
@@ -37,6 +37,20 @@ export async function ensureExtensionsIds(
...
+
+    if (!confirmed) {
+      return err('user-cancelled')
+    }
+
+    // eslint-disable-next-line no-await-in-loop
+    await migrateToUiExtension(extension.remote.id, options.appId)
+
+    output.completed(`Migrated ${extension.local.configuration.name}.`)
+  }
```
isaacroldan, line 52: I think it will be better to have 1 prompt asking for all migrations at the same time than doing it sequentially, the deploy will be cancelled anyway if you don't accept all migrations. // @MeredithCastile 
Same thing we do right now when we ask the user to confirm adding/updating an extension

packages/app/src/cli/services/environment/id-matching.ts
```diff
@@ -143,5 +145,33 @@ export async function automaticMatchmaking(
...
+function getExtensionsToMigrate(
+  localSources: LocalSource[],
+  remoteSources: RemoteSource[],
+  ids: {[key: string]: string},
+) {
+  const remoteExtensionTypesToMigrate = ['CHECKOUT_UI_EXTENSION']
+
+  return localSources.reduce<MatchResult['toMigrate']>((accumulator, localSource) => {
+    if (localSource.type === 'ui_extension') {
+      const remoteSource = remoteSources.find((source) => source.uuid === ids[localSource.configuration.name])
```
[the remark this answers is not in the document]
isaacroldan: This logic assumes you have the UUID of the extension locally right? that won't be always true.
The logic should be similar to `matchByNameAndType` but instead of matching by same type we match by `remoteExtensionTypesToMigrate` and `ui_extension`

In any case, adding a bunch of tests for different scenarios would help us detect any possible not covered edge case.

We can pair next week on this if you need to (or also maybe @Arkham can help, since he implemented most of this logic 😃 )

packages/app/src/cli/services/environment/identifiers-extensions.ts
byrichardpowell: Yeah, fair.  Lemme spike this.

packages/app/src/cli/services/environment/id-matching.ts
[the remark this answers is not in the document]
vividviolet: Does the matching by name logic account for the local name not matching the remote source's name? I think currently this can get out of date as developers can change the name through the Partner's dashboard. Do we then treat it as a different extension even though the uuid would match?

packages/app/src/cli/services/environment/identifiers-extensions.ts
```diff
@@ -37,6 +37,20 @@ export async function ensureExtensionsIds(
...
+  for (const extension of matchExtensions.toMigrate) {
+    // eslint-disable-next-line no-await-in-loop
+    const confirmed = await migratePrompt(extension.local, extension.remote)
+
+    if (!confirmed) {
+      return err('user-cancelled')
+    }
+
+    // eslint-disable-next-line no-await-in-loop
+    await migrateToUiExtension(extension.remote.id, options.appId)
```
vividviolet, line 49: Should this not be `options.app.apiKey`?

byrichardpowell:
> @byrichardpowell, are we going to have a command for migration as well as the option to edit the toml?
> In slack, we'd discussed: shopify app migrate extension

No, I think the thread in Slack ditched this in favor of adding logic to the deploy command.  The developer would edit the toml file themself and then run deploy.

packages/app/src/cli/services/environment/identifiers-extensions.ts
byrichardpowell: Any strong preferences @MeredithCastile @lemonmade @isaacroldan ?

You can checkout the last 2 commits and then test each experience using the tophat instructions in the PR description.

packages/app/src/cli/services/environment/id-matching.ts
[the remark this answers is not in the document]
byrichardpowell: I'm going to pair with @Arkham on this tomorrow, so will check this logic.

packages/app/src/cli/services/environment/identifiers-extensions.ts
byrichardpowell: Looking at how this is used elsewhere, and checking it with the env file the CLI saves, the appId is the api key

packages/app/src/cli/services/environment/identifiers-extensions.ts
Arkham: Looking at the last prompt I found it a bit confusing that it's saying it's going to update and permanently migrate the same `checkout-ui` extension.

Also that confirmation prompt is the one that users can skip by passing the `--force` flag, so I'm not sure it's the best place to display vital information that we really want developers to see. My proposal would be something like this:

```
╭─ warning ────────────────────────────────────────────────────────────────────╮
│  Extension migration detected.                                               │
│  A migration is permanent and can't be undone. All previous versions will    │
│  be migrated but current users won't be affected.                            │
│  Reference                                                                   │
│    • Migration guide ( https://shopify.dev/tutorials/migrate-extensions )    │

? Migrate these extensions in Shopify Partners?

      checkout-ui: • from 'checkout_ui_extension'
                   • to 'ui_extension'

> (y) Yes, migrate these extensions
  (n) No, cancel

? Make the following changes to your extensions in Shopify Partners?

      Update:  • checkout-ui

> (y) Yes, deploy to push changes
  (n) No, cancel
```

In the above example, the prompt about migrating extensions would show up even if the user passes `--force` to the deploy command.

byrichardpowell:
@Arkham and I paired, here is the UX now:

First we show them and Info and a confirmation for migrating:

Screenshot 2023-01-24 at 10 28 47 AM

When they select `Yes, migrate these extensions` we run the mutation.  If that succeeds we render the confirmation for the rest of the deploy:

Screenshot 2023-01-24 at 10 30 18 AM

@MeredithCastile @lemonmade if you can confirm this UX I'll add tests.

packages/app/src/cli/services/environment/identifiers-extensions.ts
byrichardpowell: Resolving this as it's now outdated.  Please see: https://github.com/Shopify/cli/pull/1179#issuecomment-1402144124 instead

byrichardpowell:
Quick video showing the experience: https://videobin.shopify.io/v/DR552j

MeredithCastile:
@byrichardpowell, @Arkham: Thanks for the screenshot of the DX. This feels directionally right.

The big thing that's missing in the DX here is a value proposition — why as a developer would I bother going through the migration?

Other notes on the proposed design tweaks (screenshot below):
- We don't need to any additional context in the confirmation modal; the question should suffice.
- We can combine the prompts.

image

([Figma](https://www.figma.com/file/doZXfSRO35QYLivxCGlUIk/App-commands?node-id=416%3A2378&t=AtatZwGNnidmJjz0-1))

Arkham:
@MeredithCastile  I'm afraid we can't combine the prompts, since the "deploy summary" one can be skipped by passing `--force` to the deploy command, and we don't want users to accidentally migrate their extensions. Also they could have added and updated other extensions (not just the ones we want to migrate), so I feel we have to have two separate prompts.

byrichardpowell:
Thanks @MeredithCastile 

Re value prop: At this point the developer has already done the hardest thing, which is manually making some changes to the toml file.  I would think that they are fairly commited at this point. Even so a soft sell might be helpful.  Any ideas?  I think the biggest reason to migrate is support for api versioning.  Thoughts @lemonmade ?

lemonmade:
Agreed that we should do a softer sell in the confirmation message. To Richard's point, the developer will have already had to follow a migration guide in order to update their dependency and config file. I can't think of a reason a developer would cancel out of the migration here, as it will be the last step of the migration guide.

MeredithCastile:
Richard and I paired yesterday.

I understand now that this will be triggered after the user had already modified their extension's config file. So I do have a value-prop in there, but it's not the main point. Richard and I talked about combining the two prompts, but he had concerns about how that would work with error states. With two prompts, it's clearer that the problem is either a migration error or a deploy error. We can perhaps revisit this, but for now, we'll go with 2 prompts.

image

Let me know if anything feels off here! Thanks, all.

byrichardpowell:
Yeah, it could be possible that my original was not valid, and this is an unrelated issue.  I'm not too familiar with this space so I would like some-one else to double check.

vividviolet:
@byrichardpowell in your example updated TOML you were missing the `api_version` so it's returning a validation error. Try adding `api_version=2022-10` and deploy again to see if that fixes the problem

vividviolet (approved):
Excellent work! I tested these flows and everything is working end to end:

- [x] Can migrate existing extensions
- [x] Can migrate existing extension even after renaming them in the dashboard
- [x] Only migrate relevant ones
- [x] Publish urls appropriately updated after migration
- [x] Canceling the deploy after migration still finish the migration and you can deploy at a later date
- [x] Deploying extensions still work after deleting all the saved uuids in the `.env`

byrichardpowell:
> @byrichardpowell in your example updated TOML you were missing the api_version so it's returning a validation error. Try adding api_version="2022-10" and deploy again to see if that fixes the problem

This was it, thank you @vividviolet

PR #1184, opened 2023-01-22
[cli-kit] Refactor testing

gonzaloriestra:

- Move testing/output module to public/node
- Move testing/ui and testing/store to private/node

### How to test your changes?

CI

### Measuring impact

### Checklist


packages/cli-kit/src/public/node/testing/output.ts
```diff
@@ -1,6 +1,21 @@
-import * as output from '../output.js'
+import * as output from '../../../output.js'
```
matteodepalo, line 1: Could this file be private instead of public? Or is the idea that we want to give adopters of `cli-kit` a tool to help them test in their code as well?

gonzaloriestra: It has to be public because it's used by other packages like `cli`.

PR #1185, opened 2023-01-23
Dev console banner

matteodepalo:

Now that we have a new dev console capable of showing different types of UI extensions together with the app link, we want to point to that directly if we can.

I've added a success banner to `dev` that displays either the app url if there are no ui extensions or the dev console url if there are. Pressing any key will now open that url.
Screenshot 2023-01-23 at 15 56 14

I've also adde some minor fixes:
- Fix `execCLI2` running tasks inside `renderConcurrent`. Previously `renderTasks` was trying to be rendered inside `renderConcurrent`, but that's not possible.
- Add bold text token to UI kit
- Banner custom sections can now have no title

### How to test your changes?

- Run `pnpm shopify app dev --path fixtures/app`
- Press any key
- Get to the new dev console

### Measuring impact

### Checklist

### DIFF
--- a/packages/app/src/cli/services/dev.ts
+++ b/packages/app/src/cli/services/dev.ts
@@ -2,7 +2,7 @@ import {ensureDevEnvironment} from './environment.js'
 import {generateFrontendURL, generatePartnersURLs, getURLs, shouldOrPromptUpdateURLs, updateURLs} from './dev/urls.js'
 import {installAppDependencies} from './dependencies.js'
 import {devUIExtensions} from './dev/extension.js'
-import {outputAppURL, outputExtensionsMessages, outputUpdateURLsResult} from './dev/output.js'
+import {outputExtensionsMessages, outputPreviewUrl, outputUpdateURLsResult} from './dev/output.js'
 import {themeExtensionArgs} from './dev/theme-extension-args.js'
 import {fetchSpecifications} from './generate/fetch-extension-specifications.js'
 import {
@@ -79,13 +79,16 @@ async function dev(options: DevOptions) {
   })
 
   const backendPort = await getAvailableTCPPort()
-
   const frontendConfig = localApp.webs.find(({configuration}) => configuration.type === WebType.Frontend)
   const backendConfig = localApp.webs.find(({configuration}) => configuration.type === WebType.Backend)
 
   /** If the app doesn't have web/ the link message is not necessary */
   const exposedUrl = usingLocalhost ? `${frontendUrl}:${frontendPort}` : frontendUrl
   let shouldUpdateURLs = false
+  const proxyTargets: ReverseHTTPProxyTarget[] = []
+  const proxyPort = usingLocalhost ? await getAvailableTCPPort() : frontendPort
+  const proxyUrl = usingLocalhost ? `${frontendUrl}:${proxyPort}` : frontendUrl
+
   if ((frontendConfig || backendConfig) && options.update) {
     const currentURLs = await getURLs(apiKey, token)
     const newURLs = generatePartnersURLs(exposedUrl, backendConfig?.configuration.authCallbackPath)
@@ -97,7 +100,6 @@ async function dev(options: DevOptions) {
     })
     if (shouldUpdateURLs) await updateURLs(newURLs, apiKey, token)
     await outputUpdateURLsResult(shouldUpdateURLs, newURLs, remoteApp)
-    outputAppURL(storeFqdn, exposedUrl)
   }
 
   // If we have a real UUID for an extension, use that instead of a random one
@@ -114,10 +116,6 @@ async function dev(options: DevOptions) {
     hostname: exposedUrl,
   }
 
-  const proxyTargets: ReverseHTTPProxyTarget[] = []
-  const proxyPort = usingLocalhost ? await getAvailableTCPPort() : frontendPort
-  const proxyUrl = usingLocalhost ? `${frontendUrl}:${proxyPort}` : frontendUrl
-
   if (localApp.extensions.ui.length > 0) {
     const devExt = await devUIExtensionsTarget({
       app: localApp,
@@ -132,7 +130,15 @@ async function dev(options: DevOptions) {
     proxyTargets.push(devExt)
   }
 
-  outputExtensionsMessages(localApp, storeFqdn, proxyUrl)
+  const previewUrl = outputPreviewUrl({
+    app: localApp,
+    storeFqdn,
+    exposedUrl,
+    proxyUrl,
+    appPreviewAvailable: Boolean(frontendConfig || backendConfig) && options.update,
+  })
+
+  outputExtensionsMessages(localApp)
 
   const additionalProcesses: output.OutputProcess[] = []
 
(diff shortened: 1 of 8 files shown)


packages/app/src/cli/utilities/app/http-reverse-proxy.ts
```diff
@@ -91,12 +93,20 @@ ${output.token.json(JSON.stringify(rules))}
   abortController.signal.addEventListener('abort', () => {
     server.close()
   })
+
+  async function openPreviewUrl() {
```
matteodepalo, line 97: Actually this doesn't work with `Ctrl+C`, I'll need to find an alternative solution.

packages/app/src/cli/services/dev/output.ts
```diff
@@ -16,45 +16,61 @@ export async function outputUpdateURLsResult(
...
 }
 
-export function outputDevConsoleURL(url: string) {
-  const title = 'Shopify extension dev console URL'
-  const heading = output.token.heading(title)
-  const devConsoleURL = `${url}/extensions/dev-console`
-  output.info(output.content`${heading}\n\n  ${devConsoleURL}\n`)
-}
+export function outputPreviewUrl({app, storeFqdn, exposedUrl, proxyUrl, appPreviewAvailable}: OutputPreviewUrlOptions) {
+  if (!appPreviewAvailable && app.extensions.ui.length === 0) {
```
isaacroldan, line 46: in which cases will `appPreviewAvailable` be false? shouldn't we always be able to `dev` the app?

packages/app/src/cli/utilities/app/http-reverse-proxy.test.ts
```diff
@@ -69,7 +70,7 @@ describe('runConcurrentHTTPProcessesAndPathForwardTraffic', () => {
     vi.mocked(getAvailableTCPPort).mockResolvedValueOnce(4000)
 
     // When
-    const got = await runConcurrentHTTPProcessesAndPathForwardTraffic(undefined, [], [])
+    const got = await runConcurrentHTTPProcessesAndPathForwardTraffic('', undefined, [], [])
```
isaacroldan, line 73: this function might improve if the params are grouped into a typed object, is weird to see `('', undefined, [], [])`
(Not necessarily needed for this PR, but if you can do it that'd be awesome)

packages/app/src/cli/services/dev/output.ts
matteodepalo: @isaacroldan do you think it still makes sense to show the dev console url if they have decided not to update the partners URLs? I think they wouldn't be able to see the dev console in that case, but the previous logic didn't account for that. Maybe we haven't gotten any bug reports because few people were going to the dev console with just `localhost`.

matteodepalo: I've greatly simplified the logic, could you please take another look?

isaacroldan: The dev console doesn't depend on the URLs being updated (the extensions should work anyway), so if there are extensions i'd say we need to show the url always

packages/cli-kit/src/ui.ts
```diff
@@ -167,23 +167,33 @@ export async function terminateBlockingPortProcessPrompt(port: number, stepDescr
...
+      }
+
+      process.stdin.pause()
+      process.stdin.setRawMode(false)
     }
 
     process.stdin.resume()
     process.stdin.setRawMode(true)
-    process.stdin.once('data', handler)
+    process.stdin.on('data', handler)
```
isaacroldan, line 197: So are we going to keep listening for keypresses here? why the change?

packages/app/src/cli/utilities/app/http-reverse-proxy.ts
```diff
@@ -91,10 +100,29 @@ ${output.token.json(JSON.stringify(rules))}
...
     renderConcurrent({
       processes: [...processes, ...additionalProcesses],
       abortController,
+      onInput(input: string) {
+        if (input === 'p' && previewUrl) {
+          // eslint-disable-next-line @typescript-eslint/no-floating-promises
+          openURL(previewUrl)
+        } else if (input === 'q') {
+          process.exit(0)
+        }
```
isaacroldan, line 114: Is this another way of opening the URL? we are not showing these options to the user no?

packages/cli-kit/src/ui.ts
isaacroldan: Could you check if this change affects the authentication flow?

packages/app/src/cli/services/dev.ts
```diff
@@ -97,7 +109,9 @@ async function dev(options: DevOptions) {
     })
     if (shouldUpdateURLs) await updateURLs(newURLs, apiKey, token)
     await outputUpdateURLsResult(shouldUpdateURLs, newURLs, remoteApp)
-    outputAppURL(storeFqdn, exposedUrl)
+    outputDevSuccess(localApp)
+    await ui.keypress('\r')
```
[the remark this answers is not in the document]
matteodepalo: I'll double check.

packages/cli-kit/src/ui.ts
matteodepalo: I believe calling `pause()` should detach this, but I'm going to double check.

packages/app/src/cli/utilities/app/http-reverse-proxy.ts
matteodepalo: There are no other ways it's only hotkeys. The user will have to press a key to continue and open the URL so there is no chance they miss this. If they close the window then they can reopen it by pressing P.

matteodepalo: I've added a preview link at the bottom of the logs.

isaacroldan: Sorry, i mean other way because we also have the `enter` one right?

matteodepalo:
@isaacroldan may I get another review? I've changed things considerably.

packages/app/src/cli/utilities/app/http-reverse-proxy.ts
```diff
@@ -91,13 +100,31 @@ ${outputToken.json(JSON.stringify(rules))}
...
+
+  if (previewUrl) {
+    renderConcurrentOptions = {
+      ...renderConcurrentOptions,
+      onInput: (input: string) => {
+        if (input === 'p' && previewUrl) {
+          // eslint-disable-next-line @typescript-eslint/no-floating-promises
+          openURL(previewUrl)
+        } else if (input === 'q') {
+          process.exit(0)
```
isaacroldan, line 117: Have you tested that no sub-processes are kept alive after exiting this way?

matteodepalo: I haven't tested, I'll check!

matteodepalo: There are a few subprocesses left hanging, but the behavior is not different from pressing ctrl+c. We might need to do some other checks here.

matteodepalo: I've investigated, process.exit wasn't killing all the child processes so I've added the `tree-kill` package to take care of that. I've tested and it does kill everything as expected.

isaacroldan: 🔥

isaacroldan (approved):

PR #1209, opened 2023-01-25
Adopt the NodeNext module resolution in Typescript

pepicrft:
The monorepo's Typescript configuration was using an old module resolution to interoperate with external dependencies that haven't yet adopted Node's standard for specifying the exported modules and types in the `package.json`. While this allowed us to use dependencies that haven't embraced the standards, we ended up having code in plugins that was able to import private modules.

I'm changing the `moduleResolution` to `NodeNext` for Typescript to resolve imports using `package.json`s' `exports` attribute. This required disabling type-checking for a few dependencies that haven't yet adopted the conventions. I think the caveat is sensible because those are dependencies that we barely updated and whose logic and module loading is already tested from our tests.

**Note:** That I have kept Hydrogen's Node resolution to `node` because the change caused some incompatibilities with `@shopify/hydrogen` and since the plugin is already being developed outside of the monorepo I don't think it's the worth the effort of going through those issues.

### How to test your changes?
All the CLIs should work, `create-app`, create-hydrogen`, and the app, hydrogen, and theme commands.

### Measuring impact

### Checklist

### DIFF
--- a/packages/app/src/cli/services/deploy.ts
+++ b/packages/app/src/cli/services/deploy.ts
@@ -20,7 +20,7 @@ import {renderInfo, renderSuccess, renderTasks} from '@shopify/cli-kit/node/ui'
 import {inTemporaryDirectory, mkdir} from '@shopify/cli-kit/node/fs'
 import {joinPath, dirname} from '@shopify/cli-kit/node/path'
 import {outputNewline, outputInfo} from '@shopify/cli-kit/node/output'
-import type {RenderAlertCutomSection, Task} from '@shopify/cli-kit/node/ui'
+import type {AlertCustomSection, Task} from '@shopify/cli-kit/node/ui'
 
 interface DeployOptions {
   /** The app to be built and uploaded */
@@ -202,7 +202,7 @@ async function outputCompletionMessage({
     ]
   }
 
-  const customSections: RenderAlertCutomSection[] = [
+  const customSections: AlertCustomSection[] = [
     {
       title: 'Summary',
       body: {


pepicrft:
[`@bugsnag/js`](https://github.com/bugsnag/bugsnag-js/tree/next/packages/node) is a big blocker. None of their packages exports an ESM module graph. [Issue](https://github.com/bugsnag/bugsnag-js)

packages/create-app/src/utils/template/npm.ts
```diff
@@ -1,6 +1,6 @@
 import {PackageManager, installNodeModules, PackageJson} from '@shopify/cli-kit/node/node-package-manager'
 import {CLI_KIT_VERSION} from '@shopify/cli-kit/common/version'
-import {Task} from '@shopify/cli-kit/src/private/node/ui/components/Tasks.js'
```
pepicrft, line 3: Here's an example of plugins importing private modules.

packages/cli-kit/src/public/node/ui.tsx
```diff
@@ -44,6 +47,7 @@ export async function renderConcurrent({
   )
 }
 
+export type RenderAlertCutomSection = CustomSection
```
matteodepalo, line 50: Could we rename this to `AlertCustomSection` to keep it consistent with the other exports? The type is more tied to the component than the render method itself.

matteodepalo (approved):
Nice! Just one comment about the name of an export.

PR #1212, opened 2023-01-25
Add downloadFile function usefull for some plugins

isaacroldan:

Fixes #0000 

### Measuring impact

### Checklist

### DIFF
--- a/packages/cli-kit/src/public/node/http.ts
+++ b/packages/cli-kit/src/public/node/http.ts
@@ -1,3 +1,5 @@
+import {dirname} from './path.js'
+import {createFileWriteStream, fileExistsSync, mkdirSync, unlinkFileSync} from './fs.js'
 import {buildHeaders, httpsAgent, sanitizedHeadersOutput} from '../../private/node/api/headers.js'
 import {outputContent, outputDebug} from '../../public/node/output.js'
 import FormData from 'form-data'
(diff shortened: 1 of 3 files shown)


packages/cli-kit/src/public/node/http.ts
```diff
@@ -73,3 +75,51 @@ ${sanitizedHeadersOutput((options?.headers ?? {}) as {[header: string]: string})
...
+    })
+
+    nodeFetch(url)
+      .then((res) => {
+        if (res.status === 302 && res.headers.get('location') !== null) {
+          const redirection = res.headers.get('location')!
+          done = false
+          file.close()
+          resolve(downloadFile(redirection, to, redirect + 1))
+        }
```
gonzaloriestra, line 117: I think we can use `nodeFetch(url, redirect: 'follow')` and forget about this logic.

gonzaloriestra (commented):
I think we can simplify the code a bit, and what about adding some tests? 🙏

packages/cli-kit/src/public/node/http.ts
isaacroldan: ah, good catch

gonzaloriestra:
Well, without the redirection logic it gets simpler, but we could add some tests anyway for the download function: check that it creates the folder, the file with the right content (mocking the nodeFetch response), that it raises an error when it fails... I wouldn't test anything about the other functions as they are just wrappers.

isaacroldan:
Simplified and added a test @gonzaloriestra 
You were right that i could create a test that tests everything except the actual download 👌

gonzaloriestra (approved):
Perfect, thanks!

PR #1233, opened 2023-01-27
Send uninstall webhook when app id changes

paulomarg:

If a developer uses 2 different apps with the same store locally, the app thinks it is already installed to the store. That happens because the database entry isn't associated with the app itself, so the app decides it's installed even if it isn't actually installed in Shopify.

To help ensure that the app starts with a fresh state, the CLI will now send an `app/uninstalled` webhook to the app after it starts, so that it can clean up the state for the store, and reinstall an app if it changes.

This only happens IF AND ONLY IF:
- the user has just selected a new app, either by running `dev --reset` or passing in a different `--api-key` flag
- the app has a new `webhooks_path` value in their backend `shopify.web.toml` file

<details>
<summary>Example output</summary>

```
Using a different app than last time, sending uninstall webhook to app server


Shareable app URL

  <super secret URL>

2023-01-26 20:52:43 | backend  |
2023-01-26 20:52:43 | backend  | > dev
2023-01-26 20:52:43 | backend  | > cross-env NODE_ENV=development nodemon index.js --ignore ./frontend
2023-01-26 20:52:43 | backend  |
2023-01-26 20:52:43 | frontend |
2023-01-26 20:52:43 | frontend | > dev
2023-01-26 20:52:43 | frontend | > vite
2023-01-26 20:52:43 | frontend |
2023-01-26 20:52:43 | backend  | [nodemon] 2.0.20
2023-01-26 20:52:43 | backend  | [nodemon] to restart at any time, enter `rs`
2023-01-26 20:52:43 | backend  | [nodemon] watching path(s): *.*
2023-01-26 20:52:43 | backend  | [nodemon] watching extensions: js,mjs,json
2023-01-26 20:52:43 | backend  | [nodemon] starting `node index.js`
2023-01-26 20:52:43 | frontend |
2023-01-26 20:52:43 | frontend |   vite v2.9.15 dev server running at:
2023-01-26 20:52:43 | frontend |
2023-01-26 20:52:43 | frontend |   > Local:    http://localhost:61733/
2023-01-26 20:52:43 | frontend |
2023-01-26 20:52:43 | frontend |   ready in 152ms.
2023-01-26 20:52:43 | frontend |
2023-01-26 20:52:43 | backend  | [shopify-api/INFO] version 6.1.0, environment Node v18.7.0
2023-01-26 20:52:45 | webhooks | Sending APP_UNINSTALLED webhook to app server
2023-01-26 20:52:45 | backend  | [shopify-api/INFO] Processing webhook request
2023-01-26 20:52:45 | backend  | [shopify-app/INFO] Webhook processed, returned status code 200
2023-01-26 20:52:45 | webhooks | APP_UNINSTALLED webhook delivered
```

</details>

### How to test your changes?

You'll need 2 Shopify apps (A and B) for this. Steps:

- Create a new app
- Add `webhooks_path: /api/webhooks` to the backend TOML file
- Run `dev`, select app A
- Stop the app, run `dev --reset`, select app B
  - You should see the new behaviour here
- Stop the app again, run `dev --api-key=<API key for app A>`
  - Behaviour should also appear here
- Run `dev`
  - No webhook should be fired

### Post-release steps

Once this is released, we should add the default webhooks path to the app template TOML files.

### Measuring impact

### Checklist

  - Is there a need to document this behaviour anywhere? It is a somewhat fringe case.


paulomarg:
I don't know why the CI builds are failing, it works locally for me, and the `output` object is being imported.

packages/app/src/cli/services/webhook/send-app-uninstalled-webhook.test.ts
```diff
@@ -0,0 +1,90 @@
...
+    headers: '{ "header": "Header Value" }',
+    success: false,
+    userErrors: [
+      {message: 'Error 1', fields: ['field1']},
+      {message: 'Error 2', fields: ['field1']},
+    ],
+  },
+}
+
+describe('sendUninstallWebhookToAppServer', () => {
```
gonzaloriestra, line 34: It would be nice to check a couple of other things:
- That the API call to get the Sample is using the expected API version (2023)
- That we are writing to stdout the expected messages

gonzaloriestra (approved):
Working like a charm! Thanks a lot for taking the time to investigate and fix this issue 👏 

The CI was failing because of some recent refactors, I pushed some changes here to fix it, I hope you don't mind. I left some suggestions as well.

And besides updating the templates, we will need to update the documentation as well. @shainaraskas could you please take care of it? Basically, we are adding a new optional field to the `shopify.web.toml` to customize the path for the webhook API. Example:

```
type="backend"
webhooks_path = "/api/webhooks"

[commands]
dev = "npm run dev"
```

packages/app/src/cli/services/dev.ts
```diff
@@ -82,6 +89,12 @@ async function dev(options: DevOptions) {
 
   const frontendConfig = localApp.webs.find(({configuration}) => configuration.type === WebType.Frontend)
   const backendConfig = localApp.webs.find(({configuration}) => configuration.type === WebType.Backend)
+  const webhooksPath = backendConfig?.configuration?.webhooksPath
```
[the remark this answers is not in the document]
paulomarg: Yes, I think that seems reasonable. We should make sure we're handling that request timing out or returning a 404 though.

shainaraskas:
@gonzaloriestra I don't think I have enough information to document this yet (need a "why"), but I'l create an issue. No guarantee I can get to this this week.

packages/app/src/cli/services/webhook/send-app-uninstalled-webhook.test.ts
paulomarg: I'm usually a bit iffy about testing messages because they make the tests a bit brittle, but I think this is unlikely to change so we can do it.

gonzaloriestra: To make it more solid, I like to use a regex to check the main part of the message instead of a full comparison. For example, `/retrying/` instead of `App isn't responding yet, retrying in 5 seconds`.

Issue #1235, opened 2023-01-28
[Bug]: Can't select --store in 3.37.0 or later


kj:
### Please confirm that you have:

- [X] Searched [existing issues](.) to see if your issue is a duplicate. (If you’ve found a duplicate issue, feel free to add additional information in a comment on it.)
- [X] Reproduced the issue in [the latest CLI version](https://www.npmjs.com/package/@shopify/cli).

### In which of these areas are you experiencing a problem?

Theme

### Expected behavior

Running `shopify theme info --store <my-store>` outputs information for the selected store.

### Actual behavior

The command outputs:

```
╭─ error ──────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│                                                                                                                      │
│  A store is required                                                                                                 │
│                                                                                                                      │
│  Specify the store passing --undefined={your_store_url} or set the SHOPIFY_FLAG_STORE environment variable.          │
│                                                                                                                      │
╰──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
```

This occurs with versions 3.37.0 and 3.38.0. Version 3.36.2 works as expected.

### Reproduction steps

1. Install `@shopify/cli` and `@shopify/theme` version 3.37.0 or 3.38.0
2. Run `shopify theme info --store <my-store>`
3. The command fails, suggesting you select the store with `--store` (although refers to the flag as `--undefined`)

### Operating System

Arch Linux

### Shopify CLI version (check your project's `package.json` if you're not sure)

3.38.0

### Shell

bash

### Node version (run `node -v` if you're not sure)

v18.12.1

amcaplan:
Note that [this](https://github.com/Shopify/cli/pull/1188) is the PR that broke it. Specifically [these lines](https://github.com/Shopify/cli/pull/1188/files#diff-e2bb7d7b733bd9f4d4bc52ce5d1aacfd937fc5de7e338b2ce277dabbce8e4dd5L15-R16), since it means that if `store` stays undefined, it will raise [here](https://github.com/Shopify/cli/pull/1188/files#diff-f0fc976ac9a3fc146ddc205bd8372f720cd43f1cb058ccec3ada7462fc198e4fR9).

PR #1243, opened 2023-01-30
Fix theme info command

isaacroldan:
With the cli-kit `store.ts` refactor we introduced a bug in the `theme info` command. This command should not require a valid store to be present.
Fixes https://github.com/Shopify/cli/issues/1235

If there is no store cached, just return "Not configured"

### How to test your changes?
`theme info` should work with and without a cached store

### Measuring impact

### Checklist


amcaplan (approved):
Makes sense. We don't really need the rest of what that function does (no stickiness is relevant as there's no flag).

amcaplan (approved):

PR #1247, opened 2023-01-30
Rename presets to environments

gonzaloriestra:

Fixes https://github.com/Shopify/internal-cli-foundations/issues/541

- Renames everything related to presets to environments, including the current file name (`shopify.environments.toml`) and flag (`--environment`).
- Rename the previous `environment` concept (like `ensureDevEnvironment`) to `context` to avoid confusion

### How to test your changes?

Create a file called `shopify.environments.toml` in the root folder of an app with the content:
```
[test]
json = true
```
And then run `yarn shopify app info --environment test`. It should show that the environment is being used, and output the info in json format.

### Measuring impact

### Checklist


gonzaloriestra:
@Arkham very good point! I totally agree.

What about renaming everything related to DevEnvironment/DeployEnvironment to DevContext/DeployContext? I prefer to keep consistency between internal/external naming, but that is only internal and I think "context" fits well.

PR #1272, opened 2023-02-02
[Feature] Set firstPartyDev to true when env is Spin

swalkinshaw:
Improves the internal Shopify developer experience by defaulting to the most common use case. 

This change could be merged on its own, but it should depend on https://github.com/Shopify/partners/pull/45140 which removes the binary choice between the Shopify or Italic apps org; instead both will be returned.

Enables `firstPartyDev` mode when in Spin.

### How to test your changes?
Run `yarn dev --reset` or `yarn deploy --reset` to see the list of partner organizations fetched.

### Checklist

### DIFF
--- a/packages/cli-kit/src/public/node/context/local.ts
+++ b/packages/cli-kit/src/public/node/context/local.ts
@@ -1,4 +1,4 @@
-import {isSpin} from './spin.js'
+import {isSpin, isSpinEnvironment} from './spin.js'
 import {isTruthy, isSet} from '../../../private/node/context/utilities.js'
 import {environmentVariables, pathConstants} from '../../../private/node/constants.js'
 import {fileExists} from '../fs.js'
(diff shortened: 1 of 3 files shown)


packages/cli-kit/src/public/node/context/local.ts
```diff
@@ -99,7 +99,7 @@ export function alwaysLogAnalytics(env = process.env): boolean {
  * @returns True if SHOPIFY_CLI_1P is truthy.
  */
 export function firstPartyDev(env = process.env): boolean {
-  return isTruthy(env[environmentVariables.firstPartyDev])
+  return isSpin(env) || isTruthy(env[environmentVariables.firstPartyDev])
```
isaacroldan, line 102: This will return true if you are running the CLI inside the Spin instance. But I guess you also want it to be true if you are running the CLI locally and connecting to a remote spin instance no?

```suggestion
  return isSpin(env) || isSpinEnvironment(env) || isTruthy(env[environmentVariables.firstPartyDev])
```

isaacroldan:
I tested this and I can see the Orgs, but no the apps inside them 🤔 Not sure why, does it work for you?

swalkinshaw:
I'll test again after making the spin env update and see 🤔

swalkinshaw:
🎩 and all seems to work as expected

image

image

image

isaacroldan:
I do see the apps I create in Italic Apps, but Shopify is empty... should I see apps there? is because of my user? 🤔

swalkinshaw:
We got to the bottom of @isaacroldan's question above ^ It just depends on apps being cached in the local partners db

PR #1274, opened 2023-02-02
Feature: Add tooltip component

kris-tremblay:
Add Tooltip Component: https://github.com/Shopify/internal-cli-foundations/issues/526
Add Tooltip to Status Column: https://github.com/Shopify/internal-cli-foundations/issues/527

Updated with keyboard navigation and status column tooltips (with dotted underline).  **Now with viewport bound checks**

https://user-images.githubusercontent.com/103125965/216848887-8475e442-97b6-43b4-ae7e-4e6134f2e259.mp4

## API/Usage (WIP)

The Tooltip component currently takes only one prop `text` and accepts a single child of either type `JSX.Element` or `string`.

```jsx
import {Tooltip} from '@components/Tooltip'

// Add a tooltip to a component

<Tooltip text="This is a tooltip!">
  <IconButton icon={SomeIcon} />
</Tooltip>

// Add a tooltip to an a string

<Tooltip text="This is a tooltip!">
  This string will have a dotted underline.
</Tooltip>

// Since Tooltip is wrapped an inline-block div, it may be used
// in a block of text

<p>
  Only <Tooltip text="This right here!">this section</Tooltip> will be underlined and trigger a tooltip.
</p>
```

## Notes:

- I have used fixed positioning and z-index in lieu of portals for simplicity.  This can be swapped out later, but at least for this application it seems to work fine.
- ~There are currently no client boundary checks, so it's possible the tooltip can be off the screen under certain circumstances.~
- ~The component has little accessibility support, save for it appears on focus (elements that are not "tabable" will not trigger with the tab key, ie. images) and the role of the container is set to "tooltip".  This can be addressed later.~

### Measuring impact


packages/ui-extensions-dev-console/src/components/Tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,62 @@
...
+    case 'show':
+      return {...state, isVisible: true}
+    case 'hide':
+      return {...state, isVisible: false}
+    case 'position':
+      return {...state, position: action.payload}
+    default:
+      throw new Error()
+  }
+}
```
byrichardpowell, line 17: It looks like there are just two actions here: 

1. Show, which sets isVisible and the position
2. Hide, which sets is visible

Is a reducer overkill?

packages/ui-extensions-dev-console/src/components/Tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,62 @@
...
+      <div
+        className={styles.Tooltip}
+        onMouseEnter={handleShow}
+        onFocus={handleShow}
+        onBlur={() => dispatch({type: 'hide'})}
+        onMouseLeave={() => dispatch({type: 'hide'})}
+        ref={ref}
+      >
+        {children}
+        {state.isVisible && <TooltipPopover position={state.position} text={text} />}
```
byrichardpowell, line 58: I think this means keyboard users won't get any value from the tooltip? How easy would it be to architect this so that keyboard users do get some value?

packages/ui-extensions-dev-console/src/sections/Extensions/Extensions.tsx
```diff
@@ -34,7 +37,14 @@ export function Extensions() {
             <th>{i18n.translate('extensionList.name')}</th>
             <th>{i18n.translate('extensionList.preview')}</th>
             <th>{i18n.translate('extensionList.mobile')}</th>
-            <th>{i18n.translate('extensionList.view')}</th>
+            <th>
+              <span className={styles.WithIcon}>
+                {i18n.translate('extensionList.view')}
+                <Tooltip text={i18n.translate('tooltips.viewColumnHeader')}>
+                  <Icon source={QuestionMarkMajor} />
```
byrichardpowell, line 44: I think the color of the Icon doesn't match the deigns, please could we update it?

packages/ui-extensions-dev-console/tsconfig.base.json
```diff
@@ -1,7 +1,7 @@
 {
   "compilerOptions": {
     "target": "ESNext",
-    "lib": ["ESNext"],
+    "lib": ["ESNext", "DOM"],
```
byrichardpowell, line 4: :+1:

packages/ui-extensions-dev-console/src/components/Tooltip/TooltipPopover.module.css
```diff
@@ -0,0 +1,12 @@
+.Popover {
+  position: fixed;
+  background-color: var(--dc-color-grey-600);
+  color: #fff;
+  padding: 1rem;
+  border-radius: 4px;
+  max-width: 250px;
```
byrichardpowell, line 7: Does the font-weight match the designs? Please could we update it if not?

PR #1277, opened 2023-02-03
Fix Banner potentially crashing on very narrow terminals

matteodepalo:

Fixes https://github.com/Shopify/internal-cli-foundations/issues/549

`repeat` throws an error if the argument is negative. This PR makes sure that doesn't happen.

### How to test your changes?

- Make your terminal very narrow, around 5 columns should do it.
- Run `deploy`
- It shouldn't crash

### Measuring impact

### Checklist


matteodepalo:
They layout logic is correct in that if the width is really small then you get that as the max width. If we provided a larger min width then we would break the layout for narrow terminals. I think we should just accept that this layout breaks for widths that are too narrow to even show the banner title.

PR #1282, opened 2023-02-03
Enable environments for themes

gonzaloriestra:

Fixes https://github.com/Shopify/internal-cli-foundations/issues/553

Enables environment for themes, reading the configuration from a `shopify.theme.toml` file.

Credits to @amcaplan, as this is mostly a refactor of his excellent previous work.

Demo: https://videobin.shopify.io/v/4QJbDx (the names in the config file are missing the `environments.` prefix)

Important details:
- It looks for the `shopify.theme.toml` file in the current directory (or from `--path`) and parent folders
- It shows a warning when the environment file/name is not found
- It shows the current flags being applied:
03-03-d88vr-x665l
- It shows the password masked with asterisks (showing the latest 4 chars)
- It allows any kind of flag:
  - If it doesn't make sense for the current command, it's ignored
  - If it's also included manually as a command flag or env variable, it's ignored
  - It allows multiple values by adding an array, like `ignore = ["sections/announcement-bar.liquid", "sections/header.liquid"]`
  - It raises an error when using a wrong type or two flags mutually exclusive
  - These special flags are ignored: `environment`, `path`, `verbose`
  - Some shared flags may have a different meaning depending on the command, like `force` or `reset`. I'd just add a warning in the documentation, but we could also disable them.
  - I didn't add a char shortcut for `--environment`, like `-e`, because it's already used by other commands like dev (alias for `--theme-editor-sync`).
  - Note that `shopify hydrogen preview` already has a `--env`/`-e` flag, which may be confusing. 

### How to test your changes?

Create a `shopify.theme.toml` in the root of a folder of a theme project:
```
[environments.env1]
name = "my-theme"
theme = "123"
store = "my-store"
password  = "shptka_123456"

[environments.env2]
store = "another-store"
ignore = "sections/announcement-bar.liquid"
output = "json"
```

Replace the flags with valid values and play with different commands and flags, like: `shopify theme dev --environment env1`

### Post-release steps

Release documentation: https://github.com/Shopify/shopify-dev/issues/31243

### Measuring impact

### Checklist


packages/app/src/cli/utilities/app-command.ts
```diff
@@ -1,19 +1,7 @@
...
-      cwd: specifiedPath,
-      type: 'file',
-    })
-    return appTOML ? dirname(appTOML) : specifiedPath
-  }
-
-  findUpForEnvironments(): boolean {
-    return false
+  environmentsFilename(): string {
+    return 'shopify.environments.toml'
```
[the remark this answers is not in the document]
gonzaloriestra: Probably not, this is on hold yet: https://docs.google.com/document/d/1yxT-PO_3Z8qG6n5BsUaoUx26Tdq0Tr2BaZfwvpr4VQE/edit?disco=AAAApPjvLnc

packages/cli-kit/src/public/node/base-command.ts
```diff
@@ -104,12 +104,16 @@ abstract class BaseCommand extends Command {
...
+    const basePath = rawFlags.path || cwd()
+    return findPathUp(this.environmentsFilename(), {
+      cwd: basePath,
+      type: 'file',
+    })
   }
 
-  protected findUpForEnvironments(): boolean {
-    return true
+  protected environmentsFilename(): string {
```
pepicrft, line 115: is it mandatory to implement this? If not, should we change the return type to `string | undefine` and remove the `throw`?

packages/cli-kit/src/public/node/environments.ts
```diff
@@ -1,36 +1,27 @@
...
-    })
-  } else {
-    const allowedEnvironmentsFilePath = joinPath(dir, environmentsFilename)
-    if (await fileExists(allowedEnvironmentsFilePath)) {
-      environmentsFilePath = allowedEnvironmentsFilePath
-    }
-  }
-  if (environmentsFilePath) {
-    return decodeToml(await readFile(environmentsFilePath)) as Environments
+export async function loadEnvironment(
```
pepicrft, line 14: I'd design this API such that Typescript can prevent invalid states from being passed. The following invocation is possible and doesn't do anything besides calling `outputWarn`:

```js
loadEnvironment("....", undefined)
```

I'd also consider moving the `findUp` logic inside this function. If it's the command knows the name of the environment file and the path, we can pass that information into this function and do the find up from here:

```js
await loadEnvironment("filename", {
  from: directory
})
```

shainaraskas:
hey @gonzaloriestra - for this one, could you fill out the [dev docs issue template](https://github.com/Shopify/shopify-dev/issues/new?assignees=&labels=discipline-content&template=new_doc.yml)? just want to make sure I don't miss any important details, and that I get this done at the right time

gonzaloriestra:
@shainaraskas here you have it: https://github.com/Shopify/shopify-dev/issues/31243. Thanks!

kfa-spyro:
Any plans for default / common "environment"? It would be a huge deal to have the possibility to setup default values for variables, same as themeKit allows it.

gonzaloriestra:
@kfa-spyro we have it in mind, thanks! But we can't provide a date yet.

PR #1312, opened 2023-02-08
Autogenerate comments for ui module

matteodepalo:
## Summary

Use the TS compiler API to automatically generate examples to put on top of `ui.tsx` render functions.

## What this PR does

I've added a new CI step, similar to the one that checks for oclif manifests, that will check if there are any changes to the files that are affected by the autogenerated docs. For now this is only one file, `ui.tsx`.

- [ ] Decrease width of examples

## What to look for in the PR

- Check out all the changes in `ui.tsx` and see if they look nice to you or not. You might have to expand the file here on Github as there are many lines that have been changed
- I've also tried to make sure that all the changes to `ui.tsx` are backwards compatible, but please let me know if I've missed something!


.github/workflows/shopify-cli.yml
```diff
@@ -154,6 +154,30 @@ jobs:
...
 
+  autogenerated-docs:
+    env:
+      CI: false
+    name: Check autogenerated docs
+    runs-on: ubuntu-latest
+    if: ${{ github.event_name == 'push' || github.event_name == 'pull_request' }}
+    timeout-minutes: 30
+    steps:
+      - uses: actions/checkout@v2
```
Arkham, line 165:
```suggestion
      - uses: actions/checkout@v3
```

packages/cli-kit/src/private/node/ui/refresh-documentation.ts
```diff
@@ -0,0 +1,77 @@
...
+  const exampleTags: {[key: string]: JSDocTag[]} = renderFunctions.reduce((acc, func) => {
+    acc[func.getNameOrThrow()] = func
+      .getJsDocs()
+      .flatMap((jsDoc) => jsDoc.getTags())
+      .filter((tag) => tag.getTagName() === 'example')
+
+    return acc
+  }, {} as {[key: string]: JSDocTag[]})
+
+  const removeTrainlingSpaces = (text: string) => text.replace(/ +$/gm, '')
```
Arkham, line 35:
```suggestion
  const removeTrailingSpaces = (text: string) => text.replace(/ +$/gm, '')
```

matteodepalo:
Thank you @Arkham I've addressed the comments 👍

packages/cli-kit/src/private/node/ui/refresh-documentation.ts
```diff
@@ -0,0 +1,77 @@
+import {examples} from './examples.js'
+import {unstyled} from '../../../public/node/output.js'
+import {JSDocTag, Project} from 'ts-morph'
+import isEqual from 'lodash/isEqual.js'
+
+async function refreshDocumentation(): Promise<void> {
+  const project = new Project({
+    tsConfigFilePath: 'tsconfig.json',
```
pepicrft, line 8: I'd use an absolute path here for two reasons:
- It's easier to reason about. It's hard to tell by looking at this line if it will use the `cli-kit`'s `tsconfig.json` or the root one.
- It makes the script work regardless of which working directory you are (assuming they are building the absolute path internally from the working directory).

packages/cli-kit/src/private/node/ui/refresh-documentation.ts
```diff
@@ -0,0 +1,77 @@
...
+      .filter((tag) => tag.getTagName() === 'example')
+      .forEach((tag) => tag.remove())
+
+    functionJsDoc.addTags(tags)
+  }
+
+  await project.save()
+}
+
+refreshDocumentation()
```
pepicrft, line 70: The Node version that we are using should support top level awaits.

packages/cli-kit/src/private/node/ui/components/AutocompletePrompt.test.tsx
```diff
@@ -436,11 +435,11 @@ describe('AutocompletePrompt', async () => {
     expect(onEnter).toHaveBeenCalledWith('fifth')
   })
 
-  test('allows selecting the first item after searching and triggering the loading state', async () => {
+  test('allows selecting the first item after searching', async () => {
     const onEnter = vi.fn()
 
     const search = async (term: string) => {
-      await new Promise((resolve) => setTimeout(resolve, 500))
+      await new Promise((resolve) => setTimeout(resolve, 300))
```
pepicrft, line 442: Is this due to flakiness?

packages/cli-kit/project.json
```diff
@@ -127,6 +127,14 @@
           "command": "pnpm tsc --noEmit",
           "cwd": "packages/cli-kit"
         }
+      },
+      "refresh-documentation": {
+        "executor": "nx:run-commands",
+        "dependsOn": ["build"],
+        "options": {
+          "command": "pnpm node dist/private/node/ui/refresh-documentation.js",
```
pepicrft, line 135: I'd recommend not to mix tooling and runtime code under `src` to be able to have access to the `TS ~> JS` build pipeline. We can place it in [`/bin`](https://github.com/Shopify/cli/tree/main/bin) and use [ts-node](https://www.npmjs.com/package/ts-node) to transpile and run the code:

```bash
pnpm ts-node-esm /bin/refresh-documentation.ts
```

pepicrft: And with that we can drop the `dependsOn: ["build"]`

pepicrft (approved):

packages/cli-kit/src/private/node/ui/alert.tsx
```diff
@@ -15,6 +16,10 @@ const typeToLogger: {[key in AlertProps['type']]: Logger} = {
   success: consoleLog,
 }
 
+export interface AlertOptions extends AlertProps {
+  renderOptions?: RenderOptions
```
pepicrft, line 20: Note that this one leaks a type from Ink through the public interface. We can trust Ink in that they won't introduce breaking changes there (which translate to breaking changes in our APIs) or not keep the trust low and export our type instead that maps to theirs. I'm inclined to the cautious because a breaking change might cause the CLI to blow up on the user end.

packages/cli-kit/src/private/node/ui/refresh-documentation.ts
matteodepalo: I tried but it wasn't a problem of the node version, bur rather the `module` not being `nodenext`. I think it's not a big deal, I'll leave it as is.

packages/cli-kit/src/private/node/ui/components/AutocompletePrompt.test.tsx
matteodepalo: I actually reduced flakiness of this test and so I was able to reduce the number of ms for search to perform.

packages/cli-kit/src/private/node/testing/ui.ts
```diff
@@ -1,5 +1,22 @@
 import {isTruthy} from '../context/utilities.js'
 import {render} from 'ink-testing-library'
+import {EventEmitter} from 'events'
+
+export class Stdin extends EventEmitter {
```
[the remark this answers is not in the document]
matteodepalo: Maybe, I think the fact that this is exported by the `testing/ui` module should be enough though.

packages/cli-kit/project.json
matteodepalo: I did go with `ts-node-esm` first but didn't want to do an extra transpilation step. I'll do as you recommended.

packages/cli-kit/src/private/node/ui/alert.tsx
matteodepalo: Not sure I understand 100%, but if we introduced our own type and Ink created a breaking change, wouldn't we need to update our type anyway to adapt to the new interface? Unless we can somehow reproduce the previous version's logic I don't think it's something we can avoid. Also this is an optional attribute exposed by a private module so I wouldn't be too worried about that.

PR #1340, opened 2023-02-09
Standardized user-facing texts in command descriptions

bhox:
User facing text for command descriptions were inconsistent

User-facing text standardizations:
- Lowercased 'Extension' -> 'extension'
- Lowercased 'Function' -> 'function'
- A command summary/descriptions end in .
- Updated 'view'/'print' -> 'Display'
- Updated 'Shopify app' -> 'app'


Arkham:
@MeredithCastile what do you feel about this? I couldn't find a place where we define conventions for command descriptions

Issue #1344, opened 2023-02-10
[Feature]: Introduce the `shopify theme console` command


karreiro:
## This command is currently released as a beta feature

Please, up-vote 👍 or down-vote 👎 for it.

---

### Overview

Read–Eval–Print Loop (REPL) tools are excellent for supporting experienced developers that want to try some snippets quickly but are also great for beginners during the learning process.

This issue aims to Introduce a REPL tool to support Liquid developers.

### Motivation

Reduce the feedback loop for theme developers.

blanklob:
Hey Can you explain in more details what this command does? and an example use case 🙂

blanklob:
Oh thats really cool, It will really save so much time, rather than output thing into HTML directly thanks for explanation!

tobi:
heh, now i'm learning about cool features we are working on from banklob. Yes, let's definitely ship that. Very cool.

ulgut:
This is so cool! 🚀

blanklob:
@karreiro  I didn't manage to launch the console, its send me to a 404 page?

I used latest build of Dawn v10

```json
"@shopify/cli": "^3.46.5",
"@shopify/theme": "^3.46.5"
```
<img width="1456" alt="Capture d’écran 2023-06-21 à 04 13 44" src="https://github.com/Shopify/cli/assets/63895927/be5a0918-7cc2-4024-8073-4fad92970d02">

octipus:
Yes please 🚀

drobinson:
Is there a way to specify a template and page context so we can for example poke at why updates to section schema aren't working in theme editor?

karreiro:
@drobinson that's no included in the prototype, but the plan is to incorporate that feature in the final version :)

blanklob:
Hey @karreiro I recieved another time the error I told about, regarding when the CLI send me to a 404 page, instead of launch the liquid terminal.

Here are the logs:

```bash
2023-07-20T02:40:34.237Z: 
Running system process:
  · Command: ruby -v
  · Working directory: /Users/youness/Development/studio/theme

2023-07-20T02:40:34.247Z: 
Running system process:
  · Command: bundle -v
  · Working directory: /Users/youness/Development/studio/theme

[...]

PR #1345, opened 2023-02-10
Introduce the hidden/beta `shopify theme console` command

karreiro:

This PR partially fixes the https://github.com/Shopify/cli/issues/1344 issue, because it does introduce the `shopify theme console` command, however this PR introduces it as a _beta_ command.

This PR introduces a introduces the new `shopify theme console` command as a beta feature.

demo

Some interesting topics this PR approaches:

#### Local evaluation vs. SFR evaluation

I've considered to evaluate snippets locally, however some filters exist only at SFR level, so we can evolve this feature to evaluate _some_ snippets locally, but it will always have some level of dependency with SFR. Also, it's quite nice that developers can explore/use real data from their stores.

#### Authentication

As this command access store data, it needs to be authenticated. The new command supports the default authentication workflow and with the Theme Access based authentication.

#### The session

Currently, there's no limit in the session, so _some_ interactions with terminal can make it a bit slower. In the official implementation we need to better handle this, still, given the nature of this feature, I consider the current implementation good enough to start.

#### Ruby

I needed to introduce the core of this command in Ruby, because it re-uses a lot of code from the development server from `shopify theme dev`. Still, the implementation is trivial to port to TypeScript as soon as we migrate the `shopify theme dev` command. Also, all UI elements follow the CLI UI kit conventions.

### Post-release steps

None.

### Measuring impact

### Checklist


packages/cli-kit/assets/cli-ruby/lib/shopify_cli/theme/dev_server/proxy.rb
```diff
@@ -78,6 +78,17 @@ def call(env)
           [response.code, headers, body]
         end
 
+        def secure_session_id
+          if secure_session_id_expired?
+            @ctx.debug("Refreshing preview _secure_session_id cookie")
```
Poitrin, line 83: What about …
```diff
-… _secure_session_id …
+… @secure_session_id …
```
or
```diff
+… secure_session_id …
```
?

packages/cli-kit/assets/cli-ruby/lib/shopify_cli/theme/repl/api.rb
```diff
@@ -0,0 +1,107 @@
...
+
+        def debug?
+          @is_debug ||= ctx.debug?
+        end
+
+        def form_data(liquid_snippet)
+          template = ["", "", liquid_snippet, "", liquid_template].join("\n")
+
+          {
+            "replace_templates[sections/announcement-bar.liquid]" => template,
```
Poitrin, line 53: Could there be themes that don't have `sections/announcement-bar.liquid`?

packages/cli-kit/assets/cli-ruby/lib/shopify_cli/theme/repl/auth_middleware.rb
```diff
@@ -0,0 +1,74 @@
...
+        end
+
+        def authenticate!
+          @repl.authenticate(storefront_session, secure_session)
+        end
+
+        def shutdown
+          Thread.new do
+            ##
+            # Web server answers the request and shutdown itself
```
Poitrin, line 57: Do we need the `##`?

packages/cli-kit/assets/cli-ruby/lib/shopify_cli/theme/repl/resources/success.html
```diff
@@ -0,0 +1,79 @@
...
+      body {
+        font-size: 26px;
+        line-height: normal;
+        margin: 0;
+        padding: 0;
+      }
+
+      button, input, optgroup, select, textarea {
+        font-family: inherit;
+      }
```
Poitrin, line 26: /nit – Do we need these styles?

packages/cli-kit/assets/cli-ruby/lib/shopify_cli/theme/repl/resources/template.liquid
```diff
@@ -0,0 +1,15 @@
+{% schema %}
+{
+  "name": "t:sections.announcement-bar.name",
```
Poitrin, line 3: I'm just curious: Why (hijacking?) the announcement-bar, and not X, Y, Z? :D

Poitrin (changes requested):
10-50-piw1s-8h9f7
10-51-48zre-ekhbq

Alas not working with my relatively Dawn-ish looking store.
Happy to share theme.

packages/cli-kit/assets/cli-ruby/lib/shopify_cli/theme/repl/resources/template.liquid
Poitrin: And if a theme has no announcement bar (if that exists)?

packages/cli-kit/assets/cli-ruby/lib/shopify_cli/theme/repl/auth_middleware.rb
karreiro: I just removed it :)

packages/cli-kit/assets/cli-ruby/lib/shopify_cli/theme/repl/api.rb
karreiro: Yes! I that's fixed in the new commit :)

Poitrin (approved):
I tried with erroneous commands, in light/dark mode, on macOS and Windows. I did not find any further errors → LGTM :+1:
Just one /nit – when I `Ctrl+D` e.g. in `python` or `irb`, exit code is `0`. In this PR’s Liquid REPL, `Ctrl+C` leads to exit code 1. Depending on our users’ feedback and the subsequent adoption of this feature, we might change the exit code :)

PR #1346, opened 2023-02-10
Bundled cli-ruby used by default

alvaro-shopify:
Some problems have appeared using the embedded version of `cli-ruby` to run `themes` commands or `dev` command with `theme-app-extensions`
Those problems seem to be related with the method used to install Ruby requirements in the local environment. There are also some problems with `theme-app-extensions` in windows as well

- Every command related with `themes` and `theme-app-extension` will use the bundled version of `cli-ruby`
- Add a the new env variable `SHOPIFY_CLI_EMBEDDED_THEME_CLI` to run the command using the embedded version.

### How to test your changes?
- Run any `theme` command with `--verbose` flag and you should see an output similar to this:

```
Running system process:
  · Command: bundle exec shopify theme check /Users/alvarogutierrez/cli-themes/first-theme --output text
  · Working directory: /Users/alvarogutierrez/cli-themes/first-theme 
```

### Measuring impact

### Checklist

### DIFF
--- a/.changeset/spotty-ways-jump.md
+++ b/.changeset/spotty-ways-jump.md
@@ -1,0 +1,5 @@
+---
+'@shopify/cli-kit': patch
+---
+
+Use bundled Ruby CLI by default
--- a/packages/cli-kit/src/public/node/ruby.test.ts
+++ b/packages/cli-kit/src/public/node/ruby.test.ts
@@ -1,8 +1,7 @@
-import {execCLI2, RubyCLIVersion} from './ruby.js'
+import {execCLI2} from './ruby.js'
 import {captureOutput} from './system.js'
 import * as system from './system.js'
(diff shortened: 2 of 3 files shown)


Michael-Gibbons:
@alvaro-shopify I believe I am having similar issues relating to this, I would appreciate any insight you can give. When attempting to deploy an app to production (hosted on digital ocean if that matters), `shopify app deploy` is run. This command also bundles and pushes theme extensions. Part of which is running theme check on the extension code. By default the server is running on 2.5 whereas shopify needs 2.7. Upgrading the ruby version server side should fix probably but this seems a bit clunky no?

image

gonzaloriestra:
@Michael-Gibbons we know depending on Ruby is annoying and we are working on that (it will take some time). But Buildpack should work with recent Ruby versions, no?

Issue #1314, opened 2023-02-08
[Bug]: [LoadError] cannot load such file -- wdm


superfein:
### Please confirm that you have:

- [X] Searched [existing issues](.) to see if your issue is a duplicate. (If you’ve found a duplicate issue, feel free to add additional information in a comment on it.)
- [X] Reproduced the issue in [the latest CLI version](https://www.npmjs.com/package/@shopify/cli).

### In which of these areas are you experiencing a problem?

Theme

### Expected behavior

When I type `shopify theme dev` in the terminal the "viewing theme" details appear with no wdm-related error message.

### Actual behavior

When I type `shopify theme dev` in the terminal the "viewing theme" details appear but below them this error appears:

```
[Note] You cannot use gems with Shopify CLI.
[LoadError] cannot load such file -- wdm
       They are disabled.
       Please don't modify the CLI locally.
       If you would like to contribute to the CLI project, please refer to
       https://github.com/Shopify/shopify-cli/blob/main/.github/CONTRIBUTING.md

  Please add the following to your Gemfile to avoid polling for changes:
    gem 'wdm', '>= 0.1.0' if Gem.win_platform?
```

### Reproduction steps

1. Type `shopify theme dev` in the terminal on Windows 10
2. Error appears below "viewing theme" details
3. Tried `gem install wdm` in the terminal which did not solve the issue. 

### Operating System

Windows 10

### Shopify CLI version (check your project's `package.json` if you're not sure)

3.40.3

### Shell

Git Bash

### Node version (run `node -v` if you're not sure)

v16.17.1

### What language and version are you using in your application?

Ruby v2.7.6

gonzaloriestra:
Hi! Thanks for creating the issue. It should be fixed in the latest version (3.43.0).

superfein:
Yes it is fixed in the latest version, thanks @gonzaloriestra!

tarin-swanky:
I'm on the latest version and it wasn't fixed for me

PR #1350, opened 2023-02-10
Added windows gem depencies to embedded cli-ruby

alvaro-shopify:
Related to https://github.com/Shopify/cli/issues/1314

Windows gem dependencies needed for the `theme-app-extension` are not installed when the embedded `cli-ruby` version is used so this error is displayed

- Added windows gem dependencies when the embedded `cli-ruby` is used

### How to test your changes?
- Run a `dev` command for an app with `theme-app-extension` in windows.

### Measuring impact

### Checklist

### DIFF
--- a/packages/cli-kit/src/public/node/ruby.test.ts
+++ b/packages/cli-kit/src/public/node/ruby.test.ts
@@ -9,7 +9,7 @@ vi.mock('./system')
 
 describe('execCLI', () => {
   it('throws an exception when Ruby is not installed', async () => {
-    vi.mocked(file.fileExists).mockResolvedValue(true)
+    mockGemfileCreation()
     vi.mocked(captureOutput).mockRejectedValue({})
 
     await expect(() => execCLI2(['args'])).rejects.toThrowError('Ruby environment not found')
@@ -17,7 +17,7 @@ describe('execCLI', () => {
 
   it('throws an exception when Ruby version requirement is not met', async () => {
     const rubyVersion = '2.2.0'
-    vi.mocked(file.fileExists).mockResolvedValue(true)
+    mockGemfileCreation()
     vi.mocked(captureOutput).mockResolvedValueOnce(rubyVersion)
 
     await expect(() => execCLI2(['args'])).rejects.toThrowError(
@@ -27,7 +27,7 @@ describe('execCLI', () => {
 
   it('throws an exception when Bundler is not installed', async () => {
     const rubyVersion = '2.7.5'
-    vi.mocked(file.fileExists).mockResolvedValue(true)
+    mockGemfileCreation()
     vi.mocked(captureOutput).mockResolvedValueOnce(rubyVersion)
     vi.mocked(captureOutput).mockRejectedValue({})
 
@@ -37,7 +37,7 @@ describe('execCLI', () => {
   it('throws an exception when Bundler version requirement is not met', async () => {
     const rubyVersion = '2.7.5'
     const bundlerVersion = '2.2.0'
-    vi.mocked(file.fileExists).mockResolvedValue(true)
+    mockGemfileCreation()
     vi.mocked(captureOutput).mockResolvedValueOnce(rubyVersion)
     vi.mocked(captureOutput).mockResolvedValueOnce(bundlerVersion)
 
@@ -53,7 +53,7 @@ describe('execCLI', () => {
     // Given
     const rubyVersion = '2.7.5'
     const bundlerVersion = '2.4.0'
-    vi.mocked(file.fileExists).mockResolvedValue(true)
+    mockGemfileCreation()
     vi.mocked(captureOutput).mockResolvedValueOnce(rubyVersion)
     vi.mocked(captureOutput).mockResolvedValueOnce(bundlerVersion)
     vi.mocked(file.mkdir).mockRejectedValue({message: 'Error'})
@@ -76,7 +76,7 @@ describe('execCLI', () => {
 
     process.env = {...originalEnv, SHOPIFY_CLI_2_0_DIRECTORY: './CLI2', SHOPIFY_RUN_AS_USER: '1'}
 
-    vi.mocked(file.fileExists).mockResolvedValue(true)
+    mockGemfileCreation()
     vi.mocked(captureOutput).mockResolvedValueOnce('2.7.5')
     vi.mocked(captureOutput).mockResolvedValueOnce('2.4.0')
 
@@ -114,7 +114,7 @@ describe('execCLI', () => {
 
     process.env = {...originalEnv, SHOPIFY_CLI_2_0_DIRECTORY: '/manual', SHOPIFY_RUN_AS_USER: '1'}
 
-    vi.mocked(file.fileExists).mockResolvedValue(true)
+    mockGemfileCreation()
     vi.mocked(captureOutput).mockResolvedValueOnce('2.7.5')
     vi.mocked(captureOutput).mockResolvedValueOnce('2.4.0')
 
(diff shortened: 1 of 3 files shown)


packages/cli-kit/src/public/node/ruby.ts
```diff
@@ -41,7 +41,9 @@ interface ExecCLI2Options {
  * @param options - Options to customize the execution of cli2.
  */
 export async function execCLI2(args: string[], options: ExecCLI2Options = {}): Promise<void> {
-  const embedded = (await isShopify()) || isTruthy(process.env.SHOPIFY_CLI_EMBEDDED_THEME_CLI)
+  const embedded =
+    ((await isShopify()) && process.env.SHOPIFY_CLI_EMBEDDED_THEME_CLI === undefined) ||
```
pepicrft, line 45: What happens if I set `SHOPIFY_CLI_EMBEDDED_THEME_CLI=false` or `SHOPIFY_CLI_EMBEDDED_THEME_CLI=0`? I'd recommend adding an `isFalsy` utility like [this one](https://github.com/Shopify/cli/blob/ad5ba3eb7260815cd8561c96770677719b16f178/packages/cli-kit/src/private/node/context/utilities.ts#L4) for truthiness and using that one instead.

packages/cli-kit/src/public/node/ruby.ts
```diff
@@ -254,14 +256,10 @@ async function createThemeCheckCLIWorkingDirectory(): Promise<void> {
...
-  const {platform} = platformAndArch()
-  if (platform === 'windows') {
-    // 'wdm' is required by 'listen', see https://github.com/Shopify/cli/issues/780
-    gemFileContent.push("gem 'wdm', '>= 0.1.0'")
-  }
-  await file.writeFile(gemPath, gemFileContent.join('\n'))
+  const directory = await shopifyCLIDirectory()
+  await addContentToGemfile(directory, "source 'https://rubygems.org'")
+  await addContentToGemfile(directory, `gem 'shopify-cli', '${RubyCLIVersion}'`)
+  await addWindowsDependencies(directory)
```
pepicrft, line 262: This can be one IO operation instead of 3:
- I create the content of the Gemfile in memory (as a string)
- Then I write the content to the `Gemfile` creating it if doesn't exist.

packages/cli-kit/src/public/node/ruby.ts
pepicrft: Also, it's a bit unclear what `embedded` means here and why the logic changed from what we had before. What's the connection between "embedded" and (is Shopifolk and SHOPIFY_CLI_EMBEDDED_THEME_CLI is not defined)?

packages/cli-kit/src/public/node/ruby.test.ts
```diff
@@ -181,4 +181,51 @@ describe('execCLI', () => {
...
+        SHOPIFY_CLI_ADMIN_AUTH_TOKEN: undefined,
+        SHOPIFY_CLI_STORE: undefined,
+        SHOPIFY_CLI_AUTH_TOKEN: 'token_0000_1111_2222_3333',
+        SHOPIFY_CLI_RUN_AS_SUBPROCESS: 'true',
+        BUNDLE_GEMFILE: `/manual/Gemfile`,
+      },
+    })
+
+    // Teardown
+    process.env = originalEnv
```
pepicrft, line 224: Please don't modify global variables like `process` because that might end up leaking state across tests.

packages/cli-kit/src/public/node/ruby.test.ts
```diff
@@ -181,4 +181,51 @@ describe('execCLI', () => {
...
+  })
 })
+
+function mockGemfileCreation() {
+  vi.mocked(file.fileExists).mockResolvedValue(true)
+  vi.mocked(file.readFile).mockResolvedValue('gemfile' as any)
+}
```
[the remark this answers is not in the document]
pepicrft: I'd reconsider the testing strategy. While mocking dependencies of the subject under test is sometimes unavoidable, like it's the case of `system.exec`, which makes tests more unmaintainable because they depend on implementation details, there's no need to mock file IO APIs and we have plenty of examples throughout the codebase. We can create a temporary working directory that the test runs against. After the subject under test runs, you can go to that directory and check if the content of the `Gemfile` is the expected one.

packages/cli-kit/src/public/node/ruby.test.ts
```diff
@@ -47,138 +54,176 @@ describe('execCLI', () => {
...
+
+      // When
+      await execCLI2(['args'], {
+        token: 'token_0000_1111_2222_3333',
+        directory: './directory',
+      })
+
+      // Then
+      validateBudleExec(execSpy, gemfilePath, joinPath(cli2Directory, 'bin', 'shopify'))
+      await validateGemFileContent(gemfilePath, !isBundled, !isWindows)
```
pepicrft, line 119: When I read this line:
```js
await validateGemFileContent(gemfilePath, !isBundled, !isWindows)
```

I can't tell what the value of `isBundled` and `isWindows` are. I saw them at the root of the module, `const isBundled = true`, and thought you added them there because you'd change the value throughout the various test scenarios, but it turns out you don't.

I'd recommend iterating on your approach to bring clarity. This is useful for your future self, for reviewers, and for contributors coming across this piece of code in the future. If we compare the above with:

```js
await validateGemFileContent(gemfilePath, {
  windows: true,
  isBundled: false,
})
```

I can immediately tell what information is being passed, and what represents.

packages/cli-kit/src/public/node/ruby.test.ts
```diff
@@ -47,138 +54,176 @@ describe('execCLI', () => {
...
+  await createGemFile(cli2Directory, existingWindowsDependency)
+  return vi.spyOn(system, 'exec')
+}
+
+function mockRubyEnvironment() {
+  vi.mocked(captureOutput).mockResolvedValueOnce('2.7.5')
+  vi.mocked(captureOutput).mockResolvedValueOnce('2.4.0')
+}
+
+function mockPlatformAndArch(windows: boolean) {
```
pepicrft, line 184: The following piece of code will make the code hard to understand:

```js
mockPlatformAndArch(true)
```

What about:

```js
mockPlatformAndArch({ isWindows: true })
```

packages/cli-kit/src/public/node/ruby.test.ts
```diff
@@ -47,138 +54,176 @@ describe('execCLI', () => {
...
+    },
+  })
+}
+
+async function validateGemFileContent(gemfilePath: string, bundled: boolean, windows: boolean) {
+  expect(file.fileExists(gemfilePath)).toBeTruthy()
+  const gemContent = await file.readFile(gemfilePath, {encoding: 'utf8'})
+  expect(gemContent).toContain("source 'https://rubygems.org'")
+  if (bundled) expect(gemContent).toContain(`gem 'shopify-cli', '${RubyCLIVersion}'`)
+  const windowsDepency = "gem 'wdm', '>= 0.1.0'"
```
pepicrft, line 221: Can we have this in a constant so that we can update its value without breaking the tests?

packages/cli-kit/src/public/node/ruby.test.ts
alvaro-shopify: I'm completely agree with you. My intention was to be a `declarative` way of mocking or validating, but your suggestion is really better

packages/cli-kit/src/public/node/ruby.test.ts
alvaro-shopify: yes, it's the same approach mentioned [here](https://github.com/Shopify/cli/pull/1350#discussion_r1109484229)

packages/cli-kit/src/public/node/ruby.ts
```diff
@@ -254,14 +254,11 @@ async function createThemeCheckCLIWorkingDirectory(): Promise<void> {
...
-  if (platform === 'windows') {
-    // 'wdm' is required by 'listen', see https://github.com/Shopify/cli/issues/780
-    gemFileContent.push("gem 'wdm', '>= 0.1.0'")
-  }
-  await file.writeFile(gemPath, gemFileContent.join('\n'))
+  const directory = await shopifyCLIDirectory()
+  const content = ["source 'https://rubygems.org'", `gem 'shopify-cli', '${RubyCLIVersion}'`].concat(
+    buildWindowsDependencies(),
+  )
+  await addContentToGemfile(directory, content)
```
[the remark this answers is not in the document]
alvaro-shopify: I'd rather use a list of lines as the content and then the `addContentToGemfile` is the method that handles how to persist it in the file.

pepicrft (approved):

PR #1351, opened 2023-02-10
fix sizing and alignment of tooltip

kris-tremblay:
Tooltip component had a temporary sizing/positioning solution in place to address some weird visual issues.

This PR cleans that up and replaces it with a pure css fix.

image

Previously this tooltip would have additional space at the end, or if the text was longer it suffered from weird breaking.


charlesdobson (approved):
Paired with Kris, lgtm

PR #1353, opened 2023-02-11
[Feature] Add `--no-input` flag to disable input prompts for development

### LINKED ISSUE #1341: [Bug]: Can't run Shopify app in development using Turbo

App

### Expected behavior

The `turbo run dev` command runs the Shopify app's dev command and should start the backend and frontend.

### Actual behavior

The Shopify app crashes and fails to start.

### Stack trace

```shell
ERROR Raw mode is not supported on the current process.stdin, which Ink uses
as input stream by default.
Read about how to prevent this error on
https://github.com/vadimdemedes/ink/#israwmodesupported
```

### Reproduction steps

1. Create monorepo with turborepo: `npx create-turbo@latest`
2. Create Shopify app in the app directory: `yarn create @shopify/app`
3. Run the dev command: `yarn dev`

### Operating System

Mac OS

3.42.0

### Shell

_No response_

14.19.1

_No response_

Cavallando:

Fixes #1341 

As stated in the original issue, when running `shopify app dev` with Vercel's turborepo (or any parent proccess that uses `stdin` i would image) the cli crashes with the following error `ERROR Raw mode is not supported on the current process.stdin, which Ink uses as input stream by default.`.

This PR adds the optional `--no-input` flag to `shopify app dev` to essentially disable `stdin` from being required. The primary use case is usage within a monorepo/parent process that might have its own `stdin` handling. This flag requires the app to first be set up/configured before using. Additionally, it is restricted from being used along side of `--reset` since that is a process that does require prompts (if there are additional commands/flags I missed please let me know!)

I'm not sure that this is the "correct" fix since it doesn't deal with `stdin` directly really, but personally felt this flag would be more friendly for developers in the less common scenario of a monorepo/parent context. Totally open to suggestions/ideas though!

### How to test your changes?

**Note: **If this change is wanted I can totally make a reproducible repo!

1. Setup a new app with `shopify app dev`
2. Run `shopify app dev --no-input` from any parent process that uses `stdin`, (ie. turbo, others) and the app will still run

### Post-release steps

I imagine there's documentation to be updated on the docs page which I am happy to do assuming this change is desired.

### Measuring impact

### Checklist

### DIFF
--- a/packages/app/src/cli/commands/app/dev.ts
+++ b/packages/app/src/cli/commands/app/dev.ts
@@ -66,6 +66,13 @@ export default class Dev extends Command {
       default: false,
       exclusive: ['tunnel-url', 'tunnel'],
     }),
+    'no-input': Flags.boolean({
+      hidden: false,
+      description: 'Disables input prompts. Requires setting up your app first before use.',
+      env: 'SHOPIFY_FLAG_NO_INPUT',
+      default: false,
+      exclusive: ['reset'],
+    }),
     tunnel: Flags.boolean({
       hidden: false,
       description: 'Use ngrok to create a tunnel to your service entry point',
@@ -112,6 +119,7 @@ export default class Dev extends Command {
       noTunnel: flags['no-tunnel'],
       theme: flags.theme,
       themeExtensionPort: flags['theme-app-extension-port'],
+      noInput: flags['no-input'],
     })
   }
 }
--- a/packages/app/src/cli/services/context.ts
+++ b/packages/app/src/cli/services/context.ts
@@ -40,6 +40,7 @@ export interface DevContextOptions {
   apiKey?: string
   storeFqdn?: string
   reset: boolean
+  noInput: boolean
 }
 
 interface DevContextOutput {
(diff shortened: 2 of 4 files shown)


Cavallando:
Hi @pepicrft thanks for getting back to me!

Your feedback totally makes sense to me and I appreciate the insight on your thought process! Happy to close this PR then as a result. (I think a more appropriate fix is to have vercel's turbo cli forward the stdin which there is an [ongoing discussion](https://github.com/vercel/turbo/issues/1235) about)

> That being said, I'm curious about what value you get out of running the CLI through Vercel, and if you've considered having a script where you can triggere the two processes concurrently. 

It is by know means a requirement for me (or most people using turbo frankly) to initiate from turbo but if i run `turbo dev` in my turborepo then it will run all workspaces' `dev` command concurrently out of the box. So in my monorepo I was just hoping for a consistent workflow where I can manage my build and dev pipelines with my own shared internal dependencies. (there is no problem building shopify apps within a turbo repo currently).

There's likely some workaround I can write for my own needs to get that to work with `dev` but worst case it isnt the end of the world if I need to run shopify outside of the turbo context!

PR #1355, opened 2023-02-12
Cache identity introspection endpoint

amcaplan:

Replaces https://github.com/Shopify/cli/pull/712
Fixes https://github.com/Shopify/internal-cli-foundations/issues/489 with the blessing of the identity team

The identity introspection endpoint is supposed to be fetched dynamically, BUT that makes less sense for the CLI which launches a new process frequently while a user sits there, vs. servers spinning up which can take a half-second hit.

This small change will make most command runs up to 0.5s faster on average (since most frequently-used commands require authentication).

Caches the identity introspection endpoint for a period of 1 week. If the existing endpoint 404s, we can retry with fetching the correct endpoint first, as we do now.

It's technically slower in the rare case of the endpoint changing, but honestly I don't think it's changed in the last year, and it's unlikely to change more than once every few months.

**BEFORE: 556ms total**

```
2023-02-12T14:29:28.254Z: Validating existing session against the scopes:
[
  "openid",
  "https://api.shopify.com/auth/shop.admin.graphql",
  "https://api.shopify.com/auth/shop.admin.themes",
  "https://api.shopify.com/auth/partners.collaborator-relationships.readonly",
  "https://api.shopify.com/auth/shop.storefront-renderer.devtools",
  "https://api.shopify.com/auth/partners.app.cli.access"
]
For applications:
{
  "partnersApi": {
    "scopes": []
  }
}

2023-02-12T14:29:28.254Z: 
Sending GET request to URL https://accounts.shopify.com/.well-known/openid-configuration.json and headers:
 - User-Agent: Shopify CLI; v=3.43.0
 - Sec-CH-UA-PLATFORM: darwin
 - X-Request-Id: e98a988f-4150-483e-8142-a360b80fe49f
 - Content-Type: application/json

2023-02-12T14:29:28.801Z: Request to https://accounts.shopify.com/.well-known/openid-configuration.json completed with status 200 in 547 ms
2023-02-12T14:29:28.810Z: Sending Identity Introspection request to URL: https://accounts.shopify.com/oauth/introspection
```

**AFTER: 1ms total**

```
2023-02-12T14:27:57.772Z: Validating existing session against the scopes:
[
  "openid",
  "https://api.shopify.com/auth/shop.admin.graphql",
  "https://api.shopify.com/auth/shop.admin.themes",
  "https://api.shopify.com/auth/partners.collaborator-relationships.readonly",
  "https://api.shopify.com/auth/shop.storefront-renderer.devtools",
  "https://api.shopify.com/auth/partners.app.cli.access"
]
For applications:
{
  "partnersApi": {
    "scopes": []
  }
}

2023-02-12T14:27:57.773Z: Sending Identity Introspection request to URL: https://accounts.shopify.com/oauth/introspection
```

### How to test your changes?

Run any authenticated command in `--verbose` mode. (NOTE: Run it twice to see the caching actually working!)

### Measuring impact

### Checklist

### DIFF
--- a/packages/cli-kit/src/private/node/conf-store.ts
+++ b/packages/cli-kit/src/private/node/conf-store.ts
@@ -57,20 +57,21 @@ export function removeSession(config: LocalStorage<ConfSchema> = cliKitStore()):
   config.delete('sessionStore')
 }
 
+type CacheValueForKey<TKey extends keyof Cache> = NonNullable<Cache[TKey]>['value']
 /**
  * Fetch from cache, or run the provided function to get the value, and cache it
  * before returning it.
  * @param key - The key to use for the cache.
- * @param fn - The function to run to get the value to cache.
- * @param timeout - The number of milliseconds to cache the value for.
+ * @param fn - The function to run to get the value to cache, if a cache miss occurs.
+ * @param timeout - The maximum valid age of a cached value, in milliseconds. If the cached value is older than this, it will be refreshed.
  * @returns The value from the cache or the result of the function.
  */
-export async function cacheFetch(
+export async function cacheRetrieveOrRepopulate(
   key: keyof Cache,
-  fn: () => Promise<NonNullable<Cache[typeof key]>['value']>,
+  fn: () => Promise<CacheValueForKey<typeof key>>,
   timeout?: number,
   config = cliKitStore(),
-): Promise<NonNullable<Cache[typeof key]>['value']> {
+): Promise<CacheValueForKey<typeof key>> {
   const cache: Cache = config.get('cache') || {}
   const cached = cache[key]
 
--- a/packages/cli-kit/src/private/node/conf-store.ts
+++ b/packages/cli-kit/src/private/node/conf-store.ts
@@ -1,16 +1,18 @@
 import {LocalStorage} from '../../public/node/local-storage.js'
 import {outputContent, outputDebug} from '@shopify/cli-kit/node/output'
 
+interface CacheValue<T> {
+  value: T
+  timestamp: number
+}
+
 interface Cache {
-  [key: string]: {
-    value: unknown
-    timestamp: number
-  }
+  'identity-introspection-url'?: CacheValue<string>
 }
 
 export interface ConfSchema {
   sessionStore: string
-  cache: Cache
+  cache?: Cache
 }
 
 let _instance: LocalStorage<ConfSchema> | undefined
@@ -63,17 +65,17 @@ export function removeSession(config: LocalStorage<ConfSchema> = cliKitStore()):
  * @param timeout - The number of milliseconds to cache the value for.
  * @returns The value from the cache or the result of the function.
  */
-export async function cacheFetch<T>(
-  key: string,
-  fn: () => Promise<T>,
+export async function cacheFetch(
+  key: keyof Cache,
+  fn: () => Promise<NonNullable<Cache[typeof key]>['value']>,
   timeout?: number,
   config = cliKitStore(),
-): Promise<T> {
+): Promise<NonNullable<Cache[typeof key]>['value']> {
   const cache: Cache = config.get('cache') || {}
   const cached = cache[key]
 
   if (cached && (timeout === undefined || Date.now() - cached.timestamp < timeout)) {
-    return cached.value as T
+    return cached.value
   }
 
   const value = await fn()
--- a/packages/cli-kit/src/private/node/conf-store.ts
+++ b/packages/cli-kit/src/private/node/conf-store.ts
@@ -1,8 +1,16 @@
 import {LocalStorage} from '../../public/node/local-storage.js'
 import {outputContent, outputDebug} from '@shopify/cli-kit/node/output'
 
+interface Cache {
+  [key: string]: {
+    value: any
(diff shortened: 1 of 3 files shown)


packages/cli-kit/src/private/node/session/identity-token-validation.ts
```diff
@@ -1,40 +1,67 @@
...
+export async function validateIdentityToken(token: string): Promise<boolean> {
   try {
-    const instrospectionURL = await getInstrospectionEndpoint()
-    const options = {
-      method: 'POST',
-      headers: {Authorization: `Bearer ${token}`, 'Content-Type': 'application/json'},
-      body: JSON.stringify({token}),
-    }
-    outputDebug(`Sending Identity Introspection request to URL: ${instrospectionURL}`)
+    return withIntrospectionURL<boolean>(async (introspectionURL: string) => {
```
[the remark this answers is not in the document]
amcaplan: Yes, that's how we do a retry, see below.

packages/cli-kit/src/private/node/conf-store.ts
```diff
@@ -3,12 +3,19 @@ import Conf, {Schema} from 'conf'
...
+interface Cache {
+  [key: string]: {
+    value: unknown
+    timestamp: number
+  }
+}
+
 export interface ConfSchema {
   sessionStore: string
+  cache: Cache
```
pepicrft, line 15: I'm a bit concerned with the approach we are taking here. Cache's schema is very loose, which makes it the perfect place to store any value with a timestamp. If that ends up happening, it won't be straightforward to know what state is stored in it, and thus, which changes that we are shipping are not breaking whatever state exists there.

I recommend keeping the configuration schema flat and strong so instead of calling this attribute `cache`, I'd call it `identityIntrospectionEndpointCachedValue` and **make it optional**. Introducing schema values that are non-optional is a breaking change.

packages/cli-kit/src/private/node/session/identity-token-validation.ts
```diff
@@ -1,40 +1,67 @@
 import {identityFqdn} from '../../../public/node/context/fqdn.js'
 import {outputDebug} from '../../../public/node/output.js'
 import {shopifyFetch} from '../../../public/node/http.js'
+import {cacheFetch} from '../conf-store.js'
 
-export async function validateIdentityToken(token: string) {
+class NotFoundError extends Error {
```
pepicrft, line 6: `Error` is not designed to be subclassed. If you try to do things like `instanceof` to check if an error is an instance of some class it'll yield unexpected values. I'd consider using [`ExtendableError`](https://github.com/Shopify/cli/blob/main/packages/cli-kit/src/public/node/error.ts#L7) or better, adjust the interface of this function to use the `Result` type, which is a better option to delegate to the caller what to do based on the obtained result.

packages/cli-kit/src/private/node/session/identity-token-validation.ts
```diff
@@ -1,40 +1,67 @@
...
 }
 
-async function getInstrospectionEndpoint(): Promise<string> {
+async function withIntrospectionURL<T>(fn: (introspectionUrl: string) => Promise<T>): Promise<T> {
+  const week = 7 * 24 * 60 * 60 * 1000
+  const introspectionURL = await cacheFetch('identity-introspection-url', getIntrospectionURL, week)
+  try {
+    return fn(introspectionURL)
+  } catch (error) {
+    if (error instanceof NotFoundError) {
```
pepicrft, line 54: Note that the `catch` block assumes that `fn` will use the error `NotFoundError` type to communicate that the token was not found. While that's an obvious assumption at the time this PR is being written, it might not be obvious for someone looking at this code in months from now. I'd recommend using `Result` types here:

```js
const result = await fn(introspectionURL)
// Type the result.error instance such that you can do an exhaustive check on the various error types.
```

packages/cli-kit/src/private/node/conf-store.ts
```diff
@@ -48,3 +57,30 @@ export function removeSession(config: Conf<ConfSchema> = cliKitStore()): void {
...
+
+/**
+ * Fetch from cache, or run the provided function to get the value, and cache it
+ * before returning it.
+ * @param key - The key to use for the cache.
+ * @param fn - The function to run to get the value to cache.
+ * @param timeout - The number of milliseconds to cache the value for.
+ * @returns The value from the cache or the result of the function.
+ */
+export async function cacheFetch(
```
pepicrft, line 69: Since we are already using classes for the store instances, have we considered having a subclass for the CLIKit Store fixing it to the CLIKit schema? Then we can move this function to that class and the ergonomics would improve:

```js
cliKitStore().cacheFetch()
```

I'd also suggest renaming the function to something that's more explicit about what it does:

```js
// Inspired by ActiveRecord APIs
store.fetchOrInvalidateFromCache()
```

amcaplan: Alright, I went with `cacheRetrieveOrRepopulate` which is maybe better.

Also, I don't currently see the major value over the current organization. I think in the past I felt that using classes on conf store ended up making the code more complex, not less.

pepicrft (approved):

PR #1357, opened 2023-02-12
Parallelize partners requests

amcaplan:

Partners requests are the heaviest part of initiating the `app dev` command. By running some of them in parallel (plus a few other small optimizations), we can shave about 1.3 seconds off command startup in most cases. This makes for a much snappier CLI experience.

1. Take Partners API GraphQL queries that can be run in parallel (i.e. they don't depend on each other) and make them parallel.
2. Use keepAlive to ensure we don't waste time re-establishing the connection to the Partners API each time
3. Eliminate waste. Don't save the session to secure storage (which takes about 20-30ms) if the session hasn't changed.

### How to test your changes?

Make sure `app dev` still works in `--reset`, cached, and non-cached cases.

### Measuring impact

### Checklist

### DIFF
--- a/packages/app/src/cli/services/context.ts
+++ b/packages/app/src/cli/services/context.ts
@@ -152,7 +152,7 @@ export async function ensureDevContext(options: DevContextOptions, token: string
   const [organization, _selectedApp, _selectedStore] = await Promise.all([
     organizationFromId(orgId, token),
     selectedApp ? selectedApp : appFromId(cachedInfo?.appId, token),
-    selectedStore ? selectedStore : storeFromFqdn(cachedInfo?.storeFqdn, orgId, token)
+    selectedStore ? selectedStore : storeFromFqdn(cachedInfo?.storeFqdn, orgId, token),
   ])
 
   if (_selectedApp) {
@@ -160,14 +160,14 @@ export async function ensureDevContext(options: DevContextOptions, token: string
   } else {
     const {apps} = await fetchOrgAndApps(orgId, token)
     const localAppName = await loadAppName(options.directory)
-    selectedApp = await selectOrCreateApp(localAppName, apps, organization!, token)
+    selectedApp = await selectOrCreateApp(localAppName, apps, organization, token)
   }
 
   if (_selectedStore) {
     selectedStore = _selectedStore
   } else {
     const allStores = await fetchAllDevStores(orgId, token)
-    selectedStore = await selectStore(allStores, organization!, token)
+    selectedStore = await selectStore(allStores, organization, token)
   }
 
   setAppInfo({
@@ -201,7 +201,11 @@ const appFromId = async (appId: string | undefined, token: string): Promise<Orga
   return app
 }
 
-const storeFromFqdn = async (storeFqdn: string | undefined, orgId: string, token: string): Promise<OrganizationStore | undefined> => {
+const storeFromFqdn = async (
+  storeFqdn: string | undefined,
+  orgId: string,
+  token: string,
+): Promise<OrganizationStore | undefined> => {
   if (!storeFqdn) return
   const result = await fetchStoreByDomain(orgId, token, storeFqdn)
   if (result?.store) {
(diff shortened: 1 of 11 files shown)


amcaplan:
I ran some more tests locally, seems like the time to reach the point of being ready to serve has dropped (on my machine, connection, etc.) from ~12.5 seconds down to ~8.5. Adding [this other PR](https://github.com/Shopify/cli/pull/1355) will drop us down to ~7 I'd guess.

packages/cli-kit/src/private/node/api/headers.ts
```diff
@@ -59,9 +61,13 @@ export function buildHeaders(token?: string): {[key: string]: string} {
...
  */
-export async function httpsAgent() {
-  return new https.Agent({rejectUnauthorized: await shouldRejectUnauthorizedRequests()})
+async function _httpsAgent(_id: string = 'default'): Promise<https.Agent> {
+  return new https.Agent({
+    rejectUnauthorized: await shouldRejectUnauthorizedRequests(),
+    keepAlive: true,
+  })
 }
+export const httpsAgent = memoize(_httpsAgent)
```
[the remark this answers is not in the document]
isaacroldan: Not sure how much time this saves, but it makes sense to only create 1 httpsAgent in the CLI lifecycle 👌

packages/app/src/cli/services/dev.ts
```diff
@@ -158,11 +158,13 @@ async function dev(options: DevOptions) {
...
-    const adminSession = await ensureAuthenticatedAdmin(storeFqdn)
-    const storefrontToken = await ensureAuthenticatedStorefront()
     const extension = localApp.extensions.theme[0]!
-    const args = await themeExtensionArgs(extension, apiKey, token, options)
-    const devExt = await devThemeExtensionTarget(args, adminSession, storefrontToken, token)
+    const [adminSession, storefrontToken, args] = await Promise.all([
+      ensureAuthenticatedAdmin(storeFqdn),
+      ensureAuthenticatedStorefront(),
+      themeExtensionArgs(extension, apiKey, token, options),
+    ])
```
isaacroldan, line 166: I don't think this is safe, if you run `ensureAuthenticatedAdmin` and `ensureAuthenticatedStorefront` at the same time, and the session is expired, it might try to trigger 2 authentication processes.

packages/app/src/cli/services/context.ts
```diff
@@ -149,47 +149,53 @@ export async function ensureDevContext(options: DevContextOptions, token: string
...
+      if (selectedStore) return selectedStore
+      if (cachedInfo?.storeFqdn) {
+        const result = await fetchStoreByDomain(orgId, token, cachedInfo.storeFqdn)
+        if (result?.store) {
+          await convertToTestStoreIfNeeded(result.store, orgId, token)
+          return result.store
+        } else {
+          throw new BugError(`Couldn't find Store with domain ${cachedInfo.storeFqdn}.`)
+        }
+      }
```
isaacroldan, line 176: could you extract each promise code into its own function? 
With the new changes (which I think make sense) it's a bit more difficult to read the code

pepicrft: Agreed. It'd have a positive impact on the readability.

packages/app/src/cli/services/dev.ts
isaacroldan: yep, that's possible, we can have a convenience function to login into all 3 APIs at the same time.
Although the storefront token is only needed if there are theme extensions, but if we request them in parallel it should be OK

packages/cli-kit/src/private/node/session.ts
```diff
@@ -134,7 +134,8 @@ ${outputToken.json(applications)}
   }
 
   const completeSession: Session = {...currentSession, ...newSession}
-  await secureStore.store(completeSession)
+  // Save the new session info if it has changed
+  if (Object.keys(newSession).length > 0) await secureStore.store(completeSession)
```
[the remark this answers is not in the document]
pepicrft: Nop! Go ahead

PR #1372, opened 2023-02-14
[Feature] Add a new type of app extension called Configuration Extension.

sunithapatel:

Currently the Shopify CLI only supports 3 types of app extensions - UI, Theme and Functions. There is another type of app extension that only has configuration which is in a `toml` file and we need a way to generate and deploy these types of extensions. It does not fit in the current 3 types.

This PR is an attempt to add this new type of extension called Configuration Extension. (Name needs to be revisited 😄)

(This is a draft; remaining sections will be filled in soon)

- Adds a new type of app extension that only needs configuration (that is, a generated `toml` file with no other code)
- Changes the preview message after running the `shopify app generate extension` command since `npm run dev` is not needed.
- When running `shopify app deploy`, it deploys the configuration for the extension that was provided in the `toml` file.

Output of `shopify app generate extension`:
(there was no prompt for supported flavors since there is no code as part of this extension type)

image

*Currently `shopify app deploy` is not passing validation. I am looking into this part.

### Measuring impact

### Checklist


vividviolet:
@sunithapatel I don't think a config extension is the right choice for this. Instead of Shopify needing to know the endpoints urls and etc to call to calculate taxes, wouldn't it make more sense of the extension to take an input and produce an output by calling whatever services it needs? This extension is better suited as a ui-less extension instead. We already have several examples of these:
- Checkout cart validation extension
- Web pixel extension
- Admin customer segmentation template extension

sunithapatel:
Hi @vividviolet! What you are describing sounds like Shopify Functions which has limitations currently that prevents us from using that today for tax calculations. This also has no UI involved, so it is a ui-less extension as you put it. However, the only extension type that the CLI seems to have for this is functions. That is why this PR is for those app extensions where the configuration is what is needed and there is no UI involved with the use of the extension.

isaacroldan:
For me `static` extensions works better than `configuration` extensions

PR #1374, opened 2023-02-14
Enable navigating directly to the embedded app when the Dev Console is embedded inside Admin

vividviolet:

Fix the navigation to App Home when the Dev Console is embedded inside Shopify Admin

https://user-images.githubusercontent.com/29458473/218873025-3c2caa87-7bee-41ab-9345-8b836b680ea3.mov

Populate the `resourceUrl` which will be used when sending the `navigate` event through the Dev Console when it's embedded as an iframe. The resource url is only populated when the Dev Console is embedded and the surface has been set to  "admin". I'm using the app's handle because only embedded apps have handles and can be opened inside of Admin.

### How to test your changes?

The unit tests should be sufficient.

### Measuring impact

### Checklist

### DIFF
--- a/packages/ui-extensions-dev-console/src/sections/Extensions/components/AppHomeRow/AppHomeRow.test.tsx
+++ b/packages/ui-extensions-dev-console/src/sections/Extensions/components/AppHomeRow/AppHomeRow.test.tsx
@@ -1,6 +1,6 @@
 import {AppHomeRow} from '.'
 import en from './translations/en.json'
-import {QRCodeModal} from '..'
+import {PreviewLink, QRCodeModal} from '..'
 import React from 'react'
 
 import {DefaultProviders} from 'tests/DefaultProviders'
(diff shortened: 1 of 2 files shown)


packages/ui-extensions-dev-console/src/sections/Extensions/components/AppHomeRow/AppHomeRow.test.tsx
```diff
@@ -53,4 +53,40 @@ describe('<AppHomeRow/>', () => {
...
+  test("renders a <PreviewLink/> with the resource url set to the app's handle if the surface has been set to 'admin'", () => {
+    const appState = {
+      app: {url: 'mock.url', title: 'Mock App Title', handle: 'my-app-handle'},
+    }
+    const container = render(<AppHomeRow />, withProviders(DefaultProviders), {
+      state: appState,
+      client: {options: {surface: 'admin'}},
+    })
+
+    expect(container).toContainReactComponent(PreviewLink, {resourceUrl: '/admin/apps/my-app-handle'})
```
isaacroldan, line 78: Isn't this test exactly the same as the one above, but the expected resourceUrl is different? 🤔

vividviolet: Oops, sorry that's an accidental change - the second one is correct. It's concerning that the first test passes though. I'll look into it

vividviolet: Actually, the test _was_ failing as expect! I'll clean it up

isaacroldan (approved):

PR #1378, opened 2023-02-15
Improve error handling in proxy 

Arkham:

Whenever one of the processes connected to the proxy crashes, the CLI crashes as well.

Add error handling to the proxy we use for web requests and websockets.

### Measuring impact

### Checklist

### DIFF
--- a/packages/app/src/cli/utilities/app/http-reverse-proxy.ts
+++ b/packages/app/src/cli/utilities/app/http-reverse-proxy.ts
@@ -27,7 +27,7 @@ export interface ReverseHTTPProxyTarget {
 
 interface Options {
   previewUrl: string | undefined
-  portNumber: number | undefined
+  portNumber: number
   proxyTargets: ReverseHTTPProxyTarget[]
   additionalProcesses: OutputProcess[]
 }
@@ -43,7 +43,7 @@ interface Options {
  */
 export async function runConcurrentHTTPProcessesAndPathForwardTraffic({
   previewUrl,
-  portNumber = undefined,
+  portNumber,
   proxyTargets,
   additionalProcesses,
 }: Options): Promise<void> {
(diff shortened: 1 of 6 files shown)


packages/app/src/cli/utilities/app/http-reverse-proxy.ts
```diff
@@ -66,18 +66,20 @@ export async function runConcurrentHTTPProcessesAndPathForwardTraffic({
...
 ${outputToken.json(JSON.stringify(rules))}
 `)
 
   const proxy = httpProxy.createProxy()
   const server = http.createServer(function (req, res) {
     const target = match(rules, req)
-    if (target) return proxy.web(req, res, {target})
+    if (target) {
+      return proxy.web(req, res, {target}, (err) => {
+        outputWarn(`Error forwarding web request: ${err}`)
```
isaacroldan, line 80: For ruby apps, does this mean we are almost always going to show this warning while the app is starting up?

Arkham: Yup!

isaacroldan (approved):

packages/app/src/cli/utilities/app/http-reverse-proxy.ts
pepicrft: What about having state scoped to the target that indicates whether it's running or not. Then we can use it to:
- Output it as a `debug` (when we pass `--verbose`) if the server is starting.
- Output it  as a warning always if the server was started and all of a sudden we started getting errors.

Arkham: makes sense, i'd also like to get it rendered with the other processes inside `renderConcurrent`

PR #1385, opened 2023-02-15
Add try/cach to extension bundle

david-mccullars:
Temporarily hack the built javascript bundle for extensions by adding a try/catch around the entire script.  This ensures that any unexpected error that may occur gets output to console log and not swallowed by our web worker.

When developing native UI extensions for POS, it is possible to get into a state where the extension fails to load. The Smart Grid tile just says "Error loading extension," but there is no further information available. Trying to use chrome://inspect shows no console error messages whatsoever, leaving a partner with no good avenue to debug.

This PR is a possible short-term solution that ensures that the extension bundle that we build is wrapped in a try/catch so that any error is properly caught and output to the console. Long-term, we should consider modifying the web worker which calls `importScript` to either do a try/catch or ensure that errors are properly reported.

### How to test your changes?

One easy way of getting into this state is to create a new POS UI extension and then to make the following change to your src/index.js file:

```javascript
throw new Error("Simulate a catastrophic error");
render('Retail::SmartGrid::Tile', () => <SmartGridTile />);
render('Retail::SmartGrid::Modal', () => <SmartGridModal />);
```

### Measuring impact

### Checklist

### DIFF
--- a/packages/app/src/cli/services/extensions/bundle.ts
+++ b/packages/app/src/cli/services/extensions/bundle.ts
@@ -56,6 +56,14 @@ export async function bundleExtension(options: BundleOptions) {
       }
     })
   }
+
+  // TEMPORARY - hack the built javascript bundle by adding a try/catch around the entire script.
+  // This ensures that any unexpected error that may occur gets output to console log and not
+  // swallowed by our web worker.
+  const fs = require('fs');
+  const mainJs = fs.readFileSync(esbuildOptions.outfile, 'utf8')
+  const safeJs = `try{${mainJs}}catch(e){console.log('Unexpected Error Loading Script',e,e.stack)}`
+  fs.writeFileSync(esbuildOptions.outfile, safeJs, { encoding: 'utf8' })
   onResult(result, options)
 }
 


vividviolet:
Can you clarify why this issue happens on POS? Should this not be happening for all extensions if it's related to importing an erroneous script? Also, I'm not sure I'm comfortable with adding a try/catch around the partner's code. This would prevent errors from being caught higher up the chain while loading extensions which we might be already doing.

Wouldn't it make more sense for POS or other hosts to do `try/catch` when loading in the partner's script instead?

vividviolet:
The proper place would be inside the worker itself before calling importScript. Basically inside the worker.load method.

david-mccullars:
> Wouldn't it make more sense for POS or other hosts to do try/catch when loading in the partner's script instead?

It absolutely would. I spent the better part of two days trying to accomplish this, but I could never get the worker code to build. (I'm not even 100% certain I was in the right place.) This PR wasn't really intended to be a real PR but was put together to help @henrytao-me see the issues I was experiencing when an extension fails to load without communicating anything meaningful to end users. However, partners are running into the same issue, and @heltisace was thinking this might provide some relief until we can figure out the long-term solution. We definitely welcome any thoughts or advice!

heltisace:
> The proper place would be inside the worker itself before calling importScript. Basically inside the worker.load method.

@vividviolet would you be able to help us do that?

vividviolet:
Yes, I can. I think you are using the same worker script from web correct? @henrytao-me do you remember?

heltisace:
Here is a cdn link to the worker we are using: https://cdn.shopify.com/shopifycloud/web/assets/v1/admin-ui-extensions-host/sandbox-worker.worker.js.

vividviolet:
Yup, that's the same worker used by web. It's strange that we can't catch errors outside of the worker though, I thought this PR to Remote UI was supposed to let us do that: https://github.com/Shopify/remote-ui/pull/173

Have you tried bumping the remote-ui version?

heltisace:
After updating the package on the host (POS), the console now throws the errors in the inspect console as expected. In this example the function wasn't imported or created.

I tried to update the client (UI Extensions) too, but it didn't work. This could have been a local issue though, because I used `yalc` to link the UI Extensions update to POS and the extension. 

Next steps: 
- I will create a PR in the POS repo to update the packages.
- We will communicate to the partners that this will be improved in the next version of POS.
- We will look into making it more clear on how to debug UI Extensions so that the partners don't expect the CLI console to present the errors. 

Thanks everyone!

cc: @vividviolet @david-mccullars @henrytao-me 

Screenshot 2023-02-23 at 3 59 46 PM

vividviolet:
>  tried to update the client (UI Extensions) too, but it didn't work.

I don't think you need to update the client libraries because the endpoint set up comes from the host code inside the worker. It should be good enough to just update POS and the `pos-next-react-native` library.

PR #1389, opened 2023-02-15
Create CLI's ESLint plugin package

isaacroldan:
We have a lot of specific and custom ESLint rules, and since more teams are collaborating with the CLI but in other repositories, makes sense to have a shared set of rules that all cli-kit consumers should use.

Fixes https://github.com/Shopify/internal-cli-foundations/issues/545

- Extract all our rules (custom and not) into a new npm package called `@shopify/eslint-plugin-cli`
- Make our main `.eslintrc` import that extend those rules (with some custom specific overrides)
 

### Measuring impact

### Checklist


packages/eslint-plugin-cli/package.json
```diff
@@ -0,0 +1,29 @@
+{
+  "name": "@shopify/eslint-plugin-cli",
```
gonzaloriestra, line 2: I'd remove the `plugin` word from the package because it has nothing to do with plugins actually. As you say in the description, this contains `Shopify CLI's ESLint rules and configs`. So what about calling it `@shopify/eslint-cli` or `@shopify/cli-eslint-rules`?

isaacroldan: The reason I used that name is because of the already existing Shopify eslint package: https://www.npmjs.com/package/@shopify/eslint-plugin

So I just appended `cli` to that. But I think I like `@shopify/cli-eslint-rules` more 🤔 
Can we name it as we like? Should we follow the convention? (most packages are called either `eslint-plugin...` or `esling-config...` ([here](https://www.npmjs.com/search?q=eslint-plugin))

gonzaloriestra: Oh, I see. That `plugin` word may be confusing in the context of the CLI, as we have other kind of plugins... What about `cli-eslint-config`? There seem to be a lot of [packages with a similar name](https://www.npmjs.com/search?q=eslint-config) as well

isaacroldan:
@gonzaloriestra @pepicrft The eslint plugin is ready to review!

@pepicrft about the naming of the package, check [this comment](https://github.com/Shopify/cli/pull/1389#discussion_r1121635888)

gonzaloriestra (approved):
👌

pepicrft (approved):

PR #1391, opened 2023-02-16
Js.ef prototype

DuncanUszkay1:
Drafting PR to see a clear diff

### DIFF
--- a/packages/app/src/cli/services/deploy/upload.ts
+++ b/packages/app/src/cli/services/deploy/upload.ts
@@ -32,6 +32,7 @@ import {functionProxyRequest, partnersRequest} from '@shopify/cli-kit/node/api/p
 import {randomUUID} from '@shopify/cli-kit/node/crypto'
 import {ensureAuthenticatedPartners} from '@shopify/cli-kit/node/session'
 import {fileExists, readFile, readSync} from '@shopify/cli-kit/node/fs'
+import {createExtension} from '../dev/create-extension.js'
 
 interface DeployThemeExtensionOptions {
   /** The application API key */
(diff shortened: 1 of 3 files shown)


packages/app/src/cli/services/deploy/upload.ts
```diff
@@ -169,6 +170,83 @@ interface UploadFunctionExtensionsOptions {
...
+
+  const variables: CreateDeploymentVariables = {
+    apiKey: options.identifiers.app,
+    uuid: deploymentUUID,
+    bundleUrl: Object.values(signedURLs)[0] || "", // we don't use this, but it must be a valid URL for deploy to work.
+    extensions: formattedExtensions,
+  }
+
+  const mutation = CreateDeployment
+  const result: CreateDeploymentSchema = await partnersRequest(mutation, options.token, variables)
```
isaacroldan, line 234: We are already doing a `CreateDeployment` query for UI extensions and Theme extensions, we should try to just add these new `fromattedExtensions` there. It's on line 118 of this same file. And the extensions object is created in `deploy.ts` line 58: you can see we create an array of extensions (we should append there the `formattedExtensions) and then we pass it to upload.

Let me know if you need help with this, we can pair!

DuncanUszkay1: This is just here so I can see the diff, there's no intent to merge.

That said, we will happily take you up on that pairing offer once we get into build 😄

isaacroldan: sure, just ping me on slack whenever you are ready to move forward with this :)

DuncanUszkay1:
It's still relevant, it's a long running prototype

PR #1401, opened 2023-02-17
Remove keytar and store the credentials serialized in the file system

pepicrft:
Fixes https://github.com/Shopify/internal-cli-foundations/issues/271

First-party developers started reporting the following error when the CLI tries to persist the credentials using `libsecret` in Linux:

```
Unable to update the secure store: Object does not exist at path "/org/freedesktop/secrets/collection/login" 
```

The investigation of the issue led me to realize that GitHub is no longer maintaining [keytar](https://www.npmjs.com/package/keytar), which we use it as a cross-platform solution to use the platform-specific tool for persisting credentials. Some developers also brought up the annoyance of Keychain prompting the user very frequently to give the CLI access to the credentials.

I'm removing keytar and adjusting the logic to serialize and store the credentials in the file system. This is the rationale that led to that decision:
- Serializing and storing credentials in the file system is **standard in the industry**. GCS users the [Config](https://github.com/boto/boto/blob/develop/boto/pyami/config.py) utility from AWS' Boto Python library to store credentials in the file system. AWS and Wrangler follow a similar approach.
- With Keytar Linux users might need to install additional components depending on their contribution. That adds friction to the first-time experience.
- Frequent Keychain prompts is a bad DX
- Keytar is no longer maintained which means sooner or later we'd have to either fork the project or create a custom solution ourselves.

**Note** that I decided not to include a migration in this work. That was a conscious decisions for two reasons:
- The migration entails adding global state (to flag that a migration has been executed) and code that will eventually get deleted.
- And I believe it's not worth it if we think about the fact that users will have to log in again, which is not a big deal.

### How to test your changes?
You should be able to run any command that requires authentication.

### Measuring impact

### Checklist


Arkham (approved):
Looks good to me, I'm looking forward to not having to type in my keychain password every time.

Arkham:
do we still need `secure-store.ts` or can we remove it?

gonzaloriestra (approved):
👌

PR #1404, opened 2023-02-17
Embedded cli-ruby opt-in by default

alvaro-shopify:
In the first try to use the `embedded cli-ruby` some problems arise so we had to [revert the activation ](https://github.com/Shopify/cli/pull/1346) and opt-it by default only for `shopifolk` users.

There were three main problems:
* Running the Ruby `shopify-cli` bin script using the `env` Ruby version. Solution [PR](https://github.com/Shopify/cli/pull/1304)
* Running `bundler` with a fixed version because of the `Gemfile.lock` presence. Solution PR](https://github.com/Shopify/cli/pull/1347)
* Missed the installation of specific gems in Windows for `theme-app-extension`. Solution [PR](https://github.com/Shopify/cli/pull/1350)

- Modify the check condition that evaluates when the `embedded cli-ruby` must be used. Now is opted-in by default and the environment variable `SHOPIFY_CLI_BUNDLED_THEME_CLI`  should be set to use the `bundled` version

### How to test your changes?
- Run any theme command using `--verbose` flag and check the exec message of the CLI2. It should be similar to this
```
Running system process:
  · Command: /Users/alvarogutierrez/src/github.com/Shopify/cli/packages/cli-kit/assets/cli-sources/bin/shopify theme check /Users/alvarogutierrez/cli-themes/first-theme --output text
  · Working directory: /Users/alvarogutierrez/cli-themes/first-theme
```

### Measuring impact

### Checklist

### DIFF
--- a/packages/cli-kit/src/public/node/environment.ts
+++ b/packages/cli-kit/src/public/node/environment.ts
@@ -1,0 +1,13 @@
+/**
+ * It returns the environment variables of the environment
+ * where the Node process is running.
+ *
+ * This function exists to prevent the access of the process
+ * global variable which is discouraged via the no-process-env
+ * ESLint rule.
+ *
+ * @returns Current process environment variables.
+ */
+export function getEnvironmentVariables(): NodeJS.ProcessEnv {
+  return {...process.env}
+}
--- a/packages/cli-kit/src/public/node/environment.ts
+++ b/packages/cli-kit/src/public/node/environment.ts
@@ -9,5 +9,5 @@
  * @returns Current process environment variables.
  */
 export function getEnvironmentVariables(): NodeJS.ProcessEnv {
-  return {...process.env}
+  return process.env
 }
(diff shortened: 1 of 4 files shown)


packages/cli-kit/src/public/node/environment.ts
```diff
@@ -0,0 +1,13 @@
...
+ *
+ * This function exists to prevent the access of the process
+ * global variable which is discouraged via the no-process-env
+ * ESLint rule.
+ *
+ * @returns Current process environment variables.
+ */
+export function getEnvironmentVariables(): NodeJS.ProcessEnv {
+  return {...process.env}
+}
```
amcaplan, line 13: I very strongly approve of this strategy. Excellent for testing, and ensures we don't accidentally write to `process.env`.

amcaplan (approved):

packages/cli-kit/src/public/node/environment.ts
isaacroldan: why not just `return process.env`?

PR #1409, opened 2023-02-20
Trim public cli-kit interface related to themes

amcaplan:

We shouldn't expose more than absolutely necessary in cli-kit. While this work isn't complete (I think cli-kit already knows too much about themes), we should at least not expose functions that aren't used anywhere.

Moves everything to private files that aren't exported from cli-kit, unless it's actually used in another package.

### How to test your changes?

Do theme stuff

### Measuring impact

### Checklist


isaacroldan:
Oh, I didn't know about this folder... Maybe it makes sense to have a `ThemeManager` in cli-kit so it can eventually be used to build other tools... but we should revisit this folder to see what belongs here and what should be in `theme`

isaacroldan (approved):

PR #1416, opened 2023-02-21
[Prototype] Render upgrade warning after command result if any GraphQL responses contain deprecations

dphm:
🛑 Do not merge. This PR is for exploration.

#gsd:33165
Issue: https://github.com/Shopify/script-service/issues/6535

[Brief proposal](https://gist.github.com/dphm/23332612d98a9cc618d07261724664f5#file-proposal-md)

https://user-images.githubusercontent.com/1707217/225382855-b72ec8bc-65c2-4ecd-b671-cda1ee4a9ac3.mov

### How to test your changes?

`partners` branch: [`prototype/deprecate-script-service-proxy`](https://github.com/Shopify/partners/pull/45978)

```sh
export SHOPIFY_SERVICE_ENV=spin; export SPIN_INSTANCE=deprecate
```

#### Without Function

```sh
rm -rf extensions && pnpm shopify app deploy --path ~/path/to/app
```

#### With Function

```sh
pnpm shopify app generate extension --path ~/path/to/app --type=payment_customization --template=rust --name demo
```

```sh
pnpm shopify app deploy --path ~/path/to/app
```


packages/cli-kit/src/private/node/api/graphql.ts
```diff
@@ -65,3 +79,11 @@ function errorHandler<T>(api: string): (error: unknown) => Error | unknown {
     }
   }
 }
+
+export function handleDeprecations<T>(response: GraphQLResponse<T>) {
+  const deprecations = (response.extensions as WithDeprecations)?.deprecations
```
pepicrft, line 84: `response.extensions` breaks the service-agnostic nature of the APIs in this module. What about the following:
- If the schema is specific to the Partners' API, we can have a version of this function, `partnersGraphQLRequest` that lives in the `@shopify/app` package (which is the only one interacting with partners).
- If the schema is specific to one of the queries and mutations, I'd move the code to where that request is happening.

dphm: We intended for the schema to be generalized beyond Partners API so other APIs could also return deprecations if necessary with the same mechanism. Maybe I can pass an optional `handleDeprecations` function to `graphqlRequest` for the caller to decide?

amcaplan:
🤔  I wonder if the strategy is overkill. Why not just directly output a deprecation warning whenever we hit it? I kind of feel like it's not our problem to minimize the number of times the message gets printed; more annoyance is a good thing. And that just happens to come with the advantage of not maintaining global state, figuring out which warning to display, etc.

By the way, will the message even appear if the user kills `dev`?

dphm:
BTW the implementation merged already: https://github.com/Shopify/cli/pull/1597
(I can make any necessary changes this week in a separate PR)

@stephszeto I'll close this Prototype PR and respond in the other PR, sorry for the confusion

PR #1418, opened 2023-02-22
Fix loading JS functions & Improve theme dev loading time

isaacroldan:
Bugs found during the release QA

- JS functions weren't loading correctly if there were other UI extensions present, fixed that.
- Theme dev was adding an extra waiting time that wasn't necessary on the first run.

### Measuring impact

### Checklist

### DIFF
--- a/packages/theme/src/cli/commands/theme/dev.ts
+++ b/packages/theme/src/cli/commands/theme/dev.ts
@@ -121,11 +121,11 @@ export default class Dev extends ThemeCommand {
       controller.abort()
       controller = new AbortController()
       // eslint-disable-next-line @typescript-eslint/no-floating-promises
-      this.execute(adminSession, flags.password, command, controller)
+      this.execute(adminSession, flags.password, command, controller, true)
     }, this.ThemeRefreshTimeoutInMs)
 
     // eslint-disable-next-line @typescript-eslint/no-floating-promises
-    this.execute(adminSession, flags.password, command, controller)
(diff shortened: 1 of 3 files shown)


packages/theme/src/cli/commands/theme/dev.ts
```diff
@@ -121,20 +121,21 @@ export default class Dev extends ThemeCommand {
...
 
   async execute(
     adminSession: AdminSession,
     password: string | undefined,
     command: string[],
     controller: AbortController,
+    shouldWait: boolean,
   ) {
-    await sleep(3)
+    if (shouldWait) await sleep(3)
```
Arkham, line 138: PS C:\Users\arkham\Desktop\shopify-apps\feb16-1745\extensions\prod-discount-ts> pnpm build

> prod-discount-ts@0.0.1 build C:\Users\arkham\Desktop\shopify-apps\feb16-1745\extensions\prod-discount-ts
> npm exec -- shopify app function build


╭─ success ────────────────────────────────────────────────────────────────────╮
│  Function built successfully.                                                │

Arkham: oh I see, it's not the end of the world then. In theory we could use functions from https://github.com/unjs/get-port-please (which we already have as a dep) to check if the port is available for use :)

isaacroldan: Yeah, it was a temporary solution, the idea is that `dev` should handle token refresh internally, but that won't happen until the command is ported from Ruby...

PR #1427, opened 2023-02-22
Remove unused shopify-cli-extensions package

Arkham:

What the title says


Arkham:
@gonzaloriestra we don't need this inside CLI3, correct?

gonzaloriestra (approved):

PR #1447, opened 2023-02-24
Disable function type validation except on deploy

isaacroldan:
When we load the local app into the CLI, we load extensions validate that they follow a specific schema and that the types are correct. For this we also use the remote ui specifications that tell us if you have access to a specific type or not.

For functions this is a bit different, all the specs are local, we validate that the type exists locally. But this same validation is then applied server-side when you try to deploy. So maybe it doesn't make sense to enforce it for all CLI operations pre-deploy.

This is what happens if you try to deploy a non-existing function type:
Captura de pantalla 2023-02-24 a las 12 30 44

- When loading functions. Accept ALL function types, do not validate if there is a spec for them or not.
- If a user modifies the type of a function to an invalid one, `app deploy` will throw an error.

### How to test your changes?
- modify the type inside the `toml` file of one function
- Try to `build, info, dev` the app, it should work
- Try to `deploy`, it should fail

### Measuring impact

### Checklist

### DIFF
--- a/packages/app/src/cli/models/app/app.test-data.ts
+++ b/packages/app/src/cli/models/app/app.test-data.ts
@@ -1,7 +1,7 @@
 import {App, AppInterface} from './app.js'
 import {FunctionExtension, ThemeExtension, UIExtension} from './extensions.js'
 import {UIExtensionInstance, UIExtensionSpec} from '../extensions/ui.js'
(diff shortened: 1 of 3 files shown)


gonzaloriestra (approved):
Makes sense! 👌

PR #1448, opened 2023-02-24
support new extension point names format

jonathanhamel4:

Adds support for both the `Admin::XX::XX` syntax and the `admin.x.x` syntax when extracting the surface of an extension. This will allow to fix the issue where `No extensions found for the surface: admin` is returned in the embedded admin console.

Just  changing the regex to support both `::` or `.` as delimiter. (Can current extension points use a `.` that would influence this?)

### Measuring impact

### Checklist

### DIFF
--- a/.changeset/neat-avocados-peel.md
+++ b/.changeset/neat-avocados-peel.md
@@ -1,0 +1,5 @@
+---
+'@shopify/app': patch
+---
+
+Adding support for the new naming convention of extension points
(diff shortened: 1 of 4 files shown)


.changeset/neat-avocados-peel.md
```diff
@@ -0,0 +1,5 @@
+---
+'@shopify/app': patch
```
vividviolet, line 2: @isaacroldan any concerns with making this a patch?

vividviolet (commented):
Thanks for doing this! Just one suggestion - can you add  an extension point with the new naming format to the payload [test](https://github.com/Shopify/cli/blob/6ab492a8f8576befa3a14b113383ce90b32bca2b/packages/app/src/cli/services/dev/extension/payload.test.ts#L136) ? This one is important for local development

.changeset/neat-avocados-peel.md
isaacroldan: nop, sounds good 👌

vividviolet (approved):

PR #1472, opened 2023-02-28
Use ESM version of Lodash

gonzaloriestra:

We were using the [CommonJS version of Lodash](https://www.npmjs.com/package/lodash), but there's another official [package exported as ESM](https://www.npmjs.com/package/lodash-es).

The ESM version is much lighter (636kB vs 1.41MB) and it will load faster.

- Replaces `lodash` with `lodash-es` in cli-kit
- Deduplicates dependencies

### How to test your changes?

CI

### Measuring impact

### Checklist

### DIFF
--- a/packages/cli-kit/package.json
+++ b/packages/cli-kit/package.json
@@ -131,7 +131,7 @@
     "kill-port-process": "3.1.0",
     "latest-version": "7.0.0",
     "liquidjs": "10.5.0",
-    "lodash": "4.17.21",
+    "lodash-es": "4.17.21",
     "macaddress": "0.5.3",
     "node-abort-controller": "3.0.1",
     "node-fetch": "3.2.4",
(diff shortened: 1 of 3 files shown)


Arkham:
I think we tried this in the past and it wasn't working. @pepicrft do you remember anything about this?

pepicrft (commented):
The usage of CJS for this particular use case was on purpose because due to the static nature of ESM and how `lodash-es` is doing exports the following:

```js
import { memoize} from 'lodash'
```

Ends up importing the whole lodash, which is something we don't want. I'd open an issue in the `lodash-es` repository and recommend exporting different modules like we are doing with `@shopify/cli-kit`:

```js
import memoize from 'lodash-es/memoize'
```

gonzaloriestra:
Oh, I see. Thanks for the explanation! Isn't the tree-shaking helping in that case?

The last update in the library was 2 years ago, I'm not sure they will help us on this, but we can try...

PR #1479, opened 2023-03-01
Fix layout width for very narrow terminals

matteodepalo:

Fixes https://github.com/Shopify/internal-cli-foundations/issues/559

Buildkit has a very narrow stdout and was producing very weird squeezed logs.

For terminals narrower than 20 characters we want to force a minimum width of 20.

### How to test your changes?

Resize your terminal to less than 20 characters and run kitchen-sink. Boxes should look a bit broken unless you resize the terminal up.

### DIFF
--- a/.changeset/great-guests-yell.md
+++ b/.changeset/great-guests-yell.md
@@ -1,0 +1,5 @@
+---
+'@shopify/cli-kit': patch
+---
+
+Fix layout width for very narrow terminals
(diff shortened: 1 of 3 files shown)


matteodepalo:
We would also need to maintain plain text versions for all our components which might become pretty tedious considering this issue happens only on buildkit and not on every CI environment (Github actions displays things just fine)

PR #1483, opened 2023-03-01
Set cloudflare as default & add tunnel flags

isaacroldan:
Contributes to https://github.com/Shopify/internal-cli-foundations/issues/307

- Set cloudflare as default tunnel solution
- Add a new (hidden) flag to select between ngrok and cloudflare
- Select the tunnel provider based on a remote feature flag

### How to test your changes?
- `app dev` should automatically create a cloudflare tunnel
- `app dev --tunnel ngrok` should create an ngrok tunnel

### Post-release steps
Update documentation saying:
- We use cloudlflare by default, there is no need for auth or anything, it'd just work
- ngrok is optional. Use it with `--tunnel ngrok`

### Measuring impact

### Checklist


packages/app/src/cli/commands/app/dev.ts
```diff
@@ -92,6 +94,7 @@ export default class Dev extends Command {
     await addPublicMetadata(() => ({
       cmd_app_dependency_installation_skipped: flags['skip-dependencies-installation'],
       cmd_app_reset_used: flags.reset,
+      cmd_dev_tunnel_type: flags['tunnel-url'] ? 'custom' : flags.tunnel,
```
isaacroldan, line 97: the tunnel type was only being reported after a successful dev execution, but I think is worth it to report it BEFORE, just in case the dev crashes. That way we can know if it crashes for a specific tunnel type

Arkham (approved):
Very nice, left a couple of questions

packages/app/src/cli/services/dev.ts
```diff
@@ -37,6 +37,8 @@ import {
 import {OutputProcess, outputInfo} from '@shopify/cli-kit/node/output'
 import {Writable} from 'stream'
 
+export type TunnelProviders = 'ngrok' | 'cloudflare' | 'none'
```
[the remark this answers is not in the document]
isaacroldan: Yeah, that was the intention of using this, but we didn't really want to expose the option of no-tunnel for now, so I removed this type. (--no-tunnel is hidden)

packages/app/src/cli/services/context.ts
```diff
@@ -47,15 +47,15 @@ interface DevContextOutput {
...
+  useCloudflareTunnels: boolean
 }
 
 /**
  * Make sure there is a valid context to execute `generate extension`
  *
  * We just need a valid app API key to access the Specifications API.
  * - If the API key is provided via flag, we use it.
- * - Else, if there is cached API key for the current directory, we use it.
+ * - Else, if there is causeCloudflareTunnelsched API key for the current directory, we use it.
```
shauns, line 58: think this is a typo

isaacroldan: ups

packages/app/src/cli/services/dev/urls.ts
```diff
@@ -86,23 +79,21 @@ export async function generateFrontendURL(options: FrontendURLOptions): Promise<
...
-    frontendPort = await getAvailableTCPPort()
-    frontendUrl = await generateURL(options.commandConfig, frontendPort)
-  } else {
+  if (options.noTunnel) {
     frontendPort = await getAvailableTCPPort()
     frontendUrl = 'http://localhost'
     usingLocalhost = true
+  } else {
+    frontendPort = await getAvailableTCPPort()
+    const provider = options.tunnelProvider || options.useCloudflareTunnels ? 'cloudflare' : 'ngrok'
```
gonzaloriestra, line 88: The second expression is missing parentheses, so the logic is not working as expected

```suggestion
    const defaultProvider = options.useCloudflareTunnels ? 'cloudflare' : 'ngrok'
    const provider = options.tunnelProvider || defaultProvider
```

packages/app/src/cli/models/organization.ts
```diff
@@ -4,6 +4,7 @@ export interface Organization {
   website?: string
   betas: {
     appUiDeployments: boolean
+    cliTunnelAlternative: boolean
```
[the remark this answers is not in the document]
isaacroldan: That makes sense, i'll update for `cliTunnelAlternative` and see if i can update it easily for `appUiDeployments`

packages/app/src/cli/api/graphql/all_orgs.ts
```diff
@@ -5,7 +5,8 @@ export interface AllOrganizationsQuerySchemaOrganization {
   businessName: string
   website: string
   betas: {
-    appUiDeployments: boolean
+    appUiDeployments?: boolean
+    cliTunnelAlternative?: boolean
```
gonzaloriestra, line 9: Do we need to retrieve the betas in all these graphql requests? It would be nice to remove it if it's not used in all the cases to avoid making them slower 🙏

gonzaloriestra (approved):
Thanks for the changes! I've tested it and it's now working fine with all the combinations of tunnel and beta flags 👌

packages/app/src/cli/api/graphql/all_orgs.ts
isaacroldan: Done @gonzaloriestra, thanks for the feedback, much cleaner now 🙇

PR #1533, opened 2023-03-02
Footer dev message not shown when exist theme app extension

alvaro-shopify:
When you run the `dev` command using a theme app extension the footer in not working properly and the output of the the theme app extension (CLI2) is messed up with the output of the other processes (backend, frontend, extensions)

There is an[ open issue ](https://github.com/Shopify/internal-cli-foundations/issues/563) to solve this problem but in the meanwhile we want to avoid the problem with the footer when the app includes a theme app extension.

- In case the app includes a theme app extension the cli will render a plain app url instead the footer

### How to test your changes?
- Create an app with a theme app extension 

### Measuring impact

### Checklist


.changeset/clean-donuts-brush.md
```diff
@@ -0,0 +1,5 @@
+---
+'@shopify/app': patch
+---
+
+Footer dev message not shown when exist theme app extension. Plain app urls is shown instead"
```
gonzaloriestra, line 5:
```suggestion
Fix URLs output for dev with theme app extensions
```

gonzaloriestra (approved):

pepicrft (changes requested):
Have we considered fixing the issues below instead of disabling the footer altogether?

- The Ruby CLI includes a frame that makes the layout break. I'd look for where in the code [this API](https://github.com/Shopify/cli/blob/067199cf6b1ea38613cc5ab1b46c25b9ffc37365/packages/cli-kit/assets/cli-ruby/lib/shopify_cli/context.rb#L96) is being used and adjust the code not to use a frame at all.
- Then, we can adjust the Node CLI to let the concurrent API format the output along with the other processes.

alvaro-shopify:
> open issue

@pepicrft  yes, there is an [open issue](https://github.com/Shopify/internal-cli-foundations/issues/563) to solve that problem, do you think it's better to solve it now instead of disabling the footer?. Anyway, I have found out that the footer is not working in the spin terminal, so we should need a way to disabling it depending on some conditions

gonzaloriestra:
I think this issue is quite urgent: theme app extensions can't be properly tested right now (the URLs can be hidden). So unless we can fix the footer quickly, I'd release this temporary solution.

amcaplan (approved):
👍

Agreed this isn't an ideal situation, but it's important to know when to do it right and when to do something that just works well enough for now. And I think that spin support is significant enough to warrant a generalized solution.

I don't think we should invest more in Ruby code, instead it will be a subpar experience but at least it won't actively block development.

Unfortunately this does mean that `Press q` won't work when theme extensions exist, which is quite unfortunate, given that Windows users experience challenges when using `ctrl+c` to exit the process. So it's worth at least marking that as something to figure out in the near term.

alvaro-shopify:
Rejected in favour of #1570 1570

PR #1541, opened 2023-03-03
Use templates specifications endpoint

gonzaloriestra:

Fixes https://github.com/Shopify/internal-cli-foundations/issues/582

Co-authored by @alvaro-shopify 

* Integrate the [new](https://github.com/Shopify/partners/pull/46021) `templateSpecifications` partners API
* Support using extensions and template specifications in the same generate command
* Right now, remote specifications only include `functions` definitions. 
* In case of using a template, there is support for including several extensions (only functions right now)

### How to test your changes?
  * `p shopify app generate extension --path YOU_APP_PATH`. UI extensions are generated from local specifications and `functions` from the remote one

### Measuring impact

### Checklist


packages/app/src/cli/constants.ts
```diff
@@ -36,14 +36,14 @@ export const blocks = {
   },
 } as const
 
-export const defaultFunctionsFlavors: {name: string; value: ExtensionFlavor}[] = [
+export const defaultFunctionsFlavors: ExtensionFlavor[] = [
   {name: 'JavaScript (developer preview)', value: 'vanilla-js'},
   {name: 'TypeScript (developer preview)', value: 'typescript'},
   {name: 'Rust', value: 'rust'},
   {name: 'Wasm', value: 'wasm'},
 ]
```
isaacroldan, line 44: The Partners PR has been merged, we can remove all references to local function specifications right?

packages/app/src/cli/services/generate/fetch-template-specifications.ts
```diff
@@ -0,0 +1,76 @@
...
+// There should be another api with the extensions specifications. Right now templates only support functions
+function resolveExtensionCustoms(_type: string): {
+  category: ExtensionCategory
+  configSchema?: ZodSchemaType<FunctionConfigType>
+} {
+  return {
+    category: 'function',
+    configSchema: BaseFunctionConfigurationSchema,
+  }
+}
```
isaacroldan, line 67: why do you need this function? 🤔

packages/app/src/cli/services/generate.ts
```diff
@@ -150,4 +144,55 @@ function formatSuccessfulRunMessage(
...
+  const specifications = await fetchSpecifications({token, apiKey, config: options.config})
+  let templateSpecifications = convertSpecificationsToTemplate(specifications)
+  const remoteTemplateSpecifications = await fetchTemplateSpecifications(token)
+  templateSpecifications = templateSpecifications.concat(remoteTemplateSpecifications)
+  // If for any reason the remote template specifications API with the functions is not available, we will load the
+  // local functions specifications
+  if (remoteTemplateSpecifications.length === 0) {
+    const functions = await loadFunctionSpecifications(options.config)
+    templateSpecifications = templateSpecifications.concat(convertSpecificationsToTemplate(functions))
+  }
```
isaacroldan, line 194: We shouldn't need to do this, the API will be available and we can't keep duplicated hardcoded info in the CLI, the idea of this project was to remove that from the CLI :)

packages/app/src/cli/services/function/common.ts
```diff
@@ -40,11 +40,12 @@ export async function inFunctionContext(
 
 export async function ensureFunctionExtensionFlavorExists(
   specification: FunctionSpec,
-  extensionFlavor: ExtensionFlavor,
+  extensionFlavor: ExtensionFlavorValue,
   templateFlavor: TemplateFlavor,
   templateDownloadDir: string,
 ): Promise<string> {
-  const templatePath = specification.templatePath(templateFlavor)
+  // Extensions resolves the template path from template flavor, but templates use extension flavor
```
isaacroldan, line 47: wat. Can we simplify this to use just one flavor? feels hard to understand what's happening here 😅

packages/app/src/cli/services/generate/fetch-template-specifications.ts
alvaro-shopify: This are the missing things either in the template spec or the extensions. 

We need to cover these requirements. Now it's hardcoded to the functions but we should add[ this information ](https://github.com/Shopify/cli/pull/1541#discussion_r1140148549) to the current extensions spec

packages/app/src/cli/services/generate.ts
alvaro-shopify: https://github.com/Shopify/cli/pull/1541#discussion_r1140142058

packages/app/src/cli/services/generate/fetch-template-specifications.ts
isaacroldan: You are not using the configSchema anywhere. But in any case and related to a previous comment, the schema doesn't belong to the template, schema is a "build" thing and the template is a "generate" thing.

packages/app/src/cli/services/function/common.ts
alvaro-shopify: the problem is that the resolution of the `templatePath` is different between how current local specifications behave and the one we load from the template specificacions:
- Local specs: Replace the token in a template url with the flavor
- Template specs: Each flavor includes it's own template url
Once we use only the remote templates this should be changed to:

```ts
const templatePath = specification.templatePath(extensionFlavor)
```

isaacroldan: Ah well, since we are removing the local specs, we can simplify this then, cool 👌

packages/app/src/cli/services/generate/fetch-template-specifications.ts
isaacroldan: I mean the one in this function, the schema is used yes :P 
The schema can still be defined locally for now, that part is not relevant for the templates

packages/app/src/cli/models/app/template.ts
```diff
@@ -0,0 +1,30 @@
+import {GenericSpecification} from './extensions.js'
+import {extensionTypesGroups} from '../../constants.js'
+
+export interface TemplateSpecification {
+  identifier: string
+  name: string
+  group: string
+  supportLinks: string[]
+  types: GenericSpecification[]
+}
```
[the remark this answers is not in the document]
isaacroldan: We can pair on monday if you want for this :)

packages/app/src/cli/services/generate/fetch-template-specifications.ts
gonzaloriestra: I've removed the schema from there.

alvaro-shopify (approved):

PR #1545, opened 2023-03-03
Update NX

isaacroldan:
Dependabot opened a bunch of PRs, but we needed some manual work

- Upgrade all nx dependencies to version 15.8.3


isaacroldan:
The changes in the `project.json` files were automatically applied by nx with `nx g @nrwl/workspace:fix-configuration --all`
But it only changed a couple of them, not all packages, so not sure if we should manually update the rest?

Arkham (approved):
Let's merge this :)

PR #1557, opened 2023-03-06
Unhide JavaScript functions flavors and commands

nickwesselman:

JavaScript functions are releasing to dev preview, but will still require use of the `@pre` release tag. We don't want the further step of enabling a feature flag to be required.

* Displays JS and TS flavors by default, with a `(developer preview)` description
* Unhides `shopify app function` topic

### How to test your changes?

Against a local build:
* `yarn shopify app generate extension --path <my app>`

image

* `yarn shopify app --help --path <my app>`

image


nickwesselman:
@Arkham No I did not know the process for that, should we engage a tech writer? Can we merge this before that's done?

Arkham:
the last couple of times I just did it myself and tagged @shainaraskas on the PR to shopify.dev Since we now release on a longer time window, it's important that it's added to the doc page by then

PR #1565, opened 2023-03-07
Add log information when API requests return an error

alvaro-shopify:
- When an API request returns an error the CLI is not generating debug log messages with neither the time execution nor the headers. This last point is quite important in case you want to trace the error request using this [tutorial](https://videobin.shopify.io/v/Nxz5l2) as we are not able to get the value of `x-request-id`

- The handling of the errors is done inside the method that executes the request, so the debug logging is run in one place

### How to test your changes?
Run the `dev` command using a wrong `api-key` with the `--verbose` flag
`yarn shopify app dev --api-key 1234 --verbose``

you should see that the correct responses and the error response should show the execution time and the desired headers

### Measuring impact

### Checklist

### DIFF
--- a/packages/cli-kit/src/private/node/api/graphql.ts
+++ b/packages/cli-kit/src/private/node/api/graphql.ts
@@ -9,7 +9,7 @@ export interface GraphQLVariables {
   [key: string]: any
 }
 
-export function graphqlRequest<T>(
+export async function graphqlRequest<T>(
   query: RequestDocument,
   api: string,
   url: string,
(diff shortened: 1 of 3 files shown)


packages/cli-kit/src/private/node/api/graphql.ts
```diff
@@ -17,20 +17,17 @@ export function graphqlRequest<T>(
...
+    )
     return response.data
-  }
-
-  if (handleErrors) {
-    return handlingErrors(api, action)
-  } else {
-    return action()
-  }
+  })()
```
Arkham, line 30: wait, why do we need the self-invoking function here?

Arkham: we have the change the function to be async, but it's returning a promise already, so I don't see why not :)

Arkham (approved):
LGTM :)

Issue #1544, opened 2023-03-03
[Bug]: Couldn't Find App With ApiKey - when application is deleted


1000p:
### Please confirm that you have:

- [X] Searched [existing issues](.) to see if your issue is a duplicate. (If you’ve found a duplicate issue, feel free to add additional information in a comment on it.)
- [X] Reproduced the issue in [the latest CLI version](https://www.npmjs.com/package/@shopify/cli).

### In which of these areas are you experiencing a problem?

Other

### Expected behavior

Initializing an application with the same application name should refresh persisted local data.

### Actual behavior

Persisted local data is not refreshed upon new initialization. An error appears when starting the application:
`Couldn't Find App With ApiKey`

### Stack trace

```
│  Couldn't find App with apiKey 1b0dea9cf4ff337a4d128c0ff968b34d.             │
│                                                                              │
│  To investigate the issue, examine this stack trace:                         │
│    at appFromId (@shopify/app/src/cli/services/context.ts:200)               │
│      if (!app) throw new BugError(`Couldn't find App with apiKey             │
│      ${appId}.`)                                                             │
│    at processTicksAndRejections (node:internal/process/task_queues:95)       │
│    at all                                                                    │
│    at async ensureDevContext (@shopify/app/src/cli/services/context.ts:152)  │
│      const [organization, _selectedApp, _selectedStore] = await              │
│      Promise.all([                                                           │
│    at async dev (@shopify/app/src/cli/services/dev.ts:74)                    │
│      } = await ensureDevContext(options, token)                              │
│    at run (@shopify/app/src/cli/commands/app/dev.ts:100)                     │
│      await dev({
```

### Reproduction steps

1. `npm init @shopify/app@latest`
2. fill out all the required initialization data, project name, template, etc.
3.  `npm run dev`
4. You will be asked to enter application name - create an application with the name "choose-your-name". Take note of the application name, it will be important.
5. Install the application on your dev. store - using P from the command line after application is started
6. Go to your Shopify Partner account and delete that application that you have just created.
7. Delete local project files

Next:

1. `npm init @shopify/app@latest`
2. fill out all the required initialization data, project name, template, etc. 
3.`npm run dev`
4. You will be asked to enter the application name - you should choose the application name of the app that you just deleted
5. The for store - choose the same one as before 
6. Error is produced: `Couldn't Find App With ApiKey`

### Operating System

Windows 10

### Shopify CLI version (check your project's `package.json` if you're not sure)

3.44.1

### Shell

[...]

PR #1572, opened 2023-03-07
Tell users to try with '--reset' when we can't find apps, orgs, or stores

Arkham:

We store information about apps using [conf](https://github.com/sindresorhus/conf). But if the user deletes their app from the local filesystem and from the partners portal, they can get in weird situations where the api key that we cached is no longer valid. 

Tell users that they can retry their command with `--reset` which should solve their problem.

### How to test your changes?

- Create a new app and run `pnpm dev`
- Press p to preview it
- Now go to your partners portal and delete the app
- Run `pnpm dev` and you should see a message that tells you to rerun the command with `--reset`

### Measuring impact

### Checklist

### DIFF
--- a/packages/app/src/cli/services/context.ts
+++ b/packages/app/src/cli/services/context.ts
@@ -188,16 +188,18 @@ export async function ensureDevContext(options: DevContextOptions, token: string
   return result
 }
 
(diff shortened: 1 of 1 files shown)


amcaplan:
Fixes https://github.com/Shopify/cli/issues/1544

PR #1581, opened 2023-03-08
Fix log logic in api responses for node version that do not support Header type

alvaro-shopify:
…eaders type

After including this [PR](https://github.com/Shopify/cli/pull/1565) in the last `pre` version, users using node v16 reported this error:

```
│  Headers is not defined                                                      │
│  To investigate the issue, examine this stack trace:                         │
│    at debugLogResponseInfo (@shopify/cli-kit/src/private/node/api.ts:22)     │
│      let headers: Headers = new Headers()                                    │
│    at shopifyFetch (@shopify/cli-kit/src/public/node/http.ts:73)             │
│      return debugLogResponseInfo({url: url.toString(), request:              │
```

The problem is that the `Headers` type does not exist in previous node versions. 
The debug log logic in api responses should work in all supported node version

- The logic that gets the value of either the success or failure responses should be generic using primitive types.

### How to test your changes?
Use any node version switcher to change between v16 and higher versions and run the dev command with each version using a wrong api-key with the --verbose flag
`yarn shopify app dev --api-key 1234 --verbose`

you should see that the correct responses and the error response should show the execution time and the desired headers

### Measuring impact

### Checklist

### DIFF
--- a/packages/cli-kit/src/private/node/api.ts
+++ b/packages/cli-kit/src/private/node/api.ts
@@ -19,14 +19,20 @@ export async function debugLogResponseInfo<T extends {headers: Headers; status: 
   errorHandler?: (error: unknown) => Error | unknown,
 ): Promise<T> {
   const t0 = performance.now()
-  let headers: Headers = new Headers()
+  const responseHeaders: {[key: string]: string} = {}
   let response: T = {} as T
   try {
     response = await request
-    headers = response.headers
+    response.headers.forEach((value, key) => {
+      if (interestingResponseHeaders.has(key)) responseHeaders[key] = value
+    })
   } catch (err) {
     if (err instanceof ClientError) {
-      headers = err.response?.headers as Headers
+      if (err.response?.headers) {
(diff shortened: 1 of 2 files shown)


packages/cli-kit/src/private/node/api.ts
```diff
@@ -19,14 +19,20 @@ export async function debugLogResponseInfo<T extends {headers: Headers; status:
...
     response = await request
-    headers = response.headers
+    response.headers.forEach((value, key) => {
+      if (interestingResponseHeaders.has(key)) responseHeaders[key] = value
+    })
   } catch (err) {
     if (err instanceof ClientError) {
-      headers = err.response?.headers as Headers
+      if (err.response?.headers) {
+        for (const [key, value] of err.response?.headers as Iterable<[string, string]>) {
```
amcaplan, line 32: Why is this flipped vs `forEach`? `(value, key)` vs `[key, value]`

isaacroldan (approved):
Tested and it works with node 16 👌

PR #1583, opened 2023-03-08
Pass api-key for Event-Bridge validation

pacocastrotech:

Because we want to validate that the event-bridge arn is owned by a known app.
See:
* https://github.com/Shopify/shopify/pull/411020
* https://github.com/Shopify/partners/pull/45998

**IMPORTANT:** this PR has to be deployed after those two

Making sure that we have an api-key if event-bridge is used and passing it to Partners
No UI changes, only prompted for app (if you belong to more than one org) if secret not set manually

### How to test your changes?

See the PRs above for testing instructions.
With that spin setup:
* execute the command with a non even-bridge url: Check the logs of partners and core to see that api_key is not sent ✅ 
* execute the command with an event-bridge url and set secret manually: api-key will be get from .env or partners and sent ✅ 
* execute the command with an event-bridge url and set secret automatically: api-key will be get from .env or partners during that step (only once) and then be sent ✅ 

### Post-release steps

Nothing yet: The parameter is just passed, but not validated to prevent breaking the flow during the rollout

### Measuring impact

### Checklist

This would be totally safe to rollback

### DIFF
--- a/packages/app/src/cli/services/webhook/request-sample.test.ts
+++ b/packages/app/src/cli/services/webhook/request-sample.test.ts
@@ -8,34 +8,35 @@ const sampleHeaders = '{ "header": "Header Value" }'
 vi.mock('@shopify/cli-kit/node/api/partners')
 
 const aToken = 'A_TOKEN'
+const anApiKey = 'AN_API_KEY'
+const inputValues = {
+  topic: 'A_TOPIC',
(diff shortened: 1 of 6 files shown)


cdarne (approved):
👍 tested 3 cases with success:
09-56-8k398-82qvw

PR #1588, opened 2023-03-10
UI kit docs

matteodepalo:

Developers wanting to use UI kit don't have a readme that can show them what is possible with it.

This PR adds a readme.md file that can be used as the starting point to use UI Kit. I've tried to avoid mentioning specific details and actual output examples here as they can become stale. For that I've recommended checking the comments at the top of the public functions which should be up to date.


docs/cli-kit/ui-kit/readme.md
```diff
@@ -0,0 +1,162 @@
...
+    - [`renderTextPrompt`](#rendertextprompt)
+  - [Async tasks](#async-tasks)
+    - [`renderConcurrent`](#renderconcurrent)
+    - [`renderTasks`](#rendertasks)
+
+# Using UI Kit
+
+## What is UI Kit
+
+UI Kit is a library that can be used to either ask for input from the user or render output to the terminal.
```
pepicrft, line 23: Nitpick: I'd say is a set of modules in the `@shopify/cli-kit` package.

docs/cli-kit/ui-kit/readme.md
```diff
@@ -0,0 +1,162 @@
...
+By using UI Kit you're making sure that whatever you're rendering in the terminal
+will be consistent with the rest of the output.
```
pepicrft, line 25: Nitpick: And also consistent across Shopify.

pepicrft (approved):

docs/cli-kit/ui-kit/readme.md
```diff
@@ -0,0 +1,162 @@
...
+}
+
+try {
+  await renderTasks([task])
+} catch {
+  console.log(task.errors.map(e => e.message).join('\n'))
+  // something went wrong0
+  // something went wrong1
+  // something went wrong2
+}
```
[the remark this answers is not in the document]
matteodepalo: I mentioned it above, in the `renderConcurrent` section as that component is only a part of the `renderConcurrent` parent component for now. If we end up making it a more generic component we can perhaps have a separate secion for it.

PR #1590, opened 2023-03-10
Pass ruby bin to cli2

Arkham:

A partner reported an issue on Slack where running theme commands would print this error:
```
Ruby is required to continue
```

This was surprising because the command executed by CLI3 was:
```
/usr/local/opt/ruby/bin/bundle exec /usr/local/opt/ruby/bin/ruby bin/shopify theme serve
  /Users/Bob/code/a-theme/dist --live-reload hot-reload --theme 1234
```

After further investigation we realized that the globally available version of `ruby` was installed through `rbenv`

image

Since CLI3 knows if a custom ruby is being used, we can pass along the information to CLI2.
If such env var is not passed, we fall back to using `ruby`.

### How to test your changes?

- Edit `/opt/homebrew/Cellar/shopify-cli/3.44.1/libexec/lib/node_modules/@shopify/theme/node_modules/@shopify/cli-kit/assets/cli-ruby/lib/shopify_cli/environment.rb`
- Find the line that says `output, status = context.capture2e("ruby", "--version")` and add a new line with `puts output`
- Run any command and you should see the global version of ruby
- after nightly homebrew is released, do the same and you will see the homebrew version of ruby

### Measuring impact

### Checklist

### DIFF
--- a/packages/cli-kit/assets/cli-ruby/lib/shopify_cli/constants.rb
+++ b/packages/cli-kit/assets/cli-ruby/lib/shopify_cli/constants.rb
@@ -68,7 +68,7 @@ module ShopifyCLI
     module SupportedVersions
       module Ruby
         FROM = "2.7.5"
-        TO = "3.2.0"
+        TO = "3.2.10"
       end
 
       module Node


packages/cli-kit/assets/cli-ruby/lib/shopify_cli/constants.rb
```diff
@@ -67,7 +68,7 @@ module EnvironmentVariables
     module SupportedVersions
       module Ruby
         FROM = "2.7.5"
-        TO = "3.2.0"
+        TO = "3.2.10"
```
gonzaloriestra, line 71: Does this work? It would be nice to allow 3.2.11 without having to change this.

```suggestion
        TO = "~> 3.2"
```

packages/cli-kit/assets/cli-ruby/lib/shopify_cli/constants.rb
```diff
@@ -67,7 +68,7 @@ module EnvironmentVariables
     module SupportedVersions
       module Ruby
         FROM = "2.7.5"
```
gonzaloriestra, line 70: I guess 2.7.0 should also work, no?

```suggestion
        FROM = "2.7.0"
```

gonzaloriestra (approved):

packages/cli-kit/assets/cli-ruby/lib/shopify_cli/constants.rb
Arkham: not sure why, but I noticed that it was 2.7.5 in our TS code as well, maybe for some security fixes?

gonzaloriestra: It was updated [here](https://github.com/Shopify/shopify-cli/pull/2572) when dropping support for 2.6 (approved by me 😅). But there's no explanation... @charlespwd do you remember if there's any good reason?

PR #1594, opened 2023-03-12
[Feature] Add `transform` command to support app codemods

### LINKED ISSUE #1542: [Feature] Migrate Polaris Migrator to Shopify CLI
## Overview

Port the Polaris Migrator and Polaris CLI over to Shopify CLI.

## Context

The Polaris team has created a [migration tool](https://polaris.shopify.com/tools/polaris-migrator) for helping Shopify app developers transform Polaris code so it is easy staying up-to-date.

This tool can better serve app developers by bringing it front-and-center through the Shopify CLI. Integrating it with the CLI will ultimately improve the migration tool DX and make the codemod platform available for libraries beyond Polaris.

## Tasks

- [ ] Port the [Polaris CLI](https://github.com/Shopify/polaris/tree/main/polaris-cli) over to the Shopify CLI
- [ ] Convert the Polaris Migrator into a generic codemod transformation tool to run various transformations

samrose3:

Fixes #1542 

Adds a `transform` command that can be used to apply codemod transformations to Shopify apps.

### Measuring impact

### Checklist


packages/app/src/cli/commands/app/transform.ts
```diff
@@ -0,0 +1,110 @@
...
+      env: 'SHOPIFY_FLAG_IGNORE',
+      default: '**/node_modules/**',
+    }),
+    cpus: Flags.string({
+      hidden: false,
+      description: 'Start at most N child processes to process source files',
+      env: 'SHOPIFY_FLAG_CPUS',
+      char: 'c',
+    }),
+    runInBand: Flags.boolean({
```
amcaplan, line 53: For consistency with other CLI commands, use `dash-case` not `camelCase`.

We should really create an ESLint rule for this.

packages/app/src/cli/commands/app/transform.ts
```diff
@@ -0,0 +1,110 @@
...
+      hidden: false,
+      description: 'Run serially in the current process',
+      env: 'SHOPIFY_FLAG_RUN_IN_BAND',
+      default: false,
+    }),
+    dry: Flags.boolean({
+      hidden: false,
+      description: 'Do a dry-run, no code will be edited.',
```
amcaplan, line 61: I don't think we use periods at the end of descriptions.

packages/app/src/cli/commands/app/transform.ts
```diff
@@ -0,0 +1,110 @@
...
+    options: Flags.string({
+      hidden: false,
+      description: 'Add transform options as a JSON string.',
+      env: 'SHOPIFY_FLAG_OPTIONS',
+    }),
+  }
+
+  async run(): Promise<void> {
+    const {flags} = await this.parse(Transform)
+    await transform({
```
amcaplan, line 94: 👍 on using services!

packages/app/package.json
```diff
@@ -53,14 +55,18 @@
     "function-runner": "4.0.2",
     "graphql-request": "5.2.0",
     "http-proxy": "1.18.1",
-    "javy-cli": "0.1.4",
+    "javy-cli": "0.1.2",
```
shauns, line 58: Was this downgrade deliberate?

samrose3: Interesting, thanks for catching this! This was not intensional and not sure how I accidentally modified it 😓 will revert this change 👍

packages/app/src/cli/commands/app/polaris/transform.ts
```diff
@@ -0,0 +1,60 @@
+import {transform} from '../../../services/transform.js'
+import Command from '../../../utilities/app-command.js'
+import {Flags} from '@oclif/core'
+import {globalFlags} from '@shopify/cli-kit/node/cli'
+
+export default class Transform extends Command {
+  static description = 'Code transformations for updating Polaris apps.'
```
amcaplan, line 7: We might consider adopting this into something like 
```suggestion
  static description = 'Update your Polaris app using code transformations.'
```
to better accord with the language in other commands.

packages/app/src/cli/commands/app/polaris/transform.ts
```diff
@@ -0,0 +1,60 @@
...
+  static flags = {
+    ...globalFlags,
+    include: Flags.string({
+      hidden: false,
+      description: 'Files or directory to transform',
```
amcaplan, line 12: This is a glob pattern, right? We should make that explicit in the description.

packages/app/src/cli/commands/app/polaris/transform.ts
```diff
@@ -0,0 +1,60 @@
...
+      env: 'SHOPIFY_FLAG_INCLUDE',
+    }),
+    package: Flags.string({
```
amcaplan, line 15: I think we can leave this out for now, until we have multiple packages which we might want to specify.

packages/app/src/cli/services/transform.ts
```diff
@@ -0,0 +1,258 @@
...
+
+export async function transform(options: TransformOptions) {
+  if (!((await isClean(cwd())) || options.dry)) {
+    if (options.force) {
+      renderWarning({
+        headline: 'Forcibly continuing',
+        body: 'You have opted to bypass the Git safety check. This may result in lost work.',
+      })
+    } else {
+      renderWarning({
```
amcaplan, line 50: Should this be a warning or an error? Feels like we didn't do the thing they asked for, so I'd expect an `AbortError`.

packages/app/src/cli/services/transform.ts
```diff
@@ -0,0 +1,258 @@
...
+    options.package ??
+    // Use the `@shopify/polaris-codemods` package as a default for now.
+    // This could be removed to prompt a list of supported codemod packages
+    '@shopify/polaris-codemods' ??
+    (await renderAutocompletePrompt({
+      message: 'Select a package',
+      choices: packages.map((pkg) => {
+        return {value: pkg, label: pkg}
+      }),
+    }))
```
amcaplan, line 74: Should this all live in code? Maybe we can just drop the prompt code for now, since it's not used yet. YAGNI etc.

packages/app/src/cli/services/transform.ts
```diff
@@ -0,0 +1,258 @@
...
+
+  const config = await fetchPackageConfig(npmPackageString, packageManager)
+  const transforms = {...config.transforms, ...config.presets}
+
+  const transform =
+    options.transform ??
+    (await renderAutocompletePrompt({
```
amcaplan, line 81: This should be extracted to a function.

packages/app/src/cli/services/transform.ts
```diff
@@ -0,0 +1,258 @@
...
+      message: 'Select a transform to apply',
+      choices: Object.keys(transforms).map((transformName) => {
+        return {
+          value: transformName,
+          label: transformName,
+          group: transformName in config.transforms! ? 'Transforms' : 'Presets',
```
amcaplan, line 87: There's probably a cleaner way to do this, maybe iterating over both `config.transforms` and `config.presets` directly.

packages/app/src/cli/services/transform.ts
```diff
@@ -0,0 +1,258 @@
...
+  if (!transformPath || !fs.existsSync(transformPath)) {
+    throw new AbortError(`No transform found for ${options.transform}`)
+  }
+
+  const transformInfo: TransformInfo = await import(transformPath)
+  const extensions = transformInfo.extensions ?? ['js', 'jsx', 'ts', 'tsx']
+  const includePath = options.include ?? './'
+  const globPattern = joinPath(includePath, `**/*.(${extensions.join('|')})`)
+
+  const filePaths = await glob(globPattern, {cwd: cwd()})
```
amcaplan, line 105: To be clear, you want this to run in the actual cwd from which the command is run, right?

packages/app/src/cli/commands/app/polaris/transform.ts
samrose3: This is useful for is we want to test a snapshot or other version of a package. For example, in Polaris we create snapshot releases before merging into `main` for release. Having this option allow the user to specify a target package to get the codemod from (ex: `@shopify/polarsi-codemods@[snapshot.version]`)

packages/app/src/cli/services/transform.ts
samrose3: Good point. I think this is what jscodeshift expects, but I will test it from a few sub directories to make sure the paths are correct.

packages/app/src/cli/services/transform.ts
```diff
@@ -0,0 +1,267 @@
...
+
+    if (info) {
+      return info
+    }
+
+    await packageManager.install(packageName, version)
+    info = packageManager.getInfo(packageName)
+
+    if (!info) {
+      throw new AbortError(`Unable to locate package files for package: '${packageName}'`)
```
amcaplan, line 181: When might we expect this to happen?

packages/app/src/cli/services/transform.ts
```diff
@@ -0,0 +1,267 @@
...
+async function fetchPackage(npmPackageString: string, packageManager: PluginManager): Promise<CodeshiftConfig> {
+  async function installPackageDeps(packageName: string, version?: string): Promise<IPluginInfo> {
+    let info = await packageManager.alreadyInstalled(packageName, version)
+
+    if (info) {
+      return info
+    }
+
+    await packageManager.install(packageName, version)
+    info = packageManager.getInfo(packageName)
```
amcaplan, line 178: I believe you can just do:

```suggestion
    info = await packageManager.install(packageName, version)
```

The return value of `install` [should be the same as `getInfo`](https://github.com/davideicardi/live-plugin-manager#pluginmanagerinstallname-string-version-string-promiseiplugininfo).

packages/app/src/cli/services/transform.ts
```diff
@@ -0,0 +1,267 @@
...
+
+    if (!info) {
+      throw new AbortError(`Unable to locate package files for package: '${packageName}'`)
+    }
+
+    const dependencies = Object.entries(info.dependencies)
+    await Promise.all(dependencies.map(([pkgName, version]) => installPackageDeps(pkgName, version)))
```
amcaplan, line 185: Why the recursion? Isn't `live-plugin-manager` supposed to install the dependencies automatically?

packages/app/src/cli/services/transform.ts
```diff
@@ -0,0 +1,209 @@
+import * as jscodeshift from 'jscodeshift/src/Runner.js'
```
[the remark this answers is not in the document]
samrose3: Oddly, this is the location [jscodeshift documents](https://github.com/facebook/jscodeshift#usage-js) for programatic usage. Since the package type is module and this is referencing a file rather than a package, the extension is needed.

packages/app/src/cli/services/transform.ts
samrose3: Theoretically, it never should. `getInfo` returns either `IPluginInfo` or `undefined`. By using `install` here instead, we can remove this check 👍

packages/app/src/cli/services/transform.ts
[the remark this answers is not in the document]
amcaplan: Screenshot 2023-05-10 at 22 24 09

🤔

PR #1605, opened 2023-03-15
Allow running deploy without extensions

matteodepalo:

Fixes https://github.com/Shopify/app-deploys/issues/349

Following up the work done [here](https://github.com/Shopify/partners/pull/46209) we want to enable deployments without extensions if you have the beta flag.

### How to test your changes?

- Enable the `appUiDeployments` beta flag
- Try deploying in an app with no extensions
- It should work
- It shouldn't work if you don't have the beta flag


isaacroldan (approved):

matteodepalo:
I'm actually going to wait before merging this PR as we do a check on `organization.app.appUiDeployments`, but the new check is going to be on the app itself so we don't need to move the `if` down after `ensureDeployContext`.

Issue #1603, opened 2023-03-15
[Bug]: npm run dev fails on default node app after `Using a different app than last time, sending uninstall webhook to app server`


buildpath-ian:
### Please confirm that you have:

- [X] Searched [existing issues](.) to see if your issue is a duplicate. (If you’ve found a duplicate issue, feel free to add additional information in a comment on it.)
- [X] Reproduced the issue in [the latest CLI version](https://www.npmjs.com/package/@shopify/cli).

### In which of these areas are you experiencing a problem?

App

### Expected behavior

When I create a new node app and run `npm run dev` I expect it to take me through the configuration and provide an app url i can visit to install the app. This is for a new app on a new dev store where I have installed no apps before.

### Actual behavior

Instead, it always tries to process an `APP_UNINSTALL` webhook on `http://localhost:[some-random-port]/api/webhooks`, receives no response, and fails. I never receive the app url output, and it is not configured in the app settings in admin (just has default example.com stuff).

`Using a different app than last time, sending uninstall webhook to app server`

### Verbose output

```shell
?  Which store would you like to use to view your project?
✔  buildpath2

2023-03-15T03:56:31.100Z: Storing app information for directory [REDACTED]inventory-sync:{
  "appId": "[REDACTED]",
  "title": "inventory-sync",
  "directory": "[REDACTED]",
  "storeFqdn": "[REDACTED]",
  "orgId": "344828"
}
2023-03-15T03:56:31.186Z: Sending "Partners" GraphQL request:
  query fetchSpecifications($api_key: String!) {
    extensionSpecifications(apiKey: $api_key) {
      name
      externalName
      externalIdentifier
      identifier
      gated
      options {
        managementExperience
        registrationLimit
      }
      features {
        argo {
          surface
        }
      }
    }
  }

With variables:
{
  "api_key": "[REDACTED]"
}

With request headers:
 - User-Agent: Shopify CLI; v=3.44.1
 - Keep-Alive: timeout=30
 - Sec-CH-UA-PLATFORM: win32
 - Content-Type: application/json

2023-03-15T03:56:31.447Z: Request to https://partners.shopify.com/api/cli/graphql completed in 253 m
With response headers:
 - cache-control: max-age=0, private, must-revalidate
 - content-type: application/json; charset=utf-8
 - etag: W/"926c769eab017ef99b12377312319ab5"
 - x-request-id: 648f560f-1c1b-44c9-b6ef-93b2fc40bdb2
    
[...]

PR #1613, opened 2023-03-16
Do no trigger app/uninstalled webhook for fresh app

matteodepalo:
Fixes https://github.com/Shopify/cli/issues/1603

`app/uninstalled` webhooks shouldn't be triggered for fresh apps.

### Testing changes

Create a new node app
cd to new app's dir
Run npm run dev
You shouldn't see `APP_UNINSTALLED webhook delivery failed` or  `Using a different app than last time, sending uninstall webhook to app server`

### DIFF
--- a/.changeset/forty-mugs-end.md
+++ b/.changeset/forty-mugs-end.md
@@ -1,0 +1,5 @@
+---
+'@shopify/app': patch
+---
+
+Fix bug related to app/uninstalled webhook being triggered for fresh apps


matteodepalo:
I see, I didn't consider that scenario. We could maybe fix this if we could know if there are apps installed before running the webhook, or we could ignore this as the message doesn't seem to cause the process to fail.

matteodepalo:
I'll close this PR then.

PR #1619, opened 2023-03-17
Add links to footnotes if terminal doesn't support hyperlinks

matteodepalo:

Fixes https://github.com/Shopify/internal-cli-foundations/issues/520
Links could wrap inside boxes and be hard to copy and paste

I've added some footnotes below the boxes that display all the links so that they can be easily copy and pasted. Here's an example.

Screenshot 2023-03-17 at 14 18 00

### How to test your changes?

Run `kitchen-sink static` and see the changes

### DIFF
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -123,7 +123,6 @@ module.exports = {
       files: ['src/private/node/ui/components/**/*.tsx'],
       extends: ['plugin:react/recommended', 'plugin:react-hooks/recommended'],
       rules: {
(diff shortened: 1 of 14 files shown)


gonzaloriestra (approved):
Working perfectly and code LGTM!

matteodepalo:
> So the problem is that there are links longer than the banner? Or we are wrapping them when we shouldn't?

Yes they can be longer than the banner and we are forced to wrap them. The alternative would be to have links inside banners "overflow" outside the borders but that would look aesthetically unappealing.

PR #1631, opened 2023-03-20
Fix flakey test in AutocompletePrompt.test.tsx

matteodepalo:

A test in `AutocompletePrompt.test.tsx` was flakey.

I've moved the `setTimeout` inside the `search` function definition instead of letting it start immediately from the outside which sometimes would cause the function to resolve before the end of the test.

### How to test your changes?

I've tried this 100 times with a script and couldn't reproduce the issue after the fix. Before the fix the issue would appear ~1/20 times.


Arkham (approved):
Thank you!

PR #1641, opened 2023-03-22
Upload drafts in dev

matteodepalo:

Fixes https://github.com/Shopify/develop-app-foundations/issues/469

As part of the unified app deployments work we want to make sure that running `dev` updates the drafts of all the extensions without custom previews.

I've split the UI extensions into two groups, previewable and non-previewable.
Non-previewable extensions need to have their draft updated whenever the source code changes or the config changes.
I've created a new `additionalProcess` that uses esbuild to watch for code changes and chokidar to watch for config changes.

### How to test your changes?

- Run `./bin/create-test-app.js --bare`
- cd into the created app
- Run `pnpm shopify app generate extension`
- Choose a web pixel extension
- Run `pnpm shopify app dev --verbose`
- See that the mutation to update the draft is being called

### DIFF
--- a/packages/app/src/cli/services/context/identifiers-extensions.ts
+++ b/packages/app/src/cli/services/context/identifiers-extensions.ts
@@ -35,7 +35,7 @@ export async function ensureExtensionsIds(
 
   for (const pending of matchExtensions.toConfirm) {
     // eslint-disable-next-line no-await-in-loop
-    const confirmed = options.force || (await matchConfirmationPrompt(pending.local, pending.remote))
+    const confirmed = await matchConfirmationPrompt(pending.local, pending.remote)
     if (!confirmed) return err('user-cancelled')
     validMatches[pending.local.localIdentifier] = pending.remote.uuid
   }
--- a/packages/app/src/cli/services/context/identifiers-extensions.test.ts
+++ b/packages/app/src/cli/services/context/identifiers-extensions.test.ts
@@ -457,7 +457,7 @@ describe('ensureExtensionsIds: asks user to confirm deploy', () => {
 
     // Then
     expect(deployConfirmationPrompt).not.toBeCalled()
-    expect(matchConfirmationPrompt).not.toBeCalled()
+    expect(matchConfirmationPrompt).toBeCalled()
   })
 })
 
--- a/packages/app/src/cli/services/context/identifiers-functions.test.ts
+++ b/packages/app/src/cli/services/context/identifiers-functions.test.ts
@@ -363,6 +363,6 @@ describe('ensureFunctionsIds: asks user to confirm deploy', () => {
 
     // Then
     expect(deployConfirmationPrompt).not.toBeCalled()
-    expect(matchConfirmationPrompt).not.toBeCalled()
+    expect(matchConfirmationPrompt).toBeCalled()
   })
 })
--- a/packages/app/src/cli/services/context/identifiers-functions.ts
+++ b/packages/app/src/cli/services/context/identifiers-functions.ts
@@ -17,7 +17,7 @@ export async function ensureFunctionsIds(
 
   for (const pending of matchFunctions.toConfirm) {
     // eslint-disable-next-line no-await-in-loop
-    const confirmed = options.force || (await matchConfirmationPrompt(pending.local, pending.remote))
+    const confirmed = await matchConfirmationPrompt(pending.local, pending.remote)
     if (!confirmed) return err('user-cancelled')
     validMatches[pending.local.localIdentifier] = pending.remote.id
   }
--- a/packages/app/src/cli/services/context/prompts.ts
+++ b/packages/app/src/cli/services/context/prompts.ts
@@ -11,9 +11,9 @@ import {
 
 export async function matchConfirmationPrompt(local: LocalSource, remote: RemoteSource) {
   return renderConfirmationPrompt({
-    message: `Deploy ${local.configuration.name} (local name) as ${remote.title} (name on Shopify Partners, ID: ${remote.id})?`,
+    message: `Match ${local.configuration.name} (local name) with ${remote.title} (name on Shopify Partners, ID: ${remote.id})?`,
     confirmationMessage: `Yes, that's right`,
-    cancellationMessage: `No, cancel deployment`,
+    cancellationMessage: `No, cancel`,
   })
 }
 


matteodepalo:
@ani-shopify this is also available for tophatting 🎩

matteodepalo:
@ani-shopify are you still have the latter issue? I recommend deleting node modules and installing the deps again.

Related to the first question, we ask it when we try to register extensions so the wording is wrong there, and we should probably not ask the question when doing dev in the first place.

ani-shopify:
I deleted `node_modules` and ran `pnpm install`. I got the following error:

image

Seems like it's not happy with my react version?

ani-shopify:
I just ran `pnpm install cli-kit@latest -w` but got the same error

matteodepalo:
>What about extensions/modules other than UI ones? I wonder if this could cause any issues when other extensions get added to CLI that aren't UI extensions (flow extensions, tax extension). What do you think about keeping a list of previewable and non-previewable extensions?

Yes we should definitely try to make this code more generic. I can add the logic once we config that the current behavior works.

Regarding your error, is that a new app?

ani-shopify:
Well here's how I was trying to test which I still couldn't get working:

- `$ pnpm create @shopify/app --path /Users/anitumanyan/Desktop/test`
- `$ pnpm shopify app generate extension --path=/Users/anitumanyan/Desktop/test/test-cli`
- `$ pnpm shopify app dev --path=/Users/anitumanyan/Desktop/test/test-cli`

But then I gave up on this approach and followed the steps @matteodepalo listed in the PR description. I was able to successfully tophat the PR without any issues.

I have a question though. When I run the `app dev` without `--verbose`

- for `Checkout UI` extension, there are some updates showing up
- for `Web pixel` extension, nothing shows up

https://user-images.githubusercontent.com/99370588/234130041-6c063fda-4ae9-4d88-b7de-313870196d70.mp4

Do you think we should show some type of message when the draft version of web pixel extension gets updated even when not running with `--verbose`?

cc @jessehoek @stephszeto

matteodepalo:
>Do you think we should show some type of message when the draft version of web pixel extension gets updated even when not running with --verbose?

Good point, I think we should add a message there.

jessehoek:
> Do you think we should show some type of message when the draft version of web pixel extension gets updated even when not running with --verbose?

Yep, I think it makes sense to have all extensions follow similar output behaviour. I don't have a lot of context of how these messages typically look, but let's try to have the message be similar to other extension update messages if possible.

matteodepalo:
@ani-shopify @mattnoakes I've added those messages to the regular logs and I've also made the tax_calculation extension follow the same behavior. Because these config-only extensions are still marked as "ui" extensions, it was just a matter of adding `isPreviewable: false` to the specification. For functions it will need to be different but that will also need to change the rest of the code quite a lot so I don't think making it generic at this stage is possible.

ani-shopify:
> @ani-shopify I've set the default to `false` and set all the other defined ui extensions to `true`.

Thanks for taking care of it! This looks good from my end. Maybe having a short demo/video recording of how things work might help to get @jessehoek and/or @MeredithCastile onboard with these changes from the DX perspective.

matteodepalo:
https://user-images.githubusercontent.com/151725/234904433-aa095e3b-f080-4f91-ba02-efa51103bc68.mp4

Sure! Here's a small video explaining the new output.

packages/app/src/cli/services/context/identifiers-extensions.ts
```diff
@@ -35,7 +35,7 @@ export async function ensureExtensionsIds(
 
   for (const pending of matchExtensions.toConfirm) {
     // eslint-disable-next-line no-await-in-loop
-    const confirmed = await matchConfirmationPrompt(pending.local, pending.remote)
+    const confirmed = options.force || (await matchConfirmationPrompt(pending.local, pending.remote))
```
matteodepalo, line 38: I was surprised to see that this wasn't here already. Any reason why not?

packages/app/src/cli/services/dev.ts
```diff
@@ -378,6 +414,55 @@ async function devUIExtensionsTarget({
...
+}: DevNonPreviewableExtensionsOptions) {
+  return {
+    prefix: 'extensions',
+    action: async (stdout: Writable, stderr: Writable, signal: AbortSignal) => {
+      await Promise.all(
+        extensions.map((extension) => {
+          const registrationId = remoteExtensions[extension.localIdentifier]
+          if (!registrationId) throw new AbortError(`Extension ${extension.localIdentifier} not found on remote app.`)
+
+          return Promise.all([
```
isaacroldan, line 446: does it make sense to have nested promise.all?
if you return here an array of promises, can we flatten that at the top level?

packages/app/src/cli/services/dev.test.ts
```diff
@@ -0,0 +1,70 @@
...
+    const stderr = new Writable()
+    const extension1 = await testUIExtension({
+      devUUID: '1',
+      directory: 'directory/path/1',
+    })
+
+    const extension2 = await testUIExtension({
+      devUUID: '2',
+      directory: 'directory/path/2',
+    })
```
isaacroldan, line 24: could you also add here an extension with isPreviewable: true, so that we test that it is not added to the non-previewable target?

packages/app/src/cli/services/context/identifiers-extensions.ts
isaacroldan: it could potentially break stuff. I think @Arkham was intentional here, if we are not sure about a match, is dangerous to force it

packages/app/src/cli/services/dev.ts
```diff
@@ -135,6 +140,15 @@ async function dev(options: DevOptions) {
   const extensionsIds = prodEnvIdentifiers.app === apiKey ? envExtensionsIds : {}
   localApp.extensions.ui.forEach((ext) => (ext.devUUID = extensionsIds[ext.localIdentifier] ?? ext.devUUID))
 
+  const {extensionIds: remoteExtensions} = await ensureDeploymentIdsPresence({
```
alvaro-shopify, line 143: Could we move this piece of code inside the following 

```ts
if (nonPreviewableExtensions.length > 0) 
```

it seems that `remoteExtensions` is only used inside thus we could avoid making API requests

packages/app/src/cli/services/context/identifiers-extensions.ts
matteodepalo: Wouldn't people expect `-f` to skip all prompts and accept the consequences? I've had to add this because in `dev` I was getting the deploy match confirmation prompt. I suppose I could solve this by changing the prompt messaged based on wether it's being used by `dev` or `deploy`.

alvaro-shopify: now that we are working on running the CLI as a service in spin we should consider an alternative to avoid prompting the user for any confirmation otherwise it won't work as expected, what do you think? @isaacroldan @matteodepalo @Arkham

packages/app/src/cli/services/dev/update-extension.ts
```diff
@@ -0,0 +1,85 @@
...
+    throw new AbortError(errorMessage)
+  }
+
+  const specification = await findSpecificationForConfig(specifications, extension.configurationPath, abort)
+
+  if (!specification) {
+    return
+  }
+
+  const configuration = await parseConfigurationFile(specification.schema, extension.configurationPath, abort)
```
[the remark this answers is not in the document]
matteodepalo: It doesn't, it only gives you the path of the watched file and some stats.

packages/app/src/cli/services/context/identifiers-extensions.ts
matteodepalo: Could we just throw instead of prompting if `force` is true?

PR #1651, opened 2023-03-23
Allow cloudflare to retry the tunnel creation

isaacroldan:
If cloudflare fails to obtain a tunnel, don't reject on the first error, keep trying until the timeout is triggered (20s right now) 

This includes 4 retries. Cloudflare will retry after 1s, then 2s, 4s and 8s, after that the timeout will be triggered and we will stop retrying.

### How to test your changes?
Make cloudflare fail, the only way to force is to disable wifi i guess 🤔 

### Measuring impact

### Checklist

### DIFF
--- a/packages/plugin-cloudflare/src/tunnel.ts
+++ b/packages/plugin-cloudflare/src/tunnel.ts
@@ -26,7 +26,8 @@ export async function hookStart(port: number): ReturnType {
 }
 
 async function tunnel(options: {port: number}): Promise<{url: string}> {
-  const args: string[] = ['tunnel', '--url', `http://localhost:${options.port}`, '--no-autoupdate']
+  const args: string[] = ['tunnel', '--url', `http://localhost:${options.port}`, '--no-autoupdate', '--retries', '3']
+  const errors: string[] = []
 
   let connected = false
   let resolved = false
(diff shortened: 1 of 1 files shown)


packages/plugin-cloudflare/scripts/postinstall.js
```diff
@@ -140,6 +147,17 @@ function sha256(filePath) {
...
+  const currentParts = current.split('.')
+  const expectedParts = expected.split('.')
+  for (let i = 0; i < currentParts.length; i++) {
+    if (parseInt(expectedParts[i]) > parseInt(currentParts[i])) {
+      return true
+    }
+  }
+  return false
+}
+
```
Arkham, line 160: could we use  `semver.gt` instead? we already have it in the monorepo

packages/plugin-cloudflare/src/tunnel.ts
```diff
@@ -36,7 +38,8 @@ async function tunnel(options: {port: number}): Promise<{url: string}> {
     setTimeout(() => {
       if (!resolved) {
         resolved = true
-        reject(new Error('Timed out waiting for a cloudflare tunnel'))
+        const lastErrors = errors.slice(-5).join('\n')
+        reject(new Error(`Timed out waiting for a cloudflare tunnel: ${lastErrors}`))
```
Arkham, line 42:
```suggestion
        reject(new Error(`Timed out while creating a cloudflare tunnel: ${lastErrors}`))
```

Arkham (approved):
Looks good, left some minor comments :)

packages/plugin-cloudflare/scripts/postinstall.js
isaacroldan: done, wasn't sure about adding a new dependency here but it's 👌

PR #1653, opened 2023-03-23
Implement disabled prompt choices

matteodepalo:

Fixes https://github.com/Shopify/internal-cli-foundations/issues/581

We want to communicate to the user better why certain extensions are not available.

Adding this functionality with the previous `SelectInput` implementation was very difficult so I've heavily refactored the state logic to:

- Use `useReducer` instead of `useState` to encapsulate it in an easier to understand reducer function.
- Not rely on `onChange` from `SelectPrompt` but instead pass an `onSubmit` function to `SelectInput`. This means that now `SelectInput` is in charge of handling the return key and the various shortcuts if `submitWithShortcuts` is true.
- `SelectInput` doesn't deal with indexes anymore but rather with a map of options that each have a `next` or `previous` properties. This way it's easy to check if the next or previous items are disabled or not.

Functionality-wise, I've removed the infinite scrolling that would reset the selected item to the top of the list when you reach the last item and press down arrow. Same for arrow up but in reverse.
I did this because if we want to properly paginate server side in the future we cannot support this behavior as pressin up arrow would try to fetch the last page, which is not something we can do with our GraphQL APIs AFAIK.

Example in `kitchen-sink prompts`:

### How to test your changes?

- Try creating an extension that has a limit of 1
- Run `extension generate` again and see that the option is greyed out and not selectable


gonzaloriestra (commented):
It looks great in `kitchen-sink`, but I can't see the disabled options in the generate command (the unavailable extensions simply don't appear).

isaacroldan:
Found another issue: The first item in the list is selected by default even if it's disabled, and you can select it:
Captura de pantalla 2023-05-29 a las 12 56 13

Captura de pantalla 2023-05-29 a las 12 56 16

matteodepalo:
@isaacroldan so strange, I can't reproduce the problem with the flavor select. Could you try with the latest commit?

isaacroldan:
@matteodepalo tested again and everything seems to be working as expected :)
btw, you need to rebase with main, this branch still has the old pnpm-lock

isaacroldan (approved):

matteodepalo:
I've fixed the issue with the initial value being disabled.
I've made it so if there are no enabled options it throws an exception because it doesn't make sense to have a prompt in that case, the user wouldn't be able to continue. Same with specifying that a default value that corresponds with a disabled option, it doesn't make sense to render the prompt in that case. One could argue that we could just autoselect the next available option instead of throwing. What do you think?

PR #1661, opened 2023-03-24
Upgrade esbuild to 0.17.10

matteodepalo:

As part of the work done here https://github.com/Shopify/cli/pull/1641 we'd like to upgrade esbuild so that we can run async functions in the event running after the build. I thought it would be better to make this upgrade separate so we can test it in isolation.

This PR upgrades esbuild to the latest version. The main change is that there is no longer a `build` function, but rather a `context` function that returns an object with `watch`, `rebuild`, etc... These functions can then be called to trigger the previous behavior. There is also no longer an `onRebuild` param as part of `watch`.
As a result, we need to add a plugin that subscribes to the `onEnd` event. This event gets run after every build, not just the second rebuild, so we don't need to trigger `onResult` manually on the first build.

### How to test your changes?

Working with `dev` and `extensions` should keep working. Try making changes to an extensions and check that the extensions are being rebuilt correctly.

### DIFF
--- a/.changeset/gorgeous-ducks-yawn.md
+++ b/.changeset/gorgeous-ducks-yawn.md
@@ -1,0 +1,5 @@
+---
+'@shopify/app': patch
+---
+
+Update esbuild to 0.17.10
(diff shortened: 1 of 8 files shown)


alvaro-shopify (approved):
LGTM!. Great job

Issue #1664, opened 2023-03-24
[Bug]: shopify theme dev suddenly stopped working


brianecook:
### Please confirm that you have:

- [X] Searched [existing issues](.) to see if your issue is a duplicate. (If you’ve found a duplicate issue, feel free to add additional information in a comment on it.)
- [X] Reproduced the issue in [the latest CLI version](https://www.npmjs.com/package/@shopify/cli).

### In which of these areas are you experiencing a problem?

Theme

### Expected behavior

I run `shopify theme dev --store store-slug` and a local development server is created for me to make changes on my machine.

### Actual behavior

Error is thrown: Cannot read properties of undefined (reading 'toString'). When I run the command with the --verbose flag, this is the last message before the error is thrown:

2023-03-24T22:55:52.557Z: Sending Identity Introspection request to URL: undefined

I'm assuming there should be a url in place of undefined.

### Verbose output

```shell
2023-03-24T22:55:52.511Z: Running command theme dev
  shopify:theme:dev init version: @oclif/core@2.1.4 argv: [ '--store', 'store-slug, '--verbose' ] +0ms
2023-03-24T22:55:52.552Z: Ensuring that the user is authenticated with the Theme API with the following scopes:
[]

2023-03-24T22:55:52.552Z: Ensuring that the user is authenticated with the Admin API with the following scopes for the store store-slug.myshopify.com:
[]

2023-03-24T22:55:52.554Z: Getting session store...
2023-03-24T22:55:52.556Z: Validating existing session against the scopes:
[
  "openid",
  "https://api.shopify.com/auth/shop.admin.graphql",
  "https://api.shopify.com/auth/shop.admin.themes",
  "https://api.shopify.com/auth/partners.collaborator-relationships.readonly",
  "https://api.shopify.com/auth/shop.storefront-renderer.devtools",
  "https://api.shopify.com/auth/partners.app.cli.access"
]
For applications:
{
  "adminApi": {
    "scopes": [],
    "storeFqdn": "oakandeden.myshopify.com"
  }
}

2023-03-24T22:55:52.557Z: Sending Identity Introspection request to URL: undefined
╭─ error ──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│                                                                                                                                          │
│  Cannot read properties of undefined (reading 'toString')                                                                                │
│                                                                                                                                          │
│  To investigate the issue, examine this stack trace:                                                                                     │
[...]

PR #1687, opened 2023-03-29
Fix introspection URL cache

gonzaloriestra:

Fixes https://github.com/Shopify/cli/issues/1664

Some users apparently got an invalid cache (missing the `value` field):

```
"cache": {
	"identity-introspection-url-accounts.shopify.com": {
		"timestamp": 1679500965290
	}
}
```

Ignore the cache when the value is not present

### How to test your changes?

- Open `~/Library/Preferences/shopify-cli-kit-nodejs/config.json` and remove the value field
- `shopify app dev`

### Measuring impact

### Checklist

### DIFF
--- a/packages/cli-kit/src/private/node/conf-store.ts
+++ b/packages/cli-kit/src/private/node/conf-store.ts
@@ -65,7 +65,8 @@ type CacheValueForKey<TKey extends keyof Cache> = NonNullable<Cache[TKey]>['valu
  * before returning it.
  * @param key - The key to use for the cache.
  * @param fn - The function to run to get the value to cache, if a cache miss occurs.
- * @param timeout - The maximum valid age of a cached value, in milliseconds. If the cached value is older than this, it will be refreshed.
+ * @param timeout - The maximum valid age of a cached value, in milliseconds.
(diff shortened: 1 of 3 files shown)


packages/cli-kit/src/private/node/conf-store.ts
```diff
@@ -77,7 +78,7 @@ export async function cacheRetrieveOrRepopulate(
   const cache: Cache = config.get('cache') || {}
   const cached = cache[key]
 
-  if (cached && (timeout === undefined || Date.now() - cached.timestamp < timeout)) {
+  if (cached?.value && (timeout === undefined || Date.now() - cached.timestamp < timeout)) {
```
amcaplan, line 81: This won't work if the cached value is `false`.

amcaplan: This is my point. If `cached.value === false` it will treat it as uncached.

gonzaloriestra: Oh, right! 🤦

amcaplan (approved):

PR #1688, opened 2023-03-29
[cli-kit] add graphqlRequest as public export and remove oxygenRequest

vincentezw:
Currently cli-kit exports two functions `oxygenRequest` and `uploadOxygenDeploymentFile`. These are remnants and aren't functional.

However we are looking to create a plugin to enable deployments to Oxygen. Since Oxygen logic is not likely to be re-used by other plugins this logic can better live in our plugin. This PR removes those two functions.

In its place it exports the `graphqlRequest` function instead, which will allow us to use this in our plugin without having to load external modules. We will require some custom headers, so the function now accepts a header object rather than a token string (previously the function called `buildHeaders` with that token).

I also added a test for it.

### Checklist


packages/cli-kit/src/public/node/api/admin.ts
```diff
@@ -18,7 +19,7 @@ export async function adminRequest<T>(query: string, session: AdminSession, vari
   const api = 'Admin'
   const version = await fetchApiVersion(session)
   const url = adminUrl(session.storeFqdn, version)
-  return graphqlRequest(query, api, url, session.token, variables)
+  return graphqlRequest(query, api, url, buildHeaders(session.token), variables)
```
isaacroldan, line 22: All graphql calls should include the default headers, so I think it'd be better for this to be inside `graphqlRequest`.
Is there a reason why oxygen wouldn't use `buildHeaders`?

packages/cli-kit/src/public/node/api/graphql.ts
```diff
@@ -0,0 +1,50 @@
...
+ * @param options - GraphQL response options.
+ */
+export async function graphqlRequest<T>(
+  query: RequestDocument,
+  api: string,
+  url: string,
+  headers: {[header: string]: string},
+  variables?: Variables,
+  options?: GraphQLResponseOptions<T>,
+): Promise<T> {
```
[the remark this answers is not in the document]
vincentezw: great suggestion, added this

packages/cli-kit/src/public/node/api/admin.ts
vincentezw: Oxygen-hub requires a header `X-Oxygen-Namespace-Name` which contains the namespace for the app that's deploying. For Hydrogen storefronts that would be a static value, but that may change should we allow non-storefront deployments. I think it's best if we have control over this header from our plugin.

vincentezw: that makes perfect sense. I've added an optional `addedHeaders` argument 👍

isaacroldan (approved):

Issue #868, opened 2022-12-06
Problems with Umlauts in home path under windows 10 (incompatible character encodings: CP850 and UTF-8)


ConcurrentHashMap:
Hey there,

I tried upgrading from shopify-cli 2 to 3 today, but faced some major errors - because of the umlaut in my lastname.

1. `shopify theme pull` did result in an error, that bundler is not installed (although it was). Running `bundler -v` directly resulted in the same `incompatible character encodings` error as below. But I managed to get this fixed by pointing the ENV `BUNDLE_USER_HOME` and `GEM_HOME` outside of my home directory. 

2. `shopify theme dev` also fails, due to the shopify-cli package using its own cache directory (still inside my home path with umlauts), so it also fails. (by the way, sending a stacktrace is even not possible due to the problem that bugsnag also fails with incompatible character encodings). Stack trace of this below:

```
C:/Users/Nils Buschhüter/AppData/Local/shopify-cli-nodejs/Cache/vendor/ruby-cli/2.32.0/ruby/3.0.0/gems/bugsnag-6.25.0/lib/bugsnag/stacktrace.rb:44:in `start_with?': incompatible character encodings: CP850 and UTF-8 (Encoding::CompatibilityError)
        from C:/Users/Nils Buschhüter/AppData/Local/shopify-cli-nodejs/Cache/vendor/ruby-cli/2.32.0/ruby/3.0.0/gems/bugsnag-6.25.0/lib/bugsnag/stacktrace.rb:44:in `block in process'
        from C:/Users/Nils Buschhüter/AppData/Local/shopify-cli-nodejs/Cache/vendor/ruby-cli/2.32.0/ruby/3.0.0/gems/bugsnag-6.25.0/lib/bugsnag/stacktrace.rb:22:in `map'
        from C:/Users/Nils Buschhüter/AppData/Local/shopify-cli-nodejs/Cache/vendor/ruby-cli/2.32.0/ruby/3.0.0/gems/bugsnag-6.25.0/lib/bugsnag/stacktrace.rb:22:in `process'
        from C:/Users/Nils Buschhüter/AppData/Local/shopify-cli-nodejs/Cache/vendor/ruby-cli/2.32.0/ruby/3.0.0/gems/bugsnag-6.25.0/lib/bugsnag/report.rb:434:in `block in generate_exception_list'
[...]

Issue #1634, opened 2023-03-21
[Bug]: Bundler not found When installing Shopify Cli 3.x


ErikZoeppig:
### Please confirm that you have:

- [X] Searched [existing issues](.) to see if your issue is a duplicate. (If you’ve found a duplicate issue, feel free to add additional information in a comment on it.)
- [X] Reproduced the issue in [the latest CLI version](https://www.npmjs.com/package/@shopify/cli).

### In which of these areas are you experiencing a problem?

Theme

### Expected behavior

When running Shopify Theme dev it should redirect me to Partner Login and create Links for Development like Shopify Theme serve on CLi 2.x

### Actual behavior

Showing up following Error:
Error: Bundler not found
    at validateBundler (file:///C:/Users/ErikZ%C3%B6ppig/AppData/Roaming/npm/node_modules/@shopify/theme/node_modules/@shopify/cli-kit/dist/public/node/ruby.js:176:15)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async validateRubyEnv (file:///C:/Users/ErikZ%C3%B6ppig/AppData/Roaming/npm/node_modules/@shopify/theme/node_modules/@shopify/cli-kit/dist/public/node/ruby.js:145:5)
    at async installCLIDependencies (file:///C:/Users/ErikZ%C3%B6ppig/AppData/Roaming/npm/node_modules/@shopify/theme/node_modules/@shopify/cli-kit/dist/public/node/ruby.js:128:5)

### Verbose output

```shell
2023-03-21T12:27:07.892Z: Running command theme dev
  shopify:theme:dev init version: @oclif/core@2.1.4 argv: [ '--verbose' ] +0ms
2023-03-21T12:27:07.919Z: Ensuring that the user is authenticated with the Theme API with the following scopes:
[]

2023-03-21T12:27:07.920Z: Ensuring that the user is authenticated with the Admin API with the following scopes for the store hofmanns-shop.myshopify.com:
[]

2023-03-21T12:27:07.922Z: Getting session store...
2023-03-21T12:27:07.926Z: Validating existing session against the scopes:
[
  "openid",
  "https://api.shopify.com/auth/shop.admin.graphql",
  "https://api.shopify.com/auth/shop.admin.themes",
  "https://api.shopify.com/auth/partners.collaborator-relationships.readonly",
  "https://api.shopify.com/auth/shop.storefront-renderer.devtools",
[...]

PR #1693, opened 2023-03-30
Set a custom BUNDLE_USER_HOME env var on Windows

Arkham:

If the username contains a UTF8 character, bundler will crash with this error:
```
PS C:\Users\Jöhnny Ärkhåm> bundle --version
C:/Ruby30-x64-utf8/lib/ruby/3.0.0/pathname.rb:50:in `match?': invalid byte sequence in UTF-8 (ArgumentError)
        from C:/Ruby30-x64-utf8/lib/ruby/3.0.0/pathname.rb:50:in `chop_basename'
        from C:/Ruby30-x64-utf8/lib/ruby/3.0.0/pathname.rb:374:in `plus'
        from C:/Ruby30-x64-utf8/lib/ruby/3.0.0/pathname.rb:354:in `+'
        from C:/Ruby30-x64-utf8/lib/ruby/3.0.0/pathname.rb:420:in `join'
```

Fixes https://github.com/Shopify/cli/issues/868 and https://github.com/Shopify/cli/issues/1634

Set a custom `BUNDLE_USER_HOME` var on Windows so that instead of using the user home folder we use `C:\Users\Public\AppData\Local\shopify-bundler-nodejs\Cache` as the Bundler user home.

### How to test your changes?

- On Windows, create a new user with UTF8 characters
- Make it an administrator
- Search for "environment variables", click edit env vars, and add `C:\Ruby30-x64\bin` (or whatever is your installed ruby) to the `PATH`
- `git clone https://github.com/Shopify/cli.git`
- `node bin/create-test-app.js -e theme`
- Witness it crashing
- Remove the app folder
- Change the branch from `main` to `fix-windows-uft8-bundler-bug`
- `node bin/create-test-app.js -e theme`and you should see it work now

### Measuring impact

### Checklist

### DIFF
--- a/packages/cli-kit/src/public/node/ruby.ts
+++ b/packages/cli-kit/src/public/node/ruby.ts
@@ -1,6 +1,6 @@
 import {AbortSignal} from './abort.js'
 import {platformAndArch} from './os.js'
-import {captureOutput, exec} from './system.js'
+import {captureOutput, exec, ExecOptions} from './system.js'
 import * as file from './fs.js'
 import {joinPath, dirname, cwd} from './path.js'
 import {AbortError, AbortSilentError} from './error.js'
@@ -68,7 +68,7 @@ export async function execCLI2(args: string[], options: ExecCLI2Options = {}): P
 
   try {
     const shopifyExecutable = embedded ? [rubyExecutable(), await embeddedCLIExecutable()] : ['shopify']
-    await exec(bundleExecutable(), ['exec', ...shopifyExecutable, ...args], {
+    await runBundler(['exec', ...shopifyExecutable, ...args], {
       ...(options.stdout === undefined && {stdio: 'inherit'}),
       cwd: options.directory ?? cwd(),
       env,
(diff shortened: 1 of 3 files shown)


packages/cli-kit/src/public/node/ruby.ts
```diff
@@ -443,15 +443,35 @@ async function getSpinEnvironmentVariables() {
...
+}
+
+/**
+ * It returns a custom BUNDLE_USER_HOME. This is required in Windows because
+ * bundler will instead crash if the username contains UTF-8 characters.
+ *
+ * @returns The value of the environment variable.
+ */
+export function bundleUserHome(): string | undefined {
+  if (platformAndArch().platform === 'windows' && process.env.PUBLIC) {
```
isaacroldan, line 462: why do you need this env var to be present?

Arkham: because otherwise I'm not sure where the `Public` folder is, since it's windows it could be `C:\Users\Public` or `D:\Users\Public` etc :)

matteodepalo (approved):
LGTM

isaacroldan (approved):

PR #1702, opened 2023-03-30
Add strict true to tsconfig for checkout extensions

Steve-D-Powell:

Fixes Shopify/temp-project-mover-megswim-20250326195306#334

When a user chooses the Typescript flavour of checkout ui extension the typing rules are too loose for the IDE, leading to frustration for Partners.

EG: Some values received from the standard extension API will always be `undefined` unless the extension renders at a specific time in the checkout process. Without the `strict` rule, this is ignored by the IDE and it looks like the requested value should have something other than `undefined`. (See issue linked above).

Adds `"strict": true` in the `compilerOptions` for the tsconfig.json that is generated for Typescript flavour checkout UI extensions.

### How to test your changes?

* Clone repo `git clone https://github.com/Shopify/cli.git `
* `cd cli` -> `pnpm install`
* `git checkout add-strict-rule-checkout-ui-extensions`
* Run `bin/create-test-app.js`
  * Note that this command created a UI extension right after the app, you can CTRL + C that part, or wait until it's finished 
* cd into app directory
* [Generate a new extension](https://shopify.dev/docs/apps/tools/cli/commands#generate-extension) - making sure to choose "Typescript React"
* Open `/app/extensions/extension-name` in your editor
* Check that `tsconfig.json` has the `"strict": true` option in  `compilerOptions`

### Measuring impact

### Checklist

### DIFF
--- a/packages/app/templates/ui-extensions/projects/checkout_ui_extension/tsconfig.json
+++ b/packages/app/templates/ui-extensions/projects/checkout_ui_extension/tsconfig.json
@@ -4,7 +4,8 @@
   // Changing options here won't affect the build of your extension
   "compilerOptions": {
     "jsx": "react-jsx",
-    "esModuleInterop": true
+    "esModuleInterop": true,
+    "strict": true
   },
   "include": ["./src"]
 }


alxclark (approved):
I didn't tophat, but it's a sane default 👍🏼  It feels a bit bad for partners that already created a bad config, but I don't think we can do much about it.

PR #1709, opened 2023-03-31
Use unifiedAppDeployment flag instead of appUiDeployments

matteodepalo:

We want to move the beta flag for unified deployments to the org level because we're getting rid of the org level one and also because the partners token doesn't have permission to fetch organizations, so the `deploy` command couldn't make use of such flags.

I've removed `appUiDeployments` from organizations and added `unifiedAppDeployment` to apps.

### How to test your changes?

- Enable the `unified_app_deployment` beta flag in partners
- The new flow should be active when deploying

### DIFF
--- a/packages/app/src/cli/api/graphql/all_orgs.ts
+++ b/packages/app/src/cli/api/graphql/all_orgs.ts
@@ -4,9 +4,6 @@ export interface AllOrganizationsQuerySchemaOrganization {
   id: string
   businessName: string
   website?: string
-  betas: {
-    cliTunnelAlternative?: boolean
-  }
 }
 
 export interface AllOrganizationsQuerySchema {
@@ -22,9 +19,6 @@ export const AllOrganizationsQuery = gql`
         id
         businessName
         website
-        betas {
-          cliTunnelAlternative
-        }
       }
     }
   }
--- a/packages/app/src/cli/api/graphql/all_orgs.ts
+++ b/packages/app/src/cli/api/graphql/all_orgs.ts
@@ -5,7 +5,7 @@ export interface AllOrganizationsQuerySchemaOrganization {
   businessName: string
   website?: string
   betas: {
-    appUiDeployments?: boolean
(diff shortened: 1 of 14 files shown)


packages/app/src/cli/api/graphql/all_orgs.ts
```diff
@@ -5,7 +5,7 @@ export interface AllOrganizationsQuerySchemaOrganization {
   businessName: string
   website?: string
   betas: {
-    appUiDeployments?: boolean
+    cliTunnelAlternative?: boolean
```
Arkham, line 8: ??

matteodepalo: @isaacroldan I've seen this flag pop up recently but I haven't seen it queried for consistently so this change brings consistency.

packages/app/src/cli/services/deploy.test.ts
```diff
@@ -234,8 +284,8 @@ async function testDeployBundle(app: AppInterface, organization?: Organization,
   await deploy({
     app,
     reset: false,
-    force: true,
-    label,
+    force: options?.force ?? false,
```
Arkham, line 287:
```suggestion
    force: Boolean(options?.force),
```

is this the same?

packages/app/src/cli/services/context.ts
```diff
@@ -317,24 +315,21 @@ export async function ensureDeployContext(options: DeployContextOptions): Promis
...
 
   await logMetadataForLoadedDeployContext(result)
   return result
 }
 
-export async function fetchOrganizationAndFetchOrCreateApp(
-  app: AppInterface,
-  token: string,
-): Promise<{partnersApp: OrganizationApp; organization: Organization}> {
+export async function fetchOrganizationAndFetchOrCreateApp(app: AppInterface, token: string): Promise<OrganizationApp> {
```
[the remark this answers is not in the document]
matteodepalo: Good point 👍

PR #1761, opened 2023-04-04
Exit prerelease

pepicrft:
I'm doing a new minor release of the CLI and therefore I need to exit the pre-release mode of changeset.


Arkham (approved):
:)

PR #1762, opened 2023-04-04
Add 3.45.0 release notes and switch changeset mode to pre-release again

pepicrft:
After having bumped the versions to 3.45.0, I'm opening a PR to add the 3.45.0 release notes and switch the changeset mode back to pre-release.


Arkham (approved):
🎊

PR #1768, opened 2023-04-04
Not upgrade the global Homebrew-managed installation of the CLI on `shopify upgrade`

pepicrft:
When a Homebrew-managed installation of the CLI tries to upgrade itself via Homebrew, it leads to [an error](https://shopify.slack.com/archives/C030LHFCA5T/p1680618843167109) because Homebrew deletes modules of the current version that OCLIF fails to load dynamically.

Upgrades and installation should be managed from a process other than the CLI's, otherwise we'll continue to find race conditions like this one where the package manager has replaced modules that are needed for the CLI process that initiates the upgrade. Because installing and upgrading packages is the package managers' responsibility, we should not have that responsibility in the CLI.

**Note:** If the issue is that `package.json` have two packages that need to be in sync, for example `@shopify/app` and `@shopify/cli`, we should investigate how to reduce that to one package such that a simply `npm update` does the job.

### Post-release steps
We need to [merge the PR](https://github.com/Shopify/shopify-dev/pull/32617) that updates the docs.

### Measuring impact

### Checklist

### DIFF
--- a/packages/cli/src/cli/services/upgrade.ts
+++ b/packages/cli/src/cli/services/upgrade.ts
@@ -17,20 +17,28 @@ type HomebrewPackageName = 'shopify-cli' | 'shopify-cli@3'
 // Canonical list of oclif plugins that should be installed globally
 const globalPlugins = ['@shopify/theme']
 
-export async function upgrade(directory: string, currentVersion: string): Promise<void> {
+interface UpgradeOptions {
+  env: NodeJS.ProcessEnv
+}
+
+export async function upgrade(
+  directory: string,
+  currentVersion: string,
+  {env}: UpgradeOptions = {env: process.env},
+): Promise<void> {
   let newestVersion: string | void
 
   const projectDir = await getProjectDir(directory)
   if (projectDir) {
     newestVersion = await upgradeLocalShopify(projectDir, currentVersion)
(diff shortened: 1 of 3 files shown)


packages/cli/src/cli/services/upgrade.ts
```diff
@@ -77,26 +77,15 @@ async function upgradeGlobalShopify(currentVersion: string): Promise<string | vo
 
   outputUpgradeMessage(currentVersion, newestVersion)
 
-  const homebrewPackage = process.env.SHOPIFY_HOMEBREW_FORMULA as HomebrewPackageName | undefined
   try {
-    await (homebrewPackage ? upgradeGlobalViaHomebrew(homebrewPackage) : upgradeGlobalViaNpm())
+    await upgradeGlobalViaNpm()
   } catch (err) {
```
isaacroldan, line 82: if we detect this is a homebrewPackage, shouldn't we cancel and warn the user to use brew?
Otherwise it will try to upgrade the global shopify using npm

pepicrft: Yep! I'm changing the code to have that behaviour instead.

isaacroldan (approved):

PR #1769, opened 2023-04-04
Async tunnels

isaacroldan:
- Tunnel creation is part of the `dev` process and can take a few seconds to initiate the tunnel.
Fixes #0000 

- Be optimistic about the tunnel, create as soon as possible in the background. When the tunnel is finally needed, poll the tunnel plugin for the URL every 0.5 seconds.
- Update cloudflare and ngrok plugins to use this new async interface

### Measuring impact

### Checklist


packages/plugin-ngrok/src/tunnel.ts
```diff
@@ -9,9 +9,21 @@ import {outputToken, outputInfo, outputContent} from '@shopify/cli-kit/node/outp
 
 export default startTunnel({provider: TUNNEL_PROVIDER, action: hookStart})
 
+let cachedPort: number
```
pepicrft, line 12: Introducing state in module is often a recipe for issues because we start developing logic on the assumption there's a single module in the graph, but that assumption might not hold true at some point. If the `dev` logic interacts with plugins to start, stop, and get the status, I think the state should be kept there and passed as needed.

packages/cli-kit/src/public/node/plugins/tunnel.ts
```diff
@@ -33,15 +47,38 @@ export interface HookReturnPerTunnelPlugin {
...
+export type TunnelStopFunction = FanoutHookFunction<'tunnel_start', ''>
+export type TunnelStatusFunction = FanoutHookFunction<'tunnel_status', ''>
+
 export type TunnelStartReturn = PluginReturnsForHook<'tunnel_start', ''>
+export type TunnelStopReturn = PluginReturnsForHook<'tunnel_start', ''>
+export type TunnelStatusReturn = PluginReturnsForHook<'tunnel_status', ''>
+
 export type TunnelStartAction = (port: number) => Promise<TunnelStartReturn>
+export type TunnelStatusAction = () => Promise<TunnelStatusReturn>
+export type TunnelStopAction = () => TunnelStopReturn
```
pepicrft, line 59: I'd be more flexible here and return `Promise<TunnelStopReturn>` here like we are doing with status and start.

packages/plugin-cloudflare/src/stop.ts
```diff
@@ -0,0 +1,5 @@
+import {TUNNEL_PROVIDER} from './provider.js'
+import {abortController} from './tunnel.js'
```
pepicrft, line 2: Similar here, I'd avoid introducing side-effects in module to store state:

```js
const tunnel = startTunnel({...})
// tunnel.port & tunnel.abortController
const status = await getTunnelStatus(tunnel)
await stopTunnel(tunnel)
```

packages/app/src/cli/services/dev.ts
```diff
@@ -65,6 +66,11 @@ interface DevWebOptions {
 }
 
 async function dev(options: DevOptions) {
+  // Be optimistic about tunnel creation and do it as early as possible
```
pepicrft, line 69: The scattering of tunnel-related logic in the function makes it hard to reason about and brittle. What about having a function like:

```js
await startTunnelOptimistically({defaultTunnel: 'cloudflare', fallback: 'ngrok'}, async (tunnelReady) => {
 // First part of the dev logic
 
 const provider = await tunnelReady 

// Continue with dev
 
})
```

pepicrft (approved):

packages/plugin-cloudflare/src/tunnel.ts
```diff
@@ -10,61 +11,67 @@ import {fileURLToPath} from 'url'
...
+    },
+  })
 
-    // eslint-disable-next-line @typescript-eslint/no-floating-promises
-    exec(getBinPathTarget(), args, {
-      stdout: customStdout,
-      stderr: customStdout,
-    })
+  // eslint-disable-next-line @typescript-eslint/no-floating-promises
+  exec(getBinPathTarget(), args, {
```
shauns, line 71: We should test this with `exit 1` -- i.e. what happens if the commands fails quickly (vs. it eventually times out, which I think we do cover for already)

packages/plugin-ngrok/src/status.ts
```diff
@@ -0,0 +1,12 @@
+import {TUNNEL_PROVIDER} from './provider.js'
+import {getCurrentStatus} from './tunnel.js'
+import {tunnelStatus} from '@shopify/cli-kit/node/plugins/tunnel'
+import {ok} from '@shopify/cli-kit/node/result'
+
+export default tunnelStatus({
+  provider: TUNNEL_PROVIDER,
+  action: async () => {
+    const aa = await getCurrentStatus()
```
shauns, line 9: Weird variable name :)

packages/plugin-ngrok/src/tunnel.ts
isaacroldan: true, removed this cachedPort as there isn't really a need for it

packages/app/src/cli/services/dev.ts
isaacroldan: Is not that straightforward, we don't use ngrok as a fallback, only if there is a flag enabled remotely that tell us to use it. This is the process:

1. Start cloudflare as early as possible
2. Authenticate User if needed
3. Fetch Org/App information
4. Check if the Org beta flags forces us to us ngrok (we added this as a safety measure)
5. If the flag tell us to use ngrok, stop cloudflare tunnel and start ngrok. If not, do nothing.
6. Later in the flow, request the tunnel URL for whatever tunnel provider is the active one.

If everything goes well, in a couple of releases we should completely remove ngrok, removing steps 4 and 5 which will simplify the tunneling logic in `dev`

packages/plugin-cloudflare/src/tunnel.ts
isaacroldan: Added an automatic retry if the process crashes before getting a valid URL and connection.

Once we get a valid URL and the connection is stablished, if it crashes we can't do anything, just let it crash with a better message. (because the CLI is already using that URL and can't tell it to use a new one)

packages/cli-kit/src/public/node/system.ts
```diff
@@ -16,6 +16,7 @@ export interface ExecOptions {
   stdio?: 'inherit'
   input?: string
   signal?: AbortSignal
+  externalErrorHandler?: (error: unknown) => Promise<void>
```
shauns, line 19: we should document what this is for (vs. wrapping the call to `exec` in a `try/catch`

pepicrft (approved):

packages/plugin-cloudflare/src/tunnel.ts
```diff
@@ -107,7 +116,7 @@ class TunnelClient {
 
         outputDebug('Cloudflared tunnel crashed, restarting...')
         // wait 1 second before restarting the tunnel, to avoid rate limiting
-        await sleep(1)
+        if (!isUnitTest) await sleep(1)
```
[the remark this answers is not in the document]
isaacroldan: I tried, but since we have two timers (this one and the global timeout), If I disable timers then the global timeout is triggered and I can't test this case. Any suggestion? 🤔

packages/app/src/cli/services/dev.ts
isaacroldan: i've simplified this a bit. We now receive a tunnelClient that we pass to the needed part of the flow, we can ask the client to stop the tunnel or to get the current status

packages/plugin-ngrok/src/tunnel.ts
```diff
@@ -1,19 +1,42 @@
...
 import ngrok from '@shopify/ngrok'
 import {renderFatalError, renderTextPrompt} from '@shopify/cli-kit/node/ui'
-import {err, ok, Result} from '@shopify/cli-kit/node/result'
+import {err, ok} from '@shopify/cli-kit/node/result'
 import {AbortError} from '@shopify/cli-kit/node/error'
 import {outputToken, outputInfo, outputContent} from '@shopify/cli-kit/node/output'
 
 export default startTunnel({provider: TUNNEL_PROVIDER, action: hookStart})
 
+export const getCurrentStatus = async (): Promise<TunnelStatusType> => {
```
shauns, line 18: I think this function can go now?

packages/cli-kit/src/public/node/plugins/tunnel.ts
```diff
@@ -1,8 +1,20 @@
-import {err, Result} from '../result.js'
 import {ExtendableError} from '../error.js'
 import {FanoutHookFunction, PluginReturnsForHook} from '../plugins.js'
+import {err, Result} from '../result.js'
 
 export type TunnelErrorType = 'invalid-provider' | 'tunnel-already-running' | 'wrong-credentials' | 'unknown'
+export interface TunnelClient {
```
shauns, line 6: ❤️

PR #1775, opened 2023-04-05
Handle DeploymentError structure from partners

matteodepalo:

Fixes https://github.com/Shopify/develop-app-foundations/issues/659

After https://github.com/Shopify/partners/pull/45686 has been merged we can now use this new `DeploymentError` structure to display errors that happen while creating a deployment. What was before coming through as validation errors is now surfaced in `userErrors`.

- Made it so `message`, the first argument of `AbortError`, can be a `TokenItem` so that we can render a more complex error message with Ink.
- If the error contains details containing the `extension_id` then we show the extension identifier next to the error like in this example, which before was a validation error:

Screenshot 2023-04-05 at 15 44 44

### How to test your changes?

- Run `./bin/create-test-app.js --bare`
- cd into the created app
- Run `pnpm shopify app generate extension`
- Choose a `Customer accounts` extension
- Run `pnpm shopify app deploy`
- You should get an error message like in the example

If you can think of other errors and would like to test them please feel free to do so.

### DIFF
--- a/packages/app/src/cli/services/deploy/upload.test.ts
+++ b/packages/app/src/cli/services/deploy/upload.test.ts
@@ -915,7 +915,7 @@ describe('uploadExtensionsBundle', () => {
           },
           {
             title: 'admin-link',
-            body: '1 error found in your extension. Fix these issues in the Partner Dashboard and try deploying again.',
+            body: '\n1 error found in your extension. Fix these issues in the Partner Dashboard and try deploying again.',
           },
         ])
       }
@@ -1028,7 +1028,7 @@ describe('deploymentErrorsToCustomSections', () => {
       },
       {
         title: 'admin-link',
-        body: '1 error found in your extension. Fix these issues in the Partner Dashboard and try deploying again.',
+        body: '\n1 error found in your extension. Fix these issues in the Partner Dashboard and try deploying again.',
       },
     ])
   })


matteodepalo:
Thank you for tophatting @ani-shopify and you raise a good point.
Previously we distinguished between `userErrors` and `validationErrors` present in the `deployedVersions`.
If there were `userErrors` we just concatenated the messages with `, ` and displayed an error box. If there were just `validationErrors` then we showed success with the errors like in your screenshot.
Since now we just look at `userErrors` we always render the error box, but we don't have a way to distinguish between the two types of messages. I wonder if there is a way to know type of error cc @gordonhirsch.

gordonhirsch:
> I wonder if there is a way to know type of error

We also added a `category` field on each user error. For validation errors, the category will be "invalid", so you could check for that. Let me know if you find that it is not working as expected.

ani-shopify:
@gordonhirsch I'm guessing this is the list where the `category` field is being fetched from:

```ruby
def category
    case object
    when User::NotFound
      :not_found
    when User::UndeletableError
      :undeletable
    when User::ValidationError
      :invalid
    when Catastrophic::EventError
      :event
    when Catastrophic::InternalError
      :internal
    when Catastrophic::PluginError
      :plugin
    when Catastrophic::UnknownError
      :unknown
    end
  end
```

@matteodepalo maybe you can use `category` and this to have a somewhat more helpful message to the end user?

matteodepalo:
Great thank you I'll use that 👍

matteodepalo:
I should have improved the error message, could you try tophatting again @ani-shopify ? 🙏

matteodepalo:
Screenshot 2023-04-21 at 15 49 25

Updated error message.

mattnoakes:
I created an admin link from the dashboard and left it in an invalid state, resulting in the top error block. Is this messaging expected?

21-43-baf75-e5i3o

mattnoakes:
<details>
<summary>
Here's what the graphql response looks like
</summary>

```json
{
    "data": {
        "deploymentCreate": {
            "deployment": null,
            "userErrors": [
                {
                    "message": "can't be blank",
                    "field": [
                        "text"
                    ],
                    "details": [
                        {
                            "type": "version",
                            "version_id": 92427681793,
                            "extension_id": 23540531201,
                            "specification_identifier": "app_link",
                            "app_id": 38781583361
                        }
                    ],
                    "category": "invalid"
                },
                {
                    "message": "is invalid",
                    "field": [
                        "url"
                    ],
                    "details": [
                        {
                            "type": "version",
                            "version_id": 92427681793,
                            "extension_id": 23540531201,
                            "specification_identifier": "app_link",
                            "app_id": 38781583361
                        }
                    ],
                    "category": "invalid"
                },
                {
                    "message": "must be a valid page",
                    "field": [
                        "context"
                    ],
                    "details": [
                        {
                            "type": "version",
                            "version_id": 92427681793,
                            "extension_id": 23540531201,
                            "specification_identifier": "app_link",
                            "app_id": 38781583361
                        }
                    ],
                    "category": "invalid"
                },
                {
                    "message": "Missing expected key(s).",
                    "field": [
                        "base"
                    ],
                    "details": [
                        {
                            "type": "version",
                            "version_id": 92427583489,
                            "extension_id": 23540465665,
                            "specification_identifier": "customer_accounts_ui_extension",
                            "app_id": 38781583361
                        }
                    ],
                    "category": "invalid"
                }
            ]
        }
    }
}
```
</details>

<details>
<summary>
Here's what the verbose logs look like
</summary>

https://user-images.githubusercontent.com/65420305/234022170-85fb7094-9820-4b7b-8d0f-071cc2507707.mp4

</details>

<details>
<summary>
Without the beta flag, deploy successfully pushes the customer accounts extension to Shopify
</summary>

24-13-3wsy1-zmiw6

</details>

matteodepalo:
Ah very interesting, that makes sense.
We can add a query to fetch dashboard managed extensions before we deploy so that we can link the extension id to them if needed in both scenarios. I'll work on this next.

jessehoek:
I think it would be a good idea to tell devs they have validation errors in their dashboard managed extensions. It would be helpful also to give a count of how many validation errors there are.

So something like:

```
There has been an error creating your deployment.

`dashboard-managed-extension-title`

N validation errors found in your extension. Fix these issues in the Partner Dashboard and try deploying again. 
<link to specific extension edit page>

`specialized-account-ext`

Validation errors found in your extension toml file
- Missing expected key(s)
```

 The unfortunate thing is, you have to click "Save" first to see the validation errors show up on the page in the dashboard. So the user would get directed to a page where they can't see the errors until they hit Save. 

I know @nikijiandani is working on a similar [issue](https://github.com/Shopify/partners/pull/46977) for showing validation errors when creating a deployment from the dashboard. In that case, we show that there are N errors on the create deployment page, and then if you click the extension name from that page, it takes you to the page and automatically displays the validation errors. I wonder if we can do the same thing when clicking a link from the CLI.

ani-shopify:
I had a quick chat with @fredriculrich about adding the extension titles to the `detail` object (as he initially was against adding anything other than IDs in the `detail` object). He agrees at least right now it would make more sense to add the title here and avoid CLI to make an additional call to partners to fetch the list of dashboard-managed extensions.

I created [this issue](https://github.com/Shopify/app-deploys/issues/421) that will expose the extension title. Will try to tackle that issue this week so we can unblock you @matteodepalo.

ani-shopify:
Pending reviews on [this PR](https://github.com/Shopify/shopify/pull/422507) to make `:title` available in the details object/hash.

matteodepalo:
It does also feel like the error sentences are meant to be displayed next to the web ui, but in this case it would be better to include the attribute name in the message. I'm not expert with partners but I know that in rails typically this is the difference between `error.messages` and `error.full_messages.` Not sure if it's possible to surface that.

gordonhirsch:
> I wonder if it's possible to surface where the error happens? Either that or the extension type?

Maybe we could differentiate between the extension ids that CLI knows about and those that it doesn't? That would be a good indicator about whether the error is in a CLI-managed extension. 

Not sure exactly which "type" we are talking about in this context, but all the version and extension details objects do contain "specification_identifier" which is an extension type of sorts.

matteodepalo:
I've pushed code that uses `field`, however I'm not able to check admin link validation errors since partners has changed and doesn't allow the creation of admin-link extensions that are not valid. @ani-shopify @gordonhirsch any tips on testing this behavior?

ani-shopify:
> I've pushed code that uses `field`, however I'm not able to check admin link validation errors since partners has changed and doesn't allow the creation of admin-link extensions that are not valid. @ani-shopify @gordonhirsch any tips on testing this behavior?

Let me see if I can reproduce. @mattnoakes might know some tricks as well.

ani-shopify:
@matteodepalo I just 🎩 ed this.

I created an app with a web pixel extension and an admin link. And got both of these two extensions in an erroneous state. I was able to see the errors in CLI.
image

This is not to do anything with CLI. But I noticed that when I added a theme app extension to the app, added an asset with en extension that is not supported (like *.webp), made sure CLI doesn't run the validations, the backend only sent theme app extension related error messages and skips to include web pixel and admin link related error messages. @gordonhirsch do you have an idea why this would be the case? Maybe we can pair on this tomorrow to figure it out.

https://github.com/Shopify/cli/assets/99370588/3f55a66b-0e60-4f13-afb3-f7e6bd3df38c


image

isaacroldan (approved):
🎩 and works as expected

PR #1776, opened 2023-04-06
Use unifiedAppDeployment flag instead of appUiDeployments

matteodepalo:

This is round number 2 for https://github.com/Shopify/cli/pull/1709 which got reverted. We now only need `read_app` to fetch beta flags so partners token won't break.

----

We want to move the beta flag for unified deployments to the ~org~ app level because we're getting rid of the org level one and also because the partners token doesn't have permission to fetch organizations, so the deploy command couldn't make use of such flags.

I've removed appUiDeployments from organizations and added unifiedAppDeployment to apps.

## How to test your changes?
Enable the unified_app_deployment beta flag in partners
The new flow should be active when deploying

### DIFF
--- a/packages/app/src/cli/services/dev/fetch.test.ts
+++ b/packages/app/src/cli/services/dev/fetch.test.ts
@@ -20,12 +20,10 @@ import {mockAndCaptureOutput} from '@shopify/cli-kit/node/testing/output'
 const ORG1: Organization = {
   id: '1',
   businessName: 'org1',
-  betas: {},
 }
 const ORG2: Organization = {
   id: '2',
   businessName: 'org2',
-  betas: {},
 }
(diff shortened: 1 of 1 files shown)


ani-shopify:
I tested the CLI version in this PR locally and I can confirm that everything works as expected both when the beta flag is ON and OFF. Details below.

- Beta flag disabled:

https://user-images.githubusercontent.com/99370588/231270802-7a85b5f6-a3f7-42af-b749-1f487624a8bf.mp4


- Beta flag enabled:

https://user-images.githubusercontent.com/99370588/231270517-72aaecef-aad7-43e6-9063-5d790c8c301f.mp4

matteodepalo:
Merging this as the previous PR had already been approved and the code didn't change.

PR #1785, opened 2023-04-10
Fix localhost link in the `shopify theme dev` command

karreiro:

The label for `http://127.0.0.1:9292` was absent.

This PR introduces the label field in the `shopify theme dev` command

### How to test your changes?

- Run `shopify theme dev`

**Before**

**After**

### Post-release steps

None.

### Measuring impact

### Checklist


karreiro:
@matteodepalo I've fixed the issue by adding the `label` field in the `shopify theme dev` command. Still, I personally feel like we should use `url` as the default label when `label` is undefined.

Let me know if you agree with that, then I can update this PR :) Otherwise, we may proceed with this minor fix.

Poitrin (approved):

matteodepalo:
@karreiro Thank you for opening this PR. I think the problem with using the url as label if the url is not present is that if the url is very long we're going to have a duplication of information if the url is quite short, like in this case. I guess that would still be better than the current state and would fix all links without labels so I'd go for your suggestion

> Still, I personally feel like we should [use url as the default label when label is undefined](https://github.com/Shopify/cli/blob/main/packages/cli-kit/src/private/node/ui/components/Link.tsx#L13-L24).

matteodepalo:
Could you make sure that there is a test that checks what happens if there is no label?

karreiro:
Thank you for the review, @matteodepalo! I've applied the changes in the new commit :)

PR #1798, opened 2023-04-12
Add a new Remix template to the CLI

byrichardpowell:

We want shopifolk to be able to create a new Remix app using the CLI.  Right now, the template that is checked out doesn't fully work since it's waiting on https://github.com/Shopify/cli/pull/1789

Add a new Remix templateto the available templates.  If you work for Shopify this allows you to run `yarn create @shopify/app` and select Remix.  Or, shopifolk can use `yarn create @shopify/app --template=remix`.

Non shopifolk do not get access to this template.

### How to test your changes?
As a shopify employee:

1. `pnpm create-app --template=remix`
2. `pnpm create-app --template=node`
3. `pnpm create-app` and select remix`
4. `pnpm create-app` and select node` 

Now fake not working for Shopify (the horror!) by changing `isShopify` to return false.  Re-run the above commands and check:

1. `pnpm create-app` should not present remix as an option in the choice list.
2.`pnpm create-app --template=remix` should throw an error.
3. `pnpm create-app --template=node` should still work.

### Post-release steps

Since this is only available to shopifolk I don't think there are any.

### Measuring impact

### Checklist

### DIFF
--- a/packages/create-app/src/commands/init.test.ts
+++ b/packages/create-app/src/commands/init.test.ts
@@ -1,5 +1,5 @@
 import Init from './init.js'
-import initPrompt, {templateURLMap} from '../prompts/init.js'
+import initPrompt, {getTemplateURLMap, templateURLMap, templateURLMapForShopifolk} from '../prompts/init.js'
 import initService from '../services/init.js'
 import {describe, expect, vi, beforeEach, test} from 'vitest'
 import {errorHandler} from '@shopify/cli-kit/node/error-handler'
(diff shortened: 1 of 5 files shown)


alvaro-shopify:
Hi @byrichardpowell . 

> The problem I found was flags are static, but we need to call an async function to now if the Remix value should be available or not. Any ideas on how to fix this? Alternatively, since the Remix template is just for Shopifolk, would it be ok if the Remix value wasn't in the flag description? What do you think?

Sorry for this, you are completely right!, descriptions are statically generated using `refresh-manifests`. I guess I don't know any workaround for this issue, so I'm agreed with your suggestion of not adding `Remix` to the template flag description

byrichardpowell:
I'm going to close this PR in favor of a much simpler PR that adds the Remix template for all users.

PR #1801, opened 2023-04-13
Fix peer dependency resolution issue with npm

alvaro-shopify:
When you generate a react ui extension using `npm` and react resolution dependency error is produced and the command fails.

On the one hand UI extension components libraries add react as a peer dependency. On the other the cli adds the react dependency as well. The problem is that `npm` can't figure out how to resolve the dependency graph when there is a different react version and you try to install both libraries at the same time.

- In case of using npm as the package manager to generate an extension, the cli will install the required dependencies one by one. In this case the list of dependencies has the correct order but you must ensure that the most priority one should be installed before the others

### How to test your changes?
- Create a new app using
- Build this branch
- Copy the content of the file `node-package-manager.js` from the `dist` folder of the branch to the `node_modules` folder of the created app.

### Measuring impact

### Checklist

### DIFF
--- a/packages/cli-kit/src/public/node/node-package-manager.ts
+++ b/packages/cli-kit/src/public/node/node-package-manager.ts
@@ -394,22 +394,31 @@ export async function addNPMDependencies(
   options: AddNPMDependenciesIfNeededOptions,
 ): Promise<void> {
   let args: string[]
-  const depedenciesWithVersion = dependencies.map((dep) => {
+  const dependenciesWithVersion = dependencies.map((dep) => {
     return dep.version ? `${dep.name}@${dep.version}` : dep.name
   })
+  options.stdout?.write(`Installing ${[dependenciesWithVersion].join(' ')} with ${options.packageManager}`)
   switch (options.packageManager) {
     case 'npm':
-      args = argumentsToAddDependenciesWithNPM(depedenciesWithVersion, options.type)
+      // npm isn't too smart when resolving the dependency tree. For example, admin ui extensions include react as
+      // a peer dependency, but npm can't figure out the relationship and fails. Installing dependencies one by one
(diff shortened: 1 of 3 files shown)


packages/cli-kit/src/public/node/node-package-manager.ts
```diff
@@ -435,7 +444,7 @@ export async function addNPMDependenciesWithoutVersionIfNeeded(
  * @param type - The dependency type.
  * @returns An array with the arguments.
  */
-function argumentsToAddDependenciesWithNPM(dependencies: string[], type: DependencyType): string[] {
+function argumentsToAddDependenciesWithNPM(dependencies: string[], type: DependencyType, force = false): string[] {
```
gonzaloriestra, line 447: Do we need this?

```suggestion
function argumentsToAddDependenciesWithNPM(dependencies: string[], type: DependencyType): string[] {
```

packages/cli-kit/src/public/node/node-package-manager.ts
```diff
@@ -394,22 +394,31 @@ export async function addNPMDependencies(
   options: AddNPMDependenciesIfNeededOptions,
 ): Promise<void> {
   let args: string[]
-  const depedenciesWithVersion = dependencies.map((dep) => {
+  const dependenciesWithVersion = dependencies.map((dep) => {
     return dep.version ? `${dep.name}@${dep.version}` : dep.name
   })
+  options.stdout?.write(`Installing ${[dependenciesWithVersion].join(' ')} with ${options.packageManager}`)
```
gonzaloriestra, line 400:
```suggestion
  options.stdout?.write(`Installing ${[dependenciesWithVersion].join(', ')} with ${options.packageManager}`)
```

Issue #1818, opened 2023-04-17
[Bug]: Single process / monoliths Shopify apps cannot set custom callback urls 


matthewattanasio:
### Please confirm that you have:

- [X] Searched [existing issues](.) to see if your issue is a duplicate. (If you’ve found a duplicate issue, feel free to add additional information in a comment on it.)
- [X] Reproduced the issue in [the latest CLI version](https://www.npmjs.com/package/@shopify/cli).

### In which of these areas are you experiencing a problem?

App

### Expected behavior

I have a monolith version of a shopify app, e.g. Node js / Koa / Nextjs all run from a single process. 

Based on the documentation [found here](https://shopify.dev/docs/apps/tools/cli/structure#shopify-web-toml), I should be able to create a single `shopify.web.toml` file in the root folder where my app project lives and I must use the `type = frontend` since my app has only one process.

The problem is when set `type = frontend` the property `auth_callback_path` is not respected. 

In my `shopify.web.toml` I have set the callback URLs to be the following.
`auth_callback_path = ["/auth", "/auth/callback", "/install/auth/callback", "/install/auth"]`

However this setting is not respected and when I run `shopify app dev` the `allowed redirection urls` on my apps settings in Shopify is now. 

```
https://cloudflare_generated_url/auth/callback
https://cloudflare_generated_url/auth/shopify/callback
https://cloudflare_generated_url/api/auth/callback
```

However, if I set `type = backend`, the `allowed redirections urls` are updated correctly on app settings, but the CLI doesn't expose a cloudflare URL that I can use to access my app.... 

I am filing this as a bug as the `auth_callback_path` property should be respected when there is no additional `shopify.web.toml` to accommodate for monolith based projects which were only a year ago the official way to create Shopify based apps. 

### Actual behavior

`allowed redirections urls` in app settings are updated incorrectly when a single `shopify.web.toml` file is used for a single process/ monolith Shopify app. 

### Verbose output

```shell
With request headers:
 - User-Agent: Shopify CLI; v=3.45.0
 - Keep-Alive: timeout=30
 - Sec-CH-UA-PLATFORM: darwin
 - Content-Type: application/json

2023-04-17T00:16:11.992Z: Request to https://partners.shopify.com/api/cli/graphql completed in 406 ms
With response headers:
 - cache-control: max-age=0, private, must-revalidate
 - content-type: application/json; charset=utf-8
 - etag: W/"6cab3adcdbff629320781d8f3f841c6a"
[...]

PR #1828, opened 2023-04-18
Use authCallbackPath in frontend configuration

matteodepalo:

Fixes https://github.com/Shopify/cli/issues/1818

As mentioned in the [docs](https://shopify.dev/docs/apps/tools/cli/structure#shopify-web-toml) it should be possible to set `auth_callback_path` in the frontend config and have the app urls be updated accordingly when running `dev`. The frontend configuration is also of type `WebConfigurationSchema` and contains that property.

I've made it so we check both frontend and backend configurations for `authCallbackPath`, giving priority to the backend.

### How to test your changes?

- Generate an app
- Update the frontend toml `auth_callback_path` with a custom array of urls
- Run `dev`
- The app urls should be updated

### DIFF
--- a/packages/app/src/cli/services/dev.ts
+++ b/packages/app/src/cli/services/dev.ts
@@ -111,7 +111,10 @@ async function dev(options: DevOptions) {
   let previewUrl
 
   if (initiateUpdateUrls) {
-    const newURLs = generatePartnersURLs(exposedUrl, backendConfig?.configuration.authCallbackPath)
+    const newURLs = generatePartnersURLs(
+      exposedUrl,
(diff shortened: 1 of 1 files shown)


amcaplan:
I'm wondering about the precedence order. Fundamentally, all apps need a frontend, some require a backend but not all. So, should we perhaps prefer frontend?

And, generally, this is yet another reason to replace the frontend/backend thing with something else...

matteodepalo:
I guess the order is more about, what happens if there are two conflicting `auth_callback_path` both in the backend andf rontend, which one wins? I'd probably expect the backend?

isaacroldan (approved):

PR #1846, opened 2023-04-24
Add abortController to async ui function parameters

matteodepalo:

At the moment there is no way for async processes to make the async prompts quit. For example in this scenario, the prompt would still remain in the terminal:

```
const templatesPromise = getLatestTemplates()
    .catch((error) => {
      renderFatalError(error);
      process.exit(1);
    });

await renderXPrompt(...);
```

image (6)

This PR adds the ability to pass an abort controller from outside. Prompts and `Tasks` will clear the output before quitting when the abort controller signal receives the `abort` event.
`ConcurrentOutput` will clear only the dynamic part, currently the preview URL and the shortcuts, but will preserve the logs.

### How to test your changes?

I've tested it modifying `kitchen-sink prompts` and replacing the first prompt with this code

```
  const controller = new AbortController()

  // after 2 seconds call abort
  setTimeout(() => {
    controller.abort()
  }, 2000)

  // renderSelectPrompt
  await renderSelectPrompt({
    message: 'Associate your project with the org Castile Ventures?',
    choices: [
      {label: 'first', value: 'first', key: 'f'},
      {label: 'second', value: 'second', key: 's'},
      {label: 'third', value: 'third'},
      {label: 'fourth', value: 'fourth'},
      {label: 'fifth', value: 'fifth', group: 'Automations', key: 'a'},
      {label: 'sixth', value: 'sixth', group: 'Automations'},
      {label: 'seventh', value: 'seventh'},
      {label: 'eighth', value: 'eighth', group: 'Merchant Admin'},
      {label: 'ninth', value: 'ninth', group: 'Merchant Admin'},
      {label: 'tenth', value: 'tenth'},
    ],
    infoTable: {add: ['new-ext'], remove: ['integrated-demand-ext', 'order-discount']},
    abortController: controller,
  })
```

The prompt should disappear.

### DIFF
--- a/packages/cli-kit/src/private/node/ui/components/TextPrompt.tsx
+++ b/packages/cli-kit/src/private/node/ui/components/TextPrompt.tsx
@@ -3,7 +3,7 @@ import {TokenizedText} from './TokenizedText.js'
 import {handleCtrlC} from '../../ui.js'
 import useLayout from '../hooks/use-layout.js'
 import {messageWithPunctuation} from '../utilities.js'
-import {AbortController} from '../../../../public/node/abort.js'
+import {AbortSignal} from '../../../../public/node/abort.js'
 import useAbortSignal from '../hooks/use-abort-signal.js'
 import React, {FunctionComponent, useCallback, useState} from 'react'
 import {Box, useApp, useInput, Text} from 'ink'
@@ -17,7 +17,7 @@ export interface TextPromptProps {
   validate?: (value: string) => string | undefined
   allowEmpty?: boolean
   emptyDisplayedValue?: string
-  abortController?: AbortController
+  abortSignal?: AbortSignal
 }
 
 const TextPrompt: FunctionComponent<TextPromptProps> = ({
@@ -28,7 +28,7 @@ const TextPrompt: FunctionComponent<TextPromptProps> = ({
   password = false,
   allowEmpty = false,
   emptyDisplayedValue = '(empty)',
-  abortController,
+  abortSignal,
 }) => {
   if (password && defaultValue) {
     throw new Error("Can't use defaultValue with password")
@@ -58,7 +58,7 @@ const TextPrompt: FunctionComponent<TextPromptProps> = ({
   const shouldShowError = submitted && error
   const color = shouldShowError ? 'red' : 'cyan'
   const underline = new Array(oneThird - 3).fill('▔')
-  const {isAborted} = useAbortSignal(abortController?.signal)
+  const {isAborted} = useAbortSignal(abortSignal)
 
   useInput((input, key) => {
     handleCtrlC(input, key)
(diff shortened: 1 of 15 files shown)


packages/cli-kit/src/public/node/ui.tsx
```diff
@@ -267,13 +273,14 @@ export interface RenderConfirmationPromptOptions extends Pick<SelectPromptProps<
...
  */
-export function renderConfirmationPrompt({
+export async function renderConfirmationPrompt({
   message,
   infoTable,
   confirmationMessage = 'Yes, confirm',
   cancellationMessage = 'No, cancel',
   renderOptions,
   defaultValue = true,
+  abortController,
```
pepicrft, line 291: What you should be passing instead is the [signal](https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal), the controller is only needed in the caller context to send an abort signal through.

packages/cli-kit/src/private/node/ui/hooks/use-abort-signal.ts
```diff
@@ -0,0 +1,17 @@
...
+import {useLayoutEffect, useState} from 'react'
+
+export default function useAbortSignal(abortSignal?: AbortSignal) {
+  const {exit: unmountInk} = useApp()
+  const [isAborted, setIsAborted] = useState(false)
+
+  useLayoutEffect(() => {
+    abortSignal?.addEventListener('abort', () => {
+      setIsAborted(true)
+      unmountInk()
```
pepicrft, line 12: What would happen if I want to abort a component (e.g. a loading bar) but continue to run the CLI using Ink for with other components? Would this make Ink unusable after aborting?

packages/cli-kit/src/public/node/ui.tsx
matteodepalo: I was debating this. The reason I chose to pass the controller is that `ConcurrentOutput` already accepts a controller as a param so I wanted to keep these functions as similar as possible. We also don't know if in the future we'll need to call abort inside the components so I wanted to future proof this API so that we don't have to change it.

packages/cli-kit/src/private/node/ui/components/TextPrompt.tsx
```diff
@@ -70,7 +75,7 @@ const TextPrompt: FunctionComponent<TextPromptProps> = ({
     }
   })
 
-  return (
+  return isAborted ? null : (
```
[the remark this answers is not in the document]
matteodepalo: We could simplify this with a component like `Abortable` 👍

packages/cli-kit/src/public/node/ui.tsx
```diff
@@ -48,10 +48,15 @@ export async function renderConcurrent({renderOptions, ...props}: RenderConcurre
...
-      ...renderOptions,
-      exitOnCtrlC: typeof props.onInput === 'undefined',
-    })
+    try {
+      await render(<ConcurrentOutput {...newProps} />, {
+        ...renderOptions,
+        exitOnCtrlC: typeof props.onInput === 'undefined',
+      })
+    } catch (error) {
+      newProps.abortController.abort()
```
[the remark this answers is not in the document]
matteodepalo: On a second pass, the only reason I added this is because the parent can also be `renderConcurrent` itself if abortController is not passed form the outside. I think we should make `abortController` optional in renderConcurrent and remove this.

matteodepalo:
I'm going to merge this as I've addressed all the comments from @pepicrft.

PR #1853, opened 2023-04-26
[cli-kit] ciPlatform add metadata

vincentezw:

We are looking to run the cli in an CI-environment and want it to be able to collect relevant metadata when executed from a workflow.

The `ciPlatform` function can indicate if the CLI is running in a CI environment such as Github, Bitbucket etc. This PR adds some metadata to the returned object, based on the appropriate environment variables used by the detected CI environment.

### Measuring impact

### Checklist


packages/cli-kit/src/public/node/context/local.ts
```diff
@@ -201,10 +201,14 @@ export async function hasGit(): Promise<boolean> {
...
  * @returns The CI platform info.
  */
-export function ciPlatform(env = process.env): {isCI: true; name: string} | {isCI: false; name?: undefined} {
+export function ciPlatform(
+  env = process.env,
+): {isCI: true; name: string; metadata: Metadata} | {isCI: false; name?: undefined; metadata?: undefined} {
   if (isTruthy(env.CI)) {
     let name = 'unknown'
-    if (isTruthy(env.CIRCLECI)) {
+    if (isTruthy(env.BITBUCKET_BUILD_NUMBER)) {
```
matteodepalo, line 209: I think this is going to fail for bitbucket unless you use `isSet`. Could you add a test to check that this behavior works with some expected bitbucket env vars?

vincentezw: good catch, you are right. I've modified this and added a test with Bitbucket environment variables set.

matteodepalo (approved):

PR #1854, opened 2023-04-26
Fix deploy app that contais only functions as extensions

alvaro-shopify:
When you `deploy` an app that only contains functions this error is returned

image

- Run the extensions build logic in the app contains functions

### How to test your changes?
- Run the deploy command with an app that contains only functions

### Measuring impact

### Checklist


packages/app/src/cli/services/deploy/bundle.ts
```diff
@@ -57,23 +64,31 @@ export async function bundleAndBuildExtensions(options: BundleOptions) {
             },
           },
         })),
-        ...options.app.extensions.function.map((functionExtension) => {
```
matteodepalo, line 60: Could we keep this here and instead do the check `if (!bundleTheme && !bundleUI) return` at the end? So something like

```
if (!bundleTheme && !bundleUI) {
  return
} else {
  const bundlePath = joinPath(options.bundleRootPath, `bundle.zip`)
    await zip({
      inputDirectory: bundleDirectory,
      outputZipPath: bundlePath,
    })
    return bundlePath
}

alvaro-shopify: done!

PR #1856, opened 2023-04-26
Add eslint rule to warn users with certain prompt messages

matteodepalo:
We had been discussing with @MeredithCastile the option of encouraging people to not use words like "choose" or "select" in prompt messages as it can conflict with the wording used in the footer of the prompt. Also these words are simply redundant.

This PR introduces a rule that will simply warn users if they include these words.


Arkham (approved):
LGTM. Can we also add another rule that checks that oclif command descriptions are capitalized and end with `.`?

PR #1860, opened 2023-04-27
Text prompt displays an specific empty value when no input is entered

alvaro-shopify:
When you deploy an app using the new unified deployment system you could add a label value to the deployment. This input is optional so if the user does not enter a value the text prompt displays nothing which looks a bit weird

image

In that case we would want to add the option of displaying en empty message

- Add an new optional parameter to the `TextPrompt` component with a default value `(empty)`. In case you don't want to display an empty message you could set it to empty in the call
- In case the user submit an empty value the empty message will be displayed with a dimmed style

label_empty

- The value submitted should be an empty string regardless the empty value is displayed

### How to test your changes?
- Activate the `unified deployment beta flag` in your app
- Run the `deploy` command and leave empty the `Deployment label` prompt

### Measuring impact

### Checklist

### DIFF
--- a/packages/cli-kit/src/private/node/ui/components/TextPrompt.tsx
+++ b/packages/cli-kit/src/private/node/ui/components/TextPrompt.tsx
@@ -14,6 +14,7 @@ export interface TextPromptProps {
   password?: boolean
   validate?: (value: string) => string | undefined
   allowEmpty?: boolean
+  emptyDisplayedValue?: string
 }
 
 const TextPrompt: FunctionComponent<TextPromptProps> = ({
@@ -23,6 +24,7 @@ const TextPrompt: FunctionComponent<TextPromptProps> = ({
   defaultValue = '',
   password = false,
   allowEmpty = false,
+  emptyDisplayedValue = '(empty)',
 }) => {
   if (password && defaultValue) {
     throw new Error("Can't use defaultValue with password")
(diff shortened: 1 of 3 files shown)


packages/cli-kit/src/private/node/ui/components/TextPrompt.tsx
```diff
@@ -44,6 +46,8 @@ const TextPrompt: FunctionComponent<TextPromptProps> = ({
   const {oneThird} = useLayout()
   const [answer, setAnswer] = useState<string>('')
   const answerOrDefault = answer.length > 0 ? answer : defaultValue
+  const useEmptyValue = answerOrDefault === ''
```
matteodepalo, line 49: Could we change this to something like `displayEmptyValue`? Usually things that begin with `use` in react are hooks and it'd be nice to keep that convention.

packages/cli-kit/src/private/node/ui/components/TextPrompt.tsx
```diff
@@ -44,6 +46,8 @@ const TextPrompt: FunctionComponent<TextPromptProps> = ({
...
+  const answerWithEmptyOrDefault = useEmptyValue ? emptyDisplayedValue : answerOrDefault
```
matteodepalo, line 50: Maybe we could rename this to just `displayedAnswer`?

matteodepalo (approved):
Just a few naming questions but nice one, great test!

PR #1861, opened 2023-04-27
Bump @types/react version in resolutions

matteodepalo:

`resolutions` version wasn't in line with what was installed.

I've bumped the resolved verison of @types/react.

### DIFF
--- a/package.json
+++ b/package.json
@@ -92,7 +92,7 @@
   },
   "version": "0.0.0",
   "resolutions": {
-    "@types/react": "16.14.0",
+    "@types/react": "17.0.2",
     "vite": "2.9.12",
     "@oclif/core": "2.1.4"
   }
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -92,7 +92,7 @@
   },
   "version": "0.0.0",
   "resolutions": {
-    "@types/react": "16.14.0",
+    "@types/react": "17.0.2",
```
gonzaloriestra, line 95: Do you know why we need this one? If cli-kit is overriding this value, which package needs it? Couldn't we move it to where it's needed?

gonzaloriestra (approved):

package.json
matteodepalo: No luck, I think the tree of dependencies doesn't manage to pick up the specific version in the packages 🤷

PR #1900, opened 2023-05-03
Remove references to rack, which we don't need to use

amcaplan:
### WHY are these changes introduced?

Fixes https://github.com/Shopify/cli/issues/1893

Theme development is currently dead in the water, due to introducing a constant reference we don't actually have access to during `theme dev`.

### WHAT is this pull request doing?

Removes the references to `Rack`, using the standard Rack `[status, headers, body]` interface instead.

### How to test your changes?

`theme dev` should work now

### Measuring impact

How do we know this change was effective? Please choose one:

- [x] n/a - this doesn't need measurement, e.g. a linting rule or a bug-fix
- [ ] Existing analytics will cater for this addition
- [ ] PR includes analytics changes to measure impact

### Checklist

- [x] I've considered possible cross-platform impacts (Mac, Linux, Windows)
- [x] I've considered possible [documentation](https://shopify.dev) changes
- [x] I've made sure that any changes to `dev` or `deploy` have been reflected in the [internal flowchart](https://www.figma.com/file/7vqUp50u6dm48Zfb4JRRn8/CLI3-Internals).

PR #1901, opened 2023-05-03
Remove references to rack, which we don't need to use

amcaplan:
https://github.com/Shopify/cli/pull/1900


karreiro (approved):
Thank you, @amcaplan.

PR #1903, opened 2023-05-03
Remove label prompt

matteodepalo:
Removing label prompt and label flag from deploy command.


ani-shopify:
I 🎩 ed and it looks good.

Arkham (approved):
LGTM!

PR #1909, opened 2023-05-03
Introduce `shopify theme update init` and `shopify theme update check` commands

karreiro:

Fixes https://github.com/Shopify/shopify/issues/420327 part (1/2).

#### New commands

This PR introduces 2 new commands:
- `shopify theme update init` that initializes an `update_extension.json` script
- `shopify theme update check` that checks if an `update_extension.json` script is valid

#### Schema validation

In order to power the `shopify theme update check` validation features, this PR relies on the `zod` schema validation library, which is currently used in `packages/cli-kit` and other CLI packages.

However, as discussed in the [thread](https://github.com/Shopify/cli/pull/1909#discussion_r1184700181) below, `zod` doesn't support loading a schema from a JSON schema file. Instead, `zod` requires its own domain-specific language (DSL) to declare schemas.

To use `zod` and benefit from its type inference, avoid adding a new library that duplicates `zod`'s functionality, and maintain a single schema declaration at [Shopify/theme-liquid-docs](https://github.com/Shopify/theme-liquid-docs), this PR introduces a new command in `packages/theme` called `pnpm run schema:generate`. This command automatically generates the `zod` schema based on the [Shopify/theme-liquid-docs](https://github.com/Shopify/theme-liquid-docs).

### How to test your changes?

- Run `shopify theme -h` and notice the new `shopify theme update` topic appears
- Run `shopify theme update -h` and notice the new commands appear
- Run `shopify theme update init` and notice it creates an `update_extension.json` script
- Run `shopify theme update check` and notice it checks if `update_extension.json` script is valid

### Post-release steps

Update documentation.

### Measuring impact

### Checklist


packages/cli-kit/package.json
```diff
@@ -100,6 +100,7 @@
     "@oclif/core": "2.1.4",
     "@types/archiver": "5.3.2",
     "abort-controller": "3.0.0",
+    "ajv": "8.12.0",
```
karreiro, line 103: @gonzaloriestra I've tried to follow the history in the repository, and I think you've added the `zod` library, so I'm assuming you have the context, still, please, feel free to forward this message if you don't.

I'm adding `ajv` because it handles JSON schemas by loading them, while `zod` supports its own DSL. 

In themes, we really want to load the schemas from the JSON schema file because we're working on introducing schemas in the [shopify/theme-liquid-docs](https://github.com/shopify/theme-liquid-docs) in the middle-term. That's powerful because we keep [shopify/theme-liquid-docs](https://github.com/shopify/theme-liquid-docs) auto synced with some internal modules, so developers can get the most updated assets without a new tooling release (we're already doing that for Intelligent Code Completion).

`zod` does have an extension to load JSON schema files, however it doesn't support [data types definitions](https://github.com/Shopify/cli/pull/1909/files#diff-4a78aede52f87ba37d16049f0ca3160c3cd9b0178a61d1d68cafd59f29f49efeR44-R53). With that in mind, I feel that `ajv` is the way to go.

Do you have any questions/comments/concerns about it?

Thank you for the early review :)

gonzaloriestra: Thanks for the detailed explanation! I'm fine with adding that library if zod doesn't cover the functionality you need. 

My only concern is that it looks we will basically have two libraries for the same thing. So it would be great to be able to use only ajv for everything (which it also seems to be faster than zod, btw). The problem is that we are exporting zod as it is from cli-kit and using it extensively, so it will require some work and also breaking changes in cli-kit...

What do you think @Shopify/cli-foundations?

alvaro-shopify: In case we need to add the new library, I would retain both of them during a period of time. We could reimplement public libraries that uses `zod` with the new library. After some time we could deprecate the `zod` based APIs
IMHO I would try to avoid adding breaking changes without giving sometime to user to adopt the new APIs

isaacroldan: i'd avoid adding a new library if possible. zod has been great so far and replacing it would be a lot of work.
I just tested a "Json schema to zod" library, it does seems to work properly, what issues did you find @karreiro?
Captura de pantalla 2023-05-08 a las 13 16 04

amcaplan: FWIW you can also convert zod to JSON schema with https://www.npmjs.com/package/zod-to-json-schema

I'm using it in the demos and it works great!

The big downside is that zod is fairly legible and more TypeScript-like compared with JSON schema which is pretty wordy and doesn't make it as easy to do code completion. I think zod also supports more complex types.

packages/theme/src/cli/commands/theme/update/run.ts
```diff
@@ -0,0 +1,16 @@
...
+  static description = `Run an 'update_extension.json' script in a theme.`
+
+  static hidden = true
+
+  static flags = {
+    ...globalFlags,
+  }
+
+  async run(): Promise<void> {
+    // empty.
```
karreiro, line 14: I'm going to introduce this command in a separate PR to keep this one from getting too long.

gonzaloriestra (approved):
LGTM. Sorry for the super-late review!

I just have a concern, which is the name of the command. The problem is that `update` is also verb, so it feels weird to me to have a command with `update init`... It may look as it's going to update something (a init/check configuration?).

Have you considered `theme update-script init` or `theme script init`?

cc: @MeredithCastile

packages/theme/scripts/schema/generate.js
```diff
@@ -0,0 +1,69 @@
...
+
+function remoteSchemaPath(schemaVersion) {
+  const repository = 'Shopify/theme-liquid-docs'
+  const schemaPath = `schemas/update/update_extension_schema_v${schemaVersion}.json`
+
+  return `https://raw.githubusercontent.com/${repository}/main/${schemaPath}`
+}
+
+function localSchemaPath(schemaVersion) {
+  return resolve(`./src/cli/services/update/schemas/update_extension_schema_v${1}.ts`)
```
mgmanzella, line 14: this might be just my gap in TS, but why pass `schemaVersion` and then use `1` to format a var in a string?

packages/theme/src/cli/services/update/check.test.ts
```diff
@@ -0,0 +1,478 @@
...
+      // When
+      // eslint-disable-next-line no-await-in-loop
+      const errors = await checkErrors(script)
+
+      // Then
+      expect(errors).toEqual([`The "operations/0/operation_x/0" doesn't have the expected attributes`])
+    }
+  })
+
+  test(`when script has additional propertiess`, async () => {
```
mgmanzella, line 334: possible typo
```suggestion
  test(`when script has additional properties`, async () => {
```

packages/theme/src/cli/services/update/check.test.ts
```diff
@@ -0,0 +1,478 @@
...
+
+    // When
+
+    const errors = await checkErrors(script)
+
+    // Then
+    expect(errors).toEqual([`The "extra_property" is not a permitted key`])
+  })
+
+  test(`when a step has additional propertiess`, async () => {
```
mgmanzella, line 354: possible typo
```suggestion
  test(`when a step has additional properties`, async () => {
```

packages/theme/src/cli/services/update/init.ts
```diff
@@ -0,0 +1,47 @@
...
+import {joinPath} from '@shopify/cli-kit/node/path'
+import {renderSuccess} from '@shopify/cli-kit/node/ui'
+
+export async function init(rawPath?: string) {
+  const path = joinPath(rawPath ?? '.', 'update_extension.json')
+
+  await createScript(path)
+
+  renderSuccess({
+    body: [`The '${path}' script has been created.`],
```
mgmanzella, line 13: /nit is this the full path or relative path? relative path may be a better experience

packages/theme/src/cli/services/update/check.ts
```diff
@@ -0,0 +1,72 @@
...
+
+  try {
+    scriptFileObj = JSON.parse(scriptFile.toString())
+  } catch (error) {
+    if (error instanceof SyntaxError) {
+      throw new AbortError(error.message)
+    }
+  }
+
+  return schemaV1.safeParse(scriptFileObj) as ValidationResult
```
mgmanzella, line 71: not relevant to this PR, im curious in the future what this looks like when validating multiple schemas? on one hand i dont think we'll have many schemas, on the other i could see this getting very verbose (eg a giant || statement 😅 )

mgmanzella (approved):
niiiice 🔥 have a few non-blocking comments, thanks @karreiro!

packages/theme/scripts/schema/generate.js
karreiro: Oops! Great catch! Thank you -- it's fixed now :)

packages/theme/src/cli/services/update/check.test.ts
karreiro: Thank you! It's fixed now :)

packages/theme/src/cli/services/update/check.test.ts
karreiro: Thank you! It's fixed now :)

packages/theme/src/cli/services/update/schemas/update_extension_schema_v1.ts
```diff
@@ -0,0 +1,123 @@
...
+
+export const schemaUrlV1 =
+  'https://raw.githubusercontent.com/Shopify/theme-liquid-docs/main/schemas/update/update_extension_schema_v1.json'
+
+export const schemaV1 = z
+  .object({
+    $schema: z.string(),
+    theme_name: z.string(),
+    theme_version: z.string(),
+    operations: z
```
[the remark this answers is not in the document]
karreiro: 🤣

packages/theme/oclif.manifest.json
```diff
@@ -1249,6 +1249,106 @@
...
+          "name": "path",
+          "type": "option",
+          "description": "The path to your theme directory.",
+          "multiple": false,
+          "default": "."
+        }
+      },
+      "args": {}
+    },
+    "theme:update:run": {
```
bahar-p, line 1319: I think this is not supposed to be in this PR?

packages/theme/oclif.manifest.json
```diff
@@ -1249,6 +1249,106 @@
...
+      "id": "theme:update:run",
+      "description": "Run an 'update_extension.json' script in a theme.",
```
bahar-p, line 1321:
```suggestion
      "description": "Run the theme update process",
```

packages/theme/src/cli/commands/theme/update/run.ts
bahar-p, line 1: Could you exclude this from this PR and add it separately?

bahar-p (approved):
Thanks LGTM! A few comments only. I'd love to know what are the steps to tophat CLI code changes like this locally (spin? or dev?)

packages/theme/src/cli/commands/theme/update/init.ts
```diff
@@ -0,0 +1,18 @@
...
+  static description = `Initialize an 'update_extension.json' script.`
+
+  static flags = {
+    ...globalFlags,
+    path: themeFlags.path,
+  }
+
+  async run(): Promise<void> {
+    const {flags} = await this.parse(UpdateInit)
+    await init(flags.path)
```
bahar-p, line 16: Should we return the newly created file path as a result of `init`? I noticed we don't. I assumed it's be in the root but i wasn't sure until i checked. What's the best practice?

packages/theme/oclif.manifest.json
karreiro: Yes, originally I added the `run` command as a hidden/empty boilerplate to be implemented [here](https://github.com/Shopify/cli/pull/1946). However, following this suggestion, I've removed for clarity :)

packages/theme/src/cli/commands/theme/update/run.ts
karreiro: I've removed it following [this thread](https://github.com/Shopify/cli/pull/1909#discussion_r1205316334).

packages/theme/src/cli/commands/theme/update/init.ts
karreiro: Yeah, we're actually returning the relative path of the file, as we may notice here:

Screenshot 2023-05-25 at 12 18 10

Initially, we were showing the absolute path, but after taking another look at the context, we realized that using the relative path is actually the best practice in this situation. You can read more about it in this [here](https://github.com/Shopify/cli/pull/1909#discussion_r1195737066) :)

PR #1911, opened 2023-05-03
POC: Command to create a shop

pobed2:


packages/cli/oclif.manifest.json
```diff
@@ -43,6 +43,17 @@
       "flags": {},
       "args": {}
     },
+    "auth:shop:create": {
```
pobed2, line 46: Auth is obviously not the right place for this. I just used it for the POC for simplicity.

PR #1913, opened 2023-05-04
Install pos_ui_extension node dependencies as exact versions (main)

isaacroldan:
Force POS extension dependencies to be installed as exact versions instead of with a caret `^`

### Measuring impact

### Checklist

### DIFF
--- a/packages/app/src/cli/services/generate/extension.ts
+++ b/packages/app/src/cli/services/generate/extension.ts
@@ -117,12 +117,11 @@ async function uiExtensionInit({
       task: async () => {
         await addResolutionOrOverrideIfNeeded(app.directory, extensionFlavor)
         const requiredDependencies = getExtensionRuntimeDependencies({specification, extensionFlavor})
-        const exact = specification.identifier === 'pos_ui_extension'
         await addNPMDependenciesIfNeeded(requiredDependencies, {
           packageManager: app.packageManager,
           type: 'prod',
           directory: app.directory,
-          exact,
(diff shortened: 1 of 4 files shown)


packages/app/src/cli/services/generate/extension.ts
```diff
@@ -121,6 +121,9 @@ async function uiExtensionInit({
           packageManager: app.packageManager,
           type: 'prod',
           directory: app.directory,
+          // This is a temporary workaround for POS extensions. By deafult all dependencies have the `^` prefix.
```
amcaplan, line 124:
```suggestion
          // This is a temporary workaround for POS extensions. By default all dependencies have the `^` prefix.
```

packages/cli-kit/src/public/node/node-package-manager.ts
```diff
@@ -444,7 +453,7 @@ export async function addNPMDependenciesWithoutVersionIfNeeded(
  * @param type - The dependency type.
  * @returns An array with the arguments.
  */
-function argumentsToAddDependenciesWithNPM(dependencies: string[], type: DependencyType): string[] {
+function argumentsToAddDependenciesWithNPM(dependencies: string[], type: DependencyType, exact = false): string[] {
```
amcaplan, line 456: Having a final boolean parameter is asking for trouble later on. Maybe this should be `{exact} = {exact: false}`?

Otherwise we break the interface if we need to add another param.

Currently it's not exported, so not the end of the world... but it could easily happen without us paying too much attention. Especially in a `public` file.

PR #1928, opened 2023-05-09
Deploy functions with unified app deployments when beta is enabled

jacobsteves:
Functions are being updated to be deployed through unified app deployments. This PR makes that possible as long as the beta flag is set.

When the beta is enabled, we deploy functions through the `CreateDeployment` mutation. Otherwise, deploy functions in the old way.

### How to test your changes?

You'll need my `js.deploy-functions-on-ef` Partners branch and `js.add-module-id-to-upload-details` Script Service branch.

Two notes:
1. There is currently no way to migrate a function outside of unified app deployments to a function on unified app deployments, but that change will come later. This means if you want to switch between the two, you'll have to remove your `.env` file.
2. Functions deployed through unified app deployments wont actually run on the server yet. We still have to hook that up, but that will not change this work. 

Testing:
1. enable the unified app deployments beta on your Partners app
2. create a function (if you haven't already)
3. deploy, make sure it succeeds

### Checklist

- [X] I've considered possible cross-platform impacts (Mac, Linux, Windows) - No impact.
- [X] I've considered possible [documentation](https://shopify.dev) changes - None needed yet.
- [ ] I've made sure that any changes to `dev` or `deploy` have been reflected in the [internal flowchart](https://www.figma.com/file/7vqUp50u6dm48Zfb4JRRn8/CLI3-Internals) - Not yet, but I can


packages/app/src/cli/models/extensions/functions.test.ts
```diff
@@ -0,0 +1,39 @@
...
+})
+
+describe('graphQLType', () => {
+  test('returns type when not using extensions framework', async () => {
+    // When
+    FUNCTION_A.usingExtensionsFramework = false
+    const got = FUNCTION_A.graphQLType
+
+    // Then
+    expect(got).toEqual('PRODUCT_DISCOUNTS')
```
DuncanUszkay1, line 28: super nit: does fetching the actual type belong more in `then` rather than `when`?

packages/app/src/cli/models/extensions/functions.ts
```diff
@@ -44,6 +44,7 @@ export class FunctionInstance<TConfiguration extends FunctionConfigType = Functi
   entrySourceFilePath?: string
   configuration: TConfiguration
   configurationPath: string
+  _usingExtensionsFramework: boolean
```
DuncanUszkay1, line 47: Why the underscore? 🤔 Is this some js/ts idiom I'm missing?

DuncanUszkay1: TIL!

isaacroldan: You can use `private usingExtensionsFramework: boolean`
It's easier to understand :)

matteodepalo:
I was able to deploy successfully, but I cannot see the function in the deployment in the dashboard. I assume there is some work required there? Not that it would prevent us from merging this PR, I just want to double check.

Screenshot 2023-05-10 at 14 23 28

matteodepalo (approved):
Great work on this one! @jacobsteves 

@isaacroldan can I get your eyes on this as well before we merge?

isaacroldan (approved):

PR #1937, opened 2023-05-12
Simplify extensions

isaacroldan:

Fixes #0000 

### Measuring impact

### Checklist


packages/app/src/cli/models/app/identifiers.ts
```diff
@@ -91,9 +91,7 @@ export function getAppIdentifiers(
       extensionsIdentifiers[extension.localIdentifier] = envVariables[extension.idEnvironmentVariableName]!
     }
   }
-  app.extensions.ui.forEach(processExtension)
-  app.extensions.function.forEach(processExtension)
-  app.extensions.theme.forEach(processExtension)
+  app.allExtensions.forEach(processExtension)
```
shauns, line 94: We're losing the ordering of tackling ui, then function, then theme. Any downside to that?

packages/app/src/cli/models/app/loader.ts
```diff
@@ -468,9 +400,7 @@ async function logMetadataForLoadedApp(
   app: App,
   loadingStrategy: {
     usedCustomLayoutForWeb: boolean
-    usedCustomLayoutForUIExtensions: boolean
-    usedCustomLayoutForFunctionExtensions: boolean
-    usedCustomLayoutForThemeExtensions: boolean
+    usedCustomLayoutForExtensions: boolean
   },
 ) {
   await metadata.addPublicMetadata(async () => {
```
shauns, line 406: Do you think there's any additional metadata it might be worth including, in light of this new approach?

packages/app/src/cli/models/extensions/specification.ts
```diff
@@ -175,6 +202,25 @@ export class UIExtensionInstance<TConfiguration extends BaseConfigContents = Bas
   hasExtensionPointTarget(target: string): boolean {
     return this.specification.hasExtensionPointTarget?.(this.configuration, target) || false
   }
+
+  // Functions specific properties
+  get buildCommand() {
+    const config = this.configuration as unknown as FunctionSchemaContents
```
shauns, line 208: If `this.configuration` isn't a Function-style one, then presumably at runtime this method (and the others) would fail? (and in a similar fashion, the UI specific properties above)

Is there more we can do to make using an `ExtensionInstance` as safe as possible?

packages/app/src/cli/models/app/identifiers.ts
isaacroldan: nop, there is no reason to have a specific order here

packages/app/src/cli/models/app/loader.ts
isaacroldan: Not that I kind think off, but we are still tracking the 3 categories, we should think of a way to track any kind of extension 🤔 Can we track extension types in an array? what would be the best format for that?

packages/app/src/cli/models/app/app.test-data.ts
```diff
@@ -54,38 +50,46 @@ export async function testUIExtension(uiExtension: Partial<UIExtension> = {}): P
...
+export async function testThemeExtensions(): Promise<ExtensionInstance & ThemeExtension> {
   const configuration = {
     name: 'theme extension name',
     type: 'theme' as const,
+    metafields: [],
   }
 
-  return new ThemeExtensionInstance({
+  const allSpecs = await loadLocalExtensionsSpecifications()
+  const specification = allSpecs.find((spec) => spec.identifier === 'theme') as ExtensionSpecification
```
Arkham, line 78: same here

packages/app/src/cli/models/app/app.test-data.ts
```diff
@@ -95,23 +99,33 @@ function defaultFunctionConfiguration(): FunctionConfigType {
...
-export async function testFunctionExtension(opts: TestFunctionExtensionOptions = {}): Promise<FunctionExtension> {
+export async function testFunctionExtension(
+  opts: TestFunctionExtensionOptions = {},
+): Promise<ExtensionInstance & FunctionExtension> {
   const directory = opts.dir ?? '/tmp/project/extensions/my-function'
   const configuration = opts.config ?? defaultFunctionConfiguration()
 
-  return new FunctionInstance({
+  const allSpecs = await loadLocalExtensionsSpecifications()
+  const specification = allSpecs.find((spec) => spec.identifier === 'function') as ExtensionSpecification
```
Arkham, line 119: and here :)

packages/app/src/cli/models/app/app.test-data.ts
```diff
@@ -54,38 +50,46 @@ export async function testUIExtension(uiExtension: Partial<UIExtension> = {}): P
     },
   }
   const configurationPath = uiExtension?.configurationPath ?? `${directory}/shopify.ui.extension.toml`
-  const entrySourceFilePath = uiExtension?.entrySourceFilePath ?? `${directory}/src/index.js`
+  const entryPath = uiExtension?.entrySourceFilePath ?? `${directory}/src/index.js`
 
   const allSpecs = await loadLocalExtensionsSpecifications()
-  const specification = allSpecs.find((spec) => spec.identifier === configuration.type) as UIExtensionSpec
+  const specification = allSpecs.find((spec) => spec.identifier === configuration.type) as ExtensionSpecification
```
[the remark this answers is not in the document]
isaacroldan: You're right, no need for this anymore

packages/app/src/cli/models/app/app.ts
```diff
@@ -95,13 +98,17 @@ export class App implements AppInterface {
...
+    extensions: ExtensionInstance[],
     usesWorkspaces: boolean,
     dotenv?: DotEnvFile,
     errors?: AppErrors,
   ) {
+    // Temporary workaround while we migrate to use appModule features.
+    const functionsExt = extensions.filter((extension) => extension.features.includes('function'))
+    const themes = extensions.filter((extension) => extension.features.includes('theme'))
+    const uis = extensions.filter(
+      (extension) => !extension.features.includes('function') && !extension.features.includes('theme'),
```
Arkham, line 110: these lines make me think that maybe it'd be nice to have a `ui` feature, so that in the future we can add new extension types which are not ui, nor theme, nor function.

isaacroldan: Oh, there is a `ui` feature, but this is just a workaround to dev/deploy keep working.
for instance web_pixel doesn't have the ui feature, but until the refactor is complete we need to consider it a ui extension still

isaacroldan: App won't expose an object with function/theme/ui in the future

packages/app/src/cli/models/extensions/schemas.ts
```diff
@@ -32,26 +32,27 @@ export const ApiVersionSchema = zod.string()
...
+export const ThemeExtensionSchema = BaseSchema.extend({
   name: zod.string(),
   type: zod.literal('theme'),
 })
 
-export const BaseFunctionConfigurationSchema = zod.object({
-  name: zod.string(),
-  type: zod.string(),
-  description: zod.string().optional().default(''),
+export const BaseFunctionConfigurationSchema = BaseSchema.extend({
```
Arkham, line 55: nitpick, why these names?
- BaseUIExtensionSchema -> UIExtensionSchema
- BaseFunctionConfigurationSchema -> FunctionSchema

or am I missing something?

isaacroldan: nop, you are right, I messed the names here during the refactor

isaacroldan: Is because Functions used to have two schemas: one for configuration and one for metadata (that no longer exists)

packages/app/src/cli/models/extensions/specification.ts
```diff
@@ -104,22 +117,35 @@ export class UIExtensionInstance<TConfiguration extends BaseConfigContents = Bas
...
-    this.outputBundlePath = joinPath(options.directory, 'dist/main.js')
     this.devUUID = `dev-${randomUUID()}`
     this.localIdentifier = basename(options.directory)
     this.idEnvironmentVariableName = `SHOPIFY_${constantize(basename(this.directory))}_ID`
+    this._usingExtensionsFramework = false
+    if (this.specification.identifier === 'theme') {
+      this.outputBundlePath = this.directory
+    } else {
+      this.outputBundlePath = joinPath(this.directory, 'dist/main.js')
+    }
```
Arkham, line 148: maybe we could still have helper functions that run this logic for you? I feel that now this file has ALL the logic from different extension types

packages/app/src/cli/models/extensions/specification.ts
```diff
@@ -219,13 +266,14 @@ export interface CreateExtensionSpecType<TConfiguration extends BaseConfigConten
...
-export function createUIExtensionSpecification<TConfiguration extends BaseConfigContents = BaseConfigContents>(
+export function createExtensionSpecification<TConfiguration extends BaseConfigContents = BaseConfigContents>(
   spec: CreateExtensionSpecType<TConfiguration>,
-): UIExtensionSpec<TConfiguration> {
+): ExtensionSpecification<TConfiguration> {
   const defaults = {
     // these two fields are going to be overridden by the extension specification API response,
     // but we need them to have a default value for tests
     externalIdentifier: `${spec.identifier}_external`,
+    additionalIdentifiers: [],
```
Arkham, line 276: it's not immediately clear that additionalIdentifiers contains the names of all variants of the extension. maybe we could call it `variantIdentifiers`? I dunno!

packages/app/src/cli/models/extensions/specifications/ui_extension.test.ts
```diff
@@ -16,7 +16,7 @@ describe('ui_extension', async () => {
   async function getTestUIExtension({directory, extensionPoints}: GetUIExtensionProps) {
     const configurationPath = joinPath(directory, configurationFileNames.extension.ui)
     const allSpecs = await loadLocalExtensionsSpecifications()
-    const specification = allSpecs.find((spec) => spec.identifier === 'ui_extension') as UIExtensionSpec
+    const specification = allSpecs.find((spec) => spec.identifier === 'ui_extension') as ExtensionSpecification
```
Arkham, line 19: another rogue `as` casting here

packages/app/src/cli/private/plugins/extension.ts
```diff
@@ -1,18 +1,11 @@
-import {UIExtensionSpec} from '../../models/extensions/ui.js'
-import {FunctionSpec} from '../../models/extensions/functions.js'
+import {ExtensionSpecification} from '../../models/extensions/specification.js'
 import {HookReturnPerExtensionPlugin} from '../../public/plugins/extension.js'
 import {Config} from '@oclif/core'
 import {getArrayRejectingUndefined} from '@shopify/cli-kit/common/array'
 import {fanoutHooks} from '@shopify/cli-kit/node/plugins'
 
-export async function loadUIExtensionSpecificiationsFromPlugins(config: Config): Promise<UIExtensionSpec[]> {
+export async function loadUIExtensionSpecificiationsFromPlugins(config: Config): Promise<ExtensionSpecification[]> {
```
Arkham, line 7: mm is this still loading only UIExtensions or all extensions? also there is a little typo in the name "specificiations" -> "specifications"

packages/app/src/cli/models/extensions/specification.ts
isaacroldan: I added this just for functions, because we will be migrating from multiple identifiers (order_discount, product...) to a single identifier: `function` and we need to be compatible with both cases, but there is a single specification for all

Arkham:
really good change, I'm happy with it, left a bunch of minor comments, LGTM

packages/app/src/cli/models/app/loader.ts
```diff
@@ -293,132 +283,74 @@ class AppLoader {
...
       const directory = dirname(configurationPath)
       const specification = await findSpecificationForConfig(this.specifications, configurationPath, this.abortOrReport)
 
-      if (!specification) {
-        return
-      }
+      if (!specification) return
 
       const configuration = await this.parseConfigurationFile(specification.schema, configurationPath)
+      const entryPath = await this.findEntryPath(directory, specification)
```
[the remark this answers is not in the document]
isaacroldan: nop, they don't need it

packages/app/src/cli/models/extensions/specification.ts
isaacroldan: Yes, there is some logic that I added here that's specific to some extensions, I plan to move this out and simplify stuff. This case is interesting, because we shouldn't need to set the outputBundlePath here, I'll improve this in future PRs 🙏

packages/app/src/cli/models/extensions/specification.ts
```diff
@@ -54,30 +66,31 @@ export interface UIExtensionSpec<TConfiguration extends BaseConfigContents = Bas
...
   entrySourceFilePath: string
-  outputBundlePath: string
   devUUID: string
   localIdentifier: string
   idEnvironmentVariableName: string
   directory: string
   configuration: TConfiguration
   configurationPath: string
+  outputBundlePath: string
+  _usingExtensionsFramework: boolean
```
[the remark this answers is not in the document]
isaacroldan: Is an old convention to define private properties. I didn't introduce this, i'll change it

matteodepalo (approved):

PR #1939, opened 2023-05-15
Use theme bundling by default

matteodepalo:
Fixes https://github.com/Shopify/app-deploys/issues/318

We want to make theme bundling opt-out instead of opt-in now that we're more confident in the feature.

## Testing

- Create a new app with `./bin/create-test-app.js -e theme`
- Try deploying, it should work
- Try introducing an error in the theme
- Try deploying, the error should be visible in the error messages banner


isaacroldan (approved):
🎩 and works as expected

PR #1946, opened 2023-05-17
[Do not merge] Introduce the `shopify theme update run` command

karreiro:
⚠️ Please, do not merge until https://github.com/Shopify/cli/pull/1909 gets merged.

Fixes [#420327](https://github.com/Shopify/shopify/issues/420327) part (2/2).

This PR introduces the `shopify theme update run` command following the contracts defined by [#420327](https://github.com/Shopify/shopify/issues/420327) and [#420326](https://github.com/Shopify/shopify/issues/420326).

### How to test your changes?

- Run `shopify theme update -h` and notice the new commands appear
- Run `shopify theme update run` and notice it checks your `update_extension.json` script, creates a theme [1], and keeps checking it until the the theme is no longer processing

[1] This step will be revisited, in the future, when we have [#420326](https://github.com/Shopify/shopify/issues/420326) concluded.

demo_run

### Post-release steps

Update documentation.

### Measuring impact

### Checklist


packages/theme/src/cli/services/update/run.ts
```diff
@@ -0,0 +1,178 @@
...
+async function triggerUpdaterAPI(
+  session: AdminSession,
+  _updateExtension: string,
+  source: number,
+  target: number,
+): Promise<Theme> {
+  const name = `updated-theme-${source}-${target}`
+  const role = 'unpublished'
+  try {
+    const theme = await createTheme({name, role, processing: true}, session)
```
albchu, line 153: Will this be a future change?

albchu (approved):
Everything looking good. Calling createTheme in the updateTheme flow is a little weird but I imagine theres a good reason for it.

packages/theme/src/cli/services/update/run.ts
```diff
@@ -0,0 +1,178 @@
...
+  }
+}
+
+async function isUpdaterIsProgress(ctx: Context, session: AdminSession, startTime: number) {
+  const theme = await fetchTheme(ctx.theme.id, session)
+
+  if (!theme) {
+    throw new AbortError(
+      'The `update_extension.json` script could not be executed due to a runtime issue.',
+      'Please check the `update_extension.json` script and retry.',
```
bahar-p, line 113: This error message should be generic. The update process could fail for many different reasons. Additionally in majority of cases there won't be any update_extension passed in from theme-partners. 
So something a long the lines of "Theme updated failed...". We can probably use the same generic error that we use in OSW on theme index.

packages/theme/src/cli/services/update/run.ts
```diff
@@ -0,0 +1,178 @@
...
+    )
+  }
+
+  const elapsedTime = Date.now() - startTime
+
+  if (theme.processing && elapsedTime >= UPDATER_TIMEOUT) {
+    throw new AbortError(
+      'The `update_extension.json` script could not be executed due to a timeout issue.',
+      'Please check the `update_extension.json` script and retry.',
```
bahar-p, line 122: same comment as above

packages/theme/src/cli/services/update/run.ts
```diff
@@ -0,0 +1,178 @@
...
+
+    return theme
+  } catch (err) {
+    if (!(err instanceof AbortError)) {
+      throw err
+    }
+
+    throw new AbortError(
+      ['The update process could not be triggered.', `Cause: ${err.message}`],
+      'Please check the `update_extension.json` script and retry.',
```
bahar-p, line 167:
```suggestion
```
remove

packages/theme/oclif.manifest.json
```diff
@@ -1315,6 +1315,67 @@
...
+          "char": "s",
+          "description": "Store URL. It can be the store prefix (johns-apparel) or the full myshopify.com URL (johns-apparel.myshopify.com, https://johns-apparel.myshopify.com).",
+          "multiple": false
+        },
+        "script": {
+          "name": "script",
+          "type": "option",
+          "description": "The path to the 'update_extension.json' script.",
+          "multiple": false,
+          "default": "./update_extension.json"
```
bahar-p, line 1361: Could this be problematic if they have an `./update_extension.json` in their directory but it was created for a different theme but they want to test an update for a different theme that doesn't need any `update_extension`. The CLI would be picking up this script to run with any update command incorrectly 🤔

packages/theme/oclif.manifest.json
```diff
@@ -1315,6 +1315,67 @@
...
+          "default": "."
+        },
+        "store": {
+          "name": "store",
+          "type": "option",
+          "char": "s",
+          "description": "Store URL. It can be the store prefix (johns-apparel) or the full myshopify.com URL (johns-apparel.myshopify.com, https://johns-apparel.myshopify.com).",
+          "multiple": false
+        },
+        "script": {
```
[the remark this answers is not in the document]
karreiro: @bahar-p, since the topic of the command is related to updates, what do you think about using `shopify theme update run --extension-script <path>`?

packages/theme/src/cli/services/update/run.ts
karreiro: > The update process could fail for many different reasons
Indeed!

> Additionally in majority of cases there won't be any update_extension passed in from theme-partners.

I'm not so sure about this part.  Since the script is still in development, it may fail due to issues in the work-in-progress extension script.

Also, in this context, there are only three elements the partner can control: the update extension script, the source theme, and the target theme.

Generally, error messages should display the cause of the error and the potential solution to the problem. With this in mind, I believe we could include information about the source and target themes. However, I don't think making the message more generic would be beneficial for our partners. What do you think about that?

packages/theme/src/cli/services/update/run.ts
karreiro: This is being handled [here](https://github.com/Shopify/cli/pull/1946#discussion_r1217700057).

packages/theme/src/cli/services/update/run.ts
karreiro: Yes. It may seem strange to call `createTheme` within the `updateTheme` flow, but we do so because we're actually creating a new-updated theme (we're not modifying an existing resource).

At the moment, this PR is not doing that as it depends on a new API. However, the plan is having that end-to-end before merging the changes.

packages/theme/oclif.manifest.json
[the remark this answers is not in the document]
milakoeva: Coming from theme extensions/theme app extensions background, I would prefer not using `--extension` or `--extension-script` alone here. Either `--script` to be short, or `--update-extension-script` to be specific work better to me, depending on the goal here.

mgmanzella (approved):
Works with the updater API, should get some more code reviews for one last look

karreiro:
✅ Thank you, @mgmanzella @lucyxiang! LGTM and works fine as well 🚀

PR #1951, opened 2023-05-17
Set current CLI3 version as the version used by the cli-ruby API requests

alvaro-shopify:

Fixes partially https://github.com/Shopify/internal-cli-foundations/issues/633

- Set a new env variable `SHOPIFY_CLI_VERSION` when a cli-ruby command is executed
- cli-ruby api requests use the SHOPIFY_CLI_VERSION version to set the proper cli version header when it makes API request

### How to test your changes?
- Create a new partners instance from this branch `spin up partners -c partners.branch=filter-request-from-old-cli-version`
- Using an app with a TAE, run the CLI3  `dev` command against spin `SHOPIFY_SERVICE_ENV=spin SPIN_INSTANCE=$(spin show -o name) yarn shopify app dev` and it should work as usual
- Using an app with a TAE, run the CLI3  `dev` command against production services `yarn shopify app dev` and  it should work as usual
- Run any `theme` command against production services and it should work as usual

### Measuring impact

### Checklist

### DIFF
--- a/packages/cli-kit/assets/cli-ruby/lib/shopify_cli/version.rb
+++ b/packages/cli-kit/assets/cli-ruby/lib/shopify_cli/version.rb
@@ -1,3 +1,3 @@
 module ShopifyCLI
-  VERSION = ENV.fetch("SHOPIFY_CLI_VERSION")
+  VERSION = ENV.fetch("SHOPIFY_CLI_VERSION", "2.34.0")
 end
--- a/packages/cli-kit/assets/cli-ruby/lib/shopify_cli/version.rb
+++ b/packages/cli-kit/assets/cli-ruby/lib/shopify_cli/version.rb
@@ -1,3 +1,3 @@
 module ShopifyCLI
-  VERSION = ENV.fetch("SHOPIFY_CLI_VERSION", "2.34.0")
+  VERSION = ENV.fetch("SHOPIFY_CLI_VERSION")
 end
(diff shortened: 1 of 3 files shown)


packages/cli-kit/assets/cli-ruby/lib/shopify_cli/version.rb
```diff
@@ -1,3 +1,3 @@
 module ShopifyCLI
-  VERSION = "2.34.0"
+  VERSION = ENV.fetch("SHOPIFY_CLI_VERSION", "2.34.0")
```
gonzaloriestra, line 2: Not sure about that fallback value... I prefer it to fail rather than setting a wrong value. And it shouldn't fail.
```suggestion
  VERSION = ENV.fetch("SHOPIFY_CLI_VERSION")
```

gonzaloriestra (approved):

packages/cli-kit/assets/cli-ruby/lib/shopify_cli/version.rb
alvaro-shopify: but the CLI3 request is not failing in case the CLI_KIT_VERSION. is wrong or missing, should we implement the same strategy here right?

gonzaloriestra: CLI3 does fail when CLI_KIT_VERSION is missing. If it's wrong, I guess both will work because we don't validate it.

isaacroldan: And In case we want to default to something, shouldn't we default to `3.x.x`? 2.x.x will be broken no?

alvaro-shopify: @isaacroldan @gonzaloriestra Added the next CLI3 version value as fallback. That should be the first version that includes support for the new env variable

PR #1961, opened 2023-05-19
Fix Update post purchase chrome extension url

alvaro-shopify:

Fixes https://github.com/Shopify/cli/pull/1959 unit tests


lizkenyon:
Thank you! I was just about to do this as well. 🙈

PR #1973, opened 2023-05-22
Use cli as a dependency of app, forwarding the command

amcaplan:

Users often get confused on upgrades, and don't upgrade both `@shopify/cli` and `@shopify/app`.

Reduces Shopify CLI to 1 single dependency, `@shopify/app`. This means the CLI gets updated atomically.

Next steps will be:

1. Update templates to remove the extra dependency
1. Decide whether to offer to auto-migrate users (since they can now drop the `@shopify/cli` dependency).

### How to test your changes?

Use `npm init` to create a new app. Point `@shopify/app` to the local copy, and delete the `@shopify/cli` dependency. Everything should still work as usual.

### Measuring impact

### Checklist

### DIFF
--- a/packages/app/bin/run.cmd
+++ b/packages/app/bin/run.cmd
@@ -1,0 +1,3 @@
+@echo off
+
+node "%~dp0\run.js" %*


alvaro-shopify (approved):
Great job!

I guess we should also need to update the `create-app` and `upgrade` commands as well

amcaplan:
Updating to Draft for now, as I see this will be controversial.

shauns:
Lets keep this open. Hopefully to be visited soon.

amcaplan:
Leave it open. It'll probably be easier to redo vs. solving all the merge conflicts, but this is a good automatic  reminder.

alvaro-shopify:
is [this](https://github.com/Shopify/cli/blob/9aca10a72e2a322e5621c6f28e2b8e52ce58344a/packages/cli/package.json#L29) still be needed?, I think that each package that includes the `cli` should "install" its own `shopify binary` right?

amcaplan:
@alvaro-shopify I only dealt with the `app` package, but the binstub still necessary for themes and h2, unless we perform the same change there.

PR #1983, opened 2023-05-25
Streamlined support for app modules (step 2)

isaacroldan:
This is part of the project to support generic app modules.

- Move the logic of build/bundle/config to the `ExtensionInstance`, each extension is responsible for building itself depending on the supported features.

### How to test your changes?
`node ./bin/create-test-app.js` to do some basic testing, after that, go to the created app folder and try to deploy or add more extensions.

### Measuring impact

### Checklist


packages/app/src/cli/services/deploy.ts
```diff
@@ -57,45 +56,21 @@ export async function deploy(options: DeployOptions) {
...
-      options.app.extensions.theme.map(async (extension) => {
-        return {
-          uuid: identifiers.extensions[extension.localIdentifier]!,
-          config: '{"theme_extension": {"files": {}}}',
-          context: '',
-        }
-      }),
-    )
-    extensions.push(...themeExtensions)
-  }
```
isaacroldan, line 80: deploy.ts no longer builds the config for each extension, that's a responsibility of the ExtensionInstance depending on the features.

packages/app/src/cli/services/generate.ts
```diff
@@ -94,13 +93,9 @@ function checkLimits(
 
 function limitReached(app: AppInterface, specifications: ExtensionSpecification[], templateType: TemplateType) {
   const type = templateType.type
-  if (type === 'function') {
-    return app.extensions.function.length >= blocks.functions.defaultRegistrationLimit
-  } else {
```
isaacroldan, line 99: There is a `function` spec already, we don't need to treat it as a special case anymore

packages/app/src/cli/models/extensions/specification.ts
```diff
@@ -225,6 +236,57 @@ export class ExtensionInstance<TConfiguration extends BaseConfigType = BaseConfi
...
+    if (!this.isFunctionExtension) return undefined
+    return this as unknown as FunctionExtension
+  }
+
+  async buildStep(options: ExtensionBuildOptions) {
+    if (this.isThemeExtension) {
+      return buildThemeExtension(this, options)
+    } else if (this.isFunctionExtension) {
+      return buildFunctionExtension(this.functionExtension!, options)
+    } else if (this.features.includes('esbuild')) {
```
matteodepalo, line 250: Breaking the pattern a bit here. Should we rename the function below to `buildExtensionWithEsbuild` or something?

packages/app/src/cli/models/extensions/specification.ts
```diff
@@ -225,6 +236,57 @@ export class ExtensionInstance<TConfiguration extends BaseConfigType = BaseConfi
...
+    }
+
+    // Workaround for tax_calculations because they remote spec NEEDS a valid js file to be included.
+    if (this.type === 'tax_calculation') {
+      await touchFile(this.outputBundlePath)
+      await writeFile(this.outputBundlePath, '(()=>{})();')
+    }
+  }
+
+  async bundleStep(options: ExtensionBuildOptions, identifiers: Identifiers, bundleDirectory: string) {
```
matteodepalo, line 261: What about naming this function `bundleCode` so that it's easier to understand in what it differs from `bundleConfig`? At the moment I'd expect the `bundleStep` to contain all the bundling instructions.

packages/app/src/cli/models/extensions/specification.ts
```diff
@@ -225,6 +236,57 @@ export class ExtensionInstance<TConfiguration extends BaseConfigType = BaseConfi
   get isJavaScript() {
     return Boolean(this.entrySourceFilePath?.endsWith('.js') || this.entrySourceFilePath?.endsWith('.ts'))
   }
+
+  get functionExtension(): FunctionExtension | undefined {
+    if (!this.isFunctionExtension) return undefined
+    return this as unknown as FunctionExtension
+  }
+
+  async buildStep(options: ExtensionBuildOptions) {
```
matteodepalo, line 245: Could we just call this `build`?

packages/app/src/cli/models/extensions/specification.ts
```diff
@@ -225,6 +236,57 @@ export class ExtensionInstance<TConfiguration extends BaseConfigType = BaseConfi
...
+    if (this.isThemeExtension) {
+      return buildThemeExtension(this, options)
+    } else if (this.isFunctionExtension) {
+      return buildFunctionExtension(this.functionExtension!, options)
+    } else if (this.features.includes('esbuild')) {
+      return buildUIExtension(this, options)
+    }
+
+    // Workaround for tax_calculations because they remote spec NEEDS a valid js file to be included.
+    if (this.type === 'tax_calculation') {
```
matteodepalo, line 255: Could we not fix the remote spec?

matteodepalo (approved):
Just a few naming comments, but the good looks good! Great refactor!

packages/app/src/cli/models/extensions/specification.ts
isaacroldan: they are going to, but will take some time, in the meantime i'll keep this here and remove it later

packages/app/src/cli/models/extensions/specification.ts
isaacroldan: yep

packages/app/src/cli/models/extensions/specification.ts
isaacroldan: Makes senses, in fact `buildStep` and `bundleStep` do mostly the same thing, the bundle step doesn't actually bundle, it just builds in a different folder that is later included in a `zip`. the only extension that is different here is `theme` that does have a bundle phase. 
So I wasn't sure how to name this

isaacroldan: what about `buildForBundle`?

packages/app/src/cli/models/extensions/specification.ts
matteodepalo: I agree actually, `buildUIExtension` does make sense if looked from that angle 👍

PR #1989, opened 2023-05-25
Add `--message` and `--version` flags to the `app deploy` command

AniTumany:

Closes https://github.com/Shopify/develop-app-foundations/issues/314

### Measuring impact

### Checklist

### DIFF
--- a/.changeset/red-comics-call.md
+++ b/.changeset/red-comics-call.md
@@ -1,0 +1,5 @@
+---
+'@shopify/app': patch
+---
+
+Allows to set a message and a version identifier when releasing a new app version
--- a/packages/app/oclif.manifest.json
+++ b/packages/app/oclif.manifest.json
@@ -119,9 +119,23 @@
         "no-release": {
           "name": "no-release",
           "type": "boolean",
-          "description": "Deploy without releasing it to the users.",
+          "description": "Creates a version but doesn't release it - it's not made available to merchants.",
           "hidden": false,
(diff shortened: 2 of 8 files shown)


ani-shopify:
Still need to work on adding unit tests

mattnoakes (commented):
Looks good, Ani! Tophatting worked as expected when running `app deploy` with and without the `message` and `version` flags for an app with deployments enabled. However, when I run `app deploy` with the `message` flag (the `version` flag works) for an app with deployments disabled, I get this error:
29-51-ybv3y-0t7hs
Is it okay to error out in this case?

ani-shopify:
> However, when I run app deploy with the message flag (the version flag works) for an app with deployments disabled, I get this error:
image
Is it okay to error out in this case?


I looked at this briefly. The issue is that [the `Apps::Models::Deployments::CreateStub` operation gets a `message` argument which it doesn't expect](https://github.com/Shopify/shopify/blob/9ef4325587b101a9bd05115cfa3e2bb70c5916ac/components/apps/app/controllers/services/internal/deployments_controller.rb#L229). 

image

Which we can easily fix in the backend. But I wonder if we should expect Shopify Core to handle these exceptions and not blow on user's face. @gordonhirsch what are your thoughts?

mattnoakes (approved):
> what do you think about merging this in and fix that bug on the backend separately?

Sounds good! This looks great!

PR #1994, opened 2023-05-26
Also run webhook requests in frontend-only apps

paulomarg:

Currently, we run the `APP_UNINSTALLED` webhook when the CLI switches apps, so that we can clean up the database and trigger a new install flow.

The issue here was that the CLI was assuming a `backend` config would always be present, which isn't necessarily true, since apps can sometimes use only `frontend` configs.

Ensuring that the webhook flow also works for frontend-only apps, preferring the backend config when present.

I've also added a 3s backoff period before firing the first attempt to give the app some time to spin up before we start trying, since it's almost guaranteed that the first try will fail if we fire it right away.

### How to test your changes?

Run `dev --reset` and ensure the webhook is fired and received.

### Post-release steps

None!

### Measuring impact

### Checklist

### DIFF
--- a/packages/app/src/cli/services/webhook/send-app-uninstalled-webhook.ts
+++ b/packages/app/src/cli/services/webhook/send-app-uninstalled-webhook.ts
@@ -31,6 +31,7 @@ export async function sendUninstallWebhookToAppServer(
 
   options.stdout.write('Sending APP_UNINSTALLED webhook to app server')
 
+  await sleep(3)
   const result = await triggerWebhook(options, sample)
 
   options.stdout.write(result ? 'APP_UNINSTALLED webhook delivered' : 'APP_UNINSTALLED webhook delivery failed')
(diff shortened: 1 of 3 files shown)


packages/app/src/cli/services/webhook/send-app-uninstalled-webhook.ts
```diff
@@ -31,6 +31,7 @@ export async function sendUninstallWebhookToAppServer(
 
   options.stdout.write('Sending APP_UNINSTALLED webhook to app server')
 
+  await sleep(3)
```
shauns, line 34: Did you mean to leave this in?

paulomarg: Yes, the webhook was firing at the same time as the actual process started, which led to it failing most of the times. This is so we back off for a few seconds to give the app time to spin up.

PR #2025, opened 2023-05-29
Fix theme dev re-authentication

gonzaloriestra:

Fixes #0000 

### Measuring impact

### Checklist


packages/theme/src/cli/commands/theme/dev.ts
```diff
@@ -131,31 +129,20 @@ export default class Dev extends ThemeCommand {
...
-      controller = new AbortController()
-      // eslint-disable-next-line @typescript-eslint/no-floating-promises
-      this.execute(adminSession, flags.password, command, controller, true)
-    }, this.ThemeRefreshTimeoutInMs)
-
     renderLinks(store, flags.theme!, flags.host, flags.port)
 
-    // eslint-disable-next-line @typescript-eslint/no-floating-promises
-    this.execute(adminSession, flags.password, command, controller, false)
+    // eslint-disable-next-line @typescript-eslint/no-misused-promises
```
gonzaloriestra, line 134: I've been reading about this, but I'm not sure if there's a better way to do this (although it works)...

karreiro (commented):
Thank you, @gonzaloriestra! Amazing stuff 🔥🚀

This approach is quite clever and it's working well with the standard authentication method (based on the `--store` flag). The only minor issue I've noticed is that,I could not authenticate via Theme Access app.

Other than that minor issue, everything else is excellent 🚀

alvaro-shopify:
> Thank you, @gonzaloriestra! Amazing stuff 🔥🚀
> 
> This approach is quite clever and it's working well with the standard authentication method (based on the `--store` flag). The only minor issue I've noticed is that,I could not authenticate via Theme Access app.
> 
> Other than that minor issue, everything else is excellent 🚀

I fixed the problem. 🎩  it and it's working. Please, could you check it again?

PR #2046, opened 2023-06-01
Add sdl file config to flow action extension

mikee649:
## Desctiption

- [Ticket](https://github.com/Shopify/flow/issues/18984)
- Adds reusable SDL file support to flow action POC
- 🎩 https://videobin.shopify.io/v/Xagp2p


nsimone (approved):
Nice! Is this included in the bundle on deploy, or will that come later?

Issue #1118, opened 2023-01-17
[Bug]: Issue with the reverse proxy for backend served architecture


bakura10:
### Please confirm that you have:

- [X] Searched [existing issues](.) to see if your issue is a duplicate. (If you’ve found a duplicate issue, feel free to add additional information in a comment on it.)
- [X] Reproduced the issue in [the latest CLI version](https://www.npmjs.com/package/@shopify/cli).

### In which of these areas are you experiencing a problem?

App

### Expected behavior

Hi,

I am trying to get a grasp of the new architecture for apps, but after many attempts with the PHP boilerplate project, I could not find the logic behind some of the choices. I have therefore tried to simplify the whole structure which makes more sense to me, managed to somewhat make it works, but now hit a blocking road.

My structure is the following:

```
- backend
  -- shopify.web.toml (with a backend type)
- frontend
  -- shopify.web.toml (with a frontend type)
```

The frontend uses ESbuild to compile, and save the result into the /backend/assets folder. The Polaris app is therefore being served by the _backend_ app.

However, I still want to have the shopify.web.toml so that when I use "shopify app dev", it runs both the backend and the frontend code that is in charge of building the code with ESBuild.

Everything starts correctly: Shopify CLI properly launch the backend and frontend commands. No error appears at this stage.

<img width="810" alt="Capture d’écran 2023-01-17 à 12 11 36" src="https://user-images.githubusercontent.com/1198915/212801694-eb453643-e2a6-469d-b7f2-ee7c8c64a3a2.png">

But as soon as I start the app, this error appears:

<img width="508" alt="image" src="https://user-images.githubusercontent.com/1198915/212801753-ee9e120c-c51f-42e1-8123-fd8ff1e6b32d.png">

I tried to read the source of the dev command, and the issue seems to be coming from the fact that the frontend apparently setup a reverse proxy, and hitting the ngrok URL hit this instead of going to the backend serving the frontend.

Said otherwise, I would like to have a "frontend" environment whose _only_ purpose is to launch the commands to build and watch the frontend environment, without setting the whole Vite and proxy server complexity. Maybe through a proxy option:

```
type="frontend"
proxy=false

[commands]
dev = "npm run dev"
build = "npm run build"
```

How would that be possible under the current architecture?

Let me know if anything is unclear.

### Actual behavior

An error is shown

### Reproduction steps

1.
2.
3.

### Operating System

OSX

### Shopify CLI version (check your project's `package.json` if you're not sure)

3.33

[...]

PR #2083, opened 2023-06-01
Unify webs

amcaplan:

Relates to #1118 among other things

We currently have strict `frontend` and `backend` roles. But real-life apps don't match this so neatly. Monoliths have a single web which is both frontend and backend, and other apps have extra processes which don't fit these roles as all.

Replaces `type` with an array of `roles` (though `type` is still honored for the time being).

Treats webs as mostly the same, with certain special treatment for the `frontend` and `backend` roles:

* `frontend` is used as the embedded app
* `backend`'s port is passed as an ENV var to all apps

There is also a new `background` role which does neither special behavior.

### How to test your changes?

Take an existing app, convert frontend and backend to use roles. You can also play with a background app, like:

```
type="background"
name="backy"

[commands]
dev = "ls -lah"
build = "echo 'building'"
```

### Post-release steps

Will be updated - need to merge a future docs PR in the next minor release. But this is safe to release without that, because it's 100% backwards compatible.

Docs PR: https://github.com/Shopify/shopify-dev/pull/34791

### Measuring impact

### Checklist

### DIFF
--- a/packages/app/src/cli/models/app/app.ts
+++ b/packages/app/src/cli/models/app/app.ts
@@ -36,8 +36,8 @@ const baseWebConfigurationSchema = zod.object({
 })
 const webTypes = zod.enum([WebType.Frontend, WebType.Backend, WebType.Background]).default(WebType.Frontend)
 export const WebConfigurationSchema = zod.union([
-  baseWebConfigurationSchema.extend({type: webTypes}),
   baseWebConfigurationSchema.extend({roles: zod.array(webTypes)}),
+  baseWebConfigurationSchema.extend({type: webTypes}),
 ])
 export const ProcessedWebConfigurationSchema = baseWebConfigurationSchema.extend({roles: zod.array(webTypes)})
 


packages/app/src/cli/services/dev.ts
```diff
@@ -107,15 +98,12 @@ async function dev(options: DevOptions) {
     localApp = await installAppDependencies(localApp)
   }
 
-  const frontendConfig = localApp.webs.find(({configuration}) => configuration.type === WebType.Frontend)
-  const backendConfig = localApp.webs.find(({configuration}) => configuration.type === WebType.Backend)
+  const backendConfig = localApp.webs.find((web) => isWebType(web, WebType.Backend))
```
paulomarg, line 101: Should we support multiple configs here, since we could technically have multiple backends?

packages/app/src/cli/models/app/app.ts
```diff
@@ -34,6 +34,11 @@ export const WebConfigurationSchema = zod.object({
     dev: zod.string(),
   }),
 })
+const webTypes = zod.enum([WebType.Frontend, WebType.Backend, WebType.Background]).default(WebType.Frontend)
+export const WebConfigurationSchema = zod.union([
+  baseWebConfigurationSchema.extend({type: webTypes}),
+  baseWebConfigurationSchema.extend({roles: zod.array(webTypes)}),
+])
```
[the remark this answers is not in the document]
amcaplan: Probably something closer to:

```js
roles = Array.from(new Set([type, ...(roles || [])))
```

but yeah, agreed in principle. Still want to validate via schema that they passed something explicit, though.

packages/app/src/cli/services/dev.ts
amcaplan: I think only one can be The Backend which the frontend iframe accesses via the `BACKEND_PORT` env var. If an app has multiple backends, these can use the `background` role. If the processes need to communicate, they should set explicit ports and hardcode in the dependent processes.

Basically the new rules are:

1. Every app has at least 1 role
2. There must be a `frontend`
3. There can only be 1 `frontend` (embedded app) and 1 `backend` (used for `BACKEND_PORT`). Unlimited webs can be `background` (generic role with no special behavior for now)
4. The same app can be both `frontend` and `backend`

packages/app/src/cli/models/app/app.ts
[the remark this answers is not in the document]
paulomarg: Yeah definitely, it'd be perfect if the schema allowed for both but combined them so the code doesn't have to worry which one is actually set.

paulomarg:
I had a thought just now: if a web uses both backend and frontend roles, the ports should probably be the same, right? Are we doing that now?

packages/app/src/cli/models/app/loader.ts
```diff
@@ -269,17 +269,35 @@ class AppLoader {
...
+          undefined,
+          websOfType[1]!.directory,
+        )
+      }
+    })
+  }
+
   async loadWeb(WebConfigurationFile: string): Promise<Web> {
+    const config = await this.parseConfigurationFile(WebConfigurationSchema, WebConfigurationFile)
+    const roles = new Set('roles' in config ? config.roles : [])
```
isaacroldan, line 295: would this work?
```suggestion
    const roles = new Set(config.roles ?? [])
```

amcaplan: @isaacroldan this is a typescript thing, because it's a union type you can never just say `config.roles` but instead you have to establish that we're in this situation where `roles` exists.

packages/app/src/cli/services/info.ts
```diff
@@ -149,8 +149,15 @@ class AppInfo {
...
+          const {name, roles} = web.configuration
+          sublevels.push([`    📂 ${name} (${roles.join(',')})`, relativePath(this.app.directory, web.directory)])
+        } else {
+          web.configuration.roles.forEach((role) => {
+            sublevels.push([`    📂 ${role}`, relativePath(this.app.directory, web.directory)])
+          })
+        }
       } else if (this.app.errors) {
         const error = this.app.errors.getError(`${web.directory}/${configurationFileNames.web}`)
         if (error) {
```
alvaro-shopify, line 163: could we do a little refactor on this just to avoid nested `if` ?

```suggestion
      if (this.app.errors && !web.configuration) {
        const error = this.app.errors.getError(`${web.directory}/${configurationFileNames.web}`)
        if (error) {
          sublevels.push([`  📂 ${UNKNOWN_TEXT}`, relativePath(this.app.directory, web.directory)])
          errors.push(error)
        }
      }

      if (web.configuration.name) {
        const {name, roles} = web.configuration
        sublevels.push([`    📂 ${name} (${roles.join(',')})`, relativePath(this.app.directory, web.directory)])
      } else {
        web.configuration.roles.forEach((role) => {
          sublevels.push([`    📂 ${role}`, relativePath(this.app.directory, web.directory)])
        })
      }
```

amcaplan: This refactor isn't the logical equivalent of what was there before. It really is an either-or. Either we identified the web properly and display information, or there was an issue identifying the web and we display `unknown`.

amcaplan: That said, I think my logic is flawed too. If there's `web.configuration` but it has no type/role, the new logic doesn't do the right thing. I'll have to fix.

alvaro-shopify: > )

You are right I missed in the second if, sorry :( 

```ts 
if (web.configuration && web.configuration.name)
```

amcaplan:
Noting for posterity: After speaking with @isaacroldan, we identified the problem and the fix is [here](https://github.com/Shopify/cli/pull/2200). Not introduced by this PR, just brought to light by this work.

packages/app/src/cli/services/info.ts
amcaplan: Fixed in https://github.com/Shopify/cli/pull/2083/commits/dd4c444a664e18154b490e249fe92f3ce624067b

PR #2093, opened 2023-06-05
Add warning message to deploy

matteodepalo:
Fixes https://github.com/Shopify/app-deploys/issues/514

We want to notify users that changes are coming to deployments. This banner still lacks the URL so either way for it if it's going to come soon or ship it without the `Reference` section for now.

Screenshot 2023-06-05 at 11 22 45


alvaro-shopify (approved):

ani-shopify:
@matteodepalo I believe @jessehoek has updated the banner to be an info banner and not a warning.

image

Would you mind updating the code to reflect that change?

ani-shopify:
@matteodepalo I see this PR is merged in with main, but the `Reference` has "Introducing Deployments 2.0" and a link. Both the name and the URL are not finalized. So we should make sure this doesn't get released until both of those two things are finalized.

cc @jessehoek @darius-gai @MeredithCastile @shainaraskas

PR #2099, opened 2023-06-06
Use more vertical space in select prompts

amcaplan:

Lets us see many more options, hooray!

Rather than limiting to displaying what's currently definitely visible, we add some logic to show what's possibly visible and just cut it off at the bottom using flexbox `overflowY="hidden"`.

Before:
before

After:
after

### How to test your changes?

Try prompts in the CLI, like:

```sh
pnpm shopify kitchen-sink prompts
pnpm shopify demo catalog
```

### Measuring impact

### Checklist


matteodepalo (approved):

packages/cli-kit/src/private/node/ui/components/SelectInput.tsx
```diff
@@ -232,20 +250,30 @@ function SelectInputInner<T>(
       </Box>
     )
   } else {
+    if (hasLimit) useStdout().stdout.write(ansiEscapes.clearTerminal)
```
matteodepalo, line 253: Remember to put `useStdout` outside of `if` conditions otherwise React will complain. I believe if you use the `eslint` plugin in vscode it will tell you that this is an error.

packages/cli-kit/src/private/node/ui/hooks/use-select-state.ts
```diff
@@ -320,7 +345,7 @@ export const useSelectState = <T>({visibleOptionCount, options, defaultValue}: U
   }, [])
 
   const visibleOptions = useMemo(() => {
-    return options.slice(state.visibleFromIndex, state.visibleToIndex)
+    return options.slice(state.visibleFromIndex)
```
matteodepalo, line 348: Do we still need visibleToIndex if we don't use it?

amcaplan: It's still used [here](https://github.com/Shopify/cli/pull/2099/files#diff-a2d15153e3c654b299aede1768af71ce7b182c9d5bb26e669affdd5f591ce30eR129).

matteodepalo: In that case, could we remove the dependency on it in this useMemo?

matteodepalo:
The solution does seem to work but I'm starting to wonder if it's worth going through all this trouble in order to support a few narrow terminals. My fear is the logic is becoming quite complicated, with a few "magic numbers" like `- 1` and `- 5`, and looking at this code again in a few months will become really difficult.

It could also be worth adding some specific tests that would fail before this PR and pass after these fixes to show why we're introducing all these changes. It's possible to stub the `stout.rows` to be less than the number of options and seeing how the old and new logic behave.

One alternative I can think of is showing that the prompt is taller than the terminal and if the user wants to see all options they should resize it. Don't get me wrong, it's really cool that we're solving this in such a clever way, I'm just worried about the increased complexity.

packages/cli-kit/src/private/node/ui/components/SelectInput.tsx
```diff
@@ -251,7 +259,7 @@ function SelectInputInner<T>(
     )
   } else {
     const optionsHeight = initialItems.length + maximumLinesLostToGroups(initialItems)
-    const maxKeyLength = initialItems
+    const maxKeyLength = itemsWithKeys
```
amcaplan, line 262: I disagree with this change. We should use `initialItems` so the width doesn't flicker as items are scrolled through.

PR #2102, opened 2023-06-06
[Feature] Support Function `[[targets]]`

jbourassa:

In #gsd:34464, we're breaking down Function APIs into multiple _targets_. The exact TOML is still being debated, but this should be reasonably close, and not be used in production until we roll out unified deployment.

Closes https://github.com/Shopify/script-service/issues/6912

Add support for a TOML that looks like:
```toml
name = "network-validation"
type = "cart_checkout_validation"
api_version = "unstable"

[[targets]]
target = "purchase.validation.one"
input_query = "path/to/one.graphql"

[[targets]]
target = "purchase.validation.two"
input_query = "path/to/two.graphql"
```

Map that targets array to a new `targets` field in the config, of type `{handle: string, input_query: string | undefined}[]`.

Without the unified deployment beta flag, having the `[[targets]]` array won't error, but it won't do anything either (the non-unified-deployment path does not read the targets array). That makes the implementation simpler and feels like a reasonable tradeoff: unified deployment will be fully rolled out soon-ish.

### How to test your changes?

It's going to be difficult, but here's how:

**Setup:**
- Checkout the branch from Shopify/shopify#428889
- Add a target for `target = "purchase.validation.perform"` in your `function.extension.toml`.
- Enable the unified deployment beta on your app
- Create a **dev** store
- Install the app on your dev store

**Action:**
- Deploy the app (app deploy)
- Create a deployment in the UI
- Setup a checkout validation in your store

**Outcome:**
The function should've ran with the input query from that was specified in `purchase.validation.perform`.

### Checklist

=> not yet, we're not rolling out the feature yet.


packages/app/src/cli/services/deploy/upload.ts
```diff
@@ -479,5 +494,17 @@ export async function functionConfiguration(
...
+  }
+}
+
+async function readInputQuery(path: string): Promise<string> {
+  if (await fileExists(path)) {
+    return readFile(path)
+  } else {
+    throw new AbortError(
+      `No input query file at ${path}.`,
+      `Create the file or remove the line referencing it in the extenion's TOML.`,
```
dphm, line 507:
```suggestion
      `Create the file or remove the line referencing it in the extension's TOML.`,
```

jbourassa:
Some changes since I opened the PR:

- Added an `export` field, because we're switching from a WASI-args based approach to named export.
- Renamed `[[targets]]` array to `[[targeting]]` because that's the leading proposal for the TOML structure. The resulting extension config is still an array of `targets` because I think internally the name makes more sense (i.e. the data structure field should not be `targeting: Target[]`, but instead `targets: Target[]`.
- The extension building code moved to a different file, so I moved the additions too.

I'm working on related PRs so the above WASI-args-to-export works, thus haven't yet tried the end-to-end solution. Will report back once that's done, but feel free to give this another review if you'd like.

PR #2106, opened 2023-06-06
[Feature] Add Flow Extensions to the CLI

MaxLeduc:

We're moving the Flow extension generation process from the Partner's Dashboard to the CLI. This is the first step towards that.

We're adding the capacity to generate both Flow actions and triggers from the CLI. This feature will be gated by a [beta flag in Partners](https://github.com/Shopify/partners/pull/48052).

### How to test your changes?

You can use my [spin instance](https://generic-extensions-1.maxime-leduc.us.spin.dev/) to test this out or create your own with these PRs from [Core](https://github.com/Shopify/shopify/pull/430734) and [Partners](https://github.com/Shopify/partners/pull/48052).

### DIFF
--- a/packages/app/src/cli/models/extensions/specifications/flow_action.ts
+++ b/packages/app/src/cli/models/extensions/specifications/flow_action.ts
@@ -1,0 +1,52 @@
+import {BaseSchema} from '../schemas.js'
+import {createExtensionSpecification} from '../specification.js'
+
+import {zod} from '@shopify/cli-kit/node/schema'
+
+const FlowActionExtensionSchema = BaseSchema.extend({
+  name: zod.string(),
+  type: zod.literal('flow_action'),
+  task: zod.object({
+    title: zod.string(),
+    description: zod.string(),
+    url: zod.string(),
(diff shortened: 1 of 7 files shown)


packages/app/src/cli/models/extensions/specifications/flow_action.ts
```diff
@@ -0,0 +1,50 @@
...
+  name: zod.string(),
+  type: zod.literal('flow_action'),
+  task: zod.object({
+    title: zod.string(),
+    description: zod.string(),
+    url: zod.string(),
+    validationUrl: zod.string().optional(),
+    customConfigurationPageUrl: zod.string().optional(),
+    customConfigurationPagePreviewUrl: zod.string().optional(),
+    fields: zod.array(
```
ryanische, line 16: Currently we can create an action with no fields - should this array be optional?

MaxLeduc: Ah yes good call.

omnisyle (approved):

PR #2115, opened 2023-06-07
Handle 'unsupported_client_version' error from partners

matteodepalo:

Fixes https://github.com/Shopify/develop-app-foundations/issues/270

We want to show a helpful error message for old clients that are trying to deploy.

This PR adds error handling in three places:
- GenerateSignedUploadUrl mutation
- UploadUrlGenerate mutation
- CreateDeployment mutation

We can merge this PR whenever we want.

### How to test your changes?

- Try to deploy with partners branch set to `app-deploy-validate-cli-version`.
- You should get this error message

Screenshot 2023-06-07 at 16 25 23


packages/app/src/cli/services/deploy/upload.ts
```diff
@@ -143,7 +145,9 @@ export async function uploadExtensionsBundle(
   }
 
   const mutation = CreateDeployment
-  const result: CreateDeploymentSchema = await partnersRequest(mutation, options.token, variables)
+  const result: CreateDeploymentSchema = await handlePartnersErrors(() =>
+    partnersRequest(mutation, options.token, variables),
```
isaacroldan, line 149: if you look in the implementation of `partnersRequest`, there is a config: `responseOptions` that accepts an `onError`. 
Can we use that and pass the errorHandler as an optional variable in this function? I think it'd be cleaner and more easy to reuse

matteodepalo: I only see `onResponse` and `handleErrors`. Am I looking in the wrong place?

PR #2117, opened 2023-06-07
[App Modules] Remove extension categories in app

isaacroldan:
As part of the App modules refactor, i'm doing some cleanup of things we no longer need.

- Remove the `extensions: {ui, theme, function}` object from the `AppInterface` and `App` object in favor of the `allExtensions` generic one. I'll eventually rename it, but in a different PR to avoid having too many changes here.
- Move function and theme validations to their respective specification
- Update some tests to use the standard `testUIExtension` 

### Measuring impact

### Checklist


packages/app/src/cli/validators/extensions/functions.ts
```diff
@@ -1,37 +0,0 @@
-import {FunctionExtension} from '../../models/app/extensions.js'
-import {fileExists} from '@shopify/cli-kit/node/fs'
-import {AbortError} from '@shopify/cli-kit/node/error'
-import {outputContent, outputToken, stringifyMessage} from '@shopify/cli-kit/node/output'
-
-const extensionLine = (extension: {id: string; path: string}): string => {
```
matteodepalo, line 6: Did you try the output before/after this removal? It seems to be different, but maybe a simpler output is what you were aiming for as well.

isaacroldan: It is different, but I wanted to simplify this validation. It now shows the error per extension, because it's weird that multiple functions have the same issue at the same time...
I can re-review this, but I think is the way to way, each extension defines and shows its own errors

PR #2124, opened 2023-06-08
Add app_release mutation

matteodepalo:

Fixes https://github.com/Shopify/app-deploys/issues/474

We want to add a `release` command to make a specific `version` active.

We've added a `app release --version=1` command to make the specified version active. If the version doesn't exist we throw an error. The user is then prompted with a diff of all the extensions from the last released version to the one specified. If the user confirms we then call the `appRelease` mutation. Releases might have to be reviewed so we handle that scenario.

It seems like the `versionsDiff` not returning what we expect for now.

### How to test your changes?

- Run `shopify app deploy --no-release`
- Copy the version number
- Run `shopify app release --version={versionNumber}`

### Post-release steps

We need to add this command to the docs.

### Measuring impact

### Checklist

### DIFF
--- a/packages/app/src/cli/commands/app/release.ts
+++ b/packages/app/src/cli/commands/app/release.ts
@@ -27,7 +27,7 @@ export default class Release extends Command {
     }),
     version: Flags.string({
       hidden: false,
-      description: 'Version tag of the version you want to release.',
+      description: 'The name of the app version to release.',
       env: 'SHOPIFY_FLAG_VERSION',
       exclusive: ['app-version-id'],
       required: true,
--- a/packages/app/src/cli/commands/app/release.ts
+++ b/packages/app/src/cli/commands/app/release.ts
@@ -30,12 +30,7 @@ export default class Release extends Command {
       description: 'Version tag of the version you want to release.',
       env: 'SHOPIFY_FLAG_VERSION',
       exclusive: ['app-version-id'],
-    }),
-    'app-version-id': Flags.string({
-      hidden: false,
-      description: 'Version identifier of the version you want to release.',
-      env: 'SHOPIFY_FLAG_VERSION_ID',
-      exclusive: ['version'],
+      required: true,
     }),
   }
 
@@ -54,7 +49,6 @@ export default class Release extends Command {
       reset: flags.reset,
       force: flags.force,
       version: flags.version,
-      versionId: flags['app-version-id'],
     })
   }
 }
--- a/packages/app/src/cli/commands/app/release.ts
+++ b/packages/app/src/cli/commands/app/release.ts
@@ -29,6 +29,13 @@ export default class Release extends Command {
       hidden: false,
       description: 'Version tag of the version you want to release.',
       env: 'SHOPIFY_FLAG_VERSION',
+      exclusive: ['app-version-id'],
+    }),
+    'app-version-id': Flags.string({
+      hidden: false,
+      description: 'Version identifier of the version you want to release.',
+      env: 'SHOPIFY_FLAG_VERSION_ID',
+      exclusive: ['version'],
     }),
   }
 
@@ -47,6 +54,7 @@ export default class Release extends Command {
       reset: flags.reset,
       force: flags.force,
       version: flags.version,
+      versionId: flags['app-version-id'],
     })
   }
 }
--- a/packages/app/src/cli/commands/app/release.ts
+++ b/packages/app/src/cli/commands/app/release.ts
@@ -1,0 +1,52 @@
+import {appFlags} from '../../flags.js'
+import {AppInterface} from '../../models/app/app.js'
+import {load as loadApp} from '../../models/app/loader.js'
+import Command from '../../utilities/app-command.js'
+import {loadExtensionsSpecifications} from '../../models/extensions/load-specifications.js'
+import {release} from '../../services/release.js'
+import {Flags} from '@oclif/core'
+import {globalFlags} from '@shopify/cli-kit/node/cli'
+import {addPublicMetadata} from '@shopify/cli-kit/node/metadata'
+
+export default class Release extends Command {
+  static description = 'Release an app version.'
+
+  static flags = {
+    ...globalFlags,
+    ...appFlags,
+    'api-key': Flags.string({
+      hidden: false,
+      description: 'The API key of your app.',
+      env: 'SHOPIFY_FLAG_APP_API_KEY',
+    }),
+    reset: Flags.boolean({
+      hidden: false,
+      description: 'Reset all your settings.',
+      env: 'SHOPIFY_FLAG_RESET',
+      default: false,
+    }),
+    version: Flags.string({
+      hidden: false,
+      description: 'Version tag of the version you want to release.',
+      env: 'SHOPIFY_FLAG_VERSION',
+    }),
+  }
+
+  async run(): Promise<void> {
+    const {flags} = await this.parse(Release)
+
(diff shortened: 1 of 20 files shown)


matteodepalo:
@ani-shopify This is still in draft but is already using the `versionsDiff` query correctly I believe.
We found a couple of issues with it though:
1. When there are no releases and we try to release the an unreleased version we get empty arrays in all sets
2. When we try to release a version that is already released we get this from the query. This also happens if we deploy an unreleased version without changing anything and try to release that.

Screenshot 2023-06-08 at 16 39 45

packages/app/src/cli/services/release.test.ts
```diff
@@ -0,0 +1,293 @@
...
+      // eslint-disable-next-line no-await-in-loop
+      await task.task({}, task)
+    }
+
+    return {}
+  })
+})
+
+describe('release', () => {
+  test('throws an error if versionDiff is empty', async () => {
```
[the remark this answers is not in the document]
ani-shopify: You are correct @matteodepalo ! Created an [issue](https://github.com/Shopify/app-deploys/issues/551) to track this.

ani-shopify:
@alvaro-shopify I know there are still some issues with showing the list of extensions when running the `app release` command. I was wondering if it would make sense to split this into two separate issue. 

1. To actually get the `app release` command to use the `appRelease` mutation with the `versionTag`. 
2. To use the `versionsDiff` to display the list of included and to-be-removed extensions.

By splitting this into two separate tickets will allow getting the command out and iterate over it by adding the prompt later on. Let me know what you think.

ani-shopify:
> On the other hand we are using the VersionDiff operation to check if the used versionTag exists or not (in case the list is empty), should we remove this validation and rely on the error returned by the operation AppRelease?.

Yeah, I don't think we should rely on `versionDiff` query to validate the `versionTag`

packages/app/src/cli/commands/app/release.ts
```diff
@@ -0,0 +1,60 @@
...
+      env: 'SHOPIFY_FLAG_VERSION',
+      exclusive: ['app-version-id'],
+    }),
+    'app-version-id': Flags.string({
+      hidden: false,
+      description: 'Version identifier of the version you want to release.',
+      env: 'SHOPIFY_FLAG_VERSION_ID',
+      exclusive: ['version'],
+    }),
+  }
```
ani-shopify, line 40: I don't think we want to make this additional flag available. 

cc @jessehoek

packages/app/src/cli/services/context.ts
```diff
@@ -331,6 +344,30 @@ export async function ensureDeployContext(options: DeployContextOptions): Promis
   return result
 }
 
+export async function ensureReleaseContext(options: ReleaseContextOptions): Promise<ReleaseContextOutput> {
+  const token = await ensureAuthenticatedPartners()
+  const [partnersApp, envIdentifiers] = await fetchAppAndIdentifiers(options, token)
+  const identifiers: Identifiers = envIdentifiers as Identifiers
+
+  if (!partnersApp.betas?.unifiedAppDeployment) {
```
ani-shopify, line 352: Can we instead use `deploymentMode` instead? Its values can be `'legacy' | 'unified' | 'unified-skip-release'`

packages/cli/src/cli/services/kitchen-sink/static.ts
```diff
@@ -157,6 +164,11 @@ export async function staticService() {
 
   renderFatalError(new AbortError('No Organization found', undefined, nextSteps))
 
+  renderError({
+    headline: "Version couldn't be released.",
+    body: 'This version needs to be submitted for review and approved by Shopify before it can be released.',
```
ani-shopify, line 169: Out of curiosity: what's the purpose of this here?

packages/app/src/cli/services/release.ts
```diff
@@ -0,0 +1,81 @@
...
+      headline: "Version couldn't be released.",
+      body: [
+        {link: {url: deployment.location, label: deployment.versionTag}},
+        `\n${deployment.message}`,
+        `\n\n${errorsMessage}`,
+      ],
+    })
+  } else {
+    renderSuccess({
+      headline: 'Version released to users.',
```
ani-shopify, line 70:
```suggestion
      headline: 'Version released to users',
```

packages/app/src/cli/services/release.ts
```diff
@@ -0,0 +1,81 @@
...
+      body: [{link: {url: deployment.location, label: deployment.versionTag}}, `\n${deployment.message}`],
```
ani-shopify, line 71: `message` can be `null`. And when it's null we do show this

image

packages/app/src/cli/services/release.ts
```diff
@@ -0,0 +1,81 @@
...
+  ]
+
+  const {
+    appRelease: {appRelease: release},
+  } = await renderTasks<Context>(tasks)
+
+  const deployment = release.deployment
+
+  if (release.userErrors?.length > 0) {
+    const errorsMessage = release.userErrors.map((error) => error.message).join(', ')
```
ani-shopify, line 58: When I tried to release an app version that couldn't me released I only saw the error message, it didn't have a header, version and the optional message in the error banner:

image

For a reference this is the mock:

image

packages/app/src/cli/commands/app/release.ts
jessehoek: So I understand correctly: is this a `--app-version-id` flag? If so, I don't think we need this? We should just have a `--version` flag

packages/app/src/cli/services/release.test.ts
```diff
@@ -0,0 +1,213 @@
...
+          },
+        },
+        '\nmessage',
+      ],
+      headline: 'Version released to users.',
+      nextSteps: [
+        [
+          'Run',
+          {
+            command: 'yarn shopify app versions list',
```
jessehoek, line 91: is this dynamic to the package manager the developer is using?

packages/app/src/cli/services/context.ts
alvaro-shopify: `deploymentMode` is a local variable [included](https://github.com/Shopify/cli/blob/63dc8c18426d35e2a817ebeaad5043948bb153fd/packages/app/src/cli/services/context.ts#L297) in the method `ensureDeployContext`.  
In the release context, the requirement about the `deploymentMode` is that the command should exit in case the app is configured as `legacy`. Thus there are some differences between the deployment context and the release one:

- `unified-skip-release` - This does not make sense in the `release` context, as we are indeed releasing :) 
- In the deploy context, in case the app is configured as ` legacy` there is a prompt to switch to `unified`. I guess that in the release flow this does not make sense, because a previous version tag in needed to create a release so at this point the app should be using `unified` deployments

Maybe it would be clearer if we do something like this:

```ts
  const deploymentMode: DeploymentMode = partnersApp.betas?.unifiedAppDeployment ? 'unified': 'legacy'
  if (deploymentMode === 'unified') {
    throw new AbortSilentError()
  }
```

however that local variable is going to be used only once in the next conditional

packages/app/src/cli/services/release.test.ts
jessehoek: Oh 🤦 that makes sense. sorry I didn't realize this was a test 😄

alvaro-shopify: no worries! 😀

Issue #2084, opened 2023-06-02
[Bug]: The option --admin is not supported


bakura10:
### Please confirm that you have:

- [X] Searched [existing issues](.) to see if your issue is a duplicate. (If you’ve found a duplicate issue, feel free to add additional information in a comment on it.)
- [X] Reproduced the issue in [the latest CLI version](https://www.npmjs.com/package/@shopify/cli).

### In which of these areas are you experiencing a problem?

Theme

### Expected behavior

Since  upgrading to Shopify CLI 3.46, I am facing this error whenever I launch the run command for themes.

I'm using a "shopify.theme.toml" file:

```
[environments.dev]
password = "shptka_xxxxxx"
path = "./dist"
store = "prestige-new-version"
theme = "yyyyyy"
```

And the command to launch:

```
shopify theme dev --environment dev
```

I had to rollback to 3.45.4, where it is working as expected.

### Actual behavior

It shows this error:

<img width="276" alt="image" src="https://github.com/Shopify/cli/assets/1198915/5beef0fd-1913-49d6-8d3c-984e6f9cc03d">

### Verbose output

```shell
/
```

### Reproduction steps

1.
2.
3.

### Operating System

OS X

### Shopify CLI version (check your project's `package.json` if you're not sure)

3.46.2

amcaplan:
Hi @bakura10, can you run with `--verbose` and paste the output? I tried reproducing and didn't encounter this error, so I'm curious to understand better where this might be coming from.

bakura10:
@amcaplan , sure!

```
2023-06-05T09:38:42.077Z: Running command theme dev
2023-06-05T09:38:42.085Z: Reading the content of file at shopify.theme.toml...
╭─ info ─────────────────────────────────────────────────────────────────────────╮
│                                                                                │
│  Using applicable flags from dev environment:                                  │
│                                                                                │
│    • password: ********2b3d                                                    │
│    • path: /Users/michaelgallego/Sites/shopify-prestige/dist                   │
│    • store: prestige-new-version.myshopify.com                                 │
│    • theme: redacted                                                       │
│                                                                                │
╰────────────────────────────────────────────────────────────────────────────────╯

2023-06-05T09:38:42.141Z: Ensuring that the user is authenticated with the Theme API with the following scopes:
[]

2023-06-05T09:38:42.151Z: 
Running system process:
  · Command: ruby -v
  · Working directory: /Users/michaelgallego/Sites/shopify-prestige

[...]

PR #2137, opened 2023-06-09
Fix theme token refresh

gonzaloriestra:

Fixes https://github.com/Shopify/cli/issues/2084

In https://github.com/Shopify/cli/pull/2025 we made a last minute change that broke the refresh system when you don't pass a password.

- Fixes recurring token refresh for `theme dev` when no password is provided
- Extracts some logic from the command to the service and adds tests

### How to test your changes?

- `p shopify theme dev` and wait for 2 hours
- `p shopify theme dev --password shptka_blabla `

### Measuring impact

### Checklist

### DIFF
--- a/packages/cli-kit/src/private/node/constants.ts
+++ b/packages/cli-kit/src/private/node/constants.ts
@@ -24,6 +24,7 @@ export const environmentVariables = {
   unitTest: 'SHOPIFY_UNIT_TEST',
   verbose: 'SHOPIFY_FLAG_VERBOSE',
   noThemeBundling: 'SHOPIFY_CLI_NO_THEME_BUNDLING',
+  bundledThemeCLI: 'SHOPIFY_CLI_BUNDLED_THEME_CLI',
   // Variables to detect if the CLI is running in a cloud environment
   codespaceName: 'CODESPACE_NAME',
   codespaces: 'CODESPACES',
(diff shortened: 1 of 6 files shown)


karreiro (commented):
Thank you very much for this PR, @gonzaloriestra!

To prevent the error mentioned in [Shopify/cli#2084](https://github.com/Shopify/cli/issues/2084), do you think it would be possible to enforce the use of the embedded CLI when utilizing the refresh-tokens mechanism? (actually, we maybe want to always enforce the use of the embedded CLI now)

What are your thoughts on this?

Thanks again for this PR!

alvaro-shopify:
Hi @karreiro. On the one hand we are finding out little issues in the integration of the embed CLI2 code, like this one. In this case users could use the env variable `SHOPIFY_CLI_BUNDLED_THEME_CLI` as a workaround to solve their problem, but on the other hand I'm completely agree with you that you are adding new features to the embed code and it's quite hard to design them total backwards compatible with the external/bundled CLI2. Apart from that, right now is more difficult to install CLI2 as the homebrew method has been disabled.
So in my hones opinion I would remove the SHOPIFY_CLI_BUNDLED_THEME_CLI env variable in the next releases

isaacroldan (approved):

PR #2139, opened 2023-06-09
Feature: Add Flow template extension in CLI

tommybae:

[FLOW2568](https://github.com/Shopify/flow-app/issues/2568)

Allows users to upload templates via the CLI

`partners` PR: https://github.com/Shopify/partners/pull/48080
`partners-web-platform`: PR: https://github.com/Shopify/partners-web-platform/pull/1198
`extensions-templates` PR: https://github.com/Shopify/extensions-templates/pull/36 (merged)
`shopify` PR : https://github.com/Shopify/shopify/pull/431925

### How to test your changes?
🎩 
 Existing spin instance: https://partners.flow-templates-v6.tomas-baena.us.spin.dev/9082/apps/1602671/extensions

or 

Create new spin instance:

> spin up flow:with-partners-and-extensions -c shopify.branch=cg/flow_templates -c cli.branch=2568/flow-template-ext/tb -c partners.branch=cg/flow_templates -c partners-web-platform.branch=cg/flow_templates

2. Enable beta flag in partners internal: `enable_cli_managed_flow_template_extensions` (Allows flow-templates to be visible in cli generate command)

3. Enable `app_unified_deployment`  beta flag on Flow (local)

4. Add optional flag on dev command in `flow` to include branch of extensions-templates
> [ -d ../cli ] && yarn cli:local app generate extension

becomes

> [ -d ../cli ] && yarn cli:local app generate extension --clone-url https://github.com/Shopify/extensions-templates#flow-templates-template/tb

5. Generate extension via command in `flow`
> dev cli-generate
- Select Shopify org and Flow (local) app

6. Add flow file to extension (ensure the file name is template.flow)
31-15-hsta1-05ufo

7. Deploy extension in `flow` with command `dev cli-deploy`

### Checklist

### DIFF
--- a/packages/app/src/cli/models/extensions/specifications/flow_template.ts
+++ b/packages/app/src/cli/models/extensions/specifications/flow_template.ts
@@ -27,7 +27,6 @@ const FlowTemplateExtensionSchema = BaseSchemaWithHandle.extend({
 const spec = createExtensionSpecification({
   identifier: 'flow_template',
   schema: FlowTemplateExtensionSchema,
-  singleEntryPath: false,
   appModuleFeatures: (_) => ['bundling'],
   deployConfig: async (config, extensionPath) => {
     return {


packages/app/src/cli/models/extensions/specifications/flow_template.ts
```diff
@@ -0,0 +1,42 @@
...
+  singleEntryPath: false,
+  appModuleFeatures: (_) => [],
+  deployConfig: async (config, directory) => {
+    return {
+      title: config.template.title,
+      description: config.template.description,
+      categories: config.template.categories,
+      pre_install_note: config.template.pre_install_note,
+      post_install_note: config.template.post_install_note,
+      definition: await loadWorkflow(directory),
```
cgidzinski, line 32: this dir might not work

packages/app/src/cli/models/templates/generic-specifications/flow_template.ts
```diff
@@ -0,0 +1,27 @@
+import {ExtensionTemplate} from '../../app/template.js'
```
cgidzinski, line 1: this folder no longer exists and needs to be pulled back one level

packages/app/templates/extensions/projects/flow_template/.flow
```diff
@@ -0,0 +1 @@
+Replace this with a .flow file exported from Production.
```
dmcvittie, line 1: Should we change the wording on this? Maybe something like "from your dev shop"

packages/app/src/cli/models/extensions/specifications/flow_template.ts
```diff
@@ -0,0 +1,85 @@
...
+      `More than one .flow file found in ${path} `,
+      'Make sure you have only one .flow file in your extension folder.',
+    )
+  }
+
+  const flowFilePath = flowFilePaths[0]
+  if (flowFilePath) {
+    return fs.readFileSync(flowFilePath, 'base64')
+  }
+}
```
RobertMic, line 83: Small style nit:

```suggestion
async function loadWorkflow(path: string) {
  const flowFilePaths = await glob(joinPath(path, '*.flow'))
  const flowFilePath = flowFilePaths[0]

  if (flowFilePath == null) {
    throw new AbortError(
      `Missing .flow file in ${path} `,
      'Make sure you have built and exported a flow file from a Shop.',
    )
  } else if (flowFilePaths.length > 1) {
    throw new AbortError(
      `More than one .flow file found in ${path} `,
      'Make sure you have only one .flow file in your extension folder.',
    )
  }

  return fs.readFileSync(flowFilePath, 'base64')
}
```

RobertMic (approved):

packages/app/src/cli/models/extensions/specifications/flow_template.ts
```diff
@@ -0,0 +1,85 @@
...
+import {BaseSchema} from '../schemas.js'
+import {createExtensionSpecification} from '../specification.js'
+import {loadLocalesConfig} from '../../../utilities/extensions/locales-configuration.js'
+import {joinPath} from '@shopify/cli-kit/node/path'
+import {zod} from '@shopify/cli-kit/node/schema'
+import {AbortError} from '@shopify/cli-kit/node/error'
+import {glob} from '@shopify/cli-kit/node/fs'
+import fs from 'fs'
+
+const FlowTemplateExtensionSchema = BaseSchema.extend({
```
MaxLeduc, line 11: We have a [PR](https://github.com/Shopify/core-build-north-star-app/pull/9) for our extensions shape in the North Star repo. IDK if you want to add your template extension there and validate the extra fields with DC. There might be a need for alignment there LMK if you need help with that.

packages/app/src/cli/models/extensions/specifications/flow_template.ts
```diff
@@ -0,0 +1,85 @@
...
+      title: extension.title,
+      description: extension.description,
+      categories: extension.categories,
+      pre_install_note: extension.pre_install_note,
+      post_install_note: extension.post_install_note,
+      require_app: extension.require_app,
+      visible: extension.visible,
+      enabled: extension.enabled,
+      localization: await loadLocalesConfig(directory, 'flow_template'),
+      definition: await loadWorkflow(directory),
```
MaxLeduc, line 59: So here you're doing a convention based approach but for both the `flow_trigger` and `flow_action` `schema` property we went for configuration. Basically the `schema` property value is the path to the file. We might need some alignment there @ryanische.

The main pro of configuration was that partners could share their schema between multiple extensions and also be able to use the folder structure that met their need (i.e. the schema didn't need to live side-by-side with the extension config file).

Also was wondering:
- Why is the template file expected to be called `.flow`?
- Could `.flow` be the format (e.g. `email-marketing-template.flow`)

packages/app/templates/extensions/projects/flow_template/shopify.extension.toml.liquid
```diff
@@ -0,0 +1,31 @@
...
+title = "My Awesome Template"
+
+# Required -- Can be overridden by translations, but used in Partner dashboard
+description = "A Template that helps increase merchant productivity"
+
+# Required -- Must include one or more categories
+categories = ["orders", "fraud"]
+
+# Required
+pre_install_note = "You must disable automatic payment capture in the Shopify Admin before using this Template"
```
MaxLeduc, line 19: Should these be localized?

packages/app/src/cli/models/extensions/specifications/flow_template.ts
```diff
@@ -0,0 +1,85 @@
...
+      type: extension.type,
+      title: extension.title,
+      description: extension.description,
+      categories: extension.categories,
+      pre_install_note: extension.pre_install_note,
+      post_install_note: extension.post_install_note,
+      require_app: extension.require_app,
+      visible: extension.visible,
+      enabled: extension.enabled,
+      localization: await loadLocalesConfig(directory, 'flow_template'),
```
MaxLeduc, line 58: How are we handling localization? Where are these saved? Will this be affected by versioning?

MaxLeduc (commented):
Looking good! Had a couple comments and questions. LMK if that makes sense!

packages/app/src/cli/models/extensions/specifications/flow_template.ts
ryanische: This is an interesting case - I'm not sure there would be a need to specify a different location for a template file so I can see the advantage of convention here.

packages/app/src/cli/models/extensions/specifications/flow_template.ts
dmcvittie: I think the check on line 68 of the suggestion needs to include `undefined` for it to work right

packages/app/templates/extensions/projects/flow_template/shopify.extension.toml.liquid
```diff
@@ -0,0 +1,31 @@
+name = "{{name}}"
+description = "A Flow Template example"
+type = "flow_template"
+
+[[extensions]]
+handle = "{{name}}"
+type = "flow_template"
+
+# Required -- Can be overridden by translations, but used in Partner dashboard
```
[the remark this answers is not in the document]
dmcvittie: Yep, we had it in there from the spec discussions earlier. Good callout to remove now.

packages/app/src/cli/models/extensions/specifications/flow_template.ts
dmcvittie: Captured in the [tech design here](https://docs.google.com/document/d/1APLY_7jPyLV43jmcEw5cVwZqI-OGgouuZ5f2QsNIaiQ/edit#heading=h.zcj4v9c5dxie)

TL;DR:
- Synced to Flow with the payload, stored in a blob in the DB
- Latest published version will override the existing one since there is only ever one version of a Template active at once
- Future work will move these to GCS as the needs of translations grow

packages/app/templates/extensions/projects/flow_template/shopify.extension.toml.liquid
dmcvittie: Yep, they are included in the localization paradigm. It's possible that we could remove them from here now and just rely entirely on translations. @cgidzinski what do you think?

packages/app/src/cli/models/extensions/specifications/flow_template.ts
```diff
@@ -0,0 +1,83 @@
...
+      title: zod.string(),
+      description: zod.string(),
+      categories: zod.array(zod.string()),
+      pre_install_note: zod.string(),
+      post_install_note: zod.string(),
+      require_app: zod.boolean(),
+      visible: zod.boolean(),
+      enabled: zod.boolean(),
+    }),
+  ),
```
isaacroldan, line 27: add `.min(1)` here so that `zod` will take care of validating that at least 1 extension exists. You can also say `max(1)`I think.

packages/app/src/cli/models/extensions/specifications/flow_template.ts
RobertMic: `== null` checks for both `undefined` and `null`

packages/app/templates/extensions/projects/flow_template/shopify.extension.toml.liquid
cgidzinski: Its one solution though i feel it could be a better experience if we expose all the possible options instead of just hiding them in the localization file 🤔. That being said if our goal in the end is to translate many things then moving everything there would make sense.

packages/app/src/cli/models/extensions/specifications/flow_template.ts
cgidzinski: I decided to make part of the change now and move the .flow files to the `./templates` folder that way we can just accept more in the future

packages/app/src/cli/models/extensions/specifications/flow_template.ts
```diff
@@ -0,0 +1,82 @@
...
+    const extension = config.extensions[0]
+    if (!extension) {
+      throw new AbortError(`Missing extension in ${directory} `, 'Make sure you have defined at least one extension.')
+    }
+    if (config.extensions.length > 1) {
+      throw new AbortError(
+        `More than one extension found in ${directory} `,
+        'Make sure you have only one extension defined. This will be expanded to support more in the future.',
+      )
+    }
```
dmcvittie, line 47: Do we need this now with the `min`/`max` changes with the zod schema?

MaxLeduc: The following check is nice for typing (since otherwise Typescript considers that `extension` can  be `undefined`)

```
if (!extension) {
  throw new AbortError(`Missing extension in ${directory} `, 'Make sure you have defined at least one extension.')
}
```

but the check for `config.extensions.length > 1` isn't necessary anymore IMO.

cgidzinski: Since its 2 different types of validation, typescript will still yell at us saying extension could be null in this case. It doesn't hurt to have the first extra check even if it is slightly redundant at the moment.

MaxLeduc: Ah you know what I had forgotten that I had removed that check in favour of coercing the typing of `extension` [here](https://github.com/Shopify/cli/pull/2139#discussion_r1254803864).

isaacroldan (changes requested):
The template definition should be added to partners.

packages/app/templates/extensions/projects/flow_template/shopify.extension.toml.liquid
```diff
@@ -0,0 +1,19 @@
...
+
+description = "A Template that helps increase merchant productivity"
+
+categories = ["orders", "fraud"]
+
+require_app = false
+
+visible = true
+
+enabled = true
```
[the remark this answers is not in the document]
tommybae: Hey @isaacroldan, are they added to partners similar to this [commit](https://github.com/Shopify/partners/commit/60b708acc4be312ca53fe9226b58e727071dd7e9) and then we store the toml and other base files in the `extensions-templates` repo?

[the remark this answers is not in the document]
tommybae: Thanks! We're not aiming for the 3.48 release so no worries 👍

isaacroldan:
@tommybae Yes please :)

packages/app/src/cli/models/extensions/specifications/flow_template.ts
```diff
@@ -0,0 +1,60 @@
...
+import fs from 'fs'
+
+const FlowTemplateExtensionSchema = BaseSchemaWithHandle.extend({
+  type: zod.literal('flow_template'),
+  title: zod.string(),
+  description: zod.string(),
+  categories: zod.array(zod.string()),
+  require_app: zod.boolean(),
+  visible: zod.boolean(),
+  enabled: zod.boolean(),
```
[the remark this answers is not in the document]
cgidzinski: Would this not conflict with the base `name` since this is the name of the flow template not the app extension.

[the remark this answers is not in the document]
cgidzinski: We will most likely move all the template based things to a nested level like https://github.com/Shopify/extensions-templates/blob/main/checkout-extension/shopify.extension.toml.liquid#L46 which from what i can see here doesn't care if the names overlap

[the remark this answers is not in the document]
cgidzinski: the first one is the name of the extension, and we will eventually support multiple templates which have their own name/title

packages/app/src/cli/models/extensions/specifications/flow_template.ts
```diff
@@ -9,12 +9,22 @@ import fs from 'fs'
...
+        description: zod.string(),
+        categories: zod.array(zod.string()),
+        require_app: zod.boolean(),
+        listed: zod.boolean(),
+        enabled: zod.boolean(),
+      }),
+    )
+    // technically we'll be able to support more than one template per extension but restricting to one for now
+    .min(1)
+    .max(1),
```
isaacroldan, line 27: this makes more sense 👏 
No need to define `name` as it's already included in the BaseSchema

packages/app/src/cli/models/extensions/specifications/flow_template.ts
```diff
@@ -24,36 +34,31 @@ const spec = createExtensionSpecification({
...
+        }),
+      ),
     }
   },
 })
 
-async function loadWorkflow(path: string, handle: string) {
-  const flowFilePaths = await glob(joinPath(path, '/template.flow'))
+async function loadWorkflow(path: string, key: string) {
+  const flowFilePaths = await glob(joinPath(path, key, `${key}.flow`))
```
isaacroldan, line 57: so each template will be in a different folder?
wouldn't it make more sense to have all the templates in the root of the extension next to the toml file? or in a `templates` folder, but 1 folder per template seems to much

davejcameron: I'd also make this be an explicit config as well, set this path to be the default but allow a developer to specify the file if they want

cgidzinski: By separating everything out it became easier to handle multiple templates as we will ended up sending them one at a time through kafka. The .flow file along with the translations will be in that folder as well.

cgidzinski: We have to separate the translations also to keep size down in the future.

isaacroldan: I thought there was going to be just the `.flow` file, if the plan is to have more files per folder then that's 👍 for me

packages/app/src/cli/models/extensions/specifications/flow_template.ts
```diff
@@ -0,0 +1,62 @@
...
+  templates: zod
+    .array(
+      zod.object({
+        key: zod.string(),
+        name: zod.string(),
+        description: zod.string(),
+        categories: zod.array(zod.string()),
+        require_app: zod.boolean(),
+        discoverable: zod.boolean(),
+        enabled: zod.boolean(),
```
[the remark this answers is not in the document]
cgidzinski: While its still in 1p land we decided to make them all mandatory, we will eventually decide if some like `discoverable` will be optional later down the line

PR #2162, opened 2023-06-13
Add post build action

isaacroldan:

We want to support post build actions with the introduction of Flow extensions since as part of their DevX we want to supply developers with a "payload preview" which will help them interact with Flow's API.

- Adding a `postBuildAction` method to the local extension specification which is trigger after the extension build is completed.
- Adding logic to support the Flow payload previews.

### How to test your changes?

Create a Flow extension and use either `deploy` or `dev` to trigger the `postBuildAction` and the payload preview.

### Checklist

- [] I've made sure that any changes to `dev` or `deploy` have been reflected in the [internal flowchart](https://www.figma.com/file/7vqUp50u6dm48Zfb4JRRn8/CLI3-Internals).


packages/app/src/cli/models/app/loader.ts
```diff
@@ -60,9 +60,7 @@ async function loadConfigurationFile(
     }
   }
   // Convert snake_case keys to camelCase before returning
-  return {
-    ...Object.fromEntries(Object.entries(configuration).map((kv) => [camelize(kv[0]), kv[1]])),
-  }
+  return convertKeysToCamelCase(configuration)
```
isaacroldan, line 63: sorry, I don't think we can do yet. There are extensions we were working around and we'll need to check that those still work

packages/app/src/cli/models/extensions/specifications/flow_action.ts
```diff
@@ -72,6 +73,22 @@ const flowActionSpecification = createExtensionSpecification({
       schema_patch: await loadSchemaPatchFromPath(extensionPath, config.task.schema),
     }
   },
+  postBuildAction: async (extension) => {
```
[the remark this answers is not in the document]
isaacroldan: Yes, it's possible to pass the app info to this function.
It's actually already available at the `build` step, we just need to include it in the signature for the post-build

packages/app/src/cli/models/extensions/extension-instance.ts
```diff
@@ -220,10 +221,13 @@ export class ExtensionInstance<TConfiguration extends BaseConfigType = BaseConfi
       await touchFile(this.outputPath)
       await writeFile(this.outputPath, '(()=>{})();')
     }
+
+    await this.specification.postBuildAction?.(this)
```
isaacroldan, line 225:
```suggestion
    await this.specification.postBuildAction?.(this, options.app)
```
The app is available here ☝️

MaxLeduc:
We made the decision to ship Flow extensions without the payload preview feature and instead work on making our API friendlier for developers. Ok if I close this @isaacroldan ?

PR #2168, opened 2023-06-14
Add version list command

isaacroldan:
- We need to list the existing deployments for the current app

Fixes https://github.com/Shopify/app-deploys/issues/437

- Add a new command `app versions list`
- Fetch the deployments for the current app (taking the apiKey from the .env file), if there is no app present, show prompts to select it.
- Allow to pass an api-key via flags.

### How to test your changes?
- `pnpm shopify app versions list --path /path/to/your/app`

### Measuring impact

### Checklist


mattnoakes:
Looks and works great!!

I have a couple notes from tophatting:
- Unless I run `deploy` once, I'm prompted to connect my local app to a remote app every time I run `versions list`. Is this expected?
- When I have 0 app versions, I get an empty list. Is this what we want to show @jessehoek?
14-45-uy5gb-w6ntj

jessehoek:
> ah! Definitely not ideal, what should we show when there aren't deployments?
Thanks for catching this Matt! We can just output `No app versions found for this app.`

ani-shopify:
Thanks for this PR @isaacroldan! Everything looks great.

I purposefully set the message for one of the app versions pretty long and this is what it looks like:

image


@jessehoek @isaacroldan not sure if we want to address this in any way? Should we truncate the error message? Split it sooner?

isaacroldan:
@mattnoakes That's a good point, do we want to cache the selected app as the "depoyable" app?
So that if you run first `versions list` and select an app, you won't be prompted again when you run `deploy`. Is that OK?

jessehoek:
> That's a good point, do we want to cache the selected app as the "depoyable" app? So that if you run first `versions list` and select an app, you won't be prompted again when you run `deploy`. Is that OK?

@isaacroldan I see how this would be nice for the versions command - but I do think it might be dangerous if you end up deploying to an app just because you ran the `versions list` command first and selected that app. 

I wonder if we can integrate with the new `config use` command here, where we use the API key as defined in the current active config.  These config commands are coming as part of the declarative app config project. @ani-shopify thoughts? If that's getting too out of scope, we could always add it as a follow up ticket?

mattnoakes:
> @mattnoakes That's a good point, do we want to cache the selected app as the "depoyable" app? So that if you run first `versions list` and select an app, you won't be prompted again when you run `deploy`. Is that OK?

I don't really have a strong opinion since, like you say, it's probably an uncommon user flow. I guess if it's easy to implement, then sure, but it doesn't seem like a big deal to leave it as is 🤷

isaacroldan:
>I wonder if we can wrap the text (make the message column a smaller width) so that the other columns don't get squished. I believe we have a max of 200 characters for the message now, so the height of the row shouldn't get crazy big.

Doesn't seem to a limit, just created new versions with 400 character messages...
I'll investigate what options do we have to wrap the text

isaacroldan:
Updated the table to be dynamic and adapt the message size to the terminal size.
If you have a really small terminal, it will still break (we can't do magic).
As a next step after this is merged we can consider showing a shorter format for other fields (date, createdby...) for small terminals.

packages/app/src/cli/commands/app/versions/list.ts
```diff
@@ -0,0 +1,35 @@
...
+export default class VersionsList extends Command {
+  static description = 'List deployed versions of your app.'
+  static examples = ['<%= config.bin %> <%= command.id %>']
+
+  static flags = {
+    ...globalFlags,
+    ...appFlags,
+    'api-key': Flags.string({
+      hidden: false,
+      description: "Application's API key that will be exposed at build time.",
```
alvaro-shopify, line 19: I guess this is not the proper description right?

packages/app/src/cli/services/versions-list.ts
```diff
@@ -0,0 +1,123 @@
...
+// eslint-disable-next-line @typescript-eslint/consistent-type-definitions
+type DeploymentLine = {
+  createdAt: string
+  createdBy?: string
+  message?: string
+  versionTag: string
+  status: string
+}
+
+function formatDate(date: Date) {
```
alvaro-shopify, line 20: could we extract this method to the `cli-kit` `string` utility module?, I mean this kind of date format could be really useful for devs

packages/app/src/cli/commands/app/versions/list.ts
```diff
@@ -0,0 +1,35 @@
...
+      env: 'SHOPIFY_FLAG_API_KEY',
+    }),
```
alvaro-shopify, line 21: I miss the `reset` flag. I have a cached spin running and I'm not able to reset it

packages/app/src/cli/services/versions-list.ts
```diff
@@ -0,0 +1,123 @@
...
+  const timeString = components[1]?.split('.')[0] ?? date.toTimeString()
+  return `${dateString} ${timeString}`
+}
+
+async function fetchDeployments(
+  token: string,
+  apiKey: string,
+): Promise<{deployments: DeploymentLine[]; totalResults: number; organizationId: string; appId: string}> {
+  const query = AppDeploymentsQuery
+  const res: AppDeploymentsQuerySchema = await partnersRequest(query, token, {apiKey})
```
alvaro-shopify, line 32: The API is returning a 500 error if the app does not have the `unified` beta enabled, is this the excepted behaviour?

packages/app/src/cli/services/versions-list.ts
```diff
@@ -0,0 +1,123 @@
...
+  const deployments = res.app.deployments.nodes.map((deployment) => {
```
alvaro-shopify, line 33: If the `appKey` does not exists (for example you deployed the app to spin) the API does not return any error but the `app` is undefined

image

packages/app/src/cli/services/versions-list.ts
```diff
@@ -0,0 +1,123 @@
...
+    }
+  })
+
+  return {
+    deployments,
+    totalResults: res.app.deployments.pageInfo.totalResults,
+    organizationId: res.app.organizationId,
+    appId: res.app.id,
+  }
+}
```
alvaro-shopify, line 81: I miss a test suite for this `version-list`
- version list -> empty content
- version list -> happy path
- version list -> Long comments
- version list -> use api flag if set
- version list -> use env value if set

packages/app/src/cli/commands/app/versions/list.ts
isaacroldan: talked about this offline, `versions list` shouldn't reset your deploy config

packages/app/src/cli/services/versions-list.ts
isaacroldan: expected for now 👍

PR #2171, opened 2023-06-14
Support enabling/disabling developer preview from CLI

alvaro-shopify:

Fixes https://github.com/Shopify/develop-app-foundations/issues/283

IMPORTANT: Disable `dev preview` when the `dev` command finishes not covered in this issue. This is the follow-up: https://github.com/Shopify/cli/pull/2203

- Integrate new `DevelopmentStorePreviewUpdate` partners API operations and add methods to enable/disable `Developer Preview` . In case of error command should not break and continue with the execution
- Enable developer preview when `dev` command starts
- Disable developer preview when `deploy` command starts
- Modify theme app extensions server instructions for enabling developer preview in the dashboard as now depends on having active new unified deployment mode
- Cloudfare tunnel is not created by default in case `--no-tunnel` flag is used

### How to test your changes?
- Create an app with a `theme app extension`

#### App configured without unified
- Run `npm run shopify app dev --verbose`
  - `enabled preview message` should not be visible when the `dev` command starts
- Run `npm run shopify app deploy`
  - `disabled preview message` should not be visible when the `deploy` command starts

#### App configured with unified deployment (active it from internals dashboard)
- Run `npm run shopify app dev --verbose`
  - `enabled preview message` should be visible when the `dev` command starts
- Run `npm run shopify app deploy`
  - `disabled preview message` should be visible when the `deploy` command starts

### Measuring impact

### Checklist

### DIFF
--- a/.changeset/clever-apples-pull.md
+++ b/.changeset/clever-apples-pull.md
@@ -3,4 +3,4 @@
 '@shopify/app': patch
 ---
 
-Support enabling/disabling developer preview from CLI
+Support enabling/disabling development store preview from CLI
--- a/packages/app/src/cli/services/context.test.ts
+++ b/packages/app/src/cli/services/context.test.ts
@@ -377,7 +377,7 @@ describe('ensureDevContext', () => {
     expect(fetchOrgAndApps).toBeCalled()
   })
 
-  test('dev enables automatically the developer preview if the unified deployments beta is enabled', async () => {
+  test('dev enables automatically the development store preview if the unified deployments beta is enabled', async () => {
     // Given
     vi.mocked(getAppInfo).mockReturnValue(undefined)
     vi.mocked(fetchOrgFromId).mockResolvedValueOnce(ORG2)
@@ -402,7 +402,7 @@ describe('ensureDevContext', () => {
     expect(partnersRequest).toHaveBeenCalledWith(DevelopmentStorePreviewUpdateQuery, 'token', {
       input: {apiKey: 'key2', enabled: true},
     })
-    expect(mockOutput.completed()).toMatchInlineSnapshot('"Developer preview enabled"')
+    expect(mockOutput.completed()).toMatchInlineSnapshot('"Development store preview enabled"')
   })
 
   test('display an error to enable dev preview if the beta is enabled in partners but an error is returned', async () => {
@@ -429,7 +429,7 @@ describe('ensureDevContext', () => {
       input: {apiKey: 'key2', enabled: true},
     })
     expect(mockOutput.warn()).toMatchInlineSnapshot(
-      '"Unable to enable developer store preview for this app. You can change this setting in the partners dashboard ( https://partners.shopify.com/1/apps/2/extensions ).\'}"',
+      '"Unable to enable development store preview for this app. You can change this setting in the Partner Dashboard ( https://partners.shopify.com/1/apps/2/extensions ).\'}"',
     )
   })
 })
@@ -604,7 +604,7 @@ describe('ensureDeployContext', () => {
     )
   })
 
-  test('deploy disables automatically the developer preview if the unified deployments beta is enabled', async () => {
+  test('deploy disables automatically the development store preview if the unified deployments beta is enabled', async () => {
     // Given
     const app = testApp()
     const identifiers = {
@@ -635,7 +635,7 @@ describe('ensureDeployContext', () => {
     expect(partnersRequest).toHaveBeenCalledWith(DevelopmentStorePreviewUpdateQuery, 'token', {
       input: {apiKey: 'key2', enabled: false},
     })
-    expect(mockOutput.completed()).toMatchInlineSnapshot('"Developer preview disabled"')
+    expect(mockOutput.completed()).toMatchInlineSnapshot('"Development store preview disabled"')
   })
 
   test('display an error to disable dev preview if the beta is enabled in partners but an error is returned', async () => {
@@ -668,7 +668,7 @@ describe('ensureDeployContext', () => {
       input: {apiKey: 'key2', enabled: false},
     })
     expect(mockOutput.warn()).toMatchInlineSnapshot(
-      '"Unable to disable developer store preview for this app. You can change this setting in the partners dashboard ( https://partners.shopify.com/1/apps/2/extensions ).\'}"',
+      '"Unable to disable development store preview for this app. You can change this setting in the Partner Dashboard ( https://partners.shopify.com/1/apps/2/extensions ).\'}"',
     )
   })
 })
--- a/packages/app/src/cli/services/context.ts
+++ b/packages/app/src/cli/services/context.ts
@@ -631,16 +631,16 @@ async function developerPreviewUpdate(app: OrganizationApp, token: string, enabl
 
         if ((result && result.developmentStorePreviewUpdate.userErrors?.length > 0) || error) {
           const previewURL = outputToken.link(
-            'partners dashboard',
+            'Partner Dashboard',
             await devPreviewURL({orgId: app.organizationId, appId: app.id}),
           )
           outputWarnError(
             outputContent`Unable to ${
               enabled ? 'enable' : 'disable'
-            } developer store preview for this app. You can change this setting in the ${previewURL}.'}`,
+            } development store preview for this app. You can change this setting in the ${previewURL}.'}`,
           )
         } else {
-          outputCompleted(`Developer preview ${enabled ? 'enabled' : 'disabled'}`)
+          outputCompleted(`Development store preview ${enabled ? 'enabled' : 'disabled'}`)
         }
       },
     },
--- a/packages/cli-kit/assets/cli-ruby/lib/shopify_cli/theme/extension/dev_server.rb
+++ b/packages/cli-kit/assets/cli-ruby/lib/shopify_cli/theme/extension/dev_server.rb
@@ -180,7 +180,7 @@ module ShopifyCLI
 
         def enable_extension_message
           if ShopifyCLI::Environment.unified_deployment?
-            "Verify your Development store preview is on"
+            "Verify that the development store preview is turned on for this app"
           else
             "Enable your theme app extension"
           end


ani-shopify:
> - Is this feature totally required?, I mean, what happens if we leave the dev preview mode enabled for an app used to develop?
> - Is it not enough to disable it when we run the deploy command?

I discussed this with @jessehoek earlier today and came to a conclusion if the main concern is there's a risk that dev preview could stay enabled because in some cases we might not detect when the command is ended, then that's fine. It should be enough to disable it as part of the `deploy` command.

@jessehoek is going to create some additional mocks to communicate this to the user.

Is the only concern that dev preview might stay enabled? Or are there overall concerns how this might affect the other running processes?

packages/app/src/cli/services/context.ts
```diff
@@ -151,7 +169,9 @@ export async function ensureDevContext(options: DevContextOptions, token: string
       orgId,
     })
 
-    return buildOutput(selectedApp, selectedStore, useCloudflareTunnels, cachedInfo)
+    await enableDeveloperPreview({app: selectedApp, token})
```
amcaplan, line 172: This seems like a blocking request; my local verbose mode shows about 700ms to execute this. Could we try a fire-and-forget approach to avoid slowing down dev startup time?

packages/app/src/cli/services/context.ts
```diff
@@ -538,3 +566,66 @@ async function logMetadataForLoadedDeployContext(env: DeployContextOutput) {
...
+      enabled,
+    },
+  }
+  const result: DevelopmentStorePreviewUpdateSchema = await partnersRequest(query, token, variables)
+
+  if (
+    result.developmentStorePreviewUpdate.userErrors?.length > 0 ||
+    result.developmentStorePreviewUpdate.app.developmentStorePreviewEnabled !== enabled
+  ) {
+    logger(
```
amcaplan, line 625: This feels like it's more significant than `outputDebug` level. If it worked yesterday and it fails today with no loud error message, that feels like it'll be a bad user experience.

packages/app/src/cli/services/dev.ts
```diff
@@ -84,7 +85,7 @@ async function dev(options: DevOptions) {
   const tunnelPort = await getAvailableTCPPort()
 
   let tunnelClient: TunnelClient | undefined
-  if (!options.tunnelUrl) {
+  if (!options.tunnelUrl && !options.noTunnel) {
```
amcaplan, line 88: Is this just a general fix that made its way into here?

packages/app/src/cli/services/dev.ts
```diff
@@ -280,12 +282,24 @@ async function dev(options: DevOptions) {
...
-  await reportAnalyticsEvent({config: options.commandConfig})
+  const abortController = new AbortController()
+  abortController.signal.addEventListener('abort', async () => {
+    setInterval(() => {
+      outputWarn('Graceful shutdown failed, killing processes ...')
+      treeKill('SIGINT')
+    }, GRACEFUL_EXIT_TIMEOUT)
+    outputNewline()
+    outputNewline()
+    outputInfo('Initiating graceful shutdown ...')
```
amcaplan, line 293: Why not:

```suggestion
    outputInfo('\n\nInitiating graceful shutdown ...')
```

packages/app/src/cli/services/dev.ts
```diff
@@ -280,12 +282,24 @@ async function dev(options: DevOptions) {
 
   await logMetadataForDev({devOptions: options, tunnelUrl: frontendUrl, shouldUpdateURLs, storeFqdn})
 
-  await reportAnalyticsEvent({config: options.commandConfig})
+  const abortController = new AbortController()
+  abortController.signal.addEventListener('abort', async () => {
+    setInterval(() => {
```
amcaplan, line 287: 1. Why `setInterval` not `setTimeout`? Either way it'll only run once because it kills the process, but I think `setTimeout` is more semantically correct.
2. Shouldn't we capture the timeout and cancel it if we shut down gracefully? And if we will always run the function because we can't be sure, we shouldn't display a message indicating that "graceful shutdown failed"

packages/app/src/cli/services/dev.ts
alvaro-shopify: @amcaplan , yes, I could create a different PR for this little fix. When you run the `dev` command using the `--not-tunnel` flag, a tunnel is created anyway

packages/app/src/cli/services/dev.ts
alvaro-shopify: > Why setInterval not setTimeout? Either way it'll only run once because it kills the process, but I think setTimeout is more semantically correct.

👍

> Shouldn't we capture the timeout and cancel it if we shut down gracefully? And if we will always run the function because we can't be sure, we shouldn't display a message indicating that "graceful shutdown failed"

yep, we should cancel it otherwise the `graceful shutdown` will never finish in less than the `GRACEFUL_EXIT_TIMEOUT` 
Regarding not adding `graceful shutdown failed` we could add it in the `--verbose` mode, just to have the chance to check it in case we need to (although it's true that we will see the exit(1)`

packages/app/src/cli/services/context.ts
alvaro-shopify: do you mean that the API has failed? or that the CLI try to enable it and it was already `enabled` remotely?  (same applies for disable). If the latest maybe we could remove the log message because it's not an error itself

alvaro-shopify:
> > * Is this feature totally required?, I mean, what happens if we leave the dev preview mode enabled for an app used to develop?
> > * Is it not enough to disable it when we run the deploy command?
> 
> I discussed this with @jessehoek earlier today and came to a conclusion if the main concern is there's a risk that dev preview could stay enabled because in some cases we might not detect when the command is ended, then that's fine. It should be enough to disable it as part of the `deploy` command.
> 
> @jessehoek is going to create some additional mocks to communicate this to the user.
> 
> Is the only concern that dev preview might stay enabled? Or are there overall concerns how this might affect the other running processes?

Hi @ani-shopify.

> if the main concern is there's a risk that dev preview could stay enabled because in some cases we might not detect when the command is ended, then that's fine.

Yes, that's the main concern, what happens if the `dev preview` remains enabled when the `dev` command finishes. What we are going to do is split this PR into two:
* PR1 (target version 3.48)
- `enabled` dev preview when the `dev` command starts
- `disable` dev preview when the `deploy` command starts
- Update `theme app extension` instructions to point to the proper site to enable dev preview if unified enabled

* PR2 (to be discussed for future or we could discard it immediately if we want)
- `disable` dev preview when the `dev` command finished

alvaro-shopify:
Hi @ani-shopify server is returning an unexpected error when we try to disable the `dev preview`. I have posted the problem  in this [thread](https://shopify.slack.com/archives/C03KJRURZKR/p1687193542209399). But that makes me wonder what should we do in case there is an error when the `dev` command tries to enable and `deploy` disable, should the CLI abort the command or continue and shows any kind of error message?

packages/app/src/cli/services/context.ts
alvaro-shopify: yesterday I made [that question](https://github.com/Shopify/cli/pull/2171#issuecomment-1597502753) to @ani-shopify :) . We are not enabling it right now so I guess if this automatic action fails we should go on in the same way than previous to this feature.

amcaplan: @isaacroldan I think it'll complete quickly enough that there's little need to stop the world over it.

If we want to fail more aggressively, we could put in an `await` later. But I think it's like the cloudflare tunnel, we can continue optimistically at first, rather than waiting nearly a second for this API call to complete.

packages/app/src/cli/services/dev.ts
alvaro-shopify: Disable `dev preview` has been removed from this PR, so everything related with using `abortController` vs `treeKill` does not apply anymore

jessehoek:
I agree with that approach @ani-shopify. A lot of extensions can be previewed just through the custom preview (which doesn't care about this toggle) so we shouldn't abort if it fails.

ani-shopify:
> PR2 (to be discussed latter or we could discard it immediately if we want)
> 
> * `disable` dev preview when the `dev` command finished

Added [an issue](https://github.com/Shopify/app-deploys/issues/576) in the CLI epic

alvaro-shopify:
> > Hi @ani-shopify server is returning an unexpected error when we try to disable the dev preview. I have posted the problem in this [thread](https://shopify.slack.com/archives/C03KJRURZKR/p1687193542209399). But that makes me wonder what should we do in case there is an error when the dev command tries to enable and deploy disable, should the CLI abort the command or continue and shows any kind of error message?
> 
> It's a [known bug](https://github.com/Shopify/app-deploys/issues/513) that @gordonhirsch is working on. But generally if the mutation comes back with an error, I'd recommend to not kill the whole `dev` process, instead we could show an warning/error message and tell the users to manually enable the dev preview from the dashboard.
> 
> @jessehoek @gordonhirsch what do you guys think?

Added this error message in case enable/disable the `dev preview` automatically` fails. There is a link to the section in the partners dashboard to change the `dev preview mode`

image

packages/app/src/cli/services/context.test.ts
```diff
@@ -545,6 +609,74 @@ describe('ensureDeployContext', () => {
       'token',
     )
   })
+
+  test('disables developer preview if the beta is enabled in partners', async () => {
```
isaacroldan, line 613: do you mean, "disabled in partners"?

alvaro-shopify: no. I mean that when you run deploy with the new unified beta enabled the developer preview is automatically disabled. 
But now that I'm reading the title carefully I'm not able to understand that properly 😅

alvaro-shopify: but `enables developer preview if the beta is enabled in partners` in under different `describe` so if you run the tests you will see.

``` 
✓ src/cli/services/context.test.ts > ensureDevContext > enables developer preview if the beta is enabled in partners
✓ src/cli/services/context.test.ts > ensureDeployContext > disables developer preview if the beta is enabled in partners
```

should that title be enough to distinguish both of them?

alvaro-shopify: Ok, I will modify the title to something similar to this

```
✓ src/cli/services/context.test.ts > ensureDevContext > dev enables developer preview if the beta is enabled in partners
✓ src/cli/services/context.test.ts > ensureDeployContext > deploy disables developer preview if the beta is enabled in partners
```

packages/app/src/cli/services/context.ts
```diff
@@ -585,9 +598,65 @@ async function logMetadataForLoadedDeployContext(env: DeployContextOutput) {
...
+          }
+          result = await partnersRequest(query, token, variables)
+          // eslint-disable-next-line no-catch-all/no-catch-all, @typescript-eslint/no-explicit-any
+        } catch (err: any) {
+          error = err.message
+        }
+
+        if ((result && result.developmentStorePreviewUpdate.userErrors?.length > 0) || error) {
+          const previewURL = outputToken.link(
+            'partners dashboard',
```
jessehoek, line 634: This is how we should refer to the partner dashboard externally 😄 
```suggestion
            'Partner Dashboard',
```

packages/app/src/cli/services/context.ts
```diff
@@ -585,9 +598,65 @@ async function logMetadataForLoadedDeployContext(env: DeployContextOutput) {
...
+            await devPreviewURL({orgId: app.organizationId, appId: app.id}),
+          )
+          outputWarnError(
+            outputContent`Unable to ${
+              enabled ? 'enable' : 'disable'
+            } developer store preview for this app. You can change this setting in the ${previewURL}.'}`,
```
jessehoek, line 640: Important to be precise in the language of this setting :D
```suggestion
            } development store preview for this app. You can change this setting in the ${previewURL}.'}`,
```

packages/cli-kit/assets/cli-ruby/lib/shopify_cli/theme/extension/dev_server.rb
```diff
@@ -158,12 +158,31 @@ def frame_title
...
+
+          parts = location.split("/")
+          index = parts.index("extensions")
+
+          parts[0..index].join("/")
+        end
+
+        def enable_extension_message
+          if ShopifyCLI::Environment.unified_deployment?
+            "Verify your Development store preview is on"
```
jessehoek, line 183:
```suggestion
            "Verify that development store preview is turned on for this app"
```

packages/app/src/cli/services/context.ts
```diff
@@ -585,9 +598,65 @@ async function logMetadataForLoadedDeployContext(env: DeployContextOutput) {
...
+          )
+        } else {
+          outputCompleted(`Developer preview ${enabled ? 'enabled' : 'disabled'}`)
```
jessehoek, line 643:
```suggestion
          outputCompleted(`Development store preview ${enabled ? 'enabled' : 'disabled'}`)
```

.changeset/clever-apples-pull.md
```diff
@@ -0,0 +1,6 @@
+---
+'@shopify/cli-kit': patch
+'@shopify/app': patch
+---
+
+Support enabling/disabling developer preview from CLI
```
jessehoek, line 6:
```suggestion
Support enabling/disabling development store preview from CLI
```

packages/app/src/cli/services/context.ts
```diff
@@ -585,9 +598,65 @@ async function logMetadataForLoadedDeployContext(env: DeployContextOutput) {
...
+export async function disableDeveloperPreview(app: OrganizationApp, token: string) {
+  return developerPreviewUpdate(app, token, false)
+}
+
+async function developerPreviewUpdate(app: OrganizationApp, token: string, enabled: boolean) {
+  if (!app.betas?.unifiedAppDeployment) return
+
+  const tasks = [
+    {
+      title: `${enabled ? 'Enabling' : 'Disabling'} developer preview...`,
```
jessehoek, line 614:
```suggestion
      title: `${enabled ? 'Enabling' : 'Disabling'} development store preview...`,
```

jessehoek (changes requested):
Can we make sure the language everywhere is `development store preview` not `developer preview`?

PR #2208, opened 2023-06-20
[wip] Implement `app config use`

rachel-carvalho:

- Closes https://github.com/Shopify/internal-cli-foundations/issues/643

Store a current configuration file name in local storage so that commands that use a `config` flag can read from it.

### Measuring impact

### Checklist


packages/app/src/cli/prompts/config.ts
```diff
@@ -25,6 +31,21 @@ export async function selectConfigName(directory: string, defaultName = ''): Pro
...
+
+  if (files.length < 2) return files[0]
+
+  return renderSelectPrompt({
+    message: 'Configuration file',
+    choices: files.map((file) => {
+      return {label: file, value: file}
+    }),
+  })
+}
```
Arkham, line 47: could we use https://github.com/Shopify/cli/blob/main/packages/cli-kit/src/public/node/result.ts so that the error case becomes clearer? happy to pair on this :)

packages/app/src/cli/prompts/config.ts
```diff
@@ -25,6 +31,21 @@ export async function selectConfigName(directory: string, defaultName = ''): Pro
   return configName
 }
 
+export async function selectConfigFile(directory: string): Promise<string | undefined> {
+  const files = (await promises.readdir(directory)).filter(
+    (file) => file.startsWith('shopify.app.') && file.endsWith('.toml'),
+  )
```
Arkham, line 37: I think we should be able to use `glob` (https://github.com/Shopify/cli/blob/main/packages/cli-kit/src/public/node/fs.ts#L473) here and write something like `await glob(path.join(directory, 'shopify.app*.toml'))`

packages/app/src/cli/services/local-storage.ts
```diff
@@ -17,13 +18,27 @@ export interface AppLocalStorageSchema {
...
+    _appLocalStorageInstance = new LocalStorage<AppLocalStorageSchema>({projectName: 'shopify-cli-app'})
+  }
+  return _appLocalStorageInstance
+}
+
+function currentConfigLocalStorage() {
+  if (!_currentConfigLocalStorageInstance) {
+    _currentConfigLocalStorageInstance = new LocalStorage<CurrentConfigLocalStorageSchema>({
+      projectName: 'shopify-cli-app',
+    })
```
Arkham, line 39: would these two storages be stored in the same json?

packages/app/src/cli/commands/app/config/use.ts
```diff
@@ -10,9 +12,20 @@ export default class ConfigUse extends Command {
...
+    reset: Flags.boolean({
+      hidden: false,
+      description: 'Reset current configuration.',
+      env: 'SHOPIFY_FLAG_RESET',
+      default: false,
+    }),
+  }
+
+  static args = {
+    config: Args.string(),
```
Arkham, line 24: we want to unify this arg with the one in `config push`, maybe extract to a separate module, and eventually move it to `appFlags`.

Arkham (commented):
Left some comments, happy to pair on stuff later today :)

packages/app/src/cli/prompts/config.ts
rachel-carvalho: Oh I didn't know we had a type for Result here too, happy to use it 👍

packages/app/src/cli/services/local-storage.ts
rachel-carvalho: We actually had the question of how `projectName` was used. If it's to determine the name of the file, I'll change it, so that keys don't collide between the two stores.

packages/app/src/cli/commands/app/config/use.ts
rezaansyed: +1 to Rachel's point. I don't think a `--config` flag makes sense for `use` because of cases like: `shopify app config use staging --config production`. It adds more confusion.

Arkham (approved):
LGTM :+1:

PR #2222, opened 2023-06-22
rfc: define legacy & new app schema via union

ryancbahan:

This PR implements a set of utils to distinguish between legacy and current app schemas, and updates the requisite commands.

### How to test your changes?

- Run `shopify app env show`, `shopify app info` on legacy toml file shape. 
- Confirm `dev` and `deploy` still work
- Run `link` and `push` on new app shape

### Measuring impact

### Checklist

### DIFF
--- a/packages/app/src/cli/services/app/config/link.ts
+++ b/packages/app/src/cli/services/app/config/link.ts
@@ -68,13 +68,6 @@ async function loadRemoteApp(localApp: AppInterface, apiKey: string | undefined)
   return app
 }
 
-interface NewAppConfig {
-  client_id?: string
-  name?: string
-  application_url?: string
-  webhook_api_version?: string
-}
-
 function mergeAppConfiguration(localApp: AppInterface, remoteApp: OrganizationApp): AppConfiguration {
   if (localApp?.configuration) {
     const configuration = localApp.configuration as CurrentAppConfiguration


packages/app/src/cli/services/app/config/push.ts
```diff
@@ -11,37 +11,45 @@ export interface Options {
...
-    abort(`${configFileName} does not contain a client_id.`)
-  }
-  const initialVariables = {
-    apiKey: configuration.client_id,
-    title: configuration.name,
-    applicationUrl: configuration.application_url,
-    redirectUrlAllowlist: configuration.redirect_url_allowlist,
-    requestedAccessScopes: configuration.requested_access_scopes,
-  }
+  if ('client_id' in configuration) {
```
ryancbahan, line 28: for values not contained in both app schema members, we'll need to always use [narrowing](https://www.typescriptlang.org/docs/handbook/2/narrowing.html)

"in" is a convenient way to do this

packages/app/src/cli/models/app/app.ts
```diff
@@ -6,17 +6,28 @@ import {getDependencies, PackageManager, readAndParsePackageJson} from '@shopify
...
   client_id: zod.string().optional(),
-  scopes: zod.string().default(''),
   application_url: zod.string().optional(),
   redirect_url_allowlist: zod.array(zod.string()).optional(),
   requested_access_scopes: zod.array(zod.string()).optional(),
 })
 
+export const AppConfigurationSchema = zod.union([legacyAppSchema, appSchema])
+
+export function isType<T extends zod.ZodTypeAny>(schema: T, item: unknown): item is zod.infer<T> {
```
ryancbahan, line 27: we could also define more specific helpers to match a particular type, if preferred. advantage to that is it's less imports in the implementation since you won't need to implement the particular schema

ryancbahan: hmm yeah, this function is asking the question "is this item this specific type" but maybe we want an updated or separate function for "treat it as new config or old config, throw if it's invalid"... or something that generally considers validity

ryancbahan: our three branches for app schema will be "is it legacy" "is it new" or "is it borked" and we might want something easy to clearly distinguish those, vs. (or in addition to) a helper of "is it this type" <-- is this what you're getting at?

packages/app/src/cli/models/app/app.ts
```diff
@@ -1,22 +1,102 @@
...
+  options?: SchemaValidationOptions,
+): item is zod.infer<typeof AppConfigurationSchema> {
+  return isLegacyAppSchema(item, {strict: options?.strict}) || isCurrentAppSchema(item, {strict: options?.strict})
+}
+
+/**
+ * Get scopes from a given app.toml config file.
+ * @param config - a configuration file
+ */
+export function getAppScopes(config: unknown) {
```
ryancbahan, line 76: I think we'll ant to talk about future formatting of scopes since we have 2 options for types now, but for now we are still treating implementation as expecting a string

packages/app/src/cli/models/app/app.ts
```diff
@@ -1,22 +1,102 @@
...
+  extension_directories: zod.array(zod.string()).optional(),
+  web_directories: zod.array(zod.string()).optional(),
+  scopes: zod.string().default(''),
+})
+
+export const AppSchema = zod.object({
   extension_directories: zod.array(zod.string()).optional(),
   web_directories: zod.array(zod.string()).optional(),
   name: zod.string().optional(),
   client_id: zod.string().optional(),
```
rezaansyed, line 21: This would need to be required right?

ryancbahan: yeah this was example config, was gonna update in a separate pr but we can do that here

packages/app/src/cli/services/app/config/link.ts
```diff
@@ -68,13 +68,20 @@ async function loadRemoteApp(localApp: AppInterface, apiKey: string | undefined)
   return app
 }
 
+interface NewAppConfig {
```
ryancbahan, line 71: I don't love the updated typing of `mergeAppConfiguration` and am happy to fix forward on it. It's kinda weird because we're not really sure if it's valid app config when we're performing the merge (bc we might be creating something new, which won't have the mandatory fields), but want to be sure in the return.

packages/app/src/cli/services/app/config/link.ts
```diff
@@ -68,13 +68,20 @@ async function loadRemoteApp(localApp: AppInterface, apiKey: string | undefined)
...
+
 function mergeAppConfiguration(localApp: AppInterface, remoteApp: OrganizationApp): AppConfiguration {
-  const configuration = localApp.configuration || {}
+  const configuration: NewAppConfig = (localApp.configuration as unknown) || {}
 
   configuration.client_id = remoteApp.apiKey
   configuration.name = remoteApp.title
   configuration.application_url = remoteApp.applicationUrl
-  configuration.redirect_url_allowlist = remoteApp.redirectUrlWhitelist
+  configuration.webhook_api_version = '2023-04'
```
rezaansyed, line 84: This should be: `configuration.webhook_api_version = remoteApp.webhookApiVersion`?

Arkham: I think a bunch of fields are missing from the `OrganizationApp` data type, so it's currently missing

packages/app/src/cli/services/app/config/push.ts
```diff
@@ -12,48 +13,68 @@ export interface Options {
...
+  const variables = {
+    // these values are mandatory, so we only read from the config file
     apiKey: configuration.client_id,
     title: configuration.name,
     applicationUrl: configuration.application_url,
-    redirectUrlAllowlist: configuration.redirect_url_allowlist,
-    requestedAccessScopes: configuration.requested_access_scopes,
+    contactEmail: configuration.api_contact_email,
+    webhookApiVersion: configuration.webhook_api_version,
+    // these values are optional, so we fall back to configured values
```
ryancbahan, line 56: this behavior will get updated

packages/app/src/cli/services/app/config/link.ts
```diff
@@ -69,12 +69,23 @@ async function loadRemoteApp(localApp: AppInterface, apiKey: string | undefined)
...
 function mergeAppConfiguration(localApp: AppInterface, remoteApp: OrganizationApp): AppConfiguration {
-  const configuration = localApp.configuration || {}
+  if (localApp?.configuration) {
+    const configuration = localApp.configuration as CurrentAppConfiguration
 
-  configuration.client_id = remoteApp.apiKey
-  configuration.name = remoteApp.title
-  configuration.application_url = remoteApp.applicationUrl
-  configuration.redirect_url_allowlist = remoteApp.redirectUrlWhitelist
+    configuration.client_id = remoteApp.apiKey
```
ryancbahan, line 75: this needs to be updated to include the full schema, but i wanted to move `use` and `link` updates to a new pr

packages/app/src/cli/services/info.ts
```diff
@@ -204,7 +204,9 @@ class AppInfo {
 
   accessScopesSection(): [string, string] {
     const title = 'Access Scopes in Root TOML File'
-    const lines = this.app.configuration.scopes.split(',').map((scope) => [scope])
+    const lines = getAppScopes(this.app.configuration)
```
RyanDJLee, line 207: Can we add support and test coverage for '' in legacy TOML being interpreted as []

RyanDJLee (approved):
This is awesome!
Just left a comment about handling the empty scope set in legacy TOML

PR #2225, opened 2023-06-23
Release validate version tag

alvaro-shopify:
Fixes https://github.com/Shopify/app-deploys/issues/581

- Validates the version flag
- Returns a different error message depending on the error type:
  - Max length 100
  - Valid characters characters are 'a-z' 'A-Z' '0-9' '.' '-' '_' and '/'
  - Strings '.' and '..' not supported

### How to test your changes?
- Valid version. Run `p shopify app release --path /path/to/your/app --version Valid-version_with.supported/characters` 

- Invalid length. Run `p shopify app release --path /path/to/your/app --version aaaaassdasdasdasdasdasdasdasdasdasdasdasdasdasdasdasdasdasdasdasdasdasdasda` 

- Invalid characters. Run `p shopify app release --path /path/to/your/app --version "Invalid$version@with(supported)characters"`

- Invalid periods. Run `p shopify app release --path /path/to/your/app --version "."` `p shopify app release --path /path/to/your/app --version ".."` 

### Measuring impact

### Checklist


packages/app/src/cli/services/release.ts
```diff
@@ -79,3 +81,31 @@ export async function release(options: ReleaseOptions) {
     })
   }
 }
+
+export function validateOptions({version}: Partial<ReleaseOptions>) {
+  if (version) {
+    const versionMaxLength = 100
+    if (version.length > versionMaxLength) {
+      throw new AbortError({
+        bold: `Version must be less than ${versionMaxLength} characters`,
```
jessehoek, line 90:
```suggestion
        bold: `Version name must be less than ${versionMaxLength} characters`,
```

jessehoek:
In the release command, this is actually an input - to fetch an existing version with the provided version name.

So shouldn't we only validate on `app deploy --version=id`? @alvaro-shopify @ani-shopify

ani-shopify:
> In the release command, this is actually an input - to fetch an existing version with the provided version name. So why would we validate if this is not a user **setting** a new version name. It's only **fetching** an existing version to release.
> 
> So shouldn't we only validate on `app deploy --version=id`? @alvaro-shopify @ani-shopify

Good catch! Yes, we need to add these validations to the `deploy` command.

amcaplan:
I moved the validations to the `deploy` command; it's probably enough to allow them on the `release` command and let them not be found.

I also made some updates to the format of the error, screenshots attached.

Screenshot 2023-06-25 at 15 11 26
Screenshot 2023-06-25 at 15 11 30
Screenshot 2023-06-25 at 15 11 38

Let me know what you think, especially @jessehoek

jessehoek:
Thanks @amcaplan these messages look great. Good call on including the version name in the message. 

For the <= 100 character validation, we should probably change the wording to say `Version name must be 100 characters or less` since the limit is inclusive of 100.

PR #2226, opened 2023-06-23
[Feature] Add support for function localization

davejcameron:

Adds support for defining a `locales` directory in your app to supply localized strings. Use `t:` prefix for your titles and description field (if applicable) in your `.toml` file to use the localized strings.

Leveraging the existing helpers for localization.

### Measuring impact

### Checklist


packages/app/src/cli/models/extensions/specifications/function.test.ts
```diff
@@ -90,6 +90,8 @@ describe('functionConfiguration', () => {
         },
         enable_creation_ui: true,
         module_id: moduleId,
+        localization: {},
```
isaacroldan, line 93: is it ok to send an empty object or do we want to not send the localization key at all if there aren't localization files?

davejcameron: This looks like the default for other extensions to send an empty `{}` once the other changes are merged this should be fine.

isaacroldan:
@davejcameron should we merge this?

PR #2238, opened 2023-06-26
Reuse CachedAppInfo interface to switch between legacy and config in code

rezaansyed:

Resolves: https://github.com/Shopify/internal-cli-foundations/issues/761

Makes it so that `app dev` reads from the chosen configuration file. Also accepts a new `--config` option to choose another config.

### How to test your changes?

- `pnpm shopify app config link`, create a new app and call the config `production`
- `pnpm shopify app config link`, create a new app and call the config `staging`
- `pnpm shopify app dev` should pick up the staging config
- make a change to the config tile
- `pnpm shopify app config push`
- now make a change to `shopify.app.production.toml`
- `pnpm shopify app config push --config production`

### Measuring impact

### Checklist

### DIFF
--- a/packages/app/src/cli/services/context.ts
+++ b/packages/app/src/cli/services/context.ts
@@ -44,7 +44,6 @@ import {
 import {getOrganization} from '@shopify/cli-kit/node/environment'
 import {writeFileSync} from '@shopify/cli-kit/node/fs'
 import {encodeToml} from '@shopify/cli-kit/node/toml'
-import {basename} from '@shopify/cli-kit/node/path'
 import {partnersRequest} from '@shopify/cli-kit/node/api/partners'
 
 export const InvalidApiKeyErrorMessage = (apiKey: string) => {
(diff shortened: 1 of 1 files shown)


packages/app/src/cli/models/app/app.ts
```diff
@@ -19,11 +19,11 @@ const LegacyAppSchema = zod
...
   .object({
     name: zod.string(),
-    api_contact_email: zod.string(),
     client_id: zod.string(),
     scopes: zod.string().optional(),
-    webhook_api_version: zod.string(),
-    application_url: zod.string(),
+    api_contact_email: zod.string().optional(),
+    webhook_api_version: zod.string().optional(),
+    application_url: zod.string().optional(),
```
rezaansyed, line 26: Hey @Arkham, what's the reason for making these optional?

Arkham: we'd need to pass them everywhere, even in tests. what makes it super annoying is that if you forget to pass one of those the test will fail with no indication that it was the cause of failure.

rezaansyed: But if they are required fields as part of app config upstream, should that not be reflected?

Arkham: are they required? I thought all fields were optional except for `client_id`

Arkham: let's address this in a follow-up PR, I just wanted to quickly unblock myself :)

Arkham (approved):
LGTM

PR #2269, opened 2023-06-29
Watch all non-hidden GrahpQL files in functions

jbourassa:

Currently, functions dev mode only watches for GraphQL files starting with `input.*`. Because we don't know how people will name their GraphQL files, it's best not to force a `input.*` prefix. The main reason for this prefix is because our current Rust crate touches a hidden GraphQL file which sends the CLI in a build loop (rebuild triggers a file change, which triggers a rebuild, rinse & repeat).

Instead of using a prefix, this commit fixes that by globbing on all non-hidden graphql file.

### How to test your changes?

Run `shopify app dev` with a function:
- change any graphql file, notice the CLI updates the draft
- change a rust file (which will touch a hidden file) and notice the CLI does not fall in loop.

### DIFF
--- a/packages/app/src/cli/models/extensions/extension-instance.test.ts
+++ b/packages/app/src/cli/models/extensions/extension-instance.test.ts
@@ -36,7 +36,7 @@ describe('watchPaths', async () => {
 
     const got = extensionInstance.watchPaths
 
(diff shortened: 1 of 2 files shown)


jbourassa:
>If the changes are user-facing, run "pnpm changeset add" to track your changes and include them in the next release CHANGELOG.

This feels closer to a bugfix so I'd skip the changelog entry, but can add one if someone feels strongly about it.

nickwesselman (approved):

PR #2278, opened 2023-06-30
Only load app configuration instead of loading the whole app

Arkham:

We used to use the loader `load` interface to load the whole app even when we only just needed to load the app configuration. This caused bugs when we skipped on passing specifications that required workarounds like [this](https://github.com/Shopify/cli/pull/2267).

- Extract the app configuration loading behaviour into a separate class
- Make it so the AppLoader uses the new AppConfigurationLoader
- Change all the callsites where we were calling `load` unnecessarily and use `loadAppConfiguration` instead.

### Measuring impact

### Checklist


isaacroldan (approved):
🎩 'ed and working great 👌

PR #2284, opened 2023-07-01
[Temp Fix] Use `handle` as context if available on extension deployConfig

omnisyle:

Resolves https://github.com/Shopify/flow/issues/19493
- Flow needs to use `handle` to simplify the trigger method for Flow triggers.
- This change is temporary until the `handle` config spec is fully developed.

- Temporarily pass `handle` from `deployConfig` to `context`.

### How to test your changes?

- Generate a Flow Action/Trigger extension and use `handle` in the config TOML.

### Measuring impact

### Checklist

### DIFF
--- a/packages/app/src/cli/models/extensions/extension-instance.ts
+++ b/packages/app/src/cli/models/extensions/extension-instance.ts
@@ -270,7 +270,14 @@ export class ExtensionInstance<TConfiguration extends BaseConfigType = BaseConfi
   async bundleConfig({identifiers, token, apiKey, unifiedDeployment}: ExtensionBundleConfigOptions) {
     const configValue = await this.deployConfig({apiKey, token, unifiedDeployment})
     if (!configValue) return undefined
-    return {uuid: identifiers.extensions[this.localIdentifier]!, config: JSON.stringify(configValue), context: ''}
+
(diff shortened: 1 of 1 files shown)


packages/app/src/cli/models/extensions/extension-instance.ts
```diff
@@ -270,7 +270,14 @@ export class ExtensionInstance<TConfiguration extends BaseConfigType = BaseConfi
...
     if (!configValue) return undefined
-    return {uuid: identifiers.extensions[this.localIdentifier]!, config: JSON.stringify(configValue), context: ''}
+
+    const contextValue = (configValue.handle as string) || ''
+
+    return {
+      uuid: identifiers.extensions[this.localIdentifier]!,
+      config: JSON.stringify(configValue),
+      context: contextValue,
+    }
```
isaacroldan, line 280: We should send an stringified json object instead of a string, to make it future proof in case we want to add more data here later.

omnisyle: this is a temporary measure so I think we can use string for now. When the proper `handle` implementation is out, this will be gone.

ryanische (approved):

isaacroldan (approved):

PR #2285, opened 2023-07-01
Revert Shopify/shopify-cli#2078

### LINKED ISSUE #2240: [Bug]: Really odd behaviour with image_url output as 'data-...' attribute on Shopify theme dev server

Theme

### Expected behavior

```
{% liquid
  assign url1 = image | image_url: width: settings.max_image_size
  assign url2 = image | image_url: width: image_size
%}
<a
  href="{{ url1 | escape }}"
  class="thumbnail"
  data-img="{{ url2 | escape }}"
  data-width="{{ image.width }}"
  data-height="{{ image.height }}">
  {%- render "image" with image: image, width: 60, placeholder: true, loading: "eager" -%}
</a>
```

Something bizarre is happening with this code when I try and test it locally.

The href tag (url1) will become `//{my-shop-url}//cdn/shop/products/image.jpg?width=1280`
However the data-img tag (url2) becomes `http://127.0.0.1:9292/cdn/shop/products/image.jpg?width=720`

This url if I try to load it doesn't work, as it appears no CDN urls work in the local test environment and have to loaded from the test site directly.

Yet if I simply print url2 outside of an attribute tag it correctly prints `//{my-shop-url}//cdn/shop/products/image.jpg?width=720`

What appears to be happening is the liquid processor, after substituting all liquid tags with their resulting outputs, then runs a kind of hacky 'find and replace' on all strings beginning with http://127.0.01, seemingly in an effort to conquer this bug mentioned in these issues (one of which I reported):
https://github.com/Shopify/cli/pull/2105
https://github.com/Shopify/cli/issues/1859
But for some reason it specifically ignores 'data-...' attributes. Using any attribute name other than 'data-...', the url is correctly output.

There seems to be some very questionable code maintenance happening behind the scenes, it all seems very hacky to me, not at all intuitive.

### Actual behavior

-

### Verbose output

```shell
-
```

### Reproduction steps

-

### Operating System

Windows 11

3.46.5

### Shell

Windows Command Prompt

18.2.0

_No response_

montalvomiguelo:

Fixes #2240 

This PR reverts the patch patch_body method since it's not needed anymore to fix CORS issues. Shopify/shopify-cli#2078

To retrieve sections asynchronously without CORS issues is preferred to use the current domain per Shopify/dawn#848 

### Measuring impact

### Checklist


karreiro (commented):
Thanks a lot for this PR, @montalvomiguelo!

I've tried the reproduction steps from the [previous PR](https://github.com/Shopify/shopify-cli/pull/2078), and noticed that removing `patch_body` would reintroduce the previous issue:

demo

Here's the steps I've followed:

- Clone the Dawn theme using `git clone git@github.com:Shopify/dawn.git`
- Switch to version 1.1.0 with `git reset 2566749e4226818500809892259b7713808319ae --hard`
- Import product data from [products.csv](https://gist.githubusercontent.com/karreiro/2cc4724245a11e2549cc7e6711def453/raw/a11c91bc0cab186af910cbd5126de24102597b34/products_export.csv) into your store
- Run `shopify theme dev`
- Interact with the Classic Varsity Top and observe the browser console

Therefore, perhaps instead of removing `patch_body`, we could adjust the regex to make it more specific to the `data-base-url` attribute. This way, we could handle both https://github.com/Shopify/shopify-cli/issues/1466 and https://github.com/Shopify/cli/issues/2240. What do you think?

Thanks again for this PR 🚀

montalvomiguelo:
Hi @karreiro (: 

It seems like the CORS issue persisted until Dawn v2.3.0 as reported in Shopify/dawn#548. Requesting the resources from the same origin solved the CORS issue per the work in Shopify/dawn#848. Prolly patching the data attributes is not needed anymore.

karreiro:
Hey @montalvomiguelo,

That's accurate indeed, but partners who have to deal with outdated Dawn themes would face that regression. With that in mind, I wonder if replacing the url only on `data-base-url` attributes would be the ideal approach to support both cases. What do you think from that perspective?

Thanks again for the awesome PR :) 🚀

karreiro (approved):
Thank you, @montalvomiguelo! 🔥🚀

PR #2335, opened 2023-07-03
Exit from `pre`

matteodepalo:
Exit from `pre` version one last time.

### DIFF
--- a/.changeset/pre.json
+++ b/.changeset/pre.json
@@ -1,5 +1,5 @@
 {
-  "mode": "pre",
+  "mode": "exit",
   "tag": "pre",
   "initialVersions": {
     "@shopify/app": "3.47.0",


.changeset/pre.json
gonzaloriestra, line 1: Do we even need this file? 🤔

gonzaloriestra: I'd say there's no need to exit, that information is updated in this file: https://github.com/changesets/changesets/blob/main/docs/prereleases.md

isaacroldan: I don't know, i'd wait for @amcaplan input as he is the expert on releases :)

amcaplan: I believe it's necessary if we have packages in a `pre` state. This file tracks the initial state of the repository before we jumped into prerelease mode.

It's certainly safest to leave it. The strategy here is incredibly messy, it's so easy to get it wrong. I'd recommend just playing it safe and leaving the file. I believe it'll get dropped when we do the next release.

amcaplan: I'm 90% sure this is fine since we never did any prereleases since the last minor release. But that 10% weighs too heavily.

This file costs us nothing to leave around until it's auto-deleted later.

PR #2342, opened 2023-07-04
Create extensions when match is cancelled

matteodepalo:

Fixes https://github.com/Shopify/internal-cli-foundations/issues/681

If someone deployed an extension, deleted it, created a new one of the same type and then run `dev` or `deploy`, they would see the matching prompt but they wouldn't be allowed to continue and create the newly added extension.

Instead of cancelling, the second option of the matching prompt allows to continue and creates a new extension.

### How to test your changes?

With unified deployments

- Add a function/extension with `extension generate`
- Deploy
- Remove the extension
- Dev

You should be allowed to continue without matching.

### DIFF
--- a/packages/app/src/cli/services/context/prompts.ts
+++ b/packages/app/src/cli/services/context/prompts.ts
@@ -14,7 +14,7 @@ export async function matchConfirmationPrompt(local: LocalSource, remote: Remote
   return renderConfirmationPrompt({
     message: `Match ${local.configuration.name} (local name) with ${remote.title} (name on Shopify Partners, ID: ${remote.id})?`,
     confirmationMessage: `Yes, that's right`,
-    cancellationMessage: `No, cancel`,
+    cancellationMessage: `No, create as a new extension`,
   })
 }
 
--- a/packages/app/src/cli/services/context/identifiers-extensions.test.ts
+++ b/packages/app/src/cli/services/context/identifiers-extensions.test.ts
@@ -366,7 +366,7 @@ describe('ensureExtensionsIds: matchmaking returns ok with some pending confirma
 })
 
 describe('ensureExtensionsIds: matchmaking returns ok with some pending confirmation', () => {
-  test('do not confirms the pending ones and fails', async () => {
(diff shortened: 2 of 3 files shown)


packages/app/src/cli/services/context/prompts.ts
```diff
@@ -13,8 +13,8 @@ import {
 export async function matchConfirmationPrompt(local: LocalSource, remote: RemoteSource) {
   return renderConfirmationPrompt({
     message: `Match ${local.configuration.name} (local name) with ${remote.title} (name on Shopify Partners, ID: ${remote.id})?`,
-    confirmationMessage: `Yes, that's right`,
-    cancellationMessage: `No, cancel`,
+    confirmationMessage: `Yes, match to existing extension`,
+    cancellationMessage: `No, create as a new extension`,
```
Arkham, line 17: I think we're reusing this prompt inside `packages/app/src/cli/services/context/identifiers-functions.ts`

Arkham:
Don't we need to adapt the same logic inside `packages/app/src/cli/services/context/identifiers-functions.ts`?

packages/app/src/cli/services/context/prompts.ts
matteodepalo: I guess for unified deployments this is not a problem, but for old style deployments it is. I'll pass something from the outside to distinguish between the two cases.

Arkham:
> I was going to do that but then I noticed that `functionsToCreate` isn't actually used anywhere. I suppose we could allow to continue instead of throwing error and then let the functions framework create the extension automatically.

Not sure what you mean, I can see it being used in `ensureFunctionsIds` which is used for legacy deployments :)

matteodepalo:
>Not sure what you mean, I can see it being used in ensureFunctionsIds which is used for legacy deployments :)

What I mean is that `functionsToCreate` doesn't actually call any mutation like for extensions, it's just part of the prompt, but then there is no different behavior if the array if populated. I'll make it so you can continue though and we don't throw error.

Arkham (approved):
LGTM!

PR #2351, opened 2023-07-05
Store config-specific env variables in separate dotenv file

Arkham:

Fixes https://github.com/Shopify/internal-cli-foundations/issues/669

When config-as-code is enabled, store the app key and the extensions ids in a config-specific dotenv file.
For example, when using `shopify.app.staging.toml`, the dotenv file is going to be `.env.staging`.
If the user is using the default `shopify.app.toml`, we'll keep using the default `.env`

### How to test your changes?

- `pnpm create-app --local --template=node --path ~/Desktop`
- `pnpm shopify app config link --path ~/Desktop/your-app` (call the file `shopify.app.staging.toml`)
- `pnpm shopify app generate extension --path ~/Desktop/your-app`
- `pnpm shopify app deploy --path ~/Desktop/your-app` (the generated `.env` file should be `.env.staging`)

### Measuring impact

### Checklist


packages/app/src/cli/models/app/loader.ts
```diff
@@ -574,19 +580,25 @@ async function logMetadataForLoadedApp(
...
+}
 
-  return configurationFileNames.app
+export function getAppConfigurationShorthand(path: string) {
+  const match = basename(path).match(appConfigurationFileNameRegex)
+  if (match) {
+    const result = match[1]?.slice(1)
+    if (result === '.') return undefined
+    return result
+  }
```
gonzaloriestra, line 603: I think the result will never be `.` because `"shopify.app.toml".match(/^shopify\.app(\.[-\w]+)?\.toml$/)[1]?.slice(1)` returns undefined, so we can simplify this.

```suggestion
  return match?.[1]?.slice(1)
```

gonzaloriestra (approved):
👌

packages/app/src/cli/models/app/loader.ts
Arkham: Oh right makes sense!

PR #2352, opened 2023-07-05
Add support for extensions array

davejcameron:

Adds support for an extensions array to unify extensions styling

Introduces some new base schemas.

### Measuring impact

### Checklist

### DIFF
--- a/packages/app/src/cli/models/app/loader.ts
+++ b/packages/app/src/cli/models/app/loader.ts
@@ -11,7 +11,7 @@ import {
 import {configurationFileNames, dotEnvFileNames} from '../../constants.js'
 import metadata from '../../metadata.js'
 import {ExtensionInstance} from '../extensions/extension-instance.js'
-import {TypeSchema} from '../extensions/schemas.js'
+import {ExtensionsArraySchema, TypeSchema, UnifiedSchema} from '../extensions/schemas.js'
 import {ExtensionSpecification} from '../extensions/specification.js'
 import {getAppInfo} from '../../services/local-storage.js'
 import {zod} from '@shopify/cli-kit/node/schema'
@@ -302,36 +302,75 @@ class AppLoader {
     extensionConfigPaths.push(`!${joinPath(appDirectory, '**/node_modules/**')}`)
     const configPaths = await glob(extensionConfigPaths)
 
-    const extensions = configPaths.map(async (configurationPath) => {
+    const extensionPromises = configPaths.map(async (configurationPath) => {
       const directory = dirname(configurationPath)
-      const specification = await findSpecificationForConfig(
-        this.specifications,
-        configurationPath,
-        this.abortOrReport.bind(this),
-      )
 
-      if (!specification) return
+      const fileContent = await readFile(configurationPath)
+      const obj = decodeToml(fileContent)
+      const {extensions} = ExtensionsArraySchema.parse(obj)
 
-      const configuration = await this.parseConfigurationFile(specification.schema, configurationPath)
-      const entryPath = await this.findEntryPath(directory, specification)
+      let extensionInstances = []
+      if (extensions) {
+        const configuration = await this.parseConfigurationFile(UnifiedSchema, configurationPath)
+        const extensionsPromises = configuration.extensions.map(async (extensionConfig) => {
+          const mergedConfig = {...configuration, ...extensionConfig}
+          const type = mergedConfig.type
+          const extensionSpecification = findSpecificationForType(this.specifications, type)
 
-      const extensionInstance = new ExtensionInstance({
-        configuration,
-        configurationPath,
-        entryPath,
-        directory,
-        specification,
-      })
+          if (!extensionSpecification) {
+            this.abortOrReport('Unknown extension type', undefined, configurationPath)
+            return
+          }
 
-      const validateResult = await extensionInstance.validate()
-      if (validateResult.isErr()) {
-        this.abortOrReport(outputContent`\n${validateResult.error}`, undefined, configurationPath)
+          const entryPath = await this.findEntryPath(directory, extensionSpecification)
+          const extensionInstance = new ExtensionInstance({
+            configuration: mergedConfig,
+            configurationPath,
+            entryPath,
+            directory,
+            specification: extensionSpecification,
+          })
+
+          const validateResult = await extensionInstance.validate()
+          if (validateResult.isErr()) {
+            this.abortOrReport(outputContent`\n${validateResult.error}`, undefined, configurationPath)
+          } else {
+            return extensionInstance
+          }
+        })
+        extensionInstances = getArrayRejectingUndefined(await Promise.all(extensionsPromises))
+      } else {
+        const specification = await findSpecificationForConfig(
+          this.specifications,
+          configurationPath,
+          this.abortOrReport.bind(this),
+        )
+
+        if (!specification) return
+        const entryPath = await this.findEntryPath(directory, specification)
+        const configuration = await this.parseConfigurationFile(specification.schema, configurationPath)
+
+        const extensionInstance = new ExtensionInstance({
+          configuration,
+          configurationPath,
+          entryPath,
+          directory,
+          specification,
+        })
+
+        const validateResult = await extensionInstance.validate()
+        if (validateResult.isErr()) {
+          this.abortOrReport(outputContent`\n${validateResult.error}`, undefined, configurationPath)
+        } else {
+          extensionInstances.push(extensionInstance)
+        }
       }
 
-      return extensionInstance
+      return extensionInstances
     })
 
-    const allExtensions = getArrayRejectingUndefined(await Promise.all(extensions))
+    const extensions = await Promise.all(extensionPromises)
+    const allExtensions = extensions.flat().filter((extension) => extension !== undefined) as ExtensionInstance[]
     return {allExtensions, usedCustomLayout: extensionDirectories !== undefined}
   }
 
--- a/packages/app/src/cli/models/extensions/extension-instance.ts
+++ b/packages/app/src/cli/models/extensions/extension-instance.ts
@@ -66,6 +66,10 @@ export class ExtensionInstance<TConfiguration extends BaseConfigType = BaseConfi
     return this.configuration.name
   }
 
+  get handle() {
+    return this.configuration.handle
+  }
+
   get dependency() {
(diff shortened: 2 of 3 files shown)


packages/app/src/cli/models/extensions/extension-instance.ts
```diff
@@ -115,8 +119,8 @@ export class ExtensionInstance<TConfiguration extends BaseConfigType = BaseConfi
     this.directory = options.directory
     this.specification = options.specification
     this.devUUID = `dev-${randomUUID()}`
-    this.localIdentifier = basename(options.directory)
-    this.idEnvironmentVariableName = `SHOPIFY_${constantize(basename(this.directory))}_ID`
+    this.localIdentifier = this.configuration.handle ? this.configuration.handle : basename(options.directory)
```
davejcameron, line 122: We will need a different fallback for a localIdentifier when a handle isn't present, otherwise both extensions will share the same one.

packages/app/src/cli/models/app/loader.test.ts
```diff
@@ -484,6 +484,59 @@ automatically_update_urls_on_dev = true
...
+      name = "my_extension_1_flow"
+      runtime_url = "https://example.com"
+
+      [settings]
+      [[settings.fields]]
+      key = "my_field"
+      name = "My Field"
+      description = "My Field Description"
+      required = true
+      type = "string"
```
davejcameron, line 511:
```suggestion
      type = "single_line_text_field"
```

packages/app/src/cli/models/extensions/schemas.ts
```diff
@@ -32,17 +37,54 @@ export const ApiVersionSchema = zod.string()
...
+    )
+    .optional(),
+})
+
+export const HandleSchema = zod
+  .string()
+  .trim()
+  .min(3, 'Handle needs to have at least 3 characters')
+  .max(30, "Handle can't exceed 30 characters")
+  .regex(/^[a-zA-Z0-9.\-_]*$/, 'Handle can only contain alphanumeric characters, periods, hyphens, and underscores')
```
nsimone, line 59: Updated guidance for handle validation (only alphanumeric and hyphens): https://github.com/Shopify/shopify/pull/435644#issuecomment-1629915032

nsimone (approved):
LGTM from the Flow side, just the one comment on the handle validation.

packages/app/src/cli/models/extensions/extension-instance.ts
```diff
@@ -115,8 +119,8 @@ export class ExtensionInstance<TConfiguration extends BaseConfigType = BaseConfi
     this.directory = options.directory
     this.specification = options.specification
     this.devUUID = `dev-${randomUUID()}`
-    this.localIdentifier = basename(options.directory)
-    this.idEnvironmentVariableName = `SHOPIFY_${constantize(basename(this.directory))}_ID`
+    this.localIdentifier = this.configuration.handle ?? basename(this.directory)
```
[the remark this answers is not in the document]
isaacroldan: For flow I added a new `BaseSchemaWithHandle` where the handle is mandatory. I'll do the same for ui_extension in the next PR and we should use that. So for extensions using the new unified config, the handle will be mandatory.

another option would be to make it mandatory for ALL extensions, and define a default value if not present (equal to `slugified(name)`, which is also equal to `basename(this.directory)`

I'll explore this tomorrow

packages/app/src/cli/models/extensions/schemas.ts
```diff
@@ -18,6 +18,11 @@ export const TypeSchema = zod.object({
   type: zod.string().default('ui_extension'),
 })
 
+export const ExtensionsArraySchema = zod.object({
+  type: zod.string().optional(),
+  extensions: zod.array(zod.any()).optional(),
```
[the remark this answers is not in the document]
isaacroldan: in the `UnifiedSchema` we also define the extensions array as an array of `any` for two reasons:
- 1. we don't care about validating the content of the array at that point, when parsing each extension then each specification schema will take care of parsing and validating that config.
- 2. I tried to use `zod.array(BaseSchema)` and it didn't work, seems like a bug in zod, couldn't figure out why.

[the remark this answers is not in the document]
vividviolet: That's fine with me

packages/app/src/cli/models/extensions/extension-instance.ts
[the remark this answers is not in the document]
vividviolet: I would prefer if we generate a default value for them as it makes things easier. Ok, that can be done in another PR.

packages/app/src/cli/models/extensions/schemas.ts
nsimone: Makes sense to me 🤷

packages/app/src/cli/models/extensions/schemas.ts
```diff
@@ -18,6 +18,11 @@ export const TypeSchema = zod.object({
   type: zod.string().default('ui_extension'),
 })
 
+export const ExtensionsArraySchema = zod.object({
+  type: zod.string().optional(),
```
nsimone, line 22: I have the same question as @MaxLeduc about `type` and `handle`, should they be on each extension in the array, or here at the higher level? https://github.com/Shopify/cli/pull/2366#discussion_r1258910061

vividviolet (changes requested):
The new format doesn't seem to work for `ui_extension` due to `extensionPoints` being required

image

isaacroldan:
@vividviolet I know i know, my plan for this PR was to just support extensions arrays and update flow extensions. And in the next PR I'll add support for ui_extensions and the new fields it requires. Didn't want to make this PR bigger 🙇

packages/app/src/cli/models/extensions/schemas.ts
isaacroldan: this specific `type` is to support extensions with the legacy format. Extensions using the unified config will have `type` and `handle` in each extension

nsimone: I see - so right now the flow specs still have handle at the top level (i.e. `config.handle`), and later on it could be inside each extension?

packages/app/src/cli/models/app/loader.ts
```diff
@@ -313,36 +355,30 @@ class AppLoader {
...
-        this.abortOrReport(outputContent`\n${validateResult.error}`, undefined, configurationPath)
+      const fileContent = await readFile(configurationPath)
+      const obj = decodeToml(fileContent)
+      const {extensions, type} = ExtensionsArraySchema.parse(obj)
+
+      if (extensions) {
+        // If the extension is an array, it's a unified toml file.
+        // Parse all extensions by merging each extension config with the global unified configuration.
+        const configuration = await this.parseConfigurationFile(UnifiedSchema, configurationPath)
+        const extensionsPromises = configuration.extensions.map(async (extensionConfig) => {
```
MaxLeduc, line 369: nit: you have an `extensionsPromises` variable inside your `extensionPromises` map which is a tad hard to read. Since `extensionsPromises` is using `createExtensionInstance` could we rename it to `extensionsInstances` or `extensionsInstancesPromises`?

packages/app/src/cli/models/extensions/schemas.ts
```diff
@@ -32,17 +37,54 @@ export const ApiVersionSchema = zod.string()
...
   name: zod.string(),
   type: zod.string(),
+  handle: HandleSchema.optional(),
   description: zod.string().optional(),
   api_version: ApiVersionSchema.optional(),
   extension_points: zod.any().optional(),
   capabilities: CapabilitiesSchema.optional(),
   metafields: zod.array(MetafieldSchema).optional().default([]),
   categories: zod.array(zod.string()).optional(),
+  extensions: zod.array(zod.any()).optional(),
```
MaxLeduc, line 71: Would we ever allow a `shopify.extensions.toml` to not contain any `extensions`? For Flow extensions we set this property to be required and of `.min(1)`.

If this is set as optional to support legacy schemas would it make sense to add custom logic so that if the `extensions` array is there we require it to contain at least one extension?

packages/app/src/cli/models/extensions/specifications/flow_trigger.ts
```diff
@@ -1,37 +1,17 @@
...
-              key: zod.string().optional(),
-              description: zod.string().optional(),
-              type: zod.string(),
-            })
-            .refine((field) => validateNonCommerceObjectShape(field, 'flow_trigger')),
-        )
-        .optional(),
-    })
-    .optional(),
+  schema: zod.string().optional(),
```
MaxLeduc, line 9: FYI we're removing the `schema` and `return_type_ref`  fields since we won't support them for triggers for Editions. Here's @mikee649's [PR](https://github.com/Shopify/cli/pull/2401) that handles that.

packages/app/templates/extensions/projects/flow_action/shopify.extension.toml.liquid
```diff
@@ -1,6 +1,5 @@
 name = "{{ name }}"
```
MaxLeduc, line 1: Will these be renamed as `shopify.extensions.toml` as part of this PR?

MaxLeduc (commented):
Looks good! Left a few comment, LMK if that makes sense.

packages/app/templates/extensions/projects/flow_action/shopify.extension.toml.liquid
isaacroldan: No, per my last conversation with Meredith, we'll continue using the singular name for now. But we can revisit that after editions

packages/app/src/cli/models/extensions/specifications/flow_trigger.ts
isaacroldan: i'll remove those in this PR as well

packages/app/src/cli/models/extensions/schemas.ts
isaacroldan: With this change, all extensions will receive a flattened config, so flow doesn't know anymore about the "extensions array", so from the specification point of view, everything is "top level". So yes, from the spec the correct way of accessing the handle is `config.handle`, but in the toml it will be part of the extensions array object (not at the top)

isaacroldan (approved):

PR #2373, opened 2023-07-07
Support multiple exports per JS Function

jbourassa:

We want to support Functions with multiple entry points. This PR brings support for that to JavaScript modules, using the new [Javy feature](http://github.com/bytecodealliance/javy/pull/412).

Closes https://github.com/Shopify/script-service/issues/7050

Split the JS building in 2 processes because building for multiple exports requires: 
- Generating a `.wit` file with a world listing all exports
- Generating a JS module that has all exports wrapped w/ Shopify Function's logic. I've extracted a `run` method in `@shopify/shopify_function`, see  https://github.com/Shopify/shopify-function-javascript/pull/6.

Limitations worth noting:
1. Following WIT standard, Wasm exports (the ones in the TOML file) must be `kebab-cased`. If we see a `camelCase` export, we'll throw a helpful error message asking to change it to kebab-case. For `kebab-case`d TOML (Wasm) exports, we'll look up a `camelCase` JS export.
    
    I favoured this over changing the TOML (Wasm) export behind the scene because the Wasm export is public and partners will see it (partners dash, when running function runner, etc.).
  
    We can reverse that decision if we see it's causing too much friction by adding complexity to our stack (accept both `kebab-case` and `camelCase`, or add a Wasm transformation phase), or potential confusion (changing it behind the scene). But the current conservative approach feels like the right first step.

2. All target exports must be explicitly defined, or none. If none are provided, we'll use the default Javy behaviour which results in a WASI module w/ `_start`. If exports are provided, we'll generate a module that w/ exports matching the JS exports (`default` can be explicitly used).

#gsd:34464

### DIFF
--- a/.changeset/brown-weeks-fly.md
+++ b/.changeset/brown-weeks-fly.md
@@ -1,0 +1,5 @@
+---
+'@shopify/app': minor
+---
+
+Support building JS Function with multiple export targets


packages/app/src/cli/services/function/build.test.ts
```diff
@@ -195,3 +195,55 @@ describe('runFunctionRunner', () => {
     )
   })
 })
+
```
jbourassa, line 198: More tests needed, will do that once I land on an approach.

packages/app/src/cli/services/function/build.ts
```diff
@@ -137,3 +156,122 @@ export async function runFunctionRunner(fun: ExtensionInstance, options: Functio
...
+  const output_obj = userfunction(input_obj);
+  const output_str = JSON.stringify(output_obj);
+  const output_data = new TextEncoder().encode(output_str);
+}`
+
+  const exports = this.exports.map((name) => {
+    const identifier = camelize("run_" + name);
+    return `
+function ${identifier}() { return run(userExports["${name}"]) }
+export { ${identifier} as "${hyphenate(name)}" }` })
```
jeffcharles, line 260: I don't think exporting as a hyphenated name will work with QuickJS unfortunately.

packages/app/src/cli/services/function/build.ts
```diff
@@ -137,3 +156,122 @@ export async function runFunctionRunner(fun: ExtensionInstance, options: Functio
...
+  const targets = fun.configuration.targeting || []
+  const withoutExport = targets.find((target) => !target.export)
+  const withExports = targets.filter((target) => Boolean(target.export))
+
+  if (withoutExport && withExports.length > 0) {
+    throw new Error(`Can't infer export name for '${withoutExport.target}'. All targets must have an export, or none.`)
+  }
+
+  return withExports.map((target) => target.export!)
+}
```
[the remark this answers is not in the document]
jeffcharles: > [..], the underlying reason is that we typically map default exports to _start in WASI. Explicitly accepting default as the name is confusing (should the behaviour diverge from the unnamed default that gets mapped to _start? what would people expect?)

Not exactly. `_start` is mapped to top-level evaluation of the JS and not invoking any functions. If we were to invoke a default exported arrow function as part of invoking `_start`, it would be a breaking change in Javy's existing behaviour. And also we wouldn't be able to support arguments in `_start`. I think we could support exporting a function named `default` for a default function export or default arrow function export. We just don't at the moment. See https://github.com/bytecodealliance/javy/issues/422.

packages/app/src/cli/services/function/build.ts
```diff
@@ -137,3 +157,121 @@ export async function runFunctionRunner(fun: ExtensionInstance, options: Functio
...
+  get entrypointContents() {
+    const prelude = `
+import * as userExports from "user-function"
+import * as fs from "javy/fs"
+
+function run(fn) {
+  const input_data = fs.readFileSync(fs.STDIO.Stdin);
+  const input_str = new TextDecoder("utf-8").decode(input_data);
+  const input_obj = JSON.parse(input_str);
+  const output_obj = userfunction(input_obj);
```
dphm, line 251: What is this `userfunction`? `userFunction`?

jbourassa: It should read `fn`, sorry. I'll update soon with a change in `@shopify/shopify_function`.

packages/app/src/cli/services/function/build.ts
```diff
@@ -137,3 +156,122 @@ export async function runFunctionRunner(fun: ExtensionInstance, options: Functio
...
+
+  const exports = this.exports.map((name) => {
+    const identifier = camelize("run_" + name);
+    return `
+function ${identifier}() { return run(userExports["${name}"]) }
+export { ${identifier} as "${hyphenate(name)}" }` })
+
+
+    return prelude + '\n' + exports.join("\n")
+  }
```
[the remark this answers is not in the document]
jbourassa: Fixed: I added javy as a peerDependency to `@shopify/shopify_function`.

packages/app/src/cli/services/function/build.ts
```diff
@@ -137,3 +157,117 @@ export async function runFunctionRunner(fun: ExtensionInstance, options: Functio
...
+
+  withExports.forEach((target) => {
+    const name = target.export!
+    if (!name.match(/^[a-z0-9-]+$/)) {
+      throw new Error(`Invalid export name: '${name}'.
+
+The TOML's exports must be kebab-case (lowercase, hyphen or numbers) to comply with WebAssembly's Component Model.
+camelCase JavaScript exports are automatically mapped to kebab-case Wasm exports.
+
+Suggestion: change the export name from '${name}' to '${hyphenate(name)}' in the TOML file.`)
```
nickwesselman, line 268: ❤️

packages/app/src/cli/services/function/build.ts
```diff
@@ -137,3 +157,121 @@ export async function runFunctionRunner(fun: ExtensionInstance, options: Functio
...
+      outfile: joinPath(fun.directory, 'dist/function.js'),
+      stdin: {
+        contents,
+        loader: 'ts',
+        resolveDir: fun.directory,
+      },
+      alias: {
+        'user-function': fun.entrySourceFilePath,
+      },
+      logLevel: 'silent',
```
jeffcharles, line 206: Will setting the log level to silent also silence what's causing the error if there is one or will that continue to be emitted?

jeffcharles (commented):
Has this been tested with a user trying to use a default JS export?

dphm:
🎩 `pnpm shopify app deploy --path ~/path/to/app`

### ✅ With default export

```
[[targeting]]
target = "purchase.validation.validate"
input_query = "./input.graphql"
```

```javascript
// @ts-check

/**
 * @typedef {import("../generated/api").InputQuery} InputQuery
 * @typedef {import("../generated/api").FunctionResult} FunctionResult
 */

export default /**
 * @param {InputQuery} input
 * @returns {FunctionResult}
 */
(input) => {
  const errors = input.cart.lines
    .filter(({ quantity }) => quantity > 1)
    .map(() => ({
      localizedMessage: "Not possible to order more than one of each",
      target: "cart",
    }));

  return {
    errors,
  };
}
```

#### With no default export

This deployed with no validation error 🤔

### ✅ With named export

```toml
[[targeting]]
target = "purchase.validation.validate"
export = "validate"
input_query = "./input.graphql"
```

```javascript
// @ts-check

/**
 * @typedef {import("../generated/api").InputQuery} InputQuery
 * @typedef {import("../generated/api").FunctionResult} FunctionResult
 */

export function /**
 * @param {InputQuery} input
 * @returns {FunctionResult}
 */
validate(input) {
  const errors = input.cart.lines
    .filter(({ quantity }) => quantity > 1)
    .map(() => ({
      localizedMessage: "Not possible to order more than one of each",
      target: "cart",
    }));

  return {
    errors,
  };
}
```

#### With no exported `validate` function

```
╭─ error ────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│                                                                                                                    │
│  Build failed with 1 error:                                                                                        │
│  <stdin>:4:9: ERROR: No matching export in "../../../../apps/shoop/extensions/ccv/src/index.js" for import         │
│  "validate"                                                                                                        │
```

packages/app/src/cli/services/function/build.ts
dphm: I don't know the answer but that behaviour isn't changing in this PR, right? 😅

```javascript
function getESBuildOptions(directory: string, entryPoint: string, userFunction: string): Parameters<typeof esBuild>[0] {
  const esbuildOptions: Parameters<typeof esBuild>[0] = {
    outfile: joinPath(directory, 'dist/function.js'),
    entryPoints: [entryPoint],
    alias: {
      'user-function': userFunction,
    },
    logLevel: 'silent',
    bundle: true,
    legalComments: 'none',
    target: 'es2022',
    format: 'esm',
  }
  return esbuildOptions
}
```

Jimmy mentioned maybe refactoring the previous build stuff, will see if I have time 🙈

jeffcharles: Ah, the diff in its present state made that difficult to see. I'll resolve this.

packages/app/src/cli/services/function/build.ts
```diff
@@ -137,3 +157,121 @@ export async function runFunctionRunner(fun: ExtensionInstance, options: Functio
...
+      const witPath = joinPath(dir, 'javy-world.wit')
+      await writeFile(witPath, witContent)
+
+      return runJavy(fun, options, ['--wit', witPath, '-n', JAVY_WORLD])
+    })
+  }
+
+  get wit() {
+    // % escapes the name to avoid conflict with reserved words, if any
+    const witExports = this.exports.map((name) => `export %${hyphenate(name)}: func()`)
```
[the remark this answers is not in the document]
dphm: <details>
<summary>Error on non-unique exports</summary>

```
│  Build failed with 2 errors:                                                 │
│  <stdin>:8:16: ERROR: Multiple exports with the same name "pleaseValidate"   │
│  <stdin>:8:16: ERROR: The symbol "pleaseValidate" has already been declared  │
│  To investigate the issue, examine this stack trace:                         │
│    at (<stdin>:8)                                                            │
│    at (<stdin>:8)                                                            │
│    at failureErrorWithLog                                                    │
│    (.pnpm/esbuild@0.18.11/node_modules/esbuild/lib/main.js:1639)             │
│      let error = new Error(`${text}${summary}`);                             │
│    at (.pnpm/esbuild@0.18.11/node_modules/esbuild/lib/main.js:1051)          │
│      const error = failureErrorWithLog("Build failed",                       │
│      originalErrors.concat(onEndErrors),                                     │
│      originalWarnings.concat(onEndWarnings));                                │
│    at (.pnpm/esbuild@0.18.11/node_modules/esbuild/lib/main.js:996)           │
│      buildOrContextContinue(null, (result, done) => done([], []), () => {    │
│    at buildResponseToResult                                                  │
│    (.pnpm/esbuild@0.18.11/node_modules/esbuild/lib/main.js:1049)             │
│      runOnEndCallbacks(result, (onEndErrors, onEndWarnings) => {             │
│    at (.pnpm/esbuild@0.18.11/node_modules/esbuild/lib/main.js:1078)          │
│      return buildResponseToResult(response, (err, res) => {                  │
│    at <computed>                                                             │
│    (.pnpm/esbuild@0.18.11/node_modules/esbuild/lib/main.js:700)              │
│      callback(error, response);                                              │
│    at handleIncomingPacket                                                   │
│    (.pnpm/esbuild@0.18.11/node_modules/esbuild/lib/main.js:755)              │
│      callback(null, packet.value);                                           │
│    at readFromStdout                                                         │
│    (.pnpm/esbuild@0.18.11/node_modules/esbuild/lib/main.js:676)              │
│      handleIncomingPacket(stdout.subarray(offset, offset + length));         │
│    at emit (node:events:513)                                                 │
│    at emit (node:domain:489)                                                 │
```
</details>

packages/app/src/cli/services/function/build.test.ts
```diff
@@ -195,3 +195,184 @@ describe('runFunctionRunner', () => {
...
+    ]
+
+    // When
+    const got = jsExports(ourFunction)
+
+    // Then
+    expect(got).toEqual(['foo-bar', 'foo-baz'])
+  })
+
+  test('errors when multiple targets present without explicit exports', async () => {
```
dphm, line 350:
```toml
[[targeting]]
target = "purchase.validation.fetch"

[[targeting]]
target = "purchase.validation.validate"
```

```
│  Can't infer export name for targets:                                        │
│  - 'purchase.validation.fetch'                                               │
│  - 'purchase.validation.validate'                                            │
│  All targets must have an export when multiple targets are present.          │
```

packages/app/src/cli/services/function/build.test.ts
```diff
@@ -195,3 +195,184 @@ describe('runFunctionRunner', () => {
...
+      {target: 'foo.biz'},
+    ]
+
+    // When & Then
+    expect(() => {
+      jsExports(ourFunction)
+    }).toThrow(/Can't infer export name for targets:\n- 'foo\.baz'\n- 'foo\.biz'/)
+  })
+
+  test('errors when exports are not kebab-case', async () => {
```
dphm, line 365:
```toml
[[targeting]]
target = "purchase.validation.fetch"
export = "pleaseFetch"

[[targeting]]
target = "purchase.validation.validate"
export = "pleaseValidate"
```

```
│  Invalid export names: 'pleaseFetch', 'pleaseValidate'.                      │
│  The TOML's exports must be kebab-case (lowercase, hyphen or numbers) to     │
│  comply with WebAssembly's Component Model.                                  │
│  JavaScript exports with camelCase names are automatically mapped to         │
│  kebab-case Wasm exports.                                                    │
│  Suggested TOML changes:                                                     │
│  - Change export for 'purchase.validation.fetch' to 'please-fetch'.          │
│  - Change export for 'purchase.validation.validate' to 'please-validate'.    │
```

SaifBenChahed (approved):
I did tophat and it works. the only thing I would say is we should add is the verification that the function.wasm exist at the end of the build before we consider it successful. I am suggesting this because when tophatting I had an issue where build succeeds but doesn't generate the wasm file and it was because the downloaded javy binary was corrupted somehow and didn't have execution rights. I had to delete the downloaded javy for it to download another one and finally works

jeffcharles:
> I am suggesting this because when tophatting I had an issue where build succeeds but doesn't generate the wasm file and it was because the downloaded javy binary was corrupted somehow and didn't have execution rights.

Interesting. @SaifBenChahed if you have more details to share about this, I would be very interested since it sounds like a bug with the `javy-cli` NPM package. Can you let me know what version of the `javy-cli` NPM package is installed for the project where you saw this happen?

jeffcharles:
Okay, that makes a little more sense as to what's happening. We added a fix in `javy-cli` version `0.1.6` to detect if the Javy process fails to execute properly and emit a non-zero exit code. Since you're running an older version, I can see why Javy is failing silently. That also said, this and a merchant report helped me track down a different bug that I wasn't aware of where we aren't checking the status code on the response we get from GitHub when downloading the Javy binary and that causes the an empty Javy binary file to be created without execute permissions which breaks the compilation process until the cached binary is deleted. I have a fix for that up at https://github.com/bytecodealliance/javy/pull/442.

PR #2409, opened 2023-07-11
Ensure reset on link logic applies to ensureGenerateContext

rezaansyed:

Fixes #0000 

### Measuring impact

### Checklist

### DIFF
--- a/packages/app/src/cli/services/context.ts
+++ b/packages/app/src/cli/services/context.ts
@@ -89,6 +89,7 @@ export async function ensureGenerateContext(options: {
   directory: string
   reset: boolean
   token: string
+  commandConfig: Config
   configName?: string
 }): Promise<string> {
   if (options.apiKey) {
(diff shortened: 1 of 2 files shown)


packages/app/src/cli/services/context.ts
```diff
@@ -584,16 +570,32 @@ async function getAppDevCachedContext({
...
   if (reset) clearAppInfo(directory)
 
   let cachedInfo = getAppInfo(directory)
 
+  if (previousCachedInfo === undefined) {
+    const explanation =
+      `\nLooks like this is the first time you're running dev for this project.\n` +
+      'Configure your preferences by answering a few questions.\n'
+    outputInfo(explanation)
+  }
```
Arkham, line 592: I think this message can go, we can also remove the similar one that says "Looks like this is the first time you're running generate for this project."

Arkham: I'd rather do it here, because the message that we print when running `generate` for the first time will be incorrect, no?

Arkham (approved):
LGTM!

PR #2416, opened 2023-07-12
Show diff of changes on config push

gonzaloriestra:

Fixes https://github.com/Shopify/internal-cli-foundations/issues/765

- Shows a confirmation prompt to confirm the changes. It ignores comments and ordering changes.
diff
- Does nothing if there are no changes
no-changes
- Skips the confirmation when passing `--force` or `-f`

### How to test your changes?

Modify the local configuration and run `p shopify app config push`

### Measuring impact

### Checklist

### DIFF
--- a/packages/app/src/cli/prompts/config.test.ts
+++ b/packages/app/src/cli/prompts/config.test.ts
@@ -1,12 +1,16 @@
 import {confirmPushChanges, selectConfigFile, selectConfigName, validate} from './config.js'
 import {PushOptions} from '../services/app/config/push.js'
-import {testOrganizationApp, testAppWithConfig} from '../models/app/app.test-data.js'
+import {testOrganizationApp, testAppWithConfig, testApp} from '../models/app/app.test-data.js'
 import {App} from '../api/graphql/get_config.js'
+import {mergeAppConfiguration} from '../services/app/config/link.js'
+import {OrganizationApp} from '../models/organization.js'
+import {AppConfiguration} from '../models/app/app.js'
 import {describe, expect, test, vi} from 'vitest'
 import {inTemporaryDirectory, writeFileSync} from '@shopify/cli-kit/node/fs'
 import {renderConfirmationPrompt, renderSelectPrompt, renderTextPrompt} from '@shopify/cli-kit/node/ui'
 import {joinPath} from '@shopify/cli-kit/node/path'
 import {err, ok} from '@shopify/cli-kit/node/result'
+import {decodeToml} from '@shopify/cli-kit/node/toml'
 
 vi.mock('@shopify/cli-kit/node/ui')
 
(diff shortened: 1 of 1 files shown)


ryancbahan:
awesome! question for future state: would it be more readable to show the diff in something more like a table view, with `prev` and `current` headers or something similar?

gonzaloriestra:
> awesome! question for future state: would it be more readable to show the diff in something more like a table view, with prev and current headers or something similar?

@ryancbahan that's something we can explore. I also find clearer two column diffs in general, but we have to take into account that the width of the terminal (or the number of chars we can print) is usually much smaller than a browser, so it's probably hard to fit it.

packages/app/src/cli/prompts/config.ts
```diff
@@ -64,3 +71,27 @@ export function validate(value: string): string | undefined {
   // Max filename size for Windows/Mac including the prefix/postfix
   if (result.length > 238) return 'The file name is too long.'
 }
+
+export async function confirmPushChanges(options: PushOptions, app: App) {
+  if (options.force) return true
+
+  const {configuration, configurationPath} = options
+  const localConfiguration = await readFile(configurationPath)
```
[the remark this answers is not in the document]
rezaansyed: hmm that's weird. The config I had was from a newly generated app

packages/app/src/cli/prompts/config.ts
```diff
@@ -63,3 +71,32 @@ export function validate(value: string): string | undefined {
...
   if (result.length > 238) return 'The file name is too long.'
 }
+
+export async function confirmPushChanges(options: PushOptions, app: App) {
+  if (options.force) return true
+
+  const {configuration, configurationPath} = options
+  const localConfigurationToml = await readFile(configurationPath)
+  const localConfiguration = decodeToml(localConfigurationToml)
+  const remoteConfiguration = mergeAppConfiguration({configuration} as AppInterface, app as OrganizationApp)
```
RyanDJLee, line 81: This should just be the current configuration values on the platform without the merge.

RyanDJLee (changes requested):
Looks great! But we may need to change to logic for what we want to compare.

packages/app/src/cli/prompts/config.ts
rezaansyed: I'm noticing that the diff captures the changes without the merge so I think it's fine:
image

rezaansyed (approved):
Tophatted and the changes look good to me! 🎉

packages/app/src/cli/prompts/config.ts
gonzaloriestra: If we compare against the raw platform configuration, then some local values would always appear in the diff (like `automatically_update_urls_on_dev`).

Also, we just have the app here, so we would need to create a configuration object to be able to compare, selecting the desired fields. Which is basically what `mergeAppConfiguration` does. Using the same logic than in `link` ensures we will never find changes when we push right after linking.

gonzaloriestra:
I've made a couple of last-minute changes:
- The app URL returns always with and ending `/` (`https://example.com/`) even if it's pushed without it (`https://example.com`), and the cloudflare URLs we are setting with dev don't have it. So I'm removing it from the remote fields.
- The link command was removing some local configuration (like `automatically_update_urls_on_dev`). I changed it to preserve it and only overwrite the remote fields.

RyanDJLee (approved):
Looks good!

PR #2417, opened 2023-07-12
Always show the footer in dev, by keeping the task display running

shauns:

Ensures that we are consistent in showing the preview footer in dev

Keeps the dev task list open, with its footer, even when the tasks have completed running.

### How to test your changes?

Init a "none" app, and add a subscription ui extension. Run dev and observe that the footer is displayed and can be previewed with (p)

### Measuring impact

### Checklist

### DIFF
--- a/packages/app/src/cli/services/dev/output.ts
+++ b/packages/app/src/cli/services/dev/output.ts
@@ -57,7 +57,7 @@ export function outputExtensionsMessages(app: AppInterface) {
   outputThemeExtensionsMessage(app.allExtensions.filter((ext) => ext.isThemeExtension))
 }
 
-export function renderDev(renderConcurrentOptions: RenderConcurrentOptions, previewUrl: string | undefined) {
+export function renderDev(renderConcurrentOptions: RenderConcurrentOptions, previewUrl: string) {
   let options = renderConcurrentOptions
 
   if (previewUrl) {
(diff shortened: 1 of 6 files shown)


packages/app/src/cli/services/dev/output.ts
```diff
@@ -86,7 +86,7 @@ export function renderDev(renderConcurrentOptions: RenderConcurrentOptions, prev
       },
     }
   }
-  return renderConcurrent(options)
+  return renderConcurrent({...options, keepOpenAfterStopping: true})
```
matteodepalo, line 89: I get this change, but could we make it so extension only projects have a promise that never resolves intead?

packages/cli-kit/src/private/node/ui/components/ConcurrentOutput.tsx
```diff
@@ -77,48 +78,41 @@ const ConcurrentOutput: FunctionComponent<ConcurrentOutputProps> = ({
...
-        const stderr = writableStream(process, index)
-
-        await process.action(stdout, stderr, abortSignal)
-      }),
-    )
-  }
+  const writableStream = useCallback(
+    (process: OutputProcess, index: number) => {
+      const concurrentColors: TextProps['color'][] = ['yellow', 'cyan', 'magenta', 'green', 'blue']
+      function lineColor(index: number) {
```
matteodepalo, line 92: I would not move this function and `concurrentColors` inside as they don't need to be instantiated for every writableStream. You can convert them to `useCallback` and `useMemo`.

packages/app/src/cli/services/dev/output.ts
matteodepalo: If not, maybe I'd change the name to something like `keepRunningAfterProcessesResolve`.

PR #2426, opened 2023-07-12
Fix registered but not released extensions appearing in Update and Missing locally sections in deploy confirmation prompt

mattnoakes:

Fixes https://github.com/Shopify/develop-app-foundations/issues/168

Context: A local extension that is registered but not included in the active version and is being deployed
This extension will appear in the in the Update section of the deploy confirmation prompt. Since we want to compare changes being deployed against the active state, this extension should appear in the Add section instead. If this extension is removed locally, it currently appears in the Removed section. Instead, it should not be referenced in the confirmation prompt.

Changes to change deploy behaviour:
- The early return when an app has no active version is removed so that `deployConfirmationPrompt()` doesn't fall back to `buildLegacyDeploymentInfoPrompt()`
- Added empty array defaults for `nonDashboardRemoteRegistrations` and `onlyRemote` for when there's no active version
- `deployConfirmationPrompt()` will only early return if the extensions `infoTable` is empty and `deploymentMode` is `legacy` so that the deploy will continue for an empty version

Test changes:
- Added tests to cover the bug cases
- Added some `describe` blocks to organize the tests and break down the long test names
- Refactored `buildSourceSummary()` and `unifiedRenderConfirmationPromptContent()` to make them reusable across the different tests

**Before**
With local extension:
with extension - main
After removing local extension:
without extension - main

**After**
With local extension:
with-local-extension
After removing local extension:
without-local-extension

### How to test your changes?

1. Create an app: `bin/create-test-app.js --bare`
2. Generate a web pixel extension: `pnpm shopify app generate extension`
3. Enable the `app_unified_deployment` app beta
4. Run dev: `pnpm shopify app dev`
5. Run deploy: `pnpm shopify app deploy`
6. Delete the web pixel extension locally
7. Run deploy: `pnpm shopify app deploy`

### Measuring impact

### Checklist

### DIFF
--- a/.changeset/twelve-crabs-switch.md
+++ b/.changeset/twelve-crabs-switch.md
@@ -1,0 +1,6 @@
+---
+'@shopify/cli-kit': patch
+'@shopify/app': patch
+---
+
+The status of registered, but unreleased, CLI-managed extensions in the deploy confirmation prompt are based on their relation to the active app version (or a blank slate if there's no active version)
--- a/packages/app/src/cli/services/context/prompts.test.ts
+++ b/packages/app/src/cli/services/context/prompts.test.ts
@@ -1,144 +1,310 @@
-import {deployConfirmationPrompt} from './prompts.js'
+import {deployConfirmationPrompt, SourceSummary} from './prompts.js'
+import {RemoteSource, LocalSource} from './identifiers.js'
+import {IdentifiersExtensions} from '../../models/app/identifiers.js'
 import {partnersRequest} from '@shopify/cli-kit/node/api/partners'
-import {InfoTableSection, renderConfirmationPrompt} from '@shopify/cli-kit/node/ui'
+import {renderConfirmationPrompt, InfoTableSection} from '@shopify/cli-kit/node/ui'
 import {describe, expect, test, vi} from 'vitest'
 
 vi.mock('@shopify/cli-kit/node/ui')
 vi.mock('@shopify/cli-kit/node/api/partners')
 
 describe('deployConfirmationPrompt', () => {
-  test('when legacy deployment mode should render confirmation prompt with the source summary', async () => {
-    // Given
-    vi.mocked(renderConfirmationPrompt).mockResolvedValue(true)
+  describe('when legacy deployment mode', () => {
+    test('renders confirmation prompt with the source summary', async () => {
+      // Given
+      vi.mocked(renderConfirmationPrompt).mockResolvedValue(true)
 
-    // When
-    const response = await deployConfirmationPrompt(buildSourceSummary(), 'legacy', 'apiKey', 'token')
+      // When
+      const response = await deployConfirmationPrompt(
+        buildSourceSummary({
+          identifiers: {...identifier1, ...identifier2},
+          toCreate: [createdExtension],
+          onlyRemote: [remoteOnlyExtension],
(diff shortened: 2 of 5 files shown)


matteodepalo:
@mattnoakes nice! I've added a couple of commits to remove the margins when `infoTable` is empty. One thing I'd like to understand is, in which cases we want to continue with an empty deployment? Could we show an info message in that case that says there is nothing to deploy?

mattnoakes:
Thanks for jumping in @matteodepalo!!

> One thing I'd like to understand is, in which cases do we want to continue with an empty deployment? Could we show an info message in that case that says there is nothing to deploy?

Our team talked about this yesterday. Whenever all extensions deleted, you should be able to continue with an empty deployment. The case here is a little odd because an extension is being deleted but we're not telling the partner about it because it's not relevant / not causing a change to end users if the version were to be released (decision in [this comment](https://github.com/Shopify/app-deploys/issues/617#issuecomment-1629509293)).

@jessehoek, what do you think should be shown in the prompt for this case? I'll create a follow-up issue for this. @ani-shopify, feel free to correct/expand on any of this.

matteodepalo:
I guess the problem I see is that the partner is being asked to confirm something that will not have any changes. If that's, the case, why not just do it without confirming and instead display a message explaining what happened?

ani-shopify:
> I guess the problem I see is that the partner is being asked to confirm something that will not have any changes, so they might be wondering what's happening behind the scenes. If that's the case, why not just do it without confirming and instead display a message explaining what happened?

@matteodepalo could you elaborate on this one? Which scenario are you referring to? Are you talking about this edge case:


> One edge case that this approach doesn't handle is this:
> 
> - The app doesn't have an active app version,
> - The app has a registration for a CLI-managed extension (maybe a different dev created it),
> - Your local code doesn't have that code locally,
> 

from [this comment](https://github.com/Shopify/app-deploys/issues/617#issuecomment-1624392434)?

Or when there are no actual impact when running the `deploy` without any extensions?

In any case, do you think that discussion can be moved into its own GH issue to unblock this PR? I have a slight preference to get this PR merged in to make sure we always compare the current state with the active version. We can always go back end enhance the UX by providing additional context what happened after the deploy was done. What do you think?

matteodepalo:
>@matteodepalo could you elaborate on this one? Which scenario are you referring to? Are you talking about this edge case:

I was referring to the scenario where the user needs to delete the registration without having deployed yet and they get the empty deployment.

> In any case, do you think that discussion can be moved into its own GH issue to unblock this PR? I have a slight preference to get this PR merged in to make sure we always compare the current state with the active version. We can always go back end enhance the UX by providing additional context what happened after the deploy was done. What do you think?

I agree, let's merge as is and we'll reconsider the UX later.

@mattnoakes I'll review the change 👍

jessehoek:
I'm thinking about this case:

Where there is no active app version, the local project has 0 extensions, but there is 1 or more registered CLI-managed extensions in the app. 

I think we need to allow the empty deployment to be created + released, so devs can get rid of lingering extension registrations. 

image

To make this more clear, I think we can include a "No extensions" in place of the info table like this.

image

I recognize this is an edge case, and we can punt this to post-editions if we have more important things to work on.

mattnoakes:
I added [this](https://github.com/Shopify/app-deploys/issues/654) to the backlog.

PR #2427, opened 2023-07-12
ml/toml and validation updates

MaxLeduc:

Did a bug bash, noticed some inconsistencies in the TOML files and validation. [Issue](https://github.com/Shopify/flow/issues/19690).

- Standardize the description property so that it is located at the extension level
- Update the generate command to generate the handle at the extension level
- Throw an error when a partner adds a key to a commerce object field
- Update trigger template to include a valid key for field
- Update the validation schema for flow actions

### How to test your changes?

Pull down the branch and create an action/trigger. Check that the generated files are correct. Test the updated validation. Deploy.

### DIFF
--- a/packages/app/src/cli/services/flow/validation.ts
+++ b/packages/app/src/cli/services/flow/validation.ts
@@ -2,30 +2,58 @@ import {ConfigField, FlowExtensionTypes} from './types.js'
 import {SUPPORTED_COMMERCE_OBJECTS} from './constants.js'
 import {zod} from '@shopify/cli-kit/node/schema'
 
-export const validateNonCommerceObjectShape = (configField: ConfigField, type: FlowExtensionTypes) => {
-  if (!Object.keys(SUPPORTED_COMMERCE_OBJECTS).includes(configField.type)) {
-    if (!configField.key) {
-      throw new zod.ZodError([
-        {
-          code: zod.ZodIssueCode.custom,
-          path: ['settings.fields.key'],
-          message: 'Key must be specified for non-commerce object fields',
-        },
-      ])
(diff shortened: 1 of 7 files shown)


packages/app/src/cli/services/flow/validation.ts
```diff
@@ -25,6 +25,28 @@ export const validateNonCommerceObjectShape = (configField: ConfigField, type: F
...
+    if (configField.name && type === 'flow_action') {
+      throw new zod.ZodError([
+        {
+          code: zod.ZodIssueCode.custom,
+          path: ['settings.fields.name'],
+          message: 'Name is not a valid property for commerce object fields',
+        },
+      ])
+    }
+  }
```
omnisyle, line 48: instead of having to validate each properties when the devs use a commerce object, I feel it's better to have a zod schema for it at this point. It's a nice to have though.

```javascript
import { z } from "zod";

const FlowActionCommerceField = z.object({
  type: z.nativeEnum(SUPPORTED_COMMERCE_OBJECTS as const),
  description: z.string().optional(),
  required: z.boolean(),
});

FlowActionCommerceField.parse(configField);

```

MaxLeduc: Good idea!

isaacroldan: Yes! Let's do this please, zod all the validations 🙏

isaacroldan: You [can](https://zod.dev/?id=strings) customize all `zod` error messages: 
```ts
type: z.nativeEnum(SUPPORTED_COMMERCE_OBJECTS as const, {message: 'type must be....'}),
```

PR #2429, opened 2023-07-13
Allow passing -t to the create-test-app script

matteodepalo:
As per title, this is useful to test templates like `none`.


Arkham (approved):
LGTM!

PR #2438, opened 2023-07-14
Update javy-cli to 0.1.7 and javy to 0.1.1

Arkham:

What it says above ☝️


isaacroldan (approved):

nickwesselman:
I introduced a failure into the build process manually (by changing the esbuild output path), looks like javy errors now output, but the build doesn't halt. Is this an issue w/ the error code being output, or is CLI not checking for it? cc @jeffcharles 

image

Should we merge and track this separately?

PR #2442, opened 2023-07-14
Change update URLs prompt to be a yes/no question

Arkham:

Fixes https://github.com/Shopify/internal-cli-foundations/issues/732

Change update URLs prompt to be a yes/no question.

**Before**
Screenshot 2023-07-14 at 14 30 56

**After**
Screenshot 2023-07-17 at 10 54 29

### How to test your changes?

`bin/create-test-app.js -e ui`

### Measuring impact

### Checklist


gonzaloriestra (approved):
LGTM!

My only concern is that current users (not opting in to config in code) that were using the Yes/No options won't have a way to do the same now. But the config way is simple enough and the default solution from now on, so I guess it's an acceptable "breaking change".

PR #2451, opened 2023-07-17
Text confirmation prompts

amcaplan:

Introduces a dangerous confirmation prompt which requires a specific string input. This is used to grab the user's intention about a particularly dangerous operation.

Screenshot 2023-07-17 at 13 20 30

Screenshot 2023-07-17 at 13 21 03

Screenshot 2023-07-17 at 13 21 22

Screenshot 2023-07-17 at 13 21 39

### How to test your changes?

Kitchen sink includes an example.

### Measuring impact

### Checklist

### DIFF
--- a/packages/cli-kit/src/private/node/ui/components/TextPrompt.tsx
+++ b/packages/cli-kit/src/private/node/ui/components/TextPrompt.tsx
@@ -1,6 +1,5 @@
 import {TextInput} from './TextInput.js'
 import {InlineToken, LinkToken, TokenItem, TokenizedText, UserInputToken} from './TokenizedText.js'
-import {InfoMessage} from './SelectPrompt.js'
 import {handleCtrlC} from '../../ui.js'
 import useLayout from '../hooks/use-layout.js'
 import {messageWithPunctuation} from '../utilities.js'
@@ -18,8 +17,9 @@ export interface TextPromptProps {
     color?: TextProps['color']
     text: TokenItem<Exclude<InlineToken, UserInputToken | LinkToken>>
   }
+  exitOnEsc?: boolean
   successMessage?: string
-  onSubmit: (value: string) => void
+  onSubmit: (value: string | undefined) => void
   defaultValue?: string
   password?: boolean
   validate?: (value: string) => string | undefined
@@ -35,6 +35,7 @@ const TextPrompt: FunctionComponent<TextPromptProps> = ({
   message,
   infoTable,
   finalInstruction,
+  exitOnEsc = false,
   successMessage,
   onSubmit,
   validate,
@@ -71,6 +72,7 @@ const TextPrompt: FunctionComponent<TextPromptProps> = ({
   const displayedAnswer = displayEmptyValue ? emptyDisplayedValue : answerOrDefault
   const {exit: unmountInk} = useApp()
   const [submitted, setSubmitted] = useState(false)
+  const [cancelled, setCancelled] = useState(false)
   const [error, setError] = useState<string | undefined>(undefined)
   const shouldShowError = submitted && error
   const color = shouldShowError ? 'red' : 'cyan'
@@ -79,6 +81,14 @@ const TextPrompt: FunctionComponent<TextPromptProps> = ({
 
   useInput((input, key) => {
     handleCtrlC(input, key)
+
+    if (exitOnEsc && key.escape) {
+      setSubmitted(true)
+      setCancelled(true)
+      setError(undefined)
+      onSubmit(undefined)
+      unmountInk()
+    }
 
     if (key.return) {
       setSubmitted(true)
@@ -128,7 +138,17 @@ const TextPrompt: FunctionComponent<TextPromptProps> = ({
           </Box>
         </Box>
       ) : null}
-      {submitted && !error ? (
+      {cancelled ? (
+        <Box>
+          <Box marginRight={2}>
+            <Text color="red">{figures.cross}</Text>
+          </Box>
+
+          <Box flexGrow={1}>
+            <Text color="red">Cancelled</Text>
+          </Box>
+        </Box>
+      ) : submitted && !error ? (
         <Box>
           <Box marginRight={2}>
             <Text color="cyan">{figures.tick}</Text>
@@ -136,9 +156,7 @@ const TextPrompt: FunctionComponent<TextPromptProps> = ({
 
           <Box flexGrow={1}>
             <Text color="cyan" dimColor={displayEmptyValue}>
-              {successMessage ? successMessage :
-                password ? '*'.repeat(answer.length) :
-                displayedAnswer}
+              {successMessage ?? (password ? '*'.repeat(answer.length) : displayedAnswer)}
             </Text>
           </Box>
         </Box>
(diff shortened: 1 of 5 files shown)


packages/cli-kit/src/private/node/ui/components/TextPrompt.tsx
```diff
@@ -53,14 +62,15 @@ const TextPrompt: FunctionComponent<TextPromptProps> = ({
...
-  const {oneThird} = useLayout()
+  const {oneThird, twoThirds} = useLayout()
   const [answer, setAnswer] = useState<string>('')
   const answerOrDefault = answer.length > 0 ? answer : defaultValue
   const displayEmptyValue = answerOrDefault === ''
   const displayedAnswer = displayEmptyValue ? emptyDisplayedValue : answerOrDefault
   const {exit: unmountInk} = useApp()
   const [submitted, setSubmitted] = useState(false)
-  const [error, setError] = useState<string | undefined>(undefined)
+  const [cancelled, setCancelled] = useState(false)
```
matteodepalo, line 72: I think it would be better to introduce an enum for the state like in AutocompletePrompt. I'm also planning to refactor the state into a separate hook.

packages/cli-kit/src/private/node/ui/components/TextPrompt.tsx
```diff
@@ -25,6 +30,10 @@ export interface TextPromptProps {
 
 const TextPrompt: FunctionComponent<TextPromptProps> = ({
   message,
+  infoTable,
+  finalInstruction,
+  exitOnEsc = false,
```
matteodepalo, line 35: I wonder if this behavior can be applied to all prompts? What was the reason for introducing it only for dangerous confirmation prompts?

packages/cli-kit/src/public/node/ui.tsx
```diff
@@ -528,7 +528,65 @@ export async function renderTextPrompt({renderOptions, ...props}: RenderTextProm
...
+  throwInNonTTY({message: props.message, stdin: renderOptions?.stdin})
+
+  // eslint-disable-next-line prefer-rest-params
+  recordUIEvent({type: 'dangerousConfirmationPrompt', properties: arguments[0]})
+
+  const finalProps = {
+    finalInstruction: ['Type', {userInput: confirmation}, 'to confirm, or press Escape to cancel.'],
+    exitOnEsc: true,
+    validate: (value: string) => {
+      return value === confirmation ? undefined : ['Value must be exactly', {userInput: confirmation}]
```
matteodepalo, line 581: Same consideration here, if there is a passed in `confirmationText` param, couldn't we automatically add this check inside the component itself?

This btw made me think that perhaps we could create a separate component from `TextPrompt` to handle this, something like `DangerousConfirmationPrompt`. The reason is that there is a lot state now in `TextPrompt` that isn't necessary to regular text prompts and I'd expect the return value to be a boolean, but it's still a string instead.

matteodepalo (commented):
A few comments, but the UI looks great!

packages/cli-kit/src/private/node/ui/components/TextPrompt.tsx
amcaplan: It definitely makes sense to use a common pattern for this. However, text prompts are unusual in that it's hard to provide a "cancel" option when all typing just goes to the text field. So I think it makes sense to start here and wait for a known use case for another prompt.

Currently we assume that all other prompts will complete successfully, or the process will be killed. I think that makes downstream code easier to write clearly.

In the unique case of a dangerous confirmation prompt, what we're promising is either `true` or `false` as a return value, so usual boolean handling rules will apply. And this is the same as the behavior of a non-dangerous confirmation prompt.

In contrast, our other prompts might return a `string` or other type, and `string or undefined` gets messy when you're in the middle of a flow. Sticking with an explicit error might be best.

packages/cli-kit/src/public/node/ui.tsx
amcaplan: A lot of the same logic applies for the actual text prompt piece; the only real difference is validation + `exitOnEsc`. I can try splitting on some reasonable faultline, but if it seems too complicated, I suggest we live with the mess for now and split when it really becomes necessary.

I do like the idea of a text prompt component that returns true/false natively, so it's worth a go.

packages/cli-kit/src/private/node/ui/components/TextPrompt.tsx
amcaplan: The non-dangerous confirmation prompt gives you the ability to exit by selecting the `no` option. But there's no such option here, so we substitute with `esc`.

amcaplan:
@matteodepalo I rewrote this as a separate component and I'm now inclined to agree this makes much more sense. So actually at this point I'm not touching `TextPrompt` at all, just copied over the code and eliminated anything irrelevant to what we're doing in this component. I think the result is quite nice. WDYT?

amcaplan:
Another change I snuck in here: adding a cursor to text prompts in no-color mode (which also exists in regular mode but isn't visible due to foreground and background being the same color). The impetus was that eslint and refresh-documentation were fighting, but I think the result is quite nice.

Before: Where is my cursor?

Screenshot 2023-07-18 at 12 51 47

After: Aha, there you are!

Screenshot 2023-07-18 at 12 51 00

packages/cli-kit/src/private/node/ui/components/TextInput.tsx
```diff
@@ -45,27 +47,39 @@ const TextInput: FunctionComponent<TextInputProps> = ({
     return chalk.inverse(value[0]) + chalk.dim(value.slice(1))
   }
 
+  const cursorChar = figures.square
+  const defaultCursor = (
+    <Text color={color} backgroundColor={color}>
```
matteodepalo, line 52: Same color and backgroundColor?

packages/cli-kit/src/private/node/ui/components/TextInput.tsx
```diff
@@ -45,27 +47,39 @@ const TextInput: FunctionComponent<TextInputProps> = ({
...
-      : chalk.inverse(' ')
+      : defaultCursor
 
   // render cursor
   renderedValue = value
     .split('')
     .map((char, index) => {
       if (index === cursorOffset) {
-        return chalk.inverse(char)
+        return shouldDisplayColors() ? chalk.inverse(char) : cursorChar
```
matteodepalo, line 69: Is it possible to add a test case for this behavior with no colors?

amcaplan: 🤔

amcaplan: I haven't had much luck with stubbing `shouldDisplayColors`, there is too much weirdness in vitest module stubbing. Instead, I added an optional `noColor` boolean parameter, which matches [something I've done in the past](https://github.com/Shopify/cli/blob/b2ce9ffed4890b8999c702be7119cf1ac94107ec/packages/cli-kit/src/private/node/ui/components/Scrollbar.tsx#L21) for testing purposes as well.

packages/cli-kit/src/private/node/ui/components/TextInput.tsx
amcaplan: Done!

matteodepalo (approved):

PR #2453, opened 2023-07-17
Implement --reset for deploy

gonzaloriestra:

Fixes https://github.com/Shopify/internal-cli-foundations/issues/687

- Links a configuration from `deploy` when it's not found

### How to test your changes?

- Create a new app
- `p shopify app deploy`

### Measuring impact

### Checklist

### DIFF
--- a/packages/app/src/cli/services/context.ts
+++ b/packages/app/src/cli/services/context.ts
@@ -223,20 +223,12 @@ function buildOutput(app: OrganizationApp, store: OrganizationStore, cachedInfo?
   }
 }
 
-export interface DeployContextOptions {
-  app: AppInterface
-  apiKey?: string
-  reset: boolean
-  force: boolean
-  noRelease: boolean
-  commitReference?: string
-}
-
 export interface ReleaseContextOptions {
   app: AppInterface
   apiKey?: string
   reset: boolean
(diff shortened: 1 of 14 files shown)


packages/app/src/cli/services/context.ts
```diff
@@ -429,28 +431,36 @@ export async function fetchAppAndIdentifiers(
...
-    const apiKey = options.apiKey ?? options.app.configuration.client_id
-    partnersApp = await appFromId(apiKey, token)
-  } else if (options.reset) {
+  if (options.reset) {
     envIdentifiers = {app: undefined, extensions: {}}
+    if (isCurrentAppSchema(app.configuration)) {
+      const configuration = await link({directory: app.directory, commandConfig: options.commandConfig})
+      app.configuration = configuration
+    }
+  }
```
Arkham, line 449: I think we changed the logic a bit here, as previously choosing `--reset` would shortcut all the other options until the `fetchOrCreateOrganizationApp` call. Or am I misreading this?

gonzaloriestra: You are right, but I'm clearing some stuff, so it should work ™️ . I had missed one scenario, it should be fine now.

PR #2464, opened 2023-07-18
Update opt-in banner condition

JonathanJChang:

Fixes #0000 

### Measuring impact

### Checklist


mattnoakes (approved):
Looks great! I think you just need to add a changeset entry.

mattnoakes:
Yeah, it looks good to me!

raluka (approved):
Looks good! Not sure what's the agreement regarding testing in CLI, especially for bug fixes.

JonathanJChang:
> Looks good! Not sure what's the agreement regarding testing in CLI, especially for bug fixes.

I'm not too familiar with that either. Does anyone know what the best approach would be in this case? There doesn't seem to be an existing test for the banner from the original PR creator. I wonder if it is necessary in this instance as its quite a small change. Although, a test does sound appropriate.

mattnoakes:
> Does anyone know what the best approach would be in this case?

I thought the change was simple and temporary enough to not need testing.

PR #2479, opened 2023-07-19
Render DangerousConfirmationPrompt when removing an extension

amcaplan:

Fixes https://github.com/Shopify/app-deploys/issues/628

We want to warn more strongly when deploying dangerously.

Regular deploys are as before.

Screenshot 2023-07-19 at 12 56 26

Deploys that remove extensions are prompted in a way that requires a more deliberate response.

Screenshot 2023-07-19 at 12 55 46

### How to test your changes?

Deploy an app with an extension, then remove the extension and deploy again. You should see the dangerous prompt.

### Measuring impact

### Checklist

### DIFF
--- a/packages/app/src/cli/services/context/prompts.ts
+++ b/packages/app/src/cli/services/context/prompts.ts
@@ -8,6 +8,7 @@ import {
   InfoTableSection,
   renderAutocompletePrompt,
   renderConfirmationPrompt,
+  renderDangerousConfirmationPrompt,
   renderInfo,
 } from '@shopify/cli-kit/node/ui'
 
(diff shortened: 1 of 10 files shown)


packages/app/src/cli/services/context/prompts.ts
```diff
@@ -69,31 +71,38 @@ export async function deployConfirmationPrompt(
...
-
-  return renderConfirmationPrompt({
-    message: question,
-    infoTable,
-    confirmationMessage,
-    cancellationMessage: 'No, cancel',
-  })
+  const promptText = {message: question, infoTable}
+  const appExists = Boolean(appTitle)
+  const isDangerous = appExists && infoTable.some((section) => section.header.match(/Remove/))
```
matteodepalo, line 76: 🤔 this check makes sense but I can see it being a bit brittle if decide to change the copy. Could we pass a boolean from outside that more clearly marks the prompt as dangerous?

amcaplan: Good feedback! Changed the code to be more stable.

matteodepalo (approved):

PR #2486, opened 2023-07-19
add cache for duration of command lifecycle

ryancbahan:
Add a command cache to store state for the duration of the command lifecycle. Use this cache to decide whether to ask the config name.

It's becoming exceeding less efficient to pass state around to 5+ functions to avoid using module state. The app cache serves an entirely different purpose -- there are a number of cases where we don't yet have an app at all but still need to understand what step we're at in the command and what preceded. A lightweight cache for the command itself is a useful abstraction for this.

### How to test your changes?

Run `link` and re-link a previously linked file. It should skip the command name step.

### Measuring impact

### Checklist

### DIFF
--- a/packages/app/src/cli/prompts/dev.test.ts
+++ b/packages/app/src/cli/prompts/dev.test.ts
@@ -107,7 +107,7 @@ describe('selectApp', () => {
     })
   })
 
-  test('includes toml names when present', async () => {
+  test.only('includes toml names when present', async () => {
     vi.mocked(getTomls).mockResolvedValueOnce({
       [APP1.apiKey]: 'shopify.app.toml',
       [APP2.apiKey]: 'shopify.app.dev.toml',
--- a/packages/app/src/cli/prompts/dev.test.ts
+++ b/packages/app/src/cli/prompts/dev.test.ts
@@ -107,7 +107,7 @@ describe('selectApp', () => {
     })
   })
 
-  test.only('includes toml names when present', async () => {
+  test('includes toml names when present', async () => {
     vi.mocked(getTomls).mockResolvedValueOnce({
       [APP1.apiKey]: 'shopify.app.toml',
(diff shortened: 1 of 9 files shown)


packages/app/src/cli/utilities/app/config/getTomls.ts
```diff
@@ -8,7 +8,7 @@ export async function getTomls(apps: OrganizationAppsResponse, appDirectory?: st
     return {}
   }
 
-  const regex = /^shopify\.app(\.[-\w]+)?\.toml$/g
+  const regex = /^shopify\.app(\.[-\w]+)?\.toml$/
```
ryancbahan, line 11: /g makes the regex test stateful in sa way we don't want

packages/app/src/cli/prompts/dev.test.ts
```diff
@@ -107,7 +107,7 @@ describe('selectApp', () => {
     })
   })
 
-  test('includes toml names when present', async () => {
+  test.only('includes toml names when present', async () => {
```
rezaansyed, line 110: remove `.only` here

packages/app/src/cli/services/dev/select-app.ts
```diff
@@ -35,6 +36,12 @@ export async function selectOrCreateApp(
     return createApp(org, localAppName, token, options)
   } else {
     const selectedAppApiKey = await selectAppPrompt(apps, org.id, token, {directory: options?.directory})
+
+    const data = getCachedCommandInfo()
+    const tomls = (data?.tomls as {[key: string]: unknown}) ?? {}
+
+    if (tomls[selectedAppApiKey]) setCachedCommandInfo({selectedToml: tomls[selectedAppApiKey], askConfigName: false})
```
rezaansyed, line 43: I don't think we're ever setting `askConfigName` to `true` here?

PR #2503, opened 2023-07-20
Run use to select config when current config file does not exist

rezaansyed:

When a config file that was cached with `link` or `use` doesn't exist anymore, we want to allow them to choose a new file.

When the cached file does not exist, we call `use` which will let users choose another file.

### How to test your changes?

1. Link a config file
2. Rename it
3. Run any command that uses the current config, like `dev`, `deploy`, etc
4. 👀 See how you can now choose the correct file

image

### Measuring impact

### Checklist

### DIFF
--- a/packages/app/src/cli/models/app/loader.test.ts
+++ b/packages/app/src/cli/models/app/loader.test.ts
@@ -22,7 +22,7 @@ import {
   pnpmWorkspaceFile,
 } from '@shopify/cli-kit/node/node-package-manager'
 import {inTemporaryDirectory, moveFile, mkdir, mkTmpDir, rmdir, writeFile} from '@shopify/cli-kit/node/fs'
-import {joinPath, dirname, cwd} from '@shopify/cli-kit/node/path'
+import {joinPath, dirname, cwd, normalizePath} from '@shopify/cli-kit/node/path'
 import {platformAndArch} from '@shopify/cli-kit/node/os'
 import {outputContent} from '@shopify/cli-kit/node/output'
 // eslint-disable-next-line no-restricted-imports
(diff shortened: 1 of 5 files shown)


packages/app/src/cli/services/app/config/use.test.ts
```diff
@@ -205,6 +205,40 @@ describe('use', () => {
...
+      const {configurationPath, configuration} = testApp({}, 'current')
+      vi.mocked(loadAppConfiguration).mockResolvedValue({directory, configurationPath, configuration})
+      vi.mocked(getAppConfigurationFileName).mockReturnValue('shopify.app.something.toml')
+      createConfigFile(directory, 'shopify.app.something.toml')
+
+      // When
+      await use({directory, configName: 'something', warningMessage: "we're doomed. DOOMED."})
+
+      // Then
+      expect(renderWarning).toHaveBeenCalledWith({headline: "we're doomed. DOOMED."})
```
rezaansyed, line 221: LOL

packages/app/src/cli/models/app/loader.test.ts
```diff
@@ -1664,8 +1671,42 @@ automatically_update_urls_on_dev = true
...
+    const result = loadApp({directory: tmpDir, specifications, configName: 'non-existent'})
+
+    // Then
+    await expect(result).rejects.toThrow(`Couldn't find shopify.app.non-existent.toml in ${tmpDir}.`)
+    expect(use).not.toHaveBeenCalled()
+  })
+
   const runningOnWindows = platformAndArch().platform === 'windows'
 
+  test.skipIf(runningOnWindows)(
```
gonzaloriestra, line 1690: What's the problem with Windows here?

packages/app/src/cli/models/app/loader.ts
```diff
@@ -497,7 +498,17 @@ class AppConfigurationLoader {
   async loaded() {
     const appDirectory = await this.getAppDirectory()
     let configSource: LinkedConfigurationSource = this.configName ? 'flag' : 'cached'
-    this.configName = this.configName ?? getCachedAppInfo(appDirectory)?.configFile
+
+    const cachedCurrentConfig = getCachedAppInfo(appDirectory)?.configFile
+    const cachedCurrentConfigPath = cachedCurrentConfig ? joinPath(appDirectory, cachedCurrentConfig) : null
+
+    if (!this.configName && cachedCurrentConfigPath && !fileExistsSync(cachedCurrentConfigPath)) {
+      const warningMessage = `Could not find config file ${cachedCurrentConfig}, please select a new config`
```
gonzaloriestra, line 506: When you have a single config file, it doesn't actually ask you to select (but it just uses it directly), so that `please select a new config` is not very accurate. What about removing it? It will show something similar in the next line if needed.

```suggestion
      const warningMessage = `Could not find config file ${cachedCurrentConfig}`
```

gonzaloriestra (approved):
It works as expected and code LGTM! Just a couple of minor comments

packages/app/src/cli/models/app/loader.ts
rezaansyed: Update based on @jenstanicak's suggestions 🙂

packages/app/src/cli/models/app/loader.test.ts
shauns: I think this test should really be happening on Windows -- the code will be run there after all and this is really fundamental functionality? `normalizePath` should cover this IIRC

rezaansyed: All tests (including Windows) passing now

PR #2531, opened 2023-07-25
Default handle to directory name

isaacroldan:
For non-unified extensions the handle in the toml file is optional and we use a "default" one derived from the name in the toml. But this means that changing the name will change the handle which has undesired side effects (as the handle is used as the local identifier now).

Make the handle default to the basename directory, which is less likely to change and works better as a default identifier.

### Measuring impact

### Checklist


isaacroldan:
Closing, this is not needed anymore.

PR #2534, opened 2023-07-25
Fix Thank You page redirect URLs

lemonmade:
Fixes errors when using the dev console with a new-style thank-you page extension target:

Screenshot 2023-07-25 at 1 45 41 PM

I fixed this by just treating these as "checkout" extensions, like we do for the order status page.

### Checklist

### DIFF
--- a/packages/app/src/cli/services/dev/extension/utilities.ts
+++ b/packages/app/src/cli/services/dev/extension/utilities.ts
@@ -33,6 +33,11 @@ export function getExtensionPointTargetSurface(extensionPointTarget: string) {
         return 'post_purchase'
       }
 
+      // The Thank You page is rendered as part of the checkout application
+      if (page === 'thank-you') {
+        return 'checkout'
+      }
+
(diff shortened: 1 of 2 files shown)


packages/app/src/cli/services/dev/extension/utilities.ts
```diff
@@ -33,6 +33,11 @@ export function getExtensionPointTargetSurface(extensionPointTarget: string) {
         return 'post_purchase'
       }
 
+      // The Thank You page is rendered as part of the checkout application
+      if (page === 'thank-you') {
+        return 'checkout'
+      }
```
alxclark, line 39: Should the default be `checkout` for all purchase extension points and we can add exceptions if that's not the case ?

Just checking quickly, we're going to have the same issue with: `purchase.cart-line-item.line-components`

lemonmade: Good call, will update

PR #2540, opened 2023-07-27
Add -w option to pnpm upgrade command

matteodepalo:

Pnpm upgrade was broken
Screenshot 2023-07-27 at 16 51 51

Fix the above issue.

### DIFF
--- a/packages/cli-kit/src/public/node/node-package-manager.ts
+++ b/packages/cli-kit/src/public/node/node-package-manager.ts
@@ -338,6 +338,9 @@ interface AddNPMDependenciesIfNeededOptions {
 
   /** Abort signal to stop the process */
   signal?: AbortSignal
+
+  /** Whether to add the dependencies to the root package.json or to the package.json of the directory */
+  addToRootDirectory?: boolean
 }
 
 /**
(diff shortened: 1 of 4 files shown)


packages/cli-kit/src/public/node/node-package-manager.ts
```diff
@@ -504,8 +504,9 @@ function argumentsToAddDependenciesWithYarn(dependencies: string[], type: Depend
  * @returns An array with the arguments.
  */
 function argumentsToAddDependenciesWithPNPM(dependencies: string[], type: DependencyType): string[] {
-  let command = ['add']
+  let command = ['add', '-w']
```
alvaro-shopify, line 507: @matteodepalo  I have tophated this and there is an error when you try to run the  `upgrade` command if the app is not configured to use `workspaces`. I guess we should only add this flag in case the app use that feature.

image

matteodepalo: Good catch I'll add this check.

alvaro-shopify (approved):
🎩 and now everything works like a charm!

PR #2547, opened 2023-07-31
Add a rate-limiter dependency for partners requests

isaacroldan:
We need a way to limit requests to the API so that we don't hit the rate limit.
Partners API is limited to 10 requests per second and there are some special use cases that might launch more than that.

- Add a bottleneck feature to partners API, only 1 request is allowed every 100ms, if say 5 requests are launched at the same time, the bottleneck will process 1 every 100ms.

### How to test your changes?
- Create an app with 30-35 functions and try to deploy with `main` and with this branch. With main you should get a 429 error. With this branch it should work.

### Post-release steps
Reduce API rate-limit in partners to 10.

### Measuring impact

### Checklist


packages/cli-kit/src/public/node/api/partners.ts
```diff
@@ -2,7 +2,14 @@ import {graphqlRequest, GraphQLVariables, GraphQLResponse} from './graphql.js'
...
 import {setNextDeprecationDate} from '../../../private/node/context/deprecations-store.js'
 import {gql} from 'graphql-request'
+import {createRequire} from 'module'
 
+const require = createRequire(import.meta.url)
+// eslint-disable-next-line import/no-extraneous-dependencies
+const limiter = require('limiter')
+
+// Allow 1 request to partners every 110ms (the limit is 10 requests per second, adding a 10% margin)
+const partnersLimiter = new limiter.RateLimiter({tokensPerInterval: 1, interval: 110})
```
matteodepalo, line 12: Can't we do something like `limiter.RateLimiter({tokensPerInterval: 10, interval: 'second'})`? This would allow multiple requests below the limit of 10 to not be affected by the limit.

PR #2609, opened 2023-08-04
Error out of versions list and release commands early when beta disabled

mattnoakes:

Fixes https://github.com/Shopify/app-deploys/issues/561

Early returns with error messages are added to the `versions list` and `release` commands for when the deployments beta is disabled.
**Before**
`versions list`
old versions list

`release`
old release

**After**
`versions list`
new versions list

`release`
new release

### How to test your changes?

1. Create an app
2. Disable the `app_unified_deployment` beta flag
3. Run `pnpm shopify app versions list --path ~/path-to-your-app `
4. Run `pnpm shopify app release --path ~/path-to-your-app --version test`

### Measuring impact

### Checklist


packages/app/src/cli/services/versions-list.ts
```diff
@@ -86,6 +87,22 @@ async function getAppApiKey(token: string, options: VersionListOptions): Promise
   if (envIdentifiers.app) return envIdentifiers.app
```
amcaplan, line 87: In my local case, `envIdentifiers.app` exists so the result returns before your check.

I think the approach here needs to be modified:

1. It doesn't work as is
2. It doesn't feel right that `getAppApiKey` is where the error pops up, when actually it should complete.

Instead, I'd recommend fetching the partners app in `versionList`, doing your check there, and then passing the result to `getAppApiKey`. That way, the check always runs, and the error is being raised by a function whose concern is actually at the level of versions listing.

amcaplan (dismissed):
Changing the status of my review as this is actually not working just yet.

packages/app/src/cli/services/versions-list.ts
mattnoakes: Makes sense, thanks!

mattnoakes: > is there no way to save the selected app after you've run the app versions list command once?

Yeah, I think if the partners app is fetched using `fetchAppAndIdentifiers()` (like in https://github.com/Shopify/cli/pull/2609/commits/4ea9e77f312329f17fb8d133c9408e3bd6871a9a), some data will be cached so that users won't have to go through the prompts to select their app each time they run the command. Can someone from @Shopify/cli-foundations confirm whether this is correct and review these changes?

packages/app/src/cli/commands/app/versions/list.ts
```diff
@@ -27,6 +27,13 @@ export default class VersionsList extends Command {
       env: 'SHOPIFY_FLAG_CLIENT_ID',
       exclusive: ['config'],
     }),
+    reset: Flags.boolean({
```
amcaplan, line 30: Can you elaborate on why this is necessary? Why not always treat as `reset: false`?

packages/app/src/cli/services/versions-list.test.ts
```diff
@@ -45,88 +61,95 @@ describe('versions-list', () => {
...
-    expect(vi.mocked(partnersRequest)).toHaveBeenCalledWith(expect.anything(), 'token', {apiKey: 'env-app-id'})
-    expect(vi.mocked(fetchOrCreateOrganizationApp)).not.toHaveBeenCalled()
+    expect(renderCurrentlyUsedConfigInfo).toHaveBeenCalledWith({
+      org: 'name of org 1',
+      appName: 'app-title',
+      configFile: undefined,
+    })
   })
 
-  test('use client_id from linked toml if no env ids or flags', async () => {
```
amcaplan, line 81: It's a bit difficult to parse the diff, but it looks like this test and the next were deleted. Was that on purpose?

packages/app/src/cli/services/versions-list.ts
```diff
@@ -80,31 +81,42 @@ async function fetchAppVersions(
...
+        {
+          link: {
+            label: 'Simplified extension deployment',
+            url: 'https://shopify.dev/docs/apps/deployment/simplified-deployment',
+          },
+        },
+      ],
+    })
+    throw new AbortSilentError()
+  }
```
alvaro-shopify, line 112: I will move this to the `context.ts` in a new method that could be named `ensureListContext`. Beasides, I will refactor the common code used by `ensureListContext` and `ensureReleaseContext` because I guess that once `unified deployment` is rolled out 100% the verification could be removed from both commands

alvaro-shopify (approved):
🎩 and working good, great contribution! 🎉

packages/app/src/cli/commands/app/versions/list.ts
```diff
@@ -42,6 +49,6 @@ export default class VersionsList extends Command {
     const specifications = await loadLocalExtensionsSpecifications(this.config)
     const app: AppInterface = await loadApp({specifications, directory: flags.path, configName: flags.config})
 
-    await versionList({app, apiKey})
+    await versionList({app, apiKey, reset: flags.reset, commandConfig: this.config})
```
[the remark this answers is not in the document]
alvaro-shopify: @amcaplan thanks for the clarification! , what a weird scenario, because that dependency is used really deep in the calling tree

packages/app/src/cli/services/versions-list.test.ts
mattnoakes: Oh yeah, the diff is even confusing me 😅 Yeah, I deleted these tests because this PR removes [the `getAppApiKey()` function](https://github.com/Shopify/cli/blob/04e6341280afad9e8dc1d735d8c1050d0c9a1932/packages/app/src/cli/services/versions-list.ts#L83-L90) (which is what I believe is being tested in these tests).

jessehoek (approved):
Tophatted and looks good to me. Thanks Matt! Just a small thing, can we make the link go to the new page now, which is https://shopify.dev/docs/apps/deployment/extension instead of https://shopify.dev/docs/apps/deployment/simplified-deployment ?

It will still redirect, but it would be nice to use the final URL that the user will end up on.

packages/app/src/cli/services/versions-list.ts
alvaro-shopify: Hi @mattnoakes, yeah that's want I meant!, we could simply it a bit more if we move 
```ts
  const token = await ensureAuthenticatedPartners()
  const [partnersApp, envIdentifiers] = await fetchAppAndIdentifiers(options, token)
```

inside `checkDeploymentsBeta` and make the method to return the required values, but for me it's absolutely ok. 
Thanks again

PR #2619, opened 2023-08-07
Add CLI readme

matteodepalo:
Oclif has this neat `readme` command that generates a README.md file with the documentation of all the commands.
This is useful not just for developers browsing the repo, but also for AIs that need to understand how the CLI works.

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -130,6 +130,7 @@
       "@shopify/app",
       "@shopify/theme",
       "@shopify/plugin-did-you-mean",
+      "@shopify/cli-hydrogen",
       "@oclif/plugin-help",
       "@oclif/plugin-plugins",
       "@oclif/plugin-commands"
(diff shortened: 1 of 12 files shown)


packages/cli/package.json
```diff
@@ -130,7 +130,6 @@
       "@shopify/app",
       "@shopify/theme",
       "@shopify/plugin-did-you-mean",
-      "@shopify/cli-hydrogen",
```
isaacroldan, line 133: why removing hydrogen here?

matteodepalo: Isn't it not used anymore? I removed it because `oclif readme` was throwing an error saying that it couldn't find the `cli-hydrogen` plugin. It's not part of the CLI anymore right?

isaacroldan: it's not part of this repo, but it's part of the CLI.
Hydrogen is still a plugin of the CLI, but installed differently, you can see it [here](https://github.com/Shopify/hydrogen/blob/ed0c7e05c4c3c8b49c4829018e584c04217cafec/templates/hello-world/package.json#L16-L17) in their template

So the CLI package needs to know that a hydrogen plugin exists

isaacroldan (approved):

PR #2636, opened 2023-08-11
Fix upgrade command using a non default app config file

alvaro-shopify:
If your app includes a valid app config file `shopify.app.production.toml` but does not the default one  `shopify.app.toml`  this error is shown when you run the `upgrade` command

```
╭─ error ────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│                                                                                                                                                        │
│  Couldn't find the configuration file for /your/app/path, are you in a Shopify project directory?                          │
│                                                                                                                                                        │
╰────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
```

- Detect as app valid path those that includes
  - `shopify.app.toml`
  - `shopify.app.*.toml`

### How to test your changes?
- Modify the name of your default app config to something similar to `shopify.invalidapp.toml`, run the command `p shopify upgrade --path /path/to/your/app` and you should see an error
- Modify the name of your default app config to its default value `shopify.app.toml`, run the command `p shopify upgrade --path /path/to/your/app` and the command should run properly
-  Modify the name of your default app config to something similar to `shopify.app.production.toml`, run the command `p shopify upgrade --path /path/to/your/app` and the command should run properly

### Measuring impact

### Checklist

### DIFF
--- a/packages/cli/src/cli/services/upgrade.ts
+++ b/packages/cli/src/cli/services/upgrade.ts
@@ -8,8 +8,8 @@ import {
   usesWorkspaces,
 } from '@shopify/cli-kit/node/node-package-manager'
 import {exec} from '@shopify/cli-kit/node/system'
-import {dirname, moduleDirectory} from '@shopify/cli-kit/node/path'
-import {findPathUp} from '@shopify/cli-kit/node/fs'
+import {dirname, joinPath, moduleDirectory} from '@shopify/cli-kit/node/path'
+import {findPathUp, glob} from '@shopify/cli-kit/node/fs'
 import {AbortError} from '@shopify/cli-kit/node/error'
 import {outputContent, outputInfo, outputSuccess, outputToken, outputWarn} from '@shopify/cli-kit/node/output'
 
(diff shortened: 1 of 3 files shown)


packages/cli/src/cli/services/upgrade.ts
```diff
@@ -48,7 +48,12 @@ export async function upgrade(
 }
 
 async function getProjectDir(directory: string): Promise<string | undefined> {
-  const configFile = await findPathUp(['shopify.app.toml', 'hydrogen.config.js', 'hydrogen.config.ts'], {
+  const configFiles = ['shopify.app{,.*}.toml', 'hydrogen.config.js', 'hydrogen.config.ts']
```
isaacroldan, line 51: what does the `,` on `{,.*}` does?

alvaro-shopify: The comma , is necessary to separate the different options within the brace expansion {], in this case the empty string and any string that starts with '.'

isaacroldan: Wouldn't it work without it? `.*` should match empty strings too

PR #2637, opened 2023-08-11
Disable dev preview on exit

alvaro-shopify:

Extracted disable dev preview when dev command finishes

- Disable developer preview when `dev` commands finishes
-  Modify `ConcurrentOutput` component to finish the command (press `q` or `ctrl + c` using the `AbortController` instead of `treeKill`
- `dev` add a graceful finishing process to close every open external process. 
  - Include a timeout to force the finishing of the command
  - dev analytics is sent when the command is finished

### How to test your changes?
- Generate a Remix app with `theme app extension`, `checkout UI extension` and one `function`
- Run the `dev` command with the `--verbose` flag
- Check in the partners dashboard that the dev preview is enabled
- Press `q` or `CTRL+C` and check debug messages about the different child processes being stopped
- Check in the partners dashboard that the dev preview is disabled

### Measuring impact

### Checklist


mattnoakes:
Quitting dev with `q` worked as expected, but it didn't work when I tried with `CTRL+C`:

https://github.com/Shopify/cli/assets/65420305/a4042cf8-07b4-4455-9400-d44ffad1539b

I also have to enter `CTRL+C` twice to get the process to fully stop.

alvaro-shopify:
> Quitting dev with `q` worked as expected, but it didn't work when I tried with `CTRL+C`:
> 
>  11-26-r593f-q3y0v.mp4 
> I also have to enter `CTRL+C` twice to get the process to fully stop.

Hey @mattnoakes 👋 , thanks for jumping into this!, I will check it again

alvaro-shopify:
> > Quitting dev with `q` worked as expected, but it didn't work when I tried with `CTRL+C`:
> > 11-26-r593f-q3y0v.mp4
> > I also have to enter `CTRL+C` twice to get the process to fully stop.
> 
> Hey @mattnoakes 👋 , thanks for jumping into this!, I will check it again

Hey @mattnoakes!, I have already fixed it. If you have some free time I'd like you could test this again.

mattnoakes (commented):
Works great!! Can/should this also handle the case when I close my terminal application? 
Also, I'd wait for another 👍  from someone with a better understanding of the cli code than me 😅

jamieguerrero:
> Works great!! Can/should this also handle the case when I close my terminal application? Also, I'd wait for another 👍 from someone with a better understanding of the cli code than me 😅

Thanks for the review @mattnoakes and @raluka! I've discussed handling terminal close with @Arkham, and it seems like handling the `SIGHUP` process doesn't allow for awaiting async processes (probably because when you close a terminal, it terminates all processes and the design most likely doesn't allow for creating hanging additional async processes). @Arkham has also done some experiments with handling node processes and it seems like the behaviour isn't reliable, so we are handling only the best case scenario - that `q` and `ctrl + c` work.

mattnoakes (approved):

packages/cli-kit/src/private/node/tree-kill.ts
```diff
@@ -1,7 +0,0 @@
-import {printEventsJson} from './demo-recorder.js'
-import originalTreeKill from 'tree-kill'
-
-export function treeKill(signal: string) {
-  printEventsJson()
-  originalTreeKill(process.pid, signal)
-}
```
Arkham, line 7: why did we remove this? did we move it to public?

Arkham:
@matteodepalo could you take a look at the `ConcurrentOutput` changes in this PR?

packages/cli-kit/src/private/node/tree-kill.ts
jamieguerrero: Yes, I believe we did move it to public.

PR #2638, opened 2023-08-11
Fix rule that detects functions that create errors

Arkham:

I noticed that the `no-error-factory-functions` rule was broken

- fix the eslint rule
- fix all failures, mostly by inlining the error or creating subclasses


gonzaloriestra (approved):
Thanks!

PR #2696, opened 2023-08-23
Remove tunnelClient.stop()

matteodepalo:

We noticed that the tunnel was already getting killed by `treeKill` so there is no need to also stop it before.

Remove the tunnelClient handler in `abort`.

### How to test your changes?

- Run `dev`
- Check your processes for the presence of `cloudflared`
- Kill `dev` (either `q` or `ctrl+c`)
- That process should have disappeared

### DIFF
--- a/packages/app/src/cli/services/dev.ts
+++ b/packages/app/src/cli/services/dev.ts
@@ -349,7 +349,6 @@ async function dev(options: DevOptions) {
       previewUrl,
       app,
       abortController,
-      tunnelClient,
     })
   } else {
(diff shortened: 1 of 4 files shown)


jamieguerrero (approved):
LGTM! iirc, abortController.abort() should kill the process that the AbortController is passed into, and tunnelClient's [stopTunnel](https://github.com/Shopify/cli/blob/main/packages/plugin-cloudflare/src/tunnel.ts#L56) method isn't doing anything additional and is redundant since we already use treeKill to stop the dev process. 

🎩 works as expected.

PR #2703, opened 2023-08-24
Add localized description to the dev server kit

alfonso-noriega:

Enabling localized description for deving ui_extensions.

Fixes [#393](https://github.com/orgs/Shopify/projects/7275/views/3#:~:text=Update%20dev%2Dserver%2Dkit%20with%20localized%20description)

The optional description field is supported on the deploy flow for UI Extensions but there was this missing pice to be able to `dev` it.

Adding `description` and the translation logic to the websocket message payloads.

### How to test your changes?
- In the cli folder checkout this branch: `git checkout update-drafts-for-all-extensions-in-dev`
- Create a new app and admin extension (or reuse existing ones)
- Add a new locales entry for the description
- Add description to the toml file: `description: 't:new_key'`
- Run `shopify app dev --path your_app_folder_here`
- Open the dev console and navigate to the extension point of your extension
- Open the web inspector network tab and check the websocket messages contain the description

### Measuring impact

### Checklist

### DIFF
--- a/packages/ui-extensions-server-kit/src/ExtensionServerClient/ExtensionServerClient.ts
+++ b/packages/ui-extensions-server-kit/src/ExtensionServerClient/ExtensionServerClient.ts
@@ -87,6 +87,7 @@ export class ExtensionServerClient implements ExtensionServer.Client {
        *  extensionPoints: [{
        *    target: 'admin.product.item.action'
        *    name: 'en name'
+       *    description: 'en description'
        *    localization: {...}
        *  }],
        * }
@@ -220,8 +221,14 @@ export class ExtensionServerClient implements ExtensionServer.Client {
         localization,
         name:
           parsedTranslation && extension.name.startsWith('t:')
-            ? this._getLocalizedName(parsedTranslation, extension.name)
+            ? this._getLocalizedValue(parsedTranslation, extension.name)
             : extension.name,
+        ...(extension.description && {
+          description:
+            parsedTranslation && extension.description?.startsWith('t:')
+              ? this._getLocalizedValue(parsedTranslation, extension.description)
+              : extension.description,
+        }),
       }
 
       this.uiExtensionsByUuid[extension.uuid] = {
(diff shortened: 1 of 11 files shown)


vividviolet (commented):
Looks great - dev flow works. We're just missing a change here to get the description deployed: https://github.com/Shopify/cli/blob/df627c5683524e00aba9bb820820566b204294da/packages/app/src/cli/models/extensions/specifications/ui_extension.ts#L50

packages/ui-extensions-server-kit/src/ExtensionServerClient/ExtensionServerClient.ts
```diff
@@ -220,8 +221,14 @@ export class ExtensionServerClient implements ExtensionServer.Client {
...
           parsedTranslation && extension.name.startsWith('t:')
-            ? this._getLocalizedName(parsedTranslation, extension.name)
+            ? this._getLocalizedValue(parsedTranslation, extension.name)
             : extension.name,
+        ...(extension.description && {
+          description:
+            parsedTranslation && extension.description?.startsWith('t:')
+              ? this._getLocalizedValue(parsedTranslation, extension.description)
+              : extension.description,
+        }),
```
isaacroldan, line 231: probably don't need to check if there is a description, this should be safe:

```suggestion
          description:
            parsedTranslation && extension.description?.startsWith('t:')
              ? this._getLocalizedValue(parsedTranslation, extension.description)
              : extension.description,
```

alfonso-noriega: Wouldn't that end up having also a `description: undefined` value in the payload if there is no description?

vividviolet: It wouldn't make a difference since `description` is always optional. Consumers of the payload will always need to check if it's present so it doesn't matter if it's `undefined` or not set.

pnpm-lock.yaml
```diff
@@ -1,4 +1,4 @@
-lockfileVersion: '6.1'
+lockfileVersion: '6.0'
```
vividviolet, line 1: Do we need this change?

vividviolet (approved):
Looks good aside from the comment about the pnpm-lock. We also need a changeset. Can you generate one please?

pnpm-lock.yaml
isaacroldan: you can merge this, pnpm did something weird and updated this to 6.1 but it should be 6.0 👌

PR #2706, opened 2023-08-25
Show developer preview state change immediately

matteodepalo:

There are currently 2 UX issues with dev preview in `dev`.
- At the start the status flickers from "off" to "on" because in the beginning dev preview is turned off and gets automatically turned on
- When pressing `d` there is a delay so the user might think that nothing is happening

By showing the real value in the backend we're optimizing for the failure state. What I mean is that if turning on/off dev preview fails then the status will be correctly displayed without extra effort.
What I propose instead is to "lie" to the user by showing instant change and only revert back if the operation fails.

- At startup we immediately show `dev` preview as turned on. If turning it on fails we show an error message AND the correct "off" status.
- When pressing `d` we immediately toggle the state, reverting it to the previous value (with an error message) only when there is a problem.

### How to test your changes?

- Run `dev`
- Dev preview should be immediately on
- Press `d`
- Dev preview should immediately toggle

### DIFF
--- a/.changeset/four-walls-report.md
+++ b/.changeset/four-walls-report.md
@@ -1,0 +1,6 @@
+---
+'@shopify/cli-kit': patch
+'@shopify/app': patch
+---
+
+Remove dev preview enabled flickering status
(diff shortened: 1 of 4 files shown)


MeredithCastile:
If it's not too much bother, @matteodepalo, could you please add a screencast of the new experience? (Optionally, the old experience for posterity.)

Thanks for working on this DX! 🙌

matteodepalo:
@MeredithCastile yes that's correct

PR #2737, opened 2023-09-01
Update dev.yml since homebrew got deprecated

Arkham:

The `homebrew` section in `dev.yml` was deprecated in favour of `packages`.

### DIFF
--- a/dev.yml
+++ b/dev.yml
@@ -8,11 +8,9 @@ up:
       version: 3.1.2
   - bundler:
       gemfile: packages/cli-kit/assets/cli-ruby/Gemfile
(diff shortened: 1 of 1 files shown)


dev.yml
```diff
@@ -8,11 +8,9 @@ up:
       version: 3.1.2
   - bundler:
       gemfile: packages/cli-kit/assets/cli-ruby/Gemfile
-  - homebrew:
+  - packages:
     - jq
-    - delve
```
Arkham, line 13: this was only useful for go binaries, which we don't use anymore

isaacroldan (approved):

PR #2777, opened 2023-09-04
Allow spaces in scopes string

amcaplan:

Fixes https://github.com/Shopify/develop-app-management/issues/192

Currently we break on legacy schema but succeed on the current schema. We should just consistently allow spaces, since many users are likely to expect them to be allowed.

Removes spaces from the scopes string when loading the app.

### How to test your changes?

To be added

### Measuring impact

### Checklist

### DIFF
--- a/packages/app/src/cli/models/app/app.ts
+++ b/packages/app/src/cli/models/app/app.ts
@@ -7,11 +7,13 @@ import {getDependencies, PackageManager, readAndParsePackageJson} from '@shopify
 import {fileRealPath, findPathUp} from '@shopify/cli-kit/node/fs'
 import {joinPath} from '@shopify/cli-kit/node/path'
 
+const scopeString = zod.string().default('').transform(scopes => scopes.replace(/\s+,\s+/g, ','))
+
 export const LegacyAppSchema = zod
   .object({
     client_id: zod.number().optional(),
     name: zod.string().optional(),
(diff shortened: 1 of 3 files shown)


packages/app/src/cli/services/app/config/link.ts
```diff
@@ -200,6 +200,7 @@ const getAccessScopes = (appConfiguration: AppConfiguration, remoteApp: Organiza
     // if we can't find scopes or have to fall back, omit setting a scope and set legacy to true
   } else {
     return {
+      scopes: '',
```
amcaplan, line 203: Added this to solve a TS complaint, but I'm not sure if it makes sense.

packages/app/src/cli/models/app/app.ts
```diff
@@ -7,11 +7,13 @@ import {getDependencies, PackageManager, readAndParsePackageJson} from '@shopify
 import {fileRealPath, findPathUp} from '@shopify/cli-kit/node/fs'
 import {joinPath} from '@shopify/cli-kit/node/path'
 
+const scopeString = zod.string().transform((scopes) => scopes.replace(/\s+,\s+/g, ','))
```
isaacroldan, line 10: this won't work for `foo, bar`, should it be: `\s*,\s*` ?

PR #2782, opened 2023-09-05
Remove show/hide button for extensions

MaxCloutier:
With the introduction of automatic draft updates during dev for extensions, toggling between local and a deployed no longer provide any value and results in a confusing experience.

Closes https://github.com/Shopify/app-ui/issues/407

Remove show/hide button for extensions

### 🎩 instructions

- Follow steps to run the console locally here https://github.com/Shopify/cli/tree/main/packages/ui-extensions-dev-console
- Confirm that the show/hide button isn't present for extensions

### DIFF
--- a/.changeset/purple-trees-sin.md
+++ b/.changeset/purple-trees-sin.md
@@ -1,0 +1,5 @@
+---
+'@shopify/ui-extensions-dev-console-app': patch
+---
+
+Remove the show/hide button from console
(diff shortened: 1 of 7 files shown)


vividviolet (commented):
Can you remove the entire column instead? It is no longer relevant for any extension. Can you also add a changeset? Ideally we'd do a patch release so that users don't have to wait until the next stable release of the CLI but we should check with the CLI team

vividviolet:
> PostPurchaseRow

All extensions will have their draft versions automatically updated so the button won't do anything. In addition Post Purchase isn't actually configured to work with the Dev Console - they are currently using a browser extension with manual set up steps to connect the Dev Server.

MaxCloutier:
@vividviolet Sounds good! Done

vividviolet (approved):
Looks good 👍 

image

PR #2797, opened 2023-09-07
Always use npm in test app script

matteodepalo:
Testing with npm will help us catch bugs faster.

### DIFF
--- a/bin/create-test-app.js
+++ b/bin/create-test-app.js
@@ -74,21 +74,12 @@ program
     const appPath = path.join(homeDir, "Desktop", appName);
 
     switch (options.packageManager) {
-      case "npm":
       case "yarn":
       case "pnpm":
         nodePackageManager = options.packageManager;
         break;
-      case undefined:
-        nodePackageManager = "npm";
-        break;
       default:
-        log(
-          `Invalid package manager: ${
-            options.packageManager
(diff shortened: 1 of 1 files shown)


bin/create-test-app.js
```diff
@@ -80,11 +80,7 @@ program
...
         break;
       case undefined:
-        if (options.install === "local") {
-          nodePackageManager = "pnpm";
-        } else if (options.install === "nightly") {
-          nodePackageManager = "npm";
-        }
+        nodePackageManager = "npm";
         break;
       default:
```
gonzaloriestra, line 85: We can just set npm in the default section and forget about undefined, right?

matteodepalo: If we don't add an undefined case then we can't omit the flag. Or did you have another fix in mind? 
I added `    let nodePackageManager = "npm";` and removed the undefined case.

gonzaloriestra: Maybe I'm missing something, but my idea was:

```
switch (options.packageManager) {
  case "yarn":
  case "pnpm":
    nodePackageManager = options.packageManager;
    break;
  default:
    nodePackageManager = "npm";
}
```

gonzaloriestra: But this is an internal script, I think it's fine and the code is simpler. But just an opinion.

PR #2799, opened 2023-09-07
Update ink to 4.4.1

matteodepalo:

4.4.1 introduces a fix that we need for Windows and the `Tasks` component `useInput` usage.

Update ink to 4.4.1

### How to test your changes?

Run all the kitchen sink commands and see that they work correctly.

- `shopify kitchen-sink prompts`
- `shopify kitchen-sink async`
- `shopify kitchen-sink static`

- Run `dev` on a new app, see that the footer includes shortcuts.
- Run `pnpm shopify app generate extension` on a new app in Windows. It should work until the end.


Arkham (approved):
Code looks good and I tested using kitchen sink and `bin/create-test-app.js` :+1`

PR #2800, opened 2023-09-07
Dev component exits with exception when error received from processes

alvaro-shopify:
Right now if you run the `dev` command with an app with a theme app extension using a wrong Ruby version, the command exists silently without any error message.

When one of the process included in the `dev` command throws an exception the CLI exists silently

The exception should managed by the global error handler so among other things the error banner is displayed

When the `Dev` component catches an exception from one of the processes, the exception is sent with the abort signal so the ink library can rethrow it once the component is unmounted.

### How to test your changes?
- Change your Ruby version to an unsupported one (< `2.7.1`) 
- Run the dev command with an app with a theme app extension
- You should see an error like this
```
╭─ error ───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│                                                                                                                                                   │
│  Ruby version 2.6.10 is not supported                                                                                                             │
│                                                                                                                                                   │
│  Make sure you have at least Ruby 2.7.5 installed on your system. Documentation.                                                                  │
│                                                                                                                                                   │
╰───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
```

### Measuring impact

### Checklist

### DIFF
--- a/.changeset/perfect-points-agree.md
+++ b/.changeset/perfect-points-agree.md
@@ -1,0 +1,6 @@
+---
+'@shopify/cli-kit': patch
+'@shopify/app': patch
+---
+
+Dev component rethrow the exception to display error banner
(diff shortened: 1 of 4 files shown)


matteodepalo:
Really nice work!

PR #2803, opened 2023-09-07
Recognize surface for legacy customer account extension targets

jrrafols:

When inferring a surface from a target, only targets prefixed with `customer-account` are currently recognized as customer account extensions.

We are in the process of migrating 1p apps from the legacy format to the new format so we need to support the legacy targets (that are prefixed with `CustomerAccount`) in the interim.

Adds a fallthrough case statement for handling `customeraccount` in addition to `customer-account`.

### How to test your changes?

* Follow up to step 7 from [this doc](https://docs.google.com/document/d/1T5yHuJxGPg37IQhT7uNOcJb8Ned5GXc_y-rCOJDRcJU/edit#heading=h.ua5trnf4kn9g) to generate a new customer account extension
* Navigate to the app directory and use `pwd` to find the path
* Return to the `cli` directory and run `pnpm shopify app dev --path {path_to_your_app}`
* Once the dev server is running, press `p` to preview. Click on the extension (with the old target name) and ensure that the preview loads properly

### Measuring impact

### Checklist

### DIFF
--- a/packages/app/src/cli/services/dev/extension/server/utilities.test.ts
+++ b/packages/app/src/cli/services/dev/extension/server/utilities.test.ts
@@ -122,6 +122,23 @@ describe('getExtensionPointRedirectUrl()', () => {
     )
   })
 
+  test('returns customer account server URL if the extension point uses legacy customer account target', () => {
+    const extension = {
+      devUUID: '123abc',
+    } as ExtensionInstance
+
+    const options = {
+      storeFqdn: 'example.myshopify.com',
+      url: 'https://localhost:8081',
+    } as unknown as ExtensionDevOptions
+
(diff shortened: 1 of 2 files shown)


jrrafols:
> Is your new ui extension using both new and old targets? Otherwise I'm not sure I understand why we need this change as the old legacy extension would not hit the same code path as the new ui extensions

Yes, initially when we migrate the 1p apps, they will still be using the legacy targets. A following step will be to replace those targets with the new targets at which point the changes in this PR can be removed.

vividviolet:
I'm wondering if that's necessary? It's a lot easier to just migrate straight to the new targets no? Is this because it's not possible on the host side to support new targets right away?

vividviolet (approved):

PR #2815, opened 2023-09-12
Add more metadata for the config commands

matteodepalo:

As per https://github.com/Shopify/develop-app-management/issues/40 we want to track more metadata for `config` commands as well.

This PR adds a new function called `logMetadataForLoadedConfigContext` that tracks `partner_id` and `api_key`.


matteodepalo:
@shauns added a test for `use`.

shauns (approved):

PR #2818, opened 2023-09-12
Fix issue of redirecting customer account ui extension to checkout

lihaokx:

Resolves: https://github.com/Shopify/core-issues/issues/60345

In cli, it redirects all targets start from `customer-account.order-status` to checkout. [Code is here.](https://github.com/Shopify/cli/blob/main/packages/app/src/cli/services/dev/extension/utilities.ts#L44)

We should only redirect [these 4 checkout targets](https://github.com/Shopify/shopify/blob/main/components/apps/app/services/apps/models/ui_extension/schema/groups/checkout_ui.rb#L222-L227) to checkout instead of all of them

### How to test your changes?
1. Enable customer account and ui extension flags for your partner account.
2. Create a new customer account ui extension locally.
3. dev clone cli and checkout to this branch.
 `pnpm shopify app dev --path your_extension_app_path --reset`
4. Test different targets in your extension (by updating `shopify.extension.toml` and changing the target in the corresponding module.)
5. Ensure that the `customer-account.order-status...` targets specified in the PR redirect to checkout when you attempt to preview, while other targets redirect to customer accounts.
 
You can also check this video for 🎩  https://videobin.shopify.io/v/9gvo8P

### Measuring impact

### Checklist

### DIFF
--- a/packages/app/src/cli/services/dev/extension/utilities.test.ts
+++ b/packages/app/src/cli/services/dev/extension/utilities.test.ts
@@ -19,9 +19,9 @@ describe('getExtensionPointTargetSurface()', () => {
     expect(getExtensionPointTargetSurface('purchase.thank-you.cart-line-list.render-after')).toBe('checkout')
   })
 
-  test('returns "checkout" for a UI extension targeting customer-account.order-status.*', async () => {
+  test('returns "checkout" for a checkout UI extension target that starts with customer-account', async () => {
     expect(getExtensionPointTargetSurface('customer-account.order-status.block.render')).toBe('checkout')
-    expect(getExtensionPointTargetSurface('customer-account.order-status.contact-information.render-after')).toBe(
+    expect(getExtensionPointTargetSurface('customer-account.order-status.customer-information.render-after')).toBe(
       'checkout',
     )
     expect(getExtensionPointTargetSurface('customer-account.order-status.cart-line-item.render-after')).toBe('checkout')
--- a/packages/app/src/cli/services/dev/extension/utilities.ts
+++ b/packages/app/src/cli/services/dev/extension/utilities.ts
@@ -40,8 +40,7 @@ export function getExtensionPointTargetSurface(extensionPointTarget: string) {
     // Covers Customer Accounts UI extensions (future)
     case 'customeraccount':
     case 'customer-account': {
-      // These targets are rendered by Checkout
-      if (page === 'order-status') {
+      if (isCheckoutRenderedCustomerAccountTarget(extensionPointTarget)) {
         return 'checkout'
       }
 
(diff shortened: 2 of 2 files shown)


packages/app/src/cli/services/dev/extension/utilities.ts
```diff
@@ -58,3 +58,14 @@ export function getExtensionPointTargetSurface(extensionPointTarget: string) {
...
+function isCheckoutTargetsBeginsWithCustomerAccount(extensionPointTarget: string) {
+  const checkoutTargetsBeginsWithCustomerAccount = new Set([
+    'customer-account.order-status.cart-line-item.render-after',
+    'customer-account.order-status.cart-line-list.render-after',
+    'customer-account.order-status.customer-information.render-after',
+    'customer-account.order-status.block.render',
+  ])
+
+  return checkoutTargetsBeginsWithCustomerAccount.has(extensionPointTarget)
+}
```
alxclark, line 71: What should the DX look like when developing these extensions ? If a shop has no order, is it proper to send you to customer account although you need to go and complete a checkout first ?

Once all of those extension points are on ODP do you expect us to align the behaviour ?

lihaokx: If a shop does not have an order, it may make sense to redirect to checkout page.
If a shop already has an order, it does not make any sense to redirect to checkout page. 

The issue is that cli does not know if the shop has an order or not. 

These are good questions. I will have a alignment with my team about this.

packages/app/src/cli/services/dev/extension/utilities.test.ts
```diff
@@ -19,9 +19,9 @@ describe('getExtensionPointTargetSurface()', () => {
     expect(getExtensionPointTargetSurface('purchase.thank-you.cart-line-list.render-after')).toBe('checkout')
   })
 
-  test('returns "checkout" for a UI extension targeting customer-account.order-status.*', async () => {
+  test('returns "checkout" for a checkout UI extension target starts with customer-account', async () => {
```
jrrafols, line 22: nit
```suggestion
  test('returns "checkout" for a checkout UI extension target that starts with customer-account', async () => {
```

jrrafols (approved):
I made a couple changes to update naming, but logic is the same as originally implemented by @lihaokx.

🎩 ✅

packages/app/src/cli/services/dev/extension/utilities.ts
```diff
@@ -58,3 +57,14 @@ export function getExtensionPointTargetSurface(extensionPointTarget: string) {
...
   }
 }
+
+function isCheckoutRenderedCustomerAccountTarget(extensionPointTarget: string) {
+  const customerAccountCheckoutRenderedTargets = new Set([
+    'customer-account.order-status.cart-line-item.render-after',
+    'customer-account.order-status.cart-line-list.render-after',
+    'customer-account.order-status.customer-information.render-after',
+    'customer-account.order-status.block.render',
+  ])
```
alxclark, line 67: Nit: Since this is a constant you could extract it to avoid creating a new set on every function call.

alxclark (approved):

packages/app/src/cli/services/dev/extension/utilities.ts
jrrafols: Moving these to a constant made the separate function redundant so the check is now being performed directly in `getExtensionPointTargetSurface`

PR #2828, opened 2023-09-14
Make it so dependabot groups together minor and patch updates into a single PR

Arkham:

Currently dependabot opens a zillion PRs

Make dependabot open a single PR with all the updates


amcaplan (approved):
looks good to me!

PR #2832, opened 2023-09-15
Fix integrated GraphiQL tool in spin

alvaro-shopify:
Running the integrated GraphiQL toll inside spin is returning an error related to shop domains validation

- When the app is run inside spin, set the proper `customShopDomains` when the `shopifyApi` is created from the `graphiQL server`

### How to test your changes?
- The PR should be merged to test it inside `spin`
- Check that the app and the graphiQL tool is working properly as usual against the PRO environment

### Measuring impact

### Checklist

### DIFF
--- a/packages/app/src/cli/services/dev/processes/graphiql.ts
+++ b/packages/app/src/cli/services/dev/processes/graphiql.ts
@@ -1,6 +1,7 @@
 import {BaseProcess, DevProcessFunction} from './types.js'
 import {urlNamespaces} from '../../../constants.js'
 import {setupGraphiQLServer} from '../graphiql/server.js'
+import {isSpinEnvironment, spinFqdn} from '@shopify/cli-kit/node/context/spin'
 
 interface GraphiQLServerProcessOptions {
   appName: string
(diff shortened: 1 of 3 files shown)


packages/app/src/cli/services/dev/processes/graphiql.ts
```diff
@@ -19,13 +21,15 @@ export interface GraphiQLServerProcess extends BaseProcess<GraphiQLServerProcess
 }
 
 export async function setupGraphiQLServerProcess(
-  options: Omit<GraphiQLServerProcessOptions, 'port'>,
+  options: Omit<GraphiQLServerProcessOptions, 'port' | 'shopCustomDomain'>,
```
shauns, line 24: Why are you passing in shopCustomDomain, but then ignoring it afterward?

alvaro-shopify: This method was using the same type for the `options` argument and for the `options` filed included in the response. The problem is that the field is mandatory for the response but not for the parameter. 
I guess that something similar was happening with the `port` field, that's why I applied the same strategy

PR #2833, opened 2023-09-15
Validate function's ui.handle matches local extension in app

mssalemi:

Fixes https://github.com/Shopify/script-service/issues/7308

This PR will add an extra validation when deploying a app that contains a function and an ui extension. 

For validating when an app that contains a function with `ui.handle`:
- The function with a `ui.handle` must contain another extension with a matching `handle` in the same local app. This will be used to link a UI extension with a function, so we can [surface that information to domains](https://github.com/Shopify/shopify/pull/452100). 
- We should ensure the matching extension is of type `ui_extension`. If we match solely on the existence of the handle, we could have a situation where partners link a function `ui_handle` to another extension type. 

This validation should check that if a function has the `ui.handle` configured, that it should match an existing extension handle. That matching extension should also be `ui_extension` type. These validation will ensure that partners are correctly linking a function with a ui_extension.

### What should reviewers focus on:
- wording of error message
- followed guidance from this [slack convo](https://shopify.slack.com/archives/C030LHFCA5T/p1694763056516339?thread_ts=1694714721.883739&cid=C030LHFCA5T) to implment

### How to test your changes?

enable `UI Extension` beta flag on organization via [partners internal](https://partners.script-service-q7k4.mehdi-salemi.us.spin.dev/internal/organizations/1).

Enable `functions_app_ui_handle` on app 

**core**
```
dev c
> api_client = ApiClient.find(1830457)
> api_client.beta.enable(:functions_app_ui_handle) 
```

Deploy an app with a function and ui Extension. The function's `ui.handle` should match the handle for the `ui extension`
```
SPIN=1 SPIN_INSTANCE=script-service-q7k4 dev spin pnpm shopify app deploy  --path=../../../../Desktop/prod-testing/product-discounts-single-ext
```
^ _updating with your spin instance urls and path to your local app_

### Measuring impact

### Checklist

### DIFF
--- a/packages/app/src/cli/models/app/app.test.ts
+++ b/packages/app/src/cli/models/app/app.test.ts
@@ -6,8 +6,11 @@ import {
   getUIExtensionRendererVersion,
   isCurrentAppSchema,
   isLegacyAppSchema,
+  validateFunctionExtensionsWithUiHandle,
 } from './app.js'
-import {DEFAULT_CONFIG, testApp, testUIExtension} from './app.test-data.js'
+import {DEFAULT_CONFIG, testApp, testUIExtension, testFunctionExtension} from './app.test-data.js'
+import {ExtensionInstance} from '../extensions/extension-instance.js'
+import {FunctionConfigType} from '../extensions/specifications/function.js'
 import {describe, expect, test} from 'vitest'
 import {inTemporaryDirectory, mkdir, writeFile} from '@shopify/cli-kit/node/fs'
 import {joinPath} from '@shopify/cli-kit/node/path'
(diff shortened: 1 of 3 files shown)


packages/app/src/cli/models/app/app.ts
```diff
@@ -252,6 +264,35 @@ export class App implements AppInterface {
...
+  const errors: string[] = []
+
+  functionExtensionsWithUiHandle.forEach((extension) => {
+    const uiHandle = extension.configuration.ui?.handle!
+
+    const matchingExtension = findExtensionByHandle(allExtension, uiHandle)
+    if (!matchingExtension) {
+      errors.push(`[${extension.name}] - Local app must contain a ui_extension with handle '${uiHandle}'`)
+    } else if (matchingExtension?.configuration?.type !== 'ui_extension') {
+      errors.push(`[${extension.name}] - Linked Extension must of type 'ui_extension'`)
```
dphm, line 285: `must of type` 😆

packages/app/src/cli/models/app/app.test.ts
```diff
@@ -200,6 +202,80 @@ describe('getAppScopesArray', () => {
...
+    type: type || 'product_discounts',
+    ui: {
+      handle: uiHandle || 'test-ui-handle',
+    },
+  })
+
+  describe('return undefined when app configuration is valid', () => {
+    test('when no function extensions exist', async () => {
+      const app = await testApp()
+      expect(validateFunctionExtensionsWithUiHandle(app.allExtensions)).toBeUndefined()
```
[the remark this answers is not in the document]
alvaro-shopify: does it make sense to run this validation if the `app` does not have the beta `functions_app_ui_handle` enabled?

alvaro-shopify (approved):

packages/app/src/cli/models/app/app.ts
```diff
@@ -241,6 +244,22 @@ export class App implements AppInterface {
     this.nodeDependencies = nodeDependencies
   }
 
+  async preDeployValidation() {
+    const functionExtensionsWithUiHandle = this.allExtensions
+      .filter((ext) => ext.isFunctionExtension)
+      .map((ext) => ext as unknown as ExtensionInstance<FunctionConfigType>)
+      .filter((ext) => ext.configuration.ui?.handle)
```
andrewhassan, line 251: The extra `map` and `filter` calls add some runtime overhead (because it's iterating over the list multiple times). Even though it's small, we should probably still remove them since they're only here for TypeScript 😛 

Could we do something like this?

```suggestion
      .filter((ext) => ext.isFunctionExtension && (ext.configuration as unknown as FunctionConfigType).ui?.handle)
```

packages/app/src/cli/models/app/app.ts
```diff
@@ -252,6 +271,32 @@ export class App implements AppInterface {
...
 }
 
+export function validateFunctionExtensionsWithUiHandle(
+  functionExtensionsWithUiHandle: ExtensionInstance<FunctionConfigType>[],
+  allExtension: ExtensionInstance[],
+): string[] | undefined {
+  const errors: string[] = []
+
+  functionExtensionsWithUiHandle.forEach((extension) => {
+    const uiHandle = extension.configuration.ui?.handle!
```
andrewhassan, line 281: Nit: We know that `ui.handle` must be in these extensions, so I think we want:

```suggestion
    const uiHandle = extension.configuration.ui!.handle!
```

packages/app/src/cli/models/app/app.ts
```diff
@@ -252,6 +271,32 @@ export class App implements AppInterface {
...
+
+    const matchingExtension = findExtensionByHandle(allExtension, uiHandle)
+    if (!matchingExtension) {
+      errors.push(`[${extension.name}] - Local app must contain a ui_extension with handle '${uiHandle}'`)
+    } else if (matchingExtension?.configuration?.type !== 'ui_extension') {
```
andrewhassan, line 286: Nit: `matchingExtension` should not be falsey here

```suggestion
    } else if (matchingExtension.configuration?.type !== 'ui_extension') {
```

PR #2836, opened 2023-09-15
Make dev2 default

Arkham:

We're pretty confident that the dev refactor did not introduce regressions.

Remove the old code pathway.

### How to test your changes?

```
bin/create-test-app.js --deploy
```

### Measuring impact

### Checklist


amcaplan:
I think our internal testing is not equivalent to external testing. What about making `dev2` the default in 3.50, and then removing old dev entirely for 3.51?

That would allow us to add all new features to `dev2` only, while keeping a failsafe in case users have some really bad unforeseen problems.

To make the final migration easier, maybe we can change for now:

```
dev => dev-old
dev2 => dev
```

So it'll be a much smaller patch change when we actually delete old dev.

Arkham:
I considered that, but since no issues have arised internally I preferred to remove the old pathway, since it's already proving a bit annoying to implement new features in both the old and new pathway. We can always fix bugs with patch releases :)

matteodepalo:
I'd agree with the bolder approach of just making the switch. If we see bugs we can fix them, but for now we already have to make sure that all changes that go into `dev` are mirrored in `dev2` and that is going to be a cause of bugs for sure if we keep doing it for long enough. @Arkham Have you manually checked all the changes that went into dev after the last update of dev2?

matteodepalo (approved):

PR #2848, opened 2023-09-18
Update URLs even without any webs, to enable GraphiQL

amcaplan:

Functions need GraphiQL to accomplish various tasks. But in extension-only apps, we are not updating OAuth callback URLs, so GraphiQL isn't available.

Prompts if necessary, and updates URLs regardless of whether there are webs. This way, extension-only apps can still run GraphiQL.

Note that at least one access scope must be defined; otherwise, GraphiQL won't appear.

### How to test your changes?

1. Create an extension-only app.
2. Add a function. (recommendation: product discount)
3. Add an access scope. (recommendation: `write_discounts`)
4. Run `dev`. You should be prompted to approve updating the URLs, and you should be able to open GraphiQL.

### Measuring impact

### Checklist


gonzaloriestra (approved):
👌

amcaplan:
I've updated this PR to be a draft, as @RyanDJLee pointed out it would break the current auth flow for extension-only apps. So instead, we're looking at alternate approaches, and this PR will likely be closed. ([slack](https://shopify.slack.com/archives/C05E3BDFDRB/p1695141003704339?thread_ts=1695127826.504449&cid=C05E3BDFDRB))

amcaplan:
I think we can close this, as we're looking at other, better options that will not break existing functionality.

PR #2863, opened 2023-09-19
Add handle persistence to app deploy mutations

pt2pham:

Part of: https://github.com/Shopify/develop-app-foundations/issues/26
Requires: https://github.com/Shopify/shopify/pull/451765, https://github.com/Shopify/partners/pull/50186, https://github.com/Shopify/partners/pull/50232

This PR is being added to persist `handle` from the CLI so that any `shopify.extension.toml` that specifies `handle` will send it up to Core so that we can persist it server-side.

### How to test your changes?

1. Use branches in the aforementioned PRs
2. Set up `spin` to create an app with the CLI and generate an Admin Action extension 
3. Run `shopify app deploy` for your app to deploy and release the changes
4. Verify in the core DB that the `handle` is persisted for an extension version

### Measuring impact

### Checklist


packages/app/src/cli/services/dev/create-extension.ts
```diff
@@ -35,6 +35,7 @@ export async function createExtension(
     title: name,
     config: JSON.stringify({}),
     context: null,
+    handle: name,
```
pt2pham, line 38: It looks like the `name` that gets passed into `createExtension` comes from either `extension.localIdentifier` or `extension.handle`. 

I took a look at the constructor for that class (`ExtensionInstance`), and it looks like `localIdentifier` is set by `handle` as well. 

https://github.com/Shopify/cli/blob/61862fec847b4843b81eb00c1c304b24121038c7/packages/app/src/cli/models/extensions/extension-instance.ts#L120-L121C10

What's the exact difference between the two?

cc: @isaacroldan

isaacroldan: the `localIdentifier` is a legacy property that we use to (as the name says) locally identify an extension. We used to use the folder name, but since the introduction of unified config we migrated to use the handle as the local identifier because a folder can contain multiple extensions now.
So right now `localIdentifier` === `handle`
And since the handle is optional, if there isn't one, we `slugify` the name.

For this specific method i'd say to pass the `handle` instead of the name to be more explicit:

```ts
export async function createExtension(
  apiKey: string,
  graphQLType: string,
  handle: string,
  token: string,
): Promise<ExtensionRegistration> {
  const query = ExtensionCreateQuery
  const variables: ExtensionCreateVariables = {
    apiKey,
    type: graphQLType,
    title: handle,
    config: JSON.stringify({}),
    context: null,
    handle: handle,
```

murray-andrew-r (approved):

pt2pham:
> If the changes are user-facing, run "pnpm changeset add" to track your changes and include them in the next release CHANGELOG.

Would it be right to say that these changes aren't user-facing? Users will benefit from the persistence of `handle`, and in particular, it will be useful for functions once they enable app module interoperability, but this change should be invisible to the user since we're not forcing any change. @murray-andrew-r @isaacroldan

PR #2867, opened 2023-09-20
Fix `cmd_app_linked_config_source` not being accurate

matteodepalo:

Fixes https://github.com/Shopify/develop-app-management/issues/187

For `dev` we were always passing `configName` from the outside into the config log functions so `cached` was never being used.

We've removed `default` because it creates no value especially considering `cmd_app_linked_config_name` exists and we can always check that its value is `shopify.app.toml`

We've also removed the cached `configName` in `dev` so we always fetch it in the config specific functions.

### How to test your changes?

- Run `dev`
- Press `q`
- See that the value is correctly `cached`
- Run it again with `--config something`
- Press `q`
- See that the value is `flag`


alvaro-shopify (approved):
LGTM

PR #2887, opened 2023-09-26
Client credentials flow for GraphiQL

amcaplan:

Fixes https://github.com/Shopify/develop-app-management/issues/1494

Replaces the existing, somewhat messy solution for authentication (OAuth with redirect to GraphiQL) with the client credentials flow that can happen entirely behind the scenes.

This means:

* No messing with OAuth callback URLs
* GraphiQL works for extension-only apps

Replaces the existing solution with a behind-the-scenes [client credentials OAuth flow](https://www.oauth.com/oauth2-servers/access-tokens/client-credentials/).

In case the user isn't logged in yet, we've added a page instructing them to click a link to install the app in a new tab. Once the app is installed, the tab will be closed and they'll be redirected to GraphiQL.

There have been a few other related changes:

* The dependency on `@shopify/shopify-api` was removed. We are now getting API versions straight from the source of truth, and requests are being forwarded to the API in a more raw fashion. (I believe this should work the same in spin, but we should be sure of that. I know the npm package did have some special spin handling.)
* Thanks to dropping the dependency, app scopes are no longer required!
* Added an explanatory warning when credentials can't be refreshed
* Styling the page required exposing from `cli-kit` the `style.css` asset we use to show simple pages like the post-login page.
* The GraphiQL URL was added to the sticky footer.
* CLI kit was modified to have some extra available functions for getting information on available API versions. This was mostly just code shuffling and exposing new functions. No APIs should have been broken.

### How to test your changes?

1. Enable the client_credentials [beta flag](https://app.shopify.com/services/internal/app_betas/client_credentials) for your app (this requirement should go away in the next few days)
2. `p shopify app dev --path /path/to/your/app`
3. Wait for the app to load, and tap `g`

Be sure to try with the app previously installed and uninstalled.

### Measuring impact

### Checklist

### DIFF
--- a/.changeset/eleven-plants-argue.md
+++ b/.changeset/eleven-plants-argue.md
@@ -1,0 +1,6 @@
+---
+'@shopify/cli-kit': minor
+'@shopify/app': minor
+---
+
+Support client credentials flow (currently used in GraphiQL)
--- a/.changeset/thick-dolls-visit.md
+++ b/.changeset/thick-dolls-visit.md
@@ -1,0 +1,5 @@
+---
+'@shopify/cli-kit': minor
+---
+
+Expose functions to get information on available Admin API versions


packages/app/src/cli/services/dev/graphiql/server.ts
```diff
@@ -87,87 +53,141 @@ export function setupGraphiQLServer({
...
-    } catch (error) {
-      return (error as HttpResponseError)?.response?.code !== 401
+      outputDebug('refreshing token', stdout)
+      _token = undefined
+      const queryString = queryStringEncode({
+        client_id: apiKey,
+        client_secret: apiSecret,
+        grant_type: 'client_credentials',
+      })
+      const tokenResponse = await fetch(`https://${storeFqdn}/admin/oauth/access_token?${queryString}`, {
```
teddyhwang, line 74: are there any plans to have this in the `@shopify/shopify-api` library?

packages/app/src/cli/services/dev/ui/components/Dev.tsx
```diff
@@ -46,7 +46,8 @@ const Dev: FunctionComponent<DevProps> = ({
   const {canEnablePreviewMode, developmentStorePreviewEnabled} = app
   const {isRawModeSupported: canUseShortcuts} = useStdin()
   const pollingInterval = useRef<NodeJS.Timeout>()
-  const [statusMessage, setStatusMessage] = useState(`Preview URL: ${previewUrl}`)
+  const defaultStatusMessage = `Preview URL: ${previewUrl}${graphiqlUrl ? `\nGraphiQL URL: ${graphiqlUrl}` : ''}`
```
shauns, line 49: Is this correct? Given that GraphiQL is not always enabled?

packages/app/src/cli/services/dev/graphiql/server.ts
shauns: I agree it's good to not have the full shopify-api dependency. Maybe worth moving it to CLI Kit though to start with?

packages/app/src/cli/services/dev/graphiql/server.ts
```diff
@@ -87,87 +53,141 @@ export function setupGraphiQLServer({
...
+  })
+
+  async function fetchApiVersionsWithTokenRefresh(): Promise<string[]> {
+    let apiVersions: string[]
+    try {
+      apiVersions = await supportedApiVersions({storeFqdn, token: await token()})
+      // eslint-disable-next-line no-catch-all/no-catch-all
+    } catch (err) {
+      // Retry once with a new token, in case the token expired or was revoked
+      await refreshToken()
```
shauns, line 103: Feel free to pass on this, not essential

We have a few places now where we have a pattern of:

```
try {
  return doAThing(token)
} catch (err) {
  token = await refreshToken()
  return doAThing(token)
}
```

I wonder if we can provide a helper for this somewhere?

shauns (approved):

packages/app/src/cli/services/dev/graphiql/server.ts
amcaplan: Meaning, it's easy to build this structure incorrectly especially with async/await. Like so:

```js
// doAThing is async, this is wrong:

async function myFunc() {
  try {
    // Usually you can return a promise fro an async function,
    // but in this case if it fails there will be no retry!
    return doAThing(token)
  } catch (err) {
    token = await refreshToken()
    return doAThing(token)
  }
}

// should be:

async function myFunc() {
  let retval
  try {
    retval = doAThing(token)
  } catch (err) {
    token = await refreshToken()
    retval = doAThing(token)
  }
  return retval
}

```

So it would be good to have as a util. Sorry I was unclear before!

shauns: Fine by me to tackle it in a separate thing

PR #2889, opened 2023-09-26
Dev Preview flag for theme check

### LINKED ISSUE #1839: Migrate @shopify/cli theme check command to @shopify/theme-check-node
```[tasklist]
### Tasks
- [ ] https://github.com/Shopify/theme-check-js/issues/93
- [ ] Migrate `theme check` commands to @shopify/theme-check-node
```

albchu:

Closes #1839. This PR introduces the typescript theme check into the CLI behind the flag `--dev-preview`. This is our new theme check 2.0. Eventually we will be sunsetting the ruby theme check and replacing it wholly with the typescript theme check.

This change to `shopify theme check` adds the new `@shopify/theme-check-node` package to use theme check 2.0 behind `--dev-preview`.

It also supports all ruby theme check flags (excluding a few cases that no longer make sense. Those exception flags are:
- `category` and `exclude-category`: These dropped support because we can control checks with better granularity through the `--config` yml
- `update-docs`: Previously this was a manual operation, in the new theme check, these sources of truth are automagically updated on `shopify theme check` invocation whenever a new docs revision exists.

https://github.com/Shopify/cli/assets/5607117/3a77b071-29ef-42f2-b019-41b81b4ec166

### How to test your changes?

```bash
# Setup CLI
pnpm install

# Setup: Add dawn to test theme check
pnpm shopify theme init --path ./tophat-theme
```

##### Default functionality should run `--output text` by default
You should get the same output for:
`pnpm shopify:run theme check --dev-preview --path ./tophat-theme`
and
`pnpm shopify:run theme check --dev-preview --path ./tophat-theme --output text`

##### Output JSON should hold structural parity with ruby theme check output
`pnpm shopify:run theme check --dev-preview --path ./tophat-theme --output json`
should give a JSON object with the same structure and keys as:
`pnpm shopify:run theme check --path ./tophat-theme --output json`

#### `--init`
`pnpm shopify:run theme check --dev-preview --path ./tophat-theme --init`
- Command should not write any new `.theme-check.yml` config to a directory that already has one. 
- Delete the existing `.theme-check.yml` file from `./tophat-theme` and run the command. It should generate a new config.

#### `--config`
- Copy the initialized yml from using the `--init` flag into the CLI repo root as `override.yml`: `cp ./tophat-theme/.theme-check.yml ./override.yml`
- Tweak this yml so that only one or none of the checks are `enabled`
- Run: `pnpm shopify:run theme check --dev-preview --path ./tophat-theme --config override.yml`
- The output should show only offenses for the enabled check.

#### `--fail-level`
`pnpm shopify:run theme check --dev-preview --path ./tophat-theme --fail-level suggestion`
should cause the CLI to exit with an error code when there are any suggestion offenses.

#### `--version`
`pnpm shopify:run theme check --dev-preview --version`
should return the version of `@shopify/theme-check-node`

This should also **not** run the full theme check scan on the theme.

#### `--auto-correct`
`pnpm shopify:run theme check --dev-preview --path ./tophat-theme --auto-correct` should run theme check on the theme and autocorrect any offenses that it can. @charlespwd I could use a bit of extra scrutiny on this bit.

#### `--print`
`pnpm shopify:run theme check --dev-preview --print` should print out the active theme check config to console.

This should also **not** run the full theme check scan on the theme.

#### `--list`
`pnpm shopify:run theme check --dev-preview --list` should show all the enabled checks that would run.

This should also **not** run the full theme check scan on the theme.

### Post-release steps
Actively monitor for bugs and to improve the experience while it sits behind the `--dev-preview` flag.

### Measuring impact

- [x] n/a - this doesn't need measurement: That comes after the `--dev-preview` period.

### Checklist

### DIFF
--- a/packages/theme/package.json
+++ b/packages/theme/package.json
@@ -36,8 +36,8 @@
   "dependencies": {
     "@oclif/core": "2.11.7",
     "@shopify/cli-kit": "3.49.1",
-    "@shopify/theme-check-node": "1.15.0",
-    "@shopify/theme-language-server-node": "1.3.0",
+    "@shopify/theme-check-node": "1.16.1",
+    "@shopify/theme-language-server-node": "1.3.2",
     "yaml": "2.3.2"
   },
   "devDependencies": {
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -699,11 +699,11 @@ importers:
         specifier: 3.49.1
         version: link:../cli-kit
       '@shopify/theme-check-node':
-        specifier: 1.15.0
-        version: 1.15.0
+        specifier: 1.16.1
+        version: 1.16.1
       '@shopify/theme-language-server-node':
-        specifier: 1.3.0
-        version: 1.3.0
+        specifier: 1.3.2
+        version: 1.3.2
       yaml:
         specifier: 2.3.2
         version: 2.3.2
@@ -4440,8 +4440,8 @@ packages:
       - encoding
     dev: false
 
-  /@shopify/theme-check-common@1.15.0:
-    resolution: {integrity: sha512-Sk1bnRbmh8M5urFhMoOZT+76ohU7Pbu2oBUf7/2mq2yad8HSDpv2Ux6K3cyybUyI7lp6GGqj8Ib2zdBYKO7XQQ==}
+  /@shopify/theme-check-common@1.16.1:
+    resolution: {integrity: sha512-n7cZ6pmjHRrNWKYW8Rruu/hnPlX6H1qEQiYY94PKONf4VMjs83q/LJlzauOJ8mND9PxeTCwWTAApZgNqHngR7g==}
     dependencies:
       '@shopify/liquid-html-parser': 1.0.0
       cross-fetch: 4.0.0
@@ -4454,25 +4454,11 @@ packages:
       - encoding
     dev: false
 
-  /@shopify/theme-check-common@1.16.0:
-    resolution: {integrity: sha512-UhaigeKOPUv3PJzTfxHvHK53fT/XZ0VTKbnR5EpvPvkXlwIgwLMjge2xpGEhO54SkIHNdwrRKtX9wGtbDCoKMQ==}
-    dependencies:
-      '@shopify/liquid-html-parser': 1.0.0
-      cross-fetch: 4.0.0
-      json-to-ast: 2.1.0
-      line-column: 1.0.2
-      lodash.set: 4.3.2
-      lodash.unset: 4.5.2
-      minimatch: 9.0.3
-    transitivePeerDependencies:
-      - encoding
-    dev: false
-
-  /@shopify/theme-check-docs-updater@1.15.0:
-    resolution: {integrity: sha512-BlzzTz3H+D6AxZn5DyIP/64tqAL9rZtXFgkwxM7GCfRlCh8l4XPndYhTKJ+YMuYkVKXFqzdelOQdpFPlJNiPQQ==}
+  /@shopify/theme-check-docs-updater@1.16.1:
+    resolution: {integrity: sha512-sNft1xUx6iRiItsh3L9M0iaj5vzi2Sf2r8v4uoyS4jeZph0gRoIQVv18vXNs16WIwVP4ucYKf/8rt3uhI3sAUw==}
     hasBin: true
     dependencies:
-      '@shopify/theme-check-common': 1.15.0
+      '@shopify/theme-check-common': 1.16.1
       ajv: 8.12.0
       env-paths: 2.2.1
       node-fetch: 2.7.0
@@ -4480,34 +4466,22 @@ packages:
       - encoding
     dev: false
 
-  /@shopify/theme-check-docs-updater@1.16.0:
-    resolution: {integrity: sha512-YbteLuRXqLF0hbetuHs7+WnXFdI00Y2kbCsB8X1F9CG7vmeRvtL1lLnB37jA9t8jgVTIu3QbmJ2OjFHAgAi11g==}
-    hasBin: true
+  /@shopify/theme-check-node@1.16.1:
+    resolution: {integrity: sha512-fJU//djQ41ltSvSjhx5LKQsxRfwIdr4m9XooLCiFKoGlZ9DUNww3SHE6rRdqdGu4v8tlRju1ccRdBQhik+eaBA==}
     dependencies:
-      '@shopify/theme-check-common': 1.16.0
-      ajv: 8.12.0
-      env-paths: 2.2.1
-      node-fetch: 2.7.0
-    transitivePeerDependencies:
-      - encoding
-    dev: false
-
-  /@shopify/theme-check-node@1.15.0:
-    resolution: {integrity: sha512-o3jNvo+YPUvC2JbMwwgn3cxZHzr4m2UKoGECvB3u16/U8nHzOIXPye2j7EDy7aLMJyhGajWuAggnnkwJ7Wdt0A==}
-    dependencies:
-      '@shopify/theme-check-common': 1.15.0
-      '@shopify/theme-check-docs-updater': 1.15.0
+      '@shopify/theme-check-common': 1.16.1
+      '@shopify/theme-check-docs-updater': 1.16.1
       glob: 8.1.0
       yaml: 2.3.2
     transitivePeerDependencies:
       - encoding
     dev: false
 
-  /@shopify/theme-language-server-common@1.3.0:
-    resolution: {integrity: sha512-VHiJ96EpyJ1wl5ghJjZca40ao/j6fEL0LIr4iAx0tgLYloGuaKi9yE2lSCKZoxhtBIVfndFN6bcFXQORNa7U/w==}
+  /@shopify/theme-language-server-common@1.3.2:
+    resolution: {integrity: sha512-56Fg4tLrAUWHXHLYvejp94IwDPUmoT2zvjHHH1kwxnFzdXfq2WPMtIBnFdtDvzs/giNrxwkzu1S80ed0Ry2xdg==}
     dependencies:
       '@shopify/liquid-html-parser': 1.0.0
-      '@shopify/theme-check-common': 1.15.0
+      '@shopify/theme-check-common': 1.16.1
       '@vscode/web-custom-data': 0.4.8
       vscode-languageserver: 8.1.0
       vscode-languageserver-textdocument: 1.0.11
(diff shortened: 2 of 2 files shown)


packages/theme/src/cli/services/check.test.ts
```diff
@@ -0,0 +1,307 @@
...
+  })
+})
+describe('formatOffensesJson', () => {
+  it('should format offenses into JSON correctly', () => {
+    const offensesByFile = {
+      '/path/to/file': [
+        {
+          type: SourceCodeType.LiquidHtml,
+          check: 'LiquidHTMLSyntaxError',
+          message: 'Attempting to close HtmlElement',
```
charlespwd, line 258: IIRC they are using vitest so you can prob use assymetric matchers here to future proof those tests and not be annoyed because of a rewording of the message. 

expect.stringMatching or something would do.

albchu: That should not affect this usecase. There are no side effect code for `formatOffensesJson`. This test would control all input and output from this function so unit test regressions here shouldnt occur from a package bump.

packages/cli-kit/src/private/node/ui/components/TokenizedText.tsx
```diff
@@ -56,6 +61,7 @@ export type Token =
   | FilePathToken
   | ListToken
   | BoldToken
+  | ColoredToken
```
isaacroldan, line 64: pinging @matteodepalo because you did all the UI stuff, just so that you are aware of this change

isaacroldan: ah! you were faster :P

packages/theme/src/cli/commands/theme/check.ts
```diff
@@ -97,6 +123,59 @@ Excludes checks matching any category when specified more than once`,
...
+        // --print should not trigger full theme check operation
+        return
+      }
+
+      if (flags.list) {
+        await outputActiveChecks(flags.config, flags.path)
+
+        // --list should not trigger full theme check operation
+        return
+      }
```
isaacroldan, line 154: Shouldn't these be subcommands of theme check instead of flags?
Like:
```bash
shopify theme check list
shopify theme check version
...
```

packages/theme/src/cli/services/check.ts
```diff
@@ -0,0 +1,359 @@
...
+  styleCount: number
+}
+export type FailLevel = 'error' | 'suggestion' | 'style'
+
+function severityToFailLevel(severity: Severity): FailLevel {
+  switch (severity) {
+    case Severity.ERROR:
+      return 'error'
+    case Severity.WARNING:
+      return 'suggestion'
```
charlespwd, line 46: More of a pet peeve than anything. If we're to render it, we should render it with the "modern" names. We support those for backwards compat, but we should print them as the new ones (`info`, `warning`, and`error`) See https://github.com/Shopify/theme-tools/blob/main/packages/theme-check-node/src/config/types.ts#L47-L56

packages/theme/src/cli/services/check.ts
```diff
@@ -0,0 +1,359 @@
...
+      meta && meta.meta
+        ? {
+            description: meta.meta.docs.description,
+            doc: meta.meta.docs.url,
+          }
+        : {}
+
+    acc[checkCode] = {
+      severity: severityToFailLevel(severity!),
+      ...metafields,
```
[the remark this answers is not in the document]
charlespwd: Ah lol that's the output from the ruby CLI not yours 🙈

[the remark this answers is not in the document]
charlespwd: Oh wow nvm it looks like we're doing that in the ruby CLI :D

packages/theme/src/cli/commands/theme/check.ts
```diff
@@ -97,6 +123,59 @@ Excludes checks matching any category when specified more than once`,
...
+        renderOffensesText(offensesByFile, flags.path)
+
+        renderInfo({
+          headline: 'Theme Check Summary.',
+          body: formatSummary(offenses, offensesByFile, theme),
+        })
+      }
+
+      if (flags.output === 'json') {
+        outputInfo(JSON.stringify(formatOffensesJson(offensesByFile), null, 2))
```
charlespwd, line 169: Looks like the CLI is printing stuff that messes with the output:
```
$ node packages/cli/bin/dev.js "theme" "check" "--dev-preview" "--path" "/Users/charles/ws/theme" --output json | jq .
jq: parse error: Invalid numeric literal at line 2, column 0
```

packages/theme/src/cli/services/check.ts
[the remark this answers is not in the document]
charlespwd: Also nvm I just got confused between --list and --print. Resolved!

packages/theme/src/cli/services/check.ts
```diff
@@ -0,0 +1,359 @@
...
+
+export async function runThemeCheck(themeRoot: string, configPath?: string) {
+  let themeCheckResults = {} as ThemeCheckRun
+
+  const themeCheckTask: Task = {
+    title: `Performing theme check. Please wait...\nEvaluating ${themeRoot}`,
+    task: async () => {
+      themeCheckResults = await themeCheckRun(themeRoot, configPath)
+    },
+  }
```
charlespwd, line 292: That's the thing that might need an option if `--output json` is selected. This pollutes STDOUT. Logging should happen on STDERR!

packages/theme/src/cli/commands/theme/check.ts
charlespwd: Also re: version, we should adopt the convention.
```
$ git --version
git version 2.42.0
hub version refs/heads/master
$ npm --version
10.1.0
$ yarn --version
1.22.19
$ gh --version
gh version 2.35.0 (2023-09-19)
https://github.com/cli/cli/releases/tag/v2.35.0
$ ruby --version
ruby 2.6.10p210 (2022-04-12 revision 67958) [universal.arm64e-darwin22]
```

packages/theme/src/cli/services/check.ts
albchu: Very good catch here. I think we probably need to drop `renderTask` in this case.

packages/cli-kit/src/private/node/ui/components/TokenizedText.tsx
```diff
@@ -163,6 +171,8 @@ const TokenizedText: FunctionComponent<TokenizedTextProps> = ({item}) => {
     return <List {...item.list} />
   } else if ('bold' in item) {
     return <Text bold>{item.bold}</Text>
+  } else if ('color' in item) {
```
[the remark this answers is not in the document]
albchu: Ive revised this change to now be three new tokens: `InfoToken`, `WarnToken` and `ErrorToken` https://github.com/Shopify/cli/pull/2889/commits/4137527413faafa2ee535f9e4af1d3c3d7236aba

packages/theme/src/cli/services/check.ts
```diff
@@ -0,0 +1,359 @@
...
+    const offensePadding = `${index === offenses.length - 1 ? '' : '\n\n'}`
+
+    return [
+      {color: severityToColor(severity), value: failLevelLabel},
+      {bold: `${check}`},
+      {subdued: `\n${message}`},
+      `\n\n${codeSnippet}`,
+      offensePadding,
+    ]
+  })
```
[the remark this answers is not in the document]
albchu: Looks like 80 is the default min width when the ui lib cannot calculate window size. 😔https://github.com/Shopify/cli/blob/main/packages/cli-kit/src/private/node/ui/hooks/use-layout.ts#L36-L48

packages/theme/src/cli/commands/theme/check.ts
```diff
@@ -44,6 +64,12 @@ Excludes checks matching any category when specified more than once`,
       env: 'SHOPIFY_FLAG_FAIL_LEVEL',
       options: ['error', 'suggestion', 'style'],
     }),
+
+    /**
+     * Typescript theme check no longer uses `--update-docs`
+     * theme check initialization verifies it has the latest revision of theme docs
+     * every time it runs, and downloads the latest revision if it doesn't.
+     */
```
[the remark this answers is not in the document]
albchu: I think it was hiding between the file diff folds: https://github.com/Shopify/cli/blob/fd821dbb1a4a7c2789b75aa17705838b89112aeb/packages/theme/src/cli/commands/theme/check.ts#L61-L66

packages/theme/src/cli/services/check.ts
```diff
@@ -0,0 +1,359 @@
...
+    // loadConfig has no clear idea of where to extend the config from.
+    // We can default to recommended to be safe.
+    extends: 'theme-check:recommended',
+
+    // Depending on how the configs were merged during loadConfig, there may be
+    // duplicate patterns to ignore. We can clean them before outputting.
+    ignore: [...new Set(ignore)],
+
+    // Dump out the active settings for all checks.
+    ...settings,
```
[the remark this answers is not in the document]
albchu: I wasnt sure if adding root here would have broke with parity before. Ive added it now though! https://github.com/Shopify/cli/pull/2889/commits/38469ffa48db325018bf9ff3f7659b4e11bb5b79

packages/theme/src/cli/services/check.ts
```diff
@@ -0,0 +1,359 @@
...
+    }
+  })
+}
+
+/**
+ * Handles the process exit based on the offenses and fail level.
+ */
+export function handleExit(offenses: Offense[], failLevel?: FailLevel) {
+  // If there is no fail level set, exit with 0
+  if (!failLevel) process.exit(0)
```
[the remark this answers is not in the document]
charlespwd: The default is set here: https://github.com/Shopify/theme-check/blob/e7bbb91eb35feafc392712cdc7e00d42ef8d558c/lib/theme_check/cli.rb#L20

The `crash` is a hardcode we use only for this to mean that even :error shouldn't exit 1. 

I think the documentation was incorrect, it should have been (crash|error|suggestion|style).

packages/theme/src/cli/services/check.ts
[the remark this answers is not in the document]
charlespwd: Ugh. Aight!

charlespwd (approved):
Only the fail-level crash and error as default are missing. Everything else is AOK from my perspective! 🙌 

Tophatted and works great :D

packages/theme/src/cli/commands/theme/check.ts
```diff
@@ -97,8 +125,81 @@ Excludes checks matching any category when specified more than once`,
...
+      }
+
+      const {offenses, theme} = await themeCheckRun(path, flags.config)
+
+      const offensesByFile = sortOffenses(offenses)
+
+      if (flags.output === 'text') {
+        renderOffensesText(offensesByFile, path)
+
+        renderInfo({
```
karreiro, line 169: Personally, I believe it would be nice if we could invoke `renderSuccess` when there are no offenses.

image

From a user's perspective, I usually appreciate the quick, rewarding green feedback when everything is fixed :)

packages/theme/package.json
```diff
@@ -36,7 +36,9 @@
   "dependencies": {
     "@oclif/core": "2.11.7",
     "@shopify/cli-kit": "3.49.1",
-    "@shopify/theme-language-server-node": "1.1.0"
+    "@shopify/theme-check-node": "1.15.0",
```
karreiro, line 39: IIUC, we're planning to update the version here to include https://github.com/Shopify/theme-tools/pull/158. Is that correct? :)

karreiro (approved):
Thank you, @albchu! Great stuff 🚀 I've left only two non-blocking comments :)

PR #2895, opened 2023-09-27
Fix extension devUUID to match remote extension UUID if exists

alfonso-noriega:
Fixes [#461](https://github.com/Shopify/app-ui/issues/461)

UI extensions, in order to have live reload, communicate with Web by web sockets. The payload of the WS does not have the uuid updated to the one in the draft. As the UUID is used for reconciliation in EH dev plugin, this results in having two extensions rendered, the one stored in the DB as a draft and the one specified by the WS payload.

The CLI already resolves the matching extensions, but this matching process was not updating the extension instance. To solve this:
- Added a new method to the App which updates all extension UUIDS to the ones provided in a map
- Force the update of the UUIDS after reconciliation of local and remote extensions0
- 
### How to test your changes?
- create an app with a block extension.
- dev the app and see the extension duplicated at the bottom of the order page.
- checkout this branch.
- dev the app again and see just one block extension rendered.

### Measuring impact

### Checklist

### DIFF
--- a/packages/app/src/cli/services/dev/processes/draftable-extension.ts
+++ b/packages/app/src/cli/services/dev/processes/draftable-extension.ts
@@ -117,7 +117,7 @@ export async function setupDraftableExtensionsProcess({
   const deploymentMode = unifiedDeployment ? 'unified' : 'legacy'
   const prodEnvIdentifiers = getAppIdentifiers({app: localApp})
 
-  const {extensionIds: remoteExtensionIds} = await ensureDeploymentIdsPresence({
+  const {extensionIds: remoteExtensionIds, extensions: extensionsUuids} = await ensureDeploymentIdsPresence({
     app: localApp,
     partnersApp: remoteApp,
     appId: apiKey,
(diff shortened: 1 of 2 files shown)


packages/app/src/cli/services/dev/processes/draftable-extension.ts
```diff
@@ -128,6 +128,8 @@ export async function setupDraftableExtensionsProcess({
     envIdentifiers: prodEnvIdentifiers,
   })
 
+  localApp.updateExtensionUUIDS(extensionsUuids)
```
alvaro-shopify, line 131: maybe a comment here to explain why the temp devUuids, generated each time the `dev` command is run, should be overwritten with the remote ones (in case they exist)

alfonso-noriega:
@Arkham this fix is addressing all the scenarios related to the dev flow, replacing the random dev UUID by the remote  UUID right after the matching between local and remote extension happens. There where some changes introduced [here](https://github.com/Shopify/cli/blob/fix_extension_devUUID_upddated_to_remote_UUID/packages/app/src/cli/services/dev.ts#L117) which were relying on the .env file values generated after the first `deploy` leaving un-deployed apps with the UUID not updated causing duplication in preview. Do you think there would be any side effects if the linked changes are removed? could you have a short look at this and give some feedback?

PR #2910, opened 2023-09-29
adding localization tracking in dev mode

MadhavMakkena:

Fixes #0000 

### Measuring impact

### Checklist


.changeset/moody-taxis-study.md
```diff
@@ -0,0 +1,5 @@
+---
+'@shopify/app': minor
```
andrewhassan, line 2: Since this is a bug fix to the localization feature, which is already released, I think we might want to include this in a patch version.

packages/app/src/cli/services/dev/extension/bundler.ts
```diff
@@ -244,20 +246,49 @@ export async function setupFunctionWatcher({
...
+  if (await fileExists(joinPath(extension.directory, 'locales'))) {
+    redeployWatchPaths.push(joinPath(extension.directory, 'locales', '**.json'))
+  }
+
+  outputDebug(
+    `Watching paths for function extension ${extension.localIdentifier}: ${rebuildAndRedeployWatchPaths.concat(
+      redeployWatchPaths,
+    )}`,
+    stdout,
+  )
```
andrewhassan, line 269: Nit: For the purposes of debugging, I wonder if it would be handy to know which paths trigger a rebuild and which trigger a redeploy.

packages/app/src/cli/services/dev/extension/bundler.ts
```diff
@@ -281,20 +312,30 @@ export async function setupFunctionWatcher({
...
-      .catch((error: any) => {
-        outputDebug(
-          `Function file watching failed to close for extension with ${extension.devUUID}: ${error.message}`,
-          stderr,
+  if (redeployWatchPaths.length > 0) {
+    const functionRedeployWatcher = chokidar.watch(redeployWatchPaths).on('change', (path) => {
+      outputDebug(`File at path ${path} changed`, stdout)
+      if (buildController) {
+        // terminate any existing builds
+        buildController.abort()
```
andrewhassan, line 322: If these two share the same controller, is there a race condition here?

For example, if there's a file that trigger both watch conditions. I think the following events could happen:
- The first watch callback runs, which creates the controller
- The second watch callback runs, which aborts the controller
- `buildFunctionExtension` resolves, but `buildSignal` has been aborted so the function is not updated

To avoid this, I think we could use a different signal for the redeploy watcher.

packages/app/src/cli/services/dev/extension/bundler.ts
MadhavMakkena: Updated the debug output message.

packages/app/src/cli/services/dev/extension/bundler.ts
MadhavMakkena: Thanks for pointing this out. I just fixed it. 

```
if (buildController) {
  // terminate any existing builds
  buildController.abort()
}
``` 
isn't required in the `functionRedeployWatcher` as it shouldn't be worried about rebuilding the function.

packages/app/src/cli/services/dev/extension/bundler.test.ts
```diff
@@ -554,6 +556,44 @@ describe('setupFunctionWatcher', () => {
     expect(chokidarOnSpy).toHaveBeenCalledWith('change', expect.any(Function))
   })
 
+  test('deploys the function on file change in locales directory', async () => {
+    // GIVEN
+    // fileExists returns true
```
andrewhassan, line 561: Nit: In other tests it looks like this is formatted differently. I also don't think we need the extra comment, it's easy to tell by looking at the code 😄 

```suggestion
    // Given
```

packages/app/src/cli/services/dev/extension/bundler.test.ts
```diff
@@ -554,6 +556,44 @@ describe('setupFunctionWatcher', () => {
...
+      watchPath: '*.rs',
+    })
+
+    const chokidarOnSpy = vi.fn().mockImplementation((_event, handler) => {
+      // Call the file watch handler immediately
+      handler(`${watchOptions.extension.directory}/locales/en.json`)
+    })
+
+    // WHEN
+    // Watcher is setup on locales directory
```
andrewhassan, line 574: Nit:

```suggestion
    // When
```

packages/app/src/cli/services/dev/extension/bundler.test.ts
```diff
@@ -554,6 +556,44 @@ describe('setupFunctionWatcher', () => {
...
+        on: vi.fn(),
+      } as any
+    })
+
+    const updateExtensionConfigSpy = vi.spyOn(updateExtension, 'updateExtensionConfig').mockResolvedValue()
+
+    await setupFunctionWatcher(watchOptions)
+    await flushPromises()
+
+    // THEN
```
andrewhassan, line 591: Nit:

```suggestion
    // Then
```

andrewhassan (approved):
Nice work!! 🎩 LGTM!

PR #2923, opened 2023-10-03
Fix checkout-cart-url flag

alvaro-shopify:
Running the `dev` command with the flag `checkout-cart-url` is not working with checkout ui extensions generated from the new template. The product shown when you visit the checkout extension preview url is different to the one you set in the flag 

- Remove the old method that built the url. It was receiving the cart url flag value and in case there are no `checkout_ui_extension`  type it returned `undefined` as the cart url. New checkout ui extensions have the generic `ui-extension` type, so we should filter them using the method `shouldFetchCartUrl` which applies to the old and new checkout ui extensions
- Remove duplicate call to the cart url build method inside the execution of the dev ui process.

### How to test your changes?
- Run the `dev` command with one product variant
```sh
p shopify app dev --path /path/to/your/app  --checkout-cart-url /cart/43823180349462:1
```
- Visit the checkout ui extension preview url and the product variant displayed in the cart should match the one you set in the flag

### Measuring impact

### Checklist

### DIFF
--- a/packages/app/src/cli/services/dev/extension.test.ts
+++ b/packages/app/src/cli/services/dev/extension.test.ts
@@ -1,4 +1,3 @@
-import * as utilities from './extension/utilities.js'
 import * as store from './extension/payload/store.js'
 import * as server from './extension/server.js'
 import * as websocket from './extension/websocket.js'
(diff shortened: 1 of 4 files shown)


isaacroldan (approved):
Good catch 👌

PR #2924, opened 2023-10-03
Support template branches when creating apps

paulomarg:

We'll have multiple options of the Remix app template going forward, and we want users to be able to select between a TS or JS version of the template.

Setting up the form so it asks a second question for the remix template, so the user can select a branch.

I implemented this as a generic way of adding branch options so that other templates can benefit from this in the future.

### How to test your changes?

- Run `dev create-app`, and select the remix template
- OR run `dev create-app --template remix`

image

### Post-release steps

Merge [this template PR](https://github.com/Shopify/shopify-app-template-remix/pull/372) which makes `main` the TS version of the code.

### Measuring impact

- [x] Existing analytics will cater for this addition, since I'm just altering the template string we use in the end.

### Checklist

### DIFF
--- a/packages/create-app/src/commands/init.ts
+++ b/packages/create-app/src/commands/init.ts
@@ -1,4 +1,4 @@
-import initPrompt, {allTemplates, visibleTemplates} from '../prompts/init.js'
+import initPrompt, {isPredefinedTemplate, visibleTemplates} from '../prompts/init.js'
 import initService from '../services/init.js'
 import {Flags} from '@oclif/core'
 import {globalFlags} from '@shopify/cli-kit/node/cli'
@@ -83,7 +83,7 @@ export default class Init extends Command {
         'Only GitHub repository references are supported, ' +
           'e.g., https://github.com/Shopify/<repository>/[subpath]#[branch]',
       )
-    if (!url && !allTemplates.includes(template))
+    if (!url && !isPredefinedTemplate(template))
       throw new AbortError(
         outputContent`Only ${visibleTemplates
           .map((alias) => outputContent`${outputToken.yellow(alias)}`.value)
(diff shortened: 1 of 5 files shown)


packages/create-app/src/commands/init.ts
```diff
@@ -83,7 +83,7 @@ export default class Init extends Command {
         'Only GitHub repository references are supported, ' +
           'e.g., https://github.com/Shopify/<repository>/[subpath]#[branch]',
       )
-    if (!url && !allTemplates.includes(template))
+    if (!url && !isPredefinedTemplate(template))
       throw new AbortError(
         outputContent`Only ${visibleTemplates
           .map((alias) => outputContent`${outputToken.yellow(alias)}`.value)
```
isaacroldan, line 89: Should we also add a new flag to select the `branch`? but called flavor like for extensions
`create-app --template remix --flavor typescript`

paulomarg: The way I set this up you'd pass in the template and get prompted for the branch, for `remix`. We could add a flag, but I (slightly) prefer the current experience.

Happy to go with whatever the team prefers.

isaacroldan: There is a "rule" that all prompts should have an alternative without user interaction (for nonTTY terminals for instance), but that's already covered by the `--template` flag if you use a full repo name like `https://github.com/Shopify/<repository>/[subpath]#[branch]` 🤔 
So not sure if it's worth it to add a new flag or not, probably a @MeredithCastile decission

paulomarg:
Re-requested review because I changed a few things!

MeredithCastile:
Copying content recommendations over from our Slack thread:

Prompt:
> For your Remix template, which language do you want to use?
> > TypeScript
> > JavaScript

Error (using the Ink error pattern in the CLI UI Kit):
> Invalid option for --flavor.
> The --flavor flag for Remix accepts only `typescript` or `javascript` options.

PR #2926, opened 2023-10-04
sends ui.handle as empty string instead of undefined when set

mssalemi:

Thanks for spotting this Detla!

Right now, we are incorrectly adding `ui.handle` to the configuration of a function when `ui.handle` is set to an empty string. 

The way we add either the ui.paths or ui.handle to the config when deploying , is from setting the ui property by:

(previously)
```
    let ui: UI | undefined

    if (config.ui?.paths) {
      ui = {
        app_bridge: {
          details_path: config.ui.paths.details,
          create_path: config.ui.paths.create,
        },
      }
    }

**This check returns false when ui.handle is emptry string**
    if (config.ui?.handle) {
      ui = {
        ...ui,
        ui_extension_handle: config.ui.handle,
      }
    }
```
**thus sending ui propery as undefined when deploying**

Updates the check above to correctly send `ui.handle` properties when they are set to empty string. 

### How to test your changes?

Added some tests for when ui.handle is `""` or `undefined`

#### 🎩 Test deploying a function with `ui.handle` as `""`. 

Screenshot 2023-10-04 at 9 41 39 AM

```
SPIN=1 SPIN_INSTANCE=script-service-q7k4 dev spin pnpm shopify app deploy  --path=../../../../Desktop/prod-testing/product-discounts-single-ext
```

This PR will not validate that value for `ui.handle`, only correctly set the configuration that will deployed. 

That validation is part of https://github.com/Shopify/shopify/pull/454992, and will check for an empty string in the `UiHandleValidator`.

### Measuring impact

### Checklist

### DIFF
--- a/packages/app/src/cli/models/extensions/specifications/function.test.ts
+++ b/packages/app/src/cli/models/extensions/specifications/function.test.ts
@@ -260,4 +260,40 @@ describe('functionConfiguration', () => {
       expect(got!.localization).toEqual(expectedLocalization)
     })
   })
+
+  test("parses ui.handle when it's an empty string", async () => {
+    await inTemporaryDirectory(async (tmpDir) => {
+      // Given
+      extension.directory = tmpDir
+      extension.configuration.ui!.handle = ''
+
+      // When
+      const got = (await extension.deployConfig({
+        apiKey,
+        token,
+        unifiedDeployment,
+      })) as unknown as {ui: {ui_extension_handle: string}}
+
+      // Then
+      expect(got.ui?.ui_extension_handle).toStrictEqual('')
+    })
+  })
+
+  test("parses ui.handle when it's undefined", async () => {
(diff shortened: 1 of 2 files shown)


dphm:
Is this to allow you to distinguish setting `ui.handle` to `""` from `ui.handle` not being present in the `toml`? i.e. to tell the dev `ui.handle can't be blank`?

DuncanUszkay1 (approved):

dphm (approved):

mssalemi:
> Is this to allow you to distinguish setting ui.handle to "" from ui.handle not being present in the toml? i.e. to tell the dev ui.handle can't be blank?

Yea. Right now there is no difference between ui.handle set to `""` and not set at all, so we would create the function in core with `ui_extension_handle = undefined` instead of `""`.

PR #2927, opened 2023-10-04
Allow enabling and disabling dev store preview mode with dashboard managed extensions

matteodepalo:

Fixes https://github.com/Shopify/develop-app-management/issues/1475

There are cases, like when developers only have dashboard managed extensions, when we weren't showing the dev preview toggle.

Allow enabling and disabling dev store preview mode with dashboard managed extensions when running `dev`.

### How to test your changes?

- Add a dashboard managed extension (e.g. admin link)
- Run `dev`
- Dev preview should be enabled and the toggling should be possible with `d`

### DIFF
--- a/.changeset/modern-snakes-occur.md
+++ b/.changeset/modern-snakes-occur.md
@@ -1,0 +1,5 @@
+---
+'@shopify/app': patch
+---
+
+Allow enabling and disabling dev store preview mode with dashboard managed extensions.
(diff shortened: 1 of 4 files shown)


matteodepalo:
Alternatively we could try to fetch remote extensions as well and add that to the condition. The problem with my proposed approach might be that there are more queries than needed when developers don't add extensions.

isaacroldan (approved):

alfonso-noriega (approved):
🎩 + LGTM!

alfonso-noriega:
@matteodepalo, great work! this definitely fixes the DX with partner managed extensions. I am curious though about the option of showing always (even if there are no remote extensions) the possibility to toggle the preview, was this discarded because of the continuous polling for the preview flag state? The benefit of this would be that the developer does not need to re-run `dev` after adding a partner managed extension, but still needs to toggle de flag on anyway 🤷🏻‍♂️

PR #2979, opened 2023-10-16
CI Improvements

isaacroldan:

Our CI runs take a lot of time, this is an effort to improve that.

**This PR reduces CI times from 12-16min to 6-9min.**

The bottleneck is in acceptance tests (in mac os), they are slow and do real network calls, if we want to further improve CI times we'd need to discuss what we want to do with them if the plan is to add more (maybe not run them in every PR?)

Revamp the github action workflows to parallelize as much as possible and remove unnecessary steps:
  - Don't make a workflow depend on any other -> run all at the same time.
  - Only install ruby if specifically needed for the current workflow -> only for acceptance tests.
  - Run acceptance tests in a new workflow -> this is what takes most of the time, no need to wait for unit tests.
  - Only fetch the last commit when cloning the repo -> we were fetch EVERYTHING.
  - Stop using nx affected -> it didn't improve the CI times and forces us to fetch all commits of the repo.
- Update some external github actions.
- Remove node 16 from the CI matrix.
- Disable some GitDiff tests that were failing on windows.
- Use the built-in cache of action/setup-node instead of our own solution

### Measuring impact

### Checklist


alvaro-shopify:
Cool!

If the problem is the acceptance test, I would move them to be executed during the `nightly` generation. In case of receiving an alert we only need to put the focus on the PRs merged the last day that should not be a big amount

pnpm-lock.yaml
```diff
@@ -9025,7 +9025,7 @@ packages:
     resolution: {integrity: sha512-rXckvqPBB0Krifk5rn/5LvQGmyXwCUpBfmTwbkQNBY9JY8RSl3b8OftBNEYxg4+SWUhEKcPifgope28uL9inlA==}
     engines: {node: '>=14.16'}
     peerDependencies:
-      '@types/react': 17.0.2
+      '@types/react': '>=18.0.0'
```
gonzaloriestra, line 9028: Are these changes related? It's not the first time I see them, it could be changing depending on our local config/versions.

gonzaloriestra (approved):
Awesome work, thanks! 👏

.github/workflows/shopify-cli.yml
```diff
@@ -118,16 +118,16 @@ jobs:
...
       - name: Setup deps
         uses: ./.github/actions/setup-cli-deps
         with:
           node-version: ${{ matrix.node }}
       - name: ${{ matrix.target }}
         run: pnpm nx run-many --all --skip-nx-cache --target=${{ matrix.target }}
 
-  oclif-readme:
-    name: '[PR] Check OCLIF readme'
+  oclif-manifests:
```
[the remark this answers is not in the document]
isaacroldan: Is not actually just refreshing docs, is checking that the manifest and readme are up to date. Does it make more sense to rename this to `oclif-checks`?

pnpm-lock.yaml
gonzaloriestra: I don't get any change now if I run `pnpm i` from main 🤷

matteodepalo (approved):
LGTM with the small comment added as caveat.

PR #3022, opened 2023-10-26
[Fix] node/git isClean to match returned type

juanpprieto:

While using git `isClean` I noticed the return type Promise<boolean> did not match the actual return. simple-git's `git.status().isClean` is a [function](https://github.com/steveukx/git-js/blob/8a31c45a1359248cb38d768d22a5e29903ceef8a/simple-git/typings/response.d.ts#L371C4-L371C11) and not a property.

### Checklist


gonzaloriestra (commented):
Good catch, thanks! Fortunately, it wasn't being used in the repository... Could you fix the tests?

juanpprieto:
👋🏼 @gonzaloriestra I think these test updates are ok?

PR #3025, opened 2023-10-27
Update base-command.ts to await error handling

jamieguerrero:

Fixes https://github.com/Shopify/develop-app-management/issues/1522 

Confirms https://github.com/Shopify/develop-app-management/issues/1528
Confirms https://github.com/Shopify/develop-app-management/issues/1527

To test:
- Make an app 
- Add a checkout extension 
- Replace the code there with garbage 
- Run `pnpm shopify app build --verbose` 
- Confirm that CLI shows the syntax error, exits, skips sending the error to monorail (because we're in development).
Screenshot 2023-10-27 at 11 36 57 PM

### Measuring impact

### Checklist

### DIFF
--- a/packages/cli-kit/src/public/node/base-command.ts
+++ b/packages/cli-kit/src/public/node/base-command.ts
@@ -28,7 +28,7 @@ abstract class BaseCommand extends Command {
   }
 
   async catch(error: Error & {exitCode?: number | undefined}): Promise<void> {
-    errorHandler(error, this.config)
+    await errorHandler(error, this.config)
     return Errors.handle(error)
   }
 
--- a/packages/cli-kit/src/public/node/cli.ts
+++ b/packages/cli-kit/src/public/node/cli.ts
@@ -59,13 +59,15 @@ export async function runCLI(options: RunCLIOptions): Promise<void> {
     settings.debug = true
   }
 
-  run(undefined, options.moduleURL)
-    .then(() => flush())
-    .then(printEventsJson)
-    .catch((error) => {
-      errorHandler(error)
-      return Errors.handle(error)
-    })
+  try {
+    await run(undefined, options.moduleURL)
+    await flush()
+    printEventsJson()
+    // eslint-disable-next-line no-catch-all/no-catch-all
+  } catch (error) {
+    await errorHandler(error as Error)
+    return Errors.handle(error as Error)
+  }
 }
 
 /**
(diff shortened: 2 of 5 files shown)


packages/cli-kit/src/public/node/base-command.ts
```diff
@@ -28,7 +28,7 @@ abstract class BaseCommand extends Command {
   }
 
   async catch(error: Error & {exitCode?: number | undefined}): Promise<void> {
-    errorHandler(error, this.config)
+    await errorHandler(error, this.config)
```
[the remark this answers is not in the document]
amcaplan: I don't think the overload signatures are helpful (since we're not using return types) but I also don't think we'd ever get an error on this. Why should TS care if it's awaited? You're always free to just fire-and-forget an async function, and we do that elsewhere. As long as you don't try and use the return value assuming it's not a Promise, you'll be fine.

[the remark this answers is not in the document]
jamieguerrero: Great, I removed the unnecessary overload signature and tightened up the return type to be either a Promise or unknown. Let me know if you have any suggestions for simplifying the return type further or any other ideas!

amcaplan (approved):
LGTM!

packages/cli-kit/src/public/node/cli.ts
```diff
@@ -62,8 +62,8 @@ export async function runCLI(options: RunCLIOptions): Promise<void> {
   run(undefined, options.moduleURL)
     .then(() => flush())
     .then(printEventsJson)
-    .catch((error) => {
-      errorHandler(error)
+    .catch(async (error) => {
```
matteodepalo, line 65: Does this work? I'm not sure you can pass an async function to `catch`, it won't be awaited.

matteodepalo: If we want to await here we'll need to convert this to a `try/catch` block so something like:

```
try {
   await run(undefined, options.moduleURL)
   await flush()
   printEventsJson()
  } catch {
   await errorHandler(error)
   return Errors.handle(error)
}
```

jamieguerrero: I updated this bit, please let me know what you think! I don't love the `// eslint-disable-next-line no-catch-all/no-catch-all` eslint ignore I had to use, but a search on the CLI repo seems like we do this quite often because we don't want to rethrow errors and interrupt the process, we want to log them instead?

matteodepalo (approved):

PR #3049, opened 2023-11-01
Generate extension features that use the network incorrectly instrumented

shauns:

Related to https://github.com/Shopify/develop-app-management/issues/1531

Wrapping a few steps from `generate extension` that are actually network time rather than active time.

### How to test your changes?

Run `generate extension` and see that the active time is lower and the network time is higher with this branch.

### Measuring impact


amcaplan (approved):
Code looks good, 🎩 seems to have the desired effect. Thanks!

PR #3067, opened 2023-11-03
Skips webhooks updates on push 

gracejychang:

Fixes #0000 

### Measuring impact

### Checklist


packages/app/src/cli/api/graphql/push_config.ts
```diff
@@ -7,7 +7,7 @@ export const PushConfig = gql`
     $applicationUrl: Url
     $redirectUrlAllowlist: [Url]
     $requestedAccessScopes: [String!]
-    $webhookApiVersion: String!
+    $webhookApiVersion: String
```
gracejychang, line 10: I had to remove this requirement otherwise net new apps can't run `push` with an undefined webhookApiVersion. We can require it on the `deploy` command.

The question is, would it break anything for net new apps if this remained `undefined`? ie, can a net new app with declarative config work properly without an api version if a developer only runs `push` and doesn't run `deploy`? We automatically set it (manually and in CLI) whenever an app is created and linked to partners.

packages/app/src/cli/services/app/config/push.ts
```diff
@@ -104,19 +105,41 @@ export async function pushConfig(options: PushOptions) {
...
+) => {
+  const getWebhookApiVersion = () => {
+    if (declarativeWebhooksBeta && app.webhookApiVersion) {
+      // if app is not new and already has webhook api version configured
+      // don't wipe out their webhook api version, just don't update it
+      return app.webhookApiVersion
+    } else if (declarativeWebhooksBeta && !app.webhookApiVersion) {
+      // if the app is net new and using declarative webhooks, keep it undefined
+      // it'll be updated by the deploy command
+      return undefined
```
gracejychang: same question as above - what do we think about keeping this as undefined?

packages/app/src/cli/services/app/config/push.ts
```diff
@@ -104,19 +105,47 @@ export async function pushConfig(options: PushOptions) {
...
+
+  if (declarativeWebhooksBeta) {
+    // if app is not new and already has webhook api version configured
+    // don't wipe out their webhook api version, just don't update it
+    if (app.webhookApiVersion) {
+      webhookApiVersion = app.webhookApiVersion
+    } else {
+      // if the app is net new and using declarative webhooks, keep it undefined
+      // it'll be updated by the deploy command
+      webhookApiVersion = undefined
```
gracejychang, line 124: same comment as above - what do we think about keeping this undefined?

packages/app/src/cli/services/app/config/push.test.ts
```diff
@@ -412,4 +420,68 @@ app_preferences > url: this url is blocked 6`)
...
+    }
+    const options: PushOptions = {
+      configuration: app.configuration,
+      force: true,
+    }
+
+    vi.mocked(partnersRequest).mockResolvedValue({
+      app: {
+        apiKey: '12345',
+        webhookApiVersion: '2023-04',
```
gracejychang, line 454: this should remain the same version as the `EXISTING_APP_CONFIG` because it skipped the `2023-01` webhookApiVersion we were trying to push up above

packages/app/src/cli/api/graphql/push_config.ts
alexanderMontague: why did we even want to make this change here now? We are also planning on storing `webhookApiVersion` as a part of the webhooks module correct? If thats the case, then im on board with doing this but I feel like this should be one of the last things we do, as we want to keep the CLI push/deploy experience pretty much the exact same right up until we release declarative webhooks.

alexanderMontague: edit: nevermind, this is toggled by beta! I totally missed that

karenxie (commented):
I'll let folks with more context in this repo comment on the code, but the logic changes make sense to me! if we don't use the organization beta here, what will it be used for?

alexanderMontague:
@karenxie I'm not entirely sure what the org beta would be used for haha... I was thinking maybe partner UI changes... but even that should still be driven off of the app beta I'm pretty sure

PR #3074, opened 2023-11-06
Increase `days-before-issue-stale` value (`21` → `42`)

karreiro:
Increase `days-before-issue-stale` value (`21` → `42`).


gonzaloriestra:
What about 30 days to have the same value for issues/PRs?

@Shopify/cli-foundations thoughts?

gonzaloriestra (approved):

PR #3122, opened 2023-11-20
Add config extension to app

alvaro-shopify:
Add support for the new version app config specifications. Each section in the app config toml will be matched with one of the new `app_cofig` extension spec. A new extension instance will be created with the content of the section. All this new `app_config` extension instances will be uploaded with the `appDeploy` API request.
Creating a new `app_cofig` extension spec should be as simple as creating a new module inside the `@shopify/app` `extensions` folder. This should be an example of adding a new section to the app config schema
```ts
import {createConfigExtensionSpecification} from '../specification.js'
import {zod} from '@shopify/cli-kit/node/schema'

const PosEmbedmentSchema = zod.object({
  pos: zod
    .object({
      embedded: zod.boolean(),
    })
    .optional(),
})

  identifier: 'pos_embedment',
  schema: PosEmbedmentSchema,
})

export default spec

```

Co-authored-by: Alvaro Gutierrez [alvaro.gutierrezblanco@shopify.com](mailto:alvaro.gutierrezblanco@shopify.com)
Co-authored-by: Raluca Badoi [raluca.badoi@shopify.com](mailto:raluca.badoi@shopify.com)

- Creates the new `ConfigExtensionSpec`. It's a simplified version of the current `ExtensionSpec`. New management experience `app_config` remote specs will be mapped to this new `ConfigExtensionSpec` type.
- Modify `app loader` logic to support new app [config format ](https://docs.google.com/document/d/1EwN1mceHGdvepj8AxrgvchIlenzkiJs3tznfr_O1VSQ/edit). The new app schema is generated dynamically from the union of the subschema of each `app_config` extension spec
- Modify `app loader` logic to load each section of the app toml as a new `configExtensions` list inside the `app` object
- Deploy command will include the current `allExtensions` list and the new `configExtensions` list

### How to test your changes?
- Create a new spin instance `spin up partners -c shopify.branch=prototype-versioned-app-config -c partners.branch=add-app-config-management`
- Create a new `cli` app and run the `app config link` command to link it to a new remote app
- Edit the `shopify.app.toml` file and leave this content
```toml
name = "<YOUR_APP_NAME>"
version = "3"
client_id = "<YOUR_CLIENT_ID>"

[pos]
embedded = true
```
- Run the `app deploy` command to release a new version.
- Open the `partners` graphiql and run the following query and check that the `config` value is ok
```
 {
     app(apiKey: "1df452484825c0a1873948f2c4acdfb4") {
      extensionRegistrations {
        id
        uuid
        title
        type
        draftVersion {
          config
        }
        activeVersion {
          config
        }
      }
      configExtensionRegistrations {
        id
        uuid
        title
        type
        draftVersion {
          config
        }
        activeVersion {
          config
        }
      }
      dashboardManagedExtensionRegistrations {
        id
        uuid
        title
        type
        activeVersion {
          config
        }
        draftVersion {
          config
        }
      }
    }
  }
```
- Edit the `shopify.app.toml` file again and switch `embedded` to `false``
- Run the `deploy` command and the `graphql` query and the value should be updated

### Measuring impact

### Checklist


packages/app/src/cli/models/extensions/specifications/app_proxy.ts
```diff
@@ -0,0 +1,19 @@
+import {createConfigExtensionSpecification} from '../specification.js'
```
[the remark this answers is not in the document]
alvaro-shopify: > Building on top of this schema, is it possible for us to maintain the existing TOML structure?

This is technically possible. For example, we can define cli local `app_preferences` spec schema like this:
```
zod.object({
  app_preferences: zod
    .object({
      url: zod.string().max(255),
    })
    .optional(),
  application_url: zod.string(),
  embedded: zod.boolean(),
})
```
The config content of this extension once this schema is parsed against the toml file will be:
```json
{
  "app_preferences": {
    "url": "https://example.com"
  },
  "application_url": "https://example.com",
  "embedded": true
}
```

> Context: trying to minimize the blast radius of our first milestone and preserve the existing TOML structure if it doesn't create extra tech debt to do so.

First questions is:
- What's the structure of the content expected by the server side?, In case the serve expect a different format, we need to add some kind of support in the CLI to transform the parsed content to the format expected by the server before running the `AppDeploy` request.

The second question is what we need to do in future releases when we need to modify the schema of the app toml to match the proposal made in the [doc](https://docs.google.com/document/d/1EwN1mceHGdvepj8AxrgvchIlenzkiJs3tznfr_O1VSQ/edit):
- Define a second version of each cli local  spec schema to match the new section formats:
- Implement a migration process to transform the current app config toml to the new format
- (optional) support all formats in the same CLI version

Let's summarise the pros and cons of implementing this in two different CLI releases:
- Pros:
  - Migration implementation should not be carried out in the first CLI release
  - Users are not impacted for managing a different app toml file
- Cons:
  - Implement a transform login between the format parsed and the format expected by the server
  - In the second release we need to define a second version of each cli local  spec schema and support previous ones.
  - Managing different schemas versions will make the code brittle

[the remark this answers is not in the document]
pt2pham: @isaacroldan Curious what your thoughts are on this given the CLI as a Client work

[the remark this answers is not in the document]
alvaro-shopify: hey @pt2pham !, regarding the topic of CLI as Client. This PR makes the task to add a new `app_config` spec (a new section in the TOML) to be as simple as adding a new file in the `packages/app/src/cli/models/extensions/specifications` with a content similar to this:
```ts
import {createConfigExtensionSpecification} from '../specification.js'
import {zod} from '@shopify/cli-kit/node/schema'

const PosEmbedmentSchema = zod.object({
  pos: zod
    .object({
      embedded: zod.boolean(),
    })
    .optional(),
})

  identifier: 'pos_embedment',
  schema: PosEmbedmentSchema,
})

export default spec
```
The project to make the CLI as a Client, means that we want to move this definition (mainly the schema definition that will be returned as a JSON schema) as new field in the Core spec definition. In that way, adding a new spec or modifying a current one will require only changes in core an 

For the matter of adding a transform layer between the format in the TOML file and the format the server expect for that app module, we were thinking about adding a transform schema that could look like this for the previous example:

```json
{
    embedded: pos.embedded
}
```
with this transformation schema definition we would have this forma in the TOML file:
```toml
[pos]
embedded = true
```
but the content the CLI will send as the `config` of the extension registration will be:
{
  embedded: true
}

in this example we are flattening the object inside the TOML when the content is sent:
The same strategy could be followed to support the current TOML file schema to send the content of the new `app_config` modules. 
For the case of the new `app_ui` app module
_TOML schema definition:_
```ts
zod.object({
  app_preferences: zod
    .object({
      url: zod.string().max(255),
    })
    .optional(),
  application_url: zod.string(),
  embedded: zod.boolean(),
})
```
_ Transform schema_
```json
{
  app_url: application_url
  embedded: embedded,
  preferences_url: app_preferences.url
}
```
Regarding the CLI as client this transform definition can also be returned from the server side. In this example we should explore if it possible to declare complex transformation rules, ie: `embedded: zod.boolean() => ui: zod.enum('embedded', 'external')`

packages/app/src/cli/models/app/loader.ts
```diff
@@ -78,6 +76,11 @@ export async function loadConfigurationFile(
   }
 }
 
+const isVersionedAppConfigurationSchema = (content: {[key: string]: unknown}) => {
+  if (content.version && content.version === '3') return true
+  return false
```
isaacroldan, line 81:
```suggestion
  return content.version === '3'
```

What is this version and where is it coming from?

packages/app/src/cli/models/extensions/extension-instance.ts
```diff
@@ -18,6 +18,52 @@ import {joinPath} from '@shopify/cli-kit/node/path'
 import {useThemebundling} from '@shopify/cli-kit/node/context/local'
 import {touchFile, writeFile} from '@shopify/cli-kit/node/fs'
 
+export class ConfigExtensionInstance<TConfiguration = unknown> {
```
isaacroldan, line 21: why create a new class? Isn't `ExtensionInstance` generic enough to support this?

alvaro-shopify: no, `ExtensionInstance` uses a different schema. The new `app_config` schema is generated dynamically from the `specs` schemas instead of having a fixed schema definition. Modifying the `ExtensionInstance` to support that required lots of changes in the code base. As this `App config` instances are quite simple, I think this is a better approach

packages/app/src/cli/models/app/loader.ts
alvaro-shopify: This is a temp solution, it's representing the `version` of the config schema. The CLI adds it automatically for the apps that use `app_config` modules. I thought it's a better way of checking the schema used instead of parsing the content against the different schemas. However it's subject to changes

packages/app/src/cli/models/extensions/load-specifications.ts
```diff
@@ -1,25 +1,38 @@
...
+    specs: [...(specs as ExtensionSpecification[]), ...plugins],
+    configSpecs: appConfigSpecs as ConfigExtensionSpecification[],
+  }
 }
 
 /**
  * Load all specifications ONLY from the local file system
  */
-export async function loadFSExtensionsSpecifications(): Promise<ExtensionSpecification[]> {
+export async function loadFSExtensionsSpecifications(): Promise<unknown[]> {
```
isaacroldan, line 35: instead of unknown, this can be `ExtensionInstance | ConfigExtensionInstance`

So that then in the groupBy above you should just do:
```
  const {cli: specs, app_config: appConfigSpecs} = groupBy(local, (spec) => spec.managementExperience)
```

packages/app/src/cli/models/extensions/load-specifications.ts
```diff
@@ -1,25 +1,38 @@
...
+import {groupBy} from '@shopify/cli-kit/common/collection'
 import {fileURLToPath} from 'url'
 
 /**
  * Load all specifications from the local file system AND plugins
  */
-export async function loadLocalExtensionsSpecifications(config: Config): Promise<ExtensionSpecification[]> {
+export async function loadLocalExtensionsSpecifications(
+  config: Config,
+): Promise<{specs: ExtensionSpecification[]; configSpecs: ConfigExtensionSpecification[]}> {
```
isaacroldan, line 16: let's try not to use abbreviations, 
```ts
{specifications: ExtensionSpecification[]; configSpecifications: ConfigExtensionSpecification[]}
```

packages/app/src/cli/models/extensions/extension-instance.ts
```diff
@@ -18,6 +18,52 @@ import {joinPath} from '@shopify/cli-kit/node/path'
...
+  get graphQLType() {
+    return this.specification.identifier.toUpperCase()
+  }
+
+  get localIdentifier() {
+    return this.specification.identifier
+  }
+
+  get handle() {
+    return slugify(this.specification.identifier)
```
isaacroldan, line 41: instead of slugifying everytime, we can build the handle in the constructor

packages/app/src/cli/models/extensions/specification.ts
```diff
@@ -15,11 +15,25 @@ export type ExtensionFeature =
   | 'single_js_entry_path'
   | 'app_config'
 
+export type ConfigExtensionFeature = 'app_config'
+
+export type ExtensionSpecificationManagementExperience = 'cli' | 'app_config'
```
isaacroldan, line 20: I think this can just be `ManagementExperience` or `ExtensionManagementExperience`

packages/app/src/cli/services/generate.ts
```diff
@@ -46,12 +46,12 @@ async function generate(options: GenerateOptions) {
   const app: AppInterface = await loadApp({
     directory: options.directory,
     configName: options.configName,
-    specifications,
+    ...specifications,
   })
-  const availableSpecifications = specifications.map((spec) => spec.identifier)
+  const availableSpecifications = specifications.specs.map((spec) => spec.identifier)
```
isaacroldan, line 51: um, this is weird and will be weirder if we call the normal specs just `specifications` 🤔 
We maybe need a new name either for the object containing both types or the non-config ones:
```ts
allSpecifications: { specifications: ..., configSpecifications: ...}
// or
allSpecifications: { extensionSpecifications: ..., configSpecifications: ...}
// or
specifications: { extensionSpecifications: ..., configSpecifications: ...}
```

packages/app/src/cli/services/deploy.ts
```diff
@@ -94,9 +94,12 @@ export async function deploy(options: DeployOptions) {
         {
           title: uploadTaskTitle,
           task: async () => {
-            const appModules = await Promise.all(
+            const appSpecModules = await Promise.all(
               options.app.allExtensions.flatMap((ext) => ext.bundleConfig({identifiers, token, apiKey})),
             )
+            const appModules = appSpecModules.concat(
```
isaacroldan, line 100: same here about namings, they are a bit confusing

packages/app/src/cli/services/dev/update-extension.ts
```diff
@@ -57,10 +56,7 @@ export async function updateExtensionDraft({
     const errors = mutationResult.extensionUpdateDraft.userErrors.map((error) => error.message).join(', ')
     stderr.write(`Error while updating drafts: ${errors}`)
   } else {
-    const successMessage = extension.isConfigExtension
-      ? `Draft config updated successfully for: ${extension.name}`
-      : `Draft updated successfully for extension: ${extension.localIdentifier}`
-    outputInfo(successMessage, stdout)
+    outputInfo(`Draft updated successfully for extension: ${extension.localIdentifier}`, stdout)
```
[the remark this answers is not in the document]
isaacroldan: in any case, i'd recommend not to touch a lot of the `dev` process, since we are working on that for consistent-dev and will change considerably :)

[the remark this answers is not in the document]
pt2pham: Yup definitely trying to get out of the way of adding new things related to `dev` and `draft versions` when we can 😄 once all versioned app config is fetched through the shards + consistent dev is released, all these problems will solve themselves

packages/app/src/cli/models/app/loader.ts
alvaro-shopify: Finally this `version` field is gone as we are going to support the current app config toml schema

PR #3142, opened 2023-11-24
remove format config option as we default to json

alexanderMontague:

- we're moving away from `xml` webhook payloads, so we want to not support the option to set this proprty, and just default to `json`
- documented more in the [decision log](https://vault.shopify.io/gsd/projects/36745/decisions/4088)
- the related core changes to the module are in https://github.com/Shopify/shopify/pull/466324

- remove the format property from the schema as we don't need it and wont process it anymore

### Measuring impact

### Checklist


karenxie (approved):
LGTM

PR #3161, opened 2023-12-03
Add support app config spec

alvaro-shopify:
Follow-up: https://github.com/Shopify/cli/pull/3160

- Add support for the new version app config specifications. 
- Each section in the app config toml will be matched with one of the new `configuration` spec. Right now, the schema of each `configuration` specs will be defined in the global `AppSchema`. Once the specs are moved to the server side, the global `AppSchema` will be dynamically generated merging the `configuration` spec schemas.
- A new `configuration` extension instance will be created with the content of the section. 
-`deploy` command 
  - will ensure that `configuration` extensions are created first
  - `configuration` instances config content will be updated with the `appDeploy` API request.
- Creating a new `configuration` spec should be as simple as creating a new module inside the `@shopify/app` `extensions` folder. This should be an example of adding a new section to the app config schema
```ts
const PosConfigurationSchema = AppSchema.pick({pos: true}).strip()

  identifier: 'pos_configuration',
  schema: PosConfigurationSchema,
  appModuleFeatures: () => ['app_config'],
})

```

Co-authored-by: Alvaro Gutierrez [alvaro.gutierrezblanco@shopify.com](mailto:alvaro.gutierrezblanco@shopify.com)
Co-authored-by: Raluca Badoi [raluca.badoi@shopify.com](mailto:raluca.badoi@shopify.com)

- Add support for a new `app module feature` called `app_config`
- Create extension instances in the `app loader` for specs that includes the new `app_config` feature
- Add new `features` methods to check if some actions can be run for a specific extension instance.
- Adapt `app deploy` command to include extension instances that includes the new `app_config` feature. Current deploy prompt filters `configuration` registrations
- Adapt `app config link` command to overwrite the current content with the merged content got from the extension registrations with the feature `app_config`
- Add support for `point_of_sale` which is the most simple `app_config` section 
- Add support to the `extension specs` to define a `transform` and `reverseTransform` methods
  - `transform` is used in the `deploy` command to map the `configuration` extension schema defined in the `toml` to the one required in the server side. By default, `configuration` specs sends the content inside the sections in the `toml` at the top level, ie, the cli sends `{embedded: true}` as the content sent for the `point_of_sale` spec with this `toml` content
    ```toml
        [pos]
        embedded = true
    ```  
  - `reverseTransform` is used in the `link` command to map the remote `configuration` extension content to the one expected in the `toml` schema. By default, `app_config` remote content is mapped inside the section defined in the spec schema, ie, when the cli receives this `{embedded: true}` inside the config of a  `point_of_sale` extension registration
    ```toml
        [pos]
        embedded = true
    ```  
- Add a new env variable `SHOPIFY_CLI_VERSIONED_APP_CONFIG` to enable the `configuration` extensions 
  - If it's not used with the `deploy` command, the `configuration` extensions are not created/updated. This means that normal users will need to use the `push` and `link` command to modify the configuration
  - if it's used with the `deploy` command, the `configuration` extensions will be created/updated. This means that `point_of_sale` configuration will be updated/fetch from the app version modules content

### How to test your changes?
- Create a new `cli` app and run the `app config link` command to link it to a new remote app. This will fill the content of `pos.embedded` with its default value
#### Versioned config disabled
- Edit the `shopify.app.toml` file and modify the value of `pos.embedded`
- Run the `npm run shopify app deploy --path /PATH/TO/YOUR/APP` command to release a new version.
- Run the `npm run shopify app config link --path /PATH/TO/YOUR/APP` command. The value should be overwritten. This means that the `deploy` command is not creating the `configuration` app module and the content is overwritten with the api client configuration value
#### Versioned config enabled
- Edit the `shopify.app.toml` file and modify the value of `pos.embedded`
- Run the `SHOPIFY_CLI_VERSIONED_APP_CONFIG=1 npm run shopify app deploy --path /PATH/TO/YOUR/APP` command to release a new version.
- Edit the `shopify.app.toml` file again modify the value of `pos.embedded`
- Run the `npm run shopify app config link --path /PATH/TO/YOUR/APP` command. The value should be overwritten with the released one. This means that it's fetching the content from the extension registration, as we haven't run the command `app config push` to update the content

### Measuring impact

### Checklist

### DIFF
--- a/packages/app/src/cli/services/app/config/link.ts
+++ b/packages/app/src/cli/services/app/config/link.ts
@@ -59,6 +59,7 @@ export default async function link(options: LinkOptions, shouldRenderSuccess = t
 
   await saveCurrentConfig({configFileName, directory})
 
+  const usingVersionedAppConfig = remoteExtensionRegistrations.app.configurationRegistrations.length > 0
   if (shouldRenderSuccess) {
     renderSuccess({
       headline: `${configFileName} is now linked to "${remoteApp.title}" on Shopify`,
@@ -67,7 +68,12 @@ export default async function link(options: LinkOptions, shouldRenderSuccess = t
         [`Make updates to ${configFileName} in your local project`],
         [
           'To upload your config, run',
-          {command: formatPackageManagerCommand(localApp.packageManager, 'shopify app deploy')},
+          {
+            command: formatPackageManagerCommand(
+              localApp.packageManager,
+              `shopify app ${usingVersionedAppConfig ? 'deploy' : 'config push'}`,
+            ),
+          },
         ],
       ],
       reference: [
--- a/packages/app/src/cli/services/app/config/link.test.ts
+++ b/packages/app/src/cli/services/app/config/link.test.ts
@@ -107,7 +107,7 @@ embedded = false
         body: 'Using shopify.app.toml as your default config.',
         nextSteps: [
           [`Make updates to shopify.app.toml in your local project`],
-          ['To upload your config, run', {command: 'npm run shopify app deploy'}],
+          ['To upload your config, run', {command: 'npm run shopify app config push'}],
         ],
         reference: [
           {
@@ -186,7 +186,7 @@ embedded = false
         body: 'Using shopify.app.staging.toml as your default config.',
         nextSteps: [
           [`Make updates to shopify.app.staging.toml in your local project`],
-          ['To upload your config, run', {command: 'yarn shopify app deploy'}],
+          ['To upload your config, run', {command: 'yarn shopify app config push'}],
         ],
         reference: [
           {
@@ -245,7 +245,7 @@ embedded = false
         body: 'Using shopify.app.toml as your default config.',
         nextSteps: [
           [`Make updates to shopify.app.toml in your local project`],
-          ['To upload your config, run', {command: 'yarn shopify app deploy'}],
+          ['To upload your config, run', {command: 'yarn shopify app config push'}],
         ],
         reference: [
           {
@@ -516,7 +516,7 @@ embedded = false
         body: 'Using shopify.app.toml as your default config.',
         nextSteps: [
           [`Make updates to shopify.app.toml in your local project`],
-          ['To upload your config, run', {command: 'npm run shopify app deploy'}],
+          ['To upload your config, run', {command: 'npm run shopify app config push'}],
         ],
         reference: [
           {
@@ -615,6 +615,22 @@ api_version = "2023-07"
 embedded = true
 `
       expect(content).toEqual(expectedContent)
+      expect(renderSuccess).toHaveBeenCalledWith({
+        headline: 'shopify.app.staging.toml is now linked to "my app" on Shopify',
+        body: 'Using shopify.app.staging.toml as your default config.',
+        nextSteps: [
+          [`Make updates to shopify.app.staging.toml in your local project`],
+          ['To upload your config, run', {command: 'yarn shopify app deploy'}],
+        ],
+        reference: [
+          {
+            link: {
+              label: 'App configuration',
+              url: 'https://shopify.dev/docs/apps/tools/cli/configuration',
+            },
+          },
+        ],
+      })
     })
   })
 })


packages/app/src/cli/api/graphql/app_active_version.ts
```diff
@@ -30,7 +30,7 @@ interface AppModuleVersionSpecification {
   identifier: string
   name: string
   options: {
-    managementExperience: 'cli' | 'custom' | 'dashboard'
+    managementExperience: 'cli' | 'custom' | 'dashboard' | 'app_config'
```
pt2pham, line 33:
```suggestion
    managementExperience: 'cli' | 'custom' | 'dashboard' | 'app_config'
```

We'll want to rely on using `experience: :configuration` instead of management experience here for this kinds of app modules

alvaro-shopify: @pt2pham , yes, this has been changed in the next PRs but I can remove it here and rebase the changes to them

alvaro-shopify: in this PR this change is only used to filter the `configure` extensions in the `deploy` prompt

alvaro-shopify: I guess I'm going to leave as it is in this PR and these are the reasons:
- This is [modified](https://github.com/Shopify/cli/pull/3186/files#diff-f06458e98cf1627074d1b1fa874185b69af5918f918da73ca8e19ba4b58ef7aaR35) in the already reviewable [Adapt deploy release prompts](https://github.com/Shopify/cli/pull/3186/) PR which is a follow up of this one
- This PR is working as expected and it can be 🎩  so if it's approve it could be merged for the `versioned app master` branch
- I have split the `prototype` branch to make the PRs easier to review so each of them are dependant among them. This is the first PR so unless the suggestion fixes something that are not solved in the next ones, my recommendation is to simply reference to the place where it's solved
@pt2pham what do you think?

pt2pham: Yeah that works for me 🚢

alvaro-shopify:
> Ok, I made a mistake when testing it, it works as expected!
> 
> The comments are only suggestions and can be tackled later if you prefer.

Great!, they are good suggestions but to follow a similar strategy mentioned [here](https://github.com/Shopify/cli/pull/3161#discussion_r1425048742)  I have created another subtask in the master PR to go with this kind of comments in a final PR

packages/app/src/cli/models/extensions/specifications/app_config_pos_configuration.ts
```diff
@@ -0,0 +1,12 @@
+import {AppSchema} from '../../app/app.js'
+import {createConfigExtensionSpecification} from '../specification.js'
+
+const PosConfigurationSchema = AppSchema.pick({pos: true}).strip()
+
+const spec = createConfigExtensionSpecification({
+  identifier: 'pos_configuration',
```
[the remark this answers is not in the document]
pt2pham: https://github.com/Shopify/shopify/pull/469812

packages/app/src/cli/models/app/identifiers.ts
```diff
@@ -22,9 +22,14 @@ export interface Identifiers {
    * The extensions' numeric identifiers (expressed as a string).
    */
   extensionIds: IdentifiersExtensions
+
+  /**
+   * The extensions' unique identifiers which uuid is not managed.
+   */
+  extensionsNonUuidManaged: IdentifiersExtensions
```
gonzaloriestra, line 29: What if we call this `configurationIds`?

packages/app/src/cli/models/extensions/extension-instance.test.ts
```diff
@@ -188,11 +188,16 @@ describe('deployConfig', async () => {
 })
 
 describe('bundleConfig', async () => {
-  test('returns the uuid value when the extension is uuid managed', async () => {
+  test('returns the uuid from extensions value when the extension is uuid managed', async () => {
```
gonzaloriestra, line 191:
```suggestion
  test('returns the uuid from extensions when the extension is uuid managed', async () => {
```

packages/app/src/cli/models/extensions/extension-instance.test.ts
```diff
@@ -202,23 +207,26 @@ describe('bundleConfig', async () => {
         uuid: 'theme-uuid',
       }),
     )
-    expect(got).not.toHaveProperty('specificationIdentifier')
   })
 
-  test('returns the specification identifier when the extension is not uuid managed', async () => {
+  test('returns the uuid from extensionsNonUuidManaged value when the extension is not uuid managed', async () => {
```
gonzaloriestra, line 212:
```suggestion
  test('returns the uuid from extensionsNonUuidManaged when the extension is not uuid managed', async () => {
```

packages/app/src/cli/services/context/identifiers-extensions.test.ts
```diff
@@ -498,13 +517,146 @@ describe('ensureExtensionsIds: excludes non uuid managed extensions', () => {
...
+    expect(createExtension).toHaveBeenCalledOnce()
+    expect(got).toEqual(
+      ok({
+        extensions: {},
+        extensionIds: {},
+        extensionsNonUuidManaged: {'point-of-sale': 'UUID_C_A'},
+      }),
+    )
+  })
+  test('when the beta flag is disable configuration extensions are not created', async () => {
```
gonzaloriestra, line 571:
```suggestion
  test('when the beta flag is disabled configuration extensions are not created', async () => {
```

packages/app/src/cli/services/context/identifiers-extensions.ts
```diff
@@ -117,17 +135,43 @@ export async function ensureExtensionsIds(
...
+    if (possibleMatch) validMatches[local.localIdentifier] = possibleMatch.uuid
+    else extensionsToCreate.push(local)
   })
+
+  if (extensionsToCreate.length > 0) {
+    const newIdentifiers = await createExtensions(extensionsToCreate, appId, false)
+    for (const [localIdentifier, registration] of Object.entries(newIdentifiers)) {
+      validMatches[localIdentifier] = registration.uuid
+    }
+  }
```
gonzaloriestra, line 162: We can only have 1 configuration registration and the only non-uuid-managed extension is the configuration, right? In that case, we could simplify this a lot with those assumptions, forgetting about arrays and so on.

packages/app/src/cli/models/app/identifiers.ts
alvaro-shopify: I wanted to introduce the concept `uuidManaged` (right now only extensions that includes `app_config` feature) as extensions which `uuids` are not going to be neither matched nor stored in the `.env` file. Maybe in the future we should extension the method `isUuidManaged` to include a different condition, so all the extensions that matches that condition will have the same behaviour. That's why this is the list of `non uuid managed`

gonzaloriestra: I just prefer a clearer name, even if it's not so generic. But it's only a name, let's move on!

packages/app/src/cli/services/app/config/link.ts
```diff
@@ -51,7 +67,7 @@ export default async function link(options: LinkOptions, shouldRenderSuccess = t
         [`Make updates to ${configFileName} in your local project`],
         [
           'To upload your config, run',
-          {command: formatPackageManagerCommand(localApp.packageManager, 'shopify app config push')},
+          {command: formatPackageManagerCommand(localApp.packageManager, 'shopify app deploy')},
```
pt2pham, line 70: Does it make sense to expose this messaging yet if this is just merging to `main`? Would a release be cut before we're ready to release?

packages/app/src/cli/models/extensions/extension-instance.ts
```diff
@@ -105,7 +109,9 @@ export class ExtensionInstance<TConfiguration extends BaseConfigType = BaseConfi
     this.directory = options.directory
     this.specification = options.specification
     this.devUUID = `dev-${randomUUID()}`
-    this.handle = this.configuration.handle ?? slugify(this.configuration.name ?? '')
+    this.handle = this.specification.appModuleFeatures().includes('app_config')
```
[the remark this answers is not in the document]
pt2pham: Is this feedback still relevant?

pt2pham (approved):
Did a higher level review since I'm not as familiar with Typescript best practices or the CLI, but the app module related code at a high level makes sense. Scope of changes going in makes sense too.

PR #3170, opened 2023-12-05
Add some non gated app config specs

alvaro-shopify:

Follow-up: https://github.com/Shopify/cli/pull/3161

NOTES:
- Rebase follow-up branch before merging it
- Edit this PR to use `main` before merging the follow-up

- Add some of the reaming `configuration` specs that are not gated (`app access`, `declarative webhooks`)
  - `app_home`
  - `app_proxy`
- Add support to a `declarative configuration` to transforma the extension configuration content to send it/receive from the server. `undefined` fields won't be mapped
  - `app_home` spec format is different in the toml than in the backend
- App loader only loads `configuration` extensions if they have some content (ie add a config section without any field inside because they are optional). This means that the extension wont be neither created in this case if it does not exists nor updated otherwise

### How to test your changes?
- Create a new `cli` app and run the `app config link` command to link it to a new remote app.
- Edit the `shopify.app.toml` file and modify some value from:
  - application_url
  - embedded
  - app_preferences.url
  - app_proxy
- Run the `SHOPIFY_CLI_VERSIONED_APP_CONFIG=1 npm run shopify app deploy --path /PATH/TO/YOUR/APP` command to release a new version.
- Edit the `shopify.app.toml` file again modify some of the previous values
- Run the `npm run shopify app config link --path /PATH/TO/YOUR/APP` command. The values should be overwritten with the released ones. This means that it's fetching the content from the extension registration, as we haven't run the command `app config push` to update the content

### Measuring impact

### Checklist


packages/app/src/cli/models/extensions/specification.ts
```diff
@@ -118,19 +123,78 @@ export function createConfigExtensionSpecification<TConfiguration extends BaseCo
...
+ * ```
+ *  and a tranform config content like this:
+ * ```json
+ * { 'target.fieldTargetA': 'source.fieldSourceA'}
+ * ```
+ * the method returns the following object:
+ * ```json
+ * { source: { fieldSourceA: 'valueA' } }
+ * ```
+ * The tranfomat can be applied in both ways depending on the reverse parameter
```
gonzaloriestra, line 172: Is that example right? 

```suggestion
 * { source: { fieldTargetA: 'valueA' } }
 * ```
 * The transformation can be applied in both ways depending on the reverse parameter
```

packages/app/src/cli/models/extensions/specifications/app_config_app_proxy.test.ts
```diff
@@ -0,0 +1,52 @@
+import spec from './app_config_app_proxy.js'
+import {describe, expect, test} from 'vitest'
+
+describe('app_config_app_proxy', () => {
+  describe('transform', () => {
```
gonzaloriestra, line 5: Do you think it's worth to test transform and reverseTransform if they use the default one?

packages/app/src/cli/models/extensions/specifications/app_config_pos_configuration.ts
```diff
@@ -6,7 +6,6 @@ const PosConfigurationSchema = AppSchema.pick({pos: true}).strip()
 const spec = createConfigExtensionSpecification({
   identifier: 'pos_configuration',
   schema: PosConfigurationSchema,
-  appModuleFeatures: () => ['app_config'],
```
gonzaloriestra, line 9: Why removing this? 🤔

packages/app/src/cli/models/extensions/specifications/app_config_privacy_compliance_webhooks.ts
```diff
@@ -0,0 +1,18 @@
+import {TransformationConfig, createConfigExtensionSpecification} from '../specification.js'
+import {AppSchema} from '../../app/app.js'
+
+const WebhookSchema = AppSchema.pick({webhooks: true}).strip()
+
+const PrivacyComplianceTransformConfig: TransformationConfig = {
+  customers_redact_url: 'webhooks.privacy_compliance.customer_deletion_url',
+  customers_data_request_url: 'webhooks.privacy_compliance.customer_data_request_url',
+  shop_redact_url: 'webhooks.privacy_compliance.shop_deletion_url',
```
gonzaloriestra, line 9: We'll finally adopt a different schema for GDPR webhooks (https://vault.shopify.io/gsd/projects/36745/decisions/4222), but we'll need to support the old one, so we can leave this for now. And we can add it the new schema later.

packages/app/src/cli/services/app/config/link.test.ts
```diff
@@ -530,7 +530,7 @@ embedded = false
     })
   })
 
-  test('overwrite the api client configuration with the remote app_config extension registrarions', async () => {
+  test('the api client configuration is depp merged with the remote app_config extension registrarions', async () => {
```
gonzaloriestra, line 533:
```suggestion
  test('the api client configuration is deep merged with the remote app_config extension registrations', async () => {
```

gonzaloriestra (approved):
Nice work! Again, all the suggestions are minor and we can tackle them later if you prefer.

The only thing that didn't work as expected was the privacy_compliance webhooks, should it work? Is the server branch ready to handle it? I got all the fields after running link, except that one.

packages/app/src/cli/models/extensions/specification.ts
```diff
@@ -118,19 +123,78 @@ export function createConfigExtensionSpecification<TConfiguration extends BaseCo
...
+    transform: resolveAppConfigTransform(spec.transformConfig),
+    reverseTransform: resolveReverseAppConfigTransform(spec.schema, spec.transformConfig),
   })
 }
 
+function resolveAppConfigTransform(transformConfig?: TransformationConfig) {
+  if (!transformConfig) return (content: {[key: string]: unknown}) => defaultAppConfigTransform(content)
+
+  return (content: {[key: string]: unknown}) => appConfigTransform(content, transformConfig)
+}
```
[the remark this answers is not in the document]
alvaro-shopify: After a pairing session we decided to keep the implementation as it is

PR #3178, opened 2023-12-07
Using the experience property nstead of relying on the management_exp…

ani-shopify:
…erience

Fixes #0000 

### Measuring impact

### Checklist

### DIFF
--- a/packages/app/src/cli/services/context/identifiers.ts
+++ b/packages/app/src/cli/services/context/identifiers.ts
@@ -218,7 +218,7 @@ async function resolveRemoteExtensionIdentifiersBreakdown(
 
   const appModuleVersionsNonConfig =
     activeAppVersion.app.activeAppVersion?.appModuleVersions.filter(
-      (module) => module.specification !== null && module.specification?.options.managementExperience !== 'app_config',
+      (module) => module.specification !== null && module.specification?.experience !== 'configuration',
     ) || []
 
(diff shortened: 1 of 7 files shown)


packages/app/src/cli/services/context/identifiers.ts
```diff
@@ -218,7 +218,7 @@ async function resolveRemoteExtensionIdentifiersBreakdown(
 
   const appModuleVersionsNonConfig =
     activeAppVersion.app.activeAppVersion?.appModuleVersions.filter(
-      (module) => module.specification !== null && module.specification?.options.managementExperience !== 'app_config',
+      (module) => module.specification !== null && module.specification?.experience !== 'configuration',
```
ani-shopify, line 221: I wonder if we could simplify the logic in this file. It seems to me that we do compare local extensions/configurations against the module versions included in the active app version. It might be helpful to create a diagram like [this one](https://www.figma.com/file/N4S2fhZ2tgpaNBjCM637bc/Partitioning-by-Deploy-Origin-and-Management-Experience?type=whiteboard&node-id=0-1&t=ZmAvqh0xUuen6cXL-0) to show what exactly we are comparing against what. That we can see if there's a way to simplify this further.

cc @alvaro-shopify

pt2pham:
Closing this as we've shipped these changes into Core already

PR #3180, opened 2023-12-07
Deploy fetch remote specs

alvaro-shopify:

Follow-up: https://github.com/Shopify/cli/pull/3170
Follow-up:  https://github.com/Shopify/cli/pull/3214

- Make `specifications` optional for the `App loader`. Set default value to an empty array so no extensions are loaded in this case.
- `deploy` command loads the app again with the extensions once the `remote app` is selected, thus, the `gated` specs are applied and in case they are not allowed the extensions are not loaded
- Temporal `app access` config module added. Supports the new section `access`
  - `direct_api_offline_access` true | false to enable Online access for Direct API

### How to test your changes?
- Create a new `cli` app and run the `app config link` command to link it to a new remote app. 

- With the org beta flag `App Access Module` disabled
  - Add the section `access` with whatever value 
  - Run the `deploy` command
  - Run the `link` command. The added `access` section content will be removed as the app module wasn't uploaded in the previous `deploy`
- With the org beta flag `App Access Module` enabled
  - Add the section `access` with whatever value 
  - Run the `deploy` command
  - Modify the value inside the `access` section
  - Run the `link` command. The `access` section content will be restored to the content it had before running the `deploy` command as it was uploaded with it
  

### Measuring impact

### Checklist

### DIFF
--- a/packages/app/src/cli/api/graphql/extension_specifications.ts
+++ b/packages/app/src/cli/api/graphql/extension_specifications.ts
@@ -32,7 +32,7 @@ export interface RemoteSpecification {
   gated: boolean
   externalIdentifier: string
   options: {
-    managementExperience: 'cli' | 'custom' | 'dashboard'
+    managementExperience: 'cli' | 'custom' | 'dashboard' | 'app_config'
     registrationLimit: number
   }
   features?: {
--- a/packages/app/src/cli/api/graphql/extension_specifications.ts
+++ b/packages/app/src/cli/api/graphql/extension_specifications.ts
@@ -8,6 +8,7 @@ export const ExtensionSpecificationsQuery = gql`
       externalIdentifier
       identifier
       gated
+      experience
       options {
         managementExperience
         registrationLimit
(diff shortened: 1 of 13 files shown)


packages/app/src/cli/models/extensions/specification.ts
```diff
@@ -189,7 +189,7 @@ function appConfigTransform(
     const originPath = reverse ? mappedPath : objectPath
     const targetPath = reverse ? objectPath : mappedPath
     const sourceValue = getPathValue(content, originPath)
-    setPathValue(transformedContent, targetPath, sourceValue)
+    if (sourceValue !== undefined) setPathValue(transformedContent, targetPath, sourceValue)
```
gonzaloriestra, line 192: If the source is not found, shouldn't it fail? Or which scenario is this check covering?

alvaro-shopify: Already discussed and approved in the previous [PR](https://github.com/Shopify/cli/pull/3170)

PR #3275, opened 2024-01-17
Remove validations on CLI so more verbose error messages are bubbled up from core

jamieguerrero:

We are doing validations for TOML values on both `cli` and `shopify`. If the invalid configs are caught by `cli`, the error messages look like this:

Screenshot 2024-01-17 at 2 08 51 PM
Screenshot 2024-01-17 at 2 07 56 PM

The issues with these error messages are:
1. ~~We are only able to see one error message at a time~~ We may be able to use report mode when loading an app to see aggregated messages. Investigating with @amcaplan.
2. The error message is not always descriptive, especially if there are multiple issues

This PR removes CLI validation in favour of hitting `shopify` to get more descriptive error messages on multiple app configuration modules for the user.

Screenshot 2024-01-17 at 2 40 51 PM

### How to test your changes?

- Link your app, and change your name/handle values to something invalid
- View error messages

### Measuring impact

### Checklist

### DIFF
--- a/packages/app/src/cli/models/extensions/specifications/app_config_branding.ts
+++ b/packages/app/src/cli/models/extensions/specifications/app_config_branding.ts
@@ -3,11 +3,7 @@ import {zod} from '@shopify/cli-kit/node/schema'
 
 const BrandingSchema = zod.object({
   name: zod.string().max(30),
-  handle: zod
-    .string()
-    .max(256)
-    .refine((value) => value && /^\w*(?!-)[a-z0-9-]+(?<!-)$/.test(value))
-    .optional(),
+  handle: zod.string().optional(),
 })
 
 const BrandingTransformConfig: TransformationConfig = {


isaacroldan:
We've always preferred to fail early, otherwise we might ask you a bunch of questions in prompts, let you think everything is going to work and then fail at the last moment.

It is possible to customize errors in [Zod](https://zod.dev/?id=strings), but since we might move to JSON Schema, I wouldn't invest to much on it.

Since the UX is better when we validate in the server, maybe is worth discussing this with Meredith.

jamieguerrero:
> We've always preferred to fail early, otherwise we might ask you a bunch of questions in prompts, let you think everything is going to work and then fail at the last moment.
> 
> It is possible to customize errors in [Zod](https://zod.dev/?id=strings), but since we might move to JSON Schema, I wouldn't invest to much on it.
> 
> Since the UX is better when we validate in the server, maybe is worth discussing this with Meredith.

Thanks for the feedback @isaacroldan! I think @jessehoek is our designer on this project so I'll defer to him quickly unless perhaps we need more eyes? 

Also, I'm curious if we have an issue to track the move from zod to JSON Schema so I can learn more about the decision making there and if folks need more hands.

jessehoek:
The way the error is presented from the server is much easier to read and see all the issues at once. But I agree with the principle of validating and failing early so that you save time. It does feel weird to see one error at a time, but I would consider improving that local validation experience vs delaying the moment you see the error. 

Question: are there some validations we can't do until we hit the remote app in the server? Like doesn't the "can't contain Shopify" rule only apply to 3P apps, and 1P can do this? Or no?

As an aside, no matter what direction we go in, I want to make sure we don't expose the config module names or groupings to end users. We present all of this as one **app configuration** in the TOML. So our name "Branding" is only an internal concept.

PR #3281, opened 2024-01-18
Add redeemable schema to payments extension

marklevi:
**Why are these changes introduced?**

This PR introduces a `redeemable` payments extension schema.

Subsequent PRs will include additional schema definitions, one for each payments extension, that will populate the schema union.

These changes will be behind a feature flag and the current CLI version `3.53` will not have `Payment Extensions` available as an option to choose from

### How to test your changes?

- View `Payments Extensions`

These changes cannot be tested with the CLI version `3.53` and is can only be tested internally at this moment.

### Measuring impact

### Checklist

### DIFF
--- a/packages/app/src/cli/models/extensions/specifications/payments_app_extension.ts
+++ b/packages/app/src/cli/models/extensions/specifications/payments_app_extension.ts
@@ -2,18 +2,18 @@ import {
   OffsitePaymentsAppExtensionConfigType,
   OffsitePaymentsAppExtensionSchema,
   offsitePaymentsAppExtensionDeployConfig,
+  OFFSITE_TARGET,
 } from './payments_app_extension_schemas/offsite_payments_app_extension_schema.js'
+import {
+  REDEEMABLE_TARGET,
+  RedeemablePaymentsAppExtensionConfigType,
+  redeemablePaymentsAppExtensionDeployConfig,
+  RedeemablePaymentsAppExtensionSchema,
+} from './payments_app_extension_schemas/redeemable_payments_app_extension_schema.js'
 import {createExtensionSpecification} from '../specification.js'
-import {BaseSchema} from '../schemas.js'
 import {zod} from '@shopify/cli-kit/node/schema'
 
-// NOTE: SamplePaymentsAppExtensionSchema is used since zod doesn't accept a single-element union. It will be removed
-// once we add a second schema for another payment app type.
-const SamplePaymentsAppExtensionSchema = BaseSchema.extend({
-  targeting: zod.array(zod.object({target: zod.literal('payments.sample.render')})).length(1),
-})
-
-const PaymentsAppExtensionSchema = zod.union([OffsitePaymentsAppExtensionSchema, SamplePaymentsAppExtensionSchema])
+const PaymentsAppExtensionSchema = zod.union([OffsitePaymentsAppExtensionSchema, RedeemablePaymentsAppExtensionSchema])
 
 export type PaymentsAppExtensionConfigType = zod.infer<typeof PaymentsAppExtensionSchema>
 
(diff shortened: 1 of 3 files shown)


packages/app/src/cli/models/extensions/specifications/payments_app_extension_schemas/redeemable_payments_app_extension_schema.ts
```diff
@@ -0,0 +1,55 @@
+import {BaseSchema} from '../../schemas.js'
+import {zod} from '@shopify/cli-kit/node/schema'
+
+export type RedeemablePaymentsAppExtensionConfigType = zod.infer<typeof RedeemablePaymentsAppExtensionSchema>
+
+const MAX_LABEL_SIZE = 50
```
apliano, line 6: Why do we have this constant (with the same value) repeated in multiple schema files? e.g: [here](https://github.com/Shopify/cli/blob/98cf1ce9258ab9b4c739a58c76fd282a4738909b/packages/app/src/cli/models/extensions/specifications/payments_app_extension_schemas/offsite_payments_app_extension_schema.ts#L6)

marklevi: I think each label size can be tailored specifically to the extension type which is why I kept them in separate files even though they are technically the same value

packages/app/src/cli/models/extensions/specifications/payments_app_extension.ts
```diff
@@ -24,6 +22,8 @@ const spec = createExtensionSpecification({
   deployConfig: async (config, _) => {
     if (config.targeting[0]!.target === 'payments.offsite.render') {
```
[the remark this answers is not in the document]
wikiti: > Is it 100% ensured that the array targeting has at least one element (and that this element has a target)?

It will, otherwise, it would fail the schema validation :)

packages/app/src/cli/models/extensions/specifications/payments_app_extension_schemas/redeemable_payments_app_extension_schema.ts
```diff
@@ -0,0 +1,55 @@
...
+
+export const RedeemablePaymentsAppExtensionSchema = BaseSchema.extend({
+  targeting: zod.array(zod.object({target: zod.literal('payments.redeemable.render')})).length(1),
```
apliano, line 9: Do we have a subset of this properties that will be common to all of the Payments Schemas? We could extract those and extend them to avoid duplication 🤔 
Maybe I'm getting ahead of myself and this will be done at a later step.

wikiti: We might; the idea is to refactor that after all schemas are defined, so we avoid as much toil work as possible 😄

wikiti (approved):

packages/app/src/cli/models/extensions/specifications/payments_app_extension_schemas/redeemable_payments_app_extension_schema.ts
```diff
@@ -0,0 +1,55 @@
...
+  void_session_url: zod.string().url().optional(),
+  balance_url: zod.string().url(),
+  merchant_label: zod.string().max(MAX_LABEL_SIZE),
+  buyer_label: zod.string().max(MAX_LABEL_SIZE).optional(),
+  buyer_label_translations: zod.array(zod.object({locale: zod.string(), label: zod.string()})).optional(),
+  supported_countries: zod.array(zod.string()),
+  supported_payment_methods: zod.array(zod.string()),
+  test_mode_available: zod.boolean(),
+  redeemable_type: zod.literal('gift_card'),
+  checkout_payment_method_fields: zod.array(zod.string()).optional(),
```
wikiti, line 23: We are missing the `ui_extension_registration_uuid` property; are we planning on adding it later?

marklevi: @wikiti part of a future work

packages/app/src/cli/models/extensions/specifications/payments_app_extension_schemas/redeemable_payments_app_extension_schema.ts
isaacroldan, line 9: I think it'd make sense to extract `'payments.redeemable.render'` to a const here that can then be imported from the specification, so that you don't need to hardcode the target there as well

PR #3292, opened 2024-01-23
Refactor and extract common fields from payment app schemas

wikiti:

To keep code maintanable, after creating the schemas for all payment app types, we decided to do a minor refactor to extract common fields into a base schema.

A new base schema, `BasePaymentsAppExtensionSchema`, is introduced with all the common fields. All existing schemas are going to extend this base schema, and add new fields when required.

Instead of extracting **strict** common fields (same types, same validations), I decided to leverage on the `omit` and `required` zod methods. This way, we can define *almost-common* fields, and make minor tweaks based on the payment app. For example, if a field is common and optional to all payment app types, but it is required within a specific type, then the `required` method would change the optional field from the base schema to be required in the sub schema. Something similar applies to `omit`; if a field is present in *almost* all payment app types, but not in all of them, the `omit` method can be used to skip that field for the specific type.

I tried to find a balance between extracting common fields and avoiding repetition. I think the current implementation doesn't create too much indirection between the base schema and the subschemas, but please let me know in comments if you find it hard to follow 🙏🏿 

I also made some minor changes to some schemas that were not properly formed, such as the custom credit card encryption certificate field

### How to test your changes?

- View `Payments Extensions`

### Measuring impact

### Checklist

### DIFF
--- a/packages/app/src/cli/models/extensions/specifications/payments_app_extension_schemas/base_payments_app_extension_schema.ts
+++ b/packages/app/src/cli/models/extensions/specifications/payments_app_extension_schemas/base_payments_app_extension_schema.ts
@@ -1,0 +1,45 @@
+import {BaseSchema} from '../../schemas.js'
+import {zod} from '@shopify/cli-kit/node/schema'
+
+const MAX_LABEL_SIZE = 50
+
+export const BasePaymentsAppExtensionSchema = BaseSchema.extend({
+  api_version: zod.string(),
+  payment_session_url: zod.string().url(),
+  refund_session_url: zod.string().url().optional(),
+  capture_session_url: zod.string().url().optional(),
+  void_session_url: zod.string().url().optional(),
+
+  supported_countries: zod.array(zod.string()),
+  supported_payment_methods: zod.array(zod.string()),
+
+  test_mode_available: zod.boolean(),
+
+  merchant_label: zod.string().max(MAX_LABEL_SIZE),
+
+  input: zod
+    .object({
+      metafield_identifiers: zod
+        .object({
+          namespace: zod.string(),
+          key: zod.string(),
+        })
+        .optional(),
+    })
+    .optional(),
+})
+
+export const BuyerLabelSchema = zod.object({
+  buyer_label: zod.string().max(MAX_LABEL_SIZE).optional(),
+  buyer_label_translations: zod.array(zod.object({locale: zod.string(), label: zod.string()})).optional(),
+})
+
+export const DeferredPaymentsSchema = zod.object({
+  supports_installments: zod.boolean(),
+  supports_deferred_payments: zod.boolean(),
+})
+
+export const ConfirmationSchema = zod.object({
+  confirmation_callback_url: zod.string().url().optional(),
+  supports_3ds: zod.boolean(),
+})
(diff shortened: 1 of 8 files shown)


packages/app/src/cli/models/extensions/specifications/payments_app_extension_schemas/credit_card_payments_app_extension_schema.ts
```diff
@@ -1,52 +1,38 @@
...
+
+export const CreditCardPaymentsAppExtensionSchema = BasePaymentsAppExtensionSchema.required({
+  refund_session_url: true,
+  capture_session_url: true,
+  void_session_url: true,
 })
+  .omit({
+    buyer_label: true,
+    buyer_label_translations: true,
+  })
```
wikiti, line 18: After a brief discussion with @isaacroldan, I'll change the usage of `omit` to instead use composed schemas (e.g. `BasePaymentsAppExtensionSchema.extend(BuyerLabelSchema).extend({....})`)

packages/app/src/cli/models/extensions/specifications/payments_app_extension_schemas/base_payments_app_extension_schema.ts
```diff
@@ -0,0 +1,45 @@
+import {BaseSchema} from '../../schemas.js'
+import {zod} from '@shopify/cli-kit/node/schema'
+
+const MAX_LABEL_SIZE = 50
+
+export const BasePaymentsAppExtensionSchema = BaseSchema.extend({
```
marklevi, line 6: I like the approach of using composition to build each schema using logical groupings like `DeferredPaymentsSchema` and `ConfirmationSchema`. However, I am not so sure if lumping a group of fields under `BasePaymentsAppExtensionSchema` in aligned with that idea. I think it's fine to have some duplicate fields in each extension schema if there isnt a logical grouping associated with those. What do you think?

wikiti: > I think it's fine to have some duplicate fields in each extension schema if there isnt a logical grouping associated with those. Nonetheless, it isnt a sticking point. What do you think?

That breaks the idea of lifting common fields into a single schema. Having schemas being defined as `BaseSchema.merge(PaymentsAppUrlExtensionsSchema).merge(MerchantLabelSchema).merge(InputSchema).merge(TestModeSchema)` seems insane to me. That's a lot of boilerplate code for just strictly sticking to composition.

I think we get the best of both worlds by using a common schema to all payment apps (`BasePaymentsAppExtensionSchema`), and then, to prevent code duplication, using schema "plugins" to extend the available fields. 

It's what we already do for the common `BaseSchema` schema, common to all extensions

marklevi (approved):

packages/app/src/cli/models/extensions/specifications/payments_app_extension_schemas/base_payments_app_extension_schema.ts
marklevi: I think the extracting out the common properties has come at a slight cost in terms of code readability but I understand the approach. approving the PR !

PR #3322, opened 2024-01-28
Add link from app dev to promote Edition

amcaplan:

Get the word out about Edition to developers by meeting them where they are - in the console!

Adds a menu item to `app dev` to open Edition. Includes link tracking, and is timed to coincide with Edition start + 1 hour, and disappear 1 week later.

If we see a good response to this, maybe we'll build something even more fun next time 😉

Screenshot 2024-01-28 at 12 12 33

Note this is Not Good Code™️ because the goal was to scrape something together that we are confident works, then revert and possibly rebuild later, depending on how the developer community responds.

### How to test your changes?

```
$ IS_EDITION_WEEK=1 p shopify app dev --path  /path/to/your/app
```

And press `e` when the menu loads

### Measuring impact

### Checklist

### DIFF
--- a/packages/app/src/cli/services/dev/ui/components/Dev.tsx
+++ b/packages/app/src/cli/services/dev/ui/components/Dev.tsx
@@ -33,6 +33,7 @@ export interface DevProps {
   }
   pollingTime?: number
   developerPreview: DeveloperPreviewController
+  isEditionWeek?: boolean
 }
 
 const Dev: FunctionComponent<DevProps> = ({
@@ -44,6 +45,7 @@ const Dev: FunctionComponent<DevProps> = ({
   app,
   pollingTime = 5000,
   developerPreview,
+  isEditionWeek,
 }) => {
   const {canEnablePreviewMode, developmentStorePreviewEnabled} = app
   const {isRawModeSupported: canUseShortcuts} = useStdin()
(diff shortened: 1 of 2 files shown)


shauns (approved):

packages/app/src/cli/services/dev/ui/components/Dev.tsx
```diff
@@ -206,6 +214,12 @@ const Dev: FunctionComponent<DevProps> = ({
         >
           {canUseShortcuts ? (
             <Box flexDirection="column">
+              {isEditionWeek ? (
+                <Text>
+                  {figures.pointerSmall} Press <Text bold>e</Text> {figures.lineVertical} check out {season} Edition
+                  {` ${year}`}, live NOW with 100+ product announcements!
```
gonzaloriestra, line 220: We will show this message for a week, so `live NOW` won't be very accurate... Why not simply showing `check out Winter Edition 2024 with 100+ product announcements!`?

amcaplan: Well, I don't think `live NOW` implies it just went live, but rather that at the current moment it's live.

Probably not as exciting after a few days, but I don't think anyone will complain... also keep in mind people might not even see it for a few days if they don't upgrade CLI. So it's only for people who are actually using the latest (creating new apps or upgrading).

PR #3339, opened 2024-01-31
link command uses local scopes to create a new remote app

alvaro-shopify:
If you create a new `remix` app from scratch, run `dev` and create a new remote app, the default scopes fetched from the `remix templates` are lost.

In this use case, the local scopes should be kept

- When the local app configuration uses the legacy format, the scopes are used to create the new remote app
- This is a regression introduced by this two PRs:
  - [link uses an empty app if the current used app is different from the remote one](https://github.com/Shopify/cli/pull/3333)
  - [dev with the default toml without client_id reuses it](https://github.com/Shopify/cli/pull/3336)
  so this fix is working for those use cases too

### How to test your changes?
- Create a new app using Remix
- Run the dev command and create a new remote app
- Check that the `write_products` scope is included in the `toml`
- Open the app and click the create products button to check it's working

### Measuring impact

### Checklist

### DIFF
--- a/packages/app/src/cli/services/app/config/link.ts
+++ b/packages/app/src/cli/services/app/config/link.ts
@@ -103,7 +103,6 @@ async function loadAppOrEmptyApp(
   remoteBetas?: BetaFlag[],
   remoteApp?: OrganizationApp,
 ): Promise<AppInterface> {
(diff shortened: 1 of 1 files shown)


rezaansyed (approved):
🎩 'd and changes LGTM!

PR #3345, opened 2024-02-01
Support graphiql query param

matteodepalo:
Hello! 👋 

It would be great if we could support more features of the GraphiQL explorer.
In particular, there is no way to pass a `query` query param to the url at the moment, but the GraphiQL explorer supports this and you can see it in action here [https://graphql.github.io/swapi-graphql?query=query](https://graphql.github.io/swapi-graphql?query=query)

This PRs simply forwards the `query` query param if present.

### How to test your changes?

- Run `dev`
- Press `g`
- Nothing should change
- Add `?query=test` to the url in the browser and refresh
- There should be a new tab with the query

### DIFF
--- a/packages/app/src/cli/services/dev/graphiql/server.ts
+++ b/packages/app/src/cli/services/dev/graphiql/server.ts
@@ -149,6 +149,7 @@ export function setupGraphiQLServer({
         {
           url: localhostUrl,
           defaultQueries: [{query: defaultQuery}],
+          query: req.query.query,
         },
       ),
     )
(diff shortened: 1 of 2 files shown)


packages/app/src/cli/services/dev/graphiql/server.ts
```diff
@@ -149,6 +151,7 @@ export function setupGraphiQLServer({
         {
           url: localhostUrl,
           defaultQueries: [{query: defaultQuery}],
+          query,
```
amcaplan, line 154: What about instead:

```suggestion
          defaultQueries: [{query: query ?? defaultQuery}],
```

And then you can skip the template changes below.

amcaplan: Hmmmm, actually I do see that the behavior is different, where if we open with a query it gets added to the currently open tabs, vs. replacing the defaults. So, I think what you did makes sense.

matteodepalo: I originally did exactly what you said :D but yeah in the end they support this separate `query` attribute for a reason.

amcaplan (approved):
Works like a charm! Excited to see what this unlocks for us in dev docs.

PR #3369, opened 2024-02-05
Fix the error message that started occurring since Ruby 3.3.0 [2]

karreiro:


packages/cli-kit/assets/cli-ruby/bin/shopify
```diff
@@ -1,5 +1,11 @@
 #!/usr/bin/env ruby
 
+module Warning
+  def self.warn(string)
+    super if ENV["DEBUG"]
```
gonzaloriestra, line 5:
```suggestion
    super if ENV["DEBUG"] || !string.include?('was loaded from the standard library')
```

What if we check the message to avoid skipping other warnings that may be useful for the users?

gonzaloriestra (approved):

packages/cli-kit/assets/cli-ruby/bin/shopify
karreiro: Thanks for the review and the comment, @gonzaloriestra!

Considering how we encapsulate the usage of Ruby in the CLI, it's a bit hard for users to handle warnings coming from Ruby. So, I believe it's a bit cleaner to hide them by default, and still keep them accessible with the `--verbose` flag. If we identify something important to warn about tho, we can try to do that in advance at the TS level, and we can show the warning with `renderWarning` (from the CLI kit).

Thanks again for bringing this up tho! It's a really valuable concern, and I hadn't thought about it before reading this suggestion! :) 🙏

gonzaloriestra: It makes sense, thanks for the explanation!

PR #3375, opened 2024-02-05
Refresh theme auth

### LINKED ISSUE #3173: [Bug]: `shopify app dev` stops authenticating development store preview

App

### Expected behavior

Running `shopify app dev` should be a simple task, but it fails frequently when using it in.

Something is causing the draft previews to update, and the error below appears in red text.

### Actual behavior

After some time, and usually quitting and re-running `shopify app dev` the error message does not appear.

When this message appears it means draft versions of app extensions, checkout extensions, theme extensions fail to update as well. It's a huge pain because it means extra steps to continuing to work. It breaks the flow when suddenly the extension is not responding to code changes (did I introduce a bug?, did I break something?) 

### Verbose output

```shell
Failed to fetch the latest status of the development store preview, trying again in 5 seconds.
```

This error appears in RED text in the console.
```


### Reproduction steps

1. run `shopify app dev`
2. sometimes the error happens, which needs a full quit and restart

### Operating System

OSX

latest

### Shell

iterm2

latest

latest

amcaplan:

Fixes #3173 

Pushing Theme App Extension drafts requires unexpired Partners auth. Unfortunately, given the Ruby subprocess system we live with (for now), nothing is refreshing authentication, so sometime within ~2 hours the authentication expires, and the theme app extension silently stops reloading on save.

1. Instead of sending a Partners auth token Node => Ruby by CLI flag, we use the hidden `theme token` command to pass a Partners token and save in the local `Shopify::DB`. That token is utilized by the `extension serve` process because we no longer pass a `token` flag.
2. The token is refreshed once before `theme extension serve` is called, and refreshed again every 3 minutes.
3. If the token fails to refresh, we make 2 more attempts after a 30-second window. Failing that, we kill the `dev` process, because we don't want to pretend things are still working when they aren't.

Note this means that if the internet connection is down for a few minutes, `dev` will crash rather than recovering smoothly when the connection is restored. I'm not sure whether that's a concern or not.

### How to test your changes?

1. Start `dev` on an app with a Theme App Extension
2. Wait several hours
3. Make a change to a file in the theme and confirm the crash in #3173 no longer happens, instead the TAE updates successfully.

### Measuring impact

### Checklist

### DIFF
--- a/packages/app/src/cli/services/dev/processes/theme-app-extension.ts
+++ b/packages/app/src/cli/services/dev/processes/theme-app-extension.ts
@@ -25,17 +25,24 @@ export const runThemeAppExtensionsServer: DevProcessFunction<PreviewThemeAppExte
   {stdout, stderr, abortSignal},
   {adminSession, themeExtensionServerArgs: args, storefrontToken},
 ) => {
-  setInterval(() => {
-    outputDebug('Refreshing theme session token...', stdout)
+  const refreshSequence = (attempt = 0) => {
+    outputDebug(`Refreshing partners token (attempt ${attempt})...`, stdout)
     refreshToken()
       .then(() => {
-        outputDebug('Refreshed theme session token successfully', stdout)
+        outputDebug('Refreshed partners token successfully', stdout)
       })
       .catch((error) => {
-        outputDebug(`Failed to refresh theme session token: ${error}`, stderr)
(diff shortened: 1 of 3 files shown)


karreiro:
**Update**: I left the development server running for some hours, and it really refreshes the token.

PR #3388, opened 2024-02-07
Show env vars in help

amcaplan:

Currently, if users want to specify an env variable, they have to look into our CLI code to know what they can specify. It's not a secret, let's make this easier for them!

Adds an `(Env: SOME_VARIABLE)` after each flag, so users know how they can specify flags via ENV vars.

See [here](https://github.com/Shopify/cli/blob/c05128e857df5c37b869093f8b7a168783abef5b/packages/cli/README.md) for how it will look.

Credit to @Arkham for the suggestion of how to present this information.

### How to test your changes?

Run `help` on a command with flags.

### Post-release steps

To be determined: Should we document this on shopify.dev as well?

### Measuring impact

- [x] Not sure there's a way to know, it'll help us to communicate more easily with users though

### Checklist

### DIFF
--- a/packages/cli/src/cli/utilities/help.ts
+++ b/packages/cli/src/cli/utilities/help.ts
@@ -1,0 +1,33 @@
+import {Command, Help, CommandHelp} from '@oclif/core'
+import chalk from 'chalk'
+
+const {dim} = chalk
+
+class CustomCommandHelp extends CommandHelp {
+  protected override flags(flags: Command.Flag.Any[]): [string, string | undefined][] | undefined {
+    if (flags.length === 0) return
+
+    return flags.map((flag) => {
+      const left = this.flagHelpLabel(flag)
+
+      let right = flag.summary || flag.description || ''
+      if (flag.type === 'option' && flag.default) {
+        right = `[default: ${flag.default}] ${right}`
+      }
+
+      if (flag.required) right = `(required) ${right}`
+
+      if (flag.type === 'option' && flag.options && !flag.helpValue && !this.opts.showFlagOptionsInTitle) {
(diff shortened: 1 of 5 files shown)


packages/cli/src/cli/utilities/help.ts
```diff
@@ -0,0 +1,33 @@
+import {Command, Help, CommandHelp} from '@oclif/core'
+import chalk from 'chalk'
+
+const {dim} = chalk
+
+class CustomCommandHelp extends CommandHelp {
+  protected override flags(flags: Command.Flag.Any[]): [string, string | undefined][] | undefined {
```
gracejychang, line 7: how come the typing here is `[string, string | undefined]` instead of just `[string | undefined]`?

packages/cli/src/cli/utilities/help.ts
```diff
@@ -0,0 +1,33 @@
...
+    if (flags.length === 0) return
+
+    return flags.map((flag) => {
+      const left = this.flagHelpLabel(flag)
+
+      let right = flag.summary || flag.description || ''
```
gracejychang, line 13: nit: the naming's a bit short and it's a bit hard to tell at first glance what this is referring to, maybe `left_column` and `right_column` could help make things more clear?

gracejychang (commented):
Thanks for putting this in Ariel! 😄  I won't approve as of now since it seems like there is some discussion of changing the content and/or formatting of the file in this [slack thread](https://shopify.slack.com/archives/C06D4Q1P99C/p1707390042132309). Feel free to ping me again for a re-review if those changes end up getting implemented though!

PR #3389, opened 2024-02-07
Remove support for extension specs plugins

isaacroldan:
- We only used this once or twice at the beginning and never since
- It forces us to pass the command config all the way down
- With the simplification efforts in the specifications this won't be necessary in the near future
- The plugins have been out-of-date for a long time.

Fixes https://github.com/Shopify/develop-app-management/issues/1606

- Remove the ability to load extension specifications from plugins

### How to test your changes?
Nothing specific, this doesn't affect any normal workflow, just that everything works.

### Measuring impact

### Checklist


packages/app/src/cli/models/extensions/load-specifications.ts
```diff
@@ -24,15 +22,6 @@ const SORTED_CONFIGURATION_SPEC_IDENTIFIERS = [
   AppHomeSpecIdentifier,
 ]
 
-/**
- * Load all specifications from the local file system AND plugins
- */
-export async function loadLocalExtensionsSpecifications(config: Config): Promise<ExtensionSpecification[]> {
```
alvaro-shopify, line 30: why not keeping this method and removing the call to `loadUIExtensionSpecificationsFromPlugins`? It make sense to me calling the `loadLocalExtensionsSpecifications` method. You avoid lot of changes in the PR :)

isaacroldan: that makes senes 👌

PR #3392, opened 2024-02-07
Add support for compliance_topics in the app TOML

gonzaloriestra:

Fixes https://github.com/Shopify/develop-app-management/issues/1600

- It allows to deploy privacy compliance webhooks through `[[webhooks.subscriptions]]` with `compliance_topics`
- It keeps working with the legacy format, by defining them under `[webhooks.privacy_compliance]`
- It raises an error if both ways are used together
- It allows defining subscriptions with `compliance_topics`, but without `topics`. It raises if none is present.
- When running link, it rewrites the TOML with the new format and simplifies it by grouping the compliance topics by URI. Only when the `declarative_webhooks` beta is enabled (code taken from https://github.com/Shopify/cli/pull/3416)

### How to test your changes?

**With `declarative_webhooks` flag:**
- Deploy with legacy format >> it gets updated in the dashboard
```
[webhooks.privacy_compliance]
customer_deletion_url = "https://app.com/customer_deletion"
customer_data_request_url = "https://app.com/customer_request"
shop_deletion_url = "https://app.com/shop_deletion"
```
- Link updates the TOML to the new format
- Update all the URIs to the same one with the new format >> it gets updated in the dashboard
- Link simplifies the TOML, merging both topics and compliance_topics with the same URI
- Deploy raises an error when both formats are present:
```
  [webhooks.privacy_compliance]
  customer_data_request_url = "https://app.com/webhook"

  [[webhooks.subscriptions]]
  uri = "https://app.com/webhook"
  compliance_topics = [ "customers/data_request" ]
```

- Deploy raises an error if topics and compliance_topics are missing:
```
  [[webhooks.subscriptions]]
  uri = "https://app.com/webhook"
```
- Deploy raises an error if there are duplicated compliance_topics:
```
  [[webhooks.subscriptions]]
  uri = "https://app.com/webhook"
  complicante_topics = [ "customers/data_request" ]
  
  [[webhooks.subscriptions]]
  uri = "https://app.com/other"
  complicante_topics = [ "customers/data_request" ]  
```
**Without `declarative_webhooks` flag:**
- Deploy with legacy format >> it gets updated in the dashboard
```
[webhooks.privacy_compliance]
customer_deletion_url = "https://app.com/customer_deletion"
customer_data_request_url = "https://app.com/customer_request"
shop_deletion_url = "https://app.com/shop_deletion"
```
- Update the URLs in the dashboard
- Link updates the TOML keeping the format

### Measuring impact

### Checklist


packages/app/src/cli/models/extensions/specifications/app_config_privacy_compliance_webhooks.test.ts
```diff
@@ -27,7 +28,67 @@ describe('privacy_compliance_webhooks', () => {
...
+
+      // Then
+      expect(result).toMatchObject({
+        customers_redact_url: 'https://example.com/customers_webhooks',
+        customers_data_request_url: 'https://example.com/customers_webhooks',
+        shop_redact_url: 'https://example.com/old',
+      })
+    })
+
+    test('should return an emptpy object if all porperties are empty', () => {
```
gracejychang, line 91:
```suggestion
    test('should return an empty object if all porperties are empty', () => {
```

packages/app/src/cli/models/extensions/specifications/app_config_privacy_compliance_webhooks.test.ts
```diff
@@ -49,31 +110,91 @@ describe('privacy_compliance_webhooks', () => {
...
+              compliance_topics: ['customers/redact'],
+              uri: 'https://example.com/customer-deletion',
+            },
+            {
+              compliance_topics: ['customers/data_request'],
+              uri: 'https://example.com/customer-data-request',
+            },
+            {
+              compliance_topics: ['shop/redact'],
+              uri: 'https://example.com/shop-deletion',
```
gracejychang, line 140: will the uri's here change to something like `uri: '/shop-deletion'` with [Alex's PR where relative paths are supported](https://github.com/Shopify/cli/pull/3402), or compliance webhooks should be unaffected here?

packages/app/src/cli/services/dev/select-app.test.ts
```diff
@@ -66,7 +66,7 @@ describe('createApp', () => {
     vi.mocked(appNamePrompt).mockResolvedValue('app-name')
     vi.mocked(partnersRequest).mockResolvedValueOnce({appCreate: {app: APP1, userErrors: []}})
     const variables = {
-      org: 2,
+      org: 1,
```
gracejychang, line 69: just for my own learnings, why are we changing the org here?

gracejychang (commented):
the code looks mostly good to me! 😄  Just some small comments here and there. Another note is that I think we should maybe start keeping track of all the places we are changing due to beta flags, as I assume we want to rollback a number of these changes once we `declarative-webhooks` beta is rolled out 100%?

gonzaloriestra:
Acceptance tests will fail until https://github.com/Shopify/partners/pull/53245 is merged

packages/app/src/cli/models/extensions/specification.ts
```diff
@@ -22,8 +22,8 @@ export interface TransformationConfig {
 }
 
 export interface CustomTransformationConfig {
-  forward?: (obj: object) => object
-  reverse?: (obj: object) => object
+  forward?: (obj: object, options?: {flags?: Flag[]}) => object
```
alvaro-shopify, line 25: Currently `transform` doesnt support the `options` params so maybe we don't need to add this param to the `forward` method

packages/app/src/cli/services/dev/select-app.test.ts
alvaro-shopify: The test has been working by chance 😓 . 
- [APP1](https://github.com/Shopify/cli/blob/main/packages/app/src/cli/services/dev/select-app.test.ts#L43) organization is set to `1`
- Variables uses `org: 2`
- This was a piece of code previous to my change where the `APP1` is modified using the `Variables` values (mutations, oh mutations 😅 )
```ts
  const createdApp: OrganizationApp = result.appCreate.app
  createdApp.organizationId = org.id
  createdApp.newApp = true
  return createdApp 
```
- The new code is not using mutation but a new copy of `APP1` with the modified values (the org is not mutable)
```
  const betas = filterDisabledBetas(result.appCreate.app.disabledBetas)
  return {...result.appCreate.app, organizationId: org.id, newApp: true, betas}
```

packages/app/src/cli/models/extensions/specification.ts
gonzaloriestra: Do you mean that we are not using the `options` param in the current forward implementations? I thought about that, but I think it's better to keep the same shape for both methods. WDYT?

packages/app/src/cli/models/extensions/specifications/app_config_privacy_compliance_webhooks.ts
```diff
@@ -21,35 +23,73 @@ export default spec
...
 }
 
-function transformFromPrivacyComplianceWebhooksModule(content: object) {
+function transformFromPrivacyComplianceWebhooksModule(content: object, options?: {betas?: BetaFlag[]}) {
   const customersRedactUrl = getPathValue(content, 'customers_redact_url') as string
   const customersDataRequestUrl = getPathValue(content, 'customers_data_request_url') as string
   const shopRedactUrl = getPathValue(content, 'shop_redact_url') as string
 
-  if (customersRedactUrl?.length > 0 || customersDataRequestUrl?.length > 0 || shopRedactUrl?.length > 0) {
+  if (options?.betas?.includes(BetaFlag.DeclarativeWebhooks)) {
```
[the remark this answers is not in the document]
gonzaloriestra: Fixed in https://github.com/Shopify/cli/pull/3392/commits/1ab13239ef5dcd1aa59f0a24e346cddb9794fcc4

packages/app/src/cli/services/app/config/link.test.ts
```diff
@@ -560,6 +560,80 @@ embedded = false
     })
   })
 
+  test('fetches the privacy compliance webhooks from the configuration module', async () => {
```
gracejychang, line 563: @alvaro-shopify 👋  I'm taking over Gonzalo's PR (mostly just taking care of merge conflicts), I noticed that you completely deleted [a similar test from this file in this PR on line ~538](https://github.com/Shopify/cli/commit/ea4983751517bc7f5490787afa1dfe46d6a09f98#diff-09c68a3f5fbcd7cb3ddbe117fc5c4876b7453f2632ac812690640f774b4ac85a) - what are your thoughts on keeping this test?

packages/app/src/cli/services/app/config/link.test.ts
```diff
@@ -560,6 +560,80 @@ embedded = false
...
+    await inTemporaryDirectory(async (tmp) => {
+      // Given
+      const options: LinkOptions = {
+        directory: tmp,
+        developerPlatformClient,
+      }
+
+      vi.mocked(loadApp).mockRejectedValue('App not found')
+      vi.mocked(fetchOrCreateOrganizationApp).mockResolvedValue(mockRemoteApp())
```
gracejychang, line 572: had to remove the prev `vi.mocked(fetchAppExtensionRegistrations).mockResolvedValue(...)` call bc it keeps saying the function doesn't exist :/ I think it's due to some changes in the `link.ts` file

packages/app/src/cli/models/extensions/specifications/validation/app_config_webhook.ts
```diff
@@ -13,14 +14,41 @@ export function webhookValidator(schema: object, ctx: zod.RefinementCtx) {
...
+      code: zod.ZodIssueCode.custom,
+      message: `The privacy_compliance section can't be used if there are subscriptions including compliance_topics`,
+    }
+  }
+
+  const ComplianceTopics = subscriptions.flatMap((subscription) => subscription.compliance_topics).filter(Boolean)
+  if (uniq(ComplianceTopics).length !== ComplianceTopics.length) {
+    return {
+      code: zod.ZodIssueCode.custom,
+      message: 'You can’t have duplicate subscriptions with the same compliance topic',
```
alexanderMontague, line 34:
```suggestion
      message: 'You can’t have multiple subscriptions with the same compliance topic',
```

but I'd run this by UX

packages/app/src/cli/models/extensions/specifications/validation/app_config_webhook.ts
```diff
@@ -13,14 +14,41 @@ export function webhookValidator(schema: object, ctx: zod.RefinementCtx) {
...
+    webhookConfig.privacy_compliance &&
+    webhookConfig.subscriptions?.some((subscription) => subscription.compliance_topics)
+  ) {
+    return {
+      code: zod.ZodIssueCode.custom,
+      message: `The privacy_compliance section can't be used if there are subscriptions including compliance_topics`,
+    }
+  }
+
+  const ComplianceTopics = subscriptions.flatMap((subscription) => subscription.compliance_topics).filter(Boolean)
```
alexanderMontague, line 30:
```suggestion
  const complianceTopics = subscriptions.flatMap((subscription) => subscription.compliance_topics).filter(Boolean)
```

alexanderMontague (approved):
This looks good, we should try to get this merged! Will need a rebase, tophat looks great too but lmk if you want to go over this again once cleaned up.

packages/app/src/cli/models/extensions/specifications/app_config_privacy_compliance_webhooks.test.ts
```diff
@@ -27,7 +28,37 @@ describe('privacy_compliance_webhooks', () => {
...
+
+      // Then
+      expect(result).toMatchObject({
+        customers_redact_url: 'https://example.com/customers_webhooks',
+        customers_data_request_url: 'https://example.com/customers_webhooks',
+        shop_redact_url: 'https://example.com/shop_webhooks',
+      })
+    })
+
+    test('should return an empty object if all porperties are empty', () => {
```
alexanderMontague, line 61:
```suggestion
    test('should return an empty object if all properties are empty', () => {
```

packages/app/src/cli/models/extensions/specifications/app_config_privacy_compliance_webhooks.ts
```diff
@@ -1,11 +1,13 @@
-import {WebhookSchema} from './app_config_webhook.js'
-import {WebhooksConfig} from './types/app_config_webhook.js'
+import {ComplianceTopic, WebhookSchema} from './app_config_webhook.js'
+import {WebhookSubscription, WebhooksConfig} from './types/app_config_webhook.js'
 import {CustomTransformationConfig, createConfigExtensionSpecification} from '../specification.js'
-import {getPathValue} from '@shopify/cli-kit/common/object'
+import {Flag} from '../../../services/dev/fetch.js'
+import {compact, getPathValue} from '@shopify/cli-kit/common/object'
 
 const PrivacyComplianceWebbhooksTransformConfig: CustomTransformationConfig = {
```
alexanderMontague, line 7:
```suggestion
const PrivacyComplianceWebhooksTransformConfig: CustomTransformationConfig = {
```

packages/app/src/cli/models/extensions/specifications/app_config_privacy_compliance_webhooks.ts
```diff
@@ -1,11 +1,13 @@
...
-  forward: (content: object) => transformToPrivacyComplianceWebhooksModule(content),
-  reverse: (content: object) => transformFromPrivacyComplianceWebhooksModule(content),
+  forward: (content: object, options?: {flags?: Flag[]}) => transformToPrivacyComplianceWebhooksModule(content),
+  reverse: (content: object, options?: {flags?: Flag[]}) =>
+    transformFromPrivacyComplianceWebhooksModule(content, options),
 }
 
 export const PrivacyComplianceWebbhooksSpecIdentifier = 'privacy_compliance_webhooks'
```
alexanderMontague, line 13:
```suggestion
export const PrivacyComplianceWebhooksSpecIdentifier = 'privacy_compliance_webhooks'
```

packages/app/src/cli/models/extensions/specifications/app_config_privacy_compliance_webhooks.ts
```diff
@@ -21,35 +23,61 @@ export default appPrivacyComplienceSpec
```
alexanderMontague, line 23: Right above this (can't comment on the exact line) 

```diff
- const appPrivacyComplienceSpec = createConfigExtensionSpecification({
+ const appPrivacyComplianceSpec = createConfigExtensionSpecification({
- identifier: PrivacyComplianceWebbhooksSpecIdentifier,
+ identifier: PrivacyComplianceWebhooksSpecIdentifier,
  schema: WebhookSchema,
- transformConfig: PrivacyComplianceWebbhooksTransformConfig,
+ transformConfig: PrivacyComplianceWebhooksTransformConfig,
})
// ...
- export default appPrivacyComplienceSpec
+ export default appPrivacyComplianceSpec

}
```

packages/app/src/cli/models/app/app.ts
```diff
@@ -322,7 +322,7 @@ function findExtensionByHandle(allExtensions: ExtensionInstance[], handle: strin
 }
```
gracejychang, line 323: I removed the previous `webhooksConfiguration(configuration: object) ` and ` simplifySubscriptions(subscriptions: WebhookSubscription[] | undefined)` functions b/c of this PR which is related to the `link` command - @alvaro-shopify does this seem correct?

gracejychang: For context, here's what the old code block looked like:

```

function webhooksConfiguration(configuration: object) {
  const webhooks = getPathValue<WebhooksConfig>(configuration, 'webhooks') as WebhooksConfig
  if (webhooks?.subscriptions) {
    const simplifiedSubscritions = simplifySubscriptions(webhooks.subscriptions)
    setPathValue(webhooks, 'subscriptions', simplifiedSubscritions)
  }
  return {webhooks}
}

function simplifySubscriptions(subscriptions: WebhookSubscription[] | undefined): WebhookSubscription[] {
  if (!subscriptions) return []

  return subscriptions.reduce((accumulator, subscription) => {
    const existingSubscription = accumulator.find(
      (sub) =>
        sub.uri === subscription.uri &&
        sub.sub_topic === subscription.sub_topic &&
        sub.include_fields === subscription.include_fields &&
        sub.metafield_namespaces === subscription.metafield_namespaces,
    )
    if (existingSubscription) {
      if (subscription.compliance_topics) {
        existingSubscription.compliance_topics ??= []
        existingSubscription.compliance_topics.push(...subscription.compliance_topics)
      }
      if (subscription.topics) {
        existingSubscription.topics ??= []
        existingSubscription.topics.push(...subscription.topics)
      }
    } else {
      accumulator.push(subscription)
    }
    return accumulator
  }, [] as WebhookSubscription[])
}
```

But all the different configuration paths has now been removed (including the `webhooksConfiguration`) so I am assuming we don't need this any more?

packages/app/src/cli/models/extensions/specifications/validation/app_config_webhook.ts
gracejychang: Ah yes agreed - I was wondering if we had made a decision I missed previously since we'd been using "duplicate" before, sorry I didn't make that clear! Deffo will await Jen's opinion here :)

jenstanicak: I think either works! I'm cool with multiple in this example, and duplicate in the other, we can have some variation in how we explain it because we're crystal clear about the actual issue (same compliance topic)

packages/app/src/cli/models/app/app.ts
gonzaloriestra: Right, that simplify logic is still required, but we have to find a new place. The purpose is that, when you have a topic and a compliance topic sharing the same URL, both should be merged in the TOML under the same subscription:
```
  [[webhooks.subscriptions]]
  topics = ["products/update"]
  compliance_topics = [ "customers/data_request" ]
``` 
Álvaro suggested a couple of alternatives:
- Restore the removed logic, but fixing the issue that [this PR](https://github.com/Shopify/cli/pull/3540) solved in a different way
- Add a `simplify` function to the specs and call each one after building the full object

I think the second way can be simpler.

packages/app/src/cli/models/extensions/specification.ts
```diff
@@ -144,10 +152,19 @@ export function createConfigExtensionSpecification<TConfiguration extends BaseCo
     appModuleFeatures,
     transform: resolveAppConfigTransform(spec.transformConfig),
     reverseTransform: resolveReverseAppConfigTransform(spec.schema, spec.transformConfig),
+    simplify: resolveSimplifyAppConfig(spec.simplify),
     experience: 'configuration',
   })
 }
 
+function resolveSimplifyAppConfig(simplifyConfig?: SimplifyConfig) {
```
gracejychang, line 160: adds new simplify method

gonzaloriestra: It looks nice!

Just one doubt: if I understand it correctly, we are iterating through the config modules, merging them to the final config and applying the simplification in that moment. And the purpose is to merge two different modules (webhooks and compliance webhooks). But what happens if the simplify method, which belongs to the webhooks module, is called before the compliance webhooks module is processed? I'm afraid it could be working by chance.

The easiest solution I can think of would be:
- Process all the modules as before, merging them in a single config
- Iterate a second time through all the modules and call the simplify method if present. That way, we ensure the config already contains all the modules.

Does it make sense?

packages/app/src/cli/models/extensions/specification.ts
```diff
@@ -144,10 +152,19 @@ export function createConfigExtensionSpecification<TConfiguration extends BaseCo
...
+  if (simplifyConfig) {
+    return simplifyConfig.simplify
+  }
+  // default, returns the configuration if there is no simplify function
+  return (content: SpecsAppConfiguration) => content
+}
```
alvaro-shopify, line 166:
```suggestion
  // Returns the configuration if there is no simplify function
  return simplifyConfig?.simplify ?? ((content: SpecsAppConfiguration) => content)
}
```

packages/app/src/cli/models/extensions/specification.ts
gracejychang: Ah yes, I think Alvaro's comment below will solve for this - I think I misunderstood where to put this code when it was mentioned that we should iterate over all the specifications! Thank you for the clarification 😄

alvaro-shopify (approved):
LGTM, amazing jog with this 🎉 !

gonzaloriestra (commented):
I've been testing it and everything works as expected 👌 

@gracejychang thanks for taking care of it, great job!

PR #3402, opened 2024-02-08
support relative paths to `application_url` for declarative webhooks

alexanderMontague:

Closes https://github.com/Shopify/develop-app-management/issues/1599

- adds support for declarative webhook `uri`s that are relative to the `application_url`
ex:

```toml
application_url = "https://my-app-url.com"

[[webhooks.subscriptions]]
topics = ["products/create", "products/update"]
uri = "/products" # https://my-app-url.com/products
```

- this is especially useful when using `app dev` as the app url changes every time

- this also transforms the config on `app config link`, and we simplify like we do with the `topics` array
ex. if you have 

```toml
application_url = "https://my-app-url.com"

[[webhooks.subscriptions]]
topics = ["products/create"]
uri = "https://my-app-url.com/products"

[[webhooks.subscriptions]]
topics = ["orders/create"]
uri = "https://my-app-url.com/orders"
```

and link, we will simplify the config to

```toml
application_url = "https://my-app-url.com"

[[webhooks.subscriptions]]
topics = ["products/create"]
uri = "/products"

[[webhooks.subscriptions]]
topics = ["orders/create"]
uri = "/orders"
```

<hr />

- There is also nothing stopping anyone from creating a bad relative path like
```toml
application_url = "https://my-app-url.com"

[[webhooks.subscriptions]]
topics = ["products/create"]
uri = "//////products" # https://my-app-url.com//////products
```

- As technically to https spec, this is still a valid URL. We could parse out extra forward slashes, but I think a developer would realize this once they see it in their TOML or on the dashboard
- The `application_url` cannot end in a forward slash so this protects against this on the app url side

### How to test your changes?

- create an app and opt into the versioned config and declarative webhooks betas (both app and dev shop)
- create some webhook subscriptions with a relative path
- deploy the version
- go to the dashboard, you should see the full URL
- you can also link and make sure the toml does not change (unless it is this simplification case like mentioned)

### Checklist

### DIFF
--- a/packages/app/src/cli/models/app/loader.test.ts
+++ b/packages/app/src/cli/models/app/loader.test.ts
@@ -2350,9 +2350,8 @@ describe('WebhooksSchema', () => {
       ],
     }
     const errorObj = {
-      validation: 'regex' as zod.ZodInvalidStringIssue['validation'],
-      code: zod.ZodIssueCode.invalid_string,
-      message: 'Invalid',
+      code: zod.ZodIssueCode.custom,
+      message: 'Must be a valid https, pubsub, or ARN URI',
       path: ['webhooks', 'subscriptions', 0, 'uri'],
     }
 
@@ -2378,9 +2377,8 @@ describe('WebhooksSchema', () => {
       subscriptions: [{uri: 'my::URI-thing::Shopify::123', topics: ['products/create']}],
     }
     const errorObj = {
-      validation: 'regex' as zod.ZodInvalidStringIssue['validation'],
-      code: zod.ZodIssueCode.invalid_string,
-      message: 'Invalid',
+      code: zod.ZodIssueCode.custom,
+      message: 'Must be a valid https, pubsub, or ARN URI',
       path: ['webhooks', 'subscriptions', 0, 'uri'],
     }
 
@@ -2426,6 +2424,22 @@ describe('WebhooksSchema', () => {
     expect(parsedConfiguration.webhooks).toMatchObject(webhookConfig)
   })
 
+  test('accepts a relative path uri', async () => {
+    const webhookConfig: WebhooksConfig = {
+      api_version: '2021-07',
+      subscriptions: [
+        {
+          uri: '/webhooks',
+          topics: ['products/create'],
+        },
+      ],
+    }
+
+    const {abortOrReport, parsedConfiguration} = await setupParsing({}, webhookConfig)
+    expect(abortOrReport).not.toHaveBeenCalled()
+    expect(parsedConfiguration.webhooks).toMatchObject(webhookConfig)
+  })
+
   test('accepts combination of uris', async () => {
     const webhookConfig: WebhooksConfig = {
       api_version: '2021-07',
@@ -2440,6 +2454,10 @@ describe('WebhooksSchema', () => {
         },
         {
           uri: 'pubsub://my-project-123:my-topic',
+          topics: ['products/create', 'products/update'],
+        },
+        {
+          uri: '/webhooks',
           topics: ['products/create', 'products/update'],
         },
       ],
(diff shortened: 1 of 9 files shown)


packages/app/src/cli/models/app/loader.ts
```diff
@@ -219,6 +220,7 @@ class AppLoader {
       specifications: this.specifications,
     })
     const {directory, configuration, configurationLoadResultMetadata, configSchema} = await configurationLoader.loaded()
+    this.fullAppConfiguration = configuration
```
alexanderMontague, line 223: This is the first time (I think) we need data from other modules in order to create another. I decided to set the full config to a property to limit the argument drilling. Is there a better way of doing this?

gonzaloriestra: The codebase is a bit mess... I can't think of a better approach without a big refactor.

The problem is that I'll need to add another parameter to the transform functions (the beta flags) in https://github.com/Shopify/cli/pull/3392 🙈

gonzaloriestra (approved):
The code LGTM, but I'm not sure about the approach as I commented [here](https://shopify.slack.com/archives/C05LVBGTYTC/p1707480040307539?thread_ts=1707411561.880419&cid=C05LVBGTYTC).

packages/app/src/cli/models/extensions/specifications/app_config_webhook.test.ts
```diff
@@ -246,5 +279,62 @@ describe('webhooks', () => {
...
+          {
+            topic: 'customers/delete',
+            uri: 'pubsub://absolute-feat-test:pub-sub-topic',
+          },
+        ],
+      }
+      const webhookSpec = spec
+
+      // When
+      const result = webhookSpec.reverseTransform!(object, {application_url: 'https://my-app-url.com'})
```
gracejychang, line 312: what happens when if the `application_url` is invalid?

gracejychang (commented):
some minor comments, but otherwise everything looks good to me!

packages/app/src/cli/models/extensions/specifications/app_config_webhook.test.ts
gracejychang: I'm not sure if this should be tested in `loader_test` or this test file

packages/app/src/cli/models/app/loader.test.ts
```diff
@@ -2263,27 +2263,6 @@ describe('parseConfigurationObject', () => {
 })
 
 describe('WebhooksSchema', () => {
-  test('throws an error if uri is not an https uri', async () => {
```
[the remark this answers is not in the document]
alexanderMontague: Its included in the simplified tests [below](https://github.com/Shopify/cli/pull/3402/files#diff-9d6527038485e8f2adcf560173fb65c807f15d4ac708394c36e1184b3a866390L2299) 😃

packages/app/src/cli/models/extensions/specifications/app_config_webhook.test.ts
alexanderMontague: Why should this be tested in loader? The essence of those tests are more integration tests. These are simply unit tests, hence why we directly call `webhookSpec.reverseTransform`. We don't really care about anything that happens above this point in the code, we can assume our other test coverage handles that (which it does as you caught in your previous comment!). We only care about testing the logic contained in this transform function.

gracejychang: I suggested `loader_test` because it's where I saw we were testing for errors after a deploy occurs on the CLI (so yes integration tests).  The tests you linked to there was testing specifically for `uri` field, but not for errors which might be thrown if the `application_url` field is invalid (ie. for example, developer starts out the TOML with relative path and `application_url`, or edits the `application_url` after the TOML is transformed)?

alexanderMontague:
wip

alexanderMontague:
wip

alexanderMontague:
https://github.com/Shopify/cli/pull/3948

Issue #3404, opened 2024-02-08
Bump @shopify/theme-check-node package to recently released 2.0.1


patrickDouglas:
Bump [@shopify/theme-check-node](https://github.com/Shopify/cli/blob/main/packages/app/package.json#L48) to the [latest version 2.0.1](https://github.com/Shopify/theme-tools/releases/tag/%40shopify%2Ftheme-check-node%402.0.1).

PR #3405, opened 2024-02-08
Bump Shopify/theme-tools packages

charlespwd:
Fixes https://github.com/Shopify/cli/issues/3404


patrickDouglas:
I think it's due to the fact that there hasn't been a new release of `@shopify/cli-kit` since that PR got merged in yesterday, AND `shopify/app` will also need to use that (currently) unreleased version: https://github.com/Shopify/cli/blob/main/packages/app/package.json#L46

isaacroldan (approved):

PR #3422, opened 2024-02-12
[Feature] Pass input through liquid renderer before parsing as TOML

jduff:

There is currently no way to template toml file if you want to supply some of the values via environment variables. I believe the recommended approach to handling multiple environments is to have multiple named configuration files. While this works, I still think it is risky to include the working production config in a repo and would rather supply the production key at deploy time in an isolated environment.

Since this project already uses Liquid I opted to use that as the templating language that can be used within the toml files. This PR passes all toml files through a Liquid renderer providing the environment variables as data to the rendered before parsing the toml file.

### How to test your changes?

I have not yet added tests to this PR, but I can if this is a feature the team is willing to accept. To test this you can use this PR and run any command that reads a toml file and verify the resulting rendered config is used.

Given shopify.app.toml:
``` toml
client_id = "{{ env.SHOPIFY_APP_API_KEY }}"
name = "{{ env.SHOPIFY_APP_NAME }}"
# the rest of the file
```

``` bash
SHOPIFY_APP_API_KEY=key SHOPIFY_APP_NAME=test npm run deploy
```

I also set this up using https://github.com/ds300/patch-package so you could use that along with [this patch file](https://gist.github.com/jduff/a88fc49b76c42cfc2332ad1f6468ee62) to test these changes locally by adding it to your `patches` folder.

### Checklist

### DIFF
--- a/packages/cli-kit/src/public/node/toml.ts
+++ b/packages/cli-kit/src/public/node/toml.ts
@@ -1,17 +1,33 @@
 import {JsonMap} from '../../private/common/json.js'
 import * as toml from '@iarna/toml'
+import {Liquid} from 'liquidjs';
+import {readAndParseDotEnv } from './dot-env.js'
+import {fileExists} from './fs.js'
+import {joinPath, cwd} from './path.js'
 
 export type JsonMapType = JsonMap
 
 /**
  * Given a TOML string, it returns a JSON object.
+ * Before parsing as TOML the string is passed through a
+ * Liquid renderer with access to environment variables.
  *
  * @param input - TOML string.
  * @returns JSON object.
  */
-export function decodeToml(input: string): object {
+export async function decodeToml(input: string): object {
+  const engine = new Liquid()
   const normalizedInput = input.replace(/\r\n$/g, '\n')
-  return toml.parse(normalizedInput)
+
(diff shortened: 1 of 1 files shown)


packages/cli-kit/src/public/node/toml.ts
```diff
@@ -1,17 +1,34 @@
...
  * @returns JSON object.
  */
-export function decodeToml(input: string): object {
+export async function decodeToml(input: string): object {
+  const engine = new Liquid()
   const normalizedInput = input.replace(/\r\n$/g, '\n')
   return toml.parse(normalizedInput)
+
+  let envData = process.env;
+  const envFile = joinPath(cwd(), '.env');
```
Kashuab, line 24: There are conventions for multiple `.env` files, see relevant information in Vite docs: https://vitejs.dev/guide/env-and-mode#env-files

Kashuab:
I wonder if it would be beneficial to require certain environment variables to be scoped, kind of like how Vite requires `VITE_*`. Some teams may have sensitive secrets in their `.env` that they wouldn't want exposed to other environments.

jduff:
I think this is still relevant, but it needs feedback from the team to know if it might be accepted before tests are added.

PR #3436, opened 2024-02-14
Migrate app info to partners adapter

amcaplan:

Includes the skeleton work on Partners adapter (branch `partners-adapter`, as well as migrating `shopify app info` as the first command to fully use the new adapter.

Closes https://github.com/Shopify/develop-app-management/issues/1615

1. Includes the skeleton `DeveloperPlatformClient` adapter with just a couple of methods, and handling implications on testing etc.
2. Removes direct Partners references from `shopify app info` and its dependencies. This amounts to a fair bit of code, especially when including tests.

### How to test your changes?

Make sure that `shopify app info` works as before. For good measure, we should try a few critical paths (`dev`, `deploy`, etc. especially with `--reset`) after each of these PRs.

### Measuring impact

### Checklist

### DIFF
--- a/packages/app/src/cli/utilities/developer-platform-client.ts
+++ b/packages/app/src/cli/utilities/developer-platform-client.ts
@@ -1,0 +1,12 @@
+import {PartnersSession} from '../../cli/services/context/partner-account-info.js'
+import {PartnersClient} from './developer-platform-client/partners-client.js'
+import {OrganizationApp} from '../models/organization.js'
+
+export function selectDeveloperPlatformClient(): DeveloperPlatformClient {
+  return new PartnersClient()
+}
+
+export interface DeveloperPlatformClient {
+  session: () => Promise<PartnersSession>
+  appFromId: (appId: string) => Promise<OrganizationApp>
+}
--- a/packages/app/src/cli/utilities/developer-platform-client.ts
+++ b/packages/app/src/cli/utilities/developer-platform-client.ts
@@ -1,6 +1,10 @@
 import {PartnersClient} from './developer-platform-client/partners-client.js'
 import {PartnersSession} from '../../cli/services/context/partner-account-info.js'
-import {OrganizationApp} from '../models/organization.js'
+import {MinimalOrganizationApp, Organization, OrganizationApp} from '../models/organization.js'
+
+export type Paginateable<T> = T & {
+  hasMorePages: boolean
+}
 
 export function selectDeveloperPlatformClient(): DeveloperPlatformClient {
   return new PartnersClient()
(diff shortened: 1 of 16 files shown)


packages/app/src/cli/utilities/developer-platform-client.ts
```diff
@@ -0,0 +1,19 @@
...
+export type Paginateable<T> = T & {
+  hasMorePages: boolean
+}
+
+export function selectDeveloperPlatformClient(): DeveloperPlatformClient {
+  return new PartnersClient()
+}
+
+export interface DeveloperPlatformClient {
+  session: () => Promise<PartnersSession>
```
isaacroldan, line 14: Do we want to define a new interface for this? `PartnersSession` is tied to the `PartnersClient` no?

amcaplan: You're correct that this is undesirable. We have left this here for now as an escape route for existing paths, so we can gradually migrate over pieces of the system. Ultimately there should be no session on the public interface.

isaacroldan: That makes sense 👌

packages/app/src/cli/utilities/developer-platform-client.ts
```diff
@@ -0,0 +1,19 @@
+import {PartnersClient} from './developer-platform-client/partners-client.js'
+import {PartnersSession} from '../../cli/services/context/partner-account-info.js'
+import {MinimalOrganizationApp, Organization, OrganizationApp} from '../models/organization.js'
+
+export type Paginateable<T> = T & {
+  hasMorePages: boolean
+}
+
+export function selectDeveloperPlatformClient(): DeveloperPlatformClient {
```
gonzaloriestra, line 9: I think this name may be confusing, as there's no choice to make, and there are many `selectX` functions that usually prompt the user.

Even when we eventually switch here the client dynamically, it will be something internal and automatic. What about something like `initializeDeveloperPlatformClient`?

packages/app/src/cli/utilities/developer-platform-client/partners-client.ts
```diff
@@ -0,0 +1,44 @@
...
+
+const resetHelpMessage = ['You can pass', {command: '--reset'}, 'to your command to reset your app configuration.']
+
+export class PartnersClient implements DeveloperPlatformClient {
+  private _session: PartnersSession | undefined
+
+  async session(): Promise<PartnersSession> {
+    if (!this._session) {
+      this._session = await fetchPartnersSession()
+    }
```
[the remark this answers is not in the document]
gonzaloriestra: It's probably because it's quite new. I think it's ok to have both expressions in the same codebase, but yeah, we can try to change a bunch of them later.

PR #3451, opened 2024-02-19
Migrate dev to Partners adapter, part 1

amcaplan:

Increments on https://github.com/Shopify/develop-app-management/issues/1621 but mostly operates on dependencies that will also be useful for https://github.com/Shopify/develop-app-management/issues/1622

Continues the work of substituting a `DeveloperPlatformClient` object where a direct connection to Partners once was.

### How to test your changes?

Make sure `dev` still works properly

### Measuring impact

### Checklist

### DIFF
--- a/packages/app/src/cli/prompts/dev.ts
+++ b/packages/app/src/cli/prompts/dev.ts
@@ -5,13 +5,16 @@ import {DeveloperPlatformClient, selectDeveloperPlatformClient} from '../utiliti
 import {renderAutocompletePrompt, renderConfirmationPrompt, renderTextPrompt} from '@shopify/cli-kit/node/ui'
 import {outputCompleted} from '@shopify/cli-kit/node/output'
 
-export async function selectOrganizationPrompt(organizations: Organization[]): Promise<Organization> {
+export async function selectOrganizationPrompt(
+  organizations: Organization[],
+  developerPlatformClient = selectDeveloperPlatformClient(),
+): Promise<Organization> {
   if (organizations.length === 1) {
     return organizations[0]!
   }
   const orgList = organizations.map((org) => ({label: org.businessName, value: org.id}))
   const id = await renderAutocompletePrompt({
-    message: 'Which Partners organization is this work for?',
+    message: `Which ${developerPlatformClient.platformTitle} organization is this work for?`,
     choices: orgList,
   })
   return organizations.find((org) => org.id === id)!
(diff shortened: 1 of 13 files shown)


packages/app/src/cli/services/context.test.ts
```diff
@@ -727,7 +729,10 @@ api_version = "2023-04"
     vi.mocked(fetchStoreByDomain).mockResolvedValue({organization: ORG1, store: STORE1})
 
     // When
-    const got = await ensureDevContext(INPUT, DEVELOPER_PLATFORM_CLIENT)
+    const dpc = developerPlatformClient()
```
gonzaloriestra, line 732: I know `developerPlatformClient` is a bit long, but I'd avoid short variable names like this and acronyms.

gonzaloriestra (approved):

packages/app/src/cli/prompts/dev.ts
```diff
@@ -5,13 +5,16 @@ import {DeveloperPlatformClient, selectDeveloperPlatformClient} from '../utiliti
...
+  organizations: Organization[],
+  developerPlatformClient = selectDeveloperPlatformClient(),
+): Promise<Organization> {
   if (organizations.length === 1) {
     return organizations[0]!
   }
   const orgList = organizations.map((org) => ({label: org.businessName, value: org.id}))
   const id = await renderAutocompletePrompt({
-    message: 'Which Partners organization is this work for?',
+    message: `Which ${developerPlatformClient.platformTitle} organization is this work for?`,
```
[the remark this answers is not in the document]
amcaplan: I think we are currently using `Partners` explicitly, in the future we will call it something else. I think `organization` might be OK, though. And it might save some confusion if one day `Partners` becomes `Shopify Developers` or something.

packages/app/src/cli/services/context.test.ts
gonzaloriestra: Sounds good. Another option is to leave the [DEVELOPER_PLATFORM_CLIENT constant](https://github.com/Shopify/cli/pull/3451/files#diff-947ada00886ffeb6daf99bafc0b67815bbca34e92222f899186ea26e35734622L156).

amcaplan: I can't leave it constant because we are spying on one of the functions, which means that using the same object each time will cause tests to fail.

packages/app/src/cli/utilities/developer-platform-client.ts
```diff
@@ -11,9 +11,12 @@ export function selectDeveloperPlatformClient(): DeveloperPlatformClient {
 }
 
 export interface DeveloperPlatformClient {
+  platformTitle: string
```
gonzaloriestra, line 14: Do we need this now?

PR #3458, opened 2024-02-20
Remove zod schemas from app info JSON format

amcaplan:

It's not valuable to the user, and it's 80% of output lines.

On a sample local app, we went from 2717 lines of JSON to 562.

Removes the Zod schemas from app info when you run `--json`

### How to test your changes?

Compare `npm run info --json` with and without this change.

### Measuring impact

### Checklist

### DIFF
--- a/.changeset/strange-drinks-move.md
+++ b/.changeset/strange-drinks-move.md
@@ -1,0 +1,5 @@
+---
+'@shopify/app': minor
+---
+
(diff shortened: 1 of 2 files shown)


gonzaloriestra (approved):
Much better now! But why do we even want to return the specifications? Why showing so much more data when choosing the JSON format? 🤔 

Also, don't we have tests for the JSON output?

PR #3471, opened 2024-02-23
Fix `shopify theme dev` issue to avoid previewing the live theme instead of the development one

karreiro:
### WHY are these changes introduced?

Fixes https://github.com/Shopify/cli/issues/3466

### WHAT is this pull request doing?

It seems like the caching policy at SFR has changed, so we're implementing this temporary solution to clean it and mitigate the issue.

### How to test your changes?

- Delete your current development theme with `shopify-dev theme delete -t <development_theme_id>`
- Run `shopify theme dev`
- Notice that the localhost no longer loads the live theme; and the theme in localhost matches with the preview theme

**Before**

**After**

### Post-release steps

None.

### Measuring impact

How do we know this change was effective? Please choose one:

- [ ] n/a - this doesn't need measurement, e.g. a linting rule or a bug-fix
- [x] Existing analytics will cater for this addition
- [ ] PR includes analytics changes to measure impact

### Checklist

- [x] I've considered possible cross-platform impacts (Mac, Linux, Windows)
- [x] I've considered possible [documentation](https://shopify.dev) changes
- [x] I've made sure that any changes to `dev` or `deploy` have been reflected in the [internal flowchart](https://www.figma.com/file/7vqUp50u6dm48Zfb4JRRn8/CLI3-Internals).

PR #3473, opened 2024-02-23
[3.56] Fix `shopify theme dev` issue to avoid previewing the live theme instead of the development one

karreiro:
Backport https://github.com/Shopify/cli/pull/3471

### DIFF
--- a/packages/cli-kit/assets/cli-ruby/lib/shopify_cli/theme/dev_server/proxy.rb
+++ b/packages/cli-kit/assets/cli-ruby/lib/shopify_cli/theme/dev_server/proxy.rb
@@ -35,10 +35,11 @@ module ShopifyCLI
           wpm
         ]
 
-        def initialize(ctx, theme, param_builder)
+        def initialize(ctx, theme, param_builder, cache_cleaned = false)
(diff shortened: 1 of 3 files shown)


isaacroldan (approved):

packages/cli-kit/assets/cli-ruby/lib/shopify_cli/theme/dev_server/proxy.rb
```diff
@@ -101,6 +105,22 @@ def secure_session_id
...
+        def clean_sfr_cache(env, query, headers)
+          return if @cache_cleaned
+
+          @cache_cleaned = true
+
+          query = query.dup
+          query << ["preview_theme_id", theme_id]
+
+          request(
+            env["REQUEST_METHOD"], env["PATH_INFO"],
```
charlespwd, line 117: I'm not 100% sure how that works. Why would doing the request twice but once with the query param fix the problem? It seems like we're throwing away the result of this request too. Do we need this to clear a remote cache (and not a local one?)

charlespwd (approved):
I believe you that it works, just not sure why :)

PR #3477, opened 2024-02-23
Migrate deploy to partners adapter, part 2

gonzaloriestra:

Part of https://github.com/Shopify/develop-app-management/issues/1622

Continuation of https://github.com/Shopify/cli/pull/3459

Replaces more of the Partners-facing logic with an adapter

### How to test your changes?

Everything about deployment (and related commands) should still work. A lot of commands are affected by these changes.

### Measuring impact

### Checklist


packages/app/src/cli/services/app/config/link.ts
```diff
@@ -60,10 +60,10 @@ export default async function link(options: LinkOptions, shouldRenderSuccess = t
   return configuration
 }
 
-async function selectRemoteApp(options: LinkOptions, developerPlatformClient: DeveloperPlatformClient) {
+async function selectRemoteApp(options: LinkOptions) {
```
amcaplan, line 63: What about something like:

```suggestion
async function selectRemoteApp(options: LinkOptions & Required<Pick<LinkOptions, "developerPlatformClient">>) {
```

Then the `!` below is unnecessary.

Alternatively, we could consider requiring `developerPlatformClient` in `LinkOptions` and fixing up any places where it isn't passed in.

packages/app/src/cli/services/deploy/upload.test.ts
```diff
@@ -83,23 +80,23 @@ describe('uploadWasmBlob', () => {
...
       vi.mocked(fetch).mockImplementation(mockedFetch)
 
       // When
-      await expect(
-        uploadWasmBlob(extension.localIdentifier, extension.outputPath, apiKey, token),
-      ).resolves.toStrictEqual({
+      const result = uploadWasmBlob(extension.localIdentifier, extension.outputPath, developerPlatformClient)
+
+      // Then
+      await expect(result).resolves.toStrictEqual({
```
amcaplan, line 96: Alternatively:

```suggestion
      const result = await uploadWasmBlob(extension.localIdentifier, extension.outputPath, developerPlatformClient)

      // Then
      expect(result).toStrictEqual({
```

packages/cli-kit/src/public/node/api/partners.ts
```diff
@@ -57,12 +57,11 @@ export interface FunctionUploadUrlGenerateResponse {
  * @returns The response of the query.
  */
 export async function getFunctionUploadUrl(token: string): Promise<FunctionUploadUrlGenerateResponse> {
-  const functionUploadUrlGenerateMutation = FunctionUploadUrlGenerateMutation
   const res: FunctionUploadUrlGenerateResponse = await partnersRequest(FunctionUploadUrlGenerateMutation, token)
   return res
 }
 
-const FunctionUploadUrlGenerateMutation = gql`
+export const FunctionUploadUrlGenerateMutation = gql`
```
amcaplan, line 64: Not a comment on this PR, just generally: should this be in cli-kit at all?

amcaplan (approved):
🚀

packages/cli-kit/src/public/node/api/partners.ts
gonzaloriestra: Maybe not... But we are also exposing other Partners functions, so not sure. We can review this later when we finish the refactor.

PR #3492, opened 2024-02-27
[WIP] `app init` prototype 1

isaacroldan:
We want to add an `app init` command to the CLI, this feature is already included in the `@shopify/create-app` package so ideally we want to reuse that.

Import the `init` command from `create-app` and re-export it as a command inside the `app package`

### Measuring impact

### Checklist

### DIFF
--- a/packages/app/oclif.manifest.json
+++ b/packages/app/oclif.manifest.json
@@ -1495,6 +1495,106 @@
       ],
       "strict": true
     },
+    "app:init": {
+      "aliases": [
+        "create-app"
+      ],
+      "args": {
+      },
+      "flags": {
+        "flavor": {
+          "description": "Which flavor of the given template to use.",
+          "env": "SHOPIFY_FLAG_TEMPLATE_FLAVOR",
+          "hasDynamicHelp": false,
+          "multiple": false,
+          "name": "flavor",
+          "type": "option"
+        },
+        "local": {
+          "allowNo": false,
+          "char": "l",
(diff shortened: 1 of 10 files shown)


gonzaloriestra (commented):
Some concerns:
- When you run the new init command from a project folder (the only way right now), it creates a nested folder with the new app, which doesn't make much sense. I can think of several alternatives when we detect that is run inside an app:
  - Ask for the destination folder
  - Show a warning suggesting to use `npm init @shopify/app` from outside 
  - Hide the command unless it's a global installation (not sure if it's possible)
- I've tested a project with `@shopify/cli/0.0.0-experimental-20240228111332` and for some reason `npm run shopify app` is not found, but it should show the help. It doesn't happen from the CLI project with this branch 🤔

isaacroldan:
>I've tested a project with @shopify/cli/0.0.0-experimental-20240228111332 and for some reason npm run shopify app is not found, but it should show the help. It doesn't happen from the CLI project with this branch 🤔

Sorry, that's because that experimental branch has other changes and the `plugin-help` is not included, so the automatic help doesn't work there.

>When you run the new init command from a project folder (the only way right now), it creates a nested folder with the new app, which doesn't make much sense

This behavior is similar to `theme init` and `hydrogen init` (both already exist), you can run them inside a project folder and it will work. Is your responsibility to use the `--path` flag to point somewhere else (or run it globally)

gonzaloriestra (approved):

PR #3494, opened 2024-02-27
[cli-kit] Support parsing multiline environment variables in .env file

aswamy:

Hydrogen CLI is introducing multiline support in their .env files
- The current library used in cli-kit does not support it
- However, we can't fully remove the old library since it still provides the ability to write to .env files (which the new one does not) 

- Adding `dotenv` library

### How to test your changes?

- Just ensure exists unit tests are run, and the new one passes as well

### Measuring impact

### Checklist


packages/cli-kit/package.json
```diff
@@ -123,6 +123,7 @@
     "cross-zip": "4.0.0",
     "deepmerge": "4.3.1",
     "del": "6.1.1",
+    "dotenv": "16.4.5",
     "env-paths": "3.0.0",
     "envfile": "6.18.0",
```
gonzaloriestra, line 128: I didn't like too much having 2 dependencies for the same thing. We already have a lot of dependencies and we should be careful about that... So I had a look at the current library, which is pretty simple, and I finally created a [PR to support multiline values](https://github.com/bevry/envfile/pull/233) that was surprisingly merged in a few minutes. So we can just upgrade `envfile` and revert the changes on dot-env.ts:

```suggestion
    "env-paths": "3.0.0",
    "envfile": "7.1.0-next.1709117147.dc6083d00c810cc6c64bba3a3e9ba728796b6976",
```

aswamy: Let me test it out and write a few tests. If it works out, ill just close this PR (or update it to use the latest `envfile`)

aswamy: Ok i did a bit of testing and I still think we should move onto `dotenv` instead of `envfile` due to the following reasons:
- the recent change to `envfile` you published isn't fully released yet
- the library does not preserve inline double quotes, which is useful if someone is storing JSON.
  - e.g. if your .env file has `SECRET={"some-key":"some-value"}`, the `parse` method is returning the value as `{some-key:some-value}`.
- true newline characters aren't supported (only escaped newlines)
- `dotenv` has higher activity and support while still remaining < 100kb unpacked

NOTE: I updated my PR so that we only include 1 library to prevent either standard from colliding

PR #3508, opened 2024-03-01
Install cloudflared at runtime

isaacroldan:
We were installing cloudflared on a postinstall script but that script was in JS, was blocking CLI installtion for some users who can't download cloudflare and slows down installation for others.

- Remove the post-install step
- Download cloudflare the first time you run `dev` (if necessary)

### How to test your changes?
- remove the cloudflare binary from `plugin-cloudflare/bin/`
- Run `dev`, it should reinstall the binary and work fine
- Run `dev` again, it should not install it this time and work fine.

### Measuring impact

### Checklist


packages/plugin-cloudflare/src/tunnel.ts
```diff
@@ -45,7 +46,13 @@ class TunnelClientInstance implements TunnelClient {
 
   constructor(port: number) {
     this.port = port
-    this.tunnel()
+    install()
+      .then(() => {
+        this.tunnel()
+      })
+      .catch((error) => {
```
MitchDickinson, line 53: I think it'd be good to have a test case for the error scenario too, to ensure that code path is exercised in test.

packages/plugin-cloudflare/src/install-cloudflared.ts
```diff
@@ -1,34 +1,36 @@
 // Install script for cloudflared, derived from https://github.com/JacobLinCool/node-cloudflared
-import * as path from 'path'
+import {basename, dirname, joinPath} from '@shopify/cli-kit/node/path'
```
MitchDickinson, line 2: I don't think this file was under test before. Getting the entire thing under test is surely out of scope for this PR, but this might be a good PR to add a baseline test and maybe capture some of the specific changes you made here. WDYT?

MitchDickinson (commented):
I'm new here so I'd strongly recommend a second review, but adding a couple comments nonetheless so I can start building the PR muscle.

packages/plugin-cloudflare/src/install-cloudflared.ts
isaacroldan: Yeah makes sense, I'll add some tests for this 👌

packages/plugin-cloudflare/src/tunnel.test.ts
```diff
@@ -133,4 +139,15 @@ describe('hookStart', () => {
       tryMessage: expect.anything(),
     })
   })
+
+  test('returns error if it fails to install cloudflared', async () => {
```
MitchDickinson, line 143: Looks great!

packages/plugin-cloudflare/src/tunnel.ts
```diff
@@ -45,7 +47,16 @@ class TunnelClientInstance implements TunnelClient {
 
   constructor(port: number) {
     this.port = port
-    this.tunnel()
+  }
+
+  async startTunnel() {
+    try {
+      await install()
```
MitchDickinson, line 54: I am a JS newbie, just curious why the change away from the callback structure?

isaacroldan: I only used then/catch initially because you can't use `await` inside a constructor, but we prefer to use `async/await` in our codebase because it's easier to read and debug. (is not a good practice to use both approaches in a codebase)

But after looking at the implementation when adding tests, I thought it made sense to actually move that from the constructor into a "setup" function, so that the constructor doesn't have side effects (only builds the object)

This way is also easier to test.

isaacroldan: Those are valid questions! 

> If anyone wants to interact with an instance of TunnelClientInstance via the TunnelClient interface, they may not have a functional tunnel because the tunnel must be started manually via startTunnel() which is not available on the TunnelClient interface

This is actually the intention, but let me explain. Consumers of this interface won't create the instance themselves, they run a hook called `tunnel_start` that calls [`hookStart`](https://github.com/Shopify/cli/blob/main/packages/plugin-cloudflare/src/tunnel.ts#L28), that function is the one that crates (and starts!) the tunnel and returns the instance to control the tunnel (TunnelClient). This is because the tunnel lives in a different package and you could potentially have multiple packages that implement the `tunnel_start` hook.

Hooks are an Oclif thing, more about them [here](https://oclif.io/docs/hooks/))


So from the point of view of the `app` package (the consumer of this interface), it [executes](https://github.com/Shopify/cli/blob/main/packages/app/src/cli/services/dev/urls.ts#L288) a `tunnel_start` hook and receives in return a `TunnelClient` with the tunnel already started :)

gonzaloriestra (approved):
Much better now! And thanks for adding the tests 👏

packages/plugin-cloudflare/src/install-cloudflared.ts
```diff
@@ -0,0 +1,148 @@
...
+  } else if (platform === 'darwin') {
+    await installMacos(fileUrlPath, binTarget)
+  } else if (platform === 'win32') {
+    await installWindows(fileUrlPath, binTarget)
+  } else {
+    throw new Error(`Unsupported platform: ${platform}`)
+  }
+}
+
+function versionIsGreaterThan(versionA: string, versionB: string) {
```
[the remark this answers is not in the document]
isaacroldan: Added tests 👌

packages/plugin-cloudflare/src/install-cloudflared.ts
```diff
@@ -71,7 +71,10 @@ function getBinPathTarget(env = process.env, platform = process.platform) {
 
 export default async function install(env = process.env, platform = process.platform, arch = process.arch) {
   // Don't install cloudflare if the SHOPIFY_CLI_IGNORE_CLOUDFLARED environment variable is set
```
gonzaloriestra, line 73: I think it's clear enough with the code below
```suggestion
```

PR #3540, opened 2024-03-08
Fix link command for apps with empty scopes

alvaro-shopify:

Fixes https://github.com/Shopify/temp-project-mover-Archetypically-20250319145332/issues/354

`managed installation` is mandatory for `only extension` apps, because they don't include a backend to negotiate the installation. To enable the `managed installation` flow the `scopes` entry should be included in the `toml` even if no specific `scopes` are required (setting them to an empty string `""`)

When a remote app has an app version with empty scopes (not `undefined` but empty string `""`) the `link` command was not writing the `scopes` entry in the `toml` so the `old installation` flow is used. This situation causes that `only extensions` app could not be installed is some specific use cases.

- The remote config content builder method includes a bug for the `app_access` section.
- Instead of using a builder method (which was just doing a 1:1 mapping) now the remote config content (parsed and validated using the `configuration` specs) will be simply casted to the typed class [SpecsAppConfiguration](https://github.com/Shopify/cli/blob/main/packages/app/src/cli/models/extensions/specifications/types/app_config.ts#L3)
- Additionally, now the `link` logic is run when you run `deploy` with `--reset` using a `legacy schema` configuration

### How to test your changes?
- Create a new `only extension` app
```sh
pnpm create-app --path <YOUR_APPS_PATH> --template none --name 0308-only-extension;cd <YOUR_APPS_PATH>/0308-only-extension
```
- Run the command `deploy` and `create a new remote app` when prompted to
```
pnpm shopify app deploy --path <YOUR_APPS_PATH>/0308-only-extension
```
- Open the partners dashboard and generate a custom installation link 
  - Access `Apps -> <YOUR-REMOTE-APP> -> Distribution ->  Custom distribution`.
  - Use you `dev` store domain
- Copy the `installation link`, open a new tab and paste it. The installation prompt should be displayed

### Measuring impact

### Checklist

### DIFF
--- a/.changeset/wise-houses-kick.md
+++ b/.changeset/wise-houses-kick.md
@@ -1,0 +1,5 @@
+---
+'@shopify/app': patch
+---
+
+Fix link command for apps with empty scopes
--- a/packages/app/src/cli/models/app/app.ts
+++ b/packages/app/src/cli/models/app/app.ts
@@ -5,7 +5,6 @@ import {isType} from '../../utilities/types.js'
 import {FunctionConfigType} from '../extensions/specifications/function.js'
 import {ExtensionSpecification} from '../extensions/specification.js'
 import {SpecsAppConfiguration} from '../extensions/specifications/types/app_config.js'
-import {WebhooksConfig} from '../extensions/specifications/types/app_config_webhook.js'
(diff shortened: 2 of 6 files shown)


gonzaloriestra (approved):
Much better now! 👏

PR #3549, opened 2024-03-12
Rename app modules UUID to UID for Developer Platform

gonzaloriestra:

Related to https://github.com/Shopify/shopify/pull/489249

Make the required changes to make it compatible with the changes in the PR above

### How to test your changes?

`USE_SHOPIFY_DEVELOPERS_CLIENT=1 bin/spin p shopify app config link`

### Measuring impact

### Checklist


amcaplan (approved):
Works like a charm 🎩, and the code makes sense.

I think down the road we should think more deeply about the right interface for `AppModuleVersion` (maybe a union type would be appropriate). But this unblocks us for now, and we can easily modify our approach later, so let's go for it!

PR #3564, opened 2024-03-15
Bump Shopify/theme-tools packages

### LINKED ISSUE #3497: Theme check: command doesn't inspect any file after CLI upgrade
**Describe the bug**
Running the command line 'shopify theme check' is not inspecting files, returning a Success message , 0 files inspected with no offenses.

**Source**
It happens on all Theme projects, even Dawn. It also happens to my other colleagues on their computers.

**Expected behaviour**
The command should output a large list of errors and suggestions.

**Actual behaviour**
The command is not inspecting files.
The command returns:
Success
Theme Check Summary.
0 files inspected with no offenses found.

**Debugging information**
 - OS Windows
 - OS Version 11
 - Shopify CLI v3.56.3

**Additional context**
The command was working fine until I upgraded shopify CLI from v3.48.x to v3.52.x. It is broken since then.

See an example with a fresh start of a Dawn theme:
image

charlespwd:

Fixes #3497

- Bumps Shopify/theme-tools dependencies
  - Fixes a bug on Windows with `shopify theme check` not globbing files properly
  - Adds support for JSON intellisense of locales and section schemas to `shopify theme language-server`


packages/app/package.json
```diff
@@ -44,12 +44,12 @@
     "@luckycatfactory/esbuild-graphql-loader": "3.8.1",
     "@oclif/core": "3.19.6",
     "@shopify/cli-kit": "3.57.0",
+    "@shopify/create-app": "3.57.0",
```
charlespwd, line 47: Hmm that shouldn't be there.

Issue #3578, opened 2024-03-19
[Bug]: CLI `dev` command fails with no port available for GraphiQL regardless of port chosen (MacOS)


mikefortney:
### Please confirm that you have:

- [X] Searched [existing issues](.) to see if your issue is a duplicate. (If you’ve found a duplicate issue, feel free to add additional information in a comment on it.)
- [X] Reproduced the issue in [the latest CLI version](https://www.npmjs.com/package/@shopify/cli).

### In which of these areas are you experiencing a problem?

App

### Expected behavior

CLI `dev` command should launch without issues on MacOS

### Actual behavior

CLI `dev` command fails with the error 'Port 3457 is not available for serving GraphiQL' and offers a command option to use a different port. No matter which port I choose, the behavior is the same. I have looked into the code a bit, and I see that you are using `get-port-please` to check port availability. There may be an issue in that package that misreports the availability of the port, OR there is something else blocking the port availability even tho it is reported as available.

### Verbose output

```shell
➜  inspired-shopify-app git:(mf-test-20241315) ✗ yarn dev --verbose
2024-03-19T17:27:38.713Z: Running command app dev
2024-03-19T17:27:38.725Z: Getting a random port...
2024-03-19T17:27:38.729Z: Random port obtained: 52870
2024-03-19T17:27:38.730Z: Ensuring that the user is authenticated with the Partners API with the following scopes:
[]

2024-03-19T17:27:38.732Z: Getting session store...
2024-03-19T17:27:38.737Z: Validating existing session against the scopes:
[
  "openid",
  "https://api.shopify.com/auth/shop.admin.graphql",
  "https://api.shopify.com/auth/shop.admin.themes",
  "https://api.shopify.com/auth/partners.collaborator-relationships.readonly",
  "https://api.shopify.com/auth/shop.storefront-renderer.devtools",
  "https://api.shopify.com/auth/partners.app.cli.access",
  "https://api.shopify.com/auth/destinations.readonly"
]
For applications:
{
  "partnersApi": {
    "scopes": []
  }
}

2024-03-19T17:27:38.738Z: Sending Identity Introspection request to URL: https://accounts.shopify.com/oauth/introspection
2024-03-19T17:27:38.739Z: Sending POST request to URL https://accounts.shopify.com/oauth/introspection
With request headers:
 - User-Agent: Shopify CLI; v=3.57.1
 - Keep-Alive: timeout=30
 - Sec-CH-UA-PLATFORM: darwin
[...]

PR #3579, opened 2024-03-20
Use a random port for GraphiQL when the default one is not available

gonzaloriestra:

Fixes https://github.com/Shopify/cli/issues/3578

Since GraphiQL was added, you can't run `dev` multiple times simultaneously without specifying a different GraphiQL port, as we always use the same default value (3457). When you try to run dev a second time, it raises an error:

20-55-wdzm7-m9r05

If the default port is in use, find an available one instead of raising an error. In that case, a warning is shown:

image

### How to test your changes?

Run `p shopify app dev` in two tabs and ensure that GraphiQL works in both places

### Measuring impact

### Checklist

### DIFF
--- a/packages/app/src/cli/services/dev.ts
+++ b/packages/app/src/cli/services/dev.ts
@@ -96,6 +96,21 @@ async function prepareForDev(commandOptions: DevOptions): Promise<DevConfig> {
   const graphiqlPort = commandOptions.graphiqlPort || (await getAvailableTCPPort(ports.graphiql))
   const {graphiqlKey} = commandOptions
 
+  if (graphiqlPort !== (commandOptions.graphiqlPort || ports.graphiql)) {
+    renderWarning({
+      headline: [
+        'A random port will be used for GraphiQL because',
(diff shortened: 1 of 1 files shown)


amcaplan:
I'm not sure this is a good idea. Certain features of GraphiQL (query history, session recovery) break if you switch ports, so I'm thinking users should have to do this manually.

gonzaloriestra:
I know, but even without history, GrahpiQL keeps working and it's not the main purpose of the dev command. It must be annoying for people having the default port already in use, or when they want to dev two different apps at the same time...

What if we show a warning when the port is random? Something like:
```
The default port for GraphiQL (3457) is not available, so a random one has been chosen (5754).

If you want to keep your GrahpiQL session, you can manually select a port by passing `--graphiql-port`.
```

gonzaloriestra:
Another alternative would be asking to kill the other process, as we do with the authentication:

21-28-vltar-bs9t3

isaacroldan (approved):

PR #3583, opened 2024-03-20
Add organization selection to ShopifyDevelopersClient

amcaplan:

For prototyping, we've hardcoded the organization ID. Instead, we'll pull the list from business platform.

Enables fetching a list of orgs to select from.

### How to test your changes?

`app config link` should prompt for an organization to connect to, and use the organization you selected. You can check by running `--verbose` and seeing that the URL includes the correct `organizationId` (2 vs 5 for the dev user).

### Measuring impact

### Checklist

### DIFF
--- a/packages/app/src/cli/utilities/developer-platform-client/partners-client.ts
+++ b/packages/app/src/cli/utilities/developer-platform-client/partners-client.ts
@@ -4,7 +4,7 @@ import {
   AllDevStoresByOrganizationQueryVariables,
   AllDevStoresByOrganizationSchema,
 } from '../../api/graphql/all_dev_stores_by_org.js'
-import {ActiveAppVersion, DeveloperPlatformClient, Paginateable} from '../developer-platform-client.js'
+import {ActiveAppVersion, AppDeployOptions, DeveloperPlatformClient, Paginateable} from '../developer-platform-client.js'
 import {fetchCurrentAccountInformation, PartnersSession} from '../../../cli/services/context/partner-account-info.js'
 import {fetchAppDetailsFromApiKey, fetchOrgAndApps, filterDisabledFlags} from '../../../cli/services/dev/fetch.js'
 import {
@@ -326,8 +326,11 @@ export class PartnersClient implements DeveloperPlatformClient {
     return this.request(ExtensionUpdateDraftMutation, extensionInput)
   }
 
-  async deploy(deployInput: AppDeployVariables): Promise<AppDeploySchema> {
-    return this.request(AppDeploy, deployInput)
+  async deploy(deployInput: AppDeployOptions): Promise<AppDeploySchema> {
+    const {organizationId, ...deployOptions} = deployInput
+    // Enforce the type
+    const variables: AppDeployVariables = deployOptions
+    return this.request(AppDeploy, variables)
   }
 
   async release(input: AppReleaseVariables): Promise<AppReleaseSchema> {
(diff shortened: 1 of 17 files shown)


packages/app/src/cli/utilities/developer-platform-client/partners-client.ts
```diff
@@ -326,8 +331,11 @@ export class PartnersClient implements DeveloperPlatformClient {
     return this.request(ExtensionUpdateDraftMutation, extensionInput)
   }
 
-  async deploy(deployInput: AppDeployVariables): Promise<AppDeploySchema> {
-    return this.request(AppDeploy, deployInput)
+  async deploy(deployInput: AppDeployOptions): Promise<AppDeploySchema> {
+    const {organizationId, ...deployOptions} = deployInput
+    // Enforce the type
+    const variables: AppDeployVariables = deployOptions
+    return this.request(AppDeploy, variables)
```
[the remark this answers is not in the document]
amcaplan: Nothing in `this.request` enforces the type of the inputs. I could also do `return this.request(AppDeploy, {invalid: "input"})` and the compiler wouldn't complain.

We saw situations in the past where the input type was changed to be incorrect and a bug was introduced. So @gonzaloriestra and I decided to make it a consistent practice to ensure inputs are assigned a type explicitly before being passed to the GraphQL request. This has prevented us from introducing new errors in client code, relying on the type system to insure inputs are correct.

Perhaps there are other ways to parse the `gql` string and ensure the inputs are aligned, but for now we don't have a way of doing this, so we need to be a bit more explicit.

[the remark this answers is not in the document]
amcaplan: More details [here](https://github.com/Shopify/cli/pull/3530)

packages/app/src/cli/utilities/developer-platform-client/shopify-developers-client.ts
```diff
@@ -134,30 +163,57 @@ export class ShopifyDevelopersClient implements DeveloperPlatformClient {
   }
 
   async organizations(): Promise<Organization[]> {
-    return [ORG1]
+    const organizationsResult = await businessPlatformRequest<OrganizationsQuerySchema>(
+      OrganizationsQuery,
+      await this.businessPlatformToken(),
+    )
+    return organizationsResult.currentUserAccount.organizations.nodes.map((org) => ({
+      id: Buffer.from(org.id, 'base64').toString('ascii').match(/\d+$/)![0],
```
amcaplan, line 171: Context: Business Platform returns the ID as a base64-encoded gid. So we base64-decode, then pull out the numeric portion at the end, which is what we'll use for calls to the Shopify Developers API.

packages/app/src/cli/utilities/developer-platform-client/shopify-developers-client.ts
```diff
@@ -134,30 +163,57 @@ export class ShopifyDevelopersClient implements DeveloperPlatformClient {
...
+      businessName: org.name,
+    }))
   }
 
   async orgFromId(orgId: string): Promise<Organization | undefined> {
-    if (orgId === '1') return ORG1
+    const gid = `gid://organization/Organization/${orgId}`
+    const base64Id = Buffer.from(gid).toString('base64')
```
amcaplan, line 178: Here you can see the opposite function: take a numeric ID, convert to a gid and base64-encode.

packages/app/src/cli/utilities/developer-platform-client/shopify-developers-client.ts
shauns: I think I'd just add the above as a comment. So someone can see at a glance what these three functions calls add up to

PR #3584, opened 2024-03-20
Fix readme generation

isaacroldan:
Apparently it has been broken for a while because we were using the `usage` property incorrectly.

- Remove `usage` unless necessary
- Use `:` as separator in `usage` because oclif needs it to properly parse the commands
- If there are multiple usages, convert to array instead of multiline string.

### How to test your changes?
Nothing, just review that the README is not broken anymore

### Measuring impact

### Checklist

### DIFF
--- a/packages/app/oclif.manifest.json
+++ b/packages/app/oclif.manifest.json
@@ -2097,8 +2097,7 @@
         "trigger.js"
       ],
       "strict": true,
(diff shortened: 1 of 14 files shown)


gonzaloriestra (approved):
I'd better remove that README, but it's ok

PR #3617, opened 2024-03-28
[DNM] [LogStreaming Prototype] Add replay command

lopert:

Part of the LogStreaming Prototype project.

- adds a new command `shopify app function replay`, which:
  - reads a folder for run logs
  - presents the list of runs in a selector
  - parses the input from the selected run
  - forwards the run to function-runner

### How to test your changes?

- checkout this branch in your local CLI
- create a `/runs/` folder under your app's extension
- (for now) manually populate it with run logs.
  - ex, for our 10% discount function, you can put this in a json file in that folder
```json
{
  "payload": {
    "input":"{\"cart\":{\"lines\":[{\"quantity\":1,\"merchandise\":{\"__typename\":\"ProductVariant\",\"id\":\"gid:\\\/\\\/shopify\\\/ProductVariant\\\/1\"}}]}}",
    "invocationId":"11111111-ed53-4377-b30f-14e8f4653cfe"
  }
}
```
  - create a few of these to have options in the selector
- run the following command from the CLI directory
```
pnpm run shopify app function replay --path=<path-to-your-ext>
```
ex:
```
pnpm run shopify app function replay --path=/Users/lopert/cli/lopert-cli-local/extensions/10-percent-off
```


lopert:
Some activity to keep the PR open, will pick this up after ATC work this week.

docs-shopify.dev/commands/app-function-replay.doc.ts
```diff
@@ -0,0 +1,34 @@
...
+  {
+    title: 'Flags',
+    description: 'The following flags are available for the `app function replay` command:',
+    type: 'appfunctionreplay',
+  },
+  ],
+  category: 'app',
+  related: [
+  ],
+}
```
DuncanUszkay1, line 32: Is this going to be publicly visible?

packages/app/src/cli/commands/app/function/replay.ts
```diff
@@ -0,0 +1,68 @@
...
+import {appFlags} from '../../../flags.js'
+import {selectRunPrompt} from '../../../prompts/dev.js'
+import Command from '@shopify/cli-kit/node/base-command'
+import {globalFlags} from '@shopify/cli-kit/node/cli'
+import {Flags} from '@oclif/core'
+import {writeFile, inTemporaryDirectory} from '@shopify/cli-kit/node/fs'
+import {joinPath} from '@shopify/cli-kit/node/path'
+
+export default class FunctionReplay extends Command {
+  static summary = 'Replays a function locally based on a FunctionRunEvent.'
```
DuncanUszkay1, line 13: We've moving away from `Event` and towards `Log`, we should do a sweep here

packages/app/src/cli/services/function/replay.ts
```diff
@@ -0,0 +1,27 @@
...
+  const runFilePaths = runFileNames.map((runFile) => joinPath(runsFolder, runFile))
+
+  // read contents
+  const runData = await Promise.all(
+    runFilePaths.map((runFile) => {
+      return readFile(runFile)
+    }),
+  )
+
+  // convert promise'd strings to json
```
DuncanUszkay1, line 24: IMO the comments here aren't necessary on `main`

packages/app/src/cli/services/function/replay.ts
```diff
@@ -0,0 +1,27 @@
...
+  const runs = runData.map((run) => JSON.parse(run))
+  return runs
```
DuncanUszkay1, line 26: Seems like something we should inline- thoughts?

DuncanUszkay1 (commented):
Thoughts:
- How can we merge this while still keeping it hidden? Not to prevent someone from finding out, but just to not mislead them into thinking that it is ready to use
- Let's talk more about the overall spec for the CLI commands on Monday

It seems like the best path forward here is to park this, start finalizing the app dev command including writing the files, and circle back once that's done. We should also have our specs cleared up by then and in every CLI issue.

packages/app/src/cli/services/function/replay.ts
```diff
@@ -0,0 +1,27 @@
+import {joinPath} from '@shopify/cli-kit/node/path'
+import {readFile} from '@shopify/cli-kit/node/fs'
+import {readdirSync} from 'fs'
+
+export async function readFunctionRunsDirectory(functionPath: string) {
```
andrewhassan, line 5: Nit: Since this is a function that's exported, I think it'd be good to also type the return type so that the reader doesn't have to grok the implementation.

packages/app/src/cli/services/function/replay.ts
```diff
@@ -0,0 +1,27 @@
...
+
+  // read contents
+  const runData = await Promise.all(
+    runFilePaths.map((runFile) => {
+      return readFile(runFile)
+    }),
+  )
+
+  // convert promise'd strings to json
+  const runs = runData.map((run) => JSON.parse(run))
```
andrewhassan, line 25: Instead of reading/parsing every file, since we're using individual files, we could probably get away with saving the files as `<uuid>.json` and returning an array of `{name: "<uuid>", updatedAt: Date}` using the file metadata.

packages/app/src/cli/services/function/replay.ts
lopert: reworked this so this is no longer exported 👍

packages/app/src/cli/services/function/replay.ts
lopert: Hmmm interesting idea. We now have the `identifier` at the end of the file, so we should be able to produce this. I'll give it a whirl.

packages/app/src/cli/commands/app/function/replay.ts
```diff
@@ -0,0 +1,83 @@
...
+      env: 'SHOPIFY_FLAG_CLIENT_ID',
+      exclusive: ['config'],
+    }),
+    export: Flags.string({
+      char: 'e',
+      hidden: false,
+      description: 'Name of the Web Assembly export to invoke.',
+      default: '_start',
+      env: 'SHOPIFY_FLAG_EXPORT',
+    }),
```
DuncanUszkay1, line 42: 💭 Isn't the export being invoked defined by the run itself? E.G. a run using `fetch` couldn't be sensibly rerun using `run`. Perhaps that should be part of the payload so you never have to specify it here?

Probably not urgent for this PR, but would be an improvement for later.

packages/app/src/cli/prompts/dev.test.ts
```diff
@@ -86,6 +125,27 @@ describe('selectOrganization', () => {
   })
 })
 
+describe('selectFunctionRun', () => {
```
DuncanUszkay1, line 128: I noticed the name of this file is `dev.test.ts`- is our prompt defined in the right place?

packages/app/src/cli/commands/app/function/replay.ts
lopert: 🤔 It's one of the flags that we can invoke function runner with so that's how it ended up here. 
The export is defined in the extension's `shopify.extension.toml`, so I think we could maybe improve this by just reading from there?

packages/app/src/cli/prompts/dev.test.ts
lopert: It is among the other prompts so I think so

packages/app/src/cli/commands/app/function/replay.ts
```diff
@@ -0,0 +1,68 @@
...
+    const functionPath = flags.path
+
+    const runs = await readFunctionRunsDirectory(functionPath)
+
+    const selectedRun = await selectRunPrompt(runs)
+
+    const input = selectedRun.payload.input
+
+    // dump the output to a file
+    await inTemporaryDirectory(async (tmpDir) => {
```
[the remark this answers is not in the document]
lopert: Turns out we can use the [simple input](https://github.com/sindresorhus/execa?tab=readme-ov-file#simple-input) to pass in a string. I've updated the replay code to use this variant of invoking `exec` without touching the existing implementation.

packages/app/src/cli/prompts/dev.ts
```diff
@@ -16,6 +17,23 @@ export async function selectOrganizationPrompt(organizations: Organization[]): P
   return organizations.find((org) => org.id === id)!
 }
 
+export async function selectFunctionRunPrompt(functionRuns: FunctionRunData[]): Promise<FunctionRunData> {
```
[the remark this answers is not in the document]
lopert: Thanks for brining this up again. For some reason reading through it again, I understood what you were both asking 🙈 . I've moved this prompt to it's own file under `.../prompts/function/replay/`

packages/app/src/cli/services/app-logs/write-app-logs.ts
```diff
@@ -32,3 +34,12 @@ export const writeAppLogsToFile = async ({
...
   }
 }
+
+function formatTimestampToFilename(timestamp: string): string {
+  // 2024-05-22T15:06:41.827379Z
+  const year = timestamp.substring(0, 10).replace(/-/g, '')
+  const time = timestamp.substring(11, 19).replace(/:/g, '')
+  const microseconds = timestamp.substring(20, 26)
+  const timezone = timestamp.substring(26)
+  return `${year}_${time}_${microseconds}${timezone}`
```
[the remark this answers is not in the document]
lopert: I attempted this initially but I couldn't see how to have the `Date` object have microsecond precision. 
```javascript
> date = new Date("2024-06-04T14:34:44.334440Z")
2024-06-04T14:34:44.334Z
> date.toISOString()
'2024-06-04T14:34:44.334Z'
```

packages/app/src/cli/services/app-logs/write-app-logs.ts
```diff
@@ -13,9 +13,11 @@ export const writeAppLogsToFile = async ({
   apiKey: string
   stdout: Writable
 }) => {
-  const fileName = `app_logs_${appLog.log_timestamp}.json`
+  const identifier = randomUUID().substring(0, 6)
+  const formattedTimestamp = formatTimestampToFilename(appLog.log_timestamp)
```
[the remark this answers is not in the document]
lopert: Going to lead this as is for now since we're going to rework the response serverside. then we can come back and update this.

packages/app/src/cli/services/function/replay.ts
```diff
@@ -0,0 +1,88 @@
...
+  const functionRunFileNames = readdirSync(functionRunsDir)
+  const functionRunFilePaths = functionRunFileNames.map((functionRunFile) => joinPath(functionRunsDir, functionRunFile))
+
+  const functionRunData = await Promise.all(
+    functionRunFilePaths.map(async (functionRunFilePath) => {
+      const fileData = await readFile(functionRunFilePath)
+      const parsedData = JSON.parse(fileData)
+      return {
+        ...parsedData,
+        identifier: getIdentifierFromFilename(functionRunFilePath) || 'no identifier',
```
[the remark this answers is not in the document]
lopert: Oh, because the timestamps contain `_` as well, if I mess with the files and delete the identifier manually, it just takes the the last part of the split, resulting in
```
?  Which function run would you like to replay locally?

>  2024-06-05T15:01:04.154838Z (failure) - 154Z.j
   2024-06-05T15:01:03.420454Z (failure) - 4a0b61
```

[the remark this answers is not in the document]
lopert: Since we should never not have an identifier, I've just removed the undefined return with `pop()!`

packages/app/src/cli/commands/app/function/replay.ts
nickwesselman: I agree with @DuncanUszkay1 that this should be part of the payload, the input you are replaying would be specific to a certain target.

DuncanUszkay1: Will add that to the payload in my upcoming PR that also adds source/source namespace

packages/app/src/cli/services/function/replay.ts
```diff
@@ -0,0 +1,83 @@
...
+      }
+    }),
+  )
+
+  return functionRunData
+}
```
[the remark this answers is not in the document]
lopert: Set a simple limit of 100 for now. We can revisit later if we need to provide an input arg to make this more flexible.

packages/app/src/cli/services/function/replay.test.ts
```diff
@@ -0,0 +1,134 @@
...
+  event_type: 'function_run',
+  cursor: '2024-05-31T15:29:47.291530Z',
+  status: 'success',
+  log_timestamp: '2024-05-31T15:29:50.741270Z',
+  identifier: '123456',
+}
+const RUN2_FILENAME = RUN2.log_timestamp.replace(/:/g, '_')
+
+describe('replay', () => {
+  test('reads the app log directory, parses the function runs, and invokes function-runner aa', async () => {
```
DuncanUszkay1, line 59: Should we add a test for the situation where the app log directory is empty?

packages/app/src/cli/services/function/replay.ts
```diff
@@ -0,0 +1,93 @@
...
+  const outputAsJson = options.json ? ['--json'] : []
+  const exportName = options.export ? ['--export', options.export] : []
+
+  return exec('npm', ['exec', '--', 'function-runner', '-f', fun.outputPath, ...outputAsJson, ...exportName], {
+    cwd: fun.directory,
+    input,
+    stdout: 'inherit',
+    stderr: 'inherit',
+  })
+}
```
DuncanUszkay1, line 68: Leaving a comment here as an anchor for an issue to start using the export in the log itself once that's available. If it's a duplicate feel free to close, but I didn't see another one.

DuncanUszkay1 (approved):
LGTM aside from dealing with the 'no logs to replay' case explicitly

packages/app/src/cli/services/function/replay.test.ts
lopert: Good callout! 
I see the other prompts return undefined in this scenario so I'll do that. 
Then here I'll throw an error showing the directory that we tried to find runs in.

packages/app/src/cli/services/app-logs/write-app-logs.ts
```diff
@@ -13,9 +13,12 @@ export const writeAppLogsToFile = async ({
   apiKey: string
   stdout: Writable
 }) => {
-  const fileName = `app_logs_${appLog.log_timestamp}.json`
+  const identifier = randomUUID().substring(0, 6)
+
+  const formattedTimestamp = formatTimestampToFilename(appLog.log_timestamp)
+  const fileName = `${formattedTimestamp}_${identifier}.json`
```
andrewhassan, line 19: Just curious: do we get an error if there's already a file with this name? I'm just wondering in the extreme off chance that there's a collision.

packages/app/src/cli/services/function/replay.test.ts
```diff
@@ -0,0 +1,163 @@
...
```
andrewhassan, line 59:
```suggestion
  test('reads the app log directory, parses the function runs, and invokes function-runner', async () => {
```

andrewhassan (approved):

PR #3633, opened 2024-04-01
[Theme] Add handling when server does not respond to bulk asset upload

jamesmengo:
- https://github.com/Shopify/develop-advanced-edits/issues/187
- This PR adds error handling for the case when the bulk asset upload endpoint responds with an empty object
  - See https://github.com/Shopify/cli/pull/3645 for more context 

- Rather than failing silently, we should return a response to the user that indicates an issue
  - I've chosen to Abort the operation rather than continuing to execute.
  - Another option is to try to execute the remaining uploads, but there is little value in doing so as this type of response indicates a larger issue (in the case linked above, bulk uploads are not supported by the theme access proxy) - this means that all the subsequent requests will also fail.

### How to test your changes?
- Checkout main
- run `pnpm shopify theme push -u --password <PASSWORD> --verbose`
- Observe that the pushed theme is empty, but the command output will indicate success
- Checkout this branch
- run `pnpm shopify theme push -u --password <PASSWORD> --verbose`
- This time, the command should fail and be aborted

image

### DIFF
--- a/packages/cli-kit/src/public/node/themes/factories.ts
+++ b/packages/cli-kit/src/public/node/themes/factories.ts
@@ -56,7 +56,7 @@ export function buildBulkUploadResults(
   bulkUploadResponse: RemoteBulkUploadResponse[],
   assets: AssetParams[],
 ): Result[] {
-  if (bulkUploadResponse.length === 0 && assets.length !== 0) {
+  if (bulkUploadResponse === undefined) {
     return assets.map((asset) => ({
       key: asset.key,
       success: false,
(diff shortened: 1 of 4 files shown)


packages/cli-kit/src/public/node/themes/factories.ts
```diff
@@ -56,18 +56,22 @@ export function buildBulkUploadResults(
   bulkUploadResponse: RemoteBulkUploadResponse[],
   assets: AssetParams[],
 ): Result[] {
-  const results: Result[] = []
-  if (!bulkUploadResponse) return results
+  if (bulkUploadResponse.length === 0 && assets.length !== 0) {
```
jamesmengo, line 59: I believe that the server will either return an array of `RemoteBulkUploadResponse` with the same length as `assets`, or it will return `[]`

I don't believe that we will encounter the case where `bulkUploadResponse.length > 0 && bulkUploadResponse.length < assets.length)

Issue #3664, opened 2024-04-05
[Bug]: Warning message since 3.58.2


msev:
### Please confirm that you have:

- [X] Searched [existing issues](.) to see if your issue is a duplicate. (If you’ve found a duplicate issue, feel free to add additional information in a comment on it.)
- [X] Reproduced the issue in [the latest CLI version](https://www.npmjs.com/package/@shopify/cli).

### In which of these areas are you experiencing a problem?

Theme

### Expected behavior

The `.shopifyignore` file contains the following rule:
```
config/settings_data.json
templates/*.json
```
Since version 3.58.2, when I run the shopify theme push command, the following message appears :

```
The pattern templates/*.json is unsafe, please use a valid regex or glob.
The pattern templates/*.json is unsafe, please use a valid regex or glob.
The pattern templates/*.json is unsafe, please use a valid regex or glob.
```

However, the documentation states that this type of wildcard can be used : [https://shopify.dev/docs/themes/tools/cli#excluding-files-from-shopify-cli](https://shopify.dev/docs/themes/tools/cli#excluding-files-from-shopify-cli)

> .shopifyignore accepts references to files in the following formats:
> 
> - Simple file names: templates/product.temp.json
> - Wildcards: config/*_secret.json, *.jpg
> - Regular expressions: /\.(txt|gif|bat)$/

### Actual behavior

A warning message is displayed

By using the `--stable` flag this message no longer appears

### Verbose output

```shell
Too long
```

### Reproduction steps

1. Add `templates/*.json` rule to `.shopifyignore` file or `shopify.theme.toml` environment 
2. Run `shopify theme push` command
3. A warning message is displayed

### Operating System

macOS 14.2.1

### Shopify CLI version (check your project's `package.json` if you're not sure)

3.58.2

### Node version (run `node -v` if you're not sure)

20.11.0

msev:
I don't understand.

I modified the `.shopifyignore` file by adding `/**/*.json` and I still get the following warning message:

```
The pattern /**/*.json is unsafe, please use a valid regex or glob.
```

msev:
Hi @jamesmengo 

I downgraded to version 3.57.1 to test the `templates/**/*.json` pattern, but it doesn't work as it should: the `*.*.json` files in the templates folder have been deleted.
I'm working with several developers, and we won't be able to use this pattern until they update Shopify CLI to 3.58.x.

Before running the `shopify theme push` command:

[...]

PR #3669, opened 2024-04-05
[Themes] Improve Glob Pattern subdirectory mismatch error handling

jamesmengo:
- https://github.com/Shopify/cli/issues/3664
- https://github.com/Shopify/develop-advanced-edits/issues/188

This warning was added to surface some work that we were always doing on the Ruby side to include subdirectories when pattern matching theme files (`.shopifyignore`, `--ignore`, `--only`)

Technically, the pattern `templates/*.json` **should not** match subdirectories, but this would be a largely breaking change, so we apply this change on the platform side when needed.
- Instead of doing this work silently in the background, the warning informs devs of the modification we were making to their patterns. 

- Fixes an issue where the warning was rendering on all pattern mismatches ( we now only render the message if the modified pattern leads to a match a hit while the original pattern does not )
  - The # of files output should only match the number of files you have in subdirectories
- Improves the clarity of the error message
- **I'm Open to input on the error message + frequency of when to report it**

### How to test your changes?
1) Start with a clean theme **that has subdirectories** (e.g. dawn with `templates/customers`)
2) `pnpm shopify theme push --path <PATH_TO_THEME> -u --ignore "templates/*.json"`
  - You can also try this with `--only` or `.shopifyignore`
image

3) Modify the pattern now to `templates/**/*.json` and observe that the messages disappear
4) In either case (warning or no warning), this **pattern should apply to subdirectories**

### Measuring impact

### Checklist

### DIFF
--- a/packages/theme/src/cli/utilities/asset-ignore.ts
+++ b/packages/theme/src/cli/utilities/asset-ignore.ts
@@ -12,15 +12,20 @@ export async function applyIgnoreFilters(
 ) {
   const shopifyIgnore = await shopifyIgnoredPatterns(themeFileSystem)
 
+  const ignoreOptions = options.ignore ?? []
+  const onlyOptions = options.only ?? []
+
+  raiseWarningForNonExplicitGlobPatterns([...shopifyIgnore, ...ignoreOptions, ...onlyOptions])
+
   return themeChecksums
     .filter(filterBy(shopifyIgnore, '.shopifyignore'))
-    .filter(filterBy(options.ignore, '--ignore'))
-    .filter(filterBy(options.only, '--only', true))
+    .filter(filterBy(ignoreOptions, '--ignore'))
+    .filter(filterBy(onlyOptions, '--only', true))
 }
 
-function filterBy(patterns: string[] | undefined, type: string, invertMatch = false) {
+function filterBy(patterns: string[], type: string, invertMatch = false) {
   return ({key}: Checksum) => {
-    if (!patterns) return true
+    if (patterns.length === 0) return true
 
     const match = patterns.some((pattern) => matchGlob(key, pattern) || regexMatch(key, pattern))
     const shouldIgnore = invertMatch ? !match : match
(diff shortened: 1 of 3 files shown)


packages/theme/src/cli/utilities/asset-ignore.ts
```diff
@@ -56,9 +56,17 @@ function matchGlob(key: string, pattern: string) {
   // When the the standard match fails and the pattern includes '/*.', we
   // replace '/*.' with '/**/*.' to emulate Shopify CLI 2.x behavior, as it was
   // based on 'File.fnmatch'.
-  if (pattern.includes('/*.')) {
-    outputWarn(`The pattern ${pattern} is unsafe, please use a valid regex or glob.`)
-    return originalMatchGlob(key, pattern.replace('/*.', '/**/*.'))
+  if (pattern.includes('/*.') && !pattern.includes('/**/*.')) {
+    const newPatternMatch = originalMatchGlob(key, pattern.replace('/*.', '/**/*.'))
+    if (newPatternMatch) {
```
karreiro, line 61: Personally, I don't think we should adopt pattern matching or not matching as our criteria for showing the warning, as the pattern is still used when it's not being matched.

What do you think about moving this warning to the `applyIgnoreFilters` level? Then, we'd display this warning once per pattern (instead of per pattern usage), and we'd still inform users about this :)

karreiro (commented):
Thank you for this PR, @jamesmengo! I've left only one minor question, please let me know what you think about it :)

packages/theme/src/cli/utilities/asset-ignore.ts
```diff
@@ -56,14 +61,31 @@ function matchGlob(key: string, pattern: string) {
...
+        `Warning: The pattern '${pattern}' does not include subdirectories. To maintain backwards compatibility, we have modified your pattern to ${pattern.replace(
+          '/*.',
+          '/**/*.',
+        )} to explicitly include subdirectories.`,
+      )
+    }
+  })
+}
+
+function shouldReplaceGlobPattern(pattern: string): boolean {
```
jamesmengo, line 85: Decision: I chose to limit this to the `templates` folder because that's the only folder where [subdirectories are supported](https://shopify.dev/docs/themes/architecture#directory-structure-and-component-types).

Some questions that could influence this
- Do you consider something like `assets/*.liquid` to be unsafe / something we need to substitute?
  - If we recommend `assets/**/*.liquid`, that could be confusing since there are no subdirectories there 

- Will we remove the automatic conversion of the glob pattern one day so that `templates/*.json` and `templates/**/*.json` mean different things in the future?
  - Some context [here](https://github.com/Shopify/cli/issues/3664)

packages/theme/src/cli/utilities/asset-ignore.ts
jamesmengo: https://github.com/Shopify/cli/pull/3669#discussion_r1555374641

packages/theme/src/cli/utilities/asset-ignore.ts
karreiro: +1 for restricting this warning to templates, as it's the only directory that may have a sub-directory :)

Regarding the removal of this warning in the future, because it would be a destructive breaking change (as in some contexts it may imply file deletion), I'd suggest making this change in the next major release

karreiro (approved):
Thank you, @jamesmengo!

PR #3689, opened 2024-04-10
Add support for ui_extension_handle on import-extensions command

bazay:

Fixes #0000 

### Measuring impact

### Checklist


packages/app/src/cli/services/payments/extension-to-toml.ts
```diff
@@ -88,6 +97,14 @@ function buildPaymentsToml<T extends BasePaymentsAppExtensionDeployConfigType>(
 
   const cliConfig = serialize(dashboardConfig)
   if (cliConfig) delete cliConfig.api_version
+  if (cliConfig && !cliConfig.ui_extension_handle && 'ui_extension_registration_uuid' in dashboardConfig) {
```
yigitozkavci, line 100: Serialising `ui_extension_handle` as null first, and then (conditionally) populating it here doesn't feel right to me, I think the `serialize` function should be responsible for serialising the whole `cliConfig`.

Also this function is pretty agnostic of individual fields we produce inside `cliConfig` (aside from the `api_version` since it's put into the root level)

packages/app/src/cli/services/import-extensions.ts
```diff
@@ -15,7 +15,7 @@ interface ImportOptions {
   apiKey?: string
   developerPlatformClient?: DeveloperPlatformClient
   extensionTypes: string[]
-  buildTomlObject: (ext: ExtensionRegistration) => string
+  buildTomlObject: (ext: ExtensionRegistration, appModules: ExtensionRegistration[]) => string
```
yigitozkavci, line 18: `appModules` here represents all of the extensions, just like the `ext` right? If so, why are we calling it an app module and the single extension `ext`? In order to be consistent, let's either call the new field something like "allExtensions" or rename the `ext` into an `appModule`. I personally prefer the former

yigitozkavci (approved):
🚀

PR #3703, opened 2024-04-12
Bundling scripts

isaacroldan:
As part of the Global CLI project we are also introducing dependency bundling. This PR adds all the necessary scripts to bundle the CLI.

- Add bundle scripts for `cli` and `create-app`.
- Add some utilities shared by both bundle scripts.
- Add new `bundle` script in package.json and project.json for both packages.

THIS PR ONLY ADDS THE SCRIPTS, the bundling is not being triggered yet anywhere,

### How to test your changes?
Nothing to test.

### Measuring impact

### Checklist

### DIFF
--- a/packages/cli/bin/bundle.js
+++ b/packages/cli/bin/bundle.js
@@ -7,6 +7,10 @@ import {build as esBuild} from 'esbuild'
 import requireResolvePlugin from '@chialab/esbuild-plugin-require-resolve'
 import {copy} from 'esbuild-plugin-copy'
 import glob from 'fast-glob'
+import {joinPath} from '@shopify/cli-kit/node/path'
+import {createRequire} from 'module'
+
+const require = createRequire(import.meta.url)
 
 const external = [
   // react-devtools-core is a dev dependency, no need to bundle it but throws errors if not included here.
@@ -17,6 +21,10 @@ const external = [
 
 // yoga wasm file is not bundled by esbuild, so we need to copy it manually
 const yogafile = glob.sync('../../node_modules/.pnpm/**/yoga.wasm')[0]
+
+// Find theme-check-node's config yml files
+const themePath = require.resolve('@shopify/theme-check-node')
+const configYmlPath = joinPath(themePath, '..', '..', 'configs/*.yml')
 
 esBuild({
   bundle: true,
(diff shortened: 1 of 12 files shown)


packages/cli/bin/bundle.js
```diff
@@ -0,0 +1,73 @@
...
+        {
+          from: ['../cli-kit/assets/**/*'],
+          to: ['./dist/assets'],
+        },
+        {
+          from: [yogafile],
+          to: ['./dist/'],
+        },
+        {
+          from: ['../../node_modules/.pnpm/node_modules/@shopify/theme-check-node/configs/*.yml'],
```
shauns, line 65: This one feels a bit fragile -- any other ways we could do this? We can change `theme-check-node` if we need to

packages/create-app/bin/bundle.js
```diff
@@ -0,0 +1,40 @@
...
+
+const external = [
+  // react-devtools-core is a dev dependency, no need to bundle it but throws errors if not included here.
+  'react-devtools-core',
+]
+
+// yoga wasm file is not bundled by esbuild, so we need to copy it manually
+const yogafile = glob.sync('../../node_modules/.pnpm/**/yoga.wasm')[0]
+
+esBuild({
```
shauns, line 15: If I change the big CLI bundle script, I really need to change this one too (e.g. maybe we realise some extra plugin is needed)

I wonder if there is a way to organise these to minimise the chance of missing one of them?

packages/cli/bin/bundle.js
```diff
@@ -0,0 +1,73 @@
...
+  // react-devtools-core is a dev dependency, no need to bundle it but throws errors if not included here.
+  'react-devtools-core',
+  // esbuild can't be bundled per design
+  'esbuild',
+]
+
+// yoga wasm file is not bundled by esbuild, so we need to copy it manually
+const yogafile = glob.sync('../../node_modules/.pnpm/**/yoga.wasm')[0]
+
+esBuild({
```
shauns, line 21: I guess a subsequent PR will add a GH action that triggers (and checks??) bundling?

packages/create-app/bin/bundle.js
isaacroldan: If we added a shared script, we will still need specific config for each package (extra plugins, defines, external dependencies) so we end up with 3 script files and that is more risk of editing the wrong one maybe?

Not sure though. The create-app one just needs to solve stuff that comes from `cli-kit` (it's only dependency)

packages/cli/bin/bundle.js
isaacroldan: A subsequent script will add the shipit config to trigger the bundling.
I didn't add a GH action to validate that bundling works but will do :)

PR #3704, opened 2024-04-12
`create-app` changes for Global CLI

isaacroldan:
Shopify CLI is now Global by default, and `create-app` should reflect that during app creation.

- When running `npm create @shopify/app` we will check if the Global CLI is installed. If not, we'll prompt the user whether to install it or keep it as a local dependency.
- `@shopify/app` is never added to the project in any case from now on as it is bundled with CLI.

### How to test your changes?
You can see this working by:
- Uninstall Global CLI if you have it (`npm uninstall -g @shopify/cli`)
- Run `pnpm create-app`

It should prompt you to install the CLI, if you select to do so we will install it and it won't appear as a dependency in the app's package.json.

### Measuring impact

### Checklist

### DIFF
--- a/packages/create-app/src/prompts/init.ts
+++ b/packages/create-app/src/prompts/init.ts
@@ -1,11 +1,14 @@
 import {generateRandomNameForSubdirectory} from '@shopify/cli-kit/node/fs'
+import {installGlobalCLIIfNeeded} from '@shopify/cli-kit/node/is-global'
+import {PackageManager} from '@shopify/cli-kit/node/node-package-manager'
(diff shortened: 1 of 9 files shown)


packages/create-app/src/prompts/init.ts
```diff
@@ -152,6 +157,9 @@ const init = async (options: InitOptions): Promise<InitOutput> => {
 
   answers.template = selectedUrl || answers.template || defaults.template
 
+  const installed = await installGlobalCLIIfNeeded(options.packageManager)
```
shauns, line 160: This will install the global CLI right? It feels a bit weird that a prompt has such a side-effect

shauns (approved):

PR #3718, opened 2024-04-15
Webhook Subscription Specification

gracejychang:
https://github.com/Shopify/develop-app-management/issues/1722

See description in issue

### How to test your changes?
See description in issue

### Measuring impact

### Checklist

### DIFF
--- a/packages/app/src/cli/models/extensions/specification.ts
+++ b/packages/app/src/cli/models/extensions/specification.ts
@@ -64,6 +64,8 @@ export interface ExtensionSpecification<TConfiguration extends BaseConfigType = 
   transform?: (content: object) => object
   reverseTransform?: (content: object, options?: {flags?: Flag[]}) => object
   simplify?: (remoteConfig: SpecsAppConfiguration) => SpecsAppConfiguration
+  hasMultipleModuleConfig?: boolean
+  multipleModuleConfigPath?: string
 }
 
 /**
@@ -118,6 +120,8 @@ export function createExtensionSpecification<TConfiguration extends BaseConfigTy
     reverseTransform: spec.reverseTransform,
     simplify: spec.simplify,
     experience: spec.experience ?? 'extension',
+    hasMultipleModuleConfig: spec.hasMultipleModuleConfig ?? false,
+    multipleModuleConfigPath: spec.multipleModuleConfigPath,
   }
   return {...defaults, ...spec}
 }
@@ -136,6 +140,8 @@ export function createConfigExtensionSpecification<TConfiguration extends BaseCo
   appModuleFeatures?: (config?: TConfiguration) => ExtensionFeature[]
   transformConfig?: TransformationConfig | CustomTransformationConfig
   simplify?: SimplifyConfig
+  hasMultipleModuleConfig?: boolean
+  multipleModuleConfigPath?: string
 }): ExtensionSpecification<TConfiguration> {
   const appModuleFeatures = spec.appModuleFeatures ?? (() => [])
   return createExtensionSpecification({
@@ -147,6 +153,8 @@ export function createConfigExtensionSpecification<TConfiguration extends BaseCo
     transform: resolveAppConfigTransform(spec.transformConfig),
     reverseTransform: resolveReverseAppConfigTransform(spec.schema, spec.transformConfig),
     simplify: resolveSimplifyAppConfig(spec.simplify),
+    hasMultipleModuleConfig: spec.hasMultipleModuleConfig,
+    multipleModuleConfigPath: spec.multipleModuleConfigPath,
     experience: 'configuration',
   })
 }
--- a/packages/app/src/cli/models/extensions/specification.ts
+++ b/packages/app/src/cli/models/extensions/specification.ts
@@ -64,7 +64,7 @@ export interface ExtensionSpecification<TConfiguration extends BaseConfigType = 
   transform?: (content: object) => object
   reverseTransform?: (content: object, options?: {flags?: Flag[]}) => object
   simplify?: (remoteConfig: SpecsAppConfiguration) => SpecsAppConfiguration
-  hasMultipleModuleConfig?: boolean
+  extensionManagedInToml?: boolean
   multipleModuleConfigPath?: string
 }
 
@@ -120,7 +120,7 @@ export function createExtensionSpecification<TConfiguration extends BaseConfigTy
     reverseTransform: spec.reverseTransform,
     simplify: spec.simplify,
     experience: spec.experience ?? 'extension',
-    hasMultipleModuleConfig: spec.hasMultipleModuleConfig ?? false,
+    extensionManagedInToml: spec.extensionManagedInToml ?? false,
     multipleModuleConfigPath: spec.multipleModuleConfigPath,
   }
   return {...defaults, ...spec}
@@ -140,7 +140,7 @@ export function createConfigExtensionSpecification<TConfiguration extends BaseCo
   appModuleFeatures?: (config?: TConfiguration) => ExtensionFeature[]
   transformConfig?: TransformationConfig | CustomTransformationConfig
   simplify?: SimplifyConfig
-  hasMultipleModuleConfig?: boolean
+  extensionManagedInToml?: boolean
   multipleModuleConfigPath?: string
 }): ExtensionSpecification<TConfiguration> {
   const appModuleFeatures = spec.appModuleFeatures ?? (() => [])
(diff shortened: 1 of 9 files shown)


packages/app/src/cli/models/extensions/specifications/app_config_webhook.ts
```diff
@@ -73,11 +27,11 @@ export default appWebhooksSpec
 function simplifyWebhooks(remoteConfig: SpecsAppConfiguration) {
```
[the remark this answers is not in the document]
karenxie: updating the `webhooks` module to not handle subscriptions anymore is not going to be part of this PR, so I vote for leaving it for now. until we do that work

packages/app/src/cli/models/extensions/specifications/app_config_webhook.ts
```diff
@@ -73,11 +27,11 @@ export default appWebhooksSpec
...
   if (!remoteConfig.webhooks?.subscriptions) return remoteConfig
 
-  remoteConfig.webhooks.subscriptions = mergeWebhooks(remoteConfig.webhooks.subscriptions)
+  remoteConfig.webhooks.subscriptions = mergePrivacyAndNonPrivacyWebhooks(remoteConfig.webhooks.subscriptions)
   return remoteConfig
 }
 
-function mergeWebhooks(subscriptions: WebhookSubscription[]): WebhookSubscription[] {
+function mergePrivacyAndNonPrivacyWebhooks(subscriptions: WebhookSubscription[]): WebhookSubscription[] {
```
karenxie, line 34: nit: I'm not a huge fan of introducing the term `NonPrivacy` to our codebase. It's not something that we're currently using (that I'm aware of) and someone looking at this code would wonder what non-privacy webhooks are. it's not a huge improvement but wdyt about `mergeAllWebhooks`?

packages/app/src/cli/models/extensions/specifications/app_config_webhook_subscription.ts
```diff
@@ -0,0 +1,69 @@
...
+  forward: (content: object) => transformFromWebhookSubscriptionConfig(content),
+  reverse: (content: object) => transformToWebhookSubscriptionConfig(content),
+}
+
+const appWebhookSubscriptionSpec = createConfigExtensionSpecification({
+  identifier: WebhookSubscriptionSpecIdentifier,
+  schema: WebhooksSchema,
+  transformConfig: WebhookSubscriptionTransformConfig,
+  simplify: WebhookSimplifyConfig,
+  hasMultipleModuleConfig: true,
```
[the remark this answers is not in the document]
karenxie: what about `extensionManagedInToml`?

packages/app/src/cli/models/extensions/specifications/app_config_webhook_subscription.ts
```diff
@@ -0,0 +1,78 @@
+import {WebhookSimplifyConfig} from './app_config_webhook.js'
+import {WebhooksConfig} from './types/app_config_webhook.js'
+import {WebhooksSchema} from './app_config_webhook_schemas/webhooks_schema.js'
+import {CustomTransformationConfig, createConfigExtensionSpecification} from '../specification.js'
+import {getPathValue} from '@shopify/cli-kit/common/object'
+
+export const WebhookSubscriptionSpecIdentifier = 'webhook_subscription'
+
+type TransformedWebhookSubscription = {
```
karenxie, line 9: I think this name is a bit misleading. It implies that this is the type of the object after you run a transform function, but based on the code, it looks like it's the type of the config in `transformToWebhookSubscriptionConfig` before it's transformed?

packages/app/src/cli/models/extensions/specifications/app_config_webhook_subscription.ts
```diff
@@ -0,0 +1,78 @@
...
+  api_version: string
+  uri: string
+  topic: string
+  compliance_topics?: string[]
+  sub_topic?: string
+  include_fields?: string[]
+}
+
+function transformFromWebhookSubscriptionConfig(content: object) {
```
karenxie, line 18: I remember us getting quite confused here around what the returned object would look like. do you think it's worthwhile to add a return type here? and for the other transform function?

packages/app/src/cli/models/extensions/specifications/app_config_webhook_subscription.ts
```diff
@@ -0,0 +1,78 @@
...
+  */
+  const webhookSubscriptions = subscriptions.flatMap((subscription) => {
+    const {uri, topics, ...optionalFields} = subscription
+    if (topics)
+      return topics.map((topic) => {
+        return {api_version, uri, topic, ...optionalFields}
+      })
+  })
+
+  return webhookSubscriptions.length > 0 ? {subscriptions: webhookSubscriptions} : {}
```
karenxie, line 40: do we have a test for the empty subscriptions case? do we want one?

karenxie (approved):
looks good! just a few minor comments

packages/app/src/cli/models/extensions/specifications/app_config_webhook_subscription.ts
gracejychang: What do you think about `RemoteWebhookSubscriptionType` or `RemoteWebhookSubscription`? The `reverseTransform` function is called when we are [building the remote app configuration](https://github.com/Shopify/cli/blob/main/packages/app/src/cli/services/app/select-app.ts#L44).

packages/app/src/cli/models/extensions/specifications/app_config_webhook_subscription.ts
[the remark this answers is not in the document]
gracejychang: that's a much clearer name - I've changed it now :) thanks Karen!

packages/app/src/cli/models/extensions/specifications/app_config_webhook_schemas/webhook_subscription_schema.ts
```diff
@@ -0,0 +1,28 @@
...
+
+export const WebhookSubscriptionSchema = zod.object({
+  topics: zod
+    .array(zod.string({invalid_type_error: 'Values within array must be a string'}), {
+      invalid_type_error: 'Value must be string[]',
+    })
+    .optional(),
+  uri: zod.preprocess(removeTrailingSlash, UriValidation, {required_error: 'Missing value at'}),
+  sub_topic: zod.string({invalid_type_error: 'Value must be a string'}).optional(),
+  include_fields: zod.array(zod.string({invalid_type_error: 'Value must be a string'})).optional(),
```
isaacroldan, line 18: Is `zod` default messaging not clear enough when the type is wrong?

packages/app/src/cli/models/extensions/specification.ts
```diff
@@ -64,6 +64,8 @@ export interface ExtensionSpecification<TConfiguration extends BaseConfigType =
   transform?: (content: object) => object
   reverseTransform?: (content: object, options?: {flags?: Flag[]}) => object
   simplify?: (remoteConfig: SpecsAppConfiguration) => SpecsAppConfiguration
+  extensionManagedInToml?: boolean
+  multipleModuleConfigPath?: string
```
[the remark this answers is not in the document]
gracejychang: since we want webhooks to have multiple modules, it's marked as an `extension` type in its specification. But we don't want the DX on app toml to change, so the first one is to indicate if there is extension data in the app toml. The `loader` loads all the extensions, but then it gets filtered out downstream based on `extension` type or `configuration` type - webhooklike when we are [creating `ExtensionInstances`](https://github.com/Shopify/cli/compare/main...extensions-configured-in-toml#diff-973c578c6a6700bec941fe428e1f488a4252f9fe356e5d002c6ce9d234137302R459). The flag lets us treat it as a `configuration` type on the CLI side since that's linked to app toml logic.

Maybe I should figure out a better name for `multipleModuleConfigPath` 😅 .. in the next PR, we want to get the config value from `webhook_subscription` specification so we can produce a module per subscription so `multipleModuleConfigPath` helps us fetch that by indicating the path: prototype is [here](https://github.com/Shopify/cli/compare/main...extensions-configured-in-toml#diff-ab169fa3e666a282671308e6c2af8e814783ccbdbfd7fe343128700590833c89R299)

packages/app/src/cli/models/extensions/specifications/app_config_webhook_schemas/webhook_subscription_schema.ts
gracejychang: This was an existing schema that I moved out of the `webhooks` specification file to its own individual file, since it's shared across 3 specifications (`webhooks`, `privacy_compliance`, and now `webhook_subscription`). So I don't have context on how this was implemented 😬

packages/app/src/cli/models/extensions/specification.ts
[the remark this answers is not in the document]
isaacroldan: I still think we shouldn't need `extensionManagedInToml`, the specification shouldn't need to know where the data is stored. As an exercise, if we assume that the specification will be migrated to the API, would a field like this make sense in the API, or is it too specific for this use case?

The differenciation between configuration/extension might go away, and we can start breaking it in CLI: the app toml should be able to have non-config modules as well without the specification telling us so :) 

About `multipleModuleConfigPath` it's true is a complex use case 🤔 I would make the same thinking exercise, if we moved this field to the API, would it make sense or is it very specific for webhooks and CLI? Could it be used for other extensions as it is? I think we might need to rethink this before adding it to the spec :/ 

I can help with all this if needed! I just want to keep it clean instead of adding new properties to the specs.

PR #3725, opened 2024-04-16
Generate and deploy extensions

amcaplan:

Related to https://github.com/Shopify/develop-app-management/issues/1757
Fixes https://github.com/Shopify/develop-app-management/issues/1710

Allows generating and deploying extensions. Includes:

* Fetching extension specifications
* Stub-out of template specifications (those will be addressed in a later PR)
* Filling in UIDs in module TOMLs, and reading them for future deploys
* Matching local and remote modules via UID
* Pushing an extension bundle to cloud storage

(Generating a version already existed, now it'll work with extensions as well.)

Functions aren't expected to work yet, as uploading those artifacts is a bit different.

### How to test your changes?

1. Create an app
2. `shopify app generate extension` (generate a few kinds, but no functions)
3. `shopify app deploy` (confirm the diff looks correct)
4. `shopify app deploy` again (confirm the diff shows no changes)
5. Generate another extension, and deploy again (confirm the diff shows changes appropriately)
6. Confirm that all extensions TOMLs have UIDs

### Measuring impact

### Checklist

### DIFF
--- a/packages/app/src/cli/models/app/identifiers.ts
+++ b/packages/app/src/cli/models/app/identifiers.ts
@@ -52,6 +52,7 @@ export async function updateAppIdentifiers(
   systemEnvironment = process.env,
 ): Promise<AppInterface> {
   if (developerPlatformClient.supportsAtomicDeployments) {
+    // We can't update the TOML files in parallel because some extensions might share the same file
     for (const extension of app.allExtensions) {
       // eslint-disable-next-line no-await-in-loop
       await addUidToToml(extension)


packages/app/src/cli/services/generate/extension.ts
```diff
@@ -106,6 +116,11 @@ async function extensionInit(options: ExtensionInitOptions) {
         await uiExtensionInit(options)
         break
     }
+    const configFile = joinPath(options.directory, 'shopify.extension.toml')
+    if (await fileExists(configFile)) {
+      const tomlContents = await readFile(configFile)
+      await writeFile(configFile, `uid = "${randomUUID()}"\n${tomlContents}`)
```
amcaplan, line 122: Do we need UID at all? Can we use handle?

packages/app/src/cli/utilities/developer-platform-client/shopify-developers-client/graphql/create-app-version.ts
```diff
@@ -30,10 +30,6 @@ export const CreateAppVersionMutation = gql`
         }
         versionTag
       }
-      userErrors {
-        field
-        message
-      }
```
amcaplan, line 36: Revert this

gonzaloriestra (approved):
Great work! I've fixed a couple of issues when working against Partners, but I think it's ready to go.

packages/app/src/cli/services/deploy.ts
```diff
@@ -96,15 +101,25 @@ export async function deploy(options: DeployOptions) {
...
-              app.allExtensions.flatMap((ext) => ext.bundleConfig({identifiers, developerPlatformClient, apiKey})),
+            const appModules: BundleConfigWithSpecificationIdentifier[] = getArrayRejectingUndefined(
+              await Promise.all(
+                app.allExtensions.flatMap(async (ext): Promise<BundleConfigWithSpecificationIdentifier | undefined> => {
+                  const bundleConfig = await ext.bundleConfig({identifiers, developerPlatformClient, apiKey})
+                  if (bundleConfig) {
+                    return {
+                      ...bundleConfig,
+                      specificationIdentifier: developerPlatformClient.toExtensionGraphQLType(ext.graphQLType),
+                    }
```
isaacroldan, line 112: Why doesn't the `bundleConfig` function directly include this `specificationIdentifier` property instead of adding it here?

packages/app/src/cli/services/context/id-matching.ts
```diff
@@ -51,6 +52,58 @@ function matchByNameAndType(
   return {matched: validMatches, pending: {local: pendingLocal, remote: pendingRemote}}
 }
 
+/**
+ * Automatically match local and remote sources if they have the same type and UID
+ */
+function matchByUID(
```
isaacroldan, line 58: In theory we shouldn't need to match anything right? just deploy with whatever UID is defined in the file. 
Or is this just to build the `diff` view?

packages/app/src/cli/models/app/identifiers.ts
```diff
@@ -43,9 +47,23 @@ interface UpdateAppIdentifiersOptions {
...
   systemEnvironment = process.env,
 ): Promise<AppInterface> {
+  if (developerPlatformClient.supportsAtomicDeployments) {
+    await Promise.all(
+      app.realExtensions.map(async (extension) => {
+        if (extension.isUuidManaged() && (await fileExists(extension.configurationPath))) {
+          const tomlContents = await readFile(extension.configurationPath)
+          const extensionConfig = decodeToml(tomlContents)
+          if ('uid' in extensionConfig) return
+          const updatedTomlContents = tomlContents.replace(/(type\s*=\s*".*")/, `$1\nuid = "${extension.uid}"`)
```
isaacroldan, line 60: ah! nice solution to insert the uid after the type :)

packages/app/src/cli/models/app/loader.ts
```diff
@@ -508,7 +508,7 @@ class AppLoader {
         // Parse all extensions by merging each extension config with the global unified configuration.
         const configuration = await this.parseConfigurationFile(UnifiedSchema, configurationPath)
         const extensionsInstancesPromises = configuration.extensions.map(async (extensionConfig) => {
-          const mergedConfig = {...configuration, ...extensionConfig}
+          const mergedConfig = {uid: (obj as {uid?: string}).uid, ...configuration, ...extensionConfig}
```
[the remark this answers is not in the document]
gonzaloriestra: I'm taking a look at your comments right now, I had forgotten about it

packages/app/src/cli/models/app/identifiers.ts
```diff
@@ -45,19 +47,22 @@ interface UpdateAppIdentifiersOptions {
...
-    }),
-  )
+  if (developerPlatformClient.supportsAtomicDeployments) {
+    await Promise.all(
+      app.realExtensions.map(async (extension) => {
+        if (extension.isUuidManaged() && (await fileExists(extension.configurationPath))) {
+          const tomlContents = await readFile(extension.configurationPath)
+          const extensionConfig = decodeToml(tomlContents)
+          if ('uid' in extensionConfig) return
+          const updatedTomlContents = tomlContents.replace(/(type\s*=\s*".*")/, `$1\nuid = "${extension.uid}"`)
```
amcaplan, line 60: Seems worth ensuring that the key is `type` and not `something_type`

```suggestion
          const updatedTomlContents = tomlContents.replace(/^(\s*type\s*=\s*".*")/, `$1\nuid = "${extension.uid}"`)
```

packages/app/src/cli/models/app/identifiers.ts
```diff
@@ -45,19 +47,22 @@ interface UpdateAppIdentifiersOptions {
...
-        const tomlContents = await readFile(extension.configurationPath)
-        const extensionConfig = decodeToml(tomlContents)
-        if ('uid' in extensionConfig) return
-        await writeFile(extension.configurationPath, `uid = "${extension.uid}"\n${tomlContents}`)
-      }
-    }),
-  )
+  if (developerPlatformClient.supportsAtomicDeployments) {
+    await Promise.all(
+      app.realExtensions.map(async (extension) => {
```
amcaplan, line 55: Does this handle gracefully a case where multiple extensions are defined in one file? Seems like that would fail somehow.

gonzaloriestra: I don't think it would work. But neither the previous solution, with a single UID for several extensions that would be deployed separately, no?

We have never had yet a template with multiple extensions, so I'm not even sure if the production code will work. We have been trying to support it for months, but it's still unused, does it make sense to keep pushing? Or should we wait and try to make it work once we are sure that it will happen? cc: @isaacroldan

gonzaloriestra: Right. The thing is that the current approach is generating a single UUID per template. But let me try to improve it.

gonzaloriestra: I've finally made it by looking for the right handle of each extension, which is unique.

packages/app/src/cli/models/app/identifiers.ts
gonzaloriestra: Thanks for the suggestion! I've followed a similar approach.

packages/app/src/cli/services/context/id-matching.ts
gonzaloriestra: I think you are right, it doesn't make sense to ask for manual matching when the UID is different. But we still need to show the created/deleted ones in the diff. Let me refactor this. Cc: @amcaplan

packages/app/src/cli/models/app/identifiers.ts
```diff
@@ -78,6 +90,33 @@ export async function updateAppIdentifiers(
   return app
 }
 
+async function addUidToToml(extension: ExtensionInstance) {
+  if (!extension.isUUIDStrategyExtension || !(await fileExists(extension.configurationPath))) return
```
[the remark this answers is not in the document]
gonzaloriestra: Actually, I took that check from your [initial commit](https://github.com/Shopify/cli/pull/3725/commits/5f77e0b25c282a37e7e2d84ff6d3e333329cbfe2#diff-0a10d012835f02ea1aad2051e85a9d7a825cb2a88ee51dae0e07ed28c3b839ffR57)... But it's a mandatory field, so I'll remove it.

https://github.com/Shopify/cli/blob/c951e175db154da1512fcdd145ea77186f635c4a/packages/app/src/cli/models/extensions/extension-instance.ts#L45

packages/app/src/cli/api/graphql/all_app_extension_registrations.ts
```diff
@@ -55,6 +55,7 @@ export interface AllAppExtensionRegistrationsQueryVariables {
 
 export interface ExtensionRegistration {
   id: string
+  uid?: string
```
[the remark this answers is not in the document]
gonzaloriestra: Totally agree.

The thing is that the Partners request is directly returning the response (which matches this interface), while for App Management we are composing a response with this same interface, but a different response format.

Ideally, I guess we should have different interfaces for each API, and then transform them to RemoteSource or whatever to keep consistency while both API live. Another refactor to the list!

packages/app/src/cli/services/context/id-manual-matching.ts
```diff
@@ -20,13 +20,10 @@ export interface ManualMatchResult {
  * @param remote - The remote sources to match
  * @returns The result of the manual matching
  */
-export async function manualMatchIds(
-  options: {
-    local: LocalSource[]
-    remote: RemoteSource[]
-  },
-  remoteIdField: 'id' | 'uuid',
```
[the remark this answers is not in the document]
isaacroldan: Functions used `id` for the matching instead of `uuid`, since they moved to AMF they all use `uuid`.

PR #3728, opened 2024-04-16
Add CLI and predeploy validations for editor extension collection

oluwatimio:

Fixes https://github.com/Shopify/core-issues/issues/68330

Adding validations for Editor extension collection

1. There must be at least two extension in a collection

Screenshot 2024-04-16 at 9 51 06 AM

2. An extension collection must have a handle defined

Screenshot 2024-04-16 at 9 51 45 AM

3. Pre-deploy validations are as follows

- The app must contain an extension with the handle specified in the collection
- The editor extension collection can only contain ui_extensions (for now)
- The editor extension collection can only contain ui_extensions but no admin targets

Screenshot 2024-04-16 at 11 04 57 AM

### How to test your changes?

**I know these steps are complex, If you need help with Tophatting, you can also schedule a quick 10-15min on my calendar!**

**If you cannot generate / deploy because its stuck let me know, I may need to restart partners on my end.** You can also go to [spin control ](https://spin-control.shopify.io/instances/collections-ui.oluwatimi-owoturo.us.spin.dev) to resume my instance.

To have an extension code locally clone Shopify/customer-accounts-ui-extension-dev and also clone CLI locally

1. Clone CLI locally and switch to this branch `add-cli-validations`
2. Clone this repo locally https://github.com/shopify/customer-accounts-ui-extension-dev
3. You can create a new instance using `spin up customer-accounts-ui-extension-dev --name collections-ui` and switch your partners branch to `no-editor-gating` so you wont have to enable the beta flag or use the cli version editor extension collection supports
4. In the CLI repo generate an editor extension collection (under the Admin group) using this command. Feel free to add the flag --reset if its gives errors related to generating a new app. Make sure to change the path in the command to the path to your cloned `customer-accounts-ui-extension-dev` locally. Also make sure to change `spin=` to your own instance name created in step 3

```
SHOPIFY_CLI_VERSIONED_APP_CONFIG=1 SHOPIFY_SERVICE_ENV=spin SPIN_INSTANCE=collections-ui pnpm shopify app generate extension —verbose --path /Users/oluwatimiowoturo/src/github.com/Shopify/customer-accounts-ui-extension-dev
```

5. Generate a bunch of other extensions (an order-discount-function,  customer account ui extension and an admin block extension).

6. For your editor extension collection config, add a random handle inside the includes array that does not point to any extensions.
7. For your editor extension collection config, add a handle in includes array that points to the order discount function.
8. For your editor extension collection config, add a handle pointing to the customer account ui extension
9. For your editor extension collection config, add a handle pointing to the admin block extension
10. Try and deploy the extension using the following. Make sure to change the path in the command to the path to your cloned `customer-accounts-ui-extension-dev` locally

```
SHOPIFY_CLI_VERSIONED_APP_CONFIG=1 SHOPIFY_SERVICE_ENV=spin SPIN_INSTANCE=collections-ui pnpm shopify app deploy —verbose --path /Users/oluwatimiowoturo/src/github.com/Shopify/customer-accounts-ui-extension-dev
```

11. You should get similar errors to this 

Screenshot 2024-04-16 at 11 04 57 AM

12. Remove all handles in the collection except from one. Try and deploy and you should get this error

Screenshot 2024-04-16 at 9 51 06 AM

13. Remove the extension collection handle and try to deploy again. You should get the following error

Screenshot 2024-04-16 at 9 51 45 AM

14. To test the success path, generate another customer account extension and point to both ui_extensions targeting either checkout or customer account surface in the includes array of editor extension collection. Deploy the app. Deployment should succeed.

### Measuring impact

### Checklist

### DIFF
--- a/.changeset/clever-baboons-taste.md
+++ b/.changeset/clever-baboons-taste.md
@@ -1,0 +1,5 @@
+---
+'@shopify/cli': minor
+---
+
+Adds validations for editor extension collection


packages/app/src/cli/models/app/app.ts
```diff
@@ -276,6 +281,17 @@ export class App implements AppInterface {
...
     }
 
+    const extensionCollections = this.allExtensions.filter(
+      (ext) => ext.isEditorExtensionCollection,
+    ) as ExtensionInstance<EditorExtensionCollectionType>[]
+
+    if (extensionCollections.length > 0) {
+      const errors = validateExtensionsHandlesInCollection(extensionCollections, this.allExtensions)
+      if (errors) {
+        throw new AbortError('Invalid editor extension collection configuration', errors.join('\n'))
```
renerbaffa, line 291: could we add another `\n` here? I found it a bit harder to see more than one error altogether. Maybe it will be a bit easier to read them if we add an empty line in between?
Screenshot 2024-04-17 at 15 12 49

renerbaffa:
I just managed to see the success now (thanks @oluwatimio).

What I had to do was to leave only customer account and checkout ui extensions and delete the others. Timi was explaining that failures like those can happen when extensions fall out of sync with core because they added new fields for example.

I recommend doing the same when 🎩ing. Leave only customer account and checkout extension when deploying to see a success!

packages/app/src/cli/models/app/app.ts
```diff
@@ -343,6 +359,45 @@ export function validateFunctionExtensionsWithUiHandle(
...
+          `[${collection.handle}] editor extension collection - Local app must contain an extension with handle '${extension.handle}'`,
+        )
+      } else if (!allowableTypesForExtensionInCollection.includes(matchingExtension.specification.identifier)) {
+        errors.push(
+          `[${collection.handle}] editor extension collection - The collection can't contain an extension of type '${matchingExtension.specification.identifier}'`,
+        )
+      } else if (matchingExtension.specification.identifier === 'ui_extension') {
+        const uiExtension = matchingExtension as ExtensionInstance<UIExtensionType>
+        uiExtension.configuration.extension_points.forEach((extensionPoint) => {
+          if (extensionPoint.target.startsWith('admin.')) {
```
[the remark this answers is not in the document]
Johnkinging: This feels a little hacky tbh, and should probably be done with the core validators.

packages/app/src/cli/models/extensions/specifications/editor_extension_collection.ts
```diff
@@ -17,17 +18,33 @@ const EditorExtensionCollectionSchema = BaseSchema.extend({
...
 const editorExtensionCollectionSpecification = createExtensionSpecification({
   identifier: 'editor_extension_collection',
   schema: EditorExtensionCollectionSchema,
   appModuleFeatures: (_) => [],
+  validate: async (config, path) => {
+    const errors: string[] = []
+    const inCollection = makeExtensionsInCollection(config)
+
+    if (!config.handle) {
+      return err(`Editor extension collection with name ${config.name} must have a handle`)
```
Johnkinging, line 32: We'll need the length validator on core, but it would be nice to have it here too since these validations fire before it tries to deploy, and it's such an easy validation.

packages/app/src/cli/models/extensions/specifications/editor_extension_collection.ts
```diff
@@ -17,17 +18,33 @@ const EditorExtensionCollectionSchema = BaseSchema.extend({
...
+    }
+
+    if (inCollection.length < 2) {
+      errors.push(`There must be at least two extensions in editor extension collection ${config.handle}`)
+    }
+
```
Johnkinging, line 38: Is there any way to validate the extension handles that are included here? If we can't here we definitely will need to in core.

packages/app/src/cli/models/extensions/specifications/editor_extension_collection.ts
oluwatimio: I believe handle already has a length validator by default

packages/app/src/cli/models/app/app.ts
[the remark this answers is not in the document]
oluwatimio: I dont think this is hacky because all the admin targets start with `admin.` that being said we decided to repeat these validations on the backend too

packages/app/src/cli/models/extensions/specifications/editor_extension_collection.ts
oluwatimio: No we cant do this here. I added validation for that already here https://github.com/Shopify/cli/pull/3728/files#diff-cf7257d7c0c4111afc303b94c039388faf0f27ff4e121ce30f89db3d0c9f2d10R377

packages/app/src/cli/models/app/app.ts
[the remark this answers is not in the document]
Johnkinging: Yeah as long we're doing that then it's all good. I thought this was going to be the only line of defence here.

packages/app/src/cli/models/extensions/specifications/editor_extension_collection.ts
Johnkinging: Oh nice, and only 30 characters!

packages/app/src/cli/models/extensions/specifications/editor_extension_collection.ts
Johnkinging: Ah okay. I was just wondering.

Johnkinging (approved):
LGTM!

packages/app/src/cli/models/extensions/specifications/editor_extension_collection.ts
```diff
@@ -17,17 +18,33 @@ const EditorExtensionCollectionSchema = BaseSchema.extend({
...
+  validate: async (config, path) => {
+    const errors: string[] = []
+    const inCollection = makeExtensionsInCollection(config)
+
+    if (!config.handle) {
+      return err(`Editor extension collection with name ${config.name} must have a handle`)
+    }
+
+    if (inCollection.length < 2) {
+      errors.push(`There must be at least two extensions in editor extension collection ${config.handle}`)
```
oluwatimio, line 36: cc @deannatron error message here

packages/app/src/cli/models/app/app.ts
```diff
@@ -343,6 +359,45 @@ export function validateFunctionExtensionsWithUiHandle(
...
+            errors.push(
+              `[${collection.handle}] editor extension collection - The collection can't contain an extension of type '${matchingExtension.specification.identifier}' with target ${extensionPoint.target}`,
```
oluwatimio, line 390: cc @deannatron error message here

packages/app/src/cli/models/extensions/specifications/editor_extension_collection.ts
isaacroldan: You don't need this validation. instead of using `BaseSchema` use `BaseSchemaWithHandle`, then `zod` will take care of making sure there is a handle always.

packages/app/src/cli/models/extensions/specifications/editor_extension_collection.ts
```diff
@@ -17,17 +18,33 @@ const EditorExtensionCollectionSchema = BaseSchema.extend({
   type: zod.literal('editor_extension_collection'),
 })
 
+export type EditorExtensionCollectionType = zod.infer<typeof EditorExtensionCollectionSchema>
```
isaacroldan, line 21: If you define your schema like this:

```ts
const EditorExtensionCollectionSchema = BaseSchema.extend({
  include: zod.array(IncludeSchema).optional().default([]),
  includes: zod.array(zod.string()).optional().default([]),
  type: zod.literal('editor_extension_collection'),
}).transform((data) => {
  const includes = data.includes.map((handle) => {
    return {handle}
  })
  return {
    ...data,
    inCollection: [...includes, ...data.include],
  }
```

You can avoid having `makeExtensionsInCollection` and all the calls to it.

isaacroldan: You can also define validations at load-time in the schema like:

```ts
 .refine((data) => data.inCollection.length > 1, {
    message: 'There must be at least two extensions in editor extension collection',
  })
```

packages/app/src/cli/models/app/app.ts
```diff
@@ -276,6 +281,17 @@ export class App implements AppInterface {
       }
     }
 
+    const extensionCollections = this.allExtensions.filter(
+      (ext) => ext.isEditorExtensionCollection,
+    ) as ExtensionInstance<EditorExtensionCollectionType>[]
+
+    if (extensionCollections.length > 0) {
+      const errors = validateExtensionsHandlesInCollection(extensionCollections, this.allExtensions)
```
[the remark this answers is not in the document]
oluwatimio: Hey Isaac I am wondering what you mean by 

> Is there a reason we want this to happen at deploy time and not at app-load time?

We need to validate this here because we need access to all the extensions in the app to do the validations. This is also similar to what is being done for functions that reference ui extension handles such as [this](https://github.com/Shopify/cli/blob/dc377a363b276f712a1b1d81e0e68ea5f864ec1d/packages/app/src/cli/models/app/app.ts#L278) and I do recall discussing this with you and it was suggested this to be done in `preDeployValidation`. We need these validations [here](https://github.com/Shopify/cli/pull/3728/files#diff-cf7257d7c0c4111afc303b94c039388faf0f27ff4e121ce30f89db3d0c9f2d10R364) because we cannot allow a partner deploy invalid configurations for what should be in the collection.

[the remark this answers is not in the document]
isaacroldan: Oh I don't question the necessity of this, what I meant is why do we prefer to this here at `preDeployValidation` and not in the `loader.ts` when we load the extensions.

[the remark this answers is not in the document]
oluwatimio: Oh I'd also prefer being consistent because If I were to move this to the loader, we have functions doing validation in preDeployValidation and then we are doing something similar somewhere else.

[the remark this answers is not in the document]
oluwatimio: I also see the loader as doing validation on schema not doing validation on specific use cases for each spec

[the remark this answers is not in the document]
isaacroldan: That makes sense 👌

packages/app/src/cli/models/extensions/specifications/editor_extension_collection.ts
oluwatimio: Thanks! Ill try out this schema change

packages/app/src/cli/models/app/app.ts
deannatron: New error copy:

>[${collection.handle}]: Remove extension of type 
'${matchingExtension.specification.identifier}' with target ${extensionPoint.target}` from this collection. This extension type is not supported in collections.

packages/app/src/cli/models/extensions/specifications/editor_extension_collection.ts
deannatron: New error copy:

>${config.handle}: This editor extension collection must include at least 2 extensions

thitemple (approved):
🎩 and LGTM

packages/app/src/cli/models/app/app.ts
oluwatimio: Sorry Deanna. Here its not the type that is the issue, its just the target. Perhaps we can remove `This extension type is not supported in collections.` from this message?

deannatron: @oluwatimio actually that would be even more helpful. Like:

>[${collection.handle}]: Remove extension {extensionName} with target ${extensionPoint.target}` from this collection. This extension target is not supported in collections.

Ignore my syntax :D

packages/app/src/cli/models/extensions/specifications/editor_extension_collection.ts
oluwatimio: Hmm just tried it again, it works!

PR #3753, opened 2024-04-18
Expose silent option in renderTasks

jamesmengo:

Fixes #0000 

### Measuring impact

### Checklist

### DIFF
--- a/packages/cli-kit/src/public/node/ui.tsx
+++ b/packages/cli-kit/src/public/node/ui.tsx
@@ -499,6 +499,7 @@ export function renderTable<T extends ScalarDict>({renderOptions, ...props}: Ren
 
 interface RenderTasksOptions {
   renderOptions?: RenderOptions
+  silent?: boolean
 }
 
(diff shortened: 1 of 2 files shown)


packages/cli-kit/src/public/node/ui.tsx
```diff
@@ -499,6 +499,7 @@ export function renderTable<T extends ScalarDict>({renderOptions, ...props}: Ren
 
 interface RenderTasksOptions {
   renderOptions?: RenderOptions
```
jamesmengo, line 501: I would prefer to add this to the renderOptions, Seems like it'd be a more widespread change

jamesmengo, line 501: I explored overriding `stdout` in renderOptions (this comes from Ink), but I felt that it felt really hacky and required users to still pipe their output

If there's a way to point a `tty.writestream` to `/dev/null` I'd be open to exploring that as well!

PR #3797, opened 2024-04-26
Dynamic log prefixes for `app dev` output

nickwesselman:

Enables extension-specific prefixes in the `app dev` output. This improves the existing `app dev` output for extensions by allowing log prefixes to match the extension handle, which is already the case for `app build`. It also enables extension-specific prefixes in the forthcoming app/function log streaming.

* `ConcurrentOutput` now looks for a prefix stashed in an `AsyncLocalStorage`, which it will use instead of the process prefix.
* `ConcurrentOutput` allows for prefix column width to be set externally and not just on known processes.
* Theme extension process has been given a new prefix since it was the only thing left outputting `extensions`.
* Implements some UX recommendations on (less) use of color in `app dev` output to better confirm to our UX principles for Shopify CLI.

The colors used in `app dev` output weren't changed, but we might consider adding to them in a follow up PR.

#### Before

CLI app dev before extension prefixes

#### After

CLI app dev after extension prefixes

### How to test your changes?

The best test is running `app dev` against an app with many extensions. I have one here:
https://github.com/Shopify/nw-scratch/tree/many-extension-app

### Measuring impact

### Checklist


packages/cli-kit/src/private/node/ui/components/ConcurrentOutput.tsx
```diff
@@ -77,43 +82,84 @@ const ConcurrentOutput: FunctionComponent<ConcurrentOutputProps> = ({
...
+  }
+
+  const writableStream = (process: OutputProcess, prefixes: string[]) => {
+    return new Writable({
+      write(chunk, _encoding, next) {
+        const parsedLog : ParsedLog = parseLog(chunk.toString('utf8'))
+        const prefix = parsedLog.prefix ?? process.prefix
+        const index = addPrefix(prefix, prefixes)
+
+        const lines = stripAnsi(parsedLog.log.replace(/(\n)$/, '')).split(/\n/)
```
isaacroldan, line 139: So we are parsing before splitting lines, do we assume chunk === line?
Could we receive log lines from different functions in the same chunk?

nickwesselman: 🤔 I don't think so? Though we could add support for that later I think by looking for the prefix marker at the start of any line?

packages/cli-kit/src/private/node/ui/components/ConcurrentOutput.tsx
```diff
@@ -71,49 +77,95 @@ function currentTime() {
...
-  const prefixColumnSize = Math.max(...processes.map((process) => process.prefix.length))
-  const concurrentColors: TextProps['color'][] = useMemo(() => ['yellow', 'cyan', 'magenta', 'green', 'blue'], [])
-  const lineColor = useCallback(
-    (index: number) => {
-      const colorIndex = index < concurrentColors.length ? index : index % concurrentColors.length
-      return concurrentColors[colorIndex]!
-    },
-    [concurrentColors],
-  )
-  const writableStream = useCallback(
```
nickwesselman, line 89: @isaacroldan The use of `useCallback` here seemed unnecessary and I ended up removing it while I was iterating on state management for the seen prefixes. I could probably reintroduce it now, but is it actually needed?

packages/cli-kit/src/private/node/ui/components/ConcurrentOutput.tsx
```diff
@@ -71,49 +77,95 @@ function currentTime() {
  */
 const ConcurrentOutput: FunctionComponent<ConcurrentOutputProps> = ({
   processes,
+  prefixColumnSize = 12,
```
nickwesselman, line 80: What should our default minimum be?

packages/cli-kit/src/private/node/ui/components/ConcurrentOutput.tsx
```diff
@@ -71,49 +77,95 @@ function currentTime() {
...
-            },
-          ])
-          next()
-        },
-      })
-    },
-    [lineColor],
-  )
+
+  const concurrentColors = [
```
nickwesselman, line 88: 🥚 Easter egg idea: Override colors with an environment variable

nickwesselman:
Needs unit tests

nickwesselman:
Need to review other uses of ConcurrentOutput as well

nickwesselman:
A couple linting fixes for non-related work ended up in here as a result of a `lint:fix` as well.

packages/app/src/cli/services/dev/processes/draftable-extension.ts
```diff
@@ -37,24 +37,48 @@ export const pushUpdatesForDraftableExtensions: DevProcessFunction<DraftableExte
...
+      const extensionErr = extension.getPrefixedLogger(stderr)
+
+      await extension.build({
+        app,
+        stdout: extensionOut,
+        stderr: extensionErr,
+        useTasks: false,
+        signal,
+        environment: 'development',
+      })
```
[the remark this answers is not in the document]
nickwesselman: I tried to keep the insertion of the extension-specific logger at the same level as much as possible. In this case, if it was done inside build, `pushUpdatesForDraftableExtensions` would need to know to _not_ wrap stdio streams for `extension.build`, but would still need to do it for `updateExtensionDraft` and `setupExtensionWatcher`. This actually seems cleaner to me? But definitely subjective.

packages/cli-kit/src/private/node/ui/components/ConcurrentOutput.tsx
nickwesselman: Linting encouraged me to add it back... I believe that using it improves rendering performance.

packages/app/src/cli/models/extensions/extension-instance.ts
```diff
@@ -227,6 +229,15 @@ export class ExtensionInstance<TConfiguration extends BaseConfigType = BaseConfi
...
   }
 
+  getPrefixedLogger(wrappedWritable: Writable) {
+    const handle = this.handle
+    return new Writable({
+      write(chunk, _encoding, callback) {
+        wrappedWritable.write(`${prefixConcurrentOutputLog(handle, chunk.toString('utf8'))}`, 'utf8', callback)
+      },
+    })
+  }
```
isaacroldan, line 239: So we are creating multiple Writable objects, but in reality we are writing to the same one and using this as a "proxy". Does this has any downsides?

packages/app/src/cli/services/dev/update-extension.ts
```diff
@@ -64,7 +64,7 @@ export async function updateExtensionDraft({
     stderr.write(`Error while updating drafts: ${errors}`)
   } else {
     const draftUpdateSuccesMessage = extension.draftMessages.successMessage
-    if (draftUpdateSuccesMessage) outputInfo(draftUpdateSuccesMessage, stdout)
+    if (draftUpdateSuccesMessage) outputInfo(`${draftUpdateSuccesMessage}`, stdout)
```
isaacroldan, line 67: do you need this change if `draftUpdateSuccesMessage` is already a string?

packages/app/src/cli/services/dev/extension/payload/store.ts
```diff
@@ -37,7 +37,15 @@ export async function getExtensionsPayloadStoreRawPayload(
...
-    extensions: await Promise.all(options.extensions.map((extension) => getUIExtensionPayload(extension, options))),
+    extensions: await Promise.all(
+      options.extensions.map((extension) =>
+        getUIExtensionPayload(extension, {
+          ...options,
+          stdout: extension.getPrefixedLogger(options.stdout),
+          stderr: extension.getPrefixedLogger(options.stderr),
+        }),
+      ),
+    ),
```
isaacroldan, line 48: Why do we need to include this in the payload? The only place I could find that we need a stdout is in `getLocalization`, correct? seems weird to be injecting a custom stdout to they payload store just for that 🤔 

I think this is another case where for me it makes sense that the extension itself decides to use the custom Writable 🤔

packages/app/src/cli/services/dev/processes/draftable-extension.ts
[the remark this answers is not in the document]
nickwesselman: Resolved with the switch to AsyncLocalStorage.

packages/cli-kit/src/private/node/ui/components/ConcurrentOutput.tsx
nickwesselman: No longer relevant with the switch to AsyncLocalStorage.

packages/cli-kit/src/private/node/ui/components/ConcurrentOutput.tsx
```diff
@@ -36,6 +38,16 @@ function currentTime() {
   return `${hours}:${minutes}:${seconds}`
 }
 
+interface ConcurrentOutputContext {
+  outputPrefix: string
+}
```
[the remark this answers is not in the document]
nickwesselman: https://github.com/Shopify/develop-app-management/issues/1801

Issue #3715, opened 2024-04-15
[Bug]: `shopify theme dev` is losing track of the development theme


karreiro:
### In which of these areas are you experiencing a problem?

Theme

### Expected behavior

When users run `shopify theme dev`, a development theme should be created and assign as `[yours]` (your development theme should never be re-assigned).

### Actual behavior

When users run `shopify theme dev`, the development theme seems to be re-assigned and in some contexts the CLI is election even non-development themes.

sameerxanand:
Here's how I'm working around it:

```bash
shopify theme list
```

Find your development theme in the list

```bash
shopify theme delete -t {ID of your development theme}
```

Run `shopify theme dev` and it should use a new development theme instead of the live theme.

---

Agree this whole issue has been a pain in the a**. Shopify pushed a new CLI update today which now shows a new warning:

```
Warning: Could not find ts-node at /opt/homebrew/Cellar/shopify-cli/3.59.0/libexec/lib/node_modules/@shopify/cli/dist/index-e6f8190d.js. Please ensure that ts-node is a devDependency. Falling back to compiled source.
```

mgmanzella:
another example: https://github.com/Shopify/cli/issues/3794

karreiro:
Thank you for sharing that context in the email, @davidhollenbeckx! That not only helps the team understand the issue itself, but it also helps us understand how developers use CLI in their workflows.

For anyone experiencing this problem, you can adopt the following steps as a temporary workaround:
- Run `shopify theme push -u -t <name_of_your_dev_theme>` to create a new theme
- Run `shopify theme dev -t <name_of_your_dev_theme>` to force the CLI to use that specific theme in the development server

As soon as we get this fixed, we will update this issue.

Thanks everyone for reporting this problem.

karreiro:
Thank you, everyone, for reporting this problem and providing details.

This issue has been fixed by https://github.com/Shopify/cli/pull/3809, so upgrading to the latest release ([3.59.2](https://github.com/Shopify/cli/releases/tag/3.59.2)) fixes this problem.

nboliver-ventureweb:
@karreiro This is definitely not fixed. See here: https://shopifypartners.slack.com/archives/CKQTKDZV2/p1714579081291059

davidhollenbeckx:
@lukeh-shopify the issue will kick in after making a change to the theme and refreshing. there are loads of reproductions and logs in this thread.

[...]

PR #3798, opened 2024-04-26
[Themes][Bugfix] - Fix unpublished themes being marked as development themes

jamesmengo:
- https://github.com/Shopify/cli/issues/3715

I left some explanations about why this is happening in the diff view
- https://github.com/Shopify/cli/pull/3798/files#r1581467420
- https://github.com/Shopify/cli/pull/3798/files#r1581330822

#### 1) Prevents unpublished themes from being marked as `[unpublished][yours]` when creating a theme via `shopify theme push -u`

| Command | Before | After |
|--------|--------|--------|
| `-d` | ✅ `[development][yours]` | ✅ `[development][yours]` | 
| `-u`  | 🐛 `[unpublished][yours]` | ✅ `[unpublished]` |
| `-u -d` | 🐛`[unpublished][yours]` | ✅ `[unpublished]` | 

#### 2) Flushes the local storage for development themes 

image

- This will reset all development themes (which are ephemeral anyway), and can still be accessed

#### Change 1
1) `pnpm shopify theme push -u --path=<PATH>`
2) `pnpm shopify theme list` should show your new theme with `[unpublished]`
3) You can also try `pnpm shopify theme push -u -d --path=<PATH>`

#### Change 2
1) Create a development theme (either corrupt or working state)
2) Run `pnpm shopify theme list`
3) `[yours]` should be empty

BEFORE
image

AFTER
image

### Post-release steps
- Maybe this isn't the only problem - need to monitor to see if this resolves the issues in `dev`
- You may need to run `shopify theme delete` and delete the existing `[unpublished][yours]` theme if it exists
- https://github.com/Shopify/cli/pull/3802

### Measuring impact

### Checklist

### DIFF
--- a/packages/theme/src/cli/commands/theme/push.test.ts
+++ b/packages/theme/src/cli/commands/theme/push.test.ts
@@ -122,12 +122,40 @@ describe('Push', () => {
     test("renders confirmation prompt if 'allow-live' flag is not provided and selected theme role is live", async () => {
       // Given
       const flags: ThemeSelectionOptions = {live: true}
+      vi.mocked(renderConfirmationPrompt).mockResolvedValue(true)
 
       // When
-      await createOrSelectTheme(adminSession, flags)
+      const theme = await createOrSelectTheme(adminSession, flags)
 
       // Then
+      expect(theme?.role).toBe(LIVE_THEME_ROLE)
       expect(renderConfirmationPrompt).toHaveBeenCalled()
+    })
+
+    test("renders confirmation prompt if 'allow-live' flag is not provided and live theme is specified via theme flag", async () => {
+      // Given
+      const flags: ThemeSelectionOptions = {theme: '3'}
+      vi.mocked(findOrSelectTheme).mockResolvedValue(buildTheme({id: 3, name: 'Live Theme', role: LIVE_THEME_ROLE})!)
+      vi.mocked(renderConfirmationPrompt).mockResolvedValue(true)
+
+      // When
+      const theme = await createOrSelectTheme(adminSession, flags)
+
+      // Then
+      expect(theme?.role).toBe(LIVE_THEME_ROLE)
+      expect(renderConfirmationPrompt).toHaveBeenCalled()
+    })
+
+    test('returns undefined if live theme confirmation prompt is not confirmed', async () => {
+      // Given
+      const flags: ThemeSelectionOptions = {live: true}
+      vi.mocked(renderConfirmationPrompt).mockResolvedValue(false)
+
+      // When
+      const theme = await createOrSelectTheme(adminSession, flags)
+
+      // Then
+      expect(theme).toBeUndefined()
     })
 
     test('renders text prompt if unpublished flag is provided and theme flag is not provided', async () => {
--- a/packages/theme/src/cli/commands/theme/push.ts
+++ b/packages/theme/src/cli/commands/theme/push.ts
@@ -14,7 +14,7 @@ import {execCLI2} from '@shopify/cli-kit/node/ruby'
 import {AdminSession, ensureAuthenticatedThemes} from '@shopify/cli-kit/node/session'
 import {useEmbeddedThemeCLI} from '@shopify/cli-kit/node/context/local'
 import {RenderConfirmationPromptOptions, renderConfirmationPrompt} from '@shopify/cli-kit/node/ui'
-import {UNPUBLISHED_THEME_ROLE, promptThemeName} from '@shopify/cli-kit/node/themes/utils'
+import {LIVE_THEME_ROLE, Role, UNPUBLISHED_THEME_ROLE, promptThemeName} from '@shopify/cli-kit/node/themes/utils'
 import {cwd, resolvePath} from '@shopify/cli-kit/node/path'
 import {Theme} from '@shopify/cli-kit/node/themes/types'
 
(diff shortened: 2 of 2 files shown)


packages/theme/src/cli/commands/theme/push.ts
```diff
@@ -149,22 +151,24 @@ export default class Push extends ThemeCommand {
       return
     }
 
-    const developmentThemeManager = new DevelopmentThemeManager(adminSession)
```
jamesmengo, line 152: `DevelopmentThemeManager` marks all themes as development - any theme created via the push command will be marked as [yours].

To address this, I've created a `ThemeManager` and renamed the base class to `BaseThemeManager`

packages/cli-kit/src/public/node/themes/base-theme-manager.ts
```diff
@@ -5,7 +5,7 @@ import {BugError} from '@shopify/cli-kit/node/error'
 import {Theme} from '@shopify/cli-kit/node/themes/types'
 import {DEVELOPMENT_THEME_ROLE, Role} from '@shopify/cli-kit/node/themes/utils'
 
-export abstract class ThemeManager {
+export abstract class BaseThemeManager {
```
jamesmengo, line 8: Renamed this to a base class

packages/theme/src/cli/utilities/unpublished-theme-manager.ts
```diff
@@ -0,0 +1,24 @@
...
+  }
+
+  async create(themeRole?: Role, themeName?: string) {
+    const role = themeRole || UNPUBLISHED_THEME_ROLE
+    return super.create(role, themeName)
+  }
+
+  protected setTheme(themeId: string): void {
+    this.themeId = themeId
+  }
```
jamesmengo, line 19: Should I be storing this or doing anything with this?

Currently this class just allows us to call `create` WITHOUT setting the development theme, so there's no immediate use case that's popping out at me

packages/theme/src/cli/commands/theme/push.ts
```diff
@@ -149,31 +150,12 @@ export default class Push extends ThemeCommand {
...
-
     if (!flags.stable) {
-      const {live, development, unpublished, path, nodelete, theme, publish, json, force, ignore, only} = flags
```
jamesmengo, line 155: I've moved this logic into selectOrCreateTheme below

packages/theme/src/cli/commands/theme/push.ts
```diff
@@ -214,6 +198,41 @@ export default class Push extends ThemeCommand {
...
+  'allow-live'?: boolean
+}
+
+export async function createOrSelectTheme(
+  adminSession: AdminSession,
+  flags: ThemeSelectionOptions,
+): Promise<Theme | undefined> {
+  const {live, development, unpublished, theme} = flags
+
+  if (development) {
```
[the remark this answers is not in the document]
jamesmengo: The entire treatment of the `-u` `-d`, `--live`, and `-t` flags are pretty smelly imo
- `-u`, `-d`, and `--live` represent the role of the theme, so I feel like the combination of these two flags can make usage confusing. Should combining theme push to ALL of those themes?
  _- My initial thought is that it _should_ create 3 themes OR we need to specify the precedence of these flags in the docs_
- the `-t` flag changes meaning when you provide it in combination with `-u` and `-d`
  - `theme push -t themename` searches for a theme with the name `themename` and fails if it doesn't exist
  - `theme push -u -t themename` will create a theme with the name `themename` even if one already exists

This pr is not the place to apply these changes (in fact, it's kind of difficult to change the semantics of these flags without careful consideration), but I think the ergonomics of these flags can be fleshed out a little more :)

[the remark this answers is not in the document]
jamesmengo: Question: Do we even want to allow for creation of a development theme with `theme push`? @karreiro 

The `--stable` flag creates a new development theme when there is none that already exist **but it looks like that's coming from the TS wrapper**. May be off base here

packages/theme/src/cli/commands/theme/push.test.ts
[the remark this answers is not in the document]
jamesmengo: imo, that falls outside the scope of this class since [yours] is detected by the [theme selector](https://github.com/Shopify/cli/blob/e715ea94c5a6ecfa64b71639e02fe38aa9882c3f/packages/theme/src/cli/utilities/theme-selector.ts#L19-L41)

Ideally, these tests should only know that a development theme is created, but I'm going to add assertions on which `ThemeManager` is used to provide some extra clarity (_don't love this, but I think this balances effort/reward_)

Maybe time for some integration tests? wdyt?

packages/theme/src/cli/services/local-storage.ts
```diff
@@ -24,7 +24,7 @@ function themeLocalStorage() {
 function developmentThemeLocalStorage() {
   if (!_developmentThemeLocalStorageInstance) {
     _developmentThemeLocalStorageInstance = new LocalStorage<DevelopmentThemeLocalStorageSchema>({
-      projectName: 'shopify-cli-development-theme-conf',
+      projectName: 'shopify-cli-development-theme-config',
```
jamesmengo, line 27: flushes local storage key

packages/theme/src/cli/commands/theme/push.ts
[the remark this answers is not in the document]
karreiro: I don't think it's related to the Ruby implementation, as we may notice [here](https://github.com/Shopify/cli/blob/2588def0a766c0273570b5e6821a50181c9a189e/packages/cli-kit/assets/cli-ruby/lib/project_types/theme/commands/push.rb#L114-L139). The Ruby implementation adopts similar reasoning with the theme ID taking precedence.

The creation of the extra theme appears to be related to the development theme creation extracted to TypeScript via the [`--development-theme-id`](https://github.com/Shopify/cli/blob/3ecdbcd78a29d2fe71f254d38f88c70309eb1e14/packages/theme/src/cli/commands/theme/push.ts#L176) flag. Thus, this seems to be more like a problem in the TS <-> Ruby interoperability than on how each implementation selects the theme.

> Question: Do we even want to allow for creation of a development theme with theme push? @karreiro

Yes, keeping [backward compatibility](https://github.com/Shopify/cli/blob/3ecdbcd78a29d2fe71f254d38f88c70309eb1e14/packages/cli-kit/assets/cli-ruby/lib/project_types/theme/commands/push.rb#L125) in mind, it's a good idea to create a development theme.

karreiro (approved):
Thank you, @jamesmengo! I've left only some minor comments not blocking to merge this PR. Great stuff!

PR #3819, opened 2024-04-30
Fix auth for 1p development in Spin

gonzaloriestra:

Running dev with TAE in Spin was failing when trying to refresh the tokens:

```
The currently available CLI credentials are invalid.  
The CLI is currently unable to prompt for reauthentication.
Restart the CLI process you were running. If in an interactive terminal, you will be prompted to reauthenticate. If in a non-interactive terminal, ensure the correct credentials are available in the program environment. 
```

Original thread: https://shopify.slack.com/archives/C030LHFCA5T/p1714414263061029

Skip the error message thrown when the credentials are invalid and we can't run the auth flow, as it doesn't make sense for 1p dev in Spin.

### How to test your changes?

Try with `0.0.0-experimental-20240430140716` in Spin

### Measuring impact

### Checklist

### DIFF
--- a/.changeset/wet-paws-smoke.md
+++ b/.changeset/wet-paws-smoke.md
@@ -1,0 +1,5 @@
+---
+'@shopify/cli-kit': patch
+---
+
+Fix auth for 1p development in Spin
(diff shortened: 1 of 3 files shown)


isaacroldan:
@gonzaloriestra Can we also check if the user is in Spin? you can be 1P in prod too and I think we want to apply this exception only in spin.

isaacroldan (approved):

PR #3847, opened 2024-05-03
[Themes][Feature] Add file search to the theme selector

jamesmengo:
- Renders a prompt with the option for users to
  - use arrow keys to select
  - filter files by name 

### How to test your changes?
Requirements
- You need > 5 themes on your shop for the search to be enabled

Run `pnpm shopify theme push --path=<PATH_TO_THEME>`  

### Measuring impact

### Checklist


packages/theme/src/cli/utilities/theme-selector.ts
```diff
@@ -25,7 +25,7 @@ export async function findOrSelectTheme(session: AdminSession, options: {header?
     return filterThemes(store, themes, filter)[0]!
   }
 
-  return renderSelectPrompt({
+  return renderAutocompletePrompt({
```
jamesmengo, line 28: the parentMethod `findOrSelectTheme` is used in numerous commands, so this will propagate to all consumers

I don't personally see an issue with this but happy to h

karreiro: It's not an issue indeed. That's actually expected, so we may keep the DX for theme selection consistent across commands :)

karreiro (approved):
Thank you, @jamesmengo! Great stuff 🚀 LGTM and works as expected as well 🎩

PR #3852, opened 2024-05-06
[proposal] Theme Dev Preview - Initialize the ThemeEnvironment

karreiro:


packages/theme/src/cli/utilities/theme-environment.ts
```diff
@@ -2,28 +2,25 @@ import {uploadTheme} from './theme-uploader.js'
 import {AdminSession} from '@shopify/cli-kit/node/session'
 import {Checksum, Theme, ThemeFileSystem} from '@shopify/cli-kit/node/themes/types'
 
-export interface ThemeEnvironmentOptions {
+export interface DevServerSession extends AdminSession {
+  storefrontToken: string
```
karreiro, line 6: The development server will also needs the storefrontToken, so we could possibly have a single session object for the dev server

Issue #3863, opened 2024-05-07
[Bug]: Logging should be done to STDERR by default. STDOUT should be reserved for scriptable output. 


charlespwd:
### Please confirm that you have:

- [X] Searched [existing issues](.) to see if your issue is a duplicate. (If you’ve found a duplicate issue, feel free to add additional information in a comment on it.)
- [X] Reproduced the issue in [the latest CLI version](https://www.npmjs.com/package/@shopify/cli).

### In which of these areas are you experiencing a problem?

Other

### Expected behavior

I noticed that logging is done by default to STDOUT (even debug logging, for theme check anyway). Shouldn’t [that be sent to STDERR by default](https://clig.dev/#the-basics:~:text=Send%20messaging%20to%20stderr.%20Log%20messages%2C%20errors%2C%20and%20so%20on%20should%20all%20be%20sent%20to%20stderr.%20This%20means%20that%20when%20commands%20are%20piped%20together%2C%20these%20messages%20are%20displayed%20to%20the%20user%20and%20not%20fed%20into%20the%20next%20command)?

Logging to STDOUT when it isn’t meant to be captured by a script makes the CLI non-composable/non-scriptable.

For instance, `shopify theme check --verbose --output json` will have non-JSON output in STDOUT and makes `| jq` break.

IMO STDOUT should be intentional, not the default.

### Actual behavior

IMO all of these should use STDERR by default. STDOUT is the wrong default. 

- [outputDebug](https://github.com/Shopify/cli/blob/6a99d8e3ed40be85e2eb9324f6a29631ec044ec9/packages/cli-kit/src/public/node/output.ts#L300) should log to STDERR by default (console.error)
- outputInfo
- outputWarn

### Verbose output

```shell
$ shopify theme check --output json --verbose | jq .
jq: parse error: Invalid numeric literal at line 1, column 14
```

### Reproduction steps

1. `shopify theme check --output json --verbose | jq .` -- get malformed JSON because debug logging isn't JSON. 

### Operating System

All

### Shopify CLI version (check your project's `package.json` if you're not sure)

3.60

knjshimi:
+1

gonzaloriestra:
I've created a draft PR with a potential solution for this: https://github.com/Shopify/cli/pull/4922, in case someone wants to continue

charlespwd:
I would even suggest having a _new_ output helper named `output` that would be the only one that logs to STDOUT.

That way you can pick from the list:
[...]

PR #3865, opened 2024-05-07
[stable] Include `theme-check-docs-updater` data folder in final bundled dist

isaacroldan:
There is a bug in theme check where some expected assets can't be found. This is because on the bundled CLI we need to explicitly copy any asset folder.

- Copy the `data` folder from `theme-check-data-updater` into the final bundled `dist` folder.

### Measuring impact

### Checklist

### DIFF
--- a/packages/theme/src/cli/commands/theme/check.ts
+++ b/packages/theme/src/cli/commands/theme/check.ts
@@ -15,7 +15,7 @@ import {
 import ThemeCommand from '../../utilities/theme-command.js'
 import {Flags} from '@oclif/core'
 import {globalFlags} from '@shopify/cli-kit/node/cli'
-import {consoleError, outputDebug, outputInfo} from '@shopify/cli-kit/node/output'
+import {outputInfo, consoleError} from '@shopify/cli-kit/node/output'
 import {renderInfo, renderSuccess} from '@shopify/cli-kit/node/ui'
 import {themeCheckRun, LegacyIdentifiers} from '@shopify/theme-check-node'
 import {findPathUp} from '@shopify/cli-kit/node/fs'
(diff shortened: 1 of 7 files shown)


packages/theme/src/cli/commands/theme/check.ts
```diff
@@ -133,7 +133,7 @@ export default class Check extends ThemeCommand {
       return
     }
 
-    const {offenses, theme} = await themeCheckRun(path, config)
+    const {offenses, theme} = await themeCheckRun(path, config, (message) => outputDebug(message))
```
charlespwd, line 136: I have two choices here:
1. Use `outputDebug` (which is nice because it does all the colouring/timestamp/--verbose/SHOPIFY_FLAG_VERBOSE good stuff)
2. Use `console.error` directly

The problem with (1) is that, because of #3863, I can't really use it unless I change the default logger of outputDebug to be `consoleError`. I wouldn't be able to use it because Shopify/theme-check-action would capture STDOUT of `shopify theme check --output json --verbose` and expect JSON. But since outputDebug dumps to STDOUT, we'd have malformed JSON to deal with. 

(2) is a more local fix if you want to make the default change in a different PR.

packages/cli-kit/src/public/node/output.ts
```diff
@@ -297,7 +297,7 @@ export function outputCompleted(content: OutputMessage, logger: Logger = console
  * @param content - The content to be output to the user.
  * @param logger - The logging function to use to output to the user.
  */
-export function outputDebug(content: OutputMessage, logger: Logger = consoleLog): void {
+export function outputDebug(content: OutputMessage, logger: Logger = consoleError): void {
```
charlespwd, line 300: This is the default change. It feels safe. But we might want to make that in a different PR.

PR #3875, opened 2024-05-09
payment apps: add support for supported buyer contexts

punkstar:

Fixes #0000 

### Measuring impact

### Checklist


marklevi (changes requested):
LGTM overall ! 
It would be great to also spin test the `pnpm shopify app import-extensions` flow ensuring that we are also importing extensions appropriately.
You could also include some tests here: packages/app/src/cli/services/payments/extension-to-toml.test.ts

marklevi:
@punkstar something to also keep in mind is that the minimum required CLI version for payments app extension support is `v3.60`. These changes would only be applicable to future CLI version releases (soonest being May 27th)

packages/app/src/cli/models/extensions/specifications/payments_app_extension_schemas/base_payments_app_extension_schema.ts
```diff
@@ -12,6 +12,7 @@ export interface BasePaymentsAppExtensionDeployConfigType {
   merchant_label: string
   supported_countries: string[]
   supported_payment_methods: string[]
+  supported_buyer_contexts?: {currency: string; countries?: [string, ...string[]]}[]
```
lacoursieresimon, line 15: Am I understanding correctly that this type definition enforces the non-empty condition? (`countries?: [string, ...string[]]` versus a simpler `string[]`)

If so, do we also need the `nonempty()` check at line 70? I guess it doesn't hurt either 🤷

packages/app/src/cli/models/extensions/specifications/payments_app_extension_schemas/custom_credit_card_payments_app_extension_schema.test.ts
```diff
@@ -150,6 +155,7 @@ describe('customCreditCardPaymentsAppExtensionDeployConfig', () => {
       supports_3ds: config.supports_3ds,
       supported_countries: config.supported_countries,
       supported_payment_methods: config.supported_payment_methods,
+      supported_buyer_contexts: config.supported_buyer_contexts,
```
lacoursieresimon, line 158: Should we also test the mapping for `customCreditCardDeployConfigToCLIConfig` ?

packages/app/src/cli/models/extensions/specifications/payments_app_extension_schemas/base_payments_app_extension_schema.ts
lacoursieresimon: If we want to make `supported_buyer_contexts` required at some point later, are we able to do so or is making it nullable today forcing lifetime support of a nullable value?

lacoursieresimon (approved):
little context on CLI, but it does seem to make sense overall 👍

packages/app/src/cli/models/extensions/specifications/payments_app_extension_schemas/base_payments_app_extension_schema.ts
punkstar: > Am I understanding correctly that this type definition enforces the non-empty condition? (`countries?: [string, ...string[]]` versus a simpler `string[]`)

Correct!

> If so, do we also need the `nonempty()` check at line 70? I guess it doesn't hurt either 🤷

We need both. `SupportedBuyerContextsSchema` is the schema definition that will validate and parse input (by calling `#parse` on it). `BasePaymentsAppExtensionDeployConfigType` is just the type definition and the output of the schema. We can compute one from the type from the schema (`type SchemaType  = zod.infer<typeof Schema>`) but there must be a reason we don't!

> If we want to make `supported_buyer_contexts` required at some point later, are we able to do so or is making it nullable today forcing lifetime support of a nullable value?

I don't think we have the concept of config schema versions right now, so we'd likely have to introduce a new pattern to support this. It may already be used by other parts of the CLI but we're not making an irreversible decision.

punkstar:
Thanks for the feedback @marklevi! I've added some tests around the TOML generation to cover the new field and I've top hatted the import extension command. All looks fine!

marklevi (approved):
💪

PR #3877, opened 2024-05-09
[Refactor][Themes] Move local disk 'delete' and 'write' operations  behind the ThemeFileSystem interface

jamesmengo:
**WHY are these changes introduced?**
- Mise en place for adding polling for theme editor changes in the `theme dev` command
- Part of https://github.com/Shopify/develop-advanced-edits/issues/203

**WHAT is this pull request doing?**
- Adds `delete` and `write` methods to the `ThemeFileSystem` interface
- replaces the `removeThemeFile` and `writeThemeFile` functions with these new methods
- creates a mockThemeFileSystem fixture function for testing
- updates the tests accordingly

**How to test your changes?**
This is a refactor so no behaviour should change
`pnpm test`

### DIFF
--- a/packages/cli-kit/src/public/node/themes/types.ts
+++ b/packages/cli-kit/src/public/node/themes/types.ts
@@ -16,6 +16,20 @@ export interface ThemeFileSystem {
    * Local theme files.
    */
   files: Map<Key, ThemeAsset>
+
+  /**
+   * Removes a file from the theme file system
+   *
+   * @param assetKey - the key of the file to remove
+   */
+  delete: (asset: ThemeAsset) => Promise<void>
+
+  /**
+   * Writes a file to the theme file system
+   *
+   * @param asset - the asset to write
+   */
+  write: (asset: ThemeAsset) => Promise<void>
 }
 
 /**
--- a/packages/cli-kit/src/public/node/themes/types.ts
+++ b/packages/cli-kit/src/public/node/themes/types.ts
@@ -18,18 +18,27 @@ export interface ThemeFileSystem {
   files: Map<Key, ThemeAsset>
 
   /**
-   * Removes a file from the theme file system
+   * Removes a file from the local disk and updates the themeFileSystem
    *
    * @param assetKey - The key of the file to remove
    */
   delete: (assetKey: string) => Promise<void>
 
   /**
-   * Writes a file to the theme file system
+   * Writes a file to the local disk and updates the themeFileSystem
    *
    * @param asset - The ThemeAsset representing the file to write
    */
   write: (asset: ThemeAsset) => Promise<void>
+
+  /**
+   * Reads a file from the local disk and updates the themeFileSystem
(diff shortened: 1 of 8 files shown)


packages/theme/src/cli/utilities/theme-fs.ts
```diff
@@ -65,6 +65,14 @@ export async function mountThemeFileSystem(root: string): Promise<ThemeFileSyste
...
+    delete: async (asset: ThemeAsset) => {
+      await removeThemeFile(root, asset.key)
+      files.delete(asset.key)
+    },
+    write: async (asset: ThemeAsset) => {
+      await writeThemeFile(root, asset)
+      files.set(asset.key, asset)
+    },
   }
 }
```
jamesmengo, line 77: todo: consider making writeThemeFile and removeThemeFile private so that there is a single entrypoint

packages/theme/src/cli/utilities/theme-fs.ts
```diff
@@ -65,10 +65,18 @@ export async function mountThemeFileSystem(root: string): Promise<ThemeFileSyste
...
 
-export async function writeThemeFile(root: string, {key, attachment, value}: ThemeAsset) {
+async function writeThemeFile(root: string, {key, attachment, value}: ThemeAsset) {
```
jamesmengo, line 79: This is being changed to `private` to avoid bugs where we remove the file but forget to update the `ThemeFileSystem`

Previously, users would have to explicitly call this, which is easy to forget
```js
await writeThemeFile(root, asset)
themeFileSystem.files.set(asset.key, asset)
```

packages/cli-kit/src/public/node/themes/types.ts
```diff
@@ -16,6 +16,20 @@ export interface ThemeFileSystem {
...
+   * @param assetKey - The key of the file to remove
+   */
+  delete: (assetKey: string) => Promise<void>
+
+  /**
+   * Writes a file to the theme file system
+   *
+   * @param asset - The ThemeAsset representing the file to write
+   */
+  write: (asset: ThemeAsset) => Promise<void>
```
jamesmengo, line 32: todo: add read

packages/theme/src/cli/utilities/theme-fs/theme-fs-mock-factory.ts
```diff
@@ -0,0 +1,14 @@
+import {ThemeAsset, ThemeFileSystem} from '@shopify/cli-kit/node/themes/types'
+
+export function mockThemeFileSystem(root: string, files: Map<string, ThemeAsset>): ThemeFileSystem {
```
karreiro, line 3: I'd personally rename this function to `fakeThemeFileSystem` to communicate even better what it does :) 🚀

karreiro (commented):
Thank you for this PR, @jamesmengo! Excellent stuff 🔥 apart from the `read` API, I've also left a minor suggestion :)

PR #3892, opened 2024-05-13
Fix webhook link logic

isaacroldan:
Fixes https://github.com/Shopify/develop-app-management/issues/1792

There are two ways of defining webhooks in the toml:
- Condensed: multiple topics in the same subscription definition (sharing the same URI)
- Expanded: each subscription has only 1 topic

When linking to a new app we need to default to the condensed state, but right now we are doing expanded.

- Remove the `simplify` function from the specification
- Update the webhook reverse transform to transform from expanded to condensed.
- Add an additional transform to the `link` command to further condense webhooks

### How to test your changes?
See the description in this [issue](https://github.com/Shopify/develop-app-management/issues/1792)

### Measuring impact

### Checklist

### DIFF
--- a/packages/app/src/cli/models/app/loader.test.ts
+++ b/packages/app/src/cli/models/app/loader.test.ts
@@ -1962,11 +1962,7 @@ wrong = "property"
           api_version: '2024-01',
           subscriptions: [
             {
-              topics: ['orders/create'],
-              uri: 'https://example.com',
-            },
-            {
-              topics: ['orders/delete'],
+              topics: ['orders/create', 'orders/delete'],
               uri: 'https://example.com',
             },
           ],
@@ -2867,27 +2863,9 @@ describe('WebhooksSchema', () => {
       ],
     }
 
-    const expectedParsedConfig = {
-      api_version: '2021-07',
-      subscriptions: [
-        {
-          uri: 'arn:aws:events:us-west-2::event-source/aws.partner/shopify.com/1234567890/SOME_PATH',
-          topics: ['products/create'],
-        },
-        {
-          uri: 'arn:aws:events:us-west-2::event-source/aws.partner/shopify.com/1234567890/SOME_PATH',
-          topics: ['products/update'],
-        },
-        {uri: 'https://example.com', topics: ['products/create']},
-        {uri: 'https://example.com', topics: ['products/update']},
-        {uri: 'pubsub://my-project-123:my-topic', topics: ['products/create']},
-        {uri: 'pubsub://my-project-123:my-topic', topics: ['products/update']},
-      ],
-    }
-
     const {abortOrReport, parsedConfiguration} = await setupParsing({}, webhookConfig)
     expect(abortOrReport).not.toHaveBeenCalled()
-    expect(parsedConfiguration.webhooks).toMatchObject(expectedParsedConfig)
+    expect(parsedConfiguration.webhooks).toMatchObject(webhookConfig)
   })
 
   test('throws an error if we have duplicate subscriptions in same topics array', async () => {
@@ -2902,7 +2880,7 @@ describe('WebhooksSchema', () => {
       code: zod.ZodIssueCode.custom,
       message: 'You can’t have duplicate subscriptions with the exact same `topic`, `uri` and `filter`',
       fatal: true,
-      path: ['webhooks', 'subscriptions', 1, 'topics', 0, 'products/create'],
+      path: ['webhooks', 'subscriptions', 0, 'topics', 1, 'products/create'],
     }
 
     const {abortOrReport, expectedFormatted} = await setupParsing(errorObj, webhookConfig)
@@ -2921,7 +2899,7 @@ describe('WebhooksSchema', () => {
       code: zod.ZodIssueCode.custom,
       message: 'You can’t have duplicate subscriptions with the exact same `topic`, `uri` and `filter`',
       fatal: true,
-      path: ['webhooks', 'subscriptions', 2, 'topics', 0, 'products/create'],
+      path: ['webhooks', 'subscriptions', 0, 'topics', 2, 'products/create'],
     }
 
     const {abortOrReport, expectedFormatted} = await setupParsing(errorObj, webhookConfig)
@@ -2941,11 +2919,8 @@ describe('WebhooksSchema', () => {
     const expectedParsedConfig = {
       api_version: '2021-07',
       subscriptions: [
-        {uri: 'https://example.com', topics: ['products/create']},
-        {uri: 'https://example.com', topics: ['products/update']},
-        {uri: 'https://example.com2', topics: ['products/create']},
-        {uri: 'https://example.com2', topics: ['products/update']},
-        {uri: 'https://example.com', topics: ['products/delete']},
+        {uri: 'https://example.com', topics: ['products/create', 'products/update', 'products/delete']},
+        {uri: 'https://example.com2', topics: ['products/create', 'products/update']},
       ],
     }
 
@@ -3026,7 +3001,7 @@ describe('WebhooksSchema', () => {
       code: zod.ZodIssueCode.custom,
       message: 'You can’t have duplicate subscriptions with the exact same `topic`, `uri` and `filter`',
       fatal: true,
-      path: ['webhooks', 'subscriptions', 1, 'topics', 0, 'products/create'],
+      path: ['webhooks', 'subscriptions', 0, 'topics', 1, 'products/create'],
     }
 
     const {abortOrReport, expectedFormatted} = await setupParsing(errorObj, webhookConfig)
@@ -3051,7 +3026,7 @@ describe('WebhooksSchema', () => {
       code: zod.ZodIssueCode.custom,
       message: 'You can’t have duplicate subscriptions with the exact same `topic`, `uri` and `filter`',
       fatal: true,
-      path: ['webhooks', 'subscriptions', 1, 'topics', 0, 'products/create'],
+      path: ['webhooks', 'subscriptions', 0, 'topics', 1, 'products/create'],
     }
 
     const {abortOrReport, expectedFormatted} = await setupParsing(errorObj, webhookConfig)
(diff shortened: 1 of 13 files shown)


packages/app/src/cli/models/extensions/specifications/transform/app_config_webhook.ts
```diff
@@ -24,12 +24,96 @@ export function transformToWebhookConfig(content: object) {
...
+  return subscriptions.sort((oneSub, twoSub) => oneSub.uri.localeCompare(twoSub.uri))
+}
+function sortTopics(subscriptions: WebhookSubscription[]) {
+  const topicsSortedSubscriptions = subscriptions.map((subscription) => {
+    const sortedTopics = sortArrayAlphabetically(subscription.topics)
+    const {topics, ...rest} = subscription
+    return {topics: sortedTopics, ...rest}
+  })
+  return topicsSortedSubscriptions
+}
```
isaacroldan, line 74: I think you can do:

```ts
subscriptions.forEach((subscription) => {
  subscription.topics = sortArrayAlphabetically(subscription.topics)
})
```
Which is simpler (same for `sortComplianceTopics`)

packages/app/src/cli/models/extensions/specifications/transform/app_config_webhook.ts
```diff
@@ -24,12 +24,96 @@ export function transformToWebhookConfig(content: object) {
...
+  if (subscriptions.length === 0) return
+  const topicSubscriptions = subscriptions
+    .filter((subscription) => subscription.topics !== undefined)
+    .map((subscription) => {
+      // eslint-disable-next-line @typescript-eslint/naming-convention
+      const {compliance_topics, ...rest} = subscription
+      return {topics: subscription.topics, ...rest}
+    })
+  const complianceSubscriptions = subscriptions
+    .filter((subscription) => subscription.topics === undefined || subscription.compliance_topics !== undefined)
```
karenxie, line 47: why do we want to check `subscription.topics === undefined`? is checking the `compliance_topics` not sufficient?

karenxie (approved):
LGTM! thanks for getting this resolved @gracejychang!

packages/app/src/cli/models/extensions/specifications/transform/app_config_webhook.ts
```diff
@@ -24,12 +24,96 @@ export function transformToWebhookConfig(content: object) {
   const serverWebhooks = getPathValue(content, 'subscriptions') as NormalizedWebhookSubscription[]
   if (!serverWebhooks) return webhooks
 
-  const webhooksSubscriptions: WebhooksConfig['subscriptions'] = []
+  const subscriptions = serverWebhooks.map((subscription) => {
+    const {topic, ...otherFields} = subscription
+    return {...otherFields, topics: [subscription.topic]} as WebhookSubscription
+  })
```
alexanderMontague, line 30: is the `topic` destructure but using `subscription.topic` and is the spread ordering important?

```suggestion
  const subscriptions = serverWebhooks.map(({topic, ...otherFields}) => {
    return {...otherFields, topics: [topic]} as WebhookSubscription
  })
```

packages/app/src/cli/models/extensions/specifications/transform/app_config_webhook.ts
```diff
@@ -24,12 +24,96 @@ export function transformToWebhookConfig(content: object) {
...
-  for (const {topic, ...otherFields} of serverWebhooks) {
-    webhooksSubscriptions.push({topics: [topic], ...otherFields})
-  }
-
-  const webhooksSubscriptionsObject = webhooksSubscriptions.length > 0 ? {subscriptions: webhooksSubscriptions} : {}
+  const webhooksSubscriptionsObject = webhooksSubscriptions ? {subscriptions: webhooksSubscriptions} : {}
   return deepMergeObjects(webhooks, {webhooks: webhooksSubscriptionsObject})
 }
+
+export function mergeAllWebhooks(subscriptions: WebhookSubscription[]): WebhookSubscription[] | undefined {
```
alexanderMontague, line 37: It might be a good idea to comment where and why this transformer is used... just looking at the above entry point, a lot of these code paths seem impossible. When they are used on the compliance side and initial schema validation, they could be

packages/app/src/cli/models/extensions/specifications/transform/app_config_webhook.ts
alexanderMontague, line 47: do you need both of these conditions?

packages/app/src/cli/models/extensions/specifications/transform/app_config_webhook.ts
```diff
@@ -24,12 +24,96 @@ export function transformToWebhookConfig(content: object) {
...
+
+export function reduceWebhooks(
+  subscriptions: WebhookSubscription[],
+  property?: keyof Pick<WebhookSubscription, 'topics' | 'compliance_topics'>,
+) {
+  return subscriptions.reduce((accumulator, subscription) => {
+    const existingSubscription = findSubscription(accumulator, subscription)
+    if (existingSubscription) {
+      if (property) {
+        ;(existingSubscription[property] ?? []).push(...(subscription[property] ?? []))
```
alexanderMontague, line 103: this is odd... syntax and all. what would you be `push`ing into an empty array for?

```suggestion
      if (property && existingSubscription[property].length) {
        ;(existingSubscription[property] ?? []).push(...(subscription[property] ?? []))
```

should help... can we also use [concat](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/concat) here or is the spread supposed to be some special overwriting under the hood?

packages/app/src/cli/models/extensions/specifications/transform/app_config_webhook.ts
```diff
@@ -24,12 +24,96 @@ export function transformToWebhookConfig(content: object) {
...
+      } else {
+        if (subscription.topics) {
+          existingSubscription.topics ??= []
+          existingSubscription.topics.push(...subscription.topics)
```
alexanderMontague, line 107: concat and below if possible

packages/app/src/cli/services/app/config/link.ts
```diff
@@ -118,7 +124,17 @@ export default async function link(options: LinkOptions, shouldRenderSuccess = t
...
+}
+
+function condenseComplianceAndNonComplianceWebhooks(config: CurrentAppConfiguration): CurrentAppConfiguration {
+  const webhooksConfig = (config.webhooks as WebhooksConfig) ?? {}
+  const subscriptionsArr = webhooksConfig.subscriptions ?? []
+  if (subscriptionsArr.length === 0) return config
+
+  const condensedSubscriptionsArr = reduceWebhooks(subscriptionsArr)
+  webhooksConfig.subscriptions = condensedSubscriptionsArr
+  return config
```
alexanderMontague, line 137: The assignment and overwriting the underlying value is pretty confusing

```suggestion
  if (!config?.webhooks?.subscriptions?.length) return config
  config.webhooks.subscriptions = reduceWebhooks(subscriptionsArr)
  return config
```

alexanderMontague (approved):
Not sure I agree with the approach but agreed lets not block and get this shipped! We can clean it up on the server for M2 - thanks for looking into this

packages/app/src/cli/services/app/config/link.ts
isaacroldan: You can't do that because you need to get the `subscriptionsArr` first, and config is not properly typed here so you need to do the initial cast.

This line does make sense though:
`config.webhooks.subscriptions = reduceWebhooks(subscriptionsArr)`

isaacroldan:
Approved 👍

packages/app/src/cli/services/app/config/link.ts
alexanderMontague: ah I missed that var, you could just do `config.webhooks.subscriptions`? Which at that point does make sense to pull this out, but still simplifies things - with edit:

```js
const subscriptionsArr = config?.webhooks?.subscriptions as WebhooksConfig["subscriptions"]
if (!subscriptionsArr.length) return config
  config.webhooks.subscriptions = reduceWebhooks(subscriptionsArr)
  return config
```

packages/app/src/cli/models/extensions/specifications/transform/app_config_webhook.ts
gracejychang: this makes sure that we're parsing `compliance_topics` from the `privacy_compliance_subsriptions` spec rather than the `webhooks` spec! so for example when we have the TOML like this:

```
[[webhooks.subscriptions]]
topics = ["products/create"]
compliance_topics = ["customers/redact"]
```

we won't run into the bug where we have 2 compliance subscriptions

packages/app/src/cli/models/extensions/specifications/transform/app_config_webhook.ts
gracejychang: Not sure, I think @isaacroldan might have a better idea about this? I think the spread does have special overwriting under the hood, cause `concat` breaks a couple of tests I had for this

Also unfortunately `existingSubscription[property].length` doesn't help with needing the empty array as this can also be `undefined` :/

packages/app/src/cli/models/extensions/specifications/transform/app_config_webhook.ts
gracejychang: `concat` also breaks things here :/

packages/app/src/cli/models/app/loader.test.ts
```diff
@@ -2828,7 +2806,7 @@ describe('WebhooksSchema', () => {
       code: zod.ZodIssueCode.custom,
       message: 'You can’t have duplicate subscriptions with the exact same `topic`, `uri` and `filter`',
       fatal: true,
-      path: ['webhooks', 'subscriptions', 1, 'topics', 0, 'products/create'],
+      path: ['webhooks', 'subscriptions', 0, 'topics', 1, 'products/create'],
```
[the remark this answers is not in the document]
gracejychang: ohh, I think this might have something to do with condensing the webhooks, instead of sorting? cause the subscriptions look like this after the transform on the schema:

```
      subscriptions: [
        {
          topics: ['products/update', 'products/update'],
          uri: 'https://example.com',
          filter: 'title:shoes',
        },
      ],
```

PR #3918, opened 2024-05-21
Add log polling to `dev` command

lopert:

For https://github.com/Shopify/shopify-functions/issues/118

Adds log polling to the app dev command.
This listens for functionRuns 

### How to test your changes? 🎩 
- Have a dev store with your app installed and function deployed
- enable log polling flag on partners org
- set local envVar to enable the log polling, `export SHOPIFY_CLI_ENABLE_APP_LOG_POLLING=1`
- From the cli directory, run app dev, i.e. `pnpm shopify app dev --path /Users/lopert/code/log-streaming/app/log-streamer`

### Measuring impact

### Checklist

### DIFF
--- a/packages/app/src/cli/api/graphql/subscribe_to_app_logs.ts
+++ b/packages/app/src/cli/api/graphql/subscribe_to_app_logs.ts
@@ -1,0 +1,25 @@
+import {gql} from 'graphql-request'
+
+export interface AppLogsSubscribeVariables {
+  shopIds: string[]
+  apiKey: string
+  token: string
+}
+
+export interface AppLogsSubscribeResponse {
+  appLogsSubscribe: {
+    success: boolean
+    errors?: string[]
+    jwtToken: string
+  }
+}
+
+export const AppLogsSubscribeMutation = gql`
+  mutation AppLogsSubscribe($apiKey: String!, $shopIds: [ID!]!) {
+    appLogsSubscribe(input: {apiKey: $apiKey, shopIds: $shopIds}) {
+      jwtToken
+      success
+      errors
+    }
+  }
+`
--- a/packages/app/src/cli/services/app-logs/poll-app-logs.test.ts
+++ b/packages/app/src/cli/services/app-logs/poll-app-logs.test.ts
@@ -1,0 +1,150 @@
+import {pollAppLogs} from './poll-app-logs.js'
+import {writeAppLogsToFile} from './write-app-logs.js'
+import {partnersFqdn} from '@shopify/cli-kit/node/context/fqdn'
+import {describe, expect, test, vi, beforeEach, afterEach} from 'vitest'
+import {fetch} from '@shopify/cli-kit/node/http'
+
+const JWT_TOKEN = 'jwtToken'
+const API_KEY = 'apiKey'
+
+vi.mock('./write-app-logs.js')
+vi.mock('@shopify/cli-kit/node/http')
+
+const FQDN = await partnersFqdn()
+const LOGS = '1\\n2\\n3\\n4\\n'
+const INPUT = {
+  cart: {
+    lines: [
+      {
+        quantity: 3,
+        merchandise: {
+          __typename: 'ProductVariant',
+          id: 'gid:\\/\\/shopify\\/ProductVariant\\/2',
+        },
+      },
+    ],
+  },
+}
+const OUTPUT = {
+  discountApplicationStrategy: 'FIRST',
+  discounts: [
+    {
+      message: '10% off',
+      value: {
+        percentage: {
+          value: 10,
+        },
+      },
+      targets: [
+        {
+          productVariant: {
+            id: 'gid://shopify/ProductVariant/2',
+          },
+        },
+      ],
+    },
+  ],
+}
+const PAYLOAD = {
+  input: JSON.stringify(INPUT),
+  input_bytes: 123,
+  output: JSON.stringify(OUTPUT),
+  output_bytes: 182,
+  function_id: 'e57b4d31-2038-49ff-a0a1-1eea532414f7',
+  logs: LOGS,
+  fuel_consumed: 512436,
+}
+const RETURNED_CURSOR = '2024-05-23T19:17:02.321773Z'
+const RESPONSE_DATA = {
+  app_logs: [
+    {
+      shop_id: 1,
+      api_client_id: 1830457,
+      payload: JSON.stringify(PAYLOAD),
+      event_type: 'function_run',
+      cursor: '2024-05-23T19:17:02.321773Z',
+      status: 'success',
+      log_timestamp: '2024-05-23T19:17:00.240053Z',
+    },
+  ],
+  cursor: RETURNED_CURSOR,
+}
+
+describe('pollAppLogs', () => {
+  let stdout: any
+
+  beforeEach(() => {
+    stdout = {write: vi.fn()}
+    vi.useFakeTimers()
+  })
+
+  afterEach(() => {
+    vi.useRealTimers()
+    vi.clearAllTimers()
+  })
+
+  test('polls and re-polls the endpoint', async () => {
+    const firstUrl = `https://${FQDN}/app_logs/poll`
+    const secondUrl = `${firstUrl}?cursor=${RETURNED_CURSOR}`
+
+    // Given
+    vi.mocked(writeAppLogsToFile)
+
+    const mockedFetch = vi
+      .fn()
+      .mockResolvedValueOnce(Response.json(RESPONSE_DATA))
+      .mockResolvedValueOnce(Response.json(RESPONSE_DATA))
+    vi.mocked(fetch).mockImplementation(mockedFetch)
+
+    // When
+    await pollAppLogs({stdout, appLogsFetchInput: {jwtToken: JWT_TOKEN}, apiKey: API_KEY})
+    await vi.advanceTimersToNextTimerAsync()
+
+    // Then
+    expect(fetch).toHaveBeenCalledWith(firstUrl, {
+      method: 'GET',
+      headers: {
+        Authorization: `Bearer ${JWT_TOKEN}`,
+      },
+    })
+
+    expect(fetch).toHaveBeenCalledWith(secondUrl, {
+      method: 'GET',
+      headers: {
+        Authorization: `Bearer ${JWT_TOKEN}`,
+      },
+    })
+
+    expect(writeAppLogsToFile).toHaveBeenCalledWith({
+      appLog: RESPONSE_DATA.app_logs[0],
+      apiKey: API_KEY,
+      stdout,
+    })
+
+    expect(stdout.write).toHaveBeenCalledWith('Function executed succesfully using 0.5124M instructions.')
+    expect(stdout.write).toHaveBeenCalledWith(LOGS)
+
+    expect(vi.getTimerCount()).toEqual(1)
+  })
+
+  test('throws error if response is not ok', async () => {
+    // Given
+    const url = `https://${FQDN}/app_logs/poll`
+
+    const response = new Response('errorMessage', {status: 500})
+    const mockedFetch = vi.fn().mockResolvedValueOnce(response)
+    vi.mocked(fetch).mockImplementation(mockedFetch)
+
+    // When/Then
+    await expect(() =>
+      pollAppLogs({stdout, appLogsFetchInput: {jwtToken: JWT_TOKEN}, apiKey: API_KEY}),
+    ).rejects.toThrowError('Error while fetching: errorMessage')
+
+    expect(fetch).toHaveBeenCalledWith(url, {
+      method: 'GET',
+      headers: {
+        Authorization: `Bearer ${JWT_TOKEN}`,
+      },
+    })
+  })
+})
--- a/packages/app/src/cli/services/app-logs/poll-app-logs.ts
+++ b/packages/app/src/cli/services/app-logs/poll-app-logs.ts
@@ -1,0 +1,97 @@
+import {writeAppLogsToFile} from './write-app-logs.js'
+import {partnersFqdn} from '@shopify/cli-kit/node/context/fqdn'
+import {fetch} from '@shopify/cli-kit/node/http'
+import {Writable} from 'stream'
+
+const POLLING_INTERVAL_MS = 450
+const ONE_MILLION = 1000000
+
+const generateFetchAppLogUrl = async (cursor?: string) => {
+  const fqdn = await partnersFqdn()
+  const url = `https://${fqdn}/app_logs/poll`
+  return url + (cursor ? `?cursor=${cursor}` : '')
+}
+
+export interface AppEventData {
+  shop_id: number
+  api_client_id: number
+  payload: string
+  event_type: string
+  cursor: string
+  status: 'success' | 'failure'
+  log_timestamp: string
+}
+
+export const pollAppLogs = async ({
+  stdout,
+  appLogsFetchInput: {jwtToken, cursor},
+  apiKey,
+}: {
+  stdout: Writable
+  appLogsFetchInput: {jwtToken: string; cursor?: string}
+  apiKey: string
+}) => {
+  const url = await generateFetchAppLogUrl(cursor)
+  const response = await fetch(url, {
+    method: 'GET',
+    headers: {
+      Authorization: `Bearer ${jwtToken}`,
+    },
+  })
+
+  if (!response.ok) {
+    // We should add some exponential backoff here to not spam partners
+
+    const responseText = await response.text()
+    throw new Error(`Error while fetching: ${responseText}`)
+  }
+
+  const data = (await response.json()) as {
+    app_logs?: AppEventData[]
+    cursor?: string
+    errors?: string[]
+  }
+
+  if (data.app_logs) {
+    const {app_logs: appLogs} = data
+
+    const functionLogs = appLogs.filter((appLog) => appLog.event_type === 'function_run')
+
+    for (const functionLog of functionLogs) {
+      const payload = JSON.parse(functionLog.payload)
+      const fuel = (payload.fuel_consumed / ONE_MILLION).toFixed(4)
+
+      if (functionLog.status === 'success') {
+        stdout.write(`Function executed succesfully using ${fuel}M instructions.`)
+      } else if (functionLog.status === 'failure') {
+        stdout.write(`❌ Function failed to execute with error: ${payload.error_type}`)
+      }
+
+      // print the logs from the appLogs as well
+      const logs = JSON.parse(functionLog.payload).logs
+      stdout.write(logs)
+
+      // eslint-disable-next-line no-await-in-loop
+      await writeAppLogsToFile({
+        appLog: functionLog,
+        apiKey,
+        stdout,
+      })
+    }
+  }
+
+  const cursorFromResponse = data?.cursor
+
+  setTimeout(() => {
+    pollAppLogs({
+      stdout,
+      appLogsFetchInput: {
+        jwtToken,
+        cursor: cursorFromResponse,
+      },
+      apiKey,
+    }).catch((error) => {
+      throw new Error(`${error} error while fetching.`)
+    })
+  }, POLLING_INTERVAL_MS)
+}
--- a/packages/app/src/cli/services/app-logs/write-app-logs.test.ts
+++ b/packages/app/src/cli/services/app-logs/write-app-logs.test.ts
@@ -1,0 +1,62 @@
+import {writeAppLogsToFile} from './write-app-logs.js'
+import {AppEventData} from './poll-app-logs.js'
+import {joinPath} from '@shopify/cli-kit/node/path'
+import {writeLog} from '@shopify/cli-kit/node/logs'
+import {describe, expect, test, vi, beforeEach} from 'vitest'
+
+vi.mock('@shopify/cli-kit/node/logs')
+
+const APP_LOG: AppEventData = {
+  shop_id: 1,
+  api_client_id: 2,
+  payload: JSON.stringify({someJson: 'someJSOn'}),
+  event_type: 'function_run',
+  cursor: '2024-05-22T15:06:43.841156Z',
+  status: 'success',
+  log_timestamp: '2024-05-22T15:06:41.827379Z',
+}
+const API_KEY = 'apiKey'
+
+describe('writeAppLogsToFile', () => {
+  let stdout: any
+
+  beforeEach(() => {
+    stdout = {write: vi.fn()}
+  })
+
+  test('calls writeLog with the right data', async () => {
+    // Given
+    const logData = expectedLogDataFromAppEvent(APP_LOG)
+
+    // determine the fileName and path
+    const fileName = `app_logs_${APP_LOG.log_timestamp}.json`
+    const path = joinPath(API_KEY, fileName)
+
+    // When
+    await writeAppLogsToFile({appLog: APP_LOG, apiKey: API_KEY, stdout})
+
+    // Then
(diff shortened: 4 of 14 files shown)


packages/app/src/cli/services/app_logs/fetch-app-logs.ts
```diff
@@ -0,0 +1,79 @@
...
+  setTimeout(async () => {
+    await fetchAppLogs({
+      stdout,
+      appLogsFetchInput: {
+        jwtToken,
+        cursor: cursorFromResponse,
+      },
+      apiKey,
+    })
+  }, 1000)
```
DuncanUszkay1, line 78: I think this should be set as a constant

packages/app/src/cli/services/app_logs/fetch-app-logs.ts
```diff
@@ -0,0 +1,79 @@
...
+  }
+
+  if (data?.errors) {
+    data?.errors?.forEach((error) => {
+      stdout.write(`${error}\n`)
+    })
+  } else if (data.app_logs) {
+    const {app_logs: appLogs} = data
+
+    await Promise.all(
```
nickwesselman, line 55: Are we doing the printing of the log during `app dev` as a separate PR?

packages/app/src/cli/services/app_logs/fetch-app-logs.ts
nickwesselman: @DuncanUszkay1 We figured that we needed 400ms to meet our SLO right?

I thought we also discussed providing some sort of server-side control over this (a header returned during the first poll?) so that we can adjust if needed. Once a CLI version is out, we can't change a hardcoded value.

DuncanUszkay1: I think it's fine to ship this (at 450ms) behind an internal beta flag for this PR- ultimately users can modify their client to poll at any rate, so it's fine if it appears in the code. The actual guard against over-polling is serverside, and it's an IP throttle.

As a follow-up, we can add in the serverside interval.

packages/app/src/cli/services/app_logs/fetch-app-logs.ts
```diff
@@ -0,0 +1,79 @@
...
+      appLogs.map(async (log) => {
+        await writeAppLogsToFile({
```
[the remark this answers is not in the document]
lopert: Hmm, I get what you are saying, and it works when I write it out in my codde, but when I try this, I get a lint violation.
```
error  Promises must be awaited, end with a call to .catch, end with a call to .then with a rejection handler or be explicitly marked as ignored with the `void` operator  @typescript-eslint/no-floating-promises
```

packages/app/src/cli/services/app_logs/fetch-app-logs.ts
```diff
@@ -0,0 +1,79 @@
...
+  stdout,
+  appLogsFetchInput: {jwtToken, cursor},
+  apiKey,
+}: {
+  stdout: Writable
+  appLogsFetchInput: {jwtToken: string; cursor?: string}
+  apiKey: string
+}) => {
+  const url = await generateFetchAppLogUrl(cursor)
+  const response = await fetch(url, {
```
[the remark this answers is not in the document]
lopert: This is using [cli-kit's fetch](https://shopify.github.io/cli/api/cli-kit/functions/node_http.fetch.html) so we get a bunch of extra debug info with `--verbose` for free.

packages/app/src/cli/services/app_logs/fetch-app-logs.ts
```diff
@@ -0,0 +1,89 @@
...
+        })
+      }),
+    )
+
+    // print the logs from the appLogs as well
+  }
+
+  const cursorFromResponse = data?.cursor
+
+  // eslint-disable-next-line @typescript-eslint/no-misused-promises
```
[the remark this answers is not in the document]
lopert: 🤔 Swapped it around to respect the lint, but I feel it's a bit weird?
First time around, we invoke fetchAppLogs directly, so if it errors out, we'd get the message from it directly.
On the second time around, would an error "bubble up" to this setTimeout's catch, and then be handled there?

Feels like the way I've set it up is wrong since ideally the error flow would always go the same way?

packages/app/src/cli/services/app_logs/fetch-app-logs.ts
```diff
@@ -0,0 +1,79 @@
...
+  shop_id: number
+  api_client_id: number
+  payload: string
+  event_type: string
+  cursor: string
+  status: 'success' | 'failure'
+  log_timestamp: string
+}
+
+export const fetchAppLogs = async ({
```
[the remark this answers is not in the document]
lopert: Yes, dev just crashes with the error message. This includes getting a bad response back from partners.
We can add error handling of the different errors to keep it retrying within https://github.com/Shopify/script-editor/issues/4145#issue-2319353974

packages/app/src/cli/services/app_logs/fetch-app-logs.ts
```diff
@@ -0,0 +1,95 @@
+import {writeAppLogsToFile} from './write-app-logs.js'
+import {partnersFqdn} from '@shopify/cli-kit/node/context/fqdn'
+import {fetch} from '@shopify/cli-kit/node/http'
+import {Writable} from 'stream'
+
+const pollingInterval = 450
```
DuncanUszkay1, line 6: nit, would be good to indicate this is ms, either in a comment or the variable name

packages/app/src/cli/services/app_logs/fetch-app-logs.ts
```diff
@@ -0,0 +1,95 @@
...
+    // We should add some exponential backoff here to not spam partners
+
+    errors.forEach((errorMessage: string) => {
+      // if the error is about the token being expired,
+      // we should perform the refresh here instead of requiring restarting app dev
+      // should be able to detect with errorMessage == 'JWT token expired'
+      stdout.write(`Error while fetching app logs: ${errorMessage}`)
+    })
+    throw new Error(`HTTP error! status: ${response.status}`)
+  }
```
DuncanUszkay1, line 54: Ticket: https://github.com/Shopify/script-editor/issues/4145

packages/app/src/cli/services/app_logs/fetch-app-logs.test.ts
```diff
@@ -0,0 +1,85 @@
...
+    stdout = {write: vi.fn()}
+    vi.useFakeTimers()
+  })
+
+  afterEach(() => {
+    vi.useRealTimers()
+    vi.clearAllTimers()
+  })
+
+  test('polls the endpoint', async () => {
```
DuncanUszkay1, line 26: Seems we're missing the error case

packages/app/src/cli/services/app_logs/write-app-logs.test.ts
```diff
@@ -0,0 +1,63 @@
...
+
+    // When
+    await writeAppLogsToFile({appLog: APP_LOG, apiKey: API_KEY, stdout})
+
+    // Then
+    expect(writeLog).toHaveBeenCalledWith(path, logData)
+    expect(stdout.write).toHaveBeenCalledWith(expect.stringContaining('Wrote log to'))
+  })
+
+  test('prints and re-throws parsing errors', async () => {
```
DuncanUszkay1, line 44: Do we need a test for the non-syntax error case handled in the code?

packages/app/src/cli/services/dev/processes/setup-dev-processes.test.ts
```diff
@@ -107,6 +108,8 @@ describe('setup-dev-processes', () => {
 
     const graphiqlKey = 'somekey'
 
+    process.env.SHOPIFY_CLI_ENABLE_APP_LOG_POLLING = '1'
```
DuncanUszkay1, line 111: Should we add a test to ensure that when this flag is 0 the setup is not performed?

DuncanUszkay1 (commented):
Code LGTM with some minor nits- I think we're missing some test cases though

packages/app/src/cli/services/app_logs/fetch-app-logs.ts
lopert: 👍 I can add `Ms` at the end. In `setTimeout` it's just `ms?: number,`

packages/app/src/cli/services/dev/processes/app-logs-polling.ts
```diff
@@ -0,0 +1,63 @@
...
+
+  if (errors.length > 0) {
+    errors.forEach((error) => {
+      stdout.write(`Error: ${error}\n`)
+    })
+  } else {
+    outputDebug(`Subscribed to App Events for shop ID(s) ${options.shopIds}`)
+    outputDebug(`Success: ${success}\n`)
+    outputDebug(`Token: ${jwtToken}\n`)
+    outputDebug(`API Key: ${options.apiKey}\n`)
```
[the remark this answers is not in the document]
lopert: This only shows up when using `--verbose`, I think it's fine to add this extra info to help debug things.

packages/app/src/cli/services/app_logs/write-app-logs.test.ts
lopert: Since we're just re-throwing here, I think I can remove the whole `try { catch` and just let it bubble up normally.

packages/app/src/cli/services/dev/processes/setup-dev-processes.test.ts
lopert: Sure thign! I can just split out our change to the dev process list into its own test.
Then when we release and remove the env flag, we just delete our test and move the check back into the main one.

packages/app/src/cli/services/app_logs/write-app-logs.ts
```diff
@@ -0,0 +1,57 @@
...
+}) => {
+  const fileName = `app_logs_${appLog.log_timestamp}.json`
+  const toSaveData = {
+    ...appLog,
+  }
+
+  try {
+    const parsedPayload = JSON.parse(appLog.payload)
+    toSaveData.payload = parsedPayload
+    // eslint-disable-next-line no-catch-all/no-catch-all
```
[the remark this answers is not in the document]
lopert: No, we'll want to surface them. I've removed the related catch-all that was near here

packages/app/src/cli/services/app_logs/write-app-logs.test.ts
```diff
@@ -0,0 +1,63 @@
...
+  test('calls writeLog with the right data', async () => {
+    // Given
+    const logData = expectedLogDataFromAppEvent(APP_LOG)
+
+    // determine the fileName and path
+    const fileName = `app_logs_${APP_LOG.log_timestamp}.json`
+    const path = joinPath(API_KEY, fileName)
+
+    // When
+    await writeAppLogsToFile({appLog: APP_LOG, apiKey: API_KEY, stdout})
```
shauns, line 37: This will write to the "real" logs location (but for a fake app). Maybe a temporary folder would be better? There are good test utils to help with that used elsewhere, that will tear down after a test finishes.

packages/app/src/cli/services/dev/processes/setup-dev-processes.test.ts
```diff
@@ -239,6 +240,98 @@ describe('setup-dev-processes', () => {
     })
   })
 
+  test('process list includes app polling when envVar is enabled', async () => {
+    process.env.SHOPIFY_CLI_ENABLE_APP_LOG_POLLING = '1'
```
shauns, line 244: We really try to avoid this, because we'd like to be able to move tests to run in parallel. Two patterns used elsewhere:
- the function that uses environment variables might have a signature like `..., env=process.env)...`, meaning you can inject your test variable via an extra parameter.
- you can wrap up the use of `process.env` in the subject function with `getEnvironmentVariables()`, and then mock the return of `getEnvironmentVariables` here.

Should be something we lint for, so I apologise for the manual intervention!

packages/cli-kit/src/public/node/logs.ts
```diff
@@ -0,0 +1,13 @@
...
+import {joinPath} from '@shopify/cli-kit/node/path'
+import {mkdir, writeFile} from '@shopify/cli-kit/node/fs'
+
+export const getLogsDir = logsFolder()
+
+export const createLogsDir = async (path: string): Promise<void> => {
+  await mkdir(joinPath(logsFolder(), path))
+}
+
+export const writeLog = async (path: string, logData: string): Promise<void> => {
```
shauns, line 11: Does this append or replace the file? Maybe best to name in such a way that the behaviour is more explicit.

shauns (changes requested):

packages/app/src/cli/services/app-logs/write-app-logs.ts
```diff
@@ -0,0 +1,35 @@
...
+  apiKey,
+  stdout,
+}: {
+  appLog: AppEventData
+  apiKey: string
+  stdout: Writable
+}) => {
+  const fileName = `app_logs_${appLog.log_timestamp}.json`
+  const path = joinPath(apiKey, fileName)
+  const fullOutputPath = joinPath(getLogsDir, path)
```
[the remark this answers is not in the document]
andrewhassan: Okay, fair enough. I also don't have a strong opinion on this so I'm fine to leave it as is 😄

packages/app/src/cli/services/dev/processes/setup-dev-processes.test.ts
lopert: I saw other test that was mocking `getEnvironmentVariables` and followed that pattern 👍

packages/app/src/cli/api/graphql/subscribe_to_app_logs.ts
```diff
@@ -0,0 +1,25 @@
+import {gql} from 'graphql-request'
+
+export interface AppLogsQueryOptions {
```
andrewhassan, line 3: Nit:

```suggestion
export interface AppLogsSubscribeVariables {
```

packages/app/src/cli/services/app-logs/write-app-logs.test.ts
```diff
@@ -0,0 +1,66 @@
...
+vi.mock('@shopify/cli-kit/node/logs')
+
+const DUMMY_JSON = {
+  someJson: 'someJSOn',
+}
+
+const APP_LOG: AppEventData = {
+  shop_id: 1,
+  api_client_id: 2,
+  payload: JSON.stringify(DUMMY_JSON),
```
andrewhassan, line 16: Nit: Since `DUMMY_JSON` is small and not used by anything else, maybe we could just inline it?

```suggestion
  payload: '{}',
```

packages/app/src/cli/services/app-logs/write-app-logs.ts
```diff
@@ -0,0 +1,35 @@
...
+  const toSaveData = {
+    ...appLog,
+  }
+
+  try {
+    const parsedPayload = JSON.parse(appLog.payload)
+    toSaveData.payload = parsedPayload
```
andrewhassan, line 25: Nit: Instead of mutating the object here, we could move the initialization down here:

```typescript
const toSaveData = {
  ...appLog,
  payload: JSON.parse(appLoad.payload),
}
```

packages/app/src/cli/services/dev/processes/app-logs-polling.ts
```diff
@@ -0,0 +1,74 @@
...
+
+import {outputDebug} from '@shopify/cli-kit/node/output'
+
+interface SubscribeAndStartPollingOptions {
+  developerPlatformClient: DeveloperPlatformClient
+  appLogsQueryOptions: AppLogsQueryOptions
+}
+
+interface AppLogsQueryOptions {
+  shopIds: [string]
```
andrewhassan, line 15: This would signify that `shopIds` is effectively a tuple with a single string value. I think we want it to be an array of strings, which would be `string[]`.

packages/app/src/cli/services/dev/processes/app-logs-polling.ts
```diff
@@ -0,0 +1,74 @@
...
+}
+
+export interface AppLogsSubscribeProcess extends BaseProcess<SubscribeAndStartPollingOptions> {
+  type: 'app-logs-subscribe'
+}
+
+interface Props {
+  developerPlatformClient: DeveloperPlatformClient
+  subscription: {
+    shopIds: [string]
```
andrewhassan, line 27: Same comment here

packages/app/src/cli/services/dev/processes/app-logs-polling.ts
```diff
@@ -0,0 +1,74 @@
...
+    apiKey: string
+  }
+}
+
+export async function setupAppLogsPollingProcess({
+  developerPlatformClient,
+  subscription: {shopIds, apiKey},
+}: Props): Promise<AppLogsSubscribeProcess | undefined> {
```
andrewhassan, line 35: In what case can we return `undefined`? It looks like we always return `AppLogsSubscribeProcess`.

packages/app/src/cli/services/dev/processes/app-logs-polling.ts
```diff
@@ -0,0 +1,74 @@
...
+  }
+}
+
+export const subscribeAndStartPolling: DevProcessFunction<SubscribeAndStartPollingOptions> = async (
+  {stdout},
+  {developerPlatformClient, appLogsQueryOptions},
+) => {
+  const result = await developerPlatformClient.subscribeToAppLogs(appLogsQueryOptions)
+  const {jwtToken, success, errors} = result.appLogsSubscribe
+  outputDebug(`Token: ${jwtToken}\n`)
```
andrewhassan, line 59: Is this something we should output, even in verbose mode? IIRC, we redact even the API keys in verbose mode.

packages/app/src/cli/services/dev/processes/setup-dev-processes.test.ts
```diff
@@ -239,6 +243,101 @@ describe('setup-dev-processes', () => {
...
+        developerPlatformClient,
+        appLogsQueryOptions: {
+          shopIds: ['123456789'],
+          apiKey: 'api-key',
+          token: 'token',
+        },
+      },
+    })
+
+    delete process.env.SHOPIFY_CLI_ENABLE_APP_LOG_POLLING
```
andrewhassan, line 338: We don't need this anymore since we're not using `process.env` directly.

packages/app/src/cli/services/app-logs/poll-app-logs.ts
```diff
@@ -0,0 +1,97 @@
...
+    headers: {
+      Authorization: `Bearer ${jwtToken}`,
+    },
+  })
+
+  if (!response.ok) {
+    // We should add some exponential backoff here to not spam partners
+
+    const responseText = await response.text()
+    throw new Error(`Error while fetching: ${responseText}`)
```
DuncanUszkay1, line 46: 💭 what does this look like to the end user? Curious if the dev would know what this error relates to

packages/app/src/cli/services/dev/processes/setup-dev-processes.test.ts
```diff
@@ -239,6 +243,101 @@ describe('setup-dev-processes', () => {
     })
   })
 
+  test('process list includes app polling when envVar is enabled', async () => {
```
andrewhassan, line 246: Should we add another test to verify that app polling is not included if the env var is falsy?

andrewhassan (approved):
LGTM aside from a few comments!

packages/app/src/cli/services/app-logs/poll-app-logs.ts
```diff
@@ -0,0 +1,97 @@
...
+
+    for (const functionLog of functionLogs) {
+      const payload = JSON.parse(functionLog.payload)
+      const fuel = (payload.fuel_consumed / ONE_MILLION).toFixed(4)
+
+      if (functionLog.status === 'success') {
+        stdout.write(`Function executed succesfully using ${fuel}M instructions.`)
+      } else if (functionLog.status === 'failure') {
+        stdout.write(`❌ Function failed to execute with error: ${payload.error_type}`)
+      }
```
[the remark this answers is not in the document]
DuncanUszkay1: Might need to include the logs portion below too 🤔 

I don't think we need to worry about it immediately, but we should aim to get it out of here eventually, maybe even as a fast follow after editions

packages/app/src/cli/services/dev/processes/setup-dev-processes.test.ts
DuncanUszkay1, line 246: I've missed a bunch of conversation on this PR so this might've already been addressed, but did we decide to not include a test to show that the process list does not include polling when the var is disabled?

DuncanUszkay1 (approved):
LGTM- I think this will be a good foothold, and we can figure out how to push more of the processing serverside later.

packages/app/src/cli/services/app-logs/poll-app-logs.test.ts
```diff
@@ -0,0 +1,150 @@
...
+
+    const mockedFetch = vi
+      .fn()
+      .mockResolvedValueOnce(Response.json(RESPONSE_DATA))
+      .mockResolvedValueOnce(Response.json(RESPONSE_DATA))
+    vi.mocked(fetch).mockImplementation(mockedFetch)
+
+    // When
+    await pollAppLogs({stdout, appLogsFetchInput: {jwtToken: JWT_TOKEN}, apiKey: API_KEY})
+    await vi.advanceTimersToNextTimerAsync()
```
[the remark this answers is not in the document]
andrewhassan: It's technically all async because we're using `await` (think of it as using promises under-the-hood). IIRC the issue is that the non-async version of this method doesn't wait for the `pollAppLogs` call to resolve before advancing the timers. So it's advanced before the timer is even enqueued.

andrewhassan:
Also don't forget to squash the commits before merging!

packages/app/src/cli/services/dev/processes/app-logs-polling.ts
lopert: These options are the same as in the gql file so I've removed this and just imported that one, which has `string[]` 👍

packages/app/src/cli/services/dev/processes/app-logs-polling.ts
lopert: The apiKey here is actually the clientID in partners, `Your client ID is your app’s public identifier and is required for interacting with Shopify APIs.`, so I still see the value in my logs with the verbose enabled. I'm not sure why it's referred to as `apiKey` in the CLI, but we're using something that already existed. `¯\_(ツ)_/¯`

I think having the token output here is fine? Is the worry that someone will steal this and then be able to submit poll requests? 🤔 I believe it has a 1 day expiry.

packages/app/src/cli/services/dev/processes/setup-dev-processes.test.ts
lopert: I think this is covered by the main process list test. 
That test checks the resulting list with hardcoded "element at this array index should be processX".
When I had originally modified that test, I had to adjust the values of the last check since it adds itself to the end.

lopert: @andrewhassan had a similar comment, see my response: https://github.com/Shopify/cli/pull/3918#discussion_r1619080831

packages/app/src/cli/services/app-logs/poll-app-logs.ts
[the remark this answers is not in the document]
nickwesselman: I think the ask was just to make it possible. There's a lot in CLI that's still extension-specific.

PR #3920, opened 2024-05-21
Feature - Add a noDelete flag to the theme dev command

jamesmengo:
- https://github.com/Shopify/develop-advanced-edits/issues/200

When the `nodelete` is passed
- **Files that are only present locally** are automatically uploaded to the remote theme (normally, we would prompt to ask the user if they want to delete the file locally)
- Files that are deleted on the remote theme are **NOT** deleted from the local disk
  - Modifications will still be downloaded

### Measuring impact

### Checklist


packages/theme/src/cli/services/dev.ts
```diff
@@ -42,31 +43,40 @@ export async function dev(options: DevOptions) {
     return
   }
 
-  if (options.flagsToPass.includes('--poll')) {
-    renderWarning({
-      body: 'The CLI flag --[flag-name] is now deprecated and will be removed in future releases. It is no longer necessary with the new implementation. Please update your usage accordingly.',
-    })
-  }
+  if (options['dev-preview']) {
```
karreiro, line 46: I don't think we need this condition, because if [the code is reaching this point](https://github.com/Shopify/cli/blob/61dfa17e6f2091d3eebf301439dd523139f844f3/packages/theme/src/cli/services/dev.ts#L37-L40), it's already running the new implementation :)

karreiro: (now, that I'm reading a bit more, I think we maybe have conflict issue on this method)

karreiro (commented):
Thank you, @jamesmengo! I've left only some minor comments, but the PR is in an excellent direction 🚀

packages/theme/src/cli/utilities/theme-environment/types.ts
```diff
@@ -52,6 +52,16 @@ export interface DevServerContext {
    * theme.
    */
   themeEditorSync: boolean
+
+  /**
+   * Additional options for the development server.
+   */
+  options?: {
```
karreiro, line 59: I believe we will make this non-optional (just leaving a reminder from our pairing) ☺️

PR #3923, opened 2024-05-22
Pull template specifications from online JSON files

amcaplan:

Allows us to update template specifications dynamically (as we do in Partners).

1. Pulls the template specifications dynamically from files in the 2 extension repos (https://github.com/Shopify/extensions-templates and https://github.com/Shopify/function-examples)
2. Respects beta flag gating (though, since we aren't accessing the BP Organizations API yet, we just default everything to `false` for now)

Before merging this, we need to merge the changes to the other repos to include the JSON files. We can probably merge without gating fully implemented, as we are defaulting to a safe strict gating, letting nothing through.

### How to test your changes?

`app generate extension` should work as before

### Measuring impact

### Checklist


packages/app/src/cli/models/app/app.test-data.ts
```diff
@@ -1154,7 +1154,7 @@ export function testDeveloperPlatformClient(stubs: Partial<DeveloperPlatformClie
     orgFromId: (_organizationId: string) => Promise.resolve(testOrganization()),
     appsForOrg: (_organizationId: string) => Promise.resolve({apps: [testOrganizationApp()], hasMorePages: false}),
     specifications: (_app: MinimalAppIdentifiers) => Promise.resolve(testRemoteSpecifications),
-    templateSpecifications: (_appId: string) => Promise.resolve(testRemoteExtensionTemplates),
+    templateSpecifications: (_app: MinimalAppIdentifiers) => Promise.resolve(testRemoteExtensionTemplates),
```
shauns, line 1157: I'm going to post about this elsewhere, but this is not good -- I saw we did it for specifications too. We are creating a dependency on having a loaded remote app, when _one is not required_. We should look into using a more narrow contract, and ideally one based on client-side loading only. I do realise the app is generally loaded already in the places where this is used today - the point stands, because we are making it harder to create efficiencies in how commands are structured.

packages/app/src/cli/utilities/developer-platform-client/app-management-client.ts
```diff
@@ -711,6 +728,15 @@ export class AppManagementClient implements DeveloperPlatformClient {
       variables,
     )
   }
+
+  private async organizationBetaFlags(_organizationId: string, allBetaFlags: string[]): Promise<{[key: string]: boolean}> {
```
shauns, line 732: Are the flags here and in the github manifest file going to be hashed? There needs to be a strong reason if not.

packages/app/src/cli/utilities/developer-platform-client/app-management-client.ts
```diff
@@ -280,8 +289,16 @@ export class AppManagementClient implements DeveloperPlatformClient {
       )
   }
 
-  async templateSpecifications(_appId: string): Promise<ExtensionTemplate[]> {
-    return stubbedExtensionTemplates()
+  async templateSpecifications({organizationId}: MinimalAppIdentifiers): Promise<ExtensionTemplate[]> {
+    const allExtensions = await Promise.all(templateJsonUrls.map(async (url) => {
+      const response = await fetch(url)
+      return response.json() as Promise<GatedExtensionTemplate[]>
```
shauns, line 295: We should handle errors here given we're taking a new dependency on github's asset server.

packages/app/src/cli/utilities/developer-platform-client/app-management-client.ts
```diff
@@ -280,8 +289,16 @@ export class AppManagementClient implements DeveloperPlatformClient {
...
+    })).then(templates2DArray => templates2DArray.flat(1))
```
shauns, line 296: If there are duplicate definitions across the two jsons (and the fact there are two repos is obviously a historical mistake), which wins? Is that an error?

packages/app/src/cli/utilities/developer-platform-client/app-management-client.ts
```diff
@@ -711,6 +728,15 @@ export class AppManagementClient implements DeveloperPlatformClient {
       variables,
     )
   }
+
+  private async organizationBetaFlags(_organizationId: string, allBetaFlags: string[]): Promise<{[key: string]: boolean}> {
+    // For now, stub everything as false
```
shauns, line 733: I'm not sure why we're intending to ship something like this -- is there a reason why the dependency (org flags, the 2nd point from the PR description) is not included _first_?

packages/app/src/cli/utilities/developer-platform-client/app-management-client.ts
shauns: I think it should be logged at least. It points at an internal team misconfiguring the manifests.

packages/app/src/cli/utilities/developer-platform-client/app-management-client.ts
```diff
@@ -280,8 +289,16 @@ export class AppManagementClient implements DeveloperPlatformClient {
       )
   }
 
-  async templateSpecifications(_appId: string): Promise<ExtensionTemplate[]> {
-    return stubbedExtensionTemplates()
+  async templateSpecifications({organizationId}: MinimalAppIdentifiers): Promise<ExtensionTemplate[]> {
```
shauns, line 292: Would like to see a test before moving out of draft.

shauns (changes requested):

packages/app/src/cli/utilities/developer-platform-client/app-management-client.ts
```diff
@@ -280,8 +289,16 @@ export class AppManagementClient implements DeveloperPlatformClient {
...
+    const allExtensions = await Promise.all(templateJsonUrls.map(async (url) => {
+      const response = await fetch(url)
+      return response.json() as Promise<GatedExtensionTemplate[]>
+    })).then(templates2DArray => templates2DArray.flat(1))
+    const allBetaFlags = Array.from(new Set(allExtensions.map(ext => ext.organizationBetaFlags ?? []).flat()))
+    const enabledBetaFlags = await this.organizationBetaFlags(organizationId, allBetaFlags)
```
gonzaloriestra, line 298: Are we going to check the CLI version in BP as well? Because we will still need to filter templates by that, right?

packages/app/src/cli/models/app/app.test-data.ts
amcaplan: The org ID won't be present if they haven't connected to an app yet. We don't yet support just fetching an org ID without creating an app. (By the way - does that mean we would store it without storing a client ID? Need to support fetching it next time, etc. Not as simple as it seems on first blush)

packages/app/src/cli/utilities/developer-platform-client/app-management-client.ts
amcaplan: We'll investigate the possibility of having just 1 file.

packages/app/src/cli/utilities/developer-platform-client/app-management-client.ts
amcaplan: Screenshot 2024-05-31 at 0 04 25

packages/app/src/cli/utilities/developer-platform-client/app-management-client.ts
gonzaloriestra: We could generate flags with obfuscated names... in public PRs explaining everything 😅

packages/app/src/cli/utilities/developer-platform-client/app-management-client.ts
amcaplan: We have concluded that we'll use 1 file, and the order in the file will be the same as the order displayed in the CLI. (The CLI will still group the options, so that can override order, but we should try and keep each group together in the file for readability.)

packages/app/src/cli/utilities/developer-platform-client/app-management-client.ts
amcaplan: I believe I've tested this reasonably thoroughly, let me know if something is missing.

shauns (approved):

PR #3929, opened 2024-05-24
App management API adjustments

gonzaloriestra:

Fixes #0000 

### Measuring impact

### Checklist

### DIFF
--- a/packages/app/src/cli/utilities/developer-platform-client/app-management-client.ts
+++ b/packages/app/src/cli/utilities/developer-platform-client/app-management-client.ts
@@ -720,34 +720,37 @@ const MAGIC_REDIRECT_URL = 'https://shopify.dev/apps/default-app-home/api/auth'
 
 function createAppVars(name: string, isLaunchable = true, scopesArray?: string[]): CreateAppMutationVariables {
   return {
-    appModules: [
-      {
-        uid: 'app_home',
-        specificationIdentifier: 'app_home',
-        config: JSON.stringify({
-          app_url: isLaunchable ? 'https://example.com' : MAGIC_URL,
-          embedded: isLaunchable,
-        }),
-      },
-      {
-        uid: 'branding',
-        specificationIdentifier: 'branding',
-        config: JSON.stringify({name}),
-      },
-      {
-        uid: 'webhooks',
-        specificationIdentifier: 'webhooks',
-        config: JSON.stringify({api_version: '2024-01'}),
-      },
-      {
-        uid: 'app_access',
-        specificationIdentifier: 'app_access',
-        config: JSON.stringify({
-          redirect_url_allowlist: isLaunchable ? ['https://example.com/api/auth'] : [MAGIC_REDIRECT_URL],
-          ...(scopesArray && {scopes: scopesArray.map((scope) => scope.trim()).join(',')}),
-        }),
-      },
-    ],
+    appSource: {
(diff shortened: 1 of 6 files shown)


packages/app/src/cli/utilities/developer-platform-client/app-management-client.ts
```diff
@@ -720,34 +720,37 @@ const MAGIC_REDIRECT_URL = 'https://shopify.dev/apps/default-app-home/api/auth'
...
-      {
-        uid: 'app_access',
-        specificationIdentifier: 'app_access',
-        config: JSON.stringify({
-          redirect_url_allowlist: isLaunchable ? ['https://example.com/api/auth'] : [MAGIC_REDIRECT_URL],
-          ...(scopesArray && {scopes: scopesArray.map((scope) => scope.trim()).join(',')}),
-        }),
-      },
-    ],
+    appSource: {
```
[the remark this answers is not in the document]
shauns: Ok, so the plan is to move this out of the client and into the platform?

packages/app/src/cli/utilities/developer-platform-client/app-management-client.ts
```diff
@@ -720,34 +723,37 @@ const MAGIC_REDIRECT_URL = 'https://shopify.dev/apps/default-app-home/api/auth'
...
+      modules: [
+        {
+          uid: randomUUID(),
```
amcaplan, line 729: Doesn't this fail the second time you deploy, because the UUID is different so it looks like you have 2 app homes? Maybe better to have a hardcoded UUID.

packages/app/src/cli/models/extensions/extension-instance.ts
```diff
@@ -24,6 +32,21 @@ import {useThemebundling} from '@shopify/cli-kit/node/context/local'
...
+  AppProxySpecIdentifier,
+  BrandingSpecIdentifier,
+  PosSpecIdentifier,
+  appPrivacyComplienceSpec,
+  WebhookSubscriptionSpecIdentifier,
+  WebhooksSpecIdentifier,
+  '0f844fe3-fee9-45ae-9b68-7e596b3eaaa9',
+  '10457bd6-aeea-4f92-ab90-a9ab1e471276',
+  '0ea86845-466c-4f0e-b7f1-9d9496d93f4a',
+  '406bb2df-eaa4-44bc-860d-c714f4b65def',
```
gonzaloriestra, line 47: @amcaplan do you remember what are these ones?

amcaplan: Maybe they were the UUIDs for the Big Four modules? The number 4 generally suggests that direction...

amcaplan: [This commit](https://github.com/Shopify/cli/commit/565ea41e7342755d487409228e93ed8352c8a1d0) certainly indicates as such!

dmerand: It's a bit hard to tell from the context of this PR whether we're talking about the new app management API, the existing Partners API, or both. The new app management API uses the `static` UID strategy from the specification for what we used to call config modules, so technically there's no need to specify a `uid` for them at all -- it will be inferred server-side to be the specification identifier.

amcaplan (approved):

PR #3932, opened 2024-05-24
[Themes][Refactor] - Theme Poll and Theme Reconciliation Cleanup

jamesmengo:
- Housekeeping

- Improve readability in `theme-poll`
- Add a `jsonOnly` option to `applyIgnoreFilters` and use this in `theme-poll` and `theme-reconciliation`

### DIFF
--- a/packages/theme/src/cli/utilities/theme-environment/theme-polling.ts
+++ b/packages/theme/src/cli/utilities/theme-environment/theme-polling.ts
@@ -50,18 +50,18 @@ export async function pollRemoteJsonChanges(
     applyFileFilters(checksums, localFileSystem, options),
   )
 
-  const assetsChangedOnRemote = getChangedAssets(previousChecksums, latestChecksums)
-  const assetsDeletedFromRemote = getDeletedAssets(latestChecksums, previousChecksums)
+  const changedAssets = getAssetsChangedOnRemote(previousChecksums, latestChecksums)
+  const deletedAssets = getAssetsDeletedFromRemote(latestChecksums, previousChecksums)
 
-  await abortIfMultipleSourcesChange(localFileSystem, assetsChangedOnRemote)
+  await abortIfMultipleSourcesChange(localFileSystem, changedAssets)
 
-  await syncChangedAssets(targetTheme, currentSession, localFileSystem, assetsChangedOnRemote)
-  await deleteRemovedAssets(localFileSystem, assetsDeletedFromRemote, options)
+  await syncChangedAssets(targetTheme, currentSession, localFileSystem, changedAssets)
+  await deleteRemovedAssets(localFileSystem, deletedAssets, options)
 
   return latestChecksums
 }
 
-function getDeletedAssets(latestChecksums: Checksum[], previousChecksums: Checksum[]) {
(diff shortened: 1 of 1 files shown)


packages/theme/src/cli/utilities/theme-environment/theme-polling.test.ts
```diff
@@ -179,6 +173,7 @@ describe('pollRemoteJsonChanges', async () => {
       await pollRemoteJsonChanges(developmentTheme, adminSession, remoteChecksums, defaultThemeFileSystem)
 
       // Then
+      expect(fetchThemeAsset).not.toHaveBeenCalled()
```
jamesmengo, line 176: this test was not failing and I realized it's because `fetchThemeAsset` is returning `undefined`

packages/theme/src/cli/utilities/theme-environment/theme-polling.ts
```diff
@@ -45,55 +40,82 @@ export async function pollRemoteJsonChanges(
   localFileSystem: ThemeFileSystem,
   options: PollingOptions = {},
 ): Promise<Checksum[]> {
-  const filteredRemoteChecksums = await applyFileFilters(remoteChecksums, localFileSystem, options)
+  const filterOptions: FilterOptions = {...options, jsonOnly: true}
+  const previousChecksums = await applyIgnoreFilters(remoteChecksums, localFileSystem, filterOptions)
```
jamesmengo, line 44: changes here:
- Use the `jsonOnly` flag
- refactor to be more declarative

packages/theme/src/cli/utilities/theme-environment/theme-polling.ts
```diff
@@ -45,62 +45,84 @@ export async function pollRemoteJsonChanges(
...
 ): Promise<Checksum[]> {
-  const filteredRemoteChecksums = await applyFileFilters(remoteChecksums, localFileSystem, options)
-
+  const previousChecksums = await applyFileFilters(remoteChecksums, localFileSystem, options)
   const latestChecksums = await fetchChecksums(targetTheme.id, currentSession).then((checksums) =>
     applyFileFilters(checksums, localFileSystem, options),
   )
 
-  const previousChecksumsMap = new Map(filteredRemoteChecksums.map((checksum) => [checksum.key, checksum]))
+  const assetsChangedOnRemote = getChangedAssets(previousChecksums, latestChecksums)
```
karreiro, line 53: What do you think if we rename the variables/methods to be more specific in the method name and less specific in the variable name? In other words, we’d have something like this here:

```ts
const changedAssets = getAssetsChangedOnRemote(previousChecksums, latestChecksums)
const deletedAssets = getAssetsDeletedFromRemote(latestChecksums, previousChecksums)
```

(I believe that being more specific in the method name may be more expressive for future readers, as we clarify the meaning of that method)

karreiro (commented):
Thank you, @jamesmengo! The PR looks great 🚀  I've left only one minor comment, please let me know what you think about it :)

karreiro (approved):
Thank you, @jamesmengo!

PR #3933, opened 2024-05-24
Pass Development Server flags down to theme environment

jamesmengo:
- https://github.com/Shopify/develop-advanced-edits/issues/205

Plumbing work for starting the development server 

### How to test your changes?
This doesn't do anything yet but you can `p test`


karreiro (commented):
Thank you, @jamesmengo!

I believe this PR is on the right path. We just need to remove the optional parameters (following what we decided with the previous PR), and then we can rebase and merge it :)

PR #3955, opened 2024-05-30
Transforms from remote to local can receive app configuration

shauns:

Includes https://github.com/Shopify/cli/pull/3956 

Pre-req for webhooks work

Changes the transform from remote to local for extensions to be given the wider app configuration too. This requires storing the app configuration on the ExtensionInstance.

Most of the PR (commit no. 2) is plumbing and filling gaps.

### How to test your changes?

TS/Linting will suffice, this doesn't actually use the new capability

### Measuring impact

### Checklist

### DIFF
--- a/packages/app/src/cli/models/app/app.ts
+++ b/packages/app/src/cli/models/app/app.ts
@@ -67,6 +67,8 @@ export const AppConfigurationSchema = zod.union([LegacyAppSchema, AppSchema])
  */
 export type AppConfiguration = zod.infer<typeof AppConfigurationSchema> & {path: string}
 
+export type AppConfigurationWithoutPath = zod.infer<typeof AppConfigurationSchema>
+
 /**
  * App configuration for a normal, linked, app. Doesn't include properties that are module derived.
  */
--- a/packages/app/src/cli/models/extensions/extension-instance.ts
+++ b/packages/app/src/cli/models/extensions/extension-instance.ts
@@ -48,7 +48,6 @@ export class ExtensionInstance<TConfiguration extends BaseConfigType = BaseConfi
   handle: string
   specification: ExtensionSpecification
   uid: string
-  appConfiguration: AppConfigurationWithoutPath
 
   get graphQLType() {
     return (this.specification.graphQLType ?? this.specification.identifier).toUpperCase()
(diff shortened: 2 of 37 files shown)


packages/app/src/cli/models/app/app.ts
```diff
@@ -67,6 +67,8 @@ export const AppConfigurationSchema = zod.union([LegacyAppSchema, AppSchema])
  */
 export type AppConfiguration = zod.infer<typeof AppConfigurationSchema> & {path: string}
 
+export type AppConfigurationWithoutPath = zod.infer<typeof AppConfigurationSchema>
```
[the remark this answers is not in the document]
shauns: It is a bit tricky. AppConfiguration encompasses having an app that is in the legacy/template state (the schema at the start of that file) as well as one in the more familiar form. Similarly an app/config can be loaded into that state too -- `AppInterface` wraps up config that can be in either format.

I'm trying to chip away at this and change so that the `Current...` config is the one most code sees, as in reality its a very narrow circumstance where we would actually want to work with with one of those old-style files now. The work in the `REDACTED` project will include (I think) a more opinionated app assembly process, where the constructed app is only ever in the up to date format. I'd like to see what breaks with that -- hopefully not much. Then we can simplify a hell of a lot of this.

As-is though, this is the right type to use because the config _can_ be in either format today.

packages/app/src/cli/models/extensions/extension-instance.ts
```diff
@@ -139,6 +142,8 @@ export class ExtensionInstance<TConfiguration extends BaseConfigType = BaseConfi
       const config = this.configuration as unknown as FunctionConfigType
       this.outputPath = joinPath(this.directory, config.build.path ?? joinPath('dist', 'index.wasm'))
     }
+
+    this.appConfiguration = options.appConfiguration
```
isaacroldan, line 146: It's a bit weird that each extension has a copy of the app configuration. Not sure how compatible this would be with `dev` where the app config can be updated, but it won't update every extension's copy 🤔

shauns: Hmm. I suppose an alternate would be to pass in the latest config when the deploy function runs? Or, reload the app when the configuration is changed at runtime -- I suppose this latter is more correct, there are other bugs you could experience if you have a modified configuration object but the appinterface is now out of date

shauns: Ok, will go for first option. I think we should think about how to wrap more of this access and mutation up in the app though. Like, we shouldn't be working with an isolated extension, or taking the config out of the app and changing it

PR #3960, opened 2024-05-31
add a root <Text> element to log files to fix wrapping (3.61 patch)

nickwesselman:
3.61 patch for #3959


shauns:
:shipit: ?

nickwesselman:
@isaacroldan This is a 3.61 patch though, do we need to apply the CI fix on this patch branch too?

isaacroldan:
Um, good point. I can remove the checks of node 18.12 for now as an alternative

isaacroldan:
Actually, those aren't required here. So you can merge 👌

PR #3966, opened 2024-06-03
Dynamic schema from contract

shauns:

Adds support for app config modules that are primarily validated by JSON Schema based contracts. Our first such module is used for custom data definitions.

Commit-by-commit is probably a good way to go through this, as there's some tidying happening too:
- `SpecsAppConfiguration` is no more, and we add `AppConfigurationUsedByCli`, which is a much narrower interface
- Extension specification schemas are no longer exposed, and methods provided instead to validate
- The loader function(s) are now more of a pipeline, which an app conifguration state method available, providing _just enough_ information to e.g. create a dev platform client, prior to loading the app proper. This is important: we need apps to only load having gone through a specification fetch.
- A JSON schema validator is added, with some support for matching Zod behaviour
- An error simplifier for union-style errors is added. It works well for structures that have e.g. a `type` key that acts as a discriminator
- Support for dynamic config entries via an env variable is dropped
- A new factory for modules that are contract based is added

### How to test your changes?

Add:
```
[metaobjects.author.fields]
name = "single_line_text_field"
```
to your config.

You should be able to deploy and `app info --json` this (the latter should have a data section), you should also be able to `link` and pull in the deployed config into the appropriate format.

Testing the contracts themselves is not currently possible as they're not available on the platform _just yet_.

### Measuring impact

### Checklist


packages/app/src/cli/models/app/app.test.ts
```diff
@@ -48,12 +47,6 @@ const CORRECT_CURRENT_APP_SCHEMA: CurrentAppConfiguration = {
     subpath: 'https://google.com',
     prefix: 'https://google.com',
   },
-  pos: {
-    embedded: false,
-  },
-  app_preferences: {
-    url: 'https://google.com',
-  },
```
isaacroldan, line 56: Why removed these?

packages/app/src/cli/utilities/json-schema.ts
```diff
@@ -0,0 +1,57 @@
...
+  merged: RemoteAwareExtensionSpecification & FlattenedRemoteSpecification,
+) {
+  if (merged.contract === undefined) {
+    return merged.parseConfigurationObject
+  }
+  const contract = JSON.parse(merged.contract)
+
+  const parseConfigurationObject = (config: object): ParseConfigurationResult<BaseConfigType> => {
+    // First we parse with zod. This may also change the format of the data.
+    const zodParse = merged.parseConfigurationObject(config)
```
isaacroldan, line 22: Is this initial parse with zod for cases where you need to get some top-keys from the app toml?
The top-keys are not included in the contract?

packages/app/src/cli/models/app/loader.ts
```diff
@@ -639,18 +598,12 @@ class AppLoader<TConfig extends AppConfiguration, TModuleSpec extends ExtensionS
...
-    const unusedExtensionInstances = unusedKeys.map((key) => {
-      // eslint-disable-next-line @typescript-eslint/no-explicit-any
-      const specConfiguration = {[key]: (appConfiguration as any)[key]}
-      return this.createExtensionInstance(key, specConfiguration, appConfiguration.path, directory)
-    })
-
-    // return all the non null extension instances, plus the unused ones
-    const nonNullExtensionInstances: ExtensionInstance[] = extensionInstancesWithKeys
+    if (unusedKeys.length > 0) {
+      throw new Error('Not all of the config was consumed')
```
isaacroldan, line 602: This error could be confusing for partners no?
Do we need to use all keys in the toml? should we show a better message if we aren't?

packages/app/src/cli/models/app/app.test.ts
shauns: They're not part of the config required by CLI -- the values were only set here -- and I figured the right shape for this object is to be minimal

packages/app/src/cli/models/extensions/specification.ts
```diff
@@ -173,6 +222,47 @@ export function createConfigExtensionSpecification<TConfiguration extends BaseCo
...
+    schema,
+    transformConfig: {
+      // outgoing config is already scoped to this module and passed directly along
+      forward(obj) {
+        return obj
+      },
+      // incoming config from the platform is included in app config as-is
+      reverse(obj) {
+        return obj
+      },
```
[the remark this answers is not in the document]
shauns: `createConfigExtensionSpecification` has default transformations that are a little different, so it's here to override those. If you don't provide any, the default transformations take over.

I guess one of the transformation approaches needs to be the default? Or am I misunderstanding your commend?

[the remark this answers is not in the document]
shauns: Yeah, the current default is to do some flattening. I don't mind changing it at all. It'll just mean updating the callers for the current config modules. That's no trouble. Your call!

isaacroldan (approved):

PR #3979, opened 2024-06-04
Revamp missing export behaviour

DuncanUszkay1:

Fixes https://github.com/Shopify/function-runner/issues/270

Our current behaviour for a missing import is rather confusing- it defaults to `_start`, and if `_start` isn't present it fails, tries to read the output, and returns a JSON parsing error.

The behaviour proposed here is as follows:
- When the user has passed in the export flag, that's always used over anything else
- When the function has a list of targets which have exports
  - If they have a list of targets and an interactive terminal, show them a prompt to pick one
  - If the terminal isn't interactive, or if there's only one option to select, select the first option
  - When exports are missing, default to `_start`
- When there's no targeting information or flag, use `_start`

I've added log statements to explain to the user why a particular export was selected.

### How to test your changes?

Pending

### Post-release steps

None, as far as I know.

### Measuring impact

### Checklist

### DIFF
--- a/packages/app/src/cli/commands/app/function/run.ts
+++ b/packages/app/src/cli/commands/app/function/run.ts
@@ -4,6 +4,10 @@ import {appFlags} from '../../../flags.js'
 import Command from '@shopify/cli-kit/node/base-command'
 import {globalFlags} from '@shopify/cli-kit/node/cli'
 import {Flags} from '@oclif/core'
+import {renderAutocompletePrompt, isTTY} from '@shopify/cli-kit/node/ui'
+import {outputDebug} from '@shopify/cli-kit/node/output'
+
+const DEFAULT_FUNCTION_EXPORT = '_start'
 
 export default class FunctionRun extends Command {
   static summary = 'Run a function locally for testing.'
@@ -25,7 +29,6 @@ export default class FunctionRun extends Command {
       char: 'e',
       hidden: false,
       description: 'Name of the WebAssembly export to invoke.',
-      default: '_start',
       env: 'SHOPIFY_FLAG_EXPORT',
     }),
     json: Flags.boolean({
@@ -42,10 +45,43 @@ export default class FunctionRun extends Command {
       path: flags.path,
       userProvidedConfigName: flags.config,
       callback: async (_app, ourFunction) => {
+        let functionExport = DEFAULT_FUNCTION_EXPORT
+
+        if (flags.export !== undefined) {
+          outputDebug(`Using export ${flags.export} from the --export flag.`)
+          functionExport = flags.export
+        } else if (
+          ourFunction.configuration.targeting !== undefined &&
+          ourFunction.configuration.targeting.length > 0
+        ) {
+          const targeting = ourFunction.configuration.targeting
+
+          if (targeting.length > 1 && isTTY({})) {
+            const targets = targeting.map((target) => ({
+              label: target.target,
+              value: target.export || DEFAULT_FUNCTION_EXPORT,
+            }))
+
+            functionExport = await renderAutocompletePrompt({
+              message: `Which target would you like to execute?`,
+              choices: targets,
+            })
+          } else {
+            functionExport = targeting?.[0]?.export || DEFAULT_FUNCTION_EXPORT
+            outputDebug(
+              `Using export '${functionExport}'. Use the --export flag or an interactive terminal to select a different export.`,
+            )
+          }
+        } else {
+          outputDebug(
+            `No targeting information found. Using the default export '${functionExport}'. Use the --export flag or an interactive terminal to select a different export.`,
+          )
+        }
+
         await runFunctionRunner(ourFunction, {
           json: flags.json,
           input: flags.input,
-          export: flags.export,
+          export: functionExport,
         })
       },
     })
--- a/packages/cli-kit/src/public/node/ui.tsx
+++ b/packages/cli-kit/src/public/node/ui.tsx
@@ -686,6 +686,15 @@ export const keypress = async (stdin = process.stdin, uiDebugOptions: UIDebugOpt
   })
 }
 
+interface IsTTYOptions {
+  stdin?: NodeJS.ReadStream
+  uiDebugOptions?: UIDebugOptions
+}
+
+export function isTTY({stdin = undefined, uiDebugOptions = defaultUIDebugOptions}: IsTTYOptions) {
+  return Boolean(uiDebugOptions.skipTTYCheck || stdin || terminalSupportsRawMode())
+}
+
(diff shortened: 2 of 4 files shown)


packages/cli-kit/src/public/node/ui.tsx
```diff
@@ -686,14 +686,18 @@ export const keypress = async (stdin = process.stdin, uiDebugOptions: UIDebugOpt
   })
 }
 
+export function isTTY(stdin: NodeJS.ReadStream | undefined = undefined, uiDebugOptions: UIDebugOptions = defaultUIDebugOptions) {
+  return uiDebugOptions.skipTTYCheck || stdin || terminalSupportsRawMode()
+}
+
```
DuncanUszkay1, line 692: Is there an existing utility for this? 

We expect folks to pass input into this method programmatically in regular use, so I don't want the command to crash out if they're not in an interactive terminal.

packages/app/src/cli/commands/app/function/run.ts
```diff
@@ -42,10 +42,43 @@ export default class FunctionRun extends Command {
       path: flags.path,
       userProvidedConfigName: flags.config,
       callback: async (_app, ourFunction) => {
+        let functionExport = '_start'
+
+        if (flags.export !== undefined) {
+          this.log('Using export ${flags.export} from the --export flag.')
```
nickwesselman, line 48: I wonder if these logs should be debug/verbose output, and the Function Runner itself should just report what export it's running.

packages/app/src/cli/commands/app/function/run.ts
```diff
@@ -42,10 +42,43 @@ export default class FunctionRun extends Command {
...
+          functionExport = flags.export
+        } else if (
+          ourFunction.configuration.targeting !== undefined &&
```
nickwesselman, line 51: Is this the indentation that the CLI linting/prettier config suggested?

packages/app/src/cli/commands/app/function/run.ts
```diff
@@ -42,10 +42,43 @@ export default class FunctionRun extends Command {
...
+            let targets = targeting.map((target) => ({
+              label: target.target,
+              value: target.export || '_start',
+            }))
+
+            functionExport = await renderAutocompletePrompt({
+              message: `Which target would you like to execute?`,
+              choices: targets,
+            })
+          } else if (targeting.length === 1) {
```
nickwesselman, line 66: You know here that it's greater than 0 and not greater than 1 so I think `=== 1` is the only possibility for this code branch

packages/app/src/cli/commands/app/function/run.ts
```diff
@@ -42,10 +42,43 @@ export default class FunctionRun extends Command {
...
+            functionExport = targeting[0]?.export || '_start'
```
nickwesselman, line 67: Maybe put `_start` in a const in the module?

packages/app/src/cli/commands/app/function/run.ts
andrewhassan, line 48: We have to use the backticks for this to be a template string.

```suggestion
          this.log(`Using export ${flags.export} from the --export flag.`)
```

andrewhassan: Also I'm not sure what `this.log` is referring to here. Do you mean for this to be a debug log?

packages/app/src/cli/commands/app/function/run.ts
```diff
@@ -42,10 +42,43 @@ export default class FunctionRun extends Command {
...
+        if (flags.export !== undefined) {
+          this.log('Using export ${flags.export} from the --export flag.')
+          functionExport = flags.export
+        } else if (
+          ourFunction.configuration.targeting !== undefined &&
+          ourFunction.configuration.targeting.length > 0
+        ) {
+          const targeting = ourFunction.configuration.targeting
+
+          if (targeting.length > 1 && isTTY()) {
```
andrewhassan, line 56: If there are > 1 targets and this isn't interactive, wouldn't it not fall into the `else if`, which means the export is `_start`. Based on the PR description, I think we want to choose the first export in that case.

packages/cli-kit/src/public/node/ui.tsx
```diff
@@ -686,14 +686,18 @@ export const keypress = async (stdin = process.stdin, uiDebugOptions: UIDebugOpt
   })
 }
 
+export function isTTY(stdin: NodeJS.ReadStream | undefined = undefined, uiDebugOptions: UIDebugOptions = defaultUIDebugOptions) {
+  return uiDebugOptions.skipTTYCheck || stdin || terminalSupportsRawMode()
```
andrewhassan, line 690: We might want to wrap this with a `Boolean` so that we don't accidentally return one of these values.

```suggestion
  return Boolean(uiDebugOptions.skipTTYCheck || stdin || terminalSupportsRawMode())
```

packages/app/src/cli/commands/app/function/run.ts
DuncanUszkay1: Hm, yeah these should all be shifted to debug

packages/app/src/cli/commands/app/function/run.ts
shauns: `:shakes-fist-at-prettier:`

packages/app/src/cli/commands/app/function/run.ts
```diff
@@ -42,10 +44,43 @@ export default class FunctionRun extends Command {
...
+              choices: targets,
+            })
+          } else {
+            functionExport = targeting?.[0]?.export || DEFAULT_FUNCTION_EXPORT
+            debug(
+              `Using export '${functionExport}'. Use the --export flag or an interactive terminal to select a different export.`,
+            )
+          }
+        } else {
+          debug(
```
shauns, line 75: These would normal be `outputDebug` -- they'll work with the normal `--verbose` option then

packages/app/src/cli/commands/app/function/run.ts
```diff
@@ -42,10 +44,43 @@ export default class FunctionRun extends Command {
...
+          debug(`Using export ${flags.export} from the --export flag.`)
+          functionExport = flags.export
+        } else if (
+          ourFunction.configuration.targeting !== undefined &&
+          ourFunction.configuration.targeting.length > 0
+        ) {
+          const targeting = ourFunction.configuration.targeting
+
+          if (targeting.length > 1 && isTTY()) {
+            let targets = targeting.map((target) => ({
```
shauns, line 59: I'm not 100% on where these come from, but if they're not ordered at the source, it might be worth sorting them first.

PR #3980, opened 2024-06-04
Resubscribe when token expires

DuncanUszkay1:

Fixes https://github.com/Shopify/shopify-functions/issues/211

Previously, when a token expired, the dev session would simply crash. Instead, we want to restart the polling process to the point prior to getting the token, so that the token will be refreshed and the session can go on perpetually.

### How to test your changes?

- Update partners in a spin environment to always return an expired token (unless you happen to have a 24 hour token for production on hand)
- Turn on app dev on verbose mode
- See that it's continually resubscribing and getting a new token

### Measuring impact

### Checklist

### DIFF
--- a/packages/app/src/cli/services/app-logs/poll-app-logs.ts
+++ b/packages/app/src/cli/services/app-logs/poll-app-logs.ts
@@ -26,10 +26,12 @@ export const pollAppLogs = async ({
   stdout,
   appLogsFetchInput: {jwtToken, cursor},
   apiKey,
+  resubscribeCallback,
 }: {
   stdout: Writable
   appLogsFetchInput: {jwtToken: string; cursor?: string}
   apiKey: string
+  resubscribeCallback: () => Promise<void>
 }) => {
   const url = await generateFetchAppLogUrl(cursor)
   const response = await fetch(url, {
@@ -38,6 +40,11 @@ export const pollAppLogs = async ({
       Authorization: `Bearer ${jwtToken}`,
     },
   })
+
+  if (response.status === 401) {
+    await resubscribeCallback()
+    return
+  }
 
   if (!response.ok) {
     // We should add some exponential backoff here to not spam partners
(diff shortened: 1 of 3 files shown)


packages/app/src/cli/services/dev/processes/app-logs-polling.ts
```diff
@@ -46,7 +46,7 @@ export async function setupAppLogsPollingProcess({
 }
 
 export const subscribeAndStartPolling: DevProcessFunction<SubscribeAndStartPollingOptions> = async (
-  {stdout},
+  {stdout, stderr, abortSignal},
```
DuncanUszkay1, line 49: The `DevProcessFunction` demands to have these available when this is called, so I'm passing them in here. We only use them to re-call this dev process function.

packages/app/src/cli/services/app-logs/poll-app-logs.ts
```diff
@@ -39,6 +41,11 @@ export const pollAppLogs = async ({
     },
   })
 
+  if (response.status == 401) {
+    await resubscribe_callback()
```
DuncanUszkay1, line 45: If we're continually getting tokens from partners then hitting 401s on the poll endpoint, this would eventually overflow the stack. That could happen if Partners was producing tokens that were already expired for instance, but shouldn't happen under any normal circumstances.

I think it's fine to just accept that risk here rather than complicating this process.

packages/app/src/cli/services/app-logs/poll-app-logs.ts
```diff
@@ -39,6 +41,11 @@ export const pollAppLogs = async ({
     },
   })
 
+  if (response.status === 401) {
+    await resubscribeCallback()
+    return
+  }
```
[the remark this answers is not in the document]
DuncanUszkay1: >since the callback ends up going through the whole flow again (subscribe -> poll app logs -> setTimeout to call -> poll app logs -> setTimeout to call -> ...), at what point does this stop awaiting so that it can return here?

In a normal scenario, it'll stop waiting after the first successful poll

lopert (approved):

DuncanUszkay1:
Don't see a clear correlation with those checks to this PR, will try merging again

PR #3981, opened 2024-06-04
App logs command

mssalemi:
Part of the LogStreaming Prototype project.

#gsd:39776

Fixes #0000 

### Measuring impact

### Checklist


packages/app/src/cli/services/app-logs/services/output-functions.ts
```diff
@@ -0,0 +1,111 @@
...
+  AppLogsPollingCommandErrorOutputFunction,
+} from '../types.js'
+import {writeAppLogsToFile} from '../write-app-logs.js'
+import {useConcurrentOutputContext} from '@shopify/cli-kit/node/ui/components'
+import {Response} from 'node-fetch'
+import {Writable} from 'stream'
+
+const POLLING_BACKOFF_INTERVAL_MS = 10000
+const ONE_MILLION = 1000000
+
```
mssalemi, line 13: Added this file to organize the code a bit better so the `pollAppLogs` can be more re-useable. 

This requires passing in 3 'output functions':
- 1 handle the main output
- 1 handle the errors output
- 1 handle the retry output

`dev` command uses the `ConcurrentOutput` component, which handles what gets written to the `stdout`. Our DX for `logs` is much different than that of `dev`, so we cannot re-use the same `ConcurrentOutput` component. 

The output functions will handle how a command will write to the `stdout`. When writing to the `stdout`, that is where the react component that CLI uses can display the data. So these functions are a bit dependent on how the CLI react components use the writeable stream. See examples for dev [here](https://github.com/Shopify/cli/blob/main/packages/cli-kit/src/private/node/ui/components/ConcurrentOutput.tsx#L127-L143), and the `Logs` component added in this PR.

packages/app/src/cli/services/app-logs/services/output-functions.ts
```diff
@@ -0,0 +1,111 @@
...
+  stdout: Writable
+  response: Response
+  apiKey?: string
+}) => {
+  const responseText = await response.text()
+  // stdout.write(`Received an error while polling for app logs.`)
+  // stdout.write(`${response.status}: ${response.statusText}`)
+  // stdout.write(responseText)
+  // stdout.write(`Retrying in ${POLLING_BACKOFF_INTERVAL_MS / 1000} seconds`)
+}
```
mssalemi, line 75: Note: Add custom retry logic here.

PR #3983, opened 2024-06-05
Improve global cli detection

isaacroldan:
Our way to detect if we were running globally or locally wasn't very robust and failed in some cases.

- Implement a new way to decide if we are running globally or not by using `npm prefix`.
  - `npm prefix` tells you the path to the root of your project, and with `argv` we know what javascript file we are executing. If `argv` starts with same path as `npm prefix` we are certain to be running a local javascript file (inside node_modules) 

- Move `sniffForPath` to a more appropriate location.
- Add `sniffForJson` to detect if the `--json` flag is being used.
- Don't show the global CLI warning if using `--json`
- Force hydrogen's `cli-kit` and `plugin-cloudflare` resolution to use the local version of those packages.

### How to test your changes?
This branch has been released to experimental: `0.0.0-experimental-20240610101312)`

You can test basic usage with an app, but basically changes in this PR are covered by tests in `is-global`.
Running `npm run shopify` inside an app project (that doesn't have a local CLI dependency) should run the global CLI and detect it has such. (previously it was detected as a local CLI)

For instance, running `app info` will show a message like `To change these, run "npm run dev -- --reset"`. That message should change if we detect a local vs a global CLI

### Measuring impact

### Checklist


packages/cli-kit/src/public/node/is-global.ts
```diff
@@ -1,18 +1,38 @@
...
+export function currentProcessIsGlobal(argv = process.argv): boolean {
+  // If we are running tests, we need to disable the cache
+  if (_isGlobal !== undefined && !isUnitTest()) return _isGlobal
+
+  // Path where the current project is (app/hydrogen)
+  const path = sniffForPath() ?? cwd()
+
+  // Closest parent directory to contain a package.json file or node_modules directory
+  // https://docs.npmjs.com/cli/v8/commands/npm-prefix#description
+  const npmPrefix = execaSync('npm', ['prefix'], {cwd: path}).stdout.trim()
```
shauns, line 26: You mentioned in the team Slack that this is more robust than looking at the environment variable -- could you expand a bit on why? Running an executable, even if quick, seems like it'd be more expensive.

Also, only tangently related -- how often do we call this? is it cached (in-memory)?

PR #4001, opened 2024-06-06
Print error and continue dev on log streaming subscription failure

andrewhassan:

Fixes https://github.com/Shopify/shopify-functions/issues/220

In cases where we encounter an error creating an app log streaming subscription (e.g. the user doesn't have the necessary permissions), the CLI prints the error and continues. This leads to a downstream error halting the `dev` process.

This PR updates the logic such that when there is an error subscribing, the error is printed out and the `dev` process continues.

### How to test your changes?

- Set the `SHOPIFY_CLI_ENABLE_APP_LOG_POLLING=1` and the organization beta flag off
- Start `app dev` and see that an error message is printed, but `dev` continues

### Screenshots

image

### Checklist

- [ ] ~I've made sure that any changes to `dev` or `deploy` have been reflected in the [internal flowchart](https://www.figma.com/file/7vqUp50u6dm48Zfb4JRRn8/CLI3-Internals).~ This link is dead 😢


DuncanUszkay1:
Makes sense to me, but we should add a test to cover this

lopert (approved):

DuncanUszkay1 (approved):
Thanks for taking the time to write up the test :+1:

PR #4008, opened 2024-06-07
Retry app log polling after 429 or 5XX

lopert:

Fixes https://github.com/Shopify/shopify-functions/issues/206

In the event of a 429 or 5XX error, enqueues another attempt at polling.
This prevents `dev` from crashing whenever there's a server error, and also allows the server time to recover instead of continuing to poll with the regular frequency.

### How to test your changes?
Set the app logs controller in partners to always return an error during polling. You can even do this while polling to see different error messages / handling.
My 🎩 

### Measuring impact

### Checklist

### DIFF
--- a/packages/app/src/cli/services/app-logs/poll-app-logs.ts
+++ b/packages/app/src/cli/services/app-logs/poll-app-logs.ts
@@ -5,6 +5,7 @@ import {fetch} from '@shopify/cli-kit/node/http'
 import {Writable} from 'stream'
 
 const POLLING_INTERVAL_MS = 450
+const POLLING_BACKOFF_INTERVAL_MS = 10000
 const ONE_MILLION = 1000000
 
 const generateFetchAppLogUrl = async (cursor?: string) => {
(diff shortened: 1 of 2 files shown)


packages/app/src/cli/services/app-logs/poll-app-logs.ts
```diff
@@ -41,15 +42,31 @@ export const pollAppLogs = async ({
...
+          stdout,
+          appLogsFetchInput: {
+            jwtToken,
+            cursor: undefined,
+          },
+          apiKey,
+          resubscribeCallback,
+        }).catch((error) => {
+          throw new Error(`${error} error while fetching.`)
+        })
```
lopert, line 66: Added this because of the lint rule, but it feels odd. There's a few different places now that all have similar catch/re-throw lines. Any thoughts on a way to clean this up while still keeping the linter happy?

DuncanUszkay1: What was the lint violation?

mssalemi (approved):
LGTM! For the lint rule, I think its ok given we do something similar later in this file.

packages/app/src/cli/services/app-logs/poll-app-logs.ts
```diff
@@ -41,15 +42,31 @@ export const pollAppLogs = async ({
...
+        pollAppLogs({
+          stdout,
+          appLogsFetchInput: {
+            jwtToken,
+            cursor: undefined,
+          },
+          apiKey,
+          resubscribeCallback,
+        }).catch((error) => {
+          throw new Error(`${error} error while fetching.`)
```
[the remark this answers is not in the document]
lopert: Doesn't look like it does, I'll look into this a bit more because the output is a bit sad:
```
Error: Error: Error while fetching: {"errors":["hadoken"]} error while fetching.
```

PR #4023, opened 2024-06-10
Cache CLI package version

amcaplan:

`shopify version` should be fast. Checking for a new version makes it slower. We should do this a max of once daily.

Limits checks to once daily, otherwise we use the cached value.

### How to test your changes?

`p shopify:run version` should become faster. No wait between displaying current local version and latest version on npm.

### Measuring impact

### Checklist

### DIFF
--- a/packages/cli-kit/src/public/node/node-package-manager.ts
+++ b/packages/cli-kit/src/public/node/node-package-manager.ts
@@ -5,11 +5,11 @@ import {fileExists, readFile, writeFile, findPathUp, glob} from './fs.js'
 import {dirname, joinPath} from './path.js'
 import {runWithTimer} from './metadata.js'
 import {outputToken, outputContent, outputDebug} from '../../public/node/output.js'
+import {PackageVersionKey, cacheRetrieveOrRepopulate} from '../../private/node/conf-store.js'
 import latestVersion from 'latest-version'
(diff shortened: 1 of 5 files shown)


gonzaloriestra (approved):
Not sure if people usually run `shopify version` more than once per day, but the code is quite simple 👍

packages/cli-kit/src/public/node/node-package-manager.ts
```diff
@@ -248,21 +249,36 @@ export async function usesWorkspaces(appDirectory: string): Promise<boolean> {
...
+  const cacheKey: PackageVersionKey = `npm-package-${dependency}`
+  let lastVersion
   try {
-    const lastVersion = await getLatestNPMPackageVersion(dependency)
-    if (lastVersion && new SemVer(currentVersion).compare(lastVersion) < 0) {
-      return lastVersion
-    } else {
-      return undefined
-    }
+    lastVersion = await cacheRetrieveOrRepopulate(cacheKey, getLatestVersion, cacheExpiryInHours * 24 * 1000)
```
gonzaloriestra, line 271: We need to convert from hours to miliseconds, no? 🤔 

```suggestion
    lastVersion = await cacheRetrieveOrRepopulate(cacheKey, getLatestVersion, cacheExpiryInHours * 3600 * 1000)
```

PR #4032, opened 2024-06-11
Function log differentiation and file linking

nickwesselman:

Multiple runs of the same function run together visually:
image

Additionally, the log file path is a bit long and frequently wraps, limiting its usefulness.

* Prefix function log output with box drawing characters, to better group related lines.
* Output function log file as a link instead of a file path, on terminals that support hyperlinks.

_VS Code_
image

_MacOS Terminal (no hyperlink support)_
image

_Windows Terminal (hyperlink disabled by default in `supports-hyperlink`_
windows terminal

_Windows Terminal w/ hyperlinks forced_
windows terminal FORCE_HYPERLINK

#### Supporting changes
* Allow hyperlink tokens to provide an explicit fallback value.
* Added a `gray` coloring output token.
* Allow disabling of ansi stripping in `ConcurrentOutput` via `useConcurrentOutputContext`.

### How to test your changes?
Run `app dev` in an app w/ functions and function log streaming enabled. Execute the function.

### Checklist

- [X] I've made sure that any changes to `dev` or `deploy` have been reflected in the [internal flowchart](https://www.figma.com/file/7vqUp50u6dm48Zfb4JRRn8/CLI3-Internals).

### DIFF
--- a/packages/app/src/cli/services/app-logs/poll-app-logs.ts
+++ b/packages/app/src/cli/services/app-logs/poll-app-logs.ts
@@ -100,17 +100,24 @@ export const pollAppLogs = async ({
 
             const logs = payload.logs
             if (logs.length > 0) {
-              stdout.write(logs)
+              stdout.write(
+                logs
+                  .split('\n')
+                  .filter(Boolean)
+                  .map((line: string) => `├ ${line}`)
+                  .join('\n'),
+              )
             }
           } else {
             stdout.write(JSON.stringify(payload))
           }
 
-          await writeAppLogsToFile({
+          const logPath = await writeAppLogsToFile({
             appLog: log,
             apiKey,
             stdout,
           })
+          stdout.write(`└ Log: ${logPath}\n`)
         })
       }
     }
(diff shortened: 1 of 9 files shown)


packages/app/src/cli/services/app-logs/poll-app-logs.ts
```diff
@@ -100,17 +100,30 @@ export const pollAppLogs = async ({
...
-          await writeAppLogsToFile({
+          const logPath = await writeAppLogsToFile({
             appLog: log,
             apiKey,
             stdout,
           })
+          stdout.write(
+            outputContent`${outputToken.gray('└ ')}${outputToken.link(
+              'Open log file',
+              `file://${logPath}`,
```
shauns, line 123: re: Windows -- might need to check that these file paths are valid there. e.g. if `logPath` is like `C:\something`.

packages/cli-kit/src/private/node/content-tokens.ts
```diff
@@ -23,16 +23,18 @@ export class RawContentToken extends ContentToken<string> {
 
 export class LinkContentToken extends ContentToken<OutputMessage> {
   link: string
+  fallback: string | undefined
 
-  constructor(value: OutputMessage, link: string) {
+  constructor(value: OutputMessage, link: string, fallback?: string) {
```
shauns, line 28: Nice 👌

shauns (approved):

isaacroldan (approved):

packages/app/src/cli/services/app-logs/poll-app-logs.ts
nickwesselman: I discovered that the forward slashes are actually due to our use of [`pathe`](https://www.npmjs.com/package/pathe), which presupposes that forward slashes are OK in Windows these days. Which I think for our use case, that would be true. So even w/ hyperlinks disabled, we are outputting forward slashes at the moment.

PR #4035, opened 2024-06-12
[Log Streaming]Add nullable log_type to AppEventData for incoming API changes

otheriault:

Fixes https://github.com/Shopify/shopify-functions/issues/171.
Blocks https://github.com/Shopify/partners/pull/55178.

Attribute currently named `event_type` is being renamed to `log_type` on the log polling API.

We are simply adding support for the new `log_type` API property and supporting both the legacy `event_type` and the new `log_type` properties simultaneously to avoid breaking changes.

### How to test your changes?

Simply use the log streaming feature as normal and everything should just keep working.
* Remember to enable the beta flag `Enable App Logs Streaming` to your organization in partners internal
* Remember to enable log streaming from the CLI with `export SHOPIFY_CLI_ENABLE_APP_LOG_POLLING=1`
* Use the CLI from this branch with command `dev shopify app dev --path [path to your app with a function in it]`
* Remember to create the object needed to trigger your function (ex: discount) and activate it.
* Trigger your function from your dev store.
* See the logs in your CLI console

### Measuring impact

### Checklist

### DIFF
--- a/packages/app/src/cli/services/app-logs/poll-app-logs.test.ts
+++ b/packages/app/src/cli/services/app-logs/poll-app-logs.test.ts
@@ -67,7 +67,7 @@ const FAILURE_PAYLOAD = {
   output_bytes: 182,
   function_id: 'e57b4d31-2038-49ff-a0a1-1eea532414f7',
   logs: LOGS,
-  error_type: FUNCTION_ERROR
+  error_type: FUNCTION_ERROR,
 }
 const OTHER_PAYLOAD = {some: 'arbitrary payload'}
 const RETURNED_CURSOR = '2024-05-23T19:17:02.321773Z'
--- a/packages/app/src/cli/services/app-logs/poll-app-logs.test.ts
+++ b/packages/app/src/cli/services/app-logs/poll-app-logs.test.ts
@@ -14,6 +14,9 @@ vi.mock('@shopify/cli-kit/node/http')
 
 const FQDN = await partnersFqdn()
 const LOGS = '1\\n2\\n3\\n4\\n'
+const FUNCTION_ERROR = 'function_error'
+const FUNCTION_RUN = 'function_run'
+
 const INPUT = {
   cart: {
     lines: [
@@ -57,6 +60,15 @@ const FUNCTION_PAYLOAD = {
   logs: LOGS,
   fuel_consumed: 512436,
 }
+const FAILURE_PAYLOAD = {
+  input: JSON.stringify(INPUT),
+  input_bytes: 123,
+  output: JSON.stringify(OUTPUT),
+  output_bytes: 182,
+  function_id: 'e57b4d31-2038-49ff-a0a1-1eea532414f7',
+  logs: LOGS,
+  error_type: FUNCTION_ERROR
+}
 const OTHER_PAYLOAD = {some: 'arbitrary payload'}
 const RETURNED_CURSOR = '2024-05-23T19:17:02.321773Z'
 const RESPONSE_DATA = {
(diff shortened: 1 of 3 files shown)


packages/app/src/cli/services/app-logs/poll-app-logs.ts
```diff
@@ -89,7 +91,8 @@ export const pollAppLogs = async ({
 
         // eslint-disable-next-line no-await-in-loop
         await useConcurrentOutputContext({outputPrefix: log.source, stripAnsi: false}, async () => {
-          if (log.event_type === 'function_run') {
+          // Use only log.log_type after https://github.com/Shopify/partners/pull/55178
+          if ((log.log_type === undefined ? log.event_type : log.log_type) === LOG_TYPE_FUNCTION_RUN) {
```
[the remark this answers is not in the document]
otheriault: That other PR will remove `event_type` from the API response, so this will mean that it will break the current behavior. I will merge the partners PR after this one, then open another CLI PR to only use `log_type`.

This way nothing *should break for anyone.

packages/app/src/cli/services/app-logs/poll-app-logs.ts
```diff
@@ -89,7 +90,8 @@ export const pollAppLogs = async ({
 
         // eslint-disable-next-line no-await-in-loop
         await useConcurrentOutputContext({outputPrefix: log.source}, async () => {
-          if (log.event_type === 'function_run') {
+          // Use only log.log_type after https://github.com/Shopify/partners/pull/55178
+          if ((log.log_type === undefined ? log.event_type : log.log_type) === 'function_run') {
```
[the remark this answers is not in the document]
otheriault: Done. Also added test case for `log.status === 'failure'`.

packages/app/src/cli/services/app-logs/poll-app-logs.ts
andrewhassan, line 95: Nit: We can use the `??` operator to simplify this 😄 

```suggestion
          if ((log.log_type ?? log.event_type) === LOG_TYPE_FUNCTION_RUN) {
```

DuncanUszkay1 (approved):

packages/app/src/cli/services/app-logs/poll-app-logs.test.ts
```diff
@@ -148,17 +182,40 @@ describe('pollAppLogs', () => {
...
+
+    // app_logs[1]
+    expect(stdout.write).toHaveBeenCalledWith(`❌ Function failed to execute with error: ${FUNCTION_ERROR}`)
+    expect(stdout.write).toHaveBeenCalledWith(expect.stringContaining(LOGS))
+    expect(stdout.write).toHaveBeenCalledWith(expect.stringContaining('Log: '))
+
+    // app_logs[2]
+    expect(stdout.write).toHaveBeenCalledWith('Function executed successfully using 0.5124M instructions.')
+    expect(stdout.write).toHaveBeenCalledWith(expect.stringContaining(LOGS))
+    expect(stdout.write).toHaveBeenCalledWith(expect.stringContaining('Log: '))
```
andrewhassan, line 214: I don't think these assertions check the call order so this one would pass even if only the `app_logs[0]` output was printed once.

You could probably use [this](https://jestjs.io/docs/expect#tohavebeennthcalledwithnthcall-arg1-arg2-) or [this](https://jestjs.io/docs/expect#tohavebeencalledtimesnumber) to validate that it was called multiple times as you expected.

packages/app/src/cli/services/app-logs/write-app-logs.test.ts
```diff
@@ -42,6 +43,25 @@ describe('writeAppLogsToFile', () => {
...
+      ...APP_LOG,
+      log_type: 'function_run',
+    }
+    const logData = expectedLogDataFromAppEvent(appLog)
+
+    // determine the fileName and path
+    const fileName = `20240522_150641_827Z_${APP_LOG.source_namespace}_${APP_LOG.source}`
+    const path = joinPath(API_KEY, fileName)
+
+    const returnedPath = await writeAppLogsToFile({appLog, apiKey: API_KEY, stdout})
```
andrewhassan, line 58: Super nitpick: I think the CLI team tries to annotate given/when/then

```suggestion
    // When
    const returnedPath = await writeAppLogsToFile({appLog, apiKey: API_KEY, stdout})
```

packages/app/src/cli/services/app-logs/write-app-logs.test.ts
```diff
@@ -42,6 +43,25 @@ describe('writeAppLogsToFile', () => {
     expect(writeLog).toHaveBeenCalledWith(expect.stringContaining(path), logData)
   })
 
+  test('calls writeLog with the right data with log_type !== undefined', async () => {
+    // Given
+    const appLog = {
+      ...APP_LOG,
+      log_type: 'function_run',
```
andrewhassan, line 50: Should this also set `event_type` to `undefined` to simulate the case where that's not returned anymore?

andrewhassan (commented):
Overall looks pretty good! Thanks for doing this 😄

mssalemi (approved):
🎩 ands code LGTM! :D

packages/app/src/cli/services/app-logs/poll-app-logs.test.ts
otheriault: So I could see that the spy was invoked 22 times, 11 times per API invocation. and I am quite confident that the above code checks the first 11 spy invocations in order.

packages/app/src/cli/services/app-logs/write-app-logs.test.ts
otheriault: Well technically the interpreter *should shortcut and not even check the `event_type` once it sees that `log_type` is defined & not null, so the value of `event_type` is not relevant for this test.

packages/app/src/cli/services/app-logs/poll-app-logs.test.ts
otheriault: done

PR #4040, opened 2024-06-12
Handle transient failures during app log polling

lopert:

Fixes https://github.com/Shopify/script-editor/issues/4145

Changes the behaviour when we encounter an unexpected error.
Previously, we would catch the error and stop the polling.
Now, we always enqueue another polling attempt regardless of the error.

### How to test your changes?

- run dev targetting your spin instance
- change partners app logs controller to return various errors, i.e.
```ruby
return render json: { errors: ['some text'] }, status: 422
```

### Measuring impact

### Checklist

### DIFF
--- a/packages/app/src/cli/services/app-logs/poll-app-logs.test.ts
+++ b/packages/app/src/cli/services/app-logs/poll-app-logs.test.ts
@@ -165,8 +165,6 @@ describe('pollAppLogs', () => {
 
   test('calls resubscribe callback if a 401 is received', async () => {
     // Given
-    const url = `https://${FQDN}/app_logs/poll`
-
     const response = new Response('errorMessage', {status: 401})
     const mockedFetch = vi.fn().mockResolvedValueOnce(response)
     vi.mocked(fetch).mockImplementation(mockedFetch)
@@ -182,14 +180,10 @@ describe('pollAppLogs', () => {
     expect(MOCKED_RESUBSCRIBE_CALLBACK).toHaveBeenCalled()
   })
 
-  test('displays error, waits, and retries if status is 429 or >500', async () => {
+  test('displays throttle message, waits, and retries if status is 429', async () => {
     // Given
-    const url = `https://${FQDN}/app_logs/poll`
-
-    const mockedFetch = vi
-      .fn()
-      .mockResolvedValueOnce(new Response('error for 429', {status: 429}))
-      .mockResolvedValueOnce(new Response('error for 500', {status: 500}))
+    const outputWarnSpy = vi.spyOn(output, 'outputWarn')
+    const mockedFetch = vi.fn().mockResolvedValueOnce(new Response('error for 429', {status: 429}))
     vi.mocked(fetch).mockImplementation(mockedFetch)
 
     // When/Then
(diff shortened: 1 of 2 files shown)


packages/app/src/cli/services/app-logs/poll-app-logs.ts
```diff
@@ -68,9 +66,9 @@ export const pollAppLogs = async ({
           }).catch((error) => {
             outputDebug(`Unexpected error during polling: ${error}}\n`)
           })
-        }, POLLING_BACKOFF_INTERVAL_MS)
+        }, POLLING_THROTTLE_RETRY_INTERVAL_MS)
       } else {
-        throw new Error(`Error while fetching: ${responseText}`)
+        throw new Error(`Unhandled bad response: ${response.status}`)
```
DuncanUszkay1, line 71: I'm a little confused here- in the PR description it says that we re-enqueue no matter what, but it seems like we're not re-queuing 500s based on this if statement

packages/app/src/cli/services/app-logs/poll-app-logs.test.ts
```diff
@@ -196,17 +189,15 @@ describe('pollAppLogs', () => {
       apiKey: API_KEY,
       resubscribeCallback: MOCKED_RESUBSCRIBE_CALLBACK,
     })
-    await vi.advanceTimersToNextTimerAsync()
+    // await vi.advanceTimersToNextTimerAsync()
```
DuncanUszkay1, line 192: I assume this should be removed?

DuncanUszkay1 (approved):

mssalemi (approved):

packages/app/src/cli/services/app-logs/poll-app-logs.ts
```diff
@@ -145,8 +143,22 @@ export const pollAppLogs = async ({
     }, POLLING_INTERVAL_MS)
     // eslint-disable-next-line no-catch-all/no-catch-all
   } catch (error) {
-    stdout.write(`Error while retrieving app logs.`)
-    stdout.write('App log streaming is no longer available in this `dev` session.')
+    outputWarn(`Error while polling app logs.`)
+    outputWarn(`Retrying in ${POLLING_ERROR_RETRY_INTERVAL_MS / 1000} seconds.`)
     outputDebug(`${error}}\n`)
+
+    setTimeout(() => {
```
otheriault, line 150: I think we should extract a reusable method to invoke `pollAppLogs` since we use it here and on line `60`. Then we can also test is standalone.

packages/app/src/cli/services/app-logs/poll-app-logs.ts
otheriault, line 71: Maybe we should throw a specific error and catch that below to reset the timeout. Then we can catch any other random errors and handle those separately - maybe kill the polling for any _unexpected_ errors for example.

packages/app/src/cli/services/app-logs/poll-app-logs.ts
lopert: Chatted on slack about this. 
- the setTimeouts are already just: `setTimeout { pollAppLogs {...}, delay}` . It’s essentially a 1 liner already.
- If I create a new retryPolling function, I just end up having to pass all the args around.

packages/app/src/cli/services/app-logs/poll-app-logs.ts
lopert: This code branch here is specifically for bad responses from the server.
When this throws, it does end up being caught by our overarching try/catch.

We had the distinction before of only retrying on these "bad response from server" errors, and stopping the polling on anything unexpected.
But sometimes the CLI would error out on other transient things that weren't covered here, ie. `ECONNRESET`, so we decided to retry on unexpected things as well.

This throw, coupled with the overarching try/catch, means that our catch can handle the retries in both scenarios (bad response from server, and any other unknown failure)

otheriault: Got it. Thanks

PR #4041, opened 2024-06-12
[LogStreaming] App logs command (without refactor)

mssalemi:

Fixes #0000 

### Measuring impact

### Checklist


packages/app/src/cli/services/app-logs/ui/components/Logs.tsx
```diff
@@ -0,0 +1,207 @@
...
+interface DetailsFunctionRunLogEvent {
+  input: string
+  inputBytes: number
+  invocationId: string
+  output: string
+  outputBytes: number
+  logs: string
+  functionId: string
+  fuelConsumed: string
+  errorMessage: string | null
```
DuncanUszkay1, line 44: Heads up- The input, output, and errorMessage types are going to change with this PR: https://github.com/Shopify/shopify/pull/513906

packages/app/src/cli/services/app-logs/ui/components/Logs.tsx
```diff
@@ -0,0 +1,207 @@
...
+  apiKey: string
+  cursor: string
+  jwtToken: string
+  filters?: {
+    status?: string
+    source?: string
+  }
+}
+
+interface DetailsFunctionRunLogEvent {
```
DuncanUszkay1, line 35: I think this is one part that we should go out of our way to share between the commands

packages/app/src/cli/services/app-logs/ui/components/Logs.tsx
```diff
@@ -0,0 +1,207 @@
...
+        errors,
+        appLogs,
+      } = await logsProcess({jwtToken: jwtTokenState || jwtToken, cursor: currentCursor, filters})
+      if (errors) {
+        if (errors.some((error) => error.includes('429'))) {
+          currentIntervalRef.current = POLLING_THROTTLE_RETRY_INTERVAL_MS
+          setErrorsState([...errors, `Retrying in ${POLLING_THROTTLE_RETRY_INTERVAL_MS / 1000} seconds.`])
+        } else if (errors.some((error) => error.includes('401'))) {
+          setJwtTokenState(null)
+          setErrorsState([...errors, 'Resubscribing to logs.'])
```
DuncanUszkay1, line 98: I don't think this message is necessary, since the resubscription happens right away

packages/app/src/cli/services/app-logs/ui/components/Logs.tsx
```diff
@@ -0,0 +1,207 @@
...
+        } else {
+          currentIntervalRef.current = POLLING_ERROR_RETRY_INTERVAL_MS
+          setErrorsState([...errors, `Retrying in ${POLLING_ERROR_RETRY_INTERVAL_MS / 1000} seconds.`])
+        }
+      }
+
+      if (newCursor) {
+        setErrorsState([])
+        currentIntervalRef.current = POLLING_INTERVAL_MS
+      }
```
DuncanUszkay1, line 108: Rather than an if here, should we just put an early return if there was an error?

packages/app/src/cli/services/context.ts
```diff
@@ -50,6 +50,131 @@ export const InvalidApiKeyErrorMessage = (apiKey: string) => {
...
+}
+
+interface LogsContextOutput {
+  remoteApp: Omit<OrganizationApp, 'apiSecretKeys'> & {apiSecret?: string}
+  storeFqdn: string
+  storeId: string
+  localApp: AppInterface
+}
+
+export async function ensureLogsContext(options: LogsContextOptions): Promise<LogsContextOutput> {
```
DuncanUszkay1, line 68: Was this replicated from what we're doing in dev? 👀

This feels like something that should be behind some kind of shared method so we don't need to implement it all just for logs- but curious to see what CLI folks think.

packages/app/src/cli/commands/app/logs.ts
```diff
@@ -0,0 +1,91 @@
...
+      description: 'The API key of your app.',
+      env: 'SHOPIFY_FLAG_APP_API_KEY',
+      exclusive: ['config'],
+    }),
+    json: Flags.boolean({
+      hidden: false,
+      description: 'Print logs in JSON format.',
+      env: 'SHOPIFY_FLAG_JSON',
+      default: false,
+    }),
```
DuncanUszkay1, line 37: This flag appears to be unsupported- should we make a follow-up ticket for this and remove it from here for now?

DuncanUszkay1 (commented):
Mainly just some nits on my end, though I'd want CLI folks to take a look before I put the ✅ since there are some elements in here that look like they should probably be generic 🤔

packages/app/src/cli/commands/app/logs.ts
mssalemi: Ah didn't know that, that is good with me.

packages/app/src/cli/services/context.ts
mssalemi: Yea, mostly just copied over the same code. I can look into this more.

shauns: Yes, can you ideally just use the `ensureDevContext` function directly? Its job is to take the command flags and produce all the options the command itself will need to run (e.g. loading the app). We definitely don't want to duplicate here

packages/app/src/cli/services/app-logs/ui/components/Logs.tsx
```diff
@@ -0,0 +1,177 @@
...
+        setJwtTokenState(jwtToken.jwtToken)
+      }
+      const {
+        cursor: newCursor,
+        errors,
+        appLogs,
+      } = await logsProcess({jwtToken: jwtTokenState || jwtToken, cursor: currentCursor, filters})
+      if (errors) {
+        if (errors.some((error) => error.includes('429'))) {
+          currentIntervalRef.current = POLLING_THROTTLE_RETRY_INTERVAL_MS
```
shauns, line 64: I'm assuming some of this -- the polling throttle, formatting the event, should be shared between this and what `dev` outputs on a function run?

packages/app/src/cli/services/app-logs/ui/components/Logs.tsx
shauns: +1

packages/app/src/cli/services/app-logs/processes/polling-app-logs.ts
```diff
@@ -0,0 +1,161 @@
...
+    token: '',
+  }
+  const result = await developerPlatformClient.subscribeToAppLogs(appLogsSubscribeVariables)
+  const {jwtToken, success, errors} = result.appLogsSubscribe
+  outputDebug(`Token: ${jwtToken}\n`)
+  outputDebug(`API Key: ${appLogsSubscribeVariables.apiKey}\n`)
+  if (errors && errors.length > 0) {
+    outputWarn(`Errors subscribing to app logs: ${errors.join(', ')}`)
+    outputWarn('App log streaming is not available in this `log` session.')
+    return
```
shauns, line 67: For this sort of thing, throw an appropriate error instead. We'll then know whether to treat this as a bug or user error.  And there's a standard format for failures. The two to use are:
- `AbortError` -- user error, i.e. not Shopify's fault
- `BugError` -- a bug, something that's our fault

I don't know the situation that causes an error here and hence which is correct.

packages/app/src/cli/services/context.ts
shauns: To expand on this though. The `logs` command is similar to `dev` in that its showing this growing output of activity on a shop. But on the other hand it's extremely different: we expect developer to use `dev` all the time and that it would be the command that might do things like guide a developer through selecting a shop/app record (that's a lot of what happens here). Is that the case for `logs`? It seems like you need very little: an api key and the shop's ID. So maybe the right thing is something substantially smaller.

packages/app/src/cli/services/app-logs/processes/polling-app-logs.ts
```diff
@@ -0,0 +1,161 @@
...
+    },
+  })
+
+  if (!response.ok) {
+    const responseText = await response.text()
+    if (response.status === 401) {
+      return {
+        errors: [`${response.status}: ${response.statusText}`],
+      }
+    } else if (response.status === 429 || response.status >= 500) {
```
shauns, line 115: Similar to elsewhere, we should be trying to unify this across what `dev` ends up consuming too.

shauns (changes requested):

packages/app/src/cli/services/app-logs/ui/components/Logs.tsx
mssalemi: No for the output, the output for the app logs will be different between `dev` and `logs` based on the current [DX](https://docs.google.com/document/d/1R_qB9-n0ahIiEkObNO7_aioWvqByNlJsKHj0_4UrxMA/edit#heading=h.ol4bdsjg0ajm).  

For the polling throttle, we handle that in `dev` [here](https://github.com/Shopify/cli/blob/main/packages/app/src/cli/services/app-logs/poll-app-logs.ts#L53-L75), and we recently updated this to use the `outputWarns` and `outputDebug` so perhaps we can do something similar instead. But with this PR's approach, we show these messages conditally (rather than always outputted to the logs). 🤔

packages/app/src/cli/services/app-logs/ui/components/Logs.tsx
mssalemi: Yea, just relizing I dont even think we need this check. If there is no cursor, then that error would have been thrown in the polling process, so would exit the terminal (since its a 400 error we dont handle).

packages/cli-kit/src/public/node/ink.ts
```diff
@@ -1 +1 @@
-export {Box, Text, useInput, useStdin} from 'ink'
+export {Static, Box, Text, useInput, useStdin} from 'ink'
```
mssalemi, line 1: @shauns Is it okay to export Static from here? I wasn't if there was a better way to use this component. I tried to setup something similar to what the [ConcurrentOutput ](https://github.com/Shopify/cli/blob/main/packages/cli-kit/src/private/node/ui/components/ConcurrentOutput.tsx#L192-L215)component uses.

packages/app/src/cli/services/app-logs/logs-command/ui/components/hooks/usePollAppLogs.ts
```diff
@@ -0,0 +1,80 @@
...
+          const fuel = (appLog.fuelConsumed / ONE_MILLION).toFixed(4)
+          const prefix = {
+            status: log.status === 'success' ? 'Success' : 'Failure',
+            source: log.source,
+            fuelConsumed: fuel,
+            functionId: appLog.functionId,
+            logTimestamp: log.log_timestamp,
+          }
+
+          setAppLogOutputs((prev) => [...prev, {appLog, prefix}])
```
DuncanUszkay1, line 67: Curious if we need to keep track of every event we've ever seen 🤔 Not urgent but something to explore as a fast-follow

packages/app/src/cli/services/app-logs/logs-command/ui/components/hooks/usePollAppLogs.test.tsx
```diff
@@ -0,0 +1,215 @@
...
+    expect(setTimeout).toHaveBeenCalledWith(expect.any(Function), POLLING_ERROR_RETRY_INTERVAL_MS)
+
+    await vi.advanceTimersToNextTimerAsync()
+    expect(hook.lastResult?.appLogOutputs).toHaveLength(1)
+    expect(hook.lastResult?.errors).toHaveLength(0)
+    expect(setTimeout).toHaveBeenCalledWith(expect.any(Function), POLLING_INTERVAL_MS)
+
+    expect(vi.getTimerCount()).toEqual(1)
+    timeoutSpy.mockRestore()
+  })
```
mssalemi, line 195: We added a `timeoutSpy.mockRestore()` to each test for react related testing.

DuncanUszkay1 (approved):

mssalemi:
Shaun requested changes which have been applied. But he is not available for re-review due to timezones. In the interest in getting this in for our dev preview at editions, we are going to move forward with the PR. Since this is behind a beta flag, we are still good to change if there is additional concerns.

lopert (approved):

packages/app/src/cli/services/app-logs/logs-command/poll-app-logs.ts
```diff
@@ -0,0 +1,43 @@
...
+      }
+    } else {
+      throw new AbortError(`${errorResponse.errors.join(', ')} while fetching app logs`)
+    }
+  }
+
+  const data = responseJson as {
+    app_logs?: AppLogData[]
+    cursor?: string
+    errors?: string[]
```
andrewhassan: In this success case, are `app_logs` and `cursor` always present? Also do we need to type `errors`?

packages/app/src/cli/commands/app/logs.ts
```diff
@@ -0,0 +1,61 @@
...
+  shopify app logs
+  \`\`\`
+  `
+
+  static description = this.descriptionWithoutMarkdown()
+
+  static flags = {
+    ...Dev.flags,
+    source: Flags.string({
+      description: 'Filters output to the specified log source (Multiple flags allowed).',
```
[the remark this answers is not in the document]
mssalemi: Ah good spot, no we do not right now. We also only allow single shop id right now.

packages/app/src/cli/services/app-logs/logs-command/ui/components/hooks/usePollAppLogs.ts
DuncanUszkay1: That's true in the browser, but is it true with ink? Does dev keep track of every line it's ever printed?

DuncanUszkay1: Might just be a gap in my understanding of ink

shauns (approved):

packages/app/src/cli/commands/app/logs.ts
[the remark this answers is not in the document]
mssalemi: Good spot, we added options for the `status`, and left the `source` unchanged (given it can be any extension handle).

PR #4049, opened 2024-06-14
Use CI environment to block CLI prompts

shauns:

Fixes an issue where CI runs that prompt on Buildkite cause the build to hang.

Buildkite's environment behaves differently to some other systems and this has revealed an issue around how we check for whether a terminal session is interactive or not. This PR adds a simpler path to that check: if the CLI is running in CI, don't prompt.

### How to test your changes?

Run `CI=1 p shopify app dev --path="path to app" --reset` for a quick example.

### Measuring impact

### Checklist

### DIFF
--- a/packages/cli-kit/src/public/node/system.ts
+++ b/packages/cli-kit/src/public/node/system.ts
@@ -2,6 +2,7 @@ import {AbortSignal} from './abort.js'
 import {ExternalError} from './error.js'
 import {cwd} from './path.js'
 import {treeKill} from './tree-kill.js'
+import {isTruthy} from './context/utilities.js'
 import {shouldDisplayColors, outputDebug} from '../../public/node/output.js'
 import {execa, ExecaChildProcess} from 'execa'
 import {ReadStream} from 'tty'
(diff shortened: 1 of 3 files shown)


packages/cli-kit/src/public/node/system.ts
```diff
@@ -133,9 +134,13 @@ export async function sleep(seconds: number): Promise<void> {
  * will be used.
  *
  * @param stdin - The standard input stream to check.
+ * @param env - Environmemnt variables.
  * @returns True in the selected input stream support raw mode.
  */
-export function terminalSupportsRawMode(stdin?: ReadStream): boolean {
+export function terminalSupportsRawMode(stdin?: ReadStream, env = process.env): boolean {
+  if (isTruthy(env.CI)) {
```
isaacroldan, line 141: Is this `CI` env-var set for all CI platforms?

shauns: Yes, its as close to a standard as something like this gets

isaacroldan (approved):

PR #4065, opened 2024-06-17
refactor polling helpers

lopert:

Fixes #0000 

### Measuring impact

### Checklist


packages/app/src/cli/services/app-logs/helpers.ts
```diff
@@ -0,0 +1,67 @@
...
+}
+
+export const fetchAppLogs = async (url: string, jwtToken: string): Promise<Response> => {
+  return fetch(url, {
+    method: 'GET',
+    headers: {
+      Authorization: `Bearer ${jwtToken}`,
+    },
+  })
+}
```
DuncanUszkay1, line 47: Thoughts on combining these two helpers into one? I can't think of a reason why we'd need one without the other

packages/app/src/cli/services/app-logs/poll-app-logs-for-dev.ts
```diff
@@ -1,22 +1,17 @@
 import {writeAppLogsToFile} from './write-app-logs.js'
-import {useConcurrentOutputContext} from '@shopify/cli-kit/node/ui/components'
-import {partnersFqdn} from '@shopify/cli-kit/node/context/fqdn'
-import {fetch} from '@shopify/cli-kit/node/http'
+import {
```
DuncanUszkay1, line 2: Thoughts about using a folder rather than a `for-dev` suffix in the filename?

packages/app/src/cli/services/app-logs/helpers.ts
mssalemi: Agreeded. I think we only want to separate these if we do not want dev to be able to filter the results, but partners can hack this pretty easily if they wanted.

mssalemi (approved):
Changes make sense to me, I will work on refactoring my app logs branch to use the subscribe and fetch app logs :D

packages/app/src/cli/services/app-logs/helpers.ts
lopert: Sounds good. I kept them separate within the helper to keep the two things clear, but had one call the other, so the result is that anything calling this only has to make the 1 call to `fetchAppLogs` with the right params.

PR #4073, opened 2024-06-18
Upgrade vitest

shauns:

Fixes https://github.com/Shopify/develop-app-inner-loop/issues/1826

Upgrades vitest & vite across the board; fixes some test incompatibilities

### How to test your changes?

- Run `p test` and `p vitest`
- Restart VS code and see that the vitest explorer is now working, and that tests can also be run directly from the editor

### Measuring impact

### Checklist


package.json
```diff
@@ -95,7 +95,7 @@
   "version": "0.0.0",
   "resolutions": {
     "@types/react": "17.0.2",
-    "vite": "4.4.9",
+    "vite": "5.3.1",
     "@oclif/core": "3.26.5",
     "whatwg-url": "14.0.0",
```
gonzaloriestra, line 100: Could you check if we can remove this? A punycode warning was shown with Node 22, and I think vite was one of the culprits.

gonzaloriestra (approved):
Nice work!

package.json
shauns: I don't think it can be removed. The extensions console uses vite to build for instance

gonzaloriestra: I meant `whatwg-url`

PR #4087, opened 2024-06-18
Function log replay correlation

nickwesselman:

Fixes https://github.com/Shopify/shopify-functions/issues/245

Makes it easier to go from seeing a function execution in `app dev`, to using it in `replay`.

Adds the log "identifier" to the `app dev` output, to further differentiate each execution: 

image

... and provide a string that can be used when searching for logs in `replay`:

image

... or to feed into the new `--log` argument of `replay` to bypass the manual selection:

image

### How to test your changes?

Use `app dev`, then copy the identifier into a `replay` execution using the `--log` argument.

### Checklist

### DIFF
--- a/packages/app/src/cli/commands/app/function/replay.ts
+++ b/packages/app/src/cli/commands/app/function/replay.ts
@@ -36,7 +36,7 @@ export default class FunctionReplay extends Command {
     log: Flags.string({
       char: 'l',
       description:
-        'Specifies a specific log identifier to replay instead of selecting from a list. The identifier is provided in the output of `shopify app dev` and is the suffix of the log file name.',
+        'Specifies a log identifier to replay instead of selecting from a list. The identifier is provided in the output of `shopify app dev` and is the suffix of the log file name.',
       env: 'SHOPIFY_FLAG_LOG',
     }),
     json: Flags.boolean({
--- a/packages/app/src/cli/services/function/replay.test.ts
+++ b/packages/app/src/cli/services/function/replay.test.ts
@@ -96,7 +96,7 @@ describe('replay', () => {
   test('does not allow selection of runs for other functions', async () => {
     // Given
     const file1 = createFunctionRunFile({handle: extension.handle})
-    const file2 = createFunctionRunFile('another-function-handle')
+    const file2 = createFunctionRunFile({handle: 'another-function-handle'})
     mockFileOperations([file1, file2])
 
     vi.mocked(selectFunctionRunPrompt).mockResolvedValue(file1.run)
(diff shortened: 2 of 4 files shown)


packages/app/src/cli/services/function/replay.test.ts
```diff
@@ -253,8 +298,8 @@ describe('replay', () => {
   })
 })
 
-function createFunctionRunFile(handle: string) {
-  const identifier = randomUUID().substring(0, 6)
+function createFunctionRunFile(handle: string, identifierOpt?: string) {
```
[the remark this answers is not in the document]
DuncanUszkay1: I ended up keeping them separate, since the identifier isn't part of the payload, it appears both at the top level and in the filename

packages/app/src/cli/commands/app/function/replay.ts
```diff
@@ -33,6 +33,12 @@ export default class FunctionReplay extends Command {
       env: 'SHOPIFY_FLAG_CLIENT_ID',
       exclusive: ['config'],
     }),
+    log: Flags.string({
+      char: 'l',
+      description:
+        'Specifies a specific log identifier to replay instead of selecting from a list. The identifier is provided in the output of `shopify app dev` and is the suffix of the log file name.',
```
andrewhassan, line 39: Super nit: I don't think we need `specific` with `specifies`
```suggestion
        'Specifies a log identifier to replay instead of selecting from a list. The identifier is provided in the output of `shopify app dev` and is the suffix of the log file name.',
```

packages/app/src/cli/services/function/replay.ts
```diff
@@ -115,12 +113,75 @@ async function runFunctionRunnerWithLogInput(
...
+  functionHandle?: string
+  identifier?: string
+}
+
+function parseLogFilename(filename: string): LogFileMetadata {
+  // Expected format: 20240522_150641_827Z_extensions_my-function_abcdef.json
+  const splitFilename = filename.split(/[_\.]/)
+
+  if (splitFilename.length < 6) {
+    return {}
```
andrewhassan, line 142: Would it make more sense to return `null` or `undefined` here? I don't think having all the fields be optional makes much sense because then it's possible to end up with an invalid state (e.g. `functionHandle` is present but `namespace` isn't).

andrewhassan (approved):

packages/app/src/cli/services/function/replay.ts
DuncanUszkay1: Yep, good point, that's nicer. Updated :+1:

DuncanUszkay1:
Going to wait until the log command ships, to avoid introducing any conflicts.

PR #4136, opened 2024-06-28
Update get-started.md

ken-britton-shopify:
Add `dev up` step to instructions.


shauns:
I've missed this post-merge, but its worth noting that `dev up` is only needed for Shopifolk, and that `dev` doesn't exist externally. Obviously most contribution is in-house and I think its fine to optimise for that, but wouldn't want to leave someone running the CLI locally externally -- e.g. to debug something -- confused.

ken-britton-shopify:
Agree, I’ll back this out when I’m back in the office.

isaacroldan:
I don't know why I thought I was reviewing the internal get-started doc 🤦 
thanks @shauns

PR #4142, opened 2024-07-01
Avoid loop when users pass the `--theme-editor-sync` flag in the `shopify theme dev` command

karreiro:

This PR updates the checksum calculation to avoid loops in when users run the `shopify theme dev --theme-editor-sync`, as before the values were never matching.

In the past, we introduced a visitor that was adding the `settings` key locally as the platform auto adds them ([context](https://github.com/Shopify/shopify-cli/commit/6d79724ca5a6516bf47e654feed98c61b668582f)):

That was a great, as we were avoiding unnecessary uploads when templates were only missing the `settings` key. However, now that we support themes blocks, that also have settings and may be deeply nested, that convenience is not longer ideal and keeping up locally may generate erratic behaviour.

Thus, this PR removes the visitor, and fixes the loop scenario that some users were facing with the `--theme-editor-sync` flag.

Additionally, the checksum logic no longer normalizes the `config/settings_schema.json` file as the platform now keeps user defined spacing on files that are not platform controlled.

### How to test your changes?

- Download [this zip file](https://github.com/user-attachments/files/16052340/Code.Base.Theme-4.0.0.zip)
- Run `shopify theme dev --theme-editor-sync`
- Notice that `config/settings_schema.json` and `templates/index.json` get into a `get`/`update` loop
- Update you local CLI to use this PR
- Notice the loop no longer happens

### Post-release steps

N/A

### Measuring impact

### Checklist

### DIFF
--- a/.changeset/warm-chicken-sin.md
+++ b/.changeset/warm-chicken-sin.md
@@ -1,0 +1,5 @@
+---
+'@shopify/theme': patch
+---
+
+Avoid loop when users pass the `--theme-editor-sync` flag in the `shopify theme dev` command
--- a/packages/cli-kit/assets/cli-ruby/lib/shopify_cli/theme/file.rb
+++ b/packages/cli-kit/assets/cli-ruby/lib/shopify_cli/theme/file.rb
@@ -76,7 +76,8 @@ module ShopifyCLI
 
       def checksum
         content = read
-        if mime_type.json?
+
+        if mime_type.json? && !settings_schema?
           # Normalize JSON to match backend
           begin
             content = normalize_json(content)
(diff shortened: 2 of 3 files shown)


jamesmengo:
Was able to 🎩 the  `settings_schema` case, but haven't been able to replicate the `templates/index.json` case mentioned

I might be missing something on my end - let's sync on this before shipping, but otherwise LGTM

karreiro:
Thank you, @jamesmengo! I've re-validated the the steps to reproduce the issue. Let's sync today because I believe I know why you couldn't face the same error (it's platform-related) :)

PR #4149, opened 2024-07-03
Notification system

gonzaloriestra:

Fixes: https://github.com/Shopify/develop-app-inner-loop/issues/1962

We want to notify CLI users about deprecations, new features or events

Adds three hidden commands:
- `notifications generate`: shows several prompts asking for the notification content, type and filters, and adds it to a local notifications.json file.
- `notifications list`: shows the current notifications from the local notifications.json
- `cache clear`: clears the CLI cache (Mac stores it in `~/Library/Preferences/shopify-cli-kit-nodejs/config.json`), including notifications stuff and other API responses that are temporarily memorized

Then, before each command we check the remote notifications.json from this repository and show the required ones. The file is cached for 1h.

Demo: https://hackdays.shopify.io/projects/18981

### How to test your changes?

Install the snapshot from this branch: `npm i -g @shopify/cli@0.0.0-snapshot-20241108163119`

I've uploaded a sample notifications.json to simplify testing:
- `SHOPIFY_CLI_NOTIFICATIONS_URL=https://shopify.link/MNbn shopify version`
- `SHOPIFY_CLI_NOTIFICATIONS_URL=https://shopify.link/MNbn shopify theme list`

When `SHOPIFY_CLI_NOTIFICATIONS_URL` is not passed, it looks for the file in the main branch of this repository (it doesn't exist yet).

To generate and list notifications:
- `shopify notifications generate`
- `shopify notifications list`

To clear the cache, so that it downloads again the configuration file and forgets if a notification was shown:
- `shopify cache clear`

### Post-release steps

Merge the internal doc: https://github.com/Shopify/vault-pages/pull/7111

### Measuring impact

### Checklist


packages/cli-kit/src/public/node/notifications-system.ts
```diff
@@ -0,0 +1,235 @@
+import {versionSatisfies} from './node-package-manager.js'
+import {renderError, renderInfo, renderWarning} from './ui.js'
+import {getCurrentCommandId} from './global-context.js'
+import {fileExists, readFile} from './fs.js'
+import {outputDebug} from './output.js'
+import {CLI_KIT_VERSION} from '../common/version.js'
+import {NotificationsKey, cacheRetrieveOrRepopulate, getCache, setCache} from '../../private/node/conf-store.js'
+
+const URL = 'https://raw.githubusercontent.com/Shopify/cli/notifications/notifications.json'
```
gonzaloriestra, line 9: We have to change this URL before merging to:

```suggestion
const URL = 'https://raw.githubusercontent.com/Shopify/cli/main/notifications.json'
```

packages/cli-kit/src/public/node/global-context.ts
```diff
@@ -0,0 +1,36 @@
+export interface GlobalContext {
+  currentCommandId: string
+  [key: string]: string
```
shauns, line 3: What's this for? It basically makes GlobalContext close to untyped -- I haven't seen a use of it. Would drop it if not used.

packages/cli-kit/src/public/node/notifications-system.ts
```diff
@@ -0,0 +1,235 @@
...
+export async function getNotifications(): Promise<Notifications> {
+  const cacheKey: NotificationsKey = `notifications-${URL}`
+  const rawNotifications = await cacheRetrieveOrRepopulate(cacheKey, fetchNotifications, 24 * 3600 * 1000)
+  return JSON.parse(rawNotifications)
+}
+
+/**
+ * Fetch notifications from GitHub.
+ */
+async function fetchNotifications(): Promise<string> {
```
shauns, line 94: I think we should look into wrapping this in a timeout. Like, if it takes longer than Xms to load, we stop waiting and move on. Given that the command is paused whilst this is happening.

packages/cli-kit/src/public/node/notifications-system.ts
```diff
@@ -0,0 +1,235 @@
...
+import {CLI_KIT_VERSION} from '../common/version.js'
+import {NotificationsKey, cacheRetrieveOrRepopulate, getCache, setCache} from '../../private/node/conf-store.js'
+
+const URL = 'https://raw.githubusercontent.com/Shopify/cli/notifications/notifications.json'
+
+export interface Notifications {
+  notifications: Notification[]
+}
+
+export interface Notification {
```
shauns, line 15: The typing here might not match the format used in the notification file -- e.g. if that file evolves but the CLI reading from it is an old one. I think it'd be helpful to defend against that -- whilst we're the authors of the notification file, we're still decoupled from it. I think I would swap this for a Zod schema, and parse the file using that. Then, if parsing fails because it doesn't match what this CLI's schema is expecting we can skip the notifications and move on. Obviously we would also try not to break the file as well.

packages/cli-kit/src/public/node/notifications-system.ts
```diff
@@ -0,0 +1,235 @@
...
+  }
+}
+
+/**
+ * Renders the notifications to the user.
+ *
+ * @param notifications - The notifications to render.
+ */
+async function renderNotifications(notifications: Notification[]) {
+  notifications.forEach((notification) => {
```
shauns, line 57: I think this is going to be quite noisy if there are multiple notifications matching on any given command run -- each section having it's own box. I'm not sure what the best alternative is. Or, how likely it is to happen. What do you think?

packages/cli-kit/src/public/node/notifications-system.ts
```diff
@@ -0,0 +1,235 @@
...
+  const cacheKey: NotificationsKey = `notifications-${URL}`
+  const rawNotifications = await cacheRetrieveOrRepopulate(cacheKey, fetchNotifications, 24 * 3600 * 1000)
+  return JSON.parse(rawNotifications)
+}
+
+/**
+ * Fetch notifications from GitHub.
+ */
+async function fetchNotifications(): Promise<string> {
+  const response = await fetch(URL)
```
shauns, line 95: There's an analytics helper to separate out network time from time with the command being active. We should wrap this up in that.

packages/cli-kit/src/public/node/notifications-system.ts
isaacroldan: That's a good idea 👌

gonzaloriestra: We are catching errors [here](https://github.com/Shopify/cli/pull/4149/files#diff-0860eecbbe74005c184dce2613e954204bd54aad1b9f68806ea996dad684715cR46), so if the parsing fails, it already moves on. Any other advantage of using Zod here?

packages/cli-kit/src/public/node/notifications-system.ts
gonzaloriestra: Limited to 2 messages finally

nickwesselman:
Could we make this easier to test by making the notification system URL overridable by an environment variable? And maybe a command to clear the notification caches?

nickwesselman:
Looks like there is a formatting issue when the message contains `\n`, we should either recognize or strip them.

```
╭─ warning ────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│                                                                                                                                          │
│  Deprecated Shopify CLI version                                                                                                          │
│                                                                                                                                          │
   December 13, 2024. Upgrade to Shopify CLI 3.67 or later.   │I version on
│                                                                                                                                          │
╰──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
```

isaacroldan (approved):

PR #4154, opened 2024-07-03
Update setup pnpm action to latest

isaacroldan:
We have a policy of pinning github actions to specific commits, our `pnpm/action-setup` was very outdated (v2.2.4) and started failing today.

Update `pnpm/action-setup` to version 4

### Measuring impact

### Checklist

### DIFF
--- a/.github/actions/setup-cli-deps/action.yml
+++ b/.github/actions/setup-cli-deps/action.yml
@@ -13,7 +13,7 @@ runs:
         git config --global user.name "App Management"
       shell: bash
     - name: Install pnpm
-      uses: pnpm/action-setup@23657c8550e9be4f7ee7e9d7fe8adda6d703bb4e # pin@v2.2.4
+      uses: pnpm/action-setup@fe02b34f77f8bc703788d5817da081398fad5dd2 # pin@v4.0.0
       with:
         version: ${{ env.PNPM_VERSION }}
         run_install: false


.github/actions/setup-cli-deps/action.yml
```diff
@@ -13,7 +13,7 @@ runs:
         git config --global user.name "App Management"
       shell: bash
     - name: Install pnpm
-      uses: pnpm/action-setup@c3b53f6a16e57305370b4ae5a540c2077a1d50dd # pin@v2.2.4
+      uses: pnpm/action-setup@fe02b34f77f8bc703788d5817da081398fad5dd2 # pin@v4.0.0
```
craigmichaelmartin, line 16: > We have a policy of pinning github actions to specific commits

Why do we pin to commits instead of versions?

If we were OK with switching to a version reference, `pnpm/action-setup` "supports npm versioning scheme, it could be an exact version (such as `6.24.1`), or a version range (such as `6`, `6.x.x`, `6.24.x`, `^6.24.1`, `*`, etc.), or latest" so we could do `uses: pnpm/action-setup@v4.0.0` to replicate the current change.

However, I wonder why we are so defensive in general. Could we `uses: pnpm/action-setup@v4.0.x` or even better just `uses: pnpm/action-setup@v4`?

craigmichaelmartin (approved):
Thanks, Isaac!

.github/actions/setup-cli-deps/action.yml
amcaplan: https://vault.shopify.io/page/GitHub-Actions~13124.md#overview

craigmichaelmartin: Thanks, Ariel! And I mis-characterized the `version` input for the `pnpm/action-setup` action which is for the version of `pnpm` as the `pnpm/action-setup` action's version; bad job out me!

craigmichaelmartin, line 16: Verified this commit hash corresponds with the [authentic repo and release](https://github.com/pnpm/action-setup/releases/tag/v4.0.0).

PR #4157, opened 2024-07-03
Clean up

gracejychang:
Resolves a part of https://github.com/Shopify/develop-app-inner-loop/issues/1846

### Measuring impact

### Checklist


packages/app/src/cli/models/extensions/extension-instance.ts
```diff
@@ -94,7 +94,7 @@ export class ExtensionInstance<TConfiguration extends BaseConfigType = BaseConfi
   }
 
   get isAppConfigExtension() {
-    return this.specification.experience === 'configuration'
+    return this.specification.uidStrategy === 'single' || this.specification.uidStrategy === 'dynamic'
```
karenxie, line 97: Is `dynamic` here because of webhooks? Will it always be the case that we want to include dynamic specs here? 

Could we also do something like `['single', 'dynamic'].includes(this.specification.uidStrategy)`?

gracejychang: yeah `dynamic` is there because of webhooksl I'm not sure if this will always be the case that we include dynamic specs but I assumed so 🤔  I thought `single` implies that it's an app module configured on the TOML, and `dynamic` is similar, but we allow for multiple modules instead of just one

karenxie: > I thought single implies that it's an app module configured on the TOML, and dynamic is similar, but we allow for multiple modules instead of just one

I don’t know if that is true. from the [github discussion](https://github.com/Shopify/extensions/discussions/3293), the dynamic strategy is that the client doesn’t need to generate the uid whereas the `uuid` strategy is where the client does generate the uid (basically what extensions do now)

karenxie (approved):

PR #4161, opened 2024-07-04
Dedup cli-kit when bundling hydrogen

isaacroldan:
Since `cli-hydrogen` is an external dependency that at the same time depends on `cli-kit` which exists in this repo, it creates a weird dependency tree that eventually causes esbuild to bundle 2 identical copies of `cli-kit`

Fixes #0000 

- Add a new ESBuild plugin to dedup `cli-kit` when bundling.

### How to test your changes?
See the snapshot in the comments, if you install that and try any app and hydrogen command, everything should work.

### Measuring impact

### Checklist

### DIFF
--- a/bin/bundling/esbuild-plugin-dedup-cli-kit.js
+++ b/bin/bundling/esbuild-plugin-dedup-cli-kit.js
@@ -1,0 +1,13 @@
+const CliKitDedupPlugin = ({require}) => {
+  return {
+    name: 'CliKitDedupPlugin',
+    setup(build) {
+      // Resolve all imports of @shopify/cli-kit to the local dependency of the package.
+      build.onResolve({filter: /@shopify\/cli-kit/}, (args) => {
+        return {path: require.resolve(args.path)}
+      })
+    }
+  }
+}
+
+export default CliKitDedupPlugin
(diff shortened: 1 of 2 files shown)


MitchDickinson:
Is this the first time we've encountered this type of scenario? I would assume we have other scenarios where dependencies are shared in this way, and curious why this is the first time we have to write some custom dedup code for it?

isaacroldan:
`pnpm` does a good job on hoisting all transitive dependencies to the root node_modules when the versions are compatible. So in most (almost all) cases we only have 1 copy of each dependency. 

If two of our dependencies use different (incompatible) versions of another sub-dependency, then `esbuild` will include both and that's correct. Example, a package depends on `zod@2.0` and another one depends on `zod@3.0`, esbuild will bundle both. Not sure if this is actually happening for any specific dependency, but it could.

The case with `cli-hydrogen` and `cli-kit` is a bit special, since cli-kit is part of this monorepo, when we use it it is not installed as a normal dependency, it is read directly from source (with a symlink).
But when we install `cli-hydrogen` as a external dependency, it tries to use an external `cli-kit` too so we need to force it to use the local one (we do this with custom resolutions in the main package.json).
But, even doing this, it still thinks they are different cli-kits, so this is why we need this plugin.

This also happens with `@shopify/plugin-cloudflare` for the same reason: is part of this monorepo but it is also used by `cli-hydrogen`. I didn't include a plugin for that case because I was having some issues to make it work and is not as important: that package only has 2 small files, is not a big deal to have them duplicated for now.

Hope is more clear now 🙏

MitchDickinson:
Very clear, thank you so much for that explanation!

PR #4173, opened 2024-07-09
Fix `nginx.conf` conflict with `partners` in spin

RobertMic:

It looks like spin provides a default `nginx.conf` file when a repository does not provide its own. This default creates a server on `cli.<FQDN>`, but the `partners` `nginx.conf.erb` does the same. The `app dev` command relies on the `partners` version of `cli.<FQDN>` to be up and running, and the conflicting `nginx.conf` causes the spin provided default to take precedence over the `partners` one, causing the dev preview of the CLI to not work in spin.

This PR adds an empty `nginx.conf.erb` file to `.spin` to prevent spin from creating the default config which leads to the conflicting server entries.

### How to test your changes?

Make a new spin instance that with a 1p app and try to run `app dev`. Make sure the dev preview is accessible. You can use the `flow` spin constellation to try this out:

```
spin up flow:with-partners-and-extensions

# Then run this inside of the `flow` directory
yarn cli:dev
```

### Measuring impact

### Checklist


isaacroldan:
Has this started to fail recently? Because it used to work 🤔

RobertMic:
> Has this started to fail recently? Because it used to work 🤔

@isaacroldan I'm not sure when it started failing, but a few days ago we tried doing 1p app development in spin with the CLI and found we couldn't hit the dev preview server. After debugging we noticed that spin creates a server for `cli.fqdn` in the`/etc/openresty/apps/shopify--cli` config. We also noticed the same server rule in the `shopify--partners` config. We tried to fix it by creating a blank config in `cli/.spin/nginx.conf.erb` (this PR) and after running `dev up` it was all working as expected and the config in `/etc/openresty/apps` was blank as expected

isaacroldan (approved):

PR #4183, opened 2024-07-10
[Themes] - Shopify Theme Console

jamesmengo:
- https://github.com/orgs/Shopify/projects/4613/views/62?pane=issue&itemId=69671709

Base branch for the `shopify theme console` TS implementation

### How to test your changes?
- There are more detailed instructions in each of the feature branches that were merged in 

1) Execute the new feature `pnpm shopify theme console --dev-preview`
2) Open an instance of the ruby implementation `shopify theme console`

Evaluate some expressions and look for parity differences

Intentional Differences:
- We return `null` rather than `nil` in TS land  

### Measuring impact

### Checklist

### DIFF
--- a/.changeset/cuddly-yaks-laugh.md
+++ b/.changeset/cuddly-yaks-laugh.md
@@ -1,0 +1,5 @@
+---
+'@shopify/theme': minor
+---
+
+Release the developer preview for the Theme Console command
--- a/docs-shopify.dev/commands/interfaces/theme-console.interface.ts
+++ b/docs-shopify.dev/commands/interfaces/theme-console.interface.ts
@@ -31,6 +31,12 @@ export interface themeconsole {
   '-s, --store <value>'?: string
 
   /**
+   * The password for storefronts with password protection.
+   * @environment SHOPIFY_FLAG_STORE_PASSWORD
+   */
+  '--store-password <value>'?: string
+
+  /**
    * The url to be used as context
    * @environment SHOPIFY_FLAG_URL
    */
--- a/docs-shopify.dev/generated/generated_docs_data.json
+++ b/docs-shopify.dev/generated/generated_docs_data.json
@@ -4258,6 +4258,15 @@
               {
                 "filePath": "docs-shopify.dev/commands/interfaces/theme-console.interface.ts",
                 "syntaxKind": "PropertySignature",
+                "name": "--store-password <value>",
+                "value": "string",
+                "description": "The password for storefronts with password protection.",
+                "isOptional": true,
+                "environmentValue": "SHOPIFY_FLAG_STORE_PASSWORD"
+              },
+              {
+                "filePath": "docs-shopify.dev/commands/interfaces/theme-console.interface.ts",
+                "syntaxKind": "PropertySignature",
                 "name": "--url <value>",
                 "value": "string",
                 "description": "The url to be used as context",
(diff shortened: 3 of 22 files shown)


packages/theme/src/cli/utilities/repl/evaluater.ts
```diff
@@ -0,0 +1,199 @@
...
+  if (isResourceNotFound(serverTiming || '')) {
+    notFoundError()
+  }
+
+  return response
+}
+
+function parseDisplayResult(result: string): string | number | undefined {
+  const match = stripHTMLContent(result)
+  if (!match || !match[1]) return
```
karreiro, line 145: (optional) We may avoid this regex if we break the line during the rendering like we're doing [here](https://github.com/Shopify/cli/blob/c0762fca1d6612469b3b2083992c975caf48e37f/packages/cli-kit/assets/cli-ruby/lib/shopify_cli/theme/repl/api.rb#L54)

Personally, I believe that adding the extra `\n` is a bit nicer/deterministic because we'd no longer depend on what's being renderer inside the section, but I cannot think on any scenario this logic breaks because the `<` character is escaped, so feel free to keep it as is if you prefer 🙏

packages/theme/src/cli/utilities/repl/evaluater.ts
```diff
@@ -0,0 +1,199 @@
...
+  }
+
+  if (isTooManyRequests(response.status)) {
+    tooManyRequestsError()
+  }
+
+  const serverTiming = response.headers.get('server-timing')
+  if (isResourceNotFound(serverTiming || '')) {
+    notFoundError()
+  }
```
karreiro, line 138: I believe we could make each method hold the logic to identify the type of response, something like this:

```suggestion
  if (isExpiredSession(response)) {
    expiredSessionError()
  }

  if (isTooManyRequests(response)) {
    tooManyRequestsError()
  }

  if (isResourceNotFound(response)) {
    notFoundError()
  }
```

To make that happen, we need to update the other methods to:
```ts
// import {Response} from '@shopify/cli-kit/node/http'
function isExpiredSession(response: Response): boolean {
  return response.status === 401 || response.status === 403
}

function isTooManyRequests(response: Response): boolean {
  return response.status === 430 || response.status === 429
}

function isResourceNotFound(response: Response): boolean {
  // We don't look for the status code here because the Section Rendering API returns 200 even on unknown paths.
  return (response.headers.get('server-timing') ?? '').includes('pageType;desc="404"')
}
```

And fix the `Response` type at `packages/cli-kit/src/public/node/http.ts` to use `Awaited`:

image

karreiro (commented):
Thank you for this PR, @jamesmengo! Great stuff. I've left some comments, but the PR looks great, and I could not find any bugs 🚀

Besides the comments, I believe we could:
- rename `evaluater.ts` to `evaluator.ts`
- move `repl-theme-manager.ts` and `password-prompt.ts` to the `packages/theme/src/cli/utilities/repl/` directory

What do you think?

Thanks again for this PR! :)

packages/theme/src/cli/utilities/repl.ts
```diff
@@ -0,0 +1,61 @@
...
+
+function shutdownReplSession(error: unknown) {
+  if (error instanceof Error) {
+    outputInfo(outputContent`${outputToken.errorText(`Shopify Liquid console error: ${error.message}`)}`)
+    outputDebug(error.stack || 'Error backtrace not found')
+  }
+}
+
+function hasDelimiter(input: string): boolean {
+  return /\{\{|\}\}|\{%|%\}/.test(input)
```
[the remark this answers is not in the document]
jamesmengo: great idea!

PR #4186, opened 2024-07-11
Clear cache on app init

gonzaloriestra:

If you create a new project in a previously used folder, it doesn't run link because the folder may be cached, so the TOML doesn't get updated.

Clear cache on app init

### How to test your changes?

- `p shopify app init --name test-project --template none`
- `p shopify app dev --path test-project`
- `cat ~/Library/Preferences/shopify-cli-app-nodejs/config.json | grep test-project -A 3` => The cache should be there
- `rm -r test-project`
- `p shopify app init --name test-project --template none`
- `cat ~/Library/Preferences/shopify-cli-app-nodejs/config.json | grep test-project -A 3` => No cache
- `p shopify app dev --path test-project`
- `cat test-project/shopify.app.toml` => The TOML should be populated

### Measuring impact

### Checklist

### DIFF
--- a/.changeset/chilly-taxis-divide.md
+++ b/.changeset/chilly-taxis-divide.md
@@ -1,0 +1,5 @@
+---
+'@shopify/create-app': patch
+---
+
+Clear cache on app init
--- a/packages/create-app/src/services/init.ts
+++ b/packages/create-app/src/services/init.ts
@@ -23,10 +23,11 @@ import {
   readFile,
   writeFile,
 } from '@shopify/cli-kit/node/fs'
-import {joinPath} from '@shopify/cli-kit/node/path'
+import {joinPath, normalizePath} from '@shopify/cli-kit/node/path'
 import {username} from '@shopify/cli-kit/node/os'
 import {AbortError} from '@shopify/cli-kit/node/error'
 import {formatPackageManagerCommand} from '@shopify/cli-kit/node/output'
+import {LocalStorage} from '@shopify/cli-kit/node/local-storage'
 
 interface InitOptions {
   name: string
@@ -47,6 +48,8 @@ async function init(options: InitOptions) {
   const githubRepo = parseGitHubRepositoryReference(options.template)
 
   await ensureAppDirectoryIsAvailable(outputDirectory, hyphenizedName)
+  // Remove cache from previous projects in the same directory
+  await clearCache(outputDirectory)
 
   renderInfo({
     body: [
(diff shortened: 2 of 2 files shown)


amcaplan (commented):
I'm a bit concerned about this approach, as it creates an implicit coupling between [the `app` package's implementation of local storage](https://github.com/Shopify/cli/blob/ce7671f85aca6f6124e35d0ba87fb05beb05e7ff/packages/app/src/cli/services/local-storage.ts) and this package. If `app` were ever to change, this would have to be changed, but nothing would make that obvious.

I suspect this is going to be a continuing trend, where `create-app` needs to reference some general `app` ideas, and a generalized solution would be worthwhile.

I don't think this information belongs in `cli-kit`, `app` is still the right place, but maybe we should consider exporting certain shared concepts.

In the past, `create-app` couldn't assume `app` was present, but now with global CLI it should always be present, right? So if we had an export, we could do something like:

```ts
import {clearCachedAppInfo} from '@shopify/app/local-storage'

// ...

// Remove cache from previous projects in the same directory
await clearCachedAppInfo(outputDirectory)
```

which would make the connection more explicit.

I presume this wouldn't work exactly this way with bundled CLI, so maybe @isaacroldan could help clarify if this approach is possible?

If this is going to be more trouble than it's worth, I'm OK with merging as is, since it's a rare bug anyway, so the effects of breakage are limited. But I think it would be nice to solve this in a clean, repeatable way.

isaacroldan:
Agree that is not the ideal approach having the local storage defined in two places but thought that the simplification was worth it.

`create-app` currently doesn't depend on the `app` package, so it is not included in its bundle (we have two bundles, one for `@shopify/cli` and one for `@shopify/create-app`). If we were to add this dependency, my guess is that `esbuild` will be smart enough to only include that specific file from `app` and nothing else but we'd need to try

gonzaloriestra:
Good point, I had the same concern. Let me check if adding the dependency doesn't include everything.

isaacroldan:
There is another problem, `app` depends on `create-app` to import and share the init command. Not sure if that circular dependency would be a problem 😬

gonzaloriestra:
Yes, it is 😅:

```
Could not execute command because the task graph has a circular dependency

create-app:build --> app:build --> create-app:build
```

isaacroldan:
The solution would be to move the init command to `app` and import it in `create-app`, but that's a bigger task

PR #4190, opened 2024-07-11
Download javy and function-runner directly

jeffcharles:

The current approach for working with Javy and function-runner is the Shopify CLI unconditionally downloads the latest released version of each tool. This means it's impossible to make breaking changes to the command line interface for these tools. Any bugs introduced in the latest released versions of these tools are also rolled out to all users of the Shopify CLI at the same time.

This change makes the Shopify CLI depend on a specific version of Javy and a specific version of function-runner. That way we can be certain that the way the Shopify CLI is invoking tools is compatible with command line interface these tools expose. This also means that if users find a regression in the version of Javy or funtion-runner that the Shopify CLI is using, they can downgrade their Shopify CLI version to use an older version of Javy or function-runner. Similarly, when a new version of Javy or function-runner is released, nightly versions of the Shopify CLI depend on these versions of Javy and function-runner and users can use a nightly version of the Shopify CLI and let us know if there are any issues that weren't caught when releasing the new versions of Javy or the function-runner.

Instead of using `npm exec javy-cli -- [args...]` or `npm exec @shopify/function-runner -- [args...]` to download and install Javy and function-runner, perform the download in the Shopify CLI, and exec the downloaded binary.

### How to test your changes?

Assuming you are using MacOS or Linux:

1. Create a Shopify app containing at least one JS Function extension if you do not already have one.
2. `pnpm shopify app build --path <path_to_app>` will perform the initial download of Javy.
3. Verify there are no errors reported.
4. `pnpm shopify app build --path <path_to_app>` will test that it uses the downloaded version of Javy.
5. Verify there are no errors reported.
6. `echo "{}" | pnpm shopify app function run --path <path_to_js_function_extension>` will test the initial download of function-runner.
7. Verify there are no errors reported.
8. `echo "{}" | pnpm shopify app function run --path <path_to_js_function_extension>` will it uses the downloaded version of function-runner.
9. Verify there are no errors reported.

If you need to re-test the original downloads and are using MacOS, run `rm -r ~/Library/Caches/javy && rm -r ~/Library/Caches/function-runner`.

### Measuring impact

### Checklist

### DIFF
--- a/packages/app/src/cli/services/function/build.ts
+++ b/packages/app/src/cli/services/function/build.ts
@@ -12,7 +12,86 @@ import {AbortSignal} from '@shopify/cli-kit/node/abort'
 import {renderTasks} from '@shopify/cli-kit/node/ui'
 import {pickBy} from '@shopify/cli-kit/common/object'
 import {runWithTimer} from '@shopify/cli-kit/node/metadata'
-import {Writable} from 'stream'
+import cachedir from 'cachedir'
+import {PipelineSource, Writable} from 'stream'
+import stream from 'stream/promises'
+import fs from 'fs'
+import * as gzip from 'zlib'
+
+const JAVY_VERSION = 'v3.0.1'
+const FUNCTION_RUNNER_VERSION = 'v5.1.3'
+
+// The logic for determining the download URL and what to do with the response stream is _coincidentally_ the same for
+// Javy and function-runner for now. Those methods may not continue to have the same logic in the future. If they
+// diverge, make `Binary` an abstract class and create subclasses to handle the different logic polymorphically.
+class DownloadableBinary {
+  readonly name: string
+  readonly version: string
+  readonly directory: string
+  readonly path: string
+  private readonly gitHubRepo: string
+
+  constructor(name: string, version: string, gitHubRepo: string) {
+    this.name = name
+    this.version = version
+    this.directory = cachedir(name)
+    this.path = joinPath(this.directory, `${this.name}-${this.version}`)
+    this.gitHubRepo = gitHubRepo
+  }
+
+  downloadUrl(processPlatform: string, processArch: string) {
+    let platform
+    let arch
+    switch (processPlatform.toLowerCase()) {
+      case 'darwin':
+        platform = 'macos'
+        break
+      case 'linux':
+        platform = 'linux'
+        break
+      case 'win32':
+        platform = 'windows'
+        break
+      default:
+        throw Error(`Unsupported platform ${processPlatform}`)
+    }
+    switch (processArch.toLowerCase()) {
+      case 'arm':
+      case 'arm64':
+        arch = 'arm'
+        break
+      // A 32 bit arch likely needs that someone has 32bit Node installed on a
+      // 64 bit system, and wasmtime doesn't support 32bit anyway.
+      case 'ia32':
+      case 'x64':
+        arch = 'x86_64'
+        break
+      default:
+        throw Error(`Unsupported architecture ${processArch}`)
+    }
+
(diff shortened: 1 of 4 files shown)


packages/app/src/cli/services/function/build.ts
```diff
@@ -12,7 +12,86 @@ import {AbortSignal} from '@shopify/cli-kit/node/abort'
 import {renderTasks} from '@shopify/cli-kit/node/ui'
 import {pickBy} from '@shopify/cli-kit/common/object'
 import {runWithTimer} from '@shopify/cli-kit/node/metadata'
-import {Writable} from 'stream'
+import cachedir from 'cachedir'
```
isaacroldan, line 15: We already have a dependency for this, is called `env-paths`.

It's available only in `cli-kit` but you can either re-export the cache methods from there.

packages/app/src/cli/services/function/build.ts
```diff
@@ -12,7 +12,86 @@ import {AbortSignal} from '@shopify/cli-kit/node/abort'
...
+  readonly version: string
+  readonly directory: string
+  readonly path: string
+  private readonly gitHubRepo: string
+
+  constructor(name: string, version: string, gitHubRepo: string) {
+    this.name = name
+    this.version = version
+    this.directory = cachedir(name)
+    this.path = joinPath(this.directory, `${this.name}-${this.version}`)
```
isaacroldan, line 38: does this need to be on a cache directory of the system? For the `cloudflared` binary we do something similar and we are saving it inside the CLI folder in `./bin`. (look for a file called `install-cloudflared.ts`)

I'm ok with using a cache directory, just wanted to understand if it's better than just storing in the same folder as the CLI.

packages/app/src/cli/services/function/build.ts
```diff
@@ -319,3 +396,30 @@ JavaScript exports with camelCase names are automatically mapped to kebab-case W
...
+    .stat(bin.path)
+    .then(() => true)
+    .catch(() => false)
+  if (isInstalled) {
+    return
+  }
+
+  const url = bin.downloadUrl(process.platform, process.arch)
+  outputDebug(`Downloading ${bin.name} ${bin.version} from ${url} to ${bin.path}`)
+  await fs.promises.mkdir(bin.directory, {recursive: true})
```
isaacroldan, line 411: We don't use `fs` directly, we have a wrapper to make sure it works across all OSes.

For this use our `await mkDir(...)`

packages/app/src/cli/services/function/build.ts
jeffcharles: It doesn't really matter where the binary is downloaded to. As long as it's persistent and we don't have to frequently re-download it, it should be fine. I can update to write it to `./bin`.

jeffcharles: It's been updated to write to `bin`.

jeffcharles:
The snapshot build works! The Javy and `function-runner` binaries were written to `~/Library/pnpm/global/5/node_modules/@shopify/cli/bin`.

packages/app/src/cli/services/function/binaries.ts
```diff
@@ -0,0 +1,115 @@
...
+      case 'ia32':
+      case 'x64':
+        arch = 'x86_64'
+        break
+      default:
+        throw Error(`Unsupported architecture ${processArch}`)
+    }
+
+    const archPlatform = `${arch}-${platform}`
+    // These are currently the same between both binaries _coincidentally_.
```
DuncanUszkay1, line 61: Is this coincidence something that might change when we migrate to a newer version?

PR #4197, opened 2024-07-12
add deploy marketing activity extension logic

billycai:
✅ Needs to wait for this ticket to ship: https://github.com/Shopify/shopify/pull/536750
Addresses CLI portion of this ticket https://github.com/Shopify/marketing-automations/issues/2042

Adds logic to migrate marketing activity extensions on deploy. Some schema changes to note:
1. We are removing the id for the fields during migration and generating it on deploy. This is not currently being used, but we decided to keep it on the object to minimize the risk of breaking changes on the BE
2. We are converting `app_api_url` to `api_path` by simply stripping the host. We will be asking partners to verify the path is correct using the documentation. When looking at the currently stored `app_api_url` compared to the base url of the app, a lot of marketing partner's base url have deviated quite a bit from the stored `app_api_url` so there's no feasible way to migrate it accurately. 

### How to test your changes?

1. Follow instructions in this [PR](https://github.com/Shopify/cli/pull/4231) to import a marketing activity extension
3. checkout this branch in `shopify` repo: `mae_model_validation`
4. in CLI command prompt run: `pnpm shopify app deploy --path  <path to app>`

NOTE: this will fail right now until we update the validation in core

### Measuring impact

### Checklist

### DIFF
--- a/packages/app/src/cli/services/context/identifiers-extensions.ts
+++ b/packages/app/src/cli/services/context/identifiers-extensions.ts
@@ -6,6 +6,7 @@ import {createExtension} from '../dev/create-extension.js'
 import {IdentifiersExtensions} from '../../models/app/identifiers.js'
 import {getUIExtensionsToMigrate, migrateExtensionsToUIExtension} from '../dev/migrate-to-ui-extension.js'
 import {getFlowExtensionsToMigrate, migrateFlowExtensions} from '../dev/migrate-flow-extension.js'
+import {getMarketingActivtyExtensionsToMigrate} from '../dev/migrate-marketing-activity-extension.js'
 import {AppInterface} from '../../models/app/app.js'
 import {DeveloperPlatformClient} from '../../utilities/developer-platform-client.js'
 import {getPaymentsExtensionsToMigrate, migrateAppModules} from '../dev/migrate-app-module.js'
@@ -15,6 +16,7 @@ import {SingleWebhookSubscriptionType} from '../../models/extensions/specificati
 import {outputCompleted} from '@shopify/cli-kit/node/output'
 import {AbortSilentError} from '@shopify/cli-kit/node/error'
 import {groupBy} from '@shopify/cli-kit/common/collection'
+import {isShopify} from '@shopify/cli-kit/node/context/local'
 
 interface AppWithExtensions {
   extensionRegistrations: RemoteSource[]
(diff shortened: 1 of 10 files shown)


packages/app/src/cli/services/dev/migrate-marketing-activity-extension.ts
```diff
@@ -0,0 +1,86 @@
...
+    const typeMatch = typesMap.get(localSource.type) === remoteSource?.type
+
+    if (remoteSource && typeMatch) {
+      accumulator.push({local: localSource, remote: remoteSource})
+    }
+    return accumulator
+  }, [])
+}
+
+export async function migrateMarketingActivityExtensions(
```
isaacroldan, line 41: can't you use the generic `migrateAppModules` function from `migrate-app-module.ts`?

packages/app/src/cli/services/context/identifiers-extensions.ts
```diff
@@ -63,6 +72,18 @@ export async function ensureExtensionsIds(
     remoteExtensions = remoteExtensions.concat(newRemoteExtensions)
   }
 
+  if (marketingActivityExtensionsToMigrate.length > 0) {
+    const confirmedMigration = await extensionMigrationPrompt(marketingActivityExtensionsToMigrate, false)
+    if (!confirmedMigration) throw new AbortSilentError()
+    const newRemoteExtensions = await migrateMarketingActivityExtensions(
```
isaacroldan, line 78: same here, you should be able to use the generic `migrateAppModules`, as payments do just above this

packages/app/src/cli/services/dev/migrate-marketing-activity-extension.ts
billycai: Good call, didn't realize the functions were so similar, updated

packages/app/src/cli/models/extensions/specifications/marketing_activity_extension.ts
```diff
@@ -1,22 +1,28 @@
...
       referring_domain: config.referring_domain,
       is_automation: config.is_automation,
       use_external_editor: config.use_external_editor,
       preview_data: config.preview_data,
-      fields: config.fields,
+      fields: config.fields.map((field) => ({
+        ...field,
+        // NOTE: we're not using this id anywhere, generating it to satisfy the schema
+        // decided not to remove it from the schema for now to minimize the risk of breaking changes
+        id: randomUUID(),
```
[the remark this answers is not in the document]
billycai: This won't be indefinite. Given the tight timeline for the migration, we decided against changing the API here as it was referenced in a lot of places in the BE code so we wanted to avoid making that change for now. We do plan on deprecating this extension altogether in favour of a new extension, but there's no hard timeline for that yet. We're hoping that will happen sometime next year.

PR #4198, opened 2024-07-12
Show prompt when running function command outside of function directory

andrewhassan:

If a function command is run outside of a function directory, the command simply fails. As an improvement to the UX, we should prompt the user to choose which function they'd like to run the command on.

Closes https://github.com/Shopify/shopify-functions/issues/229

This PR adds a prompt if the CLI is in an interactive terminal so that the user can choose the function to run a function command on (if they're not in a function directory).

### How to test your changes?

Run a function command (e.g. `function run`) outside of a function directory with an interactive terminal

image

### Measuring impact

### Checklist

### DIFF
--- a/packages/app/src/cli/services/function/common.ts
+++ b/packages/app/src/cli/services/function/common.ts
@@ -6,6 +6,8 @@ import {FunctionConfigType} from '../../models/extensions/specifications/functio
 import {resolvePath, cwd} from '@shopify/cli-kit/node/path'
 import {AbortError} from '@shopify/cli-kit/node/error'
 import {Flags} from '@oclif/core'
+import {isTerminalInteractive} from '@shopify/cli-kit/node/context/local'
+import {renderAutocompletePrompt} from '@shopify/cli-kit/node/ui'
 
 export const functionFlags = {
   path: Flags.string({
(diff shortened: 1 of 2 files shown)


packages/app/src/cli/services/function/common.ts
```diff
@@ -30,10 +32,20 @@ export async function inFunctionContext({
...
+  const allFunctions = app.allExtensions.filter(
+    (ext) => ext.isFunctionExtension,
+  ) as ExtensionInstance<FunctionConfigType>[]
+  const ourFunction = allFunctions.find((fun) => fun.directory === path)
+
   if (ourFunction) {
     return callback(app, ourFunction)
+  } else if (isTerminalInteractive()) {
+    const selectedFunction = await renderAutocompletePrompt({
+      message: 'Which function would you like to run the command on?',
```
nickwesselman, line 44: This feels a bit verbose. Thinking about this as a "conversation" with the user, how about just "Which function?"

> `shopify app function build`
>
> ? Which function?

nickwesselman (approved):

DuncanUszkay1 (approved):
Code LGTM, no opinion on message. Personally, I don't think I even register those messages when I'm using the CLI since it's clear from the options what they're asking.

packages/app/src/cli/services/function/common.ts
nickwesselman: Let's go with "Which function?"

PR #4204, opened 2024-07-15
Store API key and app ID

amcaplan:

Currently App Management is storing app ID in the place where API key should go, leading to a lot of general weirdness. It also breaks certain functional things.

Stores both the app ID and the API key in the app TOML. This is not ideal, and we should eventually replace it with API key only, especially given that we allow passing an API key as a flag for many commands.

### How to test your changes?

1. `config link`-ing an app to App Management adds `app_id` to the app TOML
2. `config link`-ing an app to Partners doesn't change
3. `USE_APP_MANAGEMENT_API=1 bin/spin pnpm shopify app info --web-env --path /path/to/your/app` works and displays API key and secret

### Measuring impact

### Checklist

### DIFF
--- a/packages/app/src/cli/utilities/developer-platform-client/app-management-client/graphql/create-app.ts
+++ b/packages/app/src/cli/utilities/developer-platform-client/app-management-client/graphql/create-app.ts
@@ -6,6 +6,7 @@ export const CreateAppMutation = gql`
     appCreate(appSource: $appSource, name: $name) {
       app {
         id
+        key
       }
       userErrors {
         category
@@ -32,6 +33,7 @@ export interface CreateAppMutationSchema {
   appCreate: {
     app: {
       id: string
+      key: string
     }
     userErrors: {
       field: string[]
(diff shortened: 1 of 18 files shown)


packages/app/src/cli/utilities/developer-platform-client/app-management-client/graphql/create-app.ts
```diff
@@ -32,6 +33,7 @@ export interface CreateAppMutationSchema {
   appCreate: {
     app: {
       id: string
+      key: string
```
isaacroldan, line 36: do we need the `activeRoot` here to get the secret?

amcaplan: Hmmmm, I guess for `dev` that might be relevant. It wouldn't be relevant in any other case I can think of, so it didn't come up in testing.

In the interest of shipping, I'll leave it out for now, and you should be able to add it in and test it in the context of `dev`.

PR #4216, opened 2024-07-17
feat: Add prop to renderTextPrompt to hide prompt message and bottom bar

jamesmengo:

The existing implementation was not a great fit for cases when we want to accept numerous inputs such as in an REPL loop. A required `prompt` message and `bottom bar` added noise to every evaluated expression (example in video below)

This change allows us to:
- accept user input in an interactive console style while staying consistent with CLI Style guidelines on accepting input
- avoid adding additional noise in the `theme console` command
- maintain visual parity with the ruby implementation

- Provides a prop that allows users to hide the prompt message and bottom bar when calling `renderTextPrompt`
_- Note: Isaac and I have aligned on using a `minimal` prop to hide both rather than individual props for the prompt message and bottom bar, though this stance is not written in stone_

**Before**
image

**After**
image

### How to test your changes?
- You can run `pnpm test` 
OR
1. Add the following snippet [here](https://github.com/Shopify/cli/blob/main/packages/theme/src/cli/commands/theme/console.ts#L40)

```
async function recursiveRenderTextPrompt() {
      await renderTextPrompt({message: '', minimal: true})
      await recursiveRenderTextPrompt()
}
    await recursiveRenderTextPrompt()
``` 
2. Execute the command `pnpm shopify theme console`
3. Notice that the prompt message (`? <MESSAGE><Punctuation>`) and bottom bar does not render

### Post-release steps
Documentation is updated in this PR
Feel free to let me know if there are additional updates needed!

### Measuring impact

### Checklist
- [x] I've considered possible cross-platform impacts (Mac, Linux, Windows

### DIFF
--- a/docs/cli-kit/ui-kit/readme.md
+++ b/docs/cli-kit/ui-kit/readme.md
@@ -193,6 +193,8 @@ Very similar to `renderSelectPrompt` with the difference that you can provide a 
 
 Shows a text field and waits for the user to input something and press `Enter` to submit.
 
+This requires a `message` to display a prompt for user input. If `skipPromptMessage` is `true`, the prompt will not be displayed.
+
 Unless `allowEmpty` is set to `true` the user will see a validation error if they attempt to press `Enter` without having typed anything. If `allowEmpty` is `true` then an empty string will be returned in case they immediately press `Enter`. This can be useful for optional fields.
 
 `defaultValue` can be used to show a default value with a dimmed text style. The user can either press `Enter` immediately if they wish to submit the default value or start typing in order to override it.
--- a/docs/cli-kit/ui-kit/readme.md
+++ b/docs/cli-kit/ui-kit/readme.md
@@ -193,15 +193,15 @@ Very similar to `renderSelectPrompt` with the difference that you can provide a 
 
 Shows a text field and waits for the user to input something and press `Enter` to submit.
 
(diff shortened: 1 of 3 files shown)


isaacroldan:
Why do we need this? 🤔 
You don't want to show any message in your prompt? why not?

jamesmengo:
> Why do we need this? 🤔 You don't want to show any message in your prompt? why not?

This enables the use of `renderTextPrompt` to accept input for the `theme console` command without adding noise to the terminal each time an expression is evaluated.

The `console` command serves a `REPL` loop to users, where the user is expected to evaluate numerous expressions rather than answering a one-off prompt. Rendering a message / prompt every time an expression is evaluated would unnecessarily clutter the terminal in my opinion

This also maintains visual parity with the current implementation

----

The current Ruby implementation looks like this

https://github.com/user-attachments/assets/7b19fad2-422d-4160-916e-aafaa9df787d

jamesmengo:
good call @isaacroldan! - I checked in with liquid storefronts UX and the consensus is that we can hide the bottom bar to maintain parity, though that stance isn't particularly held strongly.

My question for you is whether you think we should expose two separate props (one for the message, one for the bottom bart) to provide more granularity for future use cases, or to have something like a `minimal` prop that will hide both prompt message and bottom bar

PR #4217, opened 2024-07-17
list and select dev stores

gracejychang:

Resolves https://github.com/Shopify/develop-app-inner-loop/issues/1899

### How to test your changes?

See 🎩  instructions in https://github.com/Shopify/develop-app-inner-loop/issues/1899#issuecomment-2249057719

### Measuring impact

### Checklist


packages/app/src/cli/utilities/developer-platform-client/app-management-client.test.ts
```diff
@@ -113,3 +151,37 @@ describe('allowedTemplates', () => {
...
 })
+
+describe('devStoresForOrg', () => {
+  test('returns a list of OrganizationStore typed stores, from query response', async () => {
+    const mockStore1 = LIST_DEV_STORE_RESPONSE_VALUE.organization.properties.edges[0]?.node
+    const mockStore2 = LIST_DEV_STORE_RESPONSE_VALUE.organization.properties.edges[1]?.node
+
+    const mockAppManagementClient = new AppManagementClient()
+    vi.mocked(businessPlatformOrganizationsRequest).mockResolvedValue(LIST_DEV_STORE_RESPONSE_VALUE)
+
```
gracejychang, line 162: Getting a `mockResolvedValue` function doesn't exist error, though not sure why?

packages/app/src/cli/utilities/developer-platform-client/app-management-client/graphql/list-dev-stores.ts
```diff
@@ -0,0 +1,63 @@
+import {gql} from 'graphql-request'
+
+// eslint-disable-next-line no-warning-comments
+/* TODO for more options like pagination/sort/search see https://github.com/Shopify/business-platform/pull/19676#issuecomment-2165895156 */
```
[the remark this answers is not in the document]
amcaplan: I don't think we need to be able to search for the store. Generally speaking, if someone want to specify a store, they can use the `--store` flag pretty easily. It's come up exceedingly rarely that someone exceeded 50 available stores on an org and got confused, but it was never a high enough priority to really jump in and solve this, because most users in huge orgs do figure this out just fine.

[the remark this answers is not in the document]
gracejychang: I've remove this part of the query, to simplify things

packages/app/src/cli/utilities/developer-platform-client/app-management-client.test.ts
gracejychang: Turns out I have to include the mocked files, ie. `vi.mock('@shopify/cli-kit/node/api/business-platform')`

packages/app/src/cli/utilities/developer-platform-client/app-management-client.test.ts
```diff
@@ -113,3 +173,82 @@ describe('allowedTemplates', () => {
...
+describe('devStoresForOrg', () => {
+  test('returns a list of OrganizationStore typed stores, from query response', async () => {
+    const mockStore1 = STORE1_NODE
+    const mockStore2 = STORE2_NODE
+
+    const mockAppManagementClient = new AppManagementClient()
+    vi.mocked(businessPlatformOrganizationsRequest).mockResolvedValue(LIST_DEV_STORE_RESPONSE_VALUE)
+
+    const got = await mockAppManagementClient.devStoresForOrg(ORG1.id)
```
gracejychang, line 185: To me it makes sense to test this call directly here, but in this `devStoresForOrg` we call `await this.businessPlatformToken()` to get the token to call with the request. This blocks this test from running bc of 
```
    if (isUnitTest()) {
      throw new Error('AppManagementClient.businessPlatformToken() should not be invoked dynamically in a unit test')
    }
```

I want to understand why this check is here, and if there's a way to can get around it?  I've attempted to mock `ensureAuthenticatedBusinessPlatform` and I can't set `_businessPlatformToken` directly because it's a private field of `AppManagementClient`.

Thoughts? @amcaplan @isaacroldan

packages/cli-kit/src/private/node/session/scopes.ts
```diff
@@ -35,7 +35,9 @@ function defaultApiScopes(api: API, systemEnvironment = process.env): string[] {
     case 'partners':
       return ['cli']
     case 'business-platform':
-      return ['destinations']
+      return isTruthy(systemEnvironment.USE_APP_MANAGEMENT_API)
+        ? ['destinations', 'store-management']
+        : ['destinations']
```
isaacroldan, line 40: I don't think we need this, as long as the scope is added to identity, this can be enabled for both cases.

packages/app/src/cli/utilities/developer-platform-client/app-management-client.ts
```diff
@@ -357,8 +367,27 @@ export class AppManagementClient implements DeveloperPlatformClient {
...
 
-  async devStoresForOrg(_orgId: string): Promise<OrganizationStore[]> {
-    return []
+  // we are returning OrganizationStore type here because we want to keep types consistent btwn
+  // partners-client and app-management-client. Since we need transferDisabled and convertableToPartnerTest values
+  // from the Partners OrganizationStore schema, we will return this type for now
+  async devStoresForOrg(orgId: string): Promise<OrganizationStore[]> {
+    const storesResult = await businessPlatformOrganizationsRequest<ListDevStoresQuerySchema>(
+      ListDevStoresQuery,
+      this._businessPlatformToken ?? (await this.businessPlatformToken()),
```
isaacroldan, line 376: this should be just `await this.businessPlatformToken()`, that function already returns `_businessPlatformToken` if it exists.

packages/app/src/cli/utilities/developer-platform-client/app-management-client.ts
```diff
@@ -357,8 +367,27 @@ export class AppManagementClient implements DeveloperPlatformClient {
...
+
+    return storesResult.organization.properties.edges.map((edge) => {
+      const store = edge.node
+      return {
+        shopId: store.externalId,
+        link: store.primaryDomain,
+        shopDomain: store.primaryDomain,
+        shopName: store.name,
+        transferDisabled: true,
+        convertableToPartnerTest: true,
```
isaacroldan, line 388: Since you are doing this mapping in two places (here and on line 738) i'd extract this mapping to a function so we don't duplicate it. Something like `mapBPStoreToGenericStore` or probably a better name :P

packages/app/src/cli/utilities/developer-platform-client/app-management-client.test.ts
isaacroldan: @amcaplan or @gonzaloriestra have more context on why this is here, but i think we should remove that error.

Mocking `ensureAuthenticatedBusinessPlatform` seems like the right approach

packages/app/src/cli/services/context.ts
```diff
@@ -167,37 +167,37 @@ export async function ensureDevContext(options: DevContextOptions): Promise<DevC
...
   if (!selectedApp || !selectedStore) {
-    // if we have selected an app or a dev store from a command flag, we keep them
     // if not, we try to load the app or the dev store from the current config or cache
     // if that's not available, we prompt the user to choose an existing one or create a new one
-    const [_selectedApp, _selectedStore] = await Promise.all([
-      selectedApp ||
-        remoteApp ||
+    const [cachedApp, cachedStore] = await Promise.all([
+      remoteApp ||
         (cachedInfo?.appId && appFromId({apiKey: cachedInfo.appId, organizationId: orgId, developerPlatformClient})),
```
isaacroldan, line 186: you need to keep `selectedApp` here, because this code path can be executed if `selectedStore` is not present, but `selectedApp` might!

isaacroldan: same for the `selectedStore`

packages/app/src/cli/utilities/developer-platform-client/app-management-client.ts
gracejychang: oh yes, sorry I didn't remove this, I was playing around with ways to bypass that `isUnitTest()` check from above 😅

packages/app/src/cli/utilities/developer-platform-client/app-management-client/graphql/list-dev-stores.ts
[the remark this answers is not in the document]
gracejychang: These files will be removed and replaced by [this PR](https://github.com/Shopify/cli/pull/4226) :)

packages/app/src/cli/services/context.ts
```diff
@@ -167,37 +167,37 @@ export async function ensureDevContext(options: DevContextOptions): Promise<DevC
   developerPlatformClient = remoteApp?.developerPlatformClient ?? developerPlatformClient
 
   let orgId = getOrganization() || cachedInfo?.orgId
-  if (!orgId) {
+  if (!orgId || !developerPlatformClient) {
     const org = await selectOrg()
     developerPlatformClient = selectDeveloperPlatformClient({organization: org})
```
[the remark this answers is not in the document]
gracejychang: Yes, is there a way we can make this more clear? Part of my confusion comes from the fact that before we call `ensureDevContext`, we actually already set the `developerPlatformClient` in `dev` [here](https://github.com/Shopify/cli/blob/main/packages/app/src/cli/services/dev.ts#L90), based on what I assume is the local configuration? 

I guess I don't really understand why we keep re-setting it, we are selecting a client based on `configuration` first, passing that down, then selecting a client based on `organization` or `app`.. Wouldn't it be more straightforward to only select the client based on the `app` or `organization`? Vice versa, if `configuration` has to have `orgId` to point to `app-management-client`, how does `app` or `organization` change this?

Also why does `DevContextOptions` always need to have `DeveloperPlatformClient` to be present? 

Sorry for all the questions and thank you for your patience!! I will def revert this change but I just want to better understand how ``DeveloperPlatformClient` works 😅

packages/app/src/cli/utilities/developer-platform-client/app-management-client.test.ts
amcaplan: > A bit related too, I see we used to test API fetches from Partners [here](https://github.com/Shopify/cli/blob/main/packages/app/src/cli/services/dev/fetch.test.ts#L163) (Unrelated but I think this should be removed cause the corresponding function this is testing doesn't seem to be used anywhere any more). Is there a place where we documented the decision to rely on type assertions when it came to these new query calls? Either way, I think CLI should be consistent in the ways we test queries, and this should be explicitly documented for anyone who is looking to add similar code in the future.

This is a really important point. There is some scarring here from code history, and it would be sensible to clean up and document.

Before this project began, there was no `PartnersClient` either. We simply had `dev/fetch.ts` which contained both the Partners calls and the logic around them, plus some other junk that collected. And it made sense to test everything in a consistent way. So we treated the functions as units of work, and tested them whether they were just wrappers for Partners or did much more.

I think when we reorganized the code for this project, it created some confusion around the roles of `PartnersClient` and `dev/fetch.ts`. We have moved over most of the code to `PartnersClient` but a few functions are still hanging on in `dev/fetch.ts`, such as `fetchAppDetailsFromApiKey`. It would be wise for us to finish the work and fully separate API wrapping from logic. And along the way, we can reevaluate the tests, add anything missing, and remove anything that isn't creating value.

amcaplan: > (Unrelated but I think this should be removed cause the corresponding function this is testing doesn't seem to be used anywhere any more)

This is very weird. Since @shauns added `knip` we shouldn't have unused code. I do see that it's used a couple of times in tests, but not in ways that should count as "real" use.

So, for Shaun:

1. Is this expected?
2. If not, should there be some attempt to tweak knip settings to ensure these kinds of cases are caught and eliminated?
3. If it's expected, should we just manage our expectations for knip and be more active about deleting code?

shauns: I can't quite follow what the thing is that should be removed, but can talk about knip etc. a little if that's helpful

- ESLint checks for unused values within modules
- Knip checks this stuff: https://knip.dev/reference/issue-types -- which is mainly about dependencies and exports
- We have Knip configured in its default fashion: if something is used in tests, this counts as valid usage and the thing should still be `export`-ed etc.
- This can be turned off with production mode https://knip.dev/features/production-mode -- configured correctly, you can say that test usage doesn't count as _real_ usage
- We don't use this right now. It's (1) probably not correctly configured and (2) likely to be quite disruptive to introduce when its first activated and (3) at the time of introducing Knip we didn't use a version of Vitest that supported in-source testing https://vitest.dev/guide/in-source 
  - Using in-source testing, we could keep tests that are basically checking implementation details, without needing to export those functions
  - Less exporting is good

I would be up for switching to production mode, or at least seeing what it picks up.

gracejychang: > It would be wise for us to finish the work and fully separate API wrapping from logic. And along the way, we can reevaluate the tests, add anything missing, and remove anything that isn't creating value.


I've removed the tests for now, I think it sounds like we will need to follow up on both clean up for this old `fetch` file, as well as align on how we want to test API calls in the CLI. What do you think is the best way to ensure this gets done? I was thinking of creating an issue in our team board but I'm nervous this might get bogged down by other issues.

packages/app/src/cli/api/graphql/business-platform-organizations/queries/fetch_dev_store_by_domain.graphql
```diff
@@ -5,10 +5,10 @@
       properties(filters: {field: STORE_TYPE, operator: EQUALS, value: "app_development"}, search: $domain, offeringHandles: ["shop"]) {
         edges {
           node {
-            __typename
-            id
-            externalId
             ... on Shop {
+              __typename
+              id
+              externalId
```
gracejychang, line 11: Moved `--typename`, `id` and `externalId` down to the `Shop` level because we only care about `Shop` typed nodes

packages/app/src/cli/utilities/developer-platform-client/app-management-client.ts
```diff
@@ -360,8 +376,18 @@ export class AppManagementClient implements DeveloperPlatformClient {
...
+  // we are returning OrganizationStore type here because we want to keep types consistent btwn
+  // partners-client and app-management-client. Since we need transferDisabled and convertableToPartnerTest values
+  // from the Partners OrganizationStore schema, we will return this type for now
+  async devStoresForOrg(orgId: string): Promise<OrganizationStore[]> {
+    const storesResult = await businessPlatformOrganizationsRequest<ListAppDevStoresQuery>(
+      ListAppDevStores,
+      await this.businessPlatformToken(),
+      orgId,
+    )
+    const shopArray = storesResult?.organization?.properties?.edges.map((value) => value.node as ShopNode) ?? null
```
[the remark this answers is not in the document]
gracejychang: For 1, unfortunately I can't get around that because of the auto-generated type system. For some reason in the BP schema, these values can have a `null` type, even though technically it shouldn't.. 

For 2, great point! I've updated :)

packages/app/src/cli/utilities/developer-platform-client/app-management-client.ts
```diff
@@ -694,8 +720,36 @@ export class AppManagementClient implements DeveloperPlatformClient {
...
+
+    const {organization} = storesResult
+    const bpStoresArray = organization?.properties?.edges.map((value) => value.node as ShopNode) ?? []
+    const storesArray = bpStoresArray.length > 0 ? mapBusinessPlatformStoresToOrganizationStores(bpStoresArray) : []
+
+    return {
+      organizations: {
+        nodes: [
+          {
+            id: organization?.id ?? orgId,
```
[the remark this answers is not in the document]
gracejychang: would `id: organization!.id` work here?

packages/app/src/cli/utilities/developer-platform-client/app-management-client.ts
```diff
@@ -694,8 +720,36 @@ export class AppManagementClient implements DeveloperPlatformClient {
...
+            businessName: organization?.name ?? '',
+            website: 'N/A',
```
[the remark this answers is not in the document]
gracejychang: Tracking[ the issue here](https://github.com/orgs/Shopify/projects/7962/views/2?pane=issue&itemId=72646045) and assigned myself so I don't forget 😅

packages/app/src/cli/utilities/developer-platform-client/app-management-client.ts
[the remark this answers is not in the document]
isaacroldan: I think we should explicitly check and throw an error, something like:
```ts
if (!organization) {
  throw new AbortError...
}
```

isaacroldan (approved):

PR #4220, opened 2024-07-17
Add press-key-to-login to device auth flow

amcaplan:

Recently we switched to device auth by default, which has added some friction to login. This restores the simplicity of pressing any key to log in.

Before:

Screenshot 2024-07-17 at 16 51 24

After:

Screenshot 2024-07-17 at 16 50 32

We only use this flow in non-cloud TTY environments.

### How to test your changes?

Logout and do anything that requires login, e.g. `config link`

### Measuring impact

### Checklist

### DIFF
--- a/packages/cli-kit/src/private/node/session/device-authorization.ts
+++ b/packages/cli-kit/src/private/node/session/device-authorization.ts
@@ -4,7 +4,7 @@ import {IdentityToken} from './schema.js'
 import {identityFqdn} from '../../../public/node/context/fqdn.js'
 import {shopifyFetch} from '../../../public/node/http.js'
 import {outputContent, outputDebug, outputInfo, outputToken} from '../../../public/node/output.js'
-import {BugError} from '../../../public/node/error.js'
+import {AbortError, BugError} from '../../../public/node/error.js'
 import {isCloudEnvironment} from '../../../public/node/context/local.js'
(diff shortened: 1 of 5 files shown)


packages/cli-kit/src/private/node/session/device-authorization.ts
```diff
@@ -47,11 +50,25 @@ export async function requestDeviceAuthorization(scopes: string[]): Promise<Devi
...
+    outputInfo(
+      outputContent`Opened link to start the auth process: ${outputToken.link(
+        'Login link',
+        jsonResult.verification_uri_complete,
+      )}`,
+    )
+  } else {
+    outputInfo(outputContent`User verification code: ${jsonResult.user_code}`)
+    outputInfo(
+      outputContent`👉 Open this link to start the auth process: ${outputToken.link(
```
shauns, line 66: I think in this case the `green` approach is better, as we've already established this terminal might not be TTY aware anyway, and it reads a bit better with out the extra formatting we get with `link` in unsupported environments

amcaplan: Is that actually true? In cloud environments, are links generally non-clickable?

shauns (approved):
Looking good 👍

phrfpeixoto:
This change broke my development process using docker containers. Please see #5496

PR #4221, opened 2024-07-17
Add --json flag to app logs

DuncanUszkay1:

Fixes https://github.com/Shopify/shopify-functions/issues/297

We need to support a `--json` flag for clients looking to plug into logs programatically with `jq` or similar. At the same time, we were reaching a critical mass (3) of poll clients each with their own unique error handling behaviour, and it was time to simplify and refactor them together.

- Adds `--json` flag
- Refactors poll clients to use a shared error handler
- Adds missing tests

### How to test your changes?

Tophatting is required for the three poll clients:
- `pnpm run shopify app dev`
- `pnpm run shopify app logs`
- `pnpm run shopify app logs --json`

For each, we'll need to test both with successful polling and errors returned from partners.

### Checklist

  - Need to sync with product on this

### DIFF
--- a/packages/app/src/cli/services/context.ts
+++ b/packages/app/src/cli/services/context.ts
@@ -42,6 +42,7 @@ import {outputContent} from '@shopify/cli-kit/node/output'
 import {getOrganization} from '@shopify/cli-kit/node/environment'
 import {basename, joinPath} from '@shopify/cli-kit/node/path'
 import {glob} from '@shopify/cli-kit/node/fs'
+import {sniffForJson} from '@shopify/cli-kit/node/path'
 
 export const InvalidApiKeyErrorMessage = (apiKey: string) => {
   return {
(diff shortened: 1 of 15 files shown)


packages/app/src/cli/services/context.ts
```diff
@@ -816,6 +816,7 @@ interface ReusedValuesOptions {
  */
 function showReusedDevValues({organization, selectedApp, selectedStore, cachedInfo}: ReusedValuesOptions) {
   if (!cachedInfo) return
+  if (sniffForJson()) return
```
DuncanUszkay1, line 819: This ensures that it doesn't display any info boxes when the `--json` flag is in use, to avoid breaking JSON parsing libraries like `jq`

lopert (approved):
Tophatted all 3 "clients" to see them working and how they handled a 429 error from partners as well. 🎩 👍

mssalemi (approved):
LGTM! I like the separated `handleFetchAppLogsError`.

DuncanUszkay1:
Going to ship this ASAP to avoid conflicts while I focus on some emergent SLO work

PR #4191, opened 2024-07-11
[Themes] Theme Console - ensureReplEnv 1/2 - Storefront Password Prompting

jamesmengo:
### WHY are these changes introduced?
- Part of https://github.com/Shopify/develop-advanced-edits/issues/257

### WHAT is this pull request doing?

```mermaid
graph LR
    A[Start] --> B{Is the storefront password protected?}
    B -- No --> I[End]
    B -- Yes --> D{Is the --store-password flag provided?}
    D -- No --> E[Prompt the user for a password]
    D -- Yes --> F[Validate the provided --store-password flag]
    E --> G[Validate the prompted password]
    F --> H{Is the password valid?}
    G --> H{Is the password valid?}
    H -- Yes --> I[End]
    H -- No --> E[Prompt the user for a password]
```

### How to test your changes?

1. Ensure your storefront is password protected via the admin
2. `pnpm shopify theme console --dev-preview`
3. You should see a password prompt
  a. Enter a couple of incorrect passwords, then the correct one
4. `pnpm shopify theme console --dev-preview --store-password=<PASSWORD>`
  a. Try with some valid and invalid passwords.

https://github.com/user-attachments/assets/7473062a-8f3b-4b89-8306-14f86e8f9f1f

image

### Measuring impact

How do we know this change was effective? Please choose one:

- [ ] n/a - this doesn't need measurement, e.g. a linting rule or a bug-fix
- [x] Existing analytics will cater for this addition
- [ ] PR includes analytics changes to measure impact

### Checklist
- [x] I've considered possible cross-platform impacts (Mac, Linux, Windows)
- [x] I've considered possible [documentation](https://shopify.dev) changes

PR #4224, opened 2024-07-18
[Themes] Theme Console - Handle the REPL lifecycle

jamesmengo:
- https://github.com/Shopify/develop-advanced-edits/issues/259

EDIT - _This has been modified to use `readline` to accept console inputs _
https://github.com/user-attachments/assets/0645b141-be79-4c82-aff9-84797e1fb9fb

### How to test your changes?
1) Checkout the branch
2) `pnpm shopify theme console --dev-preview`
3) Play around with the console. Note that the error messages will change as the evaluator is implemented. (There are also different layers - the shutdown message, which stops the loop (hard to trigger) and the graceful message which catches the error and keeps the loop going.

Some ideas
- Delimiter - `{{ collections.first }}` or `{% x %}`
- Null value - `f`
- Something that will error out - `JSON(`
- Please feel free to suggest more cases!

### Measuring impact

### Checklist

### DIFF
--- a/packages/theme/src/cli/utilities/repl.ts
+++ b/packages/theme/src/cli/utilities/repl.ts
@@ -11,6 +11,10 @@ import {
 } from '@shopify/cli-kit/node/output'
 import {renderTextPrompt} from '@shopify/cli-kit/node/ui'
 
+interface ErrorOutput {
+  error?: string
+}
+
 // todo - combine config into a single arg
 export async function replLoop(themeSession: DevServerSession, storefrontToken: string, themeId: string, url: string) {
   try {
@@ -29,14 +33,25 @@ export async function replLoop(themeSession: DevServerSession, storefrontToken: 
   }
 }
 
-// todo - handle JSON errors
-function presentEvaluatedValue(evaluatedValue?: string) {
-  if (evaluatedValue === undefined || evaluatedValue === null) {
-    outputInfo(outputContent`${outputToken.cyan('null')}`)
+function presentEvaluatedValue(evaluatedValue?: unknown) {
+  if (hasJsonError(evaluatedValue)) {
+    consoleWarn(
+      "Object can't be printed, but you can access its fields. Read more at https://shopify.dev/docs/api/liquid.",
+    )
     return
   }
 
-  outputInfo(outputContent`${outputToken.cyan(evaluatedValue)}`)
+  if (evaluatedValue === undefined || evaluatedValue === null) {
+    presentValue('null')
+    return
+  }
+
+  const formattedOutput = JSON.stringify(evaluatedValue, null, 2)
+  presentValue(formattedOutput)
+}
+
+function presentValue(value: string) {
+  return outputInfo(outputContent`${outputToken.cyan(value)}`)
 }
 
 function shutdownReplSession(error: unknown) {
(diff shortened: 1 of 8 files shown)


packages/theme/src/cli/utilities/repl.ts
```diff
@@ -0,0 +1,39 @@
+import {DevServerSession} from './theme-environment/types.js'
+import {presentValue} from './repl/presenter.js'
+import {evaluate} from './repl/evaluater.js'
+import {AbortSilentError} from '@shopify/cli-kit/node/error'
+import {consoleWarn, consoleError, outputDebug} from '@shopify/cli-kit/node/output'
+import {renderTextPrompt} from '@shopify/cli-kit/node/ui'
+
+export async function replLoop(themeSession: DevServerSession, themeId: string, url: string) {
+  try {
```
jamesmengo, line 9: todo - test this

packages/theme/src/cli/utilities/repl.ts
```diff
@@ -0,0 +1,37 @@
+import {DevServerSession} from './theme-environment/types.js'
+import {presentValue} from './repl/presenter.js'
+import {evaluate} from './repl/evaluater.js'
+import {AbortSilentError} from '@shopify/cli-kit/node/error'
+import {consoleWarn, outputContent, outputDebug, outputInfo, outputToken} from '@shopify/cli-kit/node/output'
+import {renderTextPrompt} from '@shopify/cli-kit/node/ui'
+
+export async function replLoop(themeSession: DevServerSession, themeId: string, url: string) {
+  try {
+    const inputValue = await renderTextPrompt({message: 'Enter a value'})
```
karreiro, line 10: May we support the history commands with `renderTextPrompt` (as we may notice in the video below)?

https://github.com/user-attachments/assets/8ae95ee3-e13d-4597-931e-2cd15441c52e

While consistent in the UI components is great, it seems like `readline` provides the exact stateful features we need on the context of a REPL tool ([ref](https://github.com/Shopify/cli/pull/4191#discussion_r1686424514))

What do you think, @jamesmengo? :)

packages/theme/src/cli/utilities/repl.ts
```diff
@@ -0,0 +1,63 @@
...
+import {createInterface} from 'readline'
+
+export async function replLoop(themeSession: DevServerSession, themeId: string, url: string) {
+  consoleLog('here')
+  try {
+    if (process.stdin.isTTY) {
+      process.stdin.ref()
+    }
+
+    const rl = createInterface({
```
jamesmengo, line 22: Could use some help testing this! @karreiro

packages/theme/src/cli/utilities/repl.ts
```diff
@@ -0,0 +1,59 @@
...
+
+function shutdownReplSession(error: unknown) {
+  if (error instanceof Error) {
+    outputInfo(outputContent`${outputToken.errorText(`Shopify Liquid console error: ${error.message}`)}`)
+    outputDebug(error.stack || 'Error backtrace not found')
+  }
+}
+
+function hasDelimiter(input: string): boolean {
+  return /\{\{|\}\}|\{%|%\}/.test(input)
```
karreiro, line 58: This matches the Ruby implementation, but I wonder if we could be a bit more specific; something like this: `/^\s*(\{\{|{%)/`, this way we may avoid from failing in some specific valid use cases :)

Issue #1183, opened 2023-01-22
[Feature]: Bitbucket Integration


bobbypiper:
### What area(s) will this request affect?

Theme, Deployment

### What type of change do you want to see?

New feature

### Overview

Replicate the GitHub theme sync integration for those of us who store our corporate repositories on Bitbucket.

### Motivation

I'm trying to implement a sync solution between Shopify and Bitbucket private repositories using Bitbucket pipelines.

Unfortunately, hit a roadblock when running a '--pull' command and committing/merging the changes that are made in theme editor; if a developer makes changes in repo that diverge from the change in theme editor, additional '--pull' commands will simply overwrite the change!

JohnBarry92123:
> This issue seems inactive. If it's still relevant, please add a comment saying so. Otherwise, take no action. → If there's no activity within a week, then a bot will automatically close this. Thanks for helping to improve Shopify's dev tooling and experience.

Please can this ticket be reopened just trying to bump this. Still a critical requirement for us

evaldas-raisutis:
I agree, still pondering if we should build our own or wait for Shopify 🤔

mgmanzella:
👋 hi folks, the CLI is not coupled with the GH integration feature, we don't control if this is supported or not. I relayed this request to the team that does own GH integration internally -- keep upvoting this issue to indicate it's something you want. thanks

PR #4232, opened 2024-07-23
Remove print extension uniqueness validation

alfonso-noriega:
Remove print extension validation which should live in Core.
issue [#1183](https://github.com/orgs/Shopify/projects/7965/views/29?pane=issue&itemId=65371800#:~:text=Validation%20to%20Core-,%231183,-Edit)

**NOTE**: This shouldn't be merged until core validation is merged in this [PR](https://github.com/Shopify/shopify/pull/523523)

Remove the validation as it is implemented in cor in this [PR](https://github.com/Shopify/shopify/pull/523523)

### How to test your changes?
- Create an app with two print extensions with the same target
- In the CLI project, checkout the branch `1183-move-cli-validation-to-core`
- In an spin instance, go to shopify shell and  checkout the branch `1183-move-cli-validation-to-core`
- Try to deploy the app using the local CLI branch, it should return an error mensioning the extension handles and the target

### Measuring impact

### Checklist


vividviolet (commented):
I'm good with removing the validation but it would mean they can still `dev` an app with duplicated print targets even if the backend validation fails. Should we actually be failing the `dev` process when an error happens when attempting to update the draft version?

cc @isaacroldan

alfonso-noriega:
> ...it would mean they can still dev an app with duplicated print targets even if the backend validation fails. Should we actually be failing the dev process when an error happens when attempting to update the extension version?...

@vividviolet,`dev` fails also in with the core validation. as the `draft_version` is persisted and goes through the app_version validation plugin.

alfonso-noriega:
> ...it would mean they can still dev an app with duplicated print targets even if the backend validation fails. Should we actually be failing the dev process when an error happens when attempting to update the extension version?...

@vividviolet,this won't be an issue anymore once the new consistent dev and dev dashboard projects land as they will create a version on `dev`?

isaacroldan:
>  What I'm suggesting is for us to just fail on validation errors instead of letting them go through with dev using a "broken" extension.

I think it makes sense to stop `dev` if there is an error in the draft upload.

PR #4233, opened 2024-07-23
Run link before deploy and dev and a small refactor

MitchDickinson:

Fixes https://github.com/Shopify/develop-app-inner-loop/issues/1921

This PR fixes the first bug in https://github.com/Shopify/develop-app-inner-loop/issues/1919. Previously, if you did a `deploy` _before_ doing any kind of `dev` or `config link`, you would be in a situation where your TOML file was not populated. During the subsequent `deploy`, the CLI would default to [using the PartnerClient](https://github.com/Shopify/cli/blob/mitch/deploy-error-fixes/packages/app/src/cli/utilities/developer-platform-client.ts#L129-L132) and the request would fail since the user was not auth'd to partners.

This PR changes the code to always call `link` from both `deploy` and `dev` contexts, if the remote app has not yet been connected. In practice, if you deploy (or dev) before linking now, the CLI will prompt you for the organization + app information and it will update the TOML for you, before moving onto the rest of the deploy.

- Extracted a function called `linkIfNecessary`, which is now called from `deploy` (via `fetchAppAndIdentifiers`) and `dev` (via `getAppContext`). All the logic should remain unchanged.
- Add a new `enableLinkingPrompt` argument so the caller can opt out of the `link` flow if they want. For example, `ensureVersionsListContext` does _not_  call link even though it still uses `fetchAppAndIdentifiers`. This boolean is only set to true for `deploy` and `release` contexts.
- A lot of test mock changes, unfortunately. This is because most of the code paths use `link` now, which needs to be mocked.

### How to test your changes?
I would appreciate a second 🎩 if possible since I am still very unfamiliar with the CLI and this change required some refactors that touch a broad surface area.

You can checkout the branch and simply follow any of the flows, observing the contents of your TOML file as you work:
- `app init` => `app deploy` (should prompt + populate TOML)
- `app init` => `app dev` (should prompt + populate TOML)
- `app release --reset` (alternatively, if you know how to clear the local cache then you can do that instead of passing --reset but it should do the same thing)

### Measuring impact

### Checklist

### DIFF
--- a/packages/app/src/cli/services/context.test.ts
+++ b/packages/app/src/cli/services/context.test.ts
@@ -929,7 +929,7 @@ api_version = "2023-04"
   })
 })
 
-describe('ensureDeploy Context', () => {
+describe('ensureDeployContext', () => {
   test("fetches the app from the partners' API and returns it alongside the id when identifiers are available locally and the app has no extensions", async () => {
     // Given
     const app = testAppWithConfig({config: {client_id: APP2.apiKey}})
--- a/packages/app/src/cli/services/context.ts
+++ b/packages/app/src/cli/services/context.ts
@@ -428,7 +428,6 @@ export async function ensureDeployContext(options: DeployContextOptions): Promis
   const {reset, force, noRelease} = options
   let developerPlatformClient = options.developerPlatformClient
   // do the link here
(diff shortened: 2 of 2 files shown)


gonzaloriestra (approved):
Works as expected in Spin. Nice work! 👏

PR #4240, opened 2024-07-26
Add default_value to Settings FieldSchema

Johnkinging:

closes https://github.com/Shopify/checkout-web/issues/36347

TL;DR - UiExtensions have the ability to parse fields marked as `required`  so long as they have a "default_value`. But only `required` exists on the schema. This would have gone unnoticed as the CheckoutUiExtension CLI settings schema allowed any key to be passed: https://github.com/Shopify/cli/blob/a3dd9e5eb2a91b7fdf255feb33deac777fb76ee0/packages/app/src/cli/models/extensions/extension-specifications/checkout_ui_extension.ts#L10-L14

However in core we validate this, so including `required` necessitates including `default_value`. These keys were not explicitly defined, and the feature has [not rolled out](https://app.shopify.com/services/internal/app_betas/checkout_ui_extensions_settings_expanded_definitions) , so my guess is this is simply an understandable oversight. 

Because the default value can be any type of allowed metafield type (bool, string, date, etc) the type is set to `any`.

### How to test your changes?

Either roll the `checkout_ui_extensions_settings_expanded_definitions` flag to 100% or apply it to your app from the partners internal.

Deploy a ui_extension with a settings such as
```
[extensions.settings]
[[extensions.settings.fields]]
key = "banner_title"
type = "single_line_text_field"
name = "banner_title"
required = true
default_value = "Whatever"
```

It's probably best to have core on sb/jk/cleanup-configuration-definitions-rebased but the current main should behave the same. Currently `default_value` is removed from the config as it's uploaded to core, and the following message appears:
Screenshot 2024-07-26 at 2 02 39 PM

On this branch the extension should deploy fine, as the default value is provided alongside the required key. Removing the the default value should throw the core invalidation again.

A default value can be added without it being required.

### DIFF
--- a/packages/app/src/cli/models/extensions/schemas.ts
+++ b/packages/app/src/cli/models/extensions/schemas.ts
@@ -49,6 +49,7 @@ export const FieldSchema = zod.object({
   name: zod.string().optional(),
   description: zod.string().optional(),
   required: zod.boolean().optional(),
+  default_value: zod.any().optional(),
   type: zod.string(),
   validations: zod.array(zod.any()).optional(),
 })


Johnkinging:
> We detected some changes at either packages/*/src or packages/cli-kit/assets/cli-ruby/** and there are no updates in the .changeset. If the changes are user-facing, run "pnpm changeset add" to track your changes and include them in the next release CHANGELOG.

This isn't being rolled any time soon, but it is fully baked into core. We'll update the partner facing things when we decide to roll it out. 

Note: It's actually been a feature for a long time.

davejcameron (approved):

PR #4243, opened 2024-07-28
Remove unused and Partners-specific functions from dev/fetch

amcaplan:

It's supposed to be a generic file, but it has some lingering Partners-specific stuff, plus one entirely unused function.

The circular dependency between `dev/fetch` and `PartnersClient` is also certainly non-ideal.

1. Gets rid of `fetchAllDevStores` which wasn't being used anymore
2. Takes a couple other functions and puts any Partners-specific stuff into `PartnersClient`, and any more generic dependencies of those functions (e.g. `Flags`) into `DeveloperPlatformClient`.

The wins are:

1. Reducing the circular dependency (thought `NoOrgError` still needs addressing)
2. Keeping Partners-specific stuff out of the rest of the codebase (isolated within `PartnersClient`)
3. Net deleted LoC
4. Better testing via the injected `testDeveloperPlatformClient` instead of stubbing out functions

### How to test your changes?

Confirm that the various app command still work correctly, especially when pointing to Partners. I'm not sure there's a specific list, as these changes are small but touch critical code paths with a broad impact.

### Measuring impact

### Checklist


packages/app/src/cli/utilities/developer-platform-client/partners-client.ts
```diff
@@ -507,4 +532,17 @@ export class PartnersClient implements DeveloperPlatformClient {
     // Dev Sessions are not supported in partners client.
     throw new Error('Unsupported operation')
   }
+
+  private async fetchOrgAndApps(orgId: string, title?: string): Promise<OrgAndAppsResponse> {
+    const query = FindOrganizationQuery
+    const params: {id: string; title?: string} = {id: orgId}
+    if (title) params.title = title
+    const partnersSession = await this.session()
+    const result: FindOrganizationQuerySchema = await partnersRequest(query, partnersSession.token, params)
```
gonzaloriestra, line 541:
```suggestion
    const params: {id: string; title?: string} = {id: orgId}
    if (title) params.title = title
    const result: FindOrganizationQuerySchema = await this.request(FindOrganizationQuery, params)
```

gonzaloriestra (approved):
Nice work, thanks! 👏

PR #4246, opened 2024-07-29
use camelCase for hash keys in logs

lopert:

Fixes https://github.com/Shopify/shopify-functions/issues/271

Uses [camelcase-keys ](https://www.npmjs.com/package/camelcase-keys)to convert snake_case keys from appLogs.
I chose this because I saw it was already part of our dependencies https://github.com/Shopify/cli/blob/266908ce38dfef4cd28cb7a0eb38a7b2e8e6b4d5/pnpm-lock.yaml#L8361
... although I think I've just created another copy of it? 🤔 

### How to test your changes?
Generate an app log, click through to ensure that all keys are now in camelCase.
Run replay and see that:
- new runs look fine
- old runs now have an "undefined" timestamp, since they are missing the new `logTimestamp` (they have the old `log_timestamp`)
  - is this ok? Should we try to parse both old and new here? or is this break ok?

### Measuring impact

### Checklist


DuncanUszkay1 (approved):
Based on my grep of the lockfile, it looks like v8.0.0 is a subdependency of `meow`:
```
  /meow@6.1.1:
    resolution: {integrity: sha512-3YffViIt2QWgTy6Pale5QpopX/IvU3LPL03jOTqp6pGj3VjesdO/U8CuHMKpnQr4shCNCM5fd5XFFvIIl6JBHg==}
    engines: {node: '>=8'}
    dependencies:
      '@types/minimist': 1.2.5
      camelcase-keys: 6.2.2
```

I assume that's why another version was added when you specified 9.3

PR #4255, opened 2024-07-31
split log newlines into array

lopert:

Fixes https://github.com/Shopify/shopify-functions/issues/280

Splits logs from a function_run into an array on newline for better readability.
I followed what we were doing during the log display of poll app logs
https://github.com/Shopify/cli/blob/4ec6dc095bb2174b8bf8203beb75967089108394/packages/app/src/cli/services/app-logs/dev/poll-app-logs.ts#L144-L147

### How to test your changes?

Run a function that logs, see the array in the outputfile. Should look something like:
```json 
    "logs": [
      "First Log!",
      "Log the second!",
      "1",
      "fourth line, length should be above!"
    ],
```

### Measuring impact

### Checklist

### DIFF
--- a/packages/app/src/cli/services/app-logs/dev/write-app-logs.test.ts
+++ b/packages/app/src/cli/services/app-logs/dev/write-app-logs.test.ts
@@ -56,11 +56,23 @@ describe('writeAppLogsToFile', () => {
     })
 
     // Then
+    const expectedSaveData = {
+      shopId: APP_LOG.shop_id,
+      apiClientId: APP_LOG.api_client_id,
+      payload: {
+        logs: ['Line 1!', ' Line2!'],
+      },
+      logType: APP_LOG.log_type,
+      cursor: APP_LOG.cursor,
+      status: APP_LOG.status,
+      source: APP_LOG.source,
+      sourceNamespace: APP_LOG.source_namespace,
+      logTimestamp: APP_LOG.log_timestamp,
+    }
+    const expectedLogData = JSON.stringify(expectedSaveData, null, 2)
+
     expect(returnedPath.fullOutputPath.startsWith(path)).toBe(true)
-    expect(writeLog).toHaveBeenCalledWith(
-      expect.stringContaining(path),
-      expectedLogDataFromAppEvent(APP_LOG, FUNCTION_RUN_PAYLOAD),
-    )
+    expect(writeLog).toHaveBeenCalledWith(expect.stringContaining(path), expectedLogData)
   })
 
   test('calls writeLog with strings when no matching payload type', async () => {
@@ -86,18 +98,14 @@ describe('writeAppLogsToFile', () => {
   })
 })
 
-function expectedLogDataFromAppEvent(event: AppLogData, payload: AppLogPayload | string): string {
-  const data = camelcaseKeys(
+function expectedLogDataFromAppEvent(event: AppLogData, payload: AppLogPayload | any): string {
+  const data: any = camelcaseKeys(
     {
       ...event,
-      payload: payload as any,
+      payload,
     },
     {deep: true},
   )
-
-  if (payload instanceof FunctionRunLog) {
-    data.payload.logs = payload.logs.split('\n').filter(Boolean)
-  }
 
   return JSON.stringify(data, null, 2)
 }


packages/app/src/cli/services/app-logs/dev/write-app-logs.ts
```diff
@@ -27,10 +28,16 @@ export const writeAppLogsToFile = async ({
   const fullOutputPath = joinPath(getLogsDir(), path)
 
   try {
+    const payload = JSON.parse(appLog.payload)
```
DuncanUszkay1, line 31: I see we're already parsing the payload [here](https://github.com/Shopify/cli/blob/204542c1938e3ce55808fce6f16ea87eca4e1563/packages/app/src/cli/services/app-logs/dev/poll-app-logs.ts#L77)- is there a clean way for us to reuse that?

packages/app/src/cli/services/app-logs/dev/write-app-logs.ts
```diff
@@ -12,10 +12,13 @@ interface AppLogFile {
 
 export const writeAppLogsToFile = async ({
   appLog,
+  appLogPayload,
   apiKey,
   stdout,
 }: {
   appLog: AppLogData
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  appLogPayload: AppLogPayload | any
```
lopert, line 21: I'm not super happy with this type, but I feel it kind of conveys what we're expecting here?
The payload is `JSON.parse`'d and if the `appLog.log_type` is one we know, we convert it appropriately before passing it to the write. If we don't know it, we just pass the JSON `object`.

packages/app/src/cli/services/app-logs/dev/write-app-logs.ts
```diff
@@ -30,11 +33,15 @@ export const writeAppLogsToFile = async ({
...
         ...appLog,
-        payload: JSON.parse(appLog.payload),
+        payload: appLogPayload,
       },
       {deep: true},
     )
 
+    if (appLogPayload instanceof FunctionRunLog) {
+      toSaveData.payload.logs = appLogPayload.logs.split('\n').filter(Boolean)
+    }
```
lopert, line 43: This is a bit annoying because of the typing.
The `FunctionRunLog`, as in the type we have for the data coming from the response, has the `logs` as a string.
But here, we want the logs split out for our write, so it needs to be a `string[]`.

However, because of the `payload: appLogPayload` on line 36, the `toSaveData.payload` is our generic AppLogPayload type, which doesn't have `payload.logs`, and even if it was our `FunctionRunLog` type, `string` wouldn't match `string[]`.

Right now it works because it is getting the `| any` from the method sig, but I could see a case for putting `payload: appLogPayload as any` on line 36 here.

DuncanUszkay1: Could we reverse the order here and mutate `appLogPayload` itself before creating `toSaveData`?

lopert: I dont think this changes much because:
- to look for logs to modify, we need `appLogPayload instanceof FunctionRunLog`
- but FunctionRunLog.logs is a `string`, so we get `Type 'string[]' is not assignable to type 'string'`, if we try to mutate the appLogPayload.logs to the split string.

I have it setup as is in order to preserve the `appLogPayload`, with the idea being that that object we're going to write (toSaveData) should be the one that can handle all the different ways of being written.

packages/app/src/cli/services/app-logs/dev/write-app-logs.test.ts
```diff
@@ -27,40 +41,63 @@ describe('writeAppLogsToFile', () => {
...
       ...event,
-      payload: JSON.parse(event.payload),
+      payload: payload as any,
     },
     {deep: true},
   )
+
+  if (payload instanceof FunctionRunLog) {
+    data.payload.logs = payload.logs.split('\n').filter(Boolean)
+  }
```
DuncanUszkay1, line 100: This doesn't test much given that it's just the same code as the module it's testing- should we test the full output for these tests instead now that the conversion from input to output is non-trivial?

packages/app/src/cli/services/app-logs/dev/write-app-logs.ts
DuncanUszkay1: That said, just putting `any` on the relevant field also seems fine. My opinion isn't super strong.

Both of those options are preferable to having the `any` in the type signature IMO.

lopert: I like putting `any` on the `saveData` object, that makes the most sense to me, since we're just creating this object to be written, aka not to be reused anywhere outside of this.

I think the `any` on the type signature is a separate thing though? As in, the type should remain `AppLogPayload | any` imo because of https://github.com/Shopify/cli/pull/4255#discussion_r1700481759

DuncanUszkay1 (approved):

PR #4256, opened 2024-08-01
Show domain for stores list (app management client only)

gracejychang:
Closes https://github.com/Shopify/develop-app-inner-loop/issues/1984

### How to test your changes?
Basically https://github.com/Shopify/develop-app-inner-loop/issues/1899#issuecomment-2249057719, but use this branch instead for step 1 + in the last step, you should see the domain of the stores instead of just the store names

### Measuring impact

### Checklist

### DIFF
--- a/packages/app/src/cli/services/dev/select-store.ts
+++ b/packages/app/src/cli/services/dev/select-store.ts
@@ -39,9 +39,11 @@ export async function selectStore(
   org: Organization,
   developerPlatformClient: DeveloperPlatformClient,
 ): Promise<OrganizationStore> {
+  const clientName = developerPlatformClient.clientName
+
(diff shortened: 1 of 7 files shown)


packages/app/src/cli/services/dev/select-store.ts
```diff
@@ -39,9 +39,11 @@ export async function selectStore(
   org: Organization,
   developerPlatformClient: DeveloperPlatformClient,
 ): Promise<OrganizationStore> {
+  const clientName = developerPlatformClient.clientName
```
gracejychang, line 42: should we use the available `clientName` field, or create a new one? I think `clientName` was created for the purposes of debugging.. then again, will we always have both clients?

shauns (approved):

PR #4261, opened 2024-08-02
Don't link on deploy when api key is provided

DuncanUszkay1:

Fixes failing buildkites: https://buildkite.com/shopify/functions-platform-tests/builds/7325#01910f81-3ad3-4460-8d40-c577a22e4f1f

In non-interactive environments, CLI users pass in an `api-key` or `client-id` flag to indicate which app they want to use. When this is the case, logic in the CLI needs to not try and prompt the user to select an app, or the process will fail.

This code path was always setting `enableLinkingPrompt` to true, when really it should have been dependent on whether or not the `api-key` flag was passed in. This resulted in the CLI always trying to link the app with the remote app.

I set the value of `enableLinkingPrompt` to equal `!options.apiKey`, similar to how the `dev` codepath handles this scenario: https://github.com/Shopify/cli/blob/c7b83d837be7670622f420feae0ce89ecd15781f/packages/app/src/cli/services/context.ts#L167

### How to test your changes?

- Clear your config. The standard in OSX appears to be `~/Library/Preferences/shopify-cli-app-nodejs/config.json` based on my own machine.
- Run deploy with an api-key non-interactively (My preferred trick is to just add `| cat` at the end of the command)
- The deploy will work with this change, but will fail with "you must select an app" afterwards

### Checklist

### DIFF
--- a/.changeset/metal-foxes-tan.md
+++ b/.changeset/metal-foxes-tan.md
@@ -1,0 +1,5 @@
+---
+'@shopify/app': patch
+---
+
+Disable app linking prompt on deploy when API key is provided


packages/app/src/cli/services/context.ts
```diff
@@ -430,7 +430,7 @@ export async function ensureDeployContext(options: DeployContextOptions): Promis
   const {reset, force, noRelease} = options
   let developerPlatformClient = options.developerPlatformClient
   // do the link here
-  const [remoteApp] = await fetchAppAndIdentifiers(options, developerPlatformClient, true, true)
+  const [remoteApp] = await fetchAppAndIdentifiers(options, developerPlatformClient, true, !options.apiKey)
```
DuncanUszkay1, line 433: Positional arguments here makes it hard to see, so here's a link to the API we're calling to demonstrate that the `enableLinkingPrompt` is what changed: https://github.com/Shopify/cli/blob/18fae61a9fc8a1866e0dc80d4219fd592ece170f/packages/app/src/cli/services/context.ts#L645

MitchDickinson, line 433: Is this logic something we only want to apply for deploy? For example, `release` also works the same way. `fetchAppAndIdentifiers ` has access to `options` so in theory we could do this logic inside of `fetchAppAndIdentifiers` if we wanted to apply this business logic to anyone who is trying to use this method.

No strong opinions...just asking...

MitchDickinson (approved):
The main goal with my PR is ensuring that, post-deploy, your TOML file is correctly configured and linked to the remote app. This requires a manual prompt usually, which is what broke the functions CI tests.

When providing the api key, it still feels like we should be able to link and update the TOML correctly. The prompting step is unnecessary because the user has provided the identifier. But the rest of the linking should be doable, right?

Doing the above would likely blow up the scope of this PR, so I'm ok shipping this to unblock functions, but we should at least queue up an issue to tackle the "proper" refactor later.

Approval conditional on getting that issue created (assuming folks agree) and tophatting working, but otherwise 🚢

MitchDickinson:
Tophatted and it worked 🎩 ✅

PR #4274, opened 2024-08-06
Fix running Javy and function-runner on Windows

jeffcharles:
Fixes bug on Windows where Javy and function-runner binaries are not executing because the file names do not end with `.exe`. Also adds installing and executing both binaries to the test suite.

Javy was not was not running on Windows because the executable did not end in `.exe`.

Adds `.exe` to the Javy and function-runner binary names on Windows.

### How to test your changes?

If you're on Windows, using cmd:

1. Create a Shopify app containing at least one JS Function extension if you do not already have one.
2. `pnpm shopify app build --path <path_to_app>` will perform the initial download of Javy.
3. Verify there are no errors reported.
4. `pnpm shopify app build --path <path_to_app>` will test that it uses the downloaded version of Javy.
5. Verify there are no errors reported.
6. `echo {} | pnpm shopify app function run --path <path_to_js_function_extension>` will test the initial download of function-runner.
7. Verify there are no errors reported.
8. `echo {} | pnpm shopify app function run --path <path_to_js_function_extension>` will it uses the downloaded version of function-runner.
9. Verify there are no errors reported.

### Measuring impact

### Checklist

### DIFF
--- a/packages/app/src/cli/services/function/binaries.ts
+++ b/packages/app/src/cli/services/function/binaries.ts
@@ -22,7 +22,8 @@ class DownloadableBinary {
   constructor(name: string, version: string, gitHubRepo: string) {
     this.name = name
     this.version = version
-    this.path = joinPath(dirname(fileURLToPath(import.meta.url)), '..', 'bin', `${this.name}`)
+    const filename = process.platform === 'win32' ? `${name}.exe` : name
+    this.path = joinPath(dirname(fileURLToPath(import.meta.url)), '..', 'bin', filename)
     this.gitHubRepo = gitHubRepo
   }
 
--- a/.changeset/smart-mugs-love.md
+++ b/.changeset/smart-mugs-love.md
@@ -1,0 +1,5 @@
+---
+'@shopify/app': patch
+---
+
+Fix Javy and function-runner downloads on Windows
--- a/packages/app/src/cli/services/function/binaries.test.ts
+++ b/packages/app/src/cli/services/function/binaries.test.ts
@@ -1,4 +1,5 @@
-import {javyBinary, functionRunnerBinary} from './binaries.js'
+import {javyBinary, functionRunnerBinary, installBinary} from './binaries.js'
+import {exec} from '@shopify/cli-kit/node/system'
 import {describe, expect, test} from 'vitest'
 
 const javy = javyBinary()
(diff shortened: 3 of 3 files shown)


jeffcharles:
Should I be including a changeset entry for this change? I didn't on the original change because it _should_ have been invisible to the end user.

packages/app/src/cli/services/function/binaries.ts
```diff
@@ -22,7 +22,8 @@ class DownloadableBinary {
   constructor(name: string, version: string, gitHubRepo: string) {
     this.name = name
     this.version = version
-    this.path = joinPath(dirname(fileURLToPath(import.meta.url)), '..', 'bin', `${this.name}`)
+    const filename = process.platform === 'win32' ? `${name}.exe` : name
```
amcaplan, line 25: I would suggest we leave open the possibility of fixing this in the release itself, so the filename itself gets the `.exe` suffix. So I'd suggest something like this:

```suggestion
    const needsExe = process.platform === 'win32' && !name.endsWith('.exe')
    const filename = needsExe ? `${name}.exe` : name
```

This would prevent a situation like `filename.exe.exe`

jeffcharles: Sorry I'm confused. _name_ is _always_ going to be set to one of `javy` or `function-runner` (there are only two callsites to the constructor, one [here](https://github.com/Shopify/cli/blob/09386a083ce8773e320d0dfb4522c80130a6b42b/packages/app/src/cli/services/function/binaries.ts#L82) and one [here](https://github.com/Shopify/cli/blob/09386a083ce8773e320d0dfb4522c80130a6b42b/packages/app/src/cli/services/function/binaries.ts#L89) and the _name_ argument is hardcoded in both) so `!name.endsWith('.exe')` will always evaluate to `true` which would simplify `needsExe` to `process.platform === 'win32'`.

isaacroldan (approved):

packages/app/src/cli/services/function/binaries.ts
amcaplan: My point is that we should consider changing the file in the release to be a gzipped `.exe` file instead of a gzipped binary with no extension. But looking at the broader code, we do seem to decide the output filename not based on the release filename but based on the local hardcoded value. So this can certainly stay as is.

PR #4282, opened 2024-08-07
Only reset errors if there are errors

andrewhassan:

Fixes https://github.com/Shopify/shopify-functions/issues/364

Because React re-renders when a state changes (by reference), every time we received a successful request, we were setting the errors array. We poll roughly twice a second, which meant that we are performing a ton of unnecessary re-renders, causing flickering.

### How to test your changes?

1. Run `app logs`
2. Trigger enough function runs so that your terminal has to scroll

With this change, there should be no flickering and it should be easy to copy your function run output.

### Measuring impact

### Checklist

### DIFF
--- a/packages/app/src/cli/services/app-logs/logs-command/ui/components/hooks/usePollAppLogs.ts
+++ b/packages/app/src/cli/services/app-logs/logs-command/ui/components/hooks/usePollAppLogs.ts
@@ -52,7 +52,7 @@ export function usePollAppLogs({initialJwt, filters, resubscribeCallback}: UsePo
         }
         retryIntervalMs = result.retryIntervalMs
       } else {
-        setErrors([])
+        setErrors((errors) => (errors.length ? [] : errors))
       }
 
       const {cursor: nextCursor, appLogs} = response as SuccessResponse
--- a/packages/app/src/cli/services/app-logs/logs-command/ui/components/hooks/usePollAppLogs.test.tsx
+++ b/packages/app/src/cli/services/app-logs/logs-command/ui/components/hooks/usePollAppLogs.test.tsx
@@ -374,14 +374,14 @@ describe('usePollAppLogs', () => {
     // needed to await the render
     await vi.advanceTimersByTimeAsync(0)
 
-    // Initial invocation, 429 returned
+    // Initial invocation, 422 returned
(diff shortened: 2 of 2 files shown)


DuncanUszkay1 (approved):

packages/app/src/cli/services/app-logs/logs-command/ui/components/hooks/usePollAppLogs.ts
```diff
@@ -52,7 +52,7 @@ export function usePollAppLogs({initialJwt, filters, resubscribeCallback}: UsePo
         }
         retryIntervalMs = result.retryIntervalMs
       } else {
-        setErrors([])
+        setErrors((errors) => (errors.length ? [] : errors))
```
lopert, line 55: I don't really get how this fixed the issue?
Before, we were always setting the state to `[]` which caused the refresh.
Now, we're either setting the state to `[]` or `errors`, either of which should cause the refresh again?

I did some searching of the [useState docs](https://react.dev/reference/react/useState#updating-state-based-on-the-previous-state), but it's not clear to me if providing a "updater" function also gets us a behaviour of "if the state hasn't changed, then don't update it".

lopert (approved):
🎩 worked for me, didn't see any flickering and was able to scroll and select

packages/app/src/cli/services/app-logs/logs-command/ui/components/hooks/usePollAppLogs.ts
andrewhassan: React re-renders if a state changes, and the way it determines if the state changed is by reference. When we `setState([])`, we're setting the state with a new array. Although the contents are the same, the reference is not, so React re-renders, causing the flickering.

To avoid this, we want to reset the errors only if there are already errors. Ideally, we'd do something like this:

```typescript
if (errors.length) {
  setErrors([])
}
```

But the problem with that is that `errors` inside this function actually refers to the initial `errors` array, not the current value! So we can't simply reset the errors based on the current `errors` value.

Luckily, React exposes the current value through the setter updater function. I use that current value to decide what we should do. If there are errors, then we set the state to `[]`, triggering a re-render. If there aren't errors, we set the state to the exact same reference, which does not trigger a re-render.

PR #4302, opened 2024-08-09
[Theme] New dev command

frandiox:
New `dev` command for the theme plugin. Part of https://github.com/Shopify/develop-advanced-edits/issues/197

Closes https://github.com/Shopify/develop-advanced-edits/issues/206
Closes https://github.com/Shopify/develop-advanced-edits/issues/207
Closes https://github.com/Shopify/develop-advanced-edits/issues/209

What's left:
- Update in-memory template when local files are uploaded to the cloud or removed from there. This work depends on https://github.com/Shopify/develop-advanced-edits/issues/281 -- I've implemented the watcher in ThemeFS but it still needs to interact with the API.


packages/theme/src/cli/utilities/theme-environment/assets.ts
```diff
@@ -0,0 +1,42 @@
+import {readFile} from '@shopify/cli-kit/node/fs'
+import {lookupMimeType} from '@shopify/cli-kit/node/mimes'
+import {defineEventHandler, serveStatic} from 'h3'
+import {joinPath} from '@shopify/cli-kit/node/path'
+import {stat} from 'fs/promises'
+
+/** Forces assets to be served from local domain */
+export function replaceLocalAssets(html: string) {
```
karreiro, line 8: In order to reduce the dependency on the CDN URL structure, I believe we may use the `DevServerContext.localThemeFileSystem` to assist us.

Let's say we have `assets/base.css`, `assets/reset.css`, `assets/panel.js`, and `assets/panel.css.liquid` in the theme, and let's say that we built a regular expression that captures the URLs of all script and link tags on the page (this regex really just captures all URLs used in link and script tags, regardless of the domain). Then we get this:

```
[
  "https://store.myshopify.com/cdn/shop/t/28/assets/base.css?v=149927614001461873041712915543",
  "https://store.myshopify.com/cdn/shop/t/28/assets/reset.css?v=149927614001461873041712915232",
  "https://store.myshopify.com/cdn/shop/t/28/assets/panel.js?v=149927614001461873041712915231",
  "https://store.myshopify.com/cdn/shop/t/28/assets/panel.css?v=149927614001461873041712910984",
  "https://store.myshopify.com/cdn/shop/t/28/compiled_assets/scripts.js?7538",
  ...
]
```

Then, by using the JS/CSS files (`assets/base.css`, `assets/reset.css`, `assets/panel.js`) from the local file system, we may infer which URLs must be replaced:

```
"https://store.myshopify.com/cdn/shop/t/28/assets/base.css?v=149927614001461873041712915543",
"https://store.myshopify.com/cdn/shop/t/28/assets/reset.css?v=149927614001461873041712915232",
"https://store.myshopify.com/cdn/shop/t/28/assets/panel.js?v=149927614001461873041712915231",
```

Finally, we replace them with:
```
"/assets/base.css?v=149927614001461873041712915543",
"/assets/reset.css?v=149927614001461873041712915232",
"/assets/panel.js?v=149927614001461873041712915231",
```

This process is worthwhile because:
- We need to avoid replacing all URLs (as `https://store.myshopify.com/cdn/shop/t/28/assets/panel.css?v=149927614001461873041712910984`/`assets/panel.css.liquid` cannot be served locally and must be rendered/served remotely).
- We avoid coupling our regular expression with Shopify CDN structure that might change. So, if it changes to something like `https://store.myshopify.com/shopify_cdn/shop/t/28/assets/base.css`, this module will still work.

packages/theme/src/cli/utilities/theme-environment/hmr.ts
```diff
@@ -0,0 +1,123 @@
...
+
+      const data = JSON.parse(event.data) as HmrEvent
+      if (data.type === 'section') {
+        const id = data.content.match(/id="([^"]+)"/)?.[1]
+        if (id) {
+          const existingElement = document.getElementById(id)
+          if (existingElement) {
+            existingElement.outerHTML = data.content
+            // eslint-disable-next-line no-console
+            console.info('[HMR] Updated section:', data.key)
```
karreiro, line 111: The same section file may impact multiple DOM elements, as we may notice here:

Screenshot 2024-08-12 at 09 49 10

The `__SHOPIFY_CLI_ENV__` global variable, that supports the process above, is injected/created here:
https://github.com/Shopify/cli/blob/142797ebc1a260936762d4cddcc88102d6c72404/packages/cli-kit/assets/cli-ruby/lib/shopify_cli/theme/dev_server/hot_reload/script_injector.rb#L39-L48

(that variable helps the client side on the mapping of the expected DOM elements)

karreiro (commented):
Thank you, @frandiox! The PR is getting in excellent shape 🔥🚀

packages/theme/src/cli/utilities/theme-environment/html.ts
```diff
@@ -0,0 +1,67 @@
...
+    console.log(`${method} ${urlPath}`)
+
+    const response = await render(ctx.session, {
+      path: urlPath,
+      query: [],
+      themeId: String(theme.id),
+      cookies: headers.get('cookie') || '',
+      sectionId: '',
+      headers: getProxyStorefrontHeaders(event),
+      replaceTemplates: getInMemoryTemplates(),
```
karreiro, line 23: I've been investigating why the render function is a bit slow; in some scenarios, it takes up to 10 seconds. It appears that the primary reason is that we're passing all JSON files here. The secondary reason is the lack of memoization in the renderer's authentication, which also impacts performance, but only by around 500ms.

I'm currently working on the memoization issue, so we don't need to address that here.

Regarding the passing of all JSON files, the call [here](https://github.com/Shopify/cli/blob/0610fc4e3da7fc3ecb19c6097285d85afc9e739e/packages/theme/src/cli/utilities/theme-environment/theme-environment.ts#L13) is synchronous, so in theory, we wouldn't need to pass all those files since everything is already uploaded.

However, we do want to pass some templates; we just need to ensure we're only passing those that haven't been uploaded yet.

frandiox: Ah great catch, will fix this! I was reading the JSON files early to have an in-memory version of the existing sections. I can just move that to a different object 👍

packages/theme/src/cli/utilities/theme-fs.ts
```diff
@@ -46,28 +48,82 @@ const THEME_PARTITION_REGEX = {
...
 
 export async function mountThemeFileSystem(root: string): Promise<ThemeFileSystem> {
-  const filesPaths = await glob(THEME_DIRECTORY_PATTERNS, {
+  const files = new Map<string, ThemeAsset>()
+  const eventEmitter = new EventEmitter()
+  function emitEvent<T extends ThemeFSEventName>(eventName: T, payload: ThemeFSEvent<T>['payload']) {
+    eventEmitter.emit(eventName, payload)
+  }
+
+  const initialFilesPromise = glob(THEME_DIRECTORY_PATTERNS, {
```
frandiox, line 57: Note this is now async so that we can do other things at the same time. Before checking the contents of files, we need to await `ready()` to ensure initial files are all included.

packages/theme/package.json
```diff
@@ -44,6 +45,8 @@
     "@shopify/cli-kit": "3.65.0",
     "@shopify/theme-check-node": "2.8.0",
     "@shopify/theme-language-server-node": "1.11.4",
+    "chokidar": "3.5.3",
+    "h3": "1.12.0",
```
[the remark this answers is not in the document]
frandiox: I talked with Isaac about the versioning and it's better to start using v1. We should move the `app` to v1 as well as soon as we can.
v0 has no docs and types are rather wrong.

packages/theme/src/cli/utilities/theme-environment/hot-reload/server.ts
```diff
@@ -0,0 +1,186 @@
...
+  const handleFileUpdate = ({fileKey, contentPromise, syncPromise}: ThemeFSEvent<'add'>['payload']) => {
+    const extension = extname(fileKey)
+    const needsTemplateUpdate = ['.liquid', '.json'].includes(extension)
+    const isAsset = fileKey.startsWith('assets/')
+
+    if (needsTemplateUpdate && !isAsset) {
+      contentPromise
+        .then((content) => {
+          inMemoryTemplates[fileKey] = content
+          if (extension === '.json') parsedJsonTemplates[fileKey] = JSON.parse(content)
```
karreiro, line 52: We need to use `parseJSON` (from `'@shopify/theme-check-node'`) here instead of `JSON.parse` because `templates/*` support JSON with comments and trailing commas.

packages/cli-kit/src/public/node/themes/types.ts
```diff
@@ -39,6 +65,19 @@ export interface ThemeFileSystem {
...
    */
   read: (assetKey: string) => Promise<string | Buffer | undefined>
+
+  /**
+   * Gets the stats of a file from the local disk and updates the themeFileSystem
+   * Returns undefined if the file does not exist
+   *
+   * @param assetKey - The key of the file to read
+   */
+  stat: (assetKey: string) => Promise<Pick<Stats, 'mtime' | 'size'> | undefined>
```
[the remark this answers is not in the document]
frandiox: I was reusing the same name as in other functions (e.g. `read` or `delete`) but I'll change all of them to `fileKey` 👍

packages/theme/src/cli/utilities/theme-environment/hot-reload/server.ts
frandiox: Fixed in https://github.com/Shopify/cli/pull/4302/commits/dddf9a19fcc30264b2534982112d95d256d47e2f

packages/theme/src/cli/utilities/theme-environment/proxy.ts
```diff
@@ -0,0 +1,185 @@
...
+  // Safari adds upgrade-insecure-requests to CSP and it needs to be removed:
+  clearResponseHeaders(event, HOP_BY_HOP_HEADERS)
+
+  // Link header preloads resources from global CDN, proxy it:
+  const linkHeader = response.headers.get('Link')
+  if (linkHeader) setResponseHeader(event, 'Link', injectCdnProxy(linkHeader, ctx))
+
+  // Cookies are set for the vanity domain, fix it for localhost:
+  const setCookieHeader =
+    'raw' in response.headers ? response.headers.raw()['set-cookie'] : response.headers.getSetCookie()
```
frandiox, line 131: Note: H3 uses native `fetch` and `Response`, while cli-kit's fetch utilities are based on `node-fetch`. This means that the result of H3's `sendProxy` is a Response, whereas the result of our `render` function is a NodeResponse. The main difference is in the way we access `set-cookie` headers. They also differs in `response.body` stream implementation but we don't use that here (`response.text()` work in both).

packages/theme/src/cli/utilities/theme-fs.test.ts
```diff
@@ -157,12 +170,17 @@ describe('theme-fs', () => {
...
+      expect(file).toEqual({
         key: 'templates/404.json',
         checksum: 'f14a0bd594f4fee47b13fc09543098ff',
+        value: expect.any(String),
+        attachment: '',
       })
 
       // When
+      delete file?.value
       const content = await themeFileSystem.read(key)
```
frandiox, line 184: I had to update these tests because we now store `value` and `attachment` earlier (for free). I'm deleting the value to keep testing that `read` adds it back.

packages/theme/src/cli/utilities/theme-environment/hot-reload/server.ts
```diff
@@ -0,0 +1,232 @@
...
+import {extname} from '@shopify/cli-kit/node/path'
+import {parseJSON} from '@shopify/theme-check-node'
+import EventEmitter from 'node:events'
+import type {Theme, ThemeFSEventPayload} from '@shopify/cli-kit/node/themes/types'
+import type {DevServerContext} from '../types.js'
+
+// --- Template Replacers ---
+
+/** Store which files are currently only updated in-memory, not in remote */
+const inMemoryTemplateFiles = new Set<string>()
```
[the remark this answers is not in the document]
karreiro: Exactly, that captures the item 5 in the list below. Our final state with the `inMemoryTemplateFiles` is:

1. The CLI stores everything in the `inMemoryTemplateFiles`
2. The CLI displays links so users can preview their theme in the browser
3. The CLI uploads files and removes them from `inMemoryTemplateFiles` as each file is uploaded, helping to avoid performance impact on the server
4. The `inMemoryTemplateFiles` becomes empty when uploads are finished
5. The `inMemoryTemplateFiles` is updated when users apply changes to files

Currently, the CLI is blocking users and uploading everything in advance, but we may iterate on this when we get the `localThemeFileSystem.addEventListener` implemented 🚀

packages/theme/src/cli/utilities/theme-environment/proxy.ts
```diff
@@ -0,0 +1,185 @@
...
+ * It also replaces MainCDN-like (cdn.shopify.com/...) URLs to files that are known local assets.
+ * Other MainCDN matches are left unmodified.
+ */
+export function injectCdnProxy(originalContent: string, ctx: DevServerContext) {
+  let content = originalContent
+
+  // -- Redirect all usages to the vanity CDN to the local server:
+  const vanityCdnPath = '/cdn/'
+  const vanityCdnRE = new RegExp(`(https?:)?//${getStoreFqdnForRegEx(ctx)}${vanityCdnPath}`, 'g')
+  content = content.replace(vanityCdnRE, vanityCdnPath)
```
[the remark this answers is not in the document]
frandiox: The `readThemeFile` function returns Buffer for images, and that's what was triggering this error. I've fixed that, but I'm now also adding a way to bypass local images if they include filters 👍

packages/theme/src/cli/utilities/theme-environment/hot-reload/server.ts
```diff
@@ -0,0 +1,231 @@
...
+              if (content) replaceTemplates[fileKey] = content
+              continue
+            }
+          }
+        }
+      }
+
+      return render(ctx.session, {
+        path: '/',
+        query: [],
```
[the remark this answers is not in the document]
frandiox: I've fixed passing querystirngs but the issue seems to be something else. It even happens when there are no querystring at all, like in `/search` before searching anything.

Any other idea of why this could be happening? 🤔

packages/theme/src/cli/utilities/theme-environment/hot-reload/server.ts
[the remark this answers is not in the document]
karreiro: Absolutely! I missed capturing that in the message above. Certainly, some elements of the theme block the first rendering and should be uploaded in advance. I've updated the message to reflect that 🚀 Thank you, @frandiox!

PR #4304, opened 2024-08-09
Preparation work for consistent-dev

isaacroldan:
Some utility changes to prepare for consistent-dev and break the changes in multiple PRs to redude their size.

- Use the handle for the assets folder name in the manifest.
- Use the handle as the default identifier for the bundle output of extensions (when there isn't a local identifier)
- Add the store FQDN to the renderDev config object (will be used later)
- Normalize the store FQDN when using it in the dev session API
- Add a new function to generate a temporary directory.
- For some app-management-client queries, return a mock and show a warning instead of crashing.

### How to test your changes?
Nothing relevant to test here, all changes will be used in a future PR

### Measuring impact

### Checklist

### DIFF
--- a/packages/app/src/cli/services/deploy/bundle.ts
+++ b/packages/app/src/cli/services/deploy/bundle.ts
@@ -12,7 +12,7 @@ import {Writable} from 'stream'
 interface BundleOptions {
   app: AppInterface
   bundlePath?: string
-  identifiers: Identifiers
+  identifiers?: Identifiers
 }
 
 export async function bundleAndBuildExtensions(options: BundleOptions, systemEnvironment = process.env) {


packages/app/src/cli/models/app/app.ts
```diff
@@ -350,7 +350,7 @@ export class App<
           type: module.externalType,
           handle: module.handle,
           uid: module.uid,
-          assets: module.uid,
+          assets: module.handle,
```
gonzaloriestra, line 353: This change doesn't affect the current behavior in Partners?

gonzaloriestra (approved):

packages/app/src/cli/models/app/app.ts
isaacroldan: no, this is the manifest, is not used anywhere

PR #4310, opened 2024-08-11
Upgrade reminders

amcaplan:

To encourage faster CLI upgrades and reduce the number of support cases where users are using old CLI versions.

1. Checks for a current CLI version once daily in the background of a running command, caching the result
2. When a cached updated CLI version is found, warns the user to upgrade on each command run

Since this system exists, we can remove the check + warning to upgrade from the `version` and `app info` commands.

### How to test your changes?

1. Use the CLI normally a few times. Nothing should happen.
2. Then, change [this line](https://github.com/Shopify/cli/blob/142797ebc1a260936762d4cddcc88102d6c72404/packages/cli-kit/src/public/common/version.ts#L1) to have a lower cli-kit version.
3. On the next command, nothing should happen (though if it's a really fast command it might hang for a second at the end).
4. On the command after that, you should see a warning right away that you need to upgrade to the latest version.
5. OPTIONAL: Clear the cache by opening `~/Library/Preferences/shopify-cli-kit-nodejs/config.json`, looking for the `cache` section, and deleting the `"npm-package-@shopify/cli"` value. When you run the CLI again, the warning disappears for the first run (the run which populates the cache). In the next run, you'll see the warning again.

### Measuring impact

### Checklist

### DIFF
--- a/.changeset/dirty-berries-play.md
+++ b/.changeset/dirty-berries-play.md
@@ -1,0 +1,6 @@
+---
+'@shopify/cli-kit': minor
+---
+
+- Add cached package version checks
+- Make CLI upgrade message smarter
--- a/.changeset/silver-files-burn.md
+++ b/.changeset/silver-files-burn.md
@@ -1,0 +1,5 @@
+---
+'@shopify/cli': minor
+---
+
+Display daily upgrade reminder when a new version is available


packages/cli-kit/src/public/node/node-package-manager.test.ts
```diff
@@ -574,6 +624,7 @@ describe('checkForNewVersion', () => {
 
     // When
     await checkForNewVersion(dependency, currentVersion)
+    vi.setSystemTime(vi.getRealSystemTime() + 1000 * 3600 * 1000 - 10)
```
gonzaloriestra, line 627: You should add this when faking timers:

```
  afterEach(() => {
    vi.useRealTimers()
  })
```

packages/cli-kit/src/public/node/hooks/prerun.ts
```diff
@@ -81,3 +86,18 @@ function findAlias(aliases: string[]) {
...
+  const currentVersion = CLI_KIT_VERSION
+
+  // Check in the background, once daily
+  void checkForNewVersion(cliDependency, currentVersion, {cacheExpiryInHours: 24})
+
+  // Warn if we previously found a new version
+  const lastVersion = checkForCachedNewVersion(cliDependency, currentVersion)
+  if (lastVersion) {
+    const packageManager = packageManagerFromUserAgent()
+    outputInfo(getOutputUpdateCLIReminder(packageManager, lastVersion))
```
gonzaloriestra, line 101: I think `newVersion` makes more sense here:

```suggestion
  const newVersion = checkForCachedNewVersion(cliDependency, currentVersion)
  if (newVersion) {
    const packageManager = packageManagerFromUserAgent()
    outputInfo(getOutputUpdateCLIReminder(packageManager, newVersion))
```

packages/cli-kit/src/public/node/hooks/prerun.ts
```diff
@@ -81,3 +86,18 @@ function findAlias(aliases: string[]) {
     return existingAlias.replace(/:/g, ' ')
   }
 }
+
+function warnOnAvailableUpgrade() {
+  const cliDependency = '@shopify/cli'
+  const currentVersion = CLI_KIT_VERSION
```
gonzaloriestra, line 92: What about skipping this for nightlies and snapshots? It's probably not useful in those cases and the version comparison won't probably work as expected.

packages/cli-kit/src/private/node/conf-store.ts
```diff
@@ -89,3 +91,21 @@ export async function cacheRetrieveOrRepopulate(
...
+export function cacheRetrieve(
+  key: keyof Cache,
+  config = cliKitStore(),
+): CacheValueForKey<typeof key> | undefined {
+  const cache: Cache = config.get('cache') || {}
+  const cached = cache[key]
+  return cached?.value
+}
+
+export function cacheClear(config = cliKitStore()): void {
```
gonzaloriestra, line 109: Are we using this?

packages/cli-kit/src/public/node/hooks/prerun.ts
```diff
@@ -81,3 +86,18 @@ function findAlias(aliases: string[]) {
...
+
+  // Check in the background, once daily
+  void checkForNewVersion(cliDependency, currentVersion, {cacheExpiryInHours: 24})
+
+  // Warn if we previously found a new version
+  const lastVersion = checkForCachedNewVersion(cliDependency, currentVersion)
```
gonzaloriestra, line 98: Maybe I'm missing something, but can't we get the new version already from the previous line? Do we really need that new method to only retrieve (without repopulate)?

```suggestion
  const lastVersion = checkForNewVersion(cliDependency, currentVersion, {cacheExpiryInHours: 24})
```

packages/cli-kit/src/private/node/conf-store.ts
amcaplan: Yes, it is used for tests to reset cache state.

It's not used in `conf-store.test.ts` itself because we can always inject a conf store. But for higher-level tests, I think it makes sense to use the real thing and reset the cache between tests.

packages/cli/src/cli/services/commands/version.ts
```diff
@@ -1,15 +1,6 @@
...
-  const cliDependency = '@shopify/cli'
-  const currentVersion = CLI_KIT_VERSION
-  outputInfo(outputContent`Current Shopify CLI version: ${outputToken.yellow(currentVersion)}`.value)
-  // Check a max of once daily
-  const lastVersion = await checkForNewVersion(cliDependency, currentVersion, {cacheExpiryInHours: 24})
-  if (lastVersion) {
-    const packageManager = packageManagerFromUserAgent()
-    outputInfo(getOutputUpdateCLIReminder(packageManager, lastVersion))
-  }
+  outputInfo(outputContent`Current Shopify CLI version: ${outputToken.yellow(CLI_KIT_VERSION)}`.value)
```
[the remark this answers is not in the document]
amcaplan: I made some further adjustments, so the output string isn't colored in any way. That should make it work smoothly in scripts that might call `shopify version` and do something with that information.

I also updated the warning to use stderr instead of stdout, so that commands intended for piping elsewhere won't have the warning output getting in their way.

Screenshot 2024-08-12 at 13 44 14

packages/cli-kit/src/public/node/hooks/prerun.ts
gonzaloriestra: TIL! What happens if the command finishes before that unawaited promise resolves? Will it wait for it? Have you tested it?

packages/cli-kit/src/public/node/hooks/prerun.ts
gonzaloriestra: What about adding a comment? It might be confusing if you don't know that nightlies start with `0.0.0`

packages/cli-kit/src/public/node/hooks/prerun.ts
amcaplan: It will await the unresolved promise before exiting. This is just a single request to npmjs, and of course ctrl+c should kill it, so in nearly all cases this won't be noticeable.

jessehoek:
Looks good to me on the UX side. Thanks Ariel!

gonzaloriestra (approved):
Thanks for removing the extra step, it looks nice!

PR #4314, opened 2024-08-12
Dunk.source

DuncanUszkay1:

Fixes: https://github.com/Shopify/shopify-functions/issues/371

As per the issue, sources weren't using namespaces, didn't accept comma separated lists, and didn't validate against existing sources.

I've updated the command to split the source flag into a `sources` list, validated them against existing functions extensions (the only supported extension type ATM), and added an additional error when an app has no extensions that will receive logs. Also added in the namespace everywhere.

### How to test your changes?

Tophat:
- Create an app with no functions. Try to run app logs- it should fail with an error message
- Add a function. Try to run `shopify app logs --source=junk`, it should fail with a message indicating what the valid sources are
- Run `shopify app logs --source=extensions.your-function-name`, replacing the source with the function name
- Ensure that logs are received when the function runs
- Add an additional function, but don't configure it to run
- Run `shopify app logs --source=extensions.your-second-function-name`, replacing the source with the new function's name
- Trigger the function, ensure that the logs are filtered out and do not appear
- 
### Checklist

### DIFF
--- a/packages/app/src/cli/services/logs.test.ts
+++ b/packages/app/src/cli/services/logs.test.ts
@@ -4,8 +4,15 @@ import {ensureDevContext} from './context.js'
 import * as renderLogs from './app-logs/logs-command/ui.js'
 import * as renderJsonLogs from './app-logs/logs-command/render-json-logs.js'
 import {loadAppConfiguration} from '../models/app/loader.js'
-import {buildVersionedAppSchema, testApp, testOrganizationApp} from '../models/app/app.test-data.js'
+import {
+  buildVersionedAppSchema,
+  testApp,
+  testOrganizationApp,
+  testFunctionExtension,
+  defaultFunctionConfiguration,
+} from '../models/app/app.test-data.js'
 import {consoleLog} from '@shopify/cli-kit/node/output'
+import {AbortError} from '@shopify/cli-kit/node/error'
 import {describe, test, vi, expect} from 'vitest'
 
 vi.mock('../models/app/loader.js')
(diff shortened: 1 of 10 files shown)


packages/app/src/cli/services/logs.test.ts
```diff
@@ -57,9 +57,55 @@ describe('logs', () => {
...
+        apiKey: 'api-key',
+        directory: 'directory',
+        storeFqdn: 'store-fqdn',
+        sources: ['extensions.source'],
+        status: 'status',
+        configName: 'config-name',
+        userProvidedConfigName: 'user-provided-config-name',
+      })
+    }).rejects.toThrowError(
+      'This app has no log sources. Learn more about app logs at https://shopify.dev/docs/apps/build/functions/log-streaming-and-replay.',
```
DuncanUszkay1, line 80: cc @nickwesselman this is raised if `app logs` is called in a scenario where we know that a log will never be returned. Thoughts on copy here and where to link the user to explain why logs won't return anything?

nickwesselman: Why don't we link them to the reference docs for the `app logs` command and describe the behavior there? The URL will be

https://shopify.dev/docs/api/shopify-cli/app/app-logs

DuncanUszkay1: Sure, sounds good.

saga-dasgupta (approved):
I was able to verify, thanks for the tophatting instructions!

Screenshot 2024-08-12 at 2 46 33 PM
Screenshot 2024-08-12 at 2 46 45 PM

lopert (approved):
🎩 worked great for me! Also tested passing in a comma separated list 👍

PR #4315, opened 2024-08-12
Add root test element to ensure proper word wrapping

DuncanUszkay1:

Fixes https://github.com/Shopify/shopify-functions/issues/292

Text was being wrapped columnwise in the metadata line of the log, rather than as a unit. 

E.G. it was doing:

```
this is my fu your command ran in 
nction name 9 million instructions
```

Instead of:
```
this is my function name your com
mand ran in 9 million instructions
```

Following the pattern in [this PR](https://github.com/Shopify/cli/pull/3959/files), I added a root text element to instruct ink to wrap these as a single unit.

### How to test your changes?

- Shrink your window, increase your function handle length, or increase text size until the metadata log line does not fit in one line in the terminal
- Run `shopify app logs` and run the function
- The metadata line should wrap as a row, e.g.
```
2024-08-12T18:31:16.747589Z cv Success export "run" executed in 0.4034M
instructions
```

### Checklist


lopert (approved):
🎩 looks good to me!
Maybe a silly question but it seems weird to me to have text within text like this. Would this also work with another `Box` around it instead?
... I just tried and it didn't work, it just had the same behaviour as before. /shrug

PR #4322, opened 2024-08-13
Don't pretty print --json output for logs

DuncanUszkay1:

In a prior PR, we started pretty printing JSON for readability, which unintentionally broke the JSONL format interface that we needed for programatic users. This PR fixes that by only pretty printing when the output is a TTY.

### How to test your changes?

Run the command in TTY, it should pretty print:
```
shopify app logs --json
```

Run the command without TTY, it should return JSONL:
```
shopify app logs --json > tmp.txt
```

### Checklist


saga-dasgupta (approved):
LGTM👍🏻

DuncanUszkay1:
Spoke with Nick, he was of the opinion that it should always output JSONL, even in a TTY situation, since they can always pretty print by piping it into `jq`. Added a commit which uses that tactic.

DuncanUszkay1:
re-tophatted to validate that TTY views are also JSONL

PR #4326, opened 2024-08-14
Add query cache

gonzaloriestra:

Fixes https://github.com/Shopify/develop-app-inner-loop/issues/2021

- Adds a cache mechanism during the life of the command to run any function
- Enables cache for the specifications query for App Management API
- Re-uses the logic for the existing TOML config caching

The idea is to enable the cache for more queries later, including Partners, once we ensure it works as expected. The enabled one should be safe, as we don't expect the specifications to change during a command run.

We were fetching the specifications at least 2 times in several commands, especially when a new app is used. Now, it only makes a call per command.

### How to test your changes?

`USE_APP_MANAGEMENT_API=1 p shopify app deploy`

### Measuring impact

### Checklist

### DIFF
--- a/packages/cli-kit/src/public/node/command-cache.ts
+++ b/packages/cli-kit/src/public/node/command-cache.ts
@@ -1,6 +1,7 @@
 import {isUnitTest} from './context/local.js'
 import {LocalStorage} from './local-storage.js'
 import {outputDebug} from './output.js'
+import {tmpdir} from 'os'
 
 interface CommandLocalStorage {
   [key: string]: {[key: string]: unknown}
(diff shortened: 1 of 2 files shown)


packages/cli-kit/src/public/node/command-cache.ts
```diff
@@ -0,0 +1,111 @@
...
+ * @param fn - Function to run if the cache key is not found, and cache the result of.
+ * @param options - Options for the cache store.
+ * @returns The result of the function.
+ */
+export async function runWithCommandCache<T>(
+  cacheKey: string,
+  fn: () => T,
+  options: CommandCacheOptions = {},
+): Promise<T> {
+  // let queries = {} as {[key: string]: unknown}
```
alexanderMontague, line 95: 🗑️

alexanderMontague (commented):
This is really neat stuff, I think this could be super useful in cutting down unnecessary duplicate requests. The ultimate end goal would be to clean up those redundant requests so you don't even need the caching, but I think this is a good first step. Few suggestions/comments, but I'm generally on board with the approach!

Issue #4320, opened 2024-08-13
[Bug]: The `--theme` flag should not select themes based on parts of the name.


karreiro:
### Please confirm that you have:

- [X] Searched [existing issues](.) to see if your issue is a duplicate. (If you’ve found a duplicate issue, feel free to add additional information in a comment on it.)
- [X] Reproduced the issue in [the latest CLI version](https://www.npmjs.com/package/@shopify/cli).

### In which of these areas are you experiencing a problem?

Theme

### Expected behavior

Only the exact theme name should be used for selection.

### Actual behavior

Parts of the theme name are being used for theme selection with the `--theme` flag

### Verbose output

```shell
...
```

### Reproduction steps

1.
2.
3..

### Operating System

.

### Shopify CLI version (check your project's `package.json` if you're not sure)

.

### Shell

.

### Node version (run `node -v` if you're not sure)

.

### What language and version are you using in your application?

.

PR #4331, opened 2024-08-15
Update the theme selector to no longer match partial theme names by d…

karreiro:

Fixes https://github.com/Shopify/cli/issues/4320

This PR updates the theme selector code to avoid implicitly considering names that contain "each." While this behavior was desired for read-only scenarios (`shopify theme list`), it can lead to destructive behavior in other scenarios (`shopify theme push` updating the wrong theme).

### How to test your changes?

- Create themes with names that match.
- Notice the `shopify theme push -t <name>` only selects themes by using the exact name.
- Notice the `shopify theme list --name <partial>*` matches names using the `*`.

### Post-release steps

Update the [`shopify theme list`](https://shopify.dev/docs/api/shopify-cli/theme/theme-list) documentation to clarify the `*` pattern.

### Measuring impact

### Checklist


packages/theme/src/cli/utilities/theme-selector/filter.ts
```diff
@@ -27,7 +27,15 @@ function filterByTheme(store: string, themes: Theme[], filter: Filter) {
     const error = `The ${store} store doesn't have a theme with the "${identifier}" ID or name`
```
jamesmengo, line 28: One thing I noticed during tophat was that if we input a pattern with no matches, the err message is as follows - I think we can use this opportunity to update the message since we are now accepting patterns
`The quickstart-b88db4f0.myshopify.com store doesn't have a theme with the "f*" ID or name`       

Something along these lines is what I'm thinking, though we could add more specificity if you'd like

```suggestion
    const error = `No themes on the store ${store} match the identifier "${identifier}"`
```

jamesmengo (approved):
Behaviour lgtm!
I think we could updat the error message - non blocking.

PR #4332, opened 2024-08-15
Functions_367: Hiding irrelevant flags from app logs help command display

saga-dasgupta:
fixes https://github.com/Shopify/shopify-functions/issues/367 

We display a lot of flags for the help command for `app logs` some of which are not relevant to the `app logs` command.

Hiding flags not relevant to app logs command from being displayed to user when they run `app logs --help`. 

Screenshot 2024-08-16 at 10 36 10 AM

### How to test your changes?

Checkout my branch and run `pnpm run shopify app logs --help`

### Measuring impact

### Checklist


packages/cli/oclif.manifest.json
```diff
@@ -1567,7 +1567,7 @@
           "description": "Resource URL for checkout UI extension. Format: \"/cart/{productVariantID}:{productQuantity}\"",
           "env": "SHOPIFY_FLAG_CHECKOUT_CART_URL",
           "hasDynamicHelp": false,
-          "hidden": false,
+          "hidden": true,
```
saga-dasgupta, line 1570: I could have removed these flags from this file entirely too instead of setting `hidden` to `true` but I figured no harm in keeping them if they ever become relevant to display in future. Let me know if folks who have more context than me think otherwise.

DuncanUszkay1: I want to avoid the scenario where someone copies over this command from dev and expects it to do something, only for it to quietly do nothing- I think full removal is necessary

packages/cli/oclif.manifest.json
```diff
@@ -1652,13 +1652,14 @@
           "allowNo": false,
           "description": "Skips the Partners Dashboard URL update step.",
           "env": "SHOPIFY_FLAG_NO_UPDATE",
-          "hidden": false,
+          "hidden": true,
```
DuncanUszkay1, line 1655: This is an auto-generated file- do these changes persist through a `pnpm refresh-manifests`?

packages/app/src/cli/commands/app/logs.ts
```diff
@@ -21,8 +21,22 @@ export default class Logs extends Command {
 
   static description = this.descriptionWithoutMarkdown()
 
-  static flags = {
-    ...Dev.flags,
+  static flags: any = {
```
saga-dasgupta, line 24: Had to typecast this to `any` because the filter function I have added to filter out the undesired flags out of `dev.flags` makes it not see the `client-id` or `api-key` keys and throw a type error later in the code where we try to obtain these keys from `flags`.

packages/app/src/cli/commands/app/logs.ts
```diff
@@ -21,8 +21,22 @@ export default class Logs extends Command {
...
+            'theme',
+            'checkout-cart-url',
+            'no-update',
+            'notify',
+            'skip-dependencies-installation',
+            'subscription-product-url',
+            'theme-app-extension-port',
+            'tunnel-url',
+          ].includes(k),
+      ),
```
saga-dasgupta, line 38: I'd have like to use variables instead of this giant list inside of a hash but doing so also adds those variables to the manifest file which isn't something I want.

packages/app/src/cli/commands/app/logs.ts
```diff
@@ -51,10 +65,10 @@ export default class Logs extends Command {
 
     const sources = flags.source?.split(',')
 
-    await checkFolderIsValidApp(flags.path)
+    await checkFolderIsValidApp(flags.path!)
```
saga-dasgupta, line 68: Because of the `any` typecasting I added above, it thinks this can be either `string` or `undefined` so had to add the `!`.

packages/app/src/cli/commands/app/logs.ts
```diff
@@ -22,7 +24,16 @@ export default class Logs extends Command {
   static description = this.descriptionWithoutMarkdown()
 
   static flags = {
-    ...Dev.flags,
+    ...globalFlags,
+    ...appFlags,
+    'api-key': Dev.flags['api-key'],
+    'client-id': Dev.flags['client-id'],
+    store: Dev.flags.store,
+    reset: Dev.flags.reset,
```
DuncanUszkay1, line 32: Not an issue, but I'm curious why these flags are accessed differently 🤔

DuncanUszkay1 (commented):
That description also looks suspect given that it references functions rather than 'log sources'.

cc @nickwesselman thoughts on what the copy should be here?

My take:
```
Opens a real-time stream of detailed app logs from the selected app and store.
  Use the \`--source\` argument to limit output to a particular log source, such as a specific Shopify Function handle. Use the `shopify app logs sources` command to view a list of sources.
  Use the \`--status\` argument to specify the type of status to retrieve, either \`success\` or \`failure\`.
  \`\`\`
  shopify app logs --status=success --source=extension.discount-function
  \`\`\`

nickwesselman (approved):

PR #4347, opened 2024-08-19
Allow multiple shops to be selected in app logs command

DuncanUszkay1:

Fixes #0000 

### Measuring impact

### Checklist


packages/app/src/cli/services/context.ts
```diff
@@ -147,6 +147,169 @@ export async function ensureGenerateContext(options: GenerateContextOptions): Pr
...
+ * If there is an app from flags, we check if it's valid. If it's not, throw an error.
+ * If there is info in the cache or current configuration, check if it is still valid and return it.
+ * If there is no info (or is invalid):
+ *  - Show prompts to select an org and app
+ *  - The info will be updated in the cache or current configuration
+ *
+ * @param options - Current dev context options
+ * @returns The selected org, app and dev store
+ */
+export async function ensureLogsContext(options: LogsContextOptions): Promise<LogsContextOutput> {
```
DuncanUszkay1, line 183: There's a lot of duplicated logic between this function and `ensureDevContext`. I want to spend some time looking into some refactors to keep the two implementations more synchronized over time while still allowing logs to diverge in how it accepts shops

packages/app/src/cli/services/app-logs/logs-command/render-json-logs.ts
```diff
@@ -46,7 +48,19 @@ export async function renderJsonLogs({
 
   if (appLogs) {
     appLogs.forEach((log) => {
-      outputInfo(toFormattedAppLogJson(log, parseAppLogPayload(log.payload, log.log_type), false))
+      const storeName = storeNameById.get(log.shop_id.toString())
+      if (storeName === undefined) {
+        return
+      }
```
DuncanUszkay1, line 54: In theory, this should not occur since core should not be returning logs that belong to different shops, but that fact is not expressed by the type system. In the event that it does happen due to some unpredicted bug, I've opted to just hide that log, but open to other ideas.

packages/app/src/cli/services/logs.ts
```diff
@@ -61,28 +64,35 @@ export async function logs(commandOptions: LogsOptions) {
...
       options: {
         variables,
         developerPlatformClient: logsConfig.developerPlatformClient,
       },
       pollOptions,
+      storeNameById: logsConfig.storeNameById,
     })
   } else {
+    if (multipleStores) {
+      consoleLog(`Subscribing to additional stores: ${commandOptions.storeFqdns?.slice(1).join(', ')}\n`)
```
DuncanUszkay1, line 79: The first shop will display as part of the `info` box rendered by `ensureDevContext` in a TTY, so we only need to call out the additional stores.

Not sure if we want this at verbose level instead.

packages/app/src/cli/services/logs.ts
```diff
@@ -91,17 +101,34 @@ async function prepareForLogs(commandOptions: LogsOptions): Promise<{
...
-  const devContextOptions: DevContextOptions = {...commandOptions, developerPlatformClient}
-  const {storeId, remoteApp, localApp} = await ensureDevContext(devContextOptions)
+  const developerPlatformClient = selectDeveloperPlatformClient({configuration})
+  const primaryStoreFqdn = commandOptions.storeFqdns?.[0]
+  const devContextOptions: DevContextOptions = {
+    ...commandOptions,
+    storeFqdn: primaryStoreFqdn,
+    developerPlatformClient,
+  }
+  const {storeId, storeFqdn, remoteApp, localApp} = await ensureDevContext(devContextOptions)
```
DuncanUszkay1, line 111: We leverage `ensureDevContext` to setup the first shop, app, and org with the exact same process that dev uses. All following shops will be handled separately as additions.

packages/app/src/cli/services/logs.ts
```diff
@@ -91,17 +101,34 @@ async function prepareForLogs(commandOptions: LogsOptions): Promise<{
...
+  storeNameById.set(storeId, storeFqdn)
+  if (commandOptions.storeFqdns && commandOptions.storeFqdns.length > 1) {
+    await Promise.all(
+      commandOptions.storeFqdns?.slice(1).map((storeFqdn) => {
+        return storeFromFqdn(storeFqdn, remoteApp.organizationId, developerPlatformClient).then((store) => {
+          storeNameById.set(store.shopId, storeFqdn)
+        })
+      }),
+    )
+  }
```
DuncanUszkay1, line 123: If there are additional shops specified, we load them all in parallel

packages/app/src/cli/services/logs.ts
```diff
@@ -91,17 +92,34 @@ async function prepareForLogs(commandOptions: LogsOptions): Promise<{
...
+  const developerPlatformClient = selectDeveloperPlatformClient({configuration})
+  const primaryStoreFqdn = commandOptions.storeFqdns?.[0]
+  const devContextOptions: DevContextOptions = {
+    ...commandOptions,
+    storeFqdn: primaryStoreFqdn,
+    developerPlatformClient,
+  }
+  const {storeId, storeFqdn, remoteApp, localApp} = await ensureDevContext(devContextOptions)
 
-  developerPlatformClient = remoteApp.developerPlatformClient ?? developerPlatformClient
```
DuncanUszkay1, line 98: I inlined this action into the return value below- this allows me to use `developerPlatformClient` in a `Promise.all` block, since it's now constant.

saga-dasgupta (approved):
Code LGTM and I was able to 🎩 by creating 2 dev stores and triggering the function in both stores, first with just one dev store url in the `--store` flag and it only printed logs for that store and then with both store urls in the `--store` flag and it printed logs for both stores.

packages/app/src/cli/services/logs.ts
```diff
@@ -14,7 +14,7 @@ interface LogsOptions {
   directory: string
   reset: boolean
   apiKey?: string
-  storeFqdn?: string
+  storeFqdns?: string[] | undefined
```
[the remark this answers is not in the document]
DuncanUszkay1: Yes, good catch 👀

PR #4348, opened 2024-08-19
Add stdout check to isTTY

lopert:

https://github.com/Shopify/shopify-functions/issues/363

When piping the output of a commands such as `shopify app dev` or `shopify app logs`, we can sometimes encounter a react infinite render error. This happens when a prompt is supposed to be displayed to the user, but since we've piped the output, it's no longer interactable.

The react error doesn't really explain what's going on, and doesn't leverage the existing error we have in our [throwInNonTTY](https://github.com/Shopify/cli/blob/6c4ac20c3f430e496de7becd9898242c2768222e/packages/cli-kit/src/public/node/ui.tsx#L713-L718).

Modifies the existing `throwInNonTTY` and `isTTY` functions to also verify that `stdout` is also TTY.

### How to test your changes?

1. Cause the `ensureDevContext` check to trigger by editing 
  `/Users/<user>/Library/Preferences/shopify-cli-app-nodejs/config.json` 
  and remove the entry for your app path
2. Run `shopify app logs | more`

You should see an error explaining that things need to be run in interactive mode

```
│  Failed to prompt:                                                           │
│  Which organization is this work for?                                        │
│  This usually happens when running a command non-interactively, for example  │
│   in a CI environment, or when piping to or from another process.            │
│  To resolve this, specify the option in the command, or run the command in   │
│  an interactive environment such as your local terminal.                     │
```
### Measuring impact

### Checklist

### DIFF
--- a/packages/cli-kit/src/public/node/ui.tsx
+++ b/packages/cli-kit/src/public/node/ui.tsx
@@ -722,7 +722,7 @@ function throwInNonTTY(
 
 ${outputContent`${outputToken.cyan(promptText)}`.value}
 
-This usually happens when running a command non-interactively, for example in a CI environment, or when piping input from another process.`
+This usually happens when running a command non-interactively, for example in a CI environment, or when piping to or from another process.`
   throw new AbortError(
     errorMessage,
     'To resolve this, specify the option in the command, or run the command in an interactive environment such as your local terminal.',
--- a/packages/cli-kit/src/public/node/ui.tsx
+++ b/packages/cli-kit/src/public/node/ui.tsx
@@ -296,7 +296,7 @@ export async function renderSelectPrompt<T>(
   {renderOptions, isConfirmationPrompt, ...props}: RenderSelectPromptOptions<T>,
   uiDebugOptions: UIDebugOptions = defaultUIDebugOptions,
 ): Promise<T> {
-  throwInNonTTY({message: props.message, stdin: renderOptions?.stdin}, uiDebugOptions)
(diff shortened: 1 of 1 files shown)


saga-dasgupta (approved):
I was able to 🎩 this 👍🏻

packages/cli-kit/src/public/node/ui.tsx
```diff
@@ -689,28 +689,40 @@ export const keypress = async (stdin = process.stdin, uiDebugOptions: UIDebugOpt
...
+export function isTTY({
+  stdin = undefined,
+  stdout = undefined,
+  uiDebugOptions = defaultUIDebugOptions,
+}: IsTTYOptions = {}) {
+  return (
+    Boolean(uiDebugOptions.skipTTYCheck) ||
+    (Boolean(stdin || terminalSupportsRawMode()) && Boolean(stdout || process.stdout.isTTY))
+  )
 }
```
amcaplan, line 705: I disagree with the direction here.

Looking over the code generally, we use `terminalSupportsRawMode` as a test of TTY-like behavior. So it makes sense to fix `terminalSupportsRawMode` to look at both stdin and stdout, only returning `true` if `stdin` and `stdout` are both TTY.

And for this function, I'd expect something like:

```suggestion
export function isTTY({
  stdin = undefined,
  stdout = undefined,
  uiDebugOptions = defaultUIDebugOptions,
}: IsTTYOptions = {}) {
  return (
    Boolean(
      uiDebugOptions.skipTTYCheck ||
      (stdin && stdout) ||
      terminalSupportsRawMode(stdin ?? process.stdin, stdout ?? process.stdout)
  )
}
```

lopert:
Closed in favour of https://github.com/Shopify/cli/pull/4348

PR #4352, opened 2024-08-20
Fix `render` issue to render themes consistently

karreiro:
### Why are these changes introduced?

The `render` API could use two cookie strings (`Cookie` and `cookie`), depending on the cookies and headers used in the `DevServerRenderContext`. This was causing erratic behavior during rendering, and sometimes the wrong theme was being rendered.

### What is this pull request doing?

This PR cleans up the `render` API and prevents that ambiguity in the rendering.

### DIFF
--- a/packages/theme/src/cli/utilities/theme-environment/storefront-renderer.ts
+++ b/packages/theme/src/cli/utilities/theme-environment/storefront-renderer.ts
@@ -138,8 +138,7 @@ function themeAccessHeaders(session: DevServerSession) {
 }
 
 function cleanHeader(headers: {[key: string]: string}): {[key: string]: string} {
-  // Avoid ambiguity in case the headers object contains the
-  // 'Cookie' and 'cookie' keys.
+  // Force the use of the 'Cookie' key if consumers also provide the 'cookie' key
   delete headers.cookie
   delete headers.authorization
   return headers


packages/theme/src/cli/utilities/theme-environment/storefront-renderer.ts
```diff
@@ -136,3 +136,11 @@ function themeAccessHeaders(session: DevServerSession) {
...
   }
 }
+
+function cleanHeader(headers: {[key: string]: string}): {[key: string]: string} {
+  // Avoid ambiguity in case the headers object contains the
+  // 'Cookie' and 'cookie' keys.
+  delete headers.cookie
+  delete headers.authorization
+  return headers
+}
```
frandiox, line 146: Should we start using `new Headers(...)` to avoid problems with case sensitivity? Or would that be harder to use?

frandiox (approved):

packages/theme/src/cli/utilities/theme-environment/storefront-renderer.ts
karreiro: Thank you for the review, @frandiox! :D

I've considered the `Headers` approach (which would indeed be a bit more elegant). However, this would expose our logic to a proxy that might set both cookies, as we can see in the example below:

```
  const header = new Headers({
    Cookie: 'value1',
    cookie: 'value2',
  });

console.log('>>', header.get('cookie'));
# => value1; value2
```

I've updated the comment to clarify that we want to force the usage of the `Cookie` key and ignore the `cookie` key.

PR #4353, opened 2024-08-20
add stdout check to terminalSupportsRawMode

lopert:
https://github.com/Shopify/shopify-functions/issues/363
Alternative approach to https://github.com/Shopify/cli/pull/4348

### DIFF
--- a/packages/cli-kit/src/public/node/system.ts
+++ b/packages/cli-kit/src/public/node/system.ts
@@ -5,7 +5,6 @@ import {treeKill} from './tree-kill.js'
 import {isTruthy} from './context/utilities.js'
 import {shouldDisplayColors, outputDebug} from '../../public/node/output.js'
 import {execa, ExecaChildProcess} from 'execa'
-import {ReadStream} from 'tty'
 import type {Writable, Readable} from 'stream'
 
 export interface ExecOptions {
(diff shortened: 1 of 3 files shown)


packages/cli-kit/src/public/node/system.ts
```diff
@@ -131,17 +130,13 @@ export async function sleep(seconds: number): Promise<void> {
 }
 
 /**
- * In case an standard input stream is passed check if it supports raw mode. Otherwise default standard input stream
- * will be used.
+ * Check if the standard input and output streams supports raw mode.
```
amcaplan, line 133:
```suggestion
 * Check if the standard input and output streams support prompting.
```

packages/cli-kit/src/public/node/system.ts
```diff
@@ -131,17 +130,13 @@ export async function sleep(seconds: number): Promise<void> {
...
  *
- * @param stdin - The standard input stream to check.
- * @param env - Environmemnt variables.
- * @returns True in the selected input stream support raw mode.
+ * @returns True if the standard input and output streams support raw mode.
```
amcaplan, line 135:
```suggestion
 * @returns True if the standard input and output streams support prompting.
```

packages/cli-kit/src/public/node/system.ts
```diff
@@ -131,17 +130,13 @@ export async function sleep(seconds: number): Promise<void> {
...
  */
-export function terminalSupportsRawMode(stdin?: ReadStream, env = process.env): boolean {
-  if (isTruthy(env.CI)) {
+export function terminalSupportsRawMode(): boolean {
```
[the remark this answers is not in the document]
lopert: Renamed it 👍

amcaplan (approved):
I've requested some small comment updates, otherwise I think we're good to go!

PR #4356, opened 2024-08-20
Jm/dev

jamesmengo:

Fixes #0000 

### Measuring impact

### Checklist

### DIFF
--- a/packages/cli-kit/src/public/node/themes/types.ts
+++ b/packages/cli-kit/src/public/node/themes/types.ts
@@ -1,3 +1,4 @@
+import {AdminSession} from '../session.js'
 import type {Stats} from 'fs'
 
 /**
@@ -82,6 +83,15 @@ export interface ThemeFileSystem {
   addEventListener: {
     <T extends ThemeFSEventName>(eventName: T, cb: (params: ThemeFSEventPayload<T>) => void): void
   }
+
+  /**
+   * Starts a file watcher for the theme directory.
+   *
+   * @param themeId - The ID of the theme being watched.
+   * @param adminSession - The admin session for API communication.
+   * @returns A Promise that resolves to an FSWatcher instance.
+   */
+  startWatcher: (themeId: string, adminSession: AdminSession) => Promise<void>
 }
 
 /**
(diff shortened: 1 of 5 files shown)


packages/cli-kit/src/public/node/themes/types.ts
```diff
@@ -82,6 +83,15 @@ export interface ThemeFileSystem {
...
   }
+
+  /**
+   * Starts a file watcher for the theme directory.
+   *
+   * @param themeId - The ID of the theme being watched.
+   * @param adminSession - The admin session for API communication.
+   * @returns A Promise that resolves to an FSWatcher instance.
+   */
+  startWatcher: (themeId: string, adminSession: AdminSession) => Promise<void>
```
jamesmengo, line 94: This allows us to provide the `themeId` and `adminSession` without modifying `mountThemeFileSystem`
- I wanted to separate this logic since we are using the local theme file system in commands like `push` where we don't need watching behaviour
- The downside of this approach is that you need to explicitly call `startWatcher` - simply adding a listener is not enough

Another alternative is to include  `themeId` and `adminSession` as optional args in `mountThemeFileSystem`

```
export async function mountThemeFileSystem(root: string, themeId?: string, adminSession?: AdminSession): Promise<ThemeFileSystem> {
....existing code

if (themeId && adminSesssion) {
const watcher = chokidar.watch([...directoriesToWatch], {
        ignored: THEME_DEFAULT_IGNORE_PATTERNS,
        persistent: !process.env.SHOPIFY_UNIT_TEST,
        ignoreInitial: true,
      })

      watcher
        .on('add', handleFileUpdate.bind(null, 'add', themeId, adminSession))
        .on('change', handleFileUpdate.bind(null, 'change', themeId, adminSession))
        .on('unlink', handleFileDelete.bind(null, themeId, adminSession))
}
```

I've considered adding `themeId` and `adminSession` as args in `addEventListener` but that felt smelly - I wouldn't expect to need to provide those arguments each time I call `addEventListener`

frandiox (commented):
Thanks! I found a couple of issues that we should probably fix before merging.

Also, I think we should merge this to `main` branch directly instead? Is there any conflict with that?

karreiro (approved):
Thank you, @jamesmengo! The PR looks great! Once the comments left by @frandiox are addressed, we will be ready to merge :)

PR #4362, opened 2024-08-21
[Theme] Speed up theme synchronization

frandiox:
I think this speeds up the dev command startup by a big margin:

- Make password check async
- Make checksum async while
- Upload file batches in parallel (relying on API throttler) instead of 1 by 1
- Delete files in parallel (relying on API throttler) instead of 1 by 1
- File deletion is made async so that the server can start before finishing

---

- I think this fixes a bug in file batching because we were not using the base64 size of certain files.
- Prevents re-reading files from disk because we can reuse the content that was read beforehand to calculate checksums.

---

- Still need to defer certain upload files. That might come in a different PR.
- To review, it might be easier commit by commit.

### DIFF
--- a/.changeset/light-poems-fail.md
+++ b/.changeset/light-poems-fail.md
@@ -1,0 +1,6 @@
+---
+'@shopify/theme': patch
+'@shopify/app': patch
+---
+
+Bump Shopify/theme-tools packages
--- a/.changeset/rare-bats-marry.md
+++ b/.changeset/rare-bats-marry.md
@@ -1,0 +1,5 @@
+---
+'@shopify/app': patch
+---
+
+Add proper user-agent header for pollAppLogs
--- a/packages/app/package.json
+++ b/packages/app/package.json
@@ -57,7 +57,7 @@
     "@shopify/plugin-cloudflare": "3.66.0",
     "@shopify/polaris": "12.10.0",
     "@shopify/polaris-icons": "8.0.0",
-    "@shopify/theme-check-node": "2.8.0",
+    "@shopify/theme-check-node": "2.9.0",
     "body-parser": "1.20.2",
     "camelcase-keys": "9.1.3",
     "chokidar": "3.5.3",
--- a/packages/app/src/cli/api/graphql/partners/generated/function-upload-url-generate.ts
+++ b/packages/app/src/cli/api/graphql/partners/generated/function-upload-url-generate.ts
@@ -1,0 +1,53 @@
+/* eslint-disable @typescript-eslint/consistent-type-definitions */
+import * as Types from './types.js'
+
+import {TypedDocumentNode as DocumentNode} from '@graphql-typed-document-node/core'
+
+export type FunctionUploadUrlGenerateMutationVariables = Types.Exact<{[key: string]: never}>
+
+export type FunctionUploadUrlGenerateMutation = {
+  functionUploadUrlGenerate?: {
+    generatedUrlDetails?: {url: string; moduleId: string; headers: unknown; maxBytes: number; maxSize: string} | null
+  } | null
+}
+
+export const FunctionUploadUrlGenerate = {
+  kind: 'Document',
+  definitions: [
+    {
+      kind: 'OperationDefinition',
+      operation: 'mutation',
+      name: {kind: 'Name', value: 'functionUploadUrlGenerate'},
+      selectionSet: {
+        kind: 'SelectionSet',
+        selections: [
+          {
+            kind: 'Field',
+            name: {kind: 'Name', value: 'functionUploadUrlGenerate'},
+            selectionSet: {
+              kind: 'SelectionSet',
+              selections: [
+                {
+                  kind: 'Field',
+                  name: {kind: 'Name', value: 'generatedUrlDetails'},
+                  selectionSet: {
+                    kind: 'SelectionSet',
+                    selections: [
+                      {kind: 'Field', name: {kind: 'Name', value: 'url'}},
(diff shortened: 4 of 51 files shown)


packages/theme/src/cli/services/dev.ts
```diff
@@ -99,7 +96,12 @@ export async function dev(options: DevOptions) {
     },
   }
 
-  const server = await setupDevServer(options.theme, ctx)
+  const {server, renderProgress} = await setupDevServer(options.theme, ctx)
+
+  const storefrontPassword = await storefrontPasswordPromise
+  session.storefrontPassword = storefrontPassword
+
+  await renderProgress()
```
jamesmengo, line 104: 👍🏻 great idea!

How do you feel about a rename to clarify that this is unrelated to the `renderLinks` call below?

frandiox: Renamed to `renderThemeSyncProgress` 👍

jamesmengo (approved):
🎩 LGTM + code and works as expected 👍🏻

frandiox:
@jamesmengo I've made further changes that speed up the startup a lot. Please give it another check 🙏

packages/theme/src/cli/utilities/theme-uploader.ts
```diff
@@ -32,82 +33,171 @@ export async function uploadTheme(
...
+      title: titleGetter(),
+      task: async () => {
+        const result = await Promise.race([
+          promise,
+          new Promise((resolve) => setTimeout(() => resolve('timeout'), timeout)),
+        ])
+
+        if (result === 'timeout') {
+          addNextCheck()
+        }
```
frandiox, line 97: This adds a new task when the timeout ends until `promise` is resolved. This way we update the title with a progress %

packages/theme/src/cli/utilities/theme-uploader.ts
```diff
@@ -117,100 +207,26 @@ async function selectUploadableFiles(
...
-    }
-  })
-  return {
-    tasks,
-    updatedFileCount: runningFileCount,
-  }
-}
-
-async function createBatches(files: Checksum[], path: string): Promise<FileBatch[]> {
-  const fileSizes = await Promise.all(files.map((file) => fileSize(`${path}/${file.key}`)))
```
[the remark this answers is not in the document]
jamesmengo: Thank you for cleaning this up!
This has been an area of the codebase I was not proud of and had been hoping to revisit for a while so I'm happy to see these improvements :)

packages/theme/src/cli/services/push.ts
```diff
@@ -34,13 +34,21 @@ export async function push(theme: Theme, session: AdminSession, options: PushOpt
   const themeChecksums = await fetchChecksums(theme.id, session)
   const themeFileSystem = await mountThemeFileSystem(options.path)
 
-  const results = await uploadTheme(theme, session, themeChecksums, themeFileSystem, options)
+  const {uploadResults, renderThemeSyncProgress} = await uploadTheme(
```
jamesmengo, line 37: This method is hanging / not terminating when I 🎩 btw!

jamesmengo: I think [these changes](https://github.com/Shopify/cli/pull/4367/files#diff-d857e2b583dcee2197fc538eb4d35aabb5ce83cee18ebb970ece7854dda8ba07) will resolve that issue for us

jamesmengo (approved):
Left a comment [here](https://github.com/Shopify/cli/pull/4362/files#r1727760845) to make sure we're on the same page, but no changes needed on this particular changeset as the PR will resolve with [this](https://github.com/Shopify/cli/pull/4356/files)

frandiox:
I've force-pushed to remove the last 2 commits and move them to this follow-up PR: https://github.com/Shopify/cli/pull/4370

I think this can be merged as is and will continue in that other PR after also mergin James' changes into main branch.

packages/theme/src/cli/utilities/theme-uploader.ts
```diff
@@ -32,82 +33,183 @@ export async function uploadTheme(
...
+
+      return createBatches(fileType).map((batch) => {
+        return uploadBatch(batch, themeFileSystem, session, theme.id, uploadResults).then(() => {
+          progress.current += batch.length
+        })
+      })
+    }),
+  ).then(() => {
+    progress.current = progress.total
+  })
```
karreiro, line 200: When we flat map the files, I wonder if we're exposing the logic of the order of files, as requests for Liquid sections and JSON templates are triggered simultaneously.

The files with dependencies we have are:

1. Liquid sections need to be uploaded first so they may be referenced by templates.
2. JSON sections need to be uploaded afterward so they can reference Liquid sections.
3. JSON templates should follow so they can reference sections.
4. Contextualized templates can be uploaded after templates.
5. Config files must be the last ones, but we need to upload config/settings_schema.json first, followed by config/settings_data.json.

The files with no dependencies we have are:

- The other Liquid files (for example, snippets).
- The other JSON files (for example, locales).
- The static assets.

We do have the `handleFailedUploads` function that retries 2 times if something goes wrong. However, given the variability in theme sizes (some themes have very large Liquid files, others have large JSON files), I'm concerned that if we don't guarantee the dependency order, the retry mechanism might not be enough to ensure successful uploads.

karreiro (commented):
Thank you for this PR, @frandiox! This is awesome 🔥 It's 5x faster in my tests.

I've left one concern that might make the upload process a bit slower, but it would still be faster than the version we have on `main`. Still, I believe it's worth handling that concern to ensure we keep the uploader stable. Please let me know your thoughts about that :)

Thanks again for the amazing work!

PR #4367, opened 2024-08-22
[Themes][Feature] Dev Server - Handle theme editor sync conflicts between local and remote themes

jamesmengo:
Closes
- https://github.com/Shopify/develop-advanced-edits/issues/292
- https://github.com/Shopify/develop-advanced-edits/issues/293

With the introduction of https://github.com/Shopify/cli/pull/4356, the `theme-editor-sync` flag needs to be updated to accommodate the new upload functionality

**Note: there is a known issue where `settings_schema.json` is always detected as a mismatching file. Fixing that is not within the scope of this PR.**

#### 1) **Fixes a bug where a conflict gets detected (change on both remote and local) when the `remoteTheme watcher` pulls remote theme changes into the local FS**
  - Shares a single `ThemeFileSystem` instance between the `themeFileSystem watcher` (local FS)  and `remoteTheme watcher`, allowing them to work with a unified version of `files`

#### 2) **Prevents `themeFileSystem watcher`  from re-uploading recently downloaded files from the `remoteThemeWatcher`**
  - When the `remoteThemeWatcher` downloads a file, it updates `localThemeFileSystem.files` with the new value **and checksum** via the `write` function [here](https://github.com/Shopify/cli/blob/ab47b80789e2b08a8a824587df3a3d091d849442/packages/theme/src/cli/utilities/theme-environment/theme-polling.ts#L96)
  - This triggers the `change` or `add` handler in the `themeFileSystem watcher`
  - If the checksum is the same before and after [reading](https://github.com/Shopify/cli/blob/ab47b80789e2b08a8a824587df3a3d091d849442/packages/theme/src/cli/utilities/theme-fs.ts#L99), this means that the checksum was already updated in the `write` call above - indicating that this change event came from a remote change rather than a local FS change. In this case, we skip uploading the file

### How to test your changes?
1) `shopify-dev theme dev --dev-preview --theme-editor-sync`
2) You may be prompted about some file conflicts. Select `keep the remote version`
3) Once the server is up, click on `Customize your theme at the theme editor`
4) Apply some changes to the JSON files in your remote theme
5) Validate that these changes are downloaded locally
6) Validate through console output / logs that we DO NOT upload the changes immediately after downloading

Bonus:
- You can apply some changes to your local theme and validate that we don't pull these changes immediately or raise conflicts that both local and remote contain updates - [example error here](https://github.com/Shopify/develop-advanced-edits/issues/292)

https://github.com/user-attachments/assets/e542133b-ed48-4bec-8db4-7ce2ed1b26a9

### Measuring impact

### Checklist

### DIFF
--- a/packages/theme/src/cli/utilities/theme-environment/remote-theme-watcher.ts
+++ b/packages/theme/src/cli/utilities/theme-environment/remote-theme-watcher.ts
@@ -1,9 +1,9 @@
 import {pollThemeEditorChanges} from './theme-polling.js'
 import {reconcileJsonFiles} from './theme-reconciliation.js'
-import {outputDebug} from '@shopify/cli-kit/node/output'
 import {AdminSession} from '@shopify/cli-kit/node/session'
 import {Checksum, Theme, ThemeFileSystem} from '@shopify/cli-kit/node/themes/types'
 import {fetchChecksums} from '@shopify/cli-kit/node/themes/api'
+import {outputDebug} from '@shopify/cli-kit/node/output'
 
 export const LOCAL_STRATEGY = 'local'
 export const REMOTE_STRATEGY = 'remote'
@@ -24,6 +24,7 @@ export async function reconcileAndPollThemeEditorChanges(
   },
 ) {
   outputDebug('Initiating theme asset reconciliation process')
+  await localThemeFileSystem.ready()
 
   if (remoteChecksums.length !== 0) {
     await reconcileJsonFiles(targetTheme, session, remoteChecksums, localThemeFileSystem, options)
--- a/packages/theme/src/cli/utilities/theme-environment/remote-theme-watcher.ts
+++ b/packages/theme/src/cli/utilities/theme-environment/remote-theme-watcher.ts
@@ -1,6 +1,5 @@
 import {pollThemeEditorChanges} from './theme-polling.js'
 import {reconcileJsonFiles} from './theme-reconciliation.js'
-import {mountThemeFileSystem} from '../theme-fs.js'
 import {outputDebug} from '@shopify/cli-kit/node/output'
 import {AdminSession} from '@shopify/cli-kit/node/session'
 import {Checksum, Theme, ThemeFileSystem} from '@shopify/cli-kit/node/themes/types'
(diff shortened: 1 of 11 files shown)


frandiox (commented):
LGTM! Leaving the approval to Guilherme after that comment is addressed 👍 

Also, in case it's useful, I'm [adding](https://github.com/Shopify/cli/pull/4370/files) a new `unsyncedFileKeys: Set<Key>` to the themeFS object that contains all the currently syncing files.

packages/theme/src/cli/utilities/theme-environment/remote-theme-watcher.ts
```diff
@@ -29,8 +28,7 @@ export async function reconcileAndPollThemeEditorChanges(
 
   const updatedRemoteChecksums = await fetchChecksums(targetTheme.id, session)
 
-  const themeFileSystem = mountThemeFileSystem(localThemeFileSystem.root)
```
jamesmengo, line 32: this is a cleanup which unifies the ThemeFileSystem we're using

packages/theme/src/cli/utilities/theme-environment/remote-theme-watcher.ts
```diff
@@ -25,12 +24,12 @@ export async function reconcileAndPollThemeEditorChanges(
   },
 ) {
   outputDebug('Initiating theme asset reconciliation process')
+  await localThemeFileSystem.ready()
```
jamesmengo, line 27: this ensures that our themeFileSystem is fully mounted before we started initiating reconciliation

packages/theme/src/cli/utilities/theme-environment/theme-environment.ts
```diff
@@ -50,17 +49,6 @@ function ensureThemeEnvironmentSetup(theme: Theme, ctx: DevServerContext) {
   return {
     workPromise: uploadPromise.then((result) => result.workPromise),
     renderProgress: async () => {
-      if (ctx.options.themeEditorSync) {
```
jamesmengo, line 53: This causes the process to hang because`reconcilePromise` will not resolve.

I'm removing this for now, and if we want to add a progress bar back later for the reconciliation step, we can.

frandiox: If that's the case, then shouldn't we change `const uploadPromise = reconcilePromise.then` as well??

Can we make the promise to resolve and keep doing work in the background instead?

karreiro (commented):
Thank you for this PR, @jamesmengo! Great stuff! The conflicts are being handled as expected, but I do notice a [edge case](https://github.com/Shopify/cli/pull/4392#pullrequestreview-2271632906) involving the initialization that we may handle in the other PR.

karreiro (approved):

PR #4370, opened 2024-08-23
[Theme] Defer files to upload before server start

frandiox:
Follow up of #4362 

TL;DR The previous PR brought down the dev server startup time from 1-2min on new themes 10-20 seconds. This PR defers all the syncing work to the background and starts the dev server much faster, in about 0-5 seconds in total (including the time required by other CLI setup / auth)

- Bubble up "in memory templates" state to the themeFS under `unsyncedFileKeys` so that this information can be used everywhere.
- Uploading and deleting files is deferred and doesn't block server startup anymore
- If you open the server before the theme background sync is finished, we use `replaceTemplates` in the render API to send the local files before the cloud gets them from the Asset API.
- If there's no theme files at all in SFR, this uploads first 3 minimum files required to create the `_shopify_essential` cookie. This does not delay server startup, but might delay (~0.5s) the very first rendering request if you open the browser really fast after the server links are shown in terminal.
- Filter out locale files from the first rendering request if they are unsynced, since only 1 is needed.
- Fix status returned for 404 pages and similar.
- Fix proxying non-GET requests and returning redirects.

### DIFF
--- a/packages/theme/src/cli/utilities/theme-environment/proxy.test.ts
+++ b/packages/theme/src/cli/utilities/theme-environment/proxy.test.ts
@@ -138,7 +138,11 @@ describe('dev proxy', () => {
       )
 
       expect(event.node.res.getHeader('set-cookie')).toMatchInlineSnapshot(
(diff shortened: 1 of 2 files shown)


karreiro (approved):
Thank you, @frandiox! 🔥 🚀 Amazing work with this optimization!

During my tests I noticed two minor issues, but we can address them in following PRs:
- https://github.com/Shopify/develop-advanced-edits/issues/303
- https://github.com/Shopify/develop-advanced-edits/issues/302

Thanks again for the amazing work!

PR #4375, opened 2024-08-26
Use the theme proxy on `app dev` to support theme app extensions development

karreiro:
Work in progress.. 🚧

### DIFF
--- a/packages/theme/src/cli/utilities/theme-environment/hot-reload/client.ts
+++ b/packages/theme/src/cli/utilities/theme-environment/hot-reload/client.ts
@@ -16,6 +16,10 @@ export type HotReloadEvent =
       key: string
     }
   | {
+      type: 'ext-css'
+      key: string
+    }
+  | {
       type: 'full'
       key: string
     }
@@ -33,9 +37,14 @@ function injectFunction(fn: () => void) {
  * Therefore, do not use any imports or references to external variables here.
  */
 function hotReloadScript() {
+  let serverPid: string | undefined
+
   const prefix = '[HotReload]'
+  const evtSource = new EventSource('/__hot-reload/subscribe', {withCredentials: true})
+
   // eslint-disable-next-line no-console
   const logInfo = console.info.bind(console, prefix)
+
   // eslint-disable-next-line no-console
   const logError = console.error.bind(console, prefix)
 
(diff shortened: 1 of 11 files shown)


packages/theme/src/cli/utilities/theme-environment/storefront-renderer.ts
```diff
@@ -102,8 +103,11 @@ function buildStorefrontUrl(session: DevServerSession, {path, sectionId, query}:
     params.append(key, value)
   }
 
+  // The Section Rendering API takes precendence over the Block Rendering API.
   if (sectionId) {
     params.append('section_id', sectionId)
+  } else if (appBlockId) {
```
jamesmengo, line 109: This only happens for app embeds right? Could we clarify that if so?

jamesmengo, line 109: I expected updating an `app block` (targetting a section) to attach and `appBlockId` but for the `sectionId` to take precedence.
It seems like we only send `app_block_id` when it's an app embed being updated. If that's intended maybe we can change this to something like appEmbedBlockId?

Non-blocking.

packages/theme/src/cli/utilities/theme-environment/hot-reload/client.ts
```diff
@@ -51,70 +68,95 @@ function hotReloadScript() {
...
+      fullPageReload(data.key, error)
+    })
+  }
+
+  const refreshAppEmbedBlock = async (data: HotReloadEvent & {type: 'extAppBlock'}, block: Element) => {
+    const controller = new AbortController()
+
+    const appBlockId = block.id.replace(/^shopify-block-/, '')
+    const params = [
+      `app-block-id=${encodeURIComponent(appBlockId)}`,
```
jamesmengo, line 114: Is this going to be used for all `app blocks` or just `app embeds`?
Related to [this comment](https://github.com/Shopify/cli/pull/4375/files#r1736867327)

jamesmengo (approved):
🎩 and walked through (part) of the code, focusing on setup and hot reloading.
Functional tests went great! I left a couple of comments pertaining to naming / clarity (can be followups, though let's do another tophat after the merge conflicts are resolved)

1) The hot reload for removing / updating / adding `assets` and `liquid files` worked like a charm
2) Getting `app embed` hot reloading to trigger was a bit confusing for me - partly because I thought it would apply to **all** app blocks Stepping through the code resolved this misunderstanding. Left a comment trail starting [here](https://github.com/Shopify/cli/pull/4375/files#r1736870195)
3) I wonder if we could make the terminal output a little less noisy in a followup by rendering subdued text for the `URI` 🤔 

image

_Note: The setup flow for installing the app is **so much** better via partner dashboard_

packages/theme/src/cli/utilities/theme-environment/storefront-renderer.ts
karreiro: I had the _same_ conflict — but this was my reasoning: the API accepts the `app_block_id` as a parameter, so we should keep that consistent here, since the `storefront-renderer.ts` works as a client of the SFR API. This way, if in the future SFR supports rendering for app blocks as well, this logic wouldn’t be impacted. However, the logic in the hot reload module would be, as that’s where we’re distinguishing between section blocks and embed blocks.

packages/theme/src/cli/utilities/theme-environment/hot-reload/client.ts
karreiro: In this context, we're only handling app embeds, so it's indeed fair to rename this! :)

PR #4376, opened 2024-08-26
Refresh the theme development server session every 30 minutes

karreiro:

The remote renderer was frequently authenticating, which impacted the performance of the development server. Now, the development server session is automatically refreshed every 30 minutes.

With this change, the rendering requests become much faster:

**Before**
Screenshot 2024-08-30 at 13 06 09

**After**
Screenshot 2024-08-30 at 13 05 32

This PR extracts the `DevServerSession` creation to the `dev-server-session.ts` module, which creates and refreshes the session.

### How to test your changes?

- Update the `SESSION_TIMEOUT_IN_MS` to 10 seconds (`10000`) [here](https://github.com/Shopify/cli/blob/03e53f9d4fd929b346c61e96a5f76da04e6ff06d/packages/theme/src/cli/utilities/theme-environment/dev-server-session.ts#L8)
- Run `shopify theme dev --dev-preview --verbose`
- Notice that the dev session is updated every 10 seconds and that rendering now takes half the time 🚀

### Post-release steps

N/A

### Measuring impact

### Checklist


packages/theme/src/cli/utilities/theme-environment/storefront-renderer.test.ts
```diff
@@ -16,17 +13,16 @@ vi.mock('@shopify/cli-kit/node/http', async () => {
...
 const successResponse = {ok: true, status: 200, headers: {get: vi.fn()}} as any
-const sessionCookies = {
-  storefront_digest: '00001111222233334444',
-  _shopify_essential: ':00112233445566778899:',
-}
 
-const session = {
-  token: 'token_abc123',
+const session: DevServerSession = {
+  token: 'admin_token_abc123',
```
jamesmengo, line 18: 🤔 I wonder if it's worth renaming `token` to `adminToken`, though that would mean that `AdminSession` parent type would need to be renamed

I think we can defer this for now but if we reach a state of >=3 tokens I think we should consider renaming `token`

packages/theme/src/cli/utilities/theme-environment/storefront-renderer.ts
```diff
@@ -109,15 +104,15 @@ function buildStorefrontUrl(session: DevServerSession, {path, sectionId, query}:
...
     return 'https://theme-kit-access.shopifyapps.com/cli/sfr'
   } else {
     return `https://${session.storeFqdn}`
   }
 }
 
-function isThemeAccessSession(session: DevServerSession) {
+function isThemeAccessSession(session: AdminSession) {
   return session.token.startsWith('shptka_')
 }
```
jamesmengo, line 117: This is pretty trivial, but what do you think about moving this to `dev-server-session`?

packages/theme/src/cli/utilities/theme-environment/types.ts
```diff
@@ -23,9 +23,18 @@ export interface DevServerSession extends AdminSession {
...
+   *   - storefront_digest:  This cookie identifies an authenticated
+   *                         session and allows rendering to happen without
+   *                         the need for a password.
+   *
+   *   - _shopify_essential: This cookie essentially identifies the session,
+   *                         which is crucial to determine the theme being used
+   *                         during the rendering.
    */
-  expiresAt: Date
+  sessionCookies?: {[key: string]: string}
```
jamesmengo, line 37: Is this ever going to be actually undefined? Are there any side effects (aside from testing) of making this required?

https://github.com/Shopify/cli/blob/03e53f9d4fd929b346c61e96a5f76da04e6ff06d/packages/theme/src/cli/utilities/theme-environment/storefront-session.ts#L44

jamesmengo (approved):
Exciting!
Left some nitpicks but the tophat worked and the code makes sense to me

jamesmengo (approved):

packages/theme/src/cli/utilities/theme-environment/types.ts
karreiro: Great call! `cookieRecord` is no longer nullable :)

packages/theme/src/cli/utilities/theme-environment/storefront-renderer.test.ts
karreiro: The session is indeed handling a lot of tokens. Another possible direction could be to introduce a getter in the `DevServerSession`. For now, I'm keeping the session as it is, as this PR aims to only refresh the session object, however, that's certainly a valuable approach to consider.

PR #4379, opened 2024-08-26
Functions_378 Adding custom info box for app logs

saga-dasgupta:

Fixes https://github.com/Shopify/shopify-functions/issues/378

When user uses the app logs command, possibly with the `--store` flag, we want the info box the user sees to have relevant data such as dev stores the user entered and `update URLs` not being present as it is not relevant to the app logs command. 

Created our own custom info box that gets used instead of the default one for `app logs` command. 

Before:
App logs
Screenshot 2024-08-27 at 8 52 40 AM

App dev
Screenshot 2024-08-30 at 8 58 34 AM

After: 
App logs
Screenshot 2024-08-29 at 2 29 46 PM

App dev
Screenshot 2024-08-30 at 8 47 23 AM

### How to test your changes?

- Checkout my branch and run ` pnpm run shopify app logs --path <YOUR APP> --store <Store URL 1> <Store URL 2>`

### Measuring impact

### Checklist

### DIFF
--- a/packages/app/src/cli/services/context.test.ts
+++ b/packages/app/src/cli/services/context.test.ts
@@ -444,6 +444,8 @@ describe('ensureDevContext', async () => {
         remoteAppUpdated: true,
         updateURLs: true,
         localApp: app,
+        organization: 'org1',
+        configFile: 'shopify.app.toml',
       })
       expect(setCachedAppInfo).not.toHaveBeenCalled()
 
@@ -508,6 +510,8 @@ dev_store_url = "domain1"
         storeId: STORE1.shopId,
         remoteAppUpdated: true,
         updateURLs: true,
+        organization: 'org1',
+        configFile: 'shopify.app.toml',
       })
       expect(setCachedAppInfo).not.toHaveBeenCalled()
 
@@ -662,7 +666,7 @@ api_version = "2023-04"
           },
           ' to your command to reset your app configuration.',
         ],
-        headline: 'Using shopify.app.toml:',
+        headline: 'Using shopify.app.toml for default values:',
       })
     })
   })
@@ -683,6 +687,7 @@ api_version = "2023-04"
       storeId: STORE1.shopId,
       remoteAppUpdated: true,
       updateURLs: undefined,
+      organization: 'org1',
     })
     expect(setCachedAppInfo).toHaveBeenNthCalledWith(1, {
       appId: APP1.apiKey,
@@ -719,6 +724,8 @@ api_version = "2023-04"
       storeId: STORE1.shopId,
       remoteAppUpdated: true,
       updateURLs: undefined,
+      organization: 'org1',
+      configFile: 'shopify.app.toml',
     })
   })
 
@@ -743,6 +750,7 @@ api_version = "2023-04"
       storeId: STORE1.shopId,
       remoteAppUpdated: false,
       updateURLs: undefined,
+      organization: 'org1',
     })
     expect(fetchOrganizations).not.toHaveBeenCalled()
     expect(setCachedAppInfo).toHaveBeenNthCalledWith(1, {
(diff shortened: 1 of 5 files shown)


packages/app/src/cli/services/context.test.ts
```diff
@@ -776,6 +776,47 @@ api_version = "2023-04"
...
+        'You can pass ',
+        {
+          command: '--reset',
+        },
+        ' to your command to reset your app configuration.',
+      ],
+      headline: 'Waiting for app logs...\n\nUsing these settings:',
+    })
+  })
+
```
saga-dasgupta, line 819: I didn't write a test for the case where this flag is not passed because other tests in this file are already covering that case.

packages/app/src/cli/services/context.ts
```diff
@@ -65,6 +66,8 @@ interface DevContextOutput {
   storeId: string
   updateURLs: boolean | undefined
   localApp: AppInterface
+  organization?: string
+  configFile?: string
```
saga-dasgupta, line 70: So I decided to have `ensureDevContext` return the `organization` and `configFile` to `logs` so we don't waste time and computation on computing it again.

packages/app/src/cli/services/context.ts
```diff
@@ -251,12 +253,23 @@ export async function ensureDevContext(options: DevContextOptions): Promise<DevC
...
-  })
+  if (options.customLogsInfoBox) {
+    renderAppLogsConfigInfo({
+      devStore: selectedStore.shopDomain,
+      selectedStores: options.storeFqdns,
+      org: organization.businessName,
+      appName: selectedApp.title,
+      configFile: cachedInfo!.configFile,
+      resetMessage: resetHelpMessage,
+    })
```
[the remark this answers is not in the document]
lopert: I agree this doesn't feel ideal, and it feels like we're a bit stuck since we want to `ensureDevContext`, but don't want the post ensure render of the box it does by default.

Could we have `showReusedDevValues` be the default, but be able to provide our own render function when invoking `ensureDevContext` ?

packages/app/src/cli/services/context.ts
```diff
@@ -902,6 +923,33 @@ export function renderCurrentlyUsedConfigInfo({
...
+  }
+
+  let body: TokenItem = [{list: {items}}]
+  if (resetMessage) body = [...body, '\n', ...resetMessage]
+
+  const fileName = configFile && getAppConfigurationFileName(configFile)
+
+  renderInfo({
+    headline: `Waiting for app logs...\n\n${configFile ? `Using ${fileName}:` : 'Using these settings:'}`,
+    body,
```
lopert: A lot of this is very similar to today's render function, `renderCurrentlyUsedConfigInfo`. Is there an opportunity to reduce some of the code duplication between the two?

saga-dasgupta: Okay, I extracted some of the common code between the 2 methods out into a separate method getting called by both the render functions.

packages/app/src/cli/services/context.ts
```diff
@@ -884,18 +914,11 @@ export function renderCurrentlyUsedConfigInfo({
...
-
-  let body: TokenItem = [{list: {items}}]
-  if (resetMessage) body = [...body, '\n', ...resetMessage]
+  const devStores = []
+  if (devStore) devStores.push(devStore)
 
+  const body = formInfoBoxBody(appName, org, devStores, resetMessage, updateURLs, includeConfigOnDeploy)
   const fileName = (appDotEnv && basename(appDotEnv)) || (configFile && getAppConfigurationFileName(configFile))
-
   renderInfo({
```
DuncanUszkay1, line 922: This makes sense to me, but we should get somebody from the CLI side to take a peek since we're refactoring their code

packages/app/src/cli/services/logs.ts
```diff
@@ -131,3 +131,25 @@ async function prepareForLogs(commandOptions: LogsOptions): Promise<{
...
+  if (storeFqdns && storeFqdns.length > 0) {
+    storeFqdns.forEach((storeUrl) => devStores.push(storeUrl))
+  } else {
+    devStores.push(storeFqdn)
+  }
+  const body = formInfoBoxBody(appName, org, devStores, resetHelpMessage)
+  const fileName = configFile && getAppConfigurationFileName(configFile)
+
+  renderInfo({
+    headline: `${configFile ? `Using ${fileName} for default values:` : 'Using these settings:'}`,
```
nickwesselman, line 152: Is this message just for `app logs` or `app dev` as well? Based on our discussion with the CLI team, I think this headline should be updated for both commands.

saga-dasgupta: Oh, my bad. I thought I'm not changing anything about the `app dev` info box. To keep my changes limited to `app logs`. I will add this to `app dev` too.👍🏻

packages/app/src/cli/services/logs.test.ts
```diff
@@ -193,6 +202,64 @@ describe('logs', () => {
...
+      status: 'status',
+      configName: 'config-name',
+      userProvidedConfigName: 'user-provided-config-name',
+    })
+
+    // Then
+    expect(ensureDevContext).toHaveBeenCalledWith({
+      apiKey: 'api-key',
+      configName: 'config-name',
+      customInfoBox: true,
```
lopert, line 229: When testing for a specific thing, I like to pull it out to the `// Given` area to really highlight it.
To make it clearer that this is really what we're focused on, I'd add a `const customInfoBox = true` or something similar above.

Just a nit, feel free to disregard.

lopert (approved):
🎩 'd with both just 1 shop and multiple `--store` and it worked 👍 Just a couple of nitpicky comments.

PR #4384, opened 2024-08-27
Add timeout and retry logic to downloading Javy and function-runner

jeffcharles:

CI is sometimes marking tests around this logic as timing out. I am making an educated guess that it's the call to `fetch` that's taking more than 13 seconds to return because I don't know why ungzipping or running a subprocess's help command (the other two I/O operations) would take more than 13 seconds to complete. Adding an explicit timeout and retries to the downloading operation mean, if the downloads were to hang for whatever reason, we'd abort the download and retry 3 times. Hopefully this should be sufficient to work around this problem (assuming it is a result of the network request that's causing the timeout).

When downloading the Javy or function-runner binaries, only allow up to 5 seconds for the download to complete and then retry for a given number of times.

I'm also having the CLI download Javy and function-runner to a temporary directory first and then move the file when it's finished being processed so retrying the command won't keep trying to load the same broken binary. This should also help if multiple things are trying to download Javy or function-runner in parallel.

I'm open to suggestions on if that 5 second timeout I'm imposing should be higher to not make life difficult for someone who has a really bad network connection but needs to download and install Javy or function-runner at that point in time. We can get at least one retry in if the timeout is under 12.5 seconds.

### How to test your changes?

1. `rm packages/app/src/cli/services/bin/javy` (if it exists)
2. `rm packages/app/dist/cli/services/bin/javy ` (if it exists)
3. `pnpm shopify app build --path <path_to_app_with_shopify_function> --verbose`
4. There should've been a verbose log saying it was downloading Javy
5. `packages/app/dist/cli/services/bin/javy` should now exist

Repeat the above steps for running a function to test function-runner.

### Measuring impact

### Checklist


jeffcharles:
I've also updated the PR to download the Javy and function-runner binaries to a temporary directory first and then move the binary after it's been fully downloaded and processed so re-running the installation command will retry if something interrupted the download.

packages/app/src/cli/services/function/binaries.ts
```diff
@@ -99,18 +100,46 @@ export async function installBinary(bin: DownloadableBinary) {
...
+  if (!(await fileExists(dir))) {
+    await mkdir(dir)
   }
+  await performActionWithRetryAfterRecovery(
+    async () => {
+      const controller = new AbortController()
+      // Picking 5000 ms as the timeout because Javy is around 13 megabytes, so it
+      // will take around 4 seconds to download on a 25 Mbit connection, and also
+      // be able to retry twice in the 13 seconds we have for a test to complete
+      // before timing out.
```
andrewhassan, line 113: Our users could have slower internet, so we should probably choose a higher timeout 🤔

packages/app/src/cli/services/function/binaries.test.ts
```diff
@@ -215,7 +217,9 @@ describe('functionRunner', () => {
     })
   })
 
-  test('function-runner installs and runs', async () => {
+  // This shouldn't take 20 seconds to run but sometimes the network hangs and
+  // retries need to happen.
+  test('function-runner installs and runs', {timeout: 20000}, async () => {
```
andrewhassan, line 222: This seems like more of an integration test. I wonder if we should mock the response so that we're not dependent on an external request. @shauns Do you have opinions on this?

packages/app/src/cli/services/function/build.test.ts
```diff
@@ -155,7 +155,7 @@ describe('runJavy', () => {
 })
 
 describe('runFunctionRunner', () => {
-  test('calls function runner to execute function locally', async () => {
+  test('calls function runner to execute function locally', {timeout: 20000}, async () => {
```
andrewhassan, line 158: I also believe these should be stubbed too. In [this PR](https://github.com/Shopify/cli/pull/4404/files#diff-674e37c8b6c1dd31f06e0f74865766428e378534290b123fdd1ff6fb491a5090R13), I've refactored the `runFunction` function and stubbed the install in the test.

packages/app/src/cli/services/function/binaries.ts
andrewhassan: Yeah that makes sense to me. I'm going to remove this timeout.

shauns (approved):

lopert (approved):

PR #4382, opened 2024-08-27
Use the `--dev-preview` beta flag in the `shopify theme pull` command

karreiro:
### WHY are these changes introduced?

Use the `--dev-preview` beta flag in the `shopify theme pull` command instead of `--beta`.

### WHAT is this pull request doing?

Renames the `--beta` to  `--dev-preview`.

### How to test your changes?

- Run `shopify theme pull --dev-preview`

### Post-release steps

N/A

### Measuring impact

How do we know this change was effective? Please choose one:

- [ ] n/a - this doesn't need measurement, e.g. a linting rule or a bug-fix
- [x] Existing analytics will cater for this addition
- [ ] PR includes analytics changes to measure impact

### Checklist

- [x] I've considered possible cross-platform impacts (Mac, Linux, Windows)
- [x] I've considered possible [documentation](https://shopify.dev) changes

PR #4391, opened 2024-08-28
[Theme] Support `.shopifyignore` file, `--only` and `--ignore` flags in dev server

frandiox:
Closes https://github.com/Shopify/develop-advanced-edits/issues/283

- Refactor ignore assets logic to live in themeFS
- Support ignore patterns in dev server
- Refactor file `stats` to keep the `mtime` in memory with the file content.
- Small perf improvement for server start up

--

⚠️   It's easier to review this PR commit by commit 🙏 

To 🎩 , pass `--ignore="assets/*.css"` flag and try to update some CSS file, it should be ignored.

### DIFF
--- a/packages/theme/src/cli/utilities/theme-environment/theme-environment.ts
+++ b/packages/theme/src/cli/utilities/theme-environment/theme-environment.ts
@@ -70,17 +70,15 @@ interface DevelopmentServerInstance {
   close: () => Promise<void>
 }
 
-function createDevelopmentServer(theme: Theme, ctx: DevServerContext, initialWork?: Promise<void>) {
+function createDevelopmentServer(theme: Theme, ctx: DevServerContext, initialWork: Promise<void>) {
   const app = createApp()
 
-  if (initialWork) {
-    app.use(
-      defineLazyEventHandler(async () => {
-        await initialWork
-        return defineEventHandler(() => {})
-      }),
-    )
-  }
+  app.use(
+    defineLazyEventHandler(async () => {
+      await initialWork
+      return defineEventHandler(() => {})
+    }),
+  )
 
   if (ctx.options.liveReload !== 'off') {
     app.use(getHotReloadHandler(theme, ctx))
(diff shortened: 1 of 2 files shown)


packages/theme/src/cli/utilities/theme-environment/theme-environment.ts
```diff
@@ -72,9 +70,18 @@ interface DevelopmentServerInstance {
...
 
+  if (initialWork) {
+    app.use(
+      defineLazyEventHandler(async () => {
+        await initialWork
+        return defineEventHandler(() => {})
+      }),
+    )
+  }
+
```
frandiox, line 84: This has nothing to do with the original PR, just a small perf improvement I forgot in another PR 😅

frandiox:
>I assume that's going to be fixed when we merge [this line](https://github.com/Shopify/cli/pull/4382/files#diff-268f28d79bb8d633f4eab6cf981c7bbbe4c25842313218875ffbec4ca28a0b18R39). If my assumption is correct, feel free to proceed with the merge, and I will merge mine later :)

Yeah, that function is creating a whole new themeFS instance so we need to call `.ready()`. Alternatively, I think @jamesmengo  is removing that new instance and reusing the main one in another PR. Should still call `.ready()` though, just in case 👍

jamesmengo:
> > I assume that's going to be fixed when we merge [this line](https://github.com/Shopify/cli/pull/4382/files#diff-268f28d79bb8d633f4eab6cf981c7bbbe4c25842313218875ffbec4ca28a0b18R39). If my assumption is correct, feel free to proceed with the merge, and I will merge mine later :)
> 
> Yeah, that function is creating a whole new themeFS instance so we need to call `.ready()`. Alternatively, I think @jamesmengo is removing that new instance and reusing the main one in another PR. Should still call `.ready()` though, just in case 👍

+1 to keeping the `ready` call.
We will continue to call `mountThemeFileSystem` in the `pull` command, as we have no existing `themeFileSystem` mounted when that command is executed.

The instance I removed came from [here](https://github.com/Shopify/cli/pull/4367/files#diff-9eb3bb93a3f5e57984c31d76daeaa4528cd2fb87617a0fc28667502d3114a97fL32) - we were mounting the file system **twice** during a single command (`dev`) when the `theme-editor-sync` flag is passed.

PR #4394, opened 2024-08-29
Ensure Javy is always installed before running it

jeffcharles:

If `shopify app function build` is run for a JS Function extension before `shopify app build` is executed, then Javy is not installed before trying to execute it. This used to work when using the `javy-cli` NPM package to execute Javy because that package would implicitly install Javy if it was not already installed. But now that the CLI doesn't use that package, installation needs to be explicitly performed. I've chosen to do it in the `runJavy` function to ensure a check is always performed to see if Javy is installed and install if it isn't.

Ensuring Javy is installed before trying to run it. The `installBinary` method performs a check to see if the binary is installed before attempting an install so this won't result in any attempts to redownload the binary if it's already present.

### How to test your changes?

Ensure you don't have the `javy` binary present by running `rm packages/app/dist/cli/services/bin/javy` and `rm packages/app/dist/cli/services/bin/javy`.

Then run:

```
pnpm shopify app function build --path <path-to-js-function-extension>
```

### Measuring impact

### Checklist

### DIFF
--- a/.changeset/spicy-cheetahs-obey.md
+++ b/.changeset/spicy-cheetahs-obey.md
@@ -1,0 +1,5 @@
+---
+'@shopify/app': patch
+---
+
+Ensure Javy is always installed before running it
--- a/packages/app/src/cli/services/function/build.test.ts
+++ b/packages/app/src/cli/services/function/build.test.ts
@@ -132,7 +132,7 @@ describe('bundleExtension', () => {
 })
 
 describe('runJavy', () => {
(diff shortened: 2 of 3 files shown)


nickwesselman:
Does this also cover the auto-builds that happen during the `replay` command?

jeffcharles:
> Does this also cover the auto-builds that happen during the replay command?

Yes it will. However I did notice a bug with how function-runner is being invoked that I'll put up a separate PR to address.

PR #4410, opened 2024-09-03
Set order status targets as customer account targets instead of checkout targets

lihaokx:

Fixes #0000 

### Measuring impact

### Checklist

### DIFF
--- a/packages/app/src/cli/services/dev/extension/utilities.ts
+++ b/packages/app/src/cli/services/dev/extension/utilities.ts
@@ -2,13 +2,6 @@ import {ExtensionInstance} from '../../../models/extensions/extension-instance.j
 
 import {fetchProductVariant} from '../../../utilities/extensions/fetch-product-variant.js'
 
-const CUSTOMER_ACCOUNT_CHECKOUT_RENDERED_TARGETS = new Set([
-  'customer-account.order-status.cart-line-item.render-after',
-  'customer-account.order-status.cart-line-list.render-after',
-  'customer-account.order-status.customer-information.render-after',
-  'customer-account.order-status.block.render',
-])
-
 /**
  * To prepare UI Extensions targeting Checkout for dev'ing we need to retrieve a valid product variant ID
  * @param extensions - The UI Extensions to dev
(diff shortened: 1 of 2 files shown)


packages/app/src/cli/services/dev/extension/utilities.ts
```diff
@@ -41,12 +34,7 @@ export function getExtensionPointTargetSurface(extensionPointTarget: string) {
     }
 
     // Covers Customer Accounts UI extensions (future)
-    case 'customeraccount':
```
lihaokx, line 44: This is from the legacy customer account targets, which are not used anymore

lihaokx:
> Should corresponding core PRs be built and shipped first to handle this change of how targets are handled at the CLI-level?

In my opinion, the sequencing of shipping prs does not matter too much. Devs will be in order index page after this change, which will make the behaviour of the [8 order-status targets](https://shopify.dev/docs/api/customer-account-ui-extensions/unstable/targets) consistent. 
@jrrafols

lihaokx:
> do we any plans to update related docs as well? 

This is a good point. Since cli is version-controlled, devs may still see legacy behaviour if they are not on the latest version. I think we can change the doc after the stable version cli is released. @brianshen1990

jrrafols (approved):
> I prefer to ship this one before the core prs, because cli is version-controlled and devs could be on legacy version or new version. We need to add the change to the earlier version ASAP.

Good point on versioning. I took a look and the next release is [slated for Sept 16](https://vault.shopify.io/teams/2493/pages/Releasing~FkHb.md#release-schedule). I think it's reasonable to expect that the core changes will be available by then so I'm comfortable with this PR going out in the meantime.

As mentioned by @brianshen1990, we should also aim to have our documentation updated at around the same timeframe.

brianshen1990 (approved):

vividviolet (commented):
The changes look fine but is it possible to also make Checkout redirect to customer accounts for these targets? We can't really fix older releases of the CLI so Checkout should account for handling the legacy behaviour.

lihaokx:
> The changes look fine but is it possible to also make Checkout redirect to customer accounts for these targets?
 We can't really fix older releases of the CLI so Checkout should account for handling the legacy behaviour.

If I understand correctly, you mean: in older version, does checkout redirect to customer account for the targets?

If so, the answer is yes, devs could be redirected to customer account after completing checkout in older version.

Here is the difference between older and newer version

#### Older version:

- Clicking targets on cli console --> checkout --> completing checkout --> customer account 

#### Newer version:

- Clicking targets on cli console --> customer account

The main difference is that: in newer version, devs could be in customer account without going to checkout.

@vividviolet

PR #4411, opened 2024-09-03
Use new parameters for app management create mutations

gordonhirsch:

Part 2 of https://github.com/Shopify/develop-app-foundations/issues/1050

This PR adapts to changes in the. App Management API introduced in [part 1](https://github.com/Shopify/shopify/pull/532987) of the sequence of PRs.

The new `AppVersionInput` graphql type is used in the App Management client as a variable for in the app/version create mutations. 

On deploy, if a bundle has been uploaded, only that bundle URL is specified as `sourceUrl` in `AppVersionInput`. That means that the bundle's manifest is the sole source of app module information like config, etc.

If no bundle has been uploaded, then name and app module information is passed in the `source` field of `AppVersionInput`, similar to existing (now-deprecated) API.

It doesn't appear that there are existing tests to cover the `deploy` logic in the app management client, so I added a few new ones to test deploys and this PR in particular. 
 
### How to test your changes?

Running `app deploy` for a new app will exercise all of the changes in this PR. Note that there is currently an unrelated [issue](https://github.com/Shopify/develop-app-inner-loop/issues/2072) where an app's initial deploy will fail at the CLI level because of a switch to using the partners client for some reason. In this case, the app and initial version are still created correctly.

Testing can be run against production once [part 1](https://github.com/Shopify/shopify/pull/532987) has deployed, or against a spin instance running that PR's branch.

### Measuring impact

### Checklist

### DIFF
--- a/packages/app/src/cli/utilities/developer-platform-client/app-management-client.test.ts
+++ b/packages/app/src/cli/utilities/developer-platform-client/app-management-client.test.ts
@@ -206,6 +206,28 @@ describe('deploy', () => {
   const session = {token, accountInfo: userAccountInfo}
   const client = new AppManagementClient(session)
 
+  const mockedVersionResponse: CreateAppVersionMutationSchema = {
+    appVersionCreate: {
+      version: {
+        id: versionId,
+        appModules: [],
+        metadata: {message: '', versionTag},
+      },
+      userErrors: [],
+    },
+  }
+  const mockedReleaseResponse: ReleaseVersionMutationSchema = {
+    appReleaseCreate: {
+      release: {
+        version: {
+          id: versionId,
+          metadata: {message: '', versionTag},
+        },
+      },
+      userErrors: [],
+    },
+  }
+
   test('creates an app version from a bundle URL', async () => {
     // Given
     const bundleUrl = 'https://example.com/mybundle'
@@ -245,8 +267,6 @@ describe('deploy', () => {
         metadata: {versionTag},
       }),
     )
-    expect(got.appDeploy.appVersion.versionTag).toEqual(versionTag)
-    expect(got.appDeploy.appVersion.uuid).toEqual(versionId)
   })
 
   test('creates an app version directly', async () => {
@@ -267,21 +287,6 @@ describe('deploy', () => {
       handle: 'handle2',
     }
     const appModules = [module1, module2]
-    const mockedVersionResponse: CreateAppVersionMutationSchema = {
-      appVersionCreate: {
-        version: {
-          id: versionId,
-          appModules: appModules.map((mod) => ({
-            uuid: mod.uid,
-            handle: mod.handle,
-            config: JSON.parse(mod.config),
-            specification: {name: mod.specificationIdentifier, identifier: mod.specificationIdentifier},
-          })),
-          metadata: {message: '', versionTag},
-        },
-        userErrors: [],
-      },
-    }
 
     vi.mocked(appManagementRequest).mockResolvedValueOnce(mockedVersionResponse)
 
@@ -318,36 +323,12 @@ describe('deploy', () => {
         metadata: {versionTag},
       }),
     )
-    expect(got.appDeploy.appVersion.versionTag).toEqual(versionTag)
-    expect(got.appDeploy.appVersion.uuid).toEqual(versionId)
   })
 
   test('creates an app version and a release', async () => {
     // Given
     const name = 'my-app-name'
-    const mockedVersionResponse: CreateAppVersionMutationSchema = {
-      appVersionCreate: {
-        version: {
-          id: versionId,
-          appModules: [],
-          metadata: {message: '', versionTag},
-        },
-        userErrors: [],
-      },
-    }
     vi.mocked(appManagementRequest).mockResolvedValueOnce(mockedVersionResponse)
-
-    const mockedReleaseResponse: ReleaseVersionMutationSchema = {
-      appReleaseCreate: {
-        release: {
-          version: {
-            id: versionId,
-            metadata: {message: '', versionTag},
-          },
-        },
-        userErrors: [],
-      },
-    }
     vi.mocked(appManagementRequest).mockResolvedValueOnce(mockedReleaseResponse)
 
     // When
@@ -386,7 +367,5 @@ describe('deploy', () => {
         versionId,
       }),
     )
-    expect(got.appDeploy.appVersion.versionTag).toEqual(versionTag)
-    expect(got.appDeploy.appVersion.uuid).toEqual(versionId)
   })
 })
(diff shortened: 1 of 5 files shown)


packages/app/src/cli/utilities/developer-platform-client/app-management-client.test.ts
```diff
@@ -188,3 +193,200 @@ describe('versionDeepLink', () => {
     expect(got).toEqual('https://dev.shopify.com/dashboard/1/apps/2/versions/3')
   })
 })
+
+describe('deploy', () => {
```
amcaplan, line 197: Personal opinion.

These tests seem like lengthy restatements of the type signatures, and personally I don't find that they increase my confidence in the code beyond the type signature itself. It's just another thing to change if we need to change the type signature later.

I do see the value in checking the logic in generation of the inputs, since it varies based on inputs to `deploy`. So maybe the tests could be significantly cut down by just focusing on that piece of it, verifying that the inputs are generated correctly. But I think `appManagementRequest` can just be a mock, since the only real concern is that the inputs are right.

gordonhirsch: Hmm, ok, I could simplify the mocked response. There is some post-request code to map the response that it was meant to verify. But it's really an example of trying to do a bit to improve the tests while in there, and not strictly related to this PR.

packages/app/src/cli/utilities/developer-platform-client/app-management-client.ts
```diff
@@ -626,29 +626,39 @@ export class AppManagementClient implements DeveloperPlatformClient {
...
-    // `name` is from the package.json package name or the directory name, while
-    // the branding module reflects the current specified name in the TOML.
-    // Since it is technically valid to not have a branding module, we will default
-    // to the `name` if no branding module is present.
-    let updatedName = name
-    const brandingModule = appModules?.find((mod) => mod.specificationIdentifier === BrandingSpecIdentifier)
-    if (brandingModule) {
-      updatedName = JSON.parse(brandingModule.config).name
+    let versionInput
+    if (bundleUrl) {
```
isaacroldan, line 630: Even if there are no assets, we want to build the manifest and upload it to GCS no? so there will always be a bundleUrl? In which cases can we deploy without a bundle URL?

packages/app/src/cli/utilities/developer-platform-client/app-management-client.ts
```diff
@@ -626,29 +626,39 @@ export class AppManagementClient implements DeveloperPlatformClient {
...
+      let updatedName = name
+      const brandingModule = appModules?.find((mod) => mod.specificationIdentifier === BrandingSpecIdentifier)
+      if (brandingModule) {
+        updatedName = JSON.parse(brandingModule.config).name
+      }
+      versionInput = {
+        source: {
+          name: updatedName,
+          appModules: (appModules ?? []).map((mod) => {
+            return {
```
isaacroldan, line 648: Should we force to always use a bundleURL and not accept `versionInput` with a an appModule array? This way we can be 100% sure that all versions have a manifest and we don't need to support two formats for the API here.

packages/app/src/cli/utilities/developer-platform-client/app-management-client.ts
```diff
@@ -626,29 +626,39 @@ export class AppManagementClient implements DeveloperPlatformClient {
...
-    // Since it is technically valid to not have a branding module, we will default
-    // to the `name` if no branding module is present.
-    let updatedName = name
-    const brandingModule = appModules?.find((mod) => mod.specificationIdentifier === BrandingSpecIdentifier)
-    if (brandingModule) {
-      updatedName = JSON.parse(brandingModule.config).name
+    let versionInput
+    if (bundleUrl) {
+      versionInput = {
+        sourceUrl: bundleUrl,
```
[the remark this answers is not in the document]
gordonhirsch: Yes, the bundle manifest will contain the app name.

packages/app/src/cli/utilities/developer-platform-client/app-management-client.ts
gordonhirsch: We can do it that way, but to me, it seems a bit wasteful to require a separate bundle upload when there are no assets in the app.

packages/app/src/cli/utilities/developer-platform-client/app-management-client.ts
gordonhirsch: If/when we need to store a manifest in GCS, I think it would be easy to generate one on the back end for the no asset case. In general I think it is worthwhile to have an API for creating app versions that does not require the client to make 3 network calls for a new app version (request bundle url, upload bundle, create the version).

packages/app/src/cli/utilities/developer-platform-client/app-management-client/graphql/create-app-version.ts
```diff
@@ -32,14 +33,17 @@ export const CreateAppVersionMutation = gql`
...
-  appSource: {
-    assetsUrl?: string
-    appModules: {
-      uid: string
-      specificationIdentifier?: string
-      config: JsonMapType
-    }[]
+  version: {
+    sourceUrl?: string
+    source?: {
```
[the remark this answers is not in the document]
gordonhirsch: Good point. I think we'd want to use a GraphQL union type in the schema if we want to use it here. I'll need to revisit why it is not that way in the schema 🤔

packages/app/src/cli/utilities/developer-platform-client/app-management-client.ts
isaacroldan: It could seem that way, but is a more unified API and the code would be simpler in both CLI and backend, maybe is worth the extra resources for that? 🤔

gordonhirsch: The latest server [change](https://github.com/Shopify/shopify/pull/538723) unifies the API a bit more. Instead of a separate source type, it takes the actual manifest JSON when the app version is directly specified. The code in this PR has been updated accordingly. The back end will still have the responsibility of storing the manifest, but I think the API change removes the biggest concerns and is a reasonable compromise.

packages/app/src/cli/utilities/developer-platform-client/app-management-client.ts
```diff
@@ -628,29 +628,40 @@ export class AppManagementClient implements DeveloperPlatformClient {
...
+        source: {
+          name: updatedName,
+          modules: (appModules ?? []).map((mod) => {
+            return {
+              type: mod.specificationIdentifier,
+              uid: mod.uid ?? mod.uuid ?? mod.handle,
+              handle: mod.handle,
+              config: JSON.parse(mod.config),
+            }
+          }),
```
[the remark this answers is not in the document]
isaacroldan: I'm ok with that, as long as we don't forget about it! :)

PR #4426, opened 2024-09-09
[Themes] Theme Dev - Do not display warnings when file deletion fails

jamesmengo:
Fixes https://github.com/Shopify/develop-advanced-edits/issues/310

>The goals of this task are:
Update the logic so the file system module doesn't attempt to delete a non-existent file.
Change the logging approach from renderWarning to outputDebug since users cannot take any action on this issue.

~~- Prevents the watcher from attempting to delete files that don't exist on the local theme~~
- `outputDebug` when the file deletion fails (either due to server error or because the file already doesn't exist)

### How to test your changes?
Setup:
1) Create a theme with missing files -> `shopify theme push -d` -> delete files from the code editor
2) Execute `shopify-dev theme dev --dev-preview --theme-editor-sync --verbose` 
3) Select 'delete files from local theme'
4) Check for `verbose` logs. You should see logs like: `Failed to delete file "${fileKey}" from remote theme.`

BEFORE
image

AFTER

https://github.com/user-attachments/assets/3fe26e8a-7a97-43c5-bbdd-970b2842ba90

### Measuring impact

### Checklist

### DIFF
--- a/packages/theme/src/cli/utilities/theme-fs.test.ts
+++ b/packages/theme/src/cli/utilities/theme-fs.test.ts
@@ -16,7 +16,7 @@ import {removeFile, writeFile} from '@shopify/cli-kit/node/fs'
 import {test, describe, expect, vi, beforeEach} from 'vitest'
 import chokidar from 'chokidar'
 import {deleteThemeAsset, fetchThemeAsset} from '@shopify/cli-kit/node/themes/api'
-import {renderWarning} from '@shopify/cli-kit/node/ui'
+import {outputDebug} from '@shopify/cli-kit/node/output'
 import EventEmitter from 'events'
 import type {Checksum, ThemeAsset} from '@shopify/cli-kit/node/themes/types'
 
@@ -29,6 +29,7 @@ vi.mock('@shopify/cli-kit/node/fs', async (realImport) => {
 vi.mock('./asset-ignore.js')
 vi.mock('@shopify/cli-kit/node/themes/api')
 vi.mock('@shopify/cli-kit/node/ui')
+vi.mock('@shopify/cli-kit/node/output')
 
 beforeEach(async () => {
   vi.mocked(getPatternsFromShopifyIgnore).mockResolvedValue([])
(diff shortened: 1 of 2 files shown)


packages/theme/src/cli/utilities/theme-fs.ts
```diff
@@ -194,13 +194,20 @@ export function mountThemeFileSystem(root: string, options?: ThemeFileSystemOpti
     files.delete(fileKey)
     emitEvent('unlink', {fileKey})
 
-    deleteThemeAsset(Number(themeId), fileKey, adminSession)
-      .then(async (success) => {
-        if (!success) throw new Error('Unknown issue.')
-        outputSyncResult('delete', fileKey)
+    fetchThemeAsset(Number(themeId), fileKey, adminSession)
```
jamesmengo, line 197: I'm debating whether we should make this call explicitly, or just call `deleteThemeAsset` (which will fail) and handle the failures silently. 

This would save an additional call every time a file is deleted, but the tradeoff with that approach would mean that we are unable to discern between legitimate failures of `deleteThemeAsset` (such as a server failure) from instances where the failure comes from the asset not existing on the remote theme.

packages/theme/src/cli/utilities/theme-fs.test.ts
```diff
@@ -471,6 +478,84 @@ describe('theme-fs', () => {
...
+
+      // When
+      const themeFileSystem = mountThemeFileSystem(root)
+      await themeFileSystem.ready()
+
+      const deleteOperationPromise = new Promise<void>((resolve) => {
+        themeFileSystem.addEventListener('unlink', () => {
+          setImmediate(resolve)
+        })
+      })
```
frandiox, line 511: FYI there's also [`vi.waitFor` in Vitest](https://vitest.dev/api/vi.html#vi-waitfor) to deal with async code, but I think the promise here is actually better 👍

frandiox (approved):
Thanks for also adding the test!

PR #4404, opened 2024-08-30
Use function runner binary in replay

andrewhassan:
### WHY are these changes introduced?

The Function Runner binary to use when running functions was changed recently. However, the `replay` command was not updated, so it was using the old global binary.

### WHAT is this pull request doing?

This PR does two things (in two commits):
1. Updates the replay command to be consistent with how functions should be run.
2. Consolidates the logic for running a function so that it can be used everywhere.

### How to test your changes?

Test:
- `function replay` without watch runs function
- `function replay` with watch runs function on change
- `function run` with piped input runs function
- `function run` with input path runs function

### Measuring impact

How do we know this change was effective? Please choose one:

- [x] n/a - this doesn't need measurement, e.g. a linting rule or a bug-fix
- [ ] Existing analytics will cater for this addition
- [ ] PR includes analytics changes to measure impact

### Checklist

- [x] I've considered possible cross-platform impacts (Mac, Linux, Windows)
- [x] I've considered possible [documentation](https://shopify.dev) changes

PR #4429, opened 2024-09-10
[ShopifyVM] Update-function runner to use dynamic scaling

mssalemi:
#gsd:37875

Addresses:
- https://github.com/Shopify/script-service/issues/7502

Pre-requisite:
- https://github.com/Shopify/function-runner/pull/352
- Release new version ✅

Updates the function-runner version used by CLI. 

The new version will show an update benchmark results sections, including more information on the input and output size. Also, the color of the text in the `Benchmarks` section will be determined by the dynamic scaling limits. 

The new function runner requires two additional parameters to be provided (schema path and query path). 

Note: If incorrect arguments are provided to function for schema path and query path, then function runner will halt execution. If parameters are not passed, then it will use defaults for the limits. 

For Partners DX, there should not be an changes in the DX. They will run the same command (example):

```
npm run shopify app function run -- --input=./js_function_input.json --path=./innovative-shareholder-app   
```

A partners function will have a schema and query file. We can include the paths to those files when we call function runner from the cli.

Screenshot 2024-09-11 at 4 31 29 PM

Uses new version of function runner to provide additional information on benchmark results and a resource limits section.

From CLI we need to provide two additional paths to function-runner, `schema-path` and `query-path`.

The changes in this PR are small, but its important to review how we are determining the query / schema path we are providing. 

1. **Query Path**
- The extensions targeting configuration (specified in the toml) has input query path accessible
- The input query for functions are located in either `src/run.graphql`, or `fetch.grapql`. As per the docs, I dont know if this is a requirement or someting the partner can change. But regardless, we have access to the input query path specified. 
- And in CLI we have access to this value from `functionExtension.configuration.targeting?.[0]?.input_query`

2. **Schema Path**
- Schema gets created with the `generate schema` command, and it creates a `schema.graphql` at the root of the app directory. 
- If there is no schema file, we will call the code to generate one similar to what the `schema` command does with utilizing `generateSchemaService`. 

Making Note: CLI handles cases where query and schema are not at requires paths. 

If Schema or input is not located in the app directory, partners will get errors when deploying the function. 

Schema not located in root - CLI handles these errors on deploy like:
Screenshot 2024-09-11 at 9 13 42 AM

Query not located in specifed location in cargo toml - CLI handles these errors on deploy like:
Screenshot 2024-09-11 at 9 15 34 AM

^ verified partner can update the value of the input query file in the `shopify.extension.toml`

If there is no schema, we will call the code to generate the schema, and output a message when the schema file gets cr

Screenshot 2024-09-11 at 4 28 26 PM

### 🎩 How to test your changes?

Alex showed me and linked his [comment here](https://github.com/Shopify/cli/pull/4404#pullrequestreview-2280436279). But function-runner in the CLI gets run from a binary. So to run the code before we release the new version of function-runner. You will need to replace the binary in your local CLI.

### Measuring impact

### Checklist

### DIFF
--- a/packages/app/src/cli/commands/app/function/run.ts
+++ b/packages/app/src/cli/commands/app/function/run.ts
@@ -1,4 +1,4 @@
-import {functionFlags, inFunctionContext} from '../../../services/function/common.js'
+import {functionFlags, inFunctionContext, getOrGenerateSchemaPath} from '../../../services/function/common.js'
 import {runFunction} from '../../../services/function/runner.js'
 import {appFlags} from '../../../flags.js'
 import Command from '@shopify/cli-kit/node/base-command'
@@ -44,7 +44,7 @@ export default class FunctionRun extends Command {
     await inFunctionContext({
       path: flags.path,
       userProvidedConfigName: flags.config,
-      callback: async (_app, ourFunction) => {
+      callback: async (app, ourFunction) => {
         let functionExport = DEFAULT_FUNCTION_EXPORT
 
         if (flags.export !== undefined) {
@@ -78,12 +78,18 @@ export default class FunctionRun extends Command {
           )
         }
 
+        const inputQueryPath = ourFunction?.configuration.targeting?.[0]?.input_query
+        const queryPath = inputQueryPath && `${ourFunction?.directory}/${inputQueryPath}`
+        const schemaPath = await getOrGenerateSchemaPath(ourFunction, app)
+
         await runFunction({
           functionExtension: ourFunction,
           json: flags.json,
           inputPath: flags.input,
           export: functionExport,
           stdin: 'inherit',
+          schemaPath,
+          queryPath,
         })
       },
     })
--- a/packages/app/src/cli/services/function/common.ts
+++ b/packages/app/src/cli/services/function/common.ts
@@ -3,11 +3,13 @@ import {loadApp} from '../../models/app/loader.js'
 import {loadLocalExtensionsSpecifications} from '../../models/extensions/load-specifications.js'
 import {ExtensionInstance} from '../../models/extensions/extension-instance.js'
 import {FunctionConfigType} from '../../models/extensions/specifications/function.js'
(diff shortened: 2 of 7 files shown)


packages/app/src/cli/services/function/common.ts
```diff
@@ -50,3 +51,12 @@ export async function inFunctionContext({
     throw new AbortError('Run this command from a function directory or use `--path` to specify a function directory.')
   }
 }
+
+export async function readSchemaFile(directory: string): Promise<string | undefined> {
```
andrewhassan, line 55: This isn't really reading the file, it's just returning the schema path of the function by concatenating the given path. I think we can make this harder to misuse by ensuring it's operating on a function. What if we did something like this instead?

```typescript
export async function schemaPath(extension: ExtensionInstance<FunctionConfigType>): Promise<string | null> {
  const path = joinPath(extension.directory, 'schema.graphql');
  if (await fileExists(path)) {
    return path
  }

  return null
}
```

packages/app/src/cli/commands/app/function/run.ts
```diff
@@ -78,12 +78,18 @@ export default class FunctionRun extends Command {
           )
         }
 
+        const inputQueryPath = ourFunction?.configuration.targeting?.[0]?.input_query
+        const queryPath = inputQueryPath && `${ourFunction?.directory}/${inputQueryPath}`
+        const schemaPath = await readSchemaFile(ourFunction?.directory || '')
```
andrewhassan, line 83: If there isn't a schema, should we run the `generate schema` command first automatically?

mssalemi: Yea I think that makes sense - will update!

mssalemi: Updated this to call the generate schema.  It will output the message from the generate schema when it has been created:

Screenshot 2024-09-11 at 4 28 26 PM

PR #4432, opened 2024-09-10
[Themes] - Theme Dev - Reduce API calls and IO operations triggered from theme reconciliation step

jamesmengo:
Fixes https://github.com/Shopify/develop-advanced-edits/issues/330

While [this PR](https://github.com/Shopify/cli/pull/4426) prevents errors from being surfaced when we delete a local file during the reconciliation step, we're still making a call for each file there to check for its existence. These changes prevent us from making those requests in the first place.

https://github.com/user-attachments/assets/1c1df339-445c-4f41-8ed7-cc63873064d0

BENEFITS
- prevents us from calling API for `local filesystem` changes that come from the `reconciliation` step
- we could potentially remove the call to [check that a file exists](https://github.com/Shopify/cli/pull/4426/files#diff-d857e2b583dcee2197fc538eb4d35aabb5ce83cee18ebb970ece7854dda8ba07R197), reducing the number of API calls for each time a file is deleted from two (one to check if the file exists and one to delete it) to one. However, this needs further consideration.

TRADEOFFS
- if you modify the file system during the prompts (e.g., manually deleting a file), it won't be caught in the change set

Defers initialization of the theme watcher until after the reconciliation step has completed. This will prevent the service from trying to synchronize files changes that are coming from the reconciliation step, saving an additional API call each time.

### How to test your changes?

**Example:**
-> Create a development theme `shopify theme delete -d -f && shopify theme push -d`
-> Create a file that will only be present on your local theme
-> Initialize command `shopify theme dev --dev-preview --theme-editor-sync --verboes`
-> Select `delete files from local theme`
-> Note that we are no longer outputting a verbose log with: `File does not exist on remote theme.`

### Measuring impact

### Checklist

### DIFF
--- a/packages/theme/src/cli/utilities/theme-fs.test.ts
+++ b/packages/theme/src/cli/utilities/theme-fs.test.ts
@@ -181,6 +181,8 @@ describe('theme-fs', () => {
         key: 'assets/new_file.css',
         checksum: '1010',
         value: 'content',
+        stats: {size: 7, mtime: expect.any(Number)},
+        attachment: '',
       })
     })
 
@@ -203,6 +205,8 @@ describe('theme-fs', () => {
         key: 'assets/new_image.gif',
         checksum: '1010',
         attachment,
+        value: '',
+        stats: {size: 6, mtime: expect.any(Number)},
       })
     })
 
@@ -216,7 +220,13 @@ describe('theme-fs', () => {
 
       let filesUpdated = false
       vi.mocked(writeFile).mockImplementationOnce(() => {
-        filesUpdated = themeFileSystem.files.get(newAsset.key) === newAsset
+        expect(themeFileSystem.files.get(newAsset.key)).toEqual({
+          ...newAsset,
+          attachment: '',
+          stats: {size: 7, mtime: expect.any(Number)},
+        })
+        filesUpdated = true
+
         return Promise.resolve()
       })
 
--- a/packages/theme/src/cli/utilities/theme-fs.ts
+++ b/packages/theme/src/cli/utilities/theme-fs.ts
@@ -219,7 +219,15 @@ export function mountThemeFileSystem(root: string, options?: ThemeFileSystemOpti
       await removeThemeFile(root, fileKey)
     },
     write: async (asset: ThemeAsset) => {
-      files.set(asset.key, asset)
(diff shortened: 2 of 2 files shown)


packages/theme/src/cli/utilities/theme-environment/theme-environment.ts
```diff
@@ -12,9 +12,11 @@ import type {Checksum, Theme} from '@shopify/cli-kit/node/themes/types'
 import type {DevServerContext} from './types.js'
 
 export function setupDevServer(theme: Theme, ctx: DevServerContext) {
-  const watcherPromise = setupInMemoryTemplateWatcher(theme, ctx)
+  const watcherPromise = setupInMemoryTemplateWatcher(ctx)
   const envSetup = ensureThemeEnvironmentSetup(theme, ctx)
```
jamesmengo, line 16: There are currently 3 scenarios in which the theme reconciliation step may modify the local filesystem.
Below are how they are currently handled, and how they would be impacted with these changes.


| Case                                | Current                                                                 | With This Change                        |
|-------------------------------------|-------------------------------------------------------------------------|-----------------------------------------|
| Delete file only present on local theme | Call API to check if file exists on remote theme, stop because it doesn't.     | Save an API call to determine existence |
| Download file only present on remote theme | Handled [here](https://github.com/Shopify/cli/blob/9e75d39ed428fe286c8ff23b385aefa9abbdf953/packages/theme/src/cli/utilities/theme-fs.ts#L134-L142). Read updated checksums and choose not to trigger an upload | Save an IO call to `read`                 |
| Update file using remote theme value | Handled [here](https://github.com/Shopify/cli/blob/9e75d39ed428fe286c8ff23b385aefa9abbdf953/packages/theme/src/cli/utilities/theme-fs.ts#L134-L142). Read updated  checksums and choose not to trigger an upload | Save an IO call to `read`                 |

*_Read updated checksums and choose not to trigger an upload_ -> Changes that come from the remote theme will have matching checksums before and after we call `read`, because we [update the remote checksum when we `write`](https://github.com/Shopify/cli/blob/9e75d39ed428fe286c8ff23b385aefa9abbdf953/packages/theme/src/cli/utilities/theme-fs.ts#L227-L230)

PR #4436, opened 2024-09-11
Add specific error for invalid_target error

itsjustriley:

Resolves [#271](https://github.com/Shopify/develop-advanced-edits/issues/271)

When a user has an invalid store url, or the wrong permissions, when using theme dev in the cli they currently receive an ambiguous `invalid_target` error message. 
 

Replacing ambiguous `invalid_target` error with specific messaging. (Modified an existing error message to work in the pertinent file, added handling for 'invalid_target' specifically).

image

paired with @EvilGenius13 @AribaRajput @lukeh-shopify 

### How to test your changes?

1. Build the updated CLI `pnpm build`
2. `shopify-dev theme dev --store wrong_store`
3. Authenticate
4. Confirm updated error message is displayed in terminal

### Measuring impact

### Checklist

### DIFF
--- a/packages/cli-kit/src/private/node/session/exchange.ts
+++ b/packages/cli-kit/src/private/node/session/exchange.ts
@@ -11,6 +11,7 @@ import * as jose from 'jose'
 
 export class InvalidGrantError extends ExtendableError {}
 export class InvalidRequestError extends ExtendableError {}
+class InvalidTargetError extends AbortError {}
 
 export interface ExchangeScopes {
   admin: string[]
@@ -169,6 +170,17 @@ interface TokenRequestResult {
 }
 
 function tokenRequestErrorHandler(error: string) {
+  const invalidTargetErrorMessage =
+    'You are not authorized to use the CLI to develop in the provided store.' +
+    '\n\n' +
+    "You can't use Shopify CLI with development stores if you only have Partner " +
+    'staff member access. If you want to use Shopify CLI to work on a development store, then ' +
+    'you should be the store owner or create a staff account on the store.' +
+    '\n\n' +
+    "If you're the store owner, then you need to log in to the store directly using the " +
+    'store URL at least once before you log in using Shopify CLI.' +
+    'Logging in to the Shopify admin directly connects the development ' +
+    'store with your Shopify login.'
   if (error === 'invalid_grant') {
     // There's an scenario when Identity returns "invalid_grant" when trying to refresh the token
     // using a valid refresh token. When that happens, we take the user through the authentication flow.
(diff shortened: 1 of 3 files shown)


packages/cli-kit/src/private/node/session/exchange.ts
```diff
@@ -167,18 +167,32 @@ interface TokenRequestResult {
...
     // using a valid refresh token. When that happens, we take the user through the authentication flow.
     return new InvalidGrantError()
   }
-  if (error === 'invalid_request') {
+  if (errorMessage === 'invalid_request') {
     // There's an scenario when Identity returns "invalid_request" when exchanging an identity token.
     // This means the token is invalid. We clear the session and throw an error to let the caller know.
     return new InvalidRequestError()
   }
+  if (errorMessage === 'invalid_target') {
```
jamesmengo, line 181:
```ts
function tokenRequestErrorHandler(error: string) {
  if (error === 'invalid_grant') {
    // There's an scenario when Identity returns "invalid_grant" when trying to refresh the token
    // using a valid refresh token. When that happens, we take the user through the authentication flow.
    return new InvalidGrantError()
  }
  if (error === 'invalid_request') {
    // There's an scenario when Identity returns "invalid_request" when exchanging an identity token.
    // This means the token is invalid. We clear the session and throw an error to let the caller know.
    return new InvalidRequestError()
  }
  if (error === 'invalid_target') {
  // add a comment here
    return new InvalidTargetError(INVALID_TARGET_ERROR_MESSAGE)
  }
  // eslint-disable-next-line @shopify/cli/no-error-factory-functions
  return new AbortError(error)
}
```

jamesmengo: ^ can we try keeping these patterns as is?

jamesmengo: Ah ok thanks for the context! That's a great observation that I wasn't aware of :) 

As long as the message is in a constant, I'm happy :)

gonzaloriestra: I agree with extracting the message to a constant. And maybe there's no need to create a new error, you could just do something like `return new AbortError(INVALID_TARGET_ERROR_MESSAGE)`.

gonzaloriestra (approved):
LGTM, Thanks!

packages/cli-kit/src/private/node/session/exchange.ts
```diff
@@ -179,6 +191,9 @@ function tokenRequestErrorHandler(error: string) {
     // This means the token is invalid. We clear the session and throw an error to let the caller know.
     return new InvalidRequestError()
   }
+  if (error === 'invalid_target') {
+    return new InvalidTargetError(invalidTartgetErrorMessage)
```
EvilGenius13, line 195:
```suggestion
    return new InvalidTargetError(invalidTargetErrorMessage)
```

EvilGenius13 (approved):

PR #4439, opened 2024-09-11
[Themes] - Switch the implementation of `theme` commands to run the TypeScript by default

jamesmengo:
Closes https://github.com/Shopify/develop-advanced-edits/issues/316

Replaces the `dev-preview` flag with an inverse `stable` flag so that we default to Typescript implementation

### How to test your changes?
- Try running each command with and without the `stable` flag

### Post-Release Steps
- Communications

### Measuring impact

### Checklist


packages/app/src/cli/commands/app/logs.ts
```diff
@@ -37,7 +37,7 @@ export default class Logs extends Command {
     'no-tunnel': Dev.flags['no-tunnel'],
     'graphiql-port': Dev.flags['graphiql-port'],
     'graphiql-key': Dev.flags['graphiql-key'],
-    'dev-preview': Dev.flags['dev-preview'],
+    stable: Dev.flags.stable,
```
jamesmengo, line 40: I asked how this works [here](https://shopify.slack.com/archives/C030LHFCA5T/p1726092713391289).

packages/theme/src/cli/commands/theme/push.ts
```diff
@@ -155,7 +149,7 @@ export default class Push extends ThemeCommand {
       return
     }
 
-    if (flags['dev-preview'] || (!flags.stable && !flags.password)) {
+    if (!flags.stable && !flags.password) {
```
jamesmengo, line 152: I'm keeping the `password` check for now, meaning that if the `password` flag is provided we will still use the Ruby CLI.

This check was added to avoid breaking the `push` command for CI users - more context [here](https://github.com/Shopify/cli/pull/3765)

I think the changes introduced in the linked PR above should address those issues, however since there are numerous changes here and many users 'set and forget' their CI configs, I think it would be prudent to de-risk this PR by leaving this here.

No strong opinions on this though

packages/app/src/cli/commands/app/logs.ts
karreiro: I've answered this in the thread below :)

karreiro (approved):
Thank you for this PR, @jamesmengo! It looks great! I've made a minor commit to update the documentation and rename the flag to `--legacy` ;)

PR #4443, opened 2024-09-12
Cache user info

isaacroldan:
We want to cache data that rarely changes and avoid the extra API calls

Fixes https://github.com/Shopify/develop-app-inner-loop/issues/2026

- Cache the `hasPartnerAccount` value. For a given userId, cache it forever if true.
- Cache user account information (email, etc..) for 72h. We were fetching this on every command, but rarely changes.
  - Created a new `app-conf-store.ts` to handle this 

### How to test your changes?
Running `app info --verbose` should show a call to fetch an organization and another to fetch the user info.
If you run it again, you should see that those queries are no longer executed (we are reading from cache)

### Measuring impact

### Checklist

### DIFF
--- a/packages/cli-kit/src/private/node/conf-store.test.ts
+++ b/packages/cli-kit/src/private/node/conf-store.test.ts
@@ -9,6 +9,7 @@ import {
   getConfigStoreForPartnerStatus,
   getCachedPartnerAccountStatus,
   setCachedPartnerAccountStatus,
+  runWithRateLimit,
 } from './conf-store.js'
 import {LocalStorage} from '../../public/node/local-storage.js'
 import {afterEach, beforeEach, describe, expect, test, vi} from 'vitest'
@@ -195,7 +196,7 @@ describe('runAtMinimumInterval', () => {
 
       // When
       let taskRan = false
-      await runAtMinimumInterval(
+      const got = await runAtMinimumInterval(
         key,
         timeout,
         async () => {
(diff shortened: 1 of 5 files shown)


packages/app/src/cli/services/dev/fetch.ts
```diff
@@ -91,7 +91,8 @@ export async function fetchOrganizations(): Promise<Organization[]> {
 
   if (organizations.length === 0) {
     const developerPlatformClient = selectDeveloperPlatformClient()
-    const accountInfo = await fetchCurrentAccountInformation(developerPlatformClient)
+    const userId = await developerPlatformClient.session().then((session) => session.userId)
```
shauns, line 94: lets use await only, not then in this case

packages/app/src/cli/utilities/app-conf-store.test.ts
```diff
@@ -0,0 +1,91 @@
...
+    vi.setSystemTime(new Date('2023-01-01'))
+    await inTemporaryDirectory(async (dir) => {
+      tempDir = dir
+      process.env.XDG_CONFIG_HOME = joinPath(tempDir, '.config')
+    })
+  })
+
+  afterEach(() => {
+    vi.useRealTimers()
+    delete process.env.XDG_CONFIG_HOME
```
shauns, line 27: this may prove flaky -- any other options? can we inject preference (config folder maybe? or process.env?) into the config functions?

shauns (approved):

PR #4445, opened 2024-09-12
Functions 7452: Bundling and uploading wasm file to GCS for function extension

saga-dasgupta:
Issue: https://github.com/Shopify/script-service/issues/7452 

To have a consistent way of uploading and obtaining extension data is core using the App Module Framework. 

This PR adds the "bundling" feature to functions and sets the output path for compilation to the bundle directory to ensure the wasm file generated is zipped and uploaded to GCS like every other extension. 

The deployment plan is:
This version of CLI goes out which bundles the wasm file and uploads it (existing wasm blob upload stays in place). 
Then on Core's side we verify we are able to obtain the wasm content and upload it to our runtime-engine bucket. Then we come back here and delete the uploadWasmBlob path and in core we throw an error when we fail to upload. TLDR: existing upload path stays in place till we can verify new path works. 

### How to test your changes?
Tophat this normally from your local computer since core changes are out. Run cli for dev and deploy both for an app with functions

### Measuring impact

### Checklist


packages/app/src/cli/models/extensions/extension-instance.ts
```diff
@@ -335,11 +336,13 @@ export class ExtensionInstance<TConfiguration extends BaseConfigType = BaseConfi
 
     if (this.features.includes('bundling')) {
       // Modules that are going to be inclued in the bundle should be built in the bundle directory
-      this.outputPath = joinPath(bundleDirectory, extensionId, outputFile)
+      this.outputPath = this.isFunctionExtension ? this.outputPath : joinPath(bundleDirectory, extensionId, outputFile)
```
saga-dasgupta, line 339: Can't modify outputpath which is the function's directory before compilation.

packages/app/src/cli/services/deploy.test.ts
```diff
@@ -243,7 +243,7 @@ describe('deploy', () => {
     expect(updateAppIdentifiers).toHaveBeenCalledOnce()
   })
 
-  test('uploads the extension bundle with 1 function', async () => {
+  test('uploads the extension bundle with 1 function extension', async () => {
```
saga-dasgupta, line 246: Keeping the test names consistent.

packages/app/src/cli/services/deploy.test.ts
```diff
@@ -290,7 +290,7 @@ describe('deploy', () => {
       ],
       developerPlatformClient,
       extensionIds: {},
-      bundlePath: undefined,
+      bundlePath: expect.stringMatching(/bundle.zip$/),
```
saga-dasgupta, line 293: Bundle path now has a zip file for functions' assets.

packages/app/src/cli/models/extensions/extension-instance.ts
saga-dasgupta: Unfortunately the outputPath ends up pointing to a tmp directory so unless we are okay with not having the wasm file generated in the app's extension directory (which also might have unforeseen consequences, I haven't tried it) I think we need to avoid changing the outputPath if it is a function type extension.

andrewhassan: It looks like `outputPath` is used to know where the build artifact should be located so that it can be built/deployed/bundled/whatever. I think the problem is that this path is defined by the developer via the TOML file (since they control the build tooling), so we can't really change it ourselves.

One idea to work around this is that we could not change the `outputPath` for any extension. Then once the artifact is produced at that path, we move it into the bundle directory. This assumes that there's a single artifact produced for building an extension, and it's located at `outputPath`, but I'm not sure that's a safe assumption.

```diff
diff --git a/packages/app/src/cli/models/extensions/extension-instance.ts b/packages/app/src/cli/models/extensions/extension-instance.ts
index addb7eba72..928fd80039 100644
--- a/packages/app/src/cli/models/extensions/extension-instance.ts
+++ b/packages/app/src/cli/models/extensions/extension-instance.ts
@@ -334,15 +334,13 @@ export class ExtensionInstance<TConfiguration extends BaseConfigType = BaseConfi
     const extensionId = identifiers?.extensions[this.localIdentifier] ?? this.configuration.uid ?? this.handle
     const outputFile = this.isThemeExtension ? '' : joinPath('dist', `${this.outputFileName}`)
 
-    if (this.features.includes('bundling')) {
-      // Modules that are going to be inclued in the bundle should be built in the bundle directory
-      this.outputPath = this.isFunctionExtension ? this.outputPath : joinPath(bundleDirectory, extensionId, outputFile)
-    }
-
     await this.build(options)
-    if (this.isFunctionExtension) {
-      await copyFile(this.outputPath, joinPath(bundleDirectory, extensionId, 'index.wasm'))
+
+    if (this.features.includes('bundling')) {
+      const bundleOutputPath = joinPath(bundleDirectory, extensionId, outputFile)
+      await moveFile(this.outputPath, bundleOutputPath)
     }
+
     if (this.isThemeExtension && useThemebundling()) {
       await bundleThemeExtension(this, options)
     }
~
~
```

Alternatively, if my assumption is not true, then we could have the `buildFunctionExtension` function move the file to where it needs to be. We'd need to pass in the final location of the file though (maybe through the `buildDirectory` option?). For example:

```diff
diff --git a/packages/app/src/cli/services/build/extension.ts b/packages/app/src/cli/services/build/extension.ts
index bf81e275fb..b980baa598 100644
--- a/packages/app/src/cli/services/build/extension.ts
+++ b/packages/app/src/cli/services/build/extension.ts
@@ -146,6 +146,9 @@ export async function buildFunctionExtension(
     } else {
       await buildOtherFunction(extension, options)
     }
+
+    const buildPath = joinPath(extension.directory, extension.config.build.path ?? joinPath('dist', 'index.wasm'))
+    await copyFile(buildPath, options.buildDirectory)
   } finally {
     await releaseLock()
   }
```

@isaacroldan WDYT?

saga-dasgupta: Works for me, thank you!

packages/app/src/cli/models/extensions/extension-instance.ts
```diff
@@ -205,6 +205,7 @@ export class ExtensionInstance<TConfiguration extends BaseConfigType = BaseConfi
     apiKey,
     developerPlatformClient,
   }: ExtensionDeployConfigOptions): Promise<{[key: string]: unknown} | undefined> {
+    // To ensure this is not a breaking change we will delete this upload line on a second pass after we've ensured the new file upload process from core is working.
```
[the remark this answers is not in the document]
saga-dasgupta: >so we'll do both uploads GCS+this one?
Only till I am able to verify prod bucket access works so that would be till core is deployed and this cli version is deployed.
>Ping me on the cleanup PR as well, there is a few things we can remove after deleting this line :)
Absolutely will do, can also add you as a reviewer on my cleanup PR! :)

packages/app/src/cli/services/dev/update-extension.ts
```diff
@@ -50,12 +50,20 @@ export async function updateExtensionDraft({
...
+      uploaded_files: {'index.wasm': compiledFiles},
+    }
+  }
   const extensionInput: ExtensionUpdateDraftMutationVariables = {
     apiKey,
-    config: JSON.stringify({
-      ...config,
-      serialized_script: encodedFile,
-    }),
+    config: JSON.stringify(draftableConfig),
```
isaacroldan, line 66: Does this mean we can remove the custom wasm update then? or are we waiting until this is released and used for a bit?

saga-dasgupta: Yes, the latter!

packages/app/src/cli/services/deploy/bundle.test.ts
```diff
@@ -107,4 +107,35 @@ describe('bundleAndBuildExtensions', () => {
       await expect(file.fileExists(bundlePath)).resolves.toBeTruthy()
     })
   })
+
+  test('creates a zip file with wasm file in it for a function extension', async () => {
```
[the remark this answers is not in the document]
saga-dasgupta: What is a good way to unzip files in cli? I didn't find any prior art for this.

packages/app/src/cli/services/build/extension.ts
```diff
@@ -141,11 +142,18 @@ export async function buildFunctionExtension(
   }
 
   try {
+    const bundlePath = joinPath(extension.outputPath.split('/').slice(0, -2).join('/'), 'index.wasm')
+    extension.outputPath = joinPath(extension.directory, joinPath('dist', 'function.wasm'))
```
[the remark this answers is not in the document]
saga-dasgupta: Yes, `outputPath` ends up becoming `<bundlepath>/dist/something.js` and since `<bundlepath>/dist` does not exists it complains `No such file or directory` (remember that bundlePath is a /tmp directory). So, I essentially set it to the valid path that exists which is `<extension directory>/dist`.

packages/app/src/cli/services/build/extension.ts
```diff
@@ -141,11 +142,18 @@ export async function buildFunctionExtension(
   }
 
   try {
+    const bundlePath = joinPath(extension.outputPath.split('/').slice(0, -2).join('/'), 'index.wasm')
```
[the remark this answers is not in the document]
isaacroldan: I don't think we need to do this actually. 
just do
```ts
const bundlePath = extension.outputPath
```

and in `extension-instance.ts`, in the `get outputFileName()` function, do:
```ts
    return this.isFunctionExtension ? 'index.wasm' : `${this.handle}.js`
```

This is cleaner, and we are OK with having `isFunctionExtension` inside the `extension-instance` object

packages/app/src/cli/services/build/extension.ts
```diff
@@ -141,11 +142,18 @@ export async function buildFunctionExtension(
...
+    extension.outputPath = joinPath(extension.directory, joinPath('dist', 'function.wasm'))
     if (extension.isJavaScript) {
       await runCommandOrBuildJSFunction(extension, options)
     } else {
       await buildOtherFunction(extension, options)
     }
+    if (fileExistsSync(extension.outputPath)) {
+      const base64Contents = await readFile(extension.outputPath, {encoding: 'base64'})
+      await touchFile(bundlePath)
+      await writeFile(bundlePath, base64Contents)
```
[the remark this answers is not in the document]
saga-dasgupta: Yeah thats how I had it before but I had to change that to base64 encode this because I had to base64 encode the wasm file that I'm sending over the network for draftable dev (utf-8 encoding doesn't work for wasm files) and so the code obtaining the file in Core now expects it to be bae64 encoded.

packages/app/src/cli/services/build/extension.ts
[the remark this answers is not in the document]
saga-dasgupta: Got it, thanks! I updated and verified that works!

packages/app/src/cli/services/build/extension.ts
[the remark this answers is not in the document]
saga-dasgupta: Yes right, exactly, so then we can go back and simplify this.

packages/app/src/cli/services/build/extension.ts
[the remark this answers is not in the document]
isaacroldan: I wonder if we should remove `dist` from the output when using consistent-dev, there is really no need 🤔

saga-dasgupta:
@isaacroldan @jacobsteves I added [a commit ](https://github.com/Shopify/cli/pull/4445/commits/193ebc452f65705d574685e3ee1ea3d7e5ede8de) to remove `uploadWasmBlob` and any associated logic because after my Core changes deployed I learnt that the production service account role core uses did not have an update/overwrite file access so CLI can't be writing the same file to the bucket. So I had to add the cleanup in this PR itself but that is okay since folks using this version of CLI would bundling/uploading or sending the wasm file as text over the network so it works in sync and harmony with Core. I tophatted it all.

packages/app/src/cli/services/dev/update-extension.ts
```diff
@@ -55,10 +55,10 @@ export async function updateExtensionDraft({
     serialized_script: encodedFile,
   }
   if (extension.isFunctionExtension) {
-    const compiledFiles = await readFile(extension.outputPath, {encoding: 'base64'})
+    const compiledFiles = await readFile(extension.outputPath)
```
saga-dasgupta, line 58: Do not need the base64 encoding because for this draftable flow, the bundlePath ends up being the same as `extension.outputPath` and we already have logic in place to write base64 encoded data to the bundlepath, so outputpath here already ends up having base64 encoded content.

packages/app/src/cli/models/extensions/specifications/function.test.ts
```diff
@@ -5,17 +5,16 @@ import {
...
 import {joinPath} from '@shopify/cli-kit/node/path'
 import {AbortError} from '@shopify/cli-kit/node/error'
 import {beforeEach, describe, expect, test, vi} from 'vitest'
 import {getPathValue} from '@shopify/cli-kit/common/object'
-
-vi.mock('../../../services/deploy/upload.js')
+import {randomUUID} from '@shopify/cli-kit/node/crypto'
 
 const developerPlatformClient: DeveloperPlatformClient = testDeveloperPlatformClient()
+vi.mock('@shopify/cli-kit/node/crypto')
```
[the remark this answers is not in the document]
saga-dasgupta: It helps with json config assertions if I can mock and tell it exactly what module ID to give.

packages/app/src/cli/models/extensions/extension-instance.ts
```diff
@@ -194,18 +193,14 @@ export class ExtensionInstance<TConfiguration extends BaseConfigType = BaseConfi
...
 
-  async functionDeployConfig({
-    apiKey,
-    developerPlatformClient,
-  }: ExtensionDeployConfigOptions): Promise<{[key: string]: unknown} | undefined> {
-    const {moduleId} = await uploadWasmBlob(this.localIdentifier, this.outputPath, developerPlatformClient)
+  async functionDeployConfig(apiKey: string): Promise<{[key: string]: unknown} | undefined> {
+    const moduleId = await randomUUID()
     return this.specification.deployConfig?.(this.configuration, this.directory, apiKey, moduleId)
   }
```
isaacroldan, line 205: We need to try and remove this function completely (`functionDeployConfig`) and make the moduleId generation part of the `function.ts` specification

saga-dasgupta: Good catch, yes that is straight forward enough.

packages/app/src/cli/models/extensions/extension-instance.ts
```diff
@@ -194,21 +193,11 @@ export class ExtensionInstance<TConfiguration extends BaseConfigType = BaseConfi
 
   async deployConfig({
     apiKey,
-    developerPlatformClient,
     appConfiguration,
   }: ExtensionDeployConfigOptions): Promise<{[key: string]: unknown} | undefined> {
-    if (this.isFunctionExtension) return this.functionDeployConfig({apiKey, developerPlatformClient, appConfiguration})
     return this.commonDeployConfig(apiKey, appConfiguration)
```
isaacroldan, line 198: This is looking good!!!

I'll ask for one more change, since now `deployConfig` is basically doing nothing and just calling `commonDeployConfig` can we unify both?

Have a single `deployConfig` that does whatever `commonDeployConfig` does.
And update all calls to `commonDeployConfig` to call `deployConfig` instead. Should be straightforward :)

packages/app/src/cli/services/build/extension.ts
```diff
@@ -141,11 +142,18 @@ export async function buildFunctionExtension(
...
+    extension.outputPath = joinPath(extension.directory, joinPath('dist', 'function.wasm'))
     if (extension.isJavaScript) {
       await runCommandOrBuildJSFunction(extension, options)
     } else {
       await buildOtherFunction(extension, options)
     }
+    if (fileExistsSync(extension.outputPath) && bundlePath !== extension.outputPath) {
+      const base64Contents = await readFile(extension.outputPath, {encoding: 'base64'})
+      await touchFile(bundlePath)
+      await writeFile(bundlePath, base64Contents)
```
andrewhassan, line 155: Could we use `copyFile` here instead of reading and writing?

andrewhassan (approved):
I didn't 🎩 but overall the code LGTM! Nice work!

packages/app/src/cli/services/build/extension.ts
saga-dasgupta: I did have `copyFile` here before but I need it to be in `base64` format inside the bundlePath because code in core expects this to be base64 formatted, so `copyFIle` which copies file as is doesn't work here.

packages/app/src/cli/services/dev/update-extension.ts
```diff
@@ -47,15 +47,23 @@ export async function updateExtensionDraft({
...
+  let draftableConfig: {[key: string]: unknown} = {
+    ...config,
+    serialized_script: encodedFile,
+  }
+  if (extension.isFunctionExtension) {
+    const compiledFiles = await readFile(extension.outputPath, {encoding: 'base64'})
+    draftableConfig = {
+      ...draftableConfig,
+      uploaded_files: {'dist/index.wasm': compiledFiles},
+    }
```
isaacroldan, line 62: I believe this can be just:
```
draftableConfig.uploaded_files = {'dist/index.wasm': compiledFiles}
```

isaacroldan (approved):
Awesome work @saga-dasgupta, thank you so much for all the extra clean up 🙇

PR #4455, opened 2024-09-17
[Theme] Show error when returned theme ID does not match

frandiox:

Closes https://github.com/Shopify/develop-advanced-edits/issues/329

~~Right now this only shows an error like any other Network error. Should we stop the server completely if this is the case, even if it's just an intermitent issue (last time it was 10% of requests)?~~
The server stops when the themeId doesn't match:

image

### How to test your changes?

The easiest way to test is direclty changing the `expectedThemeId` variable to something else.

cc @Shopify/advanced-edits


packages/theme/src/cli/utilities/theme-environment/html.ts
```diff
@@ -98,3 +102,20 @@ function getErrorPage(options: {title: string; header: string; message: string;
     </body>
   </html>`
 }
+
+function assertThemeId(response: Response, html: string, expectedThemeId: string) {
+  const obtainedThemeId = html.match(/Shopify\.theme\s*=\s*{[^}]+?"id":\s*"?(\d+)"?(}|,)/)?.[1]
```
graygilmore, line 107: I wonder if it's worth adding a comment above this to explain what we're matching against? Something like:

```js
// Match against the `Shopify.theme` object that is defined in the markup like this:
// Shopify.theme = {"name":"Dawn","id":1234,"theme_store_id":null,"role":"development"};
```

packages/theme/src/cli/utilities/theme-environment/html.ts
```diff
@@ -29,6 +31,8 @@ export function getHtmlHandler(theme: Theme, ctx: DevServerContext) {
 
         html = prettifySyntaxErrors(html)
 
+        assertThemeId(response, html, String(theme.id))
```
graygilmore, line 34:
```
This is likely related to an issue in upstream Shopify APIs.
```

Is this part of the error message true? What's the most common way to hit this?

graygilmore (approved):

packages/theme/src/cli/utilities/theme-environment/html.ts
```diff
@@ -98,3 +102,33 @@ function getErrorPage(options: {title: string; header: string; message: string;
...
+   *
+   * ```
+   * <script>var Shopify = Shopify || {};
+   * Shopify.locale = "en";
+   * Shopify.theme = {"name":"Development","id":143509762348,"theme_store_id":null,"role":"development"};
+   * Shopify.theme.handle = "null";
+   * ...;</script>
+   * ```
+   */
+  const obtainedThemeId = html.match(/Shopify\.theme\s*=\s*{[^}]+?"id":\s*"?(\d+)"?(}|,)/)?.[1]
```
karreiro, line 118: One alternative way of implementing this would be relying on the JavaScript runtime.

So, we could perform this logic at `packages/theme/src/cli/utilities/theme-environment/hot-reload/client.ts` and the client would ping the server, letting it know that the process should be terminated because the theme ID is different.

I slightly prefer the JS runtime approach because we [support](https://shopify.dev/docs/api/ajax/reference/cart) the `Shopify` global object, while the shape of the declaration might change.

Still, the current approach works as expected, and the shape of the declaration is unlikely to change, so feel free to keep it as is if you prefer this approach more, @frandiox. One benefit of the current approach is that it's more self-contained.

karreiro (approved):
Thank you for this PR, @frandiox! I've left only one optional comment :)

PR #4458, opened 2024-09-17
[Theme] Fix refreshing empty, deleted and new files

frandiox:

Closes https://github.com/Shopify/develop-advanced-edits/issues/308
Closes https://github.com/Shopify/develop-advanced-edits/issues/301

### How to test your changes?

Each of the following steps should trigger a hot reload or a full refresh, depending on the information we have:

1. Make a CSS file empty and save => CSS is lost
1. Restore it => back to normal
1. Delete a CSS file => CSS is lost, and there's a 404 in network tab
1. Restore it => back to normal
1. Make a section file empty and save => section is gone
1. Restore it => back to normal
1. Delete a section file => section is gone
1. Restore it => back to normal
1. Delete any other file (template, JS asset) => full refresh

cc @Shopify/advanced-edits


packages/theme/src/cli/utilities/theme-environment/hot-reload/server.ts
```diff
@@ -187,12 +185,19 @@ export function getHotReloadHandler(theme: Theme, ctx: DevServerContext) {
...
+        const sectionTemplate = ctx.localThemeFileSystem.files.get(sectionKey)?.value
+        if (!sectionTemplate) {
+          // If the section template is not found, it means that the section has been removed.
+          // The remote version might not yet be synced so, instead of rendering it remotely,
+          // which should return an empty section, we directly return the same thing here.
+          return ''
+        }
+
+        replaceTemplates[sectionKey] = sectionTemplate
+      }
```
frandiox, line 200: @jamesmengo This is what we were missing during our pairing to fix the empty files. The server was falling back to the remote renderer because `value` was empty, and we were not passing it in `replaceTemplates`.

frandiox (commented):
@jamesmengo In order to fix deleted files hot reload, I had to fix it for empty files as well, so this PR also covers the initial work you were doing today 🙇

packages/theme/src/cli/utilities/theme-fs.ts
```diff
@@ -190,13 +189,15 @@ export function mountThemeFileSystem(root: string, options?: ThemeFileSystemOpti
...
-    unsyncedFileKeys.delete(fileKey)
+    // Optimistically delete the file from the local file system.
     files.delete(fileKey)
+    unsyncedFileKeys.add(fileKey)
     emitEvent('unlink', {fileKey})
 
     deleteThemeAsset(Number(themeId), fileKey, adminSession)
       .then(async (success) => {
         if (!success) throw new Error('Unknown issue.')
+        unsyncedFileKeys.delete(fileKey)
```
frandiox, line 200: With this change, there's now a new possible state: a file is not present in `themeFS.files` but it's still present in `themeFS.unsyncedFileKeys`. This signals that we had a file locally before but we have removed it optimistically while requesting the same thing to the remote SFR instance.

packages/theme/src/cli/utilities/theme-environment/hot-reload/server.ts
```diff
@@ -81,7 +78,7 @@ export function getInMemoryTemplates(ctx: DevServerContext, currentRoute?: strin
       } else if (!fileKey.includes('.default.')) continue
     }
 
-    inMemoryTemplates[fileKey] = content
+    inMemoryTemplates[fileKey] = ctx.localThemeFileSystem.files.get(fileKey)?.value ?? ''
```
jamesmengo, line 81: Q for curiosity: What would happen when we delete the file locally and are waiting for the remote file to be deleted?

I believe `files.get(fileKey)` would be undefined in this case. Do we still want to return an empty string? Would an early return work as well?

frandiox: >I believe files.get(fileKey) would be undefined in this case

Exactly, since it's not removed in the remote renderer yet, we pass empty string from here to make it behave at least like an empty file. For example, this way the remote renderer won't render sections that have been removed locally, even if the section file still exists in the remote.

Issue #4470, opened 2024-09-18
[Bug]: Can't pull theme anymore


ThomasB-Sunology:
### Please confirm that you have:

- [x] Searched [existing issues](.) to see if your issue is a duplicate. (If you’ve found a duplicate issue, feel free to add additional information in a comment on it.)
- [x] Reproduced the issue in [the latest CLI version](https://www.npmjs.com/package/@shopify/cli).

### In which of these areas are you experiencing a problem?

Theme

### Expected behavior

I should pull theme from a store.

### Actual behavior

It crash when I try that when I'm at around 17%.

![Image](https://github.com/user-attachments/assets/b97e2d70-7eaa-4003-adfd-d17c50f09148)

### Verbose output

2024-09-18T19:48:28.760Z: Running command theme pull
2024-09-18T19:48:28.770Z: Ensuring that the user is authenticated with the Theme API with the following scopes:
[]

2024-09-18T19:48:28.770Z: Ensuring that the user is authenticated with the Admin API with the following scopes for the store shopsunology.myshopify.com:
[]

2024-09-18T19:48:28.770Z: Getting session store...
2024-09-18T19:48:28.772Z: Validating existing session against the scopes:
[
  "openid",
  "https://api.shopify.com/auth/shop.admin.graphql",
  "https://api.shopify.com/auth/shop.admin.themes",
  "https://api.shopify.com/auth/partners.collaborator-relationships.readonly",
  "https://api.shopify.com/auth/shop.storefront-renderer.devtools",
  "https://api.shopify.com/auth/partners.app.cli.access",
  "https://api.shopify.com/auth/destinations.readonly"
]
For applications:
{
  "adminApi": {
    "scopes": [],
    "storeFqdn": "shopsunology.myshopify.com"
  }
}

2024-09-18T19:48:28.772Z: Sending Identity Introspection request to URL: https://accounts.shopify.com/oauth/introspection
2024-09-18T19:48:28.772Z: Sending POST request to URL https://accounts.shopify.com/oauth/introspection
With request headers:
 - User-Agent: Shopify CLI; v=3.67.0
 - Keep-Alive: timeout=30
 - Sec-CH-UA-PLATFORM: darwin
 - Content-Type: application/json

2024-09-18T19:48:29.009Z: Request to https://accounts.shopify.com/oauth/introspection completed in 237 ms
With response headers:
 - cache-control: no-cache, no-store, private, must-revalidate, max-age=0
 - content-type: application/json; charset=utf-8
[...]

PR #4475, opened 2024-09-18
[Themes] Account for empty assets when calculating theme asset size

jamesmengo:
Fixes https://github.com/Shopify/develop-advanced-edits/issues/339
Fixes https://github.com/Shopify/cli/issues/4470

- Add a fallback to calculate size when files are empty

### How to test your changes?
- Add an empty file to your theme via admin
- Run `shopify theme pull -t <THEME>`

### Measuring impact

### Checklist

### DIFF
--- a/packages/cli-kit/src/public/node/themes/factories.ts
+++ b/packages/cli-kit/src/public/node/themes/factories.ts
@@ -12,7 +12,8 @@ interface RemoteThemeResponse {
 interface RemoteAssetResponse {
   key: string
   checksum: string
-  attachment: string
+  attachment: string | undefined
+  // value is an empty string ('') when the file is empty
   value: string
 }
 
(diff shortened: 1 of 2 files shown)


packages/cli-kit/src/public/node/themes/factories.ts
```diff
@@ -52,7 +52,9 @@ export function buildThemeAsset(asset?: RemoteAssetResponse): ThemeAsset | undef
 
   const {key, checksum, attachment, value} = asset
   // Note: for attachments, this is the size of the base64 string, not the real length of the file
-  const stats = {size: (value || attachment).length, mtime: Date.now()}
```
frandiox, line 55: Oh, I thought they'd be empty string at this point. You can also just do:

```ts
  const stats = {size: (value || attachment || '').length, mtime: Date.now()}
```

frandiox (approved):

packages/cli-kit/src/public/node/themes/factories.ts
frandiox: In fact, the types suggest these are already strings... I guess we are passing undefined as `any` directly from what we get from the cloud, thus bypassing TS checks?

karreiro: Indeed, possibly making `value` and `attachment` nullable in the `RemoteAssetResponse` would represent a bit better their meaning.

karreiro (approved):
Thank you for this PR, @jamesmengo! Great stuff :)

packages/cli-kit/src/public/node/themes/factories.ts
jamesmengo: `size` is an empty string when the file is empty, so I changed this expression to `value ?? attachment` which returns `''`

`value || attachment` was returning `undefined`

Issue #4477, opened 2024-09-19
[Bug]: Could not use `theme dev` since CLI 3.67


bakura10:
### Please confirm that you have:

- [x] Searched [existing issues](.) to see if your issue is a duplicate. (If you’ve found a duplicate issue, feel free to add additional information in a comment on it.)
- [x] Reproduced the issue in [the latest CLI version](https://www.npmjs.com/package/@shopify/cli).

### In which of these areas are you experiencing a problem?

Theme

### Expected behavior

Hi,

After upgrading to Shopify CLI 3.67, the CLI asked for a new "store-password" for password protected storefront. I've entered it, but now I am facing a different issue that I can't understand:

```
Your development session could not be created because the "_shopify_essential" could not be defined. Please, check your internet connection.
```

Reverting to Shopify CLI 3.66 fixes the issue.
EDIT: might be related to https://github.com/Shopify/cli/issues/4471, but the error I get is different.

### Actual behavior

I can't use dev commands for theme.

Please note that I'm using the --environment option to pull the env data from a `shopify.theme.toml` file

### Verbose output

Request to https://theme-kit-access.shopifyapps.com/cli/sfr?preview_theme_id=158896423231&_fd=0&pb=0 completed in 296 ms
With response headers:
 - cache-control: no-cache
 - content-type: application/json; charset=utf-8
 - server-timing: processing;dur=6, socket_queue;dur=8.119, cfRequestDuration;dur=215.000153
 - x-request-id: f19074d8-2ee4-44d1-b664-db8129a3a277-1726734286

Error

Your development session could not be created because the "_shopify_essential" could not be defined. Please, check your internet connection.

### Reproduction steps

1. Use `shopify theme dev` command

### Operating System

OSX 15

### Shopify CLI version (check your project's `package.json` if you're not sure)

3.67

karreiro:
Thank you for reporting this issue, @bakura10 @darrencraig! We're taking a look at it.

While that's not fixed, could you please confirm if the `--legacy` flag works as a temporary workaround for you? This flag will activate the previous implementation for the command and should unblock your workflow.

Thank you for reporting!

felixmpaulus:
I am encountering the same issue when starting the dev environment for an app. 

```
"scripts": {
    "dev": ". ~/.nvm/nvm.sh && nvm use 20 && shopify app dev"
}
```

producing these logs:

```
felixpaulus@Felixs-MBP App % npm run dev                          

[...]

PR #4490, opened 2024-09-20
Fix Theme Access authentication on `shopify theme dev` and `shopify theme console` commands

karreiro:

Fixes https://github.com/Shopify/cli/issues/4477

The `shopify theme console` command was failing since [this change](https://github.com/Shopify/cli/pull/4376), because the Theme Access requests need the `Authorization` header.

This PR now just passes the `Authorization` header as we already have it at the moment the session is initialized.

This PR also avoids propagating the content-type of the SFR response, because when it's being proxied by the Theme Access app, that content-type is set to `application/json` (we're not introducing any branching code here, to protect the CLI from future changes, and in the end, the content-type should never be `application/json`).

### How to test your changes?

Use the following command passing the Theme Access token:

- Install the Theme Access app in your store
- Generate a  Theme Access app password
- Run `shopify theme dev --password <your_theme_access_password> --store <your_store>`
- Run `shopify theme console --password <your_theme_access_password> --store <your_store>`

**Before:**
Screenshot 2024-09-20 at 08 11 33

**After:**
Screenshot 2024-09-20 at 08 11 04

### Post-release steps

N/a

### Measuring impact

### Checklist

### DIFF
--- a/.changeset/hip-chefs-dance.md
+++ b/.changeset/hip-chefs-dance.md
@@ -1,0 +1,5 @@
+---
+'@shopify/theme': patch
+---
+
+Fix Theme Access authentication on `shopify theme dev` and `shopify theme console` commands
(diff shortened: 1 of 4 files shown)


frandiox (approved):
Thanks for the quick fix, it works well 🎉

PR #4494, opened 2024-09-20
[Theme] Fix serving local assets that contain non-printable characters

### LINKED ISSUE #4479: [Bug]: JS assets no longer work on `theme dev` on 127.0.0.1 since 3.67.0

- [x] Searched [existing issues](.) to see if your issue is a duplicate. (If you’ve found a duplicate issue, feel free to add additional information in a comment on it.)
- [x] Reproduced the issue in [the latest CLI version](https://www.npmjs.com/package/@shopify/cli).

Theme

### Expected behavior

With the previous version of CLI or with the `--legacy` flag, JS assets work.

### Actual behavior

JS assets no longer work on `127.0.0.1` since upgrading to Shopify CLI 3.67.0.

They're loaded, but I get a JS error in the console: 
```
Uncaught SyntaxError: missing ) after argument list (at vendors@theme.min.js?v=25063884737306468341726750322:1:131260)
```

When I go to the theme's remote url (`xxx.myshopify.com/?preview_theme_id=XXXX`), it works fine.

When I try with `theme dev --legacy`, everything works fine.

Here is the end of the JS file vendors@theme.min.js file opened from https://xxx.myshopify.com/cdn/shop/t/17/assets/vendors@theme.min.js?v=25063884737306468341726753546 (remote url):
Image

And now here's the end of the JS vendors@theme.min.js file opened from http://127.0.0.1:9292/cdn/shop/t/17/assets/vendors@theme.min.js?v=25063884737306468341726752918 (CLI server url):
Image

I notice that `);` is missing at the very end of the file.

Would CLI “cut” the file?

Thanks for your help!

### Verbose output

---

### Reproduction steps

1. Run `shopify theme dev`
2. Go to 127.0.0.1

### Operating System

macOS 14.5

3.67.0

### Shell

_No response_

v20.12.2

_No response_

frandiox:

Closes #4479

- The main issue was caused by non-printable characters, which have a different length in disk vs in memory (string format). By transforming them to Buffer before reading the length we get the real disk value, which is required for `content-length` header.

- Unrelated fixes: consider replacing CDN links in JS files with single quotes, etc.

### How to test your changes?

Added a unit test to show it.

cc @Shopify/advanced-edits

### DIFF
--- a/.changeset/hip-chefs-dance.md
+++ b/.changeset/hip-chefs-dance.md
@@ -1,0 +1,5 @@
+---
+'@shopify/theme': patch
+---
+
+Fix Theme Access authentication on `shopify theme dev` and `shopify theme console` commands
--- a/.changeset/sharp-eggs-remember.md
+++ b/.changeset/sharp-eggs-remember.md
@@ -1,0 +1,5 @@
+---
+'@shopify/theme': patch
+---
+
+Fix cart/add request in development.
--- a/packages/app/src/cli/models/extensions/specification.integration.test.ts
+++ b/packages/app/src/cli/models/extensions/specification.integration.test.ts
@@ -1,4 +1,5 @@
 import {loadLocalExtensionsSpecifications} from './load-specifications.js'
+import {createContractBasedModuleSpecification} from './specification.js'
 import {AppSchema} from '../app/app.js'
 import {describe, test, expect, beforeAll} from 'vitest'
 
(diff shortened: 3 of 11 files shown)


charlespwd (approved):
This LGTM! Curious why we have to figure out the content-length for our server. That seems like a low level detail that the server lib usually provides for us. 

(I'm really out of the loop here, so if what we're doing is is pretty low level, I understand!)

Those URL regex do be gnarly af though. Took me a couple of sips of coffee to get why they were modified :P

frandiox:
Thanks for the review! Yeah we are using a lowish level server called `h3` and it requires us passing the content-length, otherwise the browser doesn't get the files properly.

>Those URL regex do be gnarly af though. Took me a couple of sips of coffee to get why they were modified :P

Yep, just simplifying them a bit 😅  -- I think we'll probably forget about RegExps and just ask AI to change them for us 😂

PR #4500, opened 2024-09-20
[Themes] - Refactor `pull` command to prepare for public themes API

jamesmengo:

Fixes https://github.com/Shopify/develop-advanced-edits/issues/341
Fixes https://github.com/Shopify/develop-advanced-edits/issues/342

Allows users to call `pull` programatically by providing the flags as arguments

1) Creates a service function `pull` which allows users to provide command flags as args
2) Adds JSDoc to the public `pull` method args
2) Splits the CLI2 logic (this is kept in the command layer)
3) Adds some tests to the service layer

### How to test your changes?
- Hover over the call to `pull` in `packages/theme/src/cli/commands/theme/pull.ts` to see the Docs
- Call `shopify-dev theme pull` with and without the `--legacy` flag. Everything should work as before

### Measuring impact

### Checklist


packages/theme/src/cli/services/pull.ts
```diff
@@ -1,12 +1,18 @@
+/* eslint-disable tsdoc/syntax */
```
jamesmengo, line 1: Our `eslint` rules don't the period in `flags.path` 

In most of the repo, it seems like we don't outline the object properties in the JSDoc comments [example](https://github.com/Shopify/cli/blob/606087c80e07aff74306dab99c442bc97e76a1f5/packages/app/src/cli/models/app/loader.ts#L976-L984)

Since this is a complex object that consists of flags though, I am ignoring these rules.
Open to discussion!

packages/theme/src/cli/services/pull.ts
```diff
@@ -16,7 +22,78 @@ interface PullOptions {
...
+  ignore?: string[]
+  force?: boolean
+  noColor?: boolean
+  verbose?: boolean
+}
+
+/**
+ * Initiates the pull process based on provided flags.
+ *
+ * @param {PullFlags} flags - The flags for the pull operation.
```
jamesmengo, line 44: todo: remove these types - they don't get rendered when hovering anyways

packages/theme/src/cli/commands/theme/pull.ts
```diff
@@ -61,58 +54,27 @@ If no theme is specified, then you're prompted to select the theme to pull from
...
-          live,
-          theme: development ? `${developmentTheme?.id}` : flags.theme,
-        },
-      })
 
-      await pull(theme, adminSession, {path, nodelete, only, ignore, force})
-
-      return
+    const pullFlags: PullFlags = {
+      path: flags.path,
```
karreiro, line 65: What do you think about adding the documentation in the properties here instead?

Something like this:
```ts
/**
 * The directory path to download the theme.
 */
path?: string
```

The benefit I notice is that users get docs for their properties:


And the documentation of functions gets a bit easier to digest:

karreiro (commented):
Thank you, @jamesmengo! The PR is on an excellent direction. I've left only one comment about the documentation in the public API :)

packages/theme/src/cli/commands/theme/pull.ts
jamesmengo: I like this idea!

packages/theme/src/cli/services/pull.ts
```diff
@@ -16,7 +21,116 @@ interface PullOptions {
   ignore?: string[]
 }
 
-export async function pull(theme: Theme, session: AdminSession, options: PullOptions) {
+export interface PullFlags {
```
jamesmengo, line 24: https://github.com/Shopify/develop-advanced-edits/issues/345

packages/theme/src/cli/services/pull.ts
```diff
@@ -16,7 +21,116 @@ interface PullOptions {
...
+   */
+  verbose?: boolean
+}
+
+/**
+ * Initiates the pull process based on provided flags.
+ *
+ * @param flags - The flags for the pull operation.
+ * @returns Resolves when the pull operation is complete.
+ */
```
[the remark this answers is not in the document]
karreiro: +1 thank you for that suggestions, @graygilmore!

karreiro (approved):
Thank you, @jamesmengo!

PR #4501, opened 2024-09-20
Jm/api push

jamesmengo:

Fixes #0000 

### Measuring impact

### Checklist


packages/theme/src/cli/commands/theme/push.ts
```diff
@@ -139,33 +134,24 @@ export default class Push extends ThemeCommand {
 
   async run(): Promise<void> {
     const {flags} = await this.parse(Push)
-    const {path, force} = flags
-    const store = ensureThemeStore(flags)
-    const adminSession = await ensureAuthenticatedThemes(store, flags.password)
 
-    const workingDirectory = path ? resolvePath(path) : cwd()
-    if (!(await hasRequiredThemeDirectories(workingDirectory)) && !(await currentDirectoryConfirmed(force))) {
+    if (flags.password) {
```
jamesmengo, line 138: I am keeping the ruby implementation in this for now, given that this guard is for CLI users. 
This will NOT be exposed via the public package

This should be a simple removal at a later date 😄

packages/theme/src/cli/commands/theme/push.ts
```diff
@@ -197,61 +204,3 @@ export default class Push extends ThemeCommand {
...
-
-export interface ThemeSelectionOptions {
-  live?: boolean
-  development?: boolean
-  unpublished?: boolean
-  theme?: string
-  'allow-live'?: boolean
-}
-
-export async function createOrSelectTheme(
```
jamesmengo, line 209: moved to service layer

packages/theme/src/cli/commands/theme/push.ts
```diff
@@ -139,33 +134,45 @@ export default class Push extends ThemeCommand {
 
   async run(): Promise<void> {
     const {flags} = await this.parse(Push)
```
jamesmengo, line 136: The logic on the LHS here has been copied to the service layer as well as included in `execLegacyPush()`

I'm keeping it here in the command layer to service the `CLI2` logic - in this case I think it's cleaner / easier to change in the future if we have some temporary duplication here

packages/theme/src/cli/services/push.test.ts
```diff
@@ -1,44 +1,183 @@
...
-      force: false,
-    })
+    await push({...defaultFlags, publish: true})
 
     // Then
     expect(publishTheme).toHaveBeenCalledWith(theme.id, adminSession)
   })
 })
+
+describe('createOrSelectTheme', async () => {
```
jamesmengo, line 64: these tests were ripped from the command layer

packages/theme/src/cli/commands/theme/push.ts
karreiro: +100

PR #4509, opened 2024-09-23
Enforce that all commands load an app

isaacroldan:
All commands should track all app-related details.
To enforce that, we make every app command return an `AppInterface`, which means that an app was loaded at some point.

- Add a new ESLint rule to make all commands in `app/commands` extend `AppCommand`
- In `AppCommand` define the `run` function to return an object with an `AppInterface`, TS will force all child classes to use the same return type.
- Solve all the issues created by the two rules above: all commands extend `AppCommand`, all return an `AppInterface` now.

IN A FUTURE PR:
- All app commands should return a Session, to force authentication in all of them.
- Since all commands have auth + app loading -> extract that logic to a single place: `loadAppContext`

### How to test your changes?
Nothing to test in the CLI itself, you can try to modify the code of a command and see that the rules are being applied.

### Measuring impact

### Checklist

### DIFF
--- a/packages/app/src/cli/utilities/app-command.ts
+++ b/packages/app/src/cli/utilities/app-command.ts
@@ -1,8 +1,23 @@
 import {configurationFileNames} from '../constants.js'
+import {AppInterface} from '../models/app/app.js'
+import {AbortError} from '@shopify/cli-kit/node/error'
 import BaseCommand from '@shopify/cli-kit/node/base-command'
+
+/**
+ * By forcing all commands to return `AppCommandOutput` we can be sure that during the run of each command we:
+ * - Authenticate the user
+ * - Load an app
+ */
+export interface AppCommandOutput {
+  app: AppInterface
+}
 
 export default abstract class AppCommand extends BaseCommand {
   environmentsFilename(): string {
     return configurationFileNames.appEnvironments
+  }
+
+  public async run(): Promise<AppCommandOutput> {
+    throw new AbortError('You should not call this method directly, implement it in a subclass')
   }
 }
--- a/packages/app/src/cli/utilities/app-command.ts
+++ b/packages/app/src/cli/utilities/app-command.ts
@@ -1,7 +1,7 @@
 import {configurationFileNames} from '../constants.js'
-import Command from '@shopify/cli-kit/node/base-command'
+import BaseCommand from '@shopify/cli-kit/node/base-command'
 
-export default abstract class AppCommand extends Command {
+export default abstract class AppCommand extends BaseCommand {
   environmentsFilename(): string {
     return configurationFileNames.appEnvironments
   }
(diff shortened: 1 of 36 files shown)


packages/app/src/cli/utilities/app-command.ts
```diff
@@ -1,8 +1,24 @@
...
+  app: AppInterface
+}
+
+export default abstract class AppCommand extends BaseCommand {
   environmentsFilename(): string {
     return configurationFileNames.appEnvironments
   }
+
+  public async run(): Promise<AppCommandOutput> {
+    throw new AbortError('You should not call this method directly, implement it in a subclass')
```
shauns, line 22: Given the class is abstract, could the method be too? https://www.typescriptlang.org/docs/handbook/2/classes.html#abstract-classes-and-members:~:text=abstract%20getName()%3A%20string%3B

packages/cli/oclif.manifest.json
```diff
@@ -681,7 +681,6 @@
       "customPluginName": "@shopify/app",
       "description": "Compiles the function in your current directory to WebAssembly (Wasm) for testing purposes.",
       "descriptionWithMarkdown": "Compiles the function in your current directory to WebAssembly (Wasm) for testing purposes.",
-      "enableJsonFlag": false,
```
shauns, line 684: Calling out that the manifest has changed in this PR. Deliberate?

packages/eslint-plugin-cli/rules/app-commands-requirements.js
```diff
@@ -0,0 +1,53 @@
+const path = require('path')
+
+/**
+ * Every App Command should have a default export that extends `AppCommand`
+ *
+ * This way we control that all commands are constrained by what's defined in the `AppCommand` abstract class.
+ * Such as the return type for the run() function.
+ */
+module.exports = {
```
shauns, line 9: Could you use a JSDoc comment to add some type checking to this? Like: https://github.com/Shopify/cli/blob/aa8b871813d93197b6e2925a8eb5d4aa87f96267/packages/eslint-plugin-cli/rules/no-inline-graphql.js#L75

Makes it easier to later modify the rule.

packages/eslint-plugin-cli/rules/app-commands-requirements.js
```diff
@@ -0,0 +1,53 @@
...
+            if (statement.declaration.type === 'ClassDeclaration') {
+              const superClass = statement.declaration.superClass
+              if (superClass && superClass.type === 'Identifier' && superClass.name === 'AppCommand') {
+                extendsAppCommand = true
+              }
+            }
+          }
+        })
+
+        if (!hasDefaultExport || !extendsAppCommand) {
```
shauns, line 44: Out of interest: did you consider doing this by looking at the commands index file and checking _those_ sub-classes?

Issue #4528, opened 2024-09-25
[Themes] Add tests for themes local storage


jamesmengo:
I've made the change. Interested in your thoughts of what would be worth testing here. Or you can take responsibility for that piece, since this part is mainly an improvement to themes robustness.

_Originally posted by @amcaplan in https://github.com/Shopify/cli/pull/4517#discussion_r1774881055_

PR #4530, opened 2024-09-25
[Themes] - Add tests to local storage functions that call `requireThemeStore`

jamesmengo:
Adds tests for the changes introduced [Force dealing with config get returning undefined](https://github.com/Shopify/cli/pull/4517)

Closes https://github.com/Shopify/cli/issues/4528

- Test that the functions in `local-storage` throw an Error when `getThemeStore()` returns `undefined`
- Allow for a `LocalStorage<ThemeLocalStorageSchema>` instance to be injected into the functions in `local-storage`, which will be used to check in `requireThemeStore` (_this enables the testing, which was unfortunately not as straightforward as I would have liked_)

### How to test your changes?
`pnpm test`

### Measuring impact

### Checklist

cc: @Shopify/advanced-edits

### DIFF
--- a/packages/theme/src/cli/services/local-storage.ts
+++ b/packages/theme/src/cli/services/local-storage.ts
@@ -4,7 +4,7 @@ import {outputDebug, outputContent} from '@shopify/cli-kit/node/output'
 
 type DevelopmentThemeId = string
 
-interface ThemeLocalStorageSchema {
+export interface ThemeLocalStorageSchema {
   themeStore: string
 }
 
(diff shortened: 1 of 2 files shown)


packages/theme/src/cli/services/local-storage.ts
```diff
@@ -54,58 +55,85 @@ function themeStorePasswordStorage() {
...
-  const themeStore = getThemeStore()
+export function removeStorefrontPassword(
+  themeStorage: LocalStorage<ThemeLocalStorageSchema> = themeLocalStorage(),
+): void {
+  const themeStore = requireThemeStore(themeStorage)
   outputDebug(outputContent`Removing storefront password for ${themeStore}...`)
   themeStorePasswordStorage().delete(themeStore)
 }
+
+function requireThemeStore(storage: LocalStorage<ThemeLocalStorageSchema> = themeLocalStorage()): string {
```
karreiro, line 133: Personally, I believe we could rename it to `ensureThemeStore` as we have this `ensure*` pattern in some places :)

karreiro (approved):
Thank you, @jamesmengo! LGTM :) I believe that after applying the rename I suggested below (please let me know your thoughts about that) and fixing the conflict in the branch, we will be ready to merge 🚀

packages/theme/src/cli/services/local-storage.ts
jamesmengo: 🤔 
Since this check is a tripwire rather than a `set if not already set` check, I'm going to rename this to `assertThemeStoreExists`.

This avoids conflicting with `ensureThemeStore` which sets the theme store if not set in the command layer [here](https://github.com/Shopify/cli/blob/d549077b51cd5ab646ce7aa0430611e2c3f5eb28/packages/theme/src/cli/utilities/theme-store.ts#L6)

PR #4559, opened 2024-10-01
[Themes] Remove Ruby from `theme push` command

jamesmengo:

Fixes https://github.com/Shopify/develop-advanced-edits/issues/352

- Removes 
- Removes Ruby invocation for the `theme push` command

### Measuring impact

### Checklist


packages/theme/src/cli/commands/theme/push.ts
```diff
@@ -135,7 +130,7 @@ export default class Push extends ThemeCommand {
   async run(): Promise<void> {
     const {flags} = await this.parse(Push)
 
-    if (flags.password || flags.legacy) {
+    if (flags.password) {
```
jamesmengo, line 133: Blocked by: https://github.com/orgs/Shopify/projects/4613/views/62?pane=issue&itemId=80272317

50bbx:
As I can see from the [docs](https://shopify.dev/docs/api/shopify-cli) the ruby dependency is not required anymore but this changes are still in draft. Was that intended?

And it looks like ruby is actually still required (version 3.67.0): image

Is this true?

jamesmengo:
@50bbx You should be able to run the TS implementation now without the need to install Ruby (depending on the version)

Could you share which version you're on + which command you're running? The `--legacy` or `password` flag may be configured in your environment variables, which is causing it to default to Ruby.

jamesmengo:
npm snapshot

packages/theme/src/cli/commands/theme/push.ts
```diff
@@ -135,11 +121,6 @@ export default class Push extends ThemeCommand {
   async run(): Promise<void> {
     const {flags} = await this.parse(Push)
 
-    if (flags.password || flags.legacy) {
```
jamesmengo, line 138: [Validate \`Push\` command output in CI/CD workflows · Issue #337 · Shopify/develop-advanced-edits](https://github.com/Shopify/develop-advanced-edits/issues/337)

frandiox (approved):
#feelsgood

karreiro (approved):
Thank you, @jamesmengo! LGTM and works as expected! I believe we just need to remove the `cli2Flags` variable, and then we will be ready to merge :)

PR #4574, opened 2024-10-03
Warn less often about scope mismatch

shauns:

We show a warning if we think the developer may have forgotten to run deploy prior to dev, in particular if the developer has changed their app scopes in their local TOML file, but they're out of sync with what is found remotely. There are cases where this check is giving a false alarm.

Normalise the scope strings before comparing them.

### How to test your changes?

Tricky one... there's an automated test. And we have an internal app that has one of these false alarms that I'm happy to share once we have a testable patch in place.

### Measuring impact

### Checklist


dmerand (approved):
Logic looks good to me, should lead to fewer false positives on mismatch 👍🏼 I didn't tophat per the note in the PR that it's a bit tricky; the tests look to cover the case well.

shauns:
Thanks Donald -- I checked it with the snapshot build on the affected app and the warning went away.

PR #4577, opened 2024-10-03
Functions 'schema' and 'replay' commands - fix dependence on `.env`

mssalemi:

Fixes: https://github.com/Shopify/shopify-functions/issues/411

### Summary of Changes

This PR addresses an issue where the absence of a `.env` file (and consequently an API key) triggers a prompt for the user to select an app every time a command is executed. This behavior is due to the condition checked in the following code segment: https://github.com/Shopify/cli/blob/037598517bc4cc282ac57dd841fcdbe71034e346/packages/app/src/cli/services/generate-schema.ts#L40.

#### Proposed Solution

To resolve this, the solution I have currently, uses the already fetched API key (using an existing call with `fetchOrCreateOrganizationApp`), and we cache that data  using `setCachedAppInfo` (from [here](https://github.com/Shopify/cli/blob/main/packages/app/src/cli/services/local-storage.ts#L32-L69.))

With this change, once the API key is cached, subsequent command executions will not prompt the user to select an organization, thereby streamlining the workflow.

### How to Test Your Changes

1. Ensure that there is no `.env` file in your app directory.

2. Execute the schema command:
   ```
   SHOPIFY_CLI_ENABLE_APP_LOG_POLLING=1 NODE_TLS_REJECT_UNAUTHORIZED=0 SPIN=1 SPIN_INSTANCE=XXX pnpm run shopify app function schema --path=./app-name
   ```

3. Execute the replay command:
   ```
   SHOPIFY_CLI_ENABLE_APP_LOG_POLLING=1 NODE_TLS_REJECT_UNAUTHORIZED=0 SPIN=1 SPIN_INSTANCE=XXX pnpm run shopify app function replay --path=./app-name
   ```

During a new cli `replay` or `schema` command run, you will encounter the app selection prompt. However, for subsequent runs, the cli should use the cached API key, bypassing the need for manual selection. 

TODO: current just setting the apikey as the appId

### Notes:
- better way to get and store api key, do we create an `.env` for them?
- currently i just set it on the appId 

### Measuring impact

### Checklist

### DIFF
--- a/packages/app/src/cli/services/generate-schema.ts
+++ b/packages/app/src/cli/services/generate-schema.ts
@@ -28,7 +28,10 @@ export async function ensureConnectedAppFunctionContext(
   const {app} = options
   const developerPlatformClient =
     options.developerPlatformClient ?? selectDeveloperPlatformClient({configuration: options.app.configuration})
-  let apiKey = options.apiKey || getAppIdentifiers({app}, developerPlatformClient).app
+  let apiKey =
+    options.apiKey ||
+    getAppIdentifiers({app}, developerPlatformClient).app ||
+    (await getCachedAppInfo(app.directory))?.apiKey
 
   if (!apiKey) {
     if (!isTerminalInteractive()) {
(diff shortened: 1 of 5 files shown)


packages/app/src/cli/services/generate-schema.ts
```diff
@@ -37,7 +38,20 @@ export async function ensureConnectedAppFunctionContext(
...
+      const createdApp = await fetchOrCreateOrganizationApp(app.creationDefaultOptions())
+      apiKey = createdApp.apiKey
+
+      await setCachedAppInfo({
+        appId: apiKey,
+        title: createdApp.title,
+        directory: app.directory,
+        orgId: createdApp.organizationId,
+      })
+    }
```
mssalemi, line 54: TODO: refactor to simplify / clean this up.

packages/app/src/cli/services/local-storage.ts
```diff
@@ -13,6 +13,7 @@ export interface CachedAppInfo {
   storeFqdn?: string
   updateURLs?: boolean
   previousAppId?: string
+  apiKey?: string
 }
```
mssalemi, line 17: Added this, but I noticed we may use the `appId` field to store the `apiKey`, one example [here](https://github.com/Shopify/cli/blob/8da87af1592efb9cffc565fd9299e3903aaa6746/packages/app/src/cli/services/context.ts#L248-L254).  But I opted just to add `apiKey` as an optional type.

Issue #2664, opened 2023-08-17
[Bug]: shopify theme share uploads files ignored in .shopifyignore


hsdonkin:
### Please confirm that you have:

- [X] Searched [existing issues](.) to see if your issue is a duplicate. (If you’ve found a duplicate issue, feel free to add additional information in a comment on it.)
- [X] Reproduced the issue in [the latest CLI version](https://www.npmjs.com/package/@shopify/cli).

### In which of these areas are you experiencing a problem?

Theme

### Expected behavior

When uploading a theme via `shopify theme share`, files included in the `.shopifyignore` should not be uploaded

```
package.json
package-lock.json
node_modules
node_modules/
```

### Actual behavior

Ignored files are uploaded, and upload hangs due to number of files in `node_modules` 😅 
<img width="711" alt="image" src="https://github.com/Shopify/cli/assets/49598076/d0795dc0-2669-4e4a-87a7-d29e5be4eb49">

### Reproduction steps

1. Install any npm packages, i.e. `npm init && npm i preact`
2. Add `package.json` and `package-lock.json` and `node_modules` to `.shopifyignore`
3. `shopify theme share`

### Operating System

OSX 11.5.1

### Shopify CLI version (check your project's `package.json` if you're not sure)

3.4.2

johnphillipsdev:
Running into the same issue, moving the theme files into its own directory is helpless too as Shopify can't find them via subdirectories on the store connection.

karreiro:
Thank you for reporting this issue, @hsdonkin. It's being solved by https://github.com/Shopify/cli/pull/4578.

PR #4578, opened 2024-10-03
Use TS uploader for shopify theme share command

EvilGenius13:

Fixes https://github.com/Shopify/develop-advanced-edits/issues/25

The `shopify theme share` command was still calling the Ruby implementation and we wanted it to call the Typescript implementation like the other commands.

This PR switches `run` to call the TS uploader.
I have also made a `legacyRun` command that still uses the Ruby version as a fallback. We can remove the fallback in a later PR.

### How to test your changes?

Pull down the branch
Build the branch
run the `shopify theme share` command and check that the output looks the same as before.

### Measuring impact

### Checklist


packages/theme/src/cli/commands/theme/share.ts
```diff
@@ -24,12 +26,40 @@ export default class Share extends ThemeCommand {
...
+
+    const pushFlags: PushFlags = {
+      path: flags.path,
+      password: flags.password,
+      store: flags.store,
+      environment: flags.environment,
+      force: flags.force,
+      noColor: flags['no-color'],
+      verbose: flags.verbose,
+      unpublished: true,
```
EvilGenius13, line 54: The share command is essentially the push command with `unpublished: true` and the theme is given a random name which is done using `getRandomName`

packages/theme/src/cli/commands/theme/share.ts
```diff
@@ -24,12 +26,40 @@ export default class Share extends ThemeCommand {
...
+      env: 'SHOPIFY_FLAG_LEGACY',
+    }),
   }
 
   static cli2Flags = ['force']
 
   async run(): Promise<void> {
     const {flags} = await this.parse(Share)
+
+    if (flags.password || flags.legacy) {
```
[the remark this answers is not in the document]
karreiro: +1 for removing the legacy implementation from this file. The upload implementation has been validated as in `shopify theme push`, so it's safer to proceed here :)

karreiro (commented):
Thank you for this PR, @EvilGenius13! It's in the right direction 🚀 

I've associated it with this issue as well https://github.com/Shopify/cli/issues/2664 because the `shopify theme share` command should consider the `.shopifyignore` file. It was an issue in the previous implementation, and should be already fixed out-of-the-box due to the nature of the new implementation (but I've associated it to make sure we 🎩 that scenario).

Thanks again for this PR!

packages/theme/src/cli/commands/theme/share.ts
```diff
@@ -24,12 +26,40 @@ export default class Share extends ThemeCommand {
...
+      return
+    }
+
+    const pushFlags: PushFlags = {
+      path: flags.path,
+      password: flags.password,
+      store: flags.store,
+      environment: flags.environment,
+      force: flags.force,
+      noColor: flags['no-color'],
```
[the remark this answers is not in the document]
EvilGenius13: I just ran the command again without explicitly passing them in `pushFlags` and everything was still 👍

packages/theme/src/cli/commands/theme/share.ts
jamesmengo: I think we need the following flags (essentially `force` + all `themeFlags` except `environment` )
- `force`
- `store`
- `path`
- `password`

We don't need to pass anything in from `globalFlags`

jamesmengo (approved):
🎩 and code LGTM
I did test the `.shopifyignore` logic, which is working as well, but I think you should tophat that again for learning purposes!

EvilGenius13:
> 🎩 and code LGTM I did test the `.shopifyignore` logic, which is working as well, but I think you should tophat that again for learning purposes!

Confirmed as well! Files I add to the `.gitignore` are properly ignored.

.changeset/strange-tomatoes-float.md
```diff
@@ -0,0 +1,5 @@
+---
+'@shopify/cli': minor
+---
+
+Use TS uploader for shopify theme share command
```
karreiro, line 5: (optional) Personally, I'd try to keep the changelogs a bit user-centric, so I'd update with something like this:

> Use the new implementation of `shopify theme share` that no longer depends on Ruby to work. Additionally, it fixes the command so it no longer ignores `.shopifyignore`.

As a user, it's not so important that the uploader is in TS (TypeScript), it's that they no longer need Ruby on their machine. Also, it's important that `.shopifyignore` now works as expected (so I'd focus a bit more on those aspects).

EvilGenius13: I like it. I'll adjust the wording. Thank you for the feedback!

johnphillipsdev:
Hey guys, just jumping on this, I still have this issue. Could anyone point me in the right direction.
_(I'm probably being impatient, but I'm eager to get this working)_

So regardless of what I put within `.shopifyignore` the files still seem to get uploaded causing Shopify theme share to slow down.
Everything, including the theme files are within the root directory.

image

After running `shopify theme share` I run into the following...

image

Deleting node_modules resolves this, but of course, that's not practical 👀

All files are within the root directory to connect to GitHub, as of course I am unable to upload a theme via the integration using a subdirectory within Shopify.

image

johnphillipsdev:
@EvilGenius13 Sorry to be a bother! I'm reaching out as your comment was deleted. - I don't even mind if it said "you're an idiot this is what you do" 😅. 

I'm still having this issue and genuinely can't get past it. Any advice would be incredible.

EvilGenius13:
> @EvilGenius13 Sorry to be a bother! I'm reaching out as your comment was deleted. - I don't even mind if it said "you're an idiot this is what you do" 😅.
> 
> I'm still having this issue and genuinely can't get past it. Any advice would be incredible.

Hi @johnphillipsdev, no bother at all! The reason why this isn't working is because this PR shipped just after the latest version was released, so this code change isn't available. 

This should be available in the next version but in the meantime I am going to see if there is a way for you to test the changes (no promises!)

johnphillipsdev:
@EvilGenius13 Thank you for this! - No worrues about looking for an alternative way, I'll be patient! I apprirated the message and clarity. It's just nice to know I'm not doing something terribly wrong 😂.

PR #4581, opened 2024-10-04
[Theme] Avoid process exit when failing to delete a file from the remote theme

frandiox:

I found that this promise rejection was causing process exit:
image

Turning it here into an error message instead:

image

(Note in the above image I was using `renderWarning` but I've changed it to `renderError` since that's more accurate, I think)

### Measuring impact

### Checklist

### DIFF
--- a/.changeset/rotten-worms-rush.md
+++ b/.changeset/rotten-worms-rush.md
@@ -1,0 +1,5 @@
+---
+'@shopify/theme': patch
+---
+
+Avoid process exit when failing to delete a file from the remote theme.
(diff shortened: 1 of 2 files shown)


karreiro (approved):
Thank you for this PR, @frandiox!

I notice that it works as expected during startup, however, when the server is already running and users remove the `templates/gift_card.liquid`, they don't get the error message.

I don't consider that a blocker to merge tho, because the goal of this fix is to prevent the process from unexpectedly terminating, but I believe it would be nice if we could show the message in that context as well.

Thanks again for this PR!

jamesmengo (approved):

PR #4592, opened 2024-10-07
[Themes] Remove Ruby invocation from the `app dev` command

jamesmengo:
### WHY are these changes introduced?
Closes https://github.com/Shopify/develop-advanced-edits/issues/362

### WHAT is this pull request doing?
- Remove `legacy` flag from `app dev` command
- Delete `theme-app-extension.js` (legacy implementation)
- Update imports + tests

### How to test your changes?
- Try running `app dev` with the legacy flag. This should complain about a non-existent flag

### Post-release steps
- https://github.com/Shopify/cli/pull/4593
- https://github.com/Shopify/cli/pull/4595

### Measuring impact

How do we know this change was effective? Please choose one:

- [ ] n/a - this doesn't need measurement, e.g. a linting rule or a bug-fix
- [x] Existing analytics will cater for this addition
- [ ] PR includes analytics changes to measure impact

### Checklist

- [x] I've considered possible cross-platform impacts (Mac, Linux, Windows)
- [x] I've considered possible [documentation](https://shopify.dev) changes

PR #4593, opened 2024-10-07
Rename theme-app-extension-next to theme-app-extension

jamesmengo:

Follow up on - https://github.com/Shopify/cli/pull/4592

Renames a file that was deleted in the previous PR (replacing `theme-app-extension` with `theme-app-extension-next`)

I split this into a different PR to avoid polluting the diff of this change


karreiro (approved):
Thank you, @jamesmengo!

Issue #4598, opened 2024-10-08
[Bug]: CLI Timeout and fails to persistently keep renegotiating connection


JamiesonRoberts:
### Please confirm that you have:

- [x] Searched [existing issues](.) to see if your issue is a duplicate. (If you’ve found a duplicate issue, feel free to add additional information in a comment on it.)
- [x] Reproduced the issue in [the latest CLI version](https://www.npmjs.com/package/@shopify/cli).

### In which of these areas are you experiencing a problem?

Theme

### Expected behavior

If the theme CLI is left running in the background, it shouldn't timeout or just stop spontaneously and should automatically renegotiation any tokens and just keep running. 

### Actual behavior

Unknown on exact timing, but the CLI will just randomly stop working and present the following error. Sometimes it will be after a period of inactivity, sometimes it will just stop working and not re-negotiate the connection while actively using the CLI to updated files while building things

```
Error: [401] API request unauthorized error
    at request (file:///opt/homebrew/Cellar/shopify-cli/3.68.0/libexec/lib/node_modules/@shopify/cli/dist/index.js:145224:13)
    at process.processTicksAndRejections (node:internal/process/task_queues:105:5)
    at async fetchChecksums (file:///opt/homebrew/Cellar/shopify-cli/3.68.0/libexec/lib/node_modules/@shopify/cli/dist/index.js:145198:17)
    at async pollRemoteJsonChanges (file:///opt/homebrew/Cellar/shopify-cli/3.68.0/libexec/lib/node_modules/@shopify/cli/dist/index.js:148436:97)
```

### Verbose output

N/A

### Reproduction steps

1. Start up the CLI
2. Leave it running for a time period (can't get an exact timing on it, but my guess is at least an hour or 2)
3. See the timeout and failure to renegotiate

### Operating System

macOS Sonoma 14.6.1

### Shopify CLI version (check your project's `package.json` if you're not sure)

3.68.0

### Shell

iTerm + ZSH

### Node version (run `node -v` if you're not sure)

20.10.0

Issue #4576, opened 2024-10-03
[Bug]: AggregateError [ETIMEDOUT] (shopify theme dev --theme-editor-sync)


MaxDesignFR:
### Please confirm that you have:

- [x] Searched [existing issues](.) to see if your issue is a duplicate. (If you’ve found a duplicate issue, feel free to add additional information in a comment on it.)
- [x] Reproduced the issue in [the latest CLI version](https://www.npmjs.com/package/@shopify/cli).

### In which of these areas are you experiencing a problem?

Theme

### Expected behavior

It should have worked.

### Actual behavior

The CLI ends its process by itslef. Sometimes the local theme dev might work a couple of minutes before it crashes randomly (with a different error than the one in this verbose), sometimes it crashes as soon as I start it (see verbose). Using `--legacy` is not a good option either because it just can't handle the newer json comments properly when using flag `--theme-editor-sync` so it's a pain to use (see issue https://github.com/Shopify/cli/issues/4107 that is still relevant)

### Verbose output

```
maxdesign@PC-MAX:~/shopify/stores/Univers Du Japon$ store_id=$(jq -r '.store' .vscode/settings.json) || read -p 'Enter store ID: ' store_id && shopify theme dev --theme-editor-sync --store=$store_id --verbose
2024-10-03T15:42:56.437Z: Running command theme dev
2024-10-03T15:42:56.449Z: Ensuring that the user is authenticated with the Theme API with the following scopes:
[]

2024-10-03T15:42:56.449Z: Ensuring that the user is authenticated with the Admin API with the following scopes for the store japon-avenue.myshopify.com:
[]

2024-10-03T15:42:56.449Z: Getting session store...
2024-10-03T15:42:56.452Z: Validating existing session against the scopes:
[
  "openid",
  "https://api.shopify.com/auth/shop.admin.graphql",
  "https://api.shopify.com/auth/shop.admin.themes",
  "https://api.shopify.com/auth/partners.collaborator-relationships.readonly",
  "https://api.shopify.com/auth/shop.storefront-renderer.devtools",
  "https://api.shopify.com/auth/partners.app.cli.access",
  "https://api.shopify.com/auth/destinations.readonly"
]
For applications:
{
  "adminApi": {
    "scopes": [],
    "storeFqdn": "japon-avenue.myshopify.com"
  }
}

2024-10-03T15:42:56.452Z: Sending Identity Introspection request to URL: https://accounts.shopify.com/oauth/introspection
[...]

PR #4599, opened 2024-10-08
[Theme] Handle async errors

frandiox:

Related to #4576 and #4598

A proper request retry + token refresh is still needed but this PR tries to handle async errors more gracefully.

Ensure we print async errors properly with a red banner, and avoid process exit in some situations:
- On file delete / upload error in the remote theme, we show the error but the process continues. This is because the local development can continue working with local files in most cases, and is up to the developer to halt the process after seeing the error.
- On polling errors, wait for 5 errors before exiting the process. This is because it might recover from one or two failed attempts since it still compares local checksums with latest.

For example, when you lose your internet connection while polling:

image

During the initial sync, if there are too many delete errors or a single upload error:

image

### How to test your changes?

Create fake errors around the modified code or directly disconnect your internet + reconnect it within 15 seconds.

### Measuring impact

### Checklist

### DIFF
--- a/packages/theme/src/cli/utilities/errors.ts
+++ b/packages/theme/src/cli/utilities/errors.ts
@@ -3,12 +3,27 @@ import {outputDebug} from '@shopify/cli-kit/node/output'
 import {renderError, renderFatalError} from '@shopify/cli-kit/node/ui'
 
 /**
- * Renders an error banner for a failed syncing operation without interrupting the process.
+ * Renders an error banner for a thrown error with a headline.
+ * @param headline - The headline for the error.
+ * @param error - The error to render.
+ */
+export function renderThrownError(headline: string, error: Error | AbortError) {
+  if (error instanceof AbortError) {
+    error.message = `${headline}\n${error.message}`
+    renderFatalError(error)
+  } else {
+    renderError({headline, body: error.message})
+    outputDebug(`${headline}\n${error.stack ?? error.message}`)
+  }
+}
+
+/**
+ * Creates a function that renders a failed syncing operation without interrupting the process.
  * @param resource - The file that was being operated on, or a headline for the error.
  * @param preset - The type of operation that failed.
  * @returns A function that accepts an error and renders it.
  */
-export function renderCatchError(resource: string, preset?: 'delete' | 'upload') {
+export function createRenderCatchError(resource: string, preset?: 'delete' | 'upload') {
   let headline = resource
 
   if (preset) {
@@ -18,31 +33,19 @@ export function renderCatchError(resource: string, preset?: 'delete' | 'upload')
         : `Failed to upload file "${resource}" to remote theme.`
   }
 
-  return (error: Error | AbortError) => {
-    if (error instanceof AbortError) {
-      error.message = `${headline}.\n${error.message}`
-      renderFatalError(error)
-    } else {
-      renderError({headline, body: error.message})
-      outputDebug(`${headline}\n${error.stack ?? error.message}`)
-    }
+  return (error: Error) => {
+    renderThrownError(headline, error)
   }
 }
 
 /**
- * Renders a fatal error banner and exits the process without continuing.
+ * Creates a function that renders a failed syncing operation and exits the process.
  * @param headline - The headline for the error.
  * @returns A function that accepts an error and renders it.
  */
-export function abortCatchError(headline: string) {
-  return (error: Error | AbortError): never => {
-    if (error instanceof AbortError) {
-      error.message = `${headline}\n${error.message}`
-      renderFatalError(error)
-    } else {
-      renderError({headline, body: error.message})
-    }
-
+export function createAbortCatchError(headline: string) {
+  return (error: Error): never => {
+    renderThrownError(headline, error)
     process.exit(1)
   }
 }
--- a/packages/theme/src/cli/utilities/errors.ts
+++ b/packages/theme/src/cli/utilities/errors.ts
@@ -1,0 +1,30 @@
+import {AbortError} from '@shopify/cli-kit/node/error'
+import {outputDebug} from '@shopify/cli-kit/node/output'
+import {renderError, renderFatalError} from '@shopify/cli-kit/node/ui'
+
+/**
+ * Renders an error banner for a failed syncing operation without interrupting the process.
+ * @param resource - The file that was being operated on, or a headline for the error.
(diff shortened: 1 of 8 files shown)


packages/theme/src/cli/utilities/theme-environment/theme-environment.ts
```diff
@@ -28,22 +29,27 @@ export function setupDevServer(theme: Theme, ctx: DevServerContext) {
...
+    .then(async ({updatedRemoteChecksumsPromise}) => {
+      const updatedRemoteChecksums = await updatedRemoteChecksumsPromise
+      return uploadTheme(theme, ctx.session, updatedRemoteChecksums, ctx.localThemeFileSystem, {
+        nodelete: ctx.options.noDelete,
+        deferPartialWork: true,
+        backgroundWorkCatch: abort,
+      })
     })
-  })
+    .catch(abort)
```
graygilmore, line 49: Instead of `.catch()`ing on each of these could we wrap the whole thing in a `try/catch`? Maybe not with the promises?

packages/theme/src/cli/utilities/theme-fs.ts
```diff
@@ -144,27 +144,26 @@ export function mountThemeFileSystem(root: string, options?: ThemeFileSystemOpti
...
-    const syncPromise = contentPromise.then(async (content) => {
-      if (!unsyncedFileKeys.has(fileKey)) return false
+    const syncPromise = contentPromise
+      .then(async (content) => {
+        if (!unsyncedFileKeys.has(fileKey)) return false
 
-      const [result] = await bulkUploadThemeAssets(Number(themeId), [{key: fileKey, value: content}], adminSession)
+        const [result] = await bulkUploadThemeAssets(Number(themeId), [{key: fileKey, value: content}], adminSession)
+
+        if (!result?.success) {
```
graygilmore, line 153: (nit) Would prefer `result.errors` here

packages/theme/src/cli/utilities/theme-uploader.ts
```diff
@@ -46,11 +48,16 @@ export function uploadTheme(
...
-          renderWarning({headline: 'Failed to delete outdated files from remote theme.'})
-        })
+    : deleteJobPromise.then((result) => result.promise)
+
+  if (options?.backgroundWorkCatch) {
+    // Aggregate all backgorund work in a single promise and handle errors
+    Promise.all([
+      themeCreationPromise,
+      uploadJobPromise.then((result) => result.promise),
+      deleteJobPromise.then((result) => result.promise),
```
graygilmore, line 58: I think this would have already run on L51? I might be misreading this.

graygilmore (commented):
Looks good! Couple of questions but I think it all makes sense.

packages/theme/src/cli/utilities/errors.ts
```diff
@@ -0,0 +1,51 @@
...
+  }
+}
+
+/**
+ * Creates a function that renders a failed syncing operation without interrupting the process.
+ * @param resource - The file that was being operated on, or a headline for the error.
+ * @param preset - The type of operation that failed.
+ * @returns A function that accepts an error and renders it.
+ */
+export function createRenderCatchError(resource: string, preset?: 'delete' | 'upload') {
```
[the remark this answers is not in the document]
frandiox: I did consider it but since the `preset` is optional and can be extended, I thought making it generic is OK 🤔

packages/theme/src/cli/utilities/theme-environment/theme-environment.ts
frandiox: It would be much more verbose with async promises than `.catch`, though. We could also pass it as the second param in `.then(() => {}, abort)` but I think most people are more familiar with `.catch` 🤔

packages/theme/src/cli/utilities/theme-fs.ts
frandiox: I'm not familiar with the underlying API this calls in `cli-kit`, so not sure if something can have a "warning" in errors or something like that. [It looks like](https://github.com/Shopify/cli/blob/main/packages/cli-kit/src/public/node/themes/factories.ts#L71-L72) `success` is just `response.status === 200`, and errors are at least `{}` instead of undefined.
cc @jamesmengo thoughts?

packages/theme/src/cli/utilities/errors.ts
```diff
@@ -0,0 +1,51 @@
...
+  let headline = resource
+
+  if (preset) {
+    headline =
+      preset === 'delete'
+        ? `Failed to delete file "${resource}" from remote theme.`
+        : `Failed to upload file "${resource}" to remote theme.`
+  }
```
karreiro, line 34: Personally, I'd write this using a `switch` or an object (as below), because then the language server complains when we add a new kind of preset (so we may avoid to default to the update-case).

```ts
  const headlienKey = preset ?? 'default'
  const headline = {
    delete: `Failed to delete file "${resource}" from remote theme.`,
    upload: `Failed to upload file "${resource}" to remote theme.`,
    default: resource,
  }[headlienKey]
```

packages/theme/src/cli/utilities/theme-environment/theme-polling.ts
```diff
@@ -21,19 +23,54 @@ export function pollThemeEditorChanges(
...
       })
-      .catch((err) => {
-        if (err instanceof PollingError) {
-          renderError({body: err.message})
-        } else {
-          throw err
+      .catch((error: Error) => {
+        failedPollingAttempts++
+
+        if (error.message !== lastError) {
```
karreiro, line 47: ❤️

karreiro (approved):
Thank you, @frandiox! LGTM and works as expected as well 🚀

jamesmengo (approved):
Very cool! I learned lots from reading this PR - thanks Fran!

packages/theme/src/cli/utilities/errors.ts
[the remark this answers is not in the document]
frandiox: Renamed to `createSyncingCatchError`

karreiro:
@app-inner-loop, Could you please take a look at this PR? It seems like the merge is blocked by that requirement.

karreiro:
@shopify/app-inner-loop, Could you please take a look at this PR? It seems like the merge is blocked by that requirement.

gonzaloriestra (approved):

PR #4536, opened 2024-09-26
[Themes] Dev Server - Improve user logging for render requests

EvilGenius13:
### WHY are these changes introduced?
This PR adds extra information around logs when making requests

Fixes: https://github.com/Shopify/develop-advanced-edits/issues/331

### WHAT is this pull request doing?
This pull request adds more functionality to output logs when making requests to your local dev server in the `theme dev` command. 

Added:

- Timestamp
- Truncated paths if path is over 80 characters
- Request Duration

Example before changes: `GET /collections/all`
Example after changes: `• 13:00:07 Request » GET /collections/all 325ms`

Screenshot 2024-09-26 at 2 42 49 PM

### How to test your changes?
Pull down the branch
Run the build command
Run shopify theme dev
Click different paths like collections, products, etc to see the logs in action.

### Measuring impact

How do we know this change was effective? Please choose one:

- [ ] n/a - this doesn't need measurement, e.g. a linting rule or a bug-fix
- [x] Existing analytics will cater for this addition
- [ ] PR includes analytics changes to measure impact

### Checklist

- [x] I've considered possible cross-platform impacts (Mac, Linux, Windows)
- [x] I've considered possible [documentation](https://shopify.dev) changes

PR #4605, opened 2024-10-08
[Themes] Add status codes and event methods to user logging

EvilGenius13:
Iterating on the last user logging improvements

Follow up to: https://github.com/Shopify/cli/pull/4536
Closes: https://github.com/Shopify/develop-advanced-edits/issues/359

This PR adds more functionality to user logging when using the Dev Server.

Highlights:

- Pulled the `logRequestLine` function into it's own util file for use in other areas
- Added logging to proxy requests allowing us to see `POST` requests (*not including /cdn routes*)
- Added some colour distinction to the output such as status codes and response time

Output example with changes:
image

### How to test your changes?

Pull down the branch
Build the branch
run `theme dev`
click around and watch the logs

### Measuring impact

### Checklist

### DIFF
--- a/packages/theme/src/cli/utilities/theme-environment/proxy.ts
+++ b/packages/theme/src/cli/utilities/theme-environment/proxy.ts
@@ -1,4 +1,5 @@
 import {buildCookies} from './storefront-renderer.js'
+import {logRequestLine} from '../log-request-line.js'
 import {renderWarning} from '@shopify/cli-kit/node/ui'
 import {
   defineEventHandler,
@@ -23,8 +24,8 @@ import type {Response as NodeResponse} from '@shopify/cli-kit/node/http'
 import type {DevServerContext} from './types.js'
 
 const CART_PREFIX = '/cart/'
-const VANITY_CDN_PREFIX = '/cdn/'
-const EXTENSION_CDN_PREFIX = '/ext/cdn/'
+export const VANITY_CDN_PREFIX = '/cdn/'
+export const EXTENSION_CDN_PREFIX = '/ext/cdn/'
 const IGNORED_ENDPOINTS = [
   '/.well-known',
   '/shopify/monorail',
@@ -254,6 +255,8 @@ function proxyStorefrontRequest(event: H3Event, ctx: DevServerContext) {
       redirect: 'manual',
     },
     async onResponse(event, response) {
+      logRequestLine(event, response)
+
       patchProxiedResponseHeaders(ctx, event, response)
 
       const fileName = url.pathname.split('/').at(-1)
--- a/.changeset/purple-games-collect.md
+++ b/.changeset/purple-games-collect.md
@@ -1,0 +1,5 @@
+---
+'@shopify/cli': patch
+---
+
+Improve user logging by adding status codes and event methods
--- a/packages/theme/src/cli/utilities/log-request-line.test.ts
+++ b/packages/theme/src/cli/utilities/log-request-line.test.ts
@@ -1,0 +1,53 @@
+import {shouldLog} from './log-request-line.js'
+import {createEvent} from 'h3'
+import {describe, test, expect} from 'vitest'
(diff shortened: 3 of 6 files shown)


EvilGenius13:
For status codes I went with green for any 2xx status, yellow for 3xx (redirects), and red for anything 4xx-5xx.

I'm also torn on part of the logging. On every request we make a `/checkouts/internal/preloads.js?` request as well. This adds some noise to the output. I considered ignoring it but I'm not sure if that would cause a debugging issue if someone wanted to make sure that was working as intended. Happy for some opinions.

@frandiox I would love your feedback based on your original comments from the PR prior to this. Would you want padding for the Status codes to always be in line or does this seem sufficient.

packages/theme/src/cli/utilities/theme-environment/proxy.ts
```diff
@@ -254,6 +255,10 @@ function proxyStorefrontRequest(event: H3Event, ctx: DevServerContext) {
       redirect: 'manual',
     },
     async onResponse(event, response) {
+      if (!event.path.startsWith(EXTENSION_CDN_PREFIX) && !event.path.startsWith(VANITY_CDN_PREFIX)) {
```
frandiox, line 258: We should only show useful lines so perhaps we should filter this more. Some ideas:
- Filter out any pathname ending with an extension, like `.json`,`.js`, etc.
- Filter out known URLs like `/checkouts/...`
- Perhaps we should only show POST requests from the proxy logic?

Or any combination of the above? 🤔

EvilGenius13: I've added a new function that takes care of this called `shouldLog`, and it is easy enough to add or remove paths as needed. Currently we're back to just showing the important GET/POST requests.

packages/theme/src/cli/utilities/theme-environment/proxy.ts
```diff
@@ -254,6 +269,16 @@ function proxyStorefrontRequest(event: H3Event, ctx: DevServerContext) {
...
     async onResponse(event, response) {
+      if (
+        !event.path.startsWith(EXTENSION_CDN_PREFIX) &&
+        !event.path.startsWith(VANITY_CDN_PREFIX) &&
+        !event.path.endsWith('.js')
+      ) {
+        if (shouldLog(event)) {
+          logRequestLine(event, response)
+        }
+      }
```
karreiro, line 280: I believe we could just call `logRequestLine` here (and then move the `shouldLog` logic to `log-request-line.ts`).

Also, in the `log-request-line.ts`, we may add some unit tests to clarify our expectations about when we should log and when we shouldn't :)

karreiro (commented):
Thank you for this PR, @EvilGenius13! Logs look so much better 🤩 

I've left only one refactoring + unit testing suggestion, and with that we will be ready to merge :)

> Although I prefer seeing the `status code` after the event method, I can see how it aligns better with the `Synced >>` 

Loved the idea of aligning them to the right, but I strongly prefer the status codes after the event method as well 😅 

Thanks again for this PR, @EvilGenius13!

packages/theme/src/cli/utilities/theme-environment/proxy.ts
```diff
@@ -254,6 +269,16 @@ function proxyStorefrontRequest(event: H3Event, ctx: DevServerContext) {
       redirect: 'manual',
     },
     async onResponse(event, response) {
+      if (
+        !event.path.startsWith(EXTENSION_CDN_PREFIX) &&
+        !event.path.startsWith(VANITY_CDN_PREFIX) &&
+        !event.path.endsWith('.js')
+      ) {
```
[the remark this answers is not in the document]
EvilGenius13: Yes you are correct!

karreiro (approved):
Thank you, @EvilGenius13! 🚀

PR #4620, opened 2024-10-09
Add codegen for admin schema

catlee:

Add codegen for admin schema required by cli-kit to integrate with the [new theme graphQL APIs](https://shopify.dev/changelog/introducing-theme-and-theme-file-management-in-the-admin-graphql-api)

### DIFF
--- a/bin/get-graphql-schemas.js
+++ b/bin/get-graphql-schemas.js
@@ -27,6 +27,11 @@ const schemas = [
     repo: 'shopify',
     pathToFile: 'db/graphql/app_dev_schema_unstable_public.graphql',
     localPath: './packages/app/src/cli/api/graphql/app-dev/app_dev_schema.graphql',
+  },
+  {
+    repo: 'shopify',
+    pathToFile: 'db/graphql/admin_schema_unstable_public.graphql',
+    localPath: './packages/cli-kit/src/cli/api/graphql/admin/admin_schema.graphql',
   }
 ]
 
@@ -70,14 +75,14 @@ async function fetchFileForSchema(schema, octokit) {
   try {
     // Fetch the file content from the repository
     const {data} = await octokit.repos.getContent({
+      mediaType: { format: "raw" },
       owner: OWNER,
       repo: schema.repo,
       path: schema.pathToFile,
       ref: BRANCH,
     })
 
-    // Decode the base64 content
-    const content = Buffer.from(data.content, 'base64').toString('utf-8')
+    const content = Buffer.from(data).toString('utf-8')
 
     // Define the local path where the file will be saved
     const localFilePath = schema.localPath
--- a/graphql.config.ts
+++ b/graphql.config.ts
@@ -1,11 +1,11 @@
-function projectFactory(name: string, schemaName: string) {
+function projectFactory(name: string, schemaName: string, project: string = "app") {
   return {
-    schema: `./packages/app/src/cli/api/graphql/${name}/${schemaName}`,
-    documents: [`./packages/app/src/cli/api/graphql/${name}/queries/**/*.graphql`],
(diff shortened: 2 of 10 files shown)


packages/cli-kit/src/public/node/api/admin.ts
```diff
@@ -21,6 +23,34 @@ export async function adminRequest<T>(query: string, session: AdminSession, vari
...
+  }
+  const opts = {
+    url: adminUrl(session.storeFqdn, apiVersion),
+    api: 'Admin',
+    token: session.token,
+  }
+  const result = graphqlRequestDoc<TResult, TVariables>({...opts, query, variables})
+  return result
+}
+
```
karreiro, line 53: I've noticed the `fetchApiVersions` call had the `responseOptions: {handleErrors: false}`. It seems like that impacts how responses are handled and may control how they are [logged](https://github.com/Shopify/cli/blob/a2bb5689b8ad370774af9ed75dcac150d28cda03/packages/cli-kit/src/public/node/api/graphql.ts#L65).

I believe we could keep that in our new API, so I'd update this to something like this:

```suggestion
/**
 * Executes a GraphQL query against the Admin API. Uses typed documents.
 *
 * @param query - GraphQL query to execute.
 * @param session - Shopify admin session including token and Store FQDN.
 * @param variables - GraphQL variables to pass to the query.
 * @param version - API version.
 * @param responseOptions - Control how API responses will be handled.
 * @returns The response of the query of generic type <TResult>.
 */
export async function adminRequestDoc<TResult, TVariables extends Variables>(
  query: TypedDocumentNode<TResult, TVariables>,
  session: AdminSession,
  variables?: TVariables,
  version?: string,
  responseOptions?: GraphQLResponseOptions<TResult>,
): Promise<TResult> {
  let apiVersion = version
  if (!version) {
    apiVersion = await fetchLatestSupportedApiVersion(session)
  }
  const opts = {
    url: adminUrl(session.storeFqdn, apiVersion),
    api: 'Admin',
    token: session.token,
  }
  const result = graphqlRequestDoc<TResult, TVariables>({...opts, query, variables, responseOptions})
  return result
}
```

karreiro (approved):
Thank you for this PR, @catlee! LGTM, I've left only two minor suggestions :)

bin/get-graphql-schemas.js
```diff
@@ -70,14 +75,14 @@ async function fetchFileForSchema(schema, octokit) {
   try {
     // Fetch the file content from the repository
     const {data} = await octokit.repos.getContent({
+      mediaType: { format: "raw" },
```
shauns, line 78: 👍 nice

shauns (approved):
Have approved, but there's an open question about the right folder for this. Plus Guilherme's comments 👍

graphql.config.ts
```diff
@@ -73,5 +73,6 @@ export default {
     businessPlatformDestinations: projectFactory('business-platform-destinations', 'destinations_schema.graphql'),
     businessPlatformOrganizations: projectFactory('business-platform-organizations', 'organizations_schema.graphql'),
     appDev: projectFactory('app-dev', 'app_dev_schema.graphql'),
+    admin: projectFactory('admin', 'admin_schema.graphql', 'cli-kit'),
```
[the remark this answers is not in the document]
catlee: I really struggled to find a good place to put it.

Most of the APIs used by themes are defined in `packages/cli-kit/src/public/node/themes/api.ts`, and I couldn't figure out how to have the queries defined in the `theme` package but consumed by the api module in `cli-kit`.

Perhaps that argues for moving all this logic into `theme`?

packages/cli-kit/src/public/node/api/admin.ts
catlee: Great catch, thank you!

PR #4623, opened 2024-10-09
Migrate theme delete/update/publish to GraphQL 

lucyxiang:

Following the introduction of theme admin GraphQL endpoints, port the CLI to use GraphQL.

Extending what @catlee started here https://github.com/Shopify/cli/pull/4541

Migrate `themeDelete`, `themeUpdate`, and` themePublish` to use the admin graphql for both regular sessions and theme access sessions. Renamed uses of `delete/update/publishTheme` to `themeDelete/Update/Publish` for consistency.

Change themeDelete to return boolean (true on success, false on failure) instead of a theme.

To ship after Chris' code gen PR (shipped).

### How to test your changes?

`shopify theme delete`
Screenshot 2024-10-16 at 15 51 19

`shopify theme publish`
Screenshot 2024-10-16 at 15 50 52

 `shopify theme rename`
Screenshot 2024-10-16 at 15 50 32

### Measuring impact

### Checklist

### DIFF
--- a/packages/cli-kit/src/public/node/themes/api.ts
+++ b/packages/cli-kit/src/public/node/themes/api.ts
@@ -1,6 +1,10 @@
 import {storeAdminUrl} from './urls.js'
+import {composeThemeGid, parseGid} from './utils.js'
 import * as throttler from '../api/rest-api-throttler.js'
-import {restRequest, RestResponse} from '@shopify/cli-kit/node/api/admin'
+import {ThemeUpdate} from '../../../cli/api/graphql/admin/generated/theme_update.js'
+import {ThemeDelete} from '../../../cli/api/graphql/admin/generated/theme_delete.js'
+import {ThemePublish} from '../../../cli/api/graphql/admin/generated/theme_publish.js'
+import {restRequest, RestResponse, adminRequestDoc} from '@shopify/cli-kit/node/api/admin'
 import {AdminSession} from '@shopify/cli-kit/node/session'
 import {AbortError} from '@shopify/cli-kit/node/error'
 import {
@@ -77,32 +81,81 @@ export async function fetchChecksums(id: number, session: AdminSession): Promise
   return []
 }
 
-interface UpgradeThemeOptions {
-  fromTheme: number
-  toTheme: number
-  script?: string
-  session: AdminSession
+export async function themeUpdate(id: number, params: ThemeParams, session: AdminSession): Promise<Theme | undefined> {
+  const name = params.name
+  const input: {[key: string]: string} = {}
+  if (name) {
+    input.name = name
+  }
+
+  const {themeUpdate} = await adminRequestDoc(ThemeUpdate, session, {id: composeThemeGid(id), input})
+  if (!themeUpdate) {
+    // An unexpected error occurred during the GraphQL request execution
+    unexpectedGraphQLError('Failed to update theme')
+  }
+
+  const {theme, userErrors} = themeUpdate
+  if (userErrors.length) {
+    const userErrors = themeUpdate.userErrors.map((error) => error.message).join(', ')
+    throw new AbortError(userErrors)
+  }
+
+  if (!theme) {
+    // An unexpected error if neither theme nor userErrors are returned
+    unexpectedGraphQLError('Failed to update theme')
+  }
+
+  return buildTheme({
+    id: parseGid(theme.id),
+    name: theme.name,
+    role: theme.role.toLowerCase(),
+  })
 }
 
-export async function upgradeTheme(upgradeOptions: UpgradeThemeOptions): Promise<Theme | undefined> {
-  const {fromTheme, toTheme, session, script} = upgradeOptions
-  const params = {from_theme: fromTheme, to_theme: toTheme, ...(script && {script})}
-  const response = await request('POST', `/themes`, session, params)
-  return buildTheme(response.json.theme)
+export async function themePublish(id: number, session: AdminSession): Promise<Theme | undefined> {
+  const {themePublish} = await adminRequestDoc(ThemePublish, session, {id: composeThemeGid(id)})
+  if (!themePublish) {
+    // An unexpected error occurred during the GraphQL request execution
+    unexpectedGraphQLError('Failed to update theme')
+  }
+
+  const {theme, userErrors} = themePublish
+  if (userErrors.length) {
+    const userErrors = themePublish.userErrors.map((error) => error.message).join(', ')
+    throw new AbortError(userErrors)
+  }
+
+  if (!theme) {
+    // An unexpected error if neither theme nor userErrors are returned
+    unexpectedGraphQLError('Failed to update theme')
+  }
+
+  return buildTheme({
+    id: parseGid(theme.id),
+    name: theme.name,
+    role: theme.role.toLowerCase(),
+  })
 }
 
-export async function updateTheme(id: number, params: ThemeParams, session: AdminSession): Promise<Theme | undefined> {
-  const response = await request('PUT', `/themes/${id}`, session, {theme: {id, ...params}})
-  return buildTheme(response.json.theme)
-}
+export async function themeDelete(id: number, session: AdminSession): Promise<boolean | undefined> {
+  const {themeDelete} = await adminRequestDoc(ThemeDelete, session, {id: composeThemeGid(id)})
+  if (!themeDelete) {
+    // An unexpected error occurred during the GraphQL request execution
+    unexpectedGraphQLError('Failed to update theme')
+  }
 
-export async function publishTheme(id: number, session: AdminSession): Promise<Theme | undefined> {
-  return updateTheme(id, {role: 'main'}, session)
-}
+  const {deletedThemeId, userErrors} = themeDelete
+  if (userErrors.length) {
+    const userErrors = themeDelete.userErrors.map((error) => error.message).join(', ')
+    throw new AbortError(userErrors)
+  }
 
-export async function deleteTheme(id: number, session: AdminSession): Promise<Theme | undefined> {
-  const response = await request('DELETE', `/themes/${id}`, session)
-  return buildTheme(response.json.theme)
+  if (!deletedThemeId) {
+    // An unexpected error if neither theme nor userErrors are returned
+    unexpectedGraphQLError('Failed to update theme')
+  }
+
+  return true
 }
 
 async function request<T>(
@@ -198,6 +251,10 @@ function errorMessage(response: RestResponse): string {
   }
 
   return ''
+}
+
+function unexpectedGraphQLError(message: string): never {
+  throw new AbortError(message)
 }
 
 interface RetriableErrorOptions {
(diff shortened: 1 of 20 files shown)


catlee (approved):

packages/cli-kit/src/public/node/api/admin.ts
```diff
@@ -19,8 +25,9 @@ import {TypedDocumentNode} from '@graphql-typed-document-node/core'
 export async function adminRequest<T>(query: string, session: AdminSession, variables?: GraphQLVariables): Promise<T> {
   const api = 'Admin'
   const version = await fetchLatestSupportedApiVersion(session)
-  const url = adminUrl(session.storeFqdn, version)
-  return graphqlRequest({query, api, url, token: session.token, variables})
+  const url = adminUrl(session.storeFqdn, version, session)
+  const addedHeaders: {[header: string]: string} = headers(session)
```
graygilmore, line 29: The return for `headers` is already defined so we don't have to redeclare it here.
```suggestion
  const addedHeaders = headers(session)
```

packages/cli-kit/src/public/node/api/admin.ts
```diff
@@ -44,15 +51,24 @@ export async function adminRequestDoc<TResult, TVariables extends Variables>(
   if (!version) {
     apiVersion = await fetchLatestSupportedApiVersion(session)
   }
+
+  const addedHeaders: {[header: string]: string} = headers(session)
```
graygilmore, line 55: Same note as above:

```suggestion
  const addedHeaders = headers(session)
```

packages/cli-kit/src/public/node/api/admin.ts
```diff
@@ -61,7 +77,7 @@ export async function adminRequestDoc<TResult, TVariables extends Variables>(
  */
 async function fetchLatestSupportedApiVersion(session: AdminSession): Promise<string> {
   const apiVersions = await supportedApiVersions(session)
-  // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
```
graygilmore, line 64: Looks like this is still needed.

packages/cli-kit/src/public/node/themes/api.ts
```diff
@@ -74,32 +77,59 @@ export async function fetchChecksums(id: number, session: AdminSession): Promise
...
+    throw new Error(userErrors)
+  }
 
-export async function upgradeTheme(upgradeOptions: UpgradeThemeOptions): Promise<Theme | undefined> {
-  const {fromTheme, toTheme, session, script} = upgradeOptions
-  const params = {from_theme: fromTheme, to_theme: toTheme, ...(script && {script})}
-  const response = await request('POST', `/themes`, session, params)
-  return buildTheme(response.json.theme)
+  return buildTheme({
+    id: parseInt((theme.id as unknown as string).split('/').pop() as string, 10),
```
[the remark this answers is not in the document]
graygilmore: I'm kind of shocked we don't already have some `composeGid` and `parseGid` functions in the repo. Would be good to add them as a part of this transition I think. There's a `parseGid` you can rip from Hydrogen if you want: https://github.com/Shopify/hydrogen/blob/main/packages/cli/src/lib/gid.ts

packages/cli-kit/src/public/node/themes/api.ts
```diff
@@ -77,32 +80,54 @@ export async function fetchChecksums(id: number, session: AdminSession): Promise
...
-export async function publishTheme(id: number, session: AdminSession): Promise<Theme | undefined> {
-  return updateTheme(id, {role: 'main'}, session)
+export async function themePublish(id: number, session: AdminSession): Promise<Theme | undefined> {
+  const response = await adminRequestDoc(ThemePublish, session, {id: themeGid(id)})
+
+  const theme = response.themePublish?.theme
+  if (!theme) {
+    const userErrors = response.themePublish?.userErrors.map((error) => error.message).join(', ')
+    throw new AbortError(userErrors as string)
+  }
```
graygilmore, line 113: This is somewhere between a nit and a request. Ideally we're not casting `as string`. We can rearrange this a little bit:

```suggestion
  const {themePublish} = await adminRequestDoc(ThemePublish, session, {id: themeGid(id)})

  if (themePublish.userErrors.length) {
    const userErrors = themePublish.userErrors.map((error) => error.message).join(', ')
    throw new AbortError(userErrors)
  }

  if (!themePublish.theme) {
    // This is an unknown error that should never be hit but we should support for type safety
    throw new AbortError('Failed to publish theme')
  }
```

What's really weird is that `themePublish` can be `null` or `undefined` as defined in `packages/cli-kit/src/cli/api/graphql/admin/generated/theme_publish.ts` which makes me think maybe the types are being generated wrong? I feel like the object should either exist or whatever function was trying to run should have thrown an error. Does that make sense?

graygilmore (commented):
No 🎩 yet but looking good! Small concern around the types that are getting generated, would love to confirm that.

EvilGenius13:
Quick question: for the mutations, should they be in something like `api/graphql/admin/mutations/theme_delete.graphql`
instead of `api/graphql/admin/queries/theme_delete.graphql` since they are mutations and we might have queries at a later point?

packages/cli-kit/src/public/node/themes/api.ts
lucyxiang: When an error is raised during execution, graphql returns a `nil` payload (separate from `user_error`), which is why `themePublish` can be `null`/`undefined`

lucyxiang: I like your refactor, I think there I conflated `user_error` vs execution error (return nil). I will implement similarly for the other mutations 🙏

packages/cli-kit/src/public/node/themes/api.ts
```diff
@@ -77,32 +81,78 @@ export async function fetchChecksums(id: number, session: AdminSession): Promise
...
+  } else {
+    if (themeUpdate.userErrors.length) {
+      const userErrors = themeUpdate.userErrors.map((error) => error.message).join(', ')
+      throw new AbortError(userErrors)
+    }
+
+    const theme = themeUpdate.theme
+    if (!theme) {
+      // An unexpected error occured during the GraphQL request
+      throwUnexpectedGraphQLError('Failed to update theme')
```
lucyxiang, line 103: Will standardize message + update function name

packages/theme/src/cli/services/delete.test.ts
```diff
@@ -1,6 +1,6 @@
-import {deleteThemes, renderDeprecatedArgsWarning} from './delete.js'
```
lucyxiang, line 1: Mostly renaming from here on down

packages/cli-kit/src/public/node/themes/api.ts
```diff
@@ -77,32 +81,73 @@ export async function fetchChecksums(id: number, session: AdminSession): Promise
...
+        id: parseInt((theme.id as unknown as string).split('/').pop() as string, 10),
+        name: theme.name,
+        role: theme.role.toLowerCase(),
+      })
+    } else if (themeUpdate.userErrors.length) {
+      const userErrors = themeUpdate.userErrors.map((error) => error.message).join(', ')
+      throw new AbortError(userErrors)
+    } else {
+      // An unexpected error if neither theme nor userErrors are returned
+      unexpectedGraphQLError('Failed to update theme')
```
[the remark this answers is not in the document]
graygilmore: 👍🏻 ideally `adminRequestDoc` covers this case for all calls but we don't have to push that in this PR. I'm still convinced that the types in this repo aren't being generated quite right but I don't have time to poke at it yet. We'll just have to be extra defensive for now I think.

packages/cli-kit/src/public/node/themes/api.ts
```diff
@@ -77,32 +81,73 @@ export async function fetchChecksums(id: number, session: AdminSession): Promise
...
+    }
+  } else {
+    // An unexpected error occured during the GraphQL request execution
+    unexpectedGraphQLError('Failed to update theme')
+  }
 }
 
-export async function publishTheme(id: number, session: AdminSession): Promise<Theme | undefined> {
-  return updateTheme(id, {role: 'main'}, session)
+export async function themePublish(id: number, session: AdminSession): Promise<Theme | undefined> {
```
graygilmore, line 112: 🚲 🖌️ (nit) feel free to ignore completely, just sharing how I might have written this now that I see your new approach!

```ts
export async function themePublish(id: number, session: AdminSession): Promise<Theme | undefined> {
  try {
    const {themePublish} = await adminRequestDoc(ThemePublish, session, {id: themeGid(id)})

    if (!themePublish) throw new Error('')

    const {theme, userErrors} = themePublish

    if (userErrors.length) {
      throw new Error(userErrors.map((error) => error.message).join(', '))
    }

    if (!theme) throw new Error('')

    return buildTheme({
      id: themeGidToId(theme.id),
      name: theme.name,
      role: theme.role.toLowerCase(),
    })
  } catch (error: unknown) {
    if (error instanceof Error) {
      throw new AbortError(error.message)
    }
    throw new AbortError('An unknown error occurred')
  }
}
```

I like the try/catch here because it means we can just `throw` when we don't like the state and we don't end up in nested if/else statements.

graygilmore (approved):
no 🎩 but code looks good!

packages/cli-kit/src/public/node/themes/api.ts
```diff
@@ -77,32 +81,73 @@ export async function fetchChecksums(id: number, session: AdminSession): Promise
...
-  const {fromTheme, toTheme, session, script} = upgradeOptions
-  const params = {from_theme: fromTheme, to_theme: toTheme, ...(script && {script})}
-  const response = await request('POST', `/themes`, session, params)
-  return buildTheme(response.json.theme)
-}
+export async function themeUpdate(id: number, params: ThemeParams, session: AdminSession): Promise<Theme | undefined> {
+  const name = params.name
+  if (name === undefined) {
+    throw new AbortError('Theme name is required')
+  }
```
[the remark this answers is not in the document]
lucyxiang: Good call, currently you need `input` but its true `input` can be an empty hash `{}`

packages/cli-kit/src/public/node/themes/api.ts
lucyxiang: I used a mix of yours and guilherme's comments. I don't love throwing `Error('')` and then populating it later , but agreed with less if/else statements.

packages/cli-kit/src/public/node/themes/utils.ts
```diff
@@ -19,3 +19,11 @@ export async function promptThemeName(message: string) {
     defaultValue: defaultName,
   })
 }
+
+export function themeGid(id: number): string {
+  return `gid://shopify/OnlineStoreTheme/${id}`
+}
+
+export function themeGidToId(gid: string): number {
+  return parseInt((gid as unknown as string).split('/').pop() as string, 10)
```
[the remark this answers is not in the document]
lucyxiang: Good to go with what hydrogen uses, it seems more foolproof which I like. Thank you for sharing Gray!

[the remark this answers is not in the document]
lucyxiang: I missed this link in your previous comment, my bad 🙈 didn't mean to ignore it

gordonhirsch (approved):
Sorry for the delay.

PR #4628, opened 2024-10-10
Use app context in some basic commands

isaacroldan:
Use the generic app-context for all commands

Use `linkedAppContext` in these commands:
- `app-logs sources`
- `config link`
- `config use`
- `demo watcher`
- `env pull`
- `env show`
- `app info`

### How to test your changes?
Test the commands

### Measuring impact

### Checklist


packages/app/src/cli/commands/app/config/link.ts
```diff
@@ -34,14 +33,15 @@ export default class ConfigLink extends AppCommand {
...
-    const app = await loadApp({
-      specifications,
+    const {app} = await linkedAppContext({
       directory: flags.path,
-      userProvidedConfigName: undefined,
+      clientId: undefined,
+      forceRelink: false,
+      userProvidedConfigName: result.state.configurationFileName,
+      // Using report because a bad extension config shouldn't prevent the app from being linked
+      mode: 'report',
```
shauns, line 44: For all of these -- is this `mode` change accurate? Should it be? This feels like a more flexible context function. We should keep this surface as tight as we can and resist making it flexible. Like, maybe `env show` could work if your app is broken, but _should_ it?

shauns (changes requested):

packages/app/src/cli/commands/app/config/link.ts
isaacroldan: Updated, left the report mode only for `info`.

PR #4630, opened 2024-10-10
Filter app results by title

amcaplan:

Fixes https://github.com/Shopify/develop-app-inner-loop/issues/1861

Pairs with https://github.com/Shopify/shopify/pull/539421 to enable filtering by app title in the new API.

Note this should NOT be merged until that PR is merged.

Updates the format of the request to work with the new paginated field, and passes in the search term as a query for app title.

Screenshot 2024-10-09 at 20 26 55

Screenshot 2024-10-09 at 20 27 18

Note that each token is considered an independent search term, so we're searching for an app title that includes all mentioned tokens.

Screenshot 2024-10-09 at 20 28 01

### How to test your changes?

1. Point the new API to the `filter-using-pagination-system` branch (update as of 2024-11-14: this works on `main` or `dd` as well)
2. Create several apps with different names so you can see the search working (not necessary if pointing to production)
3. Run `USE_APP_MANAGEMENT_API=1 bin/spin pnpm shopify app config link --path /PATH/TO/YOUR/APP`
4. Select the option to connect to an existing app, and search for different app titles.

### Measuring impact

### Checklist

### DIFF
--- a/packages/app/src/cli/utilities/developer-platform-client/app-management-client.test.ts
+++ b/packages/app/src/cli/utilities/developer-platform-client/app-management-client.test.ts
@@ -15,6 +15,7 @@ import {CLI_KIT_VERSION} from '@shopify/cli-kit/common/version'
 import {fetch} from '@shopify/cli-kit/node/http'
 import {businessPlatformOrganizationsRequest} from '@shopify/cli-kit/node/api/business-platform'
 import {appManagementRequestDoc} from '@shopify/cli-kit/node/api/app-management'
+import {BugError} from '@shopify/cli-kit/node/error'
 
 vi.mock('@shopify/cli-kit/node/http')
 vi.mock('@shopify/cli-kit/node/api/business-platform')
@@ -240,5 +241,18 @@ describe('searching for apps', () => {
       })),
       hasMorePages: false,
     })
+  })
+
+  test("Throws a BugError if the response doesn't contain the expected data", async () => {
+    // Given
+    const orgId = '1'
+    vi.mocked(appManagementRequestDoc).mockResolvedValueOnce({})
+
+    // When
+    const client = new AppManagementClient()
+    client.token = () => Promise.resolve('token')
+
+    // Then
+    await expect(client.appsForOrg(orgId)).rejects.toThrow(BugError)
   })
 })
--- a/packages/app/src/cli/utilities/developer-platform-client/app-management-client.ts
+++ b/packages/app/src/cli/utilities/developer-platform-client/app-management-client.ts
@@ -268,7 +268,7 @@ export class AppManagementClient implements DeveloperPlatformClient {
     }
     const result = await appManagementRequestDoc(organizationId, query, await this.token(), variables)
     if (!result.appsConnection) {
-      throw new AbortError('Server failed to retrieve apps')
+      throw new BugError('Server failed to retrieve apps')
     }
     const minimalOrganizationApps = result.appsConnection.edges.map((edge) => {
       const app = edge.node


MitchDickinson:
CC @amcaplan 

Note that as part of https://github.com/Shopify/shopify/pull/539421, I introduced the connection API but I also _preserved_ the old `apps` field on the GraphQL schema. So now we have two ways to access apps:
- appsConnection
- apps
I did not want to have time pressure to ship this CLI PR while the CLI was in a broken state, so I took the safe path.

This PR is probably broken if it's using `apps` and assuming it's getting back a connection. I think the work outstanding here is going to be (in order to never break the latest CLI):
1. Ship this PR targeting `appsConnection`
2. If you want to rename `appsConnection` 
  - Update `apps` to _also_ be a connection type
  - Update CLI to use `apps`
  - Remove appsConnection from server

Or if you feel comfortable moving fast and breaking things, go ahead. I'm not opposed to that with good comms around it in our project channel. I just didn't want to take that on myself :)

MitchDickinson:
@amcaplan should we aim to get this shipped?

amcaplan:
I think this is ready to go. Currently it points to `appsConnection`. We can choose to leave it as is, or update `apps` to be a connection as well and then port this over.

pnpm-lock.yaml
```diff
@@ -279,7 +279,7 @@ importers:
         version: link:../app
       '@shopify/cli-hydrogen':
         specifier: 9.0.2
-        version: 9.0.2(react-dom@18.2.0)(react@18.2.0)
```
shauns, line 282: I don't think this file should have changed?

packages/app/src/cli/api/graphql/app-management/queries/apps.graphql
```diff
@@ -1,15 +1,22 @@
...
-    id
-    key
-    activeRelease {
-      id
-      version {
-        name
-        appModules {
-          ...ReleasedAppModule
+query listApps($query: String) {
+  appsConnection(query: $query, first: 50) {
```
shauns, line 2: Check -- we're going from all (presumably -- ignore if not) to 50. How does that match up to partners? Do we have sane ordering for these (e.g. most recent first)?

packages/app/src/cli/utilities/developer-platform-client/app-management-client.test.ts
```diff
@@ -187,3 +190,62 @@ describe('versionDeepLink', () => {
...
+        apiKey: `key-${index}`,
+        id: app.id,
+        organizationId: app.organizationId,
+        title: app.title,
+      })),
+      hasMorePages: false,
+    })
+  }
+
+  test('passes in a blank search if none is provided', async () => {
```
shauns, line 240: This is a case where a parameterised test would be best - https://vitest.dev/api/#test-each

packages/app/src/cli/utilities/developer-platform-client/app-management-client.test.ts
```diff
@@ -187,3 +190,62 @@ describe('versionDeepLink', () => {
...
+    await appSearchTest({query: undefined, queryVariable: ''})
```
shauns, line 241: Given that the default behaviour is `term = ''` can we get a test asserting that query stays empty in that case.

packages/app/src/cli/utilities/developer-platform-client/app-management-client.ts
```diff
@@ -257,10 +257,21 @@ export class AppManagementClient implements DeveloperPlatformClient {
...
+    const variables = {
+      query: term
+        .split(' ')
+        .filter((word) => word)
+        .map((word) => `title:${word}`)
+        .join(' '),
+    }
+    const result = await appManagementRequestDoc(organizationId, query, await this.token(), variables)
+    if (!result.appsConnection) {
+      throw new AbortError('Server failed to retrieve apps')
```
shauns, line 271: I think this is a `BugError`. Things are pretty broken if this happens

shauns (approved):

packages/app/src/cli/api/graphql/app-management/queries/apps.graphql
amcaplan: 1. `first` is required, so we need to pass some kind of value here.
2. 50 seems reasonable, I don't think we should have users scrolling manually beyond that.
3. I thought I recalled Partners using 50 as the max, but I can't seem to find it at the moment... That said, client-side we limit it to 25 [here](https://github.com/Shopify/cli/blob/main/packages/app/src/cli/api/graphql/find_org.ts#L9). So it's the same order of magnitude. Maybe worth checking, but it's easily changed and not a reason to hold up the PR at the moment IMO.
4. Order is the default dictated by the server, currently the most recently updated. It will match what they see in the dashboard; it feels like consistency is good.

packages/app/src/cli/utilities/developer-platform-client/app-management-client.test.ts
amcaplan: Oh, neat! I wasn't aware of how to do this.

pnpm-lock.yaml
amcaplan: I honestly have no idea sometimes why these things change... maybe my local version of pnpm is slightly off? But I'm on 8.15.7 so 🤷‍♂️

PR #4631, opened 2024-10-10
Use codegen types for App Management client

shauns:

Reimplementing contracts in GraphQL schema & Typescript types is a recipe for mistakes: let's use our codegen infrastructure instead.

- Uses codegen-ned types for all API calls in `app-management-client.ts` [^1]
- Drops manually built types
- Removes the ability to call app management API without these generated types

### How to test your changes?

CI -- functionally equivalent.

### Measuring impact

### Checklist

[^1]: Except for loading the organisation beta flags. We dynamically build the query at runtime for that one.


packages/app/src/cli/api/graphql/app_deploy.ts
isaacroldan, line 1: These files with gql definitions that are updated but not deleted, are for Partners queries? We don't want to migrate those to the new system no?

shauns: I think we should but outside of this PR. I've just been looking into this now. The manual changes made in these few files should be safe ones.

isaacroldan (approved):
Didn't 🎩 but code looks good 👌

PR #4638, opened 2024-10-11
Change CLI auth pages

davejcameron:

The current CLI auth pages are a green full screen with simple test. This doesn't match any of the colours we currently use so I wanted to just make it a card.

| Before | After |
| -- | -- |
| image |  image |

### How to test your changes?

Open the files up on your local machine, they are static

### Measuring impact

### Checklist


amcaplan:
When do we still use these screens? We should be using device auth by default now.

gonzaloriestra:
Right, they are being removed here: https://github.com/Shopify/cli/pull/4604

shauns:
Yep. We can close this PR -- device auth has been default on recent CLIs for a while and uses Shopify hosted and branded pages, and the auth flow that shows this screen is now fully deprecated. Thank you though!

PR #4639, opened 2024-10-11
Config state after linking should include the real app directory

isaacroldan:
A bug was introduced the other day when, to support app-context, we made `link` return a config state.

Config state returned from link now returns the actual directory of the app, and not the one passed by the user (which might or might not be the app directory!)

### How to test your changes?
- Have an app with an empty toml (just `scopes = ""`)
- Add a function
- Make sure that the toml is still empty, if it's not, force it.
- run 
```
pnpm shopify app function schema --stdout --path <path_to_function> --client-id <a_client_id>
```

### Measuring impact

### Checklist


andrewhassan (approved):
🎩 Thank you for fixing this!!!

shauns (approved):

PR #4658, opened 2024-10-16
Stricter ESLint ruleset

shauns:

Make ESLint rules stricter to help avoid errors creeping in

- Adds plugin:@typescript-eslint/strict and plugin:@typescript-eslint/stylistic
- Ignores some pre-existing possible-incorrect behaviour around `!` operator -- assumption is that lack of runtime errors today implies these are safe.

### Measuring impact

### Checklist


packages/app/src/cli/models/app/app.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable @typescript-eslint/no-non-null-assertion */
```
alexanderMontague, line 1: I don't love blanketing files especially when we're trying to prevent new occurrences. I know the file is large too, can we get GPT to write us up a codemod/script to add individually? Same with a few other files

alexanderMontague (approved):
This is a good start, thanks

packages/app/src/cli/models/app/app.ts
shauns: Generally I tried to exercise some judgement with these -- if there were lots of dotted around breakages of this rule I did a file ignore rather than have loads of comments. I figure this was better than sacrificing readability too much. I don't really mind switching to line ignores really. Ultimately its in the eye of the beholder: Your call?

Another option would be to make these warnings. We don't use that right now, everything is errors or off. The downside of warnings is in net-new code, they can blow right through them if they like. Ignoring errors leaves a footprint. We'd have to get warnings to show up here in GH for those to be viable, I think

packages/app/src/cli/models/app/app.test-data.ts
```diff
@@ -109,7 +109,6 @@ export function testApp(app: Partial<AppInterface> = {}, schemaType: 'current' |
     dotenv: app.dotenv,
     errors: app.errors,
     specifications: app.specifications ?? [],
-    // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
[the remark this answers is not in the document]
isaacroldan: ah, didn't realize it was just for this file 👍

PR #4668, opened 2024-10-17
Defer remote app creation during `app init` to include all default creation options.

isaacroldan:
With the changes to force a `link` after `app init`, we forgot to include the default creation options during app creation. Meaning that the app creation needs to happen after the cloning to have all the necessary context.

- Move template/prompts before org/app selection
- Force authenticate before any prompt
- Defer app creation (if you select to create a new app) until the project has been created successfully locally, that way we can know the scopes/default app url for app creation. And also don't pollute your account with zombie apps if the cloning fails.

### How to test your changes?
With the snapshot, just try `shopify app init`, try being logged out and with different templates (check that the toml is correctly filled)

### Measuring impact

### Checklist


packages/app/src/cli/commands/app/init.ts
```diff
@@ -103,30 +108,65 @@ export default class Init extends AppCommand {
...
+      forceRelink: false,
       userProvidedConfigName: undefined,
+      unsafeReportMode: false,
     })
 
     return {app}
   }
 }
+
+export type SelectAppOrNewAppNameResult =
```
shauns, line 137: 👌

shauns (approved):

PR #4669, opened 2024-10-17
[Theme] Fix Liquid asset hot reload

### LINKED ISSUE #2614: [Bug]: `assets/theme.css.liquid` and `assets/theme.js.liquid` files not hot reloading

Theme

### Expected behavior

Changes done in the theme.css.liquid should sync and reflect in the front end.

### Actual behavior

Any changes in the theme.css.liquid are not synced. Shopiy CLI shows an update has been made for assets/theme.css.liquid, but they do not show up in the front end. 

Changes can be seen in the file when looking directly at the admin file.

My current workaround is to create a new .css file (without the .liquid format) to apply styling. 

### Reproduction steps

1. Make changes to theme.css.liquid
2. Inspect front-end to see style not updated
3.

### Operating System

Mac OS Monterey 12.6.3

3.48.1

### Shell

zsh

20.5.0

Liquid, CSS

frandiox:

Fixes #2614 

When a missing .liquid asset that is referenced in the app is created anew, the server will try to hot reload it and fail.
Pairing with @jamesmengo , we found out during the first rendering, SFR renders `?1234` for the file querystring, which returns a 404.

The problem is that even after creating the file and uploading it, SFR still renders the wrong `?1234` instead of `?v=timestamp`. This makes the request fail again with 404 even though the file already exists. This might be related to caching at the SFR level, I'm not sure about it.

This PR makes 3 changes:
1. Use HotReload for CSS instead of full page reload.
2. During CSS HotReload, it replaces the wrong `?1234` with a new `?v=` timestamp.
3. The same approach doesn't work for JS assets because we only do a full refresh. Therefore, the fix here is to use our proxy to rewrite `?1234` queries to `?v=` timestamp queries. This fix would probably also work for CSS assets, but I'm keeping the previous one since it's cleaner for CSS.

### How to test your changes?

1. Reference a missing `.liquid.css` file from your theme.
2. Load the page, see it provokes a 404
3. Create the new file (or move it from another folder). It should now hot reload accordingly.

### Measuring impact

### Checklist

### DIFF
--- a/.changeset/empty-hats-own.md
+++ b/.changeset/empty-hats-own.md
@@ -1,0 +1,5 @@
+---
+'@shopify/theme': patch
+---
+
+Fix hot reloading of `.css.liquid` assets.
--- a/.changeset/empty-hats-own.md
+++ b/.changeset/empty-hats-own.md
@@ -2,4 +2,4 @@
 '@shopify/theme': patch
 ---
 
-Fix hot reloading of `.css.liquid` assets.
+Fix hot reloading of `.css.liquid` and `.js.liquid` assets.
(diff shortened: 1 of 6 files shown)


karreiro (approved):
Thank you for this PR, @frandiox! 🔥🔥🚀

PR #4635, opened 2024-10-10
[Themes] Wait for remote file deletion before emitting reload event for liquid assets

jamesmengo:
### WHY are these changes introduced?
- Fixes https://github.com/Shopify/develop-advanced-edits/issues/368

Waits for `liquid assets` to be deleted on remote theme before triggering hot reload

### WHAT is this pull request doing?
- Modifies the file deletion process to emit 'unlink' events differently for liquid and non-liquid assets
  - For non-liquid assets, 'unlink' event is emitted immediately
  - For liquid assets, 'unlink' event is emitted after the remote deletion is confirmed
- Updates tests

### How to test your changes?
- Create and set up a theme that renders a liquid asset (refer to issue for reference)
- Delete the liquid asset - the changes should be immediately visible.

**Before**

https://github.com/user-attachments/assets/567d0c5d-7aad-41bc-a3e9-b266db26f4a9

**After**

https://github.com/user-attachments/assets/77b5f314-af68-46a3-b3c4-3e7edfb202ba

### Measuring impact

How do we know this change was effective? Please choose one:

- [x] n/a - this doesn't need measurement, e.g. a linting rule or a bug-fix

### Checklist

- [x] I've considered possible cross-platform impacts (Mac, Linux, Windows)
- [x] I've considered possible [documentation](https://shopify.dev) changes

cc: @Shopify/advanced-edits

PR #4674, opened 2024-10-17
[Themes] Improve asset deletion event handling in the theme dev server

jamesmengo:

Follow up on [[Themes] Wait for remote file deletion before emitting reload event for liquid assets](https://github.com/Shopify/cli/pull/4635) with some cleanup of code

- Renames `handleFileUnlink` to `handleFileDelete` 
- Prevents emitting 'unlink' event twice for non-liquid assets

### How to test your changes?
1. Add a `console.log` in `handleFileDelete` in `server.ts`
2. Build
3. Start the development server
4. Delete a non-liquid asset file and ensure that only a single `unlink` is emitted
5. Observe that rather than a full refresh, we simply perform a hot reload in the browser

https://github.com/user-attachments/assets/732ffb6b-cc6a-4da9-b9e5-380050f128ff

### Measuring impact

### Checklist

### DIFF
--- a/packages/cli-kit/src/public/node/themes/api.test.ts
+++ b/packages/cli-kit/src/public/node/themes/api.test.ts
@@ -280,6 +280,26 @@ describe('deleteThemeAsset', () => {
     expect(restRequest).toHaveBeenCalledWith('DELETE', `/themes/${id}/assets`, session, undefined, {'asset[key]': key})
     expect(output).toBe(true)
   })
+
+  test('throws an AbortError when the server responds with a 403', async () => {
+    // Given
+    const id = 123
+    const key = 'config/settings_data.json'
+    const message = 'You are not authorized to edit themes on "my-shop.myshopify.com".'
+
+    vi.mocked(restRequest).mockResolvedValue({
+      json: {message},
+      status: 403,
+      headers: {},
+    })
+
+    // When
+    const deletePromise = () => deleteThemeAsset(id, key, session)
+
+    // Then
+    await expect(deletePromise).rejects.toThrow(new AbortError(message))
+    expect(restRequest).toHaveBeenCalledWith('DELETE', `/themes/${id}/assets`, session, undefined, {'asset[key]': key})
+  })
 })
 
 describe('deleteTheme', () => {
--- a/packages/theme/src/cli/utilities/theme-environment/hot-reload/server.test.ts
+++ b/packages/theme/src/cli/utilities/theme-environment/hot-reload/server.test.ts
@@ -135,7 +135,7 @@ describe('hot-reload server', () => {
     // We wait for syncing to finish on liquid assets before emitting a full reload event:
     await nextTick()
     expect(unlinkSyncSpy).toHaveBeenCalled()
-    expect(hotReloadEvents.at(-1)).toMatch(`data: {"type":"full","key":"${liquidAssetKey}"}`)
+    expect(hotReloadEvents.at(-1)).toMatch(`data: {"type":"css","key":"${liquidAssetKey.replace('.liquid', '')}"}`)
     // Removes the CSS Liquid file from memory:
     expect(getInMemoryTemplates(ctx)).toEqual({})
 


packages/theme/src/cli/utilities/theme-fs.ts
```diff
@@ -210,16 +204,22 @@ export function mountThemeFileSystem(root: string, options?: ThemeFileSystemOpti
...
+          syncPromise
+            .then((didSync) => {
+              if (didSync) fn()
+            })
+            .catch(() => {})
+        },
+      })
+    } else {
+      emitEvent('unlink', {fileKey})
+    }
```
frandiox, line 222: To my understanding, this is just making the `onSync` property available only for liquid assets, right?
Does it hurt to make it available to every asset, though?

I think the logic that decides when to do the hot reload is already in place in `hot-reload/server`, perhaps we shouldn't leak it here?

frandiox (commented):
Thanks for the follow up!

packages/theme/src/cli/utilities/theme-environment/hot-reload/server.ts
```diff
@@ -110,15 +116,18 @@ export function setupInMemoryTemplateWatcher(ctx: DevServerContext) {
...
-  const handleFileUnlink = ({fileKey, onSync}: ThemeFSEventPayload<'unlink'>) => {
-    const extension = extname(fileKey)
-    if (isAsset(fileKey) && extension === '.liquid') {
+  const handleFileDelete = ({fileKey, onSync}: ThemeFSEventPayload<'unlink'>) => {
+    // Liquid assets are proxied, so we need to wait until the file has been deleted on the server before reloading
+    const isLiquidAsset = isAsset(fileKey) && extname(fileKey) === '.liquid'
+    if (isLiquidAsset) {
       onSync?.(() => {
-        triggerHotReload(fileKey, ctx)
+        triggerHotReload(fileKey.replace('.liquid', ''), ctx)
```
frandiox, line 130: Do we handle deletes in the `client.ts` file? Or are we just making a new `v=<timestamp>` and getting a 404?

packages/theme/src/cli/utilities/theme-environment/hot-reload/server.ts
```diff
@@ -1,3 +1,7 @@
+/* eslint-disable @typescript-eslint/no-confusing-void-expression */
```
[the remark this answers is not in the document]
frandiox: Are we ok with disabling linting rules for our files? I try to comply when I can, except in some rules related to async promises... 🤔

packages/theme/src/cli/utilities/theme-environment/hot-reload/server.ts
jamesmengo: > are we just making a new v=<timestamp> and getting a 404?

If I'm not mistaken, this is what's happening behind the scenes. 

Also, I'm removing the `.liquid` here so that we will trigger a `css` reload for `css.liquid` files rather than a `full` reload

Alternatively, I can move the above logic into`triggerHotReload` instead of replacing the `extname`. wdyt?

packages/theme/src/cli/utilities/theme-environment/hot-reload/server.ts
```diff
@@ -287,7 +293,7 @@ function hotReloadSections(key: string, ctx: DevServerContext) {
     const content = ctx.localThemeFileSystem.files.get(key)?.value
     if (content) {
       const sections: SectionGroup | undefined = parseJSON(content, null)?.sections
-      for (const sectionName of Object.keys(sections || {})) {
+      for (const sectionName of Object.keys(sections ?? {})) {
```
jamesmengo, line 296: @frandiox can you just doublecheck that these are safe?

I'm pretty confident but would love another set of eyes, since we are intentionally using `||` in [some places](https://github.com/Shopify/cli/pull/4674/files#diff-d857e2b583dcee2197fc538eb4d35aabb5ce83cee18ebb970ece7854dda8ba07R139-R140)

frandiox: I think it should be OK. Even if the JSON is malformed and `sections` is an empty string or similar it should still work 😅

karreiro (approved):
Thank you, @jamesmengo!

shauns (approved):

PR #4686, opened 2024-10-18
Prevent rendering remote file deletion error when file has already be…

jamesmengo:
…en removed

Fixes #0000 

### Measuring impact

### Checklist


packages/theme/src/cli/utilities/theme-fs.ts
```diff
@@ -192,7 +191,14 @@ export function mountThemeFileSystem(root: string, options?: ThemeFileSystemOpti
 
     const syncPromise = deleteThemeAsset(Number(themeId), fileKey, adminSession)
       .then(async (success) => {
```
jamesmengo, line 193: context: we intentionally chose to fire off this request, even if the `handleFileDelete` call was a result of a remote file deletion

Alternatives we considered
- Adding a call the check if the file exists before initializing deletion -  (_added too much latency making 2 calls instead of 1)_
- Keeping track of file deletion source - (_this is also an option, but we decided against introducing more complexity here_)

karreiro:
Closed in favor https://github.com/Shopify/cli/pull/4691 (to avoid ambiguity). Thank you for opening both options, @jamesmengo!

PR #4691, opened 2024-10-18
Prevent rendering remote file deletion error when file has already been removed 

jamesmengo:
https://github.com/Shopify/develop-advanced-edits/issues/375

- Prevents unnecessary warning messages when a file is already absent from the remote theme. This can happen when deleting a file on the remote theme and the `--theme-editor-sync` flag is provided

### How to test your changes?
1) Run `shopify-dev theme dev --theme-editor-sync`
2) Delete a file on the remote theme via the code editor
3) Observe that we don't report the failed deletion

https://github.com/user-attachments/assets/f8016a13-f562-4756-84c4-f19d9e69dfdb

### Measuring impact

### Checklist

### DIFF
--- a/packages/cli-kit/src/public/node/themes/api.ts
+++ b/packages/cli-kit/src/public/node/themes/api.ts
@@ -51,7 +51,7 @@ export async function deleteThemeAsset(id: number, key: Key, session: AdminSessi
   const response = await request('DELETE', `/themes/${id}/assets`, session, undefined, {
     'asset[key]': key,
   })
-  return Boolean(response.json.message)
+  return response.status === 200
 }
 
 export async function bulkUploadThemeAssets(
--- a/.changeset/little-ladybugs-fix.md
+++ b/.changeset/little-ladybugs-fix.md
@@ -1,0 +1,5 @@
+---
+'@shopify/theme': patch
+---
+
+Fix an issue with `shopify theme dev --theme-editor-sync` to prevent showing a remote file deletion error when the file has already been removed
--- a/packages/cli-kit/src/public/node/themes/api.test.ts
+++ b/packages/cli-kit/src/public/node/themes/api.test.ts
@@ -261,7 +261,7 @@ describe('deleteThemeAsset', () => {
     expect(output).toBe(true)
   })
 
-  test('returns empty object when attemping to delete an nonexistent asset', async () => {
+  test('returns true when attemping to delete an nonexistent asset', async () => {
     // Given
     const id = 123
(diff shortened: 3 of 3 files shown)


packages/cli-kit/src/public/node/themes/api.ts
```diff
@@ -51,7 +51,7 @@ export async function deleteThemeAsset(id: number, key: Key, session: AdminSessi
   const response = await request('DELETE', `/themes/${id}/assets`, session, undefined, {
     'asset[key]': key,
   })
-  return Boolean(response.json.message)
+  return response.status === 200
```
graygilmore, line 54: Just needs a test updated I think

graygilmore (approved):
Love the simplicity of this one

packages/cli-kit/src/public/node/themes/api.ts
```diff
@@ -51,7 +51,7 @@ export async function deleteThemeAsset(id: number, key: Key, session: AdminSessi
   const response = await request('DELETE', `/themes/${id}/assets`, session, undefined, {
     'asset[key]': key,
   })
-  return Boolean(response.json.message)
```
jamesmengo, line 54: I've validated with the core API that this should be safe.

The `asset api` will return a `200` if the asset is deleted, but will also return a `200` if an asset already does not exist. The latter case would include an empty message.

I believe I initially made this decision to provide us with the ability to distinguish between the two cases. Seems like it may not be necessary. 

cc: @karreiro

packages/cli-kit/src/public/node/themes/api.ts
frandiox, line 54: Let's add a comment explaining the API behavior on top of this?

frandiox (approved):
If the API officially supports this use case (and looks like it does), I think we can go with this approach 👍

packages/cli-kit/src/public/node/themes/api.ts
karreiro: Personally, I believe we don't need the comment because the unit tests already document the most intriguing scenarios (which involve the two kinds of 200 responses).

I will proceed with the merge to unblock the release, but please feel free to add an extra test case to the 403 scenario and/or include the comments in a following PR, @jamesmengo :)

karreiro (approved):
Thank you for this PR and investigation reasoning, @jamesmengo! I agree this is the best option as well. The code LGTM and works as expected 🎩

karreiro:
@Shopify/app-inner-loop Could you please review this PR?(this is important to the next minor release). Thank you!

PR #4696, opened 2024-10-21
Fix the URL output to add a new product

nickwesselman:

Currently the URL output by this Checkout UI extensions error message during `shopify app dev` has an extra `https://`.

```
╭─ error ───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│                                                                                                                               │
│  Could not find a product variant                                                                                             │
│                                                                                                                               │
│  Your store needs to have at least one product to test a 'checkout_ui' extension                                              │
│                                                                                                                               │
│  You can add a new product here: https://https://nw-dd-dev-1.myshopify.com/admin/products/new                                 │
│                                                                                                                               │
╰───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
```

I assume that we expect the Store FQDN here to already have a protocol.

### DIFF
--- a/packages/app/src/cli/utilities/extensions/fetch-product-variant.ts
+++ b/packages/app/src/cli/utilities/extensions/fetch-product-variant.ts
@@ -16,7 +16,7 @@ export async function fetchProductVariant(store: string) {
     throw new AbortError(
       'Could not find a product variant',
       `Your store needs to have at least one product to test a 'checkout_ui' extension\n
-You can add a new product here: https://${store}/admin/products/new`,
+You can add a new product here: ${store}/admin/products/new`,
     )
   // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
   const variantURL = result.products.edges[0]!.node.variants.edges[0]!.node.id


shauns (changes requested):

packages/app/src/cli/utilities/extensions/fetch-product-variant.ts
```diff
@@ -12,12 +14,18 @@ export async function fetchProductVariant(store: string) {
   const adminSession = await ensureAuthenticatedAdmin(store)
   const result: FindProductVariantSchema = await adminRequest(FindProductVariantQuery, adminSession)
   const products = result.products.edges
-  if (products.length === 0)
+  if (products.length === 0) {
+    const normalizedUrl = `https://${await normalizeStoreFqdn(store)}/admin/products/new`
```
aledalgrande, line 18: I've heard that using `/admin` can create quirks with Identity and should not be used, instead we should use `admin.shopify.com` for all new code. Was it @alxclark who mentioned it maybe?

shauns: We don't have a function to convert foo.myshopify.com to admin.shopify.com/store/foo .  Am sure it's easy enough to add. There are likely a lot of places in CLI and in other repos, as well as a lot of 3P code, that aren't doing this though.

nickwesselman: Yeah I think @shaun is right that this is something that should be addressed separately as I'm seeing various `/admin` URLs throughout the CLI.

shauns (approved):

PR #4699, opened 2024-10-21
Auto create contract based specs

isaacroldan:
We want to simplify the process of defining new extension specs, this completely remove the need to update the CLI for simple contract-based extensions.

Before this change we were forced to define a local spec for every extension type, and then when fetching the remote ones, we would merge them.

Now: 
- For remote specs that don't have a local version, but define a validationSchema, automatically create a `contract-based` local extension and merge it with the remote one.
- Also: automatically include the `localization` feature if the jsonSchema expects a localization property.

### How to test your changes?
Create a marketing_activity extension, this spec only exists remotely and has a validationSchema. You should be able to deploy without problems. (you might need a beta flag, ask me)

### Measuring impact

### Checklist

### DIFF
--- a/packages/app/src/cli/services/generate/fetch-extension-specifications.ts
+++ b/packages/app/src/cli/services/generate/fetch-extension-specifications.ts
@@ -1,6 +1,10 @@
 import {loadLocalExtensionsSpecifications} from '../../models/extensions/load-specifications.js'
 import {FlattenedRemoteSpecification, RemoteSpecification} from '../../api/graphql/extension_specifications.js'
-import {ExtensionSpecification, RemoteAwareExtensionSpecification} from '../../models/extensions/specification.js'
+import {
+  createContractBasedModuleSpecification,
+  ExtensionSpecification,
+  RemoteAwareExtensionSpecification,
+} from '../../models/extensions/specification.js'
 import {DeveloperPlatformClient} from '../../utilities/developer-platform-client.js'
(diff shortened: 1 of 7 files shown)


packages/app/src/cli/services/generate/fetch-extension-specifications.ts
```diff
@@ -56,20 +61,26 @@ async function mergeLocalAndRemoteSpecs(
...
-    if (remoteSpec) {
-      const merged = {...spec, ...remoteSpec, loadedRemoteSpecs: true} as RemoteAwareExtensionSpecification &
-        FlattenedRemoteSpecification
+  // Iterate over the remote specs and merge them with the local ones
+  // If the local spec is missing, and the remote one has a validation schema, create a new local spec using contracts
+  const updated = remote.map(async (remoteSpec) => {
+    let localSpec = local.find((local) => local.identifier === remoteSpec.identifier)
+    if (!localSpec && remoteSpec.validationSchema?.jsonSchema) {
+      const normalisedSchema = await normaliseJsonSchema(remoteSpec.validationSchema.jsonSchema)
+      const hasLocalization = normalisedSchema.properties?.localization !== undefined
```
shauns, line 70: Should we be porting this to something on the specifications API instead? The CLI feature list in general I suppose. Should we block this on updating the API or proceed without?

shauns (approved):

PR #4706, opened 2024-10-21
Update contracts to support newer versions of json + bugfix

reemrazak:

For marketing extensions we'd like to make use of `unevaluatedProperties` which is supported on json 2019 and newer. This PR adds supports for 2019 + 2020 according to the ajv docs listed here: https://ajv.js.org/guide/schema-language.html#draft-2019-09-and-draft-2020-12

There is also a fix for a bug I encountered trying to test an invalid config:
```
error ──────────────────────────────────────────────────────────────────────────────╮
│                                                                                      │
│  Cannot convert undefined or null to object                                          │
│                                                                                      │
│  To investigate the issue, examine this stack trace:                                 │
│    at keys                                                                           │
│    at (cli-kit/src/public/node/json-schema.ts:195)                                   │
│      const candidatesObjectProperties =                                              │
│      Object.keys(candidateSchemaFromUnion.properties)                                │
│    at map                                                                            │
│    at simplifyUnionErrors (cli-kit/src/public/node/json-schema.ts:186) 
```

### Measuring impact

### Checklist


reemrazak:
I believe [this](https://github.com/Shopify/shopify/blob/0e73dde6c359aa776062a0656feb753259f080e5/areas/core/shopify/components/apps/essentials/app/models/apps/app_modules/specifications/contract.rb#L27) is the spot where the config is being validated in core.

Unfortunately, the [json schema gem](https://github.com/Shopify/shopify/blob/0e73dde6c359aa776062a0656feb753259f080e5/areas/core/shopify/Gemfile#L218) which is being using there is unmaintained and doesn't support newer versions of json (see [this](https://github.com/brandur/json_schema/issues/99) issue)


Although it doesn't seem to actually be blocking deploy of the config for whatever reason (still able to deploy in spin and pass it through the JSON validation), it feels risky to make this change and have an inconsistency in version support.


**We're fine with not including `unevaluatedProperties` in our schema relative to the effort of making the core changes**


I can still merge the bugfix that's in this PR

Issue #4710, opened 2024-10-22
v3.69.0 ignores files not covered by .shopifyignore file.


dev-mr:
### Please confirm that you have:

- [x] Searched [existing issues](.) to see if your issue is a duplicate. (If you’ve found a duplicate issue, feel free to add additional information in a comment on it.)
- [x] Reproduced the issue in [the latest CLI version](https://www.npmjs.com/package/@shopify/cli).

### In which of these areas are you experiencing a problem?

Theme

### Expected behavior

`shopify-cli` should correctly interpret `.shopifyignore` and follow the rules specified.

### Actual behavior

`cli` runs but ignores all theme files like `snippets`, `templates`, etc. 

None of these folders are mentioned in `.shopifyignore`

### .shopifyignore used
```
./config/settings_data.json
./sections/*.json
./templates/*.json
./templates/**/*.json
./locales/*.json
!./locales/*.schema.json
```

cli ignores `layout/theme.liquid` saying that it's ignored in `.shopifyignore` file.

> Ignoring theme file layout/theme.liquid via .shopifyignore...

### Verbose output

$ ./node_modules/.bin/shopify theme dev --environment development --verbose
2024-10-22T00:46:12.715Z: Running command theme dev
2024-10-22T00:46:12.725Z: Reading the content of file at shopify.theme.toml...
╭─ info ─────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│                                                                                                                    │
│  Using applicable flags from development environment:                                                              │
│                                                                                                                    │
│    • store: xxxxxxxxxxxxx.myshopify.com                                                                            │
│    • theme: xxxxxxxxxxxx                                                                                           │
│    • password: ********f445                                                                                        │
│                                                                                                                    │
╰────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯

2024-10-22T00:46:12.798Z: Ensuring that the user is authenticated with the Theme API with the following scopes:
[]

2024-10-22T00:46:13.326Z: Request to https://theme-kit-access.shopifyapps.com/cli/admin/api/unstable/themes.json?fields=id%2Cname%2Crole%2Cprocessing completed in 526 ms
With response headers:
 - cache-control: max-age=0, private, must-revalidate
 - content-type: application/json; charset=utf-8
 - etag: W/"9ce9ac52f171155e04e45e680c0d0bba"
 - server-timing: processing;dur=226, socket_queue;dur=3.116, cfRequestDuration;dur=434.000015
[...]

PR #4732, opened 2024-10-23
Fix bug related to using negated patterns in shopify ignore

EvilGenius13:

Fixes https://github.com/Shopify/cli/issues/4710

When a user is using `!` in their `.shopifyignore` file to ensure that file is not ignored, it had unintended side effects such as ignoring other files and folders, and not actually negating the file(s) it is supposed to.

Add some additional logic around separating negated options and normal ignored options.
Essentially, we all regular ignored logic in the `.shopifyignore` file and then run any negated logic afterwards which undoes the ignore.

Ex:
We have these files
`assets/basic.css`
`assets/complex.css`
`assets/my_special.css`

We want to ignore all files in the assets folder **except** my_special.css
Our `.shopifyignore` will look like
```git
assets/*.css,
!assets/my_special.css
```

Even though we will initially ignore the entire assets folder, we will then "undo" ignoring `my_special.css` and add it back to the files list to be synced.

This behaviour has been replicated when using flags `--only` and `--ignore`.

### How to test your changes?

Option #1: 
Run the new tests 
`npx vitest run packages/theme/src/cli/utilities/asset-ignore.test.ts`

Option #2: 
Pull down the branch
Build the branch
create a `.shopifyignore` file if you don't already have one
Ignore a folder (Ex: assets/*.css)
Negate a file from that folder (Ex: !assets/my_special.css)

### Post-release steps

I want to keep an eye out for any additional edge cases that might come up since we have many examples of developers writing different patterns/regexes to build their ignore files.

### Measuring impact

### Checklist

### DIFF
--- a/packages/theme/src/cli/utilities/asset-ignore.test.ts
+++ b/packages/theme/src/cli/utilities/asset-ignore.test.ts
@@ -174,4 +174,114 @@ describe('asset-ignore', () => {
       ])
     })
   })
+
+  describe('applyIgnoreFilters with negated patterns', () => {
+    test(`only negating a file does not produce side effects of other ignored files`, () => {
+      const options = {
+        ignoreFromFile: ['!assets/basic.css'],
+      }
+
+      const actualChecksums = applyIgnoreFilters(checksums, options)
+
+      expect(actualChecksums).toEqual([
+        {key: 'assets/basic.css', checksum: '00000000000000000000000000000000'},
+        {key: 'assets/complex.css', checksum: '11111111111111111111111111111111'},
+        {key: 'assets/image.png', checksum: '22222222222222222222222222222222'},
+        {key: 'config/settings_data.json', checksum: '33333333333333333333333333333333'},
+        {key: 'config/settings_schema.json', checksum: '44444444444444444444444444444444'},
+        {key: 'sections/announcement-bar.liquid', checksum: '55555555555555555555555555555555'},
+        {key: 'templates/404.json', checksum: '6666666666666666666666666666666'},
+        {key: 'templates/customers/account.json', checksum: '7777777777777777777777777777777'},
+      ])
+    })
+    test(`negating a specific file overrides ignoring one`, () => {
+      // Given
+      const options = {
+        ignoreFromFile: ['assets/*.css', '!assets/basic.css'],
+      }
+
+      // When
+      const actualChecksums = applyIgnoreFilters(checksums, options)
+
+      // Then
+      expect(actualChecksums).toEqual([
+        {key: 'assets/image.png', checksum: '22222222222222222222222222222222'},
+        {key: 'config/settings_data.json', checksum: '33333333333333333333333333333333'},
+        {key: 'config/settings_schema.json', checksum: '44444444444444444444444444444444'},
+        {key: 'sections/announcement-bar.liquid', checksum: '55555555555555555555555555555555'},
+        {key: 'templates/404.json', checksum: '6666666666666666666666666666666'},
+        {key: 'templates/customers/account.json', checksum: '7777777777777777777777777777777'},
+        {key: 'assets/basic.css', checksum: '00000000000000000000000000000000'},
+      ])
+    })
+
+    test('should not ignore files matching a negated pattern', () => {
+      const ignorePatterns = [
+        'assets/basic.css',
+        'sections/*.json',
+        'templates/*.json',
+        'templates/**/*.json',
+        'config/*.json',
+        '!config/*_schema.json',
+      ]
+
+      const actualChecksums = applyIgnoreFilters(checksums, {ignoreFromFile: ignorePatterns})
+
+      expect(actualChecksums).toEqual([
+        {key: 'assets/complex.css', checksum: '11111111111111111111111111111111'},
+        {key: 'assets/image.png', checksum: '22222222222222222222222222222222'},
+        {key: 'sections/announcement-bar.liquid', checksum: '55555555555555555555555555555555'},
+        {key: 'config/settings_schema.json', checksum: '44444444444444444444444444444444'},
+      ])
+    })
+  })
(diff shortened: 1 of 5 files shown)


packages/theme/src/cli/utilities/asset-ignore.test.ts
```diff
@@ -174,4 +174,65 @@ describe('asset-ignore', () => {
       ])
     })
   })
+
+  describe('applyIgnoreFilters with negated patterns', () => {
+    test(`only negating a file does not produce side effects of other ignored files`, () => {
+      const options = {
+        ignoreFromFile: ['!assets/basic.css'],
+      }
```
EvilGenius13, line 182: This test was written because the current behaviour made it so that just having a negated option ignored other files and folders.

packages/theme/src/cli/utilities/asset-ignore.ts
```diff
@@ -13,10 +13,23 @@ export function applyIgnoreFilters<T extends {key: string}>(
...
+
+  if (negatedPatterns.length > 0) {
+    filteredFiles = filteredFiles.concat(files.filter(filterBy(negatedPatterns, '.shopifyignore', true)))
+  }
+
+  filteredFiles = filteredFiles
     .filter(filterBy(ignoreOptions, '--ignore'))
     .filter(filterBy(onlyOptions, '--only', true))
+
+  const uniqueFiles = Array.from(new Set(filteredFiles.map((file) => file.key))).map((key) =>
```
EvilGenius13, line 29: I'm on the fence with this. This was an additional guard in the event any of the patterns or regexes somehow added a file twice. We comb through the file list and remove any non-unique values. 

In my tests and 🎩 I couldn't create a scenario where this happened but with how tricky regex matching can be, it could be a possibility.

I'm open to feedback on whether or not to keep this

packages/theme/src/cli/utilities/theme-fs.ts
```diff
@@ -133,7 +133,7 @@ export function mountThemeFileSystem(root: string, options?: ThemeFileSystemOpti
         unsyncedFileKeys.add(fileKey)
       }
 
-      return file.value || file.attachment || ''
+      return file.value ?? file.attachment ?? ''
```
EvilGenius13, line 136: ESLint

packages/theme/src/cli/utilities/theme-fs/theme-fs-mock-factory.ts
```diff
@@ -18,10 +18,11 @@ export function fakeThemeFileSystem(
       files.set(asset.key, asset)
     },
     read: async (fileKey: string) => {
-      return files.get(fileKey)?.value || files.get(fileKey)?.attachment
+      return files.get(fileKey)?.value ?? files.get(fileKey)?.attachment
```
EvilGenius13, line 21: ESLint

EvilGenius13:
> Thank you, @EvilGenius13! I believe this PR is heading in a great direction! Excellent unit tests :)
> 
> I think we could handle negated patterns consistently, also supporting them with the `--only` and `--ignore` flags.
> 
> Apart from that minor detail, everything else looks excellent. Thanks again for this PR! :)

`--only` and `--ignore` flags can now handle negated patterns as well

packages/theme/src/cli/utilities/asset-ignore.ts
jamesmengo: Now that we're doing more operations, wdyt about grouping the logic by source?
- Shopifyignore
- ignore
- only

packages/theme/src/cli/utilities/theme-fs/theme-fs-mock-factory.ts
EvilGenius13: No problem, I will revert this.

packages/theme/src/cli/utilities/asset-ignore.ts
```diff
@@ -13,10 +13,28 @@ export function applyIgnoreFilters<T extends {key: string}>(
...
+    filteredFiles = filteredFiles.concat(files.filter(filterBy(negatedIgnorePatterns, '--ignore', true)))
+  }
+  if (negatedOnlyPatterns.length > 0) {
+    filteredFiles = filteredFiles.filter(filterBy(negatedOnlyPatterns, '--only'))
+  }
+
+  const uniqueFiles = Array.from(new Set(filteredFiles.map((file) => file.key))).map((key) =>
+    filteredFiles.find((file) => file.key === key),
+  )
+  return uniqueFiles
```
jamesmengo, line 37: Can you try using [uniqBy](https://github.com/Shopify/cli/blob/e58493f44c616e2dae2847086158287f6f777bb0/packages/cli-kit/src/public/common/array.ts#L46-L57) for this grabbing unique files?

packages/theme/src/cli/utilities/asset-ignore.ts
EvilGenius13: Will implement that now.

packages/theme/src/cli/utilities/theme-fs.ts
```diff
@@ -241,7 +241,8 @@ export function mountThemeFileSystem(root: string, options?: ThemeFileSystemOpti
       await writeThemeFile(root, asset)
     },
     read,
-    applyIgnoreFilters: (files) => applyIgnoreFilters(files, filterPatterns),
+    applyIgnoreFilters: <T extends {key: string}>(files: T[]): T[] =>
+      applyIgnoreFilters(files, filterPatterns).filter((file): file is T => file !== undefined),
```
jamesmengo, line 245: `UniqBy` doesn't return undefined values so I believe we can remove this!

jamesmengo (approved):
Thanks for adding the test!
Let's clean up `applyIgnoreFilters` before shipping

PR #4738, opened 2024-10-24
Add import admin link extensions to CLI and migration on deployment

alfonso-noriega:
With the removal of the Partners Dashboard it is needed to migrate all dashboard managed extensions to be CLI managed. This PR adds support to import existing admin links (`app_link` and `bulk_action) and the call to the appModules migration endpoint on their deployment to update them to a CLI managed specification.

### How to test your changes?

- create a new app or reuse an existing one
- create a new app link extension on partners for the app
- in the CLI repository checkout this branch and run `pnpm shopify app import-extensions --path path/to/your/app`
- select admin links and check that in the extensions folder of your app there is a new extension with the toml file matching this structure and fields:
```
[[extensions]]
type = "admin_link"
name = "bulk action name 1"
handle = "bulk-action-name-1"

  [[extensions.targeting]]
  text = "bulk action label"
  url = "http://example.org"
  target = "admin.order.selection.link"

```

### Checklist

### DIFF
--- a/packages/app/src/cli/services/context/identifiers-extensions.ts
+++ b/packages/app/src/cli/services/context/identifiers-extensions.ts
@@ -16,6 +16,7 @@ import {
   PaymentModulesMap,
   UIModulesMap,
   SubscriptionModulesMap,
+  AdminLinkModulesMap,
 } from '../dev/migrate-app-module.js'
 import {ExtensionSpecification} from '../../models/extensions/specification.js'
 import {ExtensionInstance} from '../../models/extensions/extension-instance.js'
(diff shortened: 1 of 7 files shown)


packages/app/src/cli/services/context/identifiers-extensions.ts
```diff
@@ -45,6 +46,11 @@ export async function ensureExtensionsIds(
     dashboardOnlyExtensions,
     validIdentifiers,
   )
+  const adminLinkExtensionsToMigrate = getAdminLinkExtensionsToMigrate(
```
isaacroldan, line 49: I noticed that all `get**ExtensionsToMigrate` functions do mostly the same, do you want to unify that in a generic function? :) 
We can leave it for a future PR if not

PR #4758, opened 2024-10-28
Simplify code around migrating extensions

isaacroldan:
As we added more and more modules to be migrated, we kept adding duplicated logic. 

- Unify the logic to "Get modules to migrate by type", create a generic function and call it with different module type-mappings
- Remove duplicated logic and tests.

### How to test your changes?
Try to import extensions and then migrate them, it should work as always
Example:
- Create a flow action extension in your partner dashboard
- Run `shopify app import-extension` and import the extension you just created
- Run `shopify app deploy` and it should ask for confirmation to migrate that extension.

### Measuring impact

### Checklist


packages/app/src/cli/services/dev/migrate-app-module.test.ts
```diff
@@ -17,30 +16,55 @@ function getLocalExtension(attributes: Partial<LocalSource> = {}) {
...
     // Then
-    expect(toMigrate).toStrictEqual([{local: localExtension, remote: remoteExtension}])
+    expect(toMigrate).toStrictEqual([
+      {local: localExtension, remote: remoteExtension},
+      {local: localExtensionB, remote: remoteExtensionB},
+    ])
   })
 
-  test('matching my remote title and localIdentifier by truncating the title', () => {
+  test('matching my remote title by truncating the title', () => {
```
gracejychang, line 67:
```suggestion
  test('matching by truncated remote title and localIdentifier', () => {
```

packages/app/src/cli/services/dev/migrate-app-module.test.ts
```diff
@@ -49,62 +73,58 @@ describe('getPaymentsExtensionsToMigrate()', () => {
...
     expect(toMigrate).toStrictEqual([{local: localExtension, remote: remoteExtension}])
   })
 
-  test('matching my local and remote IDs', () => {
+  test('matching different title and localIdentifier, using uuid', () => {
     // Given
-    const localExtension = getLocalExtension({type: 'payments_extension', localIdentifier: 'my-action'})
-    const remoteExtension = getRemoteExtension({type: 'payments_app_credit_card', title: 'remote', uuid: '1234'})
+    const localExtension = getLocalExtension({type: 'payments_extension', localIdentifier: 'module-A'})
+    const remoteExtension = getRemoteExtension({type: 'payments_app_credit_card', title: 'random-title', uuid: '1234'})
```
gracejychang, line 88: To make it a bit more clear that we're matching remoteExtensions by the `defaultIdentifiers` that's being passed in to `getModulesToMigrate`, what do you think about changing how we're setting the `uuid` value here? 

```suggestion
    const remoteExtension = getRemoteExtension({type: 'payments_app_credit_card', title: 'random-title', uuid: defaultIdentifiers['moduleA']})
```

packages/app/src/cli/services/dev/migrate-app-module.test.ts
gracejychang, line 67: What do we think about changing the extension types to use other ones we've added here, such as flow?

packages/app/src/cli/services/dev/migrate-app-module.test.ts
```diff
@@ -17,30 +16,55 @@ function getLocalExtension(attributes: Partial<LocalSource> = {}) {
...
   } as unknown as RemoteSource
 }
+const defaultMap = {
+  payments_extension: [
+    'payments_app',
+    'payments_app_credit_card',
+    'payments_app_custom_credit_card',
+    'payments_app_custom_onsite',
+    'payments_app_redeemable',
+  ],
```
gracejychang, line 30: should we include the other extension types we've added? ie. flow and ui extensions

gracejychang (commented):
Mostly non-blocking comments :) I'm also wondering if we should add some type of documentation in our README on how to migrate extensions, as to avoid folks coming in and duplicating logic again.

packages/app/src/cli/services/context/identifiers-extensions.ts
```diff
@@ -30,21 +36,13 @@ export async function ensureExtensionsIds(
...
-  )
-  const paymentsExtensionsToMigrate = getPaymentsExtensionsToMigrate(
-    localExtensions,
-    dashboardOnlyExtensions,
-    validIdentifiers,
-  )
+  const uiExtensionsToMigrate = getModulesToMigrate(localExtensions, remoteExtensions, identifiers, UIModulesMap)
+  const flowExtensionsToMigrate = getModulesToMigrate(localExtensions, remoteExtensions, identifiers, FlowModulesMap)
+  const paymentsToMigrate = getModulesToMigrate(localExtensions, remoteExtensions, identifiers, PaymentModulesMap)
+  const marketingToMigrate = getModulesToMigrate(localExtensions, remoteExtensions, identifiers, MarketingModulesMap)
```
AniTumany, line 45: I think you need to pass `dashboardOnlyExtensions` to both of these, as the migration happens for dashboard only extensions which are not included in the `remoteExtensions`. Run into this issue myself when working on migrating subscription link extensions.

cc @isaacroldan

packages/app/src/cli/services/dev/migrate-app-module.test.ts
isaacroldan: The function we are testing here is generic, so by testing just a few extension types we cover all cases.

packages/app/src/cli/services/context/identifiers-extensions.ts
```diff
@@ -26,25 +32,17 @@ export async function ensureExtensionsIds(
...
-  )
-  const paymentsExtensionsToMigrate = getPaymentsExtensionsToMigrate(
-    localExtensions,
-    dashboardOnlyExtensions,
-    validIdentifiers,
-  )
+  const uiExtensionsToMigrate = getModulesToMigrate(localExtensions, remoteExtensions, identifiers, UIModulesMap)
+  const flowExtensionsToMigrate = getModulesToMigrate(localExtensions, dashboardExtensions, identifiers, FlowModulesMap)
+  const paymentsToMigrate = getModulesToMigrate(localExtensions, dashboardExtensions, identifiers, PaymentModulesMap)
+  const marketingToMigrate = getModulesToMigrate(localExtensions, dashboardExtensions, identifiers, MarketingModulesMap)
```
gonzaloriestra, line 45: Couldn't we simplify with something like this?
```
const migratableExtensions = [UIModulesMap, FlowModulesMap, PaymentModulesMap, MarketingModulesMap]
migratableExtensions.forEach((extension) => {
  const extensionsToMigrate = getModulesToMigrate(localExtensions, dashboardExtensions, identifiers, extension)
  if (extensionsToMigrate.length > 0) {
    const confirmedMigration = await extensionMigrationPrompt(extensionsToMigrate, false)
    if (!confirmedMigration) throw new AbortSilentError()
    const newRemoteExtensions = await migrateAppModules(extensionsToMigrate...)
    ...
  }
})
```

gonzaloriestra (commented):
I'm trying to migrate a flow extension, but I'm getting this error:

Monosnap -zsh 2024-11-04 13-33-16

It seems that the extension TOML is missing the name:
```
[[extensions]]
type = "flow_action"
handle = "flow-action"
runtime_url = "https://url.com/api/execute"
```

Any idea?

gonzaloriestra (approved):
I hadn't finished configuring the extension in the dashboard, thanks for your help. It works 👌

Issue #4634, opened 2024-10-10
[Bug]: Theme ID mismatch – Live theme ID returned instead of development theme


GLips:
### Please confirm that you have:

- [x] Searched [existing issues](.) to see if your issue is a duplicate. (If you’ve found a duplicate issue, feel free to add additional information in a comment on it.)
- [x] Reproduced the issue in [the latest CLI version](https://www.npmjs.com/package/@shopify/cli).

### In which of these areas are you experiencing a problem?

Theme

### Expected behavior

Shopify CLI should not crash, and always get connected to the theme ID specified by the environment in the `shopify.theme.toml` file.

### Actual behavior

CLI quits on start or page load and returns this error:
```
╭─ error ────────────────────────────────────────────────────────────────────────────────────╮
│                                                                                            │
│  Theme ID mismatch: expected 140504694981 but got 139501109445.                            │
│  Request ID: 3ebd1daa-ccfa-4b6e-a6fa-c1eb4d081a9c-1728575669                               │
│  URL: https://MY-STORE.myshopify.com/                                                      │
│                                                                                            │
│                                                                                            │
│  This is likely related to an issue in upstream Shopify APIs.                              │
│  Please try again in a few minutes and report this issue:                                  │
│  https://github.com/Shopify/cli/issues/new?template=bug-report.yml                         │
│                                                                                            │
╰────────────────────────────────────────────────────────────────────────────────────────────╯
```

### Verbose output

```
2024-10-10T16:07:14.373Z: Analytics event sent: {
  "command": "theme dev",
  "time_start": 1728576424479,
  "time_end": 1728576433729,
  "total_time": 9250,
  "success": true,
  "cli_version": "3.68.1",
  "ruby_version": "3.3.5",
  "node_version": "22.9.0",
  "is_employee": false,
  "uname": "darwin arm64",
  "env_ci": false,
  "env_plugin_installed_any_custom": false,
  "env_plugin_installed_shopify": "[\"@shopify/cli\"]",
  "env_shell": "zsh",
  "env_device_id": "ba91c82e48d5b2b1e0888eb9bc166e9c62cb811c",
  "env_cloud": "localhost",
  "env_package_manager": "pnpm",
  "env_is_global": true,
  "env_auth_method": "device_auth",
  "cmd_app_warning_api_key_deprecation_displayed": false,
  "cmd_all_timing_network_ms": 3553,
  "cmd_all_timing_prompts_ms": 4892,
[...]

PR #4759, opened 2024-10-28
Fix the theme ID mismatch error, where the live theme ID is returned instead of the development theme ID

karreiro:

Fixes https://github.com/shopify/cli/issues/4634

When the "Sign in with Shop" is activated, the `/account` endpoint behaves differently (returning the live theme ID in the body) impacting the proxy. However, following the redirect maintains the expected theme ID in the session.

It's important to note that stores without that feature were not being affected by this, and are not affected by this change as well.

This PR updates the proxy logic to proxy `/account` requests and avoid the remote-render of the `/account` endpoint.

This PR also refactors the logic in the proxy to rely on regular expressions to evaluate patterns to centralize the logic for each case on the regular expressions and prevent conditions from growing like `p.startsWith(..) && p.endsWith(..)`.

### How to test your changes?

- Run `shopify theme dev` in a store with "Sign in with Shop" is activated
- Open http://127.0.0.1:9292/account in the browser
- Notice the mismatch error no longer happens

**Before**
before

**After**
after

### Post-release steps

N/A

### Measuring impact

### Checklist

### DIFF
--- a/packages/theme/src/cli/utilities/theme-environment/proxy.test.ts
+++ b/packages/theme/src/cli/utilities/theme-environment/proxy.test.ts
@@ -227,47 +227,57 @@ describe('dev proxy', () => {
   describe('canProxyRequest', () => {
     test('should proxy non-GET requests', () => {
       const event = createH3Event('POST', '/some-path.html')
-      expect(canProxyRequest(event)).toBe(true)
+      expect(canProxyRequest(event)).toBeTruthy()
     })
 
     test('should proxy Cart requests as they are not supported by the SFR client', () => {
       const event = createH3Event('GET', '/cart/some-path')
-      expect(canProxyRequest(event)).toBe(true)
+      expect(canProxyRequest(event)).toBeTruthy()
     })
 
     test('should proxy CDN requests', () => {
       const event = createH3Event('GET', '/cdn/some-path')
-      expect(canProxyRequest(event)).toBe(true)
+      expect(canProxyRequest(event)).toBeTruthy()
     })
 
     test('should proxy CDN requests for extensions', () => {
       const event = createH3Event('GET', '/ext/cdn/some-path')
-      expect(canProxyRequest(event)).toBe(true)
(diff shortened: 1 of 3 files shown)


isaacroldan (approved):

packages/theme/src/cli/utilities/theme-environment/proxy.test.ts
```diff
@@ -227,47 +227,57 @@ describe('dev proxy', () => {
...
+      expect(canProxyRequest(event)).toBeFalsy()
     })
 
     test('should not proxy HTML requests (based on the "accept" header), allowing them to be rendered by the SFR client', () => {
       const event = createH3Event('GET', '/some-path', {accept: 'text/html'})
-      expect(canProxyRequest(event)).toBe(false)
+      expect(canProxyRequest(event)).toBeFalsy()
+    })
+
+    test('should proxy the /account requests as it may result on 302 (for users with "Sign in with Shop"), and rendering via the the SFR API would result in a broken 200', () => {
```
jamesmengo, line 273:
```suggestion
    // May result in a 302 (for users with "Sign in with Shop"), and rendering via the the SFR API would result in a broken 200
    test('should proxy the /account requests as it may result', () =>
```

jamesmengo (approved):
left 1 suggestion for test name brevity. non blocking

jamesmengo:
This fixes the issue with the server crashing

Aside - clicking `sign in with shop` opens the modal but doesn't populate any content there. Is this expected?
image

karreiro:
Thank you for the review, everyone!

> Aside - clicking sign in with shop opens the modal but doesn't populate any content there. Is this expected?

I was thinking the _same_ thing, @jamesmengo. Currently, that's expected — that's a different limitation in the proxy, as `pay.shopify.com` won't accept requests from `http://127.0.0.1:9292`. I think it's reasonable to revisit this in the future, but at the moment, I believe there are workflows in the main path of the CLI that need more attention.

PR #4768, opened 2024-10-29
Fix `app init` with custom templates.

isaacroldan:
`app init` automatically reads the scopes from the template toml to use them to create the app. This doesn't work when the template has a toml with `access_scopes`

- Use the local specifications to load the app used to get the `creationDefaultOptions`

### Measuring impact

### Checklist

### DIFF
--- a/packages/app/src/cli/services/init/init.ts
+++ b/packages/app/src/cli/services/init/init.ts
@@ -199,8 +199,7 @@ async function init(options: InitOptions) {
   let app: OrganizationApp
   if (options.selectedAppOrNameResult.result === 'new') {
     // Load the local app to get the creation options.
-    // NOTE: We need the specs just to load the scopes, if we ever remove the local scopes, we need to find a way to
-    // do this without them.
+    // NOTE: need the specs to load the scopes, if we ever remove the local specs, we need to find a way to do this without them.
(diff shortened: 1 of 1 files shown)


packages/app/src/cli/services/init/init.ts
```diff
@@ -197,8 +198,10 @@ async function init(options: InitOptions) {
 
   let app: OrganizationApp
   if (options.selectedAppOrNameResult.result === 'new') {
-    // Load the local app to get the creation options. No need for specs since we only care about Creation Options.
-    const localApp = await loadApp({specifications: [], directory: outputDirectory, userProvidedConfigName: undefined})
+    // Load the local app to get the creation options.
+    // NOTE: need the specs to load the scopes, if we ever remove the local specs, we need to find a way to do this without them.
+    const specifications = await loadLocalExtensionsSpecifications()
```
gracejychang, line 203: does this load the entire existing TOML of the app?  Is there any chance this could break things when we use `app init` just by itself?

gracejychang (approved):
Code looks good, I just have a non-blocking question 😄  I don't think tests are needed as I assume `loadLocalExtensionsSpecifications` and `loadApp` already have tests

PR #4781, opened 2024-10-31
Fix application_url being outdated in app manifest

isaacroldan:
In Dev Sessions, the first update contains an outdated application_url manifest. This is because we update the toml, but weren't updating the in-memory app configuration.

Fixes https://github.com/Shopify/develop-app-outer-loop/issues/1221

- Reload the app before starting a Dev Session, to be sure we have the latest config in all modules.

### How to test your changes?
- Start a Dev Session (dev with management API)
- If you look into your app folder in `./shopify/bundle/manifest.json`, the manifest should contain the current tunnel URL.

### Measuring impact

### Checklist

### DIFF
--- a/packages/app/src/cli/services/dev/processes/dev-session.ts
+++ b/packages/app/src/cli/services/dev/processes/dev-session.ts
@@ -1,9 +1,9 @@
 /* eslint-disable no-case-declarations */
 import {BaseProcess, DevProcessFunction} from './types.js'
 import {DeveloperPlatformClient} from '../../../utilities/developer-platform-client.js'
-import {AppInterface} from '../../../models/app/app.js'
+import {AppLinkedInterface} from '../../../models/app/app.js'
 import {getExtensionUploadURL} from '../../deploy/upload.js'
-import {AppEventWatcher, EventType} from '../app-events/app-event-watcher.js'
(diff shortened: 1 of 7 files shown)


packages/app/src/cli/services/dev/processes/dev-session.ts
```diff
@@ -60,7 +61,10 @@ export const pushUpdatesForDevSession: DevProcessFunction<DevSessionOptions> = a
   {stderr, stdout, abortSignal: signal},
   options,
 ) => {
-  const {developerPlatformClient, app} = options
+  const {developerPlatformClient} = options
+
+  // Reload the app before starting the dev session, at this point the configuration has changed (e.g. application_url)
+  const app = await reloadApp(options.app, {stderr, stdout, signal})
```
pt2pham, line 67: I'm assuming this is reloading the app state since the TOML changed? I wonder how this plays out if we try to store the tunnel URL elsewhere instead of updating all the URLs 🤔 may be a post EA thing though

pt2pham (approved):

packages/app/src/cli/services/dev/processes/dev-session.ts
isaacroldan: yeah, I didn't want to do anything weird until we decide what to do with updating the Toml or not. This was the safest solution for now.

Issue #4618, opened 2024-10-09
[Bug]: Shopify CLI crashes intermittently


rcasimmons:
### Please confirm that you have:

- [x] Searched [existing issues](.) to see if your issue is a duplicate. (If you’ve found a duplicate issue, feel free to add additional information in a comment on it.)
- [x] Reproduced the issue in [the latest CLI version](https://www.npmjs.com/package/@shopify/cli).

### In which of these areas are you experiencing a problem?

Theme

### Expected behavior

Running the CLI's Dev command should give me a running server that indefinitely runs until I close this myself.

### Actual behavior

At random intervals, the CLI will quit running the 'Dev' command and display the following error:

[401] API request unauthorized error:
![Image](https://github.com/user-attachments/assets/890185d3-1d12-4b8b-9ae9-03737f69ee0d)

The error appears to be consistent every time the CLI quits the dev command although the interval at which it happens appears to be random with no correlation between files being saved/synced.

I don't current have a --verbose log at hand but will update the issue when this is at hand.

### Verbose output

TBC

### Reproduction steps

1. Run the normal `shopify theme dev` command

### Operating System

macOS 14.6.1

### Shopify CLI version (check your project's `package.json` if you're not sure)

3.68.0

### Shell

Zsh

### Node version (run `node -v` if you're not sure)

v22

### What language and version are you using in your application?

N/A

mgmanzella:
👋 thanks for reporting this issue, to confirm when did you start experiencing this? was it on this new version of the CLI? also, does this still happen when you use the `--legacy` flag? this information will help us narrow down the root of the problem. please send along the verbose logs when you have them as well. thanks again 🙏

rcasimmons:
> looks like the team is aware of this and has some in flight fixes, in the meantime let us know if using `--legacy` is an effective workaround
> 
> [#4599](https://github.com/Shopify/cli/pull/4599) [#4606](https://github.com/Shopify/cli/pull/4606)

Hey!

This hasn't happened with the --legacy flag active.

I updated to 3.68.1 with a supposed fix for 401 but sadly this hasn't worked.

I'm still unable to get you a --verbose log just because there will be times I launch the CLI as is.

I will run the CLI using --verbose today and report back :)

frandiox:
@karreiro Am I reading this correctly?
1. The Introspection request failed with: status: 401
2. The current session is valid but needs refresh. Refreshing...
[...]

Issue #4769, opened 2024-10-29
Move `cli-kit` themes logic into the themes package


jamesmengo:
- This includes moving the existing code into `packages/theme`, and ensuring that the utilities and functions needed in the `app` package are served.

- Part of this will also require investigating `cli-kit` dependancies to ensure that we aren't breaking any existing flows

PR #4790, opened 2024-11-04
Refresh session when 401 errors happens

karreiro:

Fixes https://github.com/Shopify/cli/issues/4618

The Assets API may return 401 statuses after running for a while, and that may happen before the session gets refreshed. To prevent the development server from terminating, if we get that status, we manually refresh the session.

This PR solves this by calling `session.refresh()` when it gets a 401 error. We'd face a circular dependency if we were to get the types correctly, so this PR checks the `refresh` at runtime and just calls it (this will no longer be necessary with https://github.com/Shopify/cli/issues/4769 solved).

### How to test your changes?

- Update the `SESSION_TIMEOUT_IN_MS` value (in `dev-server-session.ts`) to `300 * 60 * 1000`
- Run `shopify theme dev --verbose`
- Leave a bash script appending a value in the sections/announcement-bar.liquid every 5 seconds
- Notice that the session will get refreshed after a while (max wait time: ~2h)
- Here's [my results](https://gist.github.com/karreiro/21cf8af828b62c3c7cee9f23d243b125) (in my test, I added some extra `console.log` logs to make things a bit clearer)

### Post-release steps

N/A

### Measuring impact

### Checklist

### DIFF
--- a/.changeset/dull-shoes-clean.md
+++ b/.changeset/dull-shoes-clean.md
@@ -2,5 +2,5 @@
 '@shopify/theme': patch
 ---
 
-Prevent the `shopify theme dev` from terminating by refreshing the session
+Prevent the `shopify theme dev` command from terminating by refreshing the session
 


EvilGenius13 (approved):

karreiro:
@Shopify/app-inner-loop Could you please take a look at this one to unblock merge? 🙇

karreiro:
(I just rebased the branch because it got a conflict — it's ready to merge again)

gonzaloriestra (approved):

PR #4799, opened 2024-11-05
WIP - port over theme files delete

catlee:

Fixes #0000 

### Measuring impact

### Checklist


graygilmore (approved):
no 🎩 yet but code part looks good, one small question about the progress bar

packages/theme/src/cli/utilities/theme-uploader.ts
```diff
@@ -131,23 +131,27 @@ function buildDeleteJob(
...
+    theme.id,
+    orderedFiles.map((file) => file.key),
+    session,
+  )
+    .then((results) => {
+      results.forEach((success, key) => {
+        if (!success) {
+          createSyncingCatchError(key, 'delete')(new Error(`Failed to delete ${key}`))
+        }
+        progress.current++
```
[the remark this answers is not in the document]
catlee: yeah, I wasn't sure about that either....

In the case where we have multiple batches, progress tracking could be useful....but the batching is being handled by the underlying function right now.

packages/cli-kit/src/public/node/themes/api.ts
graygilmore, line 1: > While 🎩 I found we run into a nil error [here](https://github.com/Shopify/shopify/blob/c75da8831e81d8714851dc5ca362157c2ed0612d/areas/core/shopify/components/online_store/app/models/graph_api/admin/mutations/online_store/theme_files_delete.rb#L207) (see observe [logs](https://observe.shopify.io/a/observe/investigate/logs?r=%7B%22from%22%3A%22now-1h%22%2C%22to%22%3A%22now%22%7D&q=%7B%22refId%22%3A%220%22%2C%22viewMode%22%3A%22builder%22%2C%22queryStep%22%3A%22%22%2C%22queryMode%22%3A%22search%22%2C%22rawDataMode%22%3Afalse%2C%22request%22%3A%7B%22datasets%22%3A%5B%22core-next%22%5D%2C%22filter_combination%22%3A%22AND%22%2C%22filters%22%3A%5B%5D%2C%22filter_group%22%3A%7B%22conjunction%22%3A%22AND%22%2C%22filter_groups%22%3A%5B%7B%22conjunction%22%3A%22AND%22%2C%22filters%22%3A%5B%7B%22column%22%3A%22request_id%22%2C%22op%22%3A%22%3D%22%2C%22value%22%3A%225574d8fb-f4ba-4eac-a66b-dd98756d5ebb-1730921691%22%7D%5D%7D%5D%2C%22filters%22%3A%5B%5D%7D%7D%2C%22templateVariables%22%3A%5B%5D%7D&ec=timestamp%7Ename%7Eapplication%7Econtroller%7Emessage%7Emsg%7Eentrypoint%7Eshop_id%7Etimestamp%7Eexception.message)). I guess the CLI (or at least in dev) doesn't have an associated user 🤔

Are we ready to look and 🎩 this PR again or are we waiting on this to be resolved?

amcaplan (commented):
The previous implementation used an API throttler, while this new implementation doesn't seem to have any kind of throttling enabled. Is that a cause for concern?

catlee:
> The previous implementation used an API throttler, while this new implementation doesn't seem to have any kind of throttling enabled. Is that a cause for concern?

Good point, I'm not sure if our graphQL implementation respects throttling. 

https://shopify.dev/docs/api/admin-graphql#status_and_error_codes says we should look for `extensions.code == THROTTLING`.

I'll file an issue for this.

catlee:
> I'll file an issue for this.

https://github.com/Shopify/cli/issues/5222

I think fixing throttling is independent from supporting deleting files with graphQL. Can we ship this PR?

packages/theme/src/cli/utilities/theme-uploader.ts
```diff
@@ -139,21 +140,32 @@ function buildDeleteJob(
...
+      batch.map((file) => file.key),
+      session,
+    ).then((results) => {
+      results.forEach((result) => {
+        uploadResults.set(result.key, result)
+        if (!result.success) {
+          const errorMessage = result.errors?.asset?.map((err) => `-${err}`).join('\n')
+          createSyncingCatchError(result.key, 'delete')(new Error(`Failed to delete ${result.key}: ${errorMessage}`))
+        }
+        progress.current += batch.length
```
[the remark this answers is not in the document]
catlee: Good catch! I think the problem was that  `progress.current += batch.length` was inside the `results.forEach` loop.

I've moved it outside (but still within the promise's .then() handler).

packages/cli-kit/src/public/node/themes/api.ts
```diff
@@ -147,11 +148,47 @@ export async function fetchThemeAssets(id: number, filenames: Key[], session: Ad
...
+    }
+
+    if (userErrors.length > 0) {
+      userErrors.forEach((error) => {
+        if (error.filename) {
+          results.push({
+            key: error.filename,
+            success: false,
+            operation: Operation.Delete,
+            errors: {asset: [error.message]},
```
lucyxiang, line 182: Any reason we need to nest errors under `asset? Or it's just the existing structure

catlee: Yeah, this is expected by the existing type structure: https://github.com/Shopify/cli/blob/main/packages/cli-kit/src/public/node/themes/types.ts#L213

catlee:
> One UI/ error msg issue but functional 🎩 lgtm

I think the old error message is misleading / wrong. We are authenticated properly, but not allowed to remove the given files.

The old error is triggered via `handleForbiddenError` here https://github.com/Shopify/cli/blob/7b3ba7720ecb0db84834442f99d0dcb7dfe6390e/packages/cli-kit/src/public/node/themes/api.ts#L397

I'm not sure why it's getting overwritten by the prompt now though...

packages/theme/src/cli/utilities/theme-uploader.ts
```diff
@@ -139,21 +140,32 @@ function buildDeleteJob(
...
-          progress.current++
-        }),
-    ),
-  ).then(() => {
+  if (orderedFiles.length === 0) {
+    return {progress, promise: Promise.resolve()}
+  }
+
+  const deleteBatches = []
+  for (let i = 0; i < orderedFiles.length; i += MAX_BATCH_FILE_COUNT) {
```
amcaplan, line 149: I think it's a mistake to create promises for all files immediately, so long as throttling doesn't exist. I think you're likely to end up with rate limiting errors, timeouts, etc. when making sweeping changes to large themes (or really anything with an automated build process resulting in lots of updates).

I also think it's a mistake to keep this a synchronous function. `async`/`await` exists for a reason, and this code is getting quite complex with nothing to catch failure, so it's possible things could just never resolve (e.g. the HTTP request totally fails and throws an error) rather than throwing an error that gets displayed to the user.

I won't hold up review as this isn't my domain of expertise, but I did want to make sure you're aware of these concerns.

PR #4801, opened 2024-11-06
Fix dev session refresh token

isaacroldan:
Dev session wasn't processing 401 errors, which means it wasn't refreshing tokens and the session crashed once the token expires.

Fixes https://github.com/Shopify/develop-app-inner-loop/issues/2295

- If a 401 error is received, instead of printing it to the terminal, re-throw it so that the retry system can catch it and refresh the token.
- Update `uncaughtException` to render fatal errors. Some errors thrown inside some contexts aren't caught by the main error handler and are instead processed by the `uncaughtException` handler. It now knows how to render fatal errors.
- Some small UX improvements for normal errors in `dev-session`.

### How to test your changes?
- Start `dev`
- Force a logout on your session (ping me on how to do this)
- Wait 1 min
- Make a change in your app -> see the error. In this case, since it was a forced logout, the CLI can recover from it and the dev-session will end.
- To test that the refresh works: leave a dev session open for more than 1h

### Measuring impact

### Checklist

### DIFF
--- a/packages/cli-kit/src/public/common/retry.ts
+++ b/packages/cli-kit/src/public/common/retry.ts
@@ -1,6 +1,3 @@
-import {FatalError} from '../node/error.js'
-import {renderFatalError} from '../node/ui.js'
-
 /**
  * Perform an action optimistically. If it fails the first time, first initiate
  * a provided recovery procedure, then retry the action. If it fails again,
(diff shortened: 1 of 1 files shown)


craigmichaelmartin (approved):
Looks great!

PR #4811, opened 2024-11-07
Report GraphQL system errors as expected

gordonhirsch:

Improve accuracy of the CLI error count SLOs by marking GraphQL failures due to internal server errors as `expected`.

Internal server errors are recorded as unexpected events by the back-end that generates them. From the point of view of the  CLI, we expect such errors will occur and we communicate the failure to the user. That's all we can do, so they should not count against the error count SLOs

Uses a common type to identify all GraphQL. Interrogates the error response info to determine if the error is internal, and reports the error as expected if it is internal. 

### Measuring impact

### Checklist

### DIFF
--- a/packages/cli-kit/src/private/node/api/graphql.ts
+++ b/packages/cli-kit/src/private/node/api/graphql.ts
@@ -1,7 +1,6 @@
 import {GraphQLClientError, sanitizedHeadersOutput} from './headers.js'
 import {sanitizeURL} from './urls.js'
 import {stringifyMessage, outputContent, outputToken, outputDebug} from '../../../public/node/output.js'
-import {AbortError} from '../../../public/node/error.js'
 import {ClientError, Variables} from 'graphql-request'
 
 export function debugLogRequestInfo(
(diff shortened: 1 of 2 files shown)


packages/cli-kit/src/private/node/api/graphql.ts
```diff
@@ -43,12 +42,7 @@ ${outputToken.json(error.response.errors)}
 Request ID: ${requestId}
 `
       }
-      let mappedError: Error
-      if (status < 500) {
-        mappedError = new GraphQLClientError(errorMessage, status, error.response.errors)
-      } else {
-        mappedError = new AbortError(errorMessage)
-      }
+      const mappedError: Error = new GraphQLClientError(errorMessage, status, error.response.errors)
```
gordonhirsch, line 45: Not strictly necessary, but using the same type here allows us to retain information that is otherwise lost. This includes the specific error messages, the status code, and stack information.

Not sure of the original intent of "Client" here. Is it meant to convey a `4xx` class error, or an error received by the GraphQL client? If the former, then this new usage would require a rename, at least as currently written,.

PR #4829, opened 2024-11-11
Update Javy and function-runner versions

jeffcharles:

The function-runner version has an Javy QuickJS provider that matches what's being used in production. I'm already in here so may as well update Javy as well. This should also slightly de-risk future updates to Javy because there will be fewer changes (compared to updating across multiple versions at the same time).

Updates the versions the CLI will download to the latest versions of Javy and the function-runner.

### How to test your changes?

- `rm -r packages/app/dist/cli/services/bin/` (clear out any cached downloaded binaries)
- `pnpm shopify app function build --path <path_to_function_extension>` (downloads and runs Javy)
- `echo "{}" | pnpm shopify app function run --path <path_to_function_extension>` (downloads and runs function-runner)

### Measuring impact

### Checklist

### DIFF
--- a/.changeset/eight-dragons-grab.md
+++ b/.changeset/eight-dragons-grab.md
@@ -1,0 +1,5 @@
+---
+'@shopify/app': patch
+---
+
+Update Javy to 3.2.0 and function-runner to 6.3.0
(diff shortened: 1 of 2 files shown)


shauns (approved):
Approving -- its straightforward enough -- though am assuming there's no change to e.g. cross-OS support for these.

jeffcharles:
Yeah there's no change to OS support for either

Issue #4713, opened 2024-10-22
[Feature]: Shopify Theme Dev Shortcuts


ceri-waterscreative:
### What area(s) will this request affect?

Theme, Running your code locally

### What type of change do you want to see?

New feature

### Overview

You know how `shopify app dev` has shortcut commands like `p` for 'preview' in your browser'?

Could we get these for `shopify theme dev` too?

I'm thinking we need shortcuts for...
Launch localhost in browser (d?)
Launch preview in browser (p?)
Launch editor in browser (e?)

### Motivation

The number of times i accidently close my preview or later need the editor and have to scroll back (past the many many logs) to find it is saddening..

nbsimmons:
I agree with this, especially when using `--verbose` - only so much history is retained.

Trying to get the customizer link can be frustrating.

PR #4830, opened 2024-11-11
[Feature] Add shortcut keys to theme dev

EvilGenius13:
Do you suffer from long logs? Do you accidentally close your browser window and have to scroll all the way up in your terminal to find the theme dev links? Look no further than the addition of shortcut keys!

Closes https://github.com/Shopify/cli/issues/4713
Closes https://github.com/Shopify/develop-advanced-edits/issues/427

With the amount of logs `theme dev` generates, it can be difficult to find the original store links such as previewing your theme, or opening the theme editor. Having shortcut keys is a simple way to open the links without needing to physically click on them in the terminal.

Add shortcut keys to the `theme dev` command
We look for keypresses, specifically:
`t` - opens the theme locally
`e` - opens the theme editor
`p` - opens the share preview
`g` - opens the gift cards preview
we also look for `ctrl` + `c` to exit the process (current behaviour)

Screenshot 2024-11-21 at 4 51 40 PM

### How to test your changes?

Pull down the branch
Build the branch
Run the `theme dev` command
Try out the shortcut keys. Click some links, save some files, and see that it does not affect anything else running, `ctrl+c` out to confirm you can exit the same as before.

### Measuring impact

### Checklist

### DIFF
--- a/packages/theme/src/cli/services/dev.ts
+++ b/packages/theme/src/cli/services/dev.ts
@@ -12,6 +12,8 @@ import {Theme} from '@shopify/cli-kit/node/themes/types'
 import {checkPortAvailability, getAvailableTCPPort} from '@shopify/cli-kit/node/tcp'
 import {AbortError} from '@shopify/cli-kit/node/error'
 import {openURL} from '@shopify/cli-kit/node/system'
+import chalk from '@shopify/cli-kit/node/colors'
+import readline from 'readline'
 
 const DEFAULT_HOST = '127.0.0.1'
 const DEFAULT_PORT = '9292'
@@ -61,6 +63,13 @@ export async function dev(options: DevOptions) {
 
   const port = options.port ?? String(await getAvailableTCPPort(Number(DEFAULT_PORT)))
 
+  const urls = {
+    local: `http://${host}:${port}`,
+    giftCard: `http://${host}:${port}/gift_cards/[store_id]/preview`,
+    themeEditor: `https://${options.store}/admin/themes/${options.theme.id}/editor`,
+    preview: `https://${options.store}/?preview_theme_id=${options.theme.id}`,
+  }
+
   const storefrontPassword = await storefrontPasswordPromise
   const session = await initializeDevServerSession(
     options.theme.id.toString(),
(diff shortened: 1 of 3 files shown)


packages/theme/src/cli/services/dev.ts
```diff
@@ -104,6 +106,33 @@ export async function dev(options: DevOptions) {
...
+        renderWarning({headline: 'Failed to open theme editor.', body: error.stack ?? error.message})
+      })
+    } else if (key.name === 'p') {
+      openURL(`https://${options.store}/?preview_theme_id=${options.theme.id}`).catch((error) => {
+        renderWarning({headline: 'Failed to open theme preview.', body: error.stack ?? error.message})
+      })
+    } else if (key.name === 'g') {
+      openURL(`http://${host}:${port}/gift_cards/${options.store}/preview`).catch((error) => {
+        renderWarning({headline: 'Failed to open gift card preview.', body: error.stack ?? error.message})
+      })
```
lukeh-shopify, line 133: These URLs are referenced elsewhere in this file for the links, correct? Could we extract them to variables or methods so there would only be one place to update them?

Same with all of the warning logs, they each have to do the same thing except for the actual `headline`. This could probably be a private method as well. I see we have at least one other `renderWarning` log on line 104 this file that could use the same private method as well

packages/theme/src/cli/services/dev.ts
```diff
@@ -104,6 +106,33 @@ export async function dev(options: DevOptions) {
...
+
+  readline.emitKeypressEvents(process.stdin)
+  if (process.stdin.isTTY) {
+    process.stdin.setRawMode(true)
+  }
+
+  process.stdin.on('keypress', (_str, key) => {
+    if (key.ctrl && key.name === 'c') {
+      process.exit()
+    } else if (key.name === 't') {
```
lukeh-shopify, line 118: Would it be cleaner if the rest of these else if conditions were changed to a switch statement? I don't really have a preference

packages/theme/src/cli/services/dev.ts
```diff
@@ -98,47 +107,93 @@ export async function dev(options: DevOptions) {
...
-  const remoteUrl = `https://${store}`
-  const localUrl = `http://${host}:${port}`
+export function renderLinks(urls: {local: string; giftCard: string; themeEditor: string; preview: string}) {
   renderSuccess({
     body: [
       {
         list: {
-          title: {bold: 'Preview your theme'},
-          items: [localUrl],
+          title: chalk.bold('Preview your theme ') + chalk.cyan('(t)'),
```
karreiro, line 160: I believe we could follow the CLI-kit standards to present the shortcuts. Thus, this line could be presented as:

```ts
title: [{bold: 'Preview your theme'}, {subdued: '(t)'}],
```

or:
```ts
title: [{bold: 'Preview your theme'}, {command: '(t)'}],
```

I believe that `subdued` fits a bit more in this context. While it's a bit tempting to be opinionated about the color, it's nice to use CLI-kit standard colors, so we may make the CLI more cohesive across commands :)

karreiro (approved):
Thank you, @EvilGenius13 🔥 🚀 Amazing stuff! I've left only one minor comment :)

packages/theme/src/cli/services/dev.ts
EvilGenius13: After doing some digging, this won't work for labels as they expect a string. We will have to make some changes in cli-kit. I have made a follow up issue [here](https://github.com/Shopify/develop-advanced-edits/issues/438) to make that fix.

PR #4831, opened 2024-11-11
Download and use Javy plugin when building JS functions

jeffcharles:

Javy is being updated to require a Javy plugin, which is a WebAssembly module exposing a specific API, when building dynamically linked modules. Not specifying a plugin when using `javy build -C dynamic` will cause Javy to exit with an error saying a plugin is required. So, the Shopify CLI, will need to be updated to download and use a Javy plugin. This change will use the default Javy plugin.

This change **WILL NOT WORK** until a new version of Javy is released that supports the `-C plugin=...` argument for dynamically linked modules. I would like to get an advance review of this change to ensure there isn't a long wait for when the Shopify CLI will be able to use future versions of Javy before I publish a new release of Javy that will include this breaking change.

1. Updates the CLI to download the default Javy plugin. This includes making `DownloadableBinary` an interface and creating two implementation classes. One is the existing `DownloadableBinary` renamed to `Executable` and the other `JavyPlugin`. This is because the logic for determining the path and download URL are different. Also renamed `installBinary` to `downloadBinary`.
2. Updates the call to `javy build` to use the plugin that was downloaded by specifying an additional argument of `-C plugin=<path_to_plugin>`.

### How to test your changes?

`pnpm shopify app function build --path <path_to_js_function_extension>`. It'll complain about Javy returning an exit code of 1 and the underlying message from Javy will be that plugins are not currently supported for dynamically linked modules.

### Measuring impact

### Checklist


packages/app/src/cli/services/function/binaries.ts
```diff
@@ -71,29 +81,56 @@ class DownloadableBinary {
...
+  readonly path: string
+
+  constructor() {
+    this.name = 'javy_quickjs_provider_v3'
+    this.version = JAVY_PLUGIN_VERSION
+    this.path = joinPath(dirname(fileURLToPath(import.meta.url)), '..', 'bin', 'javy_quickjs_provider_v3.wasm')
+  }
+
+  downloadUrl(_processPlatform: string, _processArch: string) {
+    return `https://github.com/bytecodealliance/javy/releases/download/${this.version}/plugin.wasm.gz`
```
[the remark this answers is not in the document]
jeffcharles: It does not, it's a Wasm module so it's OS and processor independent.

jeffcharles:
@lopert can you comment out the `stderr: 'inherit'` on line 192 in `build.ts` in the `exec` command under `runJavy` and try running it again and see what gets emitted?

Also I'm not sure why the contents of the stderr stream are not being displayed when it's set to `inherit` or why it was set to `inherit`. But that's separate from this PR.

jeffcharles:
@lopert it looks like the cause was I had a more recent version of Javy cached in my local dev env for the CLI package so exit code and error message emitted to stderr was different. I'll have to rebase this PR on top of the changes in main anyway so that'll cause the slightly newer Javy to be used if you run `pnpm clean` or manually clean the binary cache.

jeffcharles:
@lopert the rebased PR has an exit code of 1. You will want to run `pnpm clean` before tophatting (I've updated the tophatting instructions).

lopert (approved):
Got the right `exit code 1` this 🎩

lopert (approved):
🎩 'd with the latest again and build successfully

packages/app/src/cli/services/function/binaries.ts
```diff
@@ -71,29 +81,56 @@ class DownloadableBinary {
...
   if (!_functionRunner) {
-    _functionRunner = new DownloadableBinary('function-runner', FUNCTION_RUNNER_VERSION, 'Shopify/function-runner')
+    _functionRunner = new Executable('function-runner', FUNCTION_RUNNER_VERSION, 'Shopify/function-runner')
   }
   return _functionRunner
 }
 
-export async function installBinary(bin: DownloadableBinary) {
+export async function downloadBinary(bin: DownloadableBinary) {
   const isInstalled = await fileExists(bin.path)
```
shauns, line 134: nit: `isDownloaded` ?

packages/app/src/cli/services/function/binaries.ts
```diff
@@ -10,12 +10,22 @@ import * as gzip from 'node:zlib'
...
+interface DownloadableBinary {
+  path: string
+  name: string
+  version: string
+
+  downloadUrl(processPlatform: string, processArch: string): string
+  processResponse(responseStream: PipelineSource<unknown>, outputStream: fs.WriteStream): Promise<void>
+}
 
 // The logic for determining the download URL and what to do with the response stream is _coincidentally_ the same for
```
shauns, line 25: I think this comment is out of date -- they _have_ diverged?

shauns (approved):
Approving from app-inner-loop. I left some small naming/comment nits. I think the prior reviewers were best placed to review the purpose and the main thrust behind this change as its so specific -- I've looked at this only really from CLI stewarding.

packages/app/src/cli/services/function/binaries.ts
jeffcharles: So the logic for determining the download URL and processing the response stream is the still the same for the Javy CLI and function-runner binaries. It's different for the Javy plugin, but this comment doesn't refer to the Javy plugin. 

This comment is on top of and only applies to the `Executable` class which the Javy CLI and function-runner use. The Javy plugin is downloaded with the `JavyPlugin` class, not the `Executable` class. The `Executable` and `JavyPlugin` classes both implement the `DownloadableBinary` interface but the comment is independent of that interface.

PR #4846, opened 2024-11-12
Enable multiple asset handling

elanalynn:

Part of [conditional rendering project](https://docs.google.com/document/d/1X1NhQMxz6GnTU49rDplMd1yVmX3rsj2obIfHUFQ7WkM/edit?tab=t.0).

This PR enables the processing of a `should_render` asset, as well as other assets.

- Maps `should_render` field from TOML to extension config
- Bundles all `should_render` modules into one file (for both `dev` and `deploy` actions)
- Sends `buildManifest` to `shopify` in order centralize file path information in one place

### How to test your changes?

[Testing instructions](https://gist.github.com/elanalynn/bff3bda620cb5d0284d0cf892e713e90)

### Measuring impact

### Checklist

### DIFF
--- a/packages/app/src/cli/models/extensions/specification.ts
+++ b/packages/app/src/cli/models/extensions/specification.ts
@@ -35,6 +35,16 @@ export interface CustomTransformationConfig {
 type ExtensionExperience = 'extension' | 'configuration'
 type UidStrategy = 'single' | 'dynamic' | 'uuid'
 
+export enum AssetIdentifier {
+  ShouldRender = 'should_render',
+  Main = 'main',
+}
+
+export interface Asset {
+  identifier: AssetIdentifier
+  outputFileName: string
+  content: string
+}
 /**
  * Extension specification with all the needed properties and methods to load an extension.
  */
@@ -50,7 +60,7 @@ export interface ExtensionSpecification<TConfiguration extends BaseConfigType = 
   experience: ExtensionExperience
   dependency?: string
   graphQLType?: string
-  getBundleExtensionStdinContent?: (config: TConfiguration) => string
+  getBundleExtensionStdinContent?: (config: TConfiguration) => {main: string; assets?: Asset[]}
   deployConfig?: (
     config: TConfiguration,
     directory: string,
--- a/packages/app/src/cli/models/extensions/specifications/ui_extension.ts
+++ b/packages/app/src/cli/models/extensions/specifications/ui_extension.ts
@@ -1,4 +1,4 @@
-import {ExtensionFeature, createExtensionSpecification} from '../specification.js'
+import {Asset, AssetIdentifier, ExtensionFeature, createExtensionSpecification} from '../specification.js'
 import {NewExtensionPointSchemaType, NewExtensionPointsSchema, BaseSchema} from '../schemas.js'
 import {loadLocalesConfig} from '../../../utilities/extensions/locales-configuration.js'
 import {getExtensionPointTargetSurface} from '../../../services/dev/extension/utilities.js'
@@ -14,6 +14,21 @@ const validatePoints = (config: {extension_points?: unknown[]; targeting?: unkno
   return config.extension_points !== undefined || config.targeting !== undefined
 }
 
+export interface BuildManifest {
+  assets: {
+    // Main asset is always required
+    [AssetIdentifier.Main]: {
+      filepath: string
+      module?: string
+    }
+  } & {
+    [key in AssetIdentifier]?: {
+      filepath: string
+      module?: string
+    }
+  }
+}
+
 const missingExtensionPointsMessage = 'No extension targets defined, add a `targeting` field to your configuration'
 
 export type UIExtensionSchemaType = zod.infer<typeof UIExtensionSchema>
@@ -25,6 +40,23 @@ export const UIExtensionSchema = BaseSchema.extend({
   .refine((config) => validatePoints(config), missingExtensionPointsMessage)
   .transform((config) => {
     const extensionPoints = (config.targeting ?? config.extension_points ?? []).map((targeting) => {
+      const buildManifest: BuildManifest = {
+        assets: {
+          [AssetIdentifier.Main]: {
+            filepath: `${config.handle}.js`,
+            module: targeting.module,
+          },
+          ...(targeting.should_render?.module
+            ? {
+                [AssetIdentifier.ShouldRender]: {
+                  filepath: `${config.handle}-conditions.js`,
+                  module: targeting.should_render.module,
+                },
+              }
+            : null),
+        },
+      }
+
       return {
         target: targeting.target,
         module: targeting.module,
@@ -32,6 +64,7 @@ export const UIExtensionSchema = BaseSchema.extend({
         default_placement_reference: targeting.default_placement,
         capabilities: targeting.capabilities,
         preloads: targeting.preloads ?? {},
+        build_manifest: buildManifest,
       }
     })
     return {...config, extension_points: extensionPoints}
@@ -53,9 +86,11 @@ const uiExtensionSpec = createExtensionSpecification({
     return validateUIExtensionPointConfig(directory, config.extension_points, path)
   },
   deployConfig: async (config, directory) => {
+    const transformedExtensionPoints = config.extension_points.map(addDistPathToAssets)
+
     return {
       api_version: config.api_version,
-      extension_points: config.extension_points,
+      extension_points: transformedExtensionPoints,
       capabilities: config.capabilities,
       name: config.name,
       description: config.description,
@@ -64,7 +99,33 @@ const uiExtensionSpec = createExtensionSpecification({
     }
   },
   getBundleExtensionStdinContent: (config) => {
-    return config.extension_points.map(({module}) => `import '${module}';`).join('\n')
+    const main = config.extension_points
+      .map(({module}) => {
+        return `import '${module}'; `
+      })
+      .join('\n')
+
+    const assets: {[key: string]: Asset} = {}
+    config.extension_points.forEach((extensionPoint) => {
+      // Start of Selection
+      Object.entries(extensionPoint.build_manifest.assets).forEach(([identifier, asset]) => {
+        if (identifier === AssetIdentifier.Main) {
+          return
+        }
+
+        assets[identifier] = {
+          identifier: identifier as AssetIdentifier,
+          outputFileName: asset.filepath,
+          content: `import '${asset.module}'`,
+        }
+      })
+    })
+
+    const assetsArray = Object.values(assets)
+    return {
+      main,
+      ...(assetsArray.length ? {assets: assetsArray} : {}),
+    }
   },
   hasExtensionPointTarget: (config, requestedTarget) => {
     return (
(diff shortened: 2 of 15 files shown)


packages/app/src/cli/services/dev/extension/bundler.ts
```diff
@@ -20,6 +20,11 @@ export interface FileWatcherOptions {
   payloadStore: ExtensionsPayloadStore
 }
 
+interface Overrides {
+  contents?: string
+  outputPath?: string
+}
+
 export async function setupBundlerAndFileWatcher(options: FileWatcherOptions) {
```
elanalynn, line 28: Referencing [comment in original PR](https://github.com/Shopify/cli/pull/4812/files#r1836217274) - this function will go away. Where do we move this logic? @isaacroldan

packages/app/src/cli/services/dev/app-events/app-watcher-esbuild.ts
```diff
@@ -31,35 +32,43 @@ export class ESBuildContextManager {
   setAbortSignal(signal: AbortSignal) {
     this.signal = signal
     this.signal?.addEventListener('abort', async () => {
-      const allDispose = Object.values(this.contexts).map((context) => context.dispose())
+      const allDispose = Object.values(this.contexts).map((context) => context.map((ctxt) => ctxt.dispose()))
```
isaacroldan, line 35: This turns into an array of arrays, not sure if `Promise.all()` will handle that correctly. Better to use a `.flat()` at the end to flatten it into a single array of Promises.

packages/app/src/cli/services/dev/app-events/app-watcher-esbuild.ts
```diff
@@ -79,11 +90,36 @@ export class ESBuildContextManager {
   }
 
   async deleteContexts(extensions: ExtensionInstance[]) {
-    const promises = extensions.map((ext) => this.contexts[ext.handle]?.dispose())
+    const promises = extensions.map((ext) => this.contexts[ext.handle]?.map((context) => context.dispose()))
     await Promise.all(promises)
```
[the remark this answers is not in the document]
isaacroldan: Ah! nice bot, but I was faster 😆

packages/app/src/cli/services/dev/app-events/app-watcher-esbuild.ts
```diff
@@ -79,11 +90,36 @@ export class ESBuildContextManager {
...
     })
   }
+
+  private async extensionEsBuildOptions(
+    extension: ExtensionInstance,
+    overrides: Pick<BundleOptions, 'stdin'> & {outputFileName?: string},
+  ) {
+    const outputPath = overrides.outputFileName
+      ? joinPath(dirname(extension.outputPath), overrides.outputFileName)
+      : extension.getOutputPathForDirectory(this.outputPath)
```
isaacroldan, line 107: This is not correct, the output path for the non-main assets should be:

```ts
const mainOutputPath = extension.getOutputPathForDirectory(this.outputPath)
const assetsDirectory = dirname(mainOutputPath)
const assetOutputPath = joinPath(assetsDirectory, asset.outputFileName)
```

The way it is right now, doesn't work for Consistent-dev (it is storing each file in a different folder)

My recommendation is to remove the "overrides" concept, you should always pass options to the config:
```ts
  private async extensionEsBuildOptions(
    extension: ExtensionInstance,
    stdin: StdinOptions,
    outputFileName: string,
  )
```

vividviolet (changes requested):
I think we need to simplify this. Each asset should just be its own file - there's no need to create custom stdin content

packages/app/src/cli/models/extensions/specification.ts
```diff
@@ -50,7 +60,7 @@ export interface ExtensionSpecification<TConfiguration extends BaseConfigType =
   experience: ExtensionExperience
   dependency?: string
   graphQLType?: string
-  getBundleExtensionStdinContent?: (config: TConfiguration) => string
+  getBundleExtensionStdinContent?: (config: TConfiguration) => {main: string; assets?: Asset[]}
```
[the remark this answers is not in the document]
elanalynn: We didn't want to change the existing flow for other extensions.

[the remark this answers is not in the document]
vividviolet: But you are still changing it? It went from a string to an object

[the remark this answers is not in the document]
elanalynn: But main is still a string? Is this something we could change in a follow up?

packages/app/src/cli/services/build/extension.ts
```diff
@@ -91,12 +91,14 @@ export async function buildUIExtension(extension: ExtensionInstance, options: Ex
     env.APP_URL = options.appURL
   }
 
+  const {main, assets} = extension.getBundleExtensionStdinContent()
```
vividviolet, line 94: I think we should revert this change and not touch the stdin since we no longer need to use it for other assets. We can do a simple map for the build options since we have access to the config

packages/app/src/cli/services/build/extension.ts
```diff
@@ -106,6 +108,25 @@ export async function buildUIExtension(extension: ExtensionInstance, options: Ex
       stdout: options.stdout,
       sourceMaps: extension.isSourceMapGeneratingExtension,
     })
+    if (assets) {
```
vividviolet, line 111:
```suggestion
   if (extension.configuration.extension_points) {
      const promises: Promise<void>[] = []
      extension.configuration.extension_points.forEach((extensionPoint) => {
        // Start of Selection
        Object.entries(extensionPoint.build_manifest.assets).forEach(([identifier, asset]) => {
          if (identifier === AssetIdentifier.Main) {
            return
          }

          promises.push(
            bundleExtension({
              minify: true,
              outputPath: joinPath(dirname(asset.filepath), `${extension.configuration.handle}-conditions.js`),
              entryPoints: [joinPath(dirname(extension.outputPath), asset.module)],
              environment: options.environment,
              env,
            }),
          )
        })
      })
      await Promise.all(promises)
    }
```

packages/app/src/cli/services/build/extension.ts
vividviolet: Nevermind, it looks like `extension.getBundleExtensionStdinContent` is being called in multiple places including the `createContexts` method in `app-watcher-esbuild` which isn't even using this code path. It would be nice to refactor one day but it doesn't block this from shipping.

vividviolet (approved):
Tophat'd

packages/app/src/cli/services/dev/app-events/app-watcher-esbuild.ts
```diff
@@ -31,35 +32,46 @@ export class ESBuildContextManager {
...
+      contexts.push(context)
+
+      if (assets) {
+        await Promise.all(
+          assets.map(async (asset) => {
+            const esbuildOptions = await this.extensionEsBuildOptions(
+              {
+                contents: asset.content,
+                resolveDir: extension.directory,
+                loader: 'ts',
```
isaacroldan, line 65: Is this intended to be `ts`? can it be anything else?

packages/app/src/cli/services/dev/app-events/app-watcher-esbuild.ts
```diff
@@ -31,35 +32,46 @@ export class ESBuildContextManager {
...
-        sourceMaps: true,
-      })
-
+        mainOutputPath,
+      )
       const context = await esContext(esbuildOptions)
-      this.contexts[extension.handle] = context
+      contexts.push(context)
+
+      if (assets) {
```
isaacroldan, line 58: I didn't love how this function turned out so I rewrote here:

```ts
  async createContexts(extensions: ExtensionInstance[]) {
    const promises = extensions.map(async (extension) => {
      const {main, assets} = extension.getBundleExtensionStdinContent()
      const mainOutputPath = extension.getOutputPathForDirectory(this.outputPath)
      const esbuildOptions = await this.extensionEsBuildOptions(
        {
          contents: main,
          resolveDir: extension.directory,
          loader: 'tsx',
        },
        mainOutputPath,
      )
      const mainContexPromise = esContext(esbuildOptions)

      const assetContextPromises = (assets ?? []).map(async (asset) => {
        const esbuildOptions = await this.extensionEsBuildOptions(
          {
            contents: asset.content,
            resolveDir: extension.directory,
            loader: 'ts',
          },
          joinPath(dirname(mainOutputPath), asset.outputFileName),
        )
        return esContext(esbuildOptions)
      })

      this.contexts[extension.handle] = await Promise.all(assetContextPromises.concat(mainContexPromise))
    })

    await Promise.all(promises)
  }
```

packages/app/src/cli/services/dev/app-events/app-watcher-esbuild.ts
belalsj: Yes, for this use case we only need `ts`. But as a future improvement, this can be configurable from within the build_manifest object.

packages/app/src/cli/models/extensions/specifications/ui_extension.ts
```diff
@@ -25,13 +40,31 @@ export const UIExtensionSchema = BaseSchema.extend({
...
+      const buildManifest: BuildManifest = {
+        assets: {
+          [AssetIdentifier.Main]: {
+            filepath: `dist/${config.handle}.js`,
+            module: targeting.module,
+          },
+          ...(targeting.should_render?.module
+            ? {
+                [AssetIdentifier.ShouldRender]: {
+                  filepath: `dist/${config.handle}-conditions.js`,
```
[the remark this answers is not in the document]
belalsj: This is intentional. We wanted the full path to be driven by one source of truth (cli). As such, if we wanted to change that in the future, we don't have to make any changes to core.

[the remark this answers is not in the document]
isaacroldan: but this breaks the output of `app-watcher-esbuild`, you end up with the conditions file being nested in 
`/<extension-id>/dist/dist/...`

packages/app/src/cli/services/dev/extension/payload.ts
```diff
@@ -24,6 +25,43 @@
...
+              assets,
+            }
+          }
+
+          return extensionPoint
+        },
+      ),
+    )
+
+    const extensionPoints = getExtensionPoints(extensionPointsWithAssets, url)
```
isaacroldan, line 63: A few proposals for this function:

- Use `UIExtensionSchemaType['extension_points']` to get the right types for the extension points
- By using the new Asset interface and a for loop, you can avoid the extra `reduce`

So the implementation is simplified quite a bit:

```ts
const extensionPointsTyped: UIExtensionSchemaType['extension_points'] = extension.configuration.extension_points
    const extensionPointsWithAssets = await Promise.all(
      extensionPointsTyped.map(async (extensionPoint) => {
        const buildManifest = extensionPoint.build_manifest
        if (!buildManifest.assets) return extensionPoint
        const assets: {[key: string]: Asset} = {}

        for (const [name, asset] of Object.entries(buildManifest.assets)) {
          assets[name] = {
            name,
            url: `${url}${joinPath('/assets/', asset.filepath)}`,
            lastUpdated: (await fileLastUpdatedTimestamp(joinPath(dirname(extension.outputPath), asset.filepath))) ?? 0,
          }
        }

        return {...extensionPoint, assets}
      }),
    )
```

isaacroldan: One more thing, you might want to extract this into a proper function and call it inside `getExtensionPoints`, because `getExtensionPoints` is doing some validations on the type of the extension points, makes sense to apply this transform AFTER those validations.

packages/app/src/cli/services/build/extension.ts
```diff
@@ -106,6 +108,25 @@ export async function buildUIExtension(extension: ExtensionInstance, options: Ex
...
+      await Promise.all(
+        assets.map(async (asset) => {
+          await bundleExtension({
+            minify: true,
+            outputPath: joinPath(dirname(extension.outputPath), `${extension.configuration.handle}-conditions.js`),
```
isaacroldan, line 116: In here you are not using the output path defined in the asset, this should be `asset.filepath` right?
But then you'll have the same problem of the double nested `dist/dist` folder.

PR #4852, opened 2024-11-14
Update `shopify theme dev` to no longer display redundant URLs

karreiro:

Fixes https://github.com/Shopify/develop-advanced-edits/issues/424

We've received feedback that URLs in the `shopify theme dev` command are redundant, so this PR fixes that.

While the CLI-UI kit encourages the usage of hyperlinks, they appeared duplicated in some contexts. Additionally, most terminal emulators that do not support hyperlinks generally support clickable URLs, so this clean-up doesn't represent a change in the UX for most users, but cleans the UI for all of them.

### How to test your changes?

- Run `shopify theme dev`

Screenshot 2024-11-14 at 10 55 33

### Post-release steps

N/A

### Measuring impact

### Checklist

### DIFF
--- a/.changeset/pink-garlics-act.md
+++ b/.changeset/pink-garlics-act.md
@@ -1,0 +1,5 @@
+---
+'@shopify/theme': patch
+---
+
+Update `shopify theme dev` to no longer display redundant URLs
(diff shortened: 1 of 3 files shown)


jamesmengo (approved):

amcaplan (approved):
I think this change makes sense, however maybe it should encourage us to handle the footnote URLs differently when the URL and text are the same. The footnote system makes sense when there is meant to be clickable text that matches a URL, but if the URL is the same as the text, we shouldn't need to make this change at all.

PR #4862, opened 2024-11-15
Get creation options without loading the app during `app init`

isaacroldan:
To create during `app init`, we need to load what we call the "Creation Options" which include two key values: the default scopes and whether the app is launchable or not.

We did this by loading the local app first to check those values, but we can't do that if the template has custom extensiosn, because we don't have the specifications at this point (because there is no remote app yet!)

Instead of loading the app: just load the basic configuration from the toml and the `webs` configuration. That's all we need for the creation options.

### How to test your changes?
The easiest way is to put a breakpoint in `services/init/init.ts` line 201.
Validate that the `creationOptions` object is valid in these three cases:
- An extensions-only app
- A default remix app
- A remix template with custom extensions and scopes (ask for an example of this)

### Measuring impact

### Checklist

### DIFF
--- a/packages/app/src/cli/models/app/loader.test.ts
+++ b/packages/app/src/cli/models/app/loader.test.ts
@@ -3147,8 +3147,6 @@ describe('getAppConfigurationState', () => {
   })
 })
 
-// ... existing imports ...
-
 describe('loadConfigForAppCreation', () => {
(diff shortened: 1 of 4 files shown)


craigmichaelmartin:
> A remix template with custom extensions and scopes (ask for an example of this)

Do you have an example?

PR #4873, opened 2024-11-18
add tests for dev-session errors

isaacroldan:

Fixes #0000 

### Measuring impact

### Checklist


packages/app/src/cli/services/dev/processes/dev-session.test.ts
```diff
@@ -0,0 +1,157 @@
...
+import {AppLinkedInterface} from '../../../models/app/app.js'
+import {AppEventWatcher} from '../app-events/app-event-watcher.js'
+import {buildAppURLForWeb} from '../../../utilities/app/app-url.js'
+import {testAppLinked, testDeveloperPlatformClient, testWebhookExtensions} from '../../../models/app/app.test-data.js'
+import {formData} from '@shopify/cli-kit/node/http'
+import {describe, expect, test, vi, beforeEach} from 'vitest'
+import {AbortSignal} from '@shopify/cli-kit/node/abort'
+import {flushPromises} from '@shopify/cli-kit/node/promises'
+
+vi.mock('@shopify/cli-kit/node/fs')
```
shauns, line 12: I don't think this mock is used, or archiver. It would be good if they weren't mocked

packages/app/src/cli/services/dev/processes/dev-session.test.ts
```diff
@@ -0,0 +1,157 @@
...
+      appId: 'app123',
+      organizationId: 'org123',
+    }
+  })
+
+  test('creates a new dev session successfully when receiving the app watcher start event', async () => {
+    // When
+    await pushUpdatesForDevSession({stderr, stdout, abortSignal: new AbortSignal()}, options)
+    await appWatcher.start()
+    await flushPromises()
```
shauns, line 83: This is optional because I'm not sure how practical it is. Am wondering about how to make the watcher easier to test.

Lots of these tests are running `.start()` and then waiting for flushPromises. I wonder if `start` could return something the caller could choose to await that resolves once the watcher has reached the settled state?

```
async function start(...) {
  const watchedFinishedStartup = new Promise((resolve) => {
    //...
    resolve()
  }
  return {
    watcherFinishedStartup
  }
}
```

shauns (changes requested):

packages/app/src/cli/services/dev/processes/dev-session.test.ts
isaacroldan: Actually they are used, the upload process includes zipping the bundle folder, that's what is being mocked here with archiver (zipping) and fs (loading the zip file in memory)

packages/app/src/cli/services/dev/processes/dev-session.test.ts
isaacroldan: I tried this, but we still need the `flushPromises()`, the thing is that after "start" then the dev-session receives the event and start processing it, but we can await that, so the test continues to the `expects` before the event is processed and fails. 

I don't think there is an easy alternative to this 🤔

packages/app/src/cli/services/dev/processes/dev-session.test.ts
shauns: ok. I'd like us to start to limit this sort of mock -- I think we're best off exercising the file system in our tests -- working with files (and not arbitrary data structures) is _so_ fundamental to what the tool does

isaacroldan:
### Merge activity

* **Nov 22, 6:01 PM EST**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/Shopify/cli/4873).

PR #4874, opened 2024-11-18
Clean up app management client module interpretation quality concerns

amcaplan:

Responding to comments by @shauns regarding code quality ([1](https://github.com/Shopify/cli/pull/4777/files#r1846175012), [2](https://github.com/Shopify/cli/pull/4777/files#r1846173629))

Removes an extra field being passed around but not used (`registrationUid`), and reorganizes types to be more logical (rather than Partners-centric) which also allows DRYing up some code in `AppManagementClient` around the shared logical types.

Most importantly, `config` is now passed around as an object rather than a JSON string.

### How to test your changes?

In both Partners and App Management:

1. Create a new app
2. Generate some extensions (admin link work well)
3. Deploy a version, confirming that the diff appears correctly
4. Re-release the initial version before deploying extensions, confirming the diff appears correctly (extensions are eliminated)

### Measuring impact

### Checklist

### DIFF
--- a/packages/app/src/cli/utilities/developer-platform-client/app-management-client.ts
+++ b/packages/app/src/cli/utilities/developer-platform-client/app-management-client.ts
@@ -8,13 +8,15 @@ import {RemoteSpecification} from '../../api/graphql/extension_specifications.js
 import {
   DeveloperPlatformClient,
   Paginateable,
-  ActiveAppVersion,
+  AppVersion,
+  AppVersionWithContext,
   AppDeployOptions,
   AssetUrlSchema,
   AppVersionIdentifiers,
   DevSessionOptions,
   filterDisabledFlags,
   ClientName,
+  AppModuleVersion,
 } from '../developer-platform-client.js'
 import {PartnersSession} from '../../services/context/partner-account-info.js'
 import {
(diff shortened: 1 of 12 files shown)


packages/app/src/cli/utilities/developer-platform-client/app-management-client.ts
```diff
@@ -978,3 +942,19 @@ function mapBusinessPlatformStoresToOrganizationStores(storesArray: ShopNode[]):
     } as OrganizationStore
   })
 }
+
+function appModuleVersion(mod: ReleasedAppModuleFragment): Required<AppModuleVersion> {
+  return {
+    registrationId: mod.uuid,
```
shauns, line 948: non-blocking on this: but I would still like us to sort this particular thing out -- lots of IDs in the result format, non-obvious matching from API to result format, & app-management _not_ being the default behaviour (currently partners is setting the pattern)

amcaplan: Only Partners needs ID, with the result that eventually we will be able to drop ID and just retain UUID. (Title is not an immutable ID, but a mutable display concern.) I'm not really very worried about that now, and I don't think it's worth investing in further cleaning this up, when we'll be refactoring code just to delete it.

PR #4875, opened 2024-11-18
[ShopifyVM] Use Functions App Management API for Dev Dash

mssalemi:
Notes: See https://github.com/Shopify/shopify/pull/554957 for the Core branch and tophat instructions

https://github.com/Shopify/shopify-functions/issues/483

```
pnpm run shopify app function schema --path=./new-app
```

We added a new enpoint in core on the `dd` branch - 

This PR also updates the CLI interface for developer platform client to ensure compatibility with both the Partners Client and the App Management Client. Given the differences in their respective request structures and required parameters, this change aims to maintain a clean and functional API while adapting to the varied needs of both clients.

Notes on AMF vs Partners Flow:

**Partners Client - Flow**
- With the cli's partners-client request, the [query](https://github.com/Shopify/cli/blob/82b872fb4d9aff53031e0283366141ec6a2db7e2/packages/app/src/cli/api/graphql/functions/target_schema_definition.ts#L3-L6). Hits the partners backend [here](https://github.com/Shopify/partners/blob/main/components/partners_dashboard/app/graphql/graph_api/cli/query.rb#L150-L152).
- CLI Sends the variables:
```
{
  apiKey,
  target, (but we are really using the handle)
  version
}
```
- Which then makes its way to the ShopifyInternalApi through [here](https://github.com/Shopify/partners/blob/main/lib/shopify_internal_api/functions/apis.rb).
- Which then get to the core though the url /services/internal//functions/apis/graphql.
- This hit this our Services::Internal::Functions::Apis controller [here](https://github.com/Shopify/shopify/blob/main/areas/core/shopify/components/support/app/controllers/services/internal/functions/apis/graphql_controller.rb).
- This controller requires request be sent be partners apps.

**App Management Client - Flow**
- This core branch provides a new endpoints that app management client will use to make authenticated request with core. https://github.com/Shopify/shopify/pull/554957
- Core end points requires appId rather than api key (following existing pattern in app maangement framework)
- AMF sends variables: (makes more sense, target has other fields aswell)
```
{
  handle:
  version
}
```

- Updated `targetSchemaDefinition` and `apiSchemaDefinition` methods in the App Management client
- Modified the Partners client to maintain backwards compatibility with existing API key requirements
- Modified the DeveloperPlatformClient interface to support both implementations
- Added comments indicating the temporary nature of the API key parameter

- Partners request different slightly in that it requires an API key parameter which will be removed once fully migrated to AMF, and the variable naming is slightly different. But in both queries the the definition will be correctly returned. I opted to use the generated types from the queries for the interface. This requires some custom logic for the partners request which transforms the input and add the api Key. We can remove this once we fully migrated to app management framework as we do not need to use the partners proxy anymore. 
- App Management client uses the newer schema definition queries without API key requirement
- Maintained backwards compatibility to ensure smooth transition 

### How to test your changes?

* with your local CLI

Test Partners Request - Test in Production
```
export USE_APP_MANAGEMENT_API=1
export NODE_TLS_REJECT_UNAUTHORIZED=0
export SHOPIFY_SERVICE_ENV=spin
export SPIN_INSTANCE=$(spin show -o name)
pnpm run shopify app function schema --path=./test123
```

Test AMF Request - Test in SPIN
```
export USE_APP_MANAGEMENT_API=1
export NODE_TLS_REJECT_UNAUTHORIZED=0
export SHOPIFY_SERVICE_ENV=spin
export SPIN_INSTANCE=$(spin show -o name)
pnpm run shopify app function schema --path=./test123
```

### Measuring impact

### Checklist

### DIFF
--- a/bin/get-graphql-schemas.js
+++ b/bin/get-graphql-schemas.js
@@ -44,7 +44,13 @@ const schemas = [
     pathToFile: 'areas/core/shopify/db/graphql/webhooks_schema_unstable_public.graphql',
     localPath: './packages/app/src/cli/api/graphql/webhooks/webhooks_schema.graphql',
     branch: 'dd',
-  }
+  },
+  {
+    repo: 'shopify',
+    pathToFile: 'areas/core/shopify/db/graphql/functions_cli_api_schema_unstable_public.graphql',
+    localPath: './packages/app/src/cli/api/graphql/functions/functions_cli_schema.graphql',
+    branch: 'dd',
+  },
 ]
 
 function runCommand(command, args) {
--- a/graphql.config.ts
+++ b/graphql.config.ts
@@ -82,5 +82,6 @@ export default {
     appManagement: projectFactory('app-management', 'app_management_schema.graphql'),
     admin: projectFactory('admin', 'admin_schema.graphql', 'cli-kit'),
     webhooks: projectFactory('webhooks', 'webhooks_schema.graphql'),
+    functions: projectFactory('functions', 'functions_cli_schema.graphql', 'app'),
   },
 }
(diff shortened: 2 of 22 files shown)


saga-dasgupta (approved):
Code looks good and was able to successfully tophat using functions dev dashboard spin instance👍🏻

craigmichaelmartin (approved):

PR #4884, opened 2024-11-19
Introduce dev dash webhooks API

cdarne:

See https://github.com/Shopify/shopify/pull/554877 for the Core implementation and tophat instructions

### Measuring impact

### Checklist

### DIFF
--- a/packages/app/src/cli/services/webhook/request-sample.ts
+++ b/packages/app/src/cli/services/webhook/request-sample.ts
@@ -25,6 +25,7 @@ export interface UserErrors {
   fields: string[]
 }
 
+// eslint-disable-next-line @shopify/cli/no-inline-graphql
 export const sendSampleWebhookMutation = `
   mutation samplePayload($topic: String!, $api_version: String!, $address: String!, $delivery_method: String!, $shared_secret: String!, $api_key: String) {
     sendSampleWebhook(input: {topic: $topic, apiVersion: $api_version, address: $address, deliveryMethod: $delivery_method, sharedSecret: $shared_secret, apiKey: $api_key}) {


packages/app/project.json
```diff
@@ -80,15 +80,17 @@
         "{projectRoot}/src/cli/api/graphql/business-platform-destinations/generated/**/*.ts",
         "{projectRoot}/src/cli/api/graphql/business-platform-organizations/generated/**/*.ts",
         "{projectRoot}/src/cli/api/graphql/app-dev/generated/**/*.ts",
-        "{projectRoot}/src/cli/api/graphql/app-management/generated/**/*.ts"
+        "{projectRoot}/src/cli/api/graphql/app-management/generated/**/*.ts",
+        "{projectRoot}/src/cli/api/graphql/webhooks/generated/**/*.ts"
```
shauns, line 84: I wish we had a nicer way to do this but all the plumbing looks great 👌

packages/cli-kit/src/public/node/api/webhooks.ts
```diff
@@ -0,0 +1,43 @@
...
+  const fqdn = await appManagementFqdn()
+  const url = `https://${fqdn}/webhooks/unstable/graphql.json`
+  const result = limiter.schedule<TResult>(() =>
+    graphqlRequestDoc<TResult, TVariables>({
+      query,
+      api,
+      url,
+      token,
+      variables,
+    }),
```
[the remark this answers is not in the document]
shauns: You're right, robot is wrong. They're not taking over just yet

packages/app/src/cli/utilities/developer-platform-client/app-management-client.ts
```diff
@@ -734,25 +738,41 @@ export class AppManagementClient implements DeveloperPlatformClient {
...
+    }
+    const result = await webhooksRequest(query, await this.token(), variables)
+
     return {
       sendSampleWebhook: {
-        samplePayload: '',
-        headers: '{}',
-        success: true,
-        userErrors: [],
+        samplePayload: result.cliTesting?.samplePayload ?? '{}',
```
shauns, line 755: Lots of `?.` here. Should `cliTesting` always be present, is it an error if its not? Maybe the schema should be adjusted?

packages/app/src/cli/api/graphql/webhooks/queries/public-api-versions.graphql
```diff
@@ -0,0 +1,7 @@
+query publicApiVersions {
+  publicApiVersions {
+    displayName
+    handle
```
shauns, line 4: Given `handle` is the only thing we use of this query, should we limit what we fetch to that?

packages/app/src/cli/utilities/developer-platform-client/app-management-client.test.ts
```diff
@@ -187,3 +193,156 @@ describe('versionDeepLink', () => {
...
+    const client = new AppManagementClient()
+    client.token = () => Promise.resolve('token')
+    const topics = await client.topics({api_version: 'invalid'})
+
+    // Then
+    expect(topics.webhookTopics.length).toEqual(0)
+  })
+})
+
+describe('sendSampleWebhook', () => {
```
shauns, line 252: I think maybe a thing relevant to test here is the mapping of `input` to the mutation variables

shauns (approved):
Approving with caveat around your other PR merging and the branch name in codegen being updated.

pacocastrotech (approved):

cdarne:
Thanks @shauns for the suggestions: let me see how I can implement those 👍

packages/cli-kit/src/public/node/api/webhooks.ts
```diff
@@ -0,0 +1,43 @@
...
+ * @param variables - GraphQL variables to pass to the query.
+ * @returns The response of the query of generic type <T>.
+ */
+export async function webhooksRequest<TResult, TVariables extends Variables>(
+  query: TypedDocumentNode<TResult, TVariables>,
+  token: string,
+  variables?: TVariables,
+): Promise<TResult> {
+  const api = 'Webhooks'
+  const fqdn = await appManagementFqdn()
```
MitchDickinson, line 30: Kind of a nit but I wonder if we should rename `appManagementFqdn` since this FQDN is now shared amongst several remote services (app management, core webhooks, functions). We could still expose `appManagementFqdn` but it could directly delegate to this new name. It would avoid any ambiguity that the webhooks commands are hosted on the app management API as they are not.

I'm not sure what the best name would be though. `nonShopScopedShopifyFqdn`? I don't really like that either haha.

MitchDickinson (approved):
Great work on this 👏

packages/cli-kit/src/public/node/api/webhooks.ts
cdarne: there's `partnersFqdn` => partners.shopify.com and `developerDashboardFqdn` => dev.shopify.com. How about `appFqdn` or `appApiFqdn`?

PR #4902, opened 2024-11-21
Fix: Handle localized URLs in storefront password validation

AaronBarcos:


AaronBarcos:
Closed due to an account synchronisation error

PR #4918, opened 2024-11-25
Allow identifying apps by API key, not ID

amcaplan:

Fixes https://github.com/Shopify/develop-app-inner-loop/issues/1863 

Depends on https://github.com/Shopify/shopify/pull/556269

1. Removes app ID from `shopify.app.toml`
2. Adds a query to fetch app by API key, not ID
3. Uses that alternative query in cases where we have API key but not ID

This seems to be sufficient to accomplish the goal, at least for now, since in all cases we are fetching the active app version before doing other requests. I had originally considered e.g. adding a cache for the API key to ID conversion, but it doesn't seem necessary.

If it ever is necessary, the types should tell us as such.

### How to test your changes?

1. Work against https://github.com/Shopify/shopify/pull/556269
2. Create an app, see that `app_id` isn't in the TOML
3. Deploy, release, etc. and see that everything works fine!

### Measuring impact

### Checklist

### DIFF
--- a/packages/app/src/cli/utilities/developer-platform-client/app-management-client.ts
+++ b/packages/app/src/cli/utilities/developer-platform-client/app-management-client.ts
@@ -790,12 +790,12 @@ export class AppManagementClient implements DeveloperPlatformClient {
 
   async targetSchemaDefinition(
     input: SchemaDefinitionByTargetQueryVariables,
-    _apiKey: string,
+    apiKey: string,
     organizationId: string,
-    appId?: string,
   ): Promise<string | null> {
     try {
-      const appIdNumber = String(numberFromGid(appId!))
+      const {app} = await this.activeAppVersionRawResult({apiKey, organizationId})
+      const appIdNumber = String(numberFromGid(app.id))
       const token = await this.token()
       const result = await functionsRequestDoc<SchemaDefinitionByTargetQuery, SchemaDefinitionByTargetQueryVariables>(
         organizationId,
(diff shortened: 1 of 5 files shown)


packages/app/src/cli/models/organization.ts
```diff
@@ -18,6 +18,12 @@ export interface MinimalAppIdentifiers {
   organizationId: string
 }
 
+export interface MinimalAppIdentifiersPossiblyExcludingId {
```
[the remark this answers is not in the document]
isaacroldan: But do we need the ID here at all? can't we just remove it from the type?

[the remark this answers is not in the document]
amcaplan: If we have ID available, we would prefer to use it. However, we need to be flexible enough to handle a case when ID is unavailable but we have API key.

If you look at the logic [here](https://github.com/Shopify/cli/pull/4918/files#diff-06b22225f82056c865dd6d82b1c00f8ce938cea2285962890542d5581bd6d429R802), this will make more sense.

It's possible we could just always use API key as identifier, but we are using ID in all the other mutations (e.g. create version), so consistency seems preferable.

shauns, line 21: Are we using `MinimalAppIdentifiers` anywhere now? Should this be the one with the nice short name and `MinimalAppIdentifiers` be the one with the long obscure name?

packages/app/src/cli/services/context.ts
```diff
@@ -62,8 +62,8 @@ export const appFromId = async (options: AppFromIdOptions): Promise<Organization
       organizationId = org.id
     }
   }
-  const app = await developerPlatformClient.appFromId({
-    id: options.id ?? 'no-id-available',
+  const app = await developerPlatformClient.appFromIdentifiers({
+    id: options.id,
```
shauns, line 66: should this still be supported? I thought we were _only_ identifying by API key now?

packages/app/src/cli/utilities/developer-platform-client/app-management-client.ts
```diff
@@ -217,7 +219,9 @@ export class AppManagementClient implements DeveloperPlatformClient {
     return (await this.session()).accountInfo
   }
 
-  async appFromId(appIdentifiers: MinimalAppIdentifiers): Promise<OrganizationApp | undefined> {
+  async appFromIdentifiers(
+    appIdentifiers: MinimalAppIdentifiersPossiblyExcludingId,
+  ): Promise<OrganizationApp | undefined> {
```
isaacroldan, line 224: Can't we change `appFromIdentifiers` to need just `api-key`? (and org-id temporarily while it's removed)
There is no need for this method to have the `id` optional when most use cases will only pass api-key.

There a couple cases where we do have the `id` available, but if it's not needed, why not just simplify everything?

isaacroldan (approved):

shauns (approved):

PR #4922, opened 2024-11-26
Print logs to stderr

gonzaloriestra:

Fixes https://github.com/Shopify/cli/issues/3863

This PR moves these types of CLI output from `stdout` to `stderr`
- debug, informational, and error messages emitted though the existing `outputInfo`, `outputWarn`, `outputDebug` functions
- success and completion messages emitted through `outputSuccess` and `outputCompleted`
- rendered content emitted through `renderOnce`

Primary, non-message output is still output to `stdout`, but this intent has been more explicit by calling a new `outputResult` function. 

### Measuring impact

### Checklist


packages/cli-kit/src/public/node/output.ts
```diff
@@ -258,7 +256,7 @@ export const clearCollectedLogs = (): void => {
  * @param content - The content to be output to the user.
  * @param logger - The logging function to use to output to the user.
  */
-export function outputInfo(content: OutputMessage, logger: Logger = consoleLog): void {
+export function outputInfo(content: OutputMessage, logger: Logger = consoleWarn): void {
```
charlespwd, line 259: May I suggest a `export function output(content: outputMessage, logger: Logger = consoleLog): void {`

?

That way all the `output{log,warn,error}` would all go to stderr and we'd make it explicit which one goes to stdout?

STDOUT = scriptable output
STDERR = logging

charlespwd: idk I think output is output and logging is logging. I personally don't see a problem with these two things:
- logging that goes to STDERR and requires a loglevel 
- output goes to STDOUT and doesn't require a loglevel.

STDOUT is for piping & scripts. Why would we want a loglevel in there? What am I missing?

amcaplan: Because, while we don't currently support it, many programs have a `silent` log level which doesn't print anything. And I think we should ultimately support that. Keeping everything with log levels gives us more room to fine-tune logging behavior.

amcaplan: Here is a good example: https://docs.npmjs.com/cli/v10/using-npm/config#loglevel

Silent is just another log level.

charlespwd: Wouldn't that mean that `--json --silent` do nothing at all? There's always `&> /dev/null`? 

Whenever I'm scripting, I'm mostly annoyed by shit showing up in STDOUT and that's why I'd want `--silent`, but if it doesn't, then I don't see the point. 

One of `pnpm` or `nx` is an example offender here. If I want to run the `shopify theme language-server` with the dev build in Shopify/cli, I can only do that by running the build node package stuff because otherwise pnpm & nx are outputting crap to STDOUT that don't fit the LSP.

Same goes for `pnpm shopify theme check --json`.

charlespwd: For that reason, whenever I want to test a Shopify/theme-tools bump in vim. I have a script in my $HOME/bin that is this:

```sh
#!/usr/bin/env bash

if [[ -z $DEBUG ]]; then
  node /Users/charles/src/github.com/Shopify/cli/packages/cli/bin/dev.js "$@"
else
  node --inspect-brk /Users/charles/src/github.com/Shopify/cli/packages/cli/bin/dev.js "$@"
fi
```

I need the path of the full command because the pnpm script pollutes STDOUT and breaks the LSP.

amcaplan: Have you tried `pnpm --silent shopify:run ...`

`--silent` quiets pnpm

`shopify:run` skips the build step (you have to run `pnpm shopify build` first, but anyway you need to do that in your script)

gordonhirsch: In the end, I thought it best to have an explicit function for non-log output. I called it `outputResult`. Otherwise, to me, the use of `outputInfo` becomes somewhat error-prone. Without the separate function, for JSON and other result output, you have to remember to set the right `logger` parameter value. The use of the that parameter would also be pretty easy to miss when reading the code. 

The latest commit implements `outputResult` and calls if from the code that outputs non-log results.

packages/cli-kit/src/public/node/output.ts
```diff
@@ -267,12 +279,12 @@ export function outputInfo(content: OutputMessage, logger: Logger = consoleLog):
...
  * Outputs a success message to the user.
  * Success messages receive a special formatting to make them stand out in the console.
- * Note: Success messages are sent through the standard output.
+ * Note: Success messages are sent through the standard error.
  *
  * @param content - The content to be output to the user.
  * @param logger - The logging function to use to output to the user.
  */
-export function outputSuccess(content: OutputMessage, logger: Logger = consoleLog): void {
+export function outputSuccess(content: OutputMessage, logger: Logger = consoleWarn): void {
```
gordonhirsch, line 287: The change to `consoleWarn` here (and below in `outputCompleted`) seems in line with the intent of this PR, but let me know if I am missing something.

gordonhirsch:
@gonzaloriestra and @amcaplan, I have some tophatting left to do, but would appreciate some feedback on the general direction.

gonzaloriestra (commented):
It's looking good! Just a few suggestions:
- I see some usages of `consoleLog` and `consoleWarn` outside cli-kit, but we could replace them with `outputResult` and `outputInfo`, no? We have to simplify the options to print, because there are too many and it will be confusing for someone without enough context.
- I'd try to make `consoleLog` and friends private. Or if it's not possible, maybe a linter rule to complain if someone tries to use them and suggest `outputX`.
- I think another source of confusion is having `outputInfo`/`renderInfo`, `outputWarn`/`renderWarn`,`outputSuccess`/`renderSuccess`... What if we take the opportunity to rename the latter functions to `renderInfoBox`, `renderWarningBox`...?

gonzaloriestra:
> But I wonder if we can either remove those parameters entirely or change them to an enumeration

I think that's a good idea.


> Also wondering if there is a way we can unify the output/render functions into a single set.

That could be nice as well. I agree all of them are just printing stuff, so they should be named similarly. If we just replace all the appearances of renderX to outputXBox, it should be a simple task and a big win.

gordonhirsch:
The latest commits remove the use of `console*` and other functions from outside of `output.ts`.  The `console*` functions have been made private to `output.ts`. Others can be made private too but will require some test updates to do so, I think.

It looks like the `logger` parameter can't be removed from the `output*` functions as I had hoped because it's used to pass in custom writers. 

> What if we take the opportunity to rename the latter functions to renderInfoBox, renderWarningBox...?

Following up on this, I'm questioning this particular terminology. Using "Box" exposes a presentation detail that maybe shouldn't be exposed. I do agree that we should do something to reduce confusion.

packages/cli-kit/src/public/node/output.ts
```diff
@@ -320,11 +332,25 @@ export function outputWarn(content: OutputMessage, logger: Logger = consoleWarn)
...
 
+/**
+ * Logs an unformatted message at the given log level.
+ * Note: By default,  messages are sent through the standard error.
+ *
+ * @param content - The content to be output to the user.
+ * @param logLevel - The log level associated with the message.
+ * @param logger - The logging function to use to output to the user.
+ */
+export function output(content: OutputMessage, logLevel: LogLevel = 'info', logger: Logger = consoleWarn): void {
```
gordonhirsch, line 343: New function to support calls from the `render*` without exposing `console*` functions.

packages/cli-kit/src/public/node/ui.tsx
```diff
@@ -594,12 +583,12 @@ interface RenderTextOptions {
...
+export function renderText({text, logLevel = 'info'}: RenderTextOptions) {
   let textWithLineReturn = text
   if (!text.endsWith('\n')) textWithLineReturn += '\n'
 
-  if (isUnitTest()) collectLog(logLevel, textWithLineReturn)
-  outputWhereAppropriate(logLevel, logger, textWithLineReturn)
+  output(textWithLineReturn, logLevel)
+
   return textWithLineReturn
 }
```
gordonhirsch, line 593: I've removed the one usage of this function in this PR, so we could consider removing it altogether.

packages/cli-kit/src/public/node/output.ts
```diff
@@ -320,11 +332,25 @@ export function outputWarn(content: OutputMessage, logger: Logger = consoleWarn)
...
+  const message = stringifyMessage(content)
+  outputWhereAppropriate(logLevel, logger, message)
+}
+
 /**
  * Prints a new line in the terminal.
  */
 export function outputNewline(): void {
-  console.log()
+  console.warn()
```
amcaplan, line 353: `console.warn` should be avoided, as it may decorate output text. Instead, we should reuse `consoleWarn` which doesn't use `console` at all.

packages/cli/src/cli/commands/debug/command-flags.ts
```diff
@@ -53,7 +53,7 @@ export default class CommandFlags extends Command {
       const header = `${columns.join(',')}\n`
       const rows = data.map((obj) => columns.map((key) => obj[key]).join(',')).join('\n')
       const csvString = header + rows
-      renderText({text: csvString, logger: consoleLog})
+      outputResult(csvString)
```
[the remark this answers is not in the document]
gordonhirsch: Thanks, yes, I thought there was already a newline at the end of this string, but that was incorrect. Added in https://github.com/Shopify/cli/pull/4922/commits/247521dd6f317f78d3d49b6ba620a3271e032832

packages/cli-kit/src/public/node/output.ts
gonzaloriestra: What about moving it to `private/node/output.ts`? That way, we prevent it from being used outside of cli-kit.

packages/cli-kit/src/public/node/output.ts
```diff
@@ -365,33 +377,6 @@ export interface OutputProcess {
...
-
-/**
- * Prints a warning message in the console.
- *
- * @param message - The message to print.
- */
-export function consoleWarn(message: string): void {
-  process.stderr.write(`${withOrWithoutStyle(message)}\n`)
-}
-
```
gonzaloriestra, line 394: We are removing these public functions from cli-kit and also [this one](https://github.com/Shopify/cli/pull/4922/files#diff-eeb7a18ff165c45604a28270426f305ce28336a85960bf3e376efa1533a994e5L597), but that would be a breaking change. Is it worth keeping them instead and maybe adding some comments and a linter rule to avoid its usage?

isaacroldan: I'd say, check if Hydrogen is using them, if not, we can remove them.

packages/cli-kit/src/public/node/output.ts
```diff
@@ -245,20 +245,32 @@ export function collectLog(key: string, content: OutputMessage): void {
...
+ * Note: By default, result messages are logged at info level to stdout.
+ *
+ * @param content - The content to be output to the user.
+ * @param logger - The logging function to use to output to the user.
+ */
+export function outputResult(content: OutputMessage, logger: Logger = consoleLog): void {
+  const message = stringifyMessage(content)
+  if (isUnitTest()) collectLog('info', content)
+  outputWhereAppropriate('info', logger, message)
+}
```
amcaplan, line 263: Could this just be:

```suggestion
export function outputResult(content: OutputMessage, logger: Logger = consoleLog): void {
  output(content, 'info', logger)
}
```

(importing `output` of course)

packages/cli-kit/src/public/node/output.ts
amcaplan: This is a very good change. It means we are effectively requiring dependent packages to go through our higher-level primitives for logging.

I'm glad it's not a breaking change!

packages/features/steps/create-app.steps.ts
```diff
@@ -12,6 +12,40 @@ interface ExtensionConfiguration {
   outputPath: string
 }
 
+/**
```
amcaplan, line 15: Are these changes related?

packages/cli-kit/src/public/node/output.ts
```diff
@@ -245,20 +245,32 @@ export function collectLog(key: string, content: OutputMessage): void {
...
 /**
- * Ouputs information to the user.
+ * Outputs command result information to the user.
+ * Result messages don't get additional formatting.
+ * Note: By default, result messages are logged at info level to stdout.
+ *
+ * @param content - The content to be output to the user.
+ * @param logger - The logging function to use to output to the user.
+ */
+export function outputResult(content: OutputMessage, logger: Logger = consoleLog): void {
```
amcaplan, line 259: It's very weird that we are calling it `outputResult` and yet much of what it outputs isn't results. Do we have a plan to clean this up and move usages over to another function which will be explicitly for non-result stuff to be printed without modification to stderr?

amcaplan, line 259: I want to suggest we avoid having positional params with defaults. It's possible we'd want to add params in the future, and that means passing around a lot of `undefined`s. Especially for such a central function, a bit of extra design is very worthwhile.

As an aside, the param doesn't seem to be used at all. Is the idea to follow up with moving things to `consoleErr` gradually? A default seems to drive against that, instead encouraging continued use of stdout instead of stderr.

packages/features/steps/create-app.steps.ts
gonzaloriestra: https://github.com/Shopify/cli/pull/4922#discussion_r2088769202

packages/cli-kit/src/public/node/output.ts
gonzaloriestra: It was decided [here](https://github.com/Shopify/cli/pull/4922#discussion_r1927771073). I think it makes sense, we basically want to use this to print the important stuff, usually the "result". What are we printing that you don't think it's a result? Any other suggestion for the name?

gonzaloriestra: Yeah, it's unused. I guess Gordon added it to follow the same shape of the other outputX functions, where the logger is passed in some cases. I also agree with using named params, but the refactor is getting too big, I'd prefer to visit it later.

packages/cli-kit/src/public/node/output.ts
gonzaloriestra: Yeah, thanks!

packages/cli-kit/src/public/node/output.ts
gonzaloriestra: Well, it's not for us, but I hope it doesn't break other projects out there 😅

packages/cli-kit/src/public/node/output.ts
amcaplan: If it's unused, can we just remove it for now? Easier to make the interface thin now and then add options later.

gonzaloriestra: Done!

packages/cli-kit/src/public/node/output.ts
amcaplan: Yeah, I'm fine with breaking compatibility here, hopefully people understand that we don't provide guarantees around cli-kit backcompat and they test thoroughly on each upgrade. There aren't too many packages that depend on us directly.

amcaplan (approved):
Looks ready to me!

PR #4924, opened 2024-11-26
Improvements to App reload speed

isaacroldan:

Improves app reloading performance by caching or avoiding expensive operations.

Contributes to: https://github.com/Shopify/develop-app-outer-loop/issues/1263

- `~200ms` Introduces a new `reloadApp` function that reuses data from the previous app instance.
- `~200ms` Caches JSON schema validators to avoid recompilation on each validation. 
- `~100ms` Replaces git ignore checking with a more efficient alternative: the `ignore` package
- `~1200ms` Improves handling of CLI version warning to only check once per command execution.
- Reduces the file-watcher debouncing time to 200ms

Before this PR, time to reload an app: `1500-2000ms`
New time: `~15-20ms` ❗ 
Improvement: `100x`

### How to test your changes?

1. Run `dev` on an app with multiple extensions using the app management API.
2. Make changes to extension configuration files.
3. Verify that reloading is faster and the app continues to function correctly.

### Measuring impact

### Checklist

### DIFF
--- a/packages/app/src/cli/models/app/loader.test.ts
+++ b/packages/app/src/cli/models/app/loader.test.ts
@@ -9,7 +9,6 @@ import {
   AppLoaderMode,
   getAppConfigurationState,
   loadConfigForAppCreation,
-  resetDidCheckForGlobalCLIWarning,
 } from './loader.js'
 import {LegacyAppSchema, WebConfigurationSchema} from './app.js'
 import {DEFAULT_CONFIG, buildVersionedAppSchema, getWebhookConfig} from './app.test-data.js'
@@ -329,7 +328,6 @@ wrong = "property"
 
   test('shows warning if using global CLI but app has local dependency', async () => {
     // Given
-    resetDidCheckForGlobalCLIWarning()
     vi.mocked(globalCLIVersion).mockResolvedValue('3.68.0')
     vi.mocked(localCLIVersion).mockResolvedValue('3.65.0')
     const mockOutput = mockAndCaptureOutput()
--- a/packages/app/src/cli/models/app/loader.ts
+++ b/packages/app/src/cli/models/app/loader.ts
@@ -18,6 +18,7 @@ import {
   AppCreationDefaultOptions,
   AppLinkedInterface,
 } from './app.js'
+import {showMultipleCLIWarningIfNeeded} from './validation/multi-cli-warning.js'
 import {configurationFileNames, dotEnvFileNames} from '../../constants.js'
 import metadata from '../../metadata.js'
 import {ExtensionInstance} from '../extensions/extension-instance.js'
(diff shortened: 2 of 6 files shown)


isaacroldan:
* **#4924** <a href="https://app.graphite.dev/github/pr/Shopify/cli/4924?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Shopify/cli/4924?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`

<h2></h2>

packages/app/src/cli/services/dev/app-events/file-watcher.ts
```diff
@@ -69,9 +69,9 @@ export async function startFileWatcher(
 
   /**
    * Debounced function to emit the accumulated events.
-   * This function will be called at most once every 500ms to avoid emitting too many events in a short period.
+   * This function will be called at most once every 200ms to avoid emitting too many events in a short period.
    */
-  const debouncedEmit = debounce(emitEvents, 500)
+  const debouncedEmit = debounce(emitEvents, 200)
```
shauns, line 74: I think there's opportunity to squeeze this further, maybe.

Hypothesis: often when the file watcher is triggered, it is because the user is editing a single file (app.toml, extension.toml). Waiting for N ms before taking action, in case of any other events coming in, is wasteful in that scenario.

Maybe, we could use the `leading: true, trailing: true` debounce options? https://lodash.com/docs/4.17.15#debounce

If I'm following correctly then in the single edit scenario, the event will fire immediately, without any delay. In the scenario where many files are being edited, the first change will cause an immediate event, but then subsequent changes over the debounce window will be flattened to a single event. 

The drawback is in that latter case two events fired where previously there may have been one. So the opening hypothesis needs to be true for this to work.

packages/app/src/cli/models/app/loader.ts
```diff
@@ -319,15 +343,18 @@ class AppLoader<TConfig extends AppConfiguration, TModuleSpec extends ExtensionS
     const extensions = await this.loadExtensions(directory, configuration)
 
     const packageJSONPath = joinPath(directory, 'package.json')
-    const name = await loadAppName(directory)
-    const nodeDependencies = await getDependencies(packageJSONPath)
-    const packageManager = await getPackageManager(directory)
+
+    // These don't need to be processed again if the app is being reloaded
+    const name = this.previousApp?.name ?? (await loadAppName(directory))
+    const nodeDependencies = this.previousApp?.nodeDependencies ?? (await getDependencies(packageJSONPath))
```
shauns, line 349: Checking on this one: If I generate extension whilst dev is running, are we expecting that new extension to be picked up? It may involve dependency changes

shauns (approved):

packages/app/src/cli/models/app/loader.ts
isaacroldan: If using workspaces the dependency will be in the extension's package.json, but I guess it is true for non-worksapce apps.
In any case, just checked and `nodeDependencies` is not used for anything

packages/app/src/cli/models/app/loader.test.ts
```diff
@@ -2364,7 +2367,7 @@ wrong = "property"
       cmd_app_all_configs_any: true,
       cmd_app_all_configs_clients: JSON.stringify({'shopify.app.toml': '1234567890'}),
       cmd_app_linked_config_name: 'shopify.app.toml',
-      cmd_app_linked_config_git_tracked: false,
+      cmd_app_linked_config_git_tracked: true,
```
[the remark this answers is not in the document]
isaacroldan: I actually forced this to true to test that the new `checkIfGitTracked` works correctly

Previously it was false because `checkIfIgnoredInGitRepository` needs to load a valid git repository to work (which is slow) and was always false in this test.

PR #5020, opened 2024-12-04
Fetch notifications in background

gonzaloriestra:

We are fetching the notifications at the beginning of each command and caching the result for 1 hour. That approach has two disadvantages:
- When the network operation hangs or takes too long (the timeout is 3 seconds), the command has to wait with no feedback for the user ([error example](https://app.bugsnag.com/shopify/cli/errors/673374fe41bcaf15edcf8ac7?event_id=67407225010cf312ace50000&i=em&m=ea))
- If we want to send an urgent notification (like a downtime), we have to wait for the local cache to expire (up to 1 hour)

- Fetch the notifications in a detached process that is run when the command finishes, so the users don't have to wait, and save the result in the cache
- At the beginning of each command, we check the cache to show the required notifications

https://github.com/user-attachments/assets/ed9746d6-c2e2-4437-820d-5e764d6ff3b8

### How to test your changes?

Install the snapshot from this branch: `npm i -g @shopify/cli@0.0.0-snapshot-20241216120051`

- `shopify cache clear`
- `shopify version` => the notifications are downloaded when the command finishes
- `shopify version` => notification should be shown
- `shopify version` => no more notifications

The cache in Mac is stored here: `~/Library/Preferences/shopify-cli-kit-nodejs/config.json`

### Measuring impact

### Checklist


packages/cli-kit/bin/fetch-notifications.js
```diff
@@ -0,0 +1,21 @@
...
+    const response = await fetch(url, {signal: AbortSignal.timeout(3 * 1000)})
+    if (response.status === 200) {
+      cacheStore(`notifications-${url}`, await response.text())
+      console.log(`Notifications from ${url} cached successfully`)
+    }
+  } catch (error) {
+    const errorMessage = `Error fetching notifications: ${error.message}`
+    console.error(errorMessage)
+    const {sendErrorToBugsnag} = await import('../dist/public/node/error-handler.js')
+    await sendErrorToBugsnag(errorMessage, 'unexpected_error')
```
gonzaloriestra, line 17: Ensured that this works from here: https://app.bugsnag.com/shopify/cli/errors/675049cdac655414327f515f?event_id=675049cd010d594612b60000&i=sk&m=cm

packages/cli-kit/bin/fetch-notifications.js
```diff
@@ -0,0 +1,21 @@
+import fetch from 'node-fetch'
+import {notificationsUrl} from '../dist/public/node/notifications-system.js'
+import {cacheStore} from '../dist/private/node/conf-store.js'
```
isaacroldan, line 3: I don't think you can do this. This paths won't exist once we bundle the CLI.
You can try by doing `pnpm bundle` and then running the CLI

gonzaloriestra: Ah, good point. I forgot to try with the snapshot, let me check it...

packages/cli-kit/assets/fetch-notifications.js
```diff
@@ -0,0 +1,25 @@
+import fetch from 'node-fetch'
+import Conf from 'conf'
```
gonzaloriestra, line 2: The bundle is not able to find these imports. Removing the need of `conf` would be hard, so I'll try to convert it to TS, compile and call it from the dist folder, so that it can use dependencies.

packages/cli-kit/src/public/node/notifications-system.ts
```diff
@@ -113,25 +109,69 @@ async function renderNotifications(notifications: Notification[]) {
...
-async function fetchNotifications(): Promise<string> {
-  outputDebug(`No cached notifications found. Fetching them...`)
+export async function fetchNotifications(): Promise<Notifications> {
+  outputDebug(`Fetching  notifications...`)
   const response = await fetch(url(), {signal: AbortSignal.timeout(3 * 1000)})
   if (response.status !== 200) throw new Error(`Failed to fetch notifications: ${response.statusText}`)
-  return response.text() as unknown as string
+  const rawNotifications = await response.text()
+  await cacheNotifications(rawNotifications)
+  const notifications: object = JSON.parse(rawNotifications)
```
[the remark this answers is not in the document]
gonzaloriestra: Right, we shouldn't cache wrong values, although the JSON is in our control.

I've updated it to cache at the end. So if the parsing fails, it should throw an error and it wouldn't be cached. Is that ok?

packages/cli-kit/src/public/node/notifications-system.ts
```diff
@@ -72,14 +67,15 @@ export async function showNotificationsIfNeeded(
...
     // This is very prone to becoming a circular dependency, so we import it dynamically
     const {sendErrorToBugsnag} = await import('./error-handler.js')
     await sendErrorToBugsnag(errorMessage, 'unexpected_error')
   }
 }
 
-function skipNotifications(environment: NodeJS.ProcessEnv): boolean {
-  return isTruthy(environment.CI) || isTruthy(environment.SHOPIFY_UNIT_TEST) || jsonOutputEnabled(environment)
+function skipNotifications(environment: NodeJS.ProcessEnv = process.env): boolean {
+  return isTruthy(environment.CI) || isTruthy(environment.SHOPIFY_UNIT_TEST)
```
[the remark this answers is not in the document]
gonzaloriestra: We are using this check in two different places:
- To fetch the notifications
- To show the banner

And we want to fetch them in the background even if the command includes `--json`.

We could create two different functions (`skipFetchingNotifications` and `skipShowingNotifications`) if you think it would be clearer.

PR #5026, opened 2024-12-04
Increase file size limits for localization files

isaacroldan:

Fixes #0000 

### Measuring impact

### Checklist

### DIFF
--- a/packages/app/src/cli/utilities/extensions/locales-configuration.ts
+++ b/packages/app/src/cli/utilities/extensions/locales-configuration.ts
@@ -3,7 +3,7 @@ import {glob} from '@shopify/cli-kit/node/fs'
 import {AbortError, BugError} from '@shopify/cli-kit/node/error'
 import fs from 'fs'
 
-const L10N_FILE_SIZE_LIMIT = 16 * 1024
+const L10N_FILE_SIZE_LIMIT = 32 * 1024
 const L10N_BUNDLE_SIZE_LIMIT = 256 * 1024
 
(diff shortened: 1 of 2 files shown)


packages/app/src/cli/utilities/extensions/locales-configuration.test.ts
```diff
@@ -101,7 +101,7 @@ describe('loadLocalesConfig', () => {
       const es = joinPath(localesPath, 'es.json')
 
       await mkdir(localesPath)
-      const bigArray = JSON.stringify(new Array(3000).fill('a'))
+      const bigArray = JSON.stringify(new Array(4000).fill('a'))
```
craigmichaelmartin, line 104: Did we have to increase this? We are only increasing the per file size limit, so was only expecting to see the above test changed ("Throws if one locale is too big") and not this one ("Throws if bundle is too big")?

packages/app/src/cli/utilities/extensions/locales-configuration.ts
```diff
@@ -3,7 +3,7 @@ import {glob} from '@shopify/cli-kit/node/fs'
 import {AbortError, BugError} from '@shopify/cli-kit/node/error'
 import fs from 'fs'
 
-const L10N_FILE_SIZE_LIMIT = 16 * 1024
+const L10N_FILE_SIZE_LIMIT = 20 * 1024
```
gonzaloriestra, line 6: Why 20? Do we have any limit server-side?

PR #5033, opened 2024-12-04
Port fetchTheme / fetchThemes to graphQL

catlee:

Fixes #0000 

### Measuring impact

### Checklist

### DIFF
--- a/packages/cli-kit/src/public/node/themes/api.ts
+++ b/packages/cli-kit/src/public/node/themes/api.ts
@@ -14,6 +14,8 @@ import {
   OnlineStoreThemeFileBodyInputType,
   OnlineStoreThemeFilesUpsertFileInput,
 } from '../../../cli/api/graphql/admin/generated/types.js'
+import {GetThemes} from '../../../cli/api/graphql/admin/generated/get_themes.js'
+import {GetTheme} from '../../../cli/api/graphql/admin/generated/get_theme.js'
 import {restRequest, RestResponse, adminRequestDoc} from '@shopify/cli-kit/node/api/admin'
 import {AdminSession} from '@shopify/cli-kit/node/session'
 import {AbortError} from '@shopify/cli-kit/node/error'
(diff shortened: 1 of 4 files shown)


packages/cli-kit/src/public/node/themes/api.ts
```diff
@@ -28,15 +30,48 @@ export type ThemeParams = Partial<Pick<Theme, 'name' | 'role' | 'processing' | '
...
+    }
+    const {nodes, pageInfo} = response.themes
+    nodes.forEach((theme) => {
+      const t = buildTheme({
+        id: parseGid(theme.id),
+        processing: theme.processing,
+        role: theme.role.toLowerCase(),
+        name: theme.name,
+      })
+      if (t !== undefined) {
```
craigmichaelmartin, line 65: Any reason for the `!== undefined` check?

Could we do

```suggestion
      if (t) {
```

or can typescript not infer that theme is `Theme` without this?

packages/cli-kit/src/public/node/themes/api.ts
```diff
@@ -28,15 +30,48 @@ export type ThemeParams = Partial<Pick<Theme, 'name' | 'role' | 'processing' | '
...
+    const response = await adminRequestDoc(GetThemes, session, {after})
+    if (!response.themes) {
+      unexpectedGraphQLError('Failed to fetch themes')
+    }
+    const {nodes, pageInfo} = response.themes
+    nodes.forEach((theme) => {
+      const t = buildTheme({
+        id: parseGid(theme.id),
+        processing: theme.processing,
+        role: theme.role.toLowerCase(),
```
craigmichaelmartin, line 62: Curious about the `toLowerCase()`. This is the second time we do this for the arbitrary(?) value for `role` which we are passing to `buildTheme`. Do we expect all roles to be lower case? Should this be moved into `buildTheme`?

craigmichaelmartin (approved):

jamesmengo (approved):

packages/cli-kit/src/public/node/themes/api.ts
catlee: I'm not sure! Let me see if I can figure out an answer.

EvilGenius13 (approved):

packages/cli-kit/src/public/node/themes/api.ts
catlee: Yeah, it looks like `if (t)` should work here, since `buildTheme` returns either `Theme` or `undefined`.

lukeh-shopify (approved):

PR #5045, opened 2024-12-05
Default to using the latest patch version of shopify_function

nickwesselman:
I was going to update to use [v1.0.3](https://github.com/Shopify/shopify-function-javascript/releases/tag/v1.0.3) but shouldn't we just encourage use of the latest patch?

### DIFF
--- a/packages/app/src/cli/services/generate/extension.ts
+++ b/packages/app/src/cli/services/generate/extension.ts
@@ -246,7 +246,7 @@ async function uiExtensionInit({
 
         if (templateLanguage === 'javascript') {
           await changeIndexFileExtension(directory, srcFileExtension)
-          // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+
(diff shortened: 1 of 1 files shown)


packages/app/src/cli/services/generate/extension.ts
```diff
@@ -246,7 +246,7 @@ async function uiExtensionInit({
 
         if (templateLanguage === 'javascript') {
           await changeIndexFileExtension(directory, srcFileExtension)
-          // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
```
nickwesselman, line 249: This was apparently some auto-linting, I assume I should leave it.

nickwesselman:
@andrewhassan @saulecabrera Any concerns with auto-accepting new patch releases by default here?

isaacroldan (approved):
LGTM, just need to fix the lint issues

PR #5046, opened 2024-12-05
bump function-runner to 6.4.0

jacobsteves:
#gsd:43536

We are bumping the function input size limit. The function-runner package has been updated to reflect this.

### How to test your changes?

- `rm -r packages/app/dist/cli/services/bin/` (clear out any cached downloaded binaries)
- `pnpm shopify app function build --path <path_to_function_extension>` (downloads and runs Javy)
- `echo "{}" | pnpm shopify app function run --path <path_to_function_extension>` (downloads and runs function-runner)

### Measuring impact

### Checklist

  - none needed
  - none needed

### DIFF
--- a/packages/app/src/cli/services/function/binaries.ts
+++ b/packages/app/src/cli/services/function/binaries.ts
@@ -9,7 +9,7 @@ import fs from 'node:fs'
 import * as gzip from 'node:zlib'
 import {fileURLToPath} from 'node:url'
 
-const FUNCTION_RUNNER_VERSION = 'v6.3.0'
+const FUNCTION_RUNNER_VERSION = 'v6.4.0'
 const JAVY_VERSION = 'v4.0.0'
 // The Javy plugin version does not need to match the Javy version. It should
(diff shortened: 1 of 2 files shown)


packages/app/src/cli/services/function/binaries.ts
```diff
@@ -9,7 +9,7 @@ import fs from 'node:fs'
 import * as gzip from 'node:zlib'
 import {fileURLToPath} from 'node:url'
 
-const FUNCTION_RUNNER_VERSION = 'v6.3.0'
+const FUNCTION_RUNNER_VERSION = 'v6.4.0'
```
shauns, line 12: I understand why we're bumping this. Will someone who is already developing functions get the updated function-runner version? Do they have to make any manual step to move to 6.4.0 from whatever version they currently have?

andrewhassan: I've just triggered a snapshot so that we can validate switching versions causes it to re-download based on this version number.

PR #5084, opened 2024-12-11
Fix exceptions raised for FS level errors

shauns:

Expand the list of "these are environmental issues" with a couple more cases. I'm satisfied that the rarity of these -- very, very rare -- points at environment rather than e.g. the CLI being insufficiently defensive.

Adds a few more patterns to our list of errors that we treat as environmental.

### Measuring impact

### Checklist


craigmichaelmartin:
Is there any value in enumerating all these in this [test](https://github.com/Shopify/cli/blob/39e10174b380d2b129a23636e9f308eac67e861e/packages/cli-kit/src/public/node/error.test.ts#L71)?

PR #5099, opened 2024-12-12
[Themes] Add template generation with liquid and json support

jamesmengo:

Fixes #0000 

### Measuring impact

### Checklist

### DIFF
--- a/packages/cli/README.md
+++ b/packages/cli/README.md
@@ -1962,27 +1962,32 @@ Creates and adds a new template file to your local theme directory
 
 ```
 USAGE
-  $ shopify theme generate template [-n <value>] [--no-color] [--path <value>] [-t
-    product|collection|page|blog|article|custom] [--verbose]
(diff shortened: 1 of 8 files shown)


david-w-shopify:
I _really_ like that if a template doesn't already exist you skip the "Name of the template" question. Good thinking there!

On that note though - if a template doesn't already exist and you generate a Liquid one, it provides a section schema. It should be empty as templates don't have schemas.

Also this is scope creep, but I deleted `gift_card.liquid` and generated that as a JSON template. The default content looks fine, but includes the section `main-gift_card` which doesn't exist in my `sections` folder. I wonder if we could check if such a section exists and if not, prompt to generate that too?

---

Another thing - on selecting `robots`, `robots.txt.liquid` should be generated every time

david-w-shopify:
Since you can already use separate flags I'm not certain if this is necessary, but it would be _really_ nice and even faster to use if I could let the cli parse my decisions from a filename

```
shopify theme generate template --filename=page.landing.json
```

david-w-shopify:
Metaobject resource type is missing

PR #5109, opened 2024-12-13
Add `theme profile` command

macournoyer:

https://hackdays.shopify.io/projects/19234

We want to expose Liquid profiling via:
1. Speedscope, in the browser
2. In our [Theme VSCode extensions](https://github.com/Shopify/theme-tools/compare/sfr-profiling?expand=1)

The new `theme profile` command will be called for both. The VSCode extensions will use the `--json` flag.

Add a `shopify theme profile --url` command that calls SFR asking for profiling data (using the `Accept` header).

### Left to do

- [x] Default command to opening Speedscope in a browser window. JSON only via a `--json` arg.
- [x] Add authentication via OAuth token. Same as the `theme dev` command.
- [x] Add tests.
- [x] Test e2e when https://github.com/Shopify/storefront-renderer/pull/32295 ships
- [x] Test on a product page when https://github.com/Shopify/storefront-renderer/pull/32334 ships

### How to test your changes?

Run `dev shopify theme profile --store [YOUR STORE DOMAIN] --url /`. Should open a browser window with Speedscope in it.

### Measuring impact

### Checklist


packages/theme/src/cli/services/profile.ts
```diff
@@ -0,0 +1,51 @@
...
+  } else {
+    await openProfile(profileJson)
+  }
+}
+
+async function openProfile(profileJson: string) {
+  // Adapted from https://github.com/jlfwong/speedscope/blob/146477a8508a6d2da697cb0ea0a426ba81b3e8dc/bin/cli.js#L63
+  const __filename = fileURLToPath(import.meta.url)
+  const __dirname = dirname(__filename)
+  let urlToOpen = `file://${resolvePath(__dirname, '../../../node_modules/speedscope/dist/release/index.html')}`
```
ianks, line 32: Will this file will be available after building with vite? My understanding is that it would not, since vite will bundle everything and `node_modules` will not exist. May need to import it then use `import.meta.url` or something.

packages/theme/src/cli/services/profile.ts
```diff
@@ -0,0 +1,67 @@
...
+    return import.meta.resolve('speedscope/dist/release/index.html')
+  } else {
+    try {
+      const speedscopePath = require.resolve('speedscope/package.json')
+      const speedscopeDir = speedscopePath.replace('/package.json', '')
+      return `file://${speedscopeDir}/dist/release/index.html`
+    } catch (error) {
+      throw new Error("Can't find Speedscope package")
+    }
+  }
```
macournoyer, line 43: Seems to work both in tests and with `dev shopify`. Is there some packaging magic that could make this not work?

packages/theme/src/cli/commands/theme/profile.ts
```diff
@@ -0,0 +1,43 @@
...
+
+  This command will open a web page with the Speedscope profiler detailing the time spent executing Liquid on the given page.`
+
+  static description = this.descriptionWithoutMarkdown()
+
+  static flags = {
+    ...globalFlags,
+    store: themeFlags.store,
+    password: themeFlags.password,
+    url: Flags.string({
```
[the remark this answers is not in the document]
jamesmengo: Ah, forgot about that.
Thanks for the reminder - you're completely correct.

I'm on board with keeping the `url` flag as-is :)

.changeset/curvy-moons-judge.md
```diff
@@ -0,0 +1,5 @@
+---
+'@shopify/theme': major
```
jamesmengo, line 2: I think this is a minor change according to [semantic versioning](https://semver.org/)

```suggestion
'@shopify/theme': major
```

jamesmengo (commented):
Very cool project! Excited about this change.

I've added a couple of suggestions around changeset management, 1 of which is blocking change.

I wasn't able to get my 🎩 working, but I'll chock it up to user error and try again tommorow



Are there any edge cases you're aware of that we may want to take note of in case we need to patch / support them in the future?

macournoyer:
@EvilGenius13 what is your store?

macournoyer:
Everything should be working now!

We're set to replace https://shopify.dev/docs/storefronts/themes/tools/theme-inspector w/ this!

packages/theme/src/cli/commands/theme/profile.ts
```diff
@@ -0,0 +1,69 @@
...
+    ...globalFlags,
+    store: themeFlags.store,
+    password: themeFlags.password,
+    theme: Flags.string({
+      char: 't',
+      description: 'Theme ID or name of the remote theme.',
+      env: 'SHOPIFY_FLAG_THEME_ID',
+    }),
+    url: Flags.string({
+      description: 'The url to be used as context',
```
EvilGenius13, line 31: Minor question, when we have the `default: /` does that mean if you were to pass the `--url` flag without inputting anything after it, that it would default to using `/`? I tried that out but it didn't work and was curious if it had another meaning.

macournoyer: It's for when you don't supply the `--url` flag at all. I copied this from the `theme console` command.

jamesmengo (approved):
🎩 and code LGTM!

Thanks for the changes, and for building this!
I left a couple of small comments but nothing blocking.

packages/theme/src/cli/services/profile.test.ts
```diff
@@ -0,0 +1,107 @@
...
+      new Response(JSON.stringify(mockProfileData), {
+        status: 200,
+        headers: new Headers({
+          'content-type': 'application/json',
+        }),
+      }),
+    )
+  })
+
+  test('outputs JSON to stdout when asJson is true', async () => {
```
jamesmengo, line 37: In this repo the convention is to follow `//given //when //then`

I'm not fussed about this - feel free to ship and I can refactor that

jamesmengo:
Would like to get @karreiro input in case there's anything here that would be hard to revert once it's rolled out, such as flag semantics

packages/theme/src/cli/commands/theme/profile.ts
```diff
@@ -0,0 +1,69 @@
...
+    const theme = await findOrSelectTheme(adminSession, filter)
+
+    await profile(
+      adminSession,
+      theme.id.toString(),
+      flags.url,
+      flags.json,
+      themeAccessPassword,
+      flags['store-password'],
+    )
```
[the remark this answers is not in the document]
macournoyer: Wohaa! Nice 🌈  thank you!

package.json
```diff
@@ -246,6 +246,9 @@
           "**/{commands,hooks}/**/*.ts!",
           "**/index.ts!"
         ],
+        "ignoreDependencies": [
+          "speedscope"
+        ],
```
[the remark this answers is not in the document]
macournoyer: Trying to remove it in 93d2018, we'll see...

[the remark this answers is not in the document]
macournoyer: I think it worked! 👍

packages/theme/src/cli/services/profile.ts
macournoyer: That doesn't work w/ the packaged CLI. Reported by @karreiro:

image

karreiro (approved):
Thank you, @macournoyer! Really amazing stuff 🔥🚀

packages/theme/src/cli/commands/theme/profile.ts
```diff
@@ -0,0 +1,79 @@
...
+    url: Flags.string({
+      description: 'The url to be used as context',
+      env: 'SHOPIFY_FLAG_URL',
+      default: '/',
+    }),
+    'store-password': Flags.string({
+      description: 'The password for storefronts with password protection.',
+      env: 'SHOPIFY_FLAG_STORE_PASSWORD',
+    }),
+    json: Flags.boolean({
```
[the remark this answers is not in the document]
macournoyer: Fixed in [dbfa8b9](https://github.com/Shopify/cli/pull/5109/commits/dbfa8b900df366da6a99f48a920fa303244f1503)

karreiro:
@Shopify/app-inner-loop, could you please take a look at this?

packages/theme/src/cli/commands/theme/profile.ts
```diff
@@ -0,0 +1,75 @@
...
+
+  static usage = ['theme profile --url /products/classic-leather-jacket']
+
+  static descriptionWithMarkdown = `Profile the Shopify Liquid on a given page.
+
+  This command will open a web page with the Speedscope profiler detailing the time spent executing Liquid on the given page.`
+
+  static description = this.descriptionWithoutMarkdown()
+
+  static flags = {
```
gonzaloriestra, line 23: I think we should enable the `environment` flag for this command

packages/theme/src/cli/commands/theme/profile.ts
```diff
@@ -0,0 +1,75 @@
...
+import {renderTasksToStdErr} from '../../utilities/theme-ui.js'
+import {ensureAuthenticatedThemes} from '@shopify/cli-kit/node/session'
+import {Flags} from '@oclif/core'
+import {globalFlags, jsonFlag} from '@shopify/cli-kit/node/cli'
+import {Task} from '@shopify/cli-kit/node/ui'
+
+export default class Profile extends ThemeCommand {
+  static summary = 'Profile the Liquid rendering of a theme page.'
+
+  static usage = ['theme profile --url /products/classic-leather-jacket']
```
gonzaloriestra, line 15: The url is not mandatory, so I think we should add the usage without flags. That way, I think [this README](https://github.com/Shopify/cli/pull/5109/files#diff-15986190ef9581ab59bcd5483b2c09e7fd0bd439d6f6cddbc94b0b1de094ee51R76) won't include the flag in the command title, which is weird.

```suggestion
  static usage = ['theme profile', 'theme profile --url /products/classic-leather-jacket']
```

packages/cli/package.json
```diff
@@ -102,6 +102,7 @@
   },
   "dependencies": {
     "@ast-grep/napi": "0.33.0",
+    "speedscope": "1.21.0",
```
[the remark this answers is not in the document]
macournoyer: @jamesmengo thanks for the exploration! I'm unclear what solution you're suggesting, out of those:

> Only supported Node 20+ (could use import.meta.resolve)
> Made @shopify/theme public (dependencies would be visible)
> Didn't need to access speedscope's files directly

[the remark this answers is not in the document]
isaacroldan: That's actually not the reason why you need it there for require.resolve to work.

We bundle dependencies, once bundled, you can't access specific files with require.resolve. By adding it as a dependency of CLI, you are basically telling the CLI not to bundle this package, which will affect to the total size and speed of the CLI installation.

You need to solve this issue in the bundle script, manually copying the files you want to access with require.resolve. (cli/bin/bundle.js). 

Sorry but adding dependencies to CLI directly is not allowed as it affects all users performance.

isaacroldan (changes requested):
Adding this here to make it clear that changes are requested, can't add new dependencies to cli :)

frandiox:
Hi 👋 
I had some context about bundling in the CLI thanks to some stuff we did in Hydrogen. I'm attempting to fix Isaac's feedback in this PR: https://github.com/Shopify/cli/pull/5214

With that, we """bundle""" speedscope assets and can access them in prod.

macournoyer:
> Adding this here to make it clear that changes are requested, can't add new dependencies to cli :)

@isaacroldan would it be acceptable if I just bundled the necessary files without adding a new dependency? I just need the .html and .css files from speedscope.

isaacroldan (approved):

PR #5118, opened 2024-12-17
Make developer console support create/delete extensions mid-dev

isaacroldan:

Adds support for handling extension creation and deletion events in the developer console during consistent-dev.

Fixes https://github.com/Shopify/develop-app-outer-loop/issues/1198

- Implements handling for Created and Deleted extension events
- Adds new methods to ExtensionsPayloadStore for adding and removing extensions
- Updates the event handler to maintain the extensions list in the payload store options

KNOWN ISSUES:
- When removing an extension, you need to manually refresh the dev-console.

### How to test your changes?

1. Start dev for an app with UI extensions
2. Create a new UI extension and verify it appears in the dev console (and that the deeplink works)
3. Delete a UI extension and verify it's removed from the dev console (after refreshing)
4. Modify an existing extension and confirm updates are still working

### Measuring impact

### Checklist


packages/app/src/cli/models/extensions/extension-instance.ts
```diff
@@ -149,8 +149,8 @@ export class ExtensionInstance<TConfiguration extends BaseConfigType = BaseConfi
     this.entrySourceFilePath = options.entryPath ?? ''
     this.directory = options.directory
     this.specification = options.specification
-    this.devUUID = `dev-${randomUUID()}`
     this.handle = this.buildHandle()
+    this.devUUID = `dev-${nonRandomUUID(this.handle)}`
```
isaacroldan, line 153: With this we ensure that the `devUUID` is always the same for the same extension. Allowing us to reload it if we need it.

amcaplan: Noting that the handle can change (right?) so it's not 100% foolproof but it's close enough that it should be fine.

packages/app/src/cli/services/dev/extension.ts
```diff
@@ -109,32 +111,64 @@ export interface ExtensionDevOptions {
...
     websocketURL: getWebSocketUrl(options.url),
   }
-  const bundlePath = options.appWatcher.buildOutputPath
-  const payloadStoreRawPayload = await getExtensionsPayloadStoreRawPayload(payloadStoreOptions, bundlePath)
-  const payloadStore = new ExtensionsPayloadStore(payloadStoreRawPayload, payloadStoreOptions)
 
-  outputDebug(`Setting up the UI extensions HTTP server...`, options.stdout)
-  const httpServer = setupHTTPServer({devOptions: options, payloadStore})
+  // NOTE: Always use `payloadOptions`, never `options` directly. This way we can mutate `payloadOptions` without
+  // affecting the original `options` object and we only need to care about `payloadOptions` in this function.
```
amcaplan, line 120: Note that we could make `options` of type `Readonly<ExtensionDevOptions>` but there are also downsides as `Readonly` is a somewhat weak guarantee and might make us think code is safe when it really isn't. Not a mandatory change, but something to consider.

packages/app/src/cli/services/dev/extension.ts
```diff
@@ -109,32 +111,64 @@ export interface ExtensionDevOptions {
...
+            try {
+              const cartUrl = await buildCartURLIfNeeded(payloadOptions.extensions, payloadOptions.storeFqdn)
+              // eslint-disable-next-line require-atomic-updates
+              payloadOptions.checkoutCartUrl = cartUrl
+              // eslint-disable-next-line no-catch-all/no-catch-all
+            } catch (error) {
+              outputDebug(
+                `Failed to build a cart URL for your checkout extension. Use the --checkout-cart-url flag to set a fixed URL.`,
+                payloadOptions.stdout,
+              )
```
amcaplan, line 152: It seems like this is a unique error case because in the general case, if something fails inside `buildCartURLIfNeeded`, we can just crash dev and no harm done. I'm curious why we wouldn't:

1. Crash `dev` in this case
2. Keep going and just not provide the convenience link

Also, why would it not work? I guess just because:

1. Connection to the store fails
2. There are no products to test on

These feel like rare cases. 1 would be very odd given that we have other polling processes, and 2 would be odd not to crash or at least warn that no products exist. Note that the error message is raised inside `fetchProductVariant` so I'd expect we should at least display it.

packages/app/src/cli/models/extensions/extension-instance.ts
isaacroldan: It shouldn't change during `dev` though, which solves the issue of "reloading" the app.

In any case, this is mostly for ui-extensions, and we need to rethink why ui-extensions need a `devUUID` in the first place. Will probably do it in early january.

amcaplan (approved):

PR #5129, opened 2024-12-18
Modify changeset type to patch

jamesmengo:
Changeset type should have been a patch in the original PR -  https://github.com/Shopify/cli/pull/5123

This applies to the main branch as well

Caused the version update to look like this: https://github.com/Shopify/cli/pull/5128


alexanderMontague (approved):
nice catch

PR #5145, opened 2024-12-24
Enable search for dev stores

amcaplan:

Fixes https://github.com/Shopify/develop-app-inner-loop/issues/2455

Allows developers to find dev stores in the list even if they have > 50 stores.

Enables autocomplete search for dev stores, just as we have for apps.

Note this will only work for BP orgs, it's not implemented for Partner orgs.

### How to test your changes?

Using an org with > 50 dev stores, run `app dev` and search for a store by name that is beyond the first 50.

### Measuring impact

### Checklist

### DIFF
--- a/packages/app/src/cli/prompts/dev.test.ts
+++ b/packages/app/src/cli/prompts/dev.test.ts
@@ -49,6 +49,14 @@ const STORE2: OrganizationStore = {
   transferDisabled: false,
   convertableToPartnerTest: false,
 }
+const STORE3: OrganizationStore = {
+  shopId: '3',
+  link: 'link3',
+  shopDomain: 'domain3',
+  shopName: 'store3',
+  transferDisabled: false,
+  convertableToPartnerTest: false,
+}
 
 beforeEach(() => {
   vi.mocked(getTomls).mockResolvedValue({})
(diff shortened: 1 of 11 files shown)


packages/app/src/cli/prompts/dev.test.ts
```diff
@@ -197,6 +199,8 @@ describe('selectStore', () => {
         {label: 'store1 (domain1)', value: '1'},
         {label: 'store2 (domain2)', value: '2'},
       ],
+      hasMorePages: false,
+      search: expect.any(Function),
     })
   })
 })
```
gracejychang, line 206: we should probably add a test here for search terms (ie. renders correct store if searchTerm is not nil), wdyt?

gracejychang (commented):
For some reason, my 🎩 didn't work; I wasn't able to search for the a store which exists on the server side (test 50 exists on dev dash but I don't find it on the CLI)

image

here are the steps I did for the 🎩 , in case I missed something:
1. Revived my spinstance with 50+ dev stores: https://dev.shopify.cli-dev-stores-list.grace-chang.us.spin.dev/dashboard/2/stores (I pulled the latest from core, web, and business platform and ran `dev up`)
2. checked out `dev-store-search`, ran `dev up`
3. on the CLI, ran:
```
export USE_APP_MANAGEMENT_API=1       
export NODE_TLS_REJECT_UNAUTHORIZED=0
export SHOPIFY_SERVICE_ENV=spin
export SPIN_INSTANCE=cli-dev-stores-list
```
4. Created a new app `shopify app init`
5. Ran `shopify app dev --path [path_to_app]`

amcaplan:
@gracejychang I'm not sure why you're seeing this issue, but it definitely works in prod. Maybe a spinstance that isn't fully updated?

gracejychang (approved):
Ariel and I paired and it turns out I just wasn't using the local CLI to test so my commands weren't picking up the changes on this branch 😶‍🌫️  I was running `shopify app dev` instead of `p shopify app dev` 😅

PR #5151, opened 2025-01-03
Remove timer faking from functionWatcher tests

lopert:

Part of https://github.com/Shopify/temp-project-mover-Archetypically-20250320163947/issues/46

This test was reported as flakey, so this attempts to fix it by removing the unnecessary timer faking.

Swaps out our timer faking with a more straightforward `waitFor` that just checks the last returned result from our hook function until it matches what we expect.

I used `waitFor` from `cli-kit`'s `testing/ui`, and had to add it to the public export.
I figured this was fine, but if it's not I could also copy the code from it into our tests more directly.

I also looked at [usePollAppLogsTest](https://github.com/Shopify/cli/blob/main/packages/app/src/cli/services/app-logs/logs-command/ui/components/hooks/usePollAppLogs.test.tsx) where we used a [similar approach](https://github.com/Shopify/cli/blob/f649fa5e9006c0f6dccdc1da1e0ec824629dff88/packages/app/src/cli/services/app-logs/logs-command/ui/components/hooks/usePollAppLogs.test.tsx#L210-L211), but I chose to keep it as is over there since that test has a lot to do with timing, retries, and enqueued timers. It makes sense over there.

### How to test your changes?

```
p vitest function/ui
```

### Post-release steps

N/A

### Measuring impact

### Checklist

### DIFF
--- a/packages/cli-kit/src/public/node/testing/ui.ts
+++ b/packages/cli-kit/src/public/node/testing/ui.ts
@@ -2,7 +2,6 @@ export {
   getLastFrameAfterUnmount,
   render,
   Stdin,
-  waitFor,
   waitForInputsToBeReady,
   waitForContent,
   sendInputAndWait,


isaacroldan:
Vitest has its own `vi.waitFor` ([docs](https://vitest.dev/api/vi#vi-waitfor)) that waits until a condition is true, you can also add a expect inside it. I think is a better solution for your case and you don't need to export cli-kit's implementation.

lopert:
Still relevant, waiting on https://github.com/Shopify/develop-app-inner-loop/issues/2552

PR #5160, opened 2025-01-07
Fix extension_directories wildcards to always be recursive

isaacroldan:

To improve the handling of custom extension directory paths in app configuration by ensuring proper glob pattern support for watching changes in subfolders.

Adds functionality to automatically convert single asterisk wildcards (`*`) to double asterisks (`**`) in extension directory paths. This ensures proper recursive watching of subdirectories when using chokidar for file watching during `dev`.

### How to test your changes?

1. Create an app with and put some extensions in a folder called `custom_extensions`
2. Ensure your app.toml has `extension_directories = ['custom_extensions/*']`
2. Run `dev`
3. Verify that changes in extension files are properly detected and HMR works.

### Measuring impact

### Checklist

### DIFF
--- a/packages/app/src/cli/models/app/app.ts
+++ b/packages/app/src/cli/models/app/app.ts
@@ -59,7 +59,8 @@ function removeTrailingPathSeparator(value: string[] | undefined) {
 // If a path ends with a single asterisk, modify it to end with a double asterisk.
 // This is to support the glob pattern used by chokidar and watch for changes in subfolders.
 function fixSingleWildcards(value: string[] | undefined) {
-  return value?.map((dir) => dir.replace(/\*$/, '**'))
+  // eslint-disable-next-line no-useless-escape
+  return value?.map((dir) => dir.replace(/([^\*])\*$/, '$1**'))
 }
 
 /**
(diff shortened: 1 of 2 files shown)


packages/app/src/cli/models/app/app.ts
```diff
@@ -49,6 +55,13 @@ function removeTrailingPathSeparator(value: string[] | undefined) {
   // eslint-disable-next-line no-useless-escape
   return value?.map((dir) => dir.replace(/[\/\\]+$/, ''))
 }
+
+// If a path ends with a single asterisk, modify it to end with a double asterisk.
+// This is to support the glob pattern used by chokidar and watch for changes in subfolders.
+function fixSingleWildcards(value: string[] | undefined) {
+  return value?.map((dir) => dir.replace(/\*$/, '**'))
```
dmerand, line 62: Wouldn't this regex also convert a double-asterisk to a triple-asterisk? I think you might want a captured non-asterisk, followed by an asterisk and then end-of-line, as your regex:

```suggestion
  return value?.map((dir) => dir.replace(/([^\*])\*$/, '$1**'))
```

isaacroldan: Good catch, this will work for any kind of path type `/` or `\`
I'll add tests to it though :)

dmerand (approved):
Tophat LGTM! Thanks for the changes.

isaacroldan:
Sorry @gonzaloriestra, missed your question. I did not explicitly tested in windows, but this regex will match strings ending in an asterisk and preceded by any other character, so it should work fine with both `\` and `/`

PR #5167, opened 2025-01-07
Add support for Metafields to webhooks

rsperko:

Add support for metafield identifiers in declarative webhook subscriptions, allowing apps to obtain metafields in their webhook payloads for their subscriptions. Core support was added [here](https://github.com/Shopify/shopify/pull/565800), documentation is added [here](https://github.com/Shopify/shopify-app-js/pull/1956), this adds cli support.

Applicable CLI support is being added [here](https://github.com/Shopify/cli/pull/5167)

### Measuring impact

### Checklist

### DIFF
--- a/packages/app/src/cli/models/extensions/specifications/validation/app_config_webhook.ts
+++ b/packages/app/src/cli/models/extensions/specifications/validation/app_config_webhook.ts
@@ -13,7 +13,18 @@ export function webhookValidator(schema: object, ctx: zod.RefinementCtx) {
 }
 
 function validateSubscriptions(webhookConfig: WebhooksConfig) {
-  const {subscriptions = []} = webhookConfig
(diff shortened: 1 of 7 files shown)


packages/app/src/cli/models/extensions/specifications/validation/app_config_webhook.ts
```diff
@@ -76,3 +87,10 @@ function validateSubscriptions(webhookConfig: WebhooksConfig) {
     }
   }
 }
+
```
rsperko, line 90: @shauns I would like to get your early opinion on api version checking like this? I half wanted to add a general "api changes support" where we have a toml (?) file that we define minimum versions for changes and allow checking for that, but this is really the only place that currently needs this sort of support.

Any thoughts are appreciated.

PR #5172, opened 2025-01-08
Hide all banners that tell you to change app URL 

jamieguerrero:

Fixes https://github.com/Shopify/develop-app-inner-loop/issues/2348

Removes this app url change banner since it is no longer needed by Developer Dashboard apps or apps using consistent dev
Screenshot 2025-01-08 at 2 45 41 PM

Additionally, `isCurrentAppSchema` is always true now, so we remove some of the mentions of it from the files we are touching.

### How to test your changes?

**From within the CLI folder:**

**Try not using the App Management API first and confirm partners dashboard app url update banner doesn't show**
- Create an app `p shopify app init`
- Choose **a remix app** 
- `p shopify app dev --path ./name-of-your-app --reset`
- You should **not** see the app url banner saying you can change App URL on Partners.

**Now try using the App Management API and confirm partners dashboard app url update banner doesn't show**
- `USE_APP_MANAGEMENT_API=1 p shopify app dev --path ./name-of-your-app --reset`
  - This flag makes the CLI use the App Management API 
  - We are attempting to dev but with reset flag so you can choose the option '**create as new app**' from dropdown 
  - Choose creating a new app, and put it a new app name and create a new TOML file
  - Screenshot 2025-01-08 at 2 45 12 PM
  - You still shouldn't see any banner

**Confirm that the app config push banner is gone:**
- Confirm that in TOML you do not have `automatically_update_urls_on_dev` defined
-`USE_APP_MANAGEMENT_API=1 p shopify app dev --path ./name-of-your-app`
- When asked to update App URL's select no, never:
```
?  Have Shopify automatically update your app's URL in order to create a preview experience?
✔  No, never
```
- You should **not see** this banner
Screenshot 2025-01-13 at 3 57 13 PM

### Checklist

### DIFF
--- a/packages/app/src/cli/services/dev/ui.tsx
+++ b/packages/app/src/cli/services/dev/ui.tsx
@@ -1,69 +1,9 @@
-import {PartnersURLs} from './urls.js'
 import {Dev, DevProps} from './ui/components/Dev.js'
-import {AppInterface, isCurrentAppSchema} from '../../models/app/app.js'
-import {OrganizationApp} from '../../models/organization.js'
-import {getAppConfigurationShorthand} from '../../models/app/loader.js'
-import {ClientName, DeveloperPlatformClient} from '../../utilities/developer-platform-client.js'
 import React from 'react'
-import {partnersFqdn} from '@shopify/cli-kit/node/context/fqdn'
-import {render, renderInfo} from '@shopify/cli-kit/node/ui'
-import {basename} from '@shopify/cli-kit/node/path'
-import {formatPackageManagerCommand} from '@shopify/cli-kit/node/output'
+import {render} from '@shopify/cli-kit/node/ui'
 import {terminalSupportsPrompting} from '@shopify/cli-kit/node/system'
 import {isTruthy} from '@shopify/cli-kit/node/context/utilities'
 import {isUnitTest} from '@shopify/cli-kit/node/context/local'
-
-export async function outputUpdateURLsResult(
-  updated: boolean,
-  urls: PartnersURLs,
-  remoteApp: OrganizationApp,
-  localApp: AppInterface,
-  developerPlatformClient?: DeveloperPlatformClient,
-) {
-  const usingAppManagementClient = developerPlatformClient?.clientName === ClientName.AppManagement
-  const dashboardURL = await partnersURL(remoteApp.organizationId, remoteApp.id)
-  if (remoteApp.newApp && !usingAppManagementClient) {
-    renderInfo({
-      headline: `For your convenience, we've given your app a default URL: ${urls.applicationUrl}.`,
-      body: [
-        "You can update your app's URL anytime in the",
-        dashboardURL,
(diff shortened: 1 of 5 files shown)


amcaplan:
Honestly, I wonder if we should just get rid of this warning entirely. Now that `shopify app init` creates an app, you don't create an app the first time you run `dev` in most cases. So this warning will only appear when you do a reset. It's just not really particularly effective anymore, and it's not worth fixing if App Management and consistent dev are the way forward...

.changeset/cyan-poets-exist.md
```diff
@@ -0,0 +1,5 @@
+---
+'@shopify/app': minor
+---
+
+Hide App URL change banner when appropriate
```
amcaplan, line 5: I don't think this changeset is necessary, as it only affects app management.

amcaplan (approved):
Approving as this works. Added a couple comments, and I think the most important bit is that maybe we don't need this warning at all?

jamieguerrero:
I removed the warning altogether since I think that was Isaac's ultimate suggestion.

packages/app/src/cli/services/dev/ui.tsx
```diff
@@ -1,64 +1,31 @@
...
 
-export async function outputUpdateURLsResult(
-  updated: boolean,
-  urls: PartnersURLs,
-  remoteApp: OrganizationApp,
-  localApp: AppInterface,
-) {
-  const dashboardURL = await partnersURL(remoteApp.organizationId, remoteApp.id)
-  if (remoteApp.newApp) {
+export async function outputUpdateURLsResult(updated: boolean, urls: PartnersURLs, localApp: AppInterface) {
```
jamieguerrero, line 13: remoteApp is no longer needed to check since we are now only looking if users wanted us to update their TOML's app URLs for them during dev. If not, just let them know they can do it themselves with the tunneled URLs.

packages/app/src/cli/services/dev/ui.tsx
```diff
@@ -1,64 +1,31 @@
...
+
     renderInfo({
-      headline: `For your convenience, we've given your app a default URL: ${urls.applicationUrl}.`,
       body: [
-        "You can update your app's URL anytime in the",
-        dashboardURL,
-        'But once your app is live, updating its URL will disrupt user access.',
+        `To update URLs manually, add the following URLs to ${fileName} under auth > redirect_urls and run\n`,
+        {
+          command: formatPackageManagerCommand(localApp.packageManager, `shopify app config push`, ...pushCommandArgs),
```
isaacroldan, line 23: wait, this command no longer exists! 🤔 
It was deleted months ago

isaacroldan:
Just to confirm, we are OK with not showing any message at all related to the URLs (whether they were updated or not) right?

MitchDickinson:
> Just to confirm, we are OK with not showing any message at all related to the URLs (whether they were updated or not) right?

Do we need UX weigh in here?

PR #5178, opened 2025-01-10
Fix speedscope error on `shopify theme profile`

karreiro:
Fix the speedscope error on `shopify theme profile`

**Before**:
image

**After**:
image


macournoyer (approved):
THANK YOU! 🙇

macournoyer:
> Didn't need to access speedscope's files directly

That would make the profile feature not possible to implement, no?

Issue #5183, opened 2025-01-13
Add `.shopify/metafields.json` to `.gitignore` by default


karreiro:
In `shopify theme dev` and `shopify theme metafields pull`, the `metafields.json` file is automatically created.

While versioning should remain a developer's choice, it seems beneficial to add `.shopify/metafields.json` to `.gitignore` by default when the file exists. This aligns with existing Shopify app dev workflows and offers the convenience by minimizing the need for manual configuration.

PR #5184, opened 2025-01-13
Add `.shopify/metafields.json` to `.gitignore` by default

karreiro:

Fixes https://github.com/Shopify/cli/issues/5183

We've noticed a significant number of users discussing whether to ignore the `.shopify/metafields.json` file. While this decision ultimately rests with the developer, ignoring it by default seems like a sensible approach. This PR proposes implementing that default.

This PR updates the `.shopify/metafields.json` creation to also add it to `.gitignore`.

### How to test your changes?

- Run `shopify theme metafields pull`
- Notice that, when the .shopify/metafields.json file is created, it's also added in the gitingore

### Post-release steps

N/A

### Measuring impact

### Checklist

### DIFF
--- a/packages/theme/src/cli/services/metafields-pull.test.ts
+++ b/packages/theme/src/cli/services/metafields-pull.test.ts
@@ -6,7 +6,7 @@ import {AdminSession, ensureAuthenticatedThemes} from '@shopify/cli-kit/node/ses
 import {mockAndCaptureOutput} from '@shopify/cli-kit/node/testing/output'
 import {metafieldDefinitionsByOwnerType} from '@shopify/cli-kit/node/themes/api'
 import {describe, test, vi, beforeEach, expect, afterEach} from 'vitest'
-import {fileExists, inTemporaryDirectory, readFile, writeFileSync} from '@shopify/cli-kit/node/fs'
+import {eol, fileExists, inTemporaryDirectory, readFile, writeFileSync} from '@shopify/cli-kit/node/fs'
 
 vi.mock('../utilities/theme-store.js')
 vi.mock('../utilities/theme-ui.js')
@@ -85,9 +85,8 @@ describe('metafields-pull', () => {
 
   test('should add metafields to .gitignore', async () => {
     // Given
-    vi.mocked(metafieldDefinitionsByOwnerType).mockImplementation((type: any, _session: AdminSession) => {
-      if (type !== 'PRODUCT') return Promise.resolve([])
-      return Promise.resolve([fakeMetafieldDefinition])
+    vi.mocked(metafieldDefinitionsByOwnerType).mockImplementation(() => {
+      return Promise.resolve([])
     })
 
     await inTemporaryDirectory(async (tmpDir) => {
@@ -100,7 +99,7 @@ describe('metafields-pull', () => {
 
       // Then
       await expect(fileExists(gitIgnorePath)).resolves.toBe(true)
-      await expect(readFile(gitIgnorePath)).resolves.toBe('.DS_Store\n.shopify/metafields.json')
+      await expect(readFile(gitIgnorePath)).resolves.toBe(`.DS_Store${eol()}.shopify/metafields.json`)
     })
 
     expect(capturedOutput.info()).toContain('Metafield definitions have been successfully downloaded.')
--- a/packages/theme/src/cli/services/metafields-pull.test.ts
+++ b/packages/theme/src/cli/services/metafields-pull.test.ts
@@ -92,14 +92,14 @@ describe('metafields-pull', () => {
     await inTemporaryDirectory(async (tmpDir) => {
       // Given
       const gitIgnorePath = `${tmpDir}/.gitignore`
-      writeFileSync(gitIgnorePath, '.DS_Store')
+      writeFileSync(gitIgnorePath, '.DS_Store\n.shopify/secrets.json')
 
       // When
       await metafieldsPull({path: tmpDir})
 
       // Then
       await expect(fileExists(gitIgnorePath)).resolves.toBe(true)
-      await expect(readFile(gitIgnorePath)).resolves.toBe(`.DS_Store\n.shopify/metafields.json`)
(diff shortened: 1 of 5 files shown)


illarionvk:
It looks like your code always adds the line to .gitignore; how to opt-out of it?

karreiro:
Thank you for the review, @illarionvk!

> It looks like your code always adds the line to .gitignore; how to opt-out of it?

It adds the line only when the file gets created. In subsequent synchronizations, it will no longer update the `.gitignore`. Users who prefer to sync the file can undo the `.gitignore` change and they won't be prompted to add that anymore.

packages/theme/src/cli/services/metafields-pull.test.ts
```diff
@@ -83,6 +83,30 @@ describe('metafields-pull', () => {
     expect(capturedOutput.error()).toBeFalsy()
   })
 
+  test('should add metafields to .gitignore', async () => {
+    // Given
+    vi.mocked(metafieldDefinitionsByOwnerType).mockImplementation((type: any, _session: AdminSession) => {
+      if (type !== 'PRODUCT') return Promise.resolve([])
```
aswamy, line 89: we won't need this for this test case

packages/theme/src/cli/services/metafields-pull.ts
```diff
@@ -161,5 +161,29 @@ function writeMetafieldDefinitionsToFile(path: string, content: unknown) {
...
+  }
+
+  const gitIgnoreContent = readFileSync(gitIgnorePath)
+
+  if (gitIgnoreContent.includes(entry)) {
+    // The file already existing in the .gitignore
+    return
+  }
+
+  writeFileSync(gitIgnorePath, `${gitIgnoreContent}\n${entry}`)
```
aswamy, line 188: not sure if we need to `\r\n` for windows systems here

aswamy (approved):
❤️

packages/theme/src/cli/services/metafields-pull.ts
karreiro: Oops! 🤦 Excellent catch, @aswamy! Thanks a lot for catching that. I've fixed that in a new commit :)

packages/theme/src/cli/services/metafields-pull.ts
```diff
@@ -161,5 +161,29 @@ function writeMetafieldDefinitionsToFile(path: string, content: unknown) {
   mkdirSync(shopifyDirectory)
 
   const filePath = joinPath(shopifyDirectory, 'metafields.json')
-  writeFileSync(filePath, JSON.stringify(content, null, 2))
+  const fileContent = JSON.stringify(content, null, 2)
+
+  if (!fileExistsSync(filePath)) {
+    addToGitIgnore(path, '.shopify/metafields.json')
```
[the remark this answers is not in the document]
karreiro: Thank you for the suggestions, @isaacroldan @graygilmore! @nickwesselman also pinged me regarding the importance of consistency in the DX of this directory, so I've updated the PR to apply that change.

Thanks a lot, everyone, for raising this :)

packages/theme/src/cli/services/metafields-pull.ts
```diff
@@ -161,5 +161,31 @@ function writeMetafieldDefinitionsToFile(path: string, content: unknown) {
   mkdirSync(shopifyDirectory)
 
   const filePath = joinPath(shopifyDirectory, 'metafields.json')
-  writeFileSync(filePath, JSON.stringify(content, null, 2))
+  const fileContent = JSON.stringify(content, null, 2)
+
+  if (!fileExistsSync(filePath)) {
+    addToGitIgnore(path, '.shopify')
+  }
```
aswamy, line 168: Since we're ignoring the whole directory, i'd delete the if-statement, and move `addToGitIgnore` right below the `mkdirSync` above

packages/theme/src/cli/services/metafields-pull.ts
```diff
@@ -161,5 +161,31 @@ function writeMetafieldDefinitionsToFile(path: string, content: unknown) {
...
+
+  const gitIgnoreContent = readFileSync(gitIgnorePath)
+
+  if (gitIgnoreContent.includes(entry)) {
+    // The file already existing in the .gitignore
+    return
+  }
+
+  const eol = detectEOL(gitIgnoreContent.toString())
```
aswamy, line 188: Should we be a little safer with checking the gitignore files incase it could have similar-looking entries. E.g. `/.shopify`, `.shopify/foo`, etc.

```suggestion
  const gitIgnoreContent = readFileSync(gitIgnorePath)
  const content = gitIgnoreContent.toString()
  const eol = detectEOL(gitIgnoreContent.toString())

  if (content.split(eol).includes(entry)) {
    // The file already existing in the .gitignore
    return
  }
```

karreiro: Excellent call, @aswamy! I've applied that change :)

packages/theme/src/cli/services/metafields-pull.ts
```diff
@@ -161,5 +161,30 @@ function writeMetafieldDefinitionsToFile(path: string, content: unknown) {
...
+  const fileContent = JSON.stringify(content, null, 2)
+
+  if (!fileExistsSync(filePath)) {
+    addToGitIgnore(path, '.shopify')
+  }
+
+  writeFileSync(filePath, fileContent)
+}
+
+function addToGitIgnore(root: string, entry: string) {
```
frandiox, line 173: We should consider centralizing this logic in cli-kit 😅 
https://github.com/Shopify/hydrogen/blob/6bc6068fbffebbcd07c3d1db97956248bfa6a742/packages/cli/src/lib/shopify-config.ts#L135

frandiox (approved):
Thanks!

packages/theme/src/cli/services/metafields-pull.ts
isaacroldan: yes! can you open a PR for that please?

karreiro: I will do that in a following PR! Great call, @frandiox @isaacroldan! :)

Issue #5192, opened 2025-01-14
[Bug]: theme dev: Failed to fetch theme: <non-existent-theme-id>


jo-elbdev:
### Please confirm that you have:

- [x] Searched [existing issues](.) to see if your issue is a duplicate. (If you’ve found a duplicate issue, feel free to add additional information in a comment on it.)
- [x] Reproduced the issue in [the latest CLI version](https://www.npmjs.com/package/@shopify/cli).

### In which of these areas are you experiencing a problem?

Theme

### Expected behavior

the command shopify theme dev should start the development theme

### Actual behavior

When I use shopify theme dev the CLI fails with the error message:

Failed to fetch theme: <non-existent-theme-id>
No additional information or suggestions are provided, making it difficult for the user to understand the root cause or how to resolve the issue. The theme also does not exist, when I try to use shopify theme delete.

### Verbose output

2025-01-14T12:48:45.479Z: Running command theme dev
2025-01-14T12:48:45.484Z: Notifications to show: 0
2025-01-14T12:48:45.492Z: Ensuring that the user is authenticated with the Theme API with the following scopes:
[]

2025-01-14T12:48:45.492Z: Ensuring that the user is authenticated with the Admin API with the following scopes for the store dermalogica-shop.myshopify.com:
[]

2025-01-14T12:48:45.492Z: Getting session store...
2025-01-14T12:48:45.493Z: Validating existing session against the scopes:
[
  "openid",
  "https://api.shopify.com/auth/shop.admin.graphql",
  "https://api.shopify.com/auth/shop.admin.themes",
  "https://api.shopify.com/auth/partners.collaborator-relationships.readonly",
  "https://api.shopify.com/auth/shop.storefront-renderer.devtools",
  "https://api.shopify.com/auth/partners.app.cli.access",
  "https://api.shopify.com/auth/destinations.readonly"
]
For applications:
{
  "adminApi": {
    "scopes": [],
    "storeFqdn": "dermalogica-shop.myshopify.com"
  }
}

2025-01-14T12:48:45.493Z: Sending Identity Introspection request to URL: https://accounts.shopify.com/oauth/introspection
2025-01-14T12:48:45.493Z: Sending POST request to URL https://accounts.shopify.com/oauth/introspection
With request headers:
 - User-Agent: Shopify CLI; v=3.73.1
 - Keep-Alive: timeout=30
 - Sec-CH-UA-PLATFORM: darwin
 - Content-Type: application/json

[...]

PR #5200, opened 2025-01-15
Fix `shopify theme dev` to no longer fail when development themes expire in internationalized stores

karreiro:

Fixes https://github.com/Shopify/cli/issues/5192

Users were facing the error above because internationalized stores return internationalized error messages in the new GQL API, which were not being handled as expected in the CLI.

This PR changes the `fetchTheme` API to no longer abort or throw an error. API consumers are already handling the absence of the theme when the function returns undefined. Some abort the execution, but others create the necessary theme.

I've checked other possible similar scenarios, and I spotted no more occurrences of error messages being handled this way, so this PR should be enough.

### How to test your changes?

- Run `dev cd dawn`
- Run `shopify theme dev` and get your development theme id
- Update `~/Library/Preferences/shopify-cli-development-theme-config-nodejs/config.json` by appending `1` to your development theme id there (now, your development theme will no longer exist)
- Run `dev cd dawn`
- Run `shopify theme dev`
- Notice it just creates a new development theme

**Before**:
Screenshot 2025-01-15 at 13 43 39

**After**:
Screenshot 2025-01-15 at 13 44 10

### Post-release steps

N/A

### Measuring impact

### Checklist

### DIFF
--- a/.changeset/lemon-pants-add.md
+++ b/.changeset/lemon-pants-add.md
@@ -1,0 +1,6 @@
+---
+'@shopify/cli-kit': patch
+'@shopify/theme': patch
+---
+
+Fix `shopify theme dev` to no longer fail when development themes expire in internationalized stores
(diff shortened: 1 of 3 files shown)


EvilGenius13 (approved):
Thank you for the quick fix! 🚀

PR #5203, opened 2025-01-15
[Themes] Add debug logging for theme fetch errors

jamesmengo:

Adds debug logging to report that themeFetch failed on top of the changes introduced in #5200

This is likely inconsequential, but may come in handy for debugging something in the future (even more so because the client will handle failures silently otherwise)

@karreiro  **Do we want to do this? Would we be adding too much noise?**


karreiro (approved):
Thank you, @jamesmengo!

I had the same internal debate about this. My initial thought was that the command with the request ID would be enough to provide us some context. However, I don't think this line is harmful :)

EvilGenius13 (approved):
I'm +1 for adding this in. Adding a one liner to debug logs wouldn't be too noisy imo.

PR #5207, opened 2025-01-15
`--no-color` fix

amcaplan:

The user's expressed `--no-color` preference is not respected when logging warnings and errors.

Additional context: Some versions of node color `console.warn` output yellow and `console.error` output red unless they output is already colored explicitly. While [this has been rolled back](https://github.com/nodejs/node/pull/54677), it seems safest to stop relying on these functions and simply write directly to stdout/stderr.

`$ p shopify kitchen-sink static --no-color`

Before:

Screenshot 2025-01-15 at 23 31 49

After:

Screenshot 2025-01-15 at 23 33 46

### How to test your changes?

`$ p shopify kitchen-sink static` with and without `--no-color`

### Measuring impact

### Checklist

### DIFF
--- a/.changeset/smart-carrots-vanish.md
+++ b/.changeset/smart-carrots-vanish.md
@@ -1,0 +1,5 @@
+---
+'@shopify/cli-kit': patch
+---
+
+Ensure no-color mode is respected for warnings and errors


graygilmore (approved):
Other than the eslint errors that need fixing this looks good!

PR #5208, opened 2025-01-15
[Themes] Add error overlay for theme upload failures

jamesmengo:

To improve the developer experience by displaying theme development errors directly in the browser, making it easier for developers to identify and fix issues during theme development.

Adds an error overlay system that displays theme development errors in the browser. When errors occur during file uploads:
- Shows a modal overlay with detailed error messages
- Displays the affected file path and corresponding error messages
- Allows dismissal of the overlay via a close button
- Preserves the original page content behind the overlay

### How to test your changes?

1. Start theme development server
2. Introduce an error in a theme file (e.g., invalid Liquid syntax)
3. Save the file and verify that:
   - An overlay appears in the browser
   - The overlay shows the file path and error message
   - The overlay can be dismissed using the close button
   - The original page content remains visible behind the overlay

### Measuring impact

### Checklist


packages/theme/src/cli/utilities/theme-environment/hot-reload/error-overlay.ts
```diff
@@ -0,0 +1,74 @@
...
+        <div style="${OVERLAY_STYLES.errorItem}">
+          <strong>${fileKey}</strong>
+          ${messages.map((msg) => `<pre style="${OVERLAY_STYLES.errorMessage}">- ${msg}</pre>`).join('')}
+        </div>
+      `,
+    )
+    .join('')
+
+  return `
+      <div id="section-error-overlay" style="${OVERLAY_STYLES.container}">
```
graygilmore, line 58: Would prefer to use something like the `dialog` element so that we don't need to manually roll a bunch of accessibility features: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dialog

packages/theme/src/cli/utilities/theme-environment/hot-reload/error-overlay.ts
```diff
@@ -0,0 +1,74 @@
...
+          >
+            &times;
+          </button>
+        </div>
+      </div>
+    `
+}
+
+export function injectErrorIntoHtml(html: string, errors: Map<string, string[]>): string {
+  return html + getErrorOverlay(errors)
```
graygilmore, line 73: If I had an error and then another error would this end up rendering multiple overlays?

packages/theme/src/cli/utilities/theme-environment/html.test.ts
```diff
@@ -0,0 +1,40 @@
...
+    // Given
+    const html = '<html><head><title>My Store</title></head><body><div>content</div></body></html>'
+    const errors = new Map([['file.css', ['Error']]])
+
+    // When
+    const result = injectErrorIntoHtml(html, errors)
+
+    // Then
+    const overlay = getErrorOverlay(errors)
+    expect(result).toBe(html + overlay)
```
graygilmore, line 36: 🤔 would this imply that our overlay is rendering *after* the closing `</html>` tag? If so let's make sure we're rendering inside of the document.

packages/theme/src/cli/utilities/theme-environment/hot-reload/error-overlay.ts
```diff
@@ -0,0 +1,74 @@
+const OVERLAY_STYLES = {
```
graygilmore, line 1: I have a few opinions on the styling here but let's settle on implementation before tweaking a bunch of this!

graygilmore (commented):
Conceptually I love this. Let's see if we can put a little polish on it!

PR #5214, opened 2025-01-16
Fix speedscope bundling

frandiox:

Fixes the feedback in [this comment](https://github.com/Shopify/cli/pull/5109#discussion_r1918081002) for the `shopify theme profile` command.

Moves `speedscope` from a dependency hard to bundle to the assets system we already have in place for Hydrogen and App.

### How to test your changes?

Run `pnpm bundle`, then `ls -la packages/cli/dist/assets/speedscope` at the root and it should show all the files. Try the profile command to check that it works.


packages/cli/package.json
```diff
@@ -24,7 +24,8 @@
     ".": {
       "import": "./dist/index.js",
       "types": "./dist/index.d.ts"
-    }
+    },
+    "./package.json": "./package.json"
```
frandiox, line 28: Adding this so that we can `require.resolve('@shopify/cli/package.json')` later.
If for some reason this is not desired, we can use [`findPathUp` like we do in Hydrogen](https://github.com/Shopify/hydrogen/blob/6bc6068fbffebbcd07c3d1db97956248bfa6a742/packages/cli/src/lib/build.ts#L37-L62) (we needed that in Hydrogen because the package can be run globally bundled, or locally, so require.resolve wouldn't be enough).

packages/theme/src/cli/utilities/asset-path.ts
```diff
@@ -0,0 +1,9 @@
+import {dirname, joinPath} from '@shopify/cli-kit/node/path'
+import {createRequire} from 'node:module'
+
+const require = createRequire(import.meta.url)
+
+export async function resolveAssetPath(...subpaths: string[]) {
+  const cliRootPath = dirname(require.resolve('@shopify/cli/package.json'))
+  return joinPath(cliRootPath, 'dist', 'assets', ...subpaths)
+}
```
frandiox, line 9: Maybe this could be moved to cli-kit eventually? I see `@shopify/app` uses an [ESBuild plugin](https://github.com/Shopify/cli/blob/e2b829d0ef1f292e10fc415a12cd5567683972df/bin/bundling/esbuild-plugin-graphiql-imports.js#L13) to find the assets but perhaps this is simpler? @isaacroldan

PR #5220, opened 2025-01-16
[Themes] Store and track theme asset upload errors during `theme dev`

jamesmengo:
### WHY are these changes introduced?

This will be used to access data + render an overlay when asset upload errors occur during the `theme dev` command

These errors can come from two places. The `theme uploader`, where we do the initial theme upload when we start the command  (in the background) as well as changes made to your theme files while the development server is running.

Original attempt for context - [link](https://github.com/Shopify/cli/pull/5208)

### WHAT is this pull request doing?

Introduces a new `uploadErrors` Map to track and persist file-specific upload errors in the theme file system. This enables:
- Storage of upload errors at the file system level
- Clearing of errors when files upload successfully
- O(1) lookup to determine whether we have errors / should render an overlay

### How to test your changes?

1. No visual changes. 
2. You can run the tests in the test suite
3. Run `theme dev` in the `javascript debug terminal` - place some breakpoints in `theme-uploader` and `theme-fs`

### Measuring impact

- [x] n/a - this doesn't need measurement, e.g. a linting rule or a bug-fix

### Checklist

- [x] I've considered possible cross-platform impacts (Mac, Linux, Windows)
- [x] I've considered possible [documentation](https://shopify.dev) changes

PR #5221, opened 2025-01-16
[Feature] Render error page when `theme dev` encounters asset upload error

jamesmengo:

Fixes https://github.com/Shopify/develop-advanced-edits/issues/425

Builds on top of: https://github.com/Shopify/cli/pull/5220

Builds on top of

When using the `theme dev` command, some users were seeing discrepancies between the `localhost preview @ 127.0.01` and the remote theme via the Online Store Editor.

This scenario occurs when users are editing their assets and create an invalid change. This change prevents the file from getting uploaded to their remote theme. While this file is not saved to the remote theme, it is still being passed directly to SFR in our localhost preview (we do this to enable a snappy preview experience).

[Error Page Design](https://www.figma.com/design/yn0SBCRlexK7DjDW1VHBqS/Upload-error-page-(Theme-Dev-CLI)?node-id=13-142&p=f&t=gwklLNaRaoY8wCOg-0)

image

[Video](https://share.descript.com/view/M0zGN3HOTkX)

This change will render an error page over the when an asset upload error is encountered (this depends on when the server responds). 

These errors are asynchronous sine they come from the result of an upload operation, and can:
- exist before running `theme dev`
- be added while the `theme dev` server is running

Limitations:
- This will only work for errors that cause the asset update operation to fail. Depending on the code itself, certain changes may actually not prevent you from saving entirely, and would **not** be caught here.
- The validation code lives in core

### How to test your changes?
1)  Run `shopify-dev theme dev` in a valid theme
2) **Test errors added while server is active:** Create errors across multiple files. Note the **limitations** above.
3) Note that your page should refresh and update as you create more errors
4) Resolve the errors. 
5) Confirm that the errors are removed as you fix the files
6) **Test pre-existing errors:** Do step 2, then 1

**Errors added while server is active**

https://github.com/user-attachments/assets/100d6e8c-c909-4385-81b4-e353674bcf9e

**Pre-Existing Errors**

https://github.com/user-attachments/assets/54db7bf7-dc99-44ec-8ca5-c9d27da63f21

### Post-release steps
- Areas for future extension -> we could grab the line number and render the lines before / after, similar to how `vite` handles the error page

### Measuring impact

### Checklist


jamesmengo:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/Shopify/cli/5221?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#5221** <a href="https://app.graphite.dev/github/pr/Shopify/cli/5221?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Shopify/cli/5221?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#5220** <a href="https://app.graphite.dev/github/pr/Shopify/cli/5220?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

edmund-teh:
Great work @jamesmengo! Love this project and the improvements you're making with this error page. 

Sorry— one small pedantic change tho... can we change the icon to `AlertCircleIcon` (https://polaris.shopify.com/icons?icon=AlertCircleIcon)? You're currently using `AlertDiamondIcon`, which is inconsistent with what we're doing in the Code editor:
Error-page-reference-2025-01-20 at 9 47 40 AM

jamesmengo:
> Thanks a lot for this PR, @jamesmengo! Excellent stuff 🔥 It's great how this raises awareness on Liquid errors :)
> 
> I've noticed only one minor issue in the grouping of errors (as you may notice `/sections/collage.liquid` is not bold):
> 
> image
> 
> Here's the content of my files with errors (I'm using dawn on main):
> 
> image
> 
> Thanks again for this PR! Looking forward to seeing this feature with developers :)

packages/theme/src/cli/commands/theme/dev.ts
```diff
@@ -54,6 +54,15 @@ You can run this command only in a directory that matches the [default Shopify t
       options: ['hot-reload', 'full-page', 'off'],
       env: 'SHOPIFY_FLAG_LIVE_RELOAD',
     }),
+    'error-overlay': Flags.string({
+      description: `Controls the visibility of the error overlay when an theme asset upload fails:
+- silent Prevents the error overlay from appearing.
+- default Displays the error overlay.
+      `,
+      options: ['silent', 'default'],
```
frandiox, line 62: Instead of passing values, should we pass `allowNo: true` so that devs can use the flag as `--no-error-overlay` to disable the behavior? Or do we expect to have more options in the future?

packages/theme/src/cli/utilities/theme-environment/html.ts
```diff
@@ -31,6 +32,17 @@ export function getHtmlHandler(theme: Theme, ctx: DevServerContext) {
 
         assertThemeId(response, html, String(theme.id))
 
+        if (ctx.options.errorOverlay !== 'silent' && ctx.localThemeFileSystem.uploadErrors.size > 0) {
+          html = getErrorPage({
```
frandiox, line 36: I was going to say that maybe we should set the status here to 500. However, what we are showing is the summary of the upload errors, not the rendering errors... so perhaps we should not change the status? :confused_parrot:
Ideally we'd just inject a modal overlay on top of the rendered page so that it makes sense to use status 200 and still show errors. But that doesn't need to be done now of course.
Thoughts? cc @karreiro 

-- Thinking more about this, the observed result from the user is an error page so I think we should probably return status 500 regardless of the nature of the error until we move this to a modal 🤔

frandiox (commented):
Looking great 🔥

packages/theme/src/cli/commands/theme/dev.ts
jamesmengo: yeah, the idea is to leave room for expansion moving forward, though there's no clear timeline for that so perhaps something we could call into question

packages/theme/src/cli/utilities/theme-environment/html.ts
jamesmengo: Do you think we should try to expand the accepted types of the catch block below and use that?

frandiox: Actually, what if you do this check before calling `render` and just return the error without reaching SFR? Since it's not a real overlay yet, we don't need the actual storefront document, right?

frandiox (approved):
Great job, thanks!

PR #5243, opened 2025-01-21
[DNM] Custom data placeholder module is not required

shauns:
DNM - exploration

This shows how config modules -- modules specified in the app.toml file -- can be made solely server led. There are few small changes here:

- We no longer check that the app.toml matches the CLI-defined config zod schema precisely. We let unknown fields through
- When parsing using JSON schema, we return the JSON schema-validated data, rather than the source data
- We also remove any additional properties in this result, if `additionalProperties: false` is set. This means that JSON schema that covers a slice of the app.toml can be given the entire config, validate it, and return a data structure representing just thie piece that it validated. This is the same behaviour the internal zod schemas for app config follow.
- We remove the hard-coded custom data schema and use the server provided specification. We have to hack in that this module uses single-instance management

One side-effect of this is that completely unknown sections in the app.toml are ignored but don't cause validation to fail. This is a change. This may be desirable or not... if the reason a section is ignored is because it's called `[scopes]` then the developer probably meant `[access_scopes]` and so maybe silence isn't good. Alternatively it may be the right thing to validate things we known about and ignore things we don't.

This is tested in dev and deploy but not linking yet. We'd want to test this thoroughly I think as its a somewhat low level shift.

### DIFF
--- a/packages/app/src/cli/models/app/loader.ts
+++ b/packages/app/src/cli/models/app/loader.ts
@@ -31,7 +31,7 @@ import {WebhookSubscriptionSpecIdentifier} from '../extensions/specifications/ap
 import {WebhooksSchema} from '../extensions/specifications/app_config_webhook_schemas/webhooks_schema.js'
 import {loadLocalExtensionsSpecifications} from '../extensions/load-specifications.js'
 import {UIExtensionSchemaType} from '../extensions/specifications/ui_extension.js'
-import {deepStrict, zod} from '@shopify/cli-kit/node/schema'
+import {zod} from '@shopify/cli-kit/node/schema'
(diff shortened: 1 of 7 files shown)


packages/app/src/cli/models/app/loader.ts
```diff
@@ -905,17 +905,12 @@ async function loadAppConfigurationFromState<
           ...configState.basicConfiguration,
         }
         delete file.path
-        const appVersionedSchema = getAppVersionedSchema(specifications)
+        const appVersionedSchema = getAppVersionedSchema(specifications, true)
```
isaacroldan, line 908: Should `allowDynamicallySpecifiedConfigs` be the default for this function? 🤔 
I see we call `getAppVersionedSchema` in three other places and it defaults to false

Issue #4855, opened 2024-11-14
[Bug]: Pulling only (-o) specific folder doesn't work as expected


aleksandar89d:
### Please confirm that you have:

- [x] Searched [existing issues](.) to see if your issue is a duplicate. (If you’ve found a duplicate issue, feel free to add additional information in a comment on it.)
- [x] Reproduced the issue in [the latest CLI version](https://www.npmjs.com/package/@shopify/cli).

### In which of these areas are you experiencing a problem?

Theme

### Expected behavior

When I am using command shopify theme pull -o templates/, it should pull just templates folder and all files inside it.
This works on version 3.59.0.
This is from package json;
    "pull-hr": "shopify theme pull -o templates/ sections/*.json config/settings_data.json 
It should pull whole templates folder, json files from sections folder and settings_data.json file. And it works on 3.59.0 Cli version.

### Actual behavior

On version 3.70.0 it doesn't work as expected.
When I try to pull just specific theme folder, it doens't work.
Localy I don't have templates/ folder and it should create it, everythiing worked fine on earlier versions.
It shows that files are pulled, but folder is not pulled actually.

### Verbose output

```shell
There is no error, but files simply are not pulled.
```

### Reproduction steps

1.Try to pull just one folder, example shopify theme pull -o templates/

### Operating System

Mac OS 14.1

### Shopify CLI version (check your project's `package.json` if you're not sure)

3.70.0

### Shell

zsh

### Node version (run `node -v` if you're not sure)

v21.6.2

lucyxiang:
Thank you for raising this @aleksandar89d. I was able to replicate this issue, I'll put this issue in our backlog.

lukecharle:
Not sure if what I've posted is also related #5146 

Is there any news on this being fixed? Its a massive blocker and means I can't use the CLI in my CI/CD pipeline. How else am I supposed to manage themes?

PR #5251, opened 2025-01-21
Fixed directory only pattern when using --only flag

EvilGenius13:
This commit fixes an issue in the matchGlob to match directories when filtering through files

Closes: https://github.com/Shopify/cli/issues/4855

There was a bug in which passing a directory to the `only` flag would not work. 
It turns out it wasn't working with the `ignore` flag either.

Added some additional logic to the `matchGlob` method which looks for certain patterns. Now it also checks for folder structure patterns like `assets/`

Added tests for both `only` and `ignore` flags.

## Example

Before:

https://github.com/user-attachments/assets/6cf86a7f-d316-4dee-a827-e798d3c72250

After:

https://github.com/user-attachments/assets/fd26cc0f-311d-4318-bcf0-f5ec4dd33362

### How to test your changes?

Pull down the branch
Build the branch
If in a directory with a theme:
     - delete two different folders like `assets` and `templates`
     - run `shopify theme pull -t <your_thene> -o templates/`
     - You should only see the templates folder be synced
If in a blank directory:
     - run `shopify theme pull -t <your_theme> -o templates/`
     - You should only see the templates folder in your directory

### Measuring impact

### Checklist


jamesmengo:
🤔 
Thanks for this change! I may be bike-shedding here, but I want to raise some thoughts I have about this approach.

The issue here arises because we don't have a strong `glob` contract. 
- We're now using `minimatch` as our matching library after the TS rewrite
- For backwards compatibility reasons, we try to silently fix / replicate Ruby `glob` behaviour as a fallback [2]

My personal preference here is to diverge as little as possible from the JavaScript glob matching library moving forward.

I think we could render a warning to get the user to fix/remediate this issue rather than adding more custom behaviour. 
ALTERNATIVELY, we can raise the warning / suggestion, but still fix it for them if needed by following the pattern laid out in [2].

As always, open to discussion and changing my mind!

**[2]** 
- That's why we automatically substitute `**/*` [here](https://github.com/Shopify/cli/blob/7e34195c3068e91ee368e68d8c8c539e324c26c9/packages/theme/src/cli/utilities/asset-ignore.ts#L90-L95), since users already have `globs` that they're relying on.

We **needed** to do so because the `Ruby` library that we were previously using behaves differently than the `javascript` glob matcher 🤢

karreiro (commented):
Thank you for this PR, @EvilGenius13!

I resonate with the motivation behind this work. The `--only`/`--ignore` flags have an unusual behavior to maintain backward compatibility with the previous Ruby version.

When the Ruby CLI was migrated to the TypeScript CLI, many developers were affected by changes in this module, that was shape to avoid as much friction as possible.

Considering the change in behavior here, I would suggest waiting for a major CLI update to proceed with this approach, finally break backward-compatibility, and establish a more consistent ignore mechanism.

With that context in mind, I believe we could follow the approach suggested by @jamesmengo on this PR.

To avoid overwhelming users with warning messages, we could check the shape of the flag (as you are already doing) and the files being ignored (or selected by `--only`). If an `--ignore` flag is not effectively ignoring anything, and the shape indicates the scenario this PR addresses, we have an opportunity to warn developers.

In the warning, we could encourage users to use regexes (instead of globs) for better results, so the shape of the directory doesn't impact the pattern.

I understand this PR implements an improved behavior, but considering the potential impact, I believe this is the best path to prevent friction in a minor release.

Thanks again for this PR!

packages/theme/src/cli/utilities/asset-ignore.ts
```diff
@@ -83,9 +85,18 @@ function matchGlob(key: string, pattern: string) {
...
+  if (originalMatchGlob(key, pattern, matchOpts)) return true
 
-  if (result) return true
+  if (isRegex(pattern)) return regexMatch(key, asRegex(pattern))
+
+  if (!pattern.includes('*') && pattern.endsWith('/') && !warnedPatterns.has(pattern)) {
+    warnedPatterns.add(pattern)
+    renderWarning({
+      headline: 'Directory pattern may be misleading.',
+      body: `For more reliable matching, consider using ${pattern}* or ${pattern}*.filename instead.`,
```
karreiro, line 96: Users may have a more deterministic behavior when using strings, so I believe we could encourage them in that direction. Also, I think we mean `<file_extension>` since the `*` would represent the file name. Am I missing something?

```suggestion
      body: `For more reliable matching, consider using "${pattern}*" or "${pattern}*.<file_extension>" instead.`,
```

karreiro (commented):
Thank you for the changes, @EvilGenius13! I've left just one minor extra comment :)

packages/theme/src/cli/utilities/asset-ignore.ts
EvilGenius13: You are absolutely right. That should have been referencing `file_extension`. Thank you!

karreiro (approved):
Thank you, @EvilGenius13!

PR #5256, opened 2025-01-22
Allow `@shopify/shopify_function` version 0.x.x

andrewhassan:

This PR is a follow-up to https://github.com/Shopify/cli/pull/5153. Upon releasing that PR, we realized that some users on the latest CLI have functions using `@shopify/shopify_function` version 0.x.x, which should still work.

This PR:
1. Allows for multiple compatible major versions of `@shopify/shopify_function`
2. Doesn't wrap `AbortError`s if the function build fails. Prior to this PR, all errors were wrapped, which meant that some context was lost.

image

### How to test your changes?

- Run `app function build` for a function with version 1.x.y of `@shopify/shopify_function`.
- Run `app function build` for a function with version 0.x.y of `@shopify/shopify_function`.
- Run `app function build` for a function with version 2.x.y of `@shopify/shopify_function` (just change the version number in the `package.json` for `@shopify/shopify_function`). This should fail with the error screenshotted above.

### Measuring impact

### Checklist


gordonhirsch (approved):
Reviewing as App Inner Loop ATC. LGTM from an inner loop perspective 👍

lopert (approved):
👍 LGTM.

Confused myself during the tophatting trying to keep the versions straight.

- package.json from the function's dir is the one that `npm install` will get.
  - however, you can't install a version outside of our current "ALLOWED" set.
- package.json from the node_modules dir is the one that we're checking to ensure is valid.
  - to trigger the error, this is the one we have to edit.
  ```
  extensions/order-discount/node_modules/@shopify/shopify_function/package.json
  ```

PR #5262, opened 2025-01-23
Replace rainbow bar with moving hill for no-color mode

amcaplan:

Fixes the poor display of the loading bar in no-color terminals.

image

https://github.com/user-attachments/assets/06e49f94-7982-4e32-b3bd-67a3275ecf78

### How to test your changes?

```sh
$ p shopify kitchen-sink async
```

or any other command which uses a rainbow bar (e.g. creating a new app)

### Measuring impact

- [x] Impact is unknown, a followup PR will includes analytics changes to measure impact

### Checklist

### DIFF
--- a/packages/cli-kit/src/private/node/ui/components/Tasks.tsx
+++ b/packages/cli-kit/src/private/node/ui/components/Tasks.tsx
@@ -81,14 +81,9 @@ function Tasks<TContext>({
   const {twoThirds} = useLayout()
   let loadingBar = new Array(twoThirds).fill(loadingBarChar).join('')
   if (noColor ?? !shouldDisplayColors()) {
-    // fill loading bar with the no color chars, repeating until the length is two thirds but only using a whole number repeat of the no color chars
-    const repeatCount = Math.floor(twoThirds / gradualHillString.length)
-    if (repeatCount === 0) {
-      const shortRepeatCount = Math.floor(twoThirds / hillString.length)
-      loadingBar = hillString.repeat(Math.max(1, shortRepeatCount))
-    } else {
-      loadingBar = gradualHillString.repeat(repeatCount)
-    }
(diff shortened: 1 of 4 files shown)


packages/cli-kit/src/private/node/ui/components/Tasks.tsx
```diff
@@ -25,6 +27,7 @@ interface TasksProps<TContext> {
   silent?: boolean
   onComplete?: (ctx: TContext) => void
   abortSignal?: AbortSignal
+  noColor?: boolean
```
amcaplan, line 30: I only introduced this to inject `true` in tests because mocking `output.js` (where we get `shouldDisplayColors`) led to very weird results. It's a private interface, so I don't think this should matter all that much.

packages/cli-kit/src/private/node/ui/components/Tasks.tsx
```diff
@@ -65,9 +68,13 @@ function Tasks<TContext>({
   silent = isUnitTest(),
   onComplete = noop,
   abortSignal,
+  noColor,
 }: React.PropsWithChildren<TasksProps<TContext>>) {
   const {twoThirds} = useLayout()
-  const loadingBar = new Array(twoThirds).fill(loadingBarChar).join('')
+  let loadingBar = new Array(twoThirds).fill(loadingBarChar).join('')
+  if (noColor ?? !shouldDisplayColors()) {
+    loadingBar = hillString.repeat(Math.ceil(twoThirds / hillString.length))
```
amcaplan, line 76: Rounding up here so we have a bit of extra insurance at the end (which `TextAnimation` will truncate), rather than overtruncating immediately.

karreiro (approved):
Thank you, @amcaplan! Very neat accessibility enhancement! LGTM and works as expected as well:

tophat-2

PR #5277, opened 2025-01-27
Tidy up documentation scripts

gonzaloriestra:

We have too many similar scripts about documentation, all for different purposes, and it's really confusing:
- `docs:generate` => API documentation
- `build-docs` => shopify.dev static documentation
- `update-docs` => shopify.dev commands + static documentation
- `refresh-documentation` => autogenerated code documentation for ui.tsx

Rename them so we have:
- `build-api-docs` => API documentation
- `build-dev-docs` => shopify.dev documentation
- `refresh-code-documentation` => autogenerated code documentation for ui.tsx

### How to test your changes?

Run them

### Measuring impact

### Checklist


isaacroldan (approved):

gonzaloriestra:
### Merge activity

* **Jan 27, 7:47 AM EST**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/Shopify/cli/5277).

amcaplan:
I removed it from the queue because this will break the GitHub pages job, which uses `docs:generate` [here](https://github.com/Shopify/cli/blob/main/.github/workflows/docs.yml#L38) and needs to be updated as well. No test would catch this.

amcaplan:
Hmmmm, I am testing locally and @gonzaloriestra you're correct that `nx` is using `project.json` rather than `package.json`. But we still need the full command for the sake of `--output-style=stream` which is only appropriate for CI.

Also, I think that if we're renaming, we should be consistent throughout. So if we're calling this `build-api-docs` then it should be called that in `cli-kit`, across the `project.json` files, and in the GitHub action as well.

gonzaloriestra:
I sorted it in a separate commit, so you can check the rest. But yeah, GH and Graphite could handle that much better...

PR #5278, opened 2025-01-27
Transform dev session status into an EventEmitter class

isaacroldan:

The current dev session status management is using global variables. This leads to inefficient polling mechanisms and potential race conditions. 

This will enable more features like custom status messages from the dev-session.

Introduces a new `DevSessionStatusManager` class that extends EventEmitter to manage dev session status in a centralized way. This replaces global variables with a proper event-driven system that:
- Maintains session state (isReady and previewURL)
- Emits events when status changes
- Provides methods to update and reset the status
- Prevents unnecessary updates when status hasn't changed

### How to test your changes?

1. Start a dev-session without extensions. Verify that the shortcuts won't appear until the dev-session is ready.
2. Add a new admin-action extension and verify the URL is updated.
3. Delete the admin-action extension and verify the URL is updated.

### Checklist

### Measuring impact

### DIFF
--- a/packages/app/src/cli/services/dev/ui/components/Dev.tsx
+++ b/packages/app/src/cli/services/dev/ui/components/Dev.tsx
@@ -66,27 +66,19 @@ const Dev: FunctionComponent<DevProps> = ({
   const {isRawModeSupported: canUseShortcuts} = useStdin()
   const pollingInterval = useRef<NodeJS.Timeout>()
   const localhostGraphiqlUrl = `http://localhost:${graphiqlPort}/graphiql`
-  const defaultPreviewURL = previewUrl
-  const defaultStatusMessage = `Preview URL: ${defaultPreviewURL}${
-    graphiqlUrl ? `\nGraphiQL URL: ${localhostGraphiqlUrl}` : ''
-  }`
-  const defaultDevSessionStatus = devSessionStatusManager.status
-  const [statusMessage, setStatusMessage] = useState(defaultStatusMessage)
+  const graphiqlURLMessage = graphiqlUrl ? `GraphiQL URL: ${localhostGraphiqlUrl}` : ''
+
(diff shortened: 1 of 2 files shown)


packages/app/src/cli/services/dev/processes/dev-session-status-manager.ts
```diff
@@ -0,0 +1,36 @@
+import {deepCompare} from '@shopify/cli-kit/common/object'
```
shauns, line 1: Coverage looks to have dropped, are the tests properly covering this?

packages/app/src/cli/services/dev/ui/components/Dev.tsx
```diff
@@ -1,7 +1,7 @@
 import metadata from '../../../../metadata.js'
 import {DeveloperPlatformClient} from '../../../../utilities/developer-platform-client.js'
 import {ExtensionInstance} from '../../../../models/extensions/extension-instance.js'
-import {devSessionStatus} from '../../processes/dev-session.js'
+import {devSessionStatusManager, DevSessionStatus} from '../../processes/dev-session-status-manager.js'
```
shauns, line 4: I feel like `devSessionStatusManager` should be a prop for the component, then we can get rid of the global state? If we create it we can pass it into `dev-session` / `Dev`.

isaacroldan: Yeah, i think that makes sense, I was going to do that as part of this PR, but it would imply MANY  changes (specially in tests) to pass it down to both processes.

Is it ok if I do that in a new PR in the stack? I think it'd be cleaner

packages/app/src/cli/services/dev/processes/dev-session-status-manager.ts
isaacroldan: Totally forgot to add tests for the manager 🤦

PR #5281, opened 2025-01-27
Only rebuild docs when changed

amcaplan:

No need to rebuild on each push to `main`, only when we release a version.

The docs only should change when:

* `cli-kit` gets a version bump
* Anything in `docs` is directly updated

This limits the job runs to when those things happen.

### How to test your changes?

Gotta merge and observe

### Measuring impact

### Checklist


.github/workflows/docs.yml
```diff
@@ -3,6 +3,10 @@ name: Build and Deploy GitHub Pages Site
 on:
   push:
     branches: ["main"]
+    paths:
+      - packages/cli-kit/package.json
+      - docs/**
```
dmerand, line 8: How picky is Github workflow about path syntax being? The docs (which are [very similar to this case](https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#example-including-paths) show the paths as being quoted, but I know in other systems YAML quotes don't matter much.

.github/workflows/docs.yml
```diff
@@ -3,6 +3,10 @@ name: Build and Deploy GitHub Pages Site
 on:
   push:
     branches: ["main"]
+    paths:
+      - packages/cli-kit/package.json
```
dmerand, line 7: Technically any change to this package will trigger docs, not just a version bump but that's fine IMO.

dmerand (approved):

.github/workflows/docs.yml
amcaplan: In the linked example, the asterisk was at the beginning of the string. Asterisks have specific meaning in that case in YAML, so it needed to be quoted. I don't think this case technically needs it, but it doesn't hurt so I've added it.

.github/workflows/docs.yml
amcaplan: Yeah, the goal isn't to limit how often we're publishing, but just to skip any extra unnecessary rebuild jobs (reducing our impact on resource capacity for the repo/org). This is the right granularity to avoid 90% or more of unnecessary builds. But yes, it will rebuild when there are updates to `cli-kit` dependencies as well.

PR #5288, opened 2025-01-28
Revert "Add provenance"

amcaplan:
Reverts Shopify/cli#5285

[It doesn't work.](https://shipit.shopify.io/shopify/cli/nightly/deploys/2928667)

Apparently [we need to switch deploys to use GH Actions](https://docs.npmjs.com/generating-provenance-statements#provenance-limitations) instead of Shipit for automatic provenance generation to succeed.

So this approach won't work, we actually need to switch from Shipit to GH Actions in the next couple of months.


gonzaloriestra (approved):
Oops

amcaplan:
### Merge activity

* **Jan 28, 4:48 AM EST**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/Shopify/cli/5288).

PR #5293, opened 2025-01-28
ensure suggested app names are valid

craigmichaelmartin:

We currently have an app name length limit of 30 characters, but can suggest names longer than 30 characters.

Retries if the generated name is too long.

### Measuring impact

### Checklist


packages/cli-kit/src/public/node/fs.ts
```diff
@@ -479,14 +484,15 @@ interface GenerateRandomDirectoryOptions {
  */
 export async function generateRandomNameForSubdirectory(options: GenerateRandomDirectoryOptions): Promise<string> {
   const generated = `${getRandomName(options.family ?? 'business')}-${options.suffix}`
+
```
dmerand, line 487: Do we want to ensure that we don't end up in loops, perhaps with a fallback name after X attempts? There's recursion here with no base case other than "we will likely generate a name that works".

packages/app/src/cli/models/app/validation/common.ts
```diff
@@ -20,3 +20,9 @@ function isValidUrl(input: string, httpsOnly: boolean) {
 export function ensurePathStartsWithSlash(arg: unknown) {
   return typeof arg === 'string' && !arg.startsWith('/') ? `/${arg}` : arg
 }
+
+export const APP_NAME_MAX_LENGTH = 30
```
dmerand, line 24: Is this value shared with other constraints in the CLI? Or is it a convention we're inheriting from core?

packages/cli-kit/src/public/node/fs.ts
craigmichaelmartin: Yeah, everything about this PR is off-putting to me (I think I'm allowed to say that because it is my PR lol).

I thought about the case you are describing, but didn't want to introduce too much complexity for this edge case of convenience behavior, which would requiring adding a param for counting recursion depth, and some fallback behavior (probably couldn't be to just truncate in case the filepath for that name is already taken, so a secondary fall back to an empty string), or else having some other heavy implementation like eager avoidance of invalid names.

However, taking a step back, do you think this is the right direction? Here is roughly the landscape I see:

1. Rather than the approach of this PR to gracefully handle too-long names, we preclude their possibility: remove the one long noun (`infrastructure`) that can alone can cause the name to be greater than 30 characters, extract the max length of an app name (as done in this PR), and add a test which checks that every permutation of the name seeds is not longer than the max length. The resolves the problem this PR sets out to address, and guards against its return, with no runtime overhead. However, it introduces over a thousand new assertions and the maintenance costs of these data-validation tests; further if we wanted to swap out this set with some other 1st/3rd party seeds which are larger, we'd have to roll this back and go with option 2).
2. Keep this PR, and add a base fallback for after X attempts. However this introduces hotpath costs and code complexity to accommodate a scenario which could be prevented ahead of time, and just isn't that bad. 

What do you think?

packages/app/src/cli/models/app/validation/common.ts
craigmichaelmartin: This value had never been extracted (as far as I could tell). I extracted it to get the uses I found of this validation downstream from this single source of truth, rather than as disparate magic numbers (in `dev.ts` and wherever the app name length validation happens).

packages/cli-kit/src/public/node/fs.ts
```diff
@@ -468,8 +468,16 @@
 
   /** Type of word to use for random name. */
   family?: RandomNameFamily
+
+  /** Used to to track how many attempts have been made to generate the name. */
+  attempts?: number
+
+  /** Predicate for validating the generated name. */
+  isValidName?: (name: string) => boolean
```
isaacroldan, line 476: i'm not sure about this, the function should be simple, just generate a name following some options (suffix, family and the directory).

It should be responsibility of the caller to validate the proposed name and decide whether or not to try again (and how many times).

`attemps` should definitely not be part of the options if you want to use it to track the recursion, because it can be set externally and any caller will see this and interpret it as "max attemps" and set a number.

My proposal: 
- keep `generateRandomNameForSubdirectory` as it is now
- Make the app init validate the result name and retry again if needed
- Remove the extra long word, just to be sure.

What do you think?

craigmichaelmartin: Sounds good to me! Thanks, Isaac!

packages/app/src/cli/models/app/validation/common.ts
isaacroldan: Is this value real? the HANDLE limit is 50, I don't see why the app name limit should be less than that 🤔 
Are we validating this server side with the same value? Maybe this extraction is a good opportunity to update this number!

isaacroldan (approved):

packages/app/src/cli/models/app/validation/common.ts
craigmichaelmartin: As you noted in slack, the developer dashboard (and partners) enforces an app name limit of 30 as well.

packages/cli-kit/src/public/node/fs.ts
craigmichaelmartin: > You wouldn't need thousands of assertions

Yeah, you are right! I'm not sure why I was thinking otherwise. I merged this PR with the Current Implementation™️ which solves the user problem of our recommending an invalid app name. I'm very open to solving this problem with the other implementation (pre-verify the validity of all seed permutations by using the longest options in each group, and thereby remove the hotpath logic added here). The tradeoff would be that the solution only works for a deterministic set of seed words where we can ascertain (or specify) the longest word length, and so boxes us in if we want to swap out of homegrown seeds with a faker library. I'll leave it up to you and those with more context on the CLI (@shauns @isaacroldan) to evaluate which makes more sense, and am happy to do the work if decide to swap implementations.

PR #5306, opened 2025-01-29
Upgrade nanoid to 3.3.8

gonzaloriestra:

Fixes https://github.com/Shopify/cli/security/dependabot/126

Pins nanoid to 3.3.8

### How to test your changes?

CI

### Measuring impact

### Checklist

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -12,6 +12,7 @@ overrides:
   supports-hyperlinks: 3.1.0
   '@shopify/cli-hydrogen>@shopify/cli-kit': link:./packages/cli-kit
   '@shopify/cli-hydrogen>@shopify/plugin-cloudflare': link:./packages/plugin-cloudflare
+  nanoid: 3.3.8
 
 importers:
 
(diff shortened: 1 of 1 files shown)


isaacroldan:
Is this safe? `nanoid` seems to be a transitive dependency, if we force the resolution to a specific version we could break the dependency that actually uses it no?

gonzaloriestra:
```
❯ pnpm why nanoid
devDependencies:
@shopify/eslint-plugin-cli file:packages/eslint-plugin-cli(eslint@8.57.0)(prettier@2.8.8)(typescript@5.7.2)(vitest@1.6.0)
└─┬ eslint-plugin-vitest 0.5.4
  └─┬ vitest 1.6.0 peer
    ├─┬ vite 5.4.12
    │ └─┬ postcss 8.4.49
    │   └── nanoid 3.3.7
    └─┬ vite-node 1.6.0
      └─┬ vite 5.4.12
        └─┬ postcss 8.4.49
          └── nanoid 3.3.7
vitest 1.6.0
├─┬ vite 5.4.12
│ └─┬ postcss 8.4.49
│   └── nanoid 3.3.7
└─┬ vite-node 1.6.0
  └─┬ vite 5.4.12
    └─┬ postcss 8.4.49
      └── nanoid 3.3.7
```

The `nanoid` dependency comes from `postcss@8.4.49`, [which requires ^3.3.7](https://github.com/postcss/postcss/blob/aed8b899884db4bc89f60782771eecd8c7861b32/package.json#L88). So it should be compatible with 3.3.8.

They started requiring `^3.3.8` on `postcss@8.5`, but the [latest version of vite is still requiring 8.4](https://github.com/DannyAndres/framework/blob/c695f71ed3b8443036e8918e690b01e9972488f2/packages/vite/package.json#L43).

So the only way I can think of to update `nanoid` is to add the resolution from this PR, or alternatively a resolution to force `postcss` to `^8.5.0`. Any other idea?

isaacroldan:
Umm, if this is all because of vitest, we should definitely upgrade it, just checked and they are on version 3.0.2 now 😬

gonzaloriestra:
I'm doing that it in https://github.com/Shopify/cli/pull/5317, but that doesn't update `nanoid`.

isaacroldan (approved):

PR #5309, opened 2025-01-29
Add flag to allow —unpublished to reuse theme names

### LINKED ISSUE #2699: [Feature]: Add flag to make `theme push` idempotent

Theme

New feature

### Overview

Would it be possible to add a flag that makes `theme push` idempotent? 

Given a theme name, it would only create a single theme with that name. 

Right now, if the `--unpublished` flag is passed, it will create multiple themes with the same name. Without `--unpublished`, it will error if the theme isn't created.

What I'd like is the ability to run `theme push --theme jeff/feature-branch` multiple times and only have a single theme created.

### Motivation

To support PR reopening in a GitHub, we need the ability to push to a theme named after a branch multiple times and have only a single theme created.

In the pre-1.0 release of [Shopkeeper](https://github.com/TheBeyondGroup/shopkeeper/blob/6ec79c2df6500d7408fd1cb46c6a7c4d32d91e0e/src/cli/shopkeeper-theme-create.ts#L21-L30) I had to add a create command in order to make this possible.

I [walk through](https://github.com/TheBeyondGroup/shopkeeper/blob/jeff/add-ci-cd-to-walkthrough/docs/the_complete_guide.md#cicd-with-github-actions) how someone can use GitHub Actions to write workflows to provide a CI/CD experience for Shopify like you might have with other tools like Heroku or Vercel's branch previews.

jeffreyguenther:

Fixes #2699

```shell
shopify theme push —theme <name> --unpublished
```

Repeated calls the above command generates multiple themes with identical names. If you're using theme names to map to git branches in CI, this causes a problem. You need an idempotent way to push themes.

We want something like:

```shell
shopify theme push --theme <name> —-unpublished --upsert
```

In this PR, I add a flag to optionally allow to `--unpublished` to update a theme if there is already one with the passed name. The flag is dependent on `--unpublished` so we should get the nice OCLIF errors if it's passed without `--unpublished`

Here's my rationale:

This command pushes to a theme named "Iteration", but the theme with that name must exist. If it doesn't, an error is output.
```shell
shopify theme push --theme "Iteration"
```

So we add the `--unpublished` flag, to allow us to create a new, unpublished theme with the passed name.

```shell
shopify theme push --theme "Iteration" --unpublished
```

Then, we want to provide more information to this push to reuse the theme with the matching name if it exists.

```shell
shopify theme push --theme "Iteration" --unpublished --upsert
```

### Measuring impact

### Checklist

### DIFF
--- a/packages/theme/src/cli/commands/theme/push.ts
+++ b/packages/theme/src/cli/commands/theme/push.ts
@@ -64,6 +64,17 @@ export default class Push extends ThemeCommand {
       description: 'Create a new unpublished theme and push to it.',
       env: 'SHOPIFY_FLAG_UNPUBLISHED',
     }),
+    'unique-name': Flags.boolean({
+      description: 'Make the push to an unpublished theme reuse the name',
+      env: 'SHOPIFY_FLAG_UNIQUE_NAME',
+      default: false,
+      relationships: [
+        {
+          type: 'all',
+          flags: ['unpublished'],
+        },
+      ],
+    }),
     nodelete: Flags.boolean({
       char: 'n',
       description: `Prevent deleting remote files that don't exist locally.`,
@@ -114,6 +125,7 @@ export default class Push extends ThemeCommand {
       development: flags.development,
       live: flags.live,
       unpublished: flags.unpublished,
+      uniqueName: flags['unique-name'],
       nodelete: flags.nodelete,
       only: flags.only,
       ignore: flags.ignore,
--- a/packages/cli/README.md
+++ b/packages/cli/README.md
@@ -2208,6 +2208,7 @@ FLAGS
       --password=<value>     Password generated from the Theme Access app.
       --path=<value>         The path to your theme directory.
       --strict               Require theme check to pass without errors before pushing. Warnings are allowed.
+      --unique-name          Make the push to an unpublished theme
       --verbose              Increase the verbosity of the output.
 
 DESCRIPTION
--- a/packages/cli/oclif.manifest.json
+++ b/packages/cli/oclif.manifest.json
@@ -6281,6 +6281,21 @@
           "name": "theme",
           "type": "option"
         },
+        "unique-name": {
+          "allowNo": false,
+          "description": "Make the push to an unpublished theme",
+          "env": "SHOPIFY_FLAG_UNIQUENAME",
+          "name": "unique-name",
+          "relationships": [
+            {
+              "flags": [
+                "unpublished"
+              ],
+              "type": "all"
+            }
+          ],
+          "type": "boolean"
+        },
         "unpublished": {
           "allowNo": false,
           "char": "u",


jeffreyguenther:
Been delayed by client work, but this will finished one day.

jeffreyguenther:
@jamesmengo for your consideration.

jeffreyguenther:
@mxstbr, @karreiro or @graygilmore

jamesmengo:
Hey @jeffreyguenther!
Sorry for the late response - I'm switching teams internally so will no longer be working in this area, but I'll raise this directly to the team to make sure this gets eyes.

Thanks for contributing and for all the work you do for our tooling!

aswamy:
Hey @jeffreyguenther thanks for PR. If you wanted to push to the same theme every time, wouldn't you just want to use the theme's id instead?

```
shopify theme push --theme <id>
```

The Con of this approach is obviously it wouldn't work on the first call. i.e. you'd have to make the first theme manually, then the subsequent calls would work.

I have a few concerns with the approach in this PR, but hoping to get some light on it:
- `--unpublished` by design is known to create a theme so not sure if reusing this flag is the best idea
- if you do `shopify theme push --theme "production" --unpublished --upsert` and the theme was a production theme, would it "unpublish" the production theme?
  - Or we just say `shopify theme push --theme "production" --upsert` will always update the theme if it exists, OR create an unpublished theme by default (no `--unpublished` needed 🤷)

jeffreyguenther:
My use case for this feature is in CI to generate preview themes. Ideally, you can run an idempotent create command in the GitHub workflow so that the workflow doesn't need to care whether the theme exists. This is why I went with an "upsert" approach.

I kept `--unpublished` in because it helps communicate that the push results in a new unpublished theme. I think it can be removed if we can communicate the semantics a different way.

With regard to pushing to a live theme, I think the only way to push to the live theme should be with the flag `--allow-live`. `shopify theme push --theme "production" --unpublished --upsert` should return an error saying that the theme you've named is the live theme.

I think I may need to collect the different potential combinations of the flags on `push` as there are some funky dependencies and implied precedence in the code. Once we can see how the flags interact, I think the solution will be more obvious.

aswamy:
@jeffreyguenther I have some ideas on how we can possibly improve the current solution. Let me know what you think.

Instead of having another flag that's dependent on `--unpublished` flag, we could just make `--unpublished` a string arg instead of a boolean arg.

i.e.
- `shopify theme push --theme "test theme" --unpublished` will implicitly have "create" as a value for `--unpublished`
  - `--unpublished "create"`: create a new unpublished theme no matter what
  - `--unpublished`: has "create" as default value (same as above)
  - `--unpublished "upsert"`: creates a new unpublished theme if it doesn't exist
    - if a live theme exists with the same name, create a new one instead

This solution wouldn't break for existing CLI users since `shopify theme push --theme "test theme" --unpublished` is still a valid command and does the same operation as previous versions.

jeffreyguenther:
I like this approach. Yesterday, I went through the possible ways `--unpublished` can be used and was leaning towards something like this with a new flag like `--upsert`.

I think switching the flag from a boolean to a string is the likely the cleanest way to do this. It lets us configure `--unpublished`'s behaviour.

> --unpublished "upsert": creates a new unpublished theme if it doesn't exist
> if a live theme exists with the same name, create a new one instead

What do you think about throwing an error in this case? I think creating the additional theme with the same name breaks the upsert semantics. I would expect that we either create a new theme because it doesn't exist, or we'd update the existing one.

aswamy:
> What do you think about throwing an error in this case? I think creating the additional theme with the same name breaks the upsert semantics. I would expect that we either create a new theme because it doesn't exist, or we'd update the existing one.

Since this is explicitly saying to upsert an _unpublished_ theme, i dont think it breaks it. I'd read the following command as, "if you dont have an unpublished theme with this name, create one":
```
shopify theme push --theme "my theme" --unpublished "upsert"
```

However, I WOULD throw an error if you pass `--unpublished "upsert"` when you pass a theme ID instead of theme name since the behaviour of `--theme <id>` is always a "update"
i.e. 
```
shopify theme push --theme <id> --unpublished "upsert"
```

jeffreyguenther:
Makes sense to me. Thanks for collaboration! I'll get to work on updating the PR.

jeffreyguenther:
@aswamy Ready for review. I think I've followed Shopify's style. Standing by for feedback!

In the meantime, I'll do some manual testing.

graygilmore:
I'm not sure if running `/snapit` will work on a forked repo PR but it'll make testing a little bit easier for folks so I'll try!

graygilmore:
/snapit

```
Error: `/snapit` is not supported on pull requests from forked repositories.
```

packages/theme/src/cli/commands/theme/push.ts
```diff
@@ -59,9 +59,13 @@ export default class Push extends ThemeCommand {
       description: 'Push theme files from your remote live theme.',
       env: 'SHOPIFY_FLAG_LIVE',
     }),
-    unpublished: Flags.boolean({
+    unpublished: Flags.string({
       char: 'u',
-      description: 'Create a new unpublished theme and push to it.',
+      summary: 'Create a new unpublished theme and push to it.',
```
aswamy, line 64: update the summary as well

packages/theme/src/cli/commands/theme/push.ts
```diff
@@ -59,9 +59,13 @@ export default class Push extends ThemeCommand {
...
+      description:
+        'Pass the value "create" to always create a new theme. Pass the value "upsert" to update an existing if it exists. A new theme will be created if no theme by that name exists or the name is same as the live theme.',
+      options: Object.values(UnpublishedOption),
+      default: UnpublishedOption.Create,
```
aswamy, line 68: Doesn't this ensure `unpublished` flag is ALWAYS set - so it's almost as if we always passed in the `unpublished` flag? This changes the original functionality. See the image below where i compare the existing one (top) vs yours (bottom):

image

jeffreyguenther:
Is there any interest in me finishing this work?

aswamy (commented):
Sorry for the delay @jeffreyguenther. I was on vacation and cleared my plate on PR reviews once i returned.

1 concern, but the rest looks good. I would recommend you generate a changeset file using `pnpm changeset` so we are aware of all the packages you've touched.

jeffreyguenther:
Will return to this shortly.

jeffreyguenther:
yup, still matters

PR #5339, opened 2025-02-03
Unify token exchange for app managament client

isaacroldan:

Consolidates authentication flows for App Management and Business Platform APIs into a single function call, reducing duplicate authentication requests and improving performance.

Previously, by having two different calls to authenticate (1 for BP and 1 for AppManagement) we were triggering the whole identity token validation flow twice, now only once.

✅ This change saves up to 0.5 seconds on each command.

- Introduces a new `ensureAuthenticatedAppManagementAndBusinessPlatform` function that handles authentication for both APIs simultaneously
- Updates the `AppManagementClient` to use the new consolidated authentication method
- Ensures proper token management and session handling for both platforms (BP token is stored in session now)
- Improves token refresh logic to handle both API tokens

### How to test your changes?

0. Delete any previous session (`pnpm shopify auth logout`)
1. Run any CLI command that requires authentication
2. Verify that authentication works for both App Management and Business Platform APIs

If you want to dig further, you can run in debug and stop at the "token introspection" function. See that it should only run that once per command and not twice as it is now.

### Measuring impact

### Checklist

### DIFF
--- a/packages/app/src/cli/utilities/developer-platform-client/app-management-client.ts
+++ b/packages/app/src/cli/utilities/developer-platform-client/app-management-client.ts
@@ -171,22 +171,24 @@ export class AppManagementClient implements DeveloperPlatformClient {
         throw new Error('AppManagementClient.session() should not be invoked dynamically in a unit test')
       }
 
-      const {appToken, userId, businessPlatformToken} = await ensureAuthenticatedAppManagementAndBusinessPlatform()
-      const userAccount = await businessPlatformRequestDoc(UserInfo, businessPlatformToken)
+      const tokenResult = await ensureAuthenticatedAppManagementAndBusinessPlatform()
+      const {appManagementToken, businessPlatformToken, userId} = tokenResult
 
-      if (userAccount.currentUserAccount) {
+      const userInfoResult = await businessPlatformRequestDoc(UserInfo, businessPlatformToken)
+
+      if (userInfoResult.currentUserAccount) {
         this._session = {
-          token: appToken,
+          token: appManagementToken,
           businessPlatformToken,
           accountInfo: {
             type: 'UserAccount',
-            email: userAccount.currentUserAccount.email,
+            email: userInfoResult.currentUserAccount.email,
           },
           userId,
         }
       } else {
         this._session = {
-          token: appToken,
+          token: appManagementToken,
           businessPlatformToken,
           accountInfo: {
             type: 'UnknownAccount',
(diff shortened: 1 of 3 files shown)


packages/cli-kit/src/public/node/session.ts
```diff
@@ -62,25 +62,35 @@ ${outputToken.json(scopes)}
...
- * @param scopes - Optional array of extra scopes to authenticate with.
+ * @param appManagementScopes - Optional array of extra scopes to authenticate with.
+ * @param businessPlatformScopes - Optional array of extra scopes to authenticate with.
  * @param env - Optional environment variables to use.
  * @param options - Optional extra options to use.
  * @returns The access token for the App Management API.
  */
-export async function ensureAuthenticatedAppManagement(
-  scopes: AppManagementAPIScope[] = [],
+export async function ensureAuthenticatedAppManagementAndBusinessPlatform(
```
shauns, line 71: I find it hard to imagine this being the case, but lets be sure there's no consumers of `ensureAuthenticatedAppManagement` that are going to break.

packages/app/src/cli/utilities/developer-platform-client/app-management-client.ts
```diff
@@ -199,23 +202,22 @@ export class AppManagementClient implements DeveloperPlatformClient {
     return (await this.session()).token
   }
 
+  async businessPlatformToken(): Promise<string> {
+    return (await this.session()).businessPlatformToken
+  }
+
   async refreshToken(): Promise<string> {
-    const {token} = await ensureAuthenticatedAppManagement([], process.env, {noPrompt: true})
+    const {appToken, businessPlatformToken} = await ensureAuthenticatedAppManagementAndBusinessPlatform(
```
shauns, line 210: appManagementToken I think. appToken implies something to do with this app

packages/app/src/cli/utilities/developer-platform-client/app-management-client.ts
```diff
@@ -199,23 +202,22 @@ export class AppManagementClient implements DeveloperPlatformClient {
...
+      [],
```
shauns, line 211: Maybe worth changing the signature on the function, given how its being used? Looks like it _never_ actually cares about scopes or envs in actual usage, only prompt support. Should make this more readable.

packages/app/src/cli/utilities/developer-platform-client/app-management-client.ts
shauns: same above too

isaacroldan: Updated to `appManagementToken`

packages/app/src/cli/utilities/developer-platform-client/app-management-client.ts
isaacroldan: Good call, I've updated the function signature

PR #5358, opened 2025-02-04
Fix node install command when using yarn

wizardlyhel:

Fixing `cli-kit` when running `shopify hydrogen upgrade` with `yarn`.

The exported `installNodeModules` from `cli-kit` assumes `install` arg works with all package managers that `cli` supports.

### Measuring impact

### Checklist

### DIFF
--- a/packages/cli-kit/src/public/node/node-package-manager.ts
+++ b/packages/cli-kit/src/public/node/node-package-manager.ts
@@ -199,7 +199,7 @@ export async function installNodeModules(options: InstallNodeModulesOptions): Pr
     stderr: options.stderr,
     signal: options.signal,
   }
-  let args = ['install']
+  let args = [options.packageManager === 'yarn' ? 'add' : 'install']
   if (options.args) {
     args = args.concat(options.args)
   }
--- a/packages/cli-kit/src/public/node/node-package-manager.ts
+++ b/packages/cli-kit/src/public/node/node-package-manager.ts
@@ -199,7 +199,22 @@ export async function installNodeModules(options: InstallNodeModulesOptions): Pr
     stderr: options.stderr,
     signal: options.signal,
   }
-  let args = [options.packageManager === 'yarn' ? 'add' : 'install']
+  let args = []
+  /**
+   * If installing all packages:
+   * - `npm install`
+   * - `yarn`
+   * If installing named packages:
+   * - `npm install <package_name>`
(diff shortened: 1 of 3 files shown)


packages/cli-kit/src/public/node/node-package-manager.ts
```diff
@@ -199,7 +199,7 @@ export async function installNodeModules(options: InstallNodeModulesOptions): Pr
     stderr: options.stderr,
     signal: options.signal,
   }
-  let args = ['install']
+  let args = [options.packageManager === 'yarn' ? 'add' : 'install']
```
graygilmore, line 202: I'm not sure this is the right spot for this. I'm 90% sure this is the installation method to install *all* packages which `yarn add` wouldn't accomplish.

wizardlyhel: you mean like ...

1. if `yarn`
   1. if `options.arg` not empty, use `add`
   2. else empty `args` array
2. else `args = ['install']`

graygilmore: I'm not sure this function was ever meant to be used to install specific packages. I think that was something that Hydrogen utilized because it was available. Instead of making this function both install dependencies and add packages I think it would be better to create a separate `addNodeModulePackage` function.

What do you think?

Or since there is already a `getPackageManager` exported function could Hydrogen build the command out itself?

wizardlyhel: I don't understand what is the hold back for this change. Is there a fear of don't know how it is being used elsewhere? Is it suppose to be an internal only function? or is it that simply introducing `yarn` as a possible package manager complicate things?

packages/cli-kit/src/public/node/node-package-manager.ts
```diff
@@ -199,7 +199,22 @@ export async function installNodeModules(options: InstallNodeModulesOptions): Pr
...
-  let args = ['install']
+  let args = []
+  /**
+   * If installing all packages:
+   * - `npm install`
+   * - `yarn`
+   * If installing named packages:
+   * - `npm install <package_name>`
+   * - `yarn add <package_name>`
+   */
```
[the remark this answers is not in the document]
wizardlyhel: ah okay - that makes sense now. I'll close this PR and make the change in Hydrogen instead.

PR #5369, opened 2025-02-06
Make usePollAppLogs tests more predictable

shauns:
We have had the occasional flakey test in usePollAppLogs. This affects our build process so is worth improving; I've taken a stab at it.

I broke out the self-adjusting interval element of the polling hook into its own hook and tests

I also adjusted the app log polling hook more generally:
- I pulled out the interesting/business logic part of the hook into a method. This leaves the hook a lot easier to reason about
- Within the hook, I used React references for any non-stateful instance variables

### DIFF
--- a/packages/app/src/cli/services/app-logs/logs-command/ui/components/hooks/useSelfAdjustingInterval.ts
+++ b/packages/app/src/cli/services/app-logs/logs-command/ui/components/hooks/useSelfAdjustingInterval.ts
@@ -1,0 +1,26 @@
+import {useRef, useEffect} from 'react'
+
+export function useSelfAdjustingInterval<T extends {retryIntervalMs: number | null}>(callback: () => Promise<T>) {
+  const savedCallback = useRef(callback)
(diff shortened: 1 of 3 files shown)


packages/app/src/cli/services/app-logs/logs-command/ui/components/hooks/useSelfAdjustingInterval.ts
```diff
@@ -0,0 +1,26 @@
+import {useRef, useEffect} from 'react'
+
+export function useSelfAdjustingInterval<T extends {retryIntervalMs: number | null}>(callback: () => Promise<T>) {
```
mssalemi, line 3: Nice, I like this abstraction, I like that the whole managing interval is in its own hook now. And maybe the test were flakey due to not having the a `cleanTimeout` 🤔.

mssalemi (approved):
Changes LGTM! I think this is a good improvement.

isaacroldan (approved):

PR #5392, opened 2025-02-14
Update theme info command to support multiple environments

EvilGenius13:
Follow up from https://github.com/Shopify/cli/pull/5284
Allow `theme info` command to be called using multiple environments

These changes allow a user to pass multiple environment flags from their `shopify.theme.toml` file at once.

Before:
image

After:
image

### How to test your changes?

- Pull down the branch
- Build the branch
- If you don't already have one, create a `shopify.theme.toml` file

*Note*: You will need to create an access token through your shop(s) to bypass the need to login during theme commands.
The most basic example of the structure needed would be this:
```
[environments.env1]
store = "myfirstshop.myshopify.com"
password = "access_token"
[environments.env2]
store = "mysecondshop.myshopify.com"
password = "access_token"
```

- Run `theme info -e env1 -e env2` or whatever you named your environments.

### Post-release steps

We will be converting more commands to support multiple environments.

### Measuring impact

### Checklist

- [] I've considered possible [documentation](https://shopify.dev) changes


packages/theme/src/cli/commands/theme/info.ts
```diff
@@ -47,10 +46,8 @@ export default class Info extends ThemeCommand {
         return outputInfo(JSON.stringify(output, null, 2))
       }
 
-      const infoMessage = Object.entries(output.theme)
-        .map(([key, val]) => formatSection(key, `${val}`))
-        .join('\n\n')
-      outputInfo(infoMessage)
+      const formattedInfo = await formatThemeInfo(output, flags)
```
EvilGenius13, line 49: Moved the logic of the output into `packages/theme/src/cli/services/info.ts`

packages/theme/src/cli/services/info.test.ts
```diff
@@ -84,3 +101,54 @@ describe('info', () => {
     })
   })
 })
+
+describe('formatThemeInfo', () => {
+  beforeEach(() => {
```
EvilGenius13, line 106: Due to how our mocking works, we need to refresh when testing `formatSection`. I followed the same principles from another [test](https://github.com/Shopify/cli/blob/c9fef4107353899d4fb79a853d05d52cff462032/packages/app/src/cli/services/import-extensions.test.ts#L68).

packages/theme/src/cli/services/info.ts
```diff
@@ -84,3 +85,26 @@ function tabularSection(title: string, data: InlineToken[][]): AlertCustomSectio
     body: {tabularData: data, firstColumnSubdued: true},
   }
 }
+
```
EvilGenius13, line 88: Originally, I was going to try using `tabularData` but the main problem is the output is `key : value` so if you have `PREVIEW_URL :  https://mystoremyshopify.com?preview_theme_id=130327216212`, it will wrap onto another line when within `renderInfo`.

packages/theme/src/cli/services/info.test.ts
```diff
@@ -84,3 +101,54 @@ describe('info', () => {
...
+            'role: development',
+            'shop: test-shop.myshopify.com',
+            'editor_url: https://test-shop.myshopify.com/editor',
+            'preview_url: https://test-shop.myshopify.com/preview',
+          ].join('\n\n'),
+        },
+      ],
+    })
+
+    expect(formatSection).toHaveBeenCalledTimes(6)
```
graygilmore, line 152: How valuable of an assertion are you thinking this is? Can we skip mocking the function entirely?

packages/theme/src/cli/services/info.ts
karreiro: We could adopt an approach similar to the `shopify theme dev` command and use the `LinkToken` with a truncated label :)

packages/theme/src/cli/services/info.ts
```diff
@@ -84,3 +84,42 @@ function tabularSection(title: string, data: InlineToken[][]): AlertCustomSectio
...
+    }
+  })
+
+  return {
+    customSections: [
+      ...(flags.environment
+        ? [
+            {
+              title: `Theme information`,
+              body: [{subdued: `Environment name: ${flags.environment}`}],
```
karreiro, line 108: I don't think it is ideal to show the environment name under the theme information because the environment is not actually related to any theme.

image

I believe we could instead always display the environment name (when available) and present it in a separate section from the theme.

We may apply that enhancement in a different PR tho 🙇

karreiro (approved):
Thank you, @EvilGenius13!

I've left only one minor comment we may apply in a different PR 🚀

PR #5406, opened 2025-02-17
Add release script to GH

isaacroldan:

Fixes #0000 

### Measuring impact

### Checklist


isaacroldan:
* **#5406** <a href="https://app.graphite.dev/github/pr/Shopify/cli/5406?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Shopify/cli/5406?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

.github/workflows/release-manual.yaml
```diff
@@ -0,0 +1,45 @@
...
+env:
+  PNPM_VERSION: '8.15.7'
+  GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+jobs:
+  release:
+    name: Release
+    runs-on: shopify-ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
```
shauns, line 29: ~Is it possible to set a branch for this? It might make testing this out a bit easier.~ ignore this, I think you choose the branch when you trigger the workflow

.github/workflows/release.yml
```diff
@@ -1,4 +1,4 @@
-name: changeset
+name: Release
 
 on:
   workflow_dispatch:
```
shauns, line 4: There's a workflow dispatch option on this - should we drop it given you have the manual workflow?

.github/workflows/release.yml
gonzaloriestra, line 1: What if we rename this to "Update Release PR" or something similar? Because this is not really releasing...

bin/version
```diff
@@ -0,0 +1,21 @@
...
+  tag="latest"
+fi
+
+if [ "$tag" != "latest" ] && [ "$tag" != "nightly" ] && [ "$tag" != "experimental" ]; then
+  echo "Error: tag must be one of: latest, nightly, experimental"
+  exit 1
+fi
+
+if [ "$tag" = "latest" ]; then
+  pnpm changeset version
```
gonzaloriestra, line 16: Couldn't we include this in the `bin/release` script and get rid of this file?

gonzaloriestra (commented):
It's looking good! I like the automatic release after merge and the way to manually launch it.

.github/workflows/release.yml
isaacroldan: Yep, no need for it here

.github/workflows/release.yml
gonzaloriestra: Right, it happens when it's merged, not when this workflow is run. So I think it will be confusing to see that a "Release" workflow was run and nothing was released.

shauns (approved):

bin/version
gonzaloriestra: What about a no-op command? 
```
  version: echo "Versions are updated in the release script"
```

Just trying to simplify a bit, but we can try to improve it later.

PR #5435, opened 2025-02-19
Add no-cache flag

isaacroldan:

To provide developers with the ability to bypass the CLI's GraphQL cache when they need to fetch fresh data directly from the API.

Adds a new hidden global flag `--no-cache` that allows developers to bypass the GraphQL cache. When this flag is used, the CLI will make direct API calls instead of trying to use cached responses.

### How to test your changes?

1. Run any command that makes GraphQL requests with the `--no-cache` flag
2. Verify that the cache is bypassed and fresh data is fetched from the API
3. Run the same command without the flag to confirm the cache is still working as expected

### Measuring impact

### Checklist

### DIFF
--- a/packages/cli-kit/src/public/node/api/graphql.test.ts
+++ b/packages/cli-kit/src/public/node/api/graphql.test.ts
@@ -338,4 +338,33 @@ describe('graphqlRequest with caching', () => {
     expect(cacheRetrieveSpy).not.toHaveBeenCalled()
     expect(retryAwareSpy).toHaveBeenCalled()
   })
+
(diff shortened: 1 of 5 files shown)


isaacroldan:
@gonzaloriestra do you think this is still useful for internal testing? If you need to run a command multiple times without cache, is easier to use a flag? what do you think?

gonzaloriestra:
I think that situation won't be very common and it's not that hard to run `shopify cache clear`. So I'd discard this PR, but not a strong opinion. We can add it if you think it will be useful.

PR #5451, opened 2025-02-21
Update mutation variables for AppVersionCreate

dmerand:

Following up to https://github.com/Shopify/cli/pull/5436 per [this issue](https://github.com/Shopify/develop-app-inner-loop/issues/2530). Per these issues, we've updated the API shape on the back-end and need to update the CLI to match.

This changes the `CreateAppVersion` GraphQL call to use the updated `AppVersionInput` data shape. I've also added an `AppVersionSourceWithUrl` interface to help to ensure that we're conforming to the API locally. See the [previous PR](https://github.com/Shopify/cli/pull/5436) for a brief discussion of that.

We also didn't have any tests for the `deploy` command, so I've added some.

### How to test your changes?

- `p shopify app init`
- `p shopify app deploy` - which will exercise the non-source-url branch
- `p shopify app generate extension` and pick e.g. Admin Action or another extension that uses assets.
- `p shopify app deploy` - which will exercise the source-url branch

### Measuring impact

### Checklist

### DIFF
--- a/packages/app/src/cli/api/graphql/app-management/queries/create-app-version.graphql
+++ b/packages/app/src/cli/api/graphql/app-management/queries/create-app-version.graphql
@@ -1,5 +1,5 @@
-mutation CreateAppVersion($appId: ID!, $appSource: AppSourceInput!, $name: String!, $metadata: VersionMetadataInput) {
-  appVersionCreate(appId: $appId, appSource: $appSource, name: $name, metadata: $metadata) {
+mutation CreateAppVersion($appId: ID!, $version: AppVersionInput!, $metadata: VersionMetadataInput) {
+  appVersionCreate(appId: $appId, version: $version, metadata: $metadata) {
     version {
       id
       appModules {
--- a/packages/app/src/cli/utilities/developer-platform-client/app-management-client.ts
+++ b/packages/app/src/cli/utilities/developer-platform-client/app-management-client.ts
@@ -582,22 +582,24 @@ export class AppManagementClient implements DeveloperPlatformClient {
     if (brandingModule) {
       updatedName = JSON.parse(brandingModule.config).name
     }
-
+    const metadata = {versionTag, message, sourceControlUrl: commitReference}
+    const queryVersion: AppVersionSourceUrl | AppVersionSource = bundleUrl
+      ? {sourceUrl: bundleUrl}
+      : {
+          source: {
+            name: updatedName,
+            modules: (appModules ?? []).map((mod) => ({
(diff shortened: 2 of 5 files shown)


packages/app/src/cli/api/graphql/app-management/queries/create-app-version.graphql
```diff
@@ -1,5 +1,5 @@
-mutation CreateAppVersion($appId: ID!, $appSource: AppSourceInput!, $name: String!, $metadata: VersionMetadataInput) {
-  appVersionCreate(appId: $appId, appSource: $appSource, name: $name, metadata: $metadata) {
+mutation CreateAppVersion($appId: ID!, $version: AppVersionInput!, $metadata: VersionMetadataInput) {
+  appVersionCreate(appId: $appId, version: $version, metadata: $metadata) {
```
dmerand, line 2: I pulled `name` as well, on the theory that we're either going to pull name from the branding module (short-term), or update the manifest format to include it longer-term.

packages/app/src/cli/utilities/developer-platform-client/app-management-client.ts
```diff
@@ -582,22 +582,24 @@ export class AppManagementClient implements DeveloperPlatformClient {
     if (brandingModule) {
       updatedName = JSON.parse(brandingModule.config).name
     }
-
+    const metadata = {versionTag, message, sourceControlUrl: commitReference}
+    const queryVersion: AppVersionSourceUrl | AppVersionSource = bundleUrl
```
MitchDickinson, line 586: Not really your PR but since the function argument usage is so different between both versions of this request, I wonder if it'd be better to have two functions: i.e. `deploy()` and `deployWithUrl()` or something. I suspect it'd be easier to work with the interface if this was the case, instead of having to worry about a bunch optional function args

MitchDickinson (approved):
LGTM, just one small nit/suggestion that isn't really related to your PR directly anyway.

packages/app/src/cli/utilities/developer-platform-client/app-management-client.ts
MitchDickinson: > Perhaps easier to do post partners migration?

Makes sense - mind adding an issue to the DoD milestone on our board?

dmerand: Issue here: https://github.com/Shopify/develop-app-inner-loop/issues/2631

PR #5454, opened 2025-02-24
Controllable request behaviour

shauns:

Improves network request handling in the CLI by introducing configurable request behaviours, timeouts, and retries. This enhances reliability and responsiveness of network operations across the codebase.

These are available for fetch & shopifyFetch, and used by the GraphQL stack.

These changes are backward-compatible for public interfaces in CLI kit, and test coverage for network APIs is greatly increased.

- Introduces `requestMode` to configure network request behaviors with presets:
  - `default`: Automatic retries and timeouts
  - `non-blocking`: No retries, with timeouts
  - `slow-request`: No retries or timeouts -- i.e. neutral
- Adds configurable abort signals and timeout controls for requests
- Enhances request retry logic with better error handling
- Updates network calls across the codebase to use appropriate request modes
- Adds comprehensive test coverage using MSW for network mocking

### How to test your changes?

See the downstack PR for a test script to follow.

### Measuring impact

### Checklist

### DIFF
--- a/packages/cli-kit/src/public/node/http.test.ts
+++ b/packages/cli-kit/src/public/node/http.test.ts
@@ -1,28 +1,219 @@
+import {downloadFile, shopifyFetch, formData, requestMode, fetch} from './http.js'
+import {mockAndCaptureOutput} from './testing/output.js'
 import {fileExists, inTemporaryDirectory, readFile} from './fs.js'
-import {downloadFile} from './http.js'
 import {joinPath} from './path.js'
-import {describe, expect, test, vi} from 'vitest'
-import nodeFetch, {Response} from 'node-fetch'
-import {Readable} from 'stream'
+import {getAllPublicMetadata} from './metadata.js'
+import {platformAndArch} from './os.js'
+import {afterAll, afterEach, beforeAll, beforeEach, describe, expect, test, vi} from 'vitest'
+import {setupServer} from 'msw/node'
+import {delay, http, HttpResponse} from 'msw'
+import FormData from 'form-data'
 
-vi.mock('node-fetch', async () => {
-  const actual: any = await vi.importActual('node-fetch')
-  return {
-    ...actual,
-    default: vi.fn(),
-  }
+const DURATION_UNTIL_ABORT_IS_SEEN = 100
+
+const mockResponse = {hello: 'world!'}
+
+const handlers = [
+  http.get('https://shopify.example/working', () => {
+    return HttpResponse.json(mockResponse)
+  }),
+  http.get('https://shopify.example/a-slow-endpoint', async () => {
+    await delay(500)
+    return HttpResponse.json(mockResponse)
+  }),
+  http.get('https://shopify.example/a-blocked-endpoint', async () => {
+    await delay('infinite')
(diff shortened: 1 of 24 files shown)


packages/cli-kit/src/public/node/http.ts
```diff
@@ -20,6 +22,131 @@
...
+ *
+ * @param preset - The preset to use.
+ * @returns A request behaviour object.
+ */
+export function requestMode(preset: RequestModeInput = 'default'): RequestBehaviour {
+  const networkLevelRetryIsSupported = !skipNetworkLevelRetry()
+  switch (preset) {
+    case 'default':
+      return {
+        useNetworkLevelRetry: networkLevelRetryIsSupported && true,
```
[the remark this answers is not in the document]
shauns: I thought it was descriptive (this option is enabled but only if supported), but if you got confused, then I guess I was wrong. I'll drop it!

packages/cli-kit/src/public/node/http.ts
```diff
@@ -28,45 +155,58 @@
...
+  return innerFetch(options)
 }
 
 /**
  * A fetch function to use with Shopify services. The function ensures the right
  * TLS configuragion is used based on the environment in which the service is running
- * (e.g. Spin).
+ * (e.g. Spin). NB: headers/auth are the responsibility of the caller.
+ *
+ * By default, the CLI's fetch function's special behaviours, like automatic retries, are enabled.
```
isaacroldan, line 188: Why is it enabled here but disabled in the `fetch` above by default?
`shopifyFetch` is only used for identity i think no?

isaacroldan: I see, the only difference is that shopifyFetch will inject some custom headers. The reasoning makes sense but the truth is that we are using `fetch` also for some shopify APIs :( 
 
Requests using `fetch` that are actually connecting to shopify services:
- `admin.ts` to run queries to the Admin API.
- `storefront-session.ts` to connect to the storefront API
- `utils.ts` to fetch the app logs
- I'm almost sure that Hydrogen is also using it for internal stuff (because they don't use `shopifyFetch`)

So I think we either should:
- Migrate all these to use `shopifyFetch`
- Enable the `default` mode for `fetch` (and maybe deprecate `shopifyFetch`?)

wdyt?

packages/cli-kit/src/public/node/http.test.ts
```diff
@@ -1,28 +1,231 @@
...
+import FormData from 'form-data'
 
-vi.mock('node-fetch', async () => {
-  const actual: any = await vi.importActual('node-fetch')
+const DURATION_UNTIL_ABORT_IS_SEEN = 100
+
+vi.mock('./environment.js', async (importOriginal) => {
+  const actual = await importOriginal<typeof import('./environment.js')>()
+
+  const skipNetworkLevelRetryMock = vi.fn().mockReturnValue(false)
```
isaacroldan, line 18: instead of mocking this, we can inject `process.env` to the `requestMode` function. 

```ts
export function requestMode(preset: RequestModeInput = 'default', processEnv = process.env): RequestBehaviour {
```

isaacroldan (approved):

PR #5475, opened 2025-02-28
Use `.shopify/.gitignore` instead of appending to `.gitignore`

sambostock:

We should avoid "dirtying" the host application's `.gitignore`.

Instead of writing to the host application's `.gitignore` to hide the `.shopify` directory, we can write a `.gitignore` in the directory itself, which makes the directory self-ignoring.

### How to test your changes?

Test coverage should be good enough, but you could use the previous version of the CLI to perform a task which writes to `.gitignore`, and inspect that it wrote to it, then repeat that task using this branch and ensure the `.shopify` directory and its contents remain ignored even if `.gitignore` doesn't contain `.shopify`.

### Measuring impact

### Checklist

### DIFF
--- a/packages/cli-kit/src/public/node/hiddenFolder.ts
+++ b/packages/cli-kit/src/public/node/hiddenFolder.ts
@@ -1,0 +1,35 @@
+import {joinPath} from './path.js'
+import {mkdirSync, writeFileSync} from 'fs'
+
+const HIDDEN_FOLDER_NAME = '.shopify'
+const HIDDEN_CONFIG_PATH = 'project.json'
+
+/**
+ * Creates a git ignored hidden .shopify folder in the given app directory, and pass its path to the given callback.
+ *
+ * @param appDirectory - The directory of the app.
+ * @param callback - A callback that receives the path to the hidden folder.
+ * @returns The value returned by the callback function.
+ */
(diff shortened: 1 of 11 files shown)


packages/cli-kit/src/public/node/hiddenFolder.ts
```diff
@@ -0,0 +1,35 @@
...
+ * Creates a git ignored hidden .shopify folder in the given app directory, and pass its path to the given callback.
+ *
+ * @param appDirectory - The directory of the app.
+ * @param callback - A callback that receives the path to the hidden folder.
+ * @returns The value returned by the callback function.
+ */
+export function withHiddenShopifyFolderIn<T>(appDirectory: string, callback: (hiddenFolder: string) => T): T {
+  const hiddenFolder = joinPath(appDirectory, HIDDEN_FOLDER_NAME)
+  mkdirSync(hiddenFolder, {recursive: true})
+  writeFileSync(joinPath(hiddenFolder, '.gitignore'), `# Ignore the entire ${HIDDEN_FOLDER_NAME} directory\n*`)
```
isaacroldan, line 17: We call `mkdir` and `writeFile` everytime? is that cheaper than checking for existance of the file?

PR #5497, opened 2025-03-10
Add `appLogs` to DeveloperPlatformClient interface

mssalemi:
Part of: https://github.com/Shopify/shopify-functions/issues/605

This PR adds `appLogs` to the `DeveloperPlatformClient` interface. Currently this is not necessarily needed since when we poll for app logs, we only have one client currently in use. But to integrate app logs streaming with the developer dashboard, this will make calling the correct request (partners vs app management) simpler. 

The next step is this follow Up PR: https://github.com/Shopify/cli/pull/5498

This change enables app logs to be fetched from both the `PartnersClient` and we will update the methods for the `AppManagementClient` in a separate PR, providing a consistent interface regardless of which backend is being used.

- Moving app logs fetching from direct HTTP calls to using developerPlatformClient
- Removes direct usage of shopifyFetch in favor of developerPlatformClient.appLogs
- Adds appLogs method to the DeveloperPlatformClient interface
- Implements appLogs in PartnersClient class
- Updates all test files to use testDeveloperPlatformClient instead of mocking shopifyFetch
- Adds new FetchAppLogsOptions interface to type the parameters for app logs fetching
- Updates function signatures to accept developerPlatformClient as a parameter

Component Updates:
- Updates React components and hooks to pass through developerPlatformClient
- Modifies usePollAppLogs hook to use the new client
- Follow Up PR: https://github.com/Shopify/cli/pull/5498

### How to test your changes?

Test these changes in production using your local CLI on this branch (both commands below should work as expected):

```
pnpm run shopify app logs --path=./path-to-app

pnpm run shopify app dev --path=./path-to-app
```

### Measuring impact

### Checklist

### DIFF
--- a/packages/app/src/cli/services/app-logs/utils.ts
+++ b/packages/app/src/cli/services/app-logs/utils.ts
@@ -10,13 +10,11 @@ import {
 import {DeveloperPlatformClient} from '../../utilities/developer-platform-client.js'
 import {AppLogsSubscribeVariables} from '../../api/graphql/subscribe_to_app_logs.js'
 import {AppInterface} from '../../models/app/app.js'
-import {Response, shopifyFetch} from '@shopify/cli-kit/node/http'
 import {outputDebug, outputWarn} from '@shopify/cli-kit/node/output'
 import {partnersFqdn} from '@shopify/cli-kit/node/context/fqdn'
 import {AbortError} from '@shopify/cli-kit/node/error'
 import camelcaseKeys from 'camelcase-keys'
 import {formatLocalDate} from '@shopify/cli-kit/common/string'
-import {CLI_KIT_VERSION} from '@shopify/cli-kit/common/version'
 
 export const POLLING_INTERVAL_MS = 450
 export const POLLING_ERROR_RETRY_INTERVAL_MS = 5 * 1000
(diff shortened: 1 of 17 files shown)


packages/app/src/cli/services/app-logs/utils.ts
```diff
@@ -117,14 +117,16 @@ const generateFetchAppLogUrl = async (
...
+  cursor?: string
   filters?: {
     status?: string
     source?: string
-  },
-): Promise<Response> => {
+  }
+}
+
+export const fetchAppLogs = async ({jwtToken, cursor, filters}: FetchAppLogsOptions): Promise<Response> => {
```
shauns, line 129: I'm wondering if this should be moved into the partners client. Then you know its always called via the client.

shauns (changes requested):
Just some light suggestions. Can you put the follow-up PR into a stack with this?

mssalemi:
> Can you put the follow-up PR into a stack with this?

Sorry! Running into some issues getting these PR's stacked in graphite. Is it okay without for now? Not sure why its not showing in graphite, I tried tracking and submitting with gt 🤔.  Will try to get these stacked.

Screenshot 2025-03-14 at 9 37 23 AM

PR #5500, opened 2025-03-10
Fix GraphQL codegen PR job to actually generate PR each time

amcaplan:
Previously it would exit successfully because a PR already exists. It just happens to be a closed PR. So now we attempt to list open PRs, and if there are none, we open one.

Bottom line, the job works now, and you can see the first PR here: #5499


katiebauchman (approved):
LGTM! Thanks for explaining the logic.

amcaplan:
Skipped merge rules as this is coming to fix the processes which are breaking the merge rules at the moment 😆

PR #5518, opened 2025-03-13
Config modules schemas inherit from BaseSchema

isaacroldan:

The WHY is explained in the previous PR in the stack, but basically:

ALL extensions schemas should inherit from the same BaseSchema so that we can properly use generics and access the typed configuration correctly.

TL;DR
Thanks to this we move from:
```ts
if ('application_url' in config) {
    config.application_url = urls.applicationUrl
}
```

to just
```
  config.application_url = urls.applicationUrl
```

- Adds proper type annotations to extension specification schemas
- Extends `BaseSchema` in all app config schemas for consistency
- Adds missing transformation configurations for app proxy and point of sale extensions (these are needed so that no additional properties are accidentally included)

### How to test your changes?

1. Run the test suite to ensure all tests pass
2. Create a new app with various extension types to verify they load correctly
3. Verify that app proxy and point of sale extensions work as expected with the new transformation configurations

### Measuring impact

### Checklist

### DIFF
--- a/packages/app/src/cli/models/extensions/specifications/app_config_app_proxy.ts
+++ b/packages/app/src/cli/models/extensions/specifications/app_config_app_proxy.ts
@@ -1,8 +1,9 @@
 import {validateUrl} from '../../app/validation/common.js'
-import {ExtensionSpecification, createConfigExtensionSpecification} from '../specification.js'
+import {ExtensionSpecification, TransformationConfig, createConfigExtensionSpecification} from '../specification.js'
+import {BaseSchema} from '../schemas.js'
 import {zod} from '@shopify/cli-kit/node/schema'
 
-const AppProxySchema = zod.object({
+const AppProxySchema = BaseSchema.extend({
   app_proxy: zod
     .object({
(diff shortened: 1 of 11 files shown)


packages/app/src/cli/models/extensions/specifications/app_config_app_proxy.ts
```diff
@@ -14,11 +15,18 @@ const AppProxySchema = zod.object({
 
 export const AppProxySpecIdentifier = 'app_proxy'
 
+const AppProxyTransformConfig: TransformationConfig = {
```
shauns, line 18: I don't understand why transformation is being added here (and below) -- "these are needed so that no additional properties are accidentally included" -- what does this mean? Was something broken?

shauns: If its doing something like filtering the submitted config, do we have test coverage demonstrating that?

shauns: Its just these two modules, right?

Maybe making it mandatory for `createConfigExtensionSpecification` does the job?

Issue #5496, opened 2025-03-10
[Bug]: Interactive authentication does not work in Docker


phrfpeixoto:
### Please confirm that you have:

- [x] Searched [existing issues](.) to see if your issue is a duplicate. (If you’ve found a duplicate issue, feel free to add additional information in a comment on it.)
- [x] Reproduced the issue in [the latest CLI version](https://www.npmjs.com/package/@shopify/cli).

### In which of these areas are you experiencing a problem?

App

### Expected behavior

When running `shopify app dev` inside a container, I expect the interactive authentication process to work seamlessly, regardless of whether there is an open TTY.

The `xdg-utils` package is required for the polling to work correctly:

```
shopify-prototype-app-1  | To run this command, log in to Shopify.
shopify-prototype-app-1  | User verification code: XXXXX-XXXX
shopify-prototype-app-1  | 👉 Open this link to start the auth process: https://accounts.shopify.com/activate-with-code?device_code%5Buser_code%5D=XXXX-XXXX
shopify-prototype-app-1  | ✔ Logged in.
```

This is my Dockerfile:
```Dockerfile
FROM node:18-alpine as base

EXPOSE 3000

WORKDIR /app
COPY . .

FROM base as dev

ENV NODE_ENV=development
ENV PATH="/app/node_modules/.bin:$PATH"

RUN apk --no-cache add ca-certificates wget git gcompat vim xdg-utils

RUN npm install
RUN npm run build

CMD ["shopify", "app", "dev"]
```

### Actual behavior

Since version [3.65.0](https://github.com/Shopify/cli/releases/tag/3.65.0), we get an error regarding the terminal not being interactive:

```
shopify-prototype-app-1  | To run this command, log in to Shopify.
shopify-prototype-app-1  | ╭─ error ──────────────────────────────────────────────────────────────────────╮
shopify-prototype-app-1  | │                                                                              │
shopify-prototype-app-1  | │  Authorization is required to continue, but the current environment does     │
shopify-prototype-app-1  | │  not support interactive prompts.                                            │
shopify-prototype-app-1  | │                                                                              │
shopify-prototype-app-1  | │  To resolve this, specify credentials in your environment, or run the        │
shopify-prototype-app-1  | │  command in an interactive environment such as your local terminal.          │
shopify-prototype-app-1  | │                                                                              │
shopify-prototype-app-1  | ╰──────────────────────────────────────────────────────────────────────────────╯
```

[...]

PR #5521, opened 2025-03-14
Improve auth process for docker environments

isaacroldan:

Fixes an issue where the CLI would fail to authenticate in non-interactive environments that aren't CI environments.

Fixes https://github.com/Shopify/cli/issues/5496

- Adds a new `isCI()` function to check if the current environment is a CI environment
- Updates the device authorization flow to:
  - Only throw an error in CI environments, not in all non-TTY environments
  - Show the cloud message in both cloud environments and non-TTY environments

### How to test your changes?

1. Run the CLI in a non-interactive environment that isn't a CI environment (e.g., pipe the output to a file)
2. Verify that the authentication flow shows the cloud message instead of failing
3. Test in a regular terminal to ensure the updated prompt message is displayed correctly

### Measuring impact

### Checklist

### DIFF
--- a/packages/cli-kit/src/public/node/system.ts
+++ b/packages/cli-kit/src/public/node/system.ts
@@ -181,3 +181,12 @@ export function terminalSupportsPrompting(): boolean {
   }
   return Boolean(process.stdin.isTTY && process.stdout.isTTY)
 }
+
+/**
+ * Check if the current environment is a CI environment.
+ *
+ * @returns True if the current environment is a CI environment.
+ */
+export function isCI(): boolean {
+  return isTruthy(process.env.CI)
+}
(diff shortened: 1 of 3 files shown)


packages/cli-kit/src/public/node/system.ts
```diff
@@ -181,3 +181,12 @@ export function terminalSupportsPrompting(): boolean {
   }
   return Boolean(process.stdin.isTTY && process.stdout.isTTY)
 }
+
+/**
+ * Check if the current environment is a CI environment.
+ *
+ * @returns True if the current environment is a CI environment.
+ */
+export function isCI(): boolean {
```
gonzaloriestra, line 190: Surprising that we didn't have this function. Could you update the places where we directly use `process.env.CI`?

gonzaloriestra (approved):

PR #5542, opened 2025-03-20
Remove empty brackets and path from human readable zod error message

jamieguerrero:

Fixes https://github.com/Shopify/develop-app-inner-loop/issues/2237

This PR looks at the path that zod provides when parsing the schema. When error occurs in the 'root' of the schema, the path is an empty array. This passes `parseHumanReadableError` an empty string and causing the error to look like this:
image

After these changes, and providing a specific error message for mismatching the `LegacyAppSchema` 
Screenshot 2025-03-20 at 4 11 55 PM

### How to test your changes?

- Create an app
- Remove `client_id` or provide incorrect types for any of the types provided in `AppSchema` or `LegacyAppSchema`
- View the error message provided  

### Measuring impact

### Checklist

### DIFF
--- a/packages/app/src/cli/models/app/app.ts
+++ b/packages/app/src/cli/models/app/app.ts
@@ -53,7 +53,7 @@ export const LegacyAppSchema = zod
       })
       .optional(),
   })
-  .strict()
+  .strict({message: 'You may be missing client_id, or your TOML is malformed'})
 
 function removeTrailingPathSeparator(value: string[] | undefined) {
   // eslint-disable-next-line no-useless-escape
--- a/packages/app/src/cli/models/app/loader.ts
+++ b/packages/app/src/cli/models/app/loader.ts
@@ -122,8 +122,8 @@ export async function parseConfigurationFile<TSchema extends zod.ZodType>(
 export function parseHumanReadableError(issues: Pick<zod.ZodIssueBase, 'path' | 'message'>[]) {
   let humanReadableError = ''
   issues.forEach((issue) => {
-    const path = issue.path ? issue?.path.join('.') : 'n/a'
-    humanReadableError += `• [${path}]: ${issue.message}\n`
+    const path = issue.path?.length ? `[${issue.path.join('.')}]: ` : ''
+    humanReadableError += `• ${path}${issue.message}\n`
   })
   return humanReadableError
 }


jamieguerrero:
> Much more legible - human readable and helps the user trouble shoot more effectively.
> 
> Do we ever link to docs? `TOML is malformed` <- If devs know what this means and how to correct - no issues.

We can link to docs! Happy to change the text to maybe "make sure all required properties are present" or something like that and link out to [shopify.dev](https://shopify.dev/docs/apps/build/cli-for-apps/app-configuration)? wdyt?

packages/app/src/cli/models/app/app.ts
```diff
@@ -53,7 +53,7 @@ export const LegacyAppSchema = zod
       })
       .optional(),
   })
-  .strict()
+  .strict({message: 'You may be missing client_id, or your TOML is malformed'})
```
amcaplan, line 56: This solves the local problem, but now if someone is missing a single key from the legacy app schema, they will not know what's wrong. Is that the right tradeoff here?

packages/app/src/cli/models/app/loader.ts
```diff
@@ -122,8 +122,8 @@ export async function parseConfigurationFile<TSchema extends zod.ZodType>(
 export function parseHumanReadableError(issues: Pick<zod.ZodIssueBase, 'path' | 'message'>[]) {
   let humanReadableError = ''
   issues.forEach((issue) => {
-    const path = issue.path ? issue?.path.join('.') : 'n/a'
-    humanReadableError += `• [${path}]: ${issue.message}\n`
+    const path = issue.path?.length ? `[${issue.path.join('.')}]: ` : ''
+    humanReadableError += `• ${path}${issue.message}\n`
```
amcaplan, line 126: 👍 on this part

amcaplan (commented):
I'm sorry I have so much to say here on such a small PR, but:

1. Did you figure out why we seem to prefer the `LegacyAppSchema` in the parsing code?
2. As mentioned inline, a missing `client_id` causes a weird error, but if they really are on the `LegacyAppSchema` then the error is really just not helpful anymore. How much do we care about that situation? If it's still relevant (and I'm guessing it is), how can we support that situation better?

By the way, if we can just start preferring the new app schema, or validating against both, that might be a better fix than this strategy.

packages/app/src/cli/models/app/app.ts
isaacroldan: Yeah, I don't think strict here is the right solution.

Also, we should not be using `LegacySchema` almost never, seems like it's only being used "by accident" because if you remove the client-id the CLI thinks the app was never linked and defaults to the `LegacyAppSchema` where everything is optional.

isaacroldan:
Agree with Ariel, we need to look on usages of `LegacyAppSchema`. This probably requires a bigger refactor to completely remove `LegacyAppSchema` 🤔

PR #5560, opened 2025-03-24
Support Remote DOM ui extensions

vividviolet:

Related to https://github.com/Shopify/app-ui/issues/2158

Enables support for Remote DOM UI extensions when using the `REMOTE_DOM_EXPERIMENT` environment variable.

### How to test your changes?

1. Checkout this branch and set your `CLI_KIT_VERSION` to `3.78.0`
1. Grab the relative path to the directory for an existing app (or create one using `pnpm create-app` if you don't have one)
1. Create a remote ui extension by running `pnpm shopify app generate --path <app-directory>`
1. Select an Admin Block extension and chose the Typescript React template
1. Verify that it successfully creates the remote-ui block extension
1. Select a Checkout UI extension and chose the Typescript React template
1. Verify that it successfully creates the checkout-ui extension
1. Create a remote dom extension by running `REMOTE_DOM_EXPERIMENT=true pnpm shopify app generate REMOTE_DOM_EXPERIMENT=true pnpm shopify app generate extension --clone-url https://github.com/Shopify/extensions-templates#2025-07 --path <app-directory>`
1. Select an Admin Block extension and chose the Typescript React template
1. Verify that it successfully creates the remote-dom block extension
1. Run `REMOTE_DOM_EXPERIMENT=true pnpm shopify app dev --path ../../../apps/remote-dom-feb-2025`
1. Verify that runs successfully.
    NOTE: You will see errors relating to updating the drafts like "Error while updating drafts: "2025-07" is not a valid API version" - this is expected until `2025-07` is publicly available.
1. Open the remote-dom Block extension's TOML file and update the target
1. Verify that after you update the target the `shopify.extension.target` is typed to the new target. 
1. Verify that updating the target to be an `action` extension like `"admin.gift-card-details.action.render"` you get type errors for trying to use the `<s-admin-block>` element.
1. Verify that the remote-ui extension still works the same as before:
- changing its target should not impact the shared type file
- no type reference comment is added

Uploading Screen Recording 2025-03-27 at 3.21.21 PM.mov…

### Checklist

### DIFF
--- a/packages/app/src/cli/models/app/app.ts
+++ b/packages/app/src/cli/models/app/app.ts
@@ -17,7 +17,7 @@ import {patchAppHiddenConfigFile} from '../../services/app/patch-app-configurati
 import {ZodObjectOf, zod} from '@shopify/cli-kit/node/schema'
 import {DotEnvFile} from '@shopify/cli-kit/node/dot-env'
 import {getDependencies, PackageManager, readAndParsePackageJson} from '@shopify/cli-kit/node/node-package-manager'
-import {fileRealPath, findPathUp} from '@shopify/cli-kit/node/fs'
+import {fileExistsSync, fileRealPath, findPathUp, readFileSync, writeFileSync} from '@shopify/cli-kit/node/fs'
 import {joinPath} from '@shopify/cli-kit/node/path'
 import {AbortError} from '@shopify/cli-kit/node/error'
 import {normalizeDelimitedString} from '@shopify/cli-kit/common/string'
@@ -295,6 +295,7 @@ export interface AppInterface<
   removeExtension: (extensionUid: string) => void
   updateHiddenConfig: (values: Partial<AppHiddenConfig>) => Promise<void>
   setDevApplicationURLs: (devApplicationURLs: ApplicationURLs) => void
+  generateExtensionTypes(): Promise<void>
 }
 
 type AppConstructor<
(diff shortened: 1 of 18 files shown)


packages/app/src/cli/models/app/loader.test.ts
```diff
@@ -2377,6 +2378,31 @@ wrong = "property"
...
+    const generateTypesSpy = vi.spyOn(App.prototype, 'generateExtensionTypes')
+    // When
+    await loadTestingApp()
+    // Then
+    expect(generateTypesSpy).not.toHaveBeenCalled()
+    generateTypesSpy.mockRestore()
+  })
+
+  test('call app.generateExtensionTypes when REMOTE_DOM_EXPERIMENT is true', async () => {
+    process.env.REMOTE_DOM_EXPERIMENT = 'true'
```
isaacroldan, line 2393: We prefer not to modify the real `process.env` in a test. Because the process might crash before you can clean it up.

Instead, we should inject the `env` to the loader so it can be easily replaced in a test:
This can be easily solved with few changes:
- modify `AppLoaderConstructorArgs` to accept an optional `env?: NodeJS.ProcessEnv`
- Add a new property to the loader and set it in the constructor: `private readonly env: NodeJS.ProcessEnv` & `this.env = env ?? process.env`
- use `this.env.REMOTE_DOM_EXPERIMENT` inside the loader.
- In the test, modify `loadTestingApp` to also accept an optional env `env?: NodeJS.ProcessEnv`
- When running your tests, inject the value you need when calling `loadTestingApp`

packages/app/src/cli/models/app/loader.ts
```diff
@@ -389,6 +389,10 @@ class AppLoader<TConfig extends AppConfiguration, TModuleSpec extends ExtensionS
       usedCustomLayoutForExtensions: configuration.extension_directories !== undefined,
     })
 
+    if (process.env.REMOTE_DOM_EXPERIMENT) {
```
isaacroldan, line 392: use `isTruthy(process.env.REMOTE_DOM_EXPERIMENT)`, otherwise using `REMOTE_DOM_EXPERIMENT=false` would also pass this `if`.

(and following my previous comment about injecting the env -> `isTruthy(this.env.REMOTE_DOM_EXPERIMENT)`)

packages/app/src/cli/services/generate/extension.ts
```diff
@@ -277,6 +281,15 @@ async function uiExtensionInit({
...
   ]
+
+  // @todo: How to get an instance for the newly generated extension?
+  // if (process.env.REMOTE_DOM_EXPERIMENT) {
+  //   tasks.push({
+  //     title: 'Update shared type definition',
+  //     task: async () => app.generateExtensionTypes(),
+  //   })
+  // }
+
```
isaacroldan, line 292: You'd need to reload the app to get it

packages/app/src/cli/models/app/loader.ts
```diff
@@ -389,6 +389,10 @@ class AppLoader<TConfig extends AppConfiguration, TModuleSpec extends ExtensionS
...
+      await appClass.generateExtensionTypes()
```
isaacroldan, line 393: can we measure how long this takes? for both cases where we don't have a type file and when we already have one.
(best to use `const start = startHRTime()` and `endHRTimeInMs(start)` for accurate values)

packages/app/src/cli/models/app/app.ts
```diff
@@ -497,6 +500,38 @@ export class App<
     this.realExtensions = this.realExtensions.filter((ext) => ext.uid !== extensionUid)
   }
 
+  async generateExtensionTypes() {
+    const start = startHRTime()
```
isaacroldan, line 504: remember to delete this before merging

craigmichaelmartin: @vividviolet is this still relevant?

PR #5567, opened 2025-03-26
Enforce in-house fetch function

shauns:

To ensure that HTTP proxy support is maintained across the CLI by enforcing the use of our custom fetch implementation instead of Node.js built-in fetch.

Adds an ESLint rule to restrict the use of the global `fetch` function, requiring developers to use our custom implementation from `@shopify/cli-kit/node/http` instead. The built-in Node.js fetch doesn't support HTTP proxies, which can cause issues for users behind corporate proxies.

The PR:
- Adds a new ESLint rule `no-restricted-globals` to flag usage of the global fetch
- Adds appropriate `eslint-disable-next-line` comments to existing fetch usages that are exceptions


gonzaloriestra (approved):
Nice!

PR #5616, opened 2025-04-08
[Theme] Stop replacing request method when calling Section Rendering API without `replace_templates`

frandiox:

This is just an internal fix but it shouldn't affect anything outside.
We were incorrectly replacing the method for every request to Section Rendering API with `POST`, even though we only need to do that when `replace_templates` is sent.

### Measuring impact

### Checklist

### DIFF
--- a/packages/theme/src/cli/utilities/theme-environment/storefront-renderer.ts
+++ b/packages/theme/src/cli/utilities/theme-environment/storefront-renderer.ts
@@ -11,9 +11,9 @@ export async function render(session: DevServerSession, context: DevServerRender
   const headers = await buildHeaders(session, context)
   let response: Response
 
-  if (context.replaceTemplates) {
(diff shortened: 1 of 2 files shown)


packages/theme/src/cli/utilities/theme-environment/storefront-renderer.ts
```diff
@@ -11,9 +11,9 @@ export async function render(session: DevServerSession, context: DevServerRender
   const headers = await buildHeaders(session, context)
   let response: Response
 
-  if (context.replaceTemplates) {
-    const replaceTemplates = Object.keys({...context.replaceTemplates, ...context.replaceExtensionTemplates})
+  const replaceTemplates = Object.keys({...context.replaceTemplates, ...context.replaceExtensionTemplates})
 
+  if (replaceTemplates.length > 0) {
```
frandiox, line 16: `context.replaceTemplates` is always `{}`, so truthy

PR #5621, opened 2025-04-08
[DNH] Proto fix 401 re-subscribe issues in dev dash

mssalemi:
Related: https://github.com/shop/issues-shopifyvm/issues/249

Paired with Jacob we found an issue when the JWT token is expired and we go into a constant re-subscribe loop. 

We found the issue is that the token used when invalid and needed to be refreshed by subscribing. 

### Measuring impact

### Checklist


packages/app/src/cli/services/app-logs/utils.ts
```diff
@@ -194,6 +194,10 @@ export const subscribeToAppLogs = async (
   variables: AppLogsSubscribeMutationVariables,
   organizationId: string,
 ): Promise<string> => {
+  // Refresh the token since it might have expired since the last subscribeToAppLogs request
+  await developerPlatformClient.refreshToken()
+
+  // Now try the subscription with the fresh token
```
isaacroldan, line 200: How often is `subscribeToAppLogs` called?

PR #5630, opened 2025-04-09
[WIP] Specify use localhost port

byrichardpowell:

Fixes #0000 

### Measuring impact

### Checklist


isaacroldan:
> Run pnpm shopify app dev --path='[PATH]' --use-localhost --localhost=[PORT] in 2 separate terminals at once. The 2nd time you run it you should see a warning (see below) and a different port should be selected.

I think this case should actually crash. If you are explicitly passing a setting (a specific port) and that port is not available it makes more sense to stop the execution than to ignore the setting

packages/app/src/cli/services/dev/urls.ts
```diff
@@ -28,8 +28,17 @@ export interface ApplicationURLs {
...
+  port: number
+  tunnelUrl?: undefined
+  tunnelClient?: undefined
+}
+
+interface UseTunnelFrontendUrlOptions {
+  noTunnelUseLocalhost: false
   tunnelUrl?: string
   tunnelClient: TunnelClient | undefined
 }
```
isaacroldan, line 44: I think this can be simplified with either 

```
interface UseLocalhostFrontendUrlOptions {
  noTunnelUseLocalhost: true
  port: number
}

interface UseTunnelFrontendUrlOptions {
  noTunnelUseLocalhost: false
  tunnelUrl?: string
  tunnelClient: TunnelClient | undefined
}
```

This simplified interfaces work in `generateFrontedURL` if you do an early return when `noTunnelUseLocalhost = true`

or just keep the old Interface and generate the port inside `generateFrontendURL`

packages/cli/oclif.manifest.json
```diff
@@ -453,6 +453,18 @@
           "name": "graphiql-port",
           "type": "option"
         },
+        "localhost-port": {
+          "dependsOn": [
+            "use-localhost"
+          ],
+          "description": "Port to use for localhost. Only applicable when --use-localhost is specified.",
```
nickwesselman, line 460: Should this just enable localhost and not depend on `--use-localhost`? Is there a reason to require both?

nickwesselman: I think `--localhost-port` implying `--use-localhost` is fine.

packages/app/src/cli/services/dev.ts
```diff
@@ -524,3 +536,71 @@ async function validateCustomPorts(webConfigs: Web[], graphiqlPort: number) {
...
+  tunnelUrl,
+}: {
+  useLocalhost?: boolean
+  localhostPort?: number
+  tunnelUrl?: string
+}): Promise<TunnelMode> {
+  if (useLocalhost) {
+    const requestedPort = localhostPort ?? ports.localhost
+    const actualPort = await getAvailableTCPPort(requestedPort)
+
```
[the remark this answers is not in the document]
byrichardpowell: Yep, shifted things around.  If it doesn't get the specified port it throws.

packages/app/src/cli/services/dev.ts
```diff
@@ -524,3 +536,80 @@ async function validateCustomPorts(webConfigs: Web[], graphiqlPort: number) {
...
+        },
+      })
+
+      // The user didn't specify a port. The default isn't available. Warn
+      if (requestedPort !== actualPort) {
+        renderWarning({
+          headline: [
+            'A random port will be used for localhost because',
+            {command: `${requestedPort}`},
+            'is not available.',
```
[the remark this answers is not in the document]
byrichardpowell: Resolving since we agreed it can be a follow on. I also don't think it's critical

packages/app/src/cli/services/dev.ts
```diff
@@ -524,3 +545,136 @@ async function validateCustomPorts(webConfigs: Web[], graphiqlPort: number) {
 function setPreviousAppId(directory: string, apiKey: string) {
   setCachedAppInfo({directory, previousAppId: apiKey})
 }
+
+/**
+ * Gets the tunnel or localhost config for doing app dev
+ * @param options - Options required for the config
+ * @returns A tunnel configuration object
+ */
+export async function getTunnelMode({
```
isaacroldan, line 554: This file is already very long, can we move this to a new file? something like `cli/services/dev/tunnel-mode.ts`

packages/app/src/cli/commands/app/dev.ts
```diff
@@ -139,46 +141,22 @@ If you're using the Ruby app template, then you need to complete the following s
...
-    } else if (flags['tunnel-url']) {
-      tunnelType = 'custom'
-      tunnel = {
-        mode: 'provided',
-        url: flags['tunnel-url'],
-      }
-    }
+    // getTunnelMode can populate this array. This array is then passed to the dev function.
+    // This allows the dev function to group port warnings into a single renderWarning
+    const portWarnings: PortWarning[] = []
```
isaacroldan, line 146: I don't think you need to start this here and pass it to `getTunnelMode` and the `dev` service.

inside `prepareForDev` before calling `renderPortWarnings` you can just do:
```ts
if (tunnel.mode === 'use-localhost' && tunnel.port !== ports.localhost) {
    portWarnings.push({
      type: 'localhost',
      requestedPort: tunnel.port,
      flag: '--localhost-port',
    })
  }
```

packages/app/src/cli/services/dev.ts
```diff
@@ -524,3 +545,136 @@ async function validateCustomPorts(webConfigs: Web[], graphiqlPort: number) {
...
+ * ```ts
+ *   const items = ['apple', 'banana', 'cherry'];
+ *   const result = asHumanFriendlyList(items)
+ *
+ *   //['apple', ',', 'banana', ',', 'and', 'cherry']
+ *   console.log(result);
+ * ```
+ */
+
+function asHumanFriendlyTokenList(items: Token[]): Token[] {
```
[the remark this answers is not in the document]
byrichardpowell: Yeah, Moved to cli-kit/common/array, renamed to asHumanFriendlyArray and updated implementation to accept a generic so the return array can be explicitly typed if needed

packages/app/src/cli/services/dev.ts
byrichardpowell: Yeah, fair.  I wasn't sure what the CLI's preference is here.

I've moved renderPortWarnings() and getTunnelMode() into their own files and it's much nicer.

nickwesselman (approved):

isaacroldan (approved):

PR #5646, opened 2025-04-14
Automatically retry partners requests that fail with 401

shauns:

To improve error handling in the Partners API client by adding support for token refresh when unauthorized responses are received.

Adds an unauthorized handler to the Partners API client that automatically refreshes the token when an unauthorized response is received. This prevents API requests from failing due to expired tokens by:

1. Adding an optional `unauthorizedHandler` parameter to `partnersRequest` and `partnersRequestDoc` functions
2. Creating a handler in the `PartnersClient` class that refreshes the token when needed
3. Updating tests to expect the new handler parameter

### Measuring impact

We should see an improvement in our internal command success metrics.

### Look out for

Do we need to increase the scope of this? Other services and such.

### Checklist

### DIFF
--- a/packages/app/src/cli/utilities/developer-platform-client/partners-client.ts
+++ b/packages/app/src/cli/utilities/developer-platform-client/partners-client.ts
@@ -159,7 +159,7 @@ import {TypedDocumentNode} from '@graphql-typed-document-node/core'
 import {isUnitTest} from '@shopify/cli-kit/node/context/local'
 import {AbortError} from '@shopify/cli-kit/node/error'
 import {generateFetchAppLogUrl, partnersRequest, partnersRequestDoc} from '@shopify/cli-kit/node/api/partners'
-import {CacheOptions, GraphQLVariables} from '@shopify/cli-kit/node/api/graphql'
+import {CacheOptions, GraphQLVariables, RefreshTokenOnAuthorizedResponse} from '@shopify/cli-kit/node/api/graphql'
 import {ensureAuthenticatedPartners} from '@shopify/cli-kit/node/session'
 import {partnersFqdn} from '@shopify/cli-kit/node/context/fqdn'
 import {Response, shopifyFetch} from '@shopify/cli-kit/node/http'
@@ -218,7 +218,9 @@ export class PartnersClient implements DeveloperPlatformClient {
   public readonly supportsDevSessions = false
   public readonly supportsStoreSearch = false
   public readonly organizationSource = OrganizationSource.Partners
+
   private _session: PartnersSession | undefined
+  private tokenRefreshInProgress: undefined | Promise<string>
 
   constructor(session?: PartnersSession) {
     this._session = session
@@ -247,14 +249,14 @@ export class PartnersClient implements DeveloperPlatformClient {
     variables: GraphQLVariables | undefined = undefined,
     cacheOptions?: CacheOptions,
   ): Promise<T> {
-    return partnersRequest(query, await this.token(), variables, cacheOptions)
+    return partnersRequest(query, await this.token(), variables, cacheOptions, this.createUnauthorizedHandler())
   }
 
   async requestDoc<TResult, TVariables extends {[key: string]: unknown}>(
     document: TypedDocumentNode<TResult, TVariables>,
     variables?: TVariables,
   ): Promise<TResult> {
-    return partnersRequestDoc(document, await this.token(), variables)
+    return partnersRequestDoc(document, await this.token(), variables, this.createUnauthorizedHandler())
   }
 
   async token(): Promise<string> {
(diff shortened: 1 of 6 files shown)


packages/cli-kit/src/private/node/api.ts
```diff
@@ -68,6 +68,14 @@ type VerboseResponse<T> =
   | CanRetryErrorResponse
   | UnauthorizedErrorResponse
 
+export interface UnauthorizedHandlerThrow {
+  action: 'throw'
+}
+// `undefined` is a legacy version of this.
+type UnauthorizedHandlerContinue = {action: 'continue'} | undefined
```
isaacroldan, line 75: this `{action: 'continue'}` is not used anywhere right?

packages/cli-kit/src/public/node/api/graphql.ts
```diff
@@ -113,12 +146,35 @@ async function performGraphQLRequest<TResult>(options: PerformGraphQLRequestOpti
     }
   }
 
+  const unauthorizedHandlerFunction =
```
MitchDickinson, line 149: Since this block uses either `unauthorizedHandler` or `refreshTokenOnAuthorizedResponse`, should the "Either-or-ness" be expressed in the function signature? Currently it makes no sense to pass both of these, and I found it confusing that we accepted both arguments until I realized it was an "either or" semantic

packages/cli-kit/src/public/node/api/graphql.ts
```diff
@@ -113,12 +146,35 @@ async function performGraphQLRequest<TResult>(options: PerformGraphQLRequestOpti
...
+          const {client: newClient, headers: newHeaders} = await createGraphQLClient({
+            url,
+            addedHeaders,
+            token: refreshedToken,
+          })
+          client = newClient
+          headers = newHeaders
+          return {action: 'continue'}
+        } else {
+          return {action: 'throw'}
```
MitchDickinson, line 165: Should this `return refreshTokenResult` instead instead of hardcoding `throw`? The interface already expresses the allowed return types, and by hardcoding `throw` here we are making the assumption there will only ever be 2 possibilities, instead of just delegating that to the type (and allowing `retry` to be the special case in this component

MitchDickinson: Alternatively, is there value in reproducing the `{action: ...}` part on the `refreshTokenOnAuthorizedResponse` Function? What if that function just returns a new token or it doesn't, and then this code here could do the mapping: "If I get a new token, I continue, else I throw". 

It feels more complex to bleed the `UnauthorizedHandlerResult` type through multiple layers.

packages/app/src/cli/utilities/developer-platform-client/partners-client.ts
```diff
@@ -639,6 +641,25 @@ export class PartnersClient implements DeveloperPlatformClient {
     const appsWithOrg = org.apps.nodes.map((app) => ({...app, organizationId: org.id}))
     return {organization: parsedOrg, apps: {...org.apps, nodes: appsWithOrg}, stores: []}
   }
+
+  private createUnauthorizedHandler(): () => RefreshTokenOnAuthorizedResponse {
+    let unauthorisedRetriesUsed = 0
```
MitchDickinson, line 646: Why do we need to redefine retry limits here instead of letting the lower level systems do their job? retryAwareRequest already has a retry counter, for example. I guess we are saying we want to retry LESS times if there is this manual re-auth step in the middle? But couldn't we instead propagate a different value for `limitRetriesTo`?

MitchDickinson (commented):
I spent a lot of time with this PR this morning! Partly because I am a TypeScript n00b still, but also because I think this is somewhat complex and was tough to wrap my head around.

I think the complexity comes from having to propagate handlers all the way down the stack to the retryAwareRequest. So there's complexity in all the lambda definitions, new types we have to add, and making sure we can do the right thing from within the control flow of that `retryAwareRequest` loop.

I wonder if there a solution where retry logic can live in multiple layers? For instance, I don't _think_ a HTTP 401/403 error needs to be retried from within `retryAwareRequest`. For example, the `performGraphQLRequest` could do its own retry loop, specifically designed to handle re-auth. That might create a nice separation of the "lower level retry loop" (which could handle lower level infra failures, HTTP 503s, etc) from the "higher level retry loop" (which could handle higher level failures like HTTP 401 which require much more domain-specific handling).

Otherwise, it seems like the interface for `retryAwareRequest` bleeds across many layers of abstraction, and that feels complex.

It's cool that the CLI lets us do things like take the user through an entire blocking re-auth loop from within the control flow of a low level HTTP function. But it also feels like this results in more complex-to-understand code.

Also question: why no management-client implementation?

packages/app/src/cli/utilities/developer-platform-client/partners-client.ts
shauns: The thinking was the token refresh is a big expensive call (or multiple calls), and if it doesn't fix the 401 issue first time, it's not going to. I suppose this could be a boolean and have the same result.

The low level retry is for client errors, like a DNS issue or failing to open a connection - you might get those still after refreshing the token and it still make sense to retry the request. You might _also_ get a 401, refresh the token, and then get a rate-limited response, which implies _another_ retry.

Maybe the challenge here is there are retries on retries 🤔 
- Lowest level: client errors -- i.e. the request did not even complete
- Response level: We have this new 401-retry logic, and there is also existing logic for retry-because-rate-limited (which is for non-GraphQL calls too)
- Application level: I don't think we do this anywhere in apps, but in some places the calling code may have its own retry loop, e.g. polling

packages/app/src/cli/utilities/developer-platform-client/partners-client.ts
```diff
@@ -639,6 +641,25 @@
...
+    return async () => {
+      unauthorisedRetriesUsed++
+      if (unauthorisedRetriesUsed > 1) {
+        return {action: 'throw'}
+      }
+      if (this.tokenRefreshInProgress) {
+        const refreshedToken = await this.tokenRefreshInProgress
+        return {action: 'retry', token: refreshedToken}
+      } else {
+        this.tokenRefreshInProgress = this.refreshToken()
```
shauns, line 656: Note from a DM:  this needs an additional argument to _force_ refresh

PR #5675, opened 2025-04-22
Hot-reload `{% javascript %}` tags sections and blocks

karreiro:

These changes aim to hot-reload scripts served via compiled_assets.

This PR updates the proxy to serve those files locally.

### How to test your changes?

- Add a `{% javascript %}` to a section
- Open the local proxy at http://127.0.0.1:9292
- Update the javascript code in the section above
- Notice the page gets hot-reloaded in the browser

### Post-release steps

N/A

### Measuring impact

### Checklist


packages/theme/src/cli/utilities/theme-environment/local-assets.ts
```diff
@@ -94,35 +94,143 @@ function isCompiledAssetRequest(event: H3Event) {
...
+    case 'styles.css':
+      return handleStylesCss(ctx, event)
+    case 'block-scripts.js':
+      return handleBlockScriptsJs(ctx, event, 'block')
+    case 'snippet-scripts.js':
+      return handleBlockScriptsJs(ctx, event, 'snippet')
+    case 'section-scripts.js':
+      return handleBlockScriptsJs(ctx, event, 'section')
+    default:
+      return handleError(event)
```
frandiox, line 107: Maybe we should let it go to the server in this case instead of throwing? In case we introduce a new type or something like that?

packages/theme/src/cli/utilities/theme-environment/local-assets.ts
```diff
@@ -94,35 +94,143 @@ function isCompiledAssetRequest(event: H3Event) {
...
+    getMeta: () => ({type: 'text/css', size: stylesheet.length, mtime: new Date()}),
+  })
+}
 
-    const stylesheets = ['/*** GENERATED LOCALLY ***/\n']
-    for (const [, file] of [...sectionFiles, ...blockFiles, ...snippetFiles]) {
-      const stylesheet = getCompiledStylesheet(file)
-      if (stylesheet) stylesheets.push(stylesheet)
+function handleBlockScriptsJs(ctx: DevServerContext, event: H3Event, kind: 'block' | 'snippet' | 'section') {
+  const blockFiles = getLiquidFilesByKind(ctx, kind)
```
frandiox, line 135: Is `blockFiles` naming correct here? It can be anything, right? 🤔

packages/theme/src/cli/utilities/theme-environment/local-assets.ts
```diff
@@ -94,35 +94,143 @@ function isCompiledAssetRequest(event: H3Event) {
...
+       *    if (!__sections__['header'] && !window.DesignMode) return;
+       * ```
+       * However, this works well enough for most cases. It's preferable to have
+       * a consistent JavaScript workflow across sections, blocks, and snippets,
+       * especially when considering hot reloading. This only affects JavaScript
+       * sections in the theme editor and doesn't impact the main runtime.
+       */
+      javascripts.push(`
+        (function () {
+          if (!__${kinds}__["${baseName}"] && !Shopify.designMode) return;
```
frandiox, line 175: So this doesn't run in OSE?

packages/theme/src/cli/utilities/theme-environment/local-assets.ts
```diff
@@ -94,35 +94,143 @@ function isCompiledAssetRequest(event: H3Event) {
...
+      return key.endsWith('.liquid') && key.startsWith(`${kind}s/`)
+    })
+    .sort(([key1], [key2]) => {
+      return key1.localeCompare(key2)
     })
-  }
 }
 
-const compiledStylesheets = new Map<string, {checksum: string; content: string}>()
+function getCompiledJavaScript(file: ThemeAsset) {
```
frandiox, line 206: Not a big issue but since this function is virtually the same as `getCompiledStylesheet`, perhaps we can extract it in `getCompiledContent(file: ThemeAsset, tag: 'javascript' | 'stylesheet', cache: Map<string, {checksum: string, content: string}>)`?

frandiox (commented):
Looking good 🔥

packages/theme/src/cli/utilities/theme-environment/local-assets.ts
karreiro: Actually, it always runs on OSE (I also don't like this condition so much, but I'm trying to be as accurate as possible)

packages/theme/src/cli/utilities/theme-environment/local-assets.ts
karreiro: Oops! Excellent catch! I've applied that :)

packages/theme/src/cli/utilities/theme-environment/local-assets.ts
karreiro: Great catch! Fixed :)

packages/theme/src/cli/utilities/theme-environment/local-assets.ts
```diff
@@ -87,64 +92,160 @@ function findLocalFile(event: H3Event, ctx: DevServerContext) {
...
+
+  switch (assetPath) {
+    case 'styles.css':
+      return handleStylesCss(ctx, event)
+    case 'block-scripts.js':
+      return handleBlockScriptsJs(ctx, event, 'block')
+    case 'snippet-scripts.js':
+      return handleBlockScriptsJs(ctx, event, 'snippet')
+    case 'section-scripts.js':
+      return handleBlockScriptsJs(ctx, event, 'section')
```
frandiox, line 129: I think this one is called just `scripts.js`? It currently goes to the proxy instead of being generated locally:

image

packages/theme/src/cli/utilities/theme-environment/local-assets.ts
```diff
@@ -70,7 +75,7 @@ function findLocalFile(event: H3Event, ctx: DevServerContext) {
       const file = fileSystem.files.get(fileKey)
       const isUnsynced = fileSystem.unsyncedFileKeys.has(fileKey)
 
-      if (file || isUnsynced) {
+      if (file ?? isUnsynced) {
```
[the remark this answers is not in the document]
karreiro: Yes, I don't like that very much either, so I've explicitly added an ignore rule to _lint the linter_ itself, then we can consider removing this rule from the CLI if we notice it happening in other places as well :)

packages/theme/src/cli/utilities/theme-environment/local-assets.ts
```diff
@@ -87,64 +92,160 @@ function findLocalFile(event: H3Event, ctx: DevServerContext) {
...
 }
 
 function handleCompiledAssetRequest(event: H3Event, ctx: DevServerContext) {
-  const assetPath = parseServerEvent(event).pathname.split('/').at(-1)
+  if (!isCompiledAssetRequest(event)) return
+
+  const pathname = parseServerEvent(event).pathname
+  const assetPath = pathname.split('/').at(-1)
+
+  // isKnownCompiledAssetType already guarantees assetPath is defined.
```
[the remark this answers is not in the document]
karreiro: Removed with the refactor :)

packages/theme/src/cli/utilities/theme-environment/local-assets.ts
```diff
@@ -87,64 +92,160 @@ function findLocalFile(event: H3Event, ctx: DevServerContext) {
...
+function handleBlockScriptsJs(ctx: DevServerContext, event: H3Event, kind: 'block' | 'snippet' | 'section') {
+  const liquidFiles = getLiquidFilesByKind(ctx, kind)
+  const kinds = `${kind}s`
+
+  const javascripts = [
+    `
+      /*** GENERATED LOCALLY ***/
+
+      (function () {
+        var __${kinds}__ = {};
```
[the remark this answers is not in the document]
karreiro: Personally, I'd prefer to keep this part as is. This way, the file remains more accurate if theme developers open the local asset in their browsers. Although it might look slightly less neat here, it will be more readable and helpful for developers :)

frandiox (approved):
Thanks! LGTM 🔥

Issue #5653, opened 2025-04-15
[Bug]: `app deploy --reset` no longer resets toml file configuration


edhgoose:
### Please confirm that you have:

- [x] Searched [existing issues](.) to see if your issue is a duplicate. (If you’ve found a duplicate issue, feel free to add additional information in a comment on it.)
- [x] Reproduced the issue in [the latest CLI version](https://www.npmjs.com/package/@shopify/cli).

### In which of these areas are you experiencing a problem?

Extension

### Expected behavior

In version 3.74.0 of the Shopify CLI extension, if I ran `yarn run deploy --reset`, I'd be taken through the following prompts:

```
?  Which organization is this work for?
✔  My Company

?  Create this project as a new app on Shopify?
✔  Yes, create it as a new app

?  App name:
✔  Test

?  Configuration file name:
>  Test█
   ▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔
   shopify.app.test.toml will be generated in your root directory
```

Note that the final step here is it is asking if I'd like to create a new toml file, which is what I'd like to do.

### Actual behavior

In version 3.78 however, note the prompts now look like this with the same command (`app deploy --reset`)

```
?  Which organization is this work for?
✔  My Company

?  Create this project as a new app on Shopify?
✔  Yes, create it as a new app

?  App name:
✔  Test

╭─ success ──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│                                                                                                                                            │
│  shopify.app.existing-file.toml is now linked to "Test" on Shopify                                                                              │
│                                                                                                                                            │
│  Using shopify.app.existing-file.toml as your default config.                                                                                   │
│                                                                                                                                            │
│  Next steps                                                                                                                                │
│    • Make updates to shopify.app.existing-file.toml in your local project                                                                       │
│    • To upload your config, run `npm run shopify app deploy`                                                                               │
│                                                                                                                                            │
│  Reference                                                                                                                                 │
│    • App configuration                                                                                                                     │
│                                                                                                                                            │
╰────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
```

### Verbose output

[...]

PR #5676, opened 2025-04-22
Don't use cached config name when using `--reset`

isaacroldan:

When forcing a relink of an app, the system was still using the cached configuration file name, which made the overwrite the existing config.

Fixes https://github.com/Shopify/cli/issues/5653

Modifies the `linkedAppContext` function to set the `configName` to `undefined` when `forceRelink` is true, allowing users to be prompted for a new configuration file name instead of using the cached one.

### How to test your changes?

1. Run a command with `--reset`
2. Verify that you are prompted to select a configuration file name
3. Confirm that the new configuration file is used after relinking

### Measuring impact

### Checklist

### DIFF
--- a/.changeset/thirty-kings-add.md
+++ b/.changeset/thirty-kings-add.md
@@ -1,0 +1,5 @@
+---
+'@shopify/app': patch
+---
+
+Fix issue when using `--reset` not creating a new config toml file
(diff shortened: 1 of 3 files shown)


isaacroldan:
I've updated an existing test to check that we don't pass the cached config name.  Since `link` is mocked, what other test do you think I could add? or should I try not to mock link?

PR #5705, opened 2025-04-25
Use localhost ga

byrichardpowell:

We want to GA localhost development.

1. Make `--use-localhost` and `--localhost-port` flags public
2. No longer show a seperate renderInfo when either of these flags are present.  Add a message to the info box we show when starting dev instead: 
  
  Screenshot 2025-04-25 at 10 13 02 AM

3. Change how formInfoBox works (See commit message)
4. Move formInfoBox into its own file, rename and add tests.

### How to test your changes?

Main workflows:

1. `pnpm shopify app dev --use-localhost`: The old renderInfo box should be gone.  The app info should contain a message about localhost (see screenshot above).  The content should be sane
2. `pnpm shopify app dev --localhost-port=7891`: The old renderInfo box should be gone.  The app info should contain a message about localhost (see screenshot above).  The content should be sane

Now test I didn't break anything with the other workflow that uses render `formInfoBox()`, now `formatConfigInfoBody()`

1. `pnpm shopify app logs`

### Post-release steps

@nickwesselman will take on doc updates

### Measuring impact

### Checklist


packages/app/src/cli/commands/app/dev.ts
```diff
@@ -87,15 +87,15 @@ If you're using the Ruby app template, then you need to complete the following s
       exclusive: ['tunnel'],
     }),
     'use-localhost': Flags.boolean({
-      hidden: true,
+      hidden: false,
```
isaacroldan, line 90: false is the default value, you can remove the hidden property entirely :)

packages/app/src/cli/services/dev/tunnel-mode.ts
```diff
@@ -64,19 +63,6 @@ export async function getTunnelMode({
     requestedPort,
     actualPort,
     provideCertificate: async (appDirectory) => {
```
isaacroldan, line 65: We don't really need this to be part of the NoTunnel interface no? It's only used in one place and after validating that we are using localhost.

On line 346 of `services/dev.ts` you can just call `await generateCertificate` directly no?

byrichardpowell: Heh, yep/. Updated 👍  Thanks Isaac

packages/app/src/cli/commands/app/dev.ts
byrichardpowell: Removed.  Also removed it everywhere else we have hidden:false for the dev command flags

packages/app/src/cli/services/dev.ts
```diff
@@ -341,7 +344,11 @@ async function setupNetworkingOptions(
 
   let reverseProxyCert
   if (tunnelOptions.mode === 'use-localhost') {
-    const {keyContent, certContent, certPath} = await tunnelOptions.provideCertificate(appDirectory)
+    const {keyContent, certContent, certPath} = await generateCertificate({
+      appDirectory,
+      onRequiresConfirmation: generateCertificatePrompt,
+    })
```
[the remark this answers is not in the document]
byrichardpowell: Yeah, good point. Sorry I missed this 👍

isaacroldan (approved):

PR #5711, opened 2025-04-25
wip custom network settings for theme commands

EvilGenius13:

Fixes #0000 

### Measuring impact

### Checklist


packages/cli-kit/src/public/node/themes/api.test.ts
```diff
@@ -35,6 +35,12 @@ vi.stubGlobal('fetch', vi.fn())
 const session = {token: 'token', storeFqdn: 'my-shop.myshopify.com', refresh: async () => {}}
 const themeAccessSession = {...session, token: 'shptka_token'}
 const sessions = {CLI: session, 'Theme Access': themeAccessSession}
+const expectedApiOptions = expect.objectContaining({
+  maxRetryTimeMs: 45000,
+  timeoutMs: 45000,
+  useAbortSignal: true,
+  useNetworkLevelRetry: true,
```
EvilGenius13, line 42: I don't believe we were retrying requests before the addition of controllable request behaviours. I left the retry active, and the hard timeout and max retry time (total time of all retries) are set to 45 seconds.

packages/cli-kit/src/public/node/themes/api.ts
```diff
@@ -169,10 +176,17 @@ export async function deleteThemeAssets(id: number, filenames: Key[], session: A
   for (let i = 0; i < filenames.length; i += batchSize) {
     const batch = filenames.slice(i, i + batchSize)
     // eslint-disable-next-line no-await-in-loop
-    const {themeFilesDelete} = await adminRequestDoc(ThemeFilesDelete, session, {
-      themeId: composeThemeGid(id),
-      files: batch,
-    })
+    const {themeFilesDelete} = await adminRequestDoc(
```
EvilGenius13, line 179: I only added the custom network behaviour to api requests dealing with uploading/downloading a theme. Requests that should not take as long as file uploads such as password, metadata and metafield requests remain untouched. We can add custom options to others if we ever need

packages/cli-kit/src/public/node/themes/api.test.ts
```diff
@@ -429,7 +477,13 @@ describe('bulkUploadThemeAssets', async () => {
     const bulkUploadresults = await bulkUploadThemeAssets(id, assets, session)
 
     // Then
-    expect(adminRequestDoc).toHaveBeenCalledWith(ThemeFilesUpsert, session, {
+    const calls = vi.mocked(adminRequestDoc).mock.calls
```
EvilGenius13, line 480: Note: The standard `toHaveBeenCalledWith` assertion unexpectedly failed for the `bulkUploadThemeAssets` tests, potentially due to how the mock recorded the
arguments when multiple optional arguments (`undefined`, `undefined`, `object`) were present.
To ensure correctness, we explicitly check each argument received by the mock individually.

packages/cli-kit/src/public/node/themes/api.test.ts
```diff
@@ -35,6 +35,12 @@ vi.stubGlobal('fetch', vi.fn())
 const session = {token: 'token', storeFqdn: 'my-shop.myshopify.com', refresh: async () => {}}
 const themeAccessSession = {...session, token: 'shptka_token'}
 const sessions = {CLI: session, 'Theme Access': themeAccessSession}
+const expectedApiOptions = expect.objectContaining({
```
jamesmengo, line 38: nit: let's keep the naming for this consistent with the actual argument naming (e.g. requestBehaviourOptions)? LMK If I'm missing something though

.changeset/mighty-horses-fetch.md
```diff
@@ -0,0 +1,5 @@
+---
+'@shopify/cli-kit': patch
+---
+
+Add custom controllable request behaviour for theme related commands
```
jamesmengo, line 5: mega nit: Would love a mention of the problem we are fixing here. This almost implies that users can now control request behaviour

jamesmengo:
Can we loop in @shauns to this for visibility + an extra set of eyes?

packages/cli-kit/src/public/node/themes/api.test.ts
jamesmengo: Can you elaborate / try again? This works locally for me

packages/cli-kit/src/public/node/api/admin.ts
```diff
@@ -50,6 +51,7 @@ export async function adminRequestDoc<TResult, TVariables extends Variables>(
   variables?: TVariables,
   version?: string,
   responseOptions?: GraphQLResponseOptions<TResult>,
+  requestBehaviour?: RequestModeInput,
```
shauns, line 54: @isaacroldan did you add something like this in another PR?

packages/cli-kit/src/public/node/themes/api.ts
```diff
@@ -30,6 +31,12 @@ import {outputDebug} from '@shopify/cli-kit/node/output'
 export type ThemeParams = Partial<Pick<Theme, 'name' | 'role' | 'processing' | 'src'>>
 export type AssetParams = Pick<ThemeAsset, 'key'> & Partial<Pick<ThemeAsset, 'value' | 'attachment'>>
 const SkeletonThemeCdn = 'https://cdn.shopify.com/static/online-store/theme-skeleton.zip'
+const THEME_API_NETWORK_BEHAVIOUR: RequestModeInput = {
```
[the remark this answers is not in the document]
shauns: The effect of these will be:
- if there's a network level issue (e.g. DNS, connectivity, timeout -- anything that means the HTTP request does not complete), the CLI will retry
- it'll spend no longer than 45 seconds in total in a retry loop
- for a given request, the timeout limit is 45s, after that, the request will be stopped

Given that upload/download for themes may take a long time -- maybe the limit should be higher? If you have a limit for the retries but no limit for timeout, then it will retry for 45s if there is a connectivity issue, but once it connects it will try to complete the request for an unbounded period -- the retry limit kicks in when the retry happens, it doesn't interrupt anything.

Ultimately up to you guys.

EvilGenius13:
Thanks for chiming in @shauns & @isaacroldan! I'll take a look at the other PR, do some edits and circle back.

karreiro:
@shauns or @isaacroldan, could you please take a look at this when you have a moment?

isaacroldan (approved):
90 second timeout is ok, you could consider disabling the timeout for some specific queries if you feel they could need longer (like theme publish maybe¿)

But that's something that can be iterated on

shauns (approved):

PR #5714, opened 2025-04-28
Disable timeout for deploy requests

isaacroldan:

Fixes https://github.com/Shopify/cli/pull/5714

Improve handling of slow requests in the app management and partners clients by adding support for specifying request behavior.

This PR adds the ability to specify a `preferredBehaviour` parameter for GraphQL requests in the app management and partners clients. Specifically:

1. Adds a `preferredBehaviour` parameter to `appManagementRequestDoc`, `partnersRequest`  and `partnersRequestDoc` functions
3. Updates the app deploy requests to use the 'slow-request' behavior

These changes allow certain operations that are known to be slow (like app deploy) to use appropriate request behavior settings.

### How to test your changes?

1. Deploy an app using `shopify app deploy`
2. Verify that the deployment completes successfully without timing out
3. Create a new app version and verify it completes without timing out

### Measuring impact

### Checklist

### DIFF
--- a/packages/cli-kit/src/public/node/api/app-management.ts
+++ b/packages/cli-kit/src/public/node/api/app-management.ts
@@ -45,6 +45,10 @@ export const appManagementAppLogsUrl = async (
   return addCursorAndFiltersToAppLogsUrl(url, cursor, filters)
 }
 
+export interface RequestOptions {
+  requestMode: RequestModeInput
+}
+
 /**
  * Executes an org-scoped GraphQL query against the App Management API. Uses typed documents.
  *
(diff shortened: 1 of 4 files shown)


gonzaloriestra (approved):

packages/cli-kit/src/public/node/api/app-management.ts
```diff
@@ -60,6 +62,7 @@ export async function appManagementRequestDoc<TResult, TVariables extends Variab
   token: string,
   variables?: TVariables,
   cacheOptions?: CacheOptions,
+  preferredBehaviour?: RequestModeInput,
```
amcaplan, line 65: I don't love that we already have a few optional positional parameters here, let's not add more. Also I don't know if `preferredBehavior` is as descriptive as I'd want. Preferred behavior regarding what exactly?

isaacroldan: Agree with the naming 👌 
I'm ok with using 
```
requestOptions?: {
    requestMode?: RequestModeInput,
  }
```
too, but maybe we should refactor `appManagementRequestDoc` to receive a single object instead of positional arguments (that can be done in a different PR to not block this)

amcaplan (approved):

PR #5740, opened 2025-05-02
Automatically retry partners requests that fail with 401

MitchDickinson:

Fixes #0000 

### Measuring impact

### Checklist

### DIFF
--- a/packages/app/src/cli/utilities/developer-platform-client/partners-client.test.ts
+++ b/packages/app/src/cli/utilities/developer-platform-client/partners-client.test.ts
@@ -102,7 +102,14 @@ describe('createApp', () => {
 
     // Then
     expect(got).toEqual({...APP1, newApp: true, developerPlatformClient: partnersClient})
-    expect(partnersRequest).toHaveBeenCalledWith(CreateAppQuery, 'token', variables, undefined, undefined)
+    expect(partnersRequest).toHaveBeenCalledWith(
+      CreateAppQuery,
+      'token',
+      variables,
+      undefined,
+      undefined,
+      expect.any(Function),
+    )
   })
 
   test('creates an app with non-launchable defaults', async () => {
@@ -131,7 +138,14 @@ describe('createApp', () => {
 
     // Then
     expect(got).toEqual({...APP1, newApp: true, developerPlatformClient: partnersClient})
-    expect(partnersRequest).toHaveBeenCalledWith(CreateAppQuery, 'token', variables, undefined, undefined)
+    expect(partnersRequest).toHaveBeenCalledWith(
+      CreateAppQuery,
+      'token',
+      variables,
+      undefined,
+      undefined,
+      expect.any(Function),
+    )
   })
 
   test('throws error if requests has a user error', async () => {
@@ -162,7 +176,14 @@ describe('fetchApp', async () => {
 
     // Then
     expect(got).toEqual({organization: partnerMarkedOrg, apps: [APP1, APP2], hasMorePages: false})
-    expect(partnersRequest).toHaveBeenCalledWith(FindOrganizationQuery, 'token', {id: ORG1.id}, undefined, undefined)
+    expect(partnersRequest).toHaveBeenCalledWith(
+      FindOrganizationQuery,
+      'token',
+      {id: ORG1.id},
+      undefined,
+      undefined,
+      expect.any(Function),
+    )
   })
 
   test('throws if there are no organizations', async () => {
@@ -175,6 +196,13 @@ describe('fetchApp', async () => {
 
     // Then
     await expect(got).rejects.toThrowError(new NoOrgError(testPartnersUserSession.accountInfo))
-    expect(partnersRequest).toHaveBeenCalledWith(FindOrganizationQuery, 'token', {id: ORG1.id}, undefined, undefined)
+    expect(partnersRequest).toHaveBeenCalledWith(
+      FindOrganizationQuery,
+      'token',
+      {id: ORG1.id},
+      undefined,
+      undefined,
+      expect.any(Function),
+    )
   })
 })
--- a/packages/app/src/cli/utilities/developer-platform-client/partners-client.ts
+++ b/packages/app/src/cli/utilities/developer-platform-client/partners-client.ts
@@ -159,7 +159,7 @@ import {TypedDocumentNode} from '@graphql-typed-document-node/core'
 import {isUnitTest} from '@shopify/cli-kit/node/context/local'
 import {AbortError} from '@shopify/cli-kit/node/error'
 import {generateFetchAppLogUrl, partnersRequest, partnersRequestDoc} from '@shopify/cli-kit/node/api/partners'
-import {CacheOptions, GraphQLVariables} from '@shopify/cli-kit/node/api/graphql'
+import {CacheOptions, GraphQLVariables, RefreshTokenOnAuthorizedResponse} from '@shopify/cli-kit/node/api/graphql'
 import {ensureAuthenticatedPartners} from '@shopify/cli-kit/node/session'
 import {partnersFqdn} from '@shopify/cli-kit/node/context/fqdn'
 import {RequestModeInput, Response, shopifyFetch} from '@shopify/cli-kit/node/http'
@@ -218,7 +218,9 @@ export class PartnersClient implements DeveloperPlatformClient {
   public readonly supportsDevSessions = false
   public readonly supportsStoreSearch = false
   public readonly organizationSource = OrganizationSource.Partners
+
   private _session: PartnersSession | undefined
+  private tokenRefreshInProgress: undefined | Promise<string>
 
   constructor(session?: PartnersSession) {
     this._session = session
@@ -248,14 +250,21 @@ export class PartnersClient implements DeveloperPlatformClient {
     cacheOptions?: CacheOptions,
     preferredBehaviour?: RequestModeInput,
   ): Promise<T> {
-    return partnersRequest(query, await this.token(), variables, cacheOptions, preferredBehaviour)
+    return partnersRequest(
+      query,
+      await this.token(),
+      variables,
+      cacheOptions,
+      preferredBehaviour,
+      this.createUnauthorizedHandler(),
+    )
   }
 
   async requestDoc<TResult, TVariables extends {[key: string]: unknown}>(
     document: TypedDocumentNode<TResult, TVariables>,
     variables?: TVariables,
   ): Promise<TResult> {
-    return partnersRequestDoc(document, await this.token(), variables)
+    return partnersRequestDoc(document, await this.token(), variables, undefined, this.createUnauthorizedHandler())
   }
 
   async token(): Promise<string> {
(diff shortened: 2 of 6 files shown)


MitchDickinson:
* **#5740** <a href="https://app.graphite.dev/github/pr/Shopify/cli/5740?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Shopify/cli/5740?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/app/src/cli/utilities/developer-platform-client/partners-client.ts
```diff
@@ -640,6 +649,24 @@ export class PartnersClient implements DeveloperPlatformClient {
...
+
+  private createUnauthorizedHandler(): () => RefreshTokenOnAuthorizedResponse {
+    let unauthorisedRetriesUsed = 0
+    return async () => {
+      unauthorisedRetriesUsed++
+      if (unauthorisedRetriesUsed > 1) {
+        return {token: undefined}
+      }
+      if (this.tokenRefreshInProgress) {
+        throw new Error('Multiple simultaneous token refresh attempts are not allowed')
```
MitchDickinson, line 661: In @shauns's PR this wasn't raising, but I figured this was not an expected usage of the partner client and raising seemed safer + easier to test...but feedback welcome.

packages/app/src/cli/utilities/developer-platform-client/partners-client.test.ts
```diff
@@ -175,6 +197,85 @@ describe('fetchApp', async () => {
...
+
+    expect(refreshedToken).toBe(undefined)
+    expect(partnersClient.refreshToken).toHaveBeenCalledOnce()
+    expect(partnersRequestMock).toHaveBeenCalledTimes(1)
+  })
+
+  test('if multiple calls to partnerClient are made while token refresh is happening, the second call fails', async () => {
+    let refreshedToken: string | undefined
+    ;(partnersClient.refreshToken as Mock).mockImplementationOnce(async () => {
+      // Just sleep so the second call immediately trips the race condition guard clause
```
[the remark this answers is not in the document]
MitchDickinson: Maybe this test should be deleted since I just added a unit test for this condition? Don't want to introduce future flaky tests...

packages/app/src/cli/utilities/developer-platform-client.ts
```diff
@@ -295,3 +298,26 @@ export interface DeveloperPlatformClient {
...
+
+export function createUnauthorizedHandler(client: DeveloperPlatformClient): UnauthorizedHandler {
+  let unauthorizedRetriesUsed = 0
+  return {
+    type: 'token_refresh',
+    handler: async () => {
+      unauthorizedRetriesUsed++
+      if (unauthorizedRetriesUsed > 1) {
+        return {token: undefined}
+      } else if (client.getTokenRefreshInProgress()) {
```
shauns, line 310: I think this is not quite right. We do make simultaneous API calls -- not often, but it can happen, esp. for Partners API.

I'd expect request `A` and `B` to start, then `A` returns with a 401 and we begin the cycle of refreshing. `B` may also return with a 401, and also want to refresh. We would want B to therefore `await` for A's refresh cycle to complete.

In the prototype we did this by storing a nullable promise on the platform client class representing the refresh cycle. If a request needs to refresh and the promise is unset, set it to a promise resolving to the updated token, and then await that promise's conclusion. If a request needs to refresh and the promise is already set, await it. By making it a property of the client class, its essential a singleton value.

packages/cli-kit/src/public/node/api/graphql.ts
```diff
@@ -114,13 +146,52 @@ async function performGraphQLRequest<TResult>(options: PerformGraphQLRequestOpti
...
+        response = await retryAwareRequest(
+          {request: performRequest, url, ...requestBehaviour},
+          responseOptions?.handleErrors === false ? undefined : errorHandler(api),
+          simpleUnauthorizedHandler,
+        )
+      } catch (error) {
+        if (error instanceof ClientError && error.response.status === 401 && tokenRefreshUnauthorizedHandlerFunction) {
+          const shouldRetry = await tokenRefreshUnauthorizedHandlerFunction()
+          if (shouldRetry) {
+            response = await retryAwareRequest(
```
shauns, line 183: Can we combine this with the above, maybe a loop instead? It looks like they're separate because something different is happening, but they're intended to be the same.

packages/app/src/cli/utilities/developer-platform-client.ts
```diff
@@ -295,3 +298,26 @@ export interface DeveloperPlatformClient {
...
   getCreateDevStoreLink: (input: string) => Promise<string>
 }
+
+export function createUnauthorizedHandler(client: DeveloperPlatformClient): UnauthorizedHandler {
+  let unauthorizedRetriesUsed = 0
+  return {
+    type: 'token_refresh',
+    handler: async () => {
+      unauthorizedRetriesUsed++
+      if (unauthorizedRetriesUsed > 1) {
```
shauns, line 308: I think this is now covered in `graphql.ts` -- we only try to retry once per request

packages/app/src/cli/utilities/developer-platform-client.ts
shauns: Should add though that I like the idea of reusing this logic across the two clients. You could tweak so that there is global state like `{[client: DeveloperPlatformClient]: RefreshTokenState}`, where `type RefreshTokenState = undefined | Promise<string>`. Or have the currentlyRefreshingToken promise be part of the DeveloperPlatformClient interface. Maybe thats nicer

packages/cli-kit/src/public/node/api/graphql.test.ts
```diff
@@ -208,6 +242,77 @@ describe('graphqlRequest', () => {
       }
     `)
   })
+
+  test('Fails without retry when unauthorized and no token refresh function is provided', async () => {
```
shauns, line 246: These are really nice

packages/app/src/cli/utilities/developer-platform-client/partners-client.test.ts
```diff
@@ -175,6 +185,81 @@ describe('fetchApp', async () => {
...
     await expect(got).rejects.toThrowError(new NoOrgError(testPartnersUserSession.accountInfo))
-    expect(partnersRequest).toHaveBeenCalledWith(FindOrganizationQuery, 'token', {id: ORG1.id}, undefined, undefined)
+    expect(partnersRequest).toHaveBeenCalledWith(FindOrganizationQuery, 'token', {id: ORG1.id}, undefined, undefined, {
+      type: 'token_refresh',
+      handler: expect.any(Function),
+    })
+  })
+})
+
+describe('Token refresh', () => {
```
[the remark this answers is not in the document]
shauns: I think I'd drop them and focus on the platform client tests, imo. If we're making the logic DRY we can do the same for the tests too.

The client tests are not great right now -- they have to mock as its all about hitting the network but I think that it should be happening in a more structured way.

packages/app/src/cli/utilities/developer-platform-client/partners-client.test.ts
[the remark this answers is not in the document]
shauns: When these move to use mock-service-worker, we can use things like this https://mswjs.io/docs/api/delay

packages/app/src/cli/utilities/developer-platform-client/partners-client.test.ts
```diff
@@ -175,6 +185,81 @@ describe('fetchApp', async () => {
...
+      await new Promise((resolve) => setTimeout(resolve, 3000))
+      return newToken
+    })
+
+    const firstRequest = partnersClient.request(query, variables)
+    const secondRequest = partnersClient.request(query, variables)
+
+    await expect(secondRequest).rejects.toThrow('Multiple simultaneous token refresh attempts are not allowed')
+
+    expect(refreshedToken).toBe(undefined)
```
shauns, line 261: shouldn't this be set? should we be awaiting the first request's (successful) conclusion?

packages/app/src/cli/utilities/developer-platform-client/partners-client.test.ts
```diff
@@ -175,6 +185,81 @@ describe('fetchApp', async () => {
...
+    expect(partnersClient.refreshToken).toHaveBeenCalledOnce()
+    expect(partnersRequestMock).toHaveBeenCalledTimes(1)
+  })
+
+  test('if multiple calls to partnerClient are made while token refresh is happening, the second call fails', async () => {
+    let refreshedToken: string | undefined
+    ;(partnersClient.refreshToken as Mock).mockImplementationOnce(async () => {
+      // Just sleep so the second call immediately trips the race condition guard clause
+      await new Promise((resolve) => setTimeout(resolve, 3000))
```
shauns, line 252: In terms of correctness of sleeping, answer is: it depends. Actually sleeping for a while -- i.e ending up with a slow test -- is not ideal. We should only reach for that if truly needed.

Starting to sleep but not actually finishing the sleep, or doing a sleep that's pretty short -- generally no problem. Ultimately the test speed is what matters and these both round down to fast. If doing that to try to fudge some non-deterministic behaviour, that's not good.

If you need to test something that involves a noticeable passage of time, vitest's timer mocking is best: https://vitest.dev/guide/mocking.html#timers 

I think in this case, providing my comment below makes sense, it looks like a case for timer mocking.

shauns (changes requested):
I think low-level piece is looking great, but the piece about how the platform clients use this capability would benefit from changes.

MitchDickinson (commented):
@shauns Mind having another look now? I think I've addressed the feedback.

Another thing I don't love is that this introduces three new methods to `DeveloperPlatformClient` when the usage of those methods doesn't really need to be _publicly_ exposed. Ideally the implementation of `createUnauthorizedHandler` could live "directly" in the interface, kind of like how in Ruby you can have implementation code (and even state) live in mixins. These methods are just implementation details of the unauthorized handler.

I have a local commit I'm working on that attempts to fix this, but I _think_ it requires adding a new base class (i.e. `abstract class AbstractDeveloperPlatformClient`) that can hold the `currentlyRefreshingToken` state and then can encapsulate the logic in `createUnauthorizedHandler`. 

```ts
export abstract class AbstractDeveloperPlatformClient {
  private currentlyRefreshingToken?: Promise<string>

  protected abstract newToken(): Promise<string>

  protected unauthorizedHandler(): UnauthorizedHandler {
    return {
      type: 'token_refresh',
      handler: async () => {
        return {token: await this.refreshToken()}
      },
    }
  }

  private async refreshToken(): Promise<string> {
    if (this.currentlyRefreshingToken) {
      return this.currentlyRefreshingToken
    }

    try {
      this.currentlyRefreshingToken = this.newToken()
      return await this.currentlyRefreshingToken
    } finally {
      this.currentlyRefreshingToken = undefined
    }
  }
}
```

I'm not sure this is much better - it lets us drop the three methods from the DeveloperPlatformClient interface but it adds another base class that each client must extend. 

Do you have a sense which approach would be better?

shauns (approved):
This looks good! I left a note on your idea about better organising the logic under the client/its interface, but I don't see that as blocking as (1) its entirely internal, (2) its verbose but not hard to understand and (3) the potential value in getting this out.

PR #5761, opened 2025-05-07
Fix error handling for misconfigured multi-extension config

shauns:

Fixes bugsnag error 678f6e234714603e798bf36a

Instead of using `parse`, which can throw, use `safeParse` and the existing `abortOrReport` mechanism when loading extension config.

### DIFF
--- a/.changeset/metal-worms-check.md
+++ b/.changeset/metal-worms-check.md
@@ -1,0 +1,5 @@
+---
+'@shopify/app': patch
+---
+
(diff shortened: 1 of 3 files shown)


shauns:
### Merge activity

* **May 7, 10:09 AM EDT**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/Shopify/cli/5761).

PR #5764, opened 2025-05-07
Add trampoline run on wasm modules as part of function build

saga-dasgupta:
@lopert had most of the work done, I added some tests, cleanup and fixed the download package path.

Fixes https://github.com/shop/issues-shopifyvm/issues/28

We need to apply the trampolining process to guest modules using the Wasm API to allow for it to copy data between the memory of the provider and guest modules.

Downloading the trampoline release from github and run trampoline with input as function.wasm file during function build. 

### How to test your changes?

` pnpm run shopify app function build --path <full path to your functions app> --verbose`
verbose let me see that all the download and apply trampoline commands succeeded. 

### Measuring impact

### Checklist


adampetro (approved):
Tested out locally and it works great!

packages/app/src/cli/services/function/binaries.test.ts
```diff
@@ -313,3 +320,41 @@ describe('wasm-opt', () => {
...
+    }
+  })
+
+  test('downloadUrl returns the correct URL', () => {
+    // When
+    const url = trampoline.downloadUrl('darwin', 'x64')
+
+    // Then
+    expect(url).toMatch(
+      /https:\/\/github.com\/Shopify\/shopify-function-wasm-api\/releases\/download\/shopify_function_trampoline\/v1.0.0\/shopify-function-trampoline-x86_64-macos-v\d\.\d\.\d.gz/,
```
andrewhassan, line 343: This assertion will fail when we update the trampoline version. Do you think we should have it use `TRAMPOLINE_VERSION` in there to ensure it stays in sync automatically?

packages/app/src/cli/services/function/build.ts
```diff
@@ -242,6 +242,18 @@ export async function runWasmOpt(modulePath: string) {
...
 
+export async function runTrampoline(modulePath: string) {
+  const trampoline = trampolineBinary()
+  await downloadBinary(trampoline)
+
+  const command = trampoline.path
+
+  const args = ['-i', modulePath, '-o', modulePath]
+  outputDebug(`command: ${command}`)
+  outputDebug('Applying trampoline to the wasm binary.')
```
andrewhassan, line 253: Nit: These could be combined into a single debug output. It would probably also be useful debugging output to include the arguments here too.

andrewhassan (approved):
Code LGTM and 🎩

PR #5768, opened 2025-05-08
Fix an issue with hot-reloading the compiled assets (scripts.js, block-scripts.js, etc)

karreiro:

This PR fixes an issue related to hot-reloading compiled assets (such as scripts.js, block-scripts.js, etc).

This PR enriches the hot-reload event's payload by including details indicating whether the `javascript` or `stylesheet` tags were modified. This approach triggers a full-reload event the first time users update a section file with a `javascript` tag, but it caches this information, and subsequent updates will only trigger a full reload if the content of the `javascript` tag has changed.

### How to test your changes?

- Update your local `theme-hot-reload` package to use [this branch](https://github.com/Shopify/theme-hot-reload/pull/41)
- Run `SHOPIFY_CLI_LOCAL_HOT_RELOAD=true shopify-dev theme dev`
- Change the {% javavascript %} content within a section file
- Verify that the browser correctly performs a full refresh

Check the demo [here](https://github.com/Shopify/theme-hot-reload/pull/41)

### Post-release steps

N/A

### Measuring impact

### Checklist

### DIFF
--- a/.changeset/orange-bugs-retire.md
+++ b/.changeset/orange-bugs-retire.md
@@ -1,0 +1,5 @@
+---
+'@shopify/theme': patch
+---
+
+Fix an issue with hot-reloading the compiled assets (scripts.js, block-scripts.js, etc)
--- a/packages/theme/package.json
+++ b/packages/theme/package.json
@@ -51,7 +51,7 @@
     "yaml": "2.7.0"
   },
   "devDependencies": {
-    "@shopify/theme-hot-reload": "^0.0.11",
+    "@shopify/theme-hot-reload": "^0.0.18",
     "@vitest/coverage-istanbul": "^1.6.0",
     "node-stream-zip": "^1.15.0"
   },
--- a/packages/theme/src/cli/utilities/theme-environment/hot-reload/server.test.ts
+++ b/packages/theme/src/cli/utilities/theme-environment/hot-reload/server.test.ts
@@ -91,7 +91,7 @@ describe('Hot Reload', () => {
 
       const expectedSectionEvent = `data: {"sync":"local","themeId":"${THEME_ID}","type":"update","key":"${testSectionFileKey}","payload":{"sectionNames":["first","second"],"replaceTemplates":${JSON.stringify(
         getInMemoryTemplates(ctx),
-      )},"liquidTagsModified":{"stylesheet":false,"javascript":false}},"version":"${HOT_RELOAD_VERSION}"}`
+      )},"updatedFileParts":{"stylesheetTag":false,"javascriptTag":false}},"version":"${HOT_RELOAD_VERSION}"}`
 
       // Verify local sync event
       expect(hotReloadEvents.at(-1)).toMatch(expectedSectionEvent)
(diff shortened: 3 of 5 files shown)


frandiox (commented):
Hey, thanks so much for the PR! Looks good as a starting point. I haven't had time to do a deep review but here are a couple of early comments:

- I thiiink this doesn't "warm" the cache? Meaning, it will think the tags are always "changed" the first time, right? Maybe we should integrate this cache with the themeFS? Or perhaps that's too much coupling for this code with the CLI...
- We would also need to check if the rest of the file (outside of the tags) is modified. The reason is, even if it says `stylesheet: true`, there might be another Liquid syntax change that needs a full refresh. So basically we should get `{liquid: boolean; stylesheet: boolean; javascript: boolean}` I think?

karreiro:
👋 Hey @frandiox,

Thank you for the review!

> I thiiink this doesn't "warm" the cache? Meaning, it will think the tags are always "changed" the first time, right? Maybe we should integrate this cache with the themeFS? Or perhaps that's too much coupling for this code with the CLI...

It intentionally doesn't. Given that this impacts only sections containing `{% javascript %}` tags and that those sections inevitably require a page reload, I leaned towards accepting this trade-off of triggering an initial full refresh in the first save of those files for simplicity sake.

> We would also need to check if the rest of the file (outside of the tags) is modified. The reason is, even if it says `stylesheet: true`, there might be another Liquid syntax change that needs a full refresh. So basically we should get `{liquid: boolean; stylesheet: boolean; javascript: boolean}` I think?

Actually, the mechanism to handle Liquid updates is already in place, so I don't think we need to introduce any additional logic for that scenario. Am I missing something?

> Ah I see this is supposed to be a fix for section hot reloading, rather than the full feature we talked about!

100%. While this PR enriches the event payload tracking stylesheet changes — currently we don't take any action upon stylesheet changes.

---

Thanks again for the review and for setting up the foundations of this hot-reload mechanism, @frandiox -- it was much more straightforward for me to walk over the Liquid AST by reusing the approaches we have in the local proxy :)

frandiox (approved):
>It intentionally doesn't. Given that this impacts only sections containing {% javascript %} tags and that those sections inevitably require a page reload, I leaned towards accepting this trade-off of triggering an initial full refresh in the first save of those files for simplicity sake.

Yeah makes sense if we just do `javascript` for now. We can rewrite the cache piece once we start acting on `stylesheet` changes 👍

>Actually, the mechanism to handle Liquid updates is already in place, so I don't think we need to introduce any additional logic for that scenario. Am I missing something?

I meant this for when we send `stylesheet: true, javascript: false`. In that case we would want to only update CSS but we need to ensure that the actual Liquid (section, snippet, etc.) didn't change.

Therefore, knowing we might need this information, I'd suggest we rename `liquidTagsModified` to `filePartsModified`, `liquidUpdateDetails` or something like that? That way we can also add `liquid: Boolean` eventually (since liquid is not a "tag" per se).

Example:
- `{stylesheet: true, javascript: false, liquid: false}` => Reload only compiled CSS file
- `{stylesheet: true, javascript: true, liquid: false}` => full reload
- `{stylesheet: true, javascript: false, liquid: true}` => full reload

Eventually we might add `schema` as well? That could be useful for OSE 🤔 

---

In any case, with the change or without it we can ship this for fixing the sections hot reload. Thanks a lot! 🚀

PR #5775, opened 2025-05-08
Use shop.dev, not my.shop.dev, for shopify

bbraschi:
Won's ship - would need instead an update of `vendors`


bbraschi:
Not yet ready to be discarded

isaacroldan:
The original dev-server package still has the `my.shop.dev` value. Is there a plan to update it there?

bbraschi:
@isaacroldan I am not sure and I did not spend the time to bug the team in charge of `DevServer` because the CLI started to work, [for me], out of the box, without this patch.

bbraschi:
Closed as I did not experience problems in local dev recently

PR #5790, opened 2025-05-13
[WIP] Conditionally derive JS binary dependencies

saulecabrera:

Fixes #0000 

### Measuring impact

### Checklist

### DIFF
--- a/packages/app/src/cli/services/function/binaries.ts
+++ b/packages/app/src/cli/services/function/binaries.ts
@@ -10,11 +10,13 @@ import fs from 'node:fs'
 import * as gzip from 'node:zlib'
 import {fileURLToPath} from 'node:url'
 
-const FUNCTION_RUNNER_VERSION = 'v7.0.1'
-const JAVY_VERSION = 'v4.0.0'
+export const PREFERRED_FUNCTION_RUNNER_VERSION = 'v8.0.1'
+
+// Javy dependencies.
+const PREFERRED_JAVY_VERSION = 'v5.0.3'
 // The Javy plugin version should match the plugin version used in the
 // function-runner version specified above.
-const JAVY_PLUGIN_VERSION = 'v1'
+const PREFERRED_JAVY_PLUGIN_VERSION = 'v2'
 
 const BINARYEN_VERSION = '123.0.0'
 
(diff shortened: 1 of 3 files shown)


saga-dasgupta (approved):
I was able to tophat building shopify functions that has version 1.0.0 and 2.0.0 functions package.

packages/app/src/cli/services/function/build.ts
```diff
@@ -256,10 +270,11 @@ export async function runTrampoline(modulePath: string) {
 export async function runJavy(
   fun: ExtensionInstance<FunctionConfigType>,
   options: JSFunctionBuildOptions,
+  binaryDeps: BinaryDependencies,
   extra: string[] = [],
 ) {
-  const javy = javyBinary()
-  const plugin = javyPluginBinary()
+  const javy = javyBinary(binaryDeps.javy)
+  const plugin = javyPluginBinary(binaryDeps.javyPlugin)
```
saga-dasgupta, line 277: Not strictly related to your changes but I wonder why we have download binary for both the install javy and run javy steps, like shouldn't we just have it for the installJavy step?

packages/app/src/cli/services/function/binaries.ts
```diff
@@ -29,6 +31,32 @@ interface DownloadableBinary {
...
+  javy: string
+  javyPlugin: string
+}
+
+// Derives the binary dependencies to be used with a particular
+// `@shopify/shopify_function` package version.
+export function deriveJavaScriptBinaryDependencies(version: string): BinaryDependencies | null {
+  if (version === '0' || version === '1') {
+    return {
+      functionRunner: 'v7.0.1',
```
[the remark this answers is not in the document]
lopert: Technically, yes. We could just use the latest function-runner since it has support for all plugin versions.

@saulecabrera and I discussed this briefly, but I think we settled on "eventually we'd like to clean up function runner to not include every plugin, so we should start including this mapping so that older versions of the CLI can still know which versions work with the extension's selected npm package."

packages/app/src/cli/services/function/build.ts
lopert: I am also not sure why we have both. installJavy seems to have been added as a workaround to some issue.
https://github.com/Shopify/cli/blob/1197a18678fff028c4861b9b609e16ea36bda9e6/packages/app/src/cli/services/build.ts#L25-L27

Didn't want to delay this by digging into that too much, so just made sure that it also installed all required versions of our binairies for now.

saga-dasgupta: Yeah thats fine!

lopert:
I have also just tophatted in prod by 
- deploying an app that had both the new/old extensions, 
- enabling them in the admin, 
- running through the checkout, and seeing the expected effect,
- and finally, seeing the resulting runs in my partners dash.

PR #5807, opened 2025-05-15
Detect shopify.app.toml is using correct schema when client_id is missing 

jamieguerrero:
Screenshot 2025-05-15 at 4 57 37 PM

Fixes https://github.com/Shopify/develop-app-inner-loop/issues/2237

Prior to this PR, we assumed that if we are missing the `client_id` attribute in the TOML, that means that we are loading a TOML that is shaped like the `LegacyAppSchema`. `client_id` is required in `AppSchema`, however we should be able to validate the TOML as an `AppSchema` with a `client_id` missing. 

We determine if the schema is `LegacyAppSchema` _first_ and after we have determined that the TOML is not `LegacyAppSchema`, we can evaluate the TOML more accurately against `AppSchema` and return a correct zod validation error. (The TOML is likely not a `LegacyAppSchema` and we should rarely ever be using this zod schema since we only use it when developers are loading an app they cloned from a template or a very very old TOML that doesn't match the newest `AppSchema` TOML). 

Note that [previously](https://github.com/Shopify/cli/pull/5542) we tried to provide a better error message if the LegacyAppSchema zod Schema wasn't validated correctly but abandoned this approach. This is now a more correct solution.

### How to test your changes?

- Create an app
- Remove the `client_id` from the TOML file
- Attempt to deploy
- View the improved error

### DIFF
--- a/packages/app/src/cli/models/app/loader.test.ts
+++ b/packages/app/src/cli/models/app/loader.test.ts
@@ -3255,6 +3255,21 @@ describe('getAppConfigurationState', () => {
       expect(state).toMatchObject(resultShouldContain)
     })
   })
+
+  test('raises validation error when AppSchema is missing client_id', async () => {
+    await inTemporaryDirectory(async (tmpDir) => {
+      // We know this is a TOML file that follows the AppSchema because
+      // it can contain extra fields (something_extra). In LegacyAppSchema, all fields are optional.
+      // This content is also missing a client_id field which should throw a validation error.
+      const content = `something_extra = "some_value"`
+      const appConfigPath = joinPath(tmpDir, 'shopify.app.toml')
+      const packageJsonPath = joinPath(tmpDir, 'package.json')
+      await writeFile(appConfigPath, content)
+      await writeFile(packageJsonPath, '{}')
+
+      await expect(getAppConfigurationState(tmpDir, undefined)).rejects.toThrowError(/client_id/)
+    })
+  })
 })
 
 describe('loadConfigForAppCreation', () => {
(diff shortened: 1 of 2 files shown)


packages/app/src/cli/models/app/loader.ts
```diff
@@ -931,13 +932,15 @@
     schemaForConfigurationFile = appSchema
   }
 
+  console.log('before')
```
amcaplan, line 935: These console.log statements need to be removed, I assume they were committed accidentally 😉

packages/app/src/cli/models/app/loader.test.ts
```diff
@@ -3255,6 +3255,21 @@ describe('getAppConfigurationState', () => {
...
+      // We know this is a TOML file that follows the AppSchema because
+      // it can contain extra fields (something_extra). In LegacyAppSchema, all fields are optional.
+      // This content is also missing a client_id field which should throw a validation error.
+      const content = `something_extra = "some_value"`
+      const appConfigPath = joinPath(tmpDir, 'shopify.app.toml')
+      const packageJsonPath = joinPath(tmpDir, 'package.json')
+      await writeFile(appConfigPath, content)
+      await writeFile(packageJsonPath, '{}')
+
+      await expect(getAppConfigurationState(tmpDir, undefined)).rejects.toThrowError(/Validation errors/)
```
amcaplan, line 3270: I think it would be more helpful to assert that the _correct_ validation error is thrown, i.e. that the error mentions `client_id` (rather than the previous behavior which mentioned a bunch of unrelated stuff).

amcaplan (changes requested):
Minor issues to fix/improve

jamieguerrero:
* **#5807** <a href="https://app.graphite.dev/github/pr/Shopify/cli/5807?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Shopify/cli/5807?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #5811, opened 2025-05-16
Render `%` in progress bar when executing `theme-download`

jamesmengo:
Fixes https://github.com/Shopify/developer-tools-team/issues/457

Renders a `%` rather than `1..50/200` in the progress bar when executing `theme pull`
- Instead of following the much more complicated approach with promises in `theme-uploader`, I didn't touch the way we batch files / jobs. I just updated the way we render progress to keep things simple. I **did** try that approach earlier - [link](https://github.com/Shopify/cli/pull/5809)

This brings things in line with `theme push`

https://github.com/user-attachments/assets/8281f441-be51-4264-8085-cb6dbea9cc5b

### How to test your changes?
1) Build this branch
2) Execute `shopify-dev theme pull` in an empty directory or by passing the `--path`flag
3) Verify that the progress is a `%`
### Measuring impact

### Checklist

### DIFF
--- a/packages/theme/src/cli/utilities/theme-downloader.ts
+++ b/packages/theme/src/cli/utilities/theme-downloader.ts
@@ -64,8 +64,14 @@ function buildDownloadTasks(
 
   const filenames = checksums.map((checksum) => checksum.key)
 
+  const getProgress = (params: {current: number; total: number}) =>
+    `[${Math.round((params.current / params.total) * 100)}%]`
+
   const batches = batchedTasks(filenames, MAX_GRAPHQL_THEME_FILES, (batchedFilenames, i) => {
-    const title = `Downloading files ${i}..${i + batchedFilenames.length} / ${filenames.length} files`
+    const title = `Downloading files from remote theme ${getProgress({
+      current: i,
+      total: filenames.length,
+    })}`
     return {
       title,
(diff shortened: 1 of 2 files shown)


packages/theme/src/cli/utilities/theme-downloader.ts
```diff
@@ -64,8 +64,14 @@ function buildDownloadTasks(
 
   const filenames = checksums.map((checksum) => checksum.key)
 
+  const getProgress = (params: {current: number; total: number}) =>
```
jamesmengo, line 67: copied: https://github.com/Shopify/cli/blob/jm/progress_percent_download/packages/theme/src/cli/utilities/theme-uploader.ts#L44-L45

Decided to just duplicate for the time being

aswamy (approved):

packages/theme/src/cli/utilities/theme-downloader.ts
aswamy: Is the plan to merge these approaches later on?

jamesmengo: I think the next time we touch / duplicate this code, certainly - though I don't have a plan to do that at the current moment in case the final abstraction needs to be more general - thoughts?

PR #5820, opened 2025-05-16
Invoke `wasm-opt` with  --enable-nontrapping-float-to-int

saulecabrera:

Rust 1.87 now uses LLVM 20, which in turn enables non-trapping floating point by default
https://releases.llvm.org/20.1.0/docs/ReleaseNotes.html#changes-to-the-webassembly-backend

Without this change, any Rust functions built with 1.87 will fail to optimize.

### Measuring impact

### Checklist


davejcameron (approved):
I tested this out by running it manually with wasm-opt on the latest version of Rust

saulecabrera:
> I assume this is safe also if using an older version of Rust?

Correct.

PR #5821, opened 2025-05-16
Hardcode extension only apps to be embedded = true to make the behaviour consistent with Partners ext-only apps

jamieguerrero:

Closes https://github.com/Shopify/develop-app-inner-loop/issues/2789

Currently, all Organization owned extension-only apps are being created without a `handle`. We theorize (but have not validated yet) that having a null handle will create issues with public distribution for Organization owned apps that handle their distribution in the Partners dashboard. This is meant to unblock that before Editions. 

We may have to create a migration to backfill `handle` for Organization only apps, but this PR will avoid adding any more ApiClients without a `handle`.

The ApiClient requires an app to be embedded for the persistence model to[ generate a `handle` on save](https://github.com/shop/world/blob/05-14-add_uniqueness_check_on_name_for_eventbus_integration/areas/core/shopify/components/apps/app/models/api_client/extension.rb#L87). 

```
has_handle :handle, from: :title, sticky: true, if: :embedded?
```

This PR hardcodes `embedded = true` for all apps created with the app management client in the CLI. 

### How to test your changes?

1. Create an ext-only app in a Developer Dashboard org
2. Confirm using `dev mysql open` that an api_client record was made and the row contains a non-nil `handle` value

Confirm the same thing with an ext-only app in a Partners org.
Also confirm the same with a Remix app in both Partners and Developer Dashboard.

### DIFF
--- a/packages/app/src/cli/utilities/developer-platform-client/app-management-client.ts
+++ b/packages/app/src/cli/utilities/developer-platform-client/app-management-client.ts
@@ -1130,7 +1130,7 @@ const MAGIC_URL = 'https://shopify.dev/apps/default-app-home'
 const MAGIC_REDIRECT_URL = 'https://shopify.dev/apps/default-app-home/api/auth'
 
 function createAppVars(options: CreateAppOptions, apiVersion?: string): CreateAppMutationVariables {
-  const {isLaunchable, scopesArray, name, isEmbedded} = options
+  const {isLaunchable, scopesArray, name} = options
   const source: AppVersionSource = {
     source: {
       name,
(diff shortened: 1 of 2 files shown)


packages/app/src/cli/utilities/developer-platform-client/app-management-client.ts
```diff
@@ -1139,7 +1139,10 @@ function createAppVars(options: CreateAppOptions, apiVersion?: string): CreateAp
           type: AppHomeSpecIdentifier,
           config: {
             app_url: isLaunchable ? 'https://example.com' : MAGIC_URL,
-            embedded: isEmbedded,
+            // Ext-only apps should be embedded = false, however we are hardcoding this to
+            // match Partners behaviour for now
+            // https://github.com/Shopify/develop-app-inner-loop/issues/2789
+            embedded: true,
```
isaacroldan, line 1145: Is this a temporary fix or is the intention to default to embedded always?
Is this only needed during app creation? what happens if this value is changed later?

jamieguerrero: > Is this only needed during app creation? what happens if this value is changed later?

Yes, this is only needed during app creation. Once that app has been initialized, it is created with a `handle`. If the embedded or handle values are changed later, the app cannot remove it's handle by changing the TOML configurations so we are unconcerned.

dmerand (approved):

PR #5831, opened 2025-05-18
[Feature] Improve the extensions build process to minify files for draft extensions

keiraarts:

Currently, extension files are only minified when a branch is released, but not during development. This becomes problematic because Shopify's GraphQL API imposes a 2048KB size limit for extension files. 

As a result, many draft extensions that are larger than 2MB when unminified cannot be deployed for preview, making it impossible to test them during development. While Shopify does bundle files below 2MB during an app release, having consistent behavior between development and release environments would improve the developer experience.

This PR modifies the extension build process to apply minification to extension files in development branches, matching the behavior of released branches. This ensures that:

1. Extensions can be previewed during development even when they exceed 2MB unminified
2. The behavior is consistent between development and release environments
3. Developers can accurately test their extensions as they would appear in production

### How to test your changes?

1. Clone this branch
2. Create a new Shopify app: `npm init @shopify/app@latest`
3. Add an extension that generates a large JavaScript bundle (e.g., a UI extension with many dependencies)
4. Start a development server: `shopify app dev`
5. Verify that the extension's distribution code (/dist) is minified.
6. Check the file size of the deployed extension to confirm it's been minified during the development preview.

### Other solutions

This could be provided by app developers as a CLI flag to minify their code alternatively.

### Measuring impact

### Checklist

### DIFF
--- a/packages/app/src/cli/services/dev/app-events/app-watcher-esbuild.ts
+++ b/packages/app/src/cli/services/dev/app-events/app-watcher-esbuild.ts
@@ -118,7 +118,7 @@ export class ESBuildContextManager {
 
   private async extensionEsBuildOptions(stdin: StdinOptions, outputPath: string) {
     return getESBuildOptions({
-      minify: false,
+      minify: true,
       outputPath,
       environment: 'development',
       env: {


kpeatt:
Please merge this — my development workflow for Customer Account and Checkout extensions broke because of this.

gonzaloriestra:
Thanks a lot for the PR! It makes sense, but as you mentioned, I think it's better to have a flag to keep the current behavior. I've created another PR [here](https://github.com/Shopify/cli/pull/5864). I hope you don't mind, but this way the CI also works as expected (there are some issues with forks).

PR #5861, opened 2025-05-21
Display an info message for dev dash organizations

nickwesselman:

Make it clearer to users why they are seeing new organizations in the org list.

### DIFF
--- a/packages/app/src/cli/prompts/dev.test.ts
+++ b/packages/app/src/cli/prompts/dev.test.ts
@@ -101,7 +101,7 @@ describe('selectOrganization', () => {
     expect(renderAutocompletePrompt).not.toBeCalled()
   })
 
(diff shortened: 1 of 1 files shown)


packages/app/src/cli/prompts/dev.test.ts
```diff
@@ -95,6 +100,67 @@ describe('selectOrganization', () => {
     expect(got).toEqual(ORG2)
     expect(renderAutocompletePrompt).not.toBeCalled()
   })
+
+  test('displays info banner if user has access to a Business Platform org', async () => {
```
gonzaloriestra, line 104:
```suggestion
  test('displays info banner if there are orgs from both Business Platform and Partners', async () => {
```

gonzaloriestra (approved):
Tested that it works as expected

PR #5874, opened 2025-05-21
Allow publish command to work with multiple environments

EvilGenius13:

We are changing the publish command to allow multiple environments to be specified in a single command to allow developers with multiple stores to work more efficiently.

Changing the publish command to use our multi environment infrastructure.

### How to test your changes?

- Pull down the branch
- Build the branch
- If you don't already have one, create a `shopify.theme.toml` file

*Note*: You will need to create an access token through your shop(s) to bypass the need to login during theme commands.
The most basic example of the structure needed would be this:
```
[environments.env1]
store = "myfirstshop.myshopify.com"
password = "access_token"
theme = "your-theme-to-publish"
force = "true"
[environments.env2]
store = "mysecondshop.myshopify.com"
password = "access_token"
theme = "your-theme-to-publish"
force= "true"
```

- Run `theme publish -e env1 -e env2` or whatever you named your environments.

- Look for the confirmation output for your stores in the terminal. The order may differ from your input (i.e you might have put env1, env2 but the output is env2, env1) due to the commands running concurrently.

### Measuring impact

### Checklist

### DIFF
--- a/packages/theme/src/cli/commands/theme/publish.ts
+++ b/packages/theme/src/cli/commands/theme/publish.ts
@@ -1,10 +1,12 @@
-import {ensureThemeStore} from '../../utilities/theme-store.js'
 import ThemeCommand from '../../utilities/theme-command.js'
 import {themeFlags} from '../../flags.js'
 import {publish, renderArgumentsWarning} from '../../services/publish.js'
 import {Flags} from '@oclif/core'
 import {globalFlags} from '@shopify/cli-kit/node/cli'
-import {ensureAuthenticatedThemes} from '@shopify/cli-kit/node/session'
+import {OutputFlags} from '@oclif/core/interfaces'
+import {AdminSession} from '@shopify/cli-kit/node/session'
+
+type PublishFlags = OutputFlags<typeof Publish.flags>
 
 export default class Publish extends ThemeCommand {
   static summary = 'Set a remote theme as the live theme.'
@@ -37,16 +39,18 @@ If you want to publish your local theme, then you need to run \`shopify theme pu
     }),
   }
 
-  async run(): Promise<void> {
-    const {flags, argv} = await this.parse(Publish)
-    const store = await ensureThemeStore(flags)
-    const adminSession = await ensureAuthenticatedThemes(store, flags.password)
-    const themeId = flags.theme || argv[0]
+  static multiEnvironmentsFlags = ['store', 'password', 'theme']
 
-    if (argv.length > 0 && typeof argv[0] === 'string') {
-      renderArgumentsWarning(argv[0])
+  async command(flags: PublishFlags, adminSession: AdminSession) {
+    // Deprecated use of passing the theme id as an argument
+    // ex: `shopify theme publish <themeid>`
+    const positionalArgValue =
+      this.argv && this.argv.length > 0 && typeof this.argv[0] === 'string' ? this.argv[0] : undefined
+
+    if (!flags.theme && positionalArgValue) {
+      flags.theme = positionalArgValue
+      renderArgumentsWarning(positionalArgValue)
     }
-
-    await publish(adminSession, themeId, flags)
+    await publish(adminSession, flags)
   }
 }
--- a/packages/theme/src/cli/commands/theme/publish.ts
+++ b/packages/theme/src/cli/commands/theme/publish.ts
@@ -1,6 +1,6 @@
 import ThemeCommand from '../../utilities/theme-command.js'
 import {themeFlags} from '../../flags.js'
-import {publish, renderArgumentsWarning} from '../../services/publish.js'
+import {publish} from '../../services/publish.js'
 import {Flags} from '@oclif/core'
 import {globalFlags} from '@shopify/cli-kit/node/cli'
 import {OutputFlags} from '@oclif/core/interfaces'
@@ -21,9 +21,6 @@ If you want to publish your local theme, then you need to run \`shopify theme pu
 
   static description = this.descriptionWithoutMarkdown()
 
-  // Accept any number of args without naming them
-  static strict = false
-
   static flags = {
     ...globalFlags,
     ...themeFlags,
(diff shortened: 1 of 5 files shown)


packages/theme/src/cli/commands/theme/publish.ts
```diff
@@ -39,16 +41,18 @@ If you want to publish your local theme, then you need to run \`shopify theme pu
     environment: themeFlags.environment,
   }
 
-  async run(): Promise<void> {
-    const {flags, argv} = await this.parse(Publish)
-    const store = await ensureThemeStore(flags)
-    const adminSession = await ensureAuthenticatedThemes(store, flags.password)
-    const themeId = flags.theme || argv[0]
+  static multiEnvironmentsFlags = ['store', 'password', 'theme', 'force']
```
graygilmore: Requiring `force` feels a little bit scary to me. Could we still have a confirmation dialog that shows all of the changes that would be happening?

Thinking a little more about this: even providing a theme flag feels a little weird? The effort it would take to figure out the names or IDs of each of the themes to publish is probably the same amount of time to just hit publish on them in the admin manually.

🤔 maybe there's a good use case here that I'm not thinking of?

Would it be possible to have interactions in a multi-env command?

EvilGenius13: Super good call. Thanks to [this](https://github.com/Shopify/cli/pull/6124) PR, now we don't need the force flag. When running `theme publish` with two or more environments, we have a prompt before proceeding.
image
I've updated the code to remove `force` as a required flag

EvilGenius13: Follow up to your other question about ID/Name. You are probably right that it may not be a heavily used command for multi-env purposes. _Should_ someone be using it in CI I could see it working if they have a script pushing the theme name and passing the same theme name to the publish command. 

Interactions get broken in the CLI with how the outputs work which is why we chose to go directly to running the commands. It's definitely something we could look into though!

packages/theme/src/cli/commands/theme/publish.ts
```diff
@@ -39,16 +41,18 @@ If you want to publish your local theme, then you need to run \`shopify theme pu
...
 
-    if (argv.length > 0 && typeof argv[0] === 'string') {
-      renderArgumentsWarning(argv[0])
-    }
+  async command(flags: PublishFlags, adminSession: AdminSession) {
+    // Deprecated use of passing the theme id as an argument
```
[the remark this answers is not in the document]
dejmedus: This was discussed recently and I believe the consensus was that the deprecated positional arg logic + the warnings + strict can be removed :)

packages/theme/src/cli/services/publish.ts
```diff
@@ -18,33 +18,38 @@ export function renderArgumentsWarning(id: string) {
...
       live: false,
-      theme: themeId,
+      theme: options.theme,
     },
   })
 
-  const previewUrl = themePreviewUrl({...theme, role: 'live'} as Theme, adminSession)
+  const previewUrl = themePreviewUrl({...themeToPublish, role: 'live'} as Theme, adminSession)
 
   if (!options.force) {
```
dejmedus: The downside of how I implemented multi env batch confirmation prompts is that we will need to toggle off individual command confirmations, otherwise it will try to display both. For `delete` I handled this with a conditional, something like:
```ts
// publish.ts
async command(flags: PublishFlags, adminSession: AdminSession, multiEnvironment?: boolean) {

await publish(adminSession, flags, multiEnvironment)
```

``` ts
// services/publish.ts
export async function publish(adminSession: AdminSession, options: PublishServiceOptions, multiEnvironment?: boolean) {

if (!options.force && !multiEnvironment) {
```

EvilGenius13: Thanks, I've added those in!

packages/theme/src/cli/commands/theme/publish.ts
```diff
@@ -37,16 +39,9 @@ If you want to publish your local theme, then you need to run \`shopify theme pu
     }),
```
dejmedus, line 36: nit: now that positional arg logic has been removed I believe we can remove [static strict = false](https://oclif.io/docs/commands/#:~:text=//%20this%20makes%20the%20parser%20not%20fail%20when%20it%20receives%20invalid%20arguments%0A%20%20//%20defaults%20to%20true%0A%20%20//%20set%20it%20to%20false%20if%20you%20need%20to%20accept%20a%20variable%20number%20of%20arguments%0A%20%20static%20strict%20%3D%20false) and refresh the oclif file

packages/theme/src/cli/services/publish.ts
```diff
@@ -3,48 +3,41 @@ import {themeComponent} from '../utilities/theme-ui.js'
...
       cancellationMessage: 'No, cancel publish',
     })
     if (!accept) return
   }
 
-  await themePublish(theme.id, adminSession)
+  await themePublish(themeToPublish.id, adminSession)
 
   renderSuccess({
     body: [
```
dejmedus, line 40: We could add in some info about the environment here if there is one. Maybe something like:

```suggestion
    body: [
 ...(options.environment ? [{subdued: `Environment: ${options.environment}\n\n`}] : []),
```

dejmedus (approved):
Looks great to me + 🎩 went well! Thanks @EvilGenius13

packages/theme/src/cli/commands/theme/publish.ts
```diff
@@ -37,16 +36,9 @@ If you want to publish your local theme, then you need to run \`shopify theme pu
...
   }
 
-  async run(): Promise<void> {
-    const {flags, argv} = await this.parse(Publish)
-    const store = await ensureThemeStore(flags)
-    const adminSession = await ensureAuthenticatedThemes(store, flags.password)
```
aswamy, line 43: Do we lose anything by removing these checks?

EvilGenius13: We don't because we handle that in the `theme-command` layer [here](https://github.com/Shopify/cli/blob/3a2f619ddd070597ce4b944b58a4c9c9f2d5b8cc/packages/theme/src/cli/utilities/theme-command.ts#L259) which calls both `ensureThemeStore` and `ensureAuthenticateThemes`

EvilGenius13:
> I couldn't tophat it properly - do i need to setup anything else?
> 
> The passwords work when i publish each env individually.
> 
> image
> My toml looks like this:
> 
> ```
> [environments.env1]
> theme = "175782527140"
> store = "bottles-of-sand-2"
> password  = [redacted]
> store-password = [redacted]
> force = "true"
> 
> [environments.env2]
> theme = "153429967103"
> store = "alok-shop-experience"
> password  = [redacted]
> force = "true"
> ```

Paired with @aswamy and we realized that his shop domain wasn't being normalized. I'm going to put up a PR to fix this.

aswamy (approved):

PR #5875, opened 2025-05-21
Refresh session and retry on 401

MitchDickinson:
Reverts Shopify/cli#5852

### DIFF
--- a/packages/app/src/cli/utilities/developer-platform-client/app-management-client.ts
+++ b/packages/app/src/cli/utilities/developer-platform-client/app-management-client.ts
@@ -19,6 +19,7 @@ import {
   AppModuleVersion,
   CreateAppOptions,
   AppLogsResponse,
+  createUnauthorizedHandler,
   DevSessionUpdateOptions,
   DevSessionCreateOptions,
   DevSessionDeleteOptions,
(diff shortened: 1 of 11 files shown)


isaacroldan (approved):

packages/cli-kit/src/public/node/api/graphql.ts
```diff
@@ -40,6 +40,24 @@ export interface CacheOptions {
...
+
+interface SimpleUnauthorizedHandler {
+  type: 'simple'
+  handler: () => Promise<void>
+}
+
+interface TokenRefreshHandler {
+  type: 'token_refresh'
+  handler: () => RefreshTokenOnAuthorizedResponse
+}
```
isaacroldan, line 57: We shouldn't need a "simple" unathorized handle, it's only used from themes, be can easily modify how themes refreshes the token to use `TokenRefreshHandler`

packages/app/src/cli/utilities/developer-platform-client/app-management-client.ts
```diff
@@ -1024,6 +1079,10 @@ export class AppManagementClient implements DeveloperPlatformClient {
     )
   }
 
+  private createUnauthorizedHandler(): UnauthorizedHandler {
+    return createUnauthorizedHandler(this)
+  }
```
isaacroldan, line 1084: this is not being used, all functions call the external function not the private one. `createUnauthorizedHandler(this)`

PR #5876, opened 2025-05-21
Bump trampoline version to 1.0.2

adampetro:

Fixes an issue where a second invocation of `shopify app function build` to a Rust Function using the new Wasm API would result in an error as the trampoline binary errored in the case that a module used more than one memory. This is fixed in `1.0.2` to only error in the case when there is more than one non-imported memory.

Bump the binary version to 1.0.2

### How to test your changes?

Run `shopify app function build` against any new Rust Function (uses the Wasm API, so `shopify_function` crate version >= 1.0.0) multiple times, see that it succeeds.

### Measuring impact

### Checklist


adampetro:
seems like an unrelated error, going to retry

Issue #4431, opened 2024-09-10
[Feature]: Shopify Plus Multiple Stores + Shopify Partners Multi access


raymondmatthew:
### What area(s) will this request affect?

Theme

### What type of change do you want to see?

New feature

### Overview

Shopify CLI 3.0 is a great tool for theme development however it is hard to manage compared to Shopify themekit when it comes to handling multiple stores.

Via themekit we are able to connect a folder to a specific store without needing to switch authentication. 

### Motivation

I am handling multiple stores at a time, and doing minimal changes while working on a big feature at another store.  You can say that in a day I need to switch to around 5-10 stores.  
Then I have a client in Shopify Plus that atleast have 6 different stores in one account but still need relogging in everytime you switch.

It is very hard to switch as I need to relogin all the time just to connect with another store's CLI.
For example you are working on a big feature in store A, then store B just needs a very tiny change.
Switching to store B and switching back to A takes alot more effort than the actual change.  Unlike with themekit before where you can easily have 10 code editor opened then just switch very easily by literally just moving to the editor that is connected with the store and viola you are done.

Ideally I loved how the authenticator making our accounts more secured but with authenticator this makes it a lot harder so I was forced to remove it.

I have been using the Shopify CLI since it was released, but almost everyday I want to switch back to themekit because of this.  I hope you can find a way.

jamesmengo:
I can understand how that friction would be frustrating when switching between numerous stores!

We can look into seeing if we can make platform-level changes to make this easier, however with the number of moving parts related to this that may take a while before we can deliver something tangible.

In the meantime, would `https://shopify.dev/docs/storefronts/themes/tools/cli/environments` be helpful in removing some friction? You can pass this in using an `-e` flag.

jamesmengo:
You can set `SHOPIFY_FLAG_ENVIRONMENT` in your shell config and overriding that with the `-e` flag as needed.
Let me know how that works out for you!

gonzaloriestra:
:wave: In case you want to try, this prototype version adds multi-session support: `@shopify/cli@0.0.0-snapshot-20250530110231`. To add additional sessions or switch between them, you can just run `shopify auth login`. You can also add aliases for your accounts with `--alias`.

gonzaloriestra:
The CLI v3.85 now includes multi-session support with a new `shopify auth login` command that allows to add accounts and easily switch between them.

PR #5914, opened 2025-05-30
Multi-session support

gonzaloriestra:

Fixes https://github.com/Shopify/cli/issues/4431

Related to https://shopifypartners.slack.com/archives/CKQTKDZV2/p1747036239447309

Hack Days project: https://hackdays.shopify.io/projects/21041

https://github.com/user-attachments/assets/8d46690e-19d6-451e-b1c2-14b13ea92b1d

- Adds a new `shopify auth login` command that allows to create multiple sessions
- The command includes a `--alias` flag to customize the session names
- If no alias is provided, it will use the user ID by default
- When running commands that use Partners or App Management API, it will replace the user id with the email/org
- `shopify auth logout` clears all the sessions
- Existing sessions will not be migrated. When upgrading to this version, they will be cleared and forced to log in again.

### How to test your changes?

- `npm i -g @shopify/cli@0.0.0-snapshot-20250530110231`
- `shopify auth login` multiple times
- Any command
- `shopify auth logout`

### Measuring impact

### Checklist


packages/cli-kit/src/private/node/api/headers.ts
```diff
@@ -1,13 +1,17 @@
 import {CLI_KIT_VERSION} from '../../../public/common/version.js'
 import {firstPartyDev} from '../../../public/node/context/local.js'
 import {Environment, serviceEnvironment} from '../context/service.js'
-import {ExtendableError} from '../../../public/node/error.js'
+import {BugError} from '../../../public/node/error.js'
 import https from 'https'
 
-class RequestClientError extends ExtendableError {
+class RequestClientError extends BugError {
```
isaacroldan: I don't think this is a `BugError` no? Bug error should be something that we raise when we reach a point we know is definitely a bug (example: something that should never be undefined, it's undefined somehow)

gonzaloriestra: Right! Updated to AbortError

isaacroldan:
Tried with a normal account, but then when I run the command again I get my user ID and not my email:

gonzaloriestra:
@isaacroldan that's expected if you don't run any other command, as explained in the description:

> - If no alias is provided, it will use the user ID by default
> - When running commands that use Partners or App Management API, it will replace the user id with the email/org

isaacroldan:
ah! I see, we get this info from BP right? would it make sense for this login to authenticate you with BP directly and get that info?

gonzaloriestra:
We get the info from Partners or Business Platform, it depends on the API you are using. I guess we could add an extra query after the login to populate it, but normally you will to run at least a command after logging in, no? So I think we can skip it.

Also, there's no solution for now for theme commands, so the ID would persist in some cases. That's why I added the alias flag.

packages/cli-kit/src/public/node/session-prompt.ts
```diff
@@ -0,0 +1,95 @@
...
+  if (sessions) {
+    const sessionChoices = buildSessionChoices(sessions, fqdn)
+    choices.push(...sessionChoices)
+  }
+
+  let selectedValue = NEW_LOGIN_VALUE
+
+  if (choices.length > 0) {
+    choices.push({
+      label: 'Log in with a new account',
```
jenstanicak:
```suggestion
      label: 'Log in with a different account',
```
New made me think I was going to kick off a flow to create a brand new user name and password. Can we change it to `different` wherever we use this label?

jenstanicak (approved):
LGTM!!! Excited for this command to be added

packages/cli-kit/src/private/node/session/store.test.ts
```diff
@@ -1,62 +1,208 @@
...
+      await updateSessionAlias('user1', 'New Alias')
+
+      // Then
+      expect(setSessions).not.toHaveBeenCalled()
+    })
   })
 })
 
-function testSession(): Session {
+function testSession(): Sessions {
```
isaacroldan: this is not used anymore right? can be deleted

isaacroldan:
I've been reviewing again and 🎩 'ing, works great!

The only confusing thing for me is how the `--alias` flag works: 

If I have 2 accounts, my expectation using the `--alias` flag is that I can use it to automatically select the account with that alias without needing to answer a prompt. Instead it updates the selected account but i'd say that's not what one would expect. And there is no way to skip the prompt.

As a rule I think all prompts should have a way to skip them via flags

Proposal:
- When login with a new account: add a new mandatory prompt to assign an alias to this account (after the successful login)
- When using the `--alias` flag, it will automatically select the session with that alias. If there is no session then throw an error or ask the user "Login with a new account?"

Open questions: 
- how to update an alias? should we allow it? maybe is not that important.
- Should logout actually log you out from every account? or allow you to select the account? (maybe I only want to logout from ONE of my accounts, but the rest are working fine!)
- Again, adding a new flag for logout: `--all` if you actually want to logout from ALL accounts. Does this make sense?

What do you think?

gonzaloriestra:
@isaacroldan I like your suggestion, thanks! I've updated the PR to do that. Could you have another look?

> how to update an alias? should we allow it? maybe is not that important.

I've removed that feature. If you want to update an alias, you can log in again with the same account and it will overwrite the alias. Or just log out.

> Should logout actually log you out from every account? or allow you to select the account? (maybe I only want to logout from ONE of my accounts, but the rest are working fine!)

For now, it logs out from all the accounts. But open to improve it later. We will probably need to switch to another session (and maybe prompt?) if present.

> Again, adding a new flag for logout: --all if you actually want to logout from ALL accounts. Does this make sense?

👍

packages/cli/src/cli/commands/auth/login.ts
```diff
@@ -0,0 +1,21 @@
...
+import {promptSessionSelect} from '@shopify/cli-kit/node/session-prompt'
+import {Flags} from '@oclif/core'
+import {outputCompleted} from '@shopify/cli-kit/node/output'
+
+export default class Login extends Command {
+  static description = 'Logs you in to your Shopify account.'
+
+  static flags = {
+    alias: Flags.string({
+      description: 'An alias to identify the session.',
```
isaacroldan: maybe better something like: "Alias of the session you want to login to"?

isaacroldan (approved):
👏 👏

Issue #4692, opened 2024-10-19
[Feature]: Theme command load a default environment


cesarho:
### What area(s) will this request affect?

Theme

### What type of change do you want to see?

New feature

### Overview

The Shopify CLI already supports environment settings through `shopify.theme.toml`. I propose a small tweak to improve the default behavior by automatically loading the default environment if it is available.

**Proposed Change**
Modify the default behavior of shopify theme related command to load the environment configuration by default. Specifically:

- When a developer runs `shopify theme dev`, the CLI would effectively execute `shopify theme dev -e default` in the background.
- If the developer has created a `shopify.theme.toml` file with the default store settings, the command will automatically load these settings.
- If the `shopify.theme.toml` file doesn’t exist or doesn't contain default environment, the CLI would continue to work as it currently does —without any breaking changes.

**Example Configuration**
```toml
[environments.default]
store = "abc.myshopify.com"
```

With this change, developers could avoid unexpected deployment issues and simplify their workflow, particularly when switching between different stores.

### Motivation

Currently, the Shopify CLI remembers the last store that was worked on and keeps this stored for future sessions. However, this behavior can lead to unexpected issues: when a developer switches to another project folder and runs `shopify theme dev`, the CLI mistakenly uploads files to the previous working store instead of the intended one. This can cause significant, unintended trouble, especially if the developer is unaware of this stored preference.

Although there are alternatives, such as setting the environment variable manually or writing scripts to ensure the correct store information is set, I believe the CLI could handle this more seamlessly by improving the default behavior. A solution that prevents accidental deployment to the wrong store would benefit everyone and make the experience smoother and safer for all users.

karreiro:
@jamesmengo @graygilmore I believe this issue illustrates well how Shopify CLI aims to support multiple-store development.

kfalkiewicz:
@cesarho This is great idea and must-have feature for the freelancers that works on multiple "single" stores.

jeffreyguenther:
People are interested.

JanVormeister:
People are interested, still.

PR #5931, opened 2025-06-02
Add default environment functionality to theme commands

graygilmore:

- Closes https://github.com/Shopify/cli/issues/4692

Allows developers to set a default environment in their environments config file:

```toml
[environments.default]
store = "abc.myshopify.com"
```

This environment will automatically be used for CLI commands even if the developer does not specify the environment with the `--environment` flag. This gives developers a bit of a safety net when running commands and switching between different storefronts because this default environment will take precedence over any storefront saved to local storage.

### How to test your changes?

1. Install the snapped version:
    ```
    npm i -g @shopify/cli@0.0.0-snapshot-20250603165124 --@shopify:registry=https://registry.npmjs.org
    ```
2. Navigate to a theme folder
3. Create a `shopify.theme.toml` file if one doesn't already exist and populate it with at least a `[environments.default]` entry that has a `store` entry
4. Run a theme command like `shopify theme dev`
5. Ensure that it is using the default environment's store
6. Run the command again but specify a store `shopify theme dev --store different-store.myshopify.com`
7. Ensure that it's using the flag you passed in instead of the default store
8. Add a separate entry to the `shopify.theme.toml` file
9. Run the command again but specify your other environment `shopify theme dev -e new`
10. Ensure that it's using the new environment's store instead of the default

### Post-release steps

Merge https://github.com/Shopify/shopify-dev/pull/58447

### Checklist

### DIFF
--- a/packages/cli-kit/src/public/node/base-command.ts
+++ b/packages/cli-kit/src/public/node/base-command.ts
@@ -134,36 +134,29 @@ This flag is required in non-interactive terminal environments, such as a CI env
     options?: Input<TFlags, TGlobalFlags, TArgs>,
     argv?: string[],
   ): Promise<ParserOutput<TFlags, TGlobalFlags, TArgs>> {
-    // If no environment is specified, don't modify the results
     const flags = originalResult.flags as EnvironmentFlags
     const environmentsFileName = this.environmentsFilename()
-    let environment
-    let isDefaultEnvironment = false
 
     if (!environmentsFileName) return originalResult
 
     const environmentFileExists = await environmentFilePath(environmentsFileName, {from: flags.path})
-    const environmentSpecified = flags.environment && flags.environment.length > 0
+    const environments = flags.environment ?? []
+    const environmentSpecified = environments.length > 0
 
+    // Noop if no environment file exists and none was specified
     if (!environmentFileExists && !environmentSpecified) return originalResult
 
-    if (!environmentSpecified) {
-      // If no environment is specified attempt to load the default environment
-      environment = await loadEnvironment('default', environmentsFileName, {from: flags.path})
-      isDefaultEnvironment = true
-    }
+    // Noop if multiple environments were specified (let commands handle this)
+    if (environmentSpecified && environments.length > 1) return originalResult
 
-    if (!flags.environment && !environment) return originalResult
-
-    // If users pass multiple environments, do not load them and let each command handle it
-    if (flags.environment && flags.environment.length > 1) return originalResult
-
-    // If the specified environment isn't found, don't modify the results
-    if (!environment && flags.environment) {
-      environment = await loadEnvironment(flags.environment[0] as string, environmentsFileName, {from: flags.path})
-    }
+    const {environment, isDefaultEnvironment} = await this.loadEnvironmentForCommand(
+      flags.path,
+      environmentsFileName,
+      environments[0],
+    )
 
     if (!environment) return originalResult
+
     // Parse using noDefaultsOptions to derive a list of flags specified as
     // command-line arguments.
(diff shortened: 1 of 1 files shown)


packages/cli-kit/src/public/node/base-command.ts
graygilmore, line 1: The conditionals got a little out of hand in this file but it's a bit of a complicated thing to check for. Let me know if any of this is too complicated to read through and I can think about rearranging things.

jamesmengo (approved):
I **do** think the code in `base-command` would benefit from another pass for readability, but my tophat went well!

----

Aside from that, I wanted to raise the edge case where a user may have created an environment called 'default', which is no longer actually used in their workflow or should only be used when specified. 

Not too worried about functional impact, since we output information to the user about the environment being used. 

I wonder if that's worth making this a `minor` change though 🤔

EvilGenius13:
> I **do** think the code in `base-command` would benefit from another pass for readability, but my tophat went well!
> 
> Aside from that, I wanted to raise the edge case where a user may have created an environment called 'default', which is no longer actually used in their workflow or should only be used when specified.
> 
> Not too worried about functional impact, since we output information to the user about the environment being used.
> 
> I wonder if that's worth making this a `minor` change though 🤔

🎩 went great here too. Chiming in that I agree making it a minor change would be nice.

EvilGenius13 (approved):

graygilmore:
## Force Push Patch Notes

- Rebased with `main`

packages/cli-kit/src/public/node/base-command.ts
graygilmore: Pushed up a commit to help with this https://github.com/Shopify/cli/pull/5931/commits/fbdb994dc14a4ba9aa40fbd529a558c3850eada6

packages/cli-kit/src/public/node/base-command.ts
```diff
@@ -134,16 +134,27 @@ This flag is required in non-interactive terminal environments, such as a CI env
...
+
+    const environmentFileExists = await environmentFilePath(environmentsFileName, {from: flags.path})
+    const environments = flags.environment ?? []
+    const environmentSpecified = environments.length > 0
+
+    // Noop if no environment file exists and none was specified
+    if (!environmentFileExists && !environmentSpecified) return originalResult
+
+    // Noop if multiple environments were specified (let commands handle this)
+    if (environmentSpecified && environments.length > 1) return originalResult
```
jamesmengo, line 150: Could we skip the check for `environmentSpecified`?

graygilmore:
## Force Push Patch Notes

- Rebased with `main`

.changeset/bright-vans-rule.md
```diff
@@ -0,0 +1,9 @@
+---
+'@shopify/theme': patch
```
gonzaloriestra: Not a big deal, but new features should always be minor
```suggestion
'@shopify/theme': minor
```

.changeset/bright-vans-rule.md
```diff
@@ -0,0 +1,9 @@
...
+---
+
+Add default environment functionality
+
+You may now optionally define a default environment in your environment config
+file (e.g. `shopify.theme.toml`) that will be automatically injected into the
+running command without needing to specify it.
```
gonzaloriestra: Maybe it's not clear how to define it...

```suggestion
running command without needing to specify it. The name must be `[environments.default]`.
```

EvilGenius13 (approved):

gonzaloriestra (approved):
Code LGTM. Is the documentation ready?

graygilmore:
> Is the documentation ready?

Was waiting on approval on approach here first. I'll fire that up now and address the feedback on this PR, too.

PR #5932, opened 2025-06-02
[Feature] Display POS Extension logs within the CLI output

henryStelle:

We have had lots of feedback from [summit hackdays](https://docs.google.com/document/d/1eazVFaX-3umYTmuAUvboQh1h-rZZkQsKLflxK6caxcI/edit?tab=t.0#heading=h.1kw0jorzvrnf) that debugging POS UI Extensions is hard because the default Safari debugger misses logs or fails to start or is otherwise annoying. All of these complaits match our team's experience.

Adds support for a new event type (`log`) on the websocket connection between the POS app and CLI dev server. When log events are received, they are logged to the terminal console.

image

### How to test your changes?

Run: packages/app/src/cli/services/dev/extension/websocket/handlers.test.ts

Or, you can follow the instructions in https://github.com/Shopify/pos-next-react-native/pull/61872 to test the end-to-end flow.

### Post-release steps

1. Merge https://github.com/Shopify/extensibility/pull/464 and https://github.com/Shopify/pos-next-react-native/pull/61872 (order doesn't matter)
2. Update [debugging docs](https://shopify.dev/docs/api/pos-ui-extensions/unstable/debugging) on shopify.dev

### Measuring impact

### Checklist

### DIFF
--- a/packages/app/src/cli/services/dev/extension/websocket/handlers.test.ts
+++ b/packages/app/src/cli/services/dev/extension/websocket/handlers.test.ts
@@ -196,7 +196,7 @@ describe('getOnMessageHandler()', () => {
     wss.clients.forEach((ws) => expect(ws.send).toHaveBeenCalledWith(outgoingMessage))
   })
 
-  test('on an incoming dispatch with log type outputs the log message', () => {
+  test('on an incoming dispatch with type log calls handleLogMessage and does not notify clients', () => {
     const wss = getMockWebsocketServer()
     const options = getMockSetupWebSocketConnectionOptions()
     const data = JSON.stringify({
@@ -213,16 +213,11 @@ describe('getOnMessageHandler()', () => {
 
     getOnMessageHandler(wss, options)(data)
 
-    // Verify useConcurrentOutputContext was called with correct parameters
+    // Verify useConcurrentOutputContext (any therefore handleLogMessage) was called with correct parameters
     expect(useConcurrentOutputContext).toHaveBeenCalledWith(
       {outputPrefix: 'test-extension', stripAnsi: false},
       expect.any(Function),
     )
-    // The stdout.write() call happens inside the useConcurrentOutputContext callback
-    const contextCallback = (useConcurrentOutputContext as Mock).mock.calls[0][1]
-    contextCallback()
-
-    expect(options.stdout.write).toHaveBeenCalledWith(`INFO: Test log message`)
 
     // Verify no client messages were sent since this was a log event
     wss.clients.forEach((ws) => expect(ws.send).not.toHaveBeenCalled())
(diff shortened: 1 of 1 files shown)


packages/app/src/cli/services/dev/extension/websocket/handlers.ts
```diff
@@ -69,9 +74,29 @@ ${outputToken.json(eventData)}
         options.payloadStore.updateExtensions(eventData.extensions)
       }
     } else if (eventType === 'dispatch') {
-      const outGoingMessage = getOutgoingDispatchMessage(jsonData, options)
+      if (eventData.type === 'log') {
+        const {level, message, extensionName} = eventData.payload
 
-      notifyClients(wss, outGoingMessage, options)
+        const colouredLevel = (() => {
```
shauns: I'm not sure about this -- is the screenshot up to date? Just thinking about consistency between this extension's output and those of other extensions. There isn't much capitalisation in existing output for instance, vs. "DEBUG" and such, and the multiline `[ ... ]` seems like it stands out.

I wonder what we can do to make it better fit in. Perhaps something to try and match up on, is, what happens when I do `console.log/debug/warn` etc. from something like the template Remix app. What shows up, and how do those appear? I think thats the area where there is most existing scope for logging and hence some consistency to find.

packages/app/src/cli/services/dev/extension/websocket/handlers.ts
```diff
@@ -69,9 +74,29 @@ ${outputToken.json(eventData)}
...
+          const uppercaseLevel = level.toUpperCase()
+          switch (uppercaseLevel) {
+            case 'DEBUG':
```
shauns: I think to fit in with everything else, we should only show these when `isVerbose` returns true. They'd appear when the developer sets `--verbose`.

packages/app/src/cli/services/dev/extension/websocket/handlers.ts
henryStelle: It looks like there is precidence on the remix logs for all caps INFO/DEBUG, but the multi-line issue definitely needs work.
image

shauns: Great -- so maybe `[INFO] <log stuff>` all on one line will fit in nicely 👍

henryStelle: I've updated the logging to remove the [ .. ] syntax and format the JSON objects using the `outputToken.json` formatter.

New output:
image

packages/app/src/cli/services/dev/extension/websocket/handlers.ts
henryStelle: Responded in the above thread.

shauns (approved):

PR #5947, opened 2025-06-05
Fix payload type of dispatch log event

henryStelle:

https://github.com/Shopify/cli/pull/5932 introduced a new `log` dispatch event and the typing of the payload was inconsistent between the definition of the event and the handler.

Making the types consistent to what the handler expects in `packages/app/src/cli/services/dev/extension/websocket/handlers.ts:71`

> Question: Is there a way to import `packages/ui-extensions-server-kit/src/types.ts` in the cli so the types are not redefined (which allows them to fall out of sync as happened here)?

### How to test your changes?

See question above.

### Measuring impact

### Checklist


henryStelle:
https://github.com/Shopify/cli/pull/5949 is reverting the original PR this is patching as more substantive changes need to be made.

PR #5954, opened 2025-06-06
Update behaviour for theme command calls

EvilGenius13:

Fixes https://github.com/Shopify/developer-tools-team/issues/735

We have some issues with the way we handle multiple environments which is done in `theme-command`. 

**Environment Detection Fix:**

- Changed from `if (!flags.environment)` to `if (!flags.environment?.length)` to properly distinguish between no environments ([]) and actual environment values (["dev"])
- Added proper handling for single vs multiple environment scenarios

**Improved Multi-Environment Execution:**

- Replaced Promise.all with renderConcurrent for multiple environments to prevent terminal output conflicts
- Each environment process gets its own isolated stdout/stderr streams
- Added `CommandContext` interface to pass stream references to commands
- Integrated `AbortController` for proper cancellation handling

### How to test your changes?
- Pull the branch
- Build the branch
- Have a `shopify.theme.toml` file with a few environments

We want to test 4 different scenarios that use `theme-command.ts`' `command` method
   - No environment provided: `shopify theme list` using the cached store in your system
   - Default environment provided (To run this you will need an `environments.default` in your `.toml` file
         - `shopify theme list` which will use the default environment
   - Single environment provided: `shopify theme list -e <mystore>`
   - Multi environment provided: `shopify theme list -e <mystore> -e <myotherstore>`

Check that we get the proper output from the right environment(s)

### Post-release steps

I will monitor rollout over the week.

### Measuring impact

### Checklist

### DIFF
--- a/.changeset/itchy-houses-kick.md
+++ b/.changeset/itchy-houses-kick.md
@@ -1,0 +1,5 @@
+---
+'@shopify/theme': patch
+---
+
+fix the way we detect single or non environments & improve output for multienvironment commands
--- a/.changeset/itchy-houses-kick.md
+++ b/.changeset/itchy-houses-kick.md
@@ -2,4 +2,7 @@
 '@shopify/theme': patch
 ---
 
-fix the way we detect single or non environments & improve output for multienvironment commands
+Improvement and fixes when handling multi-environment commands
+
+- Fixes a bug where passing a single environment to multi-env commands would cause it to fail if the environment didn't have all of the required attributes for multi-env
+- Updates output when running multi-env commands to ensure the results from each command don't overlap one another
--- a/packages/theme/src/cli/utilities/theme-command.ts
+++ b/packages/theme/src/cli/utilities/theme-command.ts
@@ -4,8 +4,10 @@ import {ArgOutput, FlagOutput, Input} from '@oclif/core/lib/interfaces/parser.js
 import Command from '@shopify/cli-kit/node/base-command'
 import {AdminSession, ensureAuthenticatedThemes} from '@shopify/cli-kit/node/session'
 import {loadEnvironment} from '@shopify/cli-kit/node/environments'
-import {renderWarning} from '@shopify/cli-kit/node/ui'
+import {renderWarning, renderConcurrent} from '@shopify/cli-kit/node/ui'
 import {AbortError} from '@shopify/cli-kit/node/error'
+import {AbortController} from '@shopify/cli-kit/node/abort'
+import {Writable} from 'stream'
 
 export interface FlagValues {
   [key: string]: boolean | string | string[] | number | undefined
@@ -36,7 +38,11 @@ export default abstract class ThemeCommand extends Command {
     return configurationFileName
   }
 
-  async command(_flags: FlagValues, _session: AdminSession): Promise<void> {}
+  async command(
+    _flags: FlagValues,
+    _session: AdminSession,
+    _context?: {stdout?: Writable; stderr?: Writable},
+  ): Promise<void> {}
 
   async run<
     TFlags extends FlagOutput & {path?: string; verbose?: boolean},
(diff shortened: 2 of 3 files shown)


packages/theme/src/cli/utilities/theme-command.ts
```diff
@@ -50,20 +57,40 @@ export default abstract class ThemeCommand extends Command {
     const requiredFlags = klass.multiEnvironmentsFlags
     const {flags} = await this.parse(klass)
 
-    // Single environment
-    if (!flags.environment) {
+    // No environment provided
+    if (!flags.environment?.length) {
```
EvilGenius13, line 60: This was set more so as a pattern
When OCLIF flags have `multiple: true`, they default to `[]` instead of `undefined`. Since empty arrays are truthy, `flags.environment` would always be true.

.changeset/itchy-houses-kick.md
```diff
@@ -0,0 +1,5 @@
+---
+'@shopify/theme': patch
+---
+
+fix the way we detect single or non environments & improve output for multienvironment commands
```
graygilmore: Prefer natural language:
```suggestion
Improvement and fixes when handling multi-environment commands

- Fixes a bug where passing a single environment to multi-env commands would cause it to fail if the environment didn't have all of the required attributes for multi-env
- Updates output when running multi-env commands to ensure the results from each command don't overlap one another
```

packages/theme/src/cli/utilities/theme-command.ts
```diff
@@ -4,8 +4,10 @@ import {ArgOutput, FlagOutput, Input} from '@oclif/core/lib/interfaces/parser.js
 import Command from '@shopify/cli-kit/node/base-command'
 import {AdminSession, ensureAuthenticatedThemes} from '@shopify/cli-kit/node/session'
 import {loadEnvironment} from '@shopify/cli-kit/node/environments'
-import {renderWarning} from '@shopify/cli-kit/node/ui'
+import {renderWarning, renderConcurrent} from '@shopify/cli-kit/node/ui'
 import {AbortError} from '@shopify/cli-kit/node/error'
+import {AbortController} from '@shopify/cli-kit/node/abort'
+import {Writable} from 'stream'
```
graygilmore: Should this just be the type?

```suggestion
import {type Writable} from 'stream'
```

packages/theme/src/cli/utilities/theme-command.ts
```diff
@@ -50,50 +56,93 @@ export default abstract class ThemeCommand extends Command {
...
+    if (environments.length === 1) {
+      const environmentName = environments[0]
+      // If the environment is the default environment, the config is already available.
+      if (environmentName === 'default') {
+        const session = await this.ensureAuthenticated(flags)
+        await this.command(flags, session)
+        return
+      }
+
+      // For non-default environments, we need to load the config
```
graygilmore: Do we? Or will `await this.parse(klass)` have already done it?

graygilmore (approved):
🎩 went well

Have some open questions but not blockers if there are good answers 😄

packages/theme/src/cli/utilities/theme-command.ts
EvilGenius13: You are right!

packages/theme/src/cli/utilities/theme-command.ts
EvilGenius13: We had a redundant loading of the toml file. The parse will return the environment, whether default or not

packages/theme/src/cli/utilities/theme-command.ts
karreiro: This condition might be a little tricky for future readers to understand. What do you think about removing the "No environment" branch by using `.filter(Boolean)` instead?

```ts
/*
 * flags.environment, may be:
 *    - undefined - when nothing is passed
 *    - a string - when a single environment is passed
 *    - an array of string - when multiple environment are passed
 */
const environments = (
  Array.isArray(flags.environment) ? flags.environment : [flags.environment]
).filter(Boolean)

// Single environment
  const session = await this.ensureAuthenticated(flags)
  await this.command(flags, session)
  return
}

// Multiple environments
const sessions: {[storeFqdn: string]: AdminSession} = {}
...
```

karreiro (approved):
Thank you, @EvilGenius13! Everything looks good to me and works as expected.

I left a suggestion, but please feel free to address it in a later PR in the stack if that works better for you.

packages/theme/src/cli/utilities/theme-command.ts
EvilGenius13: I like the idea. I'm going to play around with this in a follow up!

Issue #5957, opened 2025-06-09
[Bug]: `shopify theme init` clone keeps remote repo


david-w-shopify:
### Please confirm that you have:

- [x] Searched [existing issues](.) to see if your issue is a duplicate. (If you’ve found a duplicate issue, feel free to add additional information in a comment on it.)
- [x] Reproduced the issue in [the latest CLI version](https://www.npmjs.com/package/@shopify/cli).

### In which of these areas are you experiencing a problem?

Theme

### Expected behavior

On running `git remote -v` the user should see no remote repo - it should be left for them to add

### Actual behavior

On running `git remote -v` the repo for the cloned theme is set

![Image](https://github.com/user-attachments/assets/0e42a80c-f948-46ab-9792-fffde6a297c4)

### Verbose output

<details>
  <summary>Verbose output</summary>

  ```
  Paste the output here!
  ```

</details>

### Reproduction steps

1. Run `shopify theme init`
2. CD into new theme directory
3. Run `git remote -v`
4. `origin` will be set to `https://github.com/Shopify/skeleton-theme.git`

### Operating System

Mac OS Sequoia

### Shopify CLI version (`shopify --version`)

@shopify/cli/3.80.7 darwin-arm64 node-v23.2.0

### Shell

zsh

### Node version (run `node -v` if you're not sure)

v23.2.0

PR #5963, opened 2025-06-09
Remove git remote on theme init

EvilGenius13:

Fixes https://github.com/Shopify/cli/issues/5957

This PR fixes an issue where `shopify theme init` was leaving the git remote pointing to the skeleton theme repository (`https://github.com/Shopify/skeleton-theme.git`).

Under the hood,  we are running `simple-git`'s `removeRemote` method. Before that, we check to see if any remote exists, if it doesn't we skip running `removeRemote`

Before:
image

After:
image

### How to test your changes?

- Pull down the branch
- Build the branch
- Run `shopify theme init`
- CD into the new theme folder and run `git remote -v` (you should have nothing be returned)

### Measuring impact

### Checklist

### DIFF
--- a/packages/theme/src/cli/services/init.ts
+++ b/packages/theme/src/cli/services/init.ts
@@ -1,5 +1,5 @@
 import {renderSelectPrompt, renderTasks} from '@shopify/cli-kit/node/ui'
-import {downloadGitRepository} from '@shopify/cli-kit/node/git'
+import {downloadGitRepository, removeGitRemote} from '@shopify/cli-kit/node/git'
 import {joinPath} from '@shopify/cli-kit/node/path'
 import {mkdir, writeFile} from '@shopify/cli-kit/node/fs'
 import {fetch} from '@shopify/cli-kit/node/http'
@@ -22,6 +22,7 @@ async function downloadRepository(repoUrl: string, destination: string, latestTa
           destination,
           latestTag,
         })
+        await removeGitRemote(destination)
       },
     },
   ])
--- a/packages/theme/src/cli/services/init.ts
+++ b/packages/theme/src/cli/services/init.ts
@@ -21,6 +21,7 @@ async function downloadRepository(repoUrl: string, destination: string, latestTa
           repoUrl,
           destination,
           latestTag,
+          shallow: true,
         })
         await removeGitRemote(destination)
       },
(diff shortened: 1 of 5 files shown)


packages/theme/src/cli/services/init.ts
graygilmore, line 1: I know this is fixing an existing bug but...

🤔 I wonder if there's a better way to be going about this. Currently when we're cloning I think we're getting all of the Git history? That seems a little odd to me. I wonder if instead we should be initializing a repository and then just fetching the content and writing the first commit like "Initialize a new theme".

EvilGenius13: There is an option in the `downloadGitRepository` method which `theme init` uses that allows you to do a shallow clone. That would be quicker, lighter, and less history. What do you think?

graygilmore: Definitely an improvement but I wonder if it's a step in the wrong direction? Here's an example of how Hydrogen does it: https://github.com/Shopify/hydrogen/blob/main/packages/cli/src/lib/onboarding/common.ts#L569

We don't need to block this PR, though. We can spin up a separate discussion about how we should initialize new themes.

graygilmore (approved):

packages/theme/src/cli/services/init.ts
EvilGenius13: New follow up issue opened: https://github.com/Shopify/developer-tools-team/issues/736

packages/theme/src/cli/services/init.ts
```diff
@@ -22,6 +22,7 @@ async function downloadRepository(repoUrl: string, destination: string, latestTa
           destination,
           latestTag,
         })
+        await removeGitRemote(destination)
```
aswamy, line 26: I don't really mind it being a separate call, but any reason why this isn't part of the downloadGitRepository process? We already check `ensureGitIsPresentOrAbort` in it.

i.e.
```
        await downloadGitRepository({
          repoUrl,
          destination,
          latestTag,
          removeRemote: true,
        })
```

EvilGenius13: The main reason that I didn't add it to `downloadGitRepository` is because it's used in app [init](https://github.com/Shopify/cli/blob/190e2b9f72cf0b0791009652618e721e4c8f89b4/packages/theme/src/cli/services/init.ts#L20) and app [extension ](https://github.com/Shopify/cli/blob/190e2b9f72cf0b0791009652618e721e4c8f89b4/packages/app/src/cli/services/generate/extension.ts#L102). I didn't want to change behaviour for app related commands.

Issue #5856, opened 2025-05-20
[Bug]: Images added to /assets via Shopify CLI corrupted


paul-muckypuddle:
### Please confirm that you have:

- [x] Searched [existing issues](.) to see if your issue is a duplicate. (If you’ve found a duplicate issue, feel free to add additional information in a comment on it.)
- [x] Reproduced the issue in [the latest CLI version](https://www.npmjs.com/package/@shopify/cli).

### In which of these areas are you experiencing a problem?

Theme

### Expected behavior

Image should be uploaded and viewable e.g. within the code browser in Shopify

### Actual behavior

Image is listed, but doesn't render (shows the filename and broken image error icon) - see  https://share.muckypuddle.com/7KulkngX

### Verbose output

<details>
  <summary>Verbose output</summary>

  ```
  Paste the output here!
  ```

</details>

### Reproduction steps

1. Run `shopify theme dev`
2. Add an image to the `assets` folder
3. CLI outputs `Synced » update assets/0266554465.jpeg`
4. Image is broken - see https://share.muckypuddle.com/7KulkngX
5. Same image works okay if directly uploaded in the code editor as a new asset

### Operating System

Mac OS 15.3.2 (24D81)

### Shopify CLI version (`shopify --version`)

3.80.1

### Shell

zsh

### Node version (run `node -v` if you're not sure)

v20.11.0 

### What language and version are you using in your application?

NA

aswamy:
Hey @paul-muckypuddle, just confirming, that restarting `shopify theme dev` fixes the issue of the missing asset file, right?

I was able to recreate the error you saw when i try to upload the file while `theme dev` is running; but once i restarted the command, the was gone. It is still an error, but just confirming the severity of the bug you're seeing.

paul-muckypuddle:
Hey @aswamy - just tested this. You're quite right: if the file remains and you kill / restart `shopify theme dev` it works as expected.

aswamy:
@paul-muckypuddle i think i found out what's up in the code. I'll still do a patch for the situation i've described, but hopefully you are unblocked for now.

paul-muckypuddle:
Great! Look forward to the patch.

aswamy:
The next patch version of CLI (> 3.81.1) should contain the fix. Thanks for raising!

PR #5964, opened 2025-06-09
Fix bug where `theme dev` doesnt sync images

aswamy:

Fixes https://github.com/Shopify/cli/issues/5856

- When we return the file's content, we no longer merge the raw text content (from text files) with the base64ed binary content (image files)
- By keeping them separate, we know what type of file we are uploading when calling `bulkUploadThemeAssets`
- Originally, all files were assumed to be text when doing `theme dev` because we merged text content with binary text content (see line 152 of original `packages/theme/src/cli/utilities/theme-fs.ts`)

### How to test your changes?

- Run `theme dev`
- Add an image file in `assets` folder while `theme dev` is running - you should see a message saying the image has synced in the CLI logs
- Open the theme editor, and navigate to the code editor
- You should be able to see the uploaded image
  - If you do this with the current CLI, the image would not appear in the code editor until `theme dev` was restarted

### Measuring impact

### Checklist


packages/theme/src/cli/utilities/theme-fs.ts
```diff
@@ -139,21 +139,28 @@ export function mountThemeFileSystem(root: string, options?: ThemeFileSystemOpti
...
       if (file.checksum !== previousChecksum) {
         // Sync only if the file has changed
         unsyncedFileKeys.add(fileKey)
       }
 
-      // file.value has a fallback value of '', so we want to ignore this eslint rule
-      // eslint-disable-next-line @typescript-eslint/prefer-nullish-coalescing
-      return file.value || file.attachment || ''
+      return {
+        // file.value has a fallback value of '', so we want to ignore this eslint rule
```
EvilGenius13, line 151: Is this still needed now that we don't fall back to an empty string?

aswamy: Well i want to know if `file.value` has _real_ content or not. According to that comment (made by someone else),  somewhere up the pipeline, `file.value` gets set to `""` even when it has no content.

That's why in the line below that, i explicitly do a falsy check and if it was `""`, i set `value: undefined`

.changeset/orange-humans-join.md
gonzaloriestra, line 1: If you are releasing a patch as I see in https://github.com/Shopify/cli/pull/5976, the changeset should only be there 🔥

aswamy:
Removed changeset in this branch as per @gonzaloriestra 's comment

PR #5965, opened 2025-06-10
Add ability to silence loadEnvironment warnings

graygilmore:
When we're attempting to load a default environment we don't want to render any warnings if it doesn't exist because this is happening automatically for the user.

| Before | After |
|--------|--------|
| Screenshot 2025-06-10 at 12 32 59 PM | image |

### DIFF
--- a/.changeset/force-snapshot-build.md.ignore
+++ b/.changeset/force-snapshot-build.md.ignore
@@ -1,3 +1,0 @@
----
-'@shopify/cli': patch
----
--- a/.changeset/nervous-cooks-grow.md
+++ b/.changeset/nervous-cooks-grow.md
@@ -1,5 +1,0 @@
----
-'@shopify/cli-kit': patch
----
-
-Fix a warning showing when trying to load the default environment


packages/cli-kit/src/public/node/environments.ts
```diff
@@ -24,22 +25,31 @@ export async function loadEnvironment(
...
+        body: ['No environments found in', {command: filePath}, {char: '.'}],
+      })
+    }
+    return undefined
+  }
+  const environment = environments[environmentName] as JsonMap | undefined
+
+  if (!environment) {
+    if (!options?.silent) {
+      renderWarning({
```
aswamy: Since this is repeated a lot, consider having a `renderWarningIfNotSilent(message, isSilent)` function in this file.

Looks wordy but worth it if you get annoyed at writing the if-statement 3 times.

aswamy (approved):
looks good 👍 

minor comment, but not mandatory

packages/cli-kit/src/public/node/environments.ts
gonzaloriestra: I've updated it 👍

gonzaloriestra (approved):
I've followed @aswamy suggestion and also removed the changeset, because the feature hadn't been released yet, so it's not needed.

PR #5977, opened 2025-06-12
Add back changeset file required for snapit

graygilmore:
This file was accidentally removed in https://github.com/Shopify/cli/pull/5965/commits/bd2e5a83d18332884c998079b5871707a7733f0a and is required for the `/snapit` command to work. Here's an example of the failure: https://github.com/Shopify/cli/actions/runs/15617379317/job/43993886342

### DIFF
--- a/.changeset/force-snapshot-build.md.ignore
+++ b/.changeset/force-snapshot-build.md.ignore
@@ -1,0 +1,3 @@
+---
+'@shopify/cli': patchAdd commentMore actions
+---


graygilmore:
The `/snapit` commands won't work on this PR since the action uses the yml file in `main`.

graygilmore:
> Do we care at all about the wording on the .md file?

Not that I'm aware of. I just restored it as-is

PR #5979, opened 2025-06-13
Do not require authentication for app build command

gonzaloriestra:

Fixes #0000 

### Measuring impact

### Checklist


gonzaloriestra:
* **#5980** <a href="https://app.graphite.dev/github/pr/Shopify/cli/5980?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#5979** <a href="https://app.graphite.dev/github/pr/Shopify/cli/5979?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Shopify/cli/5979?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/app/src/cli/utilities/app-command.ts
```diff
@@ -9,7 +9,7 @@ import BaseCommand from '@shopify/cli-kit/node/base-command'
  * - A remoteApp is fetched
  */
 export interface AppCommandOutput {
-  app: AppLinkedInterface
+  app: AppInterface
```
isaacroldan, line 6: I think this should still be `AppLinkedInterface` and for the commands that don't conform to that, they can force a cast. But if we relax this type for everyone we can't enforce the correct type anymore

packages/app/src/cli/services/app-context.ts
```diff
@@ -127,3 +142,26 @@ async function logMetadata(app: {apiKey: string}, organization: Organization, re
...
+  const specifications = await loadLocalExtensionsSpecifications()
+
+  // Load the local app using the specifications
+  return loadApp({
+    directory,
+    userProvidedConfigName,
+    specifications,
+    mode: unsafeReportMode ? 'report' : 'strict',
+  })
+}
```
isaacroldan, line 167: I think we should explicitly name this as something "unsafe", because `localAppContext` is very similar to `loadAppContext` and can be confusing.

I'd rename it to `unsafeLocalAppContext` or `unsafeLoadApp` (since it only returns the app)

packages/app/src/cli/utilities/app-command.ts
gonzaloriestra: I disagree 🙃. From now on, we will have two different types of app commands: with and without authentication. And I think we should properly support it, avoiding tricks like castings.

Right now we have to manually check that every new command extends from `AppCommand`, and that it returns the expected type without fake castings. And with this change, it will be similar.

To avoid confusion, what if we add a new Command class? Something like:
- `Command`: `void` (example: `shopify version`)
- `LinkedAppCommand`: `LinkedAppInterface` (example: `shopify app deploy`)
- `UnauthenticatedAppCommand`: `AppInterface` (example: `shopify app build`)

packages/app/src/cli/services/app-context.ts
gonzaloriestra: I don't like using `unsafe` here, it's perfectly safe to use it... But I agree it can be confusing to know what to return. I think [this suggestion](https://github.com/Shopify/cli/pull/5979#discussion_r2149303459) would help.

packages/app/src/cli/utilities/app-command.ts
isaacroldan: Yeah I prefer that over simplifying the existing AppCommandOutput

gonzaloriestra: Done!

packages/app/src/cli/utilities/app-unlinked-command.ts
```diff
@@ -0,0 +1,18 @@
...
+/**
+ * This command is used to run commands that exceptionally don't require authentication or a linked app.
+ */
+export default abstract class AppUnlinkedCommand extends BaseCommand {
+  environmentsFilename(): string {
+    return configurationFileNames.appEnvironments
+  }
+
+  public abstract run(): Promise<AppUnlinkedCommandOutput>
+}
```
isaacroldan: Would it make sense for the Linked class to inherit from the Unlinked one? as a ways of make it more specific.
AppLinkedCommand < AppUnlinkedCommand < BaseCommand

Because `AppLinkedCommand` also conforms to the Unlinked case beacuse `AppLinkedInterface` inherits from `AppInterface`.

does that make sense? what do you think?

That way if we ever want to add something to all commands, we can still do it in `AppUnlinkedCommand`. (maybe even rename that to `AppBaseCommand`?)

gonzaloriestra: I find it a bit forced, although it would work. If we ever need to add more staff to all app commands, I'd prefer to create a new intermediate class: 
- AppLinkedCommand < AppCommand < BaseCommand
- AppUnlinkedCommand < AppCommand < BaseCommand

Renaming it to `AppUnlinkedCommand` to `AppBaseCommand` would make it less clear that it doesn't require authentication.

But not a strong opinion!

isaacroldan: Makes sense, but maybe let's do that now then? and we can move 
```
  environmentsFilename(): string {
    return configurationFileNames.appEnvironments
  }
```
to the AppCommand?
( the only thing i'll be worried in that case is commands trying to just use `AppCommand` )

gonzaloriestra: Done! Regarding people using `AppCommand`, it's the same problem if they use `BaseCommand`... We usually copy from other commands and review it, so I think it's fine.

gonzaloriestra:
### Merge activity

* **Jun 19, 3:12 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/Shopify/cli/5979).
* **Jun 20, 4:45 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/Shopify/cli/5979).

Issue #5510, opened 2025-03-12
[Bug]: "_shopify_essential" could not be defined.


ceri-waterscreative:
### Please confirm that you have:

- [x] Searched [existing issues](.) to see if your issue is a duplicate. (If you’ve found a duplicate issue, feel free to add additional information in a comment on it.)
- [x] Reproduced the issue in [the latest CLI version](https://www.npmjs.com/package/@shopify/cli).

### In which of these areas are you experiencing a problem?

Theme

### Expected behavior

I'm unsure if this is a CLI upgrade issue, or node upgrade problem. But since upgrading both the first time i run the command `shopify theme dev` I get an error. On subsequent tries it's fine??

### Actual behavior

**Error**
"our development session could not be created because the "_shopify_essential" could not be defined. Please, check your internet"

![Image](https://github.com/user-attachments/assets/929b1e13-898d-41bf-bf2b-7a65f406eb76)

### Verbose output

<details>
  <summary>Verbose output</summary>

  ```
  Paste the output here!
  ```

</details>

### Reproduction steps

1. Install CLI version 3.67.2 & node version 23.9.0
2. Run command `shopify theme dev`
3. It breaks?
4. Run it again.. it's fine? 

### Operating System

Windows 11

### Shopify CLI version (`shopify --version`)

3.76.2

### Shell

Git Bash

### Node version (run `node -v` if you're not sure)

23.9.0

### What language and version are you using in your application?

N/A

0xtlt:
One developer of my team is having the same problem :/

0xtlt:
As @ceri-waterscreative said, 80% of the times it's working, and sometimes not, but I got those verboses when I had the bug:

(I anonymized the output)

```txt
shopify theme dev --verbose
2025-05-26T08:55:39.384Z: Running command theme dev
2025-05-26T08:55:39.386Z: Running system process in background:
  · Command: /Users/USERNAME/.nvm/versions/node/v21.7.3/bin/node /opt/homebrew/bin/shopify notifications list --ignore-errors
  · Working directory: /project/path/

2025-05-26T08:55:39.391Z: Notifications to show: 0
2025-05-26T08:55:39.399Z: Ensuring that the user is authenticated with the Theme API with the following scopes:
[]

2025-05-26T08:55:39.399Z: Ensuring that the user is authenticated with the Admin API with the following scopes for the store store-name.myshopify.com:
[]

2025-05-26T08:55:39.400Z: Getting session store...
[...]

PR #5997, opened 2025-06-18
Fix `_shopify_essential` issues related to data replication and unstable connections

karreiro:

Fixes https://github.com/Shopify/cli/issues/5510

The storefront session relies on the fact the theme is created. However, when we're working with a brand new theme, there's no guarantee it will be immediately available (see: https://kovyrin.net/2024/06/16/interview-inside-shopify-monolith). Additionally, users might experience unstable connections—either on their end or on the edge—which can cause an error in the authentication request.

We can see an example of the second case in [this report](https://github.com/Shopify/cli/issues/5510#issuecomment-2912436468), where "Request ID: unknown" is noted.

This PR updates the storefront session module to automatically retry after a failure, covering both of the situations mentioned above.

### How to test your changes?

Running `shopify theme dev` is enough, but we need to make a small change to the code to emulate the error scenario, as the the video below demonstrates:

https://github.com/user-attachments/assets/cce30118-b61e-4e2e-9df6-5da62af9afb5

### Post-release steps

N/A

### Measuring impact

### Checklist

### DIFF
--- a/.changeset/tame-mugs-arrive.md
+++ b/.changeset/tame-mugs-arrive.md
@@ -1,0 +1,5 @@
+---
+'@shopify/theme': patch
+---
+
+Fix `_shopify_essential` issues related to data replication and unstable connections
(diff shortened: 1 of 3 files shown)


EvilGenius13 (approved):
Looks Great! Thank you for putting this together 🚀

PR #6009, opened 2025-06-23
Remove deprecated  command

graygilmore:
The `theme serve` command was deprecated more than two years ago in favor of `theme dev` and is no longer available in any public documentation. This does a little bit of code clean up to remove it.


EvilGenius13:
I was wondering your opinion on this but maybe we should wait before removing this until we put a warning that this command will no longer be available after X time or X version. I can confirm that this command is still being used daily.

graygilmore:
> until we put a warning that this command

The command has already output that it's been deprecated for ~2 years now. We could create the alias in a way that could help us delete some code (https://oclif.io/docs/aliases/#command-aliases) but I'd rather just remove it. Since this is a simple word change for developers (who could create their own alias if they want!) I don't think another warning is necessary.

EvilGenius13:
> > until we put a warning that this command
> 
> The command has already output that it's been deprecated for ~2 years now. We could create the alias in a way that could help us delete some code (https://oclif.io/docs/aliases/#command-aliases) but I'd rather just remove it. Since this is a simple word change for developers (who could create their own alias if they want!) I don't think another warning is necessary.

Totally get your thoughts. My only other request would be to release this as a major patch since we're making a "breaking change". I don't think it should go in as a minor patch but I'd like to hear if anyone else has an opinion on that.

karreiro (commented):
Thank you for this proposal, @graygilmore!

I agree with @EvilGenius13 that we shouldn't introduce breaking changes to APIs in a minor release. While `shopify theme serve` has been deprecated for some time, we've had conversations with partners who have built custom tooling and commands on top of the CLI. Breaking this API could create friction for them.

I'd love to remove this command and address other API changes as well—like our current approach for ignoring files with the `--only` and `--ignore` flags—but those changes really call for a major release, as they break contracts.

I'm sure there are other areas in the CLI with similar needs, so planning for a major release might be the best path forward.

Thanks again for bringing this up; I think this is exactly the kind of seed we need 🌱 :)

graygilmore:
Going to close this for now

PR #6012, opened 2025-06-23
Ensure all theme commands accept --path

graygilmore:
While debugging some other commands during development on the CLI I noticed that some didn't accept a `--path` flag which meant that it was difficult to test some functionality. This PR ensures that all theme commands that can accept a `--path` flag do.

This can potentially expand when/how theme developers call these commands but also is handy when build the CLI to be able to do `p shopify theme info --path <path>` to utilize a `shopify.theme.toml` in another folder.

The only command that I did not add the flag to was for `language-server` but it doesn't support any theme flags.

## 🎩 Tophatting

1. Install the snapped version `npm i -g @shopify/cli@0.0.0-snapshot-20250623223113 --@shopify:registry=https://registry.npmjs.org`
2. Ensure you have a theme folder somewhere with a `shopify.theme.toml` in it
3. Run a command that previously didn't accept `--path` (e.g. `shopify theme list --path <path> -e <env>`
4. Ensure it works!

### DIFF
--- a/docs-shopify.dev/commands/interfaces/theme-check.interface.ts
+++ b/docs-shopify.dev/commands/interfaces/theme-check.interface.ts
@@ -52,7 +52,7 @@ export interface themecheck {
   '-o, --output <value>'?: string
 
   /**
-   * The path to your theme directory.
+   * The path where you want to run the command. Defaults to the current working directory.
    * @environment SHOPIFY_FLAG_PATH
    */
   '--path <value>'?: string
--- a/docs-shopify.dev/commands/interfaces/theme-console.interface.ts
+++ b/docs-shopify.dev/commands/interfaces/theme-console.interface.ts
@@ -19,7 +19,7 @@ export interface themeconsole {
   '--password <value>'?: string
 
   /**
-   * The path to your theme directory.
+   * The path where you want to run the command. Defaults to the current working directory.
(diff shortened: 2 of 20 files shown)


karreiro (commented):
Thank you for this PR, @graygilmore!

Commands like `shopify theme info`, `shopify theme console`, and `shopify theme list` don't act on the file system since they aren't dependent on a local theme. Still, it's definitely useful to be able to select the directory with the `--path` flag to load the `shopify.theme.toml` file.

May we update the documentation to be a bit more general? (as the directory doesn't necessarily have to be a theme directory for those commands)

Thanks again for this PR!

graygilmore:
> May we update the documentation to be a bit more general?

Yes! Good call

graygilmore:
@karreiro pushed up another commit!

graygilmore:
> Shouldn't something change when you set --path? Is there some magic behind the scenes I'm not aware of?

There is! For some of these commands the only reason for the `--path` flag being added is so that it can find the `shopify.theme.toml` file. This happens behind the scenes when we run `this.parse()`:

```ts
// packages/theme/src/cli/commands/theme/publish.ts
const {flags, argv} = await this.parse(Publish)
```

If you follow that through you'll find `resultWithEnvironment` that uses `flags.path` to determine where to load the environment from. That's why there aren't a lot of code changes necessary.

Issue #5951, opened 2025-06-05
[Feature]: Ignore Files When Deploying Theme Extension


sshaw:
### What area(s) will this request affect?

Extension

### What type of change do you want to see?

New feature

### Overview

Currently when running `shopify app dev` or `shopify app deploy` errors occur when backup (or other) files exist in the `extensions` directory. For example: 

```
18:12:49 │              graphiql │ GraphiQL server started on port 3457
18:12:49 │ something-amaaaazing │ Bundling theme extension something-amaaaazing...
18:12:50 │ something-amaaaazing │ Error while updating drafts: blocks directory should only contain .liquid files, snippets directory should only contain .liquid files
```

### Motivation

The file it's referring to are Emacs backups: ` some-file.liquid~`.  Instead of succeeding but failing (i.e., nothing is deployed even though command acts as if it did —another issue for another day), the command should look to `.gitignore`, `.npmignore`, or similar. 

One option it to only deploy files checked into git but this is not too developer friendly as one may not want to check in a file but just wants to test it.

charlespwd:
Does [.shopifyignore](https://shopify.dev/docs/storefronts/themes/tools/cli#excluding-files-from-shopify-cli) not work?

graygilmore:
Closing as we have support for ignore files already. Let us know if that isn't what you're asking about!

sshaw:
Thanks sounds like that will work. 

One suggestion: a `.shopifyignore` should be generated by `shopify app init`

graygilmore:
We don't want to generate one if you don't need to use it but we definitely need to do a better job at exposing the functionality for it.

Edit: though I guess we could generate one and just have it empty, I'll think about that!

sshaw:
Additional food for thought: I wouldn't add Emacs-specific ignore patterns to a repo's `.gitignore` as Emacs ignores are specific to my environment and others may use VS Code etc... I'd put patterns specific to my environment in `~/.gitignore`. The same applies here. 

If you're interested in such functionality please point me to the proper source files and I will open a PR  :neckbeard:

graygilmore:
A global ignore would be interesting! It's not something we've considered but I don't see a huge downside to the addition. The code is mostly in this file: https://github.com/Shopify/cli/blob/main/packages/theme/src/cli/utilities/asset-ignore.ts#L60

sshaw:
[...]

PR #6018, opened 2025-06-26
[Feature] Add support for a global .shopifyignore file in user's home directory

sshaw:

Stems from #5951. 

I wouldn't add Emacs-specific ignore patterns to a repo's `.gitignore` as Emacs ignores are specific to my environment and others may use VS Code etc... I'd put patterns specific to my environment in `~/.gitignore`.  Same applies to `.shopifyignore`.

Expanding the search for ignore patterns from the project root's `.shopifyignore` to include `~/.shopifyignore`.

### How to test your changes?

`pnpm vitest run packages/theme/src/cli/utilities/asset-ignore.test.ts`

### Post-release steps

N/A

### Measuring impact

🤷 

### Checklist

Can open a PR for doc changes if approved.

### DIFF
--- a/packages/theme/src/cli/utilities/asset-ignore.ts
+++ b/packages/theme/src/cli/utilities/asset-ignore.ts
@@ -1,3 +1,4 @@
+import os from 'os'
 import {uniqBy} from '@shopify/cli-kit/common/array'
 import {fileExists, readFile, matchGlob as originalMatchGlob} from '@shopify/cli-kit/node/fs'
 import {outputDebug} from '@shopify/cli-kit/node/output'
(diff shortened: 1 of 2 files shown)


packages/theme/src/cli/utilities/asset-ignore.ts
```diff
@@ -57,13 +58,18 @@ function filterBy(patterns: string[], type: string, invertMatch = false) {
...
+  const globalIgnore = joinPath(os.homedir(), SHOPIFY_IGNORE)
 
-  const shopifyIgnoreExists = await fileExists(shopifyIgnorePath)
-  if (!shopifyIgnoreExists) return []
+  const patterns = await readIgnoreFile(localIgnore)
+  return patterns.concat(await readIgnoreFile(globalIgnore))
+}
 
-  const content = await readFile(shopifyIgnorePath, {encoding: 'utf8'})
+async function readIgnoreFile(path: string) {
```
sshaw, line 75: Not really a fan of this approach, much simpler to ~due~ do via `flatMap` in `getPatternsFromShopifyIgnore` but given `await`'s semantics (`await` is an anti-pattern IMHO and this is just another example) it was a _cleaner_ approach 😉

sshaw:
Still active —and useful!

EvilGenius13:
Hi @sshaw! Thank you for your contribution. This feature is a really neat idea. I've done  a bit of testing and had a few suggestions / questions

1. Duplicate patterns.
When both local and global `.shopifyignore` files contain the same pattern, they get duplicated in the final array. Maybe we could deduplicate that.
2. Negation Precedence
I tested different negation scenarios and found that both global and local files can override each other when using a negation pattern (`!pattern`). For instance:

**Scenario A: Ignore globally, negate locally**
```
# Global ~/.shopifyignore
assets/*

# Local .shopifyignore  
!assets/important.css
``` 

**Scenario B: Ignore locally, negate globally**
```
# Local .shopifyignore
assets/*

# Global ~/.shopifyignore
!assets/important.css
```

In both cases, the negated pattern takes precedence which might be confusing for users. 

I was thinking local patterns should probably take precedence over global ones since the local `.shopifyignore` is likely where a user would want more granular control. My mental model would be that the global ignore file has the generic stuff we expect all themes to ignore (like .DS_Store, editor temp files, etc.), but you can always override that behavior locally in your project.

I would love to hear your thoughts! Thank you again for creating this PR.

sshaw:
@EvilGenius13 thanks for the feedback 💪 

> When both local and global .shopifyignore files contain the same pattern, they get duplicated in the final array. Maybe we could deduplicate that

Easy enough. 

> Negation Precedence
> I tested different negation scenarios and found that both global and local files can override each other when using a
> negation pattern (!pattern). For instance:
> I was thinking local patterns should probably take precedence over global ones since the local .shopifyignore 

Yes local should have precedence. 

Looks like I just naively put local into the array first in an attempt to give it precedence 🙃 Having a quick looksy looks like `applyIgnoreFilters` may be the spot for additional adjusting. Will try to dig in in a week or so ⏳

sshaw:
>> When both local and global .shopifyignore files contain the same pattern, they get duplicated in the final array. Maybe we could deduplicate that

> Will try to dig in in a week or so ⏳

Took a few mins to update for the uniqueness part. Let me know your thoughts. Will address the other thangz 🔜

sshaw:
Re-open! Will be committing 🔜

PR #6040, opened 2025-06-30
Slow autocomplete prompt to avoid rate limiting

amcaplan:

The autocomplete was:

- Firing too often (not debouncing properly due to React recreating the function)
- Making a query that was too expensive (no reason to return app modules)

1. Makes the query itself less expensive (not fetching app modules)
2. Replaces debounce with throttle, so we will search as the user types but in a throttled fashion

### How to test your changes?

Run `app config link` and search for an existing app

### Measuring impact

### Checklist

### DIFF
--- a/packages/app/src/cli/api/graphql/app-management/generated/apps.ts
+++ b/packages/app/src/cli/api/graphql/app-management/generated/apps.ts
@@ -1,6 +1,5 @@
 /* eslint-disable @typescript-eslint/consistent-type-definitions */
 import * as Types from './types.js'
-import {JsonMapType} from '@shopify/cli-kit/node/toml'
 
 import {TypedDocumentNode as DocumentNode} from '@graphql-typed-document-node/core'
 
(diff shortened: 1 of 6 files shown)


isaacroldan:
Tested and couldn't really tell the difference in loading times, but reducing the query complexity makes a lot of sense anyway!

About debounce vs throttle, won't throttle trigger more requests? Did you noticed an improvement on switching from one to the other?

amcaplan:
The existing logic was totally screwy. Because we weren't memoizing properly, we would:

* Set the state to loading after 100ms
* Start searching after 300ms regardless of the user's typing
* Trigger a new search on every keypress, and update results as they come in

It felt really snappy, but was sending too many backend requests. And while we accepted that on Partners, on the new dev platform we are going to be stricter about rate limiting, so we wanted to be a bit more strategic with sending requests.

When we memoized properly and started debouncing, I realized that this isn't actually the experience we want. While it makes sense logically, experientially it's wrong. You start typing and don't see any result until you've stopped, which means you don't get the nice experience of results updating in realtime as you type. (If you never stopped typing and erasing, we'd never even start the search.)

A throttle leads to a much better experience, with results updating every so often, and is quite similar to what we have now, hence you not noticing the difference.

PR #6043, opened 2025-06-30
fix: make dev store creation links clickable in CLI (fixes #114)

amcaplan:
## Summary
This PR makes the dev store creation URL clickable in the CLI when developers don't have any dev stores associated with their organization.

## What changed
- Modified `getCreateDevStoreLink()` in both `app-management-client.ts` and `partners-client.ts` to return `TokenItem` instead of plain string
- Updated the `DeveloperPlatformClient` interface to reflect the new return type
- URLs now use the CLI UI Kit's link token format: `{link: {url, label}}`

## Why
Previously, developers had to manually copy and paste the URL from the terminal. With this change, they can click the link directly in terminals that support clickable links, improving the developer experience.

## Test plan
- [ ] Run `npm dev` without any dev stores to see the clickable link
- [ ] Verify the link works in terminals that support clickable links
- [ ] Confirm fallback behavior in terminals that don't support links
- [ ] Test both App Management and Partners organization flows

Fixes #114


amcaplan (commented):
I think it would be better to use a link with a semantic label. Basically the label would be "Create one now" and then the URL would be the URL as is.

amcaplan:
Screenshot 2025-07-02 at 15 09 18
Screenshot 2025-07-02 at 15 09 45

isaacroldan (approved):

Issue #3908, opened 2024-05-19
Introduce ‘shopify theme duplicate’ command


knjshimi:
### What area(s) will this request affect?

Theme

### What type of change do you want to see?

New feature

### Overview

Introduce a theme duplicate command in the cli, so developers can duplicate a theme in the library.

### Motivation

Pushing an entire new theme via cli takes considerably longer than duplicating a pre existing theme - especially in cases where the merchant has lots of assets.

In my workflow, I never overwrite the live theme directly. Instead, when deploying, I create a new theme and allow the merchant to review it before making the decision to publish.

With the duplicate command, I could simply duplicate the live theme and push any changes to it - in that case only the files that have changed will be overwritten.

JeriverGAP:
This is a clone of the (closed) issue here:
https://github.com/Shopify/cli/issues/2000 

Just to reiterate, this would be very useful for CI/CD workflows.

jonXmack:
+1 for this!

jonXmack:
Still active

knjshimi:
still relevant

nboliver-ventureweb:
This should absolutely be a feature of the CLI. 
As @mgrbec noted, if it ever does get implemented it's imperative that translations are copied over, the same way they are when using the duplicate button in Shopify admin (related issue: https://github.com/Shopify/cli/issues/4486)

jeffreyguenther:
>Pushing an entire new theme via cli takes considerably longer than duplicating a pre existing theme - especially in cases where the merchant has lots of assets.

Would it be accurate to say that you're looking for the Admin API to support duplications?

The command you're describing would require a server-side change.

vw-anne-price:
Absolutely needed. Please add this feature.

tim-vw:
+1 vote

Maxtasy:
+1 We need this please. :)

nboliver-ventureweb:
@c-lambert See this related issue, which was closed without a fix yesterday: https://github.com/Shopify/cli/issues/4486

It seems the CLI team is recommending to use the "support channel" in the Translate and Adapt app to request this feature. Doesn't exactly inspire confidence that this is a priority for them to address...

That said, I submitted a ticket to Translate & Adapt and did get a response that they will submit a feature request, so maybe there is hope if enough people request it.

marklauyq:
+1 vote

dcrawsh:
Please do this Shopify, we need this!

hcmlopes:
[...]

PR #6050, opened 2025-07-02
Add theme duplicate command

dejmedus:
- Closes https://github.com/Shopify/cli/issues/3908

Currently theme duplication is possible via the admin, but we also want to add the ability to duplicate themes with a CLI command. This will simplify the process of copying a theme via terminal/CI flows

Adds a new theme command that will duplicate, and optionally rename, a store theme. 

```
shopify theme duplicate --theme <theme ID> --name 'A Duplicated Theme!'"
```

The command will return details of the duplicated theme. If the `--json` flag is used output will be a JSON object. When run in a CI environment or using the `--force` flag prompts and confirmations are disabled.

https://github.com/user-attachments/assets/1918083e-60b0-4dc0-8f85-fc9265080966

Also generated the associated command reference docs

### How to test your changes?

- Pull down the branch
- Running `pnpm shopify theme duplicate` should duplicate the selected theme and be visible in *admin > online store > themes > theme library*
- Flags: 
  - `theme` ID of the theme to duplicate
  - `json` should return theme info as json object
  - `force` should disable prompts/confirmations
  - `name` should name the duplicate
  - `environment` should use store/theme ID/password
- To mock a CI environment, `CI=true` can be appended to commands (should disable prompts/confirmations)

### Post-release steps

- when admin graphql v2025-10 becomes supported the hardcoded `version: 2025-10` can be removed

### Checklist


packages/theme/src/cli/services/duplicate.ts
```diff
@@ -0,0 +1,212 @@
...
+/**
+ * Initiates the duplicate process based on provided flags.
+ *
+ * @param adminSession - The admin session for the theme.
+ * @param theme - The theme to duplicate.
+ * @param flags - The flags for the duplicate operation.
+ */
+export async function duplicate(adminSession: AdminSession, themeId: string | undefined, flags: DuplicateFlags) {
+  const {rename, verbose, noColor, json, force} = flags
+  const noPrompts = force || isCI()
```
dejmedus: ESlint recommends I use `??` here, but I feel like `||` is more appropriate in this case. If this was changed to nullish and for some reason `force` was explicitly set to false the CI check would be ignored.

charlespwd (approved):
Couldn't test it but code looks good! Very clean ✨ 👍

packages/theme/src/cli/commands/theme/duplicate.ts
```diff
@@ -0,0 +1,86 @@
...
+      env: 'SHOPIFY_FLAG_FORCE',
+    }),
+  }
+
+  async run(): Promise<void> {
+    const {flags, argv} = await this.parse(Duplicate)
+    const store = ensureThemeStore(flags)
+    const adminSession = await ensureAuthenticatedThemes(store, flags.password)
+
+    const themeId = flags.theme ?? argv[0]
```
aswamy: What's the rationale behind this `argv` support? 

If i wanted to do something like the following, ill get an error because it assumes `--force` is a theme id.

image

Im pretty sure positional args are deprecated, so might as well remove that support entirely for new commands.

image

docs-shopify.dev/commands/interfaces/theme-duplicate.interface.ts
```diff
@@ -0,0 +1,56 @@
...
+   * Password generated from the Theme Access app.
+   * @environment SHOPIFY_CLI_THEME_TOKEN
+   */
+  '--password <value>'?: string
+
+  /**
+   * Rename the theme to the specified name.
+   * @environment SHOPIFY_FLAG_RENAME
+   */
+  '-r, --rename <value>'?: string
```
aswamy: `rename` implies an existing item changing its name. I'd just stick with `name` as the arg.

packages/theme/src/cli/utilities/theme-selector.ts
```diff
@@ -67,6 +67,18 @@ export async function findOrSelectTheme(session: AdminSession, options: FindOrSe
...
 }
 
+/**
+ * Finds a theme by its ID.
+ *
+ * @param session - Current Admin session
+ * @param themeId - The ID of the theme to find
+ * @returns the {@link Theme} with the given ID
+ */
+export async function findThemeById(session: AdminSession, themeId: string): Promise<Theme | undefined> {
```
aswamy, line 77: I don't think we should add something new here unless necessary. There is already a `findThemes` function that we could use here instead.

packages/theme/src/cli/services/duplicate.ts
aswamy: You're probably right that `isCI` should be checked first over `force` flag. You can either change the order of that `||`, or you can just add an esline ignore above it.

aswamy (commented):
Minor comments but overall looks great! Good stuff!

packages/theme/src/cli/commands/theme/duplicate.ts
dejmedus: I noticed this pattern in some other commands and assumed the deprecation warning was being kept around until next big CLI bump. I should have checked, removed it :)

The first error is intentional, `--force` skips both prompts and confirmations so a theme ID must be provided with `--theme` or `--environment`. I thought it made sense for force to prevent interactions to ensure scripts wouldn't accidentally hang

docs-shopify.dev/commands/interfaces/theme-duplicate.interface.ts
dejmedus: Sounds good, fixed!

packages/theme/src/cli/utilities/theme-selector.ts
dejmedus: That makes sense! I added `findThemeById` to make it easier grab themes by ID (not name/roles) and handle ID not found errors via JSON with `--json`. `findThemes` only seems to be used by delete right now and throws an AbortError

Options:

- use `findThemes` and catch/handle ID not found errors. ESLint doesn't love that so I would need to disable no-catch-all on that line
- make the `findByThemeId` helper local to duplicate file, still would be added code but would be more specific/less clutter in theme-selector

What do you think?

packages/theme/src/cli/services/duplicate.ts
dejmedus: Yes, ESLint has no problem with `CI() || force` 🤦‍♀‍ Thank you!!

packages/theme/src/cli/utilities/theme-selector.ts
aswamy: > throws an AbortError

Oh i didn't know that! If that's the case, your `findThemeById` is correct. Ignore my original comment :)

aswamy (approved):
I tophatted it and everything looks great!

I noticed there are some type issues, but don't think it was created by your PR. Not sure if it's my environment, but if you also see it, could you create a bug ticket to resolve these type issues? We can prioritize it after this PR is merged.

image
image

dejmedus:
> I noticed there are some type issues, but don't think it was created by your PR. Not sure if it's my environment, but if you also see it, could you create a bug ticket to resolve these type issues? We can prioritize it after this PR is merged.

Strange! I'm not seeing those

graygilmore (approved):

gonzaloriestra (approved):

Issue #6033, opened 2025-06-27
[Bug]: `shopify theme package` error since Shopify CLI `3.81.0`


msev:
### Please confirm that you have:

- [x] Searched [existing issues](.) to see if your issue is a duplicate. (If you’ve found a duplicate issue, feel free to add additional information in a comment on it.)
- [x] Reproduced the issue in [the latest CLI version](https://www.npmjs.com/package/@shopify/cli).

### In which of these areas are you experiencing a problem?

Theme

### Expected behavior

When I run the `shopify theme package` command, a zip archive of the theme should be created.

### Actual behavior

Since updating to version 3.81.0 and when the `default` environment is defined in the `shopify.theme.toml` file, the `shopify theme package` command returns the following error:

```
╭─ error ───────────────────────────────────────────────────────────────────────────────────╮
│                                                                                           │
│  Nonexistent flag: --environment                                                          │
│  See more help with --help                                                                │
│                                                                                           │
╰───────────────────────────────────────────────────────────────────────────────────────────╯
```
If I delete the `default` environment from `shopify.theme.toml` or rename it, the `shopify theme package` command executes correctly and a zip archive of the theme is created.

### Verbose output

<details>
  <summary>Verbose output</summary>

  ```
shopify theme package --verbose
2025-06-27T12:23:03.250Z: Running command theme package
2025-06-27T12:23:03.252Z: Running system process in background:
  · Command: /usr/local/bin/node /opt/homebrew/bin/shopify notifications list --ignore-errors
  · Working directory: /Users/maxime/Sites/unique-starter-theme/theme

2025-06-27T12:23:03.258Z: Notifications to show: 0
2025-06-27T12:23:03.261Z: Reading the content of file at shopify.theme.toml...
╭─ error ───────────────────────────────────────────────────────────────────────────────────╮
│                                                                                           │
│  Nonexistent flag: --environment                                                          │
│  See more help with --help                                                                │
│                                                                                           │
╰───────────────────────────────────────────────────────────────────────────────────────────╯

2025-06-27T12:23:03.290Z: Running system process:
  · Command: npm prefix
  · Working directory: /Users/maxime/Sites/unique-starter-theme/theme

2025-06-27T12:23:03.520Z: Obtaining the dependency manager in directory .....
2025-06-27T12:23:03.674Z: Getting session store...
2025-06-27T12:23:03.676Z: Getting session store...
[...]

PR #6054, opened 2025-07-02
Fix theme package trying to use default environment

graygilmore:

- Closes https://github.com/Shopify/cli/issues/6033

The package command does not support an environment flag and would fail if a default environment was present in a users `shopify.theme.toml` because we try to load it automatically.

This adds an early return if we're using the default environment but the command doesn't support it.


graygilmore:
## Force Push Patch Notes

- Rebased with `main`

gonzaloriestra (approved):

PR #6063, opened 2025-07-03
Change `toFile`/`toStore` etc flags to `to-file`/`to-store`

JordanVerasamy:
Resolves https://github.com/Shopify/workflows-operations/issues/3338.

We want these flags to be in kebab-case for consistency rather than lowerCamelCase, so this PR does that.

Tophat using `pnpm shopify store copy`, and provide some flags in kebab case.

image


jordanverasamy:
Replaced by https://app.graphite.dev/submit/Shopify/cli/6065.

PR #6067, opened 2025-07-03
Update authentication requirements for CLI commands (typegen + run)

MitchDickinson:

- Fixes shop/issues-develop#1311
- Updates `function typegen` to not require authentication to run
- Updates `function run` to _conditionally_ require authentication. It does this by embedding conditional app linking inside the `getOrGenerateSchemaPath` function - if the schema is already there, the linked app is not required, otherwise it is (which forces auth).

- In addition to the changes above, this PR removes the `inFunctionContext` function. This function was previously performing some boilerplate around app loading and function selecting, but it was constraining when we had commands that wanted to mix authenticated and non-authenticated environment. I realized that the only line of code `inFunctionContext` was actually saving us was the `chooseFunction` step. This seemed like not much benefit to another level of abstraction, so I removed it. Commands now have to load the app (linked or local) and invoke chooseFunction as necessary themselves. This change affected `function replay` and `function schema` commands, even though those commands have no functional changes within this PR.
- Also removed `path` from `GenerateSchemaOptions` as it was unused

### How to test your changes?
- You can create an app with a function and invoke the modified commands for that function (replay, run, schema, typegen)
- NOTE: I have not thoroughly tophatted these changes yet since I am having many issues running functions locally.
- I noticed we don't have automated/unit tests for the command files - probably something to improve later but I wasn't sure it should be scoped into this PR or not.

### DIFF
--- a/packages/app/src/cli/commands/app/function/run.ts
+++ b/packages/app/src/cli/commands/app/function/run.ts
@@ -1,7 +1,8 @@
-import {functionFlags, inFunctionContext, getOrGenerateSchemaPath} from '../../../services/function/common.js'
+import {chooseFunction, functionFlags, getOrGenerateSchemaPath} from '../../../services/function/common.js'
 import {runFunction} from '../../../services/function/runner.js'
 import {appFlags} from '../../../flags.js'
-import AppLinkedCommand, {AppLinkedCommandOutput} from '../../../utilities/app-linked-command.js'
+import AppUnlinkedCommand, {AppUnlinkedCommandOutput} from '../../../utilities/app-unlinked-command.js'
+import {localAppContext} from '../../../services/app-context.js'
 import {globalFlags, jsonFlag} from '@shopify/cli-kit/node/cli'
 import {Flags} from '@oclif/core'
 import {renderAutocompletePrompt, isTTY} from '@shopify/cli-kit/node/ui'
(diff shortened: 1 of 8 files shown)


packages/app/src/cli/services/function/common.test.ts
```diff
@@ -120,12 +73,14 @@ describe('getOrGenerateSchemaPath', () => {
   test('generates the schema file if it does not exist', async () => {
     // Given
     const expectedPath = joinPath(extension.directory, 'schema.graphql')
-    vi.mocked(fileExists).mockResolvedValue(false)
-    vi.mocked(generateSchemaService).mockResolvedValueOnce()
+    vi.mocked(fileExists).mockResolvedValueOnce(false)
```
MitchDickinson, line 76: I think this mock wasn't working correctly and the code path wasn't being correctly exercised

packages/app/src/cli/commands/app/function/run.ts
```diff
@@ -34,62 +35,64 @@ export default class FunctionRun extends AppLinkedCommand {
...
-            outputDebug(
-              `Using export '${functionExport}'. Use the --export flag or an interactive terminal to select a different export.`,
-            )
-          }
-        } else {
-          outputDebug(
-            `No targeting information found. Using the default export '${functionExport}'. Use the --export flag or an interactive terminal to select a different export.`,
-          )
-        }
+    if (flags.export !== undefined) {
```
MitchDickinson, line 50: Logic beyond here is unchanged, it's just a nesting change (no more inFunctionContext)

nickwesselman:
Is there a reason that the login from `function run` (when the schema is missing) has me click a link, instead of pressing a key to open?

image

I'd expect:
```
To run this command, log in to Shopify.
User verification code: PDZC-CXDH
👉 Press any key to open the login page on your browser
```

Otherwise, I confirmed that `app build`, `app function build`, `app function typegen`, and `app function run` all work w/out auth!

MitchDickinson:
@nickwesselman  I think that's unrelated to my change and comes from [this code](https://github.com/Shopify/cli/blob/main/packages/cli-kit/src/private/node/session/device-authorization.ts#L70-L85). For some reason when you tophatted your environment was yielding a true value for `isCloudEnvironment() || !isTTY()`

isaacroldan (approved):

MitchDickinson:
@nickwesselman did you tophat `app function schema` and `app function replay`?

nickwesselman:
@MitchDickinson I've now verified on this snapshot that when logged out, both `schema` and `replay` prompt for login then work correctly. They also work fine when already logged in.

PR #6105, opened 2025-07-14
Send uuid on app version create to do app matching

alfonso-noriega:

Enable proper identification and matching of extensions during deployment while transitioning to the new dev platform.

- Adds UUID support to app manifests (so the backend can identify the module and migrate them)
- Improves extension matching logic to handle both UUID and the new UID matching.
- Updates local extensions with remote UIDs when matches are confirmed
- Stops writing the `.env` files to disk.

### How to test your changes?

To verify everything still works fine with Partners:
- Do some deploys adding, updating and deleting extensions. 
- See that everything works as always.

To verify the new behavior in the new Dev Platform (running locally):
- Create a new app with some extensions and make a deploy
- Edit the local DB and remove the `user_identifier` from the corresponding table
- Try to make a new deploy: 
  - All extensions should still be matched automatically.

### Measuring impact

### Checklist

### DIFF
--- a/packages/app/src/cli/models/app/identifiers.test.ts
+++ b/packages/app/src/cli/models/app/identifiers.test.ts
@@ -208,12 +208,9 @@ describe('getAppIdentifiers', () => {
       })
 
       // When
-      const got = getAppIdentifiers(
-        {
-          app,
-        },
-        testDeveloperPlatformClient(),
-      )
+      const got = getAppIdentifiers({
+        app,
+      })
 
       // Then
       expect(got.app).toEqual('FOO')
@@ -238,43 +235,12 @@ describe('getAppIdentifiers', () => {
         {
           app,
         },
-        testDeveloperPlatformClient(),
         {SHOPIFY_API_KEY: 'FOO', SHOPIFY_TEST_UI_EXTENSION_ID: 'BAR'},
       )
 
       // Then
       expect(got.app).toEqual('FOO')
       expect((got.extensions ?? {})['test-ui-extension']).toEqual('BAR')
-    })
-  })
-
-  test('returns the UIDs when Atomic Deployments is enabled', async () => {
-    await inTemporaryDirectory(async (tmpDir: string) => {
-      // Given
-      const uiExtension = await testUIExtension({
-        directory: '/tmp/project/extensions/my-extension',
-        idEnvironmentVariableName: 'SHOPIFY_MY_EXTENSION_ID',
-      })
-      const app = testApp({
-        directory: tmpDir,
-        dotenv: {
-          path: joinPath(tmpDir, '.env'),
-          variables: {SHOPIFY_API_KEY: 'FOO', SHOPIFY_TEST_UI_EXTENSION_ID: 'BAR'},
-        },
-        allExtensions: [uiExtension],
-      })
-
-      // When
-      const got = getAppIdentifiers(
-        {
-          app,
-        },
-        testDeveloperPlatformClient({supportsAtomicDeployments: true}),
-      )
-
-      // Then
-      expect(got.app).toEqual('FOO')
-      expect((got.extensions ?? {})['test-ui-extension']).toEqual(uiExtension.uid)
     })
   })
 })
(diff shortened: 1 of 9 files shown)


alfonso-noriega:
* **#6135** <a href="https://app.graphite.dev/github/pr/Shopify/cli/6135?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6119** <a href="https://app.graphite.dev/github/pr/Shopify/cli/6119?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6111** <a href="https://app.graphite.dev/github/pr/Shopify/cli/6111?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6105** <a href="https://app.graphite.dev/github/pr/Shopify/cli/6105?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Shopify/cli/6105?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

isaacroldan (approved):

jamieguerrero (approved):

packages/app/src/cli/services/bundle.test.ts
```diff
@@ -25,7 +25,7 @@ describe('writeManifestToBundle', () => {
       } as unknown as AppInterface
 
       // When
-      await writeManifestToBundle(mockApp, tmpDir)
+      await writeManifestToBundle(mockApp, tmpDir, undefined)
```
jamieguerrero, line 28: Should we also test the case where we do pass in identifiers?

packages/app/src/cli/models/app/identifiers.ts
```diff
@@ -71,7 +71,10 @@ export async function updateAppIdentifiers(
   })
 
   const contentIsEqual = deepCompare(dotenvFile.variables, updatedVariables)
-  const writeToFile = !contentIsEqual && (command === 'deploy' || command === 'release')
+  const writeToFile =
+    !contentIsEqual &&
+    (command === 'deploy' || command === 'release') &&
+    !developerPlatformClient.supportsAtomicDeployments
```
jamieguerrero, line 77: Curious why we need to ensure the client doesn't support atomic deployments?

packages/app/src/cli/models/app/identifiers.test.ts
```diff
@@ -247,34 +243,4 @@ describe('getAppIdentifiers', () => {
       expect((got.extensions ?? {})['test-ui-extension']).toEqual('BAR')
     })
   })
-
-  test('returns the UIDs when Atomic Deployments is enabled', async () => {
```
jamieguerrero, line 251: I have a similar question elsewhere. Maybe I'm just missing some context here about what atomic deployments are. 😅

packages/app/src/cli/models/app/identifiers.ts
isaacroldan: atomicDeployments -> dev dash
non atomicDeployments -> partners

Is just a config variable we use to differenciate one from the other

isaacroldan: In this case, if using dev dash we don't want to write the `.env` file anymore

packages/app/src/cli/models/app/identifiers.test.ts
isaacroldan: This was injecting the UID as part of the `.env` content when using AtomicDeployments (meaning: using dev dash). But we removed this behaviour. (lines 111-113 from identifiers.ts)

packages/app/src/cli/services/bundle.test.ts
isaacroldan: That's already tested in `app.test.ts` which is the one actually creating the manifest. 
This test only ensures that the file is actually created in the filesystem, but the content is validated more extensively in the app.test.ts file.

packages/app/src/cli/models/app/identifiers.ts
jamieguerrero: Thank you!

Issue #5918, opened 2025-05-31
[Bug]: shopify theme dev hanging on certain themes


JasonHassold:
### Please confirm that you have:

- [x] Searched [existing issues](.) to see if your issue is a duplicate. (If you’ve found a duplicate issue, feel free to add additional information in a comment on it.)
- [x] Reproduced the issue in [the latest CLI version](https://www.npmjs.com/package/@shopify/cli).

### In which of these areas are you experiencing a problem?

Theme

### Expected behavior

Should sync theme quickly and then display the dev links

### Actual behavior

Hanging for 3+ minutes and eventually errors out.

```
❯ shopify theme dev
╭─ error ──────────────────────────────────────────────────────────────────────╮
│                                                                              │
│  Failed to perform the initial theme synchronization.                        │
│                                                                              │
│  The user aborted a request.                                                 │
│                                                                              │
╰──────────────────────────────────────────────────────────────────────────────╯
```

This issue is happening on a theme with a lot of reploChunk snippets and zipify pages files on the latest version 3.80.7, but doesn't happen on version 3.80.0. 

After comparing the verbose output of the two versions the main difference seems to be that it slows down and logs LiquidHTMLParsingErrors for lots of the files (below are 2 examples)

```
Error parsing Liquid file "snippets/reploChunk.0092dc1d-f311-4781-b5bd-957d223c2aeb.1.liquid" to detect updated file parts. LiquidHTMLParsingError: Attempting to close HtmlElement 'style' before it was opened
...
Error parsing Liquid file "snippets/reploChunk.0092dc1d-f311-4781-b5bd-957d223c2aeb.1.liquid" to detect updated file parts. LiquidHTMLParsingError: Attempting to close HtmlElement 'style' before it was opened
...
```

This file checking seems to be new or at least different. The files do contain issues, but at least in the case of replo it their code is intentionally split across multiple snippets and stitched together by rendering out multiple snippets so this isn't actually a problem.

I thought this was maybe apart of `theme check` so I tried adding these files to the .theme-check.yml file's ignore list but the files are still having this check run even with the files ignored.

`shopify theme push` seems to not perform this check and works without issue.

I believe the issue was introducted in [3.80.1](https://github.com/Shopify/cli/releases/tag/3.80.1)

### Verbose output

<details>
  <summary>Verbose output</summary>

  ```
  ❯ stheme dev --verbose
2025-05-31T16:31:01.192Z: Running command theme dev
[...]

PR #6112, opened 2025-07-15
No longer abort theme GraphQL requests on the client side

karreiro:
### WHY are these changes being introduced?

Re: https://github.com/Shopify/cli/issues/5918 / https://github.com/Shopify/cli/issues/6062

### WHAT does this pull request do?

Requests to the themes APIs should not be aborted on the client side and should instead rely on the server-side timeout.

Even after increasing the timeout, we have still been seeing unexpected [closed](https://github.com/Shopify/cli/issues/6062) connections. This started when the stricter `requestBehaviour` was introduced.

Theme setups can vary a lot—some have a few large files, while others have many small ones. Relying on the server-side timeout seems to be more accurate and has proven more stable. Partners did not report friction when we used this approach before.

### How to test your changes

Run `shopify theme push` on your own theme.

### Post-release steps

N/A

### Measuring impact

How will we know this change was effective? Please select one:

- [ ] n/a - this doesn't need measurement, such as a linting rule or simple bug fix
- [x] Existing analytics will cover this change
- [ ] PR includes analytics updates to measure impact

### Checklist

- [x] Considered possible cross-platform impacts (Mac, Linux, Windows)
- [x] Considered possible [documentation](https://shopify.dev) changes

### DIFF
--- a/.changeset/gorgeous-colts-retire.md
+++ b/.changeset/gorgeous-colts-retire.md
@@ -1,0 +1,6 @@
+---
+'@shopify/cli-kit': patch
+'@shopify/theme': patch
+---
+
+No longer abort theme GraphQL requests on the client side
--- a/packages/cli-kit/src/public/node/themes/api.test.ts
+++ b/packages/cli-kit/src/public/node/themes/api.test.ts
@@ -38,8 +38,7 @@ const themeAccessSession = {...session, token: 'shptka_token'}
 const sessions = {CLI: session, 'Theme Access': themeAccessSession}
 const expectedApiOptions = expect.objectContaining({
   maxRetryTimeMs: 90000,
-  timeoutMs: 90000,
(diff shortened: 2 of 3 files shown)


isaacroldan (approved):

EvilGenius13 (approved):

EvilGenius13:
Thank you @karreiro for the PR! Would adding the network behaviour to `metafieldDefinitionsByOwnerType` be beneficial at all?

karreiro:
Thank you for the review, @EvilGenius13!

Since `metafieldDefinitionsByOwnerType` is a regular GraphQL endpoint with a predictable body size and isn't used for downloading or uploading assets, we don't need a special network policy there. Still, it's a great point to think critically about each endpoint.

Thanks again for the review!

PR #6114, opened 2025-07-15
Add overwrite prompt for store export if file exists

skelley-ton:

Fixes https://github.com/Shopify/workflows-operations/issues/3364

Screenshot 2025-07-15 at 3 21 13 PM

Adds a confirmation prompt when attempting to export store data to a file that already exists. This prevents accidental overwriting of existing files by:

1. Update `confirmExportPrompt` function so that it asks users if they want to overwrite an existing file if the file exists
2. Showing the appropriate confirmation prompt based on whether the file exists or not

### How to test your changes?
run this command twice, the first time you will see the normal flow, the second time you will see the overwrite flow

`pnpm shopify store copy --from-store=source --to-file=file --mock`
### Measuring impact

### Checklist

### DIFF
--- a/packages/store/src/services/store/operations/store-export.ts
+++ b/packages/store/src/services/store/operations/store-export.ts
@@ -13,6 +13,7 @@ import {renderExportResult} from '../../../prompts/export_results.js'
 import {confirmExportPrompt} from '../../../prompts/confirm_export.js'
 import {Task, renderTasks} from '@shopify/cli-kit/node/ui'
 import {outputInfo} from '@shopify/cli-kit/node/output'
+import {fileExistsSync} from '@shopify/cli-kit/node/fs'
 
 export class StoreExportOperation implements StoreOperation {
   fromArg: string | undefined
@@ -36,7 +37,8 @@ export class StoreExportOperation implements StoreOperation {
     const apiShopId = await this.validateShop(sourceShopDomain)
 
     if (!flags['no-prompt']) {
-      if (!(await confirmExportPrompt(sourceShopDomain, toFile))) {
+      const fileExists = fileExistsSync(toFile)
+      if (!(await confirmExportPrompt(sourceShopDomain, toFile, fileExists))) {
         outputInfo('Exiting.')
         process.exit(0)
       }
--- a/packages/store/src/prompts/confirm_export.test.ts
+++ b/packages/store/src/prompts/confirm_export.test.ts
@@ -5,7 +5,7 @@ import {describe, expect, vi, test} from 'vitest'
 vi.mock('@shopify/cli-kit/node/ui')
 
 describe('confirmExportPrompt', () => {
-  test('returns true when user confirms', async () => {
+  test('returns true when user confirms export to new file', async () => {
     const toFile = 'data.sqlite'
     const fromStore = 'shop.myshopify.com'
     const message = `Export data from ${fromStore} to ${toFile}?`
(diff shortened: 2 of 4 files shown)


skelley-ton:
* **#6114** <a href="https://app.graphite.dev/github/pr/Shopify/cli/6114?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Shopify/cli/6114?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/store/src/services/store/operations/store-export.ts
```diff
@@ -36,7 +38,12 @@ export class StoreExportOperation implements StoreOperation {
     const apiShopId = await this.validateShop(sourceShopDomain)
 
     if (!flags['no-prompt']) {
-      if (!(await confirmExportPrompt(sourceShopDomain, toFile))) {
+      if (fileExistsSync(toFile)) {
+        if (!(await confirmOverwritePrompt(sourceShopDomain, toFile))) {
+          outputInfo('Exiting.')
+          process.exit(0)
+        }
+      } else if (!(await confirmExportPrompt(sourceShopDomain, toFile))) {
```
rzaharenkov: I am slightly concerned that with file conflict we are getting less informative confirmation that is only asking to override the file and export prompt is not being shown. 

I would make it so export confirmation prompt is displayed first and then when user confirmed the intent to export from a store to a file we show the next prompt to confirm file overwrite.

nit, I would extract the confirmation logic into separate function since it becomes really bulky.

rzaharenkov: Other issue I can see with it is that `no-prompt` mode will automatically overwrite the file. I would expect it to actually fail with the error stating that the file already exists. I think `no-prompt` should be still safe. E.g. when I use it in automation I would love it to not silently overwrite the file if I messed up with the logic. I don't have strong opinions on that so I would love to hear what other people think about.

skelley-ton: I paired on this with @Rusty-UX, so will let him chime in!

Rusty-UX: 👋 @rzaharenkov We discussed having the overwrite then the export prompt. IMO the proposed option in the PR is the strongest, because it clarifies the export operation will proceed, while also communicating the overwrite. IMO it's less verbose, but the information communicated is essentially the same.

On the comment of `no-prompt` I have less of an opinion on this, but I agree with you that it should fail instead of silently overwriting. LMK if that means we need to rethink the prompt content displayed for the export+overwrite.

rzaharenkov: OK, I am good to have a single confirmation, I think it's more user friendly.

nit, I would love to see some kind of extracting this logic. Maybe it could be arguments passed to the same function `confirmExportPrompt`, or it could be another function. But not a blocker for this PR.

Also would love a change to avoid silent overwrite for `no-prompt` mode. But it could definitely be a follow up since this behavior is not changing in this PR and it's positive incremental change.

rzaharenkov (approved):

packages/store/src/services/store/operations/store-export.ts
```diff
@@ -36,7 +38,12 @@ export class StoreExportOperation implements StoreOperation {
     const apiShopId = await this.validateShop(sourceShopDomain)
 
     if (!flags['no-prompt']) {
```
odacrem, line 39: I agree that this section is a little unwieldy; I would prefer to have this operation just call confirmExportPrompt(sourceShopDomain, toFile) 

and then have the prompt sort out whether it needs confirm about the overwrite or not.  

the unit test for the operation can just test that it calls the confirmExport prompt with the right parameters; 

and the confirmExportPrompt tests can confirm that asks the right questions based on no-prompt flags, file existence etc

odacrem (changes requested):

odacrem (approved):

PR #6121, opened 2025-07-16
Replace DevSessionUI footer with tabbed interface to display more app information

nickwesselman:
The new `app dev` command for Dev Platform makes the app URL less accessible. This PR adds a new App Information tab accessible via the `i` key in `app dev`, which includes the app URL and more. The tabs can also be navigated via arrow keys.

It also adds the App URL as an output to the app dev logs, via dev session messaging on the app home component.

## Demo / Screenshots

Note this video includes some resizing artifacting. We have this in `app dev` today as well. It seems impossible to avoid entirely in Ink/terminals, but this PR limits the size of he border/tabs to 90% width, which mitigates the issue.

https://github.com/user-attachments/assets/d61541bb-a1d2-41f0-abc7-03bd99072ac7

image

image

image

image

## Summary
- Replaces the simple horizontal line separator in the DevSessionUI footer with a visual tabbed interface
- Implements connected box drawing character borders (╒═╤═╕, └─┴─┘) that create a cohesive tabbed design
- Refactors input handling to use object-based tab structure with Status, App Info, and Quit tabs
- Moves keyboard shortcuts (p, g) into tab-specific data structures for better organization
- Adds comprehensive app information display including app name, URL, config path, dev store, and organization
- Always shows all tabs regardless of app readiness state for consistent UX

## Test plan
- [x] DevSessionUI tests pass (16/16 tests)
- [x] Linting passes with no errors
- [x] Visual design shows connected tabbed interface with proper box drawing characters
- [x] Tab switching works correctly (s for Status, i for App Info, q for Quit)
- [x] Keyboard shortcuts work within Status tab (p for preview, g for GraphiQL)
- [x] Terminal width calculations prevent character wrapping issues
- [x] App information displays correctly with all provided fields

🤖 Generated with [Claude Code](https://claude.ai/code)


packages/app/src/cli/services/dev/ui/components/TabPanel.test.tsx
```diff
@@ -0,0 +1,252 @@
...
+
+describe('TabPanel', () => {
+  const mockAction = vi.fn()
+  const mockShortcutAction = vi.fn()
+
+  beforeEach(() => {
+    // Reset mocks before each test
+    mockAction.mockReset()
+    mockShortcutAction.mockReset()
+  })
```
gonzaloriestra: I think you don't need this, mocks are reset by default

nickwesselman:
> The App URL and dev store should be hyperlinks (they appear underscored)

🤔 I'm not sure on this as it's meant to be informational vs clicked. Opening the App URL directly with our Remix template is actually quite confusing -- it lands you on a mock marketing page for the app. The dev store could be useful though, maybe even with an additional admin link.

nickwesselman:
Implemented PR feedback and more:
* Changed to a simple inverse for the selected tab, so that things work better in light mode and other themes
* Added arrow-based navigation
* Reformatted the quit "tab" to differentiate it as an action
* Added a hyperlink for the dev store, and a link to dev store admin
* Fixed App Info display for extension-only apps (don't display empty app info)

I also tested on Windows and Linux (via Ubuntu on WSL).

megthesmith:
> * Not sure how to fix this, but the `(q) Quit` tab is actually not a usable tab, but more like a button... Maybe a different style or place?

I agree with this feedback. If the q tab is genuinely not navigable, but more a quick action you can take to terminate the process, I'd still house it in the main screen with the other short-keys to deal with the ongoing server process. Then the tabs can remain purely about accessing the different tabbed states in the CLI, and be cleanly and consistently arrow-navigated too.

packages/app/src/cli/services/dev/ui/components/TabPanel.tsx
```diff
@@ -0,0 +1,151 @@
...
+
+interface TabPanelProps {
+  tabs: {[key: string]: Tab}
+  initialActiveTab?: string
+}
+
+export const TabPanel: React.FunctionComponent<TabPanelProps> = ({tabs, initialActiveTab}) => {
+  const {isRawModeSupported: canUseShortcuts} = useStdin()
+  const firstTabKey = Object.keys(tabs)[0]
+  const [activeTab, setActiveTab] = useState<string | undefined>(initialActiveTab ?? firstTabKey)
```
isaacroldan: Can this really be undefined? I feel like `TabPanelProps` could have `initialActiveTab` as a mandatory property.

packages/app/src/cli/services/dev/ui/components/TabPanel.tsx
```diff
@@ -0,0 +1,151 @@
...
+        // Handle arrow key navigation for tabs with content
+        if (key?.leftArrow || key?.rightArrow) {
+          const contentTabs = Object.entries(tabs).filter(([_, tab]) => tab.content)
+          if (contentTabs.length > 1) {
+            const currentIndex = contentTabs.findIndex(([tabKey]) => tabKey === activeTab)
+            if (currentIndex !== -1) {
+              let newIndex
+              // Right arrow
+              if (key?.rightArrow) {
+                newIndex = currentIndex + 1 < contentTabs.length ? currentIndex + 1 : currentIndex
```
isaacroldan: no need to change it if you prefer your way, but is this easier to follow?:
```ts
newIndex = min(currentIndex + 1, contentTabs.length - 1)
```

and

```ts
newIndex = max(currentIndex - 1, 0)
```

Also I think we can simplify this code to have less indented ifs:
```ts
        if (key?.leftArrow || key?.rightArrow) {
          const contentTabs = Object.entries(tabs).filter(([_, tab]) => tab.content)
          const currentIndex = contentTabs.findIndex(([tabKey]) => tabKey === activeTab)
          if (currentIndex === -1) return
          const modifier = key.leftArrow ? -1 : 1
          const newIndex = Math.min(Math.max(currentIndex + modifier, 0), contentTabs.length - 1)
          const newTabEntry = contentTabs[newIndex]
          if (newTabEntry) {
            setActiveTab(newTabEntry[0])
          }
        }
```

Does this make sense? is it easier or harder to follow?

isaacroldan:
Is easy to break it by resizing the terminal. Should the tab container width be limited to the size of the tabs to prevent this? 

Since it now takes the whole width, any minor resize breaks it, with a smaller container we prevent that unless you make a big resize.

nickwesselman:
> Is easy to break it by resizing the terminal. Should the tab container width be limited to the size of the tabs to prevent this?

Yeah @megthesmith has some ideas for the design that may help with this, I am going to try some iterations.

packages/app/src/cli/services/dev/ui/components/TabPanel.tsx
nickwesselman: You caught some Claude code that I didn't interrogate... 😅 ... agree your version is much more readable.

nickwesselman: Additionally updated logic to also allow looping from end to start and vice versa, and to support the tab key.

nickwesselman:
> Is easy to break it by resizing the terminal. Should the tab container width be limited to the size of the tabs to prevent this?

Latest design and use of flexbox fixes this 👍

nickwesselman:
Implemented some additional resize handling to improve display and reduce artifacts. Or experience w/ Ink though is that you can't make these go away entirely. Especially when resizing quickly.

nickwesselman:
Retested the latest UX on Windows and Linux (via WSL).

packages/app/src/cli/services/dev/ui/components/DevSessionUI.test.tsx
```diff
@@ -436,4 +439,64 @@ describe('DevSessionUI', () => {
...
+    )
+
+    await waitForInputsToBeReady()
+
+    // Then - should show Dev status tab content without interactive tabs
+    const output = renderInstance.lastFrame()!
+    expect(output).not.toContain('(d) Dev status')
+    expect(output).not.toContain('(a) App info')
+    expect(output).not.toContain('(q) Quit')
+    expect(output).not.toContain('╒═══')
```
gonzaloriestra: We can remove these references to the first iteration

packages/app/src/cli/services/dev/ui/components/DevSessionUI.tsx
```diff
@@ -207,6 +183,91 @@ const DevSessionUI: FunctionComponent<DevSesionUIProps> = ({
...
+              ['Dev store:', {link: {url: `https://${shopFqdn}`}}],
+              ['Dev store admin:', {link: {url: `https://${shopFqdn}/admin`}}],
+              ['Org:', organizationName ?? ''],
+            ].filter(([, value]) => value)}
+          />
+        </Box>
+      ),
+    },
+    q: {
+      label: 'to quit',
```
gonzaloriestra: Can we show `(q) Quit` instead to be consistent?

packages/app/src/cli/services/dev/ui/components/DevSessionUI.tsx
```diff
@@ -207,6 +183,91 @@ const DevSessionUI: FunctionComponent<DevSesionUIProps> = ({
...
+    // eslint-disable-next-line id-length
+    a: {
+      label: 'App info',
+      content: (
+        <Box flexDirection="column">
+          <TabularData
+            tabularData={[
+              ['App:', appName ?? ''],
+              ['App URL:', appURL ?? ''],
+              ['Config:', configPath?.split('/').pop() ?? ''],
```
gonzaloriestra, line 197: Would it be useful to show the full path (`/Users/Nick/tests/app2/shopify.app.toml`) instead of the file name?

gonzaloriestra (approved):
Great! Thanks for following the feedback.

One last idea. No need to tackle it now, but couldn't we remove these full URLs from the Dev status tab?

```
Preview URL: https://vacuum-cultures-computed-walter.trycloudflare.com/extensions/dev-console
GraphiQL URL: http://localhost:3457/graphiql
```

Instead, we could add hyperlinks above, so that the footer is much smaller:

```
──────────────────────────────────────────────────────────────────────────────────────────────────
│ (d) Dev status │ (a) App info │ (s) Store info │                                        (q) Quit

 ✅ Ready, watching for changes in your app

 › (p) Preview
 › (g) GraphiQL (Admin API)
```

The URL can be copied from the terminal on right click if needed, and we could show it inline when it's not supported:

```
──────────────────────────────────────────────────────────────────────────────────────────────────
│ (d) Dev status │ (a) App info │ (s) Store info │                                        (q) Quit

 ✅ Ready, watching for changes in your app

 › (p) Preview: https://vacuum-cultures-computed-walter.trycloudflare.com/extensions/dev-console
 › (g) GraphiQL (Admin API): http://localhost:3457/graphiql
```

Or if you think it's still useful to print the URLs, why not showing this latest version always?

packages/app/src/cli/services/dev/ui/components/DevSessionUI.tsx
nickwesselman: Elsewhere we use just the file name referencing the config, since they are always in the app root. So I think this is OK and less noisy.

packages/app/src/cli/services/dev/ui/components/DevSessionUI.test.tsx
nickwesselman: Good catch, fixed

packages/app/src/cli/services/dev/ui/components/DevSessionUI.tsx
nickwesselman: Yes good call. I updated the existing shortcut keys in the `Dev status` tab too

nickwesselman:
There is maybe more iteration we could do on the URL display here, but in the interest of making the next release, I think we should ship as-is. Thanks all for the great iteration on this!

nickwesselman:
Did one more round of cross-platform testing, good to 🚢

PR #6124, opened 2025-07-17
Update multi-environment theme command behaviour

dejmedus:

Previously multi-environment theme commands required `force = "true"` to be set in `shopify.theme.toml` since multi env commands must run without interactions. This PR allows flags to be provided via the CLI and adds a confirmation prompt that will run prior to concurrent command calls.

- Accepts command flags from both `shopify.theme.toml` and the CLI, with [CLI flags taking precedence](https://shopify.dev/docs/storefronts/themes/tools/cli/environments#how-environment-flags-are-applied).
- Prompts for confirmation before running commands if the command accepts force (ex. `delete`, `push`, `pull`) and `--force` has not been set.
- Catches/handles command errors in `renderConcurrent` to prefix environment name to the error message and ensure that subsequent environment's commands will still be called.

https://github.com/user-attachments/assets/e8b295de-1363-4904-abac-1ef424f75e4e

### How to test your changes?

- pull down the branch `gh checkout 6124`
- add a `shopify.theme.toml` file

```
[environments.store1]
theme = "<theme>"
store = "<store>"

[environments.store2]
theme = "<theme>"
store = "<store>"

[environments.missing]
theme = "<theme>"
store = "<store>"

[environments.fake]
theme = "1000000"
store = "fake.myshopify.com"
password  = "not-real"
```

</details> 

- `info` and `list` commands both currently work with multiple envs
- since neither allow `--force` (and confirmation prompts arent shown on commands that dont accept `--force`) the force flag can be manually added to test confirmation prompt

``` ts
// packages/theme/src/cli/commands/theme/list.ts

force: Flags.boolean({
  char: 'f',
  description: 'Skip confirmation',
  env: 'SHOPIFY_FLAG_FORCE',
}),
```
``` sh
pnpm shopify theme list -e store1 -e store2 -e missing -e fake
```
- environments missing flags should display a warning and the command shouldn't be executed
- environments with incorrect flags (ex. `password = "fake"`) should have the environment name prefixed to the error message. Commands should fail gracefully and not prevent commands from running in other environments.
- commands run with a single environment or no environment should work as normal

### Post-release steps

- more commands will be updated to allow multiple environments
- multi-env documentation

### Measuring impact

### Checklist

### DIFF
--- a/packages/theme/src/cli/utilities/theme-command.ts
+++ b/packages/theme/src/cli/utilities/theme-command.ts
@@ -4,7 +4,13 @@ import {Input} from '@oclif/core/interfaces'
 import Command, {ArgOutput, FlagOutput} from '@shopify/cli-kit/node/base-command'
 import {AdminSession, ensureAuthenticatedThemes} from '@shopify/cli-kit/node/session'
 import {loadEnvironment} from '@shopify/cli-kit/node/environments'
-import {renderWarning, renderConcurrent, renderError} from '@shopify/cli-kit/node/ui'
+import {
+  renderWarning,
+  renderConcurrent,
+  renderConfirmationPrompt,
+  RenderConfirmationPromptOptions,
+  renderError,
+} from '@shopify/cli-kit/node/ui'
 import {AbortController} from '@shopify/cli-kit/node/abort'
 import type {Writable} from 'stream'
 
@@ -53,6 +59,7 @@ export default abstract class ThemeCommand extends Command {
     // Parse command flags using the current command class definitions
     const klass = this.constructor as unknown as Input<TFlags, TGlobalFlags, TArgs> & {
       multiEnvironmentsFlags: string[]
+      flags: FlagOutput
     }
     const requiredFlags = klass.multiEnvironmentsFlags
     const {flags} = await this.parse(klass)
@@ -70,6 +77,13 @@ export default abstract class ThemeCommand extends Command {
     // Multiple environments
     const environmentsMap = await this.loadEnvironments(environments, flags)
     const validationResults = await this.validateEnvironments(environmentsMap, requiredFlags)
+
+    const commandAllowsForceFlag = 'force' in klass.flags
+
+    if (commandAllowsForceFlag && !flags.force) {
+      const confirmed = await this.showConfirmation(this.constructor.name, requiredFlags, validationResults)
+      if (!confirmed) return
+    }
 
     await this.runConcurrent(validationResults.valid)
   }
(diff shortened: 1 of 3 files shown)


packages/theme/src/cli/utilities/theme-command.ts
```diff
@@ -41,6 +47,7 @@ export default abstract class ThemeCommand extends Command {
   async command(
     _flags: FlagValues,
     _session: AdminSession,
+    _multiEnvironment = false,
```
dejmedus, line 50: This will be helpful in certain commands to toggle off behaviour like individual confirmations

packages/theme/src/cli/utilities/theme-command.ts
```diff
@@ -60,72 +67,176 @@ export default abstract class ThemeCommand extends Command {
...
-          if (!this.validConfig(environmentConfig as FlagValues, requiredFlags, environment)) return
-
-          const session = sessions[environment]
-
-          if (!session) {
-            throw new AbortError(`No session found for environment ${environment}`)
+          try {
+            await this.command(flags, session, true, {stdout, stderr})
+
+            // eslint-disable-next-line no-catch-all/no-catch-all
```
dejmedus, line 207: Without catching/handling, some command errors (like authentication failures) would prevent subsequent commands from running

packages/theme/src/cli/utilities/theme-command.ts
```diff
@@ -60,72 +61,125 @@ export default abstract class ThemeCommand extends Command {
 
     // Single environment or no environment
     if (environments.length <= 1) {
-      const session = await this.ensureAuthenticated(flags)
+      const session = await this.createSession(flags)
```
graygilmore, line 71: Nit: in the future it'd be good to split a name change into a separate commit for an easier git diff!

packages/theme/src/cli/utilities/theme-command.ts
```diff
@@ -60,72 +67,176 @@ export default abstract class ThemeCommand extends Command {
...
       await this.command(flags, session)
       return
     }
 
     // Multiple environments
-    const sessions: {[storeFqdn: string]: AdminSession} = {}
+    const environmentsMap = await this.loadEnvironments(environments, flags)
+    const validationResults = await this.validateEnvironments(environmentsMap, requiredFlags)
+
+    const commandAllowsForceFlag = 'force' in (this.constructor as typeof ThemeCommand)?.flags
```
graygilmore: Could we use `klass` here since we already type cast on L60?
```suggestion
    const commandAllowsForceFlag = 'force' in klass.flags
```

graygilmore (approved):
🎩 went great!

packages/theme/src/cli/utilities/theme-command.ts
dejmedus: Sounds good! So its easier to differentiate "cosmetic" changes from actual logic changes/additions?

packages/theme/src/cli/utilities/theme-command.ts
dejmedus: 100%, thanks!

packages/theme/src/cli/utilities/theme-command.ts
graygilmore: Exactly!

graygilmore (approved):

PR #6125, opened 2025-07-17
Fix performance issue on `shopify theme dev` command

karreiro:

Fixes https://github.com/Shopify/cli/issues/5918

This PR addresses two problems: the dev server hanging during startup and dev server requests occasionally taking up to a minute to load.

When we introduced support for hot-reloading `{% stylesheet %}` and `{% javascript %}` tags, we started using the `@shopify/liquid-html-parser` to extract JavaScript and CSS content from Liquid files and locally generate compiled assets.

Using a parser is the most accurate way to get this content, but parsing has a performance cost. Even though we used caches to reduce the impact, there was still a noticeable slowdown for large themes, especially during cache warm-up.

This PR switches from using the Liquid parser to using a regex to support hot-reloading of the `{% stylesheet %}` and `{% javascript %}` tags. We might bring the parser back later (when we get some performance gains), but it's important that the parse time stays ideally below 1ms for ~200Kb files.

### How to test your changes?

- Run `shopify theme dev` in a theme with large snippet files.

**Demos:**

<details>
<summary>Before</summary>

https://github.com/user-attachments/assets/9d321ff0-180f-4110-adef-17f247da0949

https://github.com/user-attachments/assets/f7c98e23-3f5b-4d63-b326-edfce1f095c7

</details>

<details>
<summary>After</summary>

https://github.com/user-attachments/assets/6c96946f-ab27-4941-a30a-5f37dda37ad7

https://github.com/user-attachments/assets/7a3a23bc-2186-4bd2-a52e-ac40df86fec8

</details>

### Post-release steps

N/A

### Measuring impact

### Checklist

### DIFF
--- a/packages/theme/src/cli/utilities/theme-environment/liquid-tag-content.test.ts
+++ b/packages/theme/src/cli/utilities/theme-environment/liquid-tag-content.test.ts
@@ -1,0 +1,122 @@
+import {getLiquidTagContent} from './liquid-tag-content.js'
+import {describe, expect, test} from 'vitest'
+
+const scenarios = [
+  {
+    name: 'handles simple cases with all tags present',
+    fileContent: `
+      {% stylesheet %}
+        .foo { color: red; }
+      {% endstylesheet %}
+      {% javascript %}
+        console.log('hi');
+      {% endjavascript %}
+      {% schema %}
+        { "name": "Test" }
+      {% endschema %}
+      <div>Other content</div>
+    `,
+    expected: {
+      stylesheet: '.foo { color: red; }',
+      javascript: "console.log('hi');",
+      schema: '{ "name": "Test" }',
+    },
+  },
+  {
+    name: 'handles simple cases with extra whitespace',
+    fileContent: `
+      {%    stylesheet    %}
+        .baz { color: green; }
+      {%    endstylesheet    %}
+      {%javascript%}
+        alert('hi');
+      {%endjavascript%}
+      {%   schema   %}
+        { "name": "Whitespace" }
+      {%   endschema   %}
+    `,
+    expected: {
+      stylesheet: '.baz { color: green; }',
+      javascript: "alert('hi');",
+      schema: '{ "name": "Whitespace" }',
+    },
+  },
+  {
+    name: 'handles dash cases with all tags present',
+    fileContent: `
+      {%- stylesheet -%}
+        .foo { color: red; }
+      {%- endstylesheet -%}
+      {%- javascript -%}
+        console.log('hi');
+      {%- endjavascript -%}
+      {%- schema -%}
+        { "name": "Test" }
+      {%- endschema -%}
+      <div>Other content</div>
+    `,
+    expected: {
+      stylesheet: '.foo { color: red; }',
+      javascript: "console.log('hi');",
+      schema: '{ "name": "Test" }',
+    },
+  },
+  {
+    name: 'detects dash cases with all tags present and whitespace',
+    fileContent: `
+      {%- stylesheet -%}
+        .foo { color: red; }
+      {%- endstylesheet -%}
+      {% javascript       %}
+        console.log('hi');
+      {%- endjavascript -%}
+      {%      schema -%}
+        { "name": "Test" }
+      {%-       endschema -%}
+      <div>Other content</div>
+    `,
+    expected: {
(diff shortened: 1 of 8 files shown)


packages/theme/src/cli/utilities/theme-environment/liquid-tag-content.ts
```diff
@@ -0,0 +1,35 @@
...
+ * runtime performance.
+ *
+ * @param liquid - The full Liquid template as a string.
+ * @param tag - The Liquid tag to extract content from.
+ *
+ * @returns The tag content, or undefined if not found.
+ *
+ */
+export function getLiquidTagContent(liquid: string, tag: 'javascript' | 'stylesheet' | 'schema'): string | undefined {
+  const tagRE = new RegExp(`{%-?\\s*${tag}\\s*-?%}([^%]*){%-?\\s*end${tag}\\s*-?%}`)
```
charlespwd: I suspected part of the slowdown was from this regex actually. Maybe we can do it in 2 parts rather than the slow `([^%]*){%?`. Match the start tag, then do another match starting at the endIndex of the other one?

packages/theme/src/cli/utilities/theme-environment/liquid-tag-content.ts
```diff
@@ -26,10 +26,29 @@
...
+  const closeTagRE = new RegExp(`{%-?\\s*end${tag}\\s*-?%}`)
+
+  const openMatch = openTagRE.exec(liquid)
+
+  if (!openMatch) {
+    return
+  }
+
+  const startIndex = openMatch.index + openMatch[0].length
+  const closeMatch = closeTagRE.exec(liquid)
```
charlespwd: IIRC you can use `.lastIndex` on regexes to make them start from somewhere and simplify this logic.

```
closeTagRe.lastIndex = startIndex
closeTagRe.exec(liquid); // starts from lastIndex
```

charlespwd: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/lastIndex

packages/theme/src/cli/utilities/theme-environment/liquid-tag-content.test.ts
```diff
@@ -0,0 +1,122 @@
+import {getLiquidTagContent} from './liquid-tag-content.js'
+import {describe, expect, test} from 'vitest'
+
+const scenarios = [
+  {
+    name: 'handles simple cases with all tags present',
+    fileContent: `
+      {% stylesheet %}
+        .foo { color: red; }
+      {% endstylesheet %}
```
frandiox, line 10: We have somewhat similar tests in https://github.com/Shopify/cli/blob/d1418e9388e5a09fd06434be9b2f7490a566477a/packages/theme/src/cli/utilities/theme-environment/theme-environment.test.ts#L285

In case you want to compare / dedup

(Also, the fact those tests pass is a good thing for this change)

packages/theme/src/cli/utilities/theme-environment/liquid-tag-content.ts
karreiro: Cursor proposed `lastIndex` in the first version and I wrongly assumed the `lastIndex` would require the `g` flag (what would be slower), which is not the case -- good catch! I will apply that 🙏

aswamy (approved):

packages/theme/src/cli/utilities/theme-environment/liquid-tag-content.test.ts
karreiro: Thank you, @frandiox! I did consider that, but since I expect we'll bring the parser back at some point (once we make it faster), I'd like to leave those tests there as well. I expect that we'll remove this new test suite when we switch back to the parser. But well spotted :)

packages/theme/src/cli/utilities/theme-environment/liquid-tag-content.ts
karreiro: Changes applied 🙏

karreiro:
@Shopify/app-inner-loop, could you please take a look when you have a moment?

isaacroldan (approved):

PR #6126, opened 2025-07-17
Add error handling for unauthorized errors

szluzero:

Closes https://github.com/Shopify/workflows-operations/issues/3389

Adds error handling for unauthorized errors for BDM operations

Console outputs these error messages depending on if its copy, export, or import.

Screenshot 2025-07-17 at 4 24 45 PM
Screenshot 2025-07-17 at 4 24 23 PM

Through testing I found that we can either receive `ClientError` or `GraphqlClientError`. I didn't realize `GraphqlClientError` was a thing previously, but this should cover the rest of api errors we can get from `api-client.ts` calls

### How to test your changes?

The way I tested it was to get added as a staff on an EA-enabled shop (in this case `rusty-testing-stuff`) but not have the `bulk-data` permissions enabled for me in admin. Then running the export/copy commands e.g. `pnpm shopify store copy --from-store rusty-testing-stuff` and `pnpm shopify store copy --from-store rusty-testing-stuff --to-store steve-product-set-test-1`

I was unable to test import because I don't have access to `stagedUploadCreate`, so I'm getting that error instead. Working on that
### Measuring impact

### Checklist


szluzero:
* **#6126** <a href="https://app.graphite.dev/github/pr/Shopify/cli/6126?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Shopify/cli/6126?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/store/src/services/store/api/api-client.ts
```diff
@@ -109,13 +109,23 @@
   }
 
   private handleError(error: unknown, operationName: string): OperationError | ValidationError {
-    if (error instanceof OperationError || error instanceof ValidationError) {
-      return error
-    }
```
szluzero, line 114: this doesn't actually do anything I believe, so I removed it

packages/store/src/services/store/api/api-client.ts
```diff
@@ -128,4 +138,14 @@
...
   }
+
+  private extractRequestIdFromErrorMessage(error: unknown): string | undefined {
+    if (error && typeof error === 'object' && 'message' in error) {
+      const message = (error as {message?: string}).message
+      if (message) {
+        return message.match(/Request ID: ([\w-]+)/)?.[1] ?? undefined
+      }
+    }
+    return undefined
```
rzaharenkov, line 193: nit, I think it should be OK to parse request ID from the response. However we could have generated it when submitting the request and set request headers. That way we have a guarantee to always have a request ID.

packages/store/src/services/store/api/api-client.ts
```diff
@@ -109,13 +109,23 @@ export class ApiClient implements ApiClientInterface {
...
       return new OperationError(operationName, ErrorCodes.GRAPHQL_API_ERROR, {}, requestId)
     }
+
+    if (error?.constructor?.name === 'GraphQLClientError') {
+      const requestId = this.extractRequestIdFromErrorMessage(error)
+      const errors = (error as {errors?: {extensions?: {code?: string}}[]})?.errors
+      const isUnauthorized = errors?.some((err) => err.extensions?.code === 'UNAUTHORIZED')
+
+      if (isUnauthorized) {
+        return new OperationError(operationName, ErrorCodes.UNAUTHORIZED, {}, requestId)
```
rzaharenkov: It took me a while to figure out you raise the error here without any parameters and then handle and re-raising it later. As result access error for `getStoreDetails` will not be enriched with store details.

I would like it much better if you pass down `params` to this method from the API client. I think that way code would be much more straightforward.

packages/store/src/services/store/errors/errors.ts
```diff
@@ -110,6 +116,16 @@ function generateErrorMessage(code: string, params?: ErrorParams, requestId?: st
...
+    case ErrorCodes.UNAUTHORIZED:
+      return 'You are not authorized to perform this operation'
+    case ErrorCodes.UNAUTHORIZED_EXPORT:
+      return `You are not authorized to export bulk data from store "${params?.storeName}"\n\nTo export you'll need 'bulk data export' permission`
+    case ErrorCodes.UNAUTHORIZED_IMPORT:
+      return `You are not authorized to import bulk data to store "${params?.storeName}"\n\nTo import you'll need 'bulk data import' permission`
+    case ErrorCodes.UNAUTHORIZED_COPY:
+      return `You are not authorized to copy data between these stores\n\nTo export data from "${params?.sourceStoreName}"\n• You'll need 'bulk data export' permission\n\nTo import data to "${params?.targetStoreName}"\n• You'll need 'bulk data import' permission`
+
     default:
```
skelley-ton, line 136: can we add `the` before the permission names?

rzaharenkov: I think we should either use permission handle `bulk_data_export` or the reference to permission in UI (e.g. `Bulk Data > Export`). I think this reference `bulk data export` might be a bit misleading. @Rusty-UX

packages/store/src/services/store/api/api-client.ts
szluzero: I think the issue is that the majority of api errors get converted from `ClientError` (which has request id from response header) into `GraphQLClientError` which doesn't have access to headers (can't access request or response). So if we want to make this more reliable we'd have to make changes to `cli-kit`

packages/store/src/services/store/api/api-client.ts
szluzero: oh, for some reason I was thinking `store-copy` would call the export and import apis too, whoops!

packages/store/src/services/store/operations/store-import.ts
```diff
@@ -85,24 +85,36 @@
...
 
-    if (!importResponse.bulkDataStoreImportStart.success) {
-      const errors = importResponse.bulkDataStoreImportStart.userErrors.map((error) => error.message).join(', ')
-      throw new OperationError('import', ErrorCodes.BULK_OPERATION_FAILED, {
-        errors,
-        operationType: 'import',
-      })
+      const operationId = importResponse.bulkDataStoreImportStart.operation.id
+      return this.apiClient.pollBulkDataOperation(apiShopId, operationId, bpSession)
+    } catch (error) {
```
rzaharenkov: Could this catch statement be a part of `startBulkDataStoreImport`? I think it might fir better to API client.

szluzero: you reviewed again too fast! 😆 right after this comment I just updated the PR to not touch any of the store-import/copy/export files

packages/store/src/services/store/api/api-client.ts
```diff
@@ -108,14 +115,35 @@ export class ApiClient implements ApiClientInterface {
...
+
+      if (isUnauthorized) {
+        if (operationName === 'startBulkDataStoreExport') {
+          return new OperationError(operationName, ErrorCodes.UNAUTHORIZED_EXPORT, params, requestId)
+        } else if (operationName === 'startBulkDataStoreImport') {
+          return new OperationError(operationName, ErrorCodes.UNAUTHORIZED_IMPORT, params, requestId)
+        } else if (operationName === 'startBulkDataStoreCopy') {
+          return new OperationError(operationName, ErrorCodes.UNAUTHORIZED_COPY, params, requestId)
+        }
+        return new OperationError(operationName, ErrorCodes.UNAUTHORIZED, params, requestId)
```
szluzero: I dont need this, will remove

Rusty-UX:
Tks @rzaharenkov let's go with this @szluzero 
image

rzaharenkov:
Love it!

packages/store/src/services/store/errors/errors.ts
szluzero: updated!

packages/store/src/services/store/api/api-client.ts
szluzero: will keep after all, since we can prbly get unauthorized errors from other api calls too

skelley-ton (approved):

odacrem (approved):

PR #6127, opened 2025-07-17
Add new loading mode: local

isaacroldan:

To improve the handling of unknown extension types in local development mode, allowing the app to load successfully even when unknown extension types are present.

- Adds a new `local` mode to the `AppLoaderMode` type, which ignores errors for unknown extension types
- Updates the `localAppContext` function to use the new `local` mode instead of `strict`

### How to test your changes?

For know extensions only accessible when authenticated:
1. Create a local app with an `admin_link` extensions
2. Both `app build` and `app deploy` should work. `build` will ignore the admin_link extension

For completely unknown extensions:
1. Create a local app with an `admin_link` extension.
2. Change the type of the extension to "invalid_type"
3. `app build` still works because it ignore unknown extensions, `app deploy` should crash.

### Measuring impact

### Checklist

### DIFF
--- a/packages/app/src/cli/models/app/loader.ts
+++ b/packages/app/src/cli/models/app/loader.ts
@@ -58,7 +58,13 @@ import ignore from 'ignore'
 
 const defaultExtensionDirectory = 'extensions/*'
 
-export type AppLoaderMode = 'strict' | 'report'
+/**
+ * The mode in which the app is loaded, this affects how errors are handled:
+ * - strict: If there is any kind of error, the app won't be loaded.
+ * - report: The app will be loaded as much as possible, errors will be reported afterwards.
+ * - local: Errors for unknown extensions will be ignored. Other errors will prevent the app from loading.
+ */
+export type AppLoaderMode = 'strict' | 'report' | 'local'
 
 type AbortOrReport = <T>(errorMessage: OutputMessage, fallback: T, configurationPath: string) => T
 
@@ -466,6 +472,8 @@ class AppLoader<TConfig extends AppConfiguration, TModuleSpec extends ExtensionS
 
     if (specification) {
       usedKnownSpecification = true
+    } else if (this.mode === 'local') {
+      return undefined
     } else {
       return this.abortOrReport(
         outputContent`Invalid extension type "${type}" in "${relativizePath(configurationPath)}"`,
--- a/packages/app/src/cli/services/app-context.ts
+++ b/packages/app/src/cli/services/app-context.ts
@@ -152,7 +152,6 @@ async function logMetadata(app: {apiKey: string}, organization: Organization, re
 export async function localAppContext({
   directory,
   userProvidedConfigName,
-  unsafeReportMode = false,
 }: LocalAppContextOptions): Promise<AppInterface> {
   // Load local specifications only
   const specifications = await loadLocalExtensionsSpecifications()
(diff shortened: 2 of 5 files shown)


isaacroldan:
* **#6127** <a href="https://app.graphite.dev/github/pr/Shopify/cli/6127?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Shopify/cli/6127?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/app/src/cli/models/app/loader.ts
```diff
@@ -466,6 +472,8 @@ class AppLoader<TConfig extends AppConfiguration, TModuleSpec extends ExtensionS
 
     if (specification) {
       usedKnownSpecification = true
```
MitchDickinson, line 474: It's not your change, but `usedKnownSpecification` seems pointless in this method since if there is no specification we always return early from the method, so there's no need to be checking `usedKnownSpecification` further down.

packages/app/src/cli/models/app/loader.ts
```diff
@@ -466,6 +472,8 @@ class AppLoader<TConfig extends AppConfiguration, TModuleSpec extends ExtensionS
...
+    } else if (this.mode === 'local') {
+      return undefined
```
MitchDickinson, line 476: Do we have a good sense of how the contents of the loaded app (the extensions list) are used throughout the CLI? If we knew this was only used by `build` it would be safer, but this loader could be used anywhere - how can we guarantee that all usages (current and future) will be OK with skipping an unknown extension?

packages/app/src/cli/services/app-context.ts
```diff
@@ -162,6 +161,6 @@ export async function localAppContext({
     directory,
     userProvidedConfigName,
     specifications,
-    mode: unsafeReportMode ? 'report' : 'strict',
+    mode: 'local',
```
MitchDickinson, line 164: I think since we dont' know who is using `localAppContext`, it's dangerous to assume we always want `local` here. I think we should pass an argument into `localAppContext` (similar to how it was before) so that the command can determine whether it's safe to skip unknown extensions when loading the local app. Since there might be cases where a default value of `local` is wrong and then bad things happen

packages/app/src/cli/models/app/loader.ts
isaacroldan: Actually, that's still useful when not using the `local` mode. It makes sense for the `report` mode.

packages/app/src/cli/models/app/loader.ts
isaacroldan: That is tricky, for instance the only command that uses the `report` loader mode is the `app info` command, because we want to surface errors but still load the app. 
But we can't prevent other commands from using that loading mode (unless we add a special eslint rule I guess). That's why we called it `unsafeReportMode`.

In this case, the `local` mode will only be used when using the `localAppContext`, which should only be used by unauthenticated commands (`app build` and `app function`).

packages/app/src/cli/services/app-context.ts
isaacroldan: Umm, my assumption was that if this is a local app, the expectation is that only local specifications can be used, and we should never crash in that case.

> Since there might be cases where a default value of local is wrong and then bad things happen

I can't imagine a case where you would want to load the app with the local specs, but crash if there is an unknown extension 🤔. Right now the default mode is `strict` and that's what's causing the issues that made us make this change

The way I see it:
`localAppContext` -> `local` mode
`linkedAppContext` -> `strict` or `report` modes

gonzaloriestra (approved):

PR #6147, opened 2025-07-22
Prepend application url to app proxy url if it starts with /

murray-andrew-r:

Closes https://github.com/shop/issues-develop/issues/12741#issue-3249208952

This puts us at parity with webhook subscription URLs.

Adds the application url to the start of the proxy URL if the proxy URL starts with / . 

### Measuring impact

### Checklist


isaacroldan (changes requested):
Let's investigate if this can be done server side

packages/app/src/cli/models/app/validation/common.ts
```diff
@@ -6,6 +6,10 @@ export function validateUrl(zodType: zod.ZodString, {httpsOnly = false, message
     .refine((value) => !value.includes('\n'), {message})
 }
 
+export function validateRelativeUrl(zodType: zod.ZodString, {message = 'URL must be HTTPS URL or start with /'} = {}) {
+  return zodType.refine((value) => value.startsWith('/') || isValidUrl(value, true), {message})
```
jackseok, line 10: should we worry about protocol URL's being generated here? `//`

jackseok (approved):
LGTM just left a small comment

packages/app/src/cli/models/app/validation/common.ts
murray-andrew-r: Yeah, strictly speaking the actual validation here is a bit sketchy but I'm just trying to port the existing webhooks functionality to app proxy.

Is there a risk to allowing protocol URLs aside from "The app doesn't work?"

jackseok: not that I can think of...  once we are able to get relative URLs across app config/extensions, we can place better validations then!

JonathanJChang (approved):

PR #6148, opened 2025-07-22
Handles migration of storefront_digest to essentials cookie

tsov:

Handles migration of storefront_digest to essentials cookie.

### Measuring impact

### Checklist

### DIFF
--- a/packages/theme/src/cli/utilities/theme-environment/storefront-session.ts
+++ b/packages/theme/src/cli/utilities/theme-environment/storefront-session.ts
@@ -6,6 +6,7 @@ import {outputDebug} from '@shopify/cli-kit/node/output'
 import {type AdminSession} from '@shopify/cli-kit/node/session'
 import {passwordProtected} from '@shopify/cli-kit/node/themes/api'
 import {sleep} from '@shopify/cli-kit/node/system'
+import {Response} from 'node-fetch'
 
 export class ShopifyEssentialError extends Error {}
 
@@ -41,21 +42,7 @@ export async function isStorefrontPasswordCorrect(password: string | undefined, 
     )
   }
 
-  const locationHeader = response.headers.get('location') ?? ''
-  let redirectUrl: URL
-
-  try {
-    redirectUrl = new URL(locationHeader, storeUrl)
-  } catch (error) {
-    if (error instanceof TypeError) {
-      return false
-    }
-    throw error
-  }
-
-  const storeOrigin = new URL(storeUrl).origin
-
-  return response.status === 302 && redirectUrl.origin === storeOrigin
+  return redirectsToStorefront(response, storeUrl)
 }
 
 export async function getStorefrontSessionCookies(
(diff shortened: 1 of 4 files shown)


packages/theme/src/cli/utilities/theme-environment/storefront-session.ts
```diff
@@ -6,6 +6,7 @@ import {outputDebug} from '@shopify/cli-kit/node/output'
 import {type AdminSession} from '@shopify/cli-kit/node/session'
 import {passwordProtected} from '@shopify/cli-kit/node/themes/api'
 import {sleep} from '@shopify/cli-kit/node/system'
+import {Response} from 'node-fetch'
```
karreiro: We may remove this in favor of:
```ts
import {shopifyFetch, Response} from '@shopify/cli-kit/node/http'
```

packages/theme/src/cli/utilities/theme-environment/storefront-session.ts
```diff
@@ -159,18 +148,39 @@ async function enrichSessionWithStorefrontPassword(
...
-    )
-    throw new AbortError(
-      'Your development session could not be created because the store password is invalid. Please, retry with a different password.',
-    )
+  if (storefrontDigest) {
+    return storefrontDigest
   }
 
-  return storefrontDigest
+  const newShopifyEssential = getCookie(setCookies, '_shopify_essential')
```
karreiro: When testing, `newShopifyEssential` is consistently coming up as `undefined` for me. How can I test this branch in the code? I'm trying to validate when we get a new `_shopify_essential` to understand if we need to update this value in the `session` of the command (as this is used in the rendering that powers hot-reloading)

iaineh: +1, we will need to change this. We need to update that cookie value if it has been changed.

tsov: Sorry for the delay! It should work now, since we rolled out the `storefront_digest` changes!

tsov:
@karreiro Pushed up a new commit. It makes sure we keep the newly fetched `_shopify_essential` cookie in the local/in-memory "cookie jar"

iaineh (approved):

karreiro:
@tsov Oops.. could you please include a patch changelog with `p changeset add`? I missed to mention that 🙏

karreiro (commented):
👋 @tsov, actually, I've noticed two details:

- 1. We need to include a patch changelog with `p changeset add`
- 2. We need to handle Theme Access app scenarios, I'm getting a `Your development session could not be created because the store password is invalid. Please, retry with a different password.`

Please, let know if you want to 🍐 on these details, and thanks again for this PR 🙇

karreiro (commented):
👋 Hey @tsov,

I just tried the updated version, and I've got the same message:

image

tsov:
@karreiro Let's pair on this tomorrow if you are free! Would love to patch this up 🙂

tsov:
@karreiro this should be ready now. The issue was the origin check in `redirectsToStorefront`, when using theme access kit it uses a different storeUrl, which is used to proxy to SFR. The last commit will make sure that we use the `storeFqdn` instead of the `storeUrl` in the origin check

karreiro (approved):
Thanks a lot for the updates, @tsov! LGTM and works as expected 🎩 :)

Issue #6107, opened 2025-07-14
[Bug]: `shopify theme dev` has the wrong content type for `/cart.json` requests


karreiro:
The `shopify theme dev` has the wrong content type for `/cart.json` requests:

<img width="2097" height="1572" alt="Image" src="https://github.com/user-attachments/assets/31b3b93f-5624-46a2-a700-46070c1ac79b" />

Notice that the development server returns the expected one on curl tho:

<img width="934" height="290" alt="Image" src="https://github.com/user-attachments/assets/1682d144-a92c-4143-872e-a9bd1a58c8a2" />

ConduciveMammal:
It's worth noting that I tested other Shopify JSON endpoints that also return this issue:
* `collections.json`
* `collections/handle.json`
* `collections/handle/count.json`
* `products/handle.json`

These are a few but at this point, I imagine they all return this way.

gonzaloriestra:
This should be fixed in version 3.84.0.

PR #6156, opened 2025-07-24
Fix .json files returning as text/html

EvilGenius13:

Fixes https://github.com/Shopify/cli/issues/6107

`json` files were being returned as the wrong content type in `theme dev`. Instead of `application/json` they were returning as `text/html`.

This was happening in two parts:

- we would delete the content type from any response
- We would patch responses by setting the content type to `text/html`

- No longer deleting the content type if it's `application/json`
- Not force setting the content type to `text/html` if it's `application/json`

### How to test your changes?

I have added tests in the PR.

To test the changes locally:

- Pull down the branch
- Build the branch
- Run `theme dev`
- Go to a direct route with a json file like `http://127.0.0.1:9292/cart.json`
- Inspect the page, go to the Network tab and click on the response
- Check that the content type is `application/json`
- Also check a regular page to make sure it's coming back as `text/html`

### Measuring impact

### Checklist


graygilmore (approved):
Didn't 🎩 but these changes make sense to me

PR #6138, opened 2025-07-21
Show a nice error message if user is running node <20

jordanverasamy:
### WHY are these changes introduced?

https://github.com/Shopify/workflows-operations/issues/3367

### WHAT is this pull request doing?

Show a nice error if user is running node version <20.

### How to test your changes?

```
nvm install 18
nvm use 18
pnpm shopify store copy [...]
```

PR #6158, opened 2025-07-24
Introduce pretty progress displays

JordanVerasamy:

Fixes #0000 

image

### Measuring impact

### Checklist


packages/cli-kit/src/public/node/progress-indicators.test.ts
```diff
@@ -0,0 +1,69 @@
...
+import {describe, test, expect, vi} from 'vitest'
+import {twoThirdsOfTerminalWidth} from './terminal-layout.js'
+
+vi.mock('./terminal-layout.js', () => ({
+  twoThirdsOfTerminalWidth: vi.fn(() => 80),
+}))
+
+describe('progress-indicators utilities', () => {
+  describe('createColoredProgressBar', () => {
+    const bar = (length: number) => '█'.repeat(length)
```
rzaharenkov: What is this character █ here? Would it better to use unicode character for it or at least have a comment explaining?

packages/cli-kit/src/public/node/progress-indicators.test.ts
```diff
@@ -0,0 +1,69 @@
...
+
+    test('should show all gray blocks at 0% progress', () => {
```
rzaharenkov:
```suggestion
    test('shows all gray blocks at 0% progress', () => {
```

I think we shifted away from start every test description from "should" and it seems to be aligned with other tests in the codebase.

packages/cli-kit/src/public/node/progress-indicators.test.ts
```diff
@@ -0,0 +1,69 @@
...
+    })
+  })
+
+  describe('createAnimatedDots', () => {
+    test('cycles through dot counts based on counter', () => {
+      expect(createAnimatedDots(0)).toBe('')
+      expect(createAnimatedDots(2)).toBe('.')
+      expect(createAnimatedDots(4)).toBe('..')
+      expect(createAnimatedDots(6)).toBe('...')
+      expect(createAnimatedDots(8)).toBe('')
```
rzaharenkov, line 74: I think this numbers are a bit unintuitive. I would expect it to cycle through the numbers and would make consumer responsible to increment the counter slower or faster depending on the context. 

I would naturally expect this:

```js
      expect(createAnimatedDots(0)).toBe('')
      expect(createAnimatedDots(1)).toBe('.')
      expect(createAnimatedDots(2)).toBe('..')
      expect(createAnimatedDots(3)).toBe('...')
      expect(createAnimatedDots(4)).toBe('')
      expect(createAnimatedDots(5)).toBe('.')
      // ... and so on
```

packages/cli-kit/src/public/node/progress-indicators.test.ts
```diff
@@ -0,0 +1,69 @@
...
+
+    test('should display a bar with a color gradient', async () => {
+      // mock smaller width for easier testing
+      vi.mocked(twoThirdsOfTerminalWidth).mockReturnValue(20)
+
+      // freezing time makes the gradient static
+      vi.useFakeTimers({now: new Date('2025-01-01T00:00:00.000Z')})
+
+      const gradientGreenValues = [
+        69, 77, 86, 94, 103, 111, 120, 128, 137, 145, 153, 162, 170, 179, 187, 196, 204, 213, 221, 230,
```
rzaharenkov: Since the `gradient` is a separate function I would test it here. I'd rather call `gradient` here and compare the result (I think it aligns pretty well with calling `dim` above in the previous example).

packages/cli-kit/src/public/node/progress-indicators.ts
```diff
@@ -0,0 +1,68 @@
...
+  const filledChar = '█'
+  const bar = filledChar.repeat(width)
+
+  const fastFrame = Date.now() / 100
+  const hue = (fastFrame * ANIMATION_HUE_SPEED) % 360
+  const leftColor = `hsv(${hue}, 70%, 90%)`
+  const rightColor = `hsv(${(hue + GRADIENT_HUE_SPREAD) % 360}, 70%, 90%)`
+
+  return gradient(leftColor, rightColor)(bar, {interpolation: 'hsv', hsvSpin: 'short'})
+}
```
rzaharenkov, line 48: Does it have to be a progress bar if we can't determine the progress? I though the entire point of progress bar is to show the progress toward the target. I would expect to just see a spinner or dots for such operations where the progress could not be calculated.

packages/cli-kit/src/public/node/terminal-layout.ts
```diff
@@ -0,0 +1,64 @@
...
+}
+
+/**
+ * Creates right-aligned text by padding between left and right text.
+ *
+ * @param leftText - The text to display on the left.
+ * @param rightText - The text to display on the right.
+ * @returns The formatted string with proper alignment.
+ */
+export function createRightAlignedText(leftText: string, rightText: string): string {
```
rzaharenkov: I got a bit confused by this function name. Do you mean right is sense "proper" and not in the sense of direction, right? ;)

I don't know maybe calling it something like `spreadText` would make it more straightforward.

packages/cli-kit/src/public/node/progress-indicators.ts
```diff
@@ -0,0 +1,68 @@
...
+const MAX_DOTS = 3
+
+/**
+ * Creates a colored progress bar showing completion percentage.
+ *
+ * @param percentage - The completion percentage (0-100).
+ * @returns The colored progress bar string.
+ */
+export function createColoredProgressBar(percentage: number): string {
+  const width = twoThirdsOfTerminalWidth()
```
rzaharenkov: I'd rather pass down `width` as function argument instead so progress bar could be more customizable.

packages/cli-kit/src/public/node/terminal-layout.ts
```diff
@@ -0,0 +1,64 @@
...
+  const twoThirdsWidth = twoThirdsOfTerminalWidth()
```
rzaharenkov: Same here: if you pass the expected width as the argument this function could be used in more use cases.

packages/store/src/services/store/utils/bulk-operation-progress.ts
```diff
@@ -0,0 +1,144 @@
...
+      currentOperation.organization.bulkData.operation.status === 'COMPLETED' ||
+      currentOperation.organization.bulkData.operation.status === 'FAILED'
+    )
+      return
+
+    const progressContent = renderProgress(currentOperation, dotCounter)
+    if (progressContent) {
+      updateProgress(progressContent, true)
+    }
+    dotCounter++
```
rzaharenkov: nit, I wouldn't call it a `dotCounter` since it doesn't represent the actual count of dots. IMO something like `iteration` would be more clear. 

Not for this PR but I think we might get a better abstraction with separate class that keeps track of the counter and manages animation timer.

rzaharenkov (commented):
I like solution and I appreciate your contribution to cli-kit. I have only few nit-picky comments around signatures and testing.

jordanverasamy:
* **#6192** <a href="https://app.graphite.dev/github/pr/Shopify/cli/6192?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6158** <a href="https://app.graphite.dev/github/pr/Shopify/cli/6158?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Shopify/cli/6158?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/cli-kit/src/public/node/progress-indicators.ts
jordanverasamy: Shrug, I'm not sure. Rusty wants a rainbow gradient when we don't know how far along we are, and that's what's in the Figma, so that's what I implemented.

packages/cli-kit/src/public/node/terminal-layout.ts
jordanverasamy: "Right-aligned text" is the official name of this concept, I think naming it anything else would be more confusing -- "spread text" could mean a lot of things, but "right-aligned" means precisely lining it up to the right edge, which is what we're doing. So I feel like this is the right name, but definitely open if anyone has other suggestions!

packages/store/src/services/store/utils/bulk-operation-progress.ts
jordanverasamy: Good idea, will rename :)

packages/cli-kit/src/public/node/progress-indicators.ts
jordanverasamy: good idea!

packages/cli-kit/src/public/node/terminal-layout.ts
jordanverasamy: good idea!

packages/cli-kit/src/public/node/progress-indicators.ts
jordanverasamy: Done, thanks :)

packages/cli-kit/src/public/node/terminal-layout.ts
jordanverasamy: Done, thanks :)

packages/store/src/services/store/utils/bulk-operation-progress.ts
jordanverasamy: Done, thanks :)

packages/cli-kit/src/public/node/progress-indicators.test.ts
jordanverasamy: Great question -- I extracted it to a const and gave it a nice name and comment! TY!

packages/cli-kit/src/public/node/progress-indicators.test.ts
jordanverasamy: You're totally right, good catch -- fixed!

packages/cli-kit/src/public/node/progress-indicators.test.ts
jordanverasamy: Great point -- I adjusted the animation logic based on this feedback. Now the dots and the rainbow gradient both use the same animation counter, and I just use a constant slowdown factor to make the dots animate slower:

```typescript
const DOT_ANIMATION_SLOWDOWN_FACTOR = 16
// ...
const dots = createAnimatedDots(Math.floor(animationIteration / DOT_ANIMATION_SLOWDOWN_FACTOR))
```

This actually makes it look smoother, while also giving us more control over the animation timing, _and_ making the cli-kit components more reusable and intuitive, which is sweet!

packages/cli-kit/src/public/node/progress-indicators.test.ts
jordanverasamy: Thanks for getting me to take another look at this! I ended up doing something inspired by this comment, but not exactly your suggestion -- instead of calling it in the test and comparing the result, I mocked it with a simpler version and then checked the result directly. Take a look and lmk if you think that looks ok :)

packages/cli-kit/src/public/node/terminal-layout.ts
```diff
@@ -0,0 +1,68 @@
...
+ */
+function getVisualLength(text: string): number {
+  // eslint-disable-next-line no-control-regex
+  return text.replace(/\x1b\[[0-9;]*m/g, '').length
+}
+
+/**
+ * Creates right-aligned text by padding between left and right text.
+ *
+ * @param leftText - The text to display on the left.
```
szluzero, line 21: for my own clarification, is this the `Importing...` text from your image in the pr description?

jordanverasamy: Yeah, we're using this to put `Importing...` on the left and `35 / 2450` on the right, or whatever.

packages/cli-kit/src/public/node/terminal-layout.ts
jordanverasamy: I guess this function does also print left-aligned text too... let me think harder about the name for this.

packages/store/src/services/store/operations/store-import.ts
```diff
@@ -39,47 +35,55 @@ export class StoreImportOperation implements StoreOperation {
...
 
-    const targetShopDomain = await normalizeStoreFqdn(toStore)
-    const apiShopId = await this.validateShop(targetShopDomain)
+    const {domain: targetShopDomain, id: apiShopId} = await this.normalizeAndValidateShop(toStore)
 
     if (!flags['no-prompt']) {
       if (!(await confirmImportPrompt(fromFile, targetShopDomain))) {
         outputInfo('Exiting.')
-        process.exit(0)
+        return
```
szluzero, line 43: just wondering, does this change make a difference?

packages/cli-kit/src/public/node/terminal-layout.ts
szluzero: so the truncation is in case someone else using this wants to have a super long message instead of `Importing...`?

packages/store/src/services/store/operations/store-import.ts
jordanverasamy: Nah, it's just that I got feedback in another PR that maybe we should be avoiding `process.exit`. ([comment](https://github.com/Shopify/cli/pull/6138#discussion_r2221902922)), and this behaves the same way but follows what I think the pattern is better.

Could also leave it as is if we want.

szluzero (approved):
looks good. since there are changes to cli-kit I guess someone in charge of that needs to review as well?

packages/cli-kit/src/public/node/terminal-layout.ts
jordanverasamy: Yeah!

gonzaloriestra (commented):
This is so cool, thanks! I haven't checked the code yet, but I have some initial feedback, mostly related to UX (cc: @jenstanicak @Rusty-UX).

Using a different loading bar for store commands is breaking consistency with other commands that use the rainbow bar. We are even using both in the same command:

image

I think we should only have one. We could use the new one everywhere. If I'm not wrong, you only need to update the Task component.

More things to consider:
- It should work on Windows and Linux
- The movement should be appreciated with no colors (see https://github.com/Shopify/cli/pull/5262)
- It should be hidden on non-TTY environments

packages/cli-kit/src/public/node/progress-indicators.ts
Rusty-UX: @jordanverasamy @rzaharenkov On UI — IMO this is a stop gap until we get proper progress for export. Technically we could have it show a spinner or dots, but I felt that having it display with the same overall shape makes the transition from export → import smoother, and is a step towards where we want to go UI wise.

Rusty-UX:
@gonzaloriestra From a UI standpoint, the appearance of the loading bar rectangle should be the same. 

The existing loader/progress component however is not sufficient for the needs of BDM in terms of experience. 

@jordanverasamy and co you folks are best equipped to determine best technical approach.

packages/cli-kit/src/public/node/progress-indicators.ts
rzaharenkov: I am OK to use that UI as a temporary solution. However I do believe it's not very often used type of UI and I don't think it's valuable to have it in shared library. But that's my personal opinion, not blocking on this.

packages/cli-kit/src/public/node/progress-indicators.ts
```diff
@@ -0,0 +1,59 @@
...
+
+/**
+ * Creates an animated dots string for loading indicators.
+ *
+ * @param animationIteration - The animation counter that gets incremented over time.
+ * @returns A string with 0-3 dots for animation.
+ */
+export function createAnimatedDots(animationIteration: number): string {
+  const dotCount = Math.floor(animationIteration) % (MAX_DOTS + 1)
+  return '.'.repeat(dotCount)
```
rzaharenkov, line 58: Thank you for adjusting this, I love the signature and simplicity of this function!

packages/store/src/services/store/operations/store-copy.ts
```diff
@@ -1,76 +1,82 @@
...
-  private async validateShops(sourceShopDomain: string, targetShopDomain: string): Promise<string> {
-    const sourceShop = await this.apiClient.getStoreDetails(sourceShopDomain)
-    const targetShop = await this.apiClient.getStoreDetails(targetShopDomain)
+  protected getFailureErrorCode() {
+    return ErrorCodes.COPY_FAILED
+  }
 
-    if (sourceShop.id === targetShop.id) {
-      throw new ValidationError(ErrorCodes.SAME_SHOP)
+  private readonly renderProgress = (operation: BulkDataOperationByIdResponse, dotCount: number): string => {
```
rzaharenkov, line 57:
```suggestion
  private readonly renderProgress = (operation: BulkDataOperationByIdResponse, iteration: number): string => {
```

Since function `createAnimatedDots` is now accepting `animationIteration` argument I think it might be cleaner go away from `dotCount` here and in other places.

packages/store/src/services/store/operations/store-copy.ts
```diff
@@ -1,76 +1,82 @@
...
+    const storeOps = operation.organization.bulkData.operation.storeOperations
+    if (!storeOps || storeOps.length === 0) {
+      return ''
+    }
+
+    const [exportOp, importOp] = storeOps
+    const isExportComplete = exportOp && isOperationComplete(exportOp)
```
rzaharenkov, line 64: One caveat here: I think we don't guarantee a certain order for store operations list. Although the will be always only one operation that is in progress. So maybe we could have adjusted this logic to pick an active store operation from the list and use it's type to select appropriate progress display function?

isaacroldan:
> The existing loader/progress component however is not sufficient for the needs of BDM in terms of experience.

I understand the current bar doesn't work for import, but it does por the export case no?

PR #6006, opened 2025-06-23
Add warning for non multi env commands

EvilGenius13:
### WHY are these changes introduced?
As we continue converting theme commands to use the new ThemeCommand class, we need a way to indicate which commands don't support multiple environments.

### WHAT is this pull request doing?
Allow multiEnvironmentsFlags to accept null as a value. When a command sets this to null, it will render a warning if users attempt to run it with multiple environments.

### How to test your changes?
- Ensure you have `shopify.theme.toml` file with at least two stores
- Pull down the branch
- Go into `commands/theme/list.ts` and replace `static multiEnvironmentsFlags = ['store', 'password']` with `  static multiEnvironmentsFlags = null`
- Build the branch
- Run `theme list -e first_store -e second_store`, to ensure it returns the warning
- Run `theme list -e first_store` or `theme list` to ensure the commands still work in single environments

### Post-release steps
N/A

### Measuring impact

How do we know this change was effective? Please choose one:

- [ ] n/a - this doesn't need measurement, e.g. a linting rule or a bug-fix
- [X] Existing analytics will cater for this addition
- [ ] PR includes analytics changes to measure impact

### Checklist

- [X] I've considered possible cross-platform impacts (Mac, Linux, Windows)
- [X] I've considered possible [documentation](https://shopify.dev) changes

PR #6161, opened 2025-07-24
Allow theme rename command to be called with multiple environments

dejmedus:

Previously `theme rename` command could only be run with a single `--environment` flag. This PR allows `rename` to be used in multiple environments in a single command run. 

Since multi environment commands must run without interactions we need to define which flags are required to run the command. In some cases a command will provide a few different flags that all meet the same requirement, so this PR also updates validation logic to accept "one of" required flags. 

For example: `rename` requires `store`, `password`, `name`, and one of: `live`, `theme`, or `development` (since `rename` can use any of the three to understand what to rename)

- Updates environment validation logic to understand "one of" flag requirements
- Removes flag parsing and auth from `Rename` in favour of `ThemeCommand`'s `run` logic
- Prefixes output with the environment name, when present

https://github.com/user-attachments/assets/f9cbbe15-c474-45e2-85de-954d0cf793bd

### How to test your changes?

- Pull down the branch `gh pr checkout 6161` or install the snapit version
```sh
pnpm i -g @shopify/cli@0.0.0-snapshot-20250724234104 --@shopify:registry=https://registry.npmjs.org
```

```
[environments.store1]
theme = "<theme>"
store = "<store>"

[environments.store2]
theme = "<theme>"
store = "<store>"
```

</details> 

```ts
shopify theme rename -e store1 -e store2 -n "Summer Edition" --live
```
- Rename should accept flags from both CLI and `shopify.theme.toml`
- Rename run with a single environment or no environment should work as normal
- Since rename does not allow `--force` flag, no confirmation prompt should be shown. If you manually added `force` to flags to test confirmations, missing a "one of" flag would look like: 

<details><summary>Confirmation prompt</summary>

confirmation-prompt

</details> 

- Environments missing flags should display a warning and it's theme shouldn't be renamed
- Environments with incorrect flags (ex. password = "fake") should have the environment name prefixed to the error message

> If provided both `live` and `theme`, `findOrSelectTheme` would choose role > theme (same as current single or no env runs)

### Measuring impact

### Checklist

### DIFF
--- a/packages/theme/src/cli/commands/theme/rename.ts
+++ b/packages/theme/src/cli/commands/theme/rename.ts
@@ -1,11 +1,9 @@
-import ThemeCommand from '../../utilities/theme-command.js'
-import {RenameOptions, renameTheme} from '../../services/rename.js'
-import {ensureThemeStore} from '../../utilities/theme-store.js'
+import ThemeCommand, {RequiredFlags} from '../../utilities/theme-command.js'
 import {themeFlags} from '../../flags.js'
+import {RenameOptions, renameTheme} from '../../services/rename.js'
 import {Flags} from '@oclif/core'
 import {globalFlags} from '@shopify/cli-kit/node/cli'
-import {ensureAuthenticatedThemes} from '@shopify/cli-kit/node/session'
-import {promptThemeName} from '@shopify/cli-kit/node/themes/utils'
+import {AdminSession} from '@shopify/cli-kit/node/session'
 
 export default class Rename extends ThemeCommand {
   static summary = 'Renames an existing theme.'
@@ -43,21 +41,9 @@ export default class Rename extends ThemeCommand {
     }),
   }
 
-  public async run(): Promise<void> {
-    const {flags} = await this.parse(Rename)
-    const {password, development, name, theme, live} = flags
+  static multiEnvironmentsFlags: RequiredFlags = ['store', 'password', 'name', ['live', 'development', 'theme']]
 
-    const store = ensureThemeStore(flags)
-    const adminSession = await ensureAuthenticatedThemes(store, password)
-    const newName = name || (await promptThemeName('New name for the theme'))
-
-    const renameOptions: RenameOptions = {
-      newName,
-      development,
-      theme,
-      live,
-    }
-
-    await renameTheme(adminSession, renameOptions)
+  async command(flags: RenameOptions, adminSession: AdminSession) {
+    await renameTheme(flags, adminSession)
   }
 }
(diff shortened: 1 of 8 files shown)


packages/theme/src/cli/commands/theme/rename.ts
```diff
@@ -43,21 +41,9 @@ export default class Rename extends ThemeCommand {
...
-    const adminSession = await ensureAuthenticatedThemes(store, password)
-    const newName = name || (await promptThemeName('New name for the theme'))
-
-    const renameOptions: RenameOptions = {
-      newName,
-      development,
-      theme,
-      live,
-    }
+  static multiEnvironmentsFlags: RequiredFlags = ['store', 'password', 'name', ['theme', 'live', 'development']]
```
dejmedus: The nested array looks a bit wild, but this seemed to be the easiest/cleanest way to achieve "one of" requirement groups. 

I was going to comment above `static multiEnvironmentsFlags` but I thought adding a docstring to its type could make sense for reusability, it looks like:

on-of-flags-docstring

What do we think?

packages/theme/src/cli/services/rename.ts
```diff
@@ -3,15 +3,19 @@ import {themeComponent} from '../utilities/theme-ui.js'
...
+  name?: string
   development?: boolean
   theme?: string
   live?: boolean
+  environment?: string
 }
 
-export async function renameTheme(adminSession: AdminSession, options: RenameOptions) {
+export async function renameTheme(options: RenameOptions, adminSession: AdminSession) {
+  const newName = options.name || (await promptThemeName('New name for the theme'))
```
dejmedus, line 17: ESLint is warning to change to `??` here (does pass linting though). I think it still makes sense to keep it as `||` in case of empty strings

packages/theme/src/cli/commands/theme/rename.ts
```diff
@@ -43,21 +41,9 @@ export default class Rename extends ThemeCommand {
     }),
```
dejmedus, line 41: I considered adding the `force` flag to display the multi env confirmation prompt before renaming. I held off since it might feel weird that non-multi env runs act differently (no confirmation).

graygilmore (approved):
🎩 went great!

EvilGenius13 (approved):
Awesome PR @dejmedus! 🎩 went well. This will be great for the other commands we need to convert.

packages/theme/src/cli/utilities/theme-command.ts
```diff
@@ -128,7 +143,10 @@ export default abstract class ThemeCommand extends Command {
    * @param requiredFlags - The required flags to check for
    * @returns An object containing valid and invalid environment arrays
    */
-  private async validateEnvironments(environmentMap: Map<EnvironmentName, FlagValues>, requiredFlags: string[]) {
+  private async validateEnvironments(
+    environmentMap: Map<EnvironmentName, FlagValues>,
+    requiredFlags: Exclude<RequiredFlags, null>,
```
dejmedus, line 148: After bringing in the [Add warning for non multi env commands](https://github.com/Shopify/cli/pull/6006/files) changes TS was upset because it thought requiredFlags could potentially be null, even though it couldn't be at this point

Some options were:

1. Add a non-null, but would need to ignore the ESLint error
2. Create an additional type w/ just `(string | string[])[]`, but lose the context of RequiredFlags docstring
3. I landed on: narrow the type to exclude null where it is not possible for it to exist

PR #6165, opened 2025-07-25
Run pnpm install to set up workspace

adamtroiani:

Generating an extension via pnpm is failing as we are missing `graphql-codegen-*` dependencies at the time of running the command.

Fixes shop/issues-shopifyvm#434

Adds `@graphql-codegen/cli` to dependencies to ensure we've installed these when we run the graphql code generator.

### Measuring impact

### DIFF
--- a/packages/app/src/cli/services/function/build.test.ts
+++ b/packages/app/src/cli/services/function/build.test.ts
@@ -43,7 +43,7 @@ const derivedDeps = {
   javyPlugin: PREFERRED_JAVY_PLUGIN_VERSION,
 }
 
-const app = testApp({dotenv: {variables: {VAR_FROM_ENV_FILE: 'env_file_var'}, path: ''}})
+const app = testApp({dotenv: {variables: {VAR_FROM_ENV_FILE: 'env_file_var'}, path: ''}, packageManager: 'npm'})
 
 beforeEach(async () => {
   stderr = {write: vi.fn()}
--- a/packages/app/src/cli/services/function/build.ts
+++ b/packages/app/src/cli/services/function/build.ts
@@ -129,7 +129,22 @@ export async function buildGraphqlTypes(
   }
 
   return runWithTimer('cmd_all_timing_network_ms')(async () => {
-    return exec('npm', ['exec', '--', 'graphql-code-generator', '--config', 'package.json'], {
+    const packageManager = options.app.packageManager
+    let execArgs: string[]
+
+    switch (packageManager) {
+      case 'pnpm':
+      case 'yarn':
+      case 'bun':
+        execArgs = ['exec', 'graphql-code-generator', '--config', 'package.json']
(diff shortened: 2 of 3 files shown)


nickwesselman:
Does this solve the issue for you? When I tried this it didn't seem to.

adamtroiani:
@nickwesselman it seems to work for me. Try these steps and let me know if it works for you:
1. `shopify app init`
2. cd to the shopify cli and checkout this branch 
3. `pnpm shopify app generate extension --path <path to shopify app created in step 1>`

nickwesselman:
With a Remix app, generating a Discount function, I still get the error:

```
── external error ─────────────────────────────────────────────────────────────────

Error coming from `npm exec -- graphql-code-generator --config package.json`

Command failed with exit code 1: npm exec -- graphql-code-generator --config package.json
Cannot convert undefined or null to object
[13:22:56] Parse configuration [started]
[13:22:56] Parse configuration [failed]
[13:22:56] → Cannot convert undefined or null to object

───────────────────────────────────────────────────────────────────────────────────
```

Extension-only app works fine. Partners vs Dev Dash doesn't seem to make a difference.

Environment:
CLI: `@shopify/cli/0.0.0-snapshot-20250730152204 darwin-arm64 node-v23.7.0`
pnpm: `10.13.1`

nickwesselman (changes requested):
GraphQL typegen during extension generation still fails for me w/ a Remix app.

nickwesselman:
The latest works for me on the Remix template! But it does add dependencies to the function package.json that weren't there before:

```
  "dependencies": {
    "@graphql-codegen/cli": "~5.0.0",
    "@graphql-codegen/typescript-operations": "~4.0.0",
    "@shopify/shopify_function": "2.0.0"
  },
```

This puts the versioning of `graphql-codegen` out of our control and spreads it across each extension, right?

I'll reiterate my [thought here](https://github.com/shop/issues-shopifyvm/issues/434#issuecomment-3069987342) and ask whether we can make the invocation of graphql-code-generator less fragile in general... should it just be a dependency on the CLI itself so we don't need to shell out to it?

cc @andrewhassan @otheriault

andrewhassan:
I think it depends on whether the `graphql-codegen` version should be tied to the function or the CLI. In other words, if there is a breaking change to `graphql-codegen`, tying it to the CLI would mean it fails after a CLI upgrade, and tying it to the function would mean it doesn't fail, but they do not get the latest version automatically.

andrewhassan:
Thinking about it a bit more, `graphql-codegen` is really a dependency of the CLI because it directly invokes this tool; the function itself never does. Therefore, I do agree that it should be tied to the CLI version, and not the function.

adamtroiani:
Closing in favour of this approach: https://github.com/Shopify/cli/pull/6242

PR #6196, opened 2025-07-31
adjust `exitIfOldNodeVersion` to look at engine-declared node version instead of hardcoding

jordanverasamy:

Fixes https://github.com/Shopify/workflows-operations/issues/3367.

Replaces https://github.com/Shopify/cli/pull/6138.

Instead of hardcoding the minimum node version as 18, let's check the package.json and require that version.

### How to test your changes?

Try adjusting your node version with `nvm` and watch what happens:

image

### DIFF
--- a/.changeset/shy-frogs-float.md
+++ b/.changeset/shy-frogs-float.md
@@ -1,0 +1,5 @@
+---
+'@shopify/cli-kit': major
+---
+
+cli-kit now exits with an error if the current Node version is lower than what the package.json engines require.
--- a/packages/cli-kit/src/public/node/cli.test.ts
+++ b/packages/cli-kit/src/public/node/cli.test.ts
@@ -2,63 +2,67 @@ import {clearCache, runCLI, runCreateCLI} from './cli.js'
 import {findUpAndReadPackageJson} from './node-package-manager.js'
 import {mockAndCaptureOutput} from './testing/output.js'
 import * as confStore from '../../private/node/conf-store.js'
-import {describe, expect, test, vi} from 'vitest'
+import {beforeEach, describe, expect, test, vi} from 'vitest'
 
 vi.mock('./node-package-manager.js')
 
 describe('cli', () => {
+  beforeEach(() => {
+    vi.mocked(findUpAndReadPackageJson).mockResolvedValue({content: {engines: {node: '>=18.0.0'}}} as any)
+  })
+
   test('prepares to run the CLI', async () => {
     const launchCLI = vi.fn()
-    await runCLI({moduleURL: 'test', development: false}, launchCLI)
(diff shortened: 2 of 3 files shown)


.changeset/shy-frogs-float.md
```diff
@@ -0,0 +1,5 @@
+---
+'@shopify/cli-kit': major
```
gonzaloriestra, line 2: We don't do major changes for now

```suggestion
'@shopify/cli-kit': minor
```

packages/cli-kit/src/public/node/cli.test.ts
```diff
@@ -75,6 +79,20 @@ describe('cli', () => {
     `)
   })
 
+  test('exits if running node version older than package.json engines field', async () => {
+    const launchCLI = vi.fn()
+    vi.mocked(findUpAndReadPackageJson).mockResolvedValue({content: {engines: {node: '>=22.0.0'}}} as any)
+    const run = runCLI({moduleURL: 'file:///test', development: false}, launchCLI, [], {}, {node: '21.9'} as any)
+    await expect(run).rejects.toThrow()
```
gonzaloriestra, line 86: Let's better ensure it's failing because of the right cause

```suggestion
    await expect(run).rejects.toThrow(/Node version/)
```

packages/cli-kit/src/public/node/cli.ts
```diff
@@ -15,11 +15,22 @@ interface RunCLIOptions {
...
   const nodeVersion = versions.node
   const nodeMajorVersion = Number(nodeVersion.split('.')[0])
 
-  const currentSupportedNodeVersion = 18
+  const {findUpAndReadPackageJson} = await import('./node-package-manager.js')
+  const {moduleDirectory} = await import('./path.js')
+
+  const packageJson = await findUpAndReadPackageJson(moduleDirectory(moduleURL))
+  const engines = (packageJson.content as {engines?: {node?: string}}).engines
+  const nodeRequirement = engines?.node ?? `>=${DEFAULT_MIN_NODE_VERSION}.0.0`
```
gonzaloriestra, line 29: `engines.node` should always exist in cli-kit's package.json, so I'd prefer to just log an error and return when undefined. 

Otherwise, if for some reason this is not working as expected, we would be checking against a different minimum node version here that it's outdated.

gonzaloriestra (changes requested):
Thanks for taking care of this! It's looking good, but the changeset can't be major.

packages/cli-kit/src/public/node/cli.ts
isaacroldan: Agree with Gonzalo, but also remember that cli-kit is bundled, we should test that this works on a snapshot too.
The package.json that this code will read in the released CLI will be the one under `packages/cli`

PR #6199, opened 2025-07-31
Add yarn berry (v2+) support to instalNodeModules

juanpprieto:

 The `installNodeModules` function in cli-kit hardcoded the ['install'] command for all package managers, including yarn. This caused build failures when using Yarn Berry (v2/v3+) because:

  - Yarn v1 (Classic): Uses yarn install command
  - Yarn v2/v3+ (Berry): Uses yarn add command for dependencies

  This incompatibility blocked the Hydrogen CLI upgrade command and other cli-kit consumers from working with modern yarn versions, forcing users to stay on legacy yarn v1 or face broken installations.

Enhanced installNodeModules with yarn version-aware command selection:

#### New Function: getYarnInstallCommand(directory: string)

  - Multi-tier version detection strategy with intelligent fallbacks
  - Comprehensive error handling for system availability issues

#### Enhanced installNodeModules Function
  - Zero breaking changes - npm/pnpm/bun behavior unchanged
  - Yarn-specific logic - detects version and uses appropriate command
  - Maintains all existing interfaces and stream options

#### Detection Strategy (Priority Order)
  1. packageManager field in package.json (modern standard)
  2. yarn --version command execution and parsing
  3. .yarnrc.yml file presence (Berry project indicator)
  4. Fallback to install (maximum backward compatibility)

### Measuring impact

### Checklist

### DIFF
--- a/packages/cli-kit/src/public/node/node-package-manager.ts
+++ b/packages/cli-kit/src/public/node/node-package-manager.ts
@@ -190,6 +190,71 @@ export async function installNPMDependenciesRecursively(
   }
 }
 
+const yarnVersionCache = new Map<string, string[]>()
+
+export async function getYarnInstallCommand(directory: string): Promise<string[]> {
+  if (yarnVersionCache.has(directory)) {
+    return yarnVersionCache.get(directory)!
+  }
+
+  let result: string[] = ['install']
+
+  try {
+    const packageJsonPath = joinPath(directory, 'package.json')
+    if (await fileExists(packageJsonPath)) {
+      try {
+        const packageJsonContent = await readAndParsePackageJson(packageJsonPath)
+        if (packageJsonContent.packageManager && typeof packageJsonContent.packageManager === 'string') {
+          const packageManagerField = packageJsonContent.packageManager
+          if (packageManagerField.startsWith('yarn@')) {
+            const versionMatch = packageManagerField.match(/yarn@(\d+)\./)
+            if (versionMatch) {
+              const majorVersion = parseInt(versionMatch[1], 10)
+              if (majorVersion >= 2) {
+                result = ['add']
+                yarnVersionCache.set(directory, result)
+                return result
+              } else {
+                result = ['install']
+                yarnVersionCache.set(directory, result)
+                return result
+              }
+            }
+          }
+        }
+      } catch (error) {
+      }
+    }
+
+    try {
+      const versionOutput = await captureOutput('yarn', ['--version'], {cwd: directory})
+      const version = versionOutput.trim()
+      const versionMatch = version.match(/^(\d+)\./)
+      if (versionMatch) {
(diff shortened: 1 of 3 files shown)


packages/cli-kit/src/public/node/node-package-manager.ts
```diff
@@ -207,10 +324,18 @@ export async function installNodeModules(options: InstallNodeModulesOptions): Pr
...
+  } else {
+    args = ['install']
+  }
+
   if (options.args) {
     args = args.concat(options.args)
   }
+
   await runWithTimer('cmd_all_timing_network_ms')(async () => {
     await exec(options.packageManager, args, execOptions)
```
kdaviduik, line 340: I don't think this approach works, because it looks like it would break cases where `installNodeModules` is called with just flags but no dependencies to add (eg: `yarn install --force`). From my understanding:
- both yarn v1 and yarn berry use `yarn install` to just install dependencies (ie: not add, just install)
- To ADD dependencies:
    - yarn v1 supports `yarn install <package>` OR `yarn add <package>`
    - yarn berry ONLY supports `yarn add <package>`

So this seems like a much cleaner and simpler solution:
```
  let args: string[]
  // Are all of the args just flags (like `--force`)? Or are any of them actual packages to ADD?
  const hasPackageArgs = args?.some(arg => !arg.startsWith('-'))

  if (!hasPackageArgs) {
    args = ['install']
  } else if (options.packageManager === 'yarn') {
    args = ['add']
  }

  if (options.args) {
    args = args.concat(options.args)
  }
```

packages/cli-kit/src/public/node/node-package-manager.ts
```diff
@@ -207,10 +324,18 @@ export async function installNodeModules(options: InstallNodeModulesOptions): Pr
     stderr: options.stderr,
     signal: options.signal,
   }
-  let args = ['install']
+
+  let args: string[]
+  if (options.packageManager === 'yarn') {
+    args = await determineYarnInstallCommand(options.directory)
```
kdaviduik, line 330: I don't think this approach works, because it looks like it would break cases where `installNodeModules` is called with just flags but no dependencies to add (eg: `yarn install --force`). From my understanding:
- both yarn v1 and yarn berry use `yarn install` to just install dependencies (ie: not add, just install)
- To ADD dependencies:
    - yarn v1 supports `yarn install <package>` OR `yarn add <package>`
    - yarn berry ONLY supports `yarn add <package>`

So this seems like a much cleaner and simpler solution:
```
  let args: string[]
  // Are all of the args just flags (like `--force`)? Or are any of them actual packages to ADD?
  const hasPackageArgs = args?.some(arg => !arg.startsWith('-'))

  if (hasPackageArgs && options.packageManager === 'yarn') {
    args = ['add']
  } else {
    args = ['install']
  }

  if (options.args) {
    args = args.concat(options.args)
  }
```

PR #6203, opened 2025-08-01
Improve CLI messaging for app proxy development

nickwesselman:
## Summary
- Enhanced dev session logger to:
  - Allow multiple messages from an extension
  - Improve formatting of multiple messages (use the extension handle as log source, follow log streaming formatting of box characters)
  - Skip logging if the extension was deleted
- Added specific messaging for app proxy URL updates and important deployment behaviors, so that app proxy behaviors are more transparent to developers

image

## Changes
- Updated `getDevSessionUpdateMessage` to `getDevSessionUpdateMessages` to support multiple messages
- Added app proxy-specific messages showing the proxy URL being used
- Included warning about prefix/subpath changes only applying to new installations
- Improved message formatting with proper indentation for multiple messages

## Test plan
- [ ] Follow instructions to [tophat app proxy](https://docs.google.com/document/d/1-9Yaxhf4CfCqF9MD8LfIBDudk-0AvFNw57CWX-QrxOY/edit?usp=sharing) on the Dev Platform
- [ ] Run `shopify app dev` with app proxy configuration
- [ ] Verify new messaging appears in CLI output
- [ ] Confirm existing app access scope messages still work
- [ ] Test that deleted extensions don't log messages

### DIFF
--- a/packages/app/src/cli/services/dev/processes/dev-session/dev-session-logger.ts
+++ b/packages/app/src/cli/services/dev/processes/dev-session/dev-session-logger.ts
@@ -78,6 +78,7 @@ export class DevSessionLogger {
     const extensionEvents = event.extensionEvents ?? []
     const messageArrays = await Promise.all(
       extensionEvents.map(async (eve) => {
+        // Don't log messages for deleted extensions
         if (eve.type === EventType.Deleted) return []
         const messages = await eve.extension.getDevSessionUpdateMessages()
         return messages?.map((message) => ({message, prefix: eve.extension.handle})) ?? []


packages/app/src/cli/services/dev/processes/dev-session/dev-session-logger.ts
```diff
@@ -77,16 +76,22 @@ export class DevSessionLogger {
...
     )
+    const messages = messageArrays.flat()
 
-    const logPromises = messages.map((message) => this.log(`└  ${message}`, 'app-preview'))
+    const logPromises = messages.map((message, index) => {
+      const messageContent = outputContent`${outputToken.gray(index === messages.length - 1 ? '└ ' : '│ ')}${
+        message.message
+      }`.value
+      return this.log(messageContent, message.prefix)
+    })
```
alexanderMontague, line 94: > ✅ Updated app preview ....
> | ...
>└ ...

Will be shown on every _separate_ config change right? Just thinking of the UX when you have say this proxy update, then an access change right after (or together?) and how the "dropdown prefix characters" look. If that initial text breaks em up, I think thats fine

alexanderMontague (approved):

packages/app/src/cli/services/dev/processes/dev-session/dev-session-logger.ts
nickwesselman: Yep exactly. Here's an example of:

1. Changing access scopes + proxy at the same time, then
2. Just changing access scopes, then
3. Just changing proxy

image

PR #6228, opened 2025-08-08
Block dev when there are pending extensions to migrate

gonzaloriestra:

The new dev command may behave incorrectly if the app has extensions pending to migrate to the new dev platform (missing UIDs).

Raise an error when running dev before finishing the migration with deploy

Monosnap Hyper 2025-08-08 17-17-19

### How to test your changes?

- Migrate an org
- Run `p shopify app dev`

### Measuring impact

### Checklist

### DIFF
--- a/packages/app/src/cli/services/dev.ts
+++ b/packages/app/src/cli/services/dev.ts
@@ -242,8 +242,14 @@ export async function blockIfMigrationIncomplete(devConfig: DevConfig) {
   const remoteExtensions = (await developerPlatformClient.appExtensionRegistrations(remoteApp)).app
     .extensionRegistrations
   if (developerPlatformClient.supportsDevSessions && !remoteExtensions.every((extension) => extension.id)) {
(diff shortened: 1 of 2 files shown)


packages/app/src/cli/services/dev.ts
```diff
@@ -236,6 +237,23 @@ export async function warnIfScopesDifferBeforeDev({
   }
 }
 
+export async function blockIfMigrationIncomplete(devConfig: DevConfig) {
+  const {developerPlatformClient, remoteApp} = devConfig
+  const remoteExtensions = (await developerPlatformClient.appExtensionRegistrations(remoteApp)).app
```
alfonso-noriega: This line might trigger a request to the app management api, would it be a little bit more efficient to do it only if it is DevDash? if this line is under the `supportsDevSessions` condition it will only trigger when needed potentially reducing the number of remote requests.

gonzaloriestra: Right, thanks!

PR #6233, opened 2025-08-08
Update migration mutations to accept registration id or registration uuid based on which client we are using

jamieguerrero:

Note: Do not merge until changes to [Core](https://app.graphite.dev/github/pr/shop/world/133641) and [Partners](https://app.graphite.dev/github/pr/Shopify/partners/60107) have been merged

To support extension migration using both registration ID and UUID, allowing compatibility with the App Management API.

Updates GraphQL mutations for extension migration to accept either `registrationId` or `registrationUuid` as parameters:
- Modifies `MigrateAppModuleMutation`, `MigrateFlowExtensionMutation`, and `MigrateToUiExtensionQuery` to support both ID types
- Makes `registrationId` and `registrationUuid` optional in the respective variable interfaces
- Updates migration service functions to conditionally use UUID when working with the App Management API client

### How to test your changes?

1. Test extension migration using the Partner Dashboard API (should use registrationId)
2. Test extension migration using the App Management API (should use registrationUuid)
3. Verify both migration paths successfully migrate extensions

### Measuring impact

### Checklist

### DIFF
--- a/packages/app/src/cli/services/context/identifiers-extensions.test.ts
+++ b/packages/app/src/cli/services/context/identifiers-extensions.test.ts
@@ -885,12 +885,12 @@ describe('ensureExtensionsIds: Migrates extension', () => {
       },
     })
 
-    expect(migrateExtensionsToUIExtension).toBeCalledWith(
+    expect(migrateExtensionsToUIExtension).toBeCalledWith({
       extensionsToMigrate,
-      opts.appId,
+      appId: opts.appId,
       remoteExtensions,
-      expect.any(PartnersClient),
-    )
+      migrationClient: expect.any(PartnersClient),
+    })
   })
 })
 
--- a/packages/app/src/cli/services/context/identifiers-extensions.ts
+++ b/packages/app/src/cli/services/context/identifiers-extensions.ts
@@ -61,31 +61,31 @@ export async function ensureExtensionsIds(
 
   // Migration is only supported in partners client
   const clientName = options.developerPlatformClient.clientName
-  const migrationsClient = clientName === ClientName.Partners ? options.developerPlatformClient : new PartnersClient()
+  const migrationClient = clientName === ClientName.Partners ? options.developerPlatformClient : new PartnersClient()
 
   let didMigrateDashboardExtensions = false
 
   if (uiExtensionsToMigrate.length > 0) {
     const confirmedMigration = await extensionMigrationPrompt(uiExtensionsToMigrate)
     if (!confirmedMigration) throw new AbortSilentError()
-    remoteExtensions = await migrateExtensionsToUIExtension(
-      uiExtensionsToMigrate,
-      options.appId,
+    remoteExtensions = await migrateExtensionsToUIExtension({
+      extensionsToMigrate: uiExtensionsToMigrate,
+      appId: options.appId,
       remoteExtensions,
-      migrationsClient,
-    )
+      migrationClient,
+    })
     didMigrateDashboardExtensions = true
   }
 
   if (flowExtensionsToMigrate.length > 0) {
     const confirmedMigration = await extensionMigrationPrompt(flowExtensionsToMigrate, false)
     if (!confirmedMigration) throw new AbortSilentError()
-    const newRemoteExtensions = await migrateFlowExtensions(
-      flowExtensionsToMigrate,
-      options.appId,
-      dashboardExtensions,
-      migrationsClient,
-    )
+    const newRemoteExtensions = await migrateFlowExtensions({
+      extensionsToMigrate: flowExtensionsToMigrate,
+      appId: options.appId,
+      remoteExtensions: dashboardExtensions,
+      migrationClient,
+    })
     remoteExtensions = remoteExtensions.concat(newRemoteExtensions)
     didMigrateDashboardExtensions = true
   }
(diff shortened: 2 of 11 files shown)


jamieguerrero:
* **#6233** <a href="https://app.graphite.dev/github/pr/Shopify/cli/6233?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Shopify/cli/6233?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/app/src/cli/services/dev/migrate-app-module.ts
```diff
@@ -93,10 +93,16 @@ export async function migrateAppModules(
...
   type: string,
   remoteExtensions: RemoteSource[],
+  migrationClient: DeveloperPlatformClient,
   developerPlatformClient: DeveloperPlatformClient,
 ) {
   const migratedIDs = await Promise.all(
-    extensionsToMigrate.map(({remote}) => migrateAppModule(appId, remote.id, type, developerPlatformClient)),
+    extensionsToMigrate.map(({remote}) => {
+      if (developerPlatformClient.clientName === ClientName.AppManagement) {
+        return migrateAppModule(appId, undefined, remote.uuid, type, migrationClient).then(() => remote.id)
```
dmerand: Not a TS expert: Is it conventional to pass `undefined` in calls like this?

dmerand (approved):

packages/app/src/cli/services/dev/migrate-app-module.ts
craigmichaelmartin: Could we update `migrateAppModule` to use object destructuring?

craigmichaelmartin (dismissed):
Having so many positional arguments in `migrateAppModule` (especially with some being optional) makes my head hurt lol. Maybe there is a good reason for this, but worth a discussion!

jamieguerrero:
Yeah, it feels bad but it's the tradeoff we made to avoid changing the app module domain model to expose the registration ID database column. 😓 We opted to allow searching by uuid or ID instead.

packages/app/src/cli/services/dev/migrate-app-module.ts
jamieguerrero: Not a TS expert either, but Claude says that this is the convention and a quick Google says that all JS function parameters are undefined until they are set to some value.

packages/app/src/cli/services/dev/migrate-app-module.ts
```diff
@@ -88,15 +88,35 @@ export function getModulesToMigrate(
...
+        }).then(() => remote.id)
+      }
+      return migrateAppModule({
+        apiKey: appId,
+        registrationId: remote.id,
+        registrationUuid: undefined,
+        type,
+        developerPlatformClient,
+      })
+    }),
```
isaacroldan: If you always pass the `registrationUuid` it will always work no?
This:
```ts
return migrateAppModule({
          apiKey: appId,
          registrationId: undefined,
          registrationUuid: remote.uuid,
          type,
          developerPlatformClient: migrationClient,
        }).then(() => remote.id)
```
should work for both partners and app-management right?

packages/app/src/cli/services/dev/migrate-app-module.ts
```diff
@@ -88,15 +88,35 @@ export function getModulesToMigrate(
...
-  remoteExtensions: RemoteSource[],
-  developerPlatformClient: DeveloperPlatformClient,
-) {
+export async function migrateAppModules(options: {
+  extensionsToMigrate: LocalRemoteSource[]
+  appId: string
+  type: string
+  remoteExtensions: RemoteSource[]
+  migrationClient: DeveloperPlatformClient
+  developerPlatformClient: DeveloperPlatformClient
```
isaacroldan: Why two clients? you only need the migrationClient here (which is always a PartnersClient)

alfonso-noriega:
What if we only use the `registration uuid` always? then we do not need to have conditional checks specific to the client and we just need one parameter.

packages/app/src/cli/services/context/identifiers-extensions.test.ts
```diff
@@ -885,12 +884,12 @@ describe('ensureExtensionsIds: Migrates extension', () => {
...
 
-    expect(migrateExtensionsToUIExtension).toBeCalledWith(
+    expect(migrateExtensionsToUIExtension).toBeCalledWith({
       extensionsToMigrate,
-      opts.appId,
+      appId: opts.appId,
       remoteExtensions,
-      expect.any(PartnersClient),
-    )
+      migrationClient: expect.any(Object),
```
isaacroldan: this could still be `expect.any(PartnersClient)` no?

packages/app/src/cli/services/context/identifiers-extensions.ts
```diff
@@ -68,80 +68,80 @@ export async function ensureExtensionsIds(
...
-    remoteExtensions = await migrateExtensionsToUIExtension(
-      uiExtensionsToMigrate,
-      options.appId,
+    remoteExtensions = await migrateExtensionsToUIExtension({
+      extensionsToMigrate: uiExtensionsToMigrate,
+      appId: options.appId,
       remoteExtensions,
-      migrationsClient,
-    )
+      migrationClient: migrationsClient,
```
isaacroldan: nit: migration`S`client vs migrationclient. If you unify the names there'll be less changes
not blocking :)

packages/app/src/cli/services/dev/migrate-to-ui-extension.ts
```diff
@@ -7,14 +7,23 @@ import {LocalRemoteSource} from '../context/id-matching.js'
...
+
   await Promise.all(
-    extensionsToMigrate.map(({remote}) => migrateExtensionToUIExtension(appId, remote.id, developerPlatformClient)),
+    extensionsToMigrate.map(({remote}) =>
+      migrateExtensionToUIExtension({
+        apiKey: appId,
+        registrationId: undefined,
+        registrationUuid: remote.uuid,
+        migrationClient,
+      }).then(() => remote.id),
```
isaacroldan: why do we need the `then`? it wasn't present before no?

jamieguerrero: Previously when querying for registration by uuid, I was finding that the resulting array returns back empty... I just tophatted and it no longer does that. I'll remove this.

isaacroldan (approved):

PR #6234, opened 2025-08-08
Fetch and copy AI instructions from theme-liquid-docs

aswamy:

- Feature: Fetch new AI instructions from theme-liquid-docs
- Bug Fix: Only remove `.github`, `.cursor`, and `.claude` directories if it's NOT a skeleton theme
  - If you're cloning a theme repo that isn't ours (by passing in the `--clone-url` flag), we shouldn't be removing directories

- Fetch it into a temp directory and copy content
- Add a new utility that copies content from src directory instead of overriding the whole directory

### How to test your changes?
Initializing Skeleton Theme
- `shopify theme init`
- Select to create AI instructions (try each one)
- You should have a new directory in the file system `.github`, `.cursor, or `.claude` that contains AI instructions

Initializing non-Skeleton Theme
- `shopify theme init --clone-url "https://github.com/shopify-playground/alok-sandbox"`
- Select to create AI instructions (try each one)
- Existing `.github`, `.cursor, and `.claude` directories are NOT deleted
- AI Instructions are merged in with existing content 

### Post-release steps

n/a

### Measuring impact

### Checklist


packages/theme/src/cli/services/init.ts
```diff
@@ -1,8 +1,16 @@
...
 import {joinPath} from '@shopify/cli-kit/node/path'
-import {mkdir, writeFile, rmdir, fileExists} from '@shopify/cli-kit/node/fs'
-import {fetch} from '@shopify/cli-kit/node/http'
+import {rmdir, fileExists, inTemporaryDirectory, copyDirectoryContents} from '@shopify/cli-kit/node/fs'
+import {AbortError} from '@shopify/cli-kit/node/error'
+
+export const SKELETON_THEME_URL = 'https://github.com/Shopify/skeleton-theme.git'
+
+const SUPPORTED_AI_INSTRUCTIONS = {
+  github: 'VSCode (GitHub Copilot)',
```
graygilmore:
```suggestion
  github: 'VS Code (GitHub Copilot)',
```

packages/theme/src/cli/services/init.ts
```diff
@@ -37,41 +52,42 @@ async function removeDirectory(path: string) {
   }
 }
 
-export async function promptAndCreateAIFile(destination: string) {
+export async function promptAndCreateAIInstructions(destination: string) {
   const aiChoice = await renderSelectPrompt({
     message: 'Set up AI dev support?',
```
graygilmore: (nit, feel free to ignore) I know this is old text but it isn't quite clear to me what choosing one of these options will do. I wonder if this could be tweaked at all to be more descriptive?

Maybe something like "Select which LLM instructions to include in the root folder:". That's a pretty garbage suggestion but hopefully makes my comment more understandable.

packages/theme/src/cli/services/init.ts
```diff
@@ -37,41 +52,42 @@ async function removeDirectory(path: string) {
...
+      title: `Adding AI instructions into ${destination}`,
+      task: async () => {
+        await inTemporaryDirectory(async (tempDir) => {
+          await downloadGitRepository({
+            repoUrl: `https://github.com/Shopify/theme-liquid-docs.git`,
+            destination: tempDir,
+            shallow: true,
+          })
+          const aiSrcDir = joinPath(tempDir, 'ai', aiChoice)
+          const aiDestDir = joinPath(destination, `.${aiChoice}`)
```
karreiro: I believe the Claude file needs to be in the root of the project ([source](https://www.anthropic.com/engineering/claude-code-best-practices)). Am I missing something tho? 🤔

packages/cli-kit/src/public/node/fs.ts
```diff
@@ -593,3 +593,33 @@ export function matchGlob(key: string, pattern: string, options?: MatchGlobOptio
...
+  if (!(await fileExists(srcDir))) {
+    throw new Error(`Source directory ${srcDir} does not exist`)
+  }
+
+  if (!(await fileExists(destDir))) {
+    await mkdir(destDir)
+  }
+
+  // Get all files and directories in the source directory
+  const items = await glob(joinPath(srcDir, '**/*'))
```
karreiro, line 613: Does this work on Windows? I noticed we have some glob logic mocked in the tests, so I’m wondering if the Windows CI is passing because of that. Am I missing something? 🙈

packages/theme/src/cli/services/init.ts
aswamy: The fact that this function has an "and" in the name tells you it's doing too much 😅.

Maybe i should be split up the function into two:
`promptAIInstruction(): claude | github | cursor`
`createAIInstructions(themeRoot:string, aiInstruction: claude | github | cursor): Promise<void>`

I could also put a comment above the function if it's still funky

packages/cli-kit/src/public/node/fs.ts
aswamy: Good question. The tests in `fs.test.ts` did mock glob in the whole file, and i didn't want to touch it. But for my test block, i restored it so we can do accurate testing in Windows [here](http://github.com/Shopify/cli/pull/6234/files#diff-777a2125b818bfeea9e4b3f60f51ae2467bb070d16bb678ff707709f606eb205R328-R333)

packages/theme/src/cli/services/init.ts
aswamy: Oooo i got confused by the documentation, i read the `.claude/CLAUDE.md` directory and thought it was for the project - apparently you only do that for the home directory. Good catch! Will fix before merging.

packages/theme/src/cli/services/init.ts
graygilmore: I was more referring to the wording that we show to the end user! But you can still feel free to ignore 😄

aswamy: OH poop - too early when i saw this.

YES. I like the wording a bit better. Will update.

aswamy: Had to remove the word "Select" because of some linting rule, but done

packages/cli-kit/src/public/node/fs.ts
karreiro: Thank you for the clarification, @aswamy!

karreiro (approved):

gonzaloriestra (approved):

Issue #6224, opened 2025-08-08
[Bug]: `shopify app dev` doesn't server the host theme as expected with theme app extensions


karreiro:
`shopify app dev` doesn't work as expected for theme app extension development with Horizon.

### Context

`shopify app dev` can be used for developing theme app extensions. It serves local assets from the extension's `assets/` directory, which allows for hot-reloading during development.

When building theme app extensions, developers need to install them on a theme—this is known as the host theme—so the extension can interact with and extend the theme.

However, conflicts are happening between the asset endpoints (`assets/`, `style.css`, `scripts.js`, etc.) in the host theme and those in the theme app extension. As a result, JavaScript/CSS associated stylesheet/javascript tags (such as `style.css` or `scripts.js`) are not rendering as expected in the host theme:

<img width="50%" alt="Image" src="https://github.com/user-attachments/assets/34e0217c-acf8-4993-9c91-1c8b6cf787f6" />

daviareias:
same problem here, this has being going with horizon theme for at least 3 months.

gonzaloriestra:
This should be fixed in version 3.84.0.

daviareias:
It fixed :D

PR #6237, opened 2025-08-11
Fix `shopify app dev` to serve local assets as expected

karreiro:
### Why are these changes being introduced?

Fixes https://github.com/Shopify/cli/issues/6224

Previously, when running the theme proxy in the context of theme app extension development, it attempted to serve compiled assets locally. However, these assets couldn't be generated or served because the host theme doesn't exist in the local file system.

### What does this pull request do?

This PR updates the `DevServerSession` to identify the context in which it is running and avoid asset generation when the proxy is running in the context of theme app extension development.

### How can you test these changes?

- Run `shopify app dev -t <horizon_theme_id>` in an app.
- Verify that Horizon loads as expected.

**Before:**

https://github.com/user-attachments/assets/b0d9f2b7-0876-4dd9-854c-fead3ab0c72e

**After:**

https://github.com/user-attachments/assets/2f6b69a0-4f45-47c5-8e82-47c4fc07f248

### Post-release steps

N/A

### Measuring impact

How will we know if this change was effective? Please select one:

- [ ] n/a – this doesn't need measurement, e.g., a linting rule or a bug fix
- [x] Existing analytics will capture this change

### Checklist

### DIFF
--- a/.changeset/twelve-tools-exercise.md
+++ b/.changeset/twelve-tools-exercise.md
@@ -1,0 +1,5 @@
+---
+'@shopify/theme': patch
+---
+
+Fix `shopify app dev` to serve local assets as expected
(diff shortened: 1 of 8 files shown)


EvilGenius13 (approved):
Thank you for the PR @karreiro!

PR #6244, opened 2025-08-12
Harden network retry classification

EvilGenius13:
References: https://github.com/Shopify/cli/issues/6062

- Improve resilience of transient network failures by consistently identifying retryable conditions.
- Reduce flakiness caused by casing differences and stale keep‑alive sockets.
- Fix crashes related to OTEL

- Normalize retryable error detection:
   - Compare error messages in lowercase to avoid casing mismatches.
- Refine retryable set:
   - Replace EAI_FAIL with EAI_AGAIN (temporary DNS failure is retryable, hard DNS failure is not).
   - Add EPIPE (write to stale/closed keep‑alive socket, retry establishes a fresh connection).
   - Treat `The user aborted a request` as retryable
   - Create a match for blank reasons ie: `failed, reason:` _NOTE:_ I believe the reason we are seeing these errors is that they are caused when using a VPN or Proxy that drops or resets the connection abruptly. In this event no error is logged.
- OTEL crashes
   - This is caused when a network issue happens when a promise gets rejected. We are now silently swallowing OTEL related errors
Quick breakdown on network errors:
- ECONNRESET/ECONNABORTED/ETIMEDOUT/ENETUNREACH/ECONNREFUSED: common transient transport issues.
- EAI_AGAIN: temporary DNS failure; often resolves on retry.
- EPIPE: stale socket on first write, retrying re-opens a new connection.
- Casing: providers differ in their messages, lowercasing ensures we are always catching errors.

Tests
- Added unit tests to verify:
   - Retries when the first attempt fails with “The user aborted a request.” and succeeds on retry.
   - Retries when the first attempt fails with the blank-reason wrapper “request to … failed, reason:” and succeeds on retry.

### Measuring impact

### Checklist

### DIFF
--- a/packages/cli-kit/src/private/node/api.ts
+++ b/packages/cli-kit/src/private/node/api.ts
@@ -82,7 +82,6 @@ function isARetryableNetworkError(error: unknown): boolean {
       'eai_again',
       'epipe',
       'the operation was aborted',
-      'the user aborted a request',
     ]
     const errorMessage = error.message.toLowerCase()
     const anyMatches = networkErrorMessages.some((issueMessage) => errorMessage.includes(issueMessage))
--- a/packages/cli-kit/src/private/node/api.test.ts
+++ b/packages/cli-kit/src/private/node/api.test.ts
@@ -203,7 +203,7 @@ describe('retryAwareRequest', () => {
     const mockRequestFn = vi
       .fn()
       .mockImplementationOnce(() => {
-        throw new Error('The user aborted a request.')
+        throw new Error('the operation was aborted')
       })
       .mockImplementationOnce(() => {
         return Promise.resolve({


packages/cli-kit/src/private/node/api.ts
```diff
@@ -72,18 +72,22 @@ function isARetryableNetworkError(error: unknown): boolean {
...
+      'enotfound',
+      'enetunreach',
       'network socket disconnected',
-      'ETIMEDOUT',
-      'ECONNREFUSED',
-      'EAI_FAIL',
-      'The operation was aborted.',
+      'etimedout',
+      'econnrefused',
+      'eai_again',
```
isaacroldan, line 82: is this one intentional? `eai_fail` is missing now

packages/cli-kit/src/private/node/api.ts
```diff
@@ -72,18 +72,22 @@ function isARetryableNetworkError(error: unknown): boolean {
...
+      'epipe',
+      'the operation was aborted',
+      'the user aborted a request',
```
isaacroldan: `the user aborted a request` is usually the error shown then there is a local timeout, shouldn't be triggered by node directly I believe, have you seen it here?

packages/cli-kit/src/private/node/api.ts
EvilGenius13: I did intentionally replace this. `eai_fail` is considered a non recoverable failure where as `eai_again` would be a temporary condition that can be retried. I'm happy to add back `eai_fail` again if needed!

packages/cli-kit/src/private/node/api.ts
EvilGenius13: I agree with you. I slept on this and also realized if we did see it, we might cause looping behaviour. I'm going to remove this.

EvilGenius13: I'm going to add to my original comment so we don't lose context. 
I've gone over the flow again. 
- If a request has the `AbortSignal` enabled, then we will see `the user aborted a request` show up for node-fetch and `the operation was aborted` when undici throws it
- There wouldn't be a looping behaviour as we have the `maxRetryTimeMs` and the retry count limit.

It might actually be good to keep it as an option for retries. We see this kind of behaviour more often with theme commands.

packages/cli-kit/src/private/node/api.ts
isaacroldan: we got this list from somewhere, but your reasoning makes sense, let's keep it like this

packages/cli-kit/src/private/node/api.ts
isaacroldan: But should we retry if we are the ones triggering the timeout though? Also you removed timeouts for theme queries no?

EvilGenius13: I'm going to remove it. We had some `user aborted a request` showing in theme commands and it turns out we weren't properly passing the custom request behaviour. I'm solving that in this [PR](https://github.com/Shopify/cli/pull/6265). There is always the possibility that an abort was based on a temporary network issue where the next retry would work but since we aren't using them in theme commands like you said (or we won't be soon) - I'll get rid of it.

isaacroldan (approved):

PR #6245, opened 2025-08-12
[Enhancement] Add enriched error message and path for json schema `UnionType` errors

lunataoo:

Fixes https://github.com/shop/issues-payment-partners/issues/172


**TLDR**
This PR adds more error details to the union type - so that it gives the the errored field, and specific error message.

**Problem:**
For `PaymentsAppExtensionSchema`, which is a `zod.union` type, when a required field is missing in one of the schema type, the error message is really unclear, i.e. `[]: Invalid input.`

```
╭─ error ──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│                                                                                                                                                                                                                      │
│  App configuration is not valid                                                                                                                                                                                      │
│  Validation errors in ../cli-apps/test/extensions/credit-card/shopify.extension.toml:                                                                                                                                │
│                                                                                                                                                                                                                      │
│  • []: Invalid input                                                                                                                                                                                                 │
│                                                                                                                                                                                                                      │
│                                                                                                                                                                                                                      │
╰──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯

 ELIFECYCLE  Command failed with exit code 1.
```

Improves Zod union error handling by intelligently detecting the most likely
  intended configuration variant and showing only relevant errors, instead of
  confusing users with all possible union variants.

  Key changes:
  - Add smart variant scoring algorithm that prioritizes missing fields over type
  errors
  - Show specific errors for the intended variant (e.g., "• [supports_moto]: supports_moto is required.")
  - Fallback to showing all variants only when variant detection fails
  - Add comprehensive test coverage for edge cases and scoring logic


### How to test your changes?
1. Tests added for coverage
2. local testing
  - In `cli` repo, run command `SHOPIFY_CLI_1P_DEV=1 pnpm shopify app deploy --path="/Users/<user>/src/github.com/Shopify/payment-apps/etc/cli/bogus-card-payments-app/" --config shopify.app.production.toml` on this branch
  - Since current bogus card app toml is broken, it should display the correct error code that the required field `supports_moto` is missing:
<img width="1085" height="286" alt="Monosnap loader ts — cli 2025-08-13 16-12-07" src="https://github.com/user-attachments/assets/532173f5-3db0-4bb3-8460-5659dd34ac73" />

### Post-release steps

Test invalid config changes on [Bogus payments app](https://partners.shopify.com/internal/apps/3511727). 

### Measuring impact

- [x] test on bogus payments app in production and verify a clearer error message is displayed.

### Checklist

- [x] N/A-  I've considered possible cross-platform impacts (Mac, Linux, Windows)
- [x] N/A - I've considered possible [documentation](https://shopify.dev) changes


isaacroldan:
I know that parseHumanReadableError was already in `loader.ts` so it made sense to extend that, but that file has grown a lot, can we extract this to its own file? we could create a new folder for loader utilities like this one.

packages/app/src/cli/models/app/loader.ts
```diff
@@ -126,12 +126,107 @@ export async function parseConfigurationFile<TSchema extends zod.ZodType>(
...
+
+      if (bestVariant && bestVariant.issues && bestVariant.issues.length > 0) {
+        // Show errors only for the best matching variant
+        bestVariant.issues.forEach((nestedIssue) => {
+          const path = nestedIssue.path && nestedIssue.path.length > 0 ? nestedIssue.path.map(String).join('.') : 'root'
+          humanReadableError += `• [${path}]: ${nestedIssue.message}\n`
+        })
+      } else {
+        // Fallback to showing all variants if we can't determine the best one
+        humanReadableError += `• Union validation failed. None of the possible variants matched:\n`
```
isaacroldan: I don't think message is useful to users, they don't know anything about `unions` or `variants` since these are internal concepts. Can we express this in a different way?

lunataoo: updated to ` Configuration doesn't match any expected format:` in [ff432d0](https://github.com/Shopify/cli/pull/6245/commits/ff432d01c9d93c9b27318835d49423212d2f3570)

packages/app/src/cli/models/app/loader.ts
```diff
@@ -126,15 +127,6 @@ export async function parseConfigurationFile<TSchema extends zod.ZodType>(
   return {...configuration, path: filepath}
 }
 
-export function parseHumanReadableError(issues: Pick<zod.ZodIssueBase, 'path' | 'message'>[]) {
```
delongGao, line 129: We still need to export the `parseHumanReadableError` function right? Otherwise it wouldn't be available to use from the loader.

delongGao (commented):
TY @lunataoo! One question left.

packages/app/src/cli/models/app/loader.ts
lunataoo: Good question. I think it's fine considering no other class is using the `parseHumanReadableError` function, and it's now exported from `error-parsing.ts` file [here](https://github.com/Shopify/cli/blob/ff432d01c9d93c9b27318835d49423212d2f3570/packages/app/src/cli/models/app/error-parsing.ts#L65). In `loader.ts`, it is [imported](https://github.com/Shopify/cli/blob/ff432d01c9d93c9b27318835d49423212d2f3570/packages/app/src/cli/models/app/loader.ts#L22). If I missed anything, please point it out TY! 😺

delongGao: Ah cool, if it's not used anywhere else then no need to export then, thanks!

packages/app/src/cli/models/app/error-parsing.ts
```diff
@@ -0,0 +1,103 @@
...
+          }
+        })
+      }
+    } else {
+      // Handle regular issues
+      const path = issue.path && issue.path.length > 0 ? issue.path.map(String).join('.') : 'root'
+      const message = issue.message ?? 'Unknown error'
+      humanReadableError += `• [${path}]: ${message}\n`
+    }
+  })
```
isaacroldan, line 96: The logic in this function makes sense, but there are a few code duplications and unnecessary double checks, here is a proposal for a simplified solution:

```ts
export function parseHumanReadableError(issues: ExtendedZodIssue[]) {
  let humanReadableError = ''

  function appendError(issue: ExtendedZodIssue) {
    const path = issue.path && issue.path.length > 0 ? issue.path.map(String).join('.') : 'root'
    const message = issue.message ?? 'Unknown error'
    humanReadableError += `• [${path}]: ${message}\n`
  }

  issues.forEach((issue) => {
    // Handle union errors with smart variant detection
    if (issue.code === 'invalid_union' && issue.unionErrors) {
      // Find the variant that's most likely the intended one
      const bestVariant = findBestMatchingVariant(issue.unionErrors)

      if (bestVariant?.issues?.length) {
        // Show errors only for the best matching variant
        bestVariant.issues.forEach(appendError)
      } else {
        // Fallback to showing all variants if we can't determine the best one
        humanReadableError += `• Configuration doesn't match any expected format:\n`
        issue.unionErrors.forEach((unionError, index) => {
          humanReadableError += `  Option ${index + 1}:\n`
          unionError.issues?.forEach(appendError)
        })
      }
    } else {
      // Handle regular issues
      appendError(issue)
    }
  })

  return humanReadableError
}
```

packages/app/src/cli/models/app/error-parsing.ts
```diff
@@ -0,0 +1,103 @@
...
+export function parseHumanReadableError(issues: ExtendedZodIssue[]) {
+  let humanReadableError = ''
+
+  issues.forEach((issue) => {
+    // Handle union errors with smart variant detection
+    if (issue.code === 'invalid_union' && issue.unionErrors) {
+      // Find the variant that's most likely the intended one
+      const bestVariant = findBestMatchingVariant(issue.unionErrors)
+
+      if (bestVariant && bestVariant.issues && bestVariant.issues.length > 0) {
```
isaacroldan: here you can just do `if (bestVariant?.issues?.length) {`

packages/app/src/cli/models/app/error-parsing.ts
```diff
@@ -0,0 +1,103 @@
...
+                nestedIssue.path && nestedIssue.path.length > 0 ? nestedIssue.path.map(String).join('.') : 'root'
+              humanReadableError += `    • [${path}]: ${nestedIssue.message}\n`
+            })
+          }
+        })
+      }
+    } else {
+      // Handle regular issues
+      const path = issue.path && issue.path.length > 0 ? issue.path.map(String).join('.') : 'root'
+      const message = issue.message ?? 'Unknown error'
```
isaacroldan: This piece of code appears 3 times in the function, it can be extracted to a helper function

packages/app/src/cli/models/app/error-parsing.ts
lunataoo: TY! This is great feedback, appreciate you giving code snippet examples! I've updated the code to match this clean structure

packages/app/src/cli/models/app/error-parsing.ts
lunataoo: updated!

delongGao:
Thanks for reviewing this change Isaac! I will merge this branch on behalf of Luna.

Issue #6205, opened 2025-08-02
[Bug]: Theme builds error when uploaded to the theme store


bakura10:
### Please confirm that you have:

- [x] Searched [existing issues](.) to see if your issue is a duplicate. (If you’ve found a duplicate issue, feel free to add additional information in a comment on it.)
- [x] Reproduced the issue in [the latest CLI version](https://www.npmjs.com/package/@shopify/cli).

### In which of these areas are you experiencing a problem?

Theme

### Expected behavior

Hello,

When I create a zip for a theme, when I send the zip to the theme store, I am facing the following error:

<img width="2038" height="256" alt="Image" src="https://github.com/user-attachments/assets/a4ac83e9-8945-4923-b8c3-23090448be11" />

The strange thing is that the `/listings` folder is here, it contains the folder for each preset. If I generate manually the zip using OSX archive tool, it works. I've compared the two zip (the one generated by Shopify CLI and the one I generate manually), they are exactly the same.

Even weirder, if I take the archive generated by Shopify CLI, extract it and recreate it manually, it works, confirming that all the files are here.

It seems to me that it might be an encoding issue of Shopify CLI when generating the `/listings` folder. I can't really publish the zip here as our theme is not open source, but if someone from Shopify could send me a DM on Slack (I'm bakura10) so I can send them the archive, that would help.

### Actual behavior

It didn't work

### Verbose output

<details>
  <summary>Verbose output</summary>

  ```
  Paste the output here!
  ```

</details>

### Reproduction steps

1.
2.
3.

### Operating System

Mac OS

### Shopify CLI version (`shopify --version`)

3.83.1

PR #6247, opened 2025-08-13
Fix theme package folder structure

EvilGenius13:

Closes: https://github.com/Shopify/cli/issues/6205

Uploading a zip created by `theme package` could fail validation when uploading to a shopify store. Some validators expect folders to be present as explicit entries in the zip, not just implied by the files inside them.

Right now, theme package zips the files it finds, but doesn’t add the folders themselves as entries. This can trigger validation errors.

This PR updates the process to:
- Collect the folder paths for the files we’re zipping
- Add those folders to the zip first (as metadata)
- Add the files
This makes the zip include the folder info validators look for. In the reported case, unzipping and re-zipping with the macOS archiver worked because it preserved that folder structure metadata.

#### To replicate the old and new behaviour
- Build the main branch or use a recent version of the CLI
- Run `theme package` on a theme
- Run `unzip -l <your_theme> | grep "sections"` on the zipped theme
You should see an output like:
```bash
unzip -l Lose-1.0.0.zip | grep "sections"
     1201  03-11-2025 19:12   sections/custom-section.liquid
      167  03-11-2025 19:21   sections/group.liquid
      663  03-31-2025 19:41   sections/layout-header-test.liquid
```
- Build this PR's branch `package-fix`
- Run `theme package` on a theme
- Run `unzip -l <your_theme> | grep "sections"` on the zipped theme
You should see the parent folder listed in the metadata in the output
```bash
❯ unzip -l Lose-1.0.0.zip | grep "sections"
        0  08-18-2025 18:13   sections/ <----- This should show up for you
     1201  03-11-2025 19:12   sections/custom-section.liquid
      167  03-11-2025 19:21   sections/group.liquid
      663  03-31-2025 19:41   sections/layout-header-test.liquid
```

### Measuring impact

### Checklist

### DIFF
--- a/packages/cli-kit/src/public/node/archiver.integration.test.ts
+++ b/packages/cli-kit/src/public/node/archiver.integration.test.ts
@@ -26,7 +26,7 @@ describe('zip', () => {
 
       // Then
       const archiveEntries = await readArchiveFiles(zipPath)
-      const expectedEntries = withExplicitDirectoryEntries(structure)
+      const expectedEntries = ['extensions/', 'extensions/first/', 'extensions/first/main.js', 'test.json']
       expect(expectedEntries.sort()).toEqual(archiveEntries.sort())
     })
   })
@@ -50,7 +50,10 @@ describe('zip', () => {
 
       // Then
       const archiveEntries = await readArchiveFiles(zipPath)
-      const expectedEntries = withExplicitDirectoryEntries([`extensions/first/main.js`])
+
+      expect(archiveEntries).toContain('extensions/')
+      expect(archiveEntries).toContain('extensions/first/')
+      const expectedEntries = ['extensions/', 'extensions/first/', 'extensions/first/main.js']
       expect(expectedEntries.sort()).toEqual(archiveEntries.sort())
     })
   })
@@ -158,20 +161,4 @@ async function readArchiveFiles(zipPath: string) {
   await archive.close()
 
   return archiveEntries
-}
-
-function withExplicitDirectoryEntries(files: string[]): string[] {
-  const entries = new Set<string>()
-  for (const file of files) {
-    entries.add(file)
-    let currentDir = dirname(file)
-    while (currentDir && currentDir !== '.' && currentDir !== '/') {
-      const dirEntry = currentDir.endsWith('/') ? currentDir : `${currentDir}/`
-      entries.add(dirEntry)
-      const parent = dirname(currentDir)
-      if (parent === currentDir) break
-      currentDir = parent
-    }
-  }
-  return Array.from(entries)
 }
--- a/packages/cli-kit/src/public/node/archiver.integration.test.ts
+++ b/packages/cli-kit/src/public/node/archiver.integration.test.ts
@@ -26,7 +26,8 @@ describe('zip', () => {
 
       // Then
       const archiveEntries = await readArchiveFiles(zipPath)
-      expect(structure.sort()).toEqual(archiveEntries.sort())
+      const expectedEntries = withExplicitDirectoryEntries(structure)
+      expect(expectedEntries.sort()).toEqual(archiveEntries.sort())
     })
   })
 
@@ -49,7 +50,8 @@ describe('zip', () => {
 
       // Then
       const archiveEntries = await readArchiveFiles(zipPath)
-      expect([`extensions/first/main.js`]).toEqual(archiveEntries)
+      const expectedEntries = withExplicitDirectoryEntries([`extensions/first/main.js`])
+      expect(expectedEntries.sort()).toEqual(archiveEntries.sort())
     })
   })
 })
@@ -156,4 +158,20 @@ async function readArchiveFiles(zipPath: string) {
   await archive.close()
 
   return archiveEntries
+}
+
+function withExplicitDirectoryEntries(files: string[]): string[] {
+  const entries = new Set<string>()
+  for (const file of files) {
+    entries.add(file)
+    let currentDir = dirname(file)
+    while (currentDir && currentDir !== '.' && currentDir !== '/') {
+      const dirEntry = currentDir.endsWith('/') ? currentDir : `${currentDir}/`
+      entries.add(dirEntry)
+      const parent = dirname(currentDir)
+      if (parent === currentDir) break
+      currentDir = parent
+    }
+  }
+  return Array.from(entries)
 }
--- a/packages/cli-kit/src/public/node/archiver.ts
+++ b/packages/cli-kit/src/public/node/archiver.ts
@@ -52,24 +52,21 @@ export async function zip(options: ZipOptions): Promise<void> {
     })
     archive.pipe(output)
 
-    // Find parent directories to add explicitly to the archive
     const directoriesToAdd = new Set<string>()
     for (const filePath of pathsToZip) {
-      const relPath = relativePath(inputDirectory, filePath)
-      collectParentDirectories(relPath, directoriesToAdd)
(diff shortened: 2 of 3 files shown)


packages/cli-kit/src/public/node/archiver.ts
graygilmore, line 1: 🧵 

Did this only fail for themes because there's an explicit check for the listings folder? Whereas for validation for other required files we only check for the presence of the file?

packages/cli-kit/src/public/node/archiver.ts
```diff
@@ -52,16 +52,41 @@ export async function zip(options: ZipOptions): Promise<void> {
...
+    const sortedDirs = Array.from(directoriesToAdd).sort((left, right) => left.localeCompare(right))
+    for (const dir of sortedDirs) {
+      const dirName = dir.endsWith('/') ? dir : `${dir}/`
+      archive.append(Buffer.alloc(0), {name: dirName})
+    }
+
+    // Add files
+    for (const filePath of pathsToZip) {
+      const rel = relativePath(inputDirectory, filePath)
+      if (filePath && rel) archive.file(filePath, {name: rel})
```
graygilmore: (nit) prefer using full name variables

```suggestion
      const fileRelativePath = relativePath(inputDirectory, filePath)
      if (filePath && fileRelativePath) archive.file(filePath, {name: fileRelativePath})
```

packages/cli-kit/src/public/node/archiver.integration.test.ts
```diff
@@ -49,7 +50,8 @@ describe('zip', () => {
 
       // Then
       const archiveEntries = await readArchiveFiles(zipPath)
-      expect([`extensions/first/main.js`]).toEqual(archiveEntries)
+      const expectedEntries = withExplicitDirectoryEntries([`extensions/first/main.js`])
+      expect(expectedEntries.sort()).toEqual(archiveEntries.sort())
```
graygilmore, line 57: Instead of doing this programatically can we have some explicit tests to make sure that we're seeing the directories?

packages/cli-kit/src/public/node/archiver.ts
```diff
@@ -52,16 +52,41 @@ export async function zip(options: ZipOptions): Promise<void> {
     })
     archive.pipe(output)
 
+    // Find parent directories to add explicitly to the archive
+    const directoriesToAdd = new Set<string>()
```
graygilmore, line 55: Could we have used `archiver.directory()` to help with anything here? https://github.com/archiverjs/node-archiver/blob/master/lib/core.js#L570

packages/cli-kit/src/public/node/archiver.ts
EvilGenius13: That's what it looks like. It was probably the way they run their validation to make sure each folder had a file within that corresponds to the proper listing. 

It's possible another folder structure makes its way into themes in the future and we would run into the same problem so it makes sense to fix it for everything. 

If I use the built in OSX archiver tool and run the unzip command, it adds more detailed metadata as well
image

packages/cli-kit/src/public/node/archiver.ts
EvilGenius13: Fixed

packages/cli-kit/src/public/node/archiver.integration.test.ts
EvilGenius13: Updated the test!

packages/cli-kit/src/public/node/archiver.ts
EvilGenius13: I took a look into `archiver.directory` and `archiver.glob`

- `archiver.directory` will Add a directory and files recursively but has no ability to filter so it will just add everything
- `archiver.glob` supports patterns but it only adds file and not directory entries.

I tried a few variations of the code but couldn't find a more elegant solution than what we have. I will add additional comments in the PR for what each part is doing though!

packages/cli-kit/src/public/node/archiver.ts
```diff
@@ -52,16 +52,38 @@ export async function zip(options: ZipOptions): Promise<void> {
     })
     archive.pipe(output)
 
+    const directoriesToAdd = new Set<string>()
     for (const filePath of pathsToZip) {
       const fileRelativePath = relativePath(inputDirectory, filePath)
```
EvilGenius13, line 57: At this spot we're grabbing all the relative paths for the files. 
EX:
```bash
fileRelativePath assets/asuna.png
fileRelativePath assets/scripts.js.liquid
fileRelativePath assets/style.css.liquid
fileRelativePath blocks/group.liquid
fileRelativePath blocks/text.liquid
fileRelativePath config/settings_data.json
fileRelativePath config/settings_schema.json
fileRelativePath layout/theme.liquid
fileRelativePath sections/custom-section.liquid
fileRelativePath sections/group.liquid
fileRelativePath sections/layout-header-test.liquid
fileRelativePath snippets/header.liquid
fileRelativePath snippets/temp.liquid
fileRelativePath templates/404.liquid
fileRelativePath templates/article.liquid
fileRelativePath templates/blog.liquid
fileRelativePath templates/cart.liquid
fileRelativePath templates/collection.alternate.liquid
fileRelativePath templates/collection.liquid
fileRelativePath templates/gift_card.liquid
fileRelativePath templates/index.liquid
fileRelativePath templates/list-collections.liquid
fileRelativePath templates/page.liquid
fileRelativePath templates/password.liquid
fileRelativePath templates/product.liquid
fileRelativePath templates/search.liquid
fileRelativePath listings/voltar/templates/index.json
fileRelativePath listings/doktor_frogg/templates/index.json
```

packages/cli-kit/src/public/node/archiver.ts
```diff
@@ -52,16 +52,38 @@ export async function zip(options: ZipOptions): Promise<void> {
...
-      archive.file(filePath, {name: fileRelativePath})
+      collectParentDirectories(fileRelativePath, directoriesToAdd)
+    }
```
EvilGenius13, line 59: At this point we have sourced all the directories we need to add (this is for the metadata). We make a Set
```bash
directoriesToAdd Set(12) {
  'assets',
  'blocks',
  'config',
  'layout',
  'sections',
  'snippets',
  'templates',
  'listings/voltar/templates',
  'listings/voltar',
  'listings',
  'listings/doktor_frogg/templates',
  'listings/doktor_frogg'
}
```

packages/cli-kit/src/public/node/archiver.ts
```diff
@@ -52,16 +52,38 @@ export async function zip(options: ZipOptions): Promise<void> {
...
+
+    const sortedDirs = Array.from(directoriesToAdd).sort((left, right) => left.localeCompare(right))
+    for (const dir of sortedDirs) {
+      const dirName = dir.endsWith('/') ? dir : `${dir}/`
+      archive.append(Buffer.alloc(0), {name: dirName})
+    }
```
EvilGenius13, line 65: Now we have sorted them, alphabetically and by folder structure
```bash
sortedDirs [
  'assets',
  'blocks',
  'config',
  'layout',
  'listings',
  'listings/doktor_frogg',
  'listings/doktor_frogg/templates',
  'listings/voltar',
  'listings/voltar/templates',
  'sections',
  'snippets',
  'templates'
]
```

dejmedus:
🎩 looks good!
tophat-pkg-folder-structure
tophat-inner-folders-zipped

gonzaloriestra (approved):

PR #6256, opened 2025-08-13
[Bugfix] allow relative urls to apply during `app dev` sessions

jackseok:
Resolve https://github.com/shop/issues-develop/issues/21038

Fixes a bug where relative URLs are not being applied through `dev`

### Measuring impact

### Checklist

### Tophat 🎩 
https://screenshot.click/13-17-n7y6x-r93yb.mp4

### DIFF
--- a/packages/app/src/cli/services/dev/urls.ts
+++ b/packages/app/src/cli/services/dev/urls.ts
@@ -3,6 +3,7 @@ import {AppConfigurationInterface, AppLinkedInterface, CurrentAppConfiguration} 
 import {UpdateURLsSchema, UpdateURLsVariables} from '../../api/graphql/update_urls.js'
 import {setCachedAppInfo} from '../local-storage.js'
 import {AppConfigurationUsedByCli} from '../../models/extensions/specifications/types/app_config.js'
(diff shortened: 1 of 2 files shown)


packages/app/src/cli/services/dev/urls.ts
```diff
@@ -173,7 +174,7 @@ export function generateApplicationURLs(
   const appProxy = proxyFields
     ? {
         appProxy: {
-          proxyUrl: replaceHost(proxyFields.url, baseURL),
+          proxyUrl: replaceHost(prependApplicationUrl(proxyFields.url, baseURL), baseURL),
```
pt2pham, line 177: So the main reason this was missed is because we have special logic for replacing the host for `dev` like the other app URLs, and we just need to make this support the relative URL structure too? 

Makes sense to me

JonathanJChang (approved):

packages/app/src/cli/services/dev/urls.ts
jackseok: exactly, looked like this `prependApplicationUrl` was only being called within our `deploy` flow.

PR #6193, opened 2025-07-30
Allow theme delete command to be called with multiple environments

dejmedus:
### WHY are these changes introduced?

Previously `theme delete` command could only be run with a single `--environment` flag. This PR allows `delete` to be run with multiple environments in a single command.

Since this change required refactoring `delete` command to utilize ThemeCommand's `run` method, this was a good opportunity to remove the deprecated positional arguments logic and warning.

### WHAT is this pull request doing?

- Removes flag parsing and auth from `delete` in favour of ThemeCommand's `run` logic
- Prefixes output with the environment name, when present
- Removes deprecated positional arguments logic and warning

<details><summary>Deletes multiple themes</summary>

https://github.com/user-attachments/assets/838f87a9-3b16-491f-a210-5ff81f7f7090

</details> 

<details><summary>Deletes development themes</summary>

https://github.com/user-attachments/assets/30eda153-0c74-4af4-9dff-3d1a3ff890b9

</details> 

<details><summary>Displays warning when missing flags</summary>

https://github.com/user-attachments/assets/09153711-e617-4533-b0a5-c595574fc5d5

</details> 

### How to test your changes?

- Pull down the branch `gh pr checkout` or install the snapit version
- Add a `shopify.theme.toml` file

<details><summary>Example toml</summary>

```
[environments.store1]
theme = "<theme>"
store = "<store>"
password  = "<shptka_password>"

[environments.store2]
theme = "<theme>"
store = "<store>"
password  = "<shptka_password>"
```

</details> 

```ts
shopify theme delete -e store1 -e store2
shopify theme delete -e store1 -e store2 -t "Spring Theme"
shopify theme delete -e store1 -e store2 --development
```
- Should accept flags from both CLI and `shopify.theme.toml`
- When run with a single environment or no environment should work as normal
- When `--force` flag is used the confirmation prompt should not be shown
- Environments missing required flags (store, password, and either theme or development) should display a warning and it's theme shouldn't be deleted
- Environments with incorrect flags (ex. password = "fake") should have the environment name prefixed to the error message

> If provided both `development` and `theme`, `development` will take precedence (same as current single or no env runs)

### Measuring impact

How do we know this change was effective? Please choose one:

- [x] Existing analytics will cater for this addition

### Checklist

- [x] I've considered possible cross-platform impacts (Mac, Linux, Windows)
- [x] I've considered possible [documentation](https://shopify.dev) changes

PR #6257, opened 2025-08-13
Prevent race conditions during multi environment execution

dejmedus:

Commands with multiple environment flags are run concurrently. This allows for quick multi env command runs (_`push` with 10 environments takes 2.5s run concurrently vs 12s sequentially!_) but introduces race conditions in commands that rely on local storage read/writes. Theme commands use local storage to keep track of the current store URL as well as each store's current development theme.

Examples:

- `shopify theme delete -e store1 -e store2 --development` An environment would sometimes try to lookup the current dev theme using the other env's store as key if the store file was overwritten before the dev theme read
- `shopify theme push -e store1-env -e another-store1-env --development` Two new dev themes were being created since both environment's local storage reads occurred before either's writes

> Something to keep an eye on: [the package](https://github.com/sindresorhus/conf) CLI uses to manage local storage doesn't technically support concurrent writes

- Wraps each command call in an context provider so it always knows which store its running on, eliminating the need to grab the store URL from local storage. 
- Groups environments with unique store URLs and runs those groups sequentially, preventing later called environments from reading outdated values

### How to test your changes?

- Install the snap build that contains these changes as well as `delete` multi env logic

```
npm i -g @shopify/cli@0.0.0-snapshot-20250813233940 --@shopify:registry=https://registry.npmjs.org
```

```
[environments.store1]
theme = "<theme>"
store = "<store1>"

[environments.anotherstore1]
theme = "<theme>"
store = "<store1>"

[environments.store2]
theme = "<theme>"
store = "<store2>"
```

</details> 

- Running commands with multiple environments acting on the same store URL should work as expected. Currently `list`, `info`, `rename` work with multiple environments
- Development theme ID's should be read, stored, and deleted correctly from the local storage file `/Users/<username>/Library/Preferences/shopify-cli-development-theme-config-nodejs/config.json`

```sh
shopify theme push -e store1 --development
shopify theme push -e store2 --development
shopify theme delete -e store1 -e store2 --development
```

> Heads up: `shopify theme info --development` will create a new dev theme and display it's info if one does not exist, I believe this is unintended and will open a follow up PR (it's not a result of this change) 

### Measuring impact

### Checklist

### DIFF
--- a/.changeset/many-ties-check.md
+++ b/.changeset/many-ties-check.md
@@ -1,0 +1,5 @@
+---
+'@shopify/theme': minor
+---
+
+Allow commands run in multiple environments to act on the same store URL
--- a/packages/theme/src/cli/services/local-storage.ts
+++ b/packages/theme/src/cli/services/local-storage.ts
@@ -1,6 +1,7 @@
 import {BugError} from '@shopify/cli-kit/node/error'
 import {LocalStorage} from '@shopify/cli-kit/node/local-storage'
 import {outputDebug, outputContent} from '@shopify/cli-kit/node/output'
+import {AsyncLocalStorage} from 'node:async_hooks'
 
 type DevelopmentThemeId = string
 
@@ -15,6 +16,9 @@ interface DevelopmentThemeLocalStorageSchema {
 interface ThemeStorePasswordSchema {
   [themeStore: string]: string
 }
+
+/** Preserves the theme store a command is acting on during multi environment execution */
+const themeStoreContext = new AsyncLocalStorage<{store: string}>()
 
 let _themeLocalStorageInstance: LocalStorage<ThemeLocalStorageSchema> | undefined
 let _developmentThemeLocalStorageInstance: LocalStorage<DevelopmentThemeLocalStorageSchema> | undefined
@@ -56,7 +60,8 @@ function themeStorePasswordStorage() {
 }
 
 export function getThemeStore(storage: LocalStorage<ThemeLocalStorageSchema> = themeLocalStorage()) {
-  return storage.get('themeStore')
+  const context = themeStoreContext.getStore()
+  return context?.store ? context.store : storage.get('themeStore')
 }
 
 export function setThemeStore(store: string, storage: LocalStorage<ThemeLocalStorageSchema> = themeLocalStorage()) {
@@ -143,4 +148,9 @@ function assertThemeStoreExists(storage: LocalStorage<ThemeLocalStorageSchema> =
     )
   }
   return themeStore
+}
+
+/** Provides theme store context to each environment during concurrent multi environment execution */
+export async function useThemeStoreContext<T>(store: string, callback: () => Promise<T>): Promise<T> {
+  return themeStoreContext.run({store}, callback)
 }
--- a/packages/theme/src/cli/utilities/theme-command.ts
+++ b/packages/theme/src/cli/utilities/theme-command.ts
@@ -1,6 +1,7 @@
 import {ensureThemeStore} from './theme-store.js'
 import {configurationFileName} from '../constants.js'
 import metadata from '../metadata.js'
+import {useThemeStoreContext} from '../services/local-storage.js'
 import {hashString} from '@shopify/cli-kit/node/crypto'
 import {Input} from '@oclif/core/interfaces'
 import Command, {ArgOutput, FlagOutput} from '@shopify/cli-kit/node/base-command'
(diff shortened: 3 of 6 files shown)


packages/theme/src/cli/utilities/theme-command.ts
```diff
@@ -218,30 +215,57 @@ export default abstract class ThemeCommand extends Command {
...
+    }
+  }
+
+  /**
+   * Create groups of environments with unique flags.store values to run sequentially
+   * to prevent conflicts between environments acting on the same store
+   * @param environments - The environments to group
+   * @returns The environment groups
+   */
+  private createSequentialGroups(environments: {environment: EnvironmentName; flags: FlagValues}[]) {
```
dejmedus, line 271: I assume running a command in multiple environments with the same store URL would be pretty rare, but I could picture something like `shopify theme info -e development -e production` where allowing it would be helpful

packages/theme/src/cli/services/local-storage.ts
```diff
@@ -144,3 +149,8 @@ function assertThemeStoreExists(storage: LocalStorage<ThemeLocalStorageSchema> =
   }
   return themeStore
 }
+
+/** Provides theme store context to each environment during concurrent multi environment execution */
+export async function useThemeStoreContext<T>(store: string, callback: () => Promise<T>): Promise<T> {
```
dejmedus, line 154: This pattern is also being used in [ConcurrentOutput](https://github.com/Shopify/cli/blob/7e6a0979b7c702402d6798ed02dfe6d5bf90cb7b/packages/cli-kit/src/private/node/ui/components/ConcurrentOutput.tsx#L41-L50)

.changeset/many-ties-check.md
EvilGenius13, line 1: Do we need both changesets?

dejmedus: I'm not 100% sold on having either of the changesets tbh, since this is mostly set up for _future_ changes. Or maybe changing the wording could make them more useful? 🤔 

> ~~Allow commands run in multiple environments to act on the same store URL~~
> Run environments that act on the same store URL sequentially

packages/theme/src/cli/services/local-storage.ts
```diff
@@ -56,7 +60,8 @@ function themeStorePasswordStorage() {
 }
 
 export function getThemeStore(storage: LocalStorage<ThemeLocalStorageSchema> = themeLocalStorage()) {
-  return storage.get('themeStore')
+  const context = themeStoreContext.getStore()
```
aswamy, line 63: Doesn't this _always_ return `undefined` because we are running `getStore` outside the `run` context?

image

aswamy (commented):
Hey, this might be a silly question, but is multiple `--env` flags even truely supported today?

When i try to have two envs (each from a different store), it doesn't seem to respect theme IDs I've set in the toml.

image

I can get around it by passing in `--development` flag (so it publishes to the respective development environment for each store), but it only shows the success of one store:
image

packages/theme/src/cli/services/local-storage.ts
dejmedus: Yes, good question: currently only `list`, `info`, `rename` work with multiple environments. `delete` will also work since [its branch](https://github.com/Shopify/cli/pull/6193) was included in the snapbuild . Soon<sup>TM</sup> other commands like `push`, `pull`, etc will be changed over but this and a couple other logic changes need to go in first. My PR description example was confusing!

dejmedus: Since `push` hasn't been refactored to support multi env yet (`run` needs to be swapped for `command`) it doesn't benefit from this change. Should see the context be defined in `list`, `info`, and `rename`

In multi env runs we are [wrapping the command call in the context provider](https://github.com/Shopify/cli/pull/6257/files#:~:text=store%20as%20string-,await%20useThemeStoreContext(store%2C%20async%20()%20%3D%3E%20%7B,%7D),-//%20eslint%2Ddisable%2Dnext) so it has access to the store URL


context

aswamy: > In multi env runs we are [wrapping the command call in the context provider](https://github.com/Shopify/cli/pull/6257/files#:~:text=store%20as%20string-,await%20useThemeStoreContext(store%2C%20async%20()%20%3D%3E%20%7B,%7D),-//%20eslint%2Ddisable%2Dnext) so it has access to the store URL. If we add a console log to getThemeStore it shows as undefined in single/no env runs:

OOOOH that's perfect. Thank you for the explanation - this makes a lot of sense.

PR #6264, opened 2025-08-14
fixed an error related to patched SFR responses when running theme dev

EvilGenius13:

Fixes #0000 

### Measuring impact

### Checklist


EvilGenius13:
I would be happy to have anyone with experience in the proxy double checking my logic!

packages/theme/src/cli/utilities/theme-environment/proxy.ts
```diff
@@ -171,6 +171,11 @@ export async function patchRenderingResponse(
   rawResponse: Response,
   patchCallback?: (html: string) => string | undefined,
 ): Promise<Response> {
+  // 304 (Not Modified) responses should be returned
```
graygilmore: > I would be happy to have anyone with experience in the proxy double checking my logic!

This is not me but your explanation here makes sense

graygilmore (approved):

packages/theme/src/cli/utilities/theme-environment/proxy.ts
```diff
@@ -171,6 +171,11 @@ export async function patchRenderingResponse(
...
+  if (rawResponse.status === 304) {
```
frandiox: We shouldn't be patching **any** 3xx actually, so perhaps better to check `>= 300 && <400`

frandiox (approved):
Thanks, this makes sense. The `/web-pixels@` one should be ignored completely but this makes the server logic more robust.

packages/theme/src/cli/utilities/theme-environment/proxy.ts
EvilGenius13: Sure thing! I'll make the edit

PR #6265, opened 2025-08-14
Fix custom timeouts

EvilGenius13:

It turns out the custom network behaviour (longer timeouts and no aborting requests) wasn't actually being used in theme commands.

What happened?
- Request flows through the theme api and takes in our custom behaviour called `requestBehaviour`
- It gets passed to admin where it's functions use our custom behaviour
- It bubbles all the way into the GraphQL layer where it looks for custom behaviour called `prefferedBehaviour` <-- since ours it not named that, it defaulted to the regular behaviour which included aborting requests after 30 seconds.

- Rename most of the variables we pass down to be `preferredBehaviour` for consistency and so the network behaviour actually gets passed down
- Added `requestMode` to a few extra admin calls around fetching API's where I believe users may be failing (it still defaults to the regular behaviour unless we are using a theme related command)

### How to test your changes?

One easy way is to add a console.log in `graphql.ts` 
under this line `const behaviour = requestMode(options.preferredBehaviour ?? 'default')` and run a theme command like `theme push`. It should print out an object with this:
```js
behaviour {
  useNetworkLevelRetry: true,
  useAbortSignal: false,
  maxRetryTimeMs: 90000
}
```
Note we make a lot of graphQL calls so you may see this multiple times, the important thing is you should never see `useAbortSignal: true` or `maxRetryTimeMS: 30000`

### Measuring impact

### Checklist


packages/cli-kit/src/public/node/api/graphql.ts
```diff
@@ -65,21 +65,21 @@ type PerformGraphQLRequestOptions<TResult> = GraphQLRequestBaseOptions<TResult>
...
-  requestBehaviour?: RequestModeInput
+  preferredBehaviour?: RequestModeInput
 }
 
 export type GraphQLRequestDocOptions<TResult, TVariables> = GraphQLRequestBaseOptions<TResult> & {
   query: TypedDocumentNode<TResult, TVariables> | TypedDocumentNode<TResult, Exact<{[key: string]: never}>>
   variables?: TVariables
   unauthorizedHandler?: UnauthorizedHandler
-  requestBehaviour?: RequestModeInput
+  preferredBehaviour?: RequestModeInput
```
isaacroldan: These all extend `GraphQLRequestBaseOptions` which already has a ` preferredBehaviour?: RequestModeInput` in it.

I think `requestBehaviour` behaviour was unused and don't know why it was there

EvilGenius13: Thanks @isaacroldan. You are right. I have removed those.

aswamy (approved):
> It turns out the custom network behaviour (longer timeouts and no aborting requests) wasn't actually being used in theme commands.

image

PR #6281, opened 2025-08-19
Change how Webhook Subscription UID is generated

isaacroldan:

To improve the uniqueness and readability of webhook subscription extension UIDs by generating them based on their configuration details rather than using a generic UUID.

Modifies the UID generation strategy for webhook subscription extensions:

- When using the 'dynamic' strategy, webhook subscription extensions now generate UIDs in the format `topic::filter::uri`
- Adds a maximum UID length constraint of 250 characters
- Updates tests to verify the new UID format for webhook subscriptions both with and without filters

### How to test your changes?

1. Create a webhook subscription extension with a topic and URI
2. Verify that the generated UID follows the format `topic::filter::uri`
3. Create another webhook subscription with filters and confirm the filter value is included in the UID

### Measuring impact

### Checklist

- [x] I've considered possible [documentation](https://shopify.dev)changes

### DIFF
--- a/packages/app/src/cli/models/extensions/extension-instance.ts
+++ b/packages/app/src/cli/models/extensions/extension-instance.ts
@@ -1,6 +1,6 @@
 /* eslint-disable @typescript-eslint/no-non-null-assertion */
 
-import {BaseConfigType, MAX_EXTENSION_HANDLE_LENGTH} from './schemas.js'
+import {BaseConfigType, MAX_EXTENSION_HANDLE_LENGTH, MAX_UID_LENGTH} from './schemas.js'
 import {FunctionConfigType} from './specifications/function.js'
 import {ExtensionFeature, ExtensionSpecification} from './specification.js'
 import {SingleWebhookSubscriptionType} from './specifications/app_config_webhook_schemas/webhooks_schema.js'
@@ -508,7 +508,19 @@ export class ExtensionInstance<TConfiguration extends BaseConfigType = BaseConfi
       case 'uuid':
         return this.configuration.uid ?? nonRandomUUID(this.handle)
       case 'dynamic':
-        return nonRandomUUID(this.handle)
+        // NOTE: This is a temporary special case for webhook subscriptions.
+        // We're directly checking for webhook properties and casting the configuration
+        // instead of using a proper dynamic strategy implementation.
+        // To remove this special case:
+        // 1. Implement a proper dynamic UID strategy for webhooks in the server-side specification
+        // 2. Update the CLI to use that strategy instead of this hardcoded logic
+        // Related issues: PR #559094 in old Core repo
+        if ('topic' in this.configuration && 'uri' in this.configuration) {
+          const subscription = this.configuration as unknown as SingleWebhookSubscriptionType
+          return `${subscription.topic}::${subscription.filter}::${subscription.uri}`.substring(0, MAX_UID_LENGTH)
+        } else {
+          return nonRandomUUID(JSON.stringify(this.configuration))
+        }
     }
   }
 }
--- a/packages/app/src/cli/models/extensions/extension-instance.test.ts
+++ b/packages/app/src/cli/models/extensions/extension-instance.test.ts
@@ -531,12 +531,24 @@ describe('draftMessages', async () => {
       expect(extensionInstance.uid).toBe(nonRandomUUID(extensionInstance.handle))
     })
 
-    test('returns non-random UUID based on handle when strategy is dynamic', async () => {
+    test('returns a custom string when strategy is dynamic and it is a webhook subscription extension without filters', async () => {
       // Given
       const extensionInstance = await testSingleWebhookSubscriptionExtension()
-
       // Then
-      expect(extensionInstance.uid).toBe(nonRandomUUID(extensionInstance.handle))
+      expect(extensionInstance.uid).toBe('orders/delete::undefined::https://my-app.com/webhooks')
+    })
+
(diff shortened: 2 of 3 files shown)


isaacroldan:
* **#6282** <a href="https://app.graphite.dev/github/pr/Shopify/cli/6282?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6281** <a href="https://app.graphite.dev/github/pr/Shopify/cli/6281?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Shopify/cli/6281?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/app/src/cli/models/extensions/extension-instance.ts
```diff
@@ -508,7 +508,12 @@ export class ExtensionInstance<TConfiguration extends BaseConfigType = BaseConfi
       case 'uuid':
         return this.configuration.uid ?? nonRandomUUID(this.handle)
       case 'dynamic':
-        return nonRandomUUID(this.handle)
+        if ('topic' in this.configuration && 'uri' in this.configuration) {
+          const subscription = this.configuration as unknown as SingleWebhookSubscriptionType
```
MitchDickinson, line 519: For any new consistency violation we should consider adding documentation in the code so we know how we can remove it later
- what is required to remove it?
- links to relevant issues/designs/plans

pt2pham: I believe the initial thought for UidStrategies defined server side is that we can use them to pre-compute the UID in the CLI, so for any dynamic/custom ones, we should be deferring to the custom strategy as is defined by a certain specification

packages/app/src/cli/models/extensions/extension-instance.ts
```diff
@@ -508,7 +508,12 @@ export class ExtensionInstance<TConfiguration extends BaseConfigType = BaseConfi
...
+          return `${subscription.topic}::${subscription.filter}::${subscription.uri}`.substring(0, MAX_UID_LENGTH)
+        } else {
+          return nonRandomUUID(JSON.stringify(this.configuration))
```
pt2pham, line 522: This change to `this.configuration` instead of `this.handle`; was that intended?

packages/app/src/cli/models/extensions/extension-instance.ts
```diff
@@ -508,7 +508,12 @@ export class ExtensionInstance<TConfiguration extends BaseConfigType = BaseConfi
       case 'uuid':
         return this.configuration.uid ?? nonRandomUUID(this.handle)
       case 'dynamic':
-        return nonRandomUUID(this.handle)
+        if ('topic' in this.configuration && 'uri' in this.configuration) {
```
pt2pham, line 518: I am worried about other app modules having `topic` in their config but I suppose that landmine already exists today - we should look to remove this magic logic for webhooks as soon as we can

alfonso-noriega (approved):

packages/app/src/cli/models/extensions/extension-instance.ts
isaacroldan: both are valid, is a fallback if we ever have a specification with `dynamic` strategy that's not a webhook subscribtion, it should never happen

packages/app/src/cli/models/extensions/extension-instance.ts
isaacroldan: they'll also need to have `dynamic` as their uid strategy, which right now is a hardcoded value for webhook subscriptions. The sooner we remove this custom behavior the better :)

packages/app/src/cli/models/extensions/extension-instance.ts
```diff
@@ -508,7 +508,19 @@ export class ExtensionInstance<TConfiguration extends BaseConfigType = BaseConfi
...
+        // NOTE: This is a temporary special case for webhook subscriptions.
+        // We're directly checking for webhook properties and casting the configuration
+        // instead of using a proper dynamic strategy implementation.
+        // To remove this special case:
+        // 1. Implement a proper dynamic UID strategy for webhooks in the server-side specification
+        // 2. Update the CLI to use that strategy instead of this hardcoded logic
+        // Related issues: PR #559094 in old Core repo
+        if ('topic' in this.configuration && 'uri' in this.configuration) {
+          const subscription = this.configuration as unknown as SingleWebhookSubscriptionType
+          return `${subscription.topic}::${subscription.filter}::${subscription.uri}`.substring(0, MAX_UID_LENGTH)
```
gonzaloriestra, line 520: I think the trimming here is dangerous. URLs can be very long. Imagine you have two subscriptions like this:
- `orders/create::id:123::https://very-looooo...oooong-url/1`
- `orders/create::id:123::https://very-looooo...oooong-url/2`
Both would get the same UID: `orders/create::id:123::https://very-loooooooooooooooooo`, which would cause conflicts.

Maybe it's weird to have two subscriptions with the same topic and filter, but different URL. But IMHO we should look for a more solid solution at some point.

isaacroldan: There has to be "some" limit, 250 are a lot of characters for a URL so I hope that'd be enough. 
But hopefully we can remove this soon too, would you increment the limit?

gonzaloriestra: I'd try to match it to the database limit, if it's bigger. I agree it should be enough, but we are also including filters here, I don't know how long they could be...

PR #6287, opened 2025-08-20
Fix issue when trying to import an already imported extension

isaacroldan:
# Prevent import from trying to import already imported extensions.

This PR adds functionality to filter out extensions that have already been imported into the local environment, preventing crashes and improving the extension import process.

- Implements `filterOutImportedExtensions` function to check if extensions are already imported
- Updates the `importExtensionsIfNeeded` function to only import extensions that aren't already in the local environment
- Ensures the `.env` file is updated when importing extensions
- Adds tests for the new filtering functionality

### How to test your changes?

1. Create an app with Dashboard extensions
2. manually run `import-extensions`​
3. Try to import extensions again -> should see “no extensions to migrate”
4. Run deploy against dev-dash -> should ask to migrate, but not to import.

### Measuring impact

### Checklist

- [x] I've considered possible [documentation](https://shopify.dev)changes

### DIFF
--- a/packages/app/src/cli/models/app/identifiers.ts
+++ b/packages/app/src/cli/models/app/identifiers.ts
@@ -33,7 +33,7 @@ export interface Identifiers {
 }
 
 type UuidOnlyIdentifiers = Omit<Identifiers, 'extensionIds' | 'extensionsNonUuidManaged'>
-type UpdateAppIdentifiersCommand = 'dev' | 'deploy' | 'release'
+type UpdateAppIdentifiersCommand = 'dev' | 'deploy' | 'release' | 'import-extensions'
 interface UpdateAppIdentifiersOptions {
   app: AppInterface
   identifiers: UuidOnlyIdentifiers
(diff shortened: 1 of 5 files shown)


isaacroldan:
* **#6287** <a href="https://app.graphite.dev/github/pr/Shopify/cli/6287?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Shopify/cli/6287?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/app/src/cli/models/app/identifiers.ts
```diff
@@ -72,9 +72,11 @@ export async function updateAppIdentifiers(
 
   const contentIsEqual = deepCompare(dotenvFile.variables, updatedVariables)
   const writeToFile =
-    !contentIsEqual &&
-    (command === 'deploy' || command === 'release') &&
-    !developerPlatformClient.supportsAtomicDeployments
+    (!contentIsEqual &&
+      (command === 'deploy' || command === 'release') &&
+      !developerPlatformClient.supportsAtomicDeployments) ||
+    command === 'import-extensions'
```
isaacroldan, line 78: We need the import-extensions command to always update the UUID in the `.env` file so we have something to compare with. But deploy should never write to `.env` if using the new API

gonzaloriestra, line 78: But we only want to update the `.env` in Partners, right? According to this logic, we will also update it for App Management.

```suggestion
    !contentIsEqual &&
    command !== 'dev' &&
    !developerPlatformClient.supportsAtomicDeployments)
```

isaacroldan: We need to update it for app-management too if the command is `import-extension`, because if not we won't be able to:
- match the imported extension (because we won't have a cached UUID)
- in deploy detect the imported extension and fix the original issue

gonzaloriestra: Oh, of course. Lapse.

PR #6291, opened 2025-08-20
Refactor commands that don't support multiple environments to use ThemeCommand's run method

dejmedus:

Previously `language-server`, `dev`, `console`, `open`, `profile`, `package`, and `metafields pull` implemented their own run logic. This PR refactors them to use ThemeCommand's base `run()` method. This will help to standardize command architecture and allow for out of the box authentication and environments support. 

- Refactors commands to define their own `command()` rather than `run()` so they can instead use ThemeCommand's base `run()` method
- Sets `dev`, `console`, `open`, `profile` and `metafields pull` command's `multiEnvironmentsFlags` to null to display a warning if ran with multiple environment flags

 <details><summary>dev</summary>

dev-demo

</details> 
 <details><summary>metafields pull</summary>

metafields-pull

</details> 
<details><summary>open</summary>

open

</details> 
<details><summary>console</summary>

console

</details> 
<details><summary>invalid password</summary>

password-validation

</details> 

### How to test your changes?

- Pull down the branch `gh pr checkout 6291` or install the snapit version
```sh
npm i -g --@shopify:registry=https://registry.npmjs.org @shopify/cli@0.0.0-snapshot-20250912003555
```

```
[environments.store1]
theme = "<theme>"
store = "<store>"

[environments.store2]
theme = "<theme>"
store = "<store>"
```

</details> 

- `dev`, `console`, `open`, and `metafields pull` should work as expected when provided one or no environment flags and should display a warning if provided with multiple 
- `language-server` and `package` should work as expected, but do not support any environment flags
- `profile` should work as expected, but does not currently accept a password 

### Measuring impact

### Checklist

### DIFF
--- a/.changeset/slow-buckets-battle.md
+++ b/.changeset/slow-buckets-battle.md
@@ -1,0 +1,6 @@
+---
+'@shopify/theme': minor
+'@shopify/cli': minor
+---
+
+Display warning when multiple environment flags are provided to profile, metafields pull, open, dev, or console theme commands


packages/theme/src/cli/commands/theme/console.ts
```diff
@@ -33,15 +35,12 @@ export default class Console extends ThemeCommand {
...
-
-    validateThemePassword(flags.password)
+  static multiEnvironmentsFlags: RequiredFlags = null
 
-    const store = ensureThemeStore(flags)
+  async command(flags: ConsoleFlags, adminSession: AdminSession) {
     const {url, password: themeAccessPassword} = flags
 
-    const adminSession = await ensureAuthenticatedThemes(store, themeAccessPassword)
+    validateThemePassword(themeAccessPassword)
```
dejmedus, line 43: Some command passwords are [validated](https://github.com/Shopify/cli/commit/87bedd24b4f8f51d078f7309d3ff9705032220ad#diff-dd164b3e16c813dccbb28855814a9c8633d4ebd4655ac6ffcb28e8e25a6a6032). Now that the session is being created in `ThemeCommand` it reads a bit weird (validate password after session creation) but I decided to keep it this way for a couple reasons:

- If provided a password, authenticate themes only creates a session object, it forgoes the actual auth check to api call  (so we aren't going to try to auth with an invalid password)
- password validation could be moved to `ThemeCommand`, but we would need to manage a conditional of which commands require it
- if `validateThemePassword` is not provided a password it just returns and continues on

packages/theme/src/cli/commands/theme/dev.ts
```diff
@@ -158,7 +157,7 @@ You can run this command only in a directory that matches the [default Shopify t
     await dev({
       adminSession,
       directory: flags.path,
-      store,
+      store: flags.store,
```
graygilmore, line 160: To make sure: we can skip using `ensureThemeStore` here now because `ThemeCommand` will run `createSession` which does it instead?

graygilmore (approved):
👍🏻 🎩

packages/theme/src/cli/commands/theme/dev.ts
dejmedus: Yes, exactly! ThemeCommand is now handling flag parsing and session creation which then gets passed into each command

PR #6301, opened 2025-08-21
Ensure user provided flag values override defaults in multi env commands

dejmedus:

Previously, default flag values were treated by multi environment command mapping as CLI values, which take precedence over shopify.theme.toml values. This PR ensures that flag values are selected in order:

1. User provided CLI args
2. User provided shopify.theme.toml values
3. Default flag values

- Creates each environment's flags obj by combining: CLI flags with defaults, shopify.theme.toml flags, and CLI flags without defaults in that order
- Displays a warning if `--path` flag is provided to the CLI during multi env runs (path defaults to `cwd` and should be set per environment in `shopify.theme.toml`)

<details><summary>shopify.theme.toml is respected over default values</summary>
 
https://github.com/user-attachments/assets/357dc2c5-59f8-48f4-bcbc-4b76eb684603
 
</details> 

<details><summary>path warning</summary>
 
path-errors

</details> 

### How to test your changes?

- Add a `shopify.theme.toml` file with a value that has a default like `json = true`
 
 > Heads up: globals like `no-color` and `verbose` do not seem to be respected if they come from `shopify.theme.toml` single or multi env. I'll look into this in a follow up PR

```
[environments.store1]
theme = "<theme>"
store = "<store>"
json = true

[environments.store2]
theme = "<theme>"
store = "<store>"
```
</details> 

- Pull down the branch `gh pr checkout 6301` or install the snapit version
```sh
npm i -g @shopify/cli@0.0.0-snapshot-20250822001254 ---@shopify:registry=https://registry.npmjs.org
```
- Run `list` and you should see that the environments with `json = true` defined in the .toml print out json and other environments don't
- Using`--path` flag with multi env commands should display an error

```ts
shopify theme list -e store1 -e store2
```

### Measuring impact

### Checklist

### DIFF
--- a/packages/theme/src/cli/utilities/theme-command.ts
+++ b/packages/theme/src/cli/utilities/theme-command.ts
@@ -3,7 +3,7 @@ import {configurationFileName} from '../constants.js'
 import {useThemeStoreContext} from '../services/local-storage.js'
 import {hashString} from '@shopify/cli-kit/node/crypto'
 import {Input} from '@oclif/core/interfaces'
-import Command, {ArgOutput, FlagOutput} from '@shopify/cli-kit/node/base-command'
+import Command, {ArgOutput, FlagOutput, noDefaultsOptions} from '@shopify/cli-kit/node/base-command'
 import {AdminSession, ensureAuthenticatedThemes} from '@shopify/cli-kit/node/session'
 import {loadEnvironment} from '@shopify/cli-kit/node/environments'
 import {
@@ -103,7 +103,7 @@ export default abstract class ThemeCommand extends Command {
       return
     }
 
-    const environmentsMap = await this.loadEnvironments(environments, flags)
+    const environmentsMap = await this.loadEnvironments(environments, flags, klass)
     const validationResults = await this.validateEnvironments(environmentsMap, requiredFlags)
 
     const commandAllowsForceFlag = 'force' in klass.flags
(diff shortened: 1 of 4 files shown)


packages/theme/src/cli/utilities/theme-command.ts
```diff
@@ -102,7 +102,15 @@ export default abstract class ThemeCommand extends Command {
       return
     }
 
-    const environmentsMap = await this.loadEnvironments(environments, flags, klass)
+    const {flags: flagsWithoutDefaults} = await this.parse(noDefaultsOptions(klass), this.argv)
+    if ('path' in flagsWithoutDefaults) {
+      renderWarning({
+        body: "Provide each environment's command path in your shopify.theme.toml file.",
```
graygilmore: Thinking out loud: what if I'm in `~` and I run `shopify theme command -e 1 -e2 --path ~/path/to/shopify.theme.toml`? I think that's probably the most common use case of including `--path` so we could potentially change the language of the warning maybe? Some pseudo code here:

```ts
if ('path' in flagsWithoutDefaults) {
  if (flagsWithoutDefaults.path === cwd()) {
    renderError({body: "Something about needing to run it in the same directory as the toml"});
    return;
  }

  renderError({body: "Cannot use the `--path` flag with multiple environments. Provide each environment's command path in your shopify.theme.toml file."});
  return
}
```

graygilmore: And maybe use `renderError` since we can't continue?

dejmedus: That makes sense, thanks! I updated the message

packages/theme/src/cli/utilities/theme-command.ts
```diff
@@ -102,7 +104,13 @@ export default abstract class ThemeCommand extends Command {
       return
     }
 
-    const environmentsMap = await this.loadEnvironments(environments, flags)
+    const {flags: flagsWithoutDefaults} = await this.parse(noDefaultsOptions(klass), this.argv)
+    if ('path' in flagsWithoutDefaults) {
+      this.errorOnGlobalPath()
+      return
+    }
```
graygilmore, line 112: Let's make sure we add a couple of tests for the two error scenarios

graygilmore (commented):
🎩 went great! Just need a small amount of test coverage and we'll be good to go!

EvilGenius13 (approved):
🎩 went well. Thanks for the changes!

gonzaloriestra (approved):

PR #6312, opened 2025-08-22
Only create sessions for multi env commands that require authentication

dejmedus:
- Blocked by https://github.com/Shopify/cli/pull/6301 

Previously, all commands run with multiple environments attempted to create a session object with store URL and password. This PR ensures that commands that do no accept a password and don't need to be authenticated (like `check`, `init`, `package`, and `language-server`) will not try to create a session

- Makes session creation conditional on whether a command accepts a password
- Adds `requiresAuth` boolean to valid environment type and DRYs it up by creating an `ValidEnvironment` interface

### How to test your changes?

- Install the snap build that contains these changes as well as `check` multi env logic

```
npm i -g @shopify/cli@0.0.0-snapshot-20250827192106 --@shopify:registry=https://registry.npmjs.org
```

```
[environments.store1]
theme = "<theme>"
store = "<store1>"

[environments.store2]
theme = "<theme>"
store = "<store2>"
```

</details> 

- Running `check` (from the above snapbuild) should work as expected, and not error if no password is provided
- Currently only `list`, `info`, `rename`, and `delete` work with multiple environments, these should error when provided no password or an incorrect

```sh
shopify theme check -e store1 -e store2
```

### Measuring impact

### Checklist

### DIFF
--- a/.changeset/large-otters-dream.md
+++ b/.changeset/large-otters-dream.md
@@ -1,0 +1,6 @@
+---
+'@shopify/cli-kit': minor
+'@shopify/theme': minor
+---
+
+Commands that don't require authentication should not create sessions when ran with multiple environments
--- a/packages/theme/src/cli/utilities/theme-command.ts
+++ b/packages/theme/src/cli/utilities/theme-command.ts
@@ -67,7 +67,7 @@ export default abstract class ThemeCommand extends Command {
 
   async command(
     _flags: FlagValues,
-    _session: AdminSession,
+    _session?: AdminSession,
     _multiEnvironment = false,
     _context?: {stdout?: Writable; stderr?: Writable},
   ): Promise<void> {}
@@ -84,18 +84,17 @@ export default abstract class ThemeCommand extends Command {
     }
     const requiredFlags = klass.multiEnvironmentsFlags
     const {flags} = await this.parse(klass)
-
+    const commandRequiresAuth = 'password' in klass.flags
     const environments = (Array.isArray(flags.environment) ? flags.environment : [flags.environment]).filter(Boolean)
 
     // Single environment or no environment
     if (environments.length <= 1) {
-      const session = await this.createSession(flags)
+      const session = commandRequiresAuth ? await this.createSession(flags) : undefined
       const commandName = this.constructor.name.toLowerCase()
 
       recordEvent(`theme-command:${commandName}:single-env:authenticated`)
 
       await this.command(flags, session)
-      await this.logAnalyticsData(session)
       return
     }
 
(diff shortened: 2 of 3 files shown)


.changeset/large-otters-dream.md
```diff
@@ -0,0 +1,6 @@
+---
+'@shopify/cli-kit': minor
+'@shopify/theme': minor
+---
+
+Commands that don't require authentication should not create session's when ran with multiple environments
```
graygilmore:
```suggestion
Commands that don't require authentication should not create sessions when ran with multiple environments
```

packages/theme/src/cli/utilities/theme-command.ts
```diff
@@ -84,18 +89,17 @@ export default abstract class ThemeCommand extends Command {
...
     const requiredFlags = klass.multiEnvironmentsFlags
     const {flags} = await this.parse(klass)
-
+    const commandRequiresAuth = 'password' in klass.flags
     const environments = (Array.isArray(flags.environment) ? flags.environment : [flags.environment]).filter(Boolean)
 
     // Single environment or no environment
     if (environments.length <= 1) {
-      const session = await this.createSession(flags)
+      const session = await this.createSession(flags, commandRequiresAuth)
```
graygilmore: Instead of passing a new argument to `createSession` just to return early what do you think about doing it here:

```suggestion
      const session = commandRequiresAuth ? undefined : await this.createSession(flags)
```

packages/theme/src/cli/utilities/theme-command.ts
```diff
@@ -245,13 +250,13 @@ export default abstract class ThemeCommand extends Command {
...
       await renderConcurrent({
-        processes: runGroup.map(({environment, flags}) => ({
+        processes: runGroup.map(({environment, flags, requiresAuth}) => ({
           prefix: environment,
           action: async (stdout: Writable, stderr: Writable, _signal) => {
             try {
               const store = flags.store as string
               await useThemeStoreContext(store, async () => {
-                const session = await this.createSession(flags)
+                const session = await this.createSession(flags, requiresAuth)
```
graygilmore: Ah! Answering my own question here: I'm guessing it's because we call this twice? I almost think I'd still prefer to do:
```suggestion
                const session = requiresAuth ? undefined : await this.createSession(flags)
```
...even though we now need to do it in two places. This is more of a nit than anything else but let me know what you think!

EvilGenius13 (approved):
Everything went well on testing.

.changeset/large-otters-dream.md
dejmedus: Whoops, thank you!!

packages/theme/src/cli/utilities/theme-command.ts
dejmedus: Yes, initially I stuck it in `createSession` to keep it DRY, but I think you're right! At a glance it'll be easier to understand we may not have a session with explicit `: undefined`

graygilmore (approved):

PR #6313, opened 2025-08-24
Allow theme check command to be called with multiple environments

dejmedus:

Previously `theme check` command could only be run with a single `--environment` flag. This PR allows `check` to be run with multiple environments in a single command

- Removes flag parsing from `check` in favour of ThemeCommand's `run` logic
- Prefixes output with the environment name, when present
- Prevents process exit on failure in order to allow other environments to complete
- Updates validation to ensure multi environment flag requirements are met by user provided CLI or toml values, not defaults

<details><summary>Check on multiple themes</summary>

https://github.com/user-attachments/assets/e84c5fa3-b0a9-479d-911c-dacf2dd0835a

</details> 

<details><summary>jq output</summary>

jq-output

</details> 

### How to test your changes?

- Pull down the branch `gh pr checkout 6313` or install the snapit version
```sh
npm i -g --@shopify:registry=https://registry.npmjs.org @shopify/cli@0.0.0-snapshot-20250909173916
```

```
[environments.store1]
theme = "<theme>"
store = "<store>"
path = "path/to/theme"

[environments.store2]
theme = "<theme>"
store = "<store>"
path = "path/to/theme"
```

</details> 

> Might need to increase `Terminal › Integrated: Scrollback` if running the commands in VSCode terminal

```ts
shopify theme check -e store1 -e store2
shopify theme check -e store1 -e store2 --output=json
```
- `check` should work as expected with no, single, and multiple `--environment` flags
- The environment name should be displayed on output
- If a path is not provided in `shopify.theme.toml`, the environment should not be run

### Measuring impact

### Checklist

### DIFF
--- a/packages/theme/src/cli/utilities/theme-command.ts
+++ b/packages/theme/src/cli/utilities/theme-command.ts
@@ -137,12 +137,8 @@ export default abstract class ThemeCommand extends Command {
    * @param flagsWithoutDefaults - Flags provided via the CLI
    * @returns The map of environments
    */
-  private async loadEnvironments(
-    environments: EnvironmentName[],
-    flags: FlagValues,
-    flagsWithoutDefaults: FlagValues,
-  ): Promise<Map<EnvironmentName, FlagValues>> {
-    const environmentMap = new Map<EnvironmentName, FlagValues>()
+  private async loadEnvironments(environments: EnvironmentName[], flags: FlagValues, flagsWithoutDefaults: FlagValues) {
+    const environmentMap = new Map<EnvironmentName, {flags: FlagValues; validationFlags: FlagValues}>()
 
     for (const environmentName of environments) {
       // eslint-disable-next-line no-await-in-loop
@@ -157,10 +153,13 @@ export default abstract class ThemeCommand extends Command {
       }
 
       environmentMap.set(environmentName, {
-        ...flags,
-        ...environmentFlags,
-        ...flagsWithoutDefaults,
-        environment: [environmentName],
+        flags: {
+          ...flags,
+          ...environmentFlags,
+          ...flagsWithoutDefaults,
+          environment: [environmentName],
+        },
+        validationFlags: {...environmentFlags, ...flagsWithoutDefaults} as FlagValues,
       })
     }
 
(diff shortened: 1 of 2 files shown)


packages/theme/src/cli/commands/theme/check.ts
```diff
@@ -86,11 +89,12 @@ export default class Check extends ThemeCommand {
     environment: themeFlags.environment,
   }
 
-  async run(): Promise<void> {
-    const {flags} = await this.parse(Check)
+  static multiEnvironmentsFlags = ['path']
```
graygilmore: Something might not be working here because I didn't get any warnings when running this and not including `path` in my config file.

dejmedus: Thank you! Default values were being allowed to meet flag requirements. I've updated the required flags validation to only look at user provided values. If more commands (thinking `profile` and its `url` flag) are changed over in the future, any flag that's both needed for multi env and has a default value can just be omitted from `multiEnvironmentsFlags`

packages/theme/src/cli/utilities/theme-command.ts
```diff
@@ -157,10 +153,13 @@ export default abstract class ThemeCommand extends Command {
       }
```
dejmedus, line 153: A check to resolve `path` values was added to [this PR](https://github.com/Shopify/cli/pull/6361/files#diff-5d3425a5a7069d7e5072b07c9a55347067bacd01536cc3ea6e9d101e95893262R139)

packages/theme/src/cli/utilities/theme-command.ts
graygilmore, line 1: 🧵 I'm not sure if this is a fault of this PR or not but if I specify the output as json  I think the warnings should be JSON, too?

```
➜  theme git:(main) shopify theme check -e has-path -e missing-path --output json
╭─ warning ──────────────────────────────────────────────────────────────────────────────────╮
│                                                                                            │
│  Missing required flags in environment configuration for missing-path:                     │
│    • path                                                                                  │
│                                                                                            │
╰────────────────────────────────────────────────────────────────────────────────────────────╯

[{"environment":"has-path","path":"/Users/graygilmore/repos/theme/sections/header.liqui
```

graygilmore: Which might actually open up more questions about multi env and JSON output? Or just environments in general?

charlespwd: It's fine to have the message if the message is sent to STDERR.

charlespwd: The only real concern here is that people expect to use `--json` in scripts. So if you pipe the output of a `--output json` command to another tool that accepts JSON input (e.g. `jq`), then that other tool shouldn't freak out.

IIRC we already made it so most logging goes to STDERR by default.

dejmedus: Sounds good, thanks! I took a look and `jq` seems to be okay with the multi env output:

 
<details><summary>output</summary>

Screenshot 2025-09-12 at 5 18 57 PM

</details> 
<details><summary>output with an error message</summary>

check-jq-error

</details>

PR #6314, opened 2025-08-25
Improve `shopify theme dev` to prevent accidental updates to live themes

karreiro:

Fixes https://github.com/Shopify/developer-tools-team/issues/741

This PR updates `shopify theme dev` to ask for confirmation before running the command with a live theme.

### How to test your changes?

- Run `shopify theme dev -t <live_theme_id>`
- Notice the CLI asks for confirmation now

**Before**:
image

**After**:
image

### Post-release steps

N/A

### Measuring impact

### Checklist


packages/theme/src/cli/utilities/theme-ui.ts
```diff
@@ -40,6 +41,26 @@ export async function ensureDirectoryConfirmed(
   return confirm
 }
 
+export async function ensureLiveThemeConfirmed(theme: Theme, action: string) {
+  if (theme.role !== LIVE_THEME_ROLE || !process.stdout.isTTY) {
+    return true
+  }
+
+  const message =
+    `You're about to ${action} on your live theme "${theme.name}". ` +
```
EvilGenius13, line 50:
```suggestion
      `You are about to ${action} your live theme (“${theme.name}”). ` +
      `Changes will be immediately visible to customers. Are you sure you want to proceed?`
```
This is a non-blocking and personal preference for a change. It mirrors the language we use in push and dev commands a bit more.

EvilGenius13 (approved):
Thank you for the extra guard @karreiro!

packages/theme/src/cli/utilities/theme-ui.ts
graygilmore: Small note: our style guide recommends that we use natural language so `You're` is correct in this case!

EvilGenius13: > Small note: our style guide recommends that we use natural language so `You're` is correct in this case!
I retract the first line of my suggestion!

packages/theme/src/cli/utilities/theme-ui.ts
```diff
@@ -40,6 +41,26 @@ export async function ensureDirectoryConfirmed(
   return confirm
 }
 
+export async function ensureLiveThemeConfirmed(theme: Theme, action: string) {
```
aswamy, line 44: I was wondering why we didn't just use `confirmPushToTheme` from `theme push`, but they are different enough that this is fine.

aswamy (approved):

packages/theme/src/cli/utilities/theme-ui.ts
karreiro: Good catches! Thank you! I will keep the "**You're**" than :)

packages/theme/src/cli/utilities/theme-ui.ts
karreiro: Agreed 🙂‍↕️ 

(edited -- the emoji conversion feels so weird)
image

PR #6322, opened 2025-08-26
Adding generate fixtures command

saga-dasgupta:
Vault project: https://vault.shopify.io/gsd/projects/46878-Shopify-Functions-Wasm-Level-Testing

Fixes https://github.com/shop/issues-shopifyvm/issues/538
We want to provide developers with a way to perform end to end language agnostic way of testing their function so they can avoid runtime failures.

- Providing a command to generate test fixtures from function run logs. 
- Sets up the tests directory structure with a default test file that runs the function with fixture's input and compares the function's output to the fixture's output. 
- Extracting some common code shared between replay and generate-fixtures command to the commons. 

### How to test your changes?

1. `dev cd cli`
2. `git checkout -b sd.generate_function_tes_fixtures`
3. `dev up`
4. `pnpm clean && pnpm install`
5. Create a function.
6. Run the function on your online store in app dev mode to generate logs that would be used in test fixture generation `shopify app dev`
7. Now generate tests by running `pnpm run shopify app function generate fixture --path <path to your app from the cli folder>` and follow the prompts.
8. You should see a `tests` directory created for the function with the fixture. 

**NOTE**
This PR requires a package we are in the process of creating for test helpers. Will merge after I have integrated that in.

### Post-release steps

Will need to update the Shopify documentation:

- Test and Debug Functions: [This existing section](https://shopify.dev/docs/apps/build/functions/test-debug-functions) should be updated.
- Shopify CLI App Commands: Usage information should be included in the list of commands for Shopify apps found [here](https://shopify.dev/docs/api/shopify-cli/app).

### Measuring impact

### Checklist

### DIFF
--- a/packages/app/src/cli/commands/app/function/generate-fixture.ts
+++ b/packages/app/src/cli/commands/app/function/generate-fixture.ts
@@ -11,7 +11,7 @@ export default class FunctionGenerateFixture extends AppLinkedCommand {
   static summary = 'Generates test files from a function run log.'
   static hidden = true
 
-  static descriptionWithMarkdown = `Prompts users to select a function run log and generates test fixtures from it. These test cases based on real function executions act as an E2E test suite for the function.`
+  static descriptionWithMarkdown = `Prompts users to select a function run log and generates test fixtures from it. These test cases based on real function executions allows users to test the behaviour of their function based on its input and output.`
 
   static description = this.descriptionWithoutMarkdown()
 
@@ -19,12 +19,6 @@ export default class FunctionGenerateFixture extends AppLinkedCommand {
     ...globalFlags,
     ...appFlags,
     ...functionFlags,
-    'api-key': Flags.string({
-      hidden: true,
-      description: "Application's API key",
-      env: 'SHOPIFY_FLAG_API_KEY',
-      exclusive: ['config'],
-    }),
     log: Flags.string({
       char: 'l',
       description:
@@ -35,13 +29,10 @@ export default class FunctionGenerateFixture extends AppLinkedCommand {
 
   public async run(): Promise<AppLinkedCommandOutput> {
     const {flags} = await this.parse(FunctionGenerateFixture)
-    if (flags['api-key']) {
-      await showApiKeyDeprecationWarning()
-    }
 
     const {app} = await linkedAppContext({
       directory: flags.path,
-      clientId: flags['client-id'] ?? flags['api-key'],
+      clientId: flags['client-id'],
       forceRelink: flags.reset,
       userProvidedConfigName: flags.config,
     })
--- a/packages/app/src/cli/commands/app/function/generate-fixture.ts
+++ b/packages/app/src/cli/commands/app/function/generate-fixture.ts
@@ -1,0 +1,60 @@
+import {chooseFunction, functionFlags} from '../../../services/function/common.js'
+import {generateFixture} from '../../../services/function/generate-fixture.js'
+import {appFlags} from '../../../flags.js'
+import {showApiKeyDeprecationWarning} from '../../../prompts/deprecation-warnings.js'
+import AppLinkedCommand, {AppLinkedCommandOutput} from '../../../utilities/app-linked-command.js'
+import {linkedAppContext} from '../../../services/app-context.js'
+import {globalFlags} from '@shopify/cli-kit/node/cli'
+import {Flags} from '@oclif/core'
+
+export default class FunctionGenerateFixture extends AppLinkedCommand {
+  static summary = 'Generates test files from a function run log.'
+  static hidden = true
+
+  static descriptionWithMarkdown = `Prompts users to select a function run log and generates test fixtures from it. These test cases based on real function executions act as an E2E test suite for the function.`
+
+  static description = this.descriptionWithoutMarkdown()
+
+  static flags = {
+    ...globalFlags,
+    ...appFlags,
+    ...functionFlags,
+    'api-key': Flags.string({
+      hidden: true,
+      description: "Application's API key",
+      env: 'SHOPIFY_FLAG_API_KEY',
+      exclusive: ['config'],
+    }),
+    log: Flags.string({
+      char: 'l',
+      description:
+        'Specifies a log identifier to generate test files from instead of selecting from a list. The identifier is provided in the output of `shopify app dev` and is the suffix of the log file name.',
+      env: 'SHOPIFY_FLAG_LOG',
+    }),
+  }
+
+  public async run(): Promise<AppLinkedCommandOutput> {
+    const {flags} = await this.parse(FunctionGenerateFixture)
+    if (flags['api-key']) {
+      await showApiKeyDeprecationWarning()
+    }
+
+    const {app} = await linkedAppContext({
+      directory: flags.path,
+      clientId: flags['client-id'] ?? flags['api-key'],
+      forceRelink: flags.reset,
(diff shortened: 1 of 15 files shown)


packages/app/src/cli/commands/app/function/generate-fixture.ts
```diff
@@ -0,0 +1,60 @@
...
+import {appFlags} from '../../../flags.js'
+import {showApiKeyDeprecationWarning} from '../../../prompts/deprecation-warnings.js'
+import AppLinkedCommand, {AppLinkedCommandOutput} from '../../../utilities/app-linked-command.js'
+import {linkedAppContext} from '../../../services/app-context.js'
+import {globalFlags} from '@shopify/cli-kit/node/cli'
+import {Flags} from '@oclif/core'
+
+export default class FunctionGenerateFixture extends AppLinkedCommand {
+  static summary = 'Generates test files from a function run log.'
+  static hidden = true
```
saga-dasgupta, line 12: Will keep it hidden till we also have the test command PR out. Then we can show both of them together.

packages/app/templates/function/package.json
```diff
@@ -0,0 +1,13 @@
...
+    "name": "function-tests",
+    "version": "1.0.0",
+    "description": "Test configuration for Shopify function",
+    "type": "module",
+    "scripts": {
+      "test": "vitest run"
+    },
+    "devDependencies": {
+      "vitest": "^3.0.0",
+      "graphql": "^16.8.0"
```
saga-dasgupta, line 11: I will add the package for the function test helpers here as well once that is ready.

packages/app/src/cli/commands/app/function/generate-fixture.ts
```diff
@@ -0,0 +1,60 @@
...
+
+  static descriptionWithMarkdown = `Prompts users to select a function run log and generates test fixtures from it. These test cases based on real function executions act as an E2E test suite for the function.`
```
andrewhassan: Nit: I don't think we should say "E2E test suite" since that sounds like it goes through the whole flow including Checkout. Maybe we could say something like: These test cases based on real function executions allows users to test the behaviour of their function based on its input and output.

packages/app/src/cli/commands/app/function/generate-fixture.ts
```diff
@@ -0,0 +1,60 @@
...
+        'Specifies a log identifier to generate test files from instead of selecting from a list. The identifier is provided in the output of `shopify app dev` and is the suffix of the log file name.',
+      env: 'SHOPIFY_FLAG_LOG',
+    }),
+  }
+
+  public async run(): Promise<AppLinkedCommandOutput> {
+    const {flags} = await this.parse(FunctionGenerateFixture)
+    if (flags['api-key']) {
+      await showApiKeyDeprecationWarning()
+    }
```
andrewhassan: Since this is a net new command, we should double check with the CLI team whether we even want to support `api-key` here. Maybe we can just allow `client-id`.

packages/app/src/cli/services/function/generate-fixture.ts
```diff
@@ -0,0 +1,170 @@
...
+  const functionRunsDir = joinPath(getLogsDir(), apiKey)
+
+  const selectedRun = options.log
+    ? await getRunFromIdentifier(functionRunsDir, extension.handle, options.log)
+    : await getRunFromSelector(functionRunsDir, extension.handle)
+
+  // Check if we actually got a valid run with data
+  if (!selectedRun || !selectedRun.payload) {
+    throw new AbortError(
+      `No function run logs found for function '${extension.handle}'.\n` +
```
andrewhassan: Nit: if `log` is passed in and we can't find the run, should the error message be a little different?

packages/app/src/cli/services/function/generate-fixture.ts
```diff
@@ -0,0 +1,170 @@
...
+    await mkdir(testFixturesDir)
+  }
+
+  // Create default test file only if no test files already exist
+  const testFile = joinPath(testsDir, `default.test.ts`)
+  const existingTestFiles = readdirSync(testsDir).filter(
+    (file) => file.endsWith('.test.ts') || file.endsWith('.test.js'),
+  )
+
+  if (existingTestFiles.length === 0 && !existsSync(testFile)) {
```
andrewhassan: If there are no existing tests, then I think the file existence check is redundant, right?

packages/app/src/cli/services/function/generate-fixture.ts
```diff
@@ -0,0 +1,170 @@
...
+  const packageJsonPath = joinPath(testsDir, 'package.json')
+  if (!existsSync(packageJsonPath)) {
+    const packageJsonTemplatePath = joinPath(cwd(), 'packages/app/templates/function/package.json')
+    const packageJsonContent = await readFile(packageJsonTemplatePath)
+    await writeFile(packageJsonPath, packageJsonContent)
+
+    // Install dependencies after creating package.json
+    const {exec} = await import('child_process')
+    const {promisify} = await import('util')
+    const execAsync = promisify(exec)
```
andrewhassan: Why do we want to dynamically import these here vs. just importing them at the top of this file?

packages/app/src/cli/services/function/generate-fixture.ts
```diff
@@ -0,0 +1,170 @@
...
+    inputBytes: 0,
+    outputBytes: 0,
+    functionId: '',
+    logs: '',
+    fuelConsumed: 0,
+  }
+
+  const {input, output} = payload
+
+  // Parse TOML to find the targeting section that matches the export
```
andrewhassan: Nit: should we move this into a separate function in this file to make it easier to read?

packages/app/src/cli/services/function/generate-fixture.ts
```diff
@@ -0,0 +1,170 @@
...
+  let inputQueryPath = ''
+  let target = ''
+  const tomlPath = joinPath(extension.directory, 'shopify.extension.toml')
```
andrewhassan: Instead of parsing out the TOML file here, is it possible to use `extension.configuration`?

packages/app/src/cli/commands/app/function/generate-fixture.ts
saga-dasgupta: @isaacroldan thoughts?

packages/app/src/cli/services/function/generate-fixture.ts
saga-dasgupta: Yes

saga-dasgupta: Actually I dont think this is needed at all, both the methods handle the case where no log is found with an appropriate error message already.

packages/app/src/cli/services/function/generate-fixture.ts
saga-dasgupta: Yes, true

packages/app/src/cli/commands/app/function/generate-fixture.ts
isaacroldan: yeah, no need to support `api-key`, you can remove it from the flags

packages/app/src/cli/services/function/generate-fixture.ts
saga-dasgupta: Yes, that works.

lopert:
Closing this off as we're going with a different approach.

PR #6323, opened 2025-08-26
Fix monorail reporting for theme events

EvilGenius13:

When we added custom analytics for theme events, they weren't being sent to monorail. This is because we were creating a separate container for metadata that wasn't being added to the main analytics payload.

Shop domains are also never collected for theme commands which needed to be addressed.

- Extend the global metadata container types to support the `cmd_theme_*` fields and `store_fqdn`
- Remove the theme metadata container we created

### How to test your changes?

I tested this locally and can confirm we are seeing the data in our system.

### Measuring impact

### Checklist


karreiro:
👋 @Shopify/app-inner-loop could you please take a look when you have a moment?

gonzaloriestra (approved):

PR #6333, opened 2025-08-28
Fix configName flag handling in config:link

plvaldes:

The `config:link` command is ignoring the `-c` flag.

This PR removes the field `baseConfigName` from `LinkOptions`, and replaces its uses in the code with `configName`. `baseConfigName` was never set, and it was used to get the `-c` or `--config` flag value. 

  However, this change is not enough. When a `-c` value was provided to `config:link`, it wasn't used. Instead, the `loadConfigurationFileName` function uses the cached value from `getCachedCommandInfo`, which seems to be the TOML file from the selected app, and not from the flag option. 

selectedToml_from_app

So I added a second code update to not call `loadConfigurationFileName` if the user provides a config file name in the `-c` flag, to directly get it from the flag.

### How to test your changes?

1. I added a unit test where the service is called with a `configName` different than the default (used in other tests)
2. Run `pnpm shopify app config link` with a config name that is not the app TOML's file that you will select later. Verify that the TOML file passed as a flag is linked to the selected app.

config_link_fix

### Measuring impact

### Checklist

### DIFF
--- a/packages/app/src/cli/services/app/config/link.test.ts
+++ b/packages/app/src/cli/services/app/config/link.test.ts
@@ -1697,6 +1697,85 @@ embedded = false
     })
   })
 
+  test('existing config is respected when isNewApp is true, config is current and client_id is not the same as remote app and config name is provided', async () => {
+    await inTemporaryDirectory(async (tmp) => {
+      // Given
+      const developerPlatformClient = buildDeveloperPlatformClient()
+      const options: LinkOptions = {
+        directory: tmp,
+        developerPlatformClient,
+        isNewApp: true,
+        configName: 'staging',
+      }
+      const localApp = {
+        configuration: {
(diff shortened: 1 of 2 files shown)


plvaldes:
* **#6333** <a href="https://app.graphite.dev/github/pr/Shopify/cli/6333?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Shopify/cli/6333?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

dmerand (approved):
Nice find + test. I can't find any more references to `baseConfigName` 👍🏼 . It makes sense to use provided options over the cached options when available.

Tophat LGTM!

PR #6347, opened 2025-09-02
Fixed store names not parsing correctly in multi-env theme commands

EvilGenius13:

Follow up from https://github.com/Shopify/cli/pull/5874#issuecomment-3238374866

When parsing `shopify.theme.toml`'s in multi environment mode we skip normalizing stores that happens [here](https://github.com/Shopify/cli/blob/3a2f619ddd070597ce4b944b58a4c9c9f2d5b8cc/packages/theme/src/cli/flags.ts#L27)

It should take a store like `mystore` and append `myshopify.com` to it. 

Add `normalizeStoreFqdn` to the `loadEnvironments` method in the theme command layer. It will append the right domain to the store.

### How to test your changes?
_Replicate the bug_
In the current version of the CLI (4.84.1) or on main
- Create or update your `shopify.theme.toml` by only referencing the store name
```bash
[environments.store1]
store = "first-store.myshopify.com" 
password = "shptka_*"
theme = "123456"
[environments.store2]
store = "second-store" <-- Remove "myshopify.com"
password = "shptka_*"
theme = "123456"
```
Run a command that supports multi-env like `shopify theme list -e store1 -e store2`. You should see an error with the store that doesn't have `myshopify.com`

_Test with the bug fix_
- Pull down the branch `multi-env-store-toml`
- Build the branch
- Run `shopify theme list -e store1 -e store2` with the same `.toml` file.
### Measuring impact

### Checklist

### DIFF
--- a/.changeset/pretty-clocks-develop.md
+++ b/.changeset/pretty-clocks-develop.md
@@ -1,0 +1,5 @@
+---
+'@shopify/theme': patch
+---
+
+Fixed issue with theme.toml files that have a store name without the full domain from being parsed in multi-environment mode
--- a/packages/theme/src/cli/utilities/theme-command.ts
+++ b/packages/theme/src/cli/utilities/theme-command.ts
@@ -18,6 +18,7 @@ import {recordEvent, compileData} from '@shopify/cli-kit/node/analytics'
 import {addPublicMetadata, addSensitiveMetadata} from '@shopify/cli-kit/node/metadata'
 import {cwd, joinPath} from '@shopify/cli-kit/node/path'
 import {fileExistsSync} from '@shopify/cli-kit/node/fs'
+import {normalizeStoreFqdn} from '@shopify/cli-kit/node/context/fqdn'
 import type {Writable} from 'stream'
 
 export interface FlagValues {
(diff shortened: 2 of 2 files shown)


karreiro (approved):
Thank you, @EvilGenius13!

This goes a bit beyond the scope of this PR, but should we also support multi-env handling for the `path` flag? (with something like `(input) => resolvePath(input)`)

Similar to the `store` flag, we have some logic at the `packages/theme/src/cli/flags.ts` level handling `path`, so I’m wondering if it makes sense to handle that here as well.

Thanks again for this PR!

EvilGenius13:
> Thank you, @EvilGenius13!
> 
> This goes a bit beyond the scope of this PR, but should we also support multi-env handling for the `path` flag? (with something like `(input) => resolvePath(input)`)
> 
> Similar to the `store` flag, we have some logic at the `packages/theme/src/cli/flags.ts` level handling `path`, so I’m wondering if it makes sense to handle that here as well.
> 
> Thanks again for this PR!

I will open up a separate issue for this!

graygilmore:
> I will open up a separate issue for this!

Does this solve whatever issue is being referred to? https://github.com/Shopify/cli/pull/6301

EvilGenius13:
> > I will open up a separate issue for this!
> 
> Does this solve whatever issue is being referred to? #6301

It should. There might still be a command or two that we couldn't allow the same path like `theme check --auto-correct` but it would work well for commands such as `theme push` where we take the same theme to multiple stores.

PR #6348, opened 2025-09-02
Add action type tracking to ui-extensions-server-kit state

vctrchu:

Fixes #0000 

### Measuring impact

### Checklist

### DIFF
--- a/packages/ui-extensions-server-kit/src/state/index.ts
+++ b/packages/ui-extensions-server-kit/src/state/index.ts
@@ -1,2 +1,3 @@
 export * from './actions'
 export * from './reducers'
+export type {ActionType} from './reducers/types'
--- a/packages/ui-extensions-server-kit/src/state/reducers/constants.ts
+++ b/packages/ui-extensions-server-kit/src/state/reducers/constants.ts
@@ -3,4 +3,5 @@ import type {ExtensionServerState} from './types'
 export const INITIAL_STATE: ExtensionServerState = {
   extensions: [],
   store: '',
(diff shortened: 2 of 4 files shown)


vctrchu:
* **#6348** <a href="https://app.graphite.dev/github/pr/Shopify/cli/6348?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Shopify/cli/6348?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

vividviolet (commented):
I don't think we need to make changes in the extensions-server-kit for this. It's also a little odd to track the action type in a reducer - generally the reducer just update the states depending on the action and should not "save" that anywhere.

A better place for this would be inside the dev-console-plugin. It's the one calling the reducer to get the new state and deciding what to do with that. Maybe you can add a custom config that POS can pass in to customize the behaviour as needed?  For example, instead of calling the default reducer [here](https://github.com/Shopify/extensibility/blob/eb5cfee4be281c86e12ac4f1048822a4d7ea7e70/packages/dev-console-plugin/src/devConsolePlugin.ts#L218C50-L218C69) you can just make it a no-op if POS does not care to update the state for the `refresh` action.

`refresh` is no longer used by the way - that button has been removed from the Dev Console a long time ago after some feedback from UX.

vctrchu:
@vividviolet Thank you for that recommendation, I'll take another look at altering the dev plug in instead!

PR #6353, opened 2025-09-03
adding local template json override for partners client

cpeddecord:

we allow local overrides for template specs in [app management](https://github.com/Shopify/cli/blob/main/packages/app/src/cli/utilities/developer-platform-client/app-management-client.ts#L454-L461) but not in our partners client, this brings it into parity.

### How to test your changes?

```bash
SHOPIFY_CLI_APP_TEMPLATES_JSON_PATH=/your/path/here pnpm run shopify app generate extension --path /your/path/here --clone-url=https://github.com/Shopify/extensions-templates#clp/remote-dom-validation-settings
```

### DIFF
--- a/packages/app/src/cli/utilities/developer-platform-client/partners-client.test.ts
+++ b/packages/app/src/cli/utilities/developer-platform-client/partners-client.test.ts
@@ -10,8 +10,11 @@ import {
 } from '../../models/app/app.test-data.js'
 import {appNamePrompt} from '../../prompts/dev.js'
 import {FindOrganizationQuery} from '../../api/graphql/find_org.js'
+import {ExtensionTemplate} from '../../models/app/template.js'
 import {partnersRequest} from '@shopify/cli-kit/node/api/partners'
-import {describe, expect, vi, test} from 'vitest'
+import {writeFile, removeFile} from '@shopify/cli-kit/node/fs'
+import {joinPath, cwd} from '@shopify/cli-kit/node/path'
+import {describe, expect, vi, test, beforeEach, afterEach} from 'vitest'
 
 vi.mock('../../prompts/dev.js')
 vi.mock('@shopify/cli-kit/node/api/partners')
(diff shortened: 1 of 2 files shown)


craigmichaelmartin:
We talked offline: by the time this change would be released, nearly all app development will be done via the app management client (and not this partners client). `/snapit` or using a local CLI would support working with this feature now if it is useful for testing other behavior.

packages/app/src/cli/utilities/developer-platform-client/partners-client.test.ts
```diff
@@ -201,4 +204,139 @@ describe('PartnersClient', () => {
...
+    beforeEach(() => {
+      originalEnv = process.env.SHOPIFY_CLI_APP_TEMPLATES_JSON_PATH
+      tempFilePath = joinPath(cwd(), 'test-templates.json')
+    })
+
+    afterEach(async () => {
+      if (originalEnv) {
+        process.env.SHOPIFY_CLI_APP_TEMPLATES_JSON_PATH = originalEnv
+      } else {
+        delete process.env.SHOPIFY_CLI_APP_TEMPLATES_JSON_PATH
```
isaacroldan, line 221: We try to avoid modifying `process.env` directly on tests, it could have unknown side effects if some value is leaked to the next tests or clenaup doesn't work as it should.

the env variables should always be injected.

packages/app/src/cli/utilities/developer-platform-client/partners-client.ts
```diff
@@ -354,15 +356,30 @@
...
-    const variables: RemoteTemplateSpecificationsVariables = {apiKey}
-    const result: RemoteTemplateSpecificationsSchema = await this.request(RemoteTemplateSpecificationsQuery, variables)
-    const templates = result.templateSpecifications.map((template) => {
-      const {types, ...rest} = template
-      return {
-        ...rest,
-        ...types[0],
+    let templates: ExtensionTemplate[]
+    const {templatesJsonPath} = environmentVariableNames
+    const overrideFile = process.env[templatesJsonPath]
```
isaacroldan, line 361: Best to extract this to a function where we can inject `process.env` for testing. 

We have many examples of this in `node/conext/local.ts`
Then in the tests you can mock the call to this new function and won't need to modify `process.env` directly.

PR #6356, opened 2025-09-04
Allow expose to all network interfaces

khiconit:
Support expose to all network interfaces when run in docker container

To expose to all interfaces add flag

```
export SHOPIFY_CLI_HTTP_INTERFACE="0.0.0.0"
```


khiconit:
I have signed the CLA!

PR #6360, opened 2025-09-05
don't report errors when running against non-production services

craigmichaelmartin:

The CLI depends on many other services. We don't want to report errors to our production monitoring systems which occur when using local versions of dependent services. The local services are flakey, and the use of them means we are the CLI to do local development on them, and do not when these local development issues bubbling to our production monitoring systems.

Skipping the reporting of errors which occur while using non-production services.

### Checklist

### DIFF
--- a/packages/app/src/cli/api/graphql/app-management/generated/specifications.ts
+++ b/packages/app/src/cli/api/graphql/app-management/generated/specifications.ts
@@ -16,6 +16,7 @@ export type FetchSpecificationsQuery = {
     uidStrategy:
       | {appModuleLimit: number; isClientProvided: boolean}
       | {appModuleLimit: number; isClientProvided: boolean}
+      | {appModuleLimit: number; isClientProvided: boolean}
     validationSchema?: {jsonSchema: string} | null
   }[]
 }


craigmichaelmartin:
* **#6360** <a href="https://app.graphite.dev/github/pr/Shopify/cli/6360?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Shopify/cli/6360?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

dmerand (approved):

gonzaloriestra (approved):

packages/cli-kit/src/public/node/error-handler.ts
```diff
@@ -63,7 +63,7 @@ export async function sendErrorToBugsnag(
   exitMode: Omit<CommandExitMode, 'ok'>,
 ): Promise<{reported: false; error: unknown; unhandled: unknown} | {error: Error; reported: true; unhandled: boolean}> {
   try {
-    if (settings.debug) {
+    if (process.env.SHOPIFY_SERVICE_ENV === 'local' || settings.debug) {
```
isaacroldan: Instead of checking the raw value, you can use `serviceEnvironment() === Environment.Local` or add a new `isLocalEnvironment` function to `service.ts`

isaacroldan: That way you can mock the call to `isLocalEnvironment` and don't modify the ENV in a test, which is not recommended

packages/cli-kit/src/public/node/error-handler.test.ts
```diff
@@ -22,13 +23,22 @@ vi.mock('@bugsnag/js', () => {
...
+}))
 
 beforeEach(() => {
   vi.mocked(ciPlatform).mockReturnValue({isCI: true, name: 'vitest', metadata: {}})
   vi.mocked(macAddress).mockResolvedValue('macAddress')
   vi.mocked(cloudEnvironment).mockReturnValue({platform: 'localhost', editor: false})
   vi.mocked(hashString).mockReturnValue('hashed-macaddress')
   vi.mocked(isUnitTest).mockReturnValue(true)
+  onNotify.mockClear()
+  delete process.env.SHOPIFY_SERVICE_ENV
```
isaacroldan: you are cleaning the state before, but not after, which means that this value will still be set after these tests finish, potentially affecting other tests. Always better to not modify the environment values in a test :)

packages/cli-kit/src/public/node/error-handler.ts
craigmichaelmartin: Makes sense! Thanks!!

craigmichaelmartin:
Thanks for the feedback, Isaac! Such a nice recommendation to improve the code and tests.

Issue #4833, opened 2024-11-12
[Feature]: Allow to push to multiple environments at once


andershagbard:
### What area(s) will this request affect?

Theme

### What type of change do you want to see?

New feature

### Overview

I would like for Shopify CLI to be able to push to multiple environments at once. The syntax should be very similar to Theme Kit. E.g. `theme push --env="*-production"`.

### Motivation

Theme Kit has been deprecated, but Shopify CLI does not fill all needs and use cases Theme Kit solved.

andershagbard:
Still active

andershagbard:
@EvilGenius13 Do you have an ETA on this? We have a case where we have to deploy to around 200 different stores, and we can't use themekit, cause it doesn't push the blocks folder.

EvilGenius13:
Hi @andershagbard! Thanks for reaching out and most importantly, thank you for your patience. We recently shipped changes that built up the infrastructure in the code to run multiple environments at once. We tested it by allowing the `theme list` command to allow multiple environments first. That came out with the release of `3.7.6`. 

I am currently working on the push command and will update you as soon as we have something ready!

andershagbard:
Still active

andershagbard:
Still active

SimonShirley:
Are you able to create a batch script, like:
`shopify theme push --environment=env-one && shopify theme push --environment=env-two` ?

andershagbard:
> Are you able to create a batch script, like: `shopify theme push --environment=env-one && shopify theme push --environment=env-two` ?

We ended up doing something similar

andershagbard:
I would not consider this fully solved until glob search has been implemented. Can this be opened again?

PR #6361, opened 2025-09-05
Allow theme push and share commands to be called with multiple environments

dejmedus:
- should close https://github.com/Shopify/cli/issues/4833
- should close https://github.com/Shopify/developer-tools-team/issues/816

Previously theme `push` and `share` (which is push behind the scenes)  could only be run with a single `--environment` flag. This PR allows them to be run with multiple environments in a single command

This refactor was also a good opportunity to do some clean up as `share` was the last command with leftover references to CLI2 (references have already been removed from [`pull`, `push`, and `dev`](https://github.com/Shopify/cli/commit/d2c8b9f70a2a086bd4f0b6aeaea59e57f7b78e8e#diff-894ac1bd117f7d7aff566c9a98a4fd512020062eb35d13c87b44a558312b6521))

- Removes flag parsing from `push` and `share` commands in favour of ThemeCommand's `run` logic
- During multi environment runs, the progress bar is no displayed
- Since `push` is exported publicly through [@shopify/cli](https://www.npmjs.com/package/@shopify/cli) to be used by packages like [shopkeeper](https://github.com/TheBeyondGroup/shopkeeper/blob/1b8a8ff30779d35dccb622977f0eebb473320cd1/src/utilities/theme.ts#L2), an auth session is created if one is not passed in from `ThemeCommand` (as one wouldn't be when `push` is used programatically)
- Normalizes paths provided by `shopify.theme.toml` the same way as they would be via CLI
- Removes references to `passThroughFlags` and `cli2Flags` as they are no longer used

<details><summary>push</summary>

https://github.com/user-attachments/assets/a3829d03-bb57-4a0a-906e-dca7123f6488

</details> 

<details><summary>push as imported function</summary>

https://github.com/user-attachments/assets/880cd4d2-9f8b-4ac2-9e93-ddae49dd0d66

</details> 

<details><summary>push jq output</summary>

push-json-jq

</details> 

<details><summary>share</summary>

https://github.com/user-attachments/assets/32225196-b5d6-4246-9c1a-b3aed908a8c5

</details> 

<details><summary>Allow live error</summary>

allow-live-error

</details> 

### How to test your changes?

- Pull down the branch `gh pr checkout 6361` or install the snapit version
```sh
npm i -g --@shopify:registry=https://registry.npmjs.org @shopify/cli@0.0.0-snapshot-20250912171601
```

```
[environments.store1]
theme = "<theme>"
store = "<store>"
path = "path/to/theme"

[environments.store2]
theme = "<theme>"
store = "<store>"
path = "path/to/theme"
```
</details> 

```ts
shopify theme push -e store1 -e store2
shopify theme share -e store1 -e store2
```
- `push` and `share` should work as expected with no, single, and multiple `--environment` flags
- The environment name should be displayed on output
- `push` should error if `allow-live` is not provided while pushing to the live theme

### Measuring impact

### Checklist

### DIFF
--- a/packages/theme/src/cli/utilities/theme-uploader.ts
+++ b/packages/theme/src/cli/utilities/theme-uploader.ts
@@ -80,13 +80,13 @@ export function uploadTheme(
 
       const {progress: uploadProgress, promise: uploadPromise} = await uploadJobPromise
 
-      const timeout = options.multiEnvironment ? 4000 : 1000
+      const updateInterval = options.multiEnvironment ? 4000 : 1000
 
       await renderTasksToStdErr(
         createIntervalTask({
           promise: uploadPromise,
           titleGetter: () => `Uploading files to remote theme ${getProgress(uploadProgress)}`,
-          timeout,
+          updateInterval,
         }),
         context?.stderr,
         options?.multiEnvironment,
@@ -97,7 +97,7 @@ export function uploadTheme(
         createIntervalTask({
           promise: deletePromise,
           titleGetter: () => `Cleaning your remote theme ${getProgress(deleteProgress)}`,
-          timeout,
+          updateInterval,
         }),
         context?.stderr,
         options?.multiEnvironment,
@@ -118,11 +118,11 @@ export function uploadTheme(
 function createIntervalTask({
   promise,
   titleGetter,
-  timeout,
+  updateInterval,
 }: {
   promise: Promise<unknown>
   titleGetter: () => string
-  timeout: number
+  updateInterval: number
 }) {
   const tasks: Task[] = []
 
@@ -132,7 +132,7 @@ function createIntervalTask({
       task: async () => {
         const result = await Promise.race([
           promise,
-          new Promise((resolve) => setTimeout(() => resolve('timeout'), timeout)),
+          new Promise((resolve) => setTimeout(() => resolve('timeout'), updateInterval)),
         ])
 
         if (result === 'timeout') {


packages/theme/src/cli/services/push.ts
```diff
@@ -99,57 +104,72 @@ export interface PushFlags {
...
  * Initiates the push process based on provided flags.
  *
  * @param flags - The flags for the push operation.
  */
-export async function push(flags: PushFlags): Promise<void> {
+export async function push(flags: PushFlags, adminSession?: AdminSession, multiEnvironment?: boolean) {
+  const environment = flags.environment
+
+  // when push is used programmatically, we don't have an admin session, so need to create one
+  const session =
```
dejmedus, line 127: The [push function is exported](https://github.com/Shopify/cli/blob/cbd28df0273fcbcec06277aee17325e527f41f4a/packages/theme/src/index.ts#L52) by [@shopify/cli](https://www.npmjs.com/package/@shopify/cli)

packages/theme/src/cli/services/push.ts
```diff
@@ -339,18 +364,40 @@ export async function createOrSelectTheme(adminSession: AdminSession, flags: Pus
...
+  storeFqdn: string,
+  environment?: string[],
+  multiEnvironment?: boolean,
+) {
   if (themeRole === LIVE_THEME_ROLE) {
     if (allowLive) {
       return true
     }
 
+    if (multiEnvironment) {
```
dejmedus, line 416: Since multi env commands are non-interactive, trying to push to a live theme without `allow-live` will prevent uploads and error

packages/theme/src/cli/utilities/theme-command.ts
```diff
@@ -51,22 +47,6 @@ type EnvironmentName = string
 export type RequiredFlags = (string | string[])[] | null
 
 export default abstract class ThemeCommand extends Command {
-  passThroughFlags(flags: FlagValues, {allowedFlags}: PassThroughFlagsOptions): string[] {
```
dejmedus, line 55: I believe this it leftover from CLI2, it was used in [pull/push/share](https://github.com/Shopify/cli/commit/d2c8b9f70a2a086bd4f0b6aeaea59e57f7b78e8e#diff-894ac1bd117f7d7aff566c9a98a4fd512020062eb35d13c87b44a558312b6521) but there doesnt seem to be any references to it now

passThroughFlags

packages/theme/src/cli/utilities/theme-uploader.ts
```diff
@@ -75,27 +77,85 @@ export function uploadTheme(
...
+ * stderr.write output rather than the animated progress bar
+ */
+async function renderMultiEnvironmentProgress(
+  uploadJobPromise: Promise<SyncJob>,
+  deleteJobPromise: Promise<SyncJob>,
+  options: UploadOptions,
+  getProgress: (params: {current: number; total: number}) => string,
+) {
+  const write = (message: string) =>
+    process.stderr.write(`${timestampDateFormat.format(new Date())} │ ${options.environment} │ ${message}\n`)
```
dejmedus, line 120: The output format is based off `app dev`:

app-output

packages/theme/src/cli/services/push.ts
```diff
@@ -99,58 +105,86 @@ export interface PushFlags {
...
   strict?: boolean
+
+  /** The environment to push the theme to. */
+  environment?: string[]
 }
 
 /**
  * Initiates the push process based on provided flags.
  *
  * @param flags - The flags for the push operation.
```
dejmedus, line 116: I couldn't decide if it would be helpful to add the new params to the docstring - when the function is used programmatically likely only flags would be passed in

packages/theme/src/cli/utilities/theme-ui.ts
```diff
@@ -21,11 +22,21 @@ export function themesComponent(themes: Theme[]) {
...
   force: boolean,
   message = "It doesn't seem like you're running this command in a theme directory.",
+  environment?: string,
+  multiEnvironment?: boolean,
 ) {
   if (force) {
     return true
   }
 
+  if (multiEnvironment) {
```
dejmedus, line 32: In multi environment commands we're returning false if a user can't confirm to align with other multi env command behaviour. However, in single/no env runs we're returning true and continuing

packages/cli-kit/src/private/node/ui/components/Tasks.tsx
```diff
@@ -28,6 +28,7 @@ interface TasksProps<TContext> {
   onComplete?: (ctx: TContext) => void
   abortSignal?: AbortSignal
   noColor?: boolean
+  noProgressBar?: boolean
```
graygilmore, line 25: (nit) generally prefer positive booleans like `progressBar` so that logic reads a little nicer than `!noX` which takes the brain a little bit of work but since `noColor` is already there maybe it's fine to leave it.

packages/theme/src/cli/services/push.ts
```diff
@@ -99,58 +105,86 @@ export interface PushFlags {
...
-    json: flags.json ?? false,
-    force,
-    ignore: flags.ignore ?? [],
-    only: flags.only ?? [],
-  })
+  await executePush(
+    selectedTheme,
+    session,
+    {
+      path: workingDirectory,
```
graygilmore: (super nit) since we're already touching this, feel free to alphabetize

packages/theme/src/cli/utilities/theme-command.ts
```diff
@@ -224,7 +208,9 @@ export default abstract class ThemeCommand extends Command {
         const flagDetails = requiredFlags
           .map((flag) => {
             const usedFlag = Array.isArray(flag) ? flag.find((flag) => flags[flag]) : flag
-            return usedFlag && [usedFlag.includes('password') ? usedFlag : `${usedFlag}: ${flags[usedFlag]}`]
+            if (usedFlag === 'password') return `password`
+            if (usedFlag === 'path' && typeof flags.path === 'string') return `path: ..${flags.path.slice(-12)}`
```
graygilmore: Since `path` will always be set I wonder if we should tweak this a bit? If you had a long folder name it'd be pretty cut off and you might not know what it's saying (especially since `..` has special meaning in the CLI when working with paths.

Maybe we skip adding this one entirely unless it's explicitly provided?

Just in case it's not expected: I *don't* have `path` in my `shopify.theme.toml` for either environment I'm testing this on and I see:

```
➜  aurora-theme git:(main) shopify theme push -e aurora -e farm --allow-live
?  Run push in the following environments?

   ┃  Environment
   ┃  • aurora store: <shop1>.com, password, path: ..aurora-theme, theme: <id>
   ┃  • farm store: <shop2>.myshopify.com, password, path: ..aurora-theme, theme: <id>
```

packages/theme/src/cli/utilities/theme-uploader.ts
```diff
@@ -76,22 +79,38 @@ export function uploadTheme(
       if (options?.deferPartialWork) return
 
       const {progress: uploadProgress, promise: uploadPromise} = await uploadJobPromise
+
+      const timeout = options.multiEnvironment ? 4000 : 1000
+      //
```
graygilmore:
```suggestion
```

packages/theme/src/cli/utilities/theme-command.ts
dejmedus: Previously default flag values (like path = cwd) were being allowed to meet flag requirements. During [check PR review](https://github.com/Shopify/cli/pull/6313/commits/199fa919f0c8e222a292843ced93c82d46032c6b) it was decided that this shouldn't be allowed, so once that validation change merges in we will need to explicitly provide a path. I should have stacked the PRs smarter!

I agree that the truncated path looks a bit strange, could just go back to full paths or maybe it would make more sense to only show the the last `/` segment 🤔 

new-path-truncation

packages/theme/src/cli/utilities/theme-uploader.ts
dejmedus: Thank you! 🙏🏼

packages/theme/src/cli/utilities/theme-command.ts
graygilmore: 🤔 yeah I'm not sure. Full paths might be gnarly here. I like showing the last `/` segment but we need to make sure it's clear that we're clipping the path.

dejmedus:
Fixed a conflict as progress bar has been split out from the `Tasks` component. `noProgressBar` is now being [passed down](https://github.com/Shopify/cli/pull/6361/files#diff-cf9358a83b6607effdfb4261c66f0200f09f1c7ee22f056d51352f3b64059051R16) to `LoadingBar` and its test has been moved as well

packages/theme/src/cli/utilities/theme-command.ts
dejmedus: That makes sense! I've updated it to match the screenshot: `...` and the last segment. Hopefully the three dots communicate truncation well-ish, but i'm open to changing it

packages/theme/src/cli/utilities/theme-command.ts
```diff
@@ -229,7 +213,12 @@ export default abstract class ThemeCommand extends Command {
         const flagDetails = requiredFlags
           .map((flag) => {
             const usedFlag = Array.isArray(flag) ? flag.find((flag) => flags[flag]) : flag
-            return usedFlag && [usedFlag.includes('password') ? usedFlag : `${usedFlag}: ${flags[usedFlag]}`]
+            if (usedFlag === 'password') return `password`
+            if (usedFlag === 'path' && typeof flags.path === 'string') {
+              const lastSegment = flags.path.split('/').pop() || flags.path
+              return `path: .../${lastSegment}`
```
graygilmore: (this might be too much extra for this PR, feel free to ignore)

It might be best to include both the beginning and end. Something like:

```
~/.../folder-name
/Something/.../folder-name
```

What do you think? Here's some code example:

```js
const flags = [
  'single-folder',
  '~/Users/graygilmore/repos/aurora-theme',
  'C:\\SuperCoolDriveName\\LintingRules\\pipeline-rules\\.theme-check.yml',
  '/home/runner/work/Shopify/Shopify/tools/something-else-here/cool_folder',
]

flags.forEach((flag) => {
  const splits = flag.split(/[/|\\]/);

  if (splits.length === 1) {
    console.log(flag)
    return
  }

  const first = splits[0] === '' ? `/${splits[1]}` : splits[0]
  const last = splits.at(-1)

  console.log(`${first}/.../${last}`)
})
```

Which would output:

```
single-folder
~/.../aurora-theme
C:/.../.theme-check.yml
/home/.../cool_folder
```

I'm not 100% sure if we need to specifically handle `\` for Windows here.

graygilmore (approved):

packages/theme/src/cli/utilities/theme-uploader.ts
```diff
@@ -76,22 +79,38 @@ export function uploadTheme(
       if (options?.deferPartialWork) return
 
       const {progress: uploadProgress, promise: uploadPromise} = await uploadJobPromise
+
+      const timeout = options.multiEnvironment ? 4000 : 1000
```
EvilGenius13: How did you pick the 4 second timeout for multi environment?
Side nit - although it was called timeout before, can we call it something like `updateInterval` if this is just the delay to update the progress to the terminal?

packages/theme/src/cli/utilities/theme-command.ts
dejmedus: I like this a lot! The only difference would be that the path has already been resolved, so relative `single-folder` would look something like `Users/.../single-folder` and same with `~`

Updated: paths

packages/theme/src/cli/utilities/theme-uploader.ts
dejmedus: Will do! I increased it to try to get closer to 25% jumps each time. At 1000 the logs tended to fill up shorter terminal windows, which i'm not totally against if we do want to change back.

EvilGenius13 (approved):
Awesome work! 🎩 went well

packages/theme/src/cli/utilities/theme-command.ts
gonzaloriestra: Right, those were flags we passed to the CLI2 when we used it internally during the migration.

andershagbard:
@dejmedus Have you considered implementing glob search, like Theme Kit has? `*-production` to target e.g. `int-production` and `us-production`

dejmedus:
Hi @andershagbard :wave: that makes sense! I've created an issue for the feature [here](https://github.com/Shopify/cli/issues/6481)

PR #6363, opened 2025-09-06
Refactor theme init command to use ThemeCommand's run method

dejmedus:

Previously `theme init` command implemented its own `run` logic. This PR extends ThemeCommand's `command` method to accept `args` and refactors `init` to use its logic. This will help to standardize command architecture

- Grabs `args` from `parse()` and passes it to commands
- Refactors `init` to define its own `command()` rather than `run()` so it can instead use ThemeCommand's base `run()` method

 <details><summary>init</summary>

https://github.com/user-attachments/assets/2e7e9e5f-17de-4655-a6e1-13e7a1151a0d

</details> 

### How to test your changes?

- Pull down the branch `gh pr checkout 6363` or install the snapit version
```sh
npm i -g --@shopify:registry=https://registry.npmjs.org @shopify/cli@0.0.0-snapshot-20250912003518
```
- `init` should work as expected. It does not accept any environment flags
```
shopify theme init
shopify theme init <name>
```

### Measuring impact

### Checklist

### DIFF
--- a/packages/theme/src/cli/commands/theme/init.ts
+++ b/packages/theme/src/cli/commands/theme/init.ts
@@ -13,6 +13,11 @@ import {generateRandomNameForSubdirectory} from '@shopify/cli-kit/node/fs'
 import {renderTextPrompt} from '@shopify/cli-kit/node/ui'
 import {joinPath} from '@shopify/cli-kit/node/path'
 import {terminalSupportsPrompting} from '@shopify/cli-kit/node/system'
+import {InferredArgs, InferredFlags} from '@oclif/core/interfaces'
+import {AdminSession} from '@shopify/cli-kit/node/session'
+
+type InitFlags = InferredFlags<typeof Init.flags>
+type InitArgs = InferredArgs<typeof Init.args>
 
 export default class Init extends ThemeCommand {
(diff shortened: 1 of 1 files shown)


packages/theme/src/cli/utilities/theme-command.ts
```diff
@@ -268,7 +271,7 @@ export default abstract class ThemeCommand extends Command {
                 const commandName = this.constructor.name.toLowerCase()
                 recordEvent(`theme-command:${commandName}:multi-env:authenticated`)
 
-                await this.command(flags, session, true, {stdout, stderr})
+                await this.command(flags, session, true, {}, {stdout, stderr})
```
graygilmore, line 270: Should this be `args`, too?

graygilmore (approved):
👍🏻 🎩

EvilGenius13 (approved):
Ran some 🎩 using different flags and everything went well!

packages/theme/src/cli/utilities/theme-command.ts
dejmedus: afaik `init` is the only command that accepts `args` and it doesn't accepts `environment` flags. We could add args to the environment map so it could be used here (future proofing?) but as of right now it would always be `{}` anyways

PR #6365, opened 2025-09-08
Fix sourcemaps for Observe error grouping

amcaplan:
## Summary
- Upload source files (src/) alongside compiled files so Observe can extract surrounding code
- Upload sourcemaps for ALL packages that contribute to runtime, not just published ones
- Integrate sourcemap upload into the release process

## Problem
Observe was unable to group errors by "surrounding code" and was falling back to stack trace grouping. This happened because:
1. We were only uploading compiled JS and sourcemap files, not the original TypeScript source
2. Only published packages (cli, create-app) had sourcemaps uploaded
3. Other packages like @shopify/app, @shopify/cli-kit that contribute to runtime errors had no sourcemaps

## Solution
1. Created centralized upload script (`bin/upload-sourcemaps-all.js`) that uploads all runtime packages
2. Modified upload to include `src/`, `dist/`, and `bin/` directories
3. Integrated into release process so it runs automatically
4. Deprecated individual package upload scripts

## Test Plan
- [ ] Run `/snapit` to create a test release
- [ ] Trigger an error in the test release
- [ ] Verify in Observe that errors are now grouped by "surrounding code" instead of stack trace
- [ ] Verify sourcemaps are uploaded for all packages in GCS

Related issue: https://github.com/shop/issues-app-inner-loop/issues/181

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/bin/update-bugsnag.js
+++ b/bin/update-bugsnag.js
@@ -1,73 +1,22 @@
 #!/usr/bin/env node
 
-import fs from 'fs';
-import fsPromise from 'fs/promises'
-import path from 'node:path';
+// This script is deprecated. It now just calls the centralized upload script.
+// Kept for backwards compatibility in case it's referenced elsewhere.
+
+import { spawn } from 'child_process';
+import path from 'path';
 import { fileURLToPath } from 'url';
-import { node } from "@bugsnag/source-maps";
-import reportBuild from 'bugsnag-build-reporter';
-import glob from 'fast-glob';
-import tmp from 'tmp';
 
 const __filename = fileURLToPath(import.meta.url);
 const __dirname = path.dirname(__filename);
-const appVersion = JSON.parse(fs.readFileSync(`${__dirname}/../packages/cli/package.json`)).version;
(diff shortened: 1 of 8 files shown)


bin/update-bugsnag.js
```diff
@@ -1,73 +1,22 @@
...
-          reject(e);
-        }
-      });
-    });
-
-    console.log(`Cleaning sourcemaps from @shopify/${packageName}`);
 
-    const packageDist = path.join(sourceDirectory, 'dist');
-    const sourcemaps = glob.sync(`${packageDist}/**/*.map`, {onlyFiles: true});
+console.log('Note: update-bugsnag.js is deprecated. Using centralized upload-sourcemaps-all.js instead.');
```
isaacroldan, line 13: I think the other place where this is called is the main `package.json` with `"update-bugsnag": "bin/update-bugsnag.js"`.

If we update that one we can just remove this file right?

amcaplan:
Closing as we've determined this is unnecessary.

PR #6380, opened 2025-09-10
add more event logging around graphql requests for theme commands

EvilGenius13:
🚧 👷 🚧 

Fixes #0000 

### Measuring impact

### Checklist


packages/cli-kit/src/public/node/themes/api.ts
```diff
@@ -554,6 +576,7 @@ export async function passwordProtected(session: AdminSession): Promise<boolean>
 }
 
 function unexpectedGraphQLError(message: string): never {
+  recordError(message)
   throw new AbortError(message)
```
karreiro: I believe we could also do something like this:
```ts
throw recordError(new AbortError(message))
```

EvilGenius13: I'll amend those!

PR #6385, opened 2025-09-11
Enable traversal of imported files in ui-extensions

JoviDeCroock:

Related to https://github.com/shop/issues-develop/issues/11518

The current UI extension type generation system does not handle any imports, meaning we only generate types for the extension file itself and not it's dependencies.

- **Enhanced import parsing**: Now captures all import types (ES6, require, dynamic) using `es-module-lexer`
- **TypeScript-aware resolution**: Uses `ts.resolveModuleName()` with proper `tsconfig.json` path mapping support
- **Maintains backward compatibility**: All existing functionality preserved

The system now:

1. Finds and parses the nearest `tsconfig.json` 
2. Uses TypeScript's official module resolution algorithm
3. Generates type definitions for all resolved local files (excluding node_modules)
4. Creates union types for modules shared between different targets
5. Supports complex path mapping scenarios

In the future we could add support for shared workspaces as well when a partner i.e. uses a shared workspace with common utils.

### Post-release steps

No post-release steps required.

### Measuring impact

### Checklist

### Checklist

### DIFF
--- a/packages/app/package.json
+++ b/packages/app/package.json
@@ -62,6 +62,7 @@
     "camelcase-keys": "9.1.3",
     "chokidar": "3.6.0",
     "diff": "5.2.0",
+    "es-module-lexer": "1.7.0",
     "esbuild": "0.25.5",
     "express": "4.21.2",
     "graphql-request": "6.1.0",
--- a/packages/app/package.json
+++ b/packages/app/package.json
@@ -62,7 +62,6 @@
     "camelcase-keys": "9.1.3",
     "chokidar": "3.6.0",
     "diff": "5.2.0",
-    "es-module-lexer": "1.7.0",
     "esbuild": "0.25.5",
     "express": "4.21.2",
     "graphql-request": "6.1.0",
--- a/packages/app/src/cli/models/extensions/specifications/type-generation.ts
+++ b/packages/app/src/cli/models/extensions/specifications/type-generation.ts
@@ -2,7 +2,6 @@ import {fileExists, findPathUp, readFileSync} from '@shopify/cli-kit/node/fs'
 import {dirname, joinPath, relativizePath, resolvePath} from '@shopify/cli-kit/node/path'
 import {AbortError} from '@shopify/cli-kit/node/error'
 import ts from 'typescript'
-import {init, parse} from 'es-module-lexer'
 import {createRequire} from 'module'
 
 const require = createRequire(import.meta.url)
@@ -63,21 +62,43 @@ async function fallbackResolve(importPath: string, baseDir: string): Promise<str
 
 async function parseAndResolveImports(filePath: string): Promise<string[]> {
   try {
-    await init
-
     const content = readFileSync(filePath).toString()
     const resolvedPaths: string[] = []
 
     // Load TypeScript configuration once
     const {compilerOptions} = loadTsConfig(filePath)
 
-    const [imports] = parse(content)
+    // Determine script kind based on file extension
+    let scriptKind = ts.ScriptKind.JSX
+    if (filePath.endsWith('.ts')) {
+      scriptKind = ts.ScriptKind.TS
+    } else if (filePath.endsWith('.tsx')) {
+      scriptKind = ts.ScriptKind.TSX
+    }
+
+    const sourceFile = ts.createSourceFile(filePath, content, ts.ScriptTarget.Latest, true, scriptKind)
 
     const processedImports = new Set<string>()
+    const importPaths: string[] = []
 
-    for (const pattern of imports) {
-      const importPath = pattern.n
+    const visit = (node: ts.Node): void => {
+      if (ts.isImportDeclaration(node) && node.moduleSpecifier && ts.isStringLiteral(node.moduleSpecifier)) {
+        importPaths.push(node.moduleSpecifier.text)
+      } else if (ts.isCallExpression(node) && node.expression.kind === ts.SyntaxKind.ImportKeyword) {
+        const firstArg = node.arguments[0]
+        if (firstArg && ts.isStringLiteral(firstArg)) {
+          importPaths.push(firstArg.text)
+        }
+      } else if (ts.isExportDeclaration(node) && node.moduleSpecifier && ts.isStringLiteral(node.moduleSpecifier)) {
+        importPaths.push(node.moduleSpecifier.text)
+      }
 
+      ts.forEachChild(node, visit)
+    }
+
+    visit(sourceFile)
+
+    for (const importPath of importPaths) {
       // Skip if already processed
       if (!importPath || processedImports.has(importPath)) {
         continue
@@ -149,7 +170,7 @@ export function createTypeDefinition(
         // Throw specific error for the target that failed, matching the original getSharedTypeDefinition behavior
         throw new AbortError(
           `Type reference for ${target} could not be found. You might be using the wrong @shopify/ui-extensions version.`,
-          `Fix the error by ensuring you have the correct version of @shopify/ui-extensions, for example ${year}.${month}.0, in your dependencies.`,
+          `Fix the error by ensuring you have the correct version of @shopify/ui-extensions, for example ~${year}.${month}.0, in your dependencies.`,
         )
       }
     }
@@ -160,8 +181,8 @@ export function createTypeDefinition(
       const target = targets[0] ?? ''
       return `//@ts-ignore\ndeclare module './${relativePath}' {\n  const shopify: import('@shopify/ui-extensions/${target}').Api;\n  const globalThis: { shopify: typeof shopify };\n}\n`
     } else if (targets.length > 1) {
-      const unionType = targets.map((target) => `import('@shopify/ui-extensions/${target}').Api`).join(' | ')
-      return `//@ts-ignore\ndeclare module './${relativePath}' {\n  const shopify: ${unionType};\n  const globalThis: { shopify: typeof shopify };\n}\n`
+      const unionType = targets.map((target) => `    import('@shopify/ui-extensions/${target}').Api`).join(' |\n')
+      return `//@ts-ignore\ndeclare module './${relativePath}' {\n  const shopify: \n${unionType};\n  const globalThis: { shopify: typeof shopify };\n}\n`
     }
 
     return null
@@ -173,7 +194,7 @@ export function createTypeDefinition(
     const {year, month} = parseApiVersion(apiVersion) ?? {year: 2025, month: 10}
     throw new AbortError(
       `Type reference could not be found. You might be using the wrong @shopify/ui-extensions version.`,
-      `Fix the error by ensuring you have the correct version of @shopify/ui-extensions, for example ${year}.${month}.0, in your dependencies.`,
+      `Fix the error by ensuring you have the correct version of @shopify/ui-extensions, for example ~${year}.${month}.0, in your dependencies.`,
     )
   }
 }
--- a/packages/app/src/cli/models/extensions/specifications/type-generation.ts
+++ b/packages/app/src/cli/models/extensions/specifications/type-generation.ts
@@ -1,0 +1,191 @@
+import {fileExists, findPathUp, readFileSync} from '@shopify/cli-kit/node/fs'
+import {dirname, joinPath, relativizePath, resolvePath} from '@shopify/cli-kit/node/path'
+import {AbortError} from '@shopify/cli-kit/node/error'
+import ts from 'typescript'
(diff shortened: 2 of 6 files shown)


packages/app/src/cli/models/extensions/specifications/ui_extension.ts
isaacroldan, line 1: This file has grown a lot, and it does many things... can we extract functionality? All these type generation stuff could be outside of the specificatio file

packages/app/package.json
```diff
@@ -62,6 +62,7 @@
     "camelcase-keys": "9.1.3",
     "chokidar": "3.6.0",
     "diff": "5.2.0",
+    "es-module-lexer": "1.7.0",
```
isaacroldan: @gonzaloriestra You were testing `tree-sitter` as a dependency to get the import tree of a file.
We should probably try to agree on just one dependency to do this.

JoviDeCroock: From an import discovery perspective `es-module-lexer` and `rs-module-lexer` are really convenient as they are tailored towards that.

isaacroldan: We wanted a more general solution that could support Rust too. 
This is the prototype @gonzaloriestra was working on https://github.com/Shopify/cli/pull/6078/files#diff-1b38d94d50011f9ebfefec801525f851596ded02ff74515da82cc3af73345b39

JoviDeCroock: Gotcha, yes I mean happy to refactor but I do think that this is on a timeline 😅

isaacroldan: I'm going to look into finishing the prototype using `tree-sitter` this week, is that a dependency you can use for your case? 

I understand there are timelines, but we don't want to rush things like this, because if we do we'll have 2 solutions for the same thing basically forever

JoviDeCroock: If `tree-sitter` exposes the imported modules by just running it on a source-file we can use it. The only thing I need to discover is the statically and dynamically imported modules so I can continue my traversal and generate types

gonzaloriestra: Hi! I agree it would be nice to support Rust as well, so I'm working on that PR right now. I'll ping you once it's ready.

JoviDeCroock: I have taken a middle ground for now as we'd like to have this by Oct 1, we re-use the TS parser

gonzaloriestra: I expect to have the PR ready this week, does it work?

JoviDeCroock: I am def down to wait, or alternatively we merge this and refactor with tree-sitter later. Up to y'all

gonzaloriestra: Ok, thanks. Let me try to finish the other PR first

gonzaloriestra: Ready for review! https://github.com/Shopify/cli/pull/6078

packages/app/src/cli/models/extensions/specifications/ui_extension.test.ts
```diff
@@ -1116,5 +1116,218 @@ Please check the configuration in ${uiExtension.configurationPath}`),
...
+
+        // When
+        await extension.contributeToSharedTypeFile?.(typeDefinitionsByFile)
+
+        const shopifyDtsPath = joinPath(tmpDir, 'shopify.d.ts')
+
+        // Then - should include types for imported modules when single target
+        expect(typeDefinitionsByFile.get(shopifyDtsPath)).toContain(
+          `//@ts-ignore\ndeclare module './src/utils/helper.js' {\n  const shopify: import('@shopify/ui-extensions/admin.product-details.action.render').Api;\n  const globalThis: { shopify: typeof shopify };\n}\n`,
+        )
```
kumar303: It would be nice to make this code more readable in the tests, if possible.

packages/app/src/cli/models/extensions/specifications/ui_extension.test.ts
```diff
@@ -1116,5 +1116,218 @@ Please check the configuration in ${uiExtension.configurationPath}`),
...
+
+        // When
+        await extension.contributeToSharedTypeFile?.(typeDefinitionsByFile)
+
+        const shopifyDtsPath = joinPath(tmpDir, 'shopify.d.ts')
+        const types = typeDefinitionsByFile.get(shopifyDtsPath)
+
+        // Then - should generate union type for shared module
+        expect(types).toContain(
+          `//@ts-ignore\ndeclare module './shared/utils.js' {\n  const shopify: import('@shopify/ui-extensions/admin.product-details.action.render').Api | import('@shopify/ui-extensions/admin.orders-details.block.render').Api;\n  const globalThis: { shopify: typeof shopify };\n}\n`,
```
kumar303: It would be nice to make this more readable in the actual `shopify.d.ts` file by adding line breaks between each target:

```ts
//@ts-ignore
declare module './shared/utils.js' {
  const shopify: 
    | import('@shopify/ui-extensions/admin.product-details.action.render').Api
    | import('@shopify/ui-extensions/admin.orders-details.block.render').Api;
  const globalThis: { shopify: typeof shopify };
}
```

This is because developers might be looking at this file to figure out what's going on.

packages/app/src/cli/models/extensions/specifications/ui_extension.ts
```diff
@@ -158,46 +155,101 @@ const uiExtensionSpec = createExtensionSpecification({
...
+    }
 
+    // Second pass: find all imported files from each entry point
+    for await (const extensionPoint of configuration.extension_points) {
+      const fullPath = joinPath(extension.directory, extensionPoint.module)
+      const exists = await fileExists(fullPath)
+      if (!exists) continue
+
+      // Find all imported files recursively
+      const importedFiles = await findAllImportedFiles(fullPath)
```
kumar303, line 195: Isn't there a way to hook into the existing CLI build process? I'm imagining you could run the equivalent of `shopify app build` then inspect the bundle to find all of the imports, no?

Otherwise, there's a risk that this custom introspection code could get out of sync with the actual build process.

packages/app/src/cli/models/extensions/specifications/type-generation.ts
```diff
@@ -0,0 +1,212 @@
...
+      }
+    }
+
+    return resolvedPaths
+  } catch (error) {
+    // Re-throw AbortError as-is, wrap other errors
+    if (error instanceof AbortError) {
+      throw error
+    }
+    return []
```
kumar303, line 133: Since this `catch` could cover so many different exceptions, what about adding some debug logging to show which exception was ignored?

packages/app/src/cli/models/extensions/specifications/type-generation.ts
```diff
@@ -0,0 +1,212 @@
...
+    // Validate that all targets can be resolved
+    for (const target of targets) {
+      try {
+        require.resolve(`@shopify/ui-extensions/${target}`, {paths: [fullPath, typeFilePath]})
+      } catch (_) {
+        const {year, month} = parseApiVersion(apiVersion) ?? {year: 2025, month: 10}
+        // Throw specific error for the target that failed, matching the original getSharedTypeDefinition behavior
+        throw new AbortError(
+          `Type reference for ${target} could not be found. You might be using the wrong @shopify/ui-extensions version.`,
+          `Fix the error by ensuring you have the correct version of @shopify/ui-extensions, for example ${year}.${month}.0, in your dependencies.`,
```
kumar303: nit: encourage patch fixes


```suggestion
          `Type reference for ${target} could not be found. You might be using the wrong @shopify/ui-extensions version.`,
          `Fix the error by ensuring you have the correct version of @shopify/ui-extensions, for example ~${year}.${month}.0, in your dependencies.`,
```

kumar303 (approved):
I love this feature ✨ I don't have much CLI context so here are some random suggestions. I did not tophat.

packages/app/src/cli/models/extensions/specifications/ui_extension.ts
JoviDeCroock: I don't think there's currently a way to hook into that. I don't think there is a big risk of this getting out of sync with the build process as they are unrelated to one another, we shouldn't be adding non node-module imports to the final bundle without the knowledge of the user. I can understand virtual modules/node modules but they should be typed _and_ won't be used by the user and hence the TypeScript process if they are adde by build.

JS imports should be deterministic as we are basically helping our LSP process

kumar303: ok, I just thought the source files were already available through the build (since they would be used in a source map anyway).

JoviDeCroock:
@isaacroldan @gonzaloriestra as both branches seem to be running what if we merge this one so we can get it in the next release and I create a branch PR on https://github.com/Shopify/cli/pull/6078 to use the new way. This PR is not adding a new dependency, it's using the same one as tree-sitter does (typescript) so we should be good from that point of view imho

gonzaloriestra:
@JoviDeCroock I'm just polishing https://github.com/Shopify/cli/pull/6078, so will be included in the next release, don't worry (we take care of the release and will wait if needed). If for some reason we can't finish it, we will merge this one instead.

packages/app/package.json
isaacroldan: I think we can merge and refactor in a few days when Gonzalo's PR will be fully ready

packages/app/src/cli/models/extensions/specifications/ui_extension.ts
isaacroldan: We are working on a shared way to obtain the list of imports, should be ready later this week probably. You'll need to refactor this PR to use that after 👌

isaacroldan (approved):
🎩 'ed and works as expected

JoviDeCroock:
@isaacroldan is it normal for the CI to flake like this?

isaacroldan:
Um, the errors I see seem related to your code changes no?

JoviDeCroock:
Odd they were passing before... Let me investigate, I have no windows machine tho. I think I might know what's going on

packages/app/src/cli/models/extensions/specifications/ui_extension.test.ts
```diff
@@ -1116,5 +1116,291 @@ Please check the configuration in ${uiExtension.configurationPath}`),
...
+        expect(Array.from(typeDefinitionsByFile.get(shopifyDtsPath) ?? [])).toContain(
+          `//@ts-ignore\ndeclare module './src/utils/helper.js' {\n  const shopify: import('@shopify/ui-extensions/admin.product-details.action.render').Api;\n  const globalThis: { shopify: typeof shopify };\n}\n`,
+        )
+        expect(Array.from(typeDefinitionsByFile.get(shopifyDtsPath) ?? [])).toContain(
+          `//@ts-ignore\ndeclare module './src/components/Button.jsx' {\n  const shopify: import('@shopify/ui-extensions/admin.product-details.action.render').Api;\n  const globalThis: { shopify: typeof shopify };\n}\n`,
+        )
+      })
+    })
+
+    test('does not generate shopify.d.ts at root when importing files outside extension directory', async () => {
```
vividviolet:
```suggestion
    test('generates shopify.d.ts in the extension directory when importing files outside extension directory', async () => {
```

packages/app/src/cli/models/extensions/specifications/type-generation.ts
```diff
@@ -0,0 +1,218 @@
...
+      return `//@ts-ignore\ndeclare module './${relativePath}' {\n  const shopify: \n${unionType};\n  const globalThis: { shopify: typeof shopify };\n}\n`
+    }
+
+    return null
+  } catch (error) {
+    // Re-throw AbortError as-is, wrap other errors
+    if (error instanceof AbortError) {
+      throw error
+    }
+    const {year, month} = parseApiVersion(apiVersion) ?? {year: 2025, month: 10}
```
vividviolet, line 194: Can you put this error handling into a share util instead of duplicating the code?

vividviolet (approved):
Looks great but I didn't get a chance to top hat yet

packages/app/src/cli/models/extensions/specifications/type-generation.ts
JoviDeCroock: The lint rules discourage returning an instance of abort-error hence the duplication 😅

JoviDeCroock:
Oh yikes, we might need a different `stopAt` then for the findNearestTsconfig

EDIT: I think https://github.com/Shopify/cli/compare/e31c819e7a7b2e35a7743e982a00eb5d17f4bfa9..78ac4cbb49f4e45ccda8a93b778c3830e23eeacc should have us covered

PR #6391, opened 2025-09-12
Install all function build binaries, not just Javy

jeffcharles:

There is an issue some Function developers are seeing when running `shopify app build` where trying to execute trampoline binary resulted in an `ENOENT` error message when trying to spawn `shopify-function-trampoline-v1.0.2`. While investigating, I discovered that [the `moveFile` operation we perform while downloading a binary](shopify-function-trampoline-v1.0.2) is *not* atomic. If the destination file exists, the underlying library will delete the destination file before performing the move. This _could_ result in an `ENOENT` when trying to spawn that executable file if the timing is just right. 

I've renamed `installJavy` to `installBuildTools` and have it also install `wasm-opt` and the trampoline. By installing these binaries upfront once, we won't attempt to download and move the _same_ binary file concurrently.

### How to test your changes?

`pnpm shopify app build --path <path_to_project_with_functions>`

### Measuring impact

### Checklist

### DIFF
--- a/.changeset/slow-glasses-hide.md
+++ b/.changeset/slow-glasses-hide.md
@@ -1,0 +1,5 @@
+---
+'@shopify/app': patch
+---
+
+Fix race condition with trampoline downloads


jeffcharles:
~One of the CI checks appears to be failing because https://github.com/Shopify/cli/pull/6386 hasn't been merged~

jeffcharles:
Closing this in favour of #6395 since we want to support multiple versions of the trampoline and we need to build a Wasm module to know which version of the trampoline to download.

PR #6395, opened 2025-09-12
Fix concurrency issue in downloadBinary

jeffcharles:

There is an issue some Function developers are seeing when running `shopify app build` where trying to execute trampoline binary resulted in an `ENOENT` error message when trying to spawn `shopify-function-trampoline-v1.0.2`. While investigating, I discovered that [the `moveFile` operation we perform while downloading a binary](shopify-function-trampoline-v1.0.2) is *not* atomic. If the destination file exists, the underlying library will delete the destination file before performing the move. This _could_ result in an `ENOENT` when trying to spawn that executable file if the timing is just right.

This introduces a check to ensure only one JS task at a time tries to download a specific binary and waits if there is a download in progress for that library. I've added a lot of code comments because it is likely not obvious to every reader of the code how or why this approach works without them and there's no good way to structure the code to be more self-documenting.

### How to test your changes?

`pnpm shopify app build --path <path_to_project_with_functions>`

### Measuring impact

### Checklist

### DIFF
--- a/.changeset/healthy-coins-walk.md
+++ b/.changeset/healthy-coins-walk.md
@@ -1,0 +1,5 @@
+---
+'@shopify/app': patch
+---
+
+Fix concurrency issue with downloading Functions binaries
--- a/packages/app/src/cli/services/function/binaries.test.ts
+++ b/packages/app/src/cli/services/function/binaries.test.ts
@@ -141,6 +141,26 @@ describe('javy', () => {
     expect(fetch).toHaveBeenCalledOnce()
     await expect(fileExists(javy.path)).resolves.toBeTruthy()
   })
+
+  test('handles concurrent downloads of Javy', async () => {
+    // Given
+    await removeFile(javy.path)
+    await expect(fileExists(javy.path)).resolves.toBeFalsy()
+    vi.mocked(fetch).mockResolvedValue(new Response(gzipSync('javy binary')))
+
+    // When
+    await Promise.all([
+      downloadBinary(javy),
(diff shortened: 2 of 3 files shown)


isaacroldan:
I think the solution makes sense, maybe you can add a few tests to ensure only one download is ever in progress?

jeffcharles:
> maybe you can add a few tests to ensure only one download is ever in progress?

I don't think it's possible to test deterministically only one download is ever in progress because the downloads are concurrent operations. Using automated testing to guarantee invariants on concurrent operations is inherently impossible because automated tests aren't capable of generating all possible interleavings of concurrent operations.

I can test that the expected key on `downloadsInProgress` is set but that doesn't validate that it's being awaited on appropriately. I can also write a test that enqueuing downloading the same binary for download doesn't throw an exception. But that also doesn't necessarily mean that only one download is in progress at a time. I could insert a synchronous call after entering the critical section and a synchronous call after exiting the critical section that we could mock in tests but you can still end up in a state where those calls happen in the expected sequence in tests but sometimes don't happen in that sequence outside of tests due to differences in when the download tasks are scheduled by V8.

jeffcharles:
I added a test case where I schedule 5 concurrent downloads of Javy and have a mock once on the fetch call. It doesn't guarantee there isn't a concurrency bug but better than nothing.

isaacroldan (approved):

isaacroldan:
Thanks for the detailed explanation @jeffcharles :v:

PR #6396, opened 2025-09-14
Make bind address configurable for app dev server

### LINKED ISSUE #6355: [Feature]:  Make bind address configurable for local servers

Running your code locally

Substantial change to existing feature

### Overview

**Context:**

Recent changes in the CLI enforce binding local development servers to localhost for security reasons. While this is a sensible default, it causes issues in containerized environments like Docker.

**Problem:**

When running the CLI inside a Docker container, forwarded ports from the host machine are only accessible if the server binds to 0.0.0.0. Binding to localhost makes the server unreachable from outside the container, breaking the development workflow.

**Proposal:**

Introduce a configuration option (e.g. environment variable or CLI flag) that allows developers to override the default bind address. For example:

 
`SHOPIFY_CLI_BIND_ADDRESS=0.0.0.0 yarn dev`
This would preserve the secure default (localhost) while enabling flexibility for containerized setups.

**Benefits:**

Supports Docker and other container environments

Avoids the need for custom patches or forks

Maintains security for most users while offering configurability

Thank you for your work on improving the CLI! Happy to help test or contribute if needed.

### Motivation

I’m running the app inside a Docker container. When forwarding ports from the host to the container, Docker only allows binding to `0.0.0.0`, not `localhost`. This means the development server becomes inaccessible unless it listens on all interfaces.

Would it be possible to make the bind address configurable—perhaps via an environment variable or CLI flag—so developers can choose between localhost and `0.0.0.0` depending on their setup?

lmcrean:
Adds configurable bind address support to the `shopify app dev` command to enable Docker container development workflows.

### Problem
- Recent CLI versions hardcode `localhost` binding for app dev servers for security
- Docker containers require `0.0.0.0` binding to accept connections from host machine
- Theme dev server already supports `--host` flag, but app dev server does not
- Breaks containerized development setups (working version was 3.83.3)

### Solution
- Add `--host` flag to `shopify app dev` command with `SHOPIFY_FLAG_HOST` env var support
- Update proxy server setup to use configurable host instead of hardcoded `'localhost'`
- Maintain secure default (`localhost`) while enabling Docker flexibility
- Match existing theme dev server pattern for consistency

### 1. Added host flag to app dev command
- **File**: `packages/app/src/cli/commands/app/dev.ts`
- Added `--host` flag with environment variable support
- Default: `localhost` (maintains security)
- Override: `0.0.0.0` (enables Docker containers)

### 2. Updated proxy server to use configurable host
- **File**: `packages/app/src/cli/services/dev/processes/setup-dev-processes.ts`
- Changed hardcoded `'localhost'` to use passed host parameter
- Added host parameter to proxy server setup function

### For Docker containers:
```bash
SHOPIFY_FLAG_HOST=0.0.0.0 shopify app dev
# or
shopify app dev --host=0.0.0.0
```

### For normal development (default):
```bash
shopify app dev  # Still binds to localhost by default
```

## Testing

- [ ] Verify default behavior unchanged (binds to localhost)
- [ ] Verify `--host=0.0.0.0` enables external connections
- [ ] Verify `SHOPIFY_FLAG_HOST` environment variable works
- [ ] Test Docker container setup with port forwarding
- [ ] Ensure backwards compatibility

## Risk Assessment

- **Low risk**: Changes follow existing CLI patterns
- **Backwards compatible**: Default behavior unchanged
- **Security maintained**: Still defaults to localhost
- **Consistent**: Matches theme dev server implementation

## Related Issues

- Fixes #6355
- Addresses Docker container development workflows
- Maintains parity with theme dev server host configuration

### DIFF
--- a/packages/app/src/cli/commands/app/dev.ts
+++ b/packages/app/src/cli/commands/app/dev.ts
@@ -87,6 +87,11 @@ If you're using the Ruby app template, then you need to complete the following s
       default: false,
       exclusive: ['tunnel-url'],
     }),
+    host: Flags.string({
+      description: 'Set which network interface the web server listens on. The default value is localhost.',
+      env: 'SHOPIFY_FLAG_HOST',
+      default: 'localhost',
+    }),
     'localhost-port': Flags.integer({
       description: 'Port to use for localhost.',
       env: 'SHOPIFY_FLAG_LOCALHOST_PORT',
@@ -174,6 +179,7 @@ If you're using the Ruby app template, then you need to complete the following s
       notify: flags.notify,
       graphiqlPort: flags['graphiql-port'],
       graphiqlKey: flags['graphiql-key'],
+      host: flags.host,
       tunnel: tunnelMode,
     }
 
--- a/packages/app/src/cli/services/dev.ts
+++ b/packages/app/src/cli/services/dev.ts
@@ -69,6 +69,7 @@ export interface DevOptions {
   notify?: string
   graphiqlPort?: number
   graphiqlKey?: string
+  host: string
 }
 
 export async function dev(commandOptions: DevOptions) {
--- a/packages/app/src/cli/services/dev/processes/setup-dev-processes.test.ts
+++ b/packages/app/src/cli/services/dev/processes/setup-dev-processes.test.ts
@@ -1,4 +1,4 @@
-import {DevConfig, setupDevProcesses, startProxyServer} from './setup-dev-processes.js'
+import {DevConfig, setupDevProcesses, proxyService} from './setup-dev-processes.js'
 import {subscribeAndStartPolling} from './app-logs-polling.js'
 import {sendWebhook} from './uninstall-webhook.js'
 import {WebProcess, launchWebProcess} from './web.js'
@@ -95,6 +95,7 @@ describe('setup-dev-processes', () => {
       commandConfig: new Config({root: ''}),
       skipDependenciesInstallation: false,
       tunnel: {mode: 'auto'},
+      host: 'localhost',
     }
     const network: DevConfig['network'] = {
       proxyUrl: 'https://example.com/proxy',
@@ -281,7 +282,7 @@ describe('setup-dev-processes', () => {
     expect(proxyServerProcess).toMatchObject({
       type: 'proxy-server',
       prefix: 'proxy',
-      function: startProxyServer,
+      function: proxyService,
       options: {
         port: 444,
         localhostCert: {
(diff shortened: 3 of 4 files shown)


theo-styles-radiant:
any eta when this will be looked at? It's blocking me working on an app that uses docker

lmcrean:
can someone review this PR please?

packages/app/src/cli/commands/app/dev.ts
```diff
@@ -87,6 +87,11 @@ If you're using the Ruby app template, then you need to complete the following s
       default: false,
       exclusive: ['tunnel-url'],
     }),
+    host: Flags.string({
+      description: 'Set which network interface the web server listens on. The default value is localhost.',
+      env: 'SHOPIFY_FLAG_HOST',
+      default: 'localhost',
+    }),
```
karreiro, line 94: The `shopify app dev` command can manage different types of extensions. When we introduced support for theme app extensions, we were advised to keep theme app extension-specific properties within the `'theme-app-extension-*` namespace.

With this in mind, renaming this flag to `'theme-app-extension-host'` seems appropriate. Although it's a bit more verbose, it maintains consistency with `'theme-app-extension-port'`.

Additionally, I've suggested setting the default to `127.0.0.1` to ensure alignment with `shopify theme dev`.

```suggestion
    'theme-app-extension-host': Flags.string({
      description: 'Set which network interface the web server listens on. The default value is 127.0.0.1.',
      env: 'SHOPIFY_FLAG_HOST',
      default: '127.0.0.1',
    }),
```

isaacroldan: This feature affects (and adds value to) all extensions when `dev`ing, so I think it makes more sense to keep it as `host`

lmcrean:
Thank you for the professional notes @karreiro  @isaacroldan 

Just confirming that this is in my pipeline, it appears to be nearly there, I'll be engaging with it during end of December/ early Jan

lmcrean:
still relevant -- looking at it now actually

packages/app/src/cli/commands/app/dev.ts
lmcrean: @karreiro @isaacroldan Thank you both for the feedback.  

After reviewing, I'll keep it as `--host` in this specific PR per @isaacroldan's point - this affects the main proxy server for all extensions, not just theme app extensions.

I've updated the default to `127.0.0.1` for consistency with `shopify theme dev`.

lmcrean:
Ready for review!

Kept `--host` per @isaacroldan's point about it affecting all extensions, and changed the default to `127.0.0.1` for consistency with `shopify theme dev`. 

Also fixed the lint errors and added a changeset.

lmcrean:
Still relevant, can this be reopened please?

PR #6403, opened 2025-09-15
Move app logs to app's `.shopify` directory

lopert:

Part of https://github.com/shop/issues-shopifyvm/issues/571

In order to make function logs more accessible, we want to move them to the app directory directly, instead of in the system logs.

Moves logs from one place to another.

### How to test your changes?
Logs should be created during `shopify app dev`
And accessible using `shopify app function replay`

- run dev `shopify app dev`
- execute function runs
- look in your app directory, there should be a new folder with the appname-shortId, and the run logs
- try replaying a few with `shopify app function replay`

### Measuring impact

### Checklist

### DIFF
--- a/packages/app/src/cli/services/app-logs/dev/write-app-logs.ts
+++ b/packages/app/src/cli/services/app-logs/dev/write-app-logs.ts
@@ -1,6 +1,6 @@
 import {AppLogData} from '../types.js'
 import {toFormattedAppLogJson} from '../utils.js'
-import {joinPath, dirname} from '@shopify/cli-kit/node/path'
+import {joinPath} from '@shopify/cli-kit/node/path'
 import {mkdir, writeFile} from '@shopify/cli-kit/node/fs'
 import {randomUUID} from '@shopify/cli-kit/node/crypto'
 import {Writable} from 'stream'
@@ -25,17 +25,13 @@ export const writeAppLogsToFile = async ({
   logsDir: string
 }): Promise<AppLogFile> => {
   const identifier = randomUUID().substring(0, 6)
-
   const formattedTimestamp = formatTimestampToFilename(appLog.log_timestamp)
   const fileName = `${formattedTimestamp}_${appLog.source_namespace}_${appLog.source}_${identifier}.json`
   const logContent = toFormattedAppLogJson({appLog, appLogPayload, prettyPrint: true, storeName})
-
-  // Write to the provided logs directory
   const fullOutputPath = joinPath(logsDir, fileName)
 
   try {
-    // Ensure parent directory exists and write the file
-    await mkdir(dirname(fullOutputPath))
+    await mkdir(logsDir)
     await writeFile(fullOutputPath, logContent)
 
     return {


packages/app/src/cli/models/app/app.ts
```diff
@@ -445,6 +446,14 @@ export class App<
     return this._hiddenConfig
   }
 
+  getLogsDir() {
+    // Create directory name using app name + partial client_id for uniqueness
+    const clientId = String(this.configuration.client_id || 'unknown')
+    const shortClientId = clientId.substring(0, 8)
```
lopert: This might not be needed, could just use the full client_id.
I just wanted to have the app name be part of the directory to make it easy to sort through.
... then again since these logs are all app specific, and the dir is in that app's dir, this doesn't actually do anything 🤦 

I think this could just be `./shopify/logs`, I'll update it to that

packages/app/src/cli/services/app-logs/dev/write-app-logs.ts
```diff
@@ -13,26 +13,31 @@ interface AppLogFile {
...
-  const fullOutputPath = joinPath(getLogsDir(), path)
+  const logContent = toFormattedAppLogJson({appLog, appLogPayload, prettyPrint: true, storeName})
+
+  // Write to the provided logs directory
+  const fullOutputPath = joinPath(logsDir, fileName)
 
   try {
-    await writeLog(path, toFormattedAppLogJson({appLog, appLogPayload, prettyPrint: true, storeName}))
+    // Ensure parent directory exists and write the file
+    await mkdir(dirname(fullOutputPath))
```
isaacroldan: no need to do `dirname` here, just use `logsDir` :)

isaacroldan (approved):

packages/app/src/cli/services/dev/processes/app-logs-polling.ts
```diff
@@ -76,20 +76,20 @@ export const subscribeAndStartPolling: DevProcessFunction<SubscribeAndStartPolli
       stdout,
     )
 
-    const apiKey = appLogsSubscribeVariables.apiKey
-    await createLogsDir(apiKey)
+    const logsDir = localApp.getLogsDir()
+    await mkdir(logsDir)
```
saga-dasgupta, line 80: Do you need to ensure that this doesn't already exists? It might throw if it does already exist.

lopert: Hmm, I don't think I've run into any issues starting/stopping the app dev and adding logs to an existing dir. I'll confirm.

lopert: Confirmed, this doesn't error when the directory already exists, and the logs go to it as expected.
I took a quick look through [the docs](https://nodejs.org/api/fs.html#fspromisesmkdirpath-options) to see if this behaviour was mentioned, but didn't find anything relevant.

[mkdir is what we called before when using the system logs location](https://github.com/Shopify/cli/blob/5d7e03761d2207812ff1e4975d58b2284c5fa350/packages/cli-kit/src/public/node/logs.ts#L9-L11).

saga-dasgupta (approved):

PR #6405, opened 2025-09-15
Inspect Wasm module to determine what trampoline to run

jeffcharles:

There are now multiple versions of the Shopify Functions Wasm trampoline. Version 1 translates imports from `shopify_function_v1` and version 2 translates imports from `shopify_function_v2`. So we need the CLI to select the appropriate trampoline to use.

When building a Function, we inspect the imports of the Wasm module and, if necessary, download and run the appropriate version of the trampoline.

### How to test your changes?

I'd suggest setting up a Function to use WAT by creating a Rust Function and then modifying the `shopify.extension.toml`. Note you'll need to have `wasm-tools` installed with `cargo install --locked wasm-tools`.

E.g.,

`shopify.extension.toml`:
```toml
api_version = "2025-07"

[[extensions]]
name = "t:name"
handle = "cart-checkout-validation"
type = "function"
uid = "9beb9559-a34a-eca2-aa94-2cf3baa77d52afcba76e"
description = "t:description"

  [[extensions.targeting]]
  target = "cart.validations.generate.run"
  input_query = "src/cart_validations_generate_run.graphql"
  export = "cart_validations_generate_run"

  [extensions.build]
  command = "wasm-tools parse module.wat -o module.wasm"
  path = "module.wasm"
  watch = ["module.wat"]
```

No function imports test case `module.wat`:
```wat
(module
  (import "foo" "shopify_function_input_get" (func $input_get (param i32) (result i64)))
  (memory 1)
  (func $cart_validations_generate_run
    i32.const 0
    i32.const 0
    call $input_get
    i64.store
  )
  (export "cart_validations_generate_run" (func $cart_validations_generate_run))
)
```

Run `pnpm shopify app function build --path <path_to_function>`, run `wasm-tools print <path_to>/module.wasm | less` and see that the module looks the same.

Update `module.wat` to use `shopify_function_v1`:
```wat
(module
  (import "shopify_function_v1" "shopify_function_input_get" (func $input_get (param i32) (result i64)))
  (memory 1)
  (func $cart_validations_generate_run
    i32.const 0
    i32.const 0
    call $input_get
    i64.store
  )
  (export "cart_validations_generate_run" (func $cart_validations_generate_run))
)
```

Run `pnpm shopify app function build --path <path_to_function>`, run `wasm-tools print module.wasm | less` and see that the `shopify_function_input_get` import has been changed to `_shopify_function_input_get`.

Update `module.wat` to use `shopify_function_v2`:
```wat
(module
  (import "shopify_function_v2" "shopify_function_input_get" (func $input_get (param i32) (result i64)))
  (memory 1)
  (func $cart_validations_generate_run
    i32.const 0
    i32.const 0
    call $input_get
    i64.store
  )
  (export "cart_validations_generate_run" (func $cart_validations_generate_run))
)
```

Run `pnpm shopify app function build --path <path_to_function>`, run `wasm-tools print module.wasm | less` and see that the `shopify_function_input_get` import has been changed to `_shopify_function_input_get`.

### Measuring impact

### Checklist

### DIFF
--- a/packages/app/src/cli/services/function/build.test.ts
+++ b/packages/app/src/cli/services/function/build.test.ts
@@ -297,6 +297,21 @@ describe('runTrampoline', () => {
     expect(exec).not.toHaveBeenCalled()
   })
 
+  test('does not run trampoline if Wasm module is invalid', async () => {
+    // Given
+    const ourFunction = await testFunctionExtension()
+    const modulePath = ourFunction.outputPath
+    const invalidWasmModule = Buffer.from([])
+    vi.mocked(readFileSync).mockReturnValue(invalidWasmModule)
+
+    // When
+    const got = runTrampoline(modulePath)
+
+    // Then
+    await expect(got).resolves.toBeUndefined()
+    expect(exec).not.toHaveBeenCalled()
+  })
+
   test('runs v1 trampoline on v1 module', async () => {
     // Given
     const ourFunction = await testFunctionExtension()


isaacroldan (approved):

packages/app/src/cli/services/function/build.test.ts
```diff
@@ -45,6 +49,19 @@ const derivedDeps = {
 
 const app = testApp({dotenv: {variables: {VAR_FROM_ENV_FILE: 'env_file_var'}, path: ''}})
 
+function createWasmModule(importModuleName: string): Buffer {
+  const importsModuleNameBytes = Array.from(importModuleName).map((char) => char.charCodeAt(0))
+  // Module looks like:
+  // (module
+  //   (import "${importModuleName}" "foo" (func))
+  // )
```
jacobsteves, line 57: Will it always look like this even if you build your own wasm from a random source?

jeffcharles: What do you mean by _it_? If a Wasm module is using the Wasm API, it will have to include at least one Wasm import from a `shopify_function_v` module. There would be other things in the module but they're not relevant for determining which modules are being imported.

jacobsteves: This is just a test, so I was just wondering if we needed to also test what happens when the wasm is not what we expect. They could build their own wasm without the wasm API for example, it could be unrelated to Shopify Functions.

Its fine if this is already covered elsewhere though!

jeffcharles: The case where they're not using the Wasm API should be covered by [this test case](https://github.com/Shopify/cli/pull/6405/files#diff-3325f86efa23ca32781d75c3454d6b174f358b4d0f0db692d2a73fc3d84a7871R286-R298).

packages/app/src/cli/services/function/build.ts
```diff
@@ -268,6 +282,19 @@ export async function runTrampoline(modulePath: string) {
   await exec(command, args)
 }
 
+async function importedWasmModules(modulePath: string): Promise<string[]> {
+  const moduleBytes = new Uint8Array(readFileSync(modulePath))
+  const isValid = WebAssembly.validate(moduleBytes)
+  if (!isValid) {
+    return []
+  }
```
jacobsteves, line 290: Could we add an extra test case for an invalid module?

PR #6407, opened 2025-09-16
Abort theme command if path doesn't exist

graygilmore:

We're seeing quite a few unhandled errors due to the command being run on a path that doesn't exist. Instead of blowing up let's catch the missing path early and let the user know.

Modifies `ThemeCommand` to check the path and to see if it exists. We throw two different errors depending on whether the user explicitly provided the path or not.

### How to test your changes?

> [!NOTE]
> Right now this only works for commands that have been migrated to multi-environment support but soon that will be all of them!

```sh
npm i -g --@shopify:registry=https://registry.npmjs.org @shopify/cli@0.0.0-snapshot-20250916000226
shopify theme info --path foobarbaz
```

You should see an error:

```
➜  <theme> git:(main) shopify theme info --path foooooo
╭─ error ────────────────────────────────────────────────────────────────────────────────────╮
│                                                                                            │
│  A path was explicitly provided and does not exist. Please check the path and try again:   │
│  /Users/graygilmore/repos/<theme>/foooooo                                                  │
│                                                                                            │
╰────────────────────────────────────────────────────────────────────────────────────────────╯
```

### DIFF
--- a/.changeset/big-ducks-joke.md
+++ b/.changeset/big-ducks-joke.md
@@ -1,0 +1,5 @@
+---
+'@shopify/theme': minor
+---
+
+Abort theme commands if path doesn't exist
--- a/packages/theme/src/cli/flags.ts
+++ b/packages/theme/src/cli/flags.ts
@@ -1,6 +1,8 @@
 import {Flags} from '@oclif/core'
 import {normalizeStoreFqdn} from '@shopify/cli-kit/node/context/fqdn'
 import {resolvePath, cwd} from '@shopify/cli-kit/node/path'
+import {fileExistsSync} from '@shopify/cli-kit/node/fs'
+import {renderError} from '@shopify/cli-kit/node/ui'
 
 /**
  * An object that contains the flags that
@@ -10,7 +12,21 @@ export const themeFlags = {
   path: Flags.string({
     description: 'The path where you want to run the command. Defaults to the current working directory.',
     env: 'SHOPIFY_FLAG_PATH',
-    parse: async (input) => resolvePath(input),
+    parse: async (input) => {
+      const resolvedPath = resolvePath(input)
+
+      if (fileExistsSync(resolvedPath)) {
+        return resolvedPath
+      }
+
+      // We can't use AbortError because oclif catches it and adds its own
+      // messaging that breaks our UI
+      renderError({
+        headline: "A path was explicitly provided but doesn't exist.",
+        body: [`Please check the path and try again: ${resolvedPath}`],
+      })
+      process.exit(1)
+    },
     default: async () => cwd(),
     noCacheDefault: true,
   }),
--- a/packages/theme/src/cli/utilities/theme-command.ts
+++ b/packages/theme/src/cli/utilities/theme-command.ts
@@ -14,6 +14,7 @@ import {
   renderError,
 } from '@shopify/cli-kit/node/ui'
 import {AbortController} from '@shopify/cli-kit/node/abort'
+import {AbortError} from '@shopify/cli-kit/node/error'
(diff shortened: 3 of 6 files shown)


karreiro (approved):
Thank you, @graygilmore!

> Right now this only works for commands that have been migrated to multi-environment support but soon that will be all of them!

Could we add this logic to [`flags.ts`](https://github.com/Shopify/cli/blob/32126d98e2fcff778998d8a6e8e69c08678b71fa/packages/theme/src/cli/flags.ts#L13) too handle that in all commands?

Thanks again for this PR!

.changeset/big-ducks-joke.md
```diff
@@ -0,0 +1,5 @@
+---
+'@shopify/theme': minor
```
aswamy, line 2: we can even have this as a patch since it's a bugfix

packages/theme/src/cli/utilities/theme-command.ts
```diff
@@ -100,6 +103,16 @@ export default abstract class ThemeCommand extends Command {
...
 
+      if (explicitPath && !fileExistsSync(explicitPath)) {
+        throw new AbortError(
+          `A path was explicitly provided but doesn't exist. Please check the path and try again: ${flags.path}`,
+        )
+      }
+
+      if (flags.path && !fileExistsSync(flags.path)) {
+        throw new AbortError(`Path does not exist: ${flags.path}`)
+      }
```
aswamy: Similar question to Guilherme's, but if we add it to `flags.ts` i dont _THINK_ we need it here as well since we're verifying the flag when it's passed, rather than when the command is run.

But does this logic future proof it so that we can support multiple paths based on environments?

I updated flags.ts and it seemed to work in a similar manner:
```
    parse: async (input) => {
      const resolvedPath = resolvePath(input)

      if (fileExistsSync(resolvedPath)) {
        return resolvedPath
      }

      throw new AbortError(
        `A path was explicitly provided but doesn't exist. Please check the path and try again: ${resolvedPath}`,
      )
    },
```
image

graygilmore: So long as `flags.ts` supports also handling the default location then we should be good! I'll give it a go.

aswamy: > handling the default location

oooo i can understand why you want it in the run method now - i think flags only cares about what is passed into the CLI; not what is set at runtime.

graygilmore: I ended up putting a little bit of logic in both!

packages/theme/src/cli/flags.ts
```diff
@@ -10,7 +12,17 @@ export const themeFlags = {
...
     env: 'SHOPIFY_FLAG_PATH',
-    parse: async (input) => resolvePath(input),
+    parse: async (input) => {
+      const resolvedPath = resolvePath(input)
+
+      if (fileExistsSync(resolvedPath)) {
+        return resolvedPath
+      }
+
+      throw new AbortError(
```
dejmedus: I wonder if DIYing an error here would bypass the oclif formatting? like:

```ts
     renderError({
        headline: "A path was explicitly provided but doesn't exist.",
        body: [`Please check the path and try again: ${resolvedPath}`],
      })
      process.exit(1)
```

graygilmore: Good call! https://github.com/Shopify/cli/compare/eb62fa49f5e9b182e86e0c45197c0ccc5cf08faf..2315c76ce66d1e334e96b33a2911e2e18658d928#diff-e1c5690cad583c5ca92fc5b68ff18b9101c3763dc0fdee8db29cedcfb719c36bR22-R28

PR #6413, opened 2025-09-18
Improve upload reliability with accurate bytesize calculation

EvilGenius13:

Our theme uploader could create asset batches that were larger than the 1MB limit we set. This was caused by inaccurate calculation of the files containing multi-byte UTF-8 characters such as emojis or accented letters (e.g `é`, `ü`, `ñ`). We may see this more often with internationalized content

EX: `🍕🍕🍕🍕🍕` under the current calculation is measured as `10` characters but is actually 20 bytes.

- For all text based assets (.liquid, .json, .js) we use `Buffer.byteLength` to get the true byte size of the content.
- For binary assets (.png, .jpg) we continue the way it was before - calculating the size based on the length of the Base64-encoded string.

The changes should allow us to pack batches more reliably.

### Measuring impact

### Checklist

### DIFF
--- a/packages/theme/src/cli/utilities/theme-uploader.ts
+++ b/packages/theme/src/cli/utilities/theme-uploader.ts
@@ -357,10 +357,14 @@ function calculateLocalChecksums(localThemeFileSystem: ThemeFileSystem): Checksu
   const checksums: ChecksumWithSize[] = []
 
   localThemeFileSystem.files.forEach((file, key) => {
+    // Text files: use UTF-8 byte count
+    // Binary files: use base64 length
+    const size = file.value ? Buffer.byteLength(file.value, 'utf8') : file.attachment?.length ?? 0
+
     checksums.push({
       key,
       checksum: file.checksum,
-      size: file.stats?.size ?? 0,
+      size,
     })
   })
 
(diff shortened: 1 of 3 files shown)


packages/theme/src/cli/utilities/theme-uploader.ts
```diff
@@ -357,10 +357,14 @@ function calculateLocalChecksums(localThemeFileSystem: ThemeFileSystem): Checksu
...
 
   localThemeFileSystem.files.forEach((file, key) => {
+    // Text files: use UTF-8 byte count
+    // Binary files: use base64 length
+    const size = file.value ? Buffer.byteLength(file.value, 'utf8') : file.attachment?.length ?? 0
+
     checksums.push({
       key,
       checksum: file.checksum,
-      size: file.stats?.size ?? 0,
```
graygilmore, line 363: Are we kind of saying that `file.stats` is useless? I wonder if that's something we can fix directly?

EvilGenius13: It is used in one other area for caching headers, and they have a similar comment. I would leave that as is.
`local-assets.ts`
```js
return serveStatic(event, {
      getContents: () => fileContent,
      // Note: stats.size is the length of the base64 string for attachments,
      // not the real length of the file. Use the Buffer length instead:
      getMeta: () => ({type: mimeType, size: fileContent.length, mtime: file.stats?.mtime}),
    })
```

PR #6427, opened 2025-09-24
fix issue where multiple ensureAuthenticated promises can be called

EvilGenius13:
# DO NOT MERGE
⏸️ PR IS ON HOLD ⏸️ 

We are still getting reports of users coming back to the CLI after leaving it running for a long period of time and seeing multiple authentication prompts / open browser tabs.

Added a simple authentication lock to prevent concurrent authentication flows. When `ensureAuthenticated` is called, we check if authentication is already in progress. If one is in progress, we return the existing promise instead of starting a new one. All concurrent calls to `ensureAuthenticated` will share the same auth result.

### How to test your changes?

- Pull down the branch
- Build the branch
- Run tests in `session.test.ts`
- Run `theme dev` and leave it open for as long as you're willing to see if there are any errors in re-auth (sessions refresh every 30 minutes). Also look to see if you are prompted multiple times or browser tab is opened multiple times. We should only see one if a session expires and needs manual prompting.

### Measuring impact

### Checklist

### DIFF
--- a/packages/cli-kit/src/private/node/session.ts
+++ b/packages/cli-kit/src/private/node/session.ts
@@ -118,7 +118,7 @@ type AuthMethod = 'partners_token' | 'device_auth' | 'theme_access_token' | 'cus
 
 let userId: undefined | string
 let authMethod: AuthMethod = 'none'
-let authenticationPromise: Promise<OAuthSession> | null = null
+const authenticationInProgress = new Map<string, Promise<OAuthSession>>()
 
 /**
  * Retrieves the user ID from the current session or returns 'unknown' if not found.
@@ -197,15 +197,17 @@ export async function ensureAuthenticated(
   _env?: NodeJS.ProcessEnv,
   {forceRefresh = false, noPrompt = false, forceNewSession = false}: EnsureAuthenticatedAdditionalOptions = {},
 ): Promise<OAuthSession> {
-  // If an authentication is already in progress, wait for it to complete and return its result.
-  if (authenticationPromise) {
-    outputDebug('An authentication process is already in progress. Waiting for it to complete...')
-    return authenticationPromise
+  const store = applications.adminApi?.storeFqdn ?? 'no-store'
+
+  // If an authentication is already in progress for this store, wait for it to complete and return its result.
+  const existingAuthPromise = authenticationInProgress.get(store)
+  if (existingAuthPromise) {
+    outputDebug(`Authentication already in progress for store ${store}, waiting...`)
+    return existingAuthPromise
   }
 
-  authenticationPromise = (async () => {
+  const authPromise = (async (): Promise<OAuthSession> => {
     const fqdn = await identityFqdn()
-
     const previousStoreFqdn = applications.adminApi?.storeFqdn
     if (previousStoreFqdn) {
       const normalizedStoreName = await normalizeStoreFqdn(previousStoreFqdn)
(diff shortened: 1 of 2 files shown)


packages/cli-kit/src/private/node/session.ts
```diff
@@ -196,83 +197,99 @@ export async function ensureAuthenticated(
...
-  return tokens
+    setLastSeenAuthMethod(envToken ? 'partners_token' : 'device_auth')
+    setLastSeenUserIdAfterAuth(tokens.userId)
+    return tokens
+  })()
+
+  try {
+    return await authenticationPromise
+  } finally {
+    // Release the lock.This can not be a race condition because we are waiting for the promise to resolve/reject.
```
graygilmore:
```suggestion
    // Release the lock. This can not be a race condition because we are waiting for the promise to resolve/reject.
```

graygilmore (approved):
Makes sense to me! Will snapit and let it run in a session for a while 😄

isaacroldan:
Thought this was already fixed 🤔 
The refresh-token methods already have this functionality (only 1 refresh at a time), 
why (or where) is `theme dev` calling `ensureAuthenticated` multiple times?
If you call `ensureAuthenticated` with `no-prompt`, then it won't ever open the browser on token refresh, is this still happening?

isaacroldan:
I'll review this week 🙏

packages/cli-kit/src/private/node/session.ts
```diff
@@ -196,83 +197,100 @@ export async function ensureAuthenticated(
...
+  const store = applications.adminApi?.storeFqdn ?? 'no-store'
 
-  const previousStoreFqdn = applications.adminApi?.storeFqdn
-  if (previousStoreFqdn) {
-    const normalizedStoreName = await normalizeStoreFqdn(previousStoreFqdn)
-    if (previousStoreFqdn === applications.adminApi?.storeFqdn) {
-      applications.adminApi.storeFqdn = normalizedStoreName
-    }
+  // If an authentication is already in progress for this store, wait for it to complete and return its result.
+  const existingAuthPromise = authenticationInProgress.get(store)
```
isaacroldan, line 203: Is there a use case where we want to support multiple concurrent authentication flows for different stores? 🤔 

The way the session tokens are stored, multiple auths for different stores will overwrite each other when storing the full session.

We can only have a "full session" per identity fqdn. (line 266)

EvilGenius13: I added it this way for two reasons but it was mostly a safeguard
- Make sure we never end up in a situation with running theme commands in multi-environment mode where there is a possibility overwrite two different auth flows (ie: `shopify theme push -e my_store -e my_other_store`)
- The other is why I believe we sometimes get the cascading auth prompts. Somehow the session gets dirtied, goes for a full re-auth, and another area triggers a re-auth at the same time. This ensures if for any reason two different methods call for re-auth, we just return the same session promise.

isaacroldan: Right, but you are allowing here multiple auth promises to run at the same time because the "index" is the store name 🤔 So in that case you mention, you'll get into a race condition of one auth overwriting the other no?

EvilGenius13: Sorry for the late reply!
There won't be a case where we can hit a race condition. Let's say a method asks for auth, we set the store as a key and then we create a "lock" for that promise while authentication is happening. If another method were to ask for auth from the same store, we will give that method the same auth promise currently happening for the first method. Once auth completes, we remove the lock. Since we are single threaded there should never be a time where a race condition happens. In multi-env mode, each store has it's own key and can be locked the same way during an auth flow.

isaacroldan: > In multi-env mode, each store has it's own key and can be locked the same way during an auth flow.

That's what i mean, each store has its own key, each store could trigger its own auth flow which would overwrite the main identity token every time. We can discuss in slack if its easier :)

EvilGenius13: Had a great chat with @isaacroldan this morning. I am going to pause the PR for now as we can make some more foundational changes to auth before continuing down this path.

PR #6439, opened 2025-09-26
Enable traversal of ui-extension files during type generation 

JoviDeCroock:
Backports https://github.com/Shopify/cli/pull/6385

### DIFF
--- a/.changeset/clean-squids-report.md
+++ b/.changeset/clean-squids-report.md
@@ -1,0 +1,5 @@
+---
+'@shopify/app': patch
+---
+
+Add support for generating types for files imported by the main extension file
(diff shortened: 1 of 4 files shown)


gonzaloriestra (approved):

.changeset/clean-squids-report.md
```diff
@@ -0,0 +1,5 @@
+---
+'@shopify/app': minor
```
karreiro: I think we could update this to be a patch change:
```suggestion
'@shopify/app': patch
```

karreiro: Oops! It's already done!

PR #6446, opened 2025-09-26
Add slice_name to error handler

MitchDickinson:
Add `slice_name` custom metadata attribute to bugsnag reporter. This will let us filter on "slice_name" in the Observe UI:

Screenshot 2025-09-26 at 10 25 03 AM

**To test this:**
1. You can edit any command to force an error 
```ts
// anywhere in the code
if (1==1) {
  throw new Error('test')
}
```
2. Comment out the `if (isLocalEnvironment() || settings.debug)` check in the `sendErrorToBugsnag` function

Then you can trigger any command and view the result error in Observe. [Example](https://observe.shopify.io/a/observe/errors/8769927911199070393/events/1143130c-c07a-4e2f-9f54-360b990b57c6?r=%7B%22from%22%3A%22now-1h%22%2C%22to%22%3A%22now%22%7D&tz=browser&f=%7B%22resource.service.name%22%3A%5B%5D%2C%22slice_name%22%3A%5B%5D%7D&n=test&h=1)

### DIFF
--- a/packages/cli-kit/src/public/node/error-handler.ts
+++ b/packages/cli-kit/src/public/node/error-handler.ts
@@ -21,26 +21,9 @@ import StackTracey from 'stacktracey'
 import Bugsnag, {Event} from '@bugsnag/js'
 import {realpath} from 'fs/promises'
 
-// Allowed slice names for error analytics grouping. This is populated from the
-// loaded OCLIF config at runtime.
-let ALLOWED_SLICE_NAMES: Set<string> | undefined
-
-export function computeAllowedSliceNames(config: Interfaces.Config): Set<string> {
-  const names = new Set<string>()
-  const commands = (config.commands ?? []) as Array<{id?: string}>
-  for (const cmd of commands) {
-    // Command IDs are typically colon-separated (e.g. app:build). We take the first token.
-    const id = (cmd as unknown as {id?: string}).id
-    if (!id) continue
-    const first = id.split(':')[0]?.trim()
-    if (first) names.add(first)
-  }
-  return names
-}
-
-export function initializeAllowedSliceNames(config: Interfaces.Config): void {
-  ALLOWED_SLICE_NAMES = computeAllowedSliceNames(config)
-}
+// Allowed slice names for error analytics grouping.
+// Hardcoded list per product slices to keep analytics consistent.
+const ALLOWED_SLICE_NAMES = new Set<string>(['app', 'theme', 'hydrogen', 'store'])
 
 export async function errorHandler(
   error: Error & {exitCode?: number | undefined},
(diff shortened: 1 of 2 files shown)


MitchDickinson:
* **#6446** <a href="https://app.graphite.dev/github/pr/Shopify/cli/6446?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Shopify/cli/6446?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/cli-kit/src/public/node/error-handler.ts
```diff
@@ -21,6 +21,9 @@ import StackTracey from 'stacktracey'
 import Bugsnag, {Event} from '@bugsnag/js'
 import {realpath} from 'fs/promises'
 
+// Allowed slice names for error analytics grouping; default to 'cli' when unknown
+const ALLOWED_SLICE_NAMES = new Set(['app', 'theme', 'hydrogen', 'store'])
```
MitchDickinson: This would ideally be dynamic, but not sure how to do it yet. It could always come from a future change.

gonzaloriestra: I can't think of a reliable way to do that, because there are many packages and exceptions... I think it's fine like this.

MitchDickinson:
hey @gonzaloriestra  I should have waited to tag you but I made one more change which is to make the `ALLOWED_SLICE_NAMES` dynamic (thanks Claude). Mind having another quick look? Apologies - I should have submitted as a different commit or PR, I didn't know you would have a look so quickly. My bad

isaacroldan:
Why do we need to precompute the allowed slices names? can we just include whatever the command is?

gonzaloriestra:
@isaacroldan we want to set the command "topic" as the slice. We discussed that we would like to have: `app`, `theme`, `hydrogen`, and a default category `cli` for the other commands like `version`, `auth logout`, etc.

@MitchDickinson that dynamic behavior is generating the following list of slices, which is not what we want: `app, auth, cache, commands, config, debug, demo, docs, help, hydrogen, kitchen-sink, notifications, plugins, search, store, theme, upgrade, version, webhook`. I think the static list is enough, it should hardly ever change.

MitchDickinson:
OK - I can move back to a static list. I wasn't sure if we'd want more granularity here. I wondered if more granularity wasn't a _bad_ thing, it would simply give us more options to filter in error management. 

But thinking more - I guess it depends if we are thinking about this as a `slice` per the definition of core (a large area of code with a relationship to its code owner) or if we are simply using `slice` as a proxy for `command` here.

I guess since it's called `slice`, we should probably treat it as such, and so the static list makes more sense.

MitchDickinson:
OK - should be back to the state it was in now (static list). Sorry about the extra runaround there.

packages/cli-kit/src/public/node/error-handler.ts
```diff
@@ -126,6 +130,14 @@ export async function sendErrorToBugsnag(
         const eventHandler = (event: Event) => {
           event.severity = 'error'
           event.unhandled = unhandled
+          // Attach command metadata so we know which CLI command triggered the error
+          const {commandStartOptions} = metadata.getAllSensitiveMetadata()
+          const {startCommand} = commandStartOptions ?? {}
+          if (startCommand) {
+            const firstWord = startCommand.trim().split(/\s+/)[0] ?? 'cli'
+            const sliceName = ALLOWED_SLICE_NAMES.has(firstWord) ? firstWord : 'cli'
```
isaacroldan, line 138: I believe that in the public metadata we are already saving the "topic", so this could be just:
```
const {cmd_all_topic} = metadata.getAllPublicMetadata()
const sliceName = ALLOWED_SLICE_NAMES.has(cmd_all_topic) ? cmd_all_topic : 'cli'
```

isaacroldan: Nevermind, for commands with multiple words the topic includes all but the last one

isaacroldan (approved):

PR #6451, opened 2025-09-29
Fix theme analytics logging even when a command fails

EvilGenius13:
We are not collecting analytics when a theme command fails. This is because we put our logging after our `await command` and on failure, the promise stops and we never get to our logging method.

Wrap theme commands in `try/finally` blocks to ensure analytics are logged regardless of command success or failure. 

### How to test your changes?

You can throw an error in a theme command on main and try to console log the analytics event in `theme-command`
Build this branch and do the same, you should see our logs.

### Measuring impact

### Checklist

### DIFF
--- a/packages/theme/src/cli/utilities/theme-command.ts
+++ b/packages/theme/src/cli/utilities/theme-command.ts
@@ -88,8 +88,11 @@ export default abstract class ThemeCommand extends Command {
         throw new AbortError(`Path does not exist: ${flags.path}`)
       }
 
-      await this.command(flags, session, false, args)
-      await this.logAnalyticsData(session)
+      try {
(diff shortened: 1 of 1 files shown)


EvilGenius13:
> 🫰✨ **Thanks @EvilGenius13! Your snapshot has been published to npm.**
> 
> Test the snapshot by installing your package globally:
> 
> ```shell
> npm i -g --@shopify:registry=https://registry.npmjs.org @shopify/cli@0.0.0-snapshot-20250929200550
> ```
> 
> Caution
> 
> After installing, validate the version by running just `shopify` in your terminal. If the versions don't match, you might have multiple global instances installed. Use `which shopify` to find out which one you are running and uninstall it.

This was my original snap build when I was running logs

PR #6458, opened 2025-09-30
Set the user ID for Observe

gonzaloriestra:

We are not setting the user ID for Observe, so it's using the IP as a fallback:

userId

More info: https://vault.shopify.io/page/Error-Management-FAQ~NUCh.md#how-are-unique-users-calculated

Properly set the user ID if present

30-19-04cu5-3auz4

### How to test your changes?

- `npm i -g --@shopify:registry=https://registry.npmjs.org @shopify/cli@0.0.0-snapshot-20250930111105`
- `shopify wronggg`
- Look for your error on [Observe](https://observe.shopify.io/a/observe/errors?r=%7B%22from%22%3A%22now-30m%22%2C%22to%22%3A%22now%22%7D&tz=browser&f=%7B%22resource.service.name%22%3A%5B%22cli%22%5D%2C%22handled%22%3A%5B%22false%22%5D%7D&ord=%7B%22order%22%3A%22desc%22%2C%22op%22%3A%22count%22%7D&n=wronggg) and check the `__observe_session_stability_user` field

### Measuring impact

### Checklist


packages/cli-kit/src/public/node/error-handler.ts
```diff
@@ -121,11 +122,13 @@ export async function sendErrorToBugsnag(
 
     if (report) {
       initializeBugsnag()
+      const userId = await getLastSeenUserIdAfterAuth()
```
gonzaloriestra, line 129: This is returning `unknown` when not found. For example, when running `shopify version` unauthenticated.

@MitchDickinson should we better send `undefined` in that case so Observe sets the IP in that case? Not sure what's best...

MitchDickinson: Falling back to the IP if the user ID is not found makes sense.

PR #6469, opened 2025-10-01
Use a global cache for the sessions to avoid multiple instances

gonzaloriestra:

We are caching the current session for each Developer Platform client (Partners and App Management) as an instance variable, but we are creating instances from multiple places.

That could lead in multiple login requests during the same command. I think it's not happening right now, but it could happen eventually after some refactor.

- Moves the cached sessions to a global variable, ensuring there is only one, even if there are multiple clients.
- Renames `PartnersSession` to `Session` because it was shared by both clients, and moves it to cli-kit

### How to test your changes?

Log in and run commands

### Measuring impact

### Checklist


packages/app/src/cli/services/context/partner-account-info.ts
```diff
@@ -2,37 +2,9 @@ import {getCurrentAccountInfo} from '../../api/graphql/current_account_info.js'
...
-}
-
-export function isUserAccount(account: AccountInfo): account is UserAccountInfo {
-  return account.type === 'UserAccount'
-}
-
-export function isServiceAccount(account: AccountInfo): account is ServiceAccountInfo {
-  return account.type === 'ServiceAccount'
-}
+export {AccountInfo, isUserAccount, isServiceAccount}
```
isaacroldan, line 7: No need to re-export them here no? Whoever is using these can directly import from cli-kit

gonzaloriestra:
Closing in favor of https://github.com/Shopify/cli/pull/6476

PR #6470, opened 2025-10-01
Record retryable requests to monorail for theme commands

EvilGenius13:

Follow up to: https://github.com/Shopify/cli/pull/6451

Now that we will be getting data for failed commands, we should also gather data during retryable network events.

We want to record retries for: 
- Transient level [retries](https://github.com/Shopify/cli/blob/45fe1ee39f8f08d6a2750977904ab01c98c9261c/packages/cli-kit/src/private/node/api.ts#L82)
- HTTP level retries (`429`, `401`..)

Add an optional property in `NetworkRetryBehaviour` that will record these kinds of events only if they are theme commands. App command logging will remain the same as it currently is.

### How to test your changes?

Running the tests is the easiest. `pnpm vitest run packages/cli-kit/src/private/node/api.test.ts`

### Measuring impact

### Checklist

### DIFF
--- a/.changeset/curvy-bears-raise.md
+++ b/.changeset/curvy-bears-raise.md
@@ -1,5 +1,0 @@
----
-'@shopify/cli-kit': patch
----
-
-Added optional property for recording retryable requests in theme commands
--- a/packages/cli-kit/src/private/node/api.ts
+++ b/packages/cli-kit/src/private/node/api.ts
@@ -18,11 +18,11 @@ export type NetworkRetryBehaviour =
   | {
       useNetworkLevelRetry: true
       maxRetryTimeMs: number
-      recordThemeCommandRetries?: boolean
+      recordCommandRetries?: boolean
     }
   | {
       useNetworkLevelRetry: false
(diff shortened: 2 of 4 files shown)


.changeset/curvy-bears-raise.md
gonzaloriestra, line 1: This is something internal, I think we don't need a changeset (they are only for user-facing changes)

packages/cli-kit/src/private/node/api.ts
```diff
@@ -17,9 +18,11 @@ export type NetworkRetryBehaviour =
   | {
       useNetworkLevelRetry: true
       maxRetryTimeMs: number
+      recordThemeCommandRetries?: boolean
```
gonzaloriestra: Even if this is only enabled for themes for now, why not just `recordCommandRetries`? Wouldn't this be useful for any kind of command?

EvilGenius13: Yes it definitely would!

PR #6474, opened 2025-10-02
[Theme] Avoid running cf turnstile in local dev

frandiox:

Fixes #0000 

### Measuring impact

### Checklist

### DIFF
--- a/packages/theme/src/cli/utilities/repl/evaluator.ts
+++ b/packages/theme/src/cli/utilities/repl/evaluator.ts
@@ -18,7 +18,7 @@ export interface EvaluationConfig {
 
 export async function evaluate(config: EvaluationConfig): Promise<string | number | undefined> {
   try {
-    return evaluateSnippet(config)
+    return await evaluateSnippet(config)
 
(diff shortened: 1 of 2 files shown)


packages/theme/src/cli/utilities/repl/evaluator.ts
```diff
@@ -17,17 +17,6 @@ export interface EvaluationConfig {
 }
 
 export async function evaluate(config: EvaluationConfig): Promise<string | number | undefined> {
-  try {
-    return evaluateSnippet(config)
-
-    // eslint-disable-next-line no-catch-all/no-catch-all, @typescript-eslint/no-explicit-any
-  } catch (error: any) {
```
frandiox, line 24: This was not awaiting for `evaluateSnippet`, so the `catch` would never run. Removing this doesn't change behavior.
I tried to add `await` but that changes the tested behavior, and it actually makes sense for this to throw rather than just log.
I tested this ins `shopify theme console --url=fake-url` and it would make sense for it to throw and exit when the URL doesn't exist.
However, perhaps we should consider if it should not throw for other use cases and update the code/unit tests in another PR.

PR #6475, opened 2025-10-03
Capture Conf update errors

isaacroldan:

Fixes #0000 

### Measuring impact

### Checklist

### DIFF
--- a/packages/cli-kit/src/public/node/local-storage.ts
+++ b/packages/cli-kit/src/public/node/local-storage.ts
@@ -1,3 +1,4 @@
+import {AbortError} from './error.js'
 import Config from 'conf'
 
 /**
@@ -27,24 +28,44 @@ export class LocalStorage<T extends {[key: string]: any}> {
    *
    * @param key - The key to set.
    * @param value - The value to set.
+   * @throws Error If the value cannot be set.
    */
   set<TKey extends keyof T>(key: TKey, value?: T[TKey]): void {
-    this.config.set(key, value)
+    try {
+      this.config.set(key, value)
+    } catch (error) {
+      throw new AbortError(this.errorMessage(error, 'set'))
+    }
   }
 
   /**
    * Delete a value from the local storage.
    *
    * @param key - The key to delete.
+   * @throws Error If the value cannot be deleted.
    */
   delete<TKey extends keyof T>(key: TKey): void {
-    this.config.delete(key)
+    try {
+      this.config.delete(key)
+    } catch (error) {
+      throw new AbortError(this.errorMessage(error, 'delete'))
+    }
   }
 
   /**
    * Clear the local storage (delete all values).
+   *
+   * @throws Error If the local storage cannot be cleared.
    */
   clear(): void {
-    this.config.clear()
+    try {
+      this.config.clear()
+    } catch (error) {
+      throw new AbortError(this.errorMessage(error, 'clear'))
+    }
+  }
+
+  private errorMessage(error: unknown, operation: string): string {
+    return `Failed to access local storage (${operation}). Validate that you have write permissions at ${this.config.path}. ${error}`
   }
 }


isaacroldan:
* **#6475** <a href="https://app.graphite.dev/github/pr/Shopify/cli/6475?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Shopify/cli/6475?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/cli-kit/src/public/node/local-storage.ts
MitchDickinson, line 1: Should we wrap `get` too? why/why not?

packages/cli-kit/src/public/node/local-storage.ts
```diff
@@ -27,24 +28,44 @@ export class LocalStorage<T extends {[key: string]: any}> {
    *
    * @param key - The key to set.
    * @param value - The value to set.
+   * @throws Error If the value cannot be set.
    */
   set<TKey extends keyof T>(key: TKey, value?: T[TKey]): void {
-    this.config.set(key, value)
+    try {
+      this.config.set(key, value)
+    } catch (error) {
```
MitchDickinson, line 48: I think assigning *all* failures of these methods as user errors is too broad and I worry we might miss some actual unexpected errors if we do this.

Is it possible to do something like the following?
- catch `error`
- throw `AbortError` if the error is an error we deem as a "user error" 
- otherwise re-throw `error`

Then for errors we know are related to user problems, we can update this method.

packages/cli-kit/src/public/node/local-storage.ts
MitchDickinson, line 1: Also, let's add some tests too for the changed logic

packages/cli-kit/src/public/node/local-storage.ts
```diff
@@ -27,24 +28,44 @@ export class LocalStorage<T extends {[key: string]: any}> {
...
-    this.config.clear()
+    try {
+      this.config.clear()
+    } catch (error) {
+      throw new AbortError(this.errorMessage(error, 'clear'))
+    }
+  }
+
+  private errorMessage(error: unknown, operation: string): string {
+    return `Failed to access local storage (${operation}). Validate that you have write permissions at ${this.config.path}. ${error}`
```
MitchDickinson: Similar to above feedback, I wonder if there is a way to provide a more specific error message given a specific, expected failure mode here (do we know they don't have write access, or is that mostly just a guess?)

packages/cli-kit/src/public/node/fs.ts
```diff
@@ -440,6 +442,32 @@ export async function fileHasExecutablePermissions(path: string): Promise<boolea
...
+    // eslint-disable-next-line no-catch-all/no-catch-all
+  } catch {
+    return false
+  }
+}
+
+export function fileIsOwnedByCurrentUser(path: string): boolean {
+  // process.getuid() is only available on Unix-like systems
+  if (process.platform === 'win32' || typeof process.getuid !== 'function') return true
+  if (!fileExistsSync(path)) return true
```
gonzaloriestra: I'd return false here.

packages/cli-kit/src/public/node/local-storage.ts
```diff
@@ -17,34 +20,109 @@ export class LocalStorage<T extends {[key: string]: any}> {
...
+   * Handle errors from config operations.
+   * If the error is permission-related, throw an AbortError with helpful hints.
+   * Otherwise, throw a BugError.
+   *
+   * @param error - The error that occurred.
+   * @param operation - The operation that failed.
+   * @throws AbortError if the error is permission-related.
+   * @throws BugError if the error is not permission-related.
+   */
+  private handleError(error: unknown, operation: string): never {
```
gonzaloriestra, line 94: I would send the same error for all operations to simplify the code and make the Observe grouping easier. We don't care which one is failing, all of them should work.

packages/cli-kit/src/public/node/fs.ts
```diff
@@ -440,6 +442,32 @@ export async function fileHasExecutablePermissions(path: string): Promise<boolea
...
+    return true
+    // eslint-disable-next-line no-catch-all/no-catch-all
+  } catch {
+    return false
+  }
+}
+
+export function fileIsOwnedByCurrentUser(path: string): boolean {
+  // process.getuid() is only available on Unix-like systems
+  if (process.platform === 'win32' || typeof process.getuid !== 'function') return true
```
gonzaloriestra: We shouldn't return true in this case. What if we return undefined instead?

I've checked that there's a way to find it out for Windows, but:
- All the Observe errors happen in Mac
- In Windows is not very common to have this kind of permission issues with files

So I agree we can skip it.

gonzaloriestra (approved):
Have you tested it on Linux/Windows?

packages/cli-kit/src/public/node/local-storage.ts
isaacroldan: The error contains the path to the config file, which is different for each user, so removing this won't improve the grouping :/ 
I just added it to identify which operation is throwing most errors

gonzaloriestra: What about moving the `Check that you have write permissions for ${this.config.path}` to the tryMessage? It makes more sense there.

Regarding the operation, I think it won't be very useful for the permission error. Most errors will be from `get` because it's the most common and we are doing the same checks for all of them...

packages/cli-kit/src/public/node/local-storage.test.ts
```diff
@@ -46,12 +48,90 @@ describe('storage', () => {
...
     })
   })
 })
+
+describe('error handling', () => {
+  test('throws AbortError when file lacks write permissions', async () => {
+    await inTemporaryDirectory(async (cwd) => {
+      // Given
+      const storage = new LocalStorage<TestSchema>({cwd})
+      storage.set('testValue', 'test')
```
MitchDickinson, line 67: Is it possible using our test framework to run the exact test for all affected operations, just to ensure we don't have regressions across the board? (get, delete, set, clear)

In rails you can do like
```
[:clear, :set, :get].each do |command|
  test "#{command} throws AbortError when file lacks write permissions" do 
    # test code
  end
end
```

But not sure what we get here out of the box with our JS test framework

packages/cli-kit/src/public/node/local-storage.ts
isaacroldan: Actually most errors I think will be from `set`, because many times you have read access but not write access. I think is OK to include the operation in the error message, we don't use the message for grouping and extra information is always nice to have

packages/cli-kit/src/public/node/local-storage.test.ts
isaacroldan: No, this is not possible :/ 
We can define a custom array of test cases which in some cases helps simplifying the tests (see [this PR](https://app.graphite.dev/github/pr/Shopify/cli/6497/Add-strict-validations-for-UIDs?ref=gt-pasteable-stack)), but in this case the amount of code would be very similar.

isaacroldan: I added more cases to the test to include the rest of the operations

packages/cli-kit/src/public/node/fs.ts
isaacroldan: Yeah this is more a Unix issue, that's why I defaulted to `true` here. Returning `undefined` can be confusing because for any external check, it will be treated as `false` 🤔 

My thinking here was that if we can't check for windows, is safer to assume ownership than not?

gonzaloriestra: If I use `fileIsOwnedByCurrentUser` to check a Windows file that I don't own, it will return true and that's clearly wrong.

I think there is a way to check it for Windows, but it's probably complex and unneeded for this PR, so I agree to skip it. If you don't want to return undefined or raise an error on Windows, then I'd rename the function to something like `UnixFileIsOwnedByCurrentUser`.

isaacroldan: I've refactored both the name and the return type

PR #6490, opened 2025-10-07
Fix token refresh for Business Platform API

gonzaloriestra:

Fixes https://github.com/shop/issues-develop/issues/21223

We keep seeing some [unhandled 401 errors like this](https://observe.shopify.io/goto/CctwZi6NR?orgId=1), that shouldn't be happening.

When the CLI gets a 401, it tries to get a new token (by refreshing or starting the full authentication flow) and then retries the request.

The problem is that the retry after the refresh process was always using the App Management API token, but if the query was against Business Platform API, then it was obviously getting another 401, that was reported.

Pass the API type to the refresh function, so that it returns the proper token for each one.

### How to test your changes?

The token must become invalid right before a Business Platform request, so it's not easy to reproduce. I faked it this way:
- Add a breakpoint [here](https://github.com/Shopify/cli/blob/2ef801b5b1c476e0c9510e550e5f7178eb582d45/packages/app/src/cli/commands/app/dev.ts#L134)
- Run `p shopify app dev`
- When it the debugger stops, update the value of `appContextResult.developerPlatformClient.session.businessPlatformToken` to `wrong`
- Continue
- On main, it should return a 401. With this branch, it should work.

### Measuring impact

### Checklist

### DIFF
--- a/packages/app/src/cli/utilities/developer-platform-client.ts
+++ b/packages/app/src/cli/utilities/developer-platform-client.ts
@@ -296,9 +296,18 @@ export interface DeveloperPlatformClient {
 
 const inProgressRefreshes = new WeakMap<DeveloperPlatformClient, Promise<string>>()
 
+/**
+ * Creates an unauthorized handler for a developer platform client that will refresh the token
+ * and return the appropriate token based on the token type.
+ * If the tokenType is 'businessPlatform', the handler will return the business platform token.
+ * Otherwise, it will return the default token (App Management API or Partners API).
+ * @param client - The developer platform client.
+ * @param tokenType - The type of token to return ('default' or 'businessPlatform')
+ * @returns The unauthorized handler.
+ */
 export function createUnauthorizedHandler(
   client: DeveloperPlatformClient,
-  tokenType: 'appManagement' | 'businessPlatform' = 'appManagement',
+  tokenType: 'default' | 'businessPlatform' = 'default',
 ): UnauthorizedHandler {
   return {
     type: 'token_refresh',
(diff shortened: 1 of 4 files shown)


packages/app/src/cli/utilities/developer-platform-client.ts
```diff
@@ -296,24 +296,30 @@ export interface DeveloperPlatformClient {
...
+          await tokenRefresher
         } finally {
           inProgressRefreshes.delete(client)
         }
       }
+
+      // After refresh, get the appropriate token based on the request type
+      const session = await client.session()
+      const token = tokenType === 'businessPlatform' ? session.businessPlatformToken : session.token
+      return {token}
```
isaacroldan, line 331: Partners client also goes through this code, which will "accidentally" work because we default to "appManagement" and then it uses the generic `session.token`.

Seems like a smell that we should just have different `createUnauthorizedHandler` 🤔 
But we will remove the Partners part soon so...

What do you think? maybe having a `createBusinessPlatformUnauthroziedHandler` makes everything clearer

gonzaloriestra: Good point. I've renamed the `appManagement` type to `default` and added a comment explaining it. Is it enough or do you think we should better duplicate the function?

isaacroldan: You need to also update all calls with `this.createUnauthorizedHandler('appManagement')`

I think It'd be cleaner to have different functions for different tokens. But not a strong opinion, you can leave it like this if you prefer

gonzaloriestra: Done. I'll leave it like this because we will remove Partners soon and I prefer to avoid duplicating the logic.

Issue #6481, opened 2025-10-03
[Feature] Support glob patterns in theme environment flags


dejmedus:
### What area(s) will this request affect?

Theme

### What type of change do you want to see?

New feature

### Overview

When working with multiple environments that share naming conventions (e.g., `int-production`, `us-production`, `eu-production`), we currently need to specify each environment individually:

```sh
shopify theme delete -e int-production -e us-production -e eu-production
```

Instead we could allow glob patterns in environment flags to match multiple environments:

```sh
shopify theme delete -e "*-production"
```

### Motivation

Theme Kit allowed this, but theme commands currently do not. This will make running multi environment commands easier

andershagbard:
Still an issue

andershagbard:
Still an issue

andershagbard:
Still an issue

andershagbard:
Still an issue :(

andershagbard:
Still active

andershagbard:
Still an issue

andershagbard:
Still an issue

PR #6493, opened 2025-10-07
Allow theme commands to use glob patterns for the `--environment` flag

graygilmore:

- Fixes https://github.com/Shopify/cli/issues/6481

Some theme commands accept multiple environments to be specified but it can be annoying to write them all out if you have many environments that follow a certain pattern.

This PR allows you to specify environment names using glob patterns:

```
shopify theme push -e "*-production"
```

The above example will try to match all environments defined in your `shopify.theme.toml` file that end with `-production`. We're support all pattern matching from `minimatch` including these common patterns:

- `*`: Matches zero or more characters
- `?`: Matches exactly one character
- `[]`: Matches any single character within the brackets
- `{}`: Matches any of the comma-separated patterns

### How to test your changes?

Install the snapped version:

```
npm i -g --@shopify:registry=https://registry.npmjs.org @shopify/cli@0.0.0-snapshot-20251008192534
```

Try running commands with glob patterns:

```
shopify theme info -e "test*"
```

With a `shopify.theme.toml` that looks like this:

```toml
[environments.test1]
store = "test1"
password = "asdf"

[environments.test2]
store = "test2"
password = "asdf"

[environments.test3]
store = "test3"
password = "asdf"
```

### Measuring impact

### Checklist


packages/cli-kit/src/public/node/environments.ts
```diff
@@ -111,3 +112,43 @@ export async function getEnvironmentNames(fileName: string, options?: LoadEnviro
...
+export async function expandEnvironmentPatterns(
+  patterns: string[],
+  fileName: string,
+  options?: LoadEnvironmentOptions,
+): Promise<string[]> {
+  const allEnvironments = await getEnvironmentNames(fileName, options)
+  const matchedEnvironments = new Set<string>()
+
+  for (const pattern of patterns) {
+    const isGlob = ENVIRONMENT_ALLOWED_GLOB_PATTERN.test(pattern)
```
graygilmore: Instead of matching against this we could *always* parse with `minimatch` but this would result in us ignoring the flag if it doesn't find any matches.

While this is open for review I'm going to play around with the consequences of that approach.

andershagbard:
Was this intended to close @graygilmore? It's a much needed feature

graygilmore:
Nope! I'll be opening it back up when I have some time. I found a couple of edge cases we need to cover.

Issue #6455, opened 2025-09-30
[Bug]: Restore the ability to push to published theme


bakura10:
### Please confirm that you have:

- [x] Searched [existing issues](.) to see if your issue is a duplicate. (If you’ve found a duplicate issue, feel free to add additional information in a comment on it.)
- [x] Reproduced the issue in [the latest CLI version](https://www.npmjs.com/package/@shopify/cli).

### In which of these areas are you experiencing a problem?

Theme

### Expected behavior

Starting from Shopify CLI 3.85 I'm facing this issue:

<img width="599" height="200" alt="Image" src="https://github.com/user-attachments/assets/e61e4575-2acf-4739-bdad-34bf98380a50" />

I tried to use the option --force but it does not work. Unfortunatey I can't run this method in interactive mode as I'm piping it with other commands. In our case we're developing themes so we're always on dev stores, being able to push to the live theme is extremely convenient.

Can we have a way to bypass this new thing?

Thanks

### Actual behavior

Can't push to live theme

### Verbose output

<details>
  <summary>Verbose output</summary>

  ```
  Paste the output here!
  ```

</details>

### Reproduction steps

1.
2.
3.

### Operating System

OS X

### Shopify CLI version (`shopify --version`)

3.85.2

graygilmore:
Thanks @bakura10 we'll have a look at this one.

aswamy:
Hey @bakura10 did you try using `--allow-live` flag?

Could you post your verbose logs if you're still experiencing issues after adding that flag?

bakura10:
@aswamy --allow-live is not a valid flag.

aswamy:
I need to know what command you're running. If `--allow-live` doesn't exist, i assume you're running `shopify theme dev`, not `shopify theme push`?

I checked out a previous version of the CLI and can notice there is no confirmation on pushing directly to the live theme:

i.e. `npx shopify@3.83.3 theme dev -t <live theme>` does not show any confirmation, but `npx shopify@3.83.3 theme dev -t 175926804644 --force` does.

I guess the solution would be to have something like `--allow-live` like `shopify theme push` has OR make `force` bypass the confirmation layer 🤔

bakura10:
Sorry, I should have clarified, I'm indeed talking of `dev` command :)

PR #6503, opened 2025-10-10
Add --allow-live flag to theme dev commands

dejmedus:
- closes https://github.com/Shopify/cli/issues/6455

Recently, a [confirmation prompt](https://github.com/Shopify/cli/pull/6314) was added to `theme dev` commands to prevent unintended development on live themes. In order to run `dev` command non-interactively, we need a way confirm via CLI

This PR adds an `--allow-live` flag  to skip confirmation

https://github.com/user-attachments/assets/aa888206-3495-4e6f-9b23-a130766a2e47

### How to test your changes?

- Pull down the branch `gh pr checkout 6503` or install the snapit version
```sh
npm i -g --@shopify:registry=https://registry.npmjs.org @shopify/cli@0.0.0-snapshot-20251010162922
```
```sh
shopify theme dev -t <live theme> --allow-live
```
- running `theme dev` on a:
  - non-live theme should work as expected
  - live theme without `-a` or `--allow-live` should prompt for confirmation
  - live theme with `-a` or `--allow-live` should should not prompt

### Measuring impact

### Checklist


packages/theme/src/cli/commands/theme/dev.ts
```diff
@@ -124,6 +124,12 @@ You can run this command only in a directory that matches the [default Shopify t
       description: 'The password for storefronts with password protection.',
       env: 'SHOPIFY_FLAG_STORE_PASSWORD',
     }),
+    'allow-live': Flags.boolean({
+      description: 'Allow development on a live theme.',
+      char: 'a',
+      env: 'SHOPIFY_FLAG_ALLOW_LIVE',
+      default: false,
```
aswamy, line 131: Any reason we're doing this differently than `push.ts`?

dejmedus: > default: false

The added default? Yeah, Its not necessary, if not provided it'll be false anyways. This just read as more explicit to me - but could be removed if we want?

aswamy: nah just asking why. Not sure if there is any logic in `push` that's different when it is _not_ set.

dejmedus: Ahh, sounds good! I don't believe so

Issue #5176, opened 2025-01-10
[Feature]: Customizer keyboard shortcut to go directly to the template within the customizer instead of having to navigate


rcasimmons:
### What area(s) will this request affect?

Theme, Running your code locally

### What type of change do you want to see?

New feature

### Overview

From my time working with Wordpress back in the day, there was always a helpful bar located at the top of the page that would enable you to go directly to the admin interface for the page you're on. For example, if I'm on 'About Us', I could click 'Edit page' and I'd be taken straight to the backend to edit that page.

My thought process here was, what if the CLI, now we have shortcut keys in the latest update to quickly go to the customizer using e etc, we had another that would take us directly to the customizer of the template we're viewing on localhost?

I get I'm probably being naive but I'm sure the CLI could perfect the flow here.

Essentially, if I'm viewing the product template on 127, when I hit say 'g' or whatever key, it takes me directly to the customizer with that template pre-populated as the page to customize?

This could also spread to the front-end of Shopify, if the client is logged in and previewing a theme. The white bar at the bottom could have a button that takes the user directly to the backend to edit the page and another to go straight to the customizer.

### Motivation

It can become quite tedious having to press 'e', wait for the customizer to load (especially if the client's site speed isn't that good), then scroll through the list of templates or have to search.

One key, boom, you're already in the page ready to use the customizer.

rcasimmons:
Commenting to avoid closing.

@gonzaloriestra is this something we could look to implement perhaps?

graygilmore:
Thanks @rcasimmons this is a great idea. We've also been toying around with the idea of adding some kind of developer interface to the theme preview when it's being run via the CLI that could have a number of different options/links. Would that be of interest, too?

gonzaloriestra:
Version 3.87.0 finally includes a fix for this: https://github.com/Shopify/cli/pull/6504

PR #6504, opened 2025-10-10
Open specific template URL in theme editor shortcut

graygilmore:

- Closes https://github.com/Shopify/cli/issues/5176

Uses the most recent rendered page to determine which template to open in the theme editor in the Admin. This is helpful for when you're working on a specific template (or even product) and want to make changes in the editor.

We're able to simply forward the URL because the theme editor uses the exact path in the `previewPath` query param to determine which template to load.

### How to test your changes?

```
npm i -g --@shopify:registry=https://registry.npmjs.org @shopify/cli@0.0.0-snapshot-20251010170603
```

1. Start a dev server (`shopify theme dev`)
3. While the running server is focused, hit `e`
4. You should see the theme editor open (unchanged existing behaviour)
5. Open your local preview `t` and navigate to a different template (e.g. a product)
6. While the running server is focused, hit `e`
7. You should see the theme editor open to the product template with the specific product you were looking at selected
8. Rinse and repeat for any of the other pages you want to test!

### Measuring impact

### Checklist

### DIFF
--- a/packages/theme/src/cli/utilities/theme-environment/html.ts
+++ b/packages/theme/src/cli/utilities/theme-environment/html.ts
@@ -23,6 +23,8 @@ export function getHtmlHandler(theme: Theme, ctx: DevServerContext): EventHandle
   return defineEventHandler((event) => {
     const [browserPathname = '/', browserSearch = ''] = event.path.split('?')
 
+    ctx.lastRequestedPath = event.path
+
     const shouldRenderUploadErrorPage =
       ctx.options.errorOverlay !== 'silent' && ctx.localThemeFileSystem.uploadErrors.size > 0
 
--- a/.changeset/metal-lemons-reflect.md
+++ b/.changeset/metal-lemons-reflect.md
@@ -1,0 +1,10 @@
+---
+'@shopify/theme': patch
+---
+
+Editor shortcut (e) will now navigate to your most recently viewed template
+
+Hitting `e` while your server is running (`shopify theme dev`) will now open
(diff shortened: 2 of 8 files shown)


packages/theme/src/cli/utilities/theme-environment/html.ts
```diff
@@ -23,6 +23,8 @@ export function getHtmlHandler(theme: Theme, ctx: DevServerContext): EventHandle
   return defineEventHandler((event) => {
     const [browserPathname = '/', browserSearch = ''] = event.path.split('?')
 
+    ctx.lastRequestedPath = browserPathname
```
karreiro: May we support query parameters? They’re especially important when working with pages like search, where query parameters help play an important role on setting the state of the page (they’d also be handy for product variant pages).

From what I can see, the theme editor already supports query parameters, so this might just be a matter of making sure we pass them:

image

EvilGenius13 (approved):
Thanks @gray! Happy to see we are re-using the `e` key and not making a new one. 🎩 went well.

packages/theme/src/cli/utilities/theme-environment/html.ts
graygilmore: Ooo I like it!

graygilmore:
## Force Push Patch Notes

- [Updated](https://github.com/Shopify/cli/compare/61399d3bbeae707e920e714fa89edffc7b3389ee..3ac79824746765e3a51a9456a8e63730962fc9ee) the shortcut to include query params that might have been on the page already and added a test

PR #6507, opened 2025-10-13
Update Bun lockfile name

dan-gamble:
As part of Bun v1.2 they changed the lockfile from `.lockb` to `.lock`.

I was going to change all of the lockfiles to have an array of strings so `.lock` and `.lockb` can both be supported but that was turning into quite a large change so dropping backwards support for `.lockb` (as I can't see a way to still use this after 1.2) felt fine.


gonzaloriestra:
Hi! Thanks a lot for creating this PR. The version 1.2 was released in January, so I think it's ok to drop support for the old format.

Could you update the pending references from the tests?

dan-gamble:
I've gone ahead and updated those tests and they're passing now @gonzaloriestra

PR #6509, opened 2025-10-14
Telemetry for function-testing-helpers usage tracking

saga-dasgupta:

Fixes: https://github.com/shop/issues-shopifyvm/issues/630

We want telemetry on usage and adoption of [shopify-function-test-helpers](https://github.com/Shopify/shopify-function-test-helpers). We will do so by setting an `SHOPIFY_INVOKED_BY` env variable to the `function run` or `function build` command when invoking from `shopify-function-test-helpers`. This value will then be passing to monorail through a `env_shopify_variables` field of the monorail payload. 

- Adds new `invoked-by` flag to function run command.

### How to test your changes?

I added a unit test, I dont know if we can tophat this change locally but once its out, we should be able to query this column in the sdp tables in BigQuery. 

### Measuring impact

### Checklist

### DIFF
--- a/packages/cli-kit/src/public/node/monorail.ts
+++ b/packages/cli-kit/src/public/node/monorail.ts
@@ -10,7 +10,7 @@ const url = 'https://monorail-edge.shopifysvc.com/v1/produce'
 type Optional<T> = T | null
 
 // This is the topic name of the main event we log to Monorail, the command tracker
-export const MONORAIL_COMMAND_TOPIC = 'app_cli3_command/1.19'
+export const MONORAIL_COMMAND_TOPIC = 'app_cli3_command/1.20'
 
 export interface Schemas {
   [MONORAIL_COMMAND_TOPIC]: {


packages/app/src/cli/commands/app/function/run.ts
```diff
@@ -33,11 +34,24 @@ export default class FunctionRun extends AppUnlinkedCommand {
       description: 'Name of the WebAssembly export to invoke.',
       env: 'SHOPIFY_FLAG_EXPORT',
     }),
+    'invoked-by': Flags.string({
+      hidden: true,
+      description: 'Internal flag to track which client invoked this command.',
+      env: 'SHOPIFY_FLAG_INVOKED_BY',
```
andrewhassan: This is actually really great because in the helper, we can set this environment variable and old CLI versions will just ignore it!

isaacroldan:
Does this need to be a full flag? You can just use the env var no?
The usage of this feels more like ENV stuff

packages/app/src/cli/commands/app/function/run.ts
```diff
@@ -33,6 +33,11 @@ export default class FunctionRun extends AppUnlinkedCommand {
       description: 'Name of the WebAssembly export to invoke.',
       env: 'SHOPIFY_FLAG_EXPORT',
     }),
+    'invoked-by': Flags.string({
```
isaacroldan: this flag should have `hidden: true` as it is not intended to be used externally

saga-dasgupta:
Actually @isaacroldan I realized that for developers with old version of CLI using [shopify-function-test-helpers](https://github.com/Shopify/shopify-function-test-helpers/pull/25/files) the `invoked-by` flag would throw an error, so maybe we do need it to just be set through an environment variable that we pass into monorail through either appending to an existing column or creating a new column.

packages/cli-kit/src/private/node/analytics.ts
```diff
@@ -97,9 +97,20 @@ export async function getEnvironmentData(config: Interfaces.Config): Promise<Env
...
 
+function getShopifyEnvironmentVariables() {
+  const shopifyEnvVars = new Map<string, string>()
+  for (const [key, value] of Object.entries(process.env)) {
+    if (key.startsWith('SHOPIFY_') && value !== undefined) {
+      shopifyEnvVars.set(key, value)
+    }
+  }
+  return shopifyEnvVars
+}
```
isaacroldan: Just make `getShopifyEnvironmentVariables` return a JSON, this function can be simplified to a single line:
```ts
Object.fromEntries(Object.entries(process.env).filter(([key]) => key.startsWith('SHOPIFY_')))
```

packages/cli-kit/src/public/node/monorail.ts
isaacroldan, line 1: I think you need to also update the `MONORAIL_COMMAND_TOPIC` value to use the new version (1.20)

lopert (approved):
Love the env var approach! Thanks for digging for it! 🙏

PR #6538, opened 2025-10-20
[3.86] [Theme] Avoid running cf turnstile in local dev

EvilGenius13:
Back-port : #6474


graygilmore (approved):
👍🏻 this was something we were hoping to get into the release last week

PR #6541, opened 2025-10-21
Adding function runner info command

saga-dasgupta:

Fixes https://github.com/shop/issues-shopifyvm/issues/697

Following our discussion ([see TLDR](https://shopify.slack.com/archives/C0993FRHZKN/p1761162868112669?thread_ts=1760985048.670209&cid=C0993FRHZKN)) on the best way to run test fixtures for functions fast we determined we must invoke the function-runner directly from shopify-function-testing-helpers. 
But for that we must determine function-runner binary path, wasm file path, schema and input query paths, all of which are required to run function-runner. The CLI already has this information so we decided to create a command that exposes this information to shopify-function-testing-helpers so we don't have to duplicate the logic there. We also exposed some more function fields that can be useful to developers. 

Adds `shopify app function info` command. 
Looks like this when `--json` flag is passed
Screenshot 2025-10-29 at 11 42 18 AM

without the `--json` flag
Screenshot 2025-10-29 at 11 41 34 AM

### How to test your changes?

- Install the snapshot `npm i -g --@shopify:registry=https://registry.npmjs.org @shopify/cli@0.0.0-snapshot-20251029164105`
- Run the `shopify app function info` command with and without the `--json` flag.
- Run the command for a discount function to see both targets and input query paths included in info.

### Measuring impact

### Checklist

### DIFF
--- a/docs-shopify.dev/commands/app-function-info.doc.ts
+++ b/docs-shopify.dev/commands/app-function-info.doc.ts
@@ -8,10 +8,10 @@ const data: ReferenceEntityTemplateSchema = {
   - The function handle
   - The function name
   - The function API version
-  - The function runner path
+  - The targeting configuration
   - The schema path
   - The WASM path
-  - The targeting configuration`,
+  - The function runner path`,
   overviewPreviewDescription: `Print basic information about your function.`,
   type: 'command',
   isVisualComponent: false,
--- a/packages/app/src/cli/commands/app/function/info.ts
+++ b/packages/app/src/cli/commands/app/function/info.ts
@@ -5,7 +5,7 @@ import {appFlags} from '../../../flags.js'
 import AppUnlinkedCommand, {AppUnlinkedCommandOutput} from '../../../utilities/app-unlinked-command.js'
 import {globalFlags, jsonFlag} from '@shopify/cli-kit/node/cli'
 import {outputContent, outputResult, outputToken} from '@shopify/cli-kit/node/output'
-import {renderInfo} from '@shopify/cli-kit/node/ui'
+import {InlineToken, renderInfo} from '@shopify/cli-kit/node/ui'
 
 export default class FunctionInfo extends AppUnlinkedCommand {
   static summary = 'Print basic information about your function.'
@@ -15,10 +15,10 @@ export default class FunctionInfo extends AppUnlinkedCommand {
   - The function handle
   - The function name
   - The function API version
-  - The function runner path
+  - The targeting configuration
   - The schema path
   - The WASM path
-  - The targeting configuration`
+  - The function runner path`
 
   static description = this.descriptionWithoutMarkdown()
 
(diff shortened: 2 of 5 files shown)


.changeset/hungry-ravens-sip.md
```diff
@@ -0,0 +1,5 @@
+---
+'@shopify/app': patch
```
saga-dasgupta: I kept it as patch because its a hidden internal command, which shouldn't matter to users.

packages/app/src/cli/commands/app/function/info.ts
```diff
@@ -0,0 +1,63 @@
...
+    const app = await localAppContext({
+      directory: flags.path,
+      userProvidedConfigName: flags.config,
+    })
+
+    const ourFunction = await chooseFunction(app, flags.path)
+    const functionRunner = functionRunnerBinary()
+    await downloadBinary(functionRunner)
+
+    const inputQueryPath = ourFunction?.configuration.targeting?.[0]?.input_query
```
lopert: I'm not sure what to do here with functions that have multiple targets.
With the `[0]` here, we'll always be grabbing the first one.
i.e. the `discount` template comes out of the box with two targets.
```
  [[extensions.targeting]]
  target = "cart.lines.discounts.generate.run"
  input_query = "src/cart_lines_discounts_generate_run.graphql"
  export = "cart-lines-discounts-generate-run"

  [[extensions.targeting]]
  target = "cart.delivery-options.discounts.generate.run"
  input_query = "src/cart_delivery_options_discounts_generate_run.graphql"
  export = "cart-delivery-options-discounts-generate-run"
```

I think we need to provide all the targets in our output.
That way we can map to the right input query for the right target/export for the fixture.

packages/app/src/cli/commands/app/function/info.ts
```diff
@@ -0,0 +1,63 @@
...
+import AppUnlinkedCommand, {AppUnlinkedCommandOutput} from '../../../utilities/app-unlinked-command.js'
+import {globalFlags, jsonFlag} from '@shopify/cli-kit/node/cli'
+import {outputResult} from '@shopify/cli-kit/node/output'
+
+export default class FunctionInfo extends AppUnlinkedCommand {
+  static summary = 'Get information about the function.'
+
+  static description = 'Returns information about the function.'
+
+  static hidden = true
```
lopert: We should double check if we want this to be hidden.
I think it's probably fine for this to be public?

packages/app/src/cli/commands/app/function/info.ts
```diff
@@ -0,0 +1,63 @@
...
+    const {flags} = await this.parse(FunctionInfo)
+
+    const app = await localAppContext({
+      directory: flags.path,
+      userProvidedConfigName: flags.config,
+    })
+
+    const ourFunction = await chooseFunction(app, flags.path)
+    const functionRunner = functionRunnerBinary()
+    await downloadBinary(functionRunner)
```
lopert, line 42: I tested things by deleting the binary this command was showing me, and it handled the reinstall. Nice 👍

packages/app/src/cli/commands/app/function/info.ts
andrewhassan: I agree, we shouldn't be grabbing just the first one. We should return all the targeting information and the caller can decide what to do with it.

saga-dasgupta: Oh right okay, I had grabbed this part from `shopify app function run` command, unsure how that works for multiple targets then. I will update this to return all the input query paths and targets and then we can pick the right one based on our fixture target in `default.test.ts`.

packages/app/src/cli/commands/app/function/info.ts
saga-dasgupta: Will make it public.

packages/app/src/cli/commands/app/function/info.ts
saga-dasgupta: Updated to return all target and input query pairs.

packages/app/src/cli/commands/app/function/info.ts
```diff
@@ -0,0 +1,73 @@
...
+
+    const targeting: {[key: string]: {input_query?: string; export?: string}} = {}
+    ourFunction?.configuration.targeting?.forEach((target) => {
+      if (target.target) {
+        targeting[target.target] = {
```
lopert, line 47: Is it actually possible for `target.target` to not exist? (aka is this conditional actually required?)

- The docs have [extension.targetting](https://shopify.dev/docs/apps/build/app-extensions/configure-app-extensions#:~:text=settings%20fields.-,%5B%5Bextensions.targeting%5D%5D,required,-The%20name%20of) as required, but that could just be []
- But then it says [target and module are required](https://shopify.dev/docs/apps/build/app-extensions/configure-app-extensions#:~:text=target%3A,the%20extension%20code.).
- For functions specifically, it looks like target is required, but it [might not have an input query or export (optional)](https://shopify.dev/docs/api/functions/latest#function-extension-target-types:~:text=input_query%3A,to%20_start.)

wow those last two are confusing. at the extension level, we say module is a required field, but then in functions, we say "don't use module, use export instead, also export is optional" 🙈

saga-dasgupta: I dont think anything checks at the time this command is executed whether target has been accidentally or intentionally deleted so to avoid corrupted output it doesn't hurt to perform this quick check.

packages/app/src/cli/commands/app/function/info.ts
```diff
@@ -0,0 +1,118 @@
...
+          body: {
+            list: {
+              items: targetingItems,
+            },
+          },
+        })
+      }
+
+      renderInfo({
+        headline: 'FUNCTION INFORMATION.',
```
saga-dasgupta: Do I need this headline or should I remove this because its obvious that its function information?

.changeset/hungry-ravens-sip.md
```diff
@@ -0,0 +1,5 @@
+---
+'@shopify/app': minor
+---
+
+Added shopify app function info command to display function run arguments.
```
andrewhassan: Nit: Since this is user facing, we could frame it as the improvement to the experience. 


```suggestion
Added a `function info` command to display function information.
```

docs-shopify.dev/commands/app-function-info.doc.ts
```diff
@@ -0,0 +1,34 @@
+// This is an autogenerated file. Don't edit this file manually.
+import {ReferenceEntityTemplateSchema} from '@shopify/generate-docs'
+
+const data: ReferenceEntityTemplateSchema = {
+  name: 'app function info',
+  description: `Returns information about the function.`,
```
andrewhassan: Should we add a bit more detail here like [this](https://github.com/Shopify/cli/blob/68d4128aa8d46ee7f9244c7f74f49b1a93a9fc25/docs-shopify.dev/commands/app-info.doc.ts#L6-L11)?

docs-shopify.dev/commands/app-function-info.doc.ts
```diff
@@ -0,0 +1,34 @@
...
+  overviewPreviewDescription: `Get information about the function.`,
```
andrewhassan:
```suggestion
  overviewPreviewDescription: `Print basic information about your function.`,
```

packages/app/src/cli/commands/app/function/info.ts
```diff
@@ -0,0 +1,118 @@
...
+import {appFlags} from '../../../flags.js'
+import AppUnlinkedCommand, {AppUnlinkedCommandOutput} from '../../../utilities/app-unlinked-command.js'
+import {globalFlags, jsonFlag} from '@shopify/cli-kit/node/cli'
+import {outputContent, outputResult, outputToken} from '@shopify/cli-kit/node/output'
+import {renderInfo} from '@shopify/cli-kit/node/ui'
+
+export default class FunctionInfo extends AppUnlinkedCommand {
+  static summary = 'Get information about the function.'
+
+  static description = 'Returns information about the function.'
```
andrewhassan: Same comments as above.

packages/app/src/cli/commands/app/function/info.ts
```diff
@@ -0,0 +1,118 @@
...
+            },
+          },
+        },
+        {
+          title: 'FUNCTION BUILD',
+          body: {
+            list: {
+              items: [
+                outputContent`Schema Path: ${outputToken.path(schemaPath ?? 'N/A')}`.value,
+                outputContent`WASM Path: ${outputToken.path(ourFunction.outputPath)}`.value,
```
andrewhassan: Nit:

```suggestion
                outputContent`Wasm Path: ${outputToken.path(ourFunction.outputPath)}`.value,
```

packages/app/src/cli/commands/app/function/info.ts
andrewhassan: I didn't see this comment before my last review, but I think we should remove this.

nickwesselman:
Adding a note here that @MitchDickinson and I discussed the approach and recommended this new command. I also reviewed and approved the command output.

craigmichaelmartin (approved):

PR #6547, opened 2025-10-27
observability for failures to identity service

MitchDickinson:
I am looking to capture more information about the following failure, which is flip flopping between p3 and p2: https://github.com/shop/issues-develop/issues/8058

Currently, any failure to parse a JSON response here manifests as a BugError with minimal information. This PR adds more precise introspection of the result and emits a more detailed error that will make its way to Observe so we can more precisely understand what is failing (the idea being, once we know what's happening we can add appropriate handling to retry/reduce noise/whatever)

PR was largely AI generated with a few nudges. I did tophat this against production (happy path) and JSON is successfully parsed.

### DIFF
--- a/packages/cli-kit/src/private/node/session/device-authorization.ts
+++ b/packages/cli-kit/src/private/node/session/device-authorization.ts
@@ -8,6 +8,7 @@ import {AbortError, BugError} from '../../../public/node/error.js'
 import {isCloudEnvironment} from '../../../public/node/context/local.js'
 import {isCI, openURL} from '../../../public/node/system.js'
 import {isTTY, keypress} from '../../../public/node/ui.js'
+import {Response} from 'node-fetch'
 
 export interface DeviceAuthorizationResponse {
   deviceCode: string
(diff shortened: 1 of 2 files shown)


MitchDickinson:
* **#6547** <a href="https://app.graphite.dev/github/pr/Shopify/cli/6547?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Shopify/cli/6547?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/cli-kit/src/private/node/session/device-authorization.ts
```diff
@@ -153,3 +164,35 @@ function convertRequestToParams(queryParams: {client_id: string; scope: string})
...
+/**
+ * Build a detailed error for JSON parsing failures from the authorization service.
+ * Provides context-specific error messages based on response status and content.
+ *
+ * @param response - The HTTP response object
+ * @param responseText - The raw response body text
+ * @param _parseError - The JSON parsing error (unused but passed for potential future use)
+ * @returns BugError with detailed context about the failure
+ */
+function handleAuthorizationParseError(response: Response, responseText: string, _parseError: unknown): BugError {
```
MitchDickinson: Is this overkill? 🤷

packages/cli-kit/src/private/node/session/device-authorization.ts
```diff
@@ -126,14 +139,12 @@ export async function pollForDeviceAuthorization(code: string, interval = 5): Pr
         }
         case 'slow_down':
           currentIntervalInSeconds += 5
-          {
-            startPolling()
-            return
-          }
+          startPolling()
```
MitchDickinson, line 142: Unrelated linter fixes. It seems like CI forces you to fix these if you change the file?

packages/cli-kit/src/private/node/session/device-authorization.ts
dmerand: You probably don't need to pass both response and reponseText.

MitchDickinson: My AI overlord thinks this won't work:

> If you were to modify the code to only pass `response` to `buildAuthorizationParseErrorMessage` and try to call `response.text()` again inside that function, it would fail because the response body stream has already been consumed.

Unless you object, I'll ship as is. It's a private method/implementation detail of this component so fairly low cost to pass the potentially-redundant argument, whereas I'd have to do another e2e tophat to try out `response.text()` and it's a bit of a PIA to setup the tophatting env here.

PR #6548, opened 2025-10-28
Improve git error handling to prevent false bug reports in Observe

katiebauchman:
## What

Improves git error handling in `cli-kit` to prevent expected git failures from being reported as bugs in Observe.

Fixes https://github.com/shop/issues-develop/issues/21483

## Why

Currently, when git operations fail (e.g., network issues, invalid repositories, permission problems), these errors are being tracked as bugs in Observe. This creates noise and makes it difficult to identify actual code issues.

These should be handled as expected `AbortError`s that provide helpful user feedback, not reported as application bugs.

## How

Introduced `withGit` helper function, and created a centralized error handling wrapper that:
- Wraps all git operations with try/catch
- Converts generic `Error` instances from git operations into `AbortError`
- Preserves the original error message and stack trace
- Ensures consistent error handling across all git operations

### DIFF
--- a/packages/cli-kit/src/public/node/git.ts
+++ b/packages/cli-kit/src/public/node/git.ts
@@ -27,12 +27,9 @@ export async function initializeGitRepository(directory: string, initialBranch =
   outputDebug(outputContent`Initializing git repository at ${outputToken.path(directory)}...`)
   await ensureGitIsPresentOrAbort()
   // We use init and checkout instead of `init --initial-branch` because the latter is only supported in git 2.28+
-  await withGit({
-    directory,
-    callback: async (repo) => {
-      await repo.init()
-      await repo.checkoutLocalBranch(initialBranch)
-    },
+  await withGit({directory}, async (repo) => {
+    await repo.init()
+    await repo.checkoutLocalBranch(initialBranch)
   })
 }
 
@@ -46,7 +43,7 @@ export async function initializeGitRepository(directory: string, initialBranch =
  * @returns Files ignored by the lockfile.
  */
 export async function checkIfIgnoredInGitRepository(directory: string, files: string[]): Promise<string[]> {
-  return withGit({directory, callback: (repo) => repo.checkIgnore(files)})
+  return withGit({directory}, (repo) => repo.checkIgnore(files))
 }
 
 export interface GitIgnoreTemplate {
@@ -197,12 +194,9 @@ export async function downloadGitRepository(cloneOptions: GitCloneOptions): Prom
       await git(simpleGitOptions).clone(repository!, destination, options)
 
       if (latestTag) {
-        await withGit({
-          directory: destination,
-          callback: async (localGitRepository) => {
-            const latestTag = await getLocalLatestTag(localGitRepository, repoUrl)
-            await localGitRepository.checkout(latestTag)
-          },
+        await withGit({directory: destination}, async (localGitRepository) => {
+          const latestTag = await getLocalLatestTag(localGitRepository, repoUrl)
+          await localGitRepository.checkout(latestTag)
         })
       }
     } catch (err) {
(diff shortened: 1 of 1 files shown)


packages/cli-kit/src/public/node/git.ts
```diff
@@ -379,9 +369,29 @@ export async function ensureIsClean(directory?: string): Promise<void> {
  * @returns True is the .git directory is clean.
  */
 export async function isClean(directory?: string): Promise<boolean> {
+  return (await withGit({directory, callback: (git: SimpleGit) => git.status()})).isClean()
+}
+
+async function withGit<T>({
```
amcaplan: This function should probably be at a more logical place within the file, like at the end, since it supports other functions throughout the file.

packages/cli-kit/src/public/node/git.ts
```diff
@@ -379,9 +369,29 @@ export async function ensureIsClean(directory?: string): Promise<void> {
...
+  directory,
+  callback,
+}: {
+  directory?: string
+  callback: (git: SimpleGit) => Promise<T>
```
amcaplan: I'm rethinking the function signature, and I'm starting to think it would be preferable to have something more like

```ts
async function withGit<T>({
  directory,
}: {
  directory?: string
}, callback: (git: SimpleGit) => Promise<T>): Promise<T> {
```

so the `callback` becomes a positional argument, while the options object allows `directory` to be optional.

This leads to a cleaner, more concise signature when calling the function, akin to a Ruby block which is generally a nameless parameter because there will obviously only be one block passed in.

amcaplan (commented):
Just a couple of small suggestions. I shouldn't approve as I was too involved in coming up with the idea, but let's push for review from other team members tomorrow!

packages/cli-kit/src/public/node/git.ts
katiebauchman: I like that, agreed, thank you. It's cleaner and more clearly separates the configuration (options object) from the primary action (callback). Making the callback a positional parameter makes it read more intuitively - you're performing an action with git, optionally in a specific directory. I'll make that update and get some additional feedback!

packages/cli-kit/src/public/node/git.ts
```diff
@@ -408,18 +382,39 @@
...
-  if (!remoteExists) {
-    outputDebug(outputContent`Remote ${remoteName} does not exist, no action needed`)
-    return
+async function withGit<T>(
+  {
+    directory,
+  }: {
+    directory?: string
+  },
+  callback: (git: SimpleGit) => Promise<T>,
```
alexanderMontague, line 424: Is it more correct to have this argument named `repo` or `repository` as well? Not sure what else is in `SimpleGit`, but semantically it seems we're mostly referring to it as the git repo itself

packages/cli-kit/src/public/node/git.ts
```diff
@@ -408,18 +382,39 @@ export async function removeGitRemote(directory: string, remoteName = 'origin'):
...
+): Promise<T> {
+  // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+  // @ts-ignore
+  const repo = git({baseDir: directory})
+  try {
+    return await callback(repo)
+  } catch (err) {
+    if (err instanceof Error) {
```
alexanderMontague, line 413: Will the caught error ever _not_ be an instance of `Error`? I assumed `Error` is the base error class every other error inherits from.

packages/cli-kit/src/public/node/git.ts
```diff
@@ -408,18 +382,39 @@ export async function removeGitRemote(directory: string, remoteName = 'origin'):
...
+      const abortError = new AbortError(err.message)
+      abortError.stack = err.stack
```
alexanderMontague, line 415: Assuming `AbortError` also inherits from `Error`, does [`cause`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/cause)  make sense to use here as well? I'm not sure of what we _want_, which is to "wrap" the error, or override the error completely. Overwriting the `stack` seems like this is intended, but having `cause` set also preserves this. No strong opinions, just wanted to share this approach we use on the dev dash JS side.

alexanderMontague (approved):
I think the approach is great! Super clean ✨

katiebauchman:
* **#6548** <a href="https://app.graphite.dev/github/pr/Shopify/cli/6548?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Shopify/cli/6548?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #6553, opened 2025-10-29
Avoid problems with theme extensions self built by external vite

alfonso-noriega:

The current archiving functionality in `zip` and `brotliCompress` methods doesn't handle race conditions where files might be deleted during the archiving process. This can lead to errors and failed builds when files are removed while the archiving is in progress.

Improves the archiving process to gracefully handle files that are deleted during the archiving operation:

1. Modifies the `zip` and `brotliCompress` functions to read file content immediately before adding to the archive
2. Adds error handling to skip files that are deleted during the archiving process
3. Adds comprehensive tests to verify the new behavior works correctly

### How to test your changes?

[theme-test.zip](https://github.com/user-attachments/files/23263944/theme-test.zip)

1. Download the included zip with a test them app.
2. Run `shopify app dev --path path/to./the/downloaded/app` in a terminal
3. Within the app folder run `pnpm theme-extension:dev` to run vite
4. In a new terminal window, within the theme extension folder run the script rapid-changes.sh included in the zip.
These steps will start a never ending loop of changes to the extension files in batches of a random size which will eventually trigger the error we are trying to reproduce.

Once the error is reproduced running dev with the latest CLI, to test the changes go to the CLI repo, checkout this branch, and run `pnpm shopify app dev --path path/to./the/downloaded/app` (don't forget to close the previously started dev session).

### Measuring impact

### Checklist

### DIFF
--- a/packages/cli-kit/src/public/node/archiver.ts
+++ b/packages/cli-kit/src/public/node/archiver.ts
@@ -1,5 +1,5 @@
 import {relativePath, joinPath, dirname} from './path.js'
-import {glob, removeFile} from './fs.js'
+import {glob, removeFile, readFile} from './fs.js'
 import {outputDebug, outputContent, outputToken} from '../../public/node/output.js'
 import archiver from 'archiver'
 import {createWriteStream, readFileSync, writeFileSync} from 'fs'
@@ -64,13 +64,22 @@ export async function zip(options: ZipOptions): Promise<void> {
       archive.append(Buffer.alloc(0), {name: dirName})
     }
 
-    for (const filePath of pathsToZip) {
-      const fileRelativePath = relativePath(inputDirectory, filePath)
-      if (filePath && fileRelativePath) archive.file(filePath, {name: fileRelativePath})
-    }
+    // Read all files immediately before adding to archive to prevent ENOENT errors
+    // Using archive.file() causes lazy loading which fails if files are deleted before finalize()
+    const addFilesPromises = pathsToZip.map(async (filePath) => {
+      await archiveFile(inputDirectory, filePath, archive)
+    })
 
-    // eslint-disable-next-line @typescript-eslint/no-floating-promises
-    archive.finalize()
+    // Wait for all files to be read and added before finalizing
+    Promise.all(addFilesPromises)
+      .then(() => {
+        // eslint-disable-next-line @typescript-eslint/no-floating-promises
+        archive.finalize()
+      })
+      .catch((error) => {
+        // eslint-disable-next-line @typescript-eslint/prefer-promise-reject-errors
+        reject(error)
+      })
   })
 }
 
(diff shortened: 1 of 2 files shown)


alfonso-noriega:
* **#6559** <a href="https://app.graphite.com/github/pr/Shopify/cli/6559?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6558** <a href="https://app.graphite.com/github/pr/Shopify/cli/6558?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6553** <a href="https://app.graphite.com/github/pr/Shopify/cli/6553?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Shopify/cli/6553?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/cli-kit/src/public/node/archiver.ts
```diff
@@ -147,11 +177,33 @@ export async function brotliCompress(options: BrotliOptions): Promise<void> {
...
+              if (errorMessage.includes('ENOENT')) {
+                outputDebug(
+                  outputContent`File was deleted during bundling, skipping: ${outputToken.path(fileRelativePath)}`,
+                )
+              } else {
+                outputDebug(
+                  outputContent`Failed to read file for tar archive, skipping: ${outputToken.path(
+                    fileRelativePath,
+                  )} - ${outputToken.raw(errorMessage)}`,
+                )
```
isaacroldan: Seems like the content of these "addFilesPromises" is the same on lines 70-90 and 180-200. Better to extract it to a common place

packages/cli-kit/src/public/node/archiver.ts
```diff
@@ -147,11 +181,13 @@ export async function brotliCompress(options: BrotliOptions): Promise<void> {
...
-            const fileRelativePath = relativePath(options.inputDirectory, filePath)
-            archive.file(filePath, {name: fileRelativePath})
-          }
+        .then(async (pathsToZip) => {
+          // Read all files immediately to prevent ENOENT errors during race conditions
+          const addFilesPromises = pathsToZip.map(async (filePath) => {
+            await archiveFile(options.inputDirectory, filePath, archive)
+          })
+
+          await Promise.all(addFilesPromises)
```
plvaldes: Should we do any error handling here?

packages/cli-kit/src/public/node/archiver.ts
```diff
@@ -84,6 +93,31 @@ function collectParentDirectories(fileRelativePath: string, accumulator: Set<str
...
+    const fileContent = await readFile(filePath)
+    // Use append with Buffer instead of file() to avoid lazy file reading
+    archive.append(fileContent, {name: fileRelativePath})
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any, no-catch-all/no-catch-all
+  } catch (error: any) {
+    // If file was deleted during bundling, skip it gracefully
+    const errorMessage = error instanceof Error ? error.message : String(error)
+    if (errorMessage.includes('ENOENT')) {
+      outputDebug(outputContent`File was deleted during bundling, skipping: ${outputToken.path(fileRelativePath)}`)
+    } else {
```
gonzaloriestra: We are catching all errors here and moving on without doing anything. If for some reason some files fail to be added, why do we want to continue and create an incomplete or empty zip?

I think we should raise an error when the error is different than `ENOENT`.

isaacroldan: Yeah, I think we should stop the archive process and the dev_session_upload if we fail to create the bundle 🤔 

Message i'm not sure because if this fails, is most likely because there is another process starting a new bundle, and showing an error to the user could be confusing 🤔

gonzaloriestra (approved):

PR #6554, opened 2025-10-30
Fix multiple tunnel plugins error

gonzaloriestra:

Fixes https://github.com/shop/issues-develop/issues/21321

30-50-gvxvz-gxqdi

Ignores the plugins `@shopify/app` or `@shopify/plugin-cloudflare` because they are already included in the bundle.

That way, the `multiple tunnel plugins` error won't appear when for some reason the cloudflare plugin is duplicated.

Also, it shows a warning so the users can remove those plugins that may cause conflicts:

30-52-zh53d-07vbb

### How to test your changes?

- `npm i -g --@shopify:registry=https://registry.npmjs.org @shopify/cli@0.0.0-snapshot-20251030155921`
- `shopify plugins add @shopify/app`
- `shopify app dev`

### Measuring impact

### Checklist

### DIFF
--- a/packages/cli-kit/src/public/node/base-command.test.ts
+++ b/packages/cli-kit/src/public/node/base-command.test.ts
@@ -562,6 +562,7 @@ const deleteDefaultEnvironment = async (tmpDir: string): Promise<void> => {
 
 describe('removeDuplicatedPlugins', () => {
   let capturedPlugins: Map<string, any> | undefined
+  let outputMock: ReturnType<typeof mockAndCaptureOutput>
 
   class PluginTestCommand extends MockCommand {
     async init() {
@@ -583,6 +584,8 @@ describe('removeDuplicatedPlugins', () => {
 
   beforeEach(() => {
     capturedPlugins = undefined
+    outputMock = mockAndCaptureOutput()
+    outputMock.clear()
   })
 
   test('removes @shopify/app plugin when present', async () => {
@@ -606,6 +609,11 @@ describe('removeDuplicatedPlugins', () => {
       expect(capturedPlugins.has('@shopify/plugin-ngrok')).toBe(true)
       expect(capturedPlugins.has('@shopify/plugin-did-you-mean')).toBe(true)
       expect(capturedPlugins.size).toBe(2)
+
+      // Verify warning was shown
+      expect(outputMock.output()).toMatch(/Unsupported plugins detected.*@shopify\/app/s)
+      expect(outputMock.output()).toMatch(/shopify plugins remove @shopify\/app/)
+      expect(outputMock.output()).not.toMatch(/shopify plugins remove @shopify\/plugin-cloudflare/)
     })
   })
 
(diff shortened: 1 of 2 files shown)


packages/cli-kit/src/public/node/base-command.ts
```diff
@@ -56,6 +56,7 @@ abstract class BaseCommand extends Command {
       // This function runs just prior to `run`
       await registerCleanBugsnagErrorsFromWithinPlugins(this.config)
     }
+    await removeDuplicatedPlugins(this.config)
```
isaacroldan, line 59: I think this belongs in the `prerun.ts` hook instead of the base-command class

amcaplan (commented):
Checking my understanding: This new logic means that if I have pinned `@shopify/app` in my `package.json`, while it might previously have preferred that as the source of command implementations, now it will prefer the version that comes bundled in `@shopify/cli`?

If so, that seems like a good idea, but I wanted to be sure I understood correctly.

packages/cli-kit/src/public/node/base-command.test.ts
```diff
@@ -559,3 +559,176 @@ const deleteDefaultEnvironment = async (tmpDir: string): Promise<void> => {
...
+  beforeEach(() => {
+    capturedPlugins = undefined
+  })
+
+  test('removes @shopify/app plugin when present', async () => {
+    await inTemporaryDirectory(async (tmpDir) => {
+      // Given - set up plugins to be injected
+      const mockPlugin1 = {name: '@shopify/app', version: '1.0.0'} as any
+      const mockPlugin2 = {name: '@shopify/plugin-ngrok', version: '1.0.0'} as any
+      const mockPlugin3 = {name: '@shopify/plugin-did-you-mean', version: '1.0.0'} as any
```
amcaplan, line 596: Is there a reason not to remove extra plugins? It might not raise an error, but it seems like it could lead to non-obvious behavior.

amcaplan: Actually, what about a complementary approach - what if we warn if a non-bundled `@shopify/app` plugin is installed? And then in a few minor versions it'll go warning => raise?

That seems like a good way of forcing users to migrate to the new global installation, and avoiding awkward situations where it's not clear which is the active CLI.

gonzaloriestra: Extra plugins that don't conflict with the tunnel? Why would we want to remove them?

There are some devs working on custom plugins and also some generic ones, like did-you-mean, and they would stop working.

gonzaloriestra:
@amcaplan this is not related to the `package.json` dependencies. If you include `@shopify/app` there, it's not added as a plugin.

I'm not sure about how some people are getting into this situation. The only way I found to reproduce it is to install the plugin with `shopify plugins add` as mentioned in the description.

gonzaloriestra:
@isaacroldan where exactly would you add it? I think we don't have the config in prerun, which is what we need to update.

isaacroldan:
@gonzaloriestra you do have config in prerun `options.config`, although I just tested it and the `plugins` list in that object is read-only so you can't edit it. My question is, at this point the commands have been loaded already, so does removing the plugin actually removes the commands from it?

gonzaloriestra:
Already discussed it with Isaac and this should be safe because:
- This doesn't affect plugins adding new commands
- When there are multiple plugins, the bundle ones are used

But we are hiding a weird situation: the users shouldn't have those plugins. I think I prefer to keep raising an error, but handled and explaining how to solve it: https://github.com/Shopify/cli/pull/6555

@amcaplan what do you prefer?

amcaplan:
I think we can start by rescuing the situation (because it's not known to be unsafe) but starting to warn users, and eventually we drop support for this entirely and leave it as a handled error.

packages/cli-kit/src/public/node/base-command.test.ts
amcaplan: I meant warning on any duplicated plugins. But I think I misunderstood how this happens in the first place.

dmerand (approved):
I'm aligned with this approach.

> and eventually we drop support for this entirely and leave it as a handled error.

How do we ensure this happens?

gonzaloriestra:
> How do we ensure this happens?

We don't have a set process for this. I added a reminder to our Slack channel for two months from now.

PR #6556, opened 2025-10-30
Refactoring function info command logic into service

saga-dasgupta:
Continuation of https://github.com/Shopify/cli/pull/6541 
Fixes https://github.com/shop/issues-shopifyvm/issues/697

In my previous PR I had all the changes for function info command included in the command file which goes against the pattern of including most of the logic in the service file corresponding to the command so this PR fixes that and adds unit tests for the service file.

Breaks the logic into the command file which is mostly orchestration:

- Parse flags.
- Choose function (user interaction). This is in the command file for build function command too. 
- Download binary (infrastructure).
- Generate/get schema path (file operations with user flags). This is in the command file for the run command too.
- Call service with prepared data.
- Route output to appropriate display method.

And service file which is transformation/formatting:

- Transform raw function data into display format.
- Build targeting structure.
- Format sections for text output.
- Format JSON output.

### How to test your changes?

- `npm i -g --@shopify:registry=https://registry.npmjs.org @shopify/cli@0.0.0-snapshot-20251031143515`
- `shopify app function info --path <path to your function>`
- `shopify app function info --path <path to your function> --json`

### Measuring impact

### Checklist

### DIFF
--- a/packages/app/src/cli/services/function/info.ts
+++ b/packages/app/src/cli/services/function/info.ts
@@ -2,9 +2,9 @@ import {ExtensionInstance} from '../../models/extensions/extension-instance.js'
 import {outputContent, outputToken} from '@shopify/cli-kit/node/output'
 import {InlineToken, AlertCustomSection} from '@shopify/cli-kit/node/ui'
 
-export type Format = 'json' | 'text'
+type Format = 'json' | 'text'
 
-export interface FunctionInfoOptions {
+interface FunctionInfoOptions {
   format: Format
   functionRunnerPath: string
   schemaPath?: string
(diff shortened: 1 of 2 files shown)


packages/app/src/cli/services/function/info.ts
```diff
@@ -0,0 +1,113 @@
+import {ExtensionInstance} from '../../models/extensions/extension-instance.js'
+import {outputContent, outputToken} from '@shopify/cli-kit/node/output'
+import {InlineToken, AlertCustomSection} from '@shopify/cli-kit/node/ui'
+
+export type Format = 'json' | 'text'
```
gonzaloriestra:
```suggestion
type Format = 'json' | 'text'
```

packages/app/src/cli/services/function/info.ts
```diff
@@ -0,0 +1,113 @@
...
+
+export interface FunctionInfoOptions {
```
gonzaloriestra:
```suggestion
interface FunctionInfoOptions {
```

packages/app/src/cli/services/function/info.ts
```diff
@@ -0,0 +1,113 @@
...
+        apiVersion: config.api_version,
+        targeting,
+        schemaPath,
+        wasmPath: ourFunction.outputPath,
+        functionRunnerPath,
+      },
+      null,
+      2,
+    )
+  } else {
```
gonzaloriestra: We can remove this else to simplify a bit

packages/app/src/cli/services/function/info.ts
```diff
@@ -0,0 +1,113 @@
...
+  name?: string
+  api_version?: string
+  targeting?: {
+    target: string
+    input_query?: string
+    export?: string
+  }[]
+}
+
+export function functionInfo(
```
gonzaloriestra: This function is a bit long, could you split it?

PR #6557, opened 2025-10-30
Treat transient network errors as AbortError instead of BugError

amcaplan:
## Summary

Addresses shop/issues#839

Fixes 6,941+ production errors that were incorrectly classified as `BugError` when they should be user-facing `AbortError`.

The main issue was TLS certificate validation failures (e.g., "Hostname/IP does not match certificate's altnames") being treated as bugs rather than network connectivity issues.

### 1. Renamed `isARetryableNetworkError` → `isTransientNetworkError`
- Old name was confusing in contexts where retries had already been exhausted
- New name accurately describes what these errors are: transient network issues
- Added comprehensive JSDoc explaining the purpose

### 2. Exported function for reuse
- Made function available to other modules (previously private)
- Keeps error detection logic DRY

### 3. Enhanced network error detection
Added patterns for production errors that were being missed:
- `'timeout'` - Generic network timeouts
- `'premature close'` - Connection closed prematurely
- `'certificate'`, `'cert'`, `'tls'`, `'ssl'`, `'altnames'` - TLS/certificate errors
- `'getaddrinfo'` - DNS resolution failures

### 4. Updated `fetchApiVersions` error handling
Now checks for transient network errors and classifies them as `AbortError` with a user-friendly message instead of `BugError`.

## Production Impact

Errors now properly classified as network issues (from Observe data):
- ✅ TLS certificate validation failures (6,941 occurrences - the main issue)
- ✅ ECONNRESET (connection resets)
- ✅ ENOTFOUND (DNS failures)
- ✅ ETIMEDOUT (timeouts)
- ✅ Socket hang ups
- ✅ Premature connection closes
- ✅ TLS handshake failures

## Test Plan
- [x] All existing tests pass
- [x] TypeScript compilation succeeds
- [x] Verified all 8 production error patterns are now correctly detected

### DIFF
--- a/.changeset/network-errors-as-abort-errors.md
+++ b/.changeset/network-errors-as-abort-errors.md
@@ -2,4 +2,4 @@
 '@shopify/cli-kit': patch
 ---
 
-Treat transient network errors as user-facing AbortError instead of BugError in fetchApiVersions
+Treat all network errors as user-facing AbortError instead of BugError in fetchApiVersions


EvilGenius13 (approved):
Thanks for the PR @amcaplan!

I like the new error wording explicitly stating it's a network error.

I do want to point out that some of the changes will catch more errors than just transient. For instance, `certificate` errors can be permanent as well and we will retry for no reason if we only look for the word itself. However, the trade-off might be worth it if we are helping the majority of users having blips in their network.

amcaplan:
@EvilGenius13 There's definitely a conversation to be had here. Maybe there's a wider category of network errors, and a narrower category of retryable network errors. Cert errors are unlikely to just recover, unless caused by a momentary blip in system clock accuracy. But some others in the list are worth retrying. And I think in all cases, it's correct to treat these as Handled errors since there isn't much we can do.

By the way, I suspect many of these are caused by users who set up SSL certs incorrectly for custom domains.

amcaplan:
I've updated the PR with a lengthy series of tests, as well as distinguishing between transient network errors (should be retried) vs network errors generally (which might not fix themselves but aren't a CLI bug - so should be `AbortError` not `BugError`).

EvilGenius13 (approved):
Awesome, thanks @amcaplan! 🚀

PR #6558, opened 2025-10-31
Group events by extension to reduce triggered builds

alfonso-noriega:

The current implementation of `buildExtensions` processes each extension event individually, which can lead to inefficient handling when multiple events occur for the same extension.

Optimizes the extension build process by grouping extension events by extension UID before processing them. This ensures that when multiple events occur for the same extension, we only trigger one build operation instead of multiple redundant builds.

Key changes:
- Groups extension events by extension UID before processing
- Updates all events for the same extension with the build result (success or error)
- Maintains the same error handling and output formatting as before

### How to test your changes?

1. Run `dev` with multiple extensions
2. Make simultaneous changes to the same extension
3. Verify that only one build is triggered for multiple changes to the same extension
4. Confirm that all events receive the correct build result

### Measuring impact

### Checklist

### DIFF
--- a/packages/app/src/cli/services/dev/app-events/app-event-watcher.ts
+++ b/packages/app/src/cli/services/dev/app-events/app-event-watcher.ts
@@ -10,6 +10,7 @@ import {AbortSignal} from '@shopify/cli-kit/node/abort'
 import {joinPath} from '@shopify/cli-kit/node/path'
 import {fileExists, mkdir, rmdir} from '@shopify/cli-kit/node/fs'
 import {useConcurrentOutputContext} from '@shopify/cli-kit/node/ui/components'
+import {groupBy} from '@shopify/cli-kit/common/collection'
 import {formatMessagesSync, Message} from 'esbuild'
 import {isUnitTest} from '@shopify/cli-kit/node/context/local'
 import EventEmitter from 'events'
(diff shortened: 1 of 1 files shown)


alfonso-noriega:
* **#6559** <a href="https://app.graphite.com/github/pr/Shopify/cli/6559?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6558** <a href="https://app.graphite.com/github/pr/Shopify/cli/6558?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Shopify/cli/6558?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#6553** <a href="https://app.graphite.com/github/pr/Shopify/cli/6553?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

packages/app/src/cli/services/dev/app-events/app-event-watcher.ts
```diff
@@ -234,48 +234,74 @@ export class AppEventWatcher extends EventEmitter {
...
+        } else {
+          grouped.push({
+            extension: event.extension,
+            events: [event],
+          })
+        }
+        return grouped
+      },
+      [],
+    )
```
isaacroldan: I think this can be simplified using `groupBy`:

```ts
    const groupedExtensionEvents = groupBy(extensionEvents, (event) => event.extension.handle)
```

alfonso-noriega:
The changes have to be within the debounce time of the file watcher to be grouped, that is in less than 200ms aprox. It may be that you are not doing that fast enough to see them grouped?

isaacroldan:
The first change is not debounced, the CLI starts debouncing after the second change. This is to optimize for the most common case: a single change. If we were to debounce that, we would be adding an extra 200ms every time.

isaacroldan:
We could have a smaller debouncer for the first change only, like 20-30ms to catch automatic saves that modify multiple files at once 🤔

PR #6559, opened 2025-10-31
Add resilliency to errors on concurrent build asset writes

alfonso-noriega:

Fixes an issue where the extension bundling process would fail completely if any single file couldn't be copied, preventing developers from continuing their work.

Improves error handling during extension bundling by:
- Using `Promise.allSettled()` instead of `Promise.all()` to prevent the entire process from failing if a single file copy fails
- Adding detailed error reporting for failed file copies
- Outputting debug warnings when files can't be copied instead of crashing
- Continuing the bundling process even when some files fail to copy

### How to test your changes?

1. Create a theme or UI extension with a file that can't be copied (e.g., a file with special permissions or a locked file)
2. Run `shopify app dev`
3. Verify that the build process completes successfully despite the problematic file
4. Check the debug output for warnings about files that couldn't be copied

### Measuring impact

### Checklist

### DIFF
--- a/packages/app/src/cli/services/extensions/bundle.ts
+++ b/packages/app/src/cli/services/extensions/bundle.ts
@@ -68,14 +68,7 @@ export async function bundleThemeExtension(
   options.stdout.write(`Bundling theme extension ${extension.localIdentifier}...`)
   const files = await themeExtensionFiles(extension)
 
-  await Promise.all(
-    files.map(function (filepath) {
-      const relativePathName = relativePath(extension.directory, filepath)
-      const outputFile = joinPath(extension.outputPath, relativePathName)
-      if (filepath === outputFile) return
-      return copyFile(filepath, outputFile)
-    }),
-  )
+  await copyFiles(files, extension.directory, extension.outputPath)
 }
 
 export async function copyFilesForExtension(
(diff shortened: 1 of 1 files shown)


alfonso-noriega:
* **#6559** <a href="https://app.graphite.com/github/pr/Shopify/cli/6559?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Shopify/cli/6559?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#6558** <a href="https://app.graphite.com/github/pr/Shopify/cli/6558?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6553** <a href="https://app.graphite.com/github/pr/Shopify/cli/6553?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

packages/app/src/cli/services/extensions/bundle.ts
```diff
@@ -93,14 +117,39 @@ export async function copyFilesForExtension(
     ignore: ignored,
   })
 
-  await Promise.all(
-    files.map(function (filepath) {
+  const results = await Promise.allSettled(
```
dmerand, line 112: This logic looks entirely the same as the logic added above in `bundleThemeExtension` -- is there value in DRY?

dmerand (commented):
Couldn't tophat based on the instructions (assuming `shopify app build`?) but I'm probably missing something basic there.

Otherwise I really just have the one question about DRY.

gonzaloriestra (commented):
Again, I'm not sure about the approach. But maybe I'm missing context about the problem...

Why would you want to continue if something couldn't be copied? What's the problem with throwing an error instead of a warning?

alfonso-noriega:
when you have automatic build from an external vite running in parallel with dev which is, with every change that you do, deleting the assets folder and rewriting it, it may happend that you delete something that is listed in glob to be copied

isaacroldan:
@alfonso-noriega I think @gonzaloriestra is right, because this method is also used during deploy so we need to handle errors better

PR #6561, opened 2025-11-03
Fix how app name is set from the AppLoader

plvaldes:

Fixes https://github.com/shop/issues-develop/issues/21234

The CLI was sending the package.json name as the app version name on `shopify app deploy`, which was creating some confusion to users

This PR changes the way the app version name is set and now follows these rules:
1. The `handle` from the TOML file is used
2. If there is no `handle`, the `name` is used as a fallback

### How to test your changes?

- Create an app with the CLI
- Add a handle to the TOML file
- Run `shopify app deploy` and check that this handle is the prefix for the app version name (see the output from the command)

Captura de pantalla 2025-11-04 a las 12 23 42

### Measuring impact

### Checklist

### DIFF
--- a/packages/app/src/cli/models/app/loader.test.ts
+++ b/packages/app/src/cli/models/app/loader.test.ts
@@ -64,6 +64,7 @@ describe('load', () => {
   }
 
   const appConfiguration = `
+name = "my_app"
 scopes = "read_products"
 `
   const linkedAppConfiguration = `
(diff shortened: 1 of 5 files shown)


plvaldes:
* **#6561** <a href="https://app.graphite.dev/github/pr/Shopify/cli/6561?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Shopify/cli/6561?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/app/src/cli/models/app/loader.test.ts
```diff
@@ -64,6 +64,7 @@ describe('load', () => {
   }
 
   const appConfiguration = `
+name = "my_app"
```
plvaldes, line 67: Many changes in the test files are like this one: `name` is now required in test fixtures to be able to load the app

packages/app/src/cli/models/app/loader.ts
```diff
@@ -343,7 +342,18 @@ class AppLoader<TConfig extends AppConfiguration, TModuleSpec extends ExtensionS
     const packageJSONPath = joinPath(directory, 'package.json')
 
     // These don't need to be processed again if the app is being reloaded
-    const name = this.previousApp?.name ?? (await loadAppName(directory))
+    // name is required by BrandingSchema, so it must be present in the configuration
+    const configName = 'name' in configuration ? configuration.name : undefined
```
plvaldes: I've found types a bit tricky in this PR. `name` is required in `BrandingSchema`, but the configuration is a merge from different schemas, so the typechecker does not know we are guaranteed to have a `configuration.name`. Because of this, I used some defensive code here: although we know we will have a name, because otherwise the TOML file will not be valid, I am aborting the command if for whatever reason we find that situation

packages/app/src/cli/models/app/loader.ts
```diff
@@ -1108,11 +1118,6 @@ export async function loadHiddenConfig(
   }
 }
 
-export async function loadAppName(appDirectory: string): Promise<string> {
```
plvaldes, line 1111: We will always have at least a name, so we don't need this fallback. Thus, this is dead code

PR #6576, opened 2025-11-05
added a name field to webhook subscriptions

georgge293:
closes https://github.com/shop/issues-event-foundations/issues/518
Adds support for the `name` field on webhook subscriptions when registered in the app's `shopify.app.toml`

- Adds optional `name` field to webhook subscription TypeScript types
- Adds validation for the `name` field in webhook subscription schemas (must be a non-empty string if provided)
- Adds test coverage for webhook subscriptions with the `name` attribute

Name validation:
- Field is optional
- Must be a non-empty string if provided

### How to test your changes
1. Create a webhook subscription in your `shopify.app.toml` with a `name` field while running [local app development in dev dash:](https://docs.google.com/document/d/1_fs9oHyxt4xqKnLk3WpMVIAU2gSaPUBHIVsFzcUMSlk/edit?tab=t.0)

Example:
   ```toml
   [[webhooks.subscriptions]]
   name = "products-create"
   topics = ["products/create"]
   ```
2. Run `shopify app deploy` and verify it accepts the configuration
3. Try with an empty `name = ""` and verify validation rejects it
4. Run the test suite: `pnpm vitest loader.test.ts --run`

Tophat:
- [x] Tophatted changes locally verifying that the min and max char validations are working

### DIFF
--- a/packages/app/src/cli/models/extensions/specifications/app_config_webhook_schemas/webhook_subscription_schema.ts
+++ b/packages/app/src/cli/models/extensions/specifications/app_config_webhook_schemas/webhook_subscription_schema.ts
@@ -20,7 +20,7 @@ export const WebhookSubscriptionSchema = zod.object({
   include_fields: zod.array(zod.string({invalid_type_error: 'Value must be a string'})).optional(),
   filter: zod.string({invalid_type_error: 'Value must be a string'}).optional(),
   payload_query: zod.string({invalid_type_error: 'Value must be a string'}).trim().min(1).optional(),
-  name: zod.string({invalid_type_error: 'Value must be a string'}).min(1).max(50).optional(),
+  name: zod.string({invalid_type_error: 'Value must be a string'}).trim().min(1).max(50).optional(),
 
   compliance_topics: zod
     .array(


georgge293:
I have signed the CLA!

georgge293:
Note: Shopify core changes must be deployed first before the CLI or adding the name field would fail to be recognized and error

odacrem (approved):

packages/app/src/cli/models/extensions/specifications/app_config_webhook_schemas/webhook_subscription_schema.ts
```diff
@@ -20,6 +20,7 @@ export const WebhookSubscriptionSchema = zod.object({
   include_fields: zod.array(zod.string({invalid_type_error: 'Value must be a string'})).optional(),
   filter: zod.string({invalid_type_error: 'Value must be a string'}).optional(),
   payload_query: zod.string({invalid_type_error: 'Value must be a string'}).trim().min(1).optional(),
+  name: zod.string({invalid_type_error: 'Value must be a string'}).min(1).max(50).optional(),
```
isaacroldan: I'd suggest also doing `.trim()` before `min`

georgge293: Will do

plvaldes:
> Note: Shopify core changes must be deployed first before the CLI or adding the name field would fail to be recognized and error

@georgge293 This PR is already approved but the core changes won't be deployed until after BFCM. I see some chances this PR is still accidentally merged. Which _might_ be _fine_ because afaik there won't be more CLI releases before BFCM, but core pending deploys could take a few days to deploy after BFCM and we should be careful to not release the CLI until that specific PR is deployed if we don't want to release a broken feature.

I may be a bit paranoid here, but what about moving this PR back to draft? You would keep the approvals and prevent an accidental merge from someone who hasn't read your comment on the core changes dependency

georgge293:
> > Note: Shopify core changes must be deployed first before the CLI or adding the name field would fail to be recognized and error
> 
> @georgge293 This PR is already approved but the core changes won't be deployed until after BFCM. I see some chances this PR is still accidentally merged. Which _might_ be _fine_ because afaik there won't be more CLI releases before BFCM, but core pending deploys could take a few days to deploy after BFCM and we should be careful to not release the CLI until that specific PR is deployed if we don't want to release a broken feature.
> 
> I may be a bit paranoid here, but what about moving this PR back to draft? You would keep the approvals and prevent an accidental merge from someone who hasn't read your comment on the core changes dependency

Sounds good, will do. Thanks!

gonzaloriestra (approved):

PR #6587, opened 2025-11-07
make new package for app execute

ericlee878:
Resolves: https://github.com/shop/issues-api-foundations/issues/1064

**Background**

Creating the new `shopify app execute` CLI command.  
This is just the bare bones outline of the CLI command that is basically empty. We will be implementing the command logic into this.

**To Test**
Use `pnpm shopify app execute` to test commands during development. This builds and runs the CLI from source without needing to install it globally.

**Why regenerate the manifest:**

I had to run the `pnpm refresh-manifests` command to regenerate the manifest. This is because:
- The oclif manifest and README must be updated when adding new commands. 
- They contain command metadata for discovery, help documentation, and public docs.

What it added:
- Registered app:execute in oclif.manifest.json with flags and descriptions
- Updated README.md with command documentation
- Without this, the command exists in code but isn't discoverable by the CLI.

### DIFF
--- a/packages/app/src/cli/commands/app/execute.ts
+++ b/packages/app/src/cli/commands/app/execute.ts
@@ -1,0 +1,29 @@
+import {appFlags} from '../../flags.js'
+import AppUnlinkedCommand, {AppUnlinkedCommandOutput} from '../../utilities/app-unlinked-command.js'
+import {AppInterface} from '../../models/app/app.js'
+import {globalFlags} from '@shopify/cli-kit/node/cli'
+import {renderSuccess} from '@shopify/cli-kit/node/ui'
+
+export default class Execute extends AppUnlinkedCommand {
+  static summary = 'Execute app operations.'
+
+  static description = 'Execute app operations.'
+
+  static hidden = true
+
+  static flags = {
+    ...globalFlags,
+    ...appFlags,
+  }
+
+  async run(): Promise<AppUnlinkedCommandOutput> {
+    await this.parse(Execute)
+
+    renderSuccess({
+      headline: 'Execute command ran successfully!',
+      body: 'Placeholder command. Add execution logic here.',
+    })
+
+    return {app: undefined as unknown as AppInterface}
+  }
+}
--- a/packages/app/src/cli/index.ts
+++ b/packages/app/src/cli/index.ts
@@ -8,6 +8,7 @@ import Logs from './commands/app/logs.js'
 import Sources from './commands/app/app-logs/sources.js'
 import EnvPull from './commands/app/env/pull.js'
 import EnvShow from './commands/app/env/show.js'
+import Execute from './commands/app/execute.js'
 import FunctionBuild from './commands/app/function/build.js'
 import FunctionReplay from './commands/app/function/replay.js'
 import FunctionRun from './commands/app/function/run.js'
@@ -48,6 +49,7 @@ export const commands: {[key: string]: typeof AppLinkedCommand | typeof AppUnlin
   'app:config:use': ConfigUse,
   'app:env:pull': EnvPull,
   'app:env:show': EnvShow,
+  'app:execute': Execute,
   'app:generate:schema': GenerateSchema,
   'app:function:build': FunctionBuild,
   'app:function:replay': FunctionReplay,
--- a/packages/cli/oclif.manifest.json
+++ b/packages/cli/oclif.manifest.json
@@ -808,6 +808,84 @@
       "strict": true,
       "summary": "Display app and extensions environment variables."
     },
+    "app:execute": {
+      "aliases": [
+      ],
+      "args": {
+      },
+      "customPluginName": "@shopify/app",
+      "description": "Execute app operations.",
+      "flags": {
+        "client-id": {
+          "description": "The Client ID of your app.",
+          "env": "SHOPIFY_FLAG_CLIENT_ID",
+          "exclusive": [
+            "config"
+          ],
+          "hasDynamicHelp": false,
+          "hidden": false,
+          "multiple": false,
+          "name": "client-id",
+          "type": "option"
+        },
+        "config": {
+          "char": "c",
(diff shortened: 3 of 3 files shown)


ericlee878:
* **#6587** <a href="https://app.graphite.com/github/pr/Shopify/cli/6587?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Shopify/cli/6587?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/cli/README.md
```diff
@@ -325,6 +326,28 @@ DESCRIPTION
...
+      --client-id=<value>  The Client ID of your app.
+      --no-color           Disable color output.
+      --path=<value>       The path to your app directory.
+      --reset              Reset all your settings.
+      --verbose            Increase the verbosity of the output.
+
+DESCRIPTION
+  Execute app operations.
+
+  Execute app operations.
```
jordanverasamy: Is the second `Execute app operations.` intentional? If not, let's remove.

packages/cli/README.md
```diff
@@ -325,6 +326,28 @@ DESCRIPTION
   Displays environment variables that can be used to deploy apps and app extensions.
 ```
 
+## `shopify app execute`
```
jordanverasamy: Let's make sure this description notes that the command doesn't currently do anything.

packages/app/src/cli/commands/app/execute.ts
```diff
@@ -0,0 +1,27 @@
...
+
+  async run(): Promise<AppUnlinkedCommandOutput> {
+    await this.parse(Execute)
+
+    renderSuccess({
+      headline: 'Execute command ran successfully!',
+      body: 'Placeholder command. Add execution logic here.',
+    })
+
+    return {app: undefined as unknown as AppInterface}
```
jordanverasamy, line 27: Interesting -- do we have to return something here? Did you try not returning anything?

jordanverasamy (commented):
looks pretty good, just some Qs

packages/app/src/cli/commands/app/execute.ts
```diff
@@ -0,0 +1,27 @@
+import {appFlags} from '../../flags.js'
+import AppUnlinkedCommand, {AppUnlinkedCommandOutput} from '../../utilities/app-unlinked-command.js'
+import {AppInterface} from '../../models/app/app.js'
+import {globalFlags} from '@shopify/cli-kit/node/cli'
+import {renderSuccess} from '@shopify/cli-kit/node/ui'
+
+export default class Execute extends AppUnlinkedCommand {
```
jordanverasamy, line 7: I think @nickwesselman mentioned something about marking the command as "hidden" -- should we do that to help prevent confusion for users while this is being built?

packages/cli/README.md
ericlee878: This README was automatically generated by `pnpm refresh-manifests`. I'll go in and fix that!

packages/app/src/cli/commands/app/execute.ts
ericlee878: I think so... It causes errors because of the promise. I think this was the simplest workaround I found for now... We're going to change this logic anyways right?

packages/app/src/cli/commands/app/execute.ts
ericlee878: I remember this as well. Just added the hidden functionality. This also resolved all the other issues with the README that the `pnpm refresh-manifests` command automatically created!

packages/cli/README.md
ericlee878: Resolved by the hidden functionality [here](https://app.graphite.com/github/pr/Shopify/cli/6587/make-new-package-for-app-execute#comment-PRRC_kwDOHibhHs6VF41p)

packages/app/src/cli/commands/app/execute.ts
jordanverasamy: Yeah, should be fine, just wanted to make sure we at least tried doing it in a simpler way :)

jordanverasamy (approved):
if u make me approve a 3rd time i'm defenestrating myself

packages/app/src/cli/commands/app/execute.ts
```diff
@@ -0,0 +1,29 @@
...
+
+  async run(): Promise<AppUnlinkedCommandOutput> {
+    await this.parse(Execute)
+
+    renderSuccess({
+      headline: 'Execute command ran successfully!',
+      body: 'Placeholder command. Add execution logic here.',
+    })
+
+    return {app: undefined as unknown as AppInterface}
```
alexanderMontague, line 27: I'm missing context on what this `execute` command is intended to do, but scaffolding looks good! Asking about the command intention, because that will change your app return value, ie `AppUnlinkedCommand` or `AppLinkedCommand`. Regardless, since it should be run within the context of an app (whether thats authenticated or not), you can probably just return the actual resolved app for now.


```suggestion
    const {flags} = await this.parse(Build)
    // ...
    const app = await localAppContext({
      directory: flags.path,
      userProvidedConfigName: flags.config,
    })
    // ...
    return {app}
```

ericlee878: The `execute` command is basically doing [this](https://docs.google.com/document/d/157CvTbzxv5RDkAJCCeZMcWEPrsGLEdkFc8Hd48fUSM8/edit?tab=t.0). But thanks for the feedback, will keep that in mind!

isaacroldan: This should be `AppLinkedCommand` and the app should be obtained using `linkedAppContext` (as most other commands)

ericlee878: Yes, I realized this in my next PR! Will fix this then.

PR #6588, opened 2025-11-08
Initial Setup: Bulk Operations Infrastructure for Shopify CLI


ericlee878:
Link to [ticket](https://github.com/shop/issues-api-foundations/issues/1066)

**Summary**
This PR lays the groundwork for the `shopify app execute` command in the Shopify CLI. It creates the minimal foundations of the BulkOps CLI that future PRs will build upon. By minimal foundations, we implemented just the query (not mutation), asynchronous CLI command (no synchronous progress bar), output to STDOUT (no output file). We just wanted the skeleton of the CLI command, so that future PRs could work on extending different parts of this command in parallel.

**Implementation Details**

- The `packages/app/src/cli/commands/app/execute.ts` file runs the main command. It calls on other service files.
- The `packages/app/src/cli/api/graphql/admin-bulk-operations.ts` file is a service file that stores the queries and data structures. For mutations in the future, we should add their respective queries and data structures here.
- The `packages/app/src/cli/services/bulk-operation-run-query.ts` file holds the query-specific function that holds the logic that calls the BulkOperation Query API.

**GraphQL Codegen Setup for Bulk Operations**

Added GraphQL code generation configuration for the bulk-operations API to enable type-safe GraphQL operations.

Changes:
- Added bulkOperations project to `graphql.config.ts` pointing to `admin_schema.graphql`
- Added corresponding codegen tasks to packages/app/project.json
- Run `pnpm graphql-codegen` to generate TypeScript types from GraphQL queries/mutations

This codegen allows us to have a type-safe GraphQL operations with compile-time checks that could be automatically regenerated if we change the GraphQL API.

**Testing**
1. Follow the instructions [here](https://github.com/Shopify/shopify-app-template-ruby?tab=readme-ov-file#getting-started) to setup a test app 
2. Go [here](https://dev.shopify.com/dashboard/182240880/stores) and create a store (if not already have one).
3. Use the command `pnpm shopify app execute --path=<path to test app> --query="{products{edges{node{id title handle}}}}"`
4. Get the BulkOps id that is returned and run this query in Services Internal.
```
SELECT * from bulk_operations where id = <gid>;
```
Yours should pop up. Mine was in shard 239 (for reference).

Unit tests were not added for the `execute-bulk-operation.ts` file yet because the task it was doing is currently very trivial. The main functionality of this file will come when the mutations are implemented (next PR coming!) so I will add unit tests for this file then.

### DIFF
--- a/packages/app/src/cli/commands/app/execute.ts
+++ b/packages/app/src/cli/commands/app/execute.ts
@@ -2,10 +2,8 @@ import {appFlags, bulkOperationFlags} from '../../flags.js'
 import AppLinkedCommand, {AppLinkedCommandOutput} from '../../utilities/app-linked-command.js'
 import {linkedAppContext} from '../../services/app-context.js'
 import {storeContext} from '../../services/store-context.js'
-import {runBulkOperationQuery} from '../../services/bulk-operation-run-query.js'
+import {executeBulkOperation} from '../../services/bulk-operations/execute-bulk-operation.js'
 import {globalFlags} from '@shopify/cli-kit/node/cli'
-import {renderSuccess, renderInfo, renderWarning} from '@shopify/cli-kit/node/ui'
-import {outputContent, outputToken} from '@shopify/cli-kit/node/output'
 
 export default class Execute extends AppLinkedCommand {
   static summary = 'Execute bulk operations.'
@@ -36,53 +34,11 @@ export default class Execute extends AppLinkedCommand {
       forceReselectStore: flags.reset,
     })
 
-    renderInfo({
-      headline: 'Starting bulk operation.',
-      body: `App: ${appContextResult.app.name}\nStore: ${store.shopDomain}`,
-    })
-
-    const bulkOperationResponse = await runBulkOperationQuery({
+    await executeBulkOperation({
+      app: appContextResult.app,
       storeFqdn: store.shopDomain,
       query: flags.query,
     })
-
-    if (bulkOperationResponse?.userErrors?.length) {
-      const errorMessages = bulkOperationResponse.userErrors
-        .map((error) => `${error.field?.join('.') ?? 'unknown'}: ${error.message}`)
-        .join('\n')
-      renderWarning({
-        headline: 'Bulk operation errors.',
-        body: errorMessages,
-      })
-      return {app: appContextResult.app}
-    }
-
-    const result = bulkOperationResponse?.bulkOperation
-    if (result) {
-      const infoSections = [
-        {
-          title: 'Bulk Operation Created',
-          body: [
-            {
-              list: {
-                items: [
-                  outputContent`ID: ${outputToken.cyan(result.id)}`.value,
-                  outputContent`Status: ${outputToken.yellow(result.status)}`.value,
-                  outputContent`Created: ${outputToken.gray(String(result.createdAt))}`.value,
-                ],
-              },
-            },
-          ],
-        },
-      ]
-
-      renderInfo({customSections: infoSections})
-
-      renderSuccess({
-        headline: 'Bulk operation started successfully!',
-        body: 'Congrats!',
-      })
-    }
 
     return {app: appContextResult.app}
   }
--- a/packages/app/src/cli/services/bulk-operations/execute-bulk-operation.test.ts
+++ b/packages/app/src/cli/services/bulk-operations/execute-bulk-operation.test.ts
@@ -1,0 +1,88 @@
+import {executeBulkOperation} from './execute-bulk-operation.js'
+import {runBulkOperationQuery} from './run-query.js'
+import {AppLinkedInterface} from '../../models/app/app.js'
+import {renderSuccess, renderInfo, renderWarning} from '@shopify/cli-kit/node/ui'
+import {describe, test, expect, vi} from 'vitest'
+
+vi.mock('./run-query.js')
+vi.mock('@shopify/cli-kit/node/ui')
+
+describe('executeBulkOperation', () => {
+  const mockApp = {
+    name: 'Test App',
+  } as AppLinkedInterface
+
+  const storeFqdn = 'test-store.myshopify.com'
+  const query = 'query { products { edges { node { id } } } }'
+
+  const successfulBulkOperation = {
+    id: 'gid://shopify/BulkOperation/123',
+    status: 'CREATED',
+    errorCode: null,
+    createdAt: '2024-01-01T00:00:00Z',
+    objectCount: '0',
+    fileSize: '0',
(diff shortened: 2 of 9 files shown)


ericlee878:
* **#6588** <a href="https://app.graphite.com/github/pr/Shopify/cli/6588?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#6595](https://github.com/Shopify/cli/pull/6595) <a href="https://app.graphite.com/github/pr/Shopify/cli/6595?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#6596](https://github.com/Shopify/cli/pull/6596) <a href="https://app.graphite.com/github/pr/Shopify/cli/6596?utm_source=stack-comment-icon" target="_blank">Graphite</a>) 👈 <a href="https://app.graphite.com/github/pr/Shopify/cli/6588?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/app/src/cli/api/graphql/admin-bulk-operations.ts
```diff
@@ -0,0 +1,44 @@
+import {gql} from 'graphql-request'
+
+// eslint-disable-next-line @shopify/cli/no-inline-graphql
+export const BulkOperationRunQuery = gql`
```
jordanverasamy: [nit/minor suggestion] Maybe we should consider moving this string into e.g. `bulk-operation-run-query.graphql` and we just read from there? Then we could avoid having to do `// eslint-disable-next-line @shopify/cli/no-inline-graphql`.

ericlee878:
https://github.com/shop/issues-api-foundations/issues/1066

packages/app/src/cli/api/graphql/admin-bulk-operations.ts
ericlee878: Did this. It caused more changes than I expected, but I think all of those changes were good. Also good that we did this before the mutation part. We can do something similar for the mutations.

packages/app/src/cli/commands/app/execute.ts
```diff
@@ -1,29 +1,89 @@
...
     })
 
-    return {app: undefined as unknown as AppInterface}
+    const store = await storeContext({
+      appContextResult,
+      storeFqdn: flags.store,
+      forceReselectStore: flags.reset,
+    })
+
+    renderInfo({
```
isaacroldan: The pattern we follow in every command is that we get the context (app context, store context) in the command file, but then all the logic is in a `service` file, so this file is kept very simple.

ericlee878: Good point. Implemented this.

packages/app/src/cli/services/bulk-operations/execute-bulk-operation.test.ts
```diff
@@ -0,0 +1,88 @@
...
+    id: 'gid://shopify/BulkOperation/123',
+    status: 'CREATED',
+    errorCode: null,
+    createdAt: '2024-01-01T00:00:00Z',
+    objectCount: '0',
+    fileSize: '0',
+    url: null,
+  }
+
+  test('executeBulkOperation successfully runs', async () => {
```
jordanverasamy, line 28:
```suggestion
  test('executeBulkOperation renders success message when bulk operation is started', async () => {
```

[nit] ideally unit test names should be along the lines of "if x, then y" or "y when x", where x and y are as specific as possible. that way someone reading the test name gets a better idea of what the code does!

"successfully runs" is too vague of a y, let's be more specific

packages/app/src/cli/flags.ts
```diff
@@ -33,3 +34,18 @@ export const appFlags = {
...
   }),
 }
+
+export const bulkOperationFlags = {
+  query: Flags.string({
+    char: 'q',
+    description: 'The GraphQL query, as a string.',
+    env: 'SHOPIFY_FLAG_QUERY',
+    required: true,
+  }),
```
isaacroldan, line 44: Would be nice if this query can also be passed as standard input, so that you can do:
```
shopify app execute --query 'your query' 
// or
echo 'your query' | shopify app execute
// or
cat query.txt | shopify app execute
```

the last one, being able to run a query from a file is super useful for this case where queries can be complex!

This is the same pattern that `shopify app function run` uses.
That would mean this flag is not `required` and you will need to validate manually that either the flag or stdin are present.

What do you think?

packages/app/src/cli/services/bulk-operations/execute-bulk-operation.ts
```diff
@@ -0,0 +1,65 @@
...
+      .join('\n')
+    renderWarning({
+      headline: 'Bulk operation errors.',
+      body: errorMessages,
+    })
+    return
+  }
+
+  const result = bulkOperationResponse?.bulkOperation
+  if (result) {
```
isaacroldan, line 40: Shouldn't we show a specific error if there is no result here?

packages/app/src/cli/flags.ts
ericlee878: Sounds like a good idea to me. @nickwesselman One thing - would we still need a separate `--variable-file` flag with this?

jordanverasamy: This is definitely a good idea, we should align with @nickwesselman on which pattern to follow. I'd argue that we can keep this out of scope for this PR though, and just make a ticket to improve this in a followup?

nickwesselman: Yeah for sure, I like this in `function run` and it's a nice one-liner approach for simple queries. Let's log this as a follow up.

ericlee878: [Ticket](https://github.com/shop/issues-api-foundations/issues/1047?issue=shop%7Cissues-api-foundations%7C1111) created for this

packages/app/src/cli/services/bulk-operations/execute-bulk-operation.test.ts
ericlee878: Already merged :grimacing:
[Ticket](https://github.com/shop/issues-api-foundations/issues/1047?issue=shop%7Cissues-api-foundations%7C1112) created for future fix

packages/app/src/cli/services/bulk-operations/execute-bulk-operation.ts
ericlee878: Future [ticket](https://github.com/shop/issues-api-foundations/issues/1047?issue=shop%7Cissues-api-foundations%7C1113)

PR #6595, opened 2025-11-11
[Bulk Operations CLI] show live progress updates when `--watch` is provided

jordanverasamy:
Issue: https://github.com/shop/issues-api-foundations/issues/1093

We want users to be able to pass a `--watch` flag, so they can monitor the progress and completion of their bulk operations directly from the CLI.

- Adds a new GraphQL query `GetBulkOperationById` to fetch the status and details of a bulk operation
- Adds a new function `watchBulkOperation` that polls a bulk operation and displays live progress updates, using the new `renderSingleTask `async functionality added by https://github.com/Shopify/cli/pull/6631
- Adds a `--watch` flag to the `app execute` command that controls whether to use `watchBulkOperation`

#### Bulk queries

Verify that when `--watch` is provided, we see live updates printed to the terminal, then eventually a final screen when it finishes:

```bash
pnpm shopify app execute --path=<YOUR_TEST_APP> --query="{ products { edges { node { id } } } }" --watch
```

https://github.com/user-attachments/assets/ebecb808-f5be-4cba-ad85-a4d5c1740d91

Next, verify that when `--watch` is _not_ provided, the `execute` command prints some info about the bulk operation and returns right away, just like before this PR:

```bash
pnpm shopify app execute --path=<YOUR_TEST_APP> --query="{ products { edges { node { id } } } }"
```

https://github.com/user-attachments/assets/a915a68e-d952-4486-8855-a9d2614b6e2a

Then do the same tests for bulk mutations:

#### Bulk mutations

With `--watch`:

```bash
pnpm shopify app execute \
--path=<YOUR_TEST_APP> \
--query='mutation createProduct($productSet: ProductSetInput!) { productSet(synchronous: true, input: $productSet) { product { id } } }' \
--variables='{ "productSet": { "title": "Winter hat" } }' \
--variables='{ "productSet": { "title": "Winter coat" } }' \
--variables='{ "productSet": { "title": "Winter pants" } }' \
--watch
```

https://github.com/user-attachments/assets/f826e9de-a810-433c-a6ed-500a659319fc

Without `--watch`:

```bash
pnpm shopify app execute \
--path=<YOUR_TEST_APP> \
--query='mutation createProduct($productSet: ProductSetInput!) { productSet(synchronous: true, input: $productSet) { product { id } } }' \
--variables='{ "productSet": { "title": "Winter hat" } }' \
--variables='{ "productSet": { "title": "Winter coat" } }' \
--variables='{ "productSet": { "title": "Winter pants" } }' \
```

https://github.com/user-attachments/assets/55f5d7e7-d1b4-492f-b997-2e4a1720121f

### DIFF
--- a/packages/app/src/cli/services/bulk-operations/execute-bulk-operation.ts
+++ b/packages/app/src/cli/services/bulk-operations/execute-bulk-operation.ts
@@ -97,7 +97,7 @@ function renderBulkOperationResult(operation: BulkOperation): void {
       break
     case 'COMPLETED':
       if (operation.url) {
-        const downloadMessage = outputContent`Download results ${outputToken.link('here.', operation.url)}`.value
+        const downloadMessage = outputContent`Download results ${outputToken.link('here', operation.url)}.`.value
         renderSuccess({headline, body: [downloadMessage], customSections})
       } else {
         renderSuccess({headline, customSections})
--- a/packages/app/src/cli/services/bulk-operations/format-bulk-operation-status.ts
+++ b/packages/app/src/cli/services/bulk-operations/format-bulk-operation-status.ts
@@ -12,9 +12,11 @@ export function formatBulkOperationStatus(
     case 'CREATED':
       return outputContent`Starting...`
     case 'COMPLETED':
-      return outputContent`Bulk operation succeeded. ${outputToken.gray(`(${String(operation.objectCount)} objects)`)}`
+      return outputContent`Bulk operation succeeded: ${outputToken.gray(`${String(operation.objectCount)} objects`)}`
     case 'FAILED':
-      return outputContent`Bulk operation failed. ${outputToken.gray(`(error: ${operation.errorCode ?? 'unknown'})`)}`
+      return outputContent`Bulk operation failed. ${outputToken.errorText(
+        `Error: ${operation.errorCode ?? 'unknown'}`,
+      )}`
     case 'CANCELING':
       return outputContent`Bulk operation canceling...`
     case 'CANCELED':
--- a/packages/app/src/cli/services/bulk-operations/execute-bulk-operation.test.ts
+++ b/packages/app/src/cli/services/bulk-operations/execute-bulk-operation.test.ts
@@ -345,7 +345,7 @@ describe('executeBulkOperation', () => {
     expect(watchBulkOperation).toHaveBeenCalledWith(mockAdminSession, createdBulkOperation.id)
     expect(renderSuccess).toHaveBeenCalledWith(
       expect.objectContaining({
-        headline: expect.stringContaining('Bulk operation succeeded.'),
+        headline: expect.stringContaining('Bulk operation succeeded:'),
         body: expect.arrayContaining([expect.stringContaining('https://example.com/download')]),
       }),
     )
--- a/packages/app/src/cli/services/bulk-operations/format-bulk-operation-status.test.ts
+++ b/packages/app/src/cli/services/bulk-operations/format-bulk-operation-status.test.ts
@@ -31,8 +31,8 @@ describe('formatBulkOperationStatus', () => {
 
   test('formats COMPLETED status', () => {
     const result = formatBulkOperationStatus(createMockOperation({status: 'COMPLETED', objectCount: 100}))
-    expect(result.value).toContain('Bulk operation succeeded.')
-    expect(result.value).toContain('(100 objects)')
+    expect(result.value).toContain('Bulk operation succeeded:')
+    expect(result.value).toContain('100 objects')
   })
 
   test('formats FAILED status with error code', () => {
@@ -40,13 +40,13 @@ describe('formatBulkOperationStatus', () => {
       createMockOperation({status: 'FAILED', objectCount: 10, errorCode: 'ACCESS_DENIED'}),
     )
     expect(result.value).toContain('Bulk operation failed.')
-    expect(result.value).toContain('(error: ACCESS_DENIED)')
+    expect(result.value).toContain('Error: ACCESS_DENIED')
   })
 
   test('formats FAILED status without error code', () => {
     const result = formatBulkOperationStatus(createMockOperation({status: 'FAILED', objectCount: 10, errorCode: null}))
     expect(result.value).toContain('Bulk operation failed.')
-    expect(result.value).toContain('(error: unknown)')
+    expect(result.value).toContain('Error: unknown')
   })
 
   test('formats CANCELING status', () => {


jordanverasamy:
* **#6595** <a href="https://app.graphite.com/github/pr/Shopify/cli/6595?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#6655](https://github.com/Shopify/cli/pull/6655) <a href="https://app.graphite.com/github/pr/Shopify/cli/6655?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#6656](https://github.com/Shopify/cli/pull/6656) <a href="https://app.graphite.com/github/pr/Shopify/cli/6656?utm_source=stack-comment-icon" target="_blank">Graphite</a>) 👈 <a href="https://app.graphite.com/github/pr/Shopify/cli/6595?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#6588** <a href="https://app.graphite.com/github/pr/Shopify/cli/6588?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#6596](https://github.com/Shopify/cli/pull/6596) <a href="https://app.graphite.com/github/pr/Shopify/cli/6596?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* `main`


<h2></h2>

packages/app/src/cli/api/graphql/admin-bulk-operations.ts
```diff
@@ -42,3 +42,23 @@ export interface BulkOperationRunQuerySchema {
     userErrors: BulkOperationError[]
   }
 }
+
+// eslint-disable-next-line @shopify/cli/no-inline-graphql
+export const GetBulkOperationById = gql`
```
ericlee878: Might have to move this to a separate .graphql file if we decide to organize the GraphQL commands that way. We can make this change after talking it through though.

packages/app/src/cli/flags.ts
```diff
@@ -48,4 +48,9 @@ export const bulkOperationFlags = {
     env: 'SHOPIFY_FLAG_STORE',
     parse: async (input) => normalizeStoreFqdn(input),
   }),
+  watch: Flags.boolean({
```
ericlee878, line 66: Did we ever for sure decide on the terminology for the flag here?

packages/app/src/cli/services/watch-bulk-operation.ts
```diff
@@ -0,0 +1,38 @@
...
+import {sleep} from '@shopify/cli-kit/node/system'
+import {outputInfo} from '@shopify/cli-kit/node/output'
+
+interface WatchBulkOperationOptions {
+  id: string
+  storeFqdn: string
+}
+
+const TERMINAL_STATUSES = ['COMPLETED', 'FAILED', 'CANCELED', 'EXPIRED']
+const POLL_INTERVAL_SECONDS = 5
```
ericlee878: Why 5 seconds?

packages/app/src/cli/services/bulk-operations/run-query.ts
```diff
@@ -19,6 +19,7 @@ export async function runBulkOperationQuery(
     query: BulkOperationRunQuery,
     session: adminSession,
     variables: {query},
+    version: '2026-01',
```
jordanverasamy: (Need to use `2026-01` or later to get access to `bulkOperation`. Maybe later this will be a parameter, but for now hardcoding `2026-01` is a simple way to make sure things work against the right version.)

packages/app/src/cli/api/graphql/admin-bulk-operations.ts
jordanverasamy: yea, glad we settled on that :) done!

packages/app/src/cli/flags.ts
jordanverasamy: dont think so, but easy to change in the future, so not concerned shipping this as `--watch` for now -- we can check in with Nick soon :)

packages/app/src/cli/services/watch-bulk-operation.ts
jordanverasamy: good question -- easy to adjust as we zero in on the right UX, i just went pretty conservative with the polling for now.

i also found that the numbers dont update _that_ frequently on the backend, so polling much faster than this isn't super helpful -- if it updates every 5 seconds on the backend, polling more often doesn't help.

ericlee878: Makes sense. Just was curious if there was any particular reasoning!

packages/app/src/cli/commands/app/execute.ts
```diff
@@ -38,6 +38,7 @@ export default class Execute extends AppLinkedCommand {
       app: appContextResult.app,
       storeFqdn: store.shopDomain,
       query: flags.query,
+      watch: flags.watch,
```
ericlee878, line 43: I think this might have a merge conflict with https://github.com/Shopify/cli/blob/3f5f064cc030fa64edd7180af621b7c99f4b7897/packages/app/src/cli/commands/app/execute.ts#L40 ...
Sorry :sweat_smile:

jordanverasamy: LOL no worries, easy to handle :)

packages/app/src/cli/services/bulk-operations/execute-bulk-operation.ts
```diff
@@ -49,31 +52,45 @@ export async function executeBulkOperation(input: ExecuteBulkOperationInput): Pr
...
+      renderSuccess({headline: 'Bulk operation started.', customSections})
+      break
+    case 'COMPLETED':
+      if (operation.url) {
+        const downloadMessage = outputContent`Download results ${outputToken.link('here.', operation.url)}`.value
+        renderSuccess({headline, body: [downloadMessage], customSections})
+      } else {
+        renderSuccess({headline, customSections})
+      }
+      break
```
ericlee878, line 105: [nit] Would it be better to use `renderWarning` here for other cases? Not too sure so would like to hear your opinion.

Like 
``` suggestion
      break
      case 'CANCELED':
      case 'CANCELING':
      case 'EXPIRED':
            renderWarning({headline, customSections})
            break
```

packages/app/src/cli/services/bulk-operations/watch-bulk-operation.ts
```diff
@@ -0,0 +1,68 @@
...
+    } else {
+      yield latestOperationState
+    }
+
+    // eslint-disable-next-line no-await-in-loop
+    await sleep(POLL_INTERVAL_SECONDS)
+  }
+}
+
+async function fetchBulkOperation(adminSession: AdminSession, operationId: string): Promise<GetBulkOperationByIdQuery> {
```
ericlee878, line 62: Is this necessary to make as a separate function?

packages/app/src/cli/services/bulk-operations/watch-bulk-operation.ts
```diff
@@ -0,0 +1,68 @@
...
+  return adminRequestDoc<GetBulkOperationByIdQuery, {id: string}>({
+    query: GetBulkOperationById,
+    session: adminSession,
+    variables: {id: operationId},
+    version: '2026-01',
```
ericlee878: Should we make a constant called version and pass that through here?
```
    VERSION = '2026-01'
```
```suggestion
    version: VERSION,
```

packages/app/src/cli/services/bulk-operations/execute-bulk-operation.ts
jordanverasamy: Yeah, I considered this, no strong feelings either way. Happy to change it for now if you like. Let's check in with @nickwesselman about this once he's back from Thanksgiving.

packages/app/src/cli/services/bulk-operations/watch-bulk-operation.ts
jordanverasamy: Not at all necessary, but for readability I want things split up into many small functions as much as possible :) I pulled this out to be its own function fairly late in this PR's development.

packages/app/src/cli/services/bulk-operations/watch-bulk-operation.ts
jordanverasamy: Yeah, good idea, since we know we're going to be iterating on this. I'll adjust. TY!

jordanverasamy: Done. Kept it simple for now instead of making a bigger sweeping change -- might do that in a followup though. Made [a ticket to track it](https://github.com/shop/issues-api-foundations/issues/1138) but we'll see if we get to it.

ericlee878: Thanks! Sorry if I'm being too nitpicky.

ericlee878:
Everything looks good to me! Great tophatting video btw :+1:

packages/app/src/cli/services/bulk-operations/execute-bulk-operation.ts
```diff
@@ -49,31 +52,45 @@ export async function executeBulkOperation(input: ExecuteBulkOperationInput): Pr
...
-      headline: 'Bulk operation started successfully!',
-      body: 'Congrats!',
-    })
+  const customSections = [{body: [{list: {items}}]}]
+
+  switch (operation.status) {
+    case 'CREATED':
+      renderSuccess({headline: 'Bulk operation started.', customSections})
+      break
+    case 'COMPLETED':
```
ericlee878, line 98: Would it be possible for us to land in a 'RUNNING' state? I'm thinking of a case where we don't `--wait` and the bulk operation API responds right away for us to be in a `RUNNING` state. Don't know how likely this would be though.

packages/app/src/cli/services/bulk-operations/watch-bulk-operation.ts
jordanverasamy: Not at all, please keep it up. I always have the option to just say "that's too nitpicky I won't bother", so you should err on the side of overcommunicating your thoughts :)

packages/app/src/cli/services/bulk-operations/execute-bulk-operation.ts
jordanverasamy: That's a really good question -- let's think through it carefully.

Looking at the resolver code ([query](https://github.com/shop/world/blob/main/areas/core/shopify/components/platform/bulk_operations/app/models/graph_api/admin/mutations/bulk_operation_run_mutation.rb#L192-L211) | [mutation](https://github.com/shop/world/blob/main/areas/core/shopify/components/platform/bulk_operations/app/models/graph_api/admin/mutations/bulk_operation_run_mutation.rb#L204-L213)), we return the bulk operation immediately after enqueueing the batch. That gap of time should be well under a millisecond. For us to get back an operation in the `RUNNING` state, in that sub-millisecond time interval, the following would've had to happen:

- everything in front of that bulk operation job in the queue gets picked up by a job worker
- the job payload we enqueued gets dequeued and parsed
- the job worker runs its entire `build_enumerator` method (which requires doing a lot of things, one of which is loading the bulk operation record from MySQL)
- the job worker runs its `on_start` method, which executes a transaction to update the row in MySQL

This is not going to happen anywhere near <1ms in normal operation, so I think getting `RUNNING` would be an extremely, extremely exceptional case. That said, covering that case would be extremely simple -- we'd just need to add a couple lines of straightforward code here. So we could do it, but should we?

IMO, this case is so rare, so close to impossible, that the benefits of adding it are negligible. On the other hand, there's a real cost to adding the code branch to cover it -- that might communicate to future readers that getting back a `RUNNING` operation is reasonably possible, which would confuse them.

I see that cost as being bigger than the benefit, so I don't think we should add this.

jordanverasamy: (Marking this as resolved so that reviewers don't get forced to read my wall of text, I only wrote it for overexplaining-to-Eric reasons. Feel free to respond or ask more Qs though!!)

ericlee878: Wow thanks for the long explanation LOL.
I think this makes sense. I figured it would be a rare case, but I just wanted to make sure. Thanks!

packages/app/src/cli/services/bulk-operations/execute-bulk-operation.ts
```diff
@@ -64,31 +67,45 @@ export async function executeBulkOperation(input: ExecuteBulkOperationInput): Pr
...
-    })
+  const createdOperation = bulkOperationResponse?.bulkOperation
+  if (createdOperation) {
+    if (watch) {
+      const finishedOperation = await watchBulkOperation(adminSession, createdOperation.id)
+      renderBulkOperationResult(finishedOperation)
+    } else {
+      renderBulkOperationResult(createdOperation)
+    }
+  }
```
gonzaloriestra, line 78: What about simplifying this way? Improving the error message with something more useful if possible.

```suggestion
  let operation = bulkOperationResponse?.bulkOperation
  if (!operation) throw new AbortError('Something went wrong when creating the bulk operation')
  if (watch) operation = await watchBulkOperation(adminSession, createdOperation.id)
  renderBulkOperationResult(operation)
```

packages/app/src/cli/services/bulk-operations/execute-bulk-operation.ts
gonzaloriestra: IMO, those states should be shown as errors because the query didn't finish successfully, so I think it's ok as it is.

packages/app/src/cli/services/bulk-operations/format-bulk-operation-status.ts
```diff
@@ -0,0 +1,27 @@
...
+import {outputContent, outputToken, TokenizedString} from '@shopify/cli-kit/node/output'
+
+export function formatBulkOperationStatus(
+  operation: NonNullable<GetBulkOperationByIdQuery['bulkOperation']>,
+): TokenizedString {
+  switch (operation.status) {
+    case 'RUNNING':
+      return outputContent`Bulk operation in progress... ${outputToken.gray(
+        `(${String(operation.objectCount)} objects)`,
+      )}`
```
gonzaloriestra, line 11: Where is this second ellipsis is coming from?

Image

Not sure if it's a UX decision, but I think it's a bit weird. I'd prefer any of these options: 
- `Bulk operation in progress: 17 objects`
- `Bulk operation in progress: 17 objects...`
- `Bulk operation in progress... 17 objects`

packages/app/src/cli/services/bulk-operations/execute-bulk-operation.ts
```diff
@@ -64,31 +67,45 @@ export async function executeBulkOperation(input: ExecuteBulkOperationInput): Pr
...
+  switch (operation.status) {
+    case 'CREATED':
+      renderSuccess({headline: 'Bulk operation started.', customSections})
+      break
+    case 'COMPLETED':
+      if (operation.url) {
+        const downloadMessage = outputContent`Download results ${outputToken.link('here.', operation.url)}`.value
+        renderSuccess({headline, body: [downloadMessage], customSections})
+      } else {
+        renderSuccess({headline, customSections})
```
gonzaloriestra, line 103: Maybe I'm missing something, but when there is no URL, what are you supposed to do next?

```
Bulk operation started.

• ID: gid://shopify/BulkOperation/5717783019762
• Status: CREATED
• Created at: 2025-11-26T08:51:33Z
```

I'd show some help about how to get the results. If they need to run another query, couldn't we generate them so they can copy and paste? Or print a URL they can check later?

packages/app/src/cli/services/bulk-operations/format-bulk-operation-status.ts
```diff
@@ -0,0 +1,27 @@
...
+    case 'CREATED':
+      return outputContent`Starting...`
+    case 'COMPLETED':
+      return outputContent`Bulk operation succeeded. ${outputToken.gray(`(${String(operation.objectCount)} objects)`)}`
```
gonzaloriestra: Instead of `Bulk operation succeeded. (12 objects)`, what about any of these?
- `Bulk operation succeeded: 12 objects`
- `Bulk operation succeeded (12 objects)`

packages/app/src/cli/services/bulk-operations/execute-bulk-operation.ts
```diff
@@ -64,31 +67,45 @@ export async function executeBulkOperation(input: ExecuteBulkOperationInput): Pr
...
+
+  const customSections = [{body: [{list: {items}}]}]
+
+  switch (operation.status) {
+    case 'CREATED':
+      renderSuccess({headline: 'Bulk operation started.', customSections})
+      break
+    case 'COMPLETED':
+      if (operation.url) {
+        const downloadMessage = outputContent`Download results ${outputToken.link('here.', operation.url)}`.value
```
gonzaloriestra:
```suggestion
        const downloadMessage = outputContent`Download results ${outputToken.link('here', operation.url)}.`.value
```

packages/app/src/cli/services/bulk-operations/watch-bulk-operation.ts
```diff
@@ -0,0 +1,69 @@
...
+  GetBulkOperationByIdQuery,
+} from '../../api/graphql/bulk-operations/generated/get-bulk-operation-by-id.js'
+import {adminRequestDoc} from '@shopify/cli-kit/node/api/admin'
+import {sleep} from '@shopify/cli-kit/node/system'
+import {AdminSession} from '@shopify/cli-kit/node/session'
+import {outputContent} from '@shopify/cli-kit/node/output'
+import {renderSingleTask} from '@shopify/cli-kit/node/ui'
+
+const TERMINAL_STATUSES = ['COMPLETED', 'FAILED', 'CANCELED', 'EXPIRED']
+const POLL_INTERVAL_SECONDS = 5
```
gonzaloriestra, line 13: I guess many operations will be faster than 5 seconds. Couldn't we poll more frequently?

Or even better: what about a dynamic interval? I would poll every 1-2 seconds during the first 10-20 seconds, so we show enough feedback for fast queries, and then every 5s.

Otherwise, for queries under 5s, we will just show `0 objects` and then `Completed`.

packages/app/src/cli/flags.ts
gonzaloriestra, line 39: Queries are often too complex to put them inline, why not adding a `queryFile` flag, similar to the `variableFile`?

gonzaloriestra, line 1: I think we should add a `json` flag as well, to show the output in that format.

packages/app/src/cli/services/bulk-operations/format-bulk-operation-status.ts
gonzaloriestra: Also, while the number is 0, I think I'd remove it (and only show `Bulk operation in progress...`). Otherwise, it can be a bit confusing because many times you won't see any change.

gonzaloriestra (approved):
LGTM! Tested that it works as expected.

I left a bunch of suggestions, mostly about UX.

packages/app/src/cli/services/bulk-operations/execute-bulk-operation.ts
gonzaloriestra: Now I see you will add another command: https://github.com/shop/issues-api-foundations/issues/1142. Once it's ready, I'd show a message like: 
```
You can check the status of your operation by running: `shopify app bulk status --id gid://shopify/BulkOperation/5717783019762`
```

packages/app/src/cli/services/bulk-operations/execute-bulk-operation.ts
jordanverasamy: Hmm. Fair suggestion, thanks for thinking it through!

I think I like the way it is currently more, even though it's more lines. The current structure has different names for the operation -- `finishedOperation` vs `createdOperation` -- which communicates explicitly the expected state of the operation in each case. That makes it easier to grok at a glance for me, but I dunno, maybe that's just a me thing?

jordanverasamy: I think I'm opting to leave it as is for now, but if you feel strongly let me know and I'm happy to adjust in a follow-up :)

packages/app/src/cli/services/bulk-operations/format-bulk-operation-status.ts
jordanverasamy: Yeah, the second ellipsis is coming from somewhere weird, like inside `renderSingleTask` or something. I noticed this but decided to leave it for a later round of polish. I'

The point about showing nothing when the object count is 0 makes perfect sense, we'll make that change.

I logged both of these comments [here](https://github.com/shop/issues-api-foundations/issues/1149) to make sure we come back to them.

packages/app/src/cli/services/bulk-operations/execute-bulk-operation.ts
jordanverasamy: Yeah, this is an improvement that will come after the `status` command is implemented. We'll want to:

- tell you how to use `status` if you start an operation without `--watch` (I noted this down [here](https://github.com/shop/issues-api-foundations/issues/1096#issuecomment-3582644814))
- tell you how to use `status` if you start an operation with `--watch` but then CTRL+C to quit out (I noted this down [here](https://github.com/shop/issues-api-foundations/issues/1095#issuecomment-3572995823))

packages/app/src/cli/services/bulk-operations/watch-bulk-operation.ts
jordanverasamy: That's a great idea, ty!!! Noted as a polish improvement [here](https://github.com/shop/issues-api-foundations/issues/1149), will add as a follow-up.

packages/app/src/cli/flags.ts
jordanverasamy: Yeah, great question :) This is covered by https://github.com/shop/issues-api-foundations/issues/1111, although I think we might _also_ end up adding `--queryFile`. (you can CTRL+F `query-file` in the [brief doc](https://docs.google.com/document/d/157CvTbzxv5RDkAJCCeZMcWEPrsGLEdkFc8Hd48fUSM8/edit?usp=sharing) to see that it's mentioned)

jordanverasamy: Interesting idea!!! I'll let @nickwesselman confirm but that seems like a nice idea to me for scripting, we can come back to this.

jordanverasamy: Wrote it down [here](https://github.com/shop/issues-api-foundations/issues/1150) for now just so we don't lose track of the idea. TY!

packages/app/src/cli/services/bulk-operations/execute-bulk-operation.ts
jordanverasamy: Great suggestion, made this change :)

packages/app/src/cli/services/bulk-operations/format-bulk-operation-status.ts
jordanverasamy: Great suggestion, I adjusted it to `Bulk operation succeeded: 12 objects` for now.

jordanverasamy: TY!

packages/app/src/cli/services/bulk-operations/execute-bulk-operation.ts
gonzaloriestra: No problem, I don't have a strong opinion.

What I don't like about the current approach is that it does nothing when `createdOperation` is undefined. But maybe that's not possible...

packages/app/src/cli/services/bulk-operations/format-bulk-operation-status.ts
nickwesselman: I would suggest

```
Bulk operation in progress (17 objects imported)...
Bulk operation in progress (17 objects exported)...
```

based on whether it's a query or mutation. And probably just keep 0, since it makes more sense in that case.

packages/app/src/cli/services/watch-bulk-operation.ts
nickwesselman: I think 5 seconds is an OK place to start. I worry that implementing a backoff will create a perception that the import itself is slowing down.

nickwesselman: After experiencing this myself... even a 1000 product import exits before the status can be updated. I think a 1s or even 500ms poll to start would be good.

PR #6596, opened 2025-11-11
Implement mutation support for BulkOps CLI command

ericlee878:
Resolves: https://github.com/orgs/shop/projects/208/views/34?pane=issue&itemId=137773625&issue=shop%7Cissues-api-foundations%7C1069

**Background**
The basic setup of the BulkOps CLI command was done [here](https://github.com/Shopify/cli/pull/6588). The basic command used BulkOps Query. We are implementing BulkOps Mutation in this PR. 

**Implementation Details**
- We parse the query given to us in the flag to determine if the GraphQL command is a query or a mutation. Then, we do the respective GraphQL API call.

- Bulk operation CLI accept variables via the `--variables` flag in JSON format. However, the Shopify Admin API requires these variables to be uploaded as a JSONL file rather than passed inline. So, in the `stage-file.ts` file, we needed to:

        1. Create a staged upload target via `stagedUploadsCreate` mutation.
        2. Extract the variables from JSON format and put it in a JSONL file. 
        3. Upload the JSONL file to the temporary GCS URL from staged upload.
        4. Pass the staged upload key to `bulkOperationRunMutation`.

- There will be a future PR that deals with creating a `--variable-file` flag that links to a file containing the values for bulk mutations, in JSONL format.

- Used GraphQL code generation configuration for the bulkOperationRunMutation API and stagedUpload API. Ran `pnpm graphql-codegen` to generate TypeScript types from GraphQL queries/mutations

**Architecture improvements:**
- Authentication happens once at `executeBulkOperation` level via `ensureAuthenticatedAdmin`, then the session is passed down to `runBulkOperationQuery` and `runBulkOperationMutation`
- Command flags are spread directly into `executeBulkOperation` (`...flags`) instead of parsing individual properties.

**Testing**
- Setup
        1. Follow the instructions [here](https://github.com/Shopify/shopify-app-template-ruby?tab=readme-ov-file#getting-started) to setup a test app 
        2. Go [here](https://dev.shopify.com/dashboard/182240880/stores) and create a store (if not already have one).
- Testing Mutation
To test the mutation part of the CLI, run:
```
pnpm shopify app execute --path=<path to test app> --query='mutation productUpdate($input: ProductInput!) { productUpdate(input: $input) { product { id tags } userErrors { field message } } }' --variables='[{"input": {"id": "gid://shopify/Product/<product id>", "tags": ["bulk-test"]}}]'

```

### DIFF
--- a/packages/app/src/cli/api/graphql/bulk-operations/mutations/bulk-operation-run-mutation.graphql
+++ b/packages/app/src/cli/api/graphql/bulk-operations/mutations/bulk-operation-run-mutation.graphql
@@ -1,0 +1,32 @@
+mutation BulkOperationRunMutation(
+  $mutation: String!
+  $stagedUploadPath: String!
+  $clientIdentifier: String
+) {
+  bulkOperationRunMutation(
+    mutation: $mutation
+    stagedUploadPath: $stagedUploadPath
+    clientIdentifier: $clientIdentifier
+  ) {
+    bulkOperation {
+      completedAt
+      createdAt
+      errorCode
+      fileSize
+      id
+      objectCount
+      partialDataUrl
+      query
+      rootObjectCount
+      status
+      type
+      url
+    }
+    userErrors {
+      code
+      field
+      message
+    }
+  }
+}
+
--- a/packages/app/src/cli/commands/app/execute.ts
+++ b/packages/app/src/cli/commands/app/execute.ts
@@ -38,6 +38,7 @@ export default class Execute extends AppLinkedCommand {
       app: appContextResult.app,
       storeFqdn: store.shopDomain,
       query: flags.query,
+      variables: flags.variables,
     })
 
     return {app: appContextResult.app}
--- a/packages/app/src/cli/flags.ts
+++ b/packages/app/src/cli/flags.ts
@@ -38,13 +38,20 @@ export const appFlags = {
 export const bulkOperationFlags = {
   query: Flags.string({
     char: 'q',
-    description: 'The GraphQL query, as a string.',
+    description: 'The GraphQL query or mutation to run as a bulk operation.',
     env: 'SHOPIFY_FLAG_QUERY',
     required: true,
   }),
+  variables: Flags.string({
+    char: 'v',
+    description:
+      'The values for any GraphQL variables in your mutation, in JSON format. Can be specified multiple times.',
+    env: 'SHOPIFY_FLAG_VARIABLES',
+    multiple: true,
+  }),
   store: Flags.string({
     char: 's',
-    description: 'Store URL. Must be an existing development or Shopify Plus sandbox store.',
+    description: 'The store domain. Must be an existing dev store.',
     env: 'SHOPIFY_FLAG_STORE',
     parse: async (input) => normalizeStoreFqdn(input),
   }),
--- a/packages/app/src/cli/services/bulk-operations/execute-bulk-operation.test.ts
+++ b/packages/app/src/cli/services/bulk-operations/execute-bulk-operation.test.ts
@@ -1,11 +1,15 @@
 import {executeBulkOperation} from './execute-bulk-operation.js'
 import {runBulkOperationQuery} from './run-query.js'
+import {runBulkOperationMutation} from './run-mutation.js'
 import {AppLinkedInterface} from '../../models/app/app.js'
-import {renderSuccess, renderInfo, renderWarning} from '@shopify/cli-kit/node/ui'
-import {describe, test, expect, vi} from 'vitest'
+import {renderSuccess, renderWarning} from '@shopify/cli-kit/node/ui'
+import {ensureAuthenticatedAdmin} from '@shopify/cli-kit/node/session'
+import {describe, test, expect, vi, beforeEach} from 'vitest'
 
 vi.mock('./run-query.js')
+vi.mock('./run-mutation.js')
 vi.mock('@shopify/cli-kit/node/ui')
+vi.mock('@shopify/cli-kit/node/session')
 
 describe('executeBulkOperation', () => {
   const mockApp = {
@@ -13,7 +17,7 @@ describe('executeBulkOperation', () => {
   } as AppLinkedInterface
 
   const storeFqdn = 'test-store.myshopify.com'
-  const query = 'query { products { edges { node { id } } } }'
+  const mockAdminSession = {token: 'test-token', storeFqdn}
 
   const successfulBulkOperation = {
     id: 'gid://shopify/BulkOperation/123',
@@ -25,7 +29,12 @@ describe('executeBulkOperation', () => {
     url: null,
   }
 
-  test('executeBulkOperation successfully runs', async () => {
+  beforeEach(() => {
+    vi.mocked(ensureAuthenticatedAdmin).mockResolvedValue(mockAdminSession)
+  })
+
+  test('runs query operation when GraphQL document starts with query', async () => {
+    const query = 'query { products { edges { node { id } } } }'
     const mockResponse = {
       bulkOperation: successfulBulkOperation,
       userErrors: [],
@@ -39,21 +48,91 @@ describe('executeBulkOperation', () => {
     })
 
     expect(runBulkOperationQuery).toHaveBeenCalledWith({
+      adminSession: mockAdminSession,
+      query,
+      variables: undefined,
+    })
+    expect(runBulkOperationMutation).not.toHaveBeenCalled()
+  })
+
+  test('runs query operation when GraphQL document starts with curly brace', async () => {
+    const query = '{ products { edges { node { id } } } }'
+    const mockResponse = {
+      bulkOperation: successfulBulkOperation,
+      userErrors: [],
+    }
+    vi.mocked(runBulkOperationQuery).mockResolvedValue(mockResponse as any)
+
+    await executeBulkOperation({
+      app: mockApp,
       storeFqdn,
       query,
     })
 
-    expect(renderInfo).toHaveBeenCalledWith({
-      headline: 'Starting bulk operation.',
-      body: `App: ${mockApp.name}\nStore: ${storeFqdn}`,
+    expect(runBulkOperationQuery).toHaveBeenCalledWith({
+      adminSession: mockAdminSession,
+      query,
+      variables: undefined,
+    })
+    expect(runBulkOperationMutation).not.toHaveBeenCalled()
+  })
+
+  test('runs mutation operation when GraphQL document starts with mutation', async () => {
+    const mutation = 'mutation productUpdate($input: ProductInput!) { productUpdate(input: $input) { product { id } } }'
+    const mockResponse = {
+      bulkOperation: successfulBulkOperation,
+      userErrors: [],
+    }
+    vi.mocked(runBulkOperationMutation).mockResolvedValue(mockResponse as any)
+
+    await executeBulkOperation({
+      app: mockApp,
+      storeFqdn,
+      query: mutation,
     })
 
-    expect(renderInfo).toHaveBeenCalledWith({
-      customSections: expect.arrayContaining([
-        expect.objectContaining({
-          title: 'Bulk Operation Created',
-        }),
-      ]),
+    expect(runBulkOperationMutation).toHaveBeenCalledWith({
+      adminSession: mockAdminSession,
+      query: mutation,
+      variables: undefined,
+    })
+    expect(runBulkOperationQuery).not.toHaveBeenCalled()
+  })
+
+  test('passes variables to mutation when provided with `--variables` flag', async () => {
+    const mutation = 'mutation productUpdate($input: ProductInput!) { productUpdate(input: $input) { product { id } } }'
+    const variables = ['{"input":{"id":"gid://shopify/Product/123","tags":["test"]}}']
+    const mockResponse = {
+      bulkOperation: successfulBulkOperation,
+      userErrors: [],
+    }
+    vi.mocked(runBulkOperationMutation).mockResolvedValue(mockResponse as any)
+
+    await executeBulkOperation({
+      app: mockApp,
+      storeFqdn,
+      query: mutation,
+      variables,
+    })
+
+    expect(runBulkOperationMutation).toHaveBeenCalledWith({
+      adminSession: mockAdminSession,
+      query: mutation,
+      variables,
+    })
+  })
+
+  test('renders success message when bulk operation is created', async () => {
+    const query = '{ products { edges { node { id } } } }'
+    const mockResponse = {
+      bulkOperation: successfulBulkOperation,
+      userErrors: [],
+    }
+    vi.mocked(runBulkOperationQuery).mockResolvedValue(mockResponse as any)
+    await executeBulkOperation({
+      app: mockApp,
+      storeFqdn,
+      query,
     })
 
     expect(renderSuccess).toHaveBeenCalledWith({
@@ -62,7 +141,8 @@ describe('executeBulkOperation', () => {
     })
   })
 
-  test('executeBulkOperation renders warning when user errors are present', async () => {
+  test('renders warning when user errors are present', async () => {
+    const query = '{ products { edges { node { id } } } }'
     const mockResponse = {
       bulkOperation: null,
       userErrors: [
--- a/packages/app/src/cli/services/bulk-operations/execute-bulk-operation.ts
+++ b/packages/app/src/cli/services/bulk-operations/execute-bulk-operation.ts
@@ -1,26 +1,31 @@
 import {runBulkOperationQuery} from './run-query.js'
+import {runBulkOperationMutation} from './run-mutation.js'
 import {AppLinkedInterface} from '../../models/app/app.js'
 import {renderSuccess, renderInfo, renderWarning} from '@shopify/cli-kit/node/ui'
 import {outputContent, outputToken} from '@shopify/cli-kit/node/output'
+import {ensureAuthenticatedAdmin} from '@shopify/cli-kit/node/session'
+import {parse} from 'graphql'
 
 interface ExecuteBulkOperationInput {
   app: AppLinkedInterface
   storeFqdn: string
   query: string
+  variables?: string[]
 }
 
 export async function executeBulkOperation(input: ExecuteBulkOperationInput): Promise<void> {
-  const {app, storeFqdn, query} = input
+  const {app, storeFqdn, query, variables} = input
 
   renderInfo({
     headline: 'Starting bulk operation.',
     body: `App: ${app.name}\nStore: ${storeFqdn}`,
   })
 
-  const bulkOperationResponse = await runBulkOperationQuery({
-    storeFqdn,
-    query,
-  })
+  const adminSession = await ensureAuthenticatedAdmin(storeFqdn)
+
+  const bulkOperationResponse = isMutation(query)
+    ? await runBulkOperationMutation({adminSession, query, variables})
+    : await runBulkOperationQuery({adminSession, query})
 
   if (bulkOperationResponse?.userErrors?.length) {
     const errorMessages = bulkOperationResponse.userErrors
@@ -62,4 +67,11 @@ export async function executeBulkOperation(input: ExecuteBulkOperationInput): Pr
       body: 'Congrats!',
     })
   }
+}
+
+function isMutation(graphqlOperation: string): boolean {
+  const document = parse(graphqlOperation)
+  const firstOperation = document.definitions.find((def) => def.kind === 'OperationDefinition')
+
+  return firstOperation?.kind === 'OperationDefinition' && firstOperation.operation === 'mutation'
 }
--- a/packages/app/src/cli/services/bulk-operations/stage-file.ts
+++ b/packages/app/src/cli/services/bulk-operations/stage-file.ts
@@ -1,0 +1,123 @@
+import {
+  StagedUploadsCreate,
+  StagedUploadsCreateMutation,
+  StagedUploadsCreateMutationVariables,
+} from '../../api/graphql/bulk-operations/generated/staged-uploads-create.js'
+import {adminRequestDoc} from '@shopify/cli-kit/node/api/admin'
+import {AdminSession} from '@shopify/cli-kit/node/session'
+import {formData, fetch} from '@shopify/cli-kit/node/http'
+import {AbortError} from '@shopify/cli-kit/node/error'
+
+interface StageFileOptions {
+  adminSession: AdminSession
+  jsonVariables?: string[]
+}
+
+export async function stageFile(options: StageFileOptions): Promise<string> {
+  const {adminSession, jsonVariables = []} = options
+
+  const buffer = convertJsonToJsonlBuffer(jsonVariables)
+  const filename = 'bulk-variables.jsonl'
+  const size = buffer.length
+
+  const response = await requestStagedUpload(adminSession, filename, size)
+  const target = validateStagedUploadResponse(response)
+
+  await uploadFileToStagedUrl(buffer, target.url, target.parameters, filename)
+
+  return target.stagedUploadKey
+}
+
+function convertJsonToJsonlBuffer(jsonVariables: string[]): Buffer {
+  const jsonlContent = `${jsonVariables.join('\n')}\n`
+  return Buffer.from(jsonlContent, 'utf-8')
+}
+
+async function requestStagedUpload(
+  adminSession: AdminSession,
+  filename: string,
+  size: number,
+): Promise<StagedUploadsCreateMutation> {
+  return adminRequestDoc<StagedUploadsCreateMutation, StagedUploadsCreateMutationVariables>({
+    query: StagedUploadsCreate,
+    session: adminSession,
+    variables: {
+      input: [
+        {
+          filename,
+          fileSize: size.toString(),
+          httpMethod: 'POST',
+          mimeType: 'text/jsonl',
+          resource: 'BULK_MUTATION_VARIABLES',
+        },
+      ],
+    },
+  })
+}
+
+function validateStagedUploadResponse(response: StagedUploadsCreateMutation): {
+  url: string
+  resourceUrl: string
+  parameters: {name: string; value: string}[]
+  stagedUploadKey: string
+} {
+  if (!response.stagedUploadsCreate) {
+    throw new AbortError('No response received from stagedUploadsCreate mutation')
+  }
+
+  if (response.stagedUploadsCreate.userErrors.length > 0) {
+    const errors = response.stagedUploadsCreate.userErrors
+      .map((error: {field?: string[] | null; message: string}) => error.message)
+      .join(', ')
+    throw new AbortError(`Failed to create staged upload: ${errors}`)
+  }
+
+  const target = response.stagedUploadsCreate.stagedTargets?.[0]
+  if (!target) {
+    throw new AbortError('No staged upload target returned from Shopify')
+  }
+
+  if (!target.url || !target.resourceUrl) {
+    throw new AbortError('Invalid staged upload target: missing required URLs')
+  }
+
+  const stagedUploadKey = target.parameters.find((param) => param.name === 'key')?.value
+  if (!stagedUploadKey) {
+    throw new AbortError('No key parameter found in staged upload target')
+  }
+
+  return {
+    url: target.url,
+    resourceUrl: target.resourceUrl,
+    parameters: target.parameters,
+    stagedUploadKey,
+  }
+}
+
+async function uploadFileToStagedUrl(
+  fileContents: Buffer,
+  uploadUrl: string,
+  parameters: {name: string; value: string}[],
+  filename: string,
+): Promise<void> {
+  const form = formData()
+
+  for (const param of parameters) {
+    form.append(param.name, param.value)
+  }
+
+  form.append('file', fileContents, {
+    filename,
+    contentType: 'text/jsonl',
+  })
+
+  const uploadResponse = await fetch(uploadUrl, {
+    method: 'POST',
+    body: form,
+  })
+
+  if (!uploadResponse.ok) {
+    const errorText = await uploadResponse.text()
+    throw new AbortError(`Failed to upload file to staged URL: ${uploadResponse.statusText}\n${errorText}`)
+  }
+}
--- a/packages/app/src/cli/api/graphql/bulk-operations/generated/bulk-operation-run-mutation.ts
+++ b/packages/app/src/cli/api/graphql/bulk-operations/generated/bulk-operation-run-mutation.ts
@@ -1,0 +1,125 @@
+/* eslint-disable @typescript-eslint/consistent-type-definitions, @typescript-eslint/no-redundant-type-constituents */
+import * as Types from './types.js'
+
+import {TypedDocumentNode as DocumentNode} from '@graphql-typed-document-node/core'
+
+export type BulkOperationRunMutationMutationVariables = Types.Exact<{
+  mutation: Types.Scalars['String']['input']
+  stagedUploadPath: Types.Scalars['String']['input']
+  clientIdentifier?: Types.InputMaybe<Types.Scalars['String']['input']>
+}>
+
+export type BulkOperationRunMutationMutation = {
+  bulkOperationRunMutation?: {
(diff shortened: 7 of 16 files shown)


ericlee878:
* **#6643** <a href="https://app.graphite.com/github/pr/Shopify/cli/6643?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#6663](https://github.com/Shopify/cli/pull/6663) <a href="https://app.graphite.com/github/pr/Shopify/cli/6663?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#6667](https://github.com/Shopify/cli/pull/6667) <a href="https://app.graphite.com/github/pr/Shopify/cli/6667?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#6596** <a href="https://app.graphite.com/github/pr/Shopify/cli/6596?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Shopify/cli/6596?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/app/src/cli/commands/app/execute.ts
```diff
@@ -37,7 +37,7 @@ export default class Execute extends AppLinkedCommand {
     await executeBulkOperation({
       app: appContextResult.app,
       storeFqdn: store.shopDomain,
-      query: flags.query,
+      ...flags,
```
jordanverasamy: Since we will doing more complicated flag stuff later (e.g. consuming `query` from _either_ the flag or from STDIN) it probably makes sense to continue passing each parameter to `executeBulkOperation` explicitly. It won't be the case for long that we just want every flag as-is. Thoughts?

packages/app/src/cli/services/bulk-operations/stage-file.ts
```diff
@@ -0,0 +1,123 @@
...
+import {AbortError} from '@shopify/cli-kit/node/error'
+
+interface StageFileOptions {
+  adminSession: AdminSession
+  filePath?: string
+  jsonVariables?: string
+  filename?: string
+}
+
+export async function stageFile(options: StageFileOptions): Promise<string> {
```
jordanverasamy, line 16: This is a really long function -- can we think of ways to break it up into a bunch of small functions to make it easier to read?

packages/app/src/cli/services/bulk-operations/execute-bulk-operation.ts
```diff
@@ -63,3 +67,8 @@ export async function executeBulkOperation(input: ExecuteBulkOperationInput): Pr
     })
   }
 }
+
+function isMutation(graphqlOperation: string): boolean {
```
jordanverasamy, line 72: We probably want to use a GraphQL parser to determine this instead of just checking `.startsWith('mutation')`, but I might be ok with shipping this scrappy version for now and doing that improvement separately. Thoughts? How complicated do you think that would be?

jordanverasamy (commented):
Really great start. Here's a first pass review, didn't look at tests yet. Will do a deeper look later this afternoon. We can pair as needed to address feedback. Great work so far!

packages/app/src/cli/services/bulk-operations/execute-bulk-operation.ts
```diff
@@ -1,26 +1,30 @@
...
 
-  const bulkOperationResponse = await runBulkOperationQuery({
-    storeFqdn,
-    query,
-  })
+  const adminSession = await ensureAuthenticatedAdmin(storeFqdn)
+
+  const bulkOperationResponse = isMutation(query)
+    ? await runBulkOperationMutation({adminSession, query, variables})
+    : await runBulkOperationQuery({adminSession, query, variables})
```
jordanverasamy:
```suggestion
    : await runBulkOperationQuery({adminSession, query})
```

Thoughts on just not passing this in? We could put [this check](https://github.com/Shopify/cli/pull/6622/files#r2529191925) right at the start of this function.

packages/app/src/cli/commands/app/execute.ts
ericlee878: Now that I look at it and other patterns in the codebase, I agree. Changed to reflect this!

packages/app/src/cli/services/bulk-operations/execute-bulk-operation.ts
ericlee878: Implementing the GraphQL parser wasn't too difficult, so I went ahead and did it.

jordanverasamy: Oh sweet! That looks great, ty!

packages/app/src/cli/services/bulk-operations/execute-bulk-operation.ts
```diff
@@ -63,3 +68,12 @@ export async function executeBulkOperation(input: ExecuteBulkOperationInput): Pr
     })
   }
 }
+
+function isMutation(graphqlOperation: string): boolean {
+  const document = parse(graphqlOperation)
+  const firstOperation = document.definitions.find(
```
jordanverasamy: Interesting... I forgot that one document could define multiple operations. Maybe we should have a followup ticket to error if `document.definitions != 1`? I don't think it ever makes sense to allow something like this for example:

```graphql
mutation {
  ...
}

mutation {
  ...
}
```

jordanverasamy: Not sure on the details of exactly what we should check, but just looking at this makes me really wary of how much freedom the user has to pass in subtly broken stuff here -- it's a thing we should note down and think about later.

Hmm, actually maybe we don't need to worry, because Core will probably reject documents like this?

jordanverasamy: Greg would definitely have some guidance 😂

ericlee878: Hmm okok I created a future [ticket](https://github.com/shop/issues-api-foundations/issues/1047?issue=shop%7Cissues-api-foundations%7C1114) for this!

packages/app/src/cli/services/bulk-operations/execute-bulk-operation.ts
ericlee878: Hmm ok I was thinking about both approaches. The code seemed neater this way but now that I think about it, this logic does seem like it belongs in the `execute-bulk-operation.ts` file. I will fix it in this [PR](https://app.graphite.com/github/pr/Shopify/cli/6622/Add-erroring-for-when-variables-are-given-for-a-BulkOps-query#comment-PRRC_kwDOHibhHs6WwGf1).

jordanverasamy: sweet, thanks :))

packages/app/src/cli/services/bulk-operations/stage-file.ts
ericlee878: Broken down!

packages/app/src/cli/api/graphql/bulk-operations/mutations/bulk-operation-run-mutation.graphql
```diff
@@ -0,0 +1,34 @@
+mutation BulkOperationRunMutation(
+  $mutation: String!
+  $stagedUploadPath: String!
+  $clientIdentifier: String
+) {
+  bulkOperationRunMutation(
+    mutation: $mutation
+    stagedUploadPath: $stagedUploadPath
+    # groupObjects has no effect for mutations, but we set it to false to keep consistent with query
```
jordanverasamy: [nit] is this worth it? i feel like it might be more confusing than just not having it, but i dunno i guess

packages/app/src/cli/services/bulk-operations/stage-file.ts
```diff
@@ -0,0 +1,176 @@
...
+  const fileContents = await prepareFileContents(buffer, filePath)
+  await uploadFileToStagedUrl(fileContents, target.url, target.parameters, filename)
+
+  return extractStagedUploadKey(target.parameters)
+}
+
+/**
+ * Converts JSON variables string to JSONL buffer format
+ */
+function convertJsonToJsonlBuffer(jsonVariables?: string): Buffer | undefined {
```
jordanverasamy: I'm pretty confused by this function, can you walk through it with me?

(I think?) the caller should have to provide the variables already in JSONL format, but it seems like we're manually converting from some kind of array-JSON format into JSONL here, which doesn't make sense to me.

packages/app/src/cli/services/bulk-operations/stage-file.ts
```diff
@@ -0,0 +1,176 @@
...
+  filePath?: string
+  jsonVariables?: string
+  filename?: string
+}
+
+export async function stageFile(options: StageFileOptions): Promise<string> {
+  const {adminSession, filePath, jsonVariables, filename: providedFilename} = options
+
+  const buffer = convertJsonToJsonlBuffer(jsonVariables)
+  const filename = providedFilename ?? (filePath ? basename(filePath) : 'bulk-variables.jsonl')
```
jordanverasamy: Where's all this complexity coming from? Not sure I understand this part. It seems like we're covering the possibility of having been provided a filename, but there isn't any command-level functionality to provide a filename...? Am I just missing it?

packages/app/src/cli/services/bulk-operations/stage-file.ts
```diff
@@ -0,0 +1,176 @@
...
+
+  const variablesArray: {[key: string]: unknown}[] = JSON.parse(jsonVariables)
+  const jsonlContent = `${variablesArray.map((vars) => JSON.stringify(vars)).join('\n')}\n`
+  return Buffer.from(jsonlContent, 'utf-8')
+}
+
+/**
+ * Calculates the file size from buffer or file path
+ */
+async function calculateFileSize(buffer?: Buffer, filePath?: string): Promise<number> {
```
jordanverasamy: Similar question here -- it seems like we're maintaining the possibility that we have either a buffer or a file path, but in practice we will always have a buffer, right?

packages/app/src/cli/services/bulk-operations/stage-file.ts
```diff
@@ -0,0 +1,176 @@
...
+    url: target.url,
+    resourceUrl: target.resourceUrl,
+    parameters: target.parameters,
+  }
+}
+
+/**
+ * Prepares file contents for upload by reading from buffer or file path
+ */
+async function prepareFileContents(buffer?: Buffer, filePath?: string): Promise<Buffer> {
```
jordanverasamy: Same thing here -- feels like this is more complex than it has to be for the first pass.

packages/app/src/cli/services/bulk-operations/stage-file.ts
```diff
@@ -0,0 +1,176 @@
...
+
+  return Buffer.alloc(0)
+}
+
+/**
+ * Extracts the staged upload key from target parameters
+ */
+function extractStagedUploadKey(parameters: {name: string; value: string}[]): string {
+  const stagedUploadKey = parameters.find((param) => param.name === 'key')?.value
+  if (stagedUploadKey === undefined) {
```
jordanverasamy: Should probably do this check inside the validate function, it would match the other stuff in there.

packages/app/src/cli/services/bulk-operations/stage-file.ts
```diff
@@ -0,0 +1,176 @@
...
+
+  // Type assertion: we've validated that url and resourceUrl are non-null strings
+  return {
+    url: target.url,
+    resourceUrl: target.resourceUrl,
+    parameters: target.parameters,
+  }
+}
+
+/**
```
jordanverasamy: Is having these comments describing each function standard in this codebase? Let's make sure we follow the patterns established elsewhere.

packages/app/src/cli/services/bulk-operations/stage-file.ts
jordanverasamy: Hmm, maybe what you have makes sense after all?

Here's what I see in the CLI doc:

```
  -v, --variables=<value>               The values for the GraphQL variables in bulk mutation, in JSON format. Cannot be used with bulk queries. (Multiple values allowed)
```

So maybe Nick is thinking of this being JSON and not JSONL? But we should confirm exactly how he wants it to work and why.

packages/app/src/cli/api/graphql/bulk-operations/mutations/bulk-operation-run-mutation.graphql
ericlee878: I thought it would be nice to be able to understand why it's false? Also, if someone gets confused why it's there, they can go to [`bulk-operation-run-query.graphql`](https://github.com/Shopify/cli/blob/main/packages/app/src/cli/api/graphql/bulk-operations/mutations/bulk-operation-run-query.graphql#L4) and there is a comment that explains why it's false.

packages/app/src/cli/services/bulk-operations/stage-file.ts
ericlee878: Yes! That document is what I based this off of. I assumed this design decision was made because JSON format is more convenient for users when manually inputting the variables through a CLI flag. But... I don't know how true this is. What do you think @nickwesselman ?

packages/app/src/cli/api/graphql/bulk-operations/mutations/bulk-operation-run-mutation.graphql
jordanverasamy: I feel like this would be the least confusing:

- in `bulk-operation-run-query`, pass it as false, and give the explanation
- in `bulk-operation-run-mutation`, don't pass it at all, and don't leave a comment

It doesn't matter for mutations anyway, so making someone go and read the other file where it _does_ matter just makes their life harder, I think? But very open to any argument you have for continuing to pass `false` here.

ericlee878: Wait does it not matter for mutations? I thought it did when writing this for some reason, but looking back at the [documentation](https://shopify.dev/docs/api/admin-graphql/latest/mutations/bulkoperationrunmutation), it seems like it's as necessary as any other parameter?

packages/app/src/cli/services/bulk-operations/stage-file.ts
nickwesselman: Yes the idea here was that if you are providing the values as CLI arguments, you could pass each in a separate arg, vs having to integrate newline formatting.

Maybe it's not a valid use case? My only thought is agentic demo generation case since it could avoid having to write a temp file.

If y'all don't think it's useful, I'm fine leaving off this arg.

ericlee878: I think this flag is at least useful for testing. So, I'm for keeping it. Just wanted to know if it was specified as JSON instead of JSONL format for any particular reason (since BulkOps uses JSONL format).

nickwesselman: Got it. Yeah JSONL through a command line sounds painful? But could be done more easily in a script.

packages/app/src/cli/services/bulk-operations/stage-file.ts
ericlee878: Ok this part needs to be more deeply thought through. Do you have any suggestions for a file naming convention?

packages/app/src/cli/services/bulk-operations/stage-file.ts
ericlee878: Agreed. I think for both cases, I was thinking of a possibility of adding a variable-file path but determined that was for a future PR. Will re-implement this in that future PR.

packages/app/src/cli/services/bulk-operations/stage-file.ts
ericlee878: Now that I look at it, I agree. Will refactor this file.

packages/app/src/cli/services/bulk-operations/stage-file.ts
ericlee878: Yes! Now that I look at it. This function seems unnecessary. Refactored to much cleaner code. Thanks!

packages/app/src/cli/services/bulk-operations/stage-file.ts
ericlee878: Will do. Sorry about that :sob:

packages/app/src/cli/api/graphql/bulk-operations/mutations/bulk-operation-run-mutation.graphql
jordanverasamy: Sadly I think that documentation is wrong? Also, this is confusing:

image

but I think having both "required" and "default: true" actually means its not required, and will default to true if not provided.

packages/app/src/cli/services/bulk-operations/stage-file.ts
jordanverasamy: Can you give me an example CLI command call so I can visualize what it would look like to pass e.g. 5 variable sets?

packages/app/src/cli/services/bulk-operations/stage-file.ts
jordanverasamy: Its ok no worries <3

packages/app/src/cli/services/bulk-operations/stage-file.ts
jordanverasamy: Yeah makes sense! Let's just pick one path to focus on to minimize complexity for now, and we an add the other one as a follow up :))

packages/app/src/cli/api/graphql/bulk-operations/mutations/bulk-operation-run-mutation.graphql
ericlee878: Oh ok... I'm thinking of creating a ticket for correcting this documentation. Where do you think is a good place to do this. [Here](https://github.com/orgs/shop/projects/208/views/34?pane=issue&itemId=3587480797&issue=shop%7Cissues-api-foundations%7C1045)?

packages/app/src/cli/services/bulk-operations/stage-file.ts
ericlee878: I don't know if this is a realistic example. But, I think it would look something like this?

```
shopify app execute 
--path <PATH_TO_APP> 
-query 'mutation updateProduct($id: ID!, $title: String!) { productUpdate(input: {id: $id, title: $title}) { product { id title } } }' 
-variables '[
        {"id":"gid://shopify/Product/1","title":"New Shirt"},
        {"id":"gid://shopify/Product/2","title":"Cool Pants"},
        {"id":"gid://shopify/Product/3","title":"Nice Hat"},
        {"id":"gid://shopify/Product/4","title":"Fancy Shoes"},
        {"id":"gid://shopify/Product/5","title":"Sweet Jacket"}
     ]'
```

packages/app/src/cli/services/bulk-operations/execute-bulk-operation.test.ts
```diff
@@ -25,7 +29,12 @@ describe('executeBulkOperation', () => {
     url: null,
   }
 
-  test('executeBulkOperation successfully runs', async () => {
+  beforeEach(() => {
+    vi.mocked(ensureAuthenticatedAdmin).mockResolvedValue(mockAdminSession)
+  })
+
+  test('runs query operation when GraphQL starts with query or curly brace', async () => {
```
jordanverasamy: [nit] This test only actually tests what happens when the GraphQL document starts with a curly brace. Maybe we want a second, separate test showing what happens when it starts with `query`?

packages/app/src/cli/services/bulk-operations/execute-bulk-operation.test.ts
```diff
@@ -39,21 +48,69 @@ describe('executeBulkOperation', () => {
...
     expect(runBulkOperationQuery).toHaveBeenCalledWith({
-      storeFqdn,
+      adminSession: mockAdminSession,
       query,
+      variables: undefined,
     })
+    expect(runBulkOperationMutation).not.toHaveBeenCalled()
+  })
+
+  test('runs mutation operation when GraphQL starts with mutation', async () => {
```
jordanverasamy:
```suggestion
  test('runs mutation operation when GraphQL document starts with mutation', async () => {
```

jordanverasamy (commented):
Another pass! Looking really good. I like the changes to `stageFile`, thanks for being patient with me on those :)

packages/app/src/cli/services/bulk-operations/stage-file.ts
jordanverasamy: What you have now after your last round of changes is totally good IMO!

ericlee878: Great!

packages/app/src/cli/services/bulk-operations/execute-bulk-operation.test.ts
ericlee878: Test added!

packages/app/src/cli/services/bulk-operations/stage-file.ts
jordanverasamy: I see, TY for the clean example! I understand now.

As far as I can tell, if we do it this way, here's the outcome for the user:

**Pros**: it's easier to read and type (don't need to e.g. manually type newlines, don't need to worry about escape characters or single vs double quotes

**Cons**: it's harder to script and reason about, there's more mental load (there's two different formats for different contexts, and you have to pick the right one, and you have to pay attention to whether something is "JSON variables" or "JSONL variables"

I think it's up to @nickwesselman whether that analysis is convincing, and which side of that tradeoff we should pick. For me personally the extra complexity doesn't seem worth it though.

nickwesselman: So I actually intended it to be a multi-value argument, vs a different JSON format. That way internally you can just concat the values to form JSONL.

```
shopify app bulk execute 
--path <PATH_TO_APP> 
--query 'mutation updateProduct($id: ID!, $title: String!) { productUpdate(input: {id: $id, title: $title}) { product { id title } } }' 
--variables '{"id":"gid://shopify/Product/1","title":"New Shirt"}'
--variables '{"id":"gid://shopify/Product/2","title":"Cool Pants"}'
--variables '{"id":"gid://shopify/Product/3","title":"Nice Hat"}'
--variables '{"id":"gid://shopify/Product/4","title":"Fancy Shoes"}'
--variables '{"id":"gid://shopify/Product/5","title":"Sweet Jacket"}'
```

I *think* that would cut down on the variability for both us and the user. WDYT?

packages/app/src/cli/api/graphql/bulk-operations/mutations/bulk-operation-run-mutation.graphql
jordanverasamy: I think that epic is for internal docs, but these docs are external -- I think this epic is more appropriate? https://github.com/shop/issues-api-foundations/issues/1043

packages/app/src/cli/services/bulk-operations/stage-file.ts
jordanverasamy: Oh that would be interesting and cool! If we can support that format I would be into it.

nickwesselman: Should be possible with `multiple: true` on the OCLIF flag definition:

https://oclif.io/docs/flags/

packages/app/src/cli/services/bulk-operations/stage-file.ts
```diff
@@ -0,0 +1,123 @@
...
+
+  const response = await requestStagedUpload(adminSession, filename, size)
+  const target = validateStagedUploadResponse(response)
+
+  await uploadFileToStagedUrl(buffer, target.url, target.parameters, filename)
+
+  return target.stagedUploadKey
+}
+
+function convertJsonToJsonlBuffer(jsonVariables: string[]): Buffer {
```
jordanverasamy, line 31: This is non-blocking feedback, but just thinking ahead.

When we implement the `variable-file` flag, we'll be given just straight up JSONL. So there will need to be a condition somewhere that knows how to do the right thing, regardless of whether we were given JSON or JSONL.

Where will be the easiest place for that condition to live?

I'd argue that we should unify things as soon as possible, so that we minimize how much of the code needs to know that both possibilities exist. For example, what if we immediately convert to JSONL if we were given JSON, and then that way all the rest of the code can just operate on JSONL only?

If this goes in here, then everything in the call chain before this will need to be able to pass along either JSONL or JSON. But if the conversion happens earlier, like e.g. in `execute-bulk-operation`, then the subsequent calls `run-mutation` and `stage-file` would be simpler and only care about JSON.

packages/app/src/cli/services/bulk-operations/stage-file.ts
ericlee878: Refactored to reflect this.

jordanverasamy:
🎩 :

```bash
pnpm shopify app execute \
--path="../my-test-app/my-test-app" \
--query='mutation createProduct($productSet: ProductSetInput!) { productSet(synchronous: true, input: $productSet) { product { id } } }' \
--variables='{ "productSet": { "title": "Winter hat" } }' \
--variables='{ "productSet": { "title": "Winter coat" } }' \
--variables='{ "productSet": { "title": "Winter pants" } }'
```

image

image

packages/app/src/cli/services/bulk-operations/stage-file.ts
ericlee878: Good point. We can try to minimize how much of the code needs to know in the [PR](https://github.com/orgs/shop/projects/208/views/34?pane=issue&itemId=139297173&issue=shop%7Cissues-api-foundations%7C1110) that implements `variable-file`. Will keep a note on this. Thanks for pointing this out!

ericlee878:
Thanks for adding this!

packages/app/src/cli/services/bulk-operations/execute-bulk-operation.ts
```diff
@@ -1,26 +1,31 @@
...
   renderInfo({
     headline: 'Starting bulk operation.',
     body: `App: ${app.name}\nStore: ${storeFqdn}`,
   })
 
-  const bulkOperationResponse = await runBulkOperationQuery({
-    storeFqdn,
-    query,
-  })
+  const adminSession = await ensureAuthenticatedAdmin(storeFqdn)
```
nickwesselman, line 24: We should add the `ensureAuthenticatedAdminAsApp` that was [created here](https://github.com/Shopify/cli/pull/5908/files#diff-19b0f5c6238cb4567d4897b88196fb65f7c53a3657b2b24a98ee901d507dbf4bR243-R264) for hack days. Right now you are executing with the CLI user's API key, which will not allow access to `$app` metafields and does not respect the app's configured scopes.

More details on the "why" [here](https://docs.google.com/document/d/157CvTbzxv5RDkAJCCeZMcWEPrsGLEdkFc8Hd48fUSM8/edit?tab=t.0#heading=h.1uor5oxehsdn).

packages/app/src/cli/commands/app/execute.ts
```diff
@@ -38,6 +38,7 @@ export default class Execute extends AppLinkedCommand {
       app: appContextResult.app,
       storeFqdn: store.shopDomain,
       query: flags.query,
+      variables: flags.variables,
```
nickwesselman, line 41: We should move this whole command to `shopify app bulk execute` as described [here](https://docs.google.com/document/d/157CvTbzxv5RDkAJCCeZMcWEPrsGLEdkFc8Hd48fUSM8/edit?tab=t.0#heading=h.80jr77x90oo3).

packages/app/src/cli/flags.ts
```diff
@@ -38,10 +38,17 @@ export const appFlags = {
...
+  variables: Flags.string({
+    char: 'v',
+    description:
+      'The values for any GraphQL variables in your mutation, in JSON format. Can be specified multiple times.',
+    env: 'SHOPIFY_FLAG_VARIABLES',
+    multiple: true,
+  }),
   store: Flags.string({
     char: 's',
     description: 'Store URL. Must be an existing development or Shopify Plus sandbox store.',
```
nickwesselman:
```suggestion
    description: 'The store domain. Must be an existing dev store.',
```

packages/app/src/cli/commands/app/execute.ts
jordanverasamy: Think we can do this in a followup? This PR is already scope crept 😅

ericlee878: Yes! Was thinking about making this a followup PR as well :+1:

nickwesselman: I'm OK with this as a follow up.

nickwesselman:
Mentioned elsewhere, but moving to `bulk execute` is OK as a follow up.

packages/app/src/cli/services/bulk-operations/execute-bulk-operation.ts
ericlee878: Wait just to clarify, is the `ensureAuthenticatedAdminAsApp` function not created in main yet? I can't find it anywhere. Or are you suggesting to use the `ensureAuthenticatedAdmin` function?

nickwesselman: You should create `ensureAuthenticatedAdminAsApp` if it is not on main.

jordanverasamy: Thanks for prototyping that for us Nick <3 appreciate it!

We're trying hard to split the work up into manageable chunks. Depending on how much work it is to implement `ensureAuthenticatedAdminAsApp` in a production-quality manner, are you good with us doing that as a followup too? This PR is already 900 lines, and I don't want us to get overwhelmed 😵

nickwesselman: Yeah that was @shauns's work from hack days!

Yeah if you want to stack it on that's fine.

ericlee878: Okok cool will create a follow up ticket for this as well. Thanks for being understanding!

jordanverasamy: Thanks! Just wanted to make sure we weren't eg introducing a potential security issue with this existing in its current state temporarily.

packages/app/src/cli/commands/app/execute.ts
ericlee878: Created [follow up ticket](https://github.com/shop/issues-api-foundations/issues/1127) for this.

ericlee878:
> Mentioned elsewhere, but moving to `bulk execute` is OK as a follow up.

Follow up [ticket](https://github.com/shop/issues-api-foundations/issues/1127)

packages/app/src/cli/services/bulk-operations/execute-bulk-operation.ts
ericlee878: Created [follow up ticket](https://github.com/orgs/shop/projects/208/views/34?pane=issue&itemId=140363951&issue=shop%7Cissues-api-foundations%7C1128) for this.

nickwesselman (approved):
I'm OK with my remaining requests as follow-ups

packages/app/src/cli/services/bulk-operations/execute-bulk-operation.ts
```diff
@@ -1,26 +1,31 @@
...
+
+  const bulkOperationResponse = isMutation(query)
+    ? await runBulkOperationMutation({adminSession, query, variables})
+    : await runBulkOperationQuery({adminSession, query})
```
isaacroldan, line 28: can't queries also have variables?

nickwesselman: Not bulk queries. We would need to materialize them in the client before sending them off.

PR #6605, opened 2025-11-11
Fix theme dev failing analytics

EvilGenius13:

Closes: https://github.com/Shopify/developer-tools-team/issues/902

Analytics aren't logging when we have a `theme dev` error during bootup or a background promise rejection during a session. This is because we call `process.exit(1)` (process.kill), which ends the program immediately before the `finally `block in `theme-command.ts` can fire off the analytics event.

Implementing a deferred promise pattern using `Promise.withResolvers()`. This is new to Node 22, but I've polyfilled it since we still support Node 18+ as a minimum version.

`Promise.withResolvers` allows us to pass around `resolve/reject` around and have it called outside of the original promise. This lets us create a background promise that will only be called when we have an error.

**How it works:**
`Promise.withResolvers()` returns an object with three properties:
- promise - a standard promise
- resolve - a function to resolve that promise
- reject - a function to reject that promise

These resolve/reject functions can be passed around and called from anywhere in the code, not just inside the promise. This lets us create a "control promise" that waits indefinitely unless we explicitly reject it.

**The flow:**
1. User runs `theme dev`
2. `dev.ts` calls `setupDevServer()` in `theme-environment.ts`
3. In `setupDevServer()`: We create a deferred promise that never resolves, only rejects on fatal errors:
   - `const {promise: backgroundJobPromise, reject: rejectBackgroundJob} = promiseWithResolvers<never>()`
4. We pass the `rejectBackgroundJob` function down through multiple layers:
      - `setupDevServer() → ensureThemeEnvironmentSetup() → handleThemeEditorSync() → reconcileAndPollThemeEditorChanges() → pollThemeEditorChanges()`
5. Each layer uses this reject function instead of calling `process.exit(1)` when fatal errors occur
6. Back in dev.ts: We await both the server setup AND the background promise:
```
   await Promise.all([
          backgroundJobPromise // Hangs forever unless reject() is called    
          renderDevSetupProgress().then(serverStart).then(...)   
  ])
```
7. Since the background promise never resolves, `Promise.all()` waits forever keeping the function alive. If an error occurs, `rejectBackgroundJob(error)` is called, which:
- Rejects the backgroundJobPromise
- Causes Promise.all() to reject
- Bubbles the error up to theme-command.ts
- Triggers the finally block → analytics get logged
- Process exits gracefully

### How to test your changes?

In `theme-command.ts` add some logging to the `finally` block for analytics
```
  finally {
     console.log('🔍 Analytics logging...')
     await this.logAnalyticsData(session)
     console.log('✅ Analytics logged!')
   }
```

In `theme-environment.ts` replace the `const remoteCheckSumsPromise` with 
```
const remoteChecksumsPromise = fetchChecksums(theme.id, ctx.session)
     .then(() => { throw new Error('TEST: Force failure') })
```

Run `theme dev`

### Measuring impact

### Checklist

### DIFF
--- a/packages/theme/src/cli/utilities/theme-environment/theme-environment.ts
+++ b/packages/theme/src/cli/utilities/theme-environment/theme-environment.ts
@@ -6,34 +6,12 @@ import {reconcileAndPollThemeEditorChanges} from './remote-theme-watcher.js'
 import {uploadTheme} from '../theme-uploader.js'
 import {renderTasksToStdErr} from '../theme-ui.js'
 import {renderThrownError} from '../errors.js'
+import {promiseWithResolvers} from '../../polyfills/promiseWithResolvers.js'
 import {createApp, defineEventHandler, defineLazyEventHandler, toNodeListener, handleCors} from 'h3'
 import {fetchChecksums} from '@shopify/cli-kit/node/themes/api'
 import {createServer} from 'node:http'
 import type {Checksum, Theme} from '@shopify/cli-kit/node/themes/types'
 import type {DevServerContext} from './types.js'
-
-// Polyfill for Promise.withResolvers
-// Can remove once our minimum supported Node version is 22
-interface PromiseWithResolvers<T> {
-  promise: Promise<T>
-  resolve: (value: T | PromiseLike<T>) => void
-  reject: (reason?: unknown) => void
-}
-
-function promiseWithResolvers<T>(): PromiseWithResolvers<T> {
-  if (typeof Promise.withResolvers === 'function') {
-    return Promise.withResolvers<T>()
-  }
-
-  let resolve!: (value: T | PromiseLike<T>) => void
-  let reject!: (reason?: unknown) => void
-  const promise = new Promise<T>((_resolve, _reject) => {
-    resolve = _resolve
-    reject = _reject
-  })
-
-  return {promise, resolve, reject}
-}
 
 export function setupDevServer(theme: Theme, ctx: DevServerContext) {
   const {promise: backgroundJobPromise, reject: rejectBackgroundJob} = promiseWithResolvers<never>()
--- a/packages/theme/src/cli/polyfills/promiseWithResolvers.ts
+++ b/packages/theme/src/cli/polyfills/promiseWithResolvers.ts
@@ -1,0 +1,22 @@
+interface PromiseWithResolvers<T> {
+  promise: Promise<T>
+  resolve: (value: T | PromiseLike<T>) => void
+  reject: (reason?: unknown) => void
+}
+
+// Polyfill for Promise.withResolvers
+// Can remove once our minimum supported Node version is 22
+export function promiseWithResolvers<T>(): PromiseWithResolvers<T> {
+  if (typeof Promise.withResolvers === 'function') {
+    return Promise.withResolvers<T>()
+  }
+
+  let resolve!: (value: T | PromiseLike<T>) => void
+  let reject!: (reason?: unknown) => void
+  const promise = new Promise<T>((_resolve, _reject) => {
+    resolve = _resolve
+    reject = _reject
+  })
+
+  return {promise, resolve, reject}
+}


packages/theme/src/cli/services/dev.ts
```diff
@@ -97,23 +97,7 @@ export async function dev(options: DevOptions) {
...
-  }
-
-  await renderDevSetupProgress()
-  await serverStart()
-
-  renderLinks(urls)
-  if (options.open) {
-    openURLSafely(urls.local, 'development server')
-  }
+  const {serverStart, renderDevSetupProgress, backgroundJobPromise} = setupDevServer(options.theme, ctx)
```
EvilGenius13, line 122: I removed the `if options` and `if !options` for `theme-editor-sync` as we are now awaiting `storefrontPasswordPromise` earlier in the code and this is no longer necessary.

packages/theme/src/cli/utilities/theme-environment/theme-environment.ts
```diff
@@ -25,31 +50,41 @@ export function setupDevServer(theme: Theme, ctx: DevServerContext) {
...
-  const abort = createAbortCatchError('Failed to perform the initial theme synchronization.')
+function ensureThemeEnvironmentSetup(
+  theme: Theme,
+  ctx: DevServerContext,
+  rejectBackgroundJob: (reason?: unknown) => void,
+) {
+  const abort = (error: Error): never => {
+    renderThrownError('Failed to perform the initial theme synchronization.', error)
+    rejectBackgroundJob(error)
+    // Return a permanently pending promise to stop execution without throwing
```
EvilGenius13: This stops all the layers using the `reject` from individually outputting the error and polluting the terminal.

packages/theme/src/cli/utilities/theme-environment/theme-environment.ts
```diff
@@ -25,31 +50,41 @@ export function setupDevServer(theme: Theme, ctx: DevServerContext) {
...
+    return new Promise<never>(() => {}) as never
+  }
 
-  const remoteChecksumsPromise = fetchChecksums(theme.id, ctx.session).catch(abort)
+  const remoteChecksumsPromise = fetchChecksums(theme.id, ctx.session)
```
frandiox, line 49: I thiiiink we still need the `.catch(abort)` in this one in case it fails fetching checksums?
You can make a test and mock fetcChecksums return to be a promise rejected after a timeout and see if it works or not 👍

packages/theme/src/cli/utilities/theme-environment/theme-environment.ts
frandiox: Do you know which part of the code starts showing errors if we don't return a promise here? Is it the `ctx.localThemeFileSystem.startWatcher` that we can within `setupDevServer`?

frandiox (commented):
Good job! This was precisely the suggestion 💯 

One more thing for you to think about it: if we don't like passing the reject callback down multiple levels, an alternative for this case would be relying on [AsyncLocalStorage](https://nodejs.org/api/async_context.html#class-asynclocalstorage).

Basically, you'd pass the reject callback in the context as something like

```js
const {promise, reject} = Promise.withResolvers();
const {serverStart, renderDevSetupProcess} = await asyncStorage.run({exitProcess: reject}, () => setupDevServer(...))

await Promise.all([promise, serverStart().then(...)])
```

And anywhere in the logic we could run `asyncLocalStorage.getStore().exitProcess(error)`.

In fact, maybe this is something that should go levels above in the same place we define our try/catch wrapper for the analytics: that logic would provide a context that gives access to an infinite promise in case the process should never end (like here when we run a server), and a reject callback to end the process in a controlled way.

Not saying we *should* do this, but perhaps worth considering it in this situation 👍

packages/theme/src/cli/utilities/theme-environment/theme-environment.ts
EvilGenius13: This is mainly because we have all the catch blocks and in our abort method we throw, so we throw, the next `.catch` grabs it and calls the abort error and so on.

packages/theme/src/cli/utilities/theme-environment/theme-environment.ts
EvilGenius13: I double checked and when `fetchChecksums` fails it passes down to the `reconcilePromise` and rejects properly!

EvilGenius13:
> Good job! This was precisely the suggestion 💯
> 
> One more thing for you to think about it: if we don't like passing the reject callback down multiple levels, an alternative for this case would be relying on [AsyncLocalStorage](https://nodejs.org/api/async_context.html#class-asynclocalstorage).
> 
> Basically, you'd pass the reject callback in the context as something like
> 
> ```js
> const {promise, reject} = Promise.withResolvers();
> const {serverStart, renderDevSetupProcess} = await asyncStorage.run({exitProcess: reject}, () => setupDevServer(...))
> 
> await Promise.all([promise, serverStart().then(...)])
> ```
> 
> And anywhere in the logic we could run `asyncLocalStorage.getStore().exitProcess(error)`.
> 
> In fact, maybe this is something that should go levels above in the same place we define our try/catch wrapper for the analytics: that logic would provide a context that gives access to an infinite promise in case the process should never end (like here when we run a server), and a reject callback to end the process in a controlled way.
> 
> Not saying we _should_ do this, but perhaps worth considering it in this situation 👍

Okay I checked it out. This is a super neat idea as well. IMO I feel like the current promise passdown method will be a bit easier to debug if we ever need compared to exiting via local storage where it won't be as easily traceable.

If it's okay with you I'd like to keep the current implementation.

dejmedus (commented):
🎩 went great! I've had the snapbuild dev server running since yesterday, it stayed up  and seems to be working exactly as I'd expect. I also built the branch and tried to force some errors and was able to see `logAnalyticsData` now being called

EvilGenius13:
>Do you know which part of the code starts showing errors if we don't return a promise here? Is it the ctx.localThemeFileSystem.startWatcher that we can within setupDevServer?

Without the extra promise, the rejection propagates through each of the `rejectPromise` areas which causes the error to render multiple times. This promise stop the initial chain and creates a single spot where we throw the error and you see it rendered in the terminal. I added a better comment above it to explain why it's there!

packages/theme/src/cli/utilities/theme-environment/theme-environment.ts
```diff
@@ -5,16 +5,41 @@ import {getProxyHandler} from './proxy.js'
...
+  let resolve!: (value: T | PromiseLike<T>) => void
+  let reject!: (reason?: unknown) => void
+  const promise = new Promise<T>((_resolve, _reject) => {
+    resolve = _resolve
+    reject = _reject
+  })
+
+  return {promise, resolve, reject}
+}
+
```
graygilmore: (nit) Might be worth chucking this whole function into a separate file. Something like `packages/theme/src/cli/polyfills/promise-with-resolvers.ts`. Makes it even more explicitly that this can be temporary.

packages/theme/src/cli/services/dev.ts
```diff
@@ -145,6 +129,18 @@ export async function dev(options: DevOptions) {
         break
     }
   })
+
+  await Promise.all([
```
graygilmore, line 155: > Since the background promise never resolves, Promise.all() waits forever keeping the function alive

Is there a downside to this? Could a user end up with thousands of open promises that never resolve?

EvilGenius13: Great question! In our case, no we wouldn't. The initial promise created with the `PromiseWithResolver` is passed around and used to reject errors. Every time a file changes or something like a hot-reload, it's all using the same reject function so they don't create a new error. This promise will just hold as a no-op (assuming no error) until you force quit.

packages/theme/src/cli/utilities/theme-environment/theme-environment.ts
EvilGenius13: Good idea, I moved it!

PR #6606, opened 2025-11-11
prototype identity client segmentation

alexanderMontague:

Fixes #0000 

### Measuring impact

### Checklist


MitchDickinson (commented):
Nice work. Approach looks solid 👏

packages/cli-kit/src/public/node/api/identity-client.ts
```diff
@@ -0,0 +1,300 @@
...
+  if (environment === Environment.Local) {
+    return 'e5380e02-312a-7408-5718-e07017e9cf52'
+  } else if (environment === Environment.Production) {
+    return 'fbdb2649-e327-4907-8f67-908d24cfd7e3'
+  } else {
+    return 'e5380e02-312a-7408-5718-e07017e9cf52'
+  }
+}
+
+export class ProdIdentityClient implements IdentityClientInterface {
```
MitchDickinson: IdentityServiceClient? "Prod" is a bit of a loaded term and could be confusing

packages/cli-kit/src/public/node/api/identity-client.ts
```diff
@@ -0,0 +1,300 @@
...
+
+    const payload: any = await res.json()
+
+    if (res.ok) return ok(payload)
+
+    return err({error: payload.error, store: params.store})
+  }
+}
+
+export class LocalIdentityClient implements IdentityClientInterface {
```
MitchDickinson: Maybe "FakeIdentityClient", similar comment as above.

packages/cli-kit/src/public/node/api/identity-client.ts
```diff
@@ -0,0 +1,300 @@
...
+  expiresIn: number
+  verificationUriComplete?: string
+  interval?: number
+}
+
+interface TokenRequestConfig {
+  [key: string]: string
+}
+type TokenRequestConfigResponse = Promise<Result<TokenRequestResult, {error: string; store?: string}>>
+interface IdentityClientInterface {
```
MitchDickinson: As someone who has very little identity domain knowledge, here's a couple comments when looking at the interface. I know none of this is stuff you directly introduced, but since we're introducing an abstraction here it's a good forcing function to make sure the abstraction is crystal clear

- `requestDeviceAuthorization` and `pollForDeviceAuthorization` seem to be invoked one-after-the-other, which suggests clients probably don't care about both methods existing and this is actually just one method in disguise. Having an interface where clients are required to call two methods in a particular order to do something useful is called "temporal coupling" and is a smell. Should it be just one method?
- I'm sure this is just an artifact of the refactoring, but the param/result types here are confusing. We have IdentityToken which is a zod schema type and TokenRequestConfigResponse which wraps a similarly defined (but not zod) TokenRequestResult. Is this intentional or important? If it's important, I wonder if the interface can expose a simple set of types and we can encapsulate the complexity here.
- What is TokenRequestConfig and how does it relate to IdentityToken? My assumption is that device authorization is the means to get the first token, and then that token can be refreshed/exchanged via the tokenRequest function? If that's right, I would expect a stronger link in the type system between what  tokenRequest accepts (`[key: string]: string` doesn't tell me much) and what the device authorization method returns. As a consumer of the interface it's pretty unclear how I would use tokenRequest. The type system doesn't make it obvious whether I should be calling for device authorization or tokenRequest so it feels like there's still a lot of "you must know how this works internally" going on before the interface can be used.

(I'm sure some of these comments lead to deep rabbit holes and I don't think we necessarily need to address them all in scope of this project, but good to have the convo anyway)

alexanderMontague: Will have to look into the suggestions, but at least the temporal coupling point seems very valid. Will consider this, thanks.

PR #6607, opened 2025-11-11
prevent possible CORS vulnerability

EvilGenius13:

`theme dev` had a lax CORS policy that _could_ allow any website to read authenticated data from a developer's local server.

### Two part fix:

**Part 1:**
Restrict Cors in the dev server middleware.
- Created an allowed origins list that consists of the local server address `localhost:port` and the production store `example-shop.myshopify.com`
- Allowed methods are `GET`, `PUT`, `HEAD`, `OPTIONS`.
- Added `credentials: true` as this forces the server to be explicit about which origin(s) is allowed.
- We only apply this change when an `origin` header is present. This is because the browser only sends that header when it's making a cross-origin request from another domain. 

**Part 2:**
Strip the CORS headers from proxied responses
- During testing I saw that we were getting responses back with `Access-Control-Allow-Origin: *`.
- We delete any headers involving CORS and our CORS middleware sets the correct headers.

### How to test your changes?

Build the branch and run `theme dev`

Open up your browser to a website that **isn't** your store or localhost. (i.e google.com) and open up the inspector. Open the console and paste

```
fetch('http://127.0.0.1:9292/products.json')
  .then(r => r.json())
  .then(data => console.log('✅ SUCCESS:', data))
  .catch(err => console.error('❌ BLOCKED:', err.message))
```

You should see blocked. You can try this with a current version of the CLI and it should pass (which we don't want).

### Measuring impact

### Checklist

### DIFF
--- a/.changeset/frank-lands-hunt.md
+++ b/.changeset/frank-lands-hunt.md
@@ -1,0 +1,5 @@
+---
+'@shopify/theme': patch
+---
+
+Fix a possible CORS vulnerability where a website could read data from the localhost server
--- a/packages/theme/src/cli/utilities/theme-environment/proxy.ts
+++ b/packages/theme/src/cli/utilities/theme-environment/proxy.ts
@@ -228,6 +228,15 @@ function patchProxiedResponseHeaders(ctx: DevServerContext, rawResponse: Respons
     response.headers.delete(header)
   }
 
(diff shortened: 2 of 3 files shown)


frandiox (approved):
Thanks!

PR #6612, opened 2025-11-12
Remove legacy app schema

gonzaloriestra:

We have a lot of code to support the legacy app schema in the CLI and templates, but that schema only works for the template state (before linking), so it doesn't make sense to keep it.

⚠️ These PRs for the templates must be merged first (the acceptance tests will pass then):

- https://github.com/Shopify/shopify-app-template-react-router/pull/127
- https://github.com/Shopify/shopify-app-template-react-router/pull/197
- https://github.com/Shopify/shopify-app-template-none/pull/24
- https://github.com/Shopify/shopify-app-template-remix/pull/1224
- https://github.com/Shopify/shopify-app-template-ruby/pull/194
- https://github.com/Shopify/shopify-app-template-php/pull/541
- https://github.com/Shopify/shopify-app-template-node/pull/1391

The [subscriptions-reference-app template](https://github.com/Shopify/subscriptions-reference-app/blob/main/shopify.app.toml) was ready.

- Make `client_id` required in app schema (templates must include `client_id = ""`)
- Remove support for legacy scopes field at root level (templates must include `[access_scopes]` section)
- Simplify the code to only support the new schema

### How to test your changes?

- `npm i -g --@shopify:registry=https://registry.npmjs.org @shopify/cli@0.0.0-snapshot-20260303103858`
- `shopify app init --template=https://github.com/Shopify/shopify-app-template-react-router#add-client-id-to-toml`
- `shopify app init --template=https://github.com/Shopify/shopify-app-template-react-router#add-client-id-to-toml-main-cli`
- `shopify app init --template=https://github.com/Shopify/shopify-app-template-none#add-client-id-to-toml`
- `shopify app init --template=https://github.com/Shopify/shopify-app-template-remix#add-client-id-to-toml`
- `shopify app init --template=https://github.com/Shopify/shopify-app-template-ruby#add-client-id-to-toml`
- `shopify app init --template=https://github.com/Shopify/shopify-app-template-node#add-client-id-to-toml`
- `shopify app init --template=https://github.com/Shopify/shopify-app-template-php#add-client-id-to-toml`
- `shopify app init --template https://github.com/Shopify/shopify-app-template-react-router#include-declarative-definition-and-client-id` (related to https://github.com/Shopify/cli/pull/6751)
- `shopify app init --template=https://github.com/Shopify/subscriptions-reference-app#fix-npm-i`
- `shopify app init --template none`
- `shopify app dev`, `shopify app deploy`...

### Measuring impact

### Checklist

### DIFF
--- a/packages/features/steps/create-app.steps.ts
+++ b/packages/features/steps/create-app.steps.ts
@@ -29,7 +29,7 @@ When(
         throw new Error(`Unknown app type: ${appType}`)
     }
 
-    const {stdout} = await exec(
+    const {stdout, stderr} = await exec(
       'node',
       [
         executables.createApp,
@@ -44,7 +44,8 @@ When(
       ],
       {env: {...process.env, ...this.temporaryEnv, NODE_OPTIONS: '', FORCE_COLOR: '0'}},
     )
-    const hyphenatedAppName = stdout?.match(/([\w-]+) is ready for you to build!/)?.[1] ?? appName
+    const output = `${stdout}\n${stderr}`
+    const hyphenatedAppName = output.match(/([\w-]+) is ready for you to build!/)?.[1] ?? appName
     this.appDirectory = path.join(this.temporaryDirectory, hyphenatedAppName)
 
     // Ensure .npmrc exists before appending to it


packages/features/steps/create-app.steps.ts
```diff
@@ -85,12 +51,6 @@ When(
     await fs.ensureFile(npmrcPath)
     // we need to disable this on CI otherwise pnpm will crash complaining that there is no lockfile
     await fs.appendFile(npmrcPath, 'frozen-lockfile=false\n')
-
-    // Ensure shopify.app.toml exists in the app directory
-    await ensureAppToml(this.appDirectory)
-
-    // Ensure package.json exists in the app directory
-    await ensurePackageJson(this.appDirectory)
```
gonzaloriestra, line 94: We were adding a default TOML and Json if for some reason they were not created, but that would mask actual errors.

nickwesselman:
What is the impact of doing this on previous versions? Do all versions 3.84+ support the new schema?

gonzaloriestra:
Yes, all versions above 3.84 (and much earlier) support the new schema, so this shouldn't have any impact.

packages/app/src/cli/models/app/app.ts
```diff
@@ -586,25 +496,24 @@ export class App<
   }
 
   get includeConfigOnDeploy() {
-    if (isLegacyAppSchema(this.configuration)) return false
     return this.configuration.build?.include_config_on_deploy
```
isaacroldan, line 493: I thought this was deprecated, is this only needed for 1P apps?

packages/app/src/cli/models/app/app.ts
```diff
@@ -225,9 +138,7 @@ export function getAppScopesArray(config: AppConfiguration) {
 }
 
 export function usesLegacyScopesBehavior(config: AppConfiguration) {
-  if (isLegacyAppSchema(config)) return true
-  if (isCurrentAppSchema(config)) return config.access_scopes?.use_legacy_install_flow ?? false
-  return false
+  return (config as CurrentAppConfiguration).access_scopes?.use_legacy_install_flow === true
 }
```
isaacroldan: I think this function is not used anywhere anymore

packages/app/src/cli/models/app/app.ts
```diff
@@ -540,9 +451,8 @@ export class App<
     return Boolean(frontendConfig ?? backendConfig)
   }
 
-  get appIsEmbedded() {
-    if (isCurrentAppSchema(this.configuration)) return this.configuration.embedded
-    return this.appIsLaunchable()
+  get appIsEmbedded(): boolean {
+    return (this.configuration as CurrentAppConfiguration).embedded
```
isaacroldan: it's weird that we need to cast to `CurrentAppConfiguration` everywhere, we should be able to simplify this and make `this.configuration` of type `CurrentAppConfiguration` directly no?

packages/app/src/cli/models/app/loader.ts
```diff
@@ -403,7 +400,7 @@ export async function loadAppUsingConfigurationState<TConfig extends AppConfigur
  */
 type LoadedAppConfigFromConfigState<TConfigState> = TConfigState extends AppConfigurationStateLinked
   ? CurrentAppConfiguration
-  : LegacyAppConfiguration
+  : never
```
isaacroldan: Seeing a change below, `AppConfigurationState = AppConfigurationStateLinked` then shouldn't this always be `CurrentAppConfiguration`? why the `never`?

packages/app/src/cli/models/app/loader.ts
```diff
@@ -905,16 +898,11 @@ interface AppConfigurationStateBasics {
 }
 
 export type AppConfigurationStateLinked = AppConfigurationStateBasics & {
-  state: 'connected-app'
   basicConfiguration: BasicAppConfigurationWithoutModules
+  isTemplateForm: boolean
```
isaacroldan: not sure if this is a bit contradictory, a Linked app shouldn't be in template form right?

packages/app/src/cli/services/app-context.ts
```diff
@@ -72,11 +72,10 @@ export async function linkedAppContext({
   let configState = await getAppConfigurationState(directory, userProvidedConfigName)
   let remoteApp: OrganizationApp | undefined
 
-  // If the app is not linked, force a link.
-  if (configState.state === 'template-only' || forceRelink) {
+  // If forceRelink is true, force a link.
+  if (forceRelink) {
```
isaacroldan: don't we need to still force a link if the app doesn't have a client id?

packages/app/src/cli/services/app/config/link.ts
```diff
@@ -339,7 +316,7 @@ async function loadConfigurationFileName(
   options: LinkOptions,
   localAppInfo: {
     appDirectory?: string
-    format: 'legacy' | 'current'
+    format: 'current'
```
isaacroldan: format is always `current` right? should we just remove the `format` property?

packages/app/src/cli/services/dev/app-events/file-watcher.test.ts
```diff
@@ -232,7 +233,11 @@ describe('file-watcher events', () => {
       const app = testAppLinked({
         allExtensions: [ext1, ext2, posExtension],
         directory: dir,
-        configuration: {path: joinPath(dir, '/shopify.app.toml'), scopes: ''},
+        configuration: {
+          path: joinPath(dir, '/shopify.app.toml'),
+          client_id: 'test-client-id',
+          access_scopes: {scopes: ''},
+        } as AppConfiguration,
```
isaacroldan: Why do you need the casting here? I think it should work without it :thinking_face: (same for other places where you cast to `AppConfiguration`)

packages/app/src/cli/models/app/loader.ts
```diff
@@ -905,16 +898,11 @@ interface AppConfigurationStateBasics {
...
 }
 
-type AppConfigurationStateTemplate = AppConfigurationStateBasics & {
-  state: 'template-only'
-  startingOptions: Omit<LegacyAppConfiguration, 'client_id'>
-}
-
-type AppConfigurationState = AppConfigurationStateLinked | AppConfigurationStateTemplate
+type AppConfigurationState = AppConfigurationStateLinked
```
isaacroldan: Can we just delete `AppConfigurationStateLinked`, and make everything just `AppConfigurationState`?

isaacroldan: And maybe instead of `isTemplateForm` we can call it `isLinked`if `client_id !== ''`?  
A toml might be missing the client_id and is unlinked, but doesn't mean is from a template

packages/app/src/cli/models/app/app.ts
gonzaloriestra: I think so, it's still required for Partners.

gonzaloriestra (commented):
@isaacroldan Thanks a lot for the careful review! I should have warned that there were many pending refactorings, but my plan was to tackle them later, as I already had this working and green.

Anyway, I think I solved all your comments now in [this commit](https://github.com/Shopify/cli/pull/6612/changes/7c40a281e9eb1a39d5e75cb560c823afaf9d9bd2).

Regarding the [subscriptions-reference-app template](https://github.com/Shopify/subscriptions-reference-app/blob/main/shopify.app.toml), it's not working in production. I'll ping them, but the template looks good in terms of schema, so this PR shouldn't change anything.

packages/app/src/cli/models/app/app.ts
```diff
@@ -538,9 +445,8 @@ export class App<
     return Boolean(frontendConfig ?? backendConfig)
   }
 
-  get appIsEmbedded() {
-    if (isCurrentAppSchema(this.configuration)) return this.configuration.embedded
-    return this.appIsLaunchable()
+  get appIsEmbedded(): boolean {
+    return this.configuration.embedded
```
isaacroldan, line 449: i'm not sure this is correct :thinking_face:  
  
We use appIsEmbedded to actually set the default value of `configuration.embedded`  
(see the function just below)  
  
I think the `creationDefaultOptions` should use the `appIsLaunchable` to set the default value of `embedded`

gonzaloriestra: It's ok because the embedded value is set [here](https://github.com/Shopify/cli/blob/54d4d3bae15fffcc5cec160b90a61c8ee0a04323/packages/app/src/cli/models/app/loader.ts#L230), which happens first.

If you create an app, you will see that `embedded = true` is added, even if the template doesn't have it.

gonzaloriestra:
The unrelated problem with `subscriptions-reference-app` has been fixed in https://github.com/Shopify/subscriptions-reference-app/pull/41, and I've checked that it works with this branch: `shopify app init --template=https://github.com/Shopify/subscriptions-reference-app#fix-npm-i`

packages/app/src/cli/models/app/app.ts
isaacroldan: The problem is actually with extensions-only templates, in those cases `isLaunchable = false` and therefore `embedded` should be false, but is defaulting to true (just tested with the none template)

gonzaloriestra: It happens the same with the latest release. Apparently the field is actually ignored: https://github.com/Shopify/cli/pull/5821 :man-facepalming:  
  
We need to review that and clean up.

isaacroldan: ups, not sure if that was really a temporary fix 😅 ignore my comments then

isaacroldan:
Is the only thing pending just merging the template changes? we should push to do that 🙏

packages/app/src/cli/services/app/config/link.ts
```diff
@@ -301,12 +272,10 @@ export async function loadLocalAppOptions(
...
 
     case 'loaded-template':
-      // Template with extra config keys (metafields, etc.) - treat as legacy for linking
       return {
-        state: 'legacy',
-        configFormat: 'legacy',
+        state: 'reusable-current-app',
         scopes: result.scopes,
-        localAppIdMatchedRemote: false,
+        localAppIdMatchedRemote: true,
```
ryancbahan, line 278: what does it take to get rid of this field? shouldn't this always match now?

packages/app/src/cli/services/context/breakdown-extensions.test.ts
```diff
@@ -273,7 +273,7 @@ const APP_CONFIGURATION: CurrentAppConfiguration = {
 
 const LOCAL_APP = async (
   uiExtensions: ExtensionInstance[],
-  configuration: AppConfiguration = APP_CONFIGURATION,
+  configuration: CurrentAppConfiguration = APP_CONFIGURATION,
```
ryancbahan, line 276: is `Current` still a necessary prefix?

packages/app/src/cli/services/context/identifiers-extensions.test.ts
```diff
@@ -935,7 +941,8 @@ describe('deployConfirmed: handle non existent uuid managed extensions', () => {
 
     // When
     const CONFIG_A = await testAppConfigExtensions()
-    const ensureExtensionsIdsOptions = options([], [], {configExtensions: [CONFIG_A], developerPlatformClient})
+    // Don't pass config extensions when includeConfigOnDeploy is false - they won't be in allExtensions
+    const ensureExtensionsIdsOptions = options([], [], {developerPlatformClient})
```
ryancbahan, line 945: nit: change `options` to an object arg so keys can provide context into what the args are for. `([], [])` is indirect

packages/app/src/cli/services/context/breakdown-extensions.test.ts
gonzaloriestra: I wanted to remove it, but `AppConfiguration` is already another type. I guess we could refactor somehow to simplify more, but not now, please :sweat_smile:

packages/app/src/cli/services/context/identifiers-extensions.test.ts
gonzaloriestra: Totally agree. But it's unrelated to this PR, which is big enough...

packages/app/src/cli/services/app/config/link.ts
gonzaloriestra: It's actually false when the loading fails (see line 289). Maybe there's a simpler way to handle that, we can investigate it later.

PR #6625, opened 2025-11-14
Introduce `AGENTS.md` file support in `shopify theme init`

karreiro:

This pull request updates the `shopify theme init` to consistently use `AGENTS.md` files. For consistency and compatibility, we are keeping symlinks with each editor standard.

image

Additionally, this pull request introduces a new "All" option to create symlinks for all editors:

image

This pull request:
- Updates the `init` theme service to handle the feature mentioned above
- Updates `cli-kit` to expose the symlink creation API

### How to test your changes?

- Run `shopify theme init`

### Post-release steps

None

### Measuring impact

- [ ] n/a - this doesn't need measurement, e.g., a linting rule or a bug-fix

### Checklist

- [x] I've considered possible cross-platform impacts (Mac, Linux, Windows).
- [x] I've considered possible [documentation](https://shopify.dev) changes.

### DIFF
--- a/packages/theme/src/cli/services/init.ts
+++ b/packages/theme/src/cli/services/init.ts
@@ -87,7 +87,7 @@ export async function createAIInstructions(themeRoot: string, aiInstruction: AII
 
           const instructions = (
             aiInstruction === 'all'
-              ? (Object.keys(SUPPORTED_AI_INSTRUCTIONS).filter((key) => key !== 'all') as AIInstruction[])
+              ? Object.keys(SUPPORTED_AI_INSTRUCTIONS).filter((key) => key !== 'all')
               : [aiInstruction]
           ) as Exclude<AIInstruction, 'all'>[]
 
@@ -130,7 +130,7 @@ export async function createAIInstructions(themeRoot: string, aiInstruction: AII
 export async function createAIInstructionFiles(
   themeRoot: string,
   agentsPath: string,
-  instruction: AIInstruction,
+  instruction: Exclude<AIInstruction, 'all'>,
 ): Promise<{copiedFile?: string}> {
   if (instruction === 'cursor') {
     // Cursor natively supports AGENTS.md, so no symlink needed
@@ -140,9 +140,9 @@ export async function createAIInstructionFiles(
   const symlinkMap = {
     github: 'copilot-instructions.md',
     claude: 'CLAUDE.md',
-  } as const
+  }
 
-  const symlinkName = symlinkMap[instruction as Exclude<AIInstruction, 'all' | 'cursor'>]
+  const symlinkName = symlinkMap[instruction]
   const symlinkPath = joinPath(themeRoot, symlinkName)
 
   try {
--- a/packages/theme/src/cli/services/init.ts
+++ b/packages/theme/src/cli/services/init.ts
@@ -85,10 +85,11 @@ export async function createAIInstructions(themeRoot: string, aiInstruction: AII
             shallow: true,
           })
 
-          const instructions =
+          const instructions = (
             aiInstruction === 'all'
               ? (Object.keys(SUPPORTED_AI_INSTRUCTIONS).filter((key) => key !== 'all') as AIInstruction[])
               : [aiInstruction]
+          ) as Exclude<AIInstruction, 'all'>[]
 
           try {
             const sourcePath = joinPath(tempDir, 'ai', 'github', 'copilot-instructions.md')
--- a/packages/theme/src/cli/services/init.ts
+++ b/packages/theme/src/cli/services/init.ts
@@ -136,10 +136,10 @@ export async function createAIInstructionFiles(
     return {}
   }
 
-  const symlinkMap: {[key in Exclude<AIInstruction, 'all' | 'cursor'>]: string} = {
+  const symlinkMap = {
     github: 'copilot-instructions.md',
     claude: 'CLAUDE.md',
-  }
+  } as const
 
   const symlinkName = symlinkMap[instruction as Exclude<AIInstruction, 'all' | 'cursor'>]
   const symlinkPath = joinPath(themeRoot, symlinkName)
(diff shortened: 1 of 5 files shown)


packages/theme/src/cli/services/init.ts
```diff
@@ -99,3 +100,23 @@ export async function createAIInstructions(themeRoot: string, aiInstruction: AII
...
+  const sourcePath = joinPath(tempDir, 'ai', 'github', 'copilot-instructions.md')
+  const sourceContent = await readFile(sourcePath)
+
+  const agentsPath = joinPath(themeRoot, 'AGENTS.md')
+  const agentsContent = `# AGENTS.md\n\n${sourceContent}`
+  await writeFile(agentsPath, agentsContent)
+
+  const symlinkMap: {[key in Exclude<AIInstruction, 'all'>]: string} = {
+    github: 'copilot-instructions.md',
+    cursor: '.cursorrules',
```
aswamy: I think Cursor natively supports `AGENT.md` lookup so we don't have to make a symlink for it

https://cursor.com/docs/context/rules

packages/theme/src/cli/services/init.ts
```diff
@@ -81,16 +82,16 @@ export async function createAIInstructions(themeRoot: string, aiInstruction: AII
...
           })
-          const aiSrcDir = joinPath(tempDir, 'ai', aiInstruction)
 
-          let aiDestDir = themeRoot
-
-          if (aiInstruction !== 'claude') {
-            aiDestDir = joinPath(themeRoot, `.${aiInstruction}`)
-          }
+          const instructions =
+            aiInstruction === 'all'
```
aswamy, line 89: I can't imagine wanting to support for "all" ever if i ever developed themes - but i dont see the harm 🤷

aswamy (approved):
Two optional changes, but looks good. I tophatted it on MacOS. I snapped it to test symlinks on Windows. Will report async when it's done.

aswamy (changes requested):
🚨 Tophat issue: The snap fails on Windows when creating the symlinks, but works on MacOS

image

packages/theme/src/cli/services/init.ts
karreiro: I was considering a team where members have various editor preferences.

packages/theme/src/cli/services/init.ts
karreiro: Good catch!

karreiro:
Thanks a lot for the review, @aswamy!

Windows require some specific permissions to create symbolic links, that's the reason we faced some errors. Now, we've update this PR to fallback to file creation (with a warning) when symbolic links can't be created:

Screenshot 2025-11-21 at 13 02 55

This PR is ready for re-review. Thanks again for the review!

packages/theme/src/cli/services/init.ts
```diff
@@ -81,21 +84,78 @@ export async function createAIInstructions(themeRoot: string, aiInstruction: AII
...
+): Promise<{copiedFile?: string}> {
+  if (instruction === 'cursor') {
+    // Cursor natively supports AGENTS.md, so no symlink needed
+    return {}
+  }
+
+  const symlinkMap: {[key in Exclude<AIInstruction, 'all' | 'cursor'>]: string} = {
+    github: 'copilot-instructions.md',
+    claude: 'CLAUDE.md',
+  }
```
aswamy: You don't need to define the explicit type of key here; also could you move it outside the function since it's a constant.
```suggestion
  const symlinkMap = {
    github: 'copilot-instructions.md',
    claude: 'CLAUDE.md',
  } as const
```

packages/theme/src/cli/services/init.ts
```diff
@@ -81,21 +84,78 @@ export async function createAIInstructions(themeRoot: string, aiInstruction: AII
...
-
-          let aiDestDir = themeRoot
 
-          if (aiInstruction !== 'claude') {
-            aiDestDir = joinPath(themeRoot, `.${aiInstruction}`)
-          }
+          const instructions =
+            aiInstruction === 'all'
+              ? (Object.keys(SUPPORTED_AI_INSTRUCTIONS).filter((key) => key !== 'all') as AIInstruction[])
+              : [aiInstruction]
```
aswamy:
```suggestion
          const instructions = (
            aiInstruction === 'all'
              ? (Object.keys(SUPPORTED_AI_INSTRUCTIONS).filter((key) => key !== 'all') as AIInstruction[])
              : [aiInstruction]
          ) as Exclude<AIInstruction, 'all'>[]
```
I saw a lot of casting happening in code further down the chain, and it could be resolved if we just casted it here instead.

packages/theme/src/cli/services/init.ts
```diff
@@ -81,21 +84,78 @@ export async function createAIInstructions(themeRoot: string, aiInstruction: AII
...
+        ', ',
+      )}, but your system doesn't have Developer Mode enabled or symlinks are disabled. Separate files were created instead.`,
+    })
+  }
+}
+
+export async function createAIInstructionFiles(
+  themeRoot: string,
+  agentsPath: string,
+  instruction: AIInstruction,
```
aswamy: If we changed this to `instruction: Exclude<AIInstruction, 'all'>,` we don't have to do ANY casting inside this function (and it's more accurate on what this method accepts)

aswamy (approved):
Code looks great! Tiny type changes requested. Did not tophat on Windows - seems like my VM is having issues.

packages/theme/src/cli/services/init.ts
karreiro: Great catch! Thank you, @aswamy!

karreiro:
👋 @Shopify/app-inner-loop could you please take a look when you have a moment?

gonzaloriestra (approved):
Cool!

Issue #6415, opened 2025-09-18
[Bug]: Cannot provide a custom app access token to --password flag


kennyr87:
### Please confirm that you have:

- [x] Searched [existing issues](.) to see if your issue is a duplicate. (If you’ve found a duplicate issue, feel free to add additional information in a comment on it.)
- [x] Reproduced the issue in [the latest CLI version](https://www.npmjs.com/package/@shopify/cli).

### In which of these areas are you experiencing a problem?

Theme

### Expected behavior

Provide a [custom app access token](https://shopify.dev/docs/apps/build/authentication-authorization/access-tokens/generate-app-access-tokens-admin) to authenticate with the store that you want to work on.

To authenticate using an access token, pass the ` --password` flag with each command that you want to run against the store. 

### Actual behavior

Command `shopify theme dev --password shpat_***` exits with error **Invalid password. Please generate a new password from the Theme Access app.**

Interestingly, CLI authenticates to store without errors when running other CLI commands!?

### Verbose output

<details>
  <summary>Verbose output</summary>

  ```
2025-09-18T20:31:53.271Z: Analytics event sent: {
  "command": "theme dev",
  "time_start": 1758227512286,
  "time_end": 1758227512446,
  "total_time": 160,
  "success": false,
  "cli_version": "3.84.1",
  "ruby_version": "",
  "node_version": "24.1.0",
  "is_employee": false,
  "uname": "linux amd64",
  "env_ci": false,
  "env_plugin_installed_any_custom": false,
  "env_plugin_installed_shopify": "[\"@shopify/cli\"]",
  "env_shell": "bash",
  "env_device_id": "a4998d4a8b9461fdde65077bbe4967d5c64ba448",
  "env_cloud": "localhost",
  "env_package_manager": "npm",
  "env_is_global": true,
  "env_auth_method": "none",
  "env_is_wsl": false,
  "env_build_repository": "Shopify/cli",
  "cmd_app_warning_api_key_deprecation_displayed": false,
  "cmd_all_timing_network_ms": 0,
  "cmd_all_timing_prompts_ms": 0,
  "cmd_all_launcher": "unknown",
  "cmd_all_topic": "theme",
  "cmd_all_plugin": "@shopify/theme",
  "cmd_all_force": false,
  "cmd_all_verbose": true,
  "cmd_all_path_override": true,
  "cmd_all_path_override_hash": "43473b8a8b37ed3c2f77ef0ba9648c5dc2692369",
[...]

PR #6626, opened 2025-11-14
Restore usage of custom app passwords to theme commands

graygilmore:

- Closes https://github.com/Shopify/cli/issues/6415

This was previously stripped away but theme developers are still using this flow and are forced to pin to older versions. I'm returning the functionality in this PR but calling out the specific places where it doesn't work:

- `dev` does not support hot module reloading
- `console` and `profile` commands do not work at all

I tested on all of the theme commands:

```
check - ✅
console - ❌
delete - ✅
dev - 🟡 
duplicate - ✅
info - ✅
init - ✅
language-server - ✅
list - ✅
open - ✅
package - ✅
profile - ❌
publish - ✅
pull - ✅
push - ✅
rename - ✅
share - ✅
```

For `console` and `profile` we now throw an `AbortError`. For `dev` we render a warning.

### How to test your changes?

1. Install the snapped version `npm i -g --@shopify:registry=https://registry.npmjs.org @shopify/cli@0.0.0-snapshot-20251114232105`
2. Run through as many of the commands as you'd like to test


packages/theme/src/cli/services/dev.ts
```diff
@@ -43,6 +43,13 @@ export async function dev(options: DevOptions) {
     return
   }
 
+  if (options.password?.startsWith('shpat_')) {
+    renderWarning({
+      headline: 'Admin API token missing features:',
+      body: 'Directly using an Admin API token will result in some missing features like hot module reloading. We recommend generating a password from the Theme Access app or authenticating directly.',
```
EvilGenius13: What do you think about matching up the wording for console and dev a bit? I wouldn't mind seeing something closer to console/profile with the `Alternatively, you can authenticate normally by not passing the --password flag.`

graygilmore: [Pushed up a change](https://github.com/Shopify/cli/compare/f729a755767877f0de2f5306b344841a5562a4d5..6fc48ee44d48ce6d367f12b98eaac37e6027613b)!
image

graygilmore:
## Force Push Patch Notes

- [Updated](https://github.com/Shopify/cli/compare/6fc48ee44d48ce6d367f12b98eaac37e6027613b..56c08f7c089b0c7e36537c0575a5096bd31379da) to note that password protected storefronts also don't work 

Thanks to both @dejmedus and @EvilGenius13 for the tophatting 😄

dejmedus (approved):
The commands I tried all worked as expected and the warnings look good!

<details>
<summary>🎩</summary>

Image
above with store password set, below without
Image

</details>

PR #6631, opened 2025-11-17
Refactor renderSingleTask to allow an async process to update current status

nickwesselman:

Long-running operations in the CLI don't currently have a mechanism to update their status as they run.

Allow the task provided to `renderSingleTask` to update the displayed status text. This change means that `renderSingleTask` must now always be provided a function instead of a `Promise`. I could actually find no existing use of this UI function, but this change preserves its ability to return the task function's output.

For example:

```javascript
const task = async (updateStatus: (status: string) => void) => {
  for (let i = 1; i <= 10; i++) {
    await new Promise((resolve) => setTimeout(resolve, 500))
    updateStatus(`Counting (${i} complete)`)
  }
  return 'completed'
}

const result = await renderSingleTask({
  title: `Counting`,
  task,
})

renderSuccess({
  body: `Counting: ${result}`,
})
```

https://github.com/user-attachments/assets/4ec91614-093f-480a-88a4-01fee07ae78c

https://github.com/user-attachments/assets/21901d0a-2183-4e00-b986-737f5a457775

### How to test your changes?

Create a CLI command that does something like the above.

### DIFF
--- a/packages/cli-kit/bin/documentation/examples.ts
+++ b/packages/cli-kit/bin/documentation/examples.ts
@@ -16,7 +16,7 @@ import {
   renderTextPrompt,
   renderWarning,
 } from '../../src/public/node/ui.js'
-import {unstyled} from '../../src/public/node/output.js'
+import {outputContent, unstyled} from '../../src/public/node/output.js'
 import {AbortError, BugError} from '../../src/public/node/error.js'
 import {AbortSignal} from '../../src/public/node/abort.js'
 import {Stdout} from '../../src/private/node/ui.js'
@@ -592,7 +592,7 @@ export const examples: {[key in string]: Example} = {
       const stdout = new Stdout({columns: TERMINAL_WIDTH})
 
       await renderSingleTask({
-        title: 'Loading app',
+        title: outputContent`Loading app`,
         task: async () => {
           await sleep(1)
         },


packages/cli-kit/src/private/node/ui/components/SingleTask.tsx
```diff
@@ -3,35 +3,38 @@ import {useExitOnCtrlC} from '../hooks/use-exit-on-ctrl-c.js'
 import React, {useEffect, useState} from 'react'
 import {useApp} from 'ink'
 
-interface SingleTaskProps {
+interface SingleTaskProps<T> {
   title: string
-  taskPromise: Promise<unknown>
+  task: (updateStatus: (status: string) => void) => Promise<T>
```
shauns: I think the status should probably be something renderable, so `TokenizedString`

nickwesselman: 🤔 Do you think we should include the `...` suffix by default, or leave that to the consumer? Seems more flexible w/out, especially if we are allowing `TokenizedString`

shauns: I think you're right

nickwesselman: Turns out that's in `LoadingBar` which is also used by `renderTasks` as well so I left it, can always adapt later if needed

nickwesselman:
Changed to use `TokenizedString` and added an example to `kitchen-sink` ... @isaacroldan or @gonzaloriestra would you mind taking one more look?

PR #6634, opened 2025-11-18
create identity client interface

alexanderMontague:

Part of: https://github.com/shop/issues-develop/issues/21594

- We're trying to improve our local development experience for using the CLI
- A good first step is being able to hit our APIs without the identity service running

- scaffolds out the interface for the identity client that we will use 
- defines the client swapping capabilities based on the environment and what services are running

### How to test your changes?

- included tests
- looking for feedback on a better singleton pattern if possible

### DIFF
--- a/packages/cli-kit/src/private/node/clients/identity/identity-client.ts
+++ b/packages/cli-kit/src/private/node/clients/identity/identity-client.ts
@@ -1,0 +1,23 @@
+import {ApplicationToken, IdentityToken} from '../../session/schema.js'
+import {ExchangeScopes} from '../../session/exchange.js'
+import {API} from '../../api.js'
+
+export abstract class IdentityClient {
+  abstract requestAccessToken(scopes: string[]): Promise<IdentityToken>
+
+  abstract exchangeAccessForApplicationTokens(
+    identityToken: IdentityToken,
+    scopes: ExchangeScopes,
+    store?: string,
+  ): Promise<{[x: string]: ApplicationToken}>
+
+  abstract refreshAccessToken(currentToken: IdentityToken): Promise<IdentityToken>
+
+  clientId(): string {
+    return ''
+  }
+
+  applicationId(_api: API): string {
+    return ''
+  }
+}
--- a/packages/cli-kit/src/private/node/clients/identity/instance.ts
+++ b/packages/cli-kit/src/private/node/clients/identity/instance.ts
@@ -1,0 +1,18 @@
+import {IdentityClient} from './identity-client.js'
+import {IdentityMockClient} from './identity-mock-client.js'
+import {IdentityServiceClient} from './identity-service-client.js'
+import {Environment, serviceEnvironment} from '../../context/service.js'
+import {isRunning2024} from '../../../../public/node/vendor/dev_server/dev-server-2024.js'
+
+let _identityClient: IdentityClient | undefined
+
+export function getIdentityClient() {
+  if (!_identityClient) {
+    const isLocal = serviceEnvironment() === Environment.Local
+    const identityServiceRunning = isRunning2024('identity')
+    const client = isLocal && !identityServiceRunning ? new IdentityMockClient() : new IdentityServiceClient()
+    _identityClient = client
+  }
+
+  return _identityClient
+}
--- a/packages/cli-kit/src/private/node/clients/identity/identity-mock-client.ts
+++ b/packages/cli-kit/src/private/node/clients/identity/identity-mock-client.ts
@@ -1,0 +1,21 @@
+import {IdentityClient} from './identity-client.js'
+import {ApplicationToken, IdentityToken} from '../../session/schema.js'
+import {ExchangeScopes} from '../../session/exchange.js'
+
+export class IdentityMockClient extends IdentityClient {
+  async requestAccessToken(_scopes: string[]): Promise<IdentityToken> {
+    return {} as IdentityToken
+  }
+
+  async exchangeAccessForApplicationTokens(
+    _identityToken: IdentityToken,
+    _scopes: ExchangeScopes,
+    _store?: string,
+  ): Promise<{[x: string]: ApplicationToken}> {
+    return {}
+  }
+
+  async refreshAccessToken(_currentToken: IdentityToken): Promise<IdentityToken> {
+    return {} as IdentityToken
+  }
+}
--- a/packages/cli-kit/src/private/node/clients/identity/identity-service-client.ts
+++ b/packages/cli-kit/src/private/node/clients/identity/identity-service-client.ts
@@ -1,0 +1,21 @@
+import {IdentityClient} from './identity-client.js'
+import {ApplicationToken, IdentityToken} from '../../session/schema.js'
+import {ExchangeScopes} from '../../session/exchange.js'
+
+export class IdentityServiceClient extends IdentityClient {
+  async requestAccessToken(_scopes: string[]): Promise<IdentityToken> {
+    return {} as IdentityToken
+  }
+
+  async exchangeAccessForApplicationTokens(
+    _identityToken: IdentityToken,
+    _scopes: ExchangeScopes,
+    _store?: string,
+  ): Promise<{[x: string]: ApplicationToken}> {
+    return {}
+  }
+
+  async refreshAccessToken(_currentToken: IdentityToken): Promise<IdentityToken> {
+    return {} as IdentityToken
+  }
+}
--- a/packages/cli-kit/src/private/node/clients/identity/instance.test.ts
+++ b/packages/cli-kit/src/private/node/clients/identity/instance.test.ts
@@ -1,0 +1,77 @@
+import {Environment} from '../../context/service.js'
+import {describe, expect, test, vi, beforeEach} from 'vitest'
+
+const mockServiceEnvironment = vi.fn()
+const mockIsRunning2024 = vi.fn()
+
+vi.mock('../../context/service.js', async () => {
+  const actual = await vi.importActual('../../context/service.js')
+  return {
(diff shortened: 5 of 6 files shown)


alexanderMontague:
* **#6654** <a href="https://app.graphite.com/github/pr/Shopify/cli/6654?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6635** <a href="https://app.graphite.com/github/pr/Shopify/cli/6635?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6634** <a href="https://app.graphite.com/github/pr/Shopify/cli/6634?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Shopify/cli/6634?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/cli-kit/src/private/node/clients/identity/instance.ts
```diff
@@ -0,0 +1,18 @@
+import {IdentityBaseClient} from './identity-base-client.js'
+import {IdentityMockClient} from './identity-mock-client.js'
+import {IdentityServiceClient} from './identity-service-client.js'
+import {Environment, serviceEnvironment} from '../../context/service.js'
+import {isRunning2024} from '../../../../public/node/vendor/dev_server/dev-server-2024.js'
+
+let _identityClient: IdentityBaseClient | undefined
+
+export function getIdentityClient() {
```
alexanderMontague, line 9: I really wanted this to live in `IdentityBaseClient` under a `getInstance` method like we do for our other clients, but combining that with an abstract class leads to circular dependencies. I see we use this singleton pattern in other modules, so looking for feedback on this especially.

packages/cli-kit/src/private/node/clients/identity/identity-client.ts
```diff
@@ -0,0 +1,23 @@
...
+
+  abstract exchangeAccessForApplicationTokens(
+    identityToken: IdentityToken,
+    scopes: ExchangeScopes,
+    store?: string,
+  ): Promise<{[x: string]: ApplicationToken}>
+
+  abstract refreshAccessToken(currentToken: IdentityToken): Promise<IdentityToken>
+
+  clientId(): string {
```
MitchDickinson, line 16: What are these empty string result methods for?

packages/cli-kit/src/private/node/clients/identity/identity-base-client.ts
```diff
@@ -0,0 +1,23 @@
+import {ApplicationToken, IdentityToken} from '../../session/schema.js'
+import {ExchangeScopes} from '../../session/exchange.js'
+import {API} from '../../api.js'
+
+export abstract class IdentityBaseClient {
```
MitchDickinson: IMO `IdentityClient` > `IdentityBaseClient`. The `Base` doesn't really add much. `IdentityClient` is the right abstraction

packages/cli-kit/src/private/node/clients/identity/identity-base-client.ts
```diff
@@ -0,0 +1,23 @@
...
+  abstract requestAndPollForAccessToken(_scopes: string[]): Promise<IdentityToken>
```
MitchDickinson: Do we need `_` here?

packages/cli-kit/src/private/node/clients/identity/identity-client.ts
MitchDickinson, line 1: This interface looks a lot more clear than the POC one. The types and method names are super clear and though it's not wired up, it looks like you solved the temporal coupling smell too. Nice work!

packages/cli-kit/src/private/node/clients/identity/identity-base-client.ts
MitchDickinson: Is the polling part relevant to the caller of this, or is it an implementation detail?

packages/cli-kit/src/private/node/clients/identity/identity-client.ts
```diff
@@ -0,0 +1,23 @@
...
+import {API} from '../../api.js'
+
+export abstract class IdentityBaseClient {
+  abstract requestAndPollForAccessToken(_scopes: string[]): Promise<IdentityToken>
+
+  abstract exchangeAccessForApplicationTokens(
+    identityToken: IdentityToken,
+    scopes: ExchangeScopes,
+    store?: string,
+  ): Promise<{[x: string]: ApplicationToken}>
```
MitchDickinson, line 12: I think this return type is the only slightly confusing type when I'm looking at this interface. I assume the string keys here are the same keys from the `ExchangeScopes`? It makes me wonder if we can express that relationship better in the type system. Maybe instead of using strings we could create a named type that we could use in both input and output values that makes it clear what these strings represent. 

Even if it's just a simple type alias, it's clearer to understand "this string represents an X thing" instead of having the reader have to derive the semantic of the string by reading the implementation

packages/cli-kit/src/private/node/clients/identity/identity-client.ts
MitchDickinson: Ah, I see in the follow up PR 👍

packages/cli-kit/src/private/node/clients/identity/instance.ts
isaacroldan: I think is OK to have this here 👌

packages/cli-kit/src/private/node/clients/identity/identity-base-client.ts
alexanderMontague: Agree 👍 was back and forth trying to name classes while shuffling files

packages/cli-kit/src/private/node/clients/identity/identity-base-client.ts
alexanderMontague: Probably an implementation detail 👍 `requestAccessToken` is good enough

alexanderMontague: nope, good catch

packages/cli-kit/src/private/node/clients/identity/identity-client.ts
alexanderMontague: Copy pasta'd, in the POC I did extract some generic types to explicit types. Thats probably a good thing to do here as well. Can implement this.

packages/cli-kit/src/private/node/clients/identity/identity-client.ts
alexanderMontague: Should have commented on this, just placeholders to make linter happy for now.

packages/cli-kit/src/private/node/clients/identity/instance.ts
```diff
@@ -0,0 +1,18 @@
...
+  if (!_identityClient) {
+    const isLocal = serviceEnvironment() === Environment.Local
+    const identityServiceRunning = isRunning2024('identity')
+    const client = isLocal && !identityServiceRunning ? new IdentityMockClient() : new IdentityServiceClient()
```
ryancbahan, line 13: why not just return client and get rid of the mutable variable?

packages/cli-kit/src/private/node/clients/identity/identity-client.ts
```diff
@@ -0,0 +1,23 @@
+import {ApplicationToken, IdentityToken} from '../../session/schema.js'
```
ryancbahan, line 1: What's the value of an abstract class over just using a TS interface to describe the type shapes? IMO you could avoid the circular deps and reduce LOC a bit just using types.

ryancbahan: ah, i see in followup prs we are using this class concretely

packages/cli-kit/src/private/node/clients/identity/instance.ts
alexanderMontague: Just lends itself to the singleton pattern better. Saves re-instantiating on every call, but the clients are deterministic so we could do this too. Follows existing patterns, I don't feel too strongly about this. LMK what you think

packages/cli-kit/src/private/node/clients/identity/identity-client.ts
alexanderMontague: Looked into it, the keys in `ExchangeScopes` get mapped to their application ID, so they won't unfortunately be the same. We could alias the current type, but that doesn't give us a huge benefit. Can come back to this if we feel it should be improved.

packages/cli-kit/src/private/node/clients/identity/instance.ts
ryancbahan: My opinion is that the tradeoff isn't worth it -- it introduces sequencing where the variable can be undefined, which has to be accounted for in the type as well as in all processes it's called from. The perf gains on compute will be near-zero, and may never actually become concrete unless you're calling identity multiple times in the same command lifecycle but different execution context.

alexanderMontague: How would you ever access the undefined variable?

isaacroldan: The mutable variable is private, only the `getIdentityClient` is public and will return a valid client always, regardless of using a singleton or instantiating a new client every time. 

I think is ok, to have a singleton here. Is safe and I think there is any con about it.

If eventually we decide that this client will have logic to prevent multiple refresh tokens operations from running at the same time, then we need to ensure everyone is using the same instance. (is an example, this is controlled somewhere else now)

PR #6635, opened 2025-11-18
implement identity client body

alexanderMontague:

Part of: shop/issues-develop#21594

- implementing the body of the client interface with the respective production and mocked code
- we've optimized some of the production methods
- mostly just hot swapping code, more refactoring will come in an upstack to the actual execution flow if needed

### DIFF
--- a/packages/cli-kit/src/private/node/clients/identity/identity-client.ts
+++ b/packages/cli-kit/src/private/node/clients/identity/identity-client.ts
@@ -1,23 +1,73 @@
-import {ApplicationToken, IdentityToken} from '../../session/schema.js'
-import {ExchangeScopes} from '../../session/exchange.js'
+import {IdentityToken} from '../../session/schema.js'
+import {TokenRequestResult} from '../../session/exchange.js'
 import {API} from '../../api.js'
+import {Environment, serviceEnvironment} from '../../context/service.js'
+import {BugError} from '../../../../public/node/error.js'
+import {Result} from '../../../../public/node/result.js'
 
 export abstract class IdentityClient {
   abstract requestAccessToken(scopes: string[]): Promise<IdentityToken>
 
-  abstract exchangeAccessForApplicationTokens(
-    identityToken: IdentityToken,
-    scopes: ExchangeScopes,
-    store?: string,
-  ): Promise<{[x: string]: ApplicationToken}>
+  abstract tokenRequest(params: {
+    [key: string]: string
+  }): Promise<Result<TokenRequestResult, {error: string; store?: string}>>
 
   abstract refreshAccessToken(currentToken: IdentityToken): Promise<IdentityToken>
 
-  clientId(): string {
-    return ''
-  }
+  abstract clientId(): string
 
-  applicationId(_api: API): string {
-    return ''
+  applicationId(api: API): string {
+    switch (api) {
+      case 'admin': {
+        const environment = serviceEnvironment()
+        if (environment === Environment.Local) {
+          return 'e92482cebb9bfb9fb5a0199cc770fde3de6c8d16b798ee73e36c9d815e070e52'
+        } else if (environment === Environment.Production) {
+          return '7ee65a63608843c577db8b23c4d7316ea0a01bd2f7594f8a9c06ea668c1b775c'
+        } else {
+          return 'e92482cebb9bfb9fb5a0199cc770fde3de6c8d16b798ee73e36c9d815e070e52'
+        }
+      }
+      case 'partners': {
+        const environment = serviceEnvironment()
+        if (environment === Environment.Local) {
+          return 'df89d73339ac3c6c5f0a98d9ca93260763e384d51d6038da129889c308973978'
+        } else if (environment === Environment.Production) {
+          return '271e16d403dfa18082ffb3d197bd2b5f4479c3fc32736d69296829cbb28d41a6'
+        } else {
+          return 'df89d73339ac3c6c5f0a98d9ca93260763e384d51d6038da129889c308973978'
+        }
+      }
+      case 'storefront-renderer': {
+        const environment = serviceEnvironment()
+        if (environment === Environment.Local) {
+          return '46f603de-894f-488d-9471-5b721280ff49'
+        } else if (environment === Environment.Production) {
+          return 'ee139b3d-5861-4d45-b387-1bc3ada7811c'
+        } else {
+          return '46f603de-894f-488d-9471-5b721280ff49'
+        }
+      }
+      case 'business-platform': {
+        const environment = serviceEnvironment()
+        if (environment === Environment.Local) {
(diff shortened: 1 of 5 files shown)


alexanderMontague:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/Shopify/cli/6635?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#6654** <a href="https://app.graphite.com/github/pr/Shopify/cli/6654?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6635** <a href="https://app.graphite.com/github/pr/Shopify/cli/6635?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Shopify/cli/6635?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#6634** <a href="https://app.graphite.com/github/pr/Shopify/cli/6634?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

packages/cli-kit/src/private/node/clients/identity/identity-base-client.ts
```diff
@@ -1,23 +1,82 @@
...
-  ): Promise<{[x: string]: ApplicationToken}>
+  abstract tokenRequest(params: {
+    [key: string]: string
+  }): Promise<Result<TokenRequestResult, {error: string; store?: string}>>
 
   abstract refreshAccessToken(currentToken: IdentityToken): Promise<IdentityToken>
 
   clientId(): string {
-    return ''
+    const environment = serviceEnvironment()
```
MitchDickinson: I think these methods are ripe for `replace conditional with polymorphism`:
```ts
serviceEnvironment().clientId()
```

packages/cli-kit/src/private/node/clients/identity/identity-base-client.ts
```diff
@@ -1,23 +1,82 @@
...
-    scopes: ExchangeScopes,
-    store?: string,
-  ): Promise<{[x: string]: ApplicationToken}>
+  abstract tokenRequest(params: {
+    [key: string]: string
+  }): Promise<Result<TokenRequestResult, {error: string; store?: string}>>
 
   abstract refreshAccessToken(currentToken: IdentityToken): Promise<IdentityToken>
 
   clientId(): string {
```
MitchDickinson: A smell to look out for. If nobody actually calls clientId against the abstract type (in this case your base client), then it probably doesn't belong in the abstract. You can probably just leave the two concrete implementations without having it live in the abstract - unless there is a use case coming up where it matters in the abstract

packages/cli-kit/src/private/node/clients/identity/identity-client.ts
```diff
@@ -1,23 +1,82 @@
...
+    } else if (environment === Environment.Production) {
+      return 'fbdb2649-e327-4907-8f67-908d24cfd7e3'
+    } else {
+      return 'e5380e02-312a-7408-5718-e07017e9cf52'
+    }
   }
 
-  applicationId(_api: API): string {
-    return ''
+  applicationId(api: API): string {
```
MitchDickinson, line 19: Similar here - is it only used by the mock client or is this also going to be used for the service client?

packages/cli-kit/src/private/node/clients/identity/identity-client.ts
```diff
@@ -1,23 +1,82 @@
...
+    switch (api) {
+      case 'admin': {
+        const environment = serviceEnvironment()
```
MitchDickinson, line 22: Would be good to make this polymorphic too, if possible

packages/cli-kit/src/private/node/clients/identity/identity-base-client.ts
alexanderMontague: yep good call. This can probably just be an abstract definition and the implementation can live in each client

packages/cli-kit/src/private/node/clients/identity/identity-base-client.ts
alexanderMontague: I'll think about this a bit more, but its unfortunately more nuanced since we actually _might_ want to use this client even in local development. Its now a case of:

- Prod --> `IdentityClient`
- Local + Identity service running --> `IdentityClient`
- Local + no identity --> `MockIdentityClient`

packages/cli-kit/src/private/node/clients/identity/identity-base-client.ts
MitchDickinson: If nobody calls the clientId() method against the `IdentityClient` base type though, you don't get any value from it being defined as abstract. Abstract methods are best when the root type wants to force a child to implement something because that something matters to the contract of the base type. If the clientId is simply an implementation detail of "other things" in the child type, then there is no value in having the parent enforce the existence of a clientId

Another test of this to show it's not necessary: if you delete `clientId()` from the base type, nothing breaks or complains.

packages/cli-kit/src/private/node/clients/identity/identity-base-client.ts
MitchDickinson: That could still be encapsulated behind the polymorphic interface though. If anything it's potentially even another argument in favor of it. I'm not sure if typescript enums allow definition of behaviour, but we could use a class instead (or whatever other language feature is best used to handle it)

```ts
// Conceptual pseudo code, may not actually work with enums but a conversion to a class or other type would certainly allow this kind of polymorphic definition
export enum Environment {
  Local = 'local' {
    function clientId() {
      identityRunning() ? "value" : "other_value"
    } 
    function applicationId(api) {
      switch(api) { ... }
    }
  },
  Production = 'production' {
    function clientId() {
      "production_value"
    }
    function applicationId(api) {
      switch(api) { ... }
    }
  },
}

// usage
serviceEnvironment().clientId()
serviceEnvironment().applicationId(env)
```

(this obviously isn't a huge deal at all, it's just a possible minor improvement to how we store this config data and allows you to avoid worrying about the `else` blocks and default blocks from within switch statements. I wouldn't burn much time on it tho)

packages/cli-kit/src/private/node/clients/identity/identity-service-client.ts
```diff
@@ -1,21 +1,220 @@
...
+    return identityToken
   }
 
-  async exchangeAccessForApplicationTokens(
-    _identityToken: IdentityToken,
-    _scopes: ExchangeScopes,
-    _store?: string,
-  ): Promise<{[x: string]: ApplicationToken}> {
-    return {}
+  async tokenRequest(params: {
```
alexanderMontague, line 36: Comment link for Fred

PR #6640, opened 2025-11-19
Allow non-interactive `app init`

nickwesselman:

The `app init` command has interactive prompts that are impossible to bypass. This prevents scripting and some agentic use.

(Agentic use may still be difficult as the agent needs to obtain the organization ID, but this puts us a step closer.)

- Add `organization-id` flag to suppress organization selection
- Ensure `name` flag suppresses interactive linking and name prompts
- Adds unit tests for `init` command
- Moves `SHOPIFY_FLAG_CLIENT_ID` environment definition in our GitHub actions to the acceptance test job specifically (where it's needed), since it was breaking the new `app init` tests

### How to test your changes?

The following commands should all be successful:

```
shopify app init --organization-id <org-id> --name test --template remix --flavor javascript | cat
```

```
shopify app init --organization-id <org-id> --name test --template none | cat
```

```
# no org id needed with a partners token
export SHOPIFY_CLI_PARTNERS_TOKEN=<your cli token>
shopify app init  --name test --template none | cat
```

The following should fail:

```
# with device login / no partner token
shopify app init --organization-id <NOT-your-org-id> --name test --template remix --flavor javascript | cat
```

### Measuring impact

### Checklist


nickwesselman:
Will need to figure out why unit tests are failing. Works on my machine ™️.

packages/app/src/cli/commands/app/init.ts
```diff
@@ -93,10 +100,31 @@ export default class Init extends AppLinkedCommand {
       developerPlatformClient = selectedApp.developerPlatformClient ?? developerPlatformClient
       selectAppResult = {result: 'existing', app: selectedApp}
     } else {
-      const org = await selectOrg()
+      let org: Organization
+      if (flags['organization-id']) {
+        // If an organization-id is provided, fetch all organizations and find the matching one
+        const orgs = await fetchOrganizations()
+        const matchingOrg = orgs.find((organization) => organization.id === flags['organization-id'])
```
gonzaloriestra: There's an `orgFromId` query you can use to simplify

packages/app/src/cli/commands/app/init.ts
```diff
@@ -93,10 +100,31 @@ export default class Init extends AppLinkedCommand {
...
+        if (!matchingOrg) {
+          throw new AbortError(
+            `Organization with ID ${flags['organization-id']} not found`,
+            'Please verify the organization ID and try again. You can find your organization ID in your Dev Dashboard URL: https://dev.shopify.com/dashboard/<organization-id>',
```
gonzaloriestra: What about suggesting to run `shopify auth login` to ensure they are using the right account?

packages/app/src/cli/commands/app/init.ts
nickwesselman: The default `developerPlatformClient` is still Partners so this fails, since we don't have an org yet. Maybe there's some way to work around that until we can fully remove Partners?

gonzaloriestra: This PR should fix that: https://github.com/Shopify/cli/pull/6645

nickwesselman:
* **#6640** <a href="https://app.graphite.com/github/pr/Shopify/cli/6640?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Shopify/cli/6640?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#6968** <a href="https://app.graphite.com/github/pr/Shopify/cli/6968?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

packages/app/src/cli/commands/app/init.ts
nickwesselman: Stacked on that PR and working great!

nickwesselman:
Converted PR to Draft to avoid merging before required server-side changes.

nickwesselman:
Still relevant.

nickwesselman:
Should we merge this now that we have app limits?

packages/app/src/cli/commands/app/init.ts
gonzaloriestra: The PR blocking Partners was finally reverted [here](https://github.com/Shopify/cli/pull/6708). Have you tested this again?

gonzaloriestra: Forget it, just checked that it's fine.

gonzaloriestra (approved):
Tested again, all good! 👍

packages/app/src/cli/commands/app/init.ts
```diff
@@ -93,10 +99,31 @@ export default class Init extends AppLinkedCommand {
       developerPlatformClient = selectedApp.developerPlatformClient ?? developerPlatformClient
       selectAppResult = {result: 'existing', app: selectedApp}
     } else {
-      const org = await selectOrg()
+      let org: Organization
+      if (flags['organization-id']) {
+        // If an organization-id is provided, fetch the organization directly
+        const matchingOrg = await developerPlatformClient.orgFromId(flags['organization-id'])
+        if (!matchingOrg) {
+          throw new AbortError(
```
dmerand: Nit: I did not get this helpful message when sending a non-integer as a parameter -- just got a bare GraphQL error. Perhaps we could use typing or otherwise be more defensive about input?

dmerand (approved):
This works as expected! I added a small note about error messages.

packages/app/src/cli/commands/app/init.test.ts
dmerand, line 1: 💬 **Suggestion**: The tests cover the happy path (valid org-id, empty apps list) and error path (invalid org-id), but don't cover the scenario where `--organization-id` is provided with an organization that has existing apps. This would verify that the `nameProvidedAsFlag` logic correctly bypasses the app selection prompt when both `--name` and `--organization-id` are provided but apps already exist in the org.

**Suggestion:** Add a test case that mocks `orgAndApps` to return a non-empty apps array (e.g., `apps: [testOrganizationApp()]`) and verifies that with `--organization-id`, `--name`, and `--template` flags, the command still succeeds without calling `createAsNewAppPrompt` or `selectAppPrompt`.

packages/app/src/cli/commands/app/init.test.ts
```diff
@@ -0,0 +1,156 @@
...
+        name: 'my-app',
+        packageManager: 'npm',
+        template: 'https://github.com/Shopify/shopify-app-template-remix',
+      }),
+    )
+  })
+
+  test('fails with clear error message when invalid organization-id is provided', async () => {
+    // Given
+    const validOrg = testOrganization()
```
dmerand: 🎨 **Code Style**: The variable `validOrg` is declared on line 117 but never used in this test. It appears to be a leftover from an earlier iteration when the test mocked `fetchOrganizations` to return this organization. Since the test now mocks `orgFromId` to return `undefined` directly, this variable serves no purpose.

**Suggestion:** Remove the unused variable:

```suggestion
    const mockDeveloperPlatformClient = testDeveloperPlatformClient()
```

packages/app/src/cli/commands/app/init.ts
```diff
@@ -64,6 +64,12 @@ export default class Init extends AppLinkedCommand {
       env: 'SHOPIFY_FLAG_CLIENT_ID',
       exclusive: ['config'],
     }),
+    'organization-id': Flags.string({
```
dmerand, line 70: 💬 **Suggestion**: The `client-id` flag already implies a specific app (and thus organization). While the current code handles both being provided gracefully (client-id takes precedence), marking them as mutually exclusive would make the UX clearer and prevent potential user confusion about which flag takes priority.

**Suggestion:** Add `exclusive: ['client-id']` to the organization-id flag definition:

```suggestion
    'organization-id': Flags.string({
      hidden: false,
      description:
        'The organization ID. Your organization ID can be found in your Dev Dashboard URL: https://dev.shopify.com/dashboard/<organization-id>',
      env: 'SHOPIFY_FLAG_ORGANIZATION_ID',
      exclusive: ['client-id'],
    }),
```

packages/app/src/cli/commands/app/init.ts
```diff
@@ -93,10 +99,31 @@ export default class Init extends AppLinkedCommand {
       developerPlatformClient = selectedApp.developerPlatformClient ?? developerPlatformClient
       selectAppResult = {result: 'existing', app: selectedApp}
     } else {
-      const org = await selectOrg()
+      let org: Organization
+      if (flags['organization-id']) {
+        // If an organization-id is provided, fetch the organization directly
+        const matchingOrg = await developerPlatformClient.orgFromId(flags['organization-id'])
```
dmerand: 💡 **Improvement**: The codebase already has a `fetchOrgFromId` utility in `services/dev/fetch.ts` that wraps organization lookup with comprehensive error messages via `NoOrgError`. The current implementation provides a good error message, but `fetchOrgFromId` offers additional context including clickable links to the organization, suggestions about account permissions, and guidance about partner organizations. This would provide consistency with other parts of the CLI (e.g., `services/app-context.ts`).

**Suggestion:** Consider importing and using `fetchOrgFromId` from `../../services/dev/fetch.js`:

```typescript
import {fetchOrgFromId} from '../../services/dev/fetch.js'

// Then replace lines 105-113:
org = await fetchOrgFromId(flags['organization-id'], developerPlatformClient)
```

This removes the need for custom error handling and provides consistent error messaging across the CLI.

dmerand (commented):
I have a few automated review follow-ups that might clean this up since it hasn't shipped yet.

---
_Review assisted by [pair-review](https://github.com/in-the-loop-labs/pair-review)_

alan-shoplift:
> Should we merge this now that we have app limits?

Hi 👋 - is there a public reference for what these limits are? Scoping per-PR preview envs and trying to decide between creating an app per PR vs leasing from a fixed pool.

Two quick questions if you have a sec:

1. Is the limit per-org, and does deleting an app free a slot (vs cumulative)?
2. Any CLI path to delete apps, or dev-dashboard only?

Just trying to figure out if this is something I should be worried about. Thanks for landing the non-interactive init 🙏

PR #6636, opened 2025-11-19
Add `--variable-file` flag for bulk operation mutations

ericlee878:
Resolves https://github.com/shop/issues-api-foundations/issues/1110.

### Summary
Adds a new `--variable-file` flag to shopify app execute that allows users to provide mutation variables from a JSONL file instead of passing them inline via `--variables`. This makes it easier to execute bulk operations with many variables.

### Implementation Details
- New function: `parseVariablesToJsonl()` 
  - Validates mutual exclusivity of `--variables` and `--variable-file`
  - Reads and validates file existence
  - Returns variables in JSONL format
- Restructured data flow to use JSONL strings throughout:
  - File content is passed through as JSONL
  - Reduces unnecessary conversions (no split/rejoin for files)
- Simplified `stageFile()`:
  - Now accepts variablesJsonl string parameter
  - Converts directly to Buffer for upload

### Testing
Create a file called `variables.jsonl`. This should be in the file:
```
{"input":{"id":"gid://shopify/Product/123","tags":["test"]}}
{"input":{"id":"gid://shopify/Product/456","tags":["test2"]}}
```

Run this command:
```
pnpm shopify app execute \
  --path <PATH_TO_APP> \
  -q 'mutation productUpdate($input: ProductInput!) { ... }' \
  --variable-file ./variables.jsonl
```

PR #6632, opened 2025-11-18
Add GraphQL Operation Validation for Bulk Operations CLI

ericlee878:
Resolves: https://github.com/orgs/shop/projects/208/views/34?pane=issue&itemId=139461468&issue=shop%7Cissues-api-foundations%7C1114 and https://github.com/orgs/shop/projects/208/views/34?pane=issue&itemId=139446074&issue=shop%7Cissues-api-foundations%7C1112

### Why are these changes introduced?

Improves the validation and error handling for GraphQL operations in the bulk operations service.

### Implementation Details

- Enhances GraphQL document validation to ensure exactly one operation definition is present (error handling for cases that don't have exactly one operation)
- Additional unit tests added that prove core automatically rejects malformed documents (we don't have to do this).
- Created new `validateGraphQLDocument()` function that consolidates all GraphQL validation in one place:
  - Validates exactly one operation definition exists
  - Validates --variables flag is only used with mutations
  - Checks if the operation is a mutation (returns boolean that is used later). This replaces the previous `isMutation()` function.

### Testing

Unit tests in `execute-bulk-operation.test.ts`

PR #6622, opened 2025-11-14
Add erroring for when variables are given for a BulkOps query

ericlee878:
**Resolves:** https://github.com/shop/issues-api-foundations/issues/1047?issue=shop%7Cissues-api-foundations%7C1097

### Background
Variables are only supported for [bulkOperationRunMutation](https://shopify.dev/docs/api/admin-graphql/latest/mutations/bulkoperationrunmutation), not [bulkOperationRunQuery](https://shopify.dev/docs/api/admin-graphql/latest/mutations/bulkOperationRunQuery). 

The workflow of the CLI command is shown [here](https://draw.shopify.io/embed/owaw6botp0rxvppc2z34ldsb?d=v-220.-79.2651.1443.bLH_tfkrBrIpLZ1T7vtKp). This PR is represented by the leftmost side of this diagram.

### Testing

Try running
```
pnpm shopify:run app execute \
  --path ./<TEST APP> \
  -q '{ products(first: 10) { edges { node { id title } } } }'\
  -v '[{"input": {"id": "gid://shopify/Product/<PRODUCT ID>", "tags": ["bulk-test"]}}]'
```
This should error because queries do not support variables.

While, 
```
pnpm shopify:run app execute \
  --path ./coffee-and-more \
  -q '{ products(first: 10) { edges { node { id title } } } }'\
```
should return a successful BulkOps Query.

PR #6643, opened 2025-11-19
Use `ensureAuthenticatedAdminAsApp` in `shopify app execute`

ericlee878:
Resolves: https://github.com/orgs/shop/projects/208/views/34?pane=issue&itemId=140363951&issue=shop%7Cissues-api-foundations%7C1128
Inspired by: https://github.com/Shopify/cli/pull/6596#discussion_r2543018389

### Background
The bulk operations execute command currently authenticates using the CLI user's credentials, which has two critical limitations:
1. No access to $app metafields - Cannot query or mutate app-specific metafields
2. Ignores app scopes - Does not respect the app's configured access scopes.

This prevents bulk operations from working properly with app-owned data.

### Solution
Implemented ensureAuthenticatedAdminAsApp() that authenticates as the app using OAuth 2.0 client credentials grant. This provides:
- Full access to $app metafields
- Respects the app's configured scopes
- Proper app-context authentication

This [idea](https://github.com/Shopify/cli/pull/5908/files#diff-19b0f5c6238cb4567d4897b88196fb65f7c53a3657b2b24a98ee901d507dbf4bR243-R264) originally came from a hack days.

EDIT: Actually https://github.com/Shopify/cli/pull/6649 beat us to it! This PR now just adjusts `app execute` to _use_ this feature, and doesn't have to implement it itself.

### Implementation Notes
- The API secret extraction pattern follows the existing approach in [`webhook/trigger-options.ts`](https://github.com/Shopify/cli/blob/main/packages/app/src/cli/services/webhook/trigger-options.ts#L50).
- Currently both `app` and `remoteApp` are passed to executeBulkOperation. The `app` parameter is only used for:
  - Display name (app.name)
  - Client ID (app.configuration.client_id)

Both values are also available on `remoteApp` as `title` and `apiKey`. Could be simplified in the future to only use `remoteApp` for consistency.

### Testing
To test with this new authentification, you must install the app on the store. You can do this by:
```
cd <PATH_TO_TEST_APP>
pnpm shopify app dev
```

Then, we can use the existing command works unchanged:

For queries:
```
pnpm shopify:run app execute \
  --path <PATH_TO_TEST_APPP> \
  -q '{ products(first: 10) { edges { node { id title } } } }'
```

For mutations:
```
pnpm shopify:run app execute \
  --path <PATH_TO_TEST_APPP> \
  -q 'mutation productUpdate($input: ProductInput!) { productUpdate(input: $input) { product { id tags } userErrors { field message } } }' \
  -v '[{"input": {"id": "gid://shopify/Product/9721830834416", "tags": ["bulk-test"]}}]'
```

Now authenticates as the app instead of the CLI user, enabling access to app-scoped resources.

### DIFF
--- a/packages/app/src/cli/api/graphql/app-dev/generated/types.d.ts
+++ b/packages/app/src/cli/api/graphql/app-dev/generated/types.d.ts
@@ -1,4 +1,4 @@
-/* eslint-disable @typescript-eslint/consistent-type-definitions, @typescript-eslint/naming-convention, tsdoc/syntax  */
+/* eslint-disable @typescript-eslint/consistent-type-definitions, @typescript-eslint/naming-convention, @typescript-eslint/no-explicit-any, tsdoc/syntax  */
 import {JsonMapType} from '@shopify/cli-kit/node/toml'
 
 export type Maybe<T> = T | null
@@ -16,6 +16,12 @@ export type Scalars = {
   Int: {input: number; output: number}
   Float: {input: number; output: number}
   /**
+   * Represents an [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601)-encoded date and time string.
+   * For example, 3:50 pm on September 7, 2019 in the time zone of UTC (Coordinated Universal Time) is
+   * represented as `"2019-09-07T15:50:00Z`".
+   */
+  DateTime: {input: any; output: any}
+  /**
    * A [JSON](https://www.json.org/json-en.html) object.
    *
    * Example value:
@@ -31,4 +37,12 @@ export type Scalars = {
    * }`
    */
   JSON: {input: JsonMapType | string; output: JsonMapType}
+  /**
+   * Represents an [RFC 3986](https://datatracker.ietf.org/doc/html/rfc3986) and
+   * [RFC 3987](https://datatracker.ietf.org/doc/html/rfc3987)-compliant URI string.
+   *
+   * For example, `"https://example.myshopify.com"` is a valid URL. It includes a scheme (`https`) and a host
+   * (`example.myshopify.com`).
+   */
+  URL: {input: string; output: string}
 }


ericlee878:
* **#6643** <a href="https://app.graphite.com/github/pr/Shopify/cli/6643?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#6663](https://github.com/Shopify/cli/pull/6663) <a href="https://app.graphite.com/github/pr/Shopify/cli/6663?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#6667](https://github.com/Shopify/cli/pull/6667) <a href="https://app.graphite.com/github/pr/Shopify/cli/6667?utm_source=stack-comment-icon" target="_blank">Graphite</a>) 👈 <a href="https://app.graphite.com/github/pr/Shopify/cli/6643?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#6596** <a href="https://app.graphite.com/github/pr/Shopify/cli/6596?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

packages/cli-kit/src/public/node/session.ts
```diff
@@ -227,6 +229,39 @@ ${outputToken.json(scopes)}
...
+  })
+  const normalised = await normalizeStoreFqdn(storeFqdn)
+  const tokenResponse = await shopifyFetch(`https://${normalised}/admin/oauth/access_token?${queryString}`, {
+    method: 'POST',
+    headers: {
+      'Content-Type': 'application/json',
+    },
+  })
+  const tokenJson = (await tokenResponse.json()) as {access_token: string}
+  outputDebug(outputContent`Token: ${outputToken.raw(tokenJson.access_token)}`)
```
RyanDJLee: @zzooeeyy had a great callout here that we should not be exposing this in the application logs.

ericlee878: Good point! This function is not necessary anymore because it was already created by [this](https://github.com/Shopify/cli/pull/6649), but good callout.

packages/app/src/cli/services/bulk-operations/execute-bulk-operation.ts
```diff
@@ -34,13 +36,16 @@ async function parseVariablesToJsonl(variables?: string[], variableFile?: string
...
 
   renderInfo({
     headline: 'Starting bulk operation.',
     body: `App: ${app.name}\nStore: ${storeFqdn}`,
   })
-  const adminSession = await ensureAuthenticatedAdmin(storeFqdn)
+
+  // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+  const apiSecret = remoteApp.apiSecretKeys.find((elm) => elm.secret)!.secret
+  const adminSession = await ensureAuthenticatedAdminAsApp(storeFqdn, app.configuration.client_id, apiSecret)
```
jordanverasamy: What actually is a remote app? It seems odd to me to set up a session using the client ID from `app` but the API secret from `remoteApp`... but I'm guessing this is correct for some complicated reason. Can you explain?

jordanverasamy: Ah, I see this from the PR desc:

> Both values are also available on remoteApp as title and apiKey. Could be simplified in the future to only use remoteApp for consistency.

Maybe we should adjust it to use `remoteApp` all the way through for consistency and to minimize confusion, if possible? How big of a lift is that you think?

ericlee878: I kept it for consistency. All other CLI's used app. Didn't seem like it would hurt to keep it in there?

isaacroldan: remoteApp is the remote representation of the App obtained via the API. The secret can only be obtained from the remoteApp, is not available in the "local" app

isaacroldan: you can use `remoteApp`​ to access the client_id value too yes.

ericlee878: Removed `app` and using `remoteApp` for everything.

jordanverasamy (approved):

packages/app/src/cli/services/bulk-operations/execute-bulk-operation.ts
```diff
@@ -39,13 +39,16 @@ async function parseVariablesToJsonl(variables?: string[], variableFile?: string
...
 
   renderInfo({
     headline: 'Starting bulk operation.',
-    body: `App: ${app.name}\nStore: ${storeFqdn}`,
+    body: `App: ${remoteApp.title}\nStore: ${storeFqdn}`,
   })
-  const adminSession = await ensureAuthenticatedAdmin(storeFqdn)
+
+  // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+  const apiSecret = remoteApp.apiSecretKeys.find((elm) => elm.secret)!.secret
```
gonzaloriestra:
```suggestion
  const appSecret = remoteApp.apiSecretKeys[0]?.secret
  if (!appSecret) throw new BugError('No API secret keys found for app')
```

jordanverasamy: Thanks for this feedback! I knew it was a little messy but I was just kind of letting it slide to get this out, but it's better to get a nudge to do it the right way 😅 Adjusted! TY!!

jordanverasamy:
Yeah, sorry for the outdated description -- I'll update at least the title so it'll be clearer to future readers what this PR is _actually_ doing.

Thanks for the quick review!

PR #6642, opened 2025-11-19
remove shared identity code that is now in client

alexanderMontague:
### WHY are these changes introduced?

Part of: shop/issues-develop#21594

- Remove the identity code that is now contained within the identity client

PR #6654, opened 2025-11-21
Refactor identity client code execution

alexanderMontague:

Part of: shop/issues-develop#21594

- Refactor the auth/identity flows slightly to make the client lightweight
- This also removes circular dependencies between existing auth modules and our new client

- pull shared util code into separate module
- only house identity network flows within the client

### How to test your changes?

- Have core running always
- Supported env configurations:
1. BP Online ✅  + Identity Online ✅ 
2. BP Online ✅  + Identity Offline ❌ 
3. Prod auth flows with local CLI ✅ 

- I'm running a simple `pnpm shopify auth logout && SHOPIFY_CLI_NEVER_USE_PARTNERS_API=1 SHOPIFY_SERVICE_ENV=local pnpm shopify app info`
- Then the same command without `pnpm shopify auth logout` to test without the entire flow
- Do this for all environment combinations

We are still blocked on BP + Management API auth checks, so these patches are required to allow the CLI to authenticate correctly in all non-prod scenarios

<details>
<summary>BP Patch</summary>

In `db/data/access/api_clients.yml` add

```yml
- name: shopify-cli-development
  client_id: shopify-cli-development # key change
  scopes:
  - # <existing scopes as other development client>
  environments:
  - development
  - test
```
</details>

<details>
<summary>Core Patch</summary>

In `components/apps/app/controllers/concerns/organization_authorization_concern.rb#organization_user`

```ruby
if DevelopmentSupport::HostDetection.running?(service_name: "business-platform") &&
        DevelopmentSupport::HostDetection.running?(service_name: "identity")
      @organization_user ||= 
        AccessAndAuth::FetchOrganizationUser.perform(
          # ...
        )
      )
    else
      @organization_user ||= AccessAndAuth::OrganizationUser.new( # key change to mock out user
        id: "gid://Shopify/OrganizationUser/1",
        status: "good",
        organization_permissions: [:develop_apps],
        owner: true,
        organization: AccessAndAuth::Organization.new(organization_id: 1),
        full_name: 'Dev User',
        email: "dev@shopify.com",
        administrative_permissions: []
      )
    end
```
</details>

### DIFF
--- a/packages/cli-kit/src/private/node/clients/identity/identity-client.ts
+++ b/packages/cli-kit/src/private/node/clients/identity/identity-client.ts
@@ -1,18 +1,31 @@
-import {IdentityToken} from '../../session/schema.js'
-import {TokenRequestResult} from '../../session/exchange.js'
 import {API} from '../../api.js'
 import {Environment, serviceEnvironment} from '../../context/service.js'
 import {BugError} from '../../../../public/node/error.js'
 import {Result} from '../../../../public/node/result.js'
 
-export abstract class IdentityClient {
-  abstract requestAccessToken(scopes: string[]): Promise<IdentityToken>
+export interface TokenRequestResult {
+  access_token: string
+  expires_in: number
+  refresh_token: string
+  scope: string
+  id_token?: string
+}
 
+export interface DeviceAuthorizationResponse {
+  deviceCode: string
+  userCode: string
+  verificationUri: string
+  expiresIn: number
+  verificationUriComplete?: string
+  interval?: number
+}
+
+export abstract class IdentityClient {
   abstract tokenRequest(params: {
     [key: string]: string
   }): Promise<Result<TokenRequestResult, {error: string; store?: string}>>
 
-  abstract refreshAccessToken(currentToken: IdentityToken): Promise<IdentityToken>
+  abstract requestDeviceAuthorization(scopes: string[]): Promise<DeviceAuthorizationResponse>
 
   abstract clientId(): string
 
--- a/packages/cli-kit/src/private/node/clients/identity/identity-service-client.ts
+++ b/packages/cli-kit/src/private/node/clients/identity/identity-service-client.ts
@@ -1,11 +1,4 @@
-import {IdentityClient} from './identity-client.js'
-import {IdentityToken} from '../../session/schema.js'
-import {
-  buildIdentityToken,
-  exchangeDeviceCodeForAccessToken,
-  tokenRequestErrorHandler,
-  TokenRequestResult,
-} from '../../session/exchange.js'
+import {IdentityClient, type TokenRequestResult, type DeviceAuthorizationResponse} from './identity-client.js'
 import {outputContent, outputDebug, outputInfo, outputToken} from '../../../../public/node/output.js'
 import {AbortError, BugError} from '../../../../public/node/error.js'
 import {identityFqdn} from '../../../../public/node/context/fqdn.js'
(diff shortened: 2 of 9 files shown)


alexanderMontague:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/Shopify/cli/6654?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#6654** <a href="https://app.graphite.com/github/pr/Shopify/cli/6654?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Shopify/cli/6654?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#6635** <a href="https://app.graphite.com/github/pr/Shopify/cli/6635?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6634** <a href="https://app.graphite.com/github/pr/Shopify/cli/6634?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

packages/cli-kit/src/private/node/clients/identity/identity-client.ts
alexanderMontague, line 1: I've moved only flows pertaining to network requests specifically to the client to keep it super lightweight. Shared modules in `session/*` are then pulled in to existing flows as needed, with the client only being used for the token exchanges/fetches

packages/cli-kit/src/private/node/session/token-utils.ts
alexanderMontague, line 1: This shared modules prevents circular dependencies as both the existing modules and our client pulls from them.

alexanderMontague: This is also fully tested by the other existing shared module tests, but I can add another layer of coverage here if we wanted

packages/cli-kit/src/private/node/clients/identity/identity-client.ts
```diff
@@ -1,18 +1,31 @@
...
+
+export abstract class IdentityClient {
   abstract tokenRequest(params: {
     [key: string]: string
   }): Promise<Result<TokenRequestResult, {error: string; store?: string}>>
 
-  abstract refreshAccessToken(currentToken: IdentityToken): Promise<IdentityToken>
+  abstract requestDeviceAuthorization(scopes: string[]): Promise<DeviceAuthorizationResponse>
 
   abstract clientId(): string
```
MitchDickinson, line 30: `clientId()` doesn't need to be abstract here. Instead, we should leave it as a class-specific implementation for the mock/production clients and then fix the temporal coupling issues related to it by changing the interface

packages/cli-kit/src/private/node/session/exchange.ts
```diff
@@ -20,6 +19,71 @@ export interface ExchangeScopes {
...
 }
 
+/**
+ * Request an identity token using the device authorization flow.
+ * This initiates the full flow: request device code, show to user, and poll for completion.
+ * @param scopes - The scopes to request
+ * @returns An identity token
+ */
+export async function requestAccessToken(scopes: string[]): Promise<IdentityToken> {
+  const identityClient = getIdentityClient()
```
MitchDickinson, line 29: I think there's more temporal coupling stuff going on. Example: refreshAccessToken in exchange requires clientId() to be called and then passed into tokenRequest following a specific shape. 

This method invokes pollForDeviceAuthorization which a few levels down invokes tokenRequest. For this one, we aren't propagating the identityClient that we resolve in this method (line 29), so `getIdentityClient()` is assumed to be effectively global from the POV of this class (since everyone calls out to it) which I don't think is the right thing. If it's important that the code here interacts with the same identity client through its flows, it should resolve it once and propagate it

packages/cli-kit/src/private/node/clients/identity/identity-service-client.ts
```diff
@@ -169,55 +125,4 @@ export class IdentityServiceClient extends IdentityClient {
...
-   * The endpoint will return `authorization_pending` until the user completes the auth flow in the browser.
-   * Once the user completes the auth flow, the endpoint will return the identity token.
-   *
-   * Timeout for the polling is defined by the server and is around 600 seconds.
-   *
-   * @param code - The device code obtained after starting a device identity flow
-   * @param interval - The interval to poll the token endpoint
-   * @returns The identity token
-   */
-  private async pollForDeviceAuthorization(code: string, interval = 5): Promise<IdentityToken> {
```
MitchDickinson, line 184: This feels like a loss to me - this code is directly related to the service client and the encapsulation was nice IMO

packages/cli-kit/src/private/node/clients/identity/identity-client.ts
MitchDickinson: IMO thinning out the client here weakens the abstraction:
- There's code in sesssion/exchange that is actually part of the service client "stack" but isn't part of the service client (i.e. polling code that doesn't execute if you are using the mock client)
- There's more temporal coupling going on (call this method, then that method).

Light weight isn't necessarily good in this context - I think the "right weight" (ahaha rhyming) is what we want, where the implementation code that "belongs" behind the interface lives behind the interface.

Having an abstraction just over the network part is better than having none (we **do** need the local dev solution!), but my gut here is that the circular dependency issue led us to making other changes that weakened the rest of the architecture you had established.

I strongly suspect there would be another solution to the circular dependency issue that would still allow us to keep the higher level interface with more encapsulated code behind the interface. My guess is there would be even more things we could pull behind the interface as a way to fix the circular dependency but I can't be sure without reviewing what the circular relationship actually was (is this written up somewhere I can 👀 ?)

PR #6655, opened 2025-11-21
Remove unused field selections from `BulkOperation` GraphQL documents

jordanverasamy:
Adjacent to https://github.com/shop/issues-api-foundations/issues/1093, but not directly related.

### WHY are these changes introduced?

I noticed while working on https://github.com/Shopify/cli/pull/6595 that we are asking for fields that we don't use in some of our GraphQL mutations. I want us to follow GraphQL best practices so I'm cleaning that up :)

### WHAT is this pull request doing?

Clean up the field selections of `bulkOperationRunMutation` and `bulkOperationRunQuery`, so that we don't ask for any fields that we won't actually use.

### How to test your changes?

All tests pass and all tophat steps from https://github.com/Shopify/cli/pull/6595 still work.

PR #6656, opened 2025-11-22
Introduce `--output-file` flag, write bulk operation results to STDOUT or a file

jordanverasamy:

https://github.com/shop/issues-api-foundations/issues/1096

Adds behaviour to `shopify app execute` to match the designs:

- if `--watch` and `--output-file` are both provided, write results to that file after completion
- if `--watch` is provided but `--output-file` is not, write results to STDOUT after completion
- if `--watch` is not provided, no change to behaviour

### How to test your changes?

Try running `shopify app execute` with `--watch` and `--output-file` and verify it writes to the correct file:

```
pnpm shopify app execute --path=<PATH_TO_YOUR_APP> --query="{ products { edges { node { id } } } }" --watch --output-file=my-results.jsonl
```

image

image

Then try doing the same thing without `--output-file`, and verify it prints to STDOUT:

```
pnpm shopify app execute --path=<PATH_TO_YOUR_APP> --query="{ products { edges { node { id } } } }" --watch
```

image

### DIFF
--- a/packages/app/src/cli/flags.ts
+++ b/packages/app/src/cli/flags.ts
@@ -68,4 +68,8 @@ export const bulkOperationFlags = {
     env: 'SHOPIFY_FLAG_WATCH',
     default: false,
   }),
+  'output-file': Flags.string({
+    description: 'The file path where results should be written. If not specified, results will be written to STDOUT.',
+    env: 'SHOPIFY_FLAG_OUTPUT_FILE',
+  }),
 }
--- a/packages/app/src/cli/services/bulk-operations/download-bulk-operation-results.test.ts
+++ b/packages/app/src/cli/services/bulk-operations/download-bulk-operation-results.test.ts
@@ -1,0 +1,35 @@
+import {downloadBulkOperationResults} from './download-bulk-operation-results.js'
+import {fetch} from '@shopify/cli-kit/node/http'
+import {describe, test, expect, vi} from 'vitest'
+
+vi.mock('@shopify/cli-kit/node/http')
+
+describe('downloadBulkOperationResults', () => {
+  test('returns text content when fetch is successful', async () => {
+    const mockUrl = 'https://example.com/results.jsonl'
+    const mockContent = '{"id":"gid://shopify/Product/123"}\n{"id":"gid://shopify/Product/456"}'
+
+    vi.mocked(fetch).mockResolvedValue({
+      ok: true,
+      text: async () => mockContent,
+    } as Awaited<ReturnType<typeof fetch>>)
+
+    const result = await downloadBulkOperationResults(mockUrl)
+
+    expect(fetch).toHaveBeenCalledWith(mockUrl)
+    expect(result).toBe(mockContent)
+  })
+
+  test('throws error when fetch fails', async () => {
+    const mockUrl = 'https://example.com/results.jsonl'
+
+    vi.mocked(fetch).mockResolvedValue({
+      ok: false,
+      statusText: 'Not Found',
+    } as Awaited<ReturnType<typeof fetch>>)
+
+    await expect(downloadBulkOperationResults(mockUrl)).rejects.toThrow(
+      'Failed to download bulk operation results: Not Found',
+    )
+  })
+})
--- a/packages/app/src/cli/services/bulk-operations/download-bulk-operation-results.ts
+++ b/packages/app/src/cli/services/bulk-operations/download-bulk-operation-results.ts
@@ -1,0 +1,12 @@
+import {fetch} from '@shopify/cli-kit/node/http'
+import {AbortError} from '@shopify/cli-kit/node/error'
+
+export async function downloadBulkOperationResults(url: string): Promise<string> {
+  const response = await fetch(url)
+
+  if (!response.ok) {
+    throw new AbortError(`Failed to download bulk operation results: ${response.statusText}`)
+  }
+
+  return response.text()
+}
--- a/packages/app/src/cli/services/bulk-operations/execute-bulk-operation.test.ts
+++ b/packages/app/src/cli/services/bulk-operations/execute-bulk-operation.test.ts
@@ -2,6 +2,7 @@ import {executeBulkOperation} from './execute-bulk-operation.js'
 import {runBulkOperationQuery} from './run-query.js'
 import {runBulkOperationMutation} from './run-mutation.js'
 import {watchBulkOperation} from './watch-bulk-operation.js'
+import {downloadBulkOperationResults} from './download-bulk-operation-results.js'
 import {AppLinkedInterface} from '../../models/app/app.js'
 import {BulkOperationRunQueryMutation} from '../../api/graphql/bulk-operations/generated/bulk-operation-run-query.js'
 import {BulkOperationRunMutationMutation} from '../../api/graphql/bulk-operations/generated/bulk-operation-run-mutation.js'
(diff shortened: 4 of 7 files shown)


jordanverasamy:
* **#6656** <a href="https://app.graphite.com/github/pr/Shopify/cli/6656?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Shopify/cli/6656?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#6595** <a href="https://app.graphite.com/github/pr/Shopify/cli/6595?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#6655](https://github.com/Shopify/cli/pull/6655) <a href="https://app.graphite.com/github/pr/Shopify/cli/6655?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#6588** <a href="https://app.graphite.com/github/pr/Shopify/cli/6588?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#6596](https://github.com/Shopify/cli/pull/6596) <a href="https://app.graphite.com/github/pr/Shopify/cli/6596?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* `main`


<h2></h2>

packages/app/src/cli/services/bulk-operations/download-bulk-operation-results.test.ts
```diff
@@ -0,0 +1,35 @@
...
+
+    vi.mocked(fetch).mockResolvedValue({
+      ok: true,
+      text: async () => mockContent,
+    } as Awaited<ReturnType<typeof fetch>>)
+
+    const result = await downloadBulkOperationResults(mockUrl)
+
+    expect(fetch).toHaveBeenCalledWith(mockUrl)
+    expect(result).toBe(mockContent)
```
ericlee878, line 20: Isn't this just testing the mock behavior? Seems unnecessary. Maybe I'm just missing something though.
```suggestion
```

jordanverasamy: Good question. The thing it is testing is small and easy to overlook. Let me explain.

Imagine if we accidentally broke the code, such that we do the same fetching, but then return `null`. If the code is broken like that, we want this test to fail, right? But if we remove this assertion, the test would incorrectly pass! We'd only be asserting that `fetch` was called correctly, and it was.

This line is asserting that the returned `result` is correct. (That value is flowing through the real code, we're not only testing a mock -- we got `result` by calling the actual function.) That's an important piece of behaviour and we do want to assert it.

ericlee878: Hmm ok so we're basically just testing to check `result` is not `null`. Ok makes sense. Thanks for the explanation!

jordanverasamy: Not just that it's not null -- we're asserting that `result` is _exactly_ correct. Anything else would fail the test, which is good!

ericlee878: I see. We're testing the to see if `result` is correct. Okok makes sense. Thanks for the clarification!

packages/app/src/cli/services/bulk-operations/download-bulk-operation-results.ts
```diff
@@ -0,0 +1,12 @@
...
+
+export async function downloadBulkOperationResults(url: string): Promise<string> {
+  const response = await fetch(url)
+
+  if (!response.ok) {
+    throw new AbortError(`Failed to download bulk operation results: ${response.statusText}`)
+  }
+
+  return response.text()
+}
```
isaacroldan, line 12: Why have this function in a different file? is just wrapping a call to `fetch`​ and a error? And is only used in a single place... best to put the call to fetch directly where it is needed. I'd say the tests don't add much value either (you are just testing an `if`​)

jordanverasamy: Yeah good question! It's really small and like you said, doesnt have any complexity to test. I waffled on whether it should be in its own file, and ultimately decided to, because we're going to want the same behaviour in the `bulk status` command ([more info here](https://github.com/shop/issues-api-foundations/issues/1142)) so may as well have it available to reuse.

I'm happy to just duplicate the function in both `execute` and `status` if you would prefer, since it's so simple, just a couple lines.

packages/app/src/cli/flags.ts
```diff
@@ -68,4 +68,8 @@ export const bulkOperationFlags = {
     env: 'SHOPIFY_FLAG_WATCH',
     default: false,
   }),
+  'output-file': Flags.string({
+    description: 'The file path where results should be written. If not specified, results will be written to STDOUT.',
+    env: 'SHOPIFY_FLAG_OUTPUT_FILE',
```
gonzaloriestra, line 73: In theory, if you add this it will throw an error when used without `watch`

```suggestion
    env: 'SHOPIFY_FLAG_OUTPUT_FILE',
    compatible: ['watch'],
```

packages/app/src/cli/services/bulk-operations/execute-bulk-operation.ts
```diff
@@ -97,8 +99,15 @@ function renderBulkOperationResult(operation: BulkOperation): void {
...
       if (operation.url) {
-        const downloadMessage = outputContent`Download results ${outputToken.link('here.', operation.url)}`.value
-        renderSuccess({headline, body: [downloadMessage], customSections})
+        const results = await downloadBulkOperationResults(operation.url)
+
+        if (outputFile) {
+          await writeFile(outputFile, results)
+          renderSuccess({headline, body: [`Results written to ${outputFile}`], customSections})
+        } else {
+          process.stdout.write(results)
```
gonzaloriestra: This is the standard way, and it also adds an additional break line that was missing in the end. 

```suggestion
          outputResult(results)
```

packages/app/src/cli/services/bulk-operations/execute-bulk-operation.test.ts
```diff
@@ -346,11 +350,75 @@ describe('executeBulkOperation', () => {
...
+      userErrors: [],
+    }
+    const completedOperation = {
+      ...createdBulkOperation,
+      status: 'COMPLETED' as const,
+      url: 'https://example.com/download',
+      objectCount: '2',
+    }
+
+    const stdoutWriteSpy = vi.spyOn(process.stdout, 'write').mockImplementation(() => true)
```
gonzaloriestra: You can use [this function](https://github.com/Shopify/cli/blob/21e748f1c1a28f71037e45e35ee036a845cfd51b/packages/cli-kit/src/public/node/testing/output.ts#L18) to test outputs

gonzaloriestra (approved):

packages/app/src/cli/flags.ts
jordanverasamy: Interesting. I tried this and it didn't work for me:

```
pnpm shopify app execute --path="../my-second-test-app" --query="{ products { edges { node { id variants { edges { node { id } } } } } } }" --output-file='hello.jsonl'
```

image

but maybe I'm missing something?

jordanverasamy: Let me know and I'm happy to add this in a followup though :)

packages/app/src/cli/services/bulk-operations/execute-bulk-operation.ts
jordanverasamy: Oh awesome thanks!! Updated the code to match :)

packages/app/src/cli/services/bulk-operations/execute-bulk-operation.test.ts
jordanverasamy: Awesome, that's way better, thanks :))

PR #6657, opened 2025-11-22
Allow `--development` flag to create a new theme

graygilmore:

Developers often want to create a new development theme on a per-PR basis when reviewing theme changes. This is difficult right now because our systems try to only have one development theme per user which means that GitHub's CI is considered a single user and development themes get overwritten all the time.

This PR solves this by allowing developers to provide a context by using the`--development-context` flag. The context in this case is just a name but we refer to it as a context because it's specific to development themes and we don't want to confuse with the `--theme` flag which accepts both an ID and a name.

When a user provides `--development-context` we'll search for all development themes with that name. If we find one we'll push to it otherwise we'll create a new development theme. This lets developers decide whether to keep long running contexts that get overwritten (e.g. `PR #<number>`) or new development themes on every push (e.g. `PR #<number> - <latest_sha>`).

```sh
shopify theme push --development --development-context "PR #6657"
shopify theme push -d -c "PR #6657"
```

### How to test your changes?

1. Installed the snapped version `npm i -g --@shopify:registry=https://registry.npmjs.org @shopify/cli@0.0.0-snapshot-20260313172428`
2. Run `shopify theme list --role development` to see all of your development themes
3. Run `shopify theme push -d -c test-1`
4. Run `shopify theme list --role development` again and ensure you see the new theme
5. Run `shopify theme push -d -c test-2`
6. Run `shopify theme list --role development` again and ensure a new theme is created (ie, the previous development theme isn't replaced)
7. Run `shopify theme push -d -c test-1` again
8. Run `shopify theme list --role development` again and ensure test-1 was replaced and there aren't two themes named test-1 (bonus: make sure `test-1` is marked as `[current]`!)

### DIFF
--- a/packages/cli-kit/src/public/node/themes/theme-manager.ts
+++ b/packages/cli-kit/src/public/node/themes/theme-manager.ts
@@ -15,7 +15,9 @@ export abstract class ThemeManager {
 
   async findOrCreate(name?: string, role?: Role): Promise<Theme> {
     let theme = await this.fetch(name, role)
-    if (!theme) {
+    if (theme) {
+      this.setTheme(theme.id.toString())
+    } else {
       theme = await this.create(role, name)
     }
     return theme
@@ -29,8 +31,7 @@ export abstract class ThemeManager {
     const theme =
       name && role === DEVELOPMENT_THEME_ROLE
         ? await findDevelopmentThemeByName(name, this.adminSession)
-        : // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
-          await fetchTheme(parseInt(this.themeId!, 10), this.adminSession)
+        : await fetchTheme(parseInt(this.themeId!, 10), this.adminSession)
 
     if (!theme) {
       this.removeTheme()
--- a/.changeset/every-years-study.md
+++ b/.changeset/every-years-study.md
@@ -1,0 +1,5 @@
+---
+'@shopify/cli-kit': patch
+---
+
+Set the current theme in local development on create or find
--- a/packages/cli-kit/src/public/node/themes/theme-manager.test.ts
+++ b/packages/cli-kit/src/public/node/themes/theme-manager.test.ts
@@ -68,6 +68,7 @@ describe('ThemeManager', () => {
       expect(fetchTheme).toHaveBeenCalledWith(123, session)
       expect(result).toEqual(mockTheme)
       expect(themeCreate).not.toHaveBeenCalled()
+      expect(manager.getStoredThemeId()).toBe('123')
     })
 
     test('creates a new theme when one does not exist', async () => {
@@ -105,6 +106,7 @@ describe('ThemeManager', () => {
       expect(findDevelopmentThemeByName).toHaveBeenCalledWith('Dev', session)
       expect(result).toEqual(mockTheme)
       expect(themeCreate).not.toHaveBeenCalled()
+      expect(manager.getStoredThemeId()).toBe('123')
     })
   })
 


binarymonkey84:
Hey @graygilmore 

This looks like excellent progress towards [the issue I raised on the dev forums](https://community.shopify.dev/t/create-a-unique-dev-theme-instance-with-shopify-theme-push-github-ci-action/26077), nice one!

I haven't had a chance to test the snapshotted version yet, but, some questions/notes:

* I think we'd probably use the commit ID as the name, or part of the name of the theme, to avoid collisions and guarantee I'm looking at the theme with the code at that specific commit state.
  * Maybe update the relevant docs to suggest commit ID?
* Because this is going to be used in CI, I think we need to make sure the `--json` CLI flag works so the action script in Github can grab the theme ID, theme editor link, and theme preview link easily.
  * Can this be added to the tests?
* It would be good to add clarity on how long the dev context theme will live for (before it is auto-deleted by Shopify) when it is used in this way. 
  * At the moment, [the docs say](url) dev themes are deleted when running 'shopify auth logout'. 
  * > Development themes are deleted when you run shopify auth logout. If you need a preview link that can be used after you log out, then you should [share](https://shopify.dev/docs/api/shopify-cli/theme/theme-share) your theme or [push](https://shopify.dev/docs/api/shopify-cli/theme/theme-push) to an unpublished theme on your store.  
  * How will this work in a CI context, when we'll be using a theme access password?
  * Can the docs be updated to communicate how and when these types of dev themes get deleted / how long they persist for and any rules around that? I'm assuming they don't live forever, otherwise Shopify will quickly see storage use getting out of control :)
  * Can we add a test to check the command works with theme access passwords?
  
Thanks again and I hope this is useful!

Really keen to see this go live so we can upgrade our CI process and streamline review.

packages/cli-kit/src/public/node/themes/api.ts
```diff
@@ -113,6 +114,38 @@ export async function fetchThemes(session: AdminSession): Promise<Theme[]> {
...
+    preferredBehaviour: THEME_API_NETWORK_BEHAVIOUR,
+  })
+
+  if (!themes) {
+    unexpectedGraphQLError('Failed to fetch themes')
+  }
+
+  if (themes.nodes.length > 1) {
+    throw recordError(new AbortError(`More than one development theme is named "${name}"`))
+  }
```
aswamy, line 134: Instead of fetching 2 themes with the same name, and checking whether or not more than 1 exists, could we just add the validation on the server side? This seems a bit fragile.

aswamy: If we can't make the restriction on the server-side, why block the attempt of fetching a theme if more than 1 theme with this name already exists? We could just warn the user in the CLI as a warning, and return the latest one.

packages/cli/oclif.manifest.json
```diff
@@ -7082,6 +7082,21 @@
           "name": "development",
           "type": "boolean"
         },
+        "development-context": {
```
aswamy, line 7230: Since `development-context` _needs_ `development` flag to be passed in, have you considered just converting the `development` flag from a boolean to allowing an optional context?  
  
Since they are bound together, why have two separate flags?

e.g.

```
shopify theme --development="PR#123"
```

packages/cli-kit/src/public/node/themes/theme-manager.ts
```diff
@@ -13,19 +13,24 @@ export abstract class ThemeManager {
 
   constructor(protected adminSession: AdminSession) {}
 
-  async findOrCreate(): Promise<Theme> {
-    let theme = await this.fetch()
+  async findOrCreate(name?: string): Promise<Theme> {
+    let theme = await this.fetch(name)
     if (!theme) {
-      theme = await this.create()
+      theme = await this.create(DEVELOPMENT_THEME_ROLE, name)
```
aswamy: I think `DEVELOPMENT_THEME_ROLE` should also be passed in. Feels weird that a `findOrCreate` method will _always_ create a development theme. Same thing for `fetch` below. The name of the method doesn't really express what it does.

packages/cli/oclif.manifest.json
graygilmore: Oooo that's interesting. I don't want to break any existing usage of `--development` but I think so long as the argument is optional we should be good.

packages/cli-kit/src/public/node/themes/theme-manager.ts
graygilmore: I had originally changed this because it wasn't actually a change in behavior since `create` will fallback to development:

```ts
async create(themeRole?: Role, themeName?: string) {
  const name = themeName ?? generateThemeName(this.context)
  const role = themeRole ?? DEVELOPMENT_THEME_ROLE
```

But I think being explicit here is better so I'll add the arg!

packages/cli/oclif.manifest.json
graygilmore: Loved it: https://github.com/Shopify/cli/compare/825b91ed76ca17a3f4f8919cec41f48cfb6181c5..c47c744e02ffefe09bed5b9db54e10e1c6cfb5ed#diff-21cf3d019379cf249aa5931b8651b06c6b34e73f907e728ee58e90bb89971917L60-R60

packages/cli-kit/src/public/node/themes/theme-manager.ts
graygilmore: Fixed: https://github.com/Shopify/cli/compare/c47c744e02ffefe09bed5b9db54e10e1c6cfb5ed..c78144703d371ccbee6a9f2c38c3960cd3304f03

packages/cli-kit/src/public/node/themes/api.ts
graygilmore: I think a backend validation would be overkill here and I don't want to stop *all* workflows available with the Admin API. Through the CLI this scenario shouldn't really be hit because we'll always try to reuse the same theme if we find one.

I'm happy to change the abort to a warning, though!

.changeset/puny-plants-allow.md
graygilmore, line 1: 👋🏻 @binarymonkey84! Going to respond here so we can thread replies

> Maybe update the relevant docs to suggest commit ID?

This is fully up to theme developers to decide how they want to leverage this. Maybe we could add it as an example of configuring a GitHub action, though!

> Can this be added to the tests?

This is already working! The additional functionality I've added doesn't have any effect on the `--json` flag so this should continue working as normal.

> It would be good to add clarity on how long the dev context theme will live for (before it is auto-deleted by Shopify) when it is used in this way. [...] Can the docs be updated to communicate how and when these types of dev themes get deleted / how long they persist for and any rules around that?

I'm not sure if we've ever shared the exact timing for this. I'll sync with the team to see if we can get some official documentation around this. Keep in mind, too, that there is a limit to the amount of development themes you can have (also undocumented, we'll see about having official guidelines on that, too).

> At the moment, the docs say dev themes are deleted when running 'shopify auth logout'.

This might have been true at some point but is not true now. We'll get this cleaned up.

> How will this work in a CI context, when we'll be using a theme access password?

In CI you'll be running the same command but also passing the `--password` flag. Perhaps I'm misunderstanding? Or is this in relation to the confusing wording about when development themes are deleted?

packages/theme/src/cli/utilities/theme-selector.test.ts
```diff
@@ -46,7 +46,7 @@ describe('findOrSelectTheme', () => {
     expect(theme).toBe(selectedTheme)
   })
 
-  test('flags development theme as [yours]', async () => {
+  test('flags development theme as [most-recent]', async () => {
```
aswamy:
```suggestion
  test('flags development theme as [current]', async () => {
```

packages/theme/src/cli/commands/theme/push.ts
```diff
@@ -57,11 +57,19 @@ export default class Push extends ThemeCommand {
...
       env: 'SHOPIFY_FLAG_DEVELOPMENT',
-    }),
+      parse: async (input: string | undefined) => {
+        if (input === undefined) {
+          return true
+        }
+
+        return input
+      },
+    })(),
```
aswamy: Instead of having a custom type, could we just change it to a string?  
  
Logic:

- when flag is set && value == "" => development theme with no custom name

- when flag is set && value.length > 0 => development theme with a custom name

Not sure why it's bugging out when i tophat it, but it's currently saying `--development` flag needs a value. Maybe oclif needs to be regenerated?

graygilmore: After all that it looks like you [can't have a flag that optionally accepts a value](https://oclif.io/docs/flags) 😭 

> Flag options are non-positional arguments passed to the command. Flags can either be option flags which take an argument, or boolean flags which do not. An option flag must have an argument.

aswamy: im sad you had to do so much work for this :( hoping it was mostly claude 🫣

.changeset/puny-plants-allow.md
binarymonkey84: Hey @graygilmore 

> Maybe update the relevant docs to suggest commit ID?
> 
> This is fully up to theme developers to decide how they want to leverage this. Maybe we could add it as an example of configuring a GitHub action, though!

Yeah, I guess it's just because in the PR there were already opinions on how devs could name their theme (think it was pull request name and something else), so I thought if we're making suggestions, then suggesting commit ID might be worth doing since it encourages using this for theme instances that are more likely to correspond to a PR at a specific commit. Not too fussed on this though!

> This is already working! The additional functionality I've added doesn't have any effect on the --json flag so this should continue working as normal.

Great! Wasn't sure if the --json flag was only supported with certain flags or contexts. Given this will be probably used by developers in conjunction with the command, makes sense to check it does work but if it's guaranteed to work then all good.

> It would be good to add clarity on how long the dev context theme will live for (before it is auto-deleted by Shopify) when it is used in this way. [...] Can the docs be updated to communicate how and when these types of dev themes get deleted / how long they persist for and any rules around that?
> 
> I'm not sure if we've ever shared the exact timing for this. I'll sync with the team to see if we can get some official documentation around this. Keep in mind, too, that there is a limit to the amount of development themes you can have (also undocumented, we'll see about having official guidelines on that, too).
> 
> At the moment, the docs say dev themes are deleted when running 'shopify auth logout'.
> 
> This might have been true at some point but is not true now. We'll get this cleaned up.
> 
Amazing, yes having that clarity would really help to plan and set expectations.


> How will this work in a CI context, when we'll be using a theme access password?
> 
> In CI you'll be running the same command but also passing the --password flag. Perhaps I'm misunderstanding? Or is this in relation to the confusing wording about when development themes are deleted?

It's just in relation to the confusing wording :) As in, if themes are deleted on 'shopify auth logout', but that CLI command won't be run in the CI action, it wasn't clear when the theme would actually be deleted.

Cheers for all your work on this!

binarymonkey84:
@graygilmore How's it going with this one?

graygilmore:
Hey @binarymonkey84! I've been off on a long vacation and just getting back into the swing of things over here. I'll make sure to have a look at this soon.

EvilGenius13:
Small point out: It's probably unlikely, but if anyone was using the `[yours]` in any actions or scripts, that would break for them since we changed it to `[current]`. Do you think it warrants the changesets being minor? I'm not hard set on a direction.

graygilmore:
> Do you think it warrants the changesets being minor? I'm not hard set on a direction.

Since the versions of the packages will be bumped in the same way regardless I don't mind leaving it as-is.

Issue #6293, opened 2025-08-20
[Feature]: Skip prompts when using shopify app config link -c with existing config


muchisx:
### What area(s) will this request affect?

App

### What type of change do you want to see?

New feature

### Overview

Improve the behavior of `shopify app config link -c <configName>` so that when an existing `.toml` file is specified, the CLI refreshes or updates the file directly without prompting for organization or app selection again.

Following up on:
https://github.com/Shopify/cli/issues/6202

### Motivation

Currently, even when using the `-c` or `--config` flag to target an existing config file, the CLI proceeds through the full interactive flow (org selection, app linking). This is repetitive and inefficient, especially for developers working across multiple devices or syncing environments. Supporting a streamlined update flow when the target config already exists would make `link` much more useful and aligned with developer expectations.

isaacroldan:
Hey @muchisx, I don't quite understand, the whole point of `config link` is selecting a new app (in the same or a different organization). What would you expect for this command to update if not that? It doesn't update anything else in the toml 🤔

nickwesselman:
Hi @muchisx --

I think what you are asking for is a way to sync the remote config to an existing `shopify.app.toml`. Maybe something like `shopify app config pull` would be better for that? 🤔 

I'm curious about your workflow though -- are you and your team often making changes directly in the dashboard, vs making them in the toml and running `deploy`? Can you tell us the reasons why?

muchisx:
Hi @isaacroldan @nickwesselman !

Thank you for the feedback!

@nickwesselman 
> I'm curious about your workflow though -- are you and your team often making changes directly in the dashboard, vs making them in the toml and running deploy? Can you tell us the reasons why?

Here's my workflow:
- In `.gitignore` we have `shopify.app*.toml` because:
	1.  We are multiple devs and and we don't want to commit all of the dev apps tomls
	2.  We do not want any dev connecting/running dev to the production app by mistake, so we do not commit it either.
- When we do productions deployments we need to make sure the production `app.toml` is up to date (because it's not committed). Most of the times it can be that a scope is in production but not in local, because we might deploy from different devices.
- So before deploying we always run a `config:link` for the production config, just to make sure we won't be deleting/removing any scopes or configuration that were introduced from other devices.
- Lastly, the reason for the PR is that since the toml is already in the PC, we just need to pull/update some values that might be stale (as you suggested with the `pull`), without going through the processes of selecting which org, which app, etc... Just with the config name.

[...]

PR #6662, opened 2025-11-24
fix(app): skip app prompt when config is already linked

anakafeel:
# WHY are these changes introduced?

The original issue (#6293) was about reducing unnecessary prompts when working with existing app configuration files.

After clarifications, it became clear that the behavior of:

`shopify app config link --config <file>`

should **not** be changed — even when a valid `client_id` exists, the command must still prompt for org/app selection.

The correct solution is to introduce a new command that **pulls the remote app information** for an already-linked configuration file, *without any prompts*.

---

# WHAT is this pull request doing?

This PR introduces a new command:

`shopify app config pull`

This new command:

- Detects the currently selected config (or the one specified with `--config`)
- Reads and validates the existing `client_id`
- Resolves the remote app using the already-linked org/app
- Pulls the remote configuration data without any interactive prompts
- Leaves the existing `app config link` behavior unchanged

This approach aligns with the expected behavior discussed in issue #6293.

---

# How to test your changes

### 1. Prepare a config file with a valid `client_id`

Example:

`shopify.app.staging.toml`

### 2. Run the new command

`pnpm shopify app config pull --config shopify.app.staging.toml`

###  Expected result

- No prompts appear  
- CLI resolves the remote app from the existing `client_id`  
- The remote app info is pulled successfully  
- A success message is printed  

---

### 3. Run without providing `--config`

`pnpm shopify app config pull`

- Should automatically use the currently selected config file.

---

### 4. Ensure `app config link` behavior is unchanged

`pnpm shopify app config link --config shopify.app.staging.toml`

- Prompts should still appear as they do today.

---

# Post-release steps

No additional steps required — the new command is self-contained.

### DIFF
--- a/packages/app/src/cli/services/app/config/pull.ts
+++ b/packages/app/src/cli/services/app/config/pull.ts
@@ -3,7 +3,7 @@
 import {LinkOptions, loadLocalAppOptions, overwriteLocalConfigFileWithRemoteAppConfiguration} from './link.js'
 import {CurrentAppConfiguration, isCurrentAppSchema} from '../../../models/app/app.js'
 import {OrganizationApp} from '../../../models/organization.js'
-import {AppConfigurationFileName, getAppConfigurationFileName} from '../../../models/app/loader.js'
+import {AppConfigurationFileName} from '../../../models/app/loader.js'
 import {fetchSpecifications} from '../../generate/fetch-extension-specifications.js'
 import {RemoteAwareExtensionSpecification} from '../../../models/extensions/specification.js'
 import {Flag} from '../../../utilities/developer-platform-client.js'


anakafeel:
I have signed the CLA ✅

packages/app/src/cli/commands/app/config/pull.ts
```diff
@@ -0,0 +1,55 @@
...
+    const {flags} = await this.parse(ConfigPull)
+
+    // Run the pull service (no prompts)
+    const {configuration, remoteApp} = await pull({
+      directory: flags.path,
+      configName: flags.config,
+    })
+
+    // Get local app context so the return type matches other commands
+    const app = await localAppContext({
```
gonzaloriestra: I think it would be easier and more consistent if you use `linkedAppContext` here instead (see [this example](https://github.com/Shopify/cli/blob/b6967c9ba23a9eae105a3198f763eb3e2df34cfb/packages/app/src/cli/commands/app/generate/extension.ts#L80-L85)). Then, you can pass the `app` and `remoteApp` to the pull service, and it will allow you to remove the first 3 steps there.

With that change, the command would extend from `AppLinkedCommand`.

anakafeel: Thanks @gonzaloriestra !

I’ve refactored config pull to use `linkedAppContext` and updated it to extend `AppLinkedCommand`, passing `app` and `remoteApp` directly into the pull service as suggested.

I also ran `pnpm refresh-manifests` and `pnpm knip` and cleaned up the unused exports.

Please let me know if this matches what you had in mind :)

image
image

anakafeel:
Hey @gonzaloriestra 

I’ve completed the refactor based on your suggestion:

- `app config pull` now uses `linkedAppContext` and extends `AppLinkedCommand`, aligning with the pattern from app generate extension.

- The pull service now receives configuration and `remoteApp` from the command and no longer performs duplicated app resolution internally.

- Ran `pnpm refresh-manifests` and `pnpm knip`, and cleaned up the unused exports.

- Confirmed locally that `pnpm shopify app config pull` works as expected (no prompts, pulls latest config).

Everything should now be ready for another review, thanks again for the guidance :)

packages/app/src/cli/commands/app/config/pull.ts
```diff
@@ -0,0 +1,57 @@
...
+        remoteApp.title
+      }".`,
+      nextSteps: [
+        [
+          'To deploy your updated configuration, run',
+          {
+            command: formatPackageManagerCommand(app.packageManager, 'shopify app deploy'),
+          },
+        ],
+      ],
```
gonzaloriestra: We can remove this mesage. We are fetching the remote config, so a deploy wouldn't change anything.

```suggestion
```

packages/app/src/cli/commands/app/config/pull.ts
```diff
@@ -0,0 +1,57 @@
...
+      configName: flags.config,
+      configuration: app.configuration,
+      remoteApp,
+    })
+
+    renderSuccess({
+      headline: `Pulled latest configuration for "${configuration.name}"`,
+      body: `Updated ${configuration.path ?? flags.config ?? 'app configuration'} using the already-linked app "${
+        remoteApp.title
+      }".`,
```
gonzaloriestra: We usually show only the name instead of the full path. Also, at this point, I think the configuration should always have a path, which is the updated one.

```suggestion
      body: `Updated ${basename(configuration.path)} with the remote data".`,
```

packages/app/src/cli/services/app/config/pull.ts
```diff
@@ -0,0 +1,76 @@
...
+  }
+
+  const localAppOptions = await loadLocalAppOptions(linkOptions, specifications, flags, remoteApp.apiKey)
+
+  // Decide which config file to overwrite:
+  // - if config has a path, reuse that file
+  // - otherwise, fallback to --config or default app config name
+  const configFileName: AppConfigurationFileName =
+    (configuration.path && (basename(configuration.path) as AppConfigurationFileName)) ||
+    getAppConfigurationFileName(configName ?? configurationFileNames.app)
```
gonzaloriestra: Same here, I think the configuration path should always exist, so we can simplify:

```suggestion
  const configFileName: AppConfigurationFileName = basename(configuration.path) as AppConfigurationFileName
```

packages/app/src/cli/commands/app/config/pull.ts
anakafeel: Fixed as suggested

packages/app/src/cli/services/app/config/pull.ts
anakafeel: Done the same here

packages/app/src/cli/commands/app/config/pull.ts
anakafeel: Fixed as suggested

anakafeel:
@gonzaloriestra 
Thanks for the review!
I’ve applied all the suggestions:
– Updated success message
– Removed deploy next step
– Simplified config file name logic
– Rebuilt dev docs
Let me know if you’d like any further adjustments or if this is ready for merge :)

anakafeel:
@gonzaloriestra 
Oh my bad ! i completely forgot to run the linter to check for unused imports.
i've removed the unused imports now and everything should be good for a merge now :)

PR #6663, opened 2025-11-24
Add stdin support for `shopify app execute`

ericlee878:
Resolves: https://github.com/orgs/shop/projects/208/views/34?pane=issue&itemId=139444955&issue=shop%7Cissues-api-foundations%7C1111

Inspired by: https://app.graphite.com/github/pr/Shopify/cli/6588/Initial-Setup-Bulk-Operations-Infrastructure-for-Shopify-CLI#comment-PRRC_kwDOHibhHs6WnBvj

### Summary
This PR enables users to pipe GraphQL queries into `shopify app execute` via stdin, improving the developer experience for bulk operations.

### Why This Change Matters
**Before**: Users had to pass queries inline via --query, which is cumbersome for complex multi-line GraphQL:
```
shopify app execute --query 'query { products(first: 100) { edges { node { id title variants(first: 10) { edges { node { id price } } } } } } }'
```
**After**: Users can pipe from files or other commands:
```
cat my-query.graphql | shopify app execute
```
This aligns with Unix conventions and how developers already work with GraphQL tooling.

### Design Decisions
1. Stdin utilities in `cli-kit/system.ts`
   - Placing `isStdinPiped()` and `readStdin()` in `system.ts` alongside other process utilities makes them discoverable and reusable for future commands.
2. Early return when TTY
    -  `readStdin()` returns undefined immediately when stdin is interactive. This prevents the command from hanging waiting for user input — a common pitfall in CLI tools.

### Thoughts for Future
- Consider adding to command help text — The flag description mentions stdin, but --help output could include an example.

### DIFF
--- a/packages/cli-kit/src/public/node/system.ts
+++ b/packages/cli-kit/src/public/node/system.ts
@@ -9,6 +9,7 @@ import {shouldDisplayColors, outputDebug} from '../../public/node/output.js'
 import {execa, ExecaChildProcess} from 'execa'
 import which from 'which'
 import {delimiter} from 'pathe'
+import {fstatSync} from 'fs'
 import type {Writable, Readable} from 'stream'
 
 export interface ExecOptions {
@@ -199,4 +200,44 @@ export function isCI(): boolean {
 export async function isWsl(): Promise<boolean> {
   const wsl = await import('is-wsl')
   return wsl.default
+}
+
+/**
+ * Check if stdin has piped data available.
+ * This distinguishes between actual piped input (e.g., `echo "query" | cmd`)
+ * and non-TTY environments without input (e.g., CI).
+ *
+ * @returns True if stdin is receiving piped data or file redirect, false otherwise.
+ */
+export function isStdinPiped(): boolean {
+  try {
+    const stats = fstatSync(0)
+    return stats.isFIFO() || stats.isFile()
+    // eslint-disable-next-line no-catch-all/no-catch-all
+  } catch {
+    return false
+  }
+}
+
+/**
+ * Reads all data from stdin and returns it as a string.
+ * This is useful for commands that accept input via piping.
+ *
+ * @example
+ * // Usage: echo "your query" | shopify app execute
+ * const query = await readStdin()
+ *
+ * @returns A promise that resolves with the stdin content, or undefined if stdin is a TTY.
+ */
+export async function readStdin(): Promise<string | undefined> {
+  if (!isStdinPiped()) {
+    return undefined
+  }
+
+  let data = ''
+  process.stdin.setEncoding('utf8')
+  for await (const chunk of process.stdin) {
+    data += String(chunk)
+  }
+  return data.trim()
 }
--- a/packages/app/src/cli/commands/app/execute.ts
+++ b/packages/app/src/cli/commands/app/execute.ts
@@ -4,6 +4,8 @@ import {linkedAppContext} from '../../services/app-context.js'
 import {storeContext} from '../../services/store-context.js'
 import {executeBulkOperation} from '../../services/bulk-operations/execute-bulk-operation.js'
 import {globalFlags} from '@shopify/cli-kit/node/cli'
+import {readStdin} from '@shopify/cli-kit/node/system'
+import {AbortError} from '@shopify/cli-kit/node/error'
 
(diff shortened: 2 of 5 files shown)


ericlee878:
* **#6666** <a href="https://app.graphite.com/github/pr/Shopify/cli/6666?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6663** <a href="https://app.graphite.com/github/pr/Shopify/cli/6663?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Shopify/cli/6663?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#6643** <a href="https://app.graphite.com/github/pr/Shopify/cli/6643?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#6667](https://github.com/Shopify/cli/pull/6667) <a href="https://app.graphite.com/github/pr/Shopify/cli/6667?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#6596** <a href="https://app.graphite.com/github/pr/Shopify/cli/6596?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

packages/cli-kit/src/public/node/system.ts
```diff
@@ -200,3 +200,36 @@ export async function isWsl(): Promise<boolean> {
...
+
+/**
+ * Check if stdin has piped data available (not a TTY).
+ * This is useful for detecting when a command is receiving input via piping.
+ *
+ * @returns True if stdin is being piped, false if it's interactive.
+ */
+export function isStdinPiped(): boolean {
+  return !process.stdin.isTTY
+}
```
isaacroldan, line 220: I think that `piped implies non TTY`, but `non TTY doesn't always mean piped` 🤔
I might be wrong, but CI environments are non TTY by default and that doesn't mean you are receiving an input via stdin

ericlee878: You are right! I changed it so that the function now checks what stdin actually is using fstatSync(0): isFIFO() for pipes, isFile() for file redirects.

packages/app/src/cli/services/bulk-operations/execute-bulk-operation.ts
```diff
@@ -8,19 +8,40 @@ import {renderSuccess, renderInfo, renderError, renderWarning} from '@shopify/cl
...
+/**
+ * Resolves the query from the provided flag or stdin.
+ * Follows the same pattern as parseVariablesToJsonl for consistency.
+ */
+async function resolveQuery(query?: string): Promise<string> {
+  if (query) {
+    return query
+  }
+
+  const stdinContent = await readStdin()
```
dmerand: I wonder if this is the right layer to read from STDIN? Perhaps the command layer e.g. `execute`​ where input flags etc. are being parsed would be the right moment to check whether we've been given an input text/file?

packages/cli-kit/src/public/node/system.ts
```diff
@@ -200,3 +201,43 @@ export async function isWsl(): Promise<boolean> {
...
+ * Reads all data from stdin and returns it as a string.
+ * This is useful for commands that accept input via piping.
+ *
+ * @example
+ * // Usage: echo "your query" | shopify app execute
+ * const query = await readStdin()
+ *
+ * @returns A promise that resolves with the stdin content, or undefined if stdin is a TTY.
+ */
+export async function readStdin(): Promise<string | undefined> {
```
dmerand, line 232: Nit: Since we're assuming string content from STDIN here with the encoding (rather than allowing arbitrary encodings), perhaps we should call this `readStdinString`​ or something similar.

packages/app/src/cli/services/bulk-operations/execute-bulk-operation.ts
ericlee878: Good call. I refactored it this way and made the code way more organized. Thanks!

packages/cli-kit/src/public/node/system.ts
ericlee878: Makes sense. Changed!

PR #6678, opened 2025-12-02
Guide users to use `bulk status` when an operation is still running in the background

jordanverasamy:

https://github.com/shop/issues-api-foundations/issues/1095

We want to make sure it's clear to users when things are still running in the background.

Let's helpfully print a nice message telling users when they can use the `bulk status` command, even if they CTRL+C out of an operation midway!

We did this by:

- implementing a new optional `onAbort` callback to `renderSingleTask`, which you can use to configure custom behaviour when the user quits mid-way through a single task
- teaching `executeBulkOperation` and `watchBulkOperation` to use that new `onAbort` callback to display a nice message when the user quits while watching a bulk operation

### How to test your changes?

Without `--watch`, you get a nice help message:

image

With `--watch`, if you abort using CTRL+C before it's done, you get a nice help message:

https://github.com/user-attachments/assets/457a47d3-e375-43d7-8e65-3ce9f5433869

With `--watch`, without aborting, same behaviour as before this PR:

image

### DIFF
--- a/packages/app/src/cli/services/bulk-operations/execute-bulk-operation.test.ts
+++ b/packages/app/src/cli/services/bulk-operations/execute-bulk-operation.test.ts
@@ -6,7 +6,7 @@ import {downloadBulkOperationResults} from './download-bulk-operation-results.js
 import {BulkOperationRunQueryMutation} from '../../api/graphql/bulk-operations/generated/bulk-operation-run-query.js'
 import {BulkOperationRunMutationMutation} from '../../api/graphql/bulk-operations/generated/bulk-operation-run-mutation.js'
 import {OrganizationApp} from '../../models/organization.js'
-import {renderSuccess, renderWarning, renderError} from '@shopify/cli-kit/node/ui'
+import {renderSuccess, renderWarning, renderError, renderInfo} from '@shopify/cli-kit/node/ui'
 import {ensureAuthenticatedAdminAsApp} from '@shopify/cli-kit/node/session'
 import {inTemporaryDirectory, writeFile} from '@shopify/cli-kit/node/fs'
 import {joinPath} from '@shopify/cli-kit/node/path'
@@ -355,12 +355,43 @@ describe('executeBulkOperation', () => {
       watch: true,
     })
 
-    expect(watchBulkOperation).toHaveBeenCalledWith(mockAdminSession, createdBulkOperation.id)
     expect(renderSuccess).toHaveBeenCalledWith(
       expect.objectContaining({
         headline: expect.stringContaining('Bulk operation succeeded:'),
       }),
     )
+  })
+
+  test('renders help message in an info banner when watch is provided and user aborts', async () => {
+    const query = '{ products { edges { node { id } } } }'
+    const initialResponse: BulkOperationRunQueryMutation['bulkOperationRunQuery'] = {
+      bulkOperation: createdBulkOperation,
+      userErrors: [],
+    }
+    const runningOperation = {
+      ...createdBulkOperation,
+      status: 'RUNNING' as const,
+      objectCount: '100',
+    }
+
+    vi.mocked(runBulkOperationQuery).mockResolvedValue(initialResponse)
+    vi.mocked(watchBulkOperation).mockImplementation(async (_session, _id, signal, onAbort) => {
+      onAbort()
+      return runningOperation
+    })
+
+    await executeBulkOperation({
+      remoteApp: mockRemoteApp,
+      storeFqdn,
+      query,
+      watch: true,
+    })
+
+    expect(renderInfo).toHaveBeenCalledWith({
+      headline: `Bulk operation ${createdBulkOperation.id} is still running in the background.`,
+      body: ['Monitor its progress with:', {command: expect.stringContaining('shopify app bulk status')}],
+    })
+    expect(downloadBulkOperationResults).not.toHaveBeenCalled()
   })
 
   test('writes results to file when --output-file flag is provided', async () => {
@@ -450,7 +481,6 @@ describe('executeBulkOperation', () => {
         watch: true,
       })
 
-      expect(watchBulkOperation).toHaveBeenCalledWith(mockAdminSession, createdBulkOperation.id)
       expect(renderError).toHaveBeenCalledWith(
         expect.objectContaining({
           headline: expect.any(String),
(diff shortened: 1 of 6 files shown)


jordanverasamy:
* **#6678** <a href="https://app.graphite.com/github/pr/Shopify/cli/6678?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Shopify/cli/6678?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/app/src/cli/services/bulk-operations/execute-bulk-operation.test.ts
```diff
@@ -355,14 +355,45 @@ describe('executeBulkOperation', () => {
       watch: true,
     })
 
-    expect(watchBulkOperation).toHaveBeenCalledWith(mockAdminSession, createdBulkOperation.id)
```
jordanverasamy, line 358: (I don't think this assertion was providing actual value, since the thing we actually care about is that the success banner was rendered; I opted to just remove it instead of fixing the parameters on the assertion.)

packages/app/src/cli/services/bulk-operations/execute-bulk-operation.test.ts
```diff
@@ -450,7 +481,6 @@ describe('executeBulkOperation', () => {
         watch: true,
       })
 
-      expect(watchBulkOperation).toHaveBeenCalledWith(mockAdminSession, createdBulkOperation.id)
```
jordanverasamy, line 453: (Same as https://github.com/Shopify/cli/pull/6678/files#r2587231786.)

packages/app/src/cli/services/bulk-operations/execute-bulk-operation.ts
```diff
@@ -76,8 +77,19 @@ export async function executeBulkOperation(input: ExecuteBulkOperationInput): Pr
...
     if (watch) {
-      const finishedOperation = await watchBulkOperation(adminSession, createdOperation.id)
-      await renderBulkOperationResult(finishedOperation, outputFile)
+      const abortController = new AbortController()
+      const operation = await watchBulkOperation(adminSession, createdOperation.id, abortController.signal, () =>
+        abortController.abort(),
+      )
+
+      if (abortController.signal.aborted) {
+        renderInfo({
```
jordanverasamy, line 86: Made this an info banner per [the brief](https://docs.google.com/document/d/157CvTbzxv5RDkAJCCeZMcWEPrsGLEdkFc8Hd48fUSM8/edit?tab=t.0#bookmark=id.sjyndives3h9), but I do think it looks a bit odd being all grey, maybe a success banner would more appropriately convey to the user "nothing went wrong, all is still good and running nicely". I'm not really sure. Let me know what you think is correct!

packages/cli-kit/src/private/node/ui/components/SingleTask.tsx
```diff
@@ -1,21 +1,33 @@
...
+
+  useInput(
+    (input, key) => {
+      if (onAbort) {
+        handleCtrlC(input, key, onAbort)
+      } else {
+        handleCtrlC(input, key)
+      }
+    },
+    {isActive: Boolean(isRawModeSupported)},
```
jordanverasamy, line 29: This part is basically saying: only activate this behaviour if we're in an interactive terminal.

If STDIN isn't a TTY, for example if input is being piped, we're running in CI, or it's otherwise non-interactive, then `isRawModeSupported` will be false and the `useInput` hook will helpfully not do anything.

packages/app/src/cli/services/bulk-operations/execute-bulk-operation.ts
```diff
@@ -147,6 +163,11 @@ function validateGraphQLDocument(graphqlOperation: string, variablesJsonl?: stri
   }
 }
 
+function statusCommandHelpMessage(operationId: string): string {
+  const command = outputToken.cyan(`shopify app bulk status --id="${operationId}"`)
+  return outputContent`Monitor its progress with:\n${command}`.value
```
gonzaloriestra: Instead of manually choosing a different color, there's a standard way to print commands, see [this example](https://github.com/Shopify/cli/blob/a59fc98bc06f2c0eef3398e5df74b308d136aa13/packages/app/src/cli/services/dev-clean.ts#L28-L35).

packages/app/src/cli/services/bulk-operations/execute-bulk-operation.ts
gonzaloriestra: I agree a success banner would be better. Actually, why not showing exactly the same message as when you run it without `--watch`?

```
╭─ success ─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│  Bulk operation started.                                                                                                                                  │
│  Monitor its progress with:                                                                                                                               │
│  shopify app bulk status --id="gid://shopify/BulkOperation/5749231780082"                                                                                 │
│    • ID: gid://shopify/BulkOperation/5749231780082                                                                                                        │
│    • Status: CREATED                                                                                                                                      │
│    • Created at: 2025-12-04T10:37:57Z                                                                                                                     │
```

jordanverasamy: Yeah, I considered this but decided it was a slightly different context so the most helpful possible messaging to the user is slightly different. I'll [log this for Nick to think about](https://github.com/shop/issues-api-foundations/issues/1149) though! TY!

jordanverasamy:
> When you abort, it seems that it waits for the current polling sleep (up to 5 seconds) before actually finishing the command:

That's very interesting, it shouldn't... I specifically used `Promise.race` to make sure we can get interrupted mid-sleep. I confirmed that I'm able to responsively and quickly exit mid-sleep on my machine. Wonder what's causing a difference between our environments 🤔 

> Also, maybe you are already aware, but it's showing a duplicated ellipsis in the Starting task:

Yeah, that extra ellipsis will get cleaned up, you pointed it out before so we've got it tracked by https://github.com/shop/issues-api-foundations/issues/1149. Thanks!!

packages/app/src/cli/services/bulk-operations/execute-bulk-operation.ts
jordanverasamy: Wow, that's so nice, didn't know that! Adjusted, TY :)

packages/app/src/cli/services/bulk-operations/execute-bulk-operation.ts
nickwesselman: Since this is the Ctrl+C banner, I think info is correct, since it’s just informing you that the query is still running.

PR #6684, opened 2025-12-04
introduce `--version` flag to Bulk Operations CLI

jordanverasamy:
### WHY are these changes introduced?

https://github.com/shop/issues-api-foundations/issues/1138

Users should be able to run their bulk operations against any valid and supported API version.

### WHAT is this pull request doing?

Allows users to run their bulk operations against a configurable version. If not provided, defaults to latest stable version.

### How to test your changes?

Try running `shopify app bulk execute` with various values of `--version`.

Invalid version, e.g. `--version=1998-01`:

image

Valid stable version, e.g. `--version=2025-04`:

image

`--version=unstable`:

image

`--version` not provided at all:

image

PR #6686, opened 2025-12-04
Add `app execute` command for GraphQL operations

nickwesselman:

To provide developers with a convenient way to execute GraphQL queries and mutations directly from the CLI, making it easier to interact with Shopify's Admin API.

Adds a new `shopify app execute` command that allows developers to:

- Execute GraphQL queries and mutations against a store
- Provide variables in JSON format
- Output results to STDOUT or a file
- Specify API version

This PR also:

- Refactors shared code between the new `execute` command and existing `bulk execute` command
- Creates helper utilities for preparing execution context
- Adds documentation for the new command

### How to test your changes?

1. Run a simple query:
2. Run a query with variables:
3. Pipe a query via stdin:
4. Output to a file:
5. Verify correct output streams:

`echo "query { shop { name } }" | shopify app execute 2>/dev/null | jq .shop.name`

### Measuring impact

### Checklist

### DIFF
--- a/packages/app/src/cli/utilities/execute-command-helpers.ts
+++ b/packages/app/src/cli/utilities/execute-command-helpers.ts
@@ -65,7 +65,7 @@ export async function prepareExecuteContext(
   if (!query) {
     throw new AbortError(
       'No query provided. Use the --query flag or pipe input via stdin.',
-      `Example: echo "query { shop { name } }" | shopify app ${commandName}`,
+      `Example: echo "query { ... }" | shopify app ${commandName}`,
     )
   }
 


nickwesselman:
* **#6689** <a href="https://app.graphite.com/github/pr/Shopify/cli/6689?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6686** <a href="https://app.graphite.com/github/pr/Shopify/cli/6686?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Shopify/cli/6686?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/app/src/cli/flags.ts
```diff
@@ -73,3 +73,32 @@ export const bulkOperationFlags = {
...
 }
+
+export const operationFlags = {
+  query: Flags.string({
+    char: 'q',
+    description: 'The GraphQL query or mutation, as a string.',
+    env: 'SHOPIFY_FLAG_QUERY',
+    required: false,
+  }),
+  variables: Flags.string({
```
ericlee878, line 88: Maybe we could rename this to `variable` since operations that aren't bulk operations should always have just 1 variable? Same for the description.

ericlee878: Actually, not too sure if this is true. @jordanverasamy could you verify?

packages/app/src/cli/services/execute-operation.ts
```diff
@@ -0,0 +1,94 @@
...
+interface ExecuteOperationInput {
+  remoteApp: OrganizationApp
+  storeFqdn: string
+  query: string
+  variables?: string
+  apiVersion?: string
+  outputFile?: string
+}
+
+async function parseVariables(variables?: string): Promise<{[key: string]: unknown} | undefined> {
```
ericlee878, line 20: Should we check if there is more than one variable here? Like we suggested here. And return an error if there is more than one variable.

ericlee878:
Thanks for the helper functions. The code seems much cleaner now!

packages/app/src/cli/services/bulk-operations/execute-bulk-operation.test.ts
```diff
@@ -190,57 +190,6 @@ describe('executeBulkOperation', () => {
     expect(renderSuccess).not.toHaveBeenCalled()
   })
 
-  test('throws GraphQL syntax error when given malformed GraphQL document', async () => {
```
ericlee878, line 200: Why were these tests removed?

packages/app/src/cli/flags.ts
jordanverasamy: Good thinking to ask! But no, `variables` is the appropriate name here.

Let's look at the example variables in our documentation for `productCreate` ([link](https://shopify.dev/docs/api/admin-graphql/latest/mutations/productCreate)):

```json
{
   "title":"Cool socks",
   "productOptions":[
      {
         "name":"Color",
         "values":[
            {
               "name":"Red"
            },
            {
               "name":"Blue"
            }
         ]
      },
      {
         "name":"Size",
         "values":[
            {
               "name":"Small"
            },
            {
               "name":"Large"
            }
         ]
      }
   ]
}
```

There are multiple different keys in there (e.g. `title` and `productOptions`, so we call that JSON object "variables".

If you ran this mutation in bulk, you'd get many of these objects instead of just one of them. But even if you just look at one, there are multiple variables inside of it. Make sense?

packages/app/src/cli/services/execute-operation.ts
jordanverasamy: I think it's ok to just pass whatever we're given along to the API. We had all that fancy handling sketched out because we thought we wanted to smartly handle both cases in one command -- given that we aren't doing that, this command can just be "dumber" and not do anything sophisticated.

packages/app/src/cli/services/bulk-operations/execute-bulk-operation.test.ts
jordanverasamy: good q, wondering the same

packages/app/src/cli/services/bulk-operations/execute-bulk-operation.test.ts
```diff
@@ -506,7 +455,7 @@ describe('executeBulkOperation', () => {
     ).rejects.toThrow('Bulk operation response returned null with no error message.')
 
     expect(renderWarning).toHaveBeenCalledWith({
-      headline: 'Bulk operation not created succesfully.',
+      headline: 'Bulk operation not created successfully.',
```
jordanverasamy, line 466: 🙏🏾

packages/app/src/cli/flags.ts
ericlee878: I see. So, there could be multiple different types of variables, instead of multiple versions of the same variable (if that makes sense). Thanks for the explanation!

packages/app/src/cli/services/execute-operation.ts
ericlee878: Makes sense! Thanks for the explanation!

packages/app/src/cli/services/dev/fetch.test.ts
```diff
@@ -121,7 +121,11 @@ describe('fetchStore', () => {
     const got = fetchStore(ORG1, 'domain1', developerPlatformClient)
 
     // Then
-    await expect(got).rejects.toThrow(new AbortError(`Could not find Store for domain domain1 in Organization org1.`))
+    await expect(got).rejects.toThrow(
+      new AbortError(
+        `Could not find store for domain domain1 in organization org1. Ensure you've provided the correct store domain, that the store is a Dev Store, and that you have access to the store.`,
```
gonzaloriestra:
```suggestion
        `Could not find store for domain domain1 in organization org1.`,
        `Ensure you've provided the correct store domain, that the store is a Dev Store, and that you have access to the store.`,
```

packages/app/src/cli/services/dev/fetch.ts
```diff
@@ -125,7 +125,10 @@ export async function fetchStore(
 ): Promise<OrganizationStore> {
   const store = await developerPlatformClient.storeByDomain(org.id, storeFqdn)
 
-  if (!store) throw new AbortError(`Could not find Store for domain ${storeFqdn} in Organization ${org.businessName}.`)
+  if (!store)
+    throw new AbortError(
+      `Could not find store for domain ${storeFqdn} in organization ${org.businessName}. Ensure you've provided the correct store domain, that the store is a Dev Store, and that you have access to the store.`,
```
gonzaloriestra:
```suggestion
      `Could not find store for domain ${storeFqdn} in organization ${org.businessName}.`,
      `Ensure you've provided the correct store domain, that the store is a Dev Store, and that you have access to the store.`,
```

gonzaloriestra: The second argument is the "try message" or "next steps"

packages/app/src/cli/services/execute-operation.ts
```diff
@@ -0,0 +1,94 @@
...
+      await writeFile(outputFile, resultString)
+      renderSuccess({
+        headline: 'Operation completed successfully.',
+        body: `Results written to ${outputFile}`,
+      })
+    } else {
+      renderSuccess({
+        headline: 'Operation completed successfully.',
+      })
+      outputResult(resultString)
```
gonzaloriestra, line 87: The bulk command with `--watch` is printing the result before the success message. Also, it shows `operation succeeded` instead. What about being more consistent between both?

I wonder if we should better include the results inside the success block. @jenstanicak?

Current bulk output:



Current single output:

packages/app/src/cli/utilities/execute-command-helpers.ts
```diff
@@ -0,0 +1,75 @@
...
+ */
+export async function prepareExecuteContext(
+  flags: ExecuteCommandFlags,
+  commandName = 'execute',
+): Promise<ExecuteContext> {
+  const query = flags.query ?? (await readStdinString())
+  if (!query) {
+    throw new AbortError(
+      'No query provided. Use the --query flag or pipe input via stdin.',
+      `Example: echo "query { shop { name } }" | shopify app ${commandName}`,
```
gonzaloriestra: Is the bulk command allowing stdin input? Both commands should allow the flag in the same ways and we should show the same option in the examples.

packages/app/src/cli/services/bulk-operations/execute-bulk-operation.test.ts
nickwesselman: This was 100% Claude!

packages/app/src/cli/services/bulk-operations/execute-bulk-operation.test.ts
nickwesselman: These should be covered by [tests on the helpers](https://app.graphite.com/github/pr/Shopify/cli/6686/Add-%60app-execute%60-command-for-GraphQL-operations#lines-packages/app/src/cli/services/graphql/common.test.ts-R53-R53) now, unless y'all would like this to be verified at this layer as well? But it's no longer a responsibility of `executeBulkOperation` itself.

packages/app/src/cli/flags.ts
nickwesselman: Yeah the difference is that with `bulk execute`, you can have multiple sets of `variables`!

packages/app/src/cli/services/execute-operation.ts
nickwesselman: We want the results outside the block so they can be piped/redirected. @ericlee878 @jordanverasamy what do you think is the better approach here? For `app execute` I moved the success message before the result, since it's the result itself that you are going to care about reading.  
  
I'm good with `Operation succeeded` here to be consistent with the message itself.

packages/app/src/cli/utilities/execute-command-helpers.ts
nickwesselman: Yep I extracted this helper from `bulk execute` logic, it's now shared by both `app execute` and `app bulk execute`

packages/app/src/cli/services/execute-operation.ts
gonzaloriestra: Ah, makes sense!

packages/app/src/cli/services/bulk-operations/execute-bulk-operation.test.ts
jordanverasamy: mmm, got it, you right! thanks! my bad not putting that together 😅

packages/app/src/cli/services/dev/fetch.test.ts
jordanverasamy: Is `Dev Store` the correct/accepted stylization rather than `dev store` or `development store`?

packages/app/src/cli/services/execute-operation.ts
jordanverasamy: Yeah, outside of the block is correct since (AFAIK) the results are being printed to STDOUT whereas UI info etc is all going through STDERR.

I have no strong thoughts about whether the results should be printed before or after the success banner. (I agree they should be consistent one way or the other though!) Go with your gut :)

packages/app/src/cli/services/dev/fetch.test.ts
nickwesselman: Actually it should be `dev store` and not `Dev Store`:  
https://shopify.dev/docs/apps/build/dev-dashboard/development-stores  
  
I'll update :+1:

packages/app/src/cli/services/execute-operation.ts
nickwesselman: I will update `bulk execute` to render the banner first for consistency :+1:

packages/app/src/cli/services/execute-operation.ts
```diff
@@ -0,0 +1,94 @@
...
+import {ClientError} from 'graphql-request'
+import {parse} from 'graphql'
+import {writeFile} from '@shopify/cli-kit/node/fs'
+
+interface ExecuteOperationInput {
+  remoteApp: OrganizationApp
+  storeFqdn: string
+  query: string
+  variables?: string
+  apiVersion?: string
```
jordanverasamy: [nit] possibly this should just be `version` for consistency with https://github.com/Shopify/cli/pull/6684? sorry for shipping it and making you have to rebase 😭 if you want to just ship things as-is with the lowest-effort rebase and then we unify everything in a followup, that's ok with me

packages/app/src/cli/services/execute-operation.ts
```diff
@@ -0,0 +1,94 @@
...
+async function parseVariables(variables?: string): Promise<{[key: string]: unknown} | undefined> {
+  if (!variables) return undefined
+
+  try {
+    return JSON.parse(variables)
+  } catch (error) {
+    const errorMessage = error instanceof Error ? error.message : 'Unknown error'
+    throw new AbortError(
+      outputContent`Invalid JSON in ${outputToken.yellow('--variables')} flag: ${errorMessage}`,
+      'Please provide valid JSON format.',
```
jordanverasamy, line 29: Man, I've never been a fan of my tools/docs saying "please" (I think they should just be matter-of-fact) but I do see a lot of occurrences of "please" in the codebase so I guess this is the correct pattern-matching choice 🙈

packages/app/src/cli/services/execute-operation.ts
```diff
@@ -0,0 +1,94 @@
...
+    )
+  }
+}
+
+export async function executeOperation(input: ExecuteOperationInput): Promise<void> {
+  const {remoteApp, storeFqdn, query, variables, apiVersion, outputFile} = input
+
+  renderInfo({
```
jordanverasamy, line 37: [nit] Now that version is configurable I think it's nice to print it so the user knows what's being executed, can reuse my code from here: https://github.com/Shopify/cli/pull/6684/files#diff-6040bd49d6a47d1cfe82752ae20e0e5502ce7764dcb7ec1b38fa904eea5bbc6fR58-R71

Happy with that being a followup though if makes your life easier :)

jordanverasamy (approved):
left some nits/comments but nothing blocking, this looks fantastic, thanks for putting it together :)

packages/app/src/cli/services/execute-operation.ts
nickwesselman: Yeah I think generally we optimize for kindness :grinning:   
  
Please try again :laughing:

packages/app/src/cli/services/execute-operation.ts
nickwesselman: Not afraid of a rebase... I'll update my version handling to be consistent. You have the validation there too that I can move into helpers and reuse.

packages/app/src/cli/services/execute-operation.ts
nickwesselman: Made this consistent too!

ericlee878 (approved):

nickwesselman:
I did some extensive e2e testing after these last changes, I think we are in good shape!

PR #6687, opened 2025-12-04
Quick-watch: Catch early bulk operation errors without `--watch`

ericlee878:
Resolves: https://github.com/orgs/shop/projects/208/views/34?filterQuery=&pane=issue&itemId=140129246&issue=shop%7Cissues-api-foundations%7C1123

Inspired by: https://github.com/shop/issues-api-foundations/issues/1123#issuecomment-3603641001

### Background
When users run bulk operations with `--variables` or `--variable-file`, invalid JSON or malformed inputs aren't validated upfront. Without the `--watch` flag, users would:
1. Submit a potentially invalid job
2. Walk away thinking it's running
3. Come back hours later to find errors in the result file

This is a frustrating experience, especially for long-running bulk operations. 

We considered validating JSON on the client before sending to the API, but:
- There are too many ways for Core to reject input (incorrect query, incorrect variables, incorrect data format, etc.). There could also be errors we cannot catch from client side, making it possible for there to sometimes be errors caught on the client side and sometimes given in the output file (potentially causing confusion).
- It's impractical to catch every type of error at a validation stage.
- Full validation of large imports would hurt performance.

**The solution: Quick-watch**
Instead of duplicating validation logic, we leverage the fact that errors tend to surface in the first 2-3 seconds of a bulk operation. By polling briefly after creation, we can catch most "obvious issues" (like malformed JSON) and present them immediately—without requiring users to run a second command or wait for hours.

This approach:
- Catches early errors without duplicating Core's validation logic
- Doesn't significantly impact performance (only 3 seconds of polling)
- Works for the common case where validation errors appear quickly

### Implementation Details
**Detecting mutation-level errors in results**
When a bulk operation completes `(status: COMPLETED)`, we check the downloaded results for `userErrors` using a simple string check: `results.includes('"userErrors":[{').`

This catches cases where the operation "succeeds" but individual mutations failed (e.g., malformed JSON variables). Instead of showing a misleading "Bulk operation succeeded" message, we show:
- `renderWarning` with "Bulk operation completed with errors."
- If `--output-file` was provided: includes the file path so users know where to check

**Why `.includes()` instead of `?.userErrors?.length:` like in line 62 of `execute-bulk-operation.ts`?**
The result file is a raw JSONL string, not a JavaScript object. Parsing every line would be expensive for large files — a simple string search is O(n) with no allocations. Checking for "userErrors":[{ only matches non-empty arrays ("userErrors":[] won't match).

**Flag behavior**
Quick watch only happens when the ``--watch`` flag isn't given. This decision is explained [here](https://shopify.slack.com/archives/C08RN29AQLX/p1764874226066419?thread_ts=1764796222.127799&cid=C08RN29AQLX).

**Polling configuration**

``QUICK_WATCH_TIMEOUT_MS = 3000`` - Total time to poll
``QUICK_WATCH_POLL_INTERVAL_MS = 300`` - Poll every 300ms (~10 polls max)

**No null checks in ``quickWatchBulkOperation``:**
We use ``bulkOperation!`` (non-null assertion) because:
- ``executeBulkOperation`` already checks ``userErrors`` before calling quick-watch
- If ``bulkOperation`` is null, ``userErrors`` would exist and we'd return early

### Tophatting
**For catching errors early:**
```
pnpm shopify:run app bulk execute \
  --path <PATH_TO_TEST_APP> \
  --output-file ./results.jsonl \
  -q 'mutation productUpdate($input: ProductInput!) { productUpdate(input: $input) { product { id tags } userErrors { field message } } }' \
  -v '{"input": invalid json}'
```

This is what my `results.jsonl` looked like:
```
{"data":{"productUpdate":{"userErrors":[{"message":"unexpected character: 'invalid' at line 1 column 11"}]}},"__lineNumber":0}
```

**For super quickly finished bulk operations that successfully finish within the quick-watch:**
```
pnpm shopify:run app bulk execute \
  --path <PATH_TO_TEST_APP>\
  -q '{ products(first: 1) { edges { node { id title } } } }'
```

**For bulk operations that do not finish in quick-watch:**

Create a variable file:

```
cat > /tmp/bulk-products.jsonl << 'EOF'
{"input":{"title":"Test Product 1","descriptionHtml":"<p>This is test product 1</p>","productType":"Test Type","vendor":"Test Vendor","tags":["bulk","test"],"metafields":[{"namespace":"custom","key":"field1","value":"value1","type":"single_line_text_field"},{"namespace":"custom","key":"field2","value":"value2","type":"single_line_text_field"},{"namespace":"custom","key":"field3","value":"value3","type":"single_line_text_field"}]}}
{"input":{"title":"Test Product 2","descriptionHtml":"<p>This is test product 2</p>","productType":"Test Type","vendor":"Test Vendor","tags":["bulk","test"],"metafields":[{"namespace":"custom","key":"field1","value":"value1","type":"single_line_text_field"},{"namespace":"custom","key":"field2","value":"value2","type":"single_line_text_field"},{"namespace":"custom","key":"field3","value":"value3","type":"single_line_text_field"}]}}
{"input":{"title":"Test Product 3","descriptionHtml":"<p>This is test product 3</p>","productType":"Test Type","vendor":"Test Vendor","tags":["bulk","test"],"metafields":[{"namespace":"custom","key":"field1","value":"value1","type":"single_line_text_field"},{"namespace":"custom","key":"field2","value":"value2","type":"single_line_text_field"},{"namespace":"custom","key":"field3","value":"value3","type":"single_line_text_field"}]}}
{"input":{"title":"Test Product 4","descriptionHtml":"<p>This is test product 4</p>","productType":"Test Type","vendor":"Test Vendor","tags":["bulk","test"],"metafields":[{"namespace":"custom","key":"field1","value":"value1","type":"single_line_text_field"},{"namespace":"custom","key":"field2","value":"value2","type":"single_line_text_field"},{"namespace":"custom","key":"field3","value":"value3","type":"single_line_text_field"}]}}
{"input":{"title":"Test Product 5","descriptionHtml":"<p>This is test product 5</p>","productType":"Test Type","vendor":"Test Vendor","tags":["bulk","test"],"metafields":[{"namespace":"custom","key":"field1","value":"value1","type":"single_line_text_field"},{"namespace":"custom","key":"field2","value":"value2","type":"single_line_text_field"},{"namespace":"custom","key":"field3","value":"value3","type":"single_line_text_field"}]}}
{"input":{"title":"Test Product 6","descriptionHtml":"<p>This is test product 6</p>","productType":"Test Type","vendor":"Test Vendor","tags":["bulk","test"],"metafields":[{"namespace":"custom","key":"field1","value":"value1","type":"single_line_text_field"},{"namespace":"custom","key":"field2","value":"value2","type":"single_line_text_field"},{"namespace":"custom","key":"field3","value":"value3","type":"single_line_text_field"}]}}
{"input":{"title":"Test Product 7","descriptionHtml":"<p>This is test product 7</p>","productType":"Test Type","vendor":"Test Vendor","tags":["bulk","test"],"metafields":[{"namespace":"custom","key":"field1","value":"value1","type":"single_line_text_field"},{"namespace":"custom","key":"field2","value":"value2","type":"single_line_text_field"},{"namespace":"custom","key":"field3","value":"value3","type":"single_line_text_field"}]}}
{"input":{"title":"Test Product 8","descriptionHtml":"<p>This is test product 8</p>","productType":"Test Type","vendor":"Test Vendor","tags":["bulk","test"],"metafields":[{"namespace":"custom","key":"field1","value":"value1","type":"single_line_text_field"},{"namespace":"custom","key":"field2","value":"value2","type":"single_line_text_field"},{"namespace":"custom","key":"field3","value":"value3","type":"single_line_text_field"}]}}
{"input":{"title":"Test Product 9","descriptionHtml":"<p>This is test product 9</p>","productType":"Test Type","vendor":"Test Vendor","tags":["bulk","test"],"metafields":[{"namespace":"custom","key":"field1","value":"value1","type":"single_line_text_field"},{"namespace":"custom","key":"field2","value":"value2","type":"single_line_text_field"},{"namespace":"custom","key":"field3","value":"value3","type":"single_line_text_field"}]}}
{"input":{"title":"Test Product 10","descriptionHtml":"<p>This is test product 10</p>","productType":"Test Type","vendor":"Test Vendor","tags":["bulk","test"],"metafields":[{"namespace":"custom","key":"field1","value":"value1","type":"single_line_text_field"},{"namespace":"custom","key":"field2","value":"value2","type":"single_line_text_field"},{"namespace":"custom","key":"field3","value":"value3","type":"single_line_text_field"}]}}
EOF
```

Run the long bulk mutation:
```
pnpm shopify:run app bulk execute \
  --path ./coffee-and-more \
  --variable-file /tmp/bulk-products.jsonl \
  -q 'mutation productCreate($input: ProductInput!) {
      product {
        id
        title
        metafields(first: 10) {
          edges {
            node {
              id
              namespace
              key
              value
            }
          }
        }
      }
      userErrors {
        field
        message
      }
    }
  }'
```

### DIFF
--- a/packages/app/src/cli/services/bulk-operations/execute-bulk-operation.ts
+++ b/packages/app/src/cli/services/bulk-operations/execute-bulk-operation.ts
@@ -1,6 +1,6 @@
 import {runBulkOperationQuery} from './run-query.js'
 import {runBulkOperationMutation} from './run-mutation.js'
-import {watchBulkOperation, type BulkOperation} from './watch-bulk-operation.js'
+import {watchBulkOperation, shortBulkOperationPoll, type BulkOperation} from './watch-bulk-operation.js'
 import {formatBulkOperationStatus} from './format-bulk-operation-status.js'
 import {downloadBulkOperationResults} from './download-bulk-operation-results.js'
 import {extractBulkOperationId} from './bulk-operation-status.js'
@@ -104,7 +104,8 @@ export async function executeBulkOperation(input: ExecuteBulkOperationInput): Pr
         await renderBulkOperationResult(operation, outputFile)
       }
     } else {
-      await renderBulkOperationResult(createdOperation, outputFile)
+      const operation = await shortBulkOperationPoll(adminSession, createdOperation.id)
+      await renderBulkOperationResult(operation, outputFile)
     }
   } else {
     renderWarning({
@@ -136,16 +137,38 @@ async function renderBulkOperationResult(operation: BulkOperation, outputFile?: 
         customSections,
       })
       break
+    case 'RUNNING':
+      renderSuccess({
+        headline: 'Bulk operation is running.',
+        body: statusCommandHelpMessage(operation.id),
+        customSections,
+      })
+      break
     case 'COMPLETED':
       if (operation.url) {
         const results = await downloadBulkOperationResults(operation.url)
+        const hasUserErrors = resultsContainUserErrors(results)
 
         if (outputFile) {
           await writeFile(outputFile, results)
-          renderSuccess({headline, body: [`Results written to ${outputFile}`], customSections})
         } else {
-          renderSuccess({headline, customSections})
           outputResult(results)
+        }
+
+        if (hasUserErrors) {
+          renderWarning({
+            headline: 'Bulk operation completed with errors.',
+            body: outputFile
+              ? `Results written to ${outputFile}. Check file for error details.`
+              : 'Check results for error details.',
+            customSections,
+          })
+        } else {
+          renderSuccess({
+            headline,
+            body: outputFile ? [`Results written to ${outputFile}`] : undefined,
+            customSections,
+          })
         }
       } else {
         renderSuccess({headline, customSections})
@@ -155,6 +178,17 @@ async function renderBulkOperationResult(operation: BulkOperation, outputFile?: 
       renderError({headline, customSections})
       break
   }
+}
+
+function resultsContainUserErrors(results: string): boolean {
+  const lines = results.trim().split('\n')
+
+  return lines.some((line) => {
+    const parsed = JSON.parse(line)
+    if (!parsed.data) return false
+    const result = Object.values(parsed.data)[0] as {userErrors?: unknown[]} | undefined
+    return result?.userErrors !== undefined && result.userErrors.length > 0
+  })
 }
 
 function validateGraphQLDocument(graphqlOperation: string, variablesJsonl?: string): void {
--- a/packages/app/src/cli/services/bulk-operations/watch-bulk-operation.ts
+++ b/packages/app/src/cli/services/bulk-operations/watch-bulk-operation.ts
@@ -16,7 +16,30 @@ const REGULAR_POLL_INTERVAL_SECONDS = 5
 const INITIAL_POLL_COUNT = 10
 const API_VERSION = '2026-01'
 
+export const QUICK_WATCH_TIMEOUT_MS = 3000
+export const QUICK_WATCH_POLL_INTERVAL_MS = 300
+
 export type BulkOperation = NonNullable<GetBulkOperationByIdQuery['bulkOperation']>
+
+export async function shortBulkOperationPoll(adminSession: AdminSession, operationId: string): Promise<BulkOperation> {
+  const startTime = Date.now()
+  const poller = pollBulkOperation({
+    adminSession,
+    operationId,
+    pollIntervalSeconds: QUICK_WATCH_POLL_INTERVAL_MS / 1000,
+  })
+
+  let latestOperationState: BulkOperation | undefined
+
+  do {
+    // eslint-disable-next-line no-await-in-loop
+    const {value, done} = await poller.next()
+    latestOperationState = value
+    if (done) return latestOperationState
+  } while (Date.now() - startTime < QUICK_WATCH_TIMEOUT_MS)
+
+  return latestOperationState
+}
 
 export async function watchBulkOperation(
   adminSession: AdminSession,
@@ -27,7 +50,15 @@ export async function watchBulkOperation(
   return renderSingleTask<BulkOperation>({
     title: outputContent`Polling bulk operation...`,
     task: async (updateStatus) => {
-      const poller = pollBulkOperation(adminSession, operationId, abortSignal)
+      const poller = pollBulkOperation({
+        adminSession,
+        operationId,
+        pollIntervalSeconds: REGULAR_POLL_INTERVAL_SECONDS,
+        initialPollIntervalSeconds: INITIAL_POLL_INTERVAL_SECONDS,
+        initialPollCount: INITIAL_POLL_COUNT,
+        useAdaptivePolling: true,
+        abortSignal,
+      })
 
       while (true) {
         // eslint-disable-next-line no-await-in-loop
@@ -44,11 +75,28 @@ export async function watchBulkOperation(
   })
 }
 
-async function* pollBulkOperation(
-  adminSession: AdminSession,
-  operationId: string,
-  abortSignal: AbortSignal,
-): AsyncGenerator<BulkOperation, BulkOperation> {
+interface PollBulkOperationOptions {
+  adminSession: AdminSession
+  operationId: string
+  pollIntervalSeconds: number
+  /** When true, polls faster initially then slows to pollIntervalSeconds */
+  useAdaptivePolling?: boolean
+  /** Poll interval in seconds for initial fast polls (default: 1) */
+  initialPollIntervalSeconds?: number
+  /** Number of fast polls before switching to regular interval (default: 10) */
+  initialPollCount?: number
+  abortSignal?: AbortSignal
+}
+
+async function* pollBulkOperation({
+  adminSession,
+  operationId,
+  pollIntervalSeconds,
+  useAdaptivePolling = false,
+  initialPollIntervalSeconds = INITIAL_POLL_INTERVAL_SECONDS,
+  initialPollCount = INITIAL_POLL_COUNT,
+  abortSignal,
+}: PollBulkOperationOptions): AsyncGenerator<BulkOperation, BulkOperation> {
   let pollCount = 0
 
   while (true) {
@@ -61,19 +109,28 @@ async function* pollBulkOperation(
 
     const latestOperationState = response.bulkOperation
 
-    if (TERMINAL_STATUSES.includes(latestOperationState.status) || abortSignal.aborted) {
+    if (TERMINAL_STATUSES.includes(latestOperationState.status) || abortSignal?.aborted) {
       return latestOperationState
     } else {
       yield latestOperationState
     }
 
     pollCount++
+    let pollInterval = pollIntervalSeconds
+    if (useAdaptivePolling && pollCount <= initialPollCount) {
+      pollInterval = initialPollIntervalSeconds
+    }
 
-    // Use shorter interval for the first 10 polls, then switch to regular interval
-    const pollInterval = pollCount <= INITIAL_POLL_COUNT ? INITIAL_POLL_INTERVAL_SECONDS : REGULAR_POLL_INTERVAL_SECONDS
-
-    // eslint-disable-next-line no-await-in-loop
-    await Promise.race([sleep(pollInterval), new Promise((resolve) => abortSignal.addEventListener('abort', resolve))])
+    if (abortSignal) {
+      // eslint-disable-next-line no-await-in-loop
+      await Promise.race([
+        sleep(pollInterval),
+        new Promise((resolve) => abortSignal.addEventListener('abort', resolve)),
+      ])
+    } else {
+      // eslint-disable-next-line no-await-in-loop
+      await sleep(pollInterval)
+    }
   }
 }
 
--- a/packages/app/src/cli/services/bulk-operations/execute-bulk-operation.test.ts
+++ b/packages/app/src/cli/services/bulk-operations/execute-bulk-operation.test.ts
@@ -1,7 +1,7 @@
 import {executeBulkOperation} from './execute-bulk-operation.js'
 import {runBulkOperationQuery} from './run-query.js'
 import {runBulkOperationMutation} from './run-mutation.js'
-import {watchBulkOperation} from './watch-bulk-operation.js'
+import {watchBulkOperation, shortBulkOperationPoll} from './watch-bulk-operation.js'
 import {downloadBulkOperationResults} from './download-bulk-operation-results.js'
 import {validateApiVersion} from '../graphql/common.js'
 import {BulkOperationRunQueryMutation} from '../../api/graphql/bulk-operations/generated/bulk-operation-run-query.js'
@@ -67,6 +67,7 @@ describe('executeBulkOperation', () => {
 
   beforeEach(() => {
     vi.mocked(ensureAuthenticatedAdminAsApp).mockResolvedValue(mockAdminSession)
+    vi.mocked(shortBulkOperationPoll).mockResolvedValue(createdBulkOperation)
   })
 
   afterEach(() => {
@@ -305,7 +306,7 @@ describe('executeBulkOperation', () => {
     })
   })
 
-  test('waits for operation to finish and renders success when watch is provided and operation finishes with COMPLETED status', async () => {
+  test('uses watchBulkOperation (not quickWatchBulkOperation) when watch flag is true', async () => {
     const query = '{ products { edges { node { id } } } }'
     const initialResponse: BulkOperationRunQueryMutation['bulkOperationRunQuery'] = {
       bulkOperation: createdBulkOperation,
@@ -320,7 +321,9 @@ describe('executeBulkOperation', () => {
 
     vi.mocked(runBulkOperationQuery).mockResolvedValue(initialResponse)
     vi.mocked(watchBulkOperation).mockResolvedValue(completedOperation)
-    vi.mocked(downloadBulkOperationResults).mockResolvedValue('{"id":"gid://shopify/Product/123"}')
+    vi.mocked(downloadBulkOperationResults).mockResolvedValue(
+      '{"data":{"products":{"edges":[{"node":{"id":"gid://shopify/Product/123"}}],"userErrors":[]}},"__lineNumber":0}',
+    )
 
     await executeBulkOperation({
       organization: mockOrganization,
@@ -330,6 +333,13 @@ describe('executeBulkOperation', () => {
       watch: true,
     })
 
+    expect(watchBulkOperation).toHaveBeenCalledWith(
+      mockAdminSession,
+      createdBulkOperation.id,
+      expect.any(Object),
+      expect.any(Function),
+    )
+    expect(shortBulkOperationPoll).not.toHaveBeenCalled()
     expect(renderSuccess).toHaveBeenCalledWith(
       expect.objectContaining({
         headline: expect.stringContaining('Bulk operation succeeded:'),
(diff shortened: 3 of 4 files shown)


ericlee878:
* **#6700** <a href="https://app.graphite.com/github/pr/Shopify/cli/6700?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6687** <a href="https://app.graphite.com/github/pr/Shopify/cli/6687?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Shopify/cli/6687?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/app/src/cli/services/bulk-operations/watch-bulk-operation.ts
```diff
@@ -14,8 +14,33 @@ const TERMINAL_STATUSES = ['COMPLETED', 'FAILED', 'CANCELED', 'EXPIRED']
...
 
+export async function quickWatchBulkOperation(adminSession: AdminSession, operationId: string): Promise<BulkOperation> {
+  const startTime = Date.now()
+
+  while (Date.now() - startTime < QUICK_WATCH_TIMEOUT_MS) {
+    // eslint-disable-next-line no-await-in-loop
+    const {bulkOperation} = await fetchBulkOperation(adminSession, operationId)
+
+    if (bulkOperation && TERMINAL_STATUSES.includes(bulkOperation.status)) {
+      return bulkOperation
```
allancalix: I think this could just be:

```typescript
TERMINAL_STATUSES.includes(bulkOperation?.status || "")
```

packages/app/src/cli/services/bulk-operations/watch-bulk-operation.ts
```diff
@@ -14,8 +14,33 @@ const TERMINAL_STATUSES = ['COMPLETED', 'FAILED', 'CANCELED', 'EXPIRED']
...
+    }
+
+    // eslint-disable-next-line no-await-in-loop
+    await sleep(QUICK_WATCH_POLL_INTERVAL_MS / 1000)
+  }
+
+  const {bulkOperation} = await fetchBulkOperation(adminSession, operationId)
```
allancalix: Why are you fetching it again here? At this point you've already fetched it several times

packages/app/src/cli/services/bulk-operations/execute-bulk-operation.ts
```diff
@@ -144,6 +160,10 @@ async function renderBulkOperationResult(operation: BulkOperation, outputFile?:
   }
 }
 
+function resultsContainUserErrors(results: string): boolean {
+  return results.includes('"userErrors":[{')
```
allancalix: Why not parse it at this point? This is coupled pretty hard to our json serializer and not our api response

packages/app/src/cli/services/bulk-operations/watch-bulk-operation.ts
ericlee878: We continue fetching it to see if the state of the bulkOperation has been updated since we last fetched it.

packages/app/src/cli/services/bulk-operations/execute-bulk-operation.ts
ericlee878: Would that not cause the performance to be very very bad?

packages/app/src/cli/services/bulk-operations/watch-bulk-operation.ts
ericlee878: Just for my own learning, could you explain why this is better/cleaner?

packages/app/src/cli/services/bulk-operations/execute-bulk-operation.ts
jordanverasamy: good question, i'm not sure -- i wonder if you can benchmark this! e.g. make a 10MB result file and measure how long it takes to parse, then measure how long it takes to search for this substring

packages/app/src/cli/services/bulk-operations/execute-bulk-operation.ts
```diff
@@ -91,7 +91,8 @@ export async function executeBulkOperation(input: ExecuteBulkOperationInput): Pr
         await renderBulkOperationResult(operation, outputFile)
       }
     } else {
-      await renderBulkOperationResult(createdOperation, outputFile)
+      const operation = await quickWatchBulkOperation(adminSession, createdOperation.id)
+      await renderBulkOperationResult(operation, outputFile)
```
ericlee878, line 108: Maybe we need to change this function to properly deal with the `RUNNING` state if we want to call it here now. I remember we talked about this a while ago @jordanverasamy

packages/app/src/cli/services/bulk-operations/watch-bulk-operation.ts
allancalix: I dont know that it is but one weird quirk about javascript is you will end up passing a type you might not expect to the function depending on the type of the input.

```typescript
let bulkOperation = 10 // some other type that returns truthy
bulkOperation && bulkOperation.status // returns undefined
```

So if `status.includes(undefined)` fails you'll get some weird error. The way I mentioned always returns a known value if the former type is something unexpected. Maybe its not something that happens in typescript idk but just for your consideration.

packages/app/src/cli/services/bulk-operations/execute-bulk-operation.ts
allancalix: You should check with the upper limit is on what you can even return in the 3 second window i'm almost positive its not more than like 40KB of output data

packages/app/src/cli/services/bulk-operations/watch-bulk-operation.ts
ericlee878: Okok I see.
Then, could this also be a way of writing things:

```
bulkOperation.status && TERMINAL_STATUSES.includes(bulkOperation.status)
```

packages/app/src/cli/services/bulk-operations/execute-bulk-operation.ts
jordanverasamy: yea if we know it'll be a small file we can certainly get away with parsing, it'll be blazing fast at these scales

packages/app/src/cli/services/bulk-operations/execute-bulk-operation.ts
jordanverasamy: yeah! i think now that we have quick-watch we need to handle that :)

packages/app/src/cli/services/bulk-operations/execute-bulk-operation.ts
ericlee878: After a mini-investigation with Allan, determined that there is a very very little chance that parsing will cause a performance bottleneck. So, will change to JSON parsing!

packages/app/src/cli/services/bulk-operations/watch-bulk-operation.ts
ericlee878: Wait, I just realized. This is typescript not javascript. Typescript guarantees that `bulkOperation` is either a proper BulkOperation , null, or undefined. So, the original way should be fine?

packages/app/src/cli/services/bulk-operations/execute-bulk-operation.ts
ericlee878: Changed to JSON parsing!

packages/app/src/cli/services/bulk-operations/execute-bulk-operation.ts
ericlee878: Added `RUNNING` functionality. Also, added tophatting for `RUNNING`.

ericlee878:
Are there any impactful consequences with extra polling? Does it slow anything down or put too much pressure on the platform or something? Tbh, I don't think I know enough about the speed of the API, the user/agent's needs, or the impact of the extra polling to decide this. Happy to refactor either way though!

packages/app/src/cli/services/bulk-operations/watch-bulk-operation.ts
```diff
@@ -14,8 +14,33 @@ const TERMINAL_STATUSES = ['COMPLETED', 'FAILED', 'CANCELED', 'EXPIRED']
 const POLL_INTERVAL_SECONDS = 5
 const API_VERSION = '2026-01'
 
+export const QUICK_WATCH_TIMEOUT_MS = 3000
+export const QUICK_WATCH_POLL_INTERVAL_MS = 300
+
 export type BulkOperation = NonNullable<GetBulkOperationByIdQuery['bulkOperation']>
 
+export async function quickWatchBulkOperation(adminSession: AdminSession, operationId: string): Promise<BulkOperation> {
```
jordanverasamy: I'm not sure about this name -- if I imagine being a new developer reading the code and I see "quickWatchBulkOperation" I don't think I would be able to intuit what it actually does. What does it mean to quickly watch something?

We aren't really "watching" anything (we used `watch` in the name of the other method because it displays progress, but this function doesn't!) and we don't do anything to make anything go faster, so `quick` doesn't seem appropriate.

How can we boil down what this function does to its purest essence? I think I'd describe it as: "poll the bulk operation for a short period". Maybe the name should be more along those lines?

packages/app/src/cli/services/bulk-operations/execute-bulk-operation.ts
```diff
@@ -105,7 +91,8 @@ export async function executeBulkOperation(input: ExecuteBulkOperationInput): Pr
         await renderBulkOperationResult(operation, outputFile)
       }
     } else {
-      await renderBulkOperationResult(createdOperation, outputFile)
+      const operation = await quickWatchBulkOperation(adminSession, createdOperation.id)
```
jordanverasamy: Curious why you opted to make this a new function. Other options might have been:

- do a fetch inline right here
- make it an option on the normal `watchBulkOperation` (e.g. a `maxWatchTime` parameter)

packages/app/src/cli/services/bulk-operations/watch-bulk-operation.ts
jordanverasamy: (on that note, I'm curious about your decision to not display anything to the user while this polling is happening! do you think that's correct or should we display something? why?)

ericlee878: Regarding the name, the function could probably use a better name now that I look at it. Maybe something like `shortBulkOperationPoll`?

Regarding displaying something to the user while polling, I thought this was unnecessary because it was only 3 seconds, but maybe it is? Happy to do both though.

packages/app/src/cli/services/bulk-operations/execute-bulk-operation.ts
ericlee878: I thought it was the simplest way to go about this. Definitely thought it was better than doing a fetch inline right there (if we wanted to do multiple polling at least). 

Did not think about making it an option on the normal `watchBulkOperation` function... I think this might be the best way to go about this now that I think about it. 

What do you think?

packages/app/src/cli/services/bulk-operations/watch-bulk-operation.ts
ericlee878: Renamed to `shortBulkOperationPoll`!

ericlee878: Created [ticket](https://github.com/orgs/shop/projects/208/views/34?pane=issue&itemId=3708308214&issue=shop%7Cissues-api-foundations%7C1160) for user display while shortPoll.

packages/app/src/cli/services/bulk-operations/execute-bulk-operation.ts
ericlee878: I'm keeping `shortBulkOperationPoll` and `watchBulkOperation` as separate functions rather than consolidating them. Having two separate functions communicate intent better. You immediately know whether you're doing a full watch or a quick poll, and the purpose of those functions are easier to understand. Don't have too strong of an opinion here though, so am happy to change if you want me to.

jordanverasamy: That makes sense to me, thanks :)

packages/app/src/cli/services/bulk-operations/execute-bulk-operation.ts
```diff
@@ -133,17 +134,39 @@ async function renderBulkOperationResult(operation: BulkOperation, outputFile?:
...
 
         if (outputFile) {
           await writeFile(outputFile, results)
-          renderSuccess({headline, body: [`Results written to ${outputFile}`], customSections})
         } else {
-          renderSuccess({headline, customSections})
           outputResult(results)
         }
+
+        if (hasUserErrors) {
```
jordanverasamy, line 158: I'm a really big fan of this section. There's subtle complexity here because of the way the flags and cases interact (you need to print the right thing whether there were errors or not AND whether the results went to the file or STDOUT) but you were able to write it in a way that feels clean and easy to understand. Great job!! 🚀

packages/app/src/cli/services/bulk-operations/execute-bulk-operation.ts
```diff
@@ -154,6 +177,22 @@ async function renderBulkOperationResult(operation: BulkOperation, outputFile?:
...
 
+function resultsContainUserErrors(results: string): boolean {
+  const lines = results.trim().split('\n')
+
+  return lines.some((line) => {
+    try {
+      const parsed = JSON.parse(line)
+      if (!parsed.data) return false
+
+      const result = Object.values(parsed.data)[0] as any
```
jordanverasamy: are we able to do better typing than `as any` here? if not, why?

packages/app/src/cli/services/bulk-operations/watch-bulk-operation.ts
```diff
@@ -14,8 +14,35 @@ const TERMINAL_STATUSES = ['COMPLETED', 'FAILED', 'CANCELED', 'EXPIRED']
...
 export type BulkOperation = NonNullable<GetBulkOperationByIdQuery['bulkOperation']>
 
+export async function shortBulkOperationPoll(adminSession: AdminSession, operationId: string): Promise<BulkOperation> {
+  const startTime = Date.now()
+
+  let bulkOperation: BulkOperation | undefined
+
+  do {
+    // eslint-disable-next-line no-await-in-loop
+    const response = await fetchBulkOperation(adminSession, operationId)
```
jordanverasamy: i wonder if you'd be able to reuse `pollBulkOperation` here instead of calling `fetchBulkOperation` directly? this would get you nice things for free like e.g. the use of `Promise.race` to let CTRL+C gracefully interrupt you, and not needing to re-implement the `TERMINAL_STATUSES` check. but you'd also need to make the poll duration configurable so hmm... maybe not worth it?

packages/app/src/cli/services/bulk-operations/execute-bulk-operation.ts
ericlee878: resolved during pairing

packages/app/src/cli/services/bulk-operations/watch-bulk-operation.ts
jordanverasamy: We paired and were able to do this nicely!

jordanverasamy (approved):

jordanverasamy:
> Are there any impactful consequences with extra polling? Does it slow anything down or put too much pressure on the platform or something? Tbh, I don't think I know enough about the speed of the API, the user/agent's needs, or the impact of the extra polling to decide this. Happy to refactor either way though!

Not really, it's not that many requests in the grand scheme of things. IMO the real cost there is code complexity and developer brainpower, not platform pressure. But that said I'm happy with keeping it and like where we landed, so don't worry :)

packages/app/src/cli/services/bulk-operations/execute-bulk-operation.ts
ericlee878: Thanks!! :smile:

PR #6698, opened 2025-12-09
Abort theme command when an invalid environment is provided

EvilGenius13:

When running theme commands with an environment that doesn't exist, we fall back to using the cached store data instead (which can be a different store entirely). 

See below pictures.
Legitimate Environment
Screenshot 2025-12-09 at 1 37 04 PM
We get a warning that the environment isn't found but still use the prior store
Screenshot 2025-12-09 at 1 37 23 PM

When running a theme command, we will check if the environment has the store flag which is needed, if it doesn't exist, stop the command and return a warning. 

I chose this way as we already run `loadEnvironment` in the base command and it only renders a warning that the environment doesn't exist. 
image

### Measuring impact

### Checklist

### DIFF
--- a/packages/theme/src/cli/utilities/theme-command.ts
+++ b/packages/theme/src/cli/utilities/theme-command.ts
@@ -79,6 +79,10 @@ export default abstract class ThemeCommand extends Command {
 
     // Single environment or no environment
     if (environments.length <= 1) {
+      if (environments[0] && !flags.store) {
+        throw new AbortError(`Please provide a valid environment.`)
+      }
+
       const session = commandRequiresAuth ? await this.createSession(flags) : undefined
       const commandName = this.constructor.name.toLowerCase()
 
--- a/.changeset/fifty-plants-ask.md
+++ b/.changeset/fifty-plants-ask.md
@@ -1,0 +1,5 @@
+---
+'@shopify/theme': patch
+---
+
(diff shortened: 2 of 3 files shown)


packages/theme/src/cli/utilities/theme-command.ts
```diff
@@ -79,6 +79,10 @@ export default abstract class ThemeCommand extends Command {
 
     // Single environment or no environment
     if (environments.length <= 1) {
+      if (environments[0] && !flags.store) {
```
graygilmore, line 82: I've re-read this a few times and I don't think I understand how this works. Wouldn't `environments[0]` imply that there *is* an environment?

EvilGenius13: Yup! So when you pass in an environment like `-e my-store` or `--environment my-store` it's extracted in the base command as the environment before it checks if it's valid or not. Even after it checks and is found to not be a valid environment, the `flag.environment` still has the value. 

So in our case we check if an environment was actually set (no running a regular command like `theme list` still works without), and if the store flag is missing from the env. If it is, we know the command would not work.

graygilmore: Ah! Because we'd populate `flags.store` based on the environment. I gotcha now!

PR #6703, opened 2025-12-10
Enable custom headers in CLI GraphiQL

amcaplan:
## Summary
- Re-enables the GraphiQL header editor UI that was previously disabled
- Adds header filtering to block problematic browser/hop-by-hop headers
- Allows custom headers like `Shopify-Search-Query-Debug=1` to pass through to the Admin API

## Problem
The CLI's built-in GraphiQL didn't allow passing custom headers. This feature was explicitly disabled because the proxy couldn't distinguish between user-set headers and browser defaults, causing conflicts.

## Solution
- Enable `isHeadersEditorEnabled: true` in the GraphiQL component
- Implement a blocklist of headers that should NOT be forwarded:
  - Hop-by-hop headers (RFC 7230): connection, keep-alive, transfer-encoding, etc.
  - Proxy-controlled headers: host, content-type, accept, authorization, cookie, etc.
- Filter incoming request headers and merge safe custom headers with required headers

## Test plan
- [x] All 9 new tests pass for header filtering logic
- [x] Type-check passes
- [x] Lint passes
- [x] Manual testing: Run `shopify app dev`, open GraphiQL, add custom header, verify it reaches the API

Fixes: https://github.com/shop/issues-develop/issues/21688

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/.changeset/stale-wolves-follow.md
+++ b/.changeset/stale-wolves-follow.md
@@ -1,0 +1,5 @@
+---
+'@shopify/app': minor
+---
+
+Enable custom headers in CLI GraphiQL. Users can now set custom headers like `Shopify-Search-Query-Debug=1` in the GraphiQL interface to pass debugging headers to the Admin API.


packages/app/src/cli/services/dev/graphiql/server.ts
```diff
@@ -15,6 +15,49 @@ import {createRequire} from 'module'
...
+])
+
+/**
+ * Filters request headers to extract only custom headers that are safe to forward.
+ * Blocked headers and non-string values are excluded.
+ */
+export function filterCustomHeaders(headers: {[key: string]: string | string[] | undefined}): {[key: string]: string} {
+  const customHeaders: {[key: string]: string} = {}
+  for (const [key, value] of Object.entries(headers)) {
+    if (!BLOCKED_HEADERS.has(key.toLowerCase()) && typeof value === 'string') {
```
ravangen: If we're worried about security of which headers get passed through, and we have all of the `req.headers`, can't we invert this check and do `if ALLOWED_HEADERS.has`?

ravangen (commented):
Should ideally also allow:
- `Shopify-GraphQL-Cost-Debug` (https://shopify.dev/docs/api/usage/limits#graphql-response)
- `Shopify-Debug-Headers`
- `X-Shopify-Profile`, `X-Shopify-Profile-Auth`
- `Shopify-Force-Canary`, `Shopify-Force-Alternate-Backend` ([docs](https://vault.shopify.io/page/Canary-deploys-of-Shopify-Core~dhb21d2.md))

There are [other ones](https://github.com/Shopify/ubertool/blob/main/src/known-headers.ts), but not sure if they are applicable here.

packages/app/src/cli/services/dev/graphiql/server.ts
amcaplan: This isn't really a security concern - anything they can generate in GraphiQL, they can generate manually as well via Postman, cURL, etc. The issue is to prevent situations seen previously where default browser headers are passed through and the admin API breaks because it only wants to talk to an API client, not a browser directly.

shauns (approved):

PR #6706, opened 2025-12-11
`shopify app bulk cancel` command to cancel bulkops through CLI

ericlee878:
Issue: https://github.com/shop/issues-api-foundations/issues/1164

### Summary
Adds a new `shopify app bulk cancel` command to cancel running bulk operations, complementing the existing `bulk execute` and `bulk status` commands.

### Usage
```
# Cancel by positional argument
shopify app bulk cancel gid://shopify/BulkOperation/123

# Cancel by flag
shopify app bulk cancel --id gid://shopify/BulkOperation/123

# With store flag
shopify app bulk cancel gid://shopify/BulkOperation/123 --store my-store.myshopify.com
```

### Implementation Details
- ID accepts both flag and positional argument
   - The operation ID can be provided as --id flag or as a positional argument for ergonomics:
    ```
    shopify app bulk cancel gid://shopify/BulkOperation/123  # positional
    shopify app bulk cancel --id gid://shopify/BulkOperation/123  # flag
    ``` 

- Status-aware result messaging
  - The command renders different UI based on what state the operation is in after the cancel request.

- API version hardcoded `2026-01`
  - This can be changed when the API version handling PR is merged.

- Should we add a `--watch` flag to this?

### Testing

1. Create a large Bulk Operation
- Ask cursor to build a large variable-file. It should look like this but with much more variables:
 ```
{"input":{"title":"Test Product 1","descriptionHtml":"<p>This is test product 1</p>","productType":"Test Type","vendor":"Test Vendor","tags":["bulk","test"],"metafields":[{"namespace":"custom","key":"field1","value":"value1","type":"single_line_text_field"},{"namespace":"custom","key":"field2","value":"value2","type":"single_line_text_field"},{"namespace":"custom","key":"field3","value":"value3","type":"single_line_text_field"}]}}
{"input":{"title":"Test Product 2","descriptionHtml":"<p>This is test product 2</p>","productType":"Test Type","vendor":"Test Vendor","tags":["bulk","test"],"metafields":[{"namespace":"custom","key":"field1","value":"value1","type":"single_line_text_field"},{"namespace":"custom","key":"field2","value":"value2","type":"single_line_text_field"},{"namespace":"custom","key":"field3","value":"value3","type":"single_line_text_field"}]}}
```
  - Run this command:
```
pnpm shopify:run app bulk execute --path ./coffee-and-more --variable-file large-variables.jsonl -q 'mutation productCreate($input: ProductInput!) {
      product {
        id
        title
        metafields(first: 10) {
          edges {
            node {
              id
              namespace
              key
              value
            }
          }
        }
      }
      userErrors {
        field
        message
      }
    }
  }'
```
This should show. 

Save the bulk operation id somewhere.

2. Cancel the Bulk Operation
- Run the command
```
pnpm shopify:run app bulk cancel \
  --path ./coffee-and-more \
  --id gid://shopify/BulkOperation/<bulk_operation_id>
```
This should show.

3. Check the status of the bulk operation
- Run the command
```
pnpm shopify:run app bulk status \
  --path ./coffee-and-more \
  --id gid://shopify/BulkOperation/6578182226092
```

This should show.

### DIFF
--- a/packages/app/src/cli/commands/app/bulk/cancel.ts
+++ b/packages/app/src/cli/commands/app/bulk/cancel.ts
@@ -1,0 +1,47 @@
+import {appFlags} from '../../../flags.js'
+import AppLinkedCommand, {AppLinkedCommandOutput} from '../../../utilities/app-linked-command.js'
+import {prepareAppStoreContext} from '../../../utilities/execute-command-helpers.js'
+import {cancelBulkOperation} from '../../../services/bulk-operations/cancel-bulk-operation.js'
+import {normalizeBulkOperationId} from '../../../services/bulk-operations/bulk-operation-status.js'
+import {Flags} from '@oclif/core'
+import {globalFlags} from '@shopify/cli-kit/node/cli'
+import {normalizeStoreFqdn} from '@shopify/cli-kit/node/context/fqdn'
+
+export default class BulkCancel extends AppLinkedCommand {
+  static summary = 'Cancel a bulk operation.'
+
+  static description = 'Cancels a running bulk operation by ID.'
+
+  static hidden = true
+
+  static flags = {
+    ...globalFlags,
+    ...appFlags,
+    id: Flags.string({
+      description: 'The bulk operation ID to cancel (numeric ID or full GID).',
+      env: 'SHOPIFY_FLAG_ID',
+      required: true,
+    }),
+    store: Flags.string({
+      char: 's',
+      description: 'The store domain. Must be an existing dev store.',
+      env: 'SHOPIFY_FLAG_STORE',
+      parse: async (input) => normalizeStoreFqdn(input),
+    }),
+  }
+
+  async run(): Promise<AppLinkedCommandOutput> {
+    const {flags} = await this.parse(BulkCancel)
+
+    const {appContextResult, store} = await prepareAppStoreContext(flags)
+
+    await cancelBulkOperation({
+      organization: appContextResult.organization,
+      storeFqdn: store.shopDomain,
+      operationId: normalizeBulkOperationId(flags.id),
+      remoteApp: appContextResult.remoteApp,
+    })
+
+    return {app: appContextResult.app}
+  }
+}
--- a/packages/app/src/cli/api/graphql/bulk-operations/generated/bulk-operation-cancel.ts
+++ b/packages/app/src/cli/api/graphql/bulk-operations/generated/bulk-operation-cancel.ts
@@ -1,0 +1,102 @@
+/* eslint-disable @typescript-eslint/consistent-type-definitions, @typescript-eslint/no-redundant-type-constituents */
+import * as Types from './types.js'
+
+import {TypedDocumentNode as DocumentNode} from '@graphql-typed-document-node/core'
+
+export type BulkOperationCancelMutationVariables = Types.Exact<{
+  id: Types.Scalars['ID']['input']
+}>
+
+export type BulkOperationCancelMutation = {
+  bulkOperationCancel?: {
+    bulkOperation?: {
+      completedAt?: unknown | null
+      createdAt: unknown
+      errorCode?: Types.BulkOperationErrorCode | null
+      fileSize?: unknown | null
+      id: string
+      objectCount: unknown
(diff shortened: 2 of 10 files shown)

### CODE REVIEW

**Comment by ericlee878** [comment]
* **#6706** <a href="https://app.graphite.com/github/pr/Shopify/cli/6706?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Shopify/cli/6706?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

File: packages/app/src/cli/commands/app/bulk/cancel.ts [hunk 1]
```diff
@@ -0,0 +1,57 @@
...
+    store: Flags.string({
+      char: 's',
+      description: 'The store domain. Must be an existing dev store.',
+      env: 'SHOPIFY_FLAG_STORE',
+      parse: async (input) => normalizeStoreFqdn(input),
+    }),
+  }
+
+  static args = {
+    id: Args.string({
```
gonzaloriestra: I think we should choose between arg and flag, offering both ways is a bit weird.

What I would do:  
\- Only use the flag to be consistent with bulk status  
\- When not provided, show a list with the current operations to choose from there

ericlee878: I like this idea. Before I implement it, do you have any thoughts on this @nickwesselman ?

nickwesselman: Yeah we should definitely stick to the flag and get rid of the argument.  
  
I'm not sure we need the list, since you can get that list from the `status` command. We can keep this command simple.

ericlee878: Made the changes!

ericlee878:
Not too sure about this. I think @allancalix will have more context. But for the limit in the number of simultaneous bulk operations part, I'm pretty sure you can only run one bulk operation per shop at a time.

allancalix:
@gonzaloriestra We allow at most 5 bulk mutations at any time per app. Most of the bulk operation use cases are async and offline, we didn't consider how a recent change we made to allow bulk mutation jobs to run longer would cause delays for cancellation (cc: @teddyhwang). There are things we can do to improve this but this is the first we have heard of cancellation delay causing issues so we'll have to make changes to lower the latency on it.

gonzaloriestra:
> We allow at most 5 bulk mutations at any time per app. 

In that case, I suspect this command would almost never be used... But it's ok to keep it.

Also, if the cancellation is going to be fast, I think the command should be sync (similar to `app execute`) instead of creating another bulk operation that you have to monitor. What do you think?

nickwesselman:
Initiating a cancel is sync, but the process is async. It will take time for the job to respond to the cancel "request." I think the output of cancel in the screenshot above though looks too much like a new job has been created.

Let's use the info banner to confirm inputs -- especially in the case that there is an error due to a bad ID or having the wrong app/store context. Then use the success banner to confirm that the cancellation was initiated, set expectations, and tell them how to check on it.

```
┌─ info ───────────────────────────────────────────────────────┐
│                                                              │
│  Canceling bulk operation.                                   │
│                                                              │
│  • ID: gid://shopify/BulkOperation/6578182226092             │
│  • Organization: Coffee & More                               │
│  • App: coffee-and-more                                      │
│  • Store: coffee-and-more-1851.myshopify.com                 │
│                                                              │
└──────────────────────────────────────────────────────────────┘

┌─ success ────────────────────────────────────────────────────┐
│                                                              │
│  Bulk operation is being cancelled.                          │
│                                                              │
│  This may take a few moments to complete. To check the       │
│  status of the operation, run:                               │
│  `shopify app bulk status --id=6578182226092`                │
│                                                              │
└──────────────────────────────────────────────────────────────┘
```

ericlee878:
@allancalix Should we do a quick-watch feature here to confirm inputs for the info banner? Or do you think there is a better way to do this for cancel?

ericlee878:
@nickwesselman Made the changes to reflect this!

PR #6711, opened 2025-12-11
Create a slight delay in on shortcut keys so you can't spam theme in theme dev

EvilGenius13:

Fixes https://github.com/Shopify/developer-tools-team/issues/918

If you copy and paste in the terminal when running `theme dev` you can end up with a ton of tabs opening.
Since we are watching for certain keys like `t` to open localhost, if you were to copy and paste text that had multiple `t|p|e|g`<-- shortcut keys, you can end up opening a lot of tabs.

When a key is pressed we create a short lived lock, and any other keys pressed during that time will be ignored. The delay is set to 100ms so we shouldn't see any problems with someone quickly doing something like opening localhost and then the editor for themselves.

I also added some extra keypress tests that we didn't originally have.

### How to test your changes?
This is a fun one to test.
- On a current shopify version, run `theme dev` and in the console, paste `tteeppggtt`. This should open 10 tabs.
- Build this branch and do the same. Only the first key should register (`t`) and open localhost once.
- You can also test if it feels snappy enough by manually pressing a couple keys like opening localhost, then the editor, then the preview pane to see how it feels.

### Measuring impact

### Checklist

### DIFF
--- a/packages/theme/src/cli/services/dev.ts
+++ b/packages/theme/src/cli/services/dev.ts
@@ -12,6 +12,7 @@ import {Theme} from '@shopify/cli-kit/node/themes/types'
 import {checkPortAvailability, getAvailableTCPPort} from '@shopify/cli-kit/node/tcp'
 import {AbortError} from '@shopify/cli-kit/node/error'
 import {openURL} from '@shopify/cli-kit/node/system'
+import {debounce} from '@shopify/cli-kit/common/function'
 import chalk from '@shopify/cli-kit/node/colors'
 import readline from 'readline'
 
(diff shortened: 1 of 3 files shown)


packages/theme/src/cli/services/dev.ts
```diff
@@ -165,6 +175,48 @@ export async function dev(options: DevOptions) {
   ])
 }
 
+export function createKeypressHandler(
```
graygilmore, line 146: Maybe I'm not reading something right but I don't see this function actually used anywhere?

packages/theme/src/cli/services/dev.ts
```diff
@@ -150,6 +154,12 @@ export async function dev(options: DevOptions) {
         openURLSafely(urls.giftCard, 'gift card preview')
         break
     }
+
+    // This prevents the keypress handler from being called multiple times in a short span
+    keyBlock = true
+    setTimeout(() => {
```
graygilmore: Are you familiar with [debouncing](https://developer.mozilla.org/en-US/docs/Glossary/Debounce)? We have a `debounce` helper function that we could potentially use here instead? Might be overkill for what you're doing but it's there if you want! https://github.com/Shopify/cli/blob/main/packages/cli-kit/src/public/common/function.ts#L39

We  could use `leading: true` to make sure we don't delay the first keypress.

packages/theme/src/cli/services/dev.ts
EvilGenius13: I had exported it so I could test it directly in the test suite as well.

packages/theme/src/cli/services/dev.ts
EvilGenius13: Sure I could use the helper since it's already available. Let me check this out!

packages/theme/src/cli/services/dev.ts
graygilmore: Hmmmm but we're testing something that's not actually being used so I'm not sure about the value there.

EvilGenius13: I apologize, I left in the other keypress setup. 🤦 Fixing that up now

packages/theme/src/cli/services/dev.ts
```diff
@@ -126,43 +127,55 @@ export async function dev(options: DevOptions) {
...
+
+  return (_str: string, key: {ctrl?: boolean; name?: string}) => {
     if (key.ctrl && key.name === 'c') {
       process.exit()
     }
 
     switch (key.name) {
       case 't':
-        openURLSafely(urls.local, 'localhost')
+        debouncedOpenURL(urls.local, 'localhost')
```
graygilmore, line 159: I might've debounced the click handler instead of the URL opening itself but this works, too!

dejmedus (approved):
This was a super fun 🎩 Looks great!

PR #6736, opened 2025-12-19
[Feature] Include `Request ID` at the footer of `shopify theme *` command errors

fredericoo:
This will improve our relationship with the community, as we frequently need to deal with the back and forth of asking for the --verbose output.

Fixes https://github.com/Shopify/developer-tools-team/issues/467

We make use of the `catch` method in `Command` to append the graphql request id made if present. It only affects `ThemeCommand` and leaves other parts of the cli untouched.

### How to test your changes?

Besides running the written unit tests, you may:
- checkout to repo
- call `pnpm shopify theme pull --store <VALID_STORE_ID>.myshopify.com --theme 999999999` for an error _with_ a request id
- call `pnpm shopify theme list --store fake-store-that-does-not-exist.myshopify.com` for an error _without_ a request id (it fails during auth)

### Measuring impact

### Checklist

### DIFF
--- a/.changeset/dark-mirrors-appear.md
+++ b/.changeset/dark-mirrors-appear.md
@@ -1,0 +1,5 @@
+---
+'@shopify/theme': patch
+---
+
+Command errors in the console will now include the GraphQL Request ID when present


karreiro (approved):
Thank you for this PR, @fredericoo! Neat approach :) LGTM and works as expected too:

Image

I've added @EvilGenius13 because he has more context around `@shopify/cli-kit/node/metadata`, but I believe `cmd_all_last_graphql_request_id` is the way to go :)

Thanks again for this PR!

EvilGenius13 (approved):
Looks good! `cmd_all_last_graphql_request_id` is definitely what we want in this scenario.

PR #6738, opened 2025-12-23
Add support for executing GraphQL queries on non-dev stores

ericlee878:
# Add support for executing GraphQL queries on non-dev stores

Came from: https://app.graphite.com/github/pr/Shopify/cli/6689/Add-support-for-executing-GraphQL-queries-on-non-dev-stores

Made a new PR because this task was handed off to me and the rebasing caused major changes.

Resolves: https://github.com/orgs/shop/projects/208/views/34?pane=issue&itemId=144171469&issue=shop%7Cissues-api-foundations%7C1165

**What**
- Added a new GraphQL query to fetch stores by domain without filtering by store type
- Updates the executeBulkOperation and executeOperation services to accept a store object instead of just the FQDN. This is so that we can determine the storeType of the store. The storeType is either dev or non-dev.
- Modifies the store context to optionally include all store types, not just development stores
- Using the storeType, we only let dev stores use bulkOperation mutations. BulkOperation queries are allowed for all stores.
- The already existing mutation validation was integrated with the part where only let dev stores do bulkOperation mutations.
- Added tests to verify mutation validation works correctly.
- Added helper functions to check if a GraphQL operation is a mutation

### How to test your changes?

1. Try executing a query on a production store:
   ```
   shopify app execute --store=your-production-store.myshopify.com --query "{ shop { name } }"
   ```
   This should succeed.

2. Try executing a mutation on a production store:
   ```
   shopify app execute --store=your-production-store.myshopify.com --query "mutation { productUpdate(input: {}) { product { id } } }"
   ```
   This should fail with an error message indicating mutations can only be executed on Dev Stores.

3. Try executing a mutation on a development store:
   ```
   shopify app execute --store=your-dev-store.myshopify.com --query  "mutation { productUpdate(input: {}) { product { id } } }"
   ```
   This should succeed.


ericlee878:
* **#6738** <a href="https://app.graphite.com/github/pr/Shopify/cli/6738?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Shopify/cli/6738?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/app/src/cli/services/bulk-operations/bulk-operation-status.test.ts
```diff
@@ -203,6 +203,19 @@ describe('getBulkOperationStatus', () => {
     )
   })
 
+  test('uses resolved API version in admin request', async () => {
```
jordanverasamy, line 206: We didn't change `bulk-operation-status.ts`, so I wouldn't expect that we need new tests in `bulk-operation-status.test.ts`. Did we discover something that made us add this?

packages/app/src/cli/utilities/developer-platform-client/partners-client.ts
```diff
@@ -513,7 +513,12 @@ export class PartnersClient implements DeveloperPlatformClient {
     return this.request(ConvertDevToTransferDisabledStoreQuery, input)
   }
 
-  async storeByDomain(orgId: string, shopDomain: string): Promise<OrganizationStore | undefined> {
+  async storeByDomain(
+    orgId: string,
+    shopDomain: string,
+    _includeAllStores = false,
+  ): Promise<OrganizationStore | undefined> {
+    // Note: includeAllStores parameter not implemented for PartnersClient
```
jordanverasamy: [typescript noob question] Could we have just not modified this signature at all, or does it have to match the `app-management-client` signature exactly for some reason?

jordanverasamy (commented):
seems generally reasonable :) tophatting now!

jordanverasamy:
🎩  on a production store successfully:

Queries work:

image

Mutations don't:

image

jordanverasamy:
My one question is whether it would've been possible to accomplish the same stuff with slightly fewer code changes, but I'm sure a reviewer who's more deeply knowledgeable about the CLI will be able to weigh in on that :) from a Bulk Ops perspective this change def works!

packages/app/src/cli/services/bulk-operations/bulk-operation-status.test.ts
ericlee878: I think Nick added these tests to improve test coverage for `resolveApiVersion` [here](https://app.graphite.com/github/pr/Shopify/cli/6689/Add-support-for-executing-GraphQL-queries-on-non-dev-stores). I think the test helps better test the `resolveApiVersion` use.

packages/app/src/cli/utilities/developer-platform-client/partners-client.ts
ericlee878: This is from Nick's original PR (not the rebase), but I think he added it for explicitness. The `_includeAllStores` param + comment makes it clear this parameter exists in the interface but isn't implemented for `PartnersClient`.
That said, it's not strictly required. Happy to simplify if you'd prefer a smaller diff, or we can keep it as-is for the documentation value.

jordanverasamy: No strong opinion tbh, just wanted to ask -- fine with leaving it personally. TY!

ericlee878:
I think the boolean was added as a quick toggle between two hardcoded queries: one filtered to `APP_DEVELOPMENT` stores, one unfiltered. I can see how using `storeTypes: Store[]` would be better. Refactored this way!

packages/app/src/cli/api/graphql/business-platform-organizations/queries/fetch_dev_store_by_domain.graphql
```diff
@@ -1,22 +1,22 @@
...
-          node {
-            id
-            externalId
-            name
-            storeType
-            primaryDomain
-            shortName
-            url
-          }
+query FetchDevStoreByDomain($domain: String) {
```
ryancbahan: what's the purpose of adding a new query that hardcodes the filter rather than just reusing `FetchStoreByDomain` and passing the relevant store types at the callsite?

packages/theme/src/cli/polyfills/promiseWithResolvers.ts
```diff
@@ -1,22 +1,22 @@
...
-    withResolvers?<T>(): PromiseWithResolvers<T>
-  }
-}
-
 // Polyfill for Promise.withResolvers
 // Can remove once our minimum supported Node version is 22
-export function promiseWithResolvers<T>(): PromiseWithResolvers<T> {
-  if (Promise.withResolvers) {
-    return Promise.withResolvers<T>()
+export function promiseWithResolvers<T>(): PromiseWithResolversResult<T> {
```
ryancbahan: is this necessary to your change? feels orthogonal to patch the polyfill

packages/app/src/cli/services/graphql/common.ts
```diff
@@ -115,3 +115,29 @@ export function formatOperationInfo(options: {
...
+}
+
+/**
+ * Validates that mutations can only be executed on dev stores.
+ *
+ * @param graphqlOperation - The GraphQL operation to validate.
+ * @param store - The store where the operation will be executed.
+ * @throws AbortError if attempting to run a mutation on a non-dev store.
+ */
+export function validateMutationStore(graphqlOperation: string, store: OrganizationStore): void {
```
ryancbahan, line 139: Does this need to be in the cli? Can you return a 4xx from the server mutation and let the cli be a thin client over the request?

packages/app/src/cli/api/graphql/business-platform-organizations/queries/fetch_dev_store_by_domain.graphql
ericlee878: Are you referring to the `fetch_dev_store_by_domain.graphql` query? That query was used before I refactored after you left this [comment](https://app.graphite.com/github/pr/Shopify/cli/6738/Add-support-for-executing-GraphQL-queries-on-non-dev-stores#review-PRR_kwDOHibhHs7ZFN1a). It's not being used anymore. I'll go ahead and remove that file. Let me know if you think this makes sense!

packages/app/src/cli/services/graphql/common.ts
ericlee878: I think the current implementation makes sense because bulk mutations are restricted to dev stores only in the CLI. @jordanverasamy  I might be completely wrong here so please correct me if I'm wrong.

jordanverasamy: IMO: this is fundamentally a CLI concern and not a server level concern. In general, mutations should be allowed on prod stores; however, _specifically from the CLI_, mutations should not be allowed against prod stores. So this is the right layer for this check, in my opinion. Can let @nickwesselman weigh in if he likes too!

packages/theme/src/cli/polyfills/promiseWithResolvers.ts
ericlee878: You are totally right. I think this file was changed due to some linting change. Reverted!

packages/app/src/cli/services/graphql/common.ts
ryancbahan: > mutations should not be allowed against prod stores

Is this a pre-existing opinion? I get the high-level rationale for preventing footguns, but similar to GitHub allowing merges/rebases to `main` I'm curious of if this is something we should actually enforce. If I'm a solo entrepreneur working on something, why should Shopify stop me from mutating prod from the CLI if that's my intent?

ryancbahan: Not trying to be pedantic here -- it just seems like an important product consideration across the CLI if all mutations should be gated behind a storetype validation check.

ericlee878: I think the CLI is trending toward more agentic workflows, which raises a safety concern: an agent with CLI access could potentially run bulk mutations against a store.

ericlee878: That being said, I don't think I'm the right person to ask about this. @nickwesselman probably has better input on this.

ryancbahan (approved):
Code lgtm. I didn't tophat, but as an external stakeholder it all makes sense. I still have the question of whether the cli itself should do mutation validation, but I recognize that's plausibly a longer conversation.

packages/app/src/cli/services/graphql/common.ts
nickwesselman: Hey @ryancbahan -- This is a product requirement aligned w/ Foundations SLT. There is a lot of concern about agentic use causing accidental damage to production, especially given that our permission model doesn't provide a way to lock out a developer from a production app (and the shops it is installed on). Even without agents, it's a big potential footgun. Maybe something we can relax if we ever have more granular app access permissions.

PR #6740, opened 2026-01-05
Update ink to 5.2.1

matteodepalo:
Ink had two major version bumps, v5 and v6. v6 requires React 19+ but we are tied to react 18+ because of Polaris.
This PR updates ink to version 5. This introduced only a minor issue with spacing for certain characters which I fixed.

### DIFF
--- a/packages/cli-kit/bin/documentation/examples.ts
+++ b/packages/cli-kit/bin/documentation/examples.ts
@@ -565,43 +565,64 @@ export const examples: {[key in string]: Example} = {
   renderTasks: {
     type: 'async',
     basic: async () => {
-      const stdout = new Stdout({columns: TERMINAL_WIDTH})
+      // Force colors for deterministic output (solid blocks vs animated wave)
+      const originalForceColor = process.env.FORCE_COLOR
+      process.env.FORCE_COLOR = '1'
+      try {
+        const stdout = new Stdout({columns: TERMINAL_WIDTH})
 
-      const tasks = [
-        {
-          title: 'Installing dependencies',
-          task: async () => {
-            await new Promise((resolve) => setTimeout(resolve, 2000))
+        const tasks = [
+          {
+            title: 'Installing dependencies',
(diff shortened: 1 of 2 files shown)


packages/app/src/cli/services/dev/app-events/app-event-watcher-handler.ts
```diff
@@ -135,6 +135,8 @@ async function reload(app: AppLinkedInterface): Promise<AppLinkedInterface> {
     return newApp
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
   } catch (error: any) {
-    throw new Error(`Error reloading app: ${error.message}`, {cause: 'validation-error'})
+    const err = new Error(`Error reloading app: ${error.message}`)
+    ;(err as Error & {cause: string}).cause = 'validation-error'
```
craigmichaelmartin: Using a leading semicolon in the middle of a script we are writing to appease our lint rule is pretty wild to me. I've only ever seen this defensively or by tooling, though maybe this is widely used in this codebase? My vote would be to eslint disable the first line and end it with a semicolon.

Though we could also just reformat it a bit:

```suggestion
    const err = new Error(`Error reloading app: ${error.message}`) as Error & {cause: string}
    err.cause = 'validation-error'
```

packages/theme/src/cli/utilities/theme-environment/proxy.ts
```diff
@@ -114,7 +114,7 @@ export function canProxyRequest(event: H3Event) {
 }
 
 function getStoreFqdnForRegEx(ctx: DevServerContext) {
-  return ctx.session.storeFqdn.replaceAll('.', '\\.')
+  return ctx.session.storeFqdn.replace(/\./g, '\\.')
```
EvilGenius13: 🎩 'd and didn't run into any issues. I'm not sure if there are edge cases that would cause a problem here when the old code was essentially doing the same.

PR #6746, opened 2026-01-08
Fix Promise.withResolvers TypeScript signature conflict

jordanverasamy:
## Summary

#6740 introduced a global declaration for `Promise.withResolvers` ([this line](https://github.com/Shopify/cli/pull/6740/changes#diff-0ce173cf5766df7690ae612e050b1a974672ff3f6a350e319426a2cd3b208da8R9)) but incorrectly marked it as optional with `?`. this conflicts with node 24's native type definition which marks it as required, causing typescript to error:

image

this PR fixes it by removing that optional modifier in the global `PromiseConstructor` interface declaration.

## Test plan

- [x] `pnpm nx run theme:type-check` passes
- [x] `pnpm refresh-manifests` completes successfully

### DIFF
--- a/packages/theme/src/cli/polyfills/promiseWithResolvers.ts
+++ b/packages/theme/src/cli/polyfills/promiseWithResolvers.ts
@@ -6,7 +6,7 @@ interface PromiseWithResolvers<T> {
 
 declare global {
   interface PromiseConstructor {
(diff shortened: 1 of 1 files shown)


jordanverasamy:
HAHA jk no bug i'm just a goober. `dev up` resolves this locally for me

PR #6755, opened 2026-01-13
Update `app bulk execute` to fail if given a mutation document but no variables

jordanverasamy:
Small bugfix related to https://github.com/shop/issues-api-foundations/issues/1047. Found it while testing the command looking for bugs (https://github.com/shop/issues-api-foundations/issues/1081).

## Summary
Without variables, bulk mutations have no input data to operate on. So, this PR adds validation that bulk mutations must include variables (via `--variables` or `--variable-file`). If not, the user now gets a clear error message instead of a cryptic XML error from the staged upload API

## Tophatting

On `main`, running `app bulk execute` against a dev store with a mutation but no variables:

image

On this branch, doing the same:

image

### DIFF
--- a/packages/app/src/cli/services/bulk-operations/execute-bulk-operation.ts
+++ b/packages/app/src/cli/services/bulk-operations/execute-bulk-operation.ts
@@ -207,10 +207,15 @@ function resultsContainUserErrors(results: string): boolean {
   })
 }
 
-/**
- * Validates bulk operation-specific constraints for variables.
- */
 function validateBulkOperationVariables(graphqlOperation: string, variablesJsonl?: string): void {
+  if (isMutation(graphqlOperation) && !variablesJsonl) {
+    throw new AbortError(
+      outputContent`Bulk mutations require variables. Provide a JSONL file with ${outputToken.yellow(
+        '--variable-file',
+      )} or individual JSON objects with ${outputToken.yellow('--variables')}.`,
+    )
+  }
+
   if (!isMutation(graphqlOperation) && variablesJsonl) {
     throw new AbortError(
       outputContent`The ${outputToken.yellow('--variables')} and ${outputToken.yellow(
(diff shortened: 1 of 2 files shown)


packages/app/src/cli/services/bulk-operations/execute-bulk-operation.ts
```diff
@@ -207,10 +207,15 @@ function resultsContainUserErrors(results: string): boolean {
...
- */
 function validateBulkOperationVariables(graphqlOperation: string, variablesJsonl?: string): void {
+  if (isMutation(graphqlOperation) && !variablesJsonl) {
+    throw new AbortError(
+      outputContent`Bulk mutations require variables. Provide a JSONL file with ${outputToken.yellow(
+        '--variable-file',
+      )} or individual JSON objects with ${outputToken.yellow('--variables')}.`,
+    )
+  }
+
```
alexanderMontague, line 218: This is not the correct spot for this in my opinion. OCLIF already has this validation functionality built in via the `exactlyOne` modifier that I already see is being used for the `buildOperationFlags`. Can we do this instead? Less code, proper use of the framework, and setting a better pattern!

jordanverasamy: We need to know whether the provided GraphQL executable document is a query or a mutation in order to determine whether the provided flags are valid. That means we have to read the flag value, parse it (being done here with `isMutation`), and only then do we know what to validate.

As far as I understand, that means `exactlyOne` can't be used here, right? I'm still open to a better/more on-pattern way to do this validation if OCLIF provides a mechanism for it, but not sure what that would be.

alexanderMontague: ahh, I gotcha. Depends on the runtime value of the query flag. I don't think OCLIF supports this. This runtime approach is fine in that case, but another question. Some GQL mutations don't _require_ variables. Will we always expect to have them?

ericlee878:
Good catch!

packages/app/src/cli/services/bulk-operations/execute-bulk-operation.ts
jordanverasamy: Good question! Some GQL mutations don't require variables, but if that's the case, you would only need to run it once. The only reason to run a mutation in _bulk_ is to run it against N different variable hashes, and so bulk mutations do in fact [strictly require an input variable JSONL file](https://shopify.dev/docs/api/usage/bulk-operations/imports#create-a-jsonl-file-and-include-graphql-variables).

To match this, `app execute` allows having no variables, but `app bulk execute` requires them.

alexanderMontague: Perfect, thanks for the explanation. Looks great then!

PR #6756, opened 2026-01-13
Use native `FormData` for bulk mutation variable file uploads

jordanverasamy:
Small improvement related to https://github.com/shop/issues-api-foundations/issues/1047. Found it while testing the command looking for bugs (https://github.com/shop/issues-api-foundations/issues/1081).

## Summary

The `form-data` npm package triggers a deprecation warning when passed as a request body to node-fetch 3.x. This PR fixes it by switching bulk operation file uploads to use native `FormData` + `Blob` instead of the `form-data` package.

### Tophatting

On `main`, running a bulk mutation:

image

On this branch:

image

### DIFF
--- a/packages/app/src/cli/services/bulk-operations/stage-file.ts
+++ b/packages/app/src/cli/services/bulk-operations/stage-file.ts
@@ -5,7 +5,7 @@ import {
 } from '../../api/graphql/bulk-operations/generated/staged-uploads-create.js'
 import {adminRequestDoc} from '@shopify/cli-kit/node/api/admin'
 import {AdminSession} from '@shopify/cli-kit/node/session'
-import {formData, fetch} from '@shopify/cli-kit/node/http'
+import {fetch} from '@shopify/cli-kit/node/http'
 import {AbortError} from '@shopify/cli-kit/node/error'
 import {outputContent} from '@shopify/cli-kit/node/output'
 import {renderSingleTask} from '@shopify/cli-kit/node/ui'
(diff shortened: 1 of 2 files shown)


packages/app/src/cli/services/bulk-operations/stage-file.ts
```diff
@@ -5,7 +5,7 @@ import {
 } from '../../api/graphql/bulk-operations/generated/staged-uploads-create.js'
 import {adminRequestDoc} from '@shopify/cli-kit/node/api/admin'
 import {AdminSession} from '@shopify/cli-kit/node/session'
-import {formData, fetch} from '@shopify/cli-kit/node/http'
```
alexanderMontague, line 8: Can we remove this package from `cli-kit` now? I'd love to see a follow up PR if that is the case, reducing the bundle! Thanks

alexanderMontague (approved):

packages/app/src/cli/services/bulk-operations/stage-file.ts
jordanverasamy: Actually yeah :) Claude did that at first, but I asked him to minimize the diff and make this PR smaller, because I wasn't sure if removing it entirely was productive. I can absolutely do that in a follow-up!

jordanverasamy: Aw, looks like we can't actually remove `form-data` from `cli-kit` yet.

I started sketching out the PR to fully remove it ([link](https://github.com/Shopify/cli/pull/6760)), but I discovered `@shopify/oxygen-cli` still imports `formData` from `@shopify/cli-kit/node/http`. (See [here](https://github.com/Shopify/oxygen-cli/blob/main/src/deploy/upload-files.ts#L4).)

We'd need to update `oxygen-cli` to use native `FormData` first, then revisit this. 😭

PR #6758, opened 2026-01-14
Update feature flag gating for extension templates

dmerand:

The current implementation uses `hasFeatureFlag` which does not delegate to Experiments Platform -- it only uses Business Platform-native feature flags, which are deprecated. Per [this Slack discussion](https://shopify.slack.com/archives/C07UJ7UNMTK/p1768406545165449), the `enabledFlags` API does delegate to Verdict and should be used for new feature flags.

However, we need to maintain backward compatibility with existing beta flags while enabling new templates to use Verdict-delegated flags. More on the reasoning for that [here](https://docs.google.com/document/d/1ApIELHXU2mZlkpXMxmzwhutKTSwK5n6c-zJDXcNeCMw/edit?tab=t.0).

This PR implements a dual flag strategy that allows gradual migration from legacy beta flags to modern exp flags:

#### Dual Flag System

- **`organizationBetaFlags`** (Legacy) → Uses `hasFeatureFlag` without Verdict delegation
- **`organizationExpFlags`** (Modern) → Uses `enabledFlags` with Verdict delegation

NOTE That Business Platform doesn't dynamically retrieve ExP flags, so you'll still need to submit a PR to `organizations` [flag seeds](https://github.com/shop/world/blob/main/areas/platforms/organizations/db/data/verdict/flags.yml#L28) to track Verdict flags there.

Templates can now specify either or both flag types:

```typescript
// Legacy templates (existing)
{organizationBetaFlags: ['oldFlag']}

// Modern templates (new - uses short hash instead of flag name)
{organizationExpFlags: ['2d5e9acc']}

// Transition templates (requires both)
{
  organizationBetaFlags: ['oldFlag'],
  organizationExpFlags: ['2d5e9acc']
}
```

**Benefits:**
- ✅ **Backward compatible** - Existing beta flags continue working unchanged
- ✅ **Forward compatible** - New templates can use Verdict-delegated flags. We're also using hashed flags which leak less internal information 👍🏼 
- ✅ **Gradual migration** - Teams can migrate at their own pace
- ✅ **No breaking changes** - All existing functionality preserved
- ✅ **Performance optimized** - Both flag types fetched in parallel

#### Test with a custom templates.json file

I used [this flag](https://experiments.shopify.com/flags/f_block_new_merchant_custom_apps) which is already rolled out to test. You want to use the short flag hash in `templates.json`.

```bash
# Download the production templates as a base
curl https://cdn.shopify.com/static/cli/extensions/templates.json > /tmp/test-templates.json

# Edit /tmp/test-templates.json and update an existing template with a flag to match:
{
  "identifier": "whatever",
  "name": "Whatever",
  "organizationExpFlags": ["2d5e9acc"],
  //etc.
}
```

```bash
SHOPIFY_CLI_APP_TEMPLATES_JSON_PATH=/tmp/test-templates.json \
  shopify app generate extension
```

You should see whichever template you've enabled by the flag show up!

### Measuring impact

### Checklist

- [x] I've added comprehensive test coverage for the new functionality

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -14384,15 +14384,6 @@ snapshots:
       msw: 2.8.7(@types/node@24.7.0)(typescript@5.8.3)
       vite: 6.4.1(@types/node@18.19.70)(jiti@2.4.2)(sass@1.89.1)(yaml@2.8.1)
 
-  '@vitest/mocker@3.2.1(msw@2.8.7(@types/node@24.7.0)(typescript@5.8.3))(vite@6.4.1(@types/node@24.7.0)(jiti@2.4.2)(sass@1.89.1)(yaml@2.8.1))':
-    dependencies:
-      '@vitest/spy': 3.2.1
-      estree-walker: 3.0.3
-      magic-string: 0.30.17
-    optionalDependencies:
-      msw: 2.8.7(@types/node@24.7.0)(typescript@5.8.3)
-      vite: 6.4.1(@types/node@24.7.0)(jiti@2.4.2)(sass@1.89.1)(yaml@2.8.1)
-
   '@vitest/pretty-format@3.2.1':
     dependencies:
       tinyrainbow: 2.0.0
@@ -19992,7 +19983,7 @@ snapshots:
     dependencies:
       '@types/chai': 5.2.2
       '@vitest/expect': 3.2.1
-      '@vitest/mocker': 3.2.1(msw@2.8.7(@types/node@24.7.0)(typescript@5.8.3))(vite@6.4.1(@types/node@24.7.0)(jiti@2.4.2)(sass@1.89.1)(yaml@2.8.1))
+      '@vitest/mocker': 3.2.1(msw@2.8.7(@types/node@24.7.0)(typescript@5.8.3))(vite@6.4.1(@types/node@18.19.70)(jiti@2.4.2)(sass@1.89.1)(yaml@2.8.1))
       '@vitest/pretty-format': 3.2.1
       '@vitest/runner': 3.2.1
       '@vitest/snapshot': 3.2.1


dmerand:
* **#6758** <a href="https://app.graphite.com/github/pr/Shopify/cli/6758?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Shopify/cli/6758?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

pnpm-lock.yaml
gonzaloriestra, line 1: I guess it should be fine, but this wasn't intended, no?

gonzaloriestra (commented):
Thanks for taking a look at this! It looks good, but I have an alternative proposal: what about directly switching to the new system?

If I'm not wrong, it would require updating 8 templates [here](https://cdn.shopify.com/static/cli/extensions/templates.json) that include `organizationBetaFlags` to add `deprecatedFromCliVersion: X` , and add them again, but with the new encoded flags and `minimumCliVersion: X`.

That way, we can switch here to only fetch the new flags. Simpler code, less confusion and we can try to deprecate the old flags sooner. WDYT?

kbav:
> I have an alternative proposal: what about directly switching to the new system?

While I like this idea — it's a simpler mental model — I think this would require a fair amount of work upfront to prepare for.

For example:
The "Order routing location rule" function template. It has `"organizationBetaFlags": ["order_routing_extensibility_partner"]`. However, that flag is [not in Verdict/ExP](https://experiments.shopify.com/search?q=order_routing_extensibility_partner&type=Flag%2CExperiment%2CMetric%2CBandit). It's a BP flag, that a user manually assigns at `https://business-platform.shopifycloud.com/internal/organizations/{YOUR_ORG_ID_HERE}?tab=feature_flags`.

If I understand it right, the alternative proposal would require migrating BP flags over to ExP... but we'd still have to keep the old BP flags around as well for the sake of old CLI clients. And now I'm wondering if the two different cloned flags would essentially become out-of-sync (two different systems tracking rollout % and manual assignments).

I could be way off of this, but if I'm not, we may _have_ to have two systems until we can entirely get rid of BP flags.

dmerand:
I do like the idea of switching to the new system, and [wrote up what it will take here](https://docs.google.com/document/d/1ApIELHXU2mZlkpXMxmzwhutKTSwK5n6c-zJDXcNeCMw/edit?tab=t.0). In my mind this is the first step in either case: make both systems workable. 

All new app module gating should happen through ExP, and these app module types gated through the legacy setup can roll out to 100% and naturally "die off" at that point. ExP is easier to use, so it shouldn't be a hard sell.

I do worry that some of the existing gating is being treated as a "program membership" and therefore might end up being longer-lived. For example, some of those payment extensions already have more than 500 flag activations :/

kbav:
@dmerand @gonzaloriestra Update: it's looking like this (or some other way for CLI to check ExP flags) will be blocking the addition of the `admin.app.tools.data` extension template ([PR](https://github.com/Shopify/extensions-templates/pull/358)).

I checked with our team, and we do want for this new extension template to be gated by a controlled rollout. And it looks like legacy BP flags are a dead end for us (documentation says "[DO NOT USE](https://github.com/shop/world/blob/main/areas/platforms/organizations/documentation/guides/feature_flags.md#feature-flags---deprecated-do-not-use)" and it seems [the old maintenance task used for creating these flags](https://github.com/shop/world/pull/67415#:~:text=How%20I%20top%20hatted%20the%20flag%20part) is [gone](https://business-platform.myshopify.io/maintenance_tasks/tasks/Maintenance::CreateFeatureFlagTask)).

gonzaloriestra:
I hadn't realized about the problem of out-of-sync flags, which is not so easy to solve... I have another idea:
- Migrate the flags and the templates
- Remove the old flags
- Add a warning in the `extension generate` command for the old versions (using the notification system) with something like: "If you want to create any of these gated extensions, please upgrade the CLI first: X, Y, Z."

Anyway, I agree to do this as a first step if there's a plan to fully migrate.

PR #6760, opened 2026-01-14
Remove `form-data` package from `cli-kit` entirely

jordanverasamy:
## Summary

Context: https://github.com/Shopify/cli/pull/6756#discussion_r2691988658

Following up on #6756 which switched bulk uploads to use native `FormData`, this removes the `form-data` package entirely from `cli-kit`:

- Remove `form-data` package from cli-kit dependencies (reduces bundle size)
- Remove `formData()` export and FormData import from `http.ts`
- Remove `formData` test from `http.test.ts`
- Simplify `uploadToGCS` to not use `FormData` (it was being used incorrectly anyway - creating a `FormData`, appending data, but then sending raw buffer as the body with multipart headers)
- Replace `Headers` type import from form-data with a minimal `ResponseHeaders` interface that works with both `node-fetch` and `graphql-request`

## Tophatting

[wip]

### DIFF
--- a/packages/app/src/cli/services/bundle.ts
+++ b/packages/app/src/cli/services/bundle.ts
@@ -4,7 +4,7 @@ import {AssetUrlSchema, DeveloperPlatformClient} from '../utilities/developer-pl
 import {MinimalAppIdentifiers} from '../models/organization.js'
 import {joinPath} from '@shopify/cli-kit/node/path'
 import {brotliCompress, zip} from '@shopify/cli-kit/node/archiver'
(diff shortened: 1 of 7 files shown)


jordanverasamy:
Aw, turns out we can't remove `form-data` from `cli-kit` yet.

`@shopify/oxygen-cli` still imports `formData` from `@shopify/cli-kit/node/http`. (See [here](https://github.com/Shopify/oxygen-cli/blob/main/src/deploy/upload-files.ts#L4).)

We'd need to update `oxygen-cli` to use native `FormData` first, then revisit this. 😭

Issue #947, opened 2022-12-16
Remove Checkbox functionality


byrichardpowell:
- [x] Remove Checkbox in left column
- [x] Remove Actions in Table head 
- [ ] Match left and right padding on table to designs

PR #6794, opened 2026-01-23
fix SSRF vulnerability in proxy.ts

FKauwe:

Fixes [#947](https://github.com/Shopify/developer-tools-team/issues/947#issuecomment-3772789077) 

Fixes the vulnerability when proxying by introducing a check where if the two host names don't match, the request is not executed and a 502 error is surfaced to the user. 
I could have identified hosts preceded by `// `and then performed the matching logic or even tried to correct the incorrect host but we (I and Josh) chose to throw an error because it's simpler to do, and the alternate method of rewriting the `//` path would involve regexes, and we don't want to try loading these requests. This change ensures we block the request. We also chose to throw an error instead of outputWarn as this is a rare but real vulnerability and if it happens, we want the surfaced error message to be loud

Baseline behavior:

Malicious request in logs and browser:
Screenshot 2026-01-22 at 3 45 10 PM
Screenshot 2026-01-22 at 3 45 42 PM

### How to test your changes?

`p build` (in editor terminal) on main branch
Start the dev server in an external terminal 

When the server is running, use the `t `shortcut to trigger a legitimate request and see the healthy logs output
To trigger the SSRF detection,  visit the malicious URL directly in your browser:
`http://127.0.0.1:9292//pokeapi.co/api/v2/pokemon/gengar`
and see the browser behavior change

Then switch to my branch to see the fix in action:
before testing the new build, kill the server in the external terminal `ctrl-c`
Then `p build`
Then restart the server again

Then run the malicious request again and see the new browser and log behavior: 
Screenshot 2026-01-22 at 6 47 33 PM
Screenshot 2026-01-22 at 9 00 46 PM
Screenshot 2026-01-22 at 9 05 32 PM
Screenshot 2026-01-22 at 9 06 07 PM

wrote a new unittest to verify the behavior and all proxy tests pass and all storefront-renderer tests pass

###Update
During this PR lifecycle, Chrome released a browser update which changed the behavior of the `//` . the vulnerability of a replacement of primary host by a secondary host following the `// `no longer exists in the most recent version of Chrome (as far as we can tell, the patch was released 1.27.26). This also affects the browser behavior for our fix.  On newer versions of Chrome, users will no longer see the 502 gateway or the terminal output, they will see: 
Screenshot 2026-01-29 at 12 02 23 PM
Screenshot 2026-01-29 at 2 24 01 PM

The fix still protects against a malicious attack, but the browser response looks different.

### DIFF
--- a/packages/theme/src/cli/utilities/theme-environment/proxy.ts
+++ b/packages/theme/src/cli/utilities/theme-environment/proxy.ts
@@ -293,6 +293,13 @@ export function proxyStorefrontRequest(event: H3Event, ctx: DevServerContext): P
   const host = event.path.startsWith(EXTENSION_CDN_PREFIX) ? 'cdn.shopify.com' : ctx.session.storeFqdn
   const url = new URL(path, `https://${host}`)
 
+  // Check that we aren't redirecting to external hosts
+  if (url.hostname !== host) {
+    return Promise.reject(
+      new Error(`Request failed: Hostname mismatch. Expected host: ${host}. Resulting URL hostname: ${url.hostname}`),
+    )
+  }
+
   // When a .css.liquid or .js.liquid file is requested but it doesn't exist in SFR,
   // it will be rendered with a query string like `assets/file.css?1234`.
   // For some reason, after refreshing, this rendered URL keeps the wrong `?1234`
--- a/.changeset/rich-swans-draw.md
+++ b/.changeset/rich-swans-draw.md
@@ -1,0 +1,5 @@
+---
+'@shopify/theme': patch
+---
+
+Protect SSRF vulnerability in proxy requests when hosts don't match
--- a/packages/theme/src/cli/utilities/theme-environment/proxy.test.ts
+++ b/packages/theme/src/cli/utilities/theme-environment/proxy.test.ts
@@ -1,4 +1,10 @@
-import {canProxyRequest, getProxyStorefrontHeaders, injectCdnProxy, patchRenderingResponse} from './proxy.js'
+import {
+  canProxyRequest,
+  getProxyStorefrontHeaders,
+  injectCdnProxy,
+  patchRenderingResponse,
+  proxyStorefrontRequest,
+} from './proxy.js'
 import {describe, test, expect} from 'vitest'
 import {createEvent} from 'h3'
 import {IncomingMessage, ServerResponse} from 'node:http'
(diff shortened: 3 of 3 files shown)


packages/theme/src/cli/utilities/theme-environment/storefront-renderer.ts
```diff
@@ -117,6 +117,7 @@ export function buildCookies(session: DevServerSession, ctx: Pick<DevServerRende
 function buildStorefrontUrl(session: DevServerSession, {path, sectionId, appBlockId, query}: DevServerRenderContext) {
   const baseUrl = buildBaseStorefrontUrl(session)
   const url = `${baseUrl}${path}`
+
```
EvilGenius13: We can remove the newline here 😄 (prob left over from logging)

packages/theme/src/cli/utilities/theme-environment/proxy.ts
```diff
@@ -293,6 +293,11 @@ export function proxyStorefrontRequest(event: H3Event, ctx: DevServerContext): P
   const host = event.path.startsWith(EXTENSION_CDN_PREFIX) ? 'cdn.shopify.com' : ctx.session.storeFqdn
   const url = new URL(path, `https://${host}`)
 
+  // Check that we aren't redirecting to external hosts
+  if (url.hostname !== host) {
+    return Promise.reject(new Error('Request failed: Hostname mismatch'))
```
EvilGenius13: I think it would be nice if we could show the mismatched hosts in the error so it's a bit more clear what happened (`host` vs `url.hostname`)

FKauwe:
@EvilGenius13 pointed out there is conditional branching logic in the `proxyStorefrontRequest` function when the host is being set by the `event.path` object and I was only testing one path. 
I understood the need to have test coverage for both branches of the conditional logic but didn't understand exactly what the conditions were or how the urls were being built based on the presence or absence of the prefix. I dug into this, here are my learnings:
`createH3Event()` is a function in the `proxy.ts.test` file that creates a fake HTTP request event. The second argument becomes `event.path` — that's the URL path someone is requesting and `event.path` is used in the `proxyStorefrontRequest` function.

These are the questions I identified I need to answer in my investigation:
what is the difference between this: `createH3Event('GET', '//evil.com/some-path')`
and this: `createH3Event('GET', '/ext/cdn//evil.com/some-path')`
and how does that relate to: `const host = event.path.startsWith(EXTENSION_CDN_PREFIX) ? 'cdn.shopify.com' : ctx.session.storeFqdn`

And the answers I came up:
The second one: `'/ext/cdn//[evil.com/some-path](http://evil.com/some-path)' ` since it is a different path, will force an execution of the other ternary branching logic where the condition: `event.path.startsWith(EXTENSION_CDN_PREFIX)` is found to be TRUE
Both paths will go through the workflow of the function and two different urls will be created, which are: 
```
new URL('//evil.com/some-path', 'https://my-store.myshopify.com')
Result URL: https://evil.com/some-path
url.hostname = 'evil.com'
```
(for the path without the ext)
And:
```
new URL('//evil.com/some-path', 'https://cdn.shopify.com')
Result URL: https://evil.com/some-path
url.hostname = 'evil.com'
For the path *with the ext
```
Both paths will ultimately fail in the new check that I wrote, since in both cases, the hostname will be `‘evil.com’` which will not match the respective hosts of the urls, which are `my-store.myshopify.com` , `cdn.shopify.com` but the mismatches will be caused by two different hosts
I added the new test to cover both hosts being created by the conditional logic and I renamed them to make it clear that the prefix name and the host mismatching are being tested.

PR #6797, opened 2026-01-26
make remote dev console the default case

ryancbahan:

Addresses https://github.com/shop/issues-develop/issues/21458 

Makes remote dev console the default, only doing local console for 1p devs.

Replaces the optional skip logic with opt-in logic for the local dev console.

### How to test your changes?

`CLIP_1P_DEV=1 pnpm shopify app dev`

### Measuring impact

### Checklist


craigmichaelmartin:
Rather than invert `skipLocalDevConsole` I think we should just remove it. If you are a first party app, you see the local dev console; otherwise you skip past it. Once first party apps are migrated, we can release a patch removing this conditional entirely.

Alternatively, we could put this behind an organization-scoped ExP flag. That way we have more control with the timing as well, and not just that the new dev console work must be finished before the next CLI version which includes this change (the new dev console can work either way - with or without this change - but this change requires the new dev console). If we went this route we could also handle supporting first party apps when the time comes, without a new release. The problem with this though is the whole time thing. Both because first party apps may not be migrated within the life of a flag, and also because the flag platform returns false for deleted flags, which I imagine is a problem for support of old CLI versions.

PR #6798, opened 2026-01-28
Add `doctor` command

EvilGenius13:
                                                                                                                                                                                                                                                                                                                                                                    
We need an automated way to validate that CLI commands work correctly after changes or new releases. Currently, there's no easy way to:                                                             
  - Test commands end-to-end before releasing to npm                                                                                                                                                        
  - Catch regressions when changes affect command behavior                                                                                                                                                  
  - Verify that a fresh install works as expected                                                                                                                                                  
  - Run QA workflows consistently across different scenarios                                                                                                                                                
                                                                                                                                                                                                            
This doctor command provides the foundation for automated command testing.                                                                                                                                 
                                                                                                                                                          
Introduces the `shopify doctor` command with initial theme command tests:                                                                                                                                  
                                                                                                                                                                                                            Core Framework:                                                                                                                                                                                           - Adds shopify doctor theme command for testing theme-related commands                                                                                                                                     - Implements test framework with suites, assertions, and context management                                                                                                                                                                                                                                                   - Sets up structure for future expansion (app commands)                                                                                                                             
                                                                                                                                                                                                            Initial Tests:                                                                                                                                                                                            - We are testing `theme init` and `theme push` commands. Follow up PR's will add the remaining commands and update existing tests and assertions.                                                                                                                                               
                                                                                                                                                                                                            Note:                                                                                                                                                                                             - Tests run in order. If a test relies on another command, that command should run first (Ex: `theme init` to create a theme and then running `theme push`.                                                                                                                                                               

### How to test your changes?

Pull down the branch
Build the branch
Run `shopify doctor theme -e <your env>` in a folder that has a `shopify.theme.toml` file.

- Check to see a new theme folder is initialized in the folder you ran the doctor command
- Check your store to see if that same theme was pushed to your store

### Post-release steps

Much like the CLI multi-env command work, the doctor command will be built up over many additional PR's.

### Measuring impact

### Checklist


packages/cli/src/cli/services/audit/framework.ts
```diff
@@ -0,0 +1,398 @@
...
+   *
+   * @example
+   * const result = await this.run('shopify theme init my-theme')
+   * const result = await this.run('shopify theme push --json')
+   */
+  protected async run(
+    command: string,
+    options?: {cwd?: string; env?: {[key: string]: string}},
+  ): Promise<CommandResult> {
+    const parts = command.split(' ').filter(Boolean)
```
karreiro: It seems like we have an issue here for flags with spaces. For example:

```ts
const command = 'shopify theme push --theme "My Theme Name"';
const parts = command.split(' ').filter(Boolean);
// Expected: ['shopify', 'theme', 'push', '--theme', 'My Theme Name']
// Actual:   ['shopify', 'theme', 'push', '--theme', '"My', 'Theme', 'Name"']
```

In the `packages/cli/src/cli/services/audit/theme/tests/init.ts` file, tests stop working if the API user updates the theme name to this:

```ts
this.themeName = `My audit-theme-${getRandomName('creative')}`
```

Since the theme name can include spaces, the current `split(' ')` parsing ends up breaking the `--theme` value into multiple parts.

packages/cli/src/cli/services/audit/framework.ts
```diff
@@ -0,0 +1,398 @@
...
+
+    let stdout = ''
+    let stderr = ''
+    let exitCode = 0
+
+    try {
+      stdout = await captureOutput(cmd, args, {cwd, env: options?.env})
+      // eslint-disable-next-line no-catch-all/no-catch-all
+    } catch (error) {
+      // captureOutput throws on non-zero exit
```
karreiro: It doesn't seems like we're showing enough debugging information for API users, so it's a bit hard to act on the fixes:

Image

I wonder if it's a problem in the test code or in the infrastructure of the testing framework. What do you think, @EvilGenius13?

packages/cli/src/cli/services/audit/theme/tests/init.ts
```diff
@@ -0,0 +1,57 @@
...
+/**
+ * Tests for `shopify theme init` command
+ */
+export default class ThemeInitTests extends AuditSuite {
+  static description = 'Tests the theme init command creates a valid theme structure'
+
+  private themeName = ''
+  private themePath = ''
+
+  // eslint-disable-next-line @typescript-eslint/naming-convention
```
karreiro: I don't think we should promote the pattern of always eslint-ignoring all test cases.

We may solve this by adopting a pattern like this for test definition:

```js
test('test init creates theme directory', async () => {
  …
})
```

Or, we may ignore this warning globally in this module (still, I personally prefer the `test(...)` API as it follows a more well-established convention when it comes to unit tests in JS/TS frameworks).

packages/cli/src/cli/services/audit/framework.ts
karreiro: Actually, I don't think we need to cover this comment.

It's a responsibility from the test author to provide the optional message string with context

packages/cli/src/cli/services/audit/framework.ts
```diff
@@ -0,0 +1,398 @@
...
+      if (error instanceof Error) {
+        stderr = error.message
+        // Try to extract exit code from error
+        const match = /exit code (\d+)/i.exec(error.message)
```
karreiro: It seems a bit fragile to extract the error code based on the output string.

The `captureOutput` function calls `buildExec` that results the output and the exit code.

I wonder if we may update `packages/cli-kit/src/public/node/system.ts` with a new function that returns the output and the exit status code.

packages/cli/src/cli/services/audit/framework.ts
EvilGenius13: Fixed parsing to gather themes with spaces.

packages/cli/src/cli/services/audit/theme/tests/init.ts
EvilGenius13: switched to the test method!

packages/cli/src/cli/services/audit/framework.ts
EvilGenius13: Fixed.

packages/cli/src/cli/commands/doctor-release/doctor-release.ts
```diff
@@ -0,0 +1,21 @@
...
+    renderInfo({
+      headline: 'Shopify CLI Doctor.',
+      body: [
+        'Available doctor commands:',
+        '',
+        '  shopify doctor theme -e <environment>   Run all theme command tests',
+        '',
+        'The -e/--environment flag is required to specify the store configuration.',
+        'Use --help with any command for more options.',
+      ],
```
karreiro, line 23: It doesn't seem like the output here is formatted as expected (we can use `TokenItem` to break lines and highlight commands):

Image

Image

(we may handle this in follow-up PR)

packages/cli/src/cli/services/doctor/theme/tests/init.ts
```diff
@@ -0,0 +1,55 @@
...
+      this.context.themePath = this.themePath
+    })
+
+    this.test('essential theme files exist', async () => {
+      const essentialFiles = ['layout/theme.liquid', 'config/settings_schema.json', 'templates/index.json']
+
+      for (const file of essentialFiles) {
+        // eslint-disable-next-line no-await-in-loop
+        await this.assertFile(joinPath(this.themePath, file))
+      }
```
karreiro: May we resolve promises here to avoid the the eslint ignore?

packages/cli/src/cli/services/doctor/theme/tests/init.ts
```diff
@@ -0,0 +1,55 @@
...
+    })
+
+    this.test('theme directories exist', async () => {
+      const directories = ['sections', 'snippets', 'assets', 'locales']
+
+      for (const dir of directories) {
+        // eslint-disable-next-line no-await-in-loop
```
karreiro: May we resolve promises here too?

packages/cli/src/cli/services/doctor-release/theme/tests/push.ts
```diff
@@ -0,0 +1,62 @@
...
+
+      if (this.context.environment) {
+        cmd += ` -e ${this.context.environment}`
+      }
+      if (this.context.store) {
+        cmd += ` -s ${this.context.store}`
+      }
+      if (this.context.password) {
+        cmd += ` --password ${this.context.password}`
+      }
```
karreiro, line 42: Given the framework requires the `-e` flag to run the test suite, and that this is something that's going to happen with all commands that require authentication, I believe we should try to extract this early to another place.

(we may apply this in a follow-up PR tho)

packages/cli/src/cli/commands/doctor-release/doctor-release.ts
EvilGenius13: Yes absolutely, will handle in a follow up PR.

packages/cli/src/cli/services/doctor/theme/tests/init.ts
EvilGenius13: Originally I had them this way in case of consistent failures, the results would come back in order (such as missing files) but considering this is handling the order of a missing file or folder on an error, I switched to `Promise.all`

packages/cli/src/cli/services/doctor/theme/tests/init.ts
EvilGenius13: Same as above comment

packages/cli/src/cli/services/doctor-release/theme/tests/push.ts
EvilGenius13: I can add this in a follow up.

EvilGenius13:
Added an extra commit changing how we handle filepaths in the command as it would fail on windows. The latest commit is working cross OS
image

nickwesselman:
hmm, should we consider some command naming that is more aligned with its internal use, even though it is hidden? People do discover these things and I'm not sure we want people running our tests?

Any additional knee-high walls? Requiring 1P login, or setting an environment variable for the tests to execute?

EvilGenius13:
> hmm, should we consider some command naming that is more aligned with its internal use, even though it is hidden? People do discover these things and I'm not sure we want people running our tests?
> 
> Any additional knee-high walls? Requiring 1P login, or setting an environment variable for the tests to execute?

Absolutely. We will rename to `doctor-release` and set behind an ENV var.

packages/cli-kit/src/public/node/system.ts
```diff
@@ -57,6 +65,165 @@ export async function captureOutput(command: string, args: string[], options?: E
...
+  }
+}
+
+/**
+ * Runs a command string asynchronously (parses command and arguments from the string).
+ *
+ * @param command - Full command string to be executed (e.g., 'ls -la "my folder"').
+ * @param options - Optional settings for how to run the command.
+ */
+export async function execCommand(command: string, options?: ExecOptions): Promise<void> {
```
gonzaloriestra, line 201: I wonder why we didn't do this from the beginning with `exec` 😄

packages/cli-kit/src/public/node/doctor/framework.ts
gonzaloriestra, line 1: As there are other CLIs building on top of cli-kit (like hydrogen), what about moving these files there so they can use it?

packages/cli/src/cli/commands/doctor-release/doctor-release.ts
```diff
@@ -0,0 +1,26 @@
+import Command from '@shopify/cli-kit/node/base-command'
+import {firstPartyDev} from '@shopify/cli-kit/node/context/local'
+import {renderInfo} from '@shopify/cli-kit/node/ui'
+
+export default class DoctorRelease extends Command {
+  static description = 'Run CLI doctor-release tests'
+  static hidden = true
+
+  async run(): Promise<void> {
+    if (!firstPartyDev()) {
```
gonzaloriestra: Why this?

packages/cli-kit/src/public/node/doctor/framework.ts
EvilGenius13: Let me look into doing that! This might end up being a follow up PR.

packages/cli/src/cli/commands/doctor-release/doctor-release.ts
EvilGenius13: For now it's an added piece of security. Even though the command is hidden, in the event someone stumbled on the command, they would still need to run with the additional environment variable just to make sure they were intentional about using the command. (It's mostly to protect anyone as the command will start to run more commands that will change remote themes in a person's shop (like running theme delete, theme dev etc)

gonzaloriestra: What about checking `isShopify()` instead? Or alternatively, creating a new env var for this (they are free!)

EvilGenius13: I went with a new env var `SHOPIFY_CLI_DOCTOR`

packages/cli-kit/src/public/node/doctor/framework.ts
EvilGenius13: I take it back, I'll move it before merging.

PR #6813, opened 2026-02-02
remove `reverse: true` in `listBulkOperations`

jordanverasamy:

This was actually the intended sort order all along. See the [brief], where we show newest-first sorting: (https://docs.google.com/document/d/157CvTbzxv5RDkAJCCeZMcWEPrsGLEdkFc8Hd48fUSM8/edit?usp=sharing):

image

I messed this up when initially implementing -- I'd assumed the default sort would be oldest first, so we'd need to reverse it. Turns out [the default sort is newest first](https://github.com/shop/world/blob/main/areas/core/shopify/components/platform/bulk_operations/app/models/pagination/bulk_operations_query.rb#L10), so we don't! (We didn't realize the issue before because without the auth fixes, operations would stop being visible to this command fairly quickly.)

Let's just... not reverse the sort anymore!

### How to test your changes?

Run `app bulk status` now and verify that newest are at the top:

image


ericlee878 (approved):
🎩 successfully! LGTM!!

PR #6827, opened 2026-02-05
Fix custom-oclif-loader.ts hydrogen loader strategy for hydrogen-cli local development

andguy95:

During local development of the `@shopify/cli` in the [`Shopify/hydrogen`](https://github.com/Shopify/hydrogen) you are no longer able to see your changes when running the `npx shopify hydrogen` commands because the `custom-ocliff-loader.ts` no longer loads the local `Shopify/hydrgen` cli in the `Shopify/hydrogen` monorepo.

These versions below are good:
- 3.80.6 ✅
- 3.80.7 ✅
- 3.81.0 ✅
- 3.81.1 ✅
- 3.81.2 ✅
- 3.82.0 ✅
- 3.82.1 ✅
- 3.83.0 ❌ - Looks like the hydrogen mono repo awareness stopped working here (looks like major bump of OCLIF)

In OCLIF v4, `determinePriority` is a standalone utility function (not an instance method), so we can't override it to customize command priority  in `custom-ocliff-loader.ts`.  

Change the `custom-ocliff-loader.ts` to manually replace bundled hydrogen commands with external ones after loading completes if in dev mode.

### How to test your changes?

1. Go to `dev.ts` in `hydrogen/packages/cli/src/commands/hydrogen/dev.ts`
2. Make arbitrary update to the `run()`
3. In the `cli` project run `npm run build` 
4. In the `template/skeleton` project run `npx shopify hydrogen dev`
    - To test locally i created an alias `shopify-dev` pointing to a locally built version of the `Shopiy/cli`
  
```
# Create the following file in ~/bin/shopify:

#!/usr/bin/env node

process.env.SHOPIFY_RUBY_BINDIR = "/opt/rubies/3.1.2/bin"
process.env.SHOPIFY_HOMEBREW_FORMULA = "shopify-cli"

import("/Users/<your-user>/src/github.com/Shopify/cli/packages/cli/bin/run.js")
```

`alias shopify-dev='/Users/$USER/bin/shopify'`

### Measuring impact

### Checklist

### DIFF
--- a/packages/app/src/cli/api/graphql/business-platform-organizations/generated/types.d.ts
+++ b/packages/app/src/cli/api/graphql/business-platform-organizations/generated/types.d.ts
@@ -86,6 +86,7 @@ export type OrganizationUserProvisionShopAccessInput = {
 export type Store =
   | 'APP_DEVELOPMENT'
   | 'CLIENT_TRANSFER'
+  | 'COLLABORATOR'
   | 'DEVELOPMENT'
   | 'DEVELOPMENT_SUPERSET'
   | 'PRODUCTION';


EvilGenius13 (approved):

EvilGenius13:
Changes make sense, thank you for the pairing @andguy95. We'll want to make sure @Shopify/app-inner-loop takes a peek as they have historically made the changes to the file.

packages/cli-kit/src/public/node/custom-oclif-loader.ts
```diff
@@ -25,69 +26,36 @@ export class ShopifyConfig extends Config {
           core: ['@shopify/cli-hydrogen'],
           path,
         }
+        // Force OCLIF to ignore manifests so commands are loaded dynamically
+        // to be replaced later
+        options.ignoreManifest = true
```
kdaviduik: I'm not an expert here but several Claude Opus agents independently flagged this as dead code with a misleading comment

packages/cli-kit/src/public/node/custom-oclif-loader.ts
```diff
@@ -2,7 +2,7 @@ import {fileExistsSync} from './fs.js'
 import {cwd, joinPath, sniffForPath} from './path.js'
 import {isDevelopment} from './context/local.js'
 import {execaSync} from 'execa'
-import {Command, Config} from '@oclif/core'
+import {Config} from '@oclif/core'
```
kdaviduik, line 5: (threading)

I think we should add tests here too. Otherwise we aren't fixing the *true* root cause, which is regressions in this file can break this functionality for the Hydrogen CLI, and no tests or CI checks will fail as a result. 

I think we should add at minimum:
- A test verifying external hydrogen commands replace bundled ones after load()
- A test verifying non-dev mode does NOT trigger replacement
- A test that would catch future oclif API changes

kdaviduik, line 5: (threading)

The PR only replaces commands by command.id. If hydrogen commands have aliases registered in _commands, those alias entries still point to bundled versions. There are currently no hydrogen command aliases (checked all 26 commands in oclif.manifest.json), so this is not a bug today but it's a latent issue.

I would recommend adding a comment documenting this limitation, or handling aliases proactively, but there might be a better way to do this overall that avoids this issue:

oclif's own insertLegacyPlugins method uses a delete-then-loadCommands pattern. I think using this.loadCommands(externalHydrogenPlugin) after deleting bundled commands would be a bit cleaner since it handles aliases and permutations through oclif's own code, though both options would involve the `ts-expect-error` from accessing internal properties

packages/cli-kit/src/public/node/custom-oclif-loader.ts
```diff
@@ -25,69 +26,36 @@ export class ShopifyConfig extends Config {
...
+      // commands with external ones after loading completes.
+      const externalHydrogenPlugin = Array.from(this.plugins.values()).find(
+        (plugin) => plugin.name === '@shopify/cli-hydrogen' && !plugin.isRoot,
+      )
+
+      if (externalHydrogenPlugin) {
+        for (const command of externalHydrogenPlugin.commands) {
+          if (command.id.startsWith('hydrogen')) {
+            // @ts-expect-error: _commands is private but we need to replace bundled commands
+            if (this._commands.has(command.id)) {
```
kdaviduik: I think we should add in something like:

```
if (typeof (this as Record<string, unknown>)._commands === 'undefined') {
    throw new Error('ShopifyConfig: oclif internals changed. _commands is no longer available.')
  }
```

since `_commands` is a private property that Oclif may change in patch/minor versions. We don't want this to silently do nothing in that situation

packages/cli-kit/src/public/node/custom-oclif-loader.ts
```diff
@@ -25,69 +26,36 @@ export class ShopifyConfig extends Config {
...
+    if (isDevelopment()) {
+      // Let OCLIF load all commands first, then manually replace bundled hydrogen
+      // commands with external ones after loading completes.
+      const externalHydrogenPlugin = Array.from(this.plugins.values()).find(
+        (plugin) => plugin.name === '@shopify/cli-hydrogen' && !plugin.isRoot,
+      )
+
+      if (externalHydrogenPlugin) {
+        for (const command of externalHydrogenPlugin.commands) {
+          if (command.id.startsWith('hydrogen')) {
```
kdaviduik: overall this logic has 5 layers of nesting. I think we can cut this down a bit with early returns to make this code cleaner :) 

https://www.youtube.com/watch?v=CFRhGnuXG-4

packages/cli-kit/src/public/node/custom-oclif-loader.ts
andguy95: @kdaviduik I poked into this a little. But it looks like this is needed for the `load` function below. It forces oclif to skip the cache and load commands dynamically oclif _might_ fall back to the bundled plugin's manifest and never use the live local version during dev

packages/cli-kit/src/public/node/custom-oclif-loader.ts
andguy95: Updated the logic to match the delete and load pattern found in the insert legacy plugin. This should also account for the aliasing!

packages/cli-kit/src/public/node/custom-oclif-loader.ts
```diff
@@ -25,69 +26,51 @@ export class ShopifyConfig extends Config {
...
-      // if b is a jit plugin and a is not sort a first
-      if (bCommand.pluginType === 'jit' && aCommand.pluginType !== 'jit') {
-        return -1
+    // Delete all bundled hydrogen command entries (canonical IDs, aliases, and hidden aliases)
+    // before reloading from the external plugin. This mirrors oclif's own insertLegacyPlugins
+    // pattern and ensures alias entries don't continue pointing to the bundled version.
+    for (const command of externalHydrogenPlugin.commands) {
+      if (!command.id.startsWith('hydrogen')) continue
+      internalCommands.delete(command.id)
+      const allAliases = [...(command.aliases ?? []), ...((command as {hiddenAliases?: string[]}).hiddenAliases ?? [])]
```
andguy95: Here is the code where we delete the aliased commands so we don't have a stale/cached version if aliased

packages/cli-kit/src/public/node/custom-oclif-loader.test.ts
```diff
@@ -0,0 +1,287 @@
...
+        'ShopifyConfig: oclif internals changed. _commands is no longer available.',
+      )
+    })
+  })
+
+  // These tests use the REAL @oclif/core (via vi.importActual) so they will fail
+  // if oclif removes or renames the private APIs that ShopifyConfig depends on.
+  // The mock above intentionally replaces oclif for logic isolation; this block
+  // provides the missing contract check against the installed package version.
+  describe('oclif API contract', () => {
```
andguy95, line 301: Added test to check for possible regression: 
- `loadCommands` needs to be on the config 
- `_commands` needs to be on the config
  - This will also be caught in the error check in the actual loader

packages/cli-kit/src/public/node/custom-oclif-loader.test.ts
```diff
@@ -0,0 +1,287 @@
...
+      process.env.IGNORE_HYDROGEN_MONOREPO = '1'
+
+      const _config = new ShopifyConfig({root: '/workspace'} as Options)
+
+      expect(execaSync).not.toHaveBeenCalled()
+    })
+  })
+
+  describe('load()', () => {
+    test('does not replace commands when not in dev mode', async () => {
```
andguy95, line 161: A test verifying non-dev mode does NOT trigger replacement

packages/cli-kit/src/public/node/custom-oclif-loader.test.ts
```diff
@@ -0,0 +1,287 @@
...
+      asMock(config).plugins.set('@shopify/cli-hydrogen', {name: '@shopify/cli-hydrogen', isRoot: true, commands: []})
+      asMock(config)._commands!.set('hydrogen:dev', {bundled: true})
+
+      await config.load()
+
+      expect(asMock(config)._commands!.has('hydrogen:dev')).toBe(true)
+      expect(asMock(config).loadCommandsCalls).toHaveLength(0)
+    })
+
+    test('removes bundled hydrogen commands and reloads from external plugin', async () => {
```
andguy95, line 204: A test verifying external hydrogen commands replace bundled ones after load()

packages/cli-kit/src/public/node/custom-oclif-loader.ts
kdaviduik: Thanks for looking into this!

kdaviduik (commented):
Needs a changeset but other than that LGTM :) 

note about [this](https://github.com/Shopify/cli/pull/6827#issuecomment-3930600353), it seems to be a false alarm as apparently:
- customPriority had zero callers anywhere in the codebase
- customPriority was already non-functional since oclif v3.83.0
- The sole consumer (cli-launcher.ts) already calls config.load() — no caller changes needed

packages/cli-kit/src/public/node/custom-oclif-loader.ts
```diff
@@ -25,69 +26,51 @@ export class ShopifyConfig extends Config {
...
```
kdaviduik: non-blocking: `hiddenAliases` is an unguarded third private API dependency

The type assertion to `{hiddenAliases?: string[]}` accesses an undocumented oclif property. The contract tests verify `_commands` and `loadCommands` on the real Config but do NOT verify `hiddenAliases` on command objects. If oclif renames or removes `hiddenAliases`, the `?? []` fallback silently produces an empty array and hidden aliases for bundled commands remain in `_commands`, causing the bundled version to win when invoked via a hidden alias.

I think adding a third contract test would close this gap:

```typescript
test('Command objects still expose hiddenAliases', async () => {
  const {Config: RealConfig} = await vi.importActual<typeof import('@oclif/core')>('@oclif/core')
  const config = new (RealConfig as new (options: {root: string}) => OclifConfig)({
    root: fileURLToPath(new URL('.', import.meta.url)),
  })
  await config.load()
  const someCommand = Array.from(config.commands)[0]
  if (someCommand) {
    expect('hiddenAliases' in someCommand).toBe(true)
  }
})
```

I'd also suggest extracting the inline type cast to a named interface for readability:

```typescript
interface OclifCommand {
  id: string
  aliases?: string[]
  hiddenAliases?: string[]
}
```

.changeset/pretty-swans-smell.md
```diff
@@ -0,0 +1,5 @@
+---
+'@shopify/cli-kit': patch
+---
+
+Fix hydrogen local dev plugin loader to correctly replace bundled hydrogen commands with external plugin commands by switching to a post-load strategy
```
kdaviduik: blocking: `@shopify/cli-kit` exports `custom-oclif-loader.js` as a public entry point. The type diff shows `customPriority` was removed from the public declaration and `load()` was added. While `customPriority` was effectively dead code in oclif v4 (the framework no longer called it), it was part of the public type declaration.

The changeset is marked as `patch`, which I think is appropriate if there are no external consumers, but I think the changeset body should also mention the `customPriority` removal for transparency, even if it's dead code - consumers upgrading `@shopify/cli-kit` should be aware.

packages/cli-kit/src/public/node/custom-oclif-loader.ts
```diff
@@ -25,69 +26,51 @@ export class ShopifyConfig extends Config {
...
-        return 1
-      }
+    if (!externalHydrogenPlugin) return
 
-      // if a is a core plugin and b is not sort a first
-      if (aCommand.pluginType === 'core' && bCommand.pluginType !== 'core') {
-        return -1
-      }
+    if (typeof (this as unknown as {[key: string]: unknown})._commands === 'undefined') {
+      throw new Error('ShopifyConfig: oclif internals changed. _commands is no longer available.')
```
kdaviduik, line 62: non-blocking: I think we should also check `instanceof Map` here

The `typeof` check catches `_commands` being removed, but if oclif changes `_commands` from `Map` to a plain object or other collection type, the code would crash at `.delete()`. An `instanceof Map` check should throw (same pattern as this guard) rather than silently skipping:

```typescript
// After the typeof check, add:
const internalCommands = this._commands
if (!(internalCommands instanceof Map)) {
  throw new Error('ShopifyConfig: oclif internals changed. _commands is no longer a Map.')
}
```

packages/cli-kit/src/public/node/custom-oclif-loader.ts
```diff
@@ -25,69 +26,51 @@ export class ShopifyConfig extends Config {
           core: ['@shopify/cli-hydrogen'],
           path,
         }
+        // Force OCLIF to ignore manifests so commands are loaded dynamically
+        // to be replaced later
+        options.ignoreManifest = true
```
kdaviduik: non-blocking: I think this comment could be more precise about the tradeoff - this applies to ALL plugins, not just hydrogen:

```suggestion
        // Skip the oclif manifest cache so external plugin commands are loaded
        // from disk rather than from a potentially stale manifest. This has a
        // startup performance cost, but only applies during development.
        options.ignoreManifest = true
```

isaacroldan: This is dangerous change, ignoring the manifest will make the CLI slower at startup on every command, because it needs to discover every command again on each run. Please do a comparison to check how speed is affected with and without this :pray:.  
  
When is this line code executed? will it affect most users or just internal hydrogen development?

isaacroldan:
Is bringing `hydrogen-cli` back to this repo a possible alternative to this change? We are doing hacks to support a weird repo structure :/

packages/cli-kit/src/public/node/custom-oclif-loader.ts
andguy95: @isaacroldan Thank you, this is a good callout. I will test. This ideally should be executed in development mode but only when in the Hydrogen Monorepo. 

In this instance we could probably move the hydrogen monorepo check where the `isDevelopment` check is to ensure this ignoreManifest only runs in the expected context.

packages/cli-kit/src/public/node/custom-oclif-loader.ts
andguy95: Good call out. Updated!

packages/cli-kit/src/public/node/custom-oclif-loader.ts
andguy95: Hey @isaacroldan. Here's what I found:

### Who is affected?
`ignoreManifest` only fires inside `isDevelopment()`, which requires `SHOPIFY_CLI_ENV=development`. The majority of public facing end users shouldn't be affected. The only people who hit this path are CLI developers
running from source with the `SHOPIFY_CLI_ENV` set.

### The gate fix
Even so, the original code was broader than needed, as it would set `ignoreManifest = true` for any dev running from source in any directory with a `package.json` and `SHOPIFY_CLI_ENV=development`. 

I've tightened it so that `ignoreManifest` is now gated behind `useHydrogenMonorepo` (the same
`/(shopify|hydrogen)\/hydrogen/i` path check already used for the `npm` prefix call). It only fires when you're actually inside the Hydrogen monorepo.

### Tests
Either way with and without the gate, there wasn't any noticible perf hit running outside or inside the Hydrogen Monorepo. I used `hyperfine` to run these tests:

image

andguy95:
> Is bringing `hydrogen-cli` back to this repo a possible alternative to this change? We are doing hacks to support a weird repo structure :/

@isaacroldan, there have been some discussions on how we can better manage this strange repo dependencies. We are going to explore patch packages as well as seeing if we can possible change the dependencies of the cli-hydrogen to avoid this hack if possible 🤞

andguy95:
@isaacroldan I put up a draft PR in the hydrogen monorepo (https://github.com/Shopify/hydrogen/pull/3531) for a strategy to offload the logic for local hydrogen monorepo development. 

This custom oclif loader I believe would become simplified to still maintain that logic for hydrogen backwards compatibility as noted in the comment. (But honestly I would love your opinion to see if we could kill this off all together)

isaacroldan:
Love it, I think it makes the most sense for you to own this and be part of the hydrogen repo itself.

About the backwards compatibility, I think the global CLI has been out for 2 years or more, IMHO it's time to remove that part too and basically stop having a custom oclif loader :)

andguy95:
Going to close this in favor of this new strategy

PR #6795, opened 2026-01-23
Fix the default environments infrastructure so it doesn’t fail when running commands that don’t require authentication

karreiro:
### WHY are these changes introduced?

`shopify theme check` currently fails when the default environment is set in `shopify.theme.toml`.

### WHAT is this pull request doing?

This PR updates the multi-env infrastructure so it doesn’t fail when `--store` isn’t included in the list of required flags.

### How to test your changes?

- Go to an empty directory
- Create a `shopify.theme.toml` file
  ```toml
  [environments.default]
  store = "xyz.myshopify.com"
  ```
- Run `shopify theme check`
- Notice it no longer fails

### Post-release steps

N/A

### Measuring impact

How do we know this change was effective? Please choose one:

- [ ] n/a - this doesn't need measurement, e.g. a linting rule or a bug-fix
- [x] Existing analytics will cater for this addition
- [ ] PR includes analytics changes to measure impact

### Checklist

- [x] I've considered possible cross-platform impacts (Mac, Linux, Windows)
- [x] I've considered possible [documentation](https://shopify.dev) changes

PR #6828, opened 2026-02-05
fix issue for undefined sessions in theme commands

EvilGenius13:

Fixes https://github.com/Shopify/developer-tools-team/issues/1024?reload=1
Follow up from the [PR](https://github.com/Shopify/cli/pull/6795) to fix default environments

CLI main is failing on commands that don't need a store flag to run such as `theme profile`. A few other commands are failing with an error `Cannot read properties of undefined` such as `theme init`

Two parter:
1. I have reverted some of the logic from the other PR back to checking if a command needs authorization and not checking against store flags
2. The commands failing with error were missing `static multiEnvironmentsFlags: RequiredFlags = null` which is our way of telling whether these commands support multi env or not. I have added them where needed.

### How to test your changes?
Unfortunately we need to test this in a few configurations to make sure we caught every instance.

Run the commands using regular auth (email / password) login
- `shopify theme dev` Should pass
- `shopify theme list` Should pass
- `shopify theme package` Should pass
- `shopify theme check` Should pass
- `shopify theme profile` Should pass

Run the commands using a default environment (`[environments.default]`) in your `shopify.theme.toml` **Note**: Using the default environment means you do not need to explicitly pass it as an environment.
- `shopify theme dev` Should pass
- `shopify theme list` Should pass
- `shopify theme package` Should pass
- `shopify theme check` Should pass
- `shopify theme profile` Should fail (can only run using regular auth)

Run the commands using an environment variable (-e my_store) from your `shopify.theme.toml`
- `shopify theme dev` Should pass
- `shopify theme list` Should pass
- `shopify theme package` Should fail (`package` doesn't have an environment flag)
- `shopify theme check` Should pass
- `shopify theme profile` Should fail (can only run using regular auth)

Run the commands using two environment variables (-e my_store, -e my_other_store ) from your `shopify.theme.toml`
- `shopify theme dev` Should fail (can't run in multi env mode)
- `shopify theme list` Should pass
I shortened this one since only a few commands run in multi env mode

### Measuring impact

### Checklist

### DIFF
--- a/.changeset/goofy-colts-sink.md
+++ b/.changeset/goofy-colts-sink.md
@@ -1,0 +1,5 @@
+---
+'@shopify/theme': patch
+---
+
+Fix issue where certain theme commands were failing to have sessions created
--- a/packages/cli/oclif.manifest.json
+++ b/packages/cli/oclif.manifest.json
@@ -6326,6 +6326,7 @@
       "hiddenAliases": [
       ],
       "id": "theme:init",
+      "multiEnvironmentsFlags": null,
       "pluginAlias": "@shopify/cli",
       "pluginName": "@shopify/cli",
       "pluginType": "core",
@@ -6363,6 +6364,7 @@
       "hiddenAliases": [
       ],
       "id": "theme:language-server",
+      "multiEnvironmentsFlags": null,
       "pluginAlias": "@shopify/cli",
       "pluginName": "@shopify/cli",
       "pluginType": "core",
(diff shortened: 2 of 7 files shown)


aswamy (approved):
Amazing stuff! 🥇

Tophatted and everything looks great.

✅ Run commands with email/password
✅ Run commands with default environment
✅ Run commands with `-e` flag
✅ Run commands with multiple `-e` flags

Question: `theme check` doesn't need you to be logged in - but does take environment as a flag? What kind of configuration can you have when you pass in an env?

EvilGenius13:
> Question: `theme check` doesn't need you to be logged in - but does take environment as a flag? What kind of configuration can you have when you pass in an env?

It lets you set a path to the theme folder you want to check.

karreiro (approved):
Thank you for this PR, @EvilGenius13! LGTM and works as expected as well 🎩 

I believe we just need to regenerate OCLIF manifests and we will be ready to merge 🚀

PR #6831, opened 2026-02-06
Upgrade Ink to v6 and React to v19

tewson:

Ink v6 was released with React 19 as a peer dependency. Upgrading keeps the CLI on the latest supported versions of both libraries, benefiting from React 19 performance improvements and Ink 6 fixes.

**Dependency bumps:**
- `ink`: 5.2.1 → 6.2.0 (`cli-kit`)
- `react`: ^18.2.0 → ^19.0.0 (`cli-kit`, `app`)
- `@types/react`: ^18.2.0 → ^19.0.0 (`cli-kit`, `app`)
- `react-dom`: 18.3.1 → ^19.0.0 (`app`)
- `@types/react-dom`: ^18.2.0 → ^19.0.0 (`app`)

**Code changes (all in `SelectInput.tsx`):**
- Removed `forwardRef` wrapper and custom module augmentation — `ref` is now a regular prop in React 19
- Updated `inputFixedAreaRef` type from `React.RefObject<DOMElement>` to `React.Ref<DOMElement>` for React 19 compatibility

No other code changes were required. The codebase was already clean of deprecated React 18 patterns (no `defaultProps`, `propTypes`, string refs, legacy context, or implicit `children` on `FunctionComponent`).

### How to test your changes?

1. `pnpm install`
2. `pnpm --filter @shopify/cli-kit type-check` — passes
3. `pnpm --filter @shopify/app type-check` — passes
4. Run the CLI and verify interactive prompts (select, autocomplete, text input) work correctly

### Measuring impact

### Checklist

### DIFF
--- a/packages/cli-kit/src/private/node/ui/components/AutocompletePrompt.test.tsx
+++ b/packages/cli-kit/src/private/node/ui/components/AutocompletePrompt.test.tsx
@@ -448,7 +448,7 @@ describe('AutocompletePrompt', async () => {
       "
     `)
 
-    await sendInputAndWaitForChange(renderInstance, DELETE)
+    await sendInputAndWaitForContent(renderInstance, 'ype to search', DELETE)
 
     expect(renderInstance.lastFrame()).toMatchInlineSnapshot(`
       "?  Associate your project with the org Castile Ventures?   [36m[7mT[27m[2mype to search...[22m[39m
@@ -697,7 +697,7 @@ describe('AutocompletePrompt', async () => {
       "
     `)
 
-    await sendInputAndWaitForChange(renderInstance, DELETE)
+    await sendInputAndWaitForContent(renderInstance, 'ype to search', DELETE)
 
     expect(renderInstance.lastFrame()).toMatchInlineSnapshot(`
       "?  Associate your project with the org Castile Ventures?   [36m[7mT[27m[2mype to search...[22m[39m
--- a/packages/cli-kit/src/private/node/ui/components/TextInput.tsx
+++ b/packages/cli-kit/src/private/node/ui/components/TextInput.tsx
@@ -1,6 +1,6 @@
 /* eslint-disable no-nested-ternary */
 import {shouldDisplayColors} from '../../../../public/node/output.js'
-import React, {useEffect, useState} from 'react'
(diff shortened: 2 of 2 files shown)


tewson:
@gonzaloriestra Thank you for generating the snapshot build. What are the next steps for me? I tested the snapshot with the `shopify theme dev` command and it seemed fine but is there a more thorough test I should do?

gonzaloriestra:
You don't need to do anything else. I just need some time to review and test it a bit. Thanks a lot for taking care of this!

packages/app/src/cli/services/dev/ui/components/Dev.tsx
```diff
@@ -274,6 +274,11 @@ const Dev: FunctionComponent<DevProps> = ({
           {error ? <Text color="red">{error}</Text> : null}
         </Box>
       ) : null}
+      {isAborted && isShuttingDownMessage ? (
+        <Box flexDirection="column">
+          <Text>{isShuttingDownMessage}</Text>
```
gonzaloriestra: Why this? Aren't we showing that in line 259?

Just tested it:
- In production we show `Shutting down dev ...` temporarily, but it disappears once the command actually finishes
- With this change, it does the same, but it doesn't disappear later

Is that intended?

packages/app/src/cli/services/dev/ui/components/DevSessionUI.tsx
```diff
@@ -278,6 +278,11 @@ const DevSessionUI: FunctionComponent<DevSesionUIProps> = ({
           )}
         </Box>
       ) : null}
+      {isAborted && isShuttingDownMessage ? (
+        <Box flexDirection="column">
+          <Text>{isShuttingDownMessage}</Text>
```
gonzaloriestra: Same here

packages/app/src/cli/services/dev/ui/components/Dev.tsx
isaacroldan: Maybe this was added to pass the test that expects the `Shutting down..` string?  
  
I think it should be reverted for the reasons @gonzaloriestra points out.

tewson:
@isaacroldan I've removed it now in the latest commit

PR #6848, opened 2026-02-11
Auto-upgrade

gonzaloriestra:
HackDays project: https://vault.shopify.io/hackdays/153/projects/22255

We want our devs to always use the latest CLI version and avoid fragmentation

Adds a prompt in `shopify upgrade` to ask about enabling auto-update:

24-34-eshor-470hq

Then, it checks once per day if there is a new CLI version available and automatically upgrades unless:
- Disabled
- Local CLI installation
- CI
- Snapshot or nightly version
- Major version change
- Development environment

The way to upgrade is to simply run the appropriated package manager command for the user (like `npm i -g @shopify/cli@latest` for NPM).

### How to test your changes?

Install the snapsot with different package managers and operating systems:
- `npm i -g --@shopify:registry=https://registry.npmjs.org @shopify/cli@0.0.0-snapshot-20260224152342`
- `shopify cache clear` => removes cached version check
- `shopify version` => does nothing because it skips auto-update in snapshots
- `SHOPIFY_CLI_FORCE_AUTO_UPGRADE=1 shopify version` => does nothing because auto-update is disabled
- `shopify upgrade` => enable auto-upgrade
- `npm i -g --@shopify:registry=https://registry.npmjs.org @shopify/cli@0.0.0-snapshot-20260224152342` => install the snapshot again because the previous command installed the latest version
- `shopify version` => does nothing because it skips auto-update in snapshots
- `SHOPIFY_CLI_FORCE_AUTO_UPGRADE=1 shopify version` => auto-updates

### Measuring impact

### Checklist

### DIFF
--- a/packages/cli-kit/src/public/node/upgrade.ts
+++ b/packages/cli-kit/src/public/node/upgrade.ts
@@ -1,6 +1,22 @@
-import {currentProcessIsGlobal, inferPackageManagerForGlobalCLI} from './is-global.js'
-import {packageManagerFromUserAgent} from './node-package-manager.js'
-import {outputContent, outputToken} from './output.js'
+import {isDevelopment} from './context/local.js'
+import {currentProcessIsGlobal, inferPackageManagerForGlobalCLI, getProjectDir} from './is-global.js'
+import {
+  checkForCachedNewVersion,
+  findUpAndReadPackageJson,
+  PackageJson,
+  checkForNewVersion,
+  DependencyType,
+  usesWorkspaces,
+  addNPMDependencies,
+  getPackageManager,
+} from './node-package-manager.js'
+import {outputContent, outputDebug, outputInfo, outputToken} from './output.js'
+import {cwd, moduleDirectory, sniffForPath} from './path.js'
+import {exec, isCI} from './system.js'
+import {renderConfirmationPrompt} from './ui.js'
+import {isPreReleaseVersion} from './version.js'
+import {getAutoUpgradeEnabled, setAutoUpgradeEnabled} from '../../private/node/conf-store.js'
+import {CLI_KIT_VERSION} from '../common/version.js'
 
 /**
  * Utility function for generating an install command for the user to run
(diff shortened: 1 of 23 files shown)


gonzaloriestra:
* **#6848** <a href="https://app.graphite.com/github/pr/Shopify/cli/6848?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Shopify/cli/6848?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/cli-kit/src/public/node/upgrade.ts
```diff
@@ -1,38 +1,220 @@
...
+  const packageManager = inferPackageManagerForGlobalCLI()
+  if (!packageManager) return undefined
+
+  if (packageManager === 'homebrew') {
+    return 'brew upgrade shopify-cli'
+  } else if (packageManager === 'yarn') {
+    return `${packageManager} global add @shopify/cli@latest`
+  } else {
+    const verb = packageManager === 'pnpm' ? 'add' : 'install'
+    return `${packageManager} ${verb} -g @shopify/cli@latest`
```
isaacroldan, line 37: it's possible that the new CLI requires a newer version of node than the currently installed in the system. Maybe we could try to be smart and detect that and warn the user instead of failing forever :thinking_face:

gonzaloriestra: It's a good idea to improve later. But if I'm not wrong, the user would see that specific error at the end of each command, so I guess it's not a big deal as if it failed silently.

craigmichaelmartin: > it's possible that the new CLI requires a newer version of node

Upgrading our required Node version would come in a major version, which is already handled, right?

gonzaloriestra: If we also switch to SemVer, yes

craigmichaelmartin: Gotcha - I was thinking that was a dependency of this.

gonzaloriestra: Ideally yes, but we could ship this with the current system as well.

gonzaloriestra:
Closing in favor of https://github.com/Shopify/cli/pull/7104/

PR #6853, opened 2026-02-12
Fix hot-reloading for the {% javascript %} tag when serving compiled assets scripts with multibyte characters

karreiro:

When hot-reloading compiled JS generated from `{% javascript %}` blocks, the response `Content-Length` header could be computed using the string character count rather than the UTF‑8 byte length. This becomes incorrect as soon as the output contains multi-byte characters (e.g., an em dash `—`).

This PR updates the compiled_assets script response path to set `content-length` based on the UTF‑8 byte length.

### How to test your changes?

- Run `shopify theme dev` with a snippet like this:
```
Hey
{% javascript %}
console.log("Hey, hey")
// ...not full cart — derive count from response

console.log("Hey, hey, heeeey!")
{% endjavascript %}
```

### Post-release steps

n/a

### Measuring impact

### Checklist

### DIFF
--- a/.changeset/lovely-terms-tell.md
+++ b/.changeset/lovely-terms-tell.md
@@ -1,0 +1,5 @@
+---
+'@shopify/theme': patch
+---
+
+Fix hot-reloading for the {% javascript %} tag when serving compiled assets scripts with multibyte characters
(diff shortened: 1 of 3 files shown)


karreiro:
👋 Hey @Shopify/app-inner-loop, could you please take a look at this when you have a moment?

PS: I've proposed https://github.com/Shopify/cli/pull/6818 to prevent similar review requests in the future 🙇

PR #6851, opened 2026-02-11
Log on metrics submission errors

dmerand:
### WHY are these changes introduced?

Related to incident #inv-18682 (February 2026 OTLP failures)

When telemetry collection fails (e.g., OTLP endpoint unavailable), the CLI currently only logs to debug output, which means these failures go unnoticed unless users run with `--verbose`. This leaves the operations team blind to systemic telemetry infrastructure issues.

During the February 2026 incident, OTLP failures were silently occurring and only discovered when users manually reported CLI issues. We need visibility into telemetry failures without impacting the developer experience.

### WHAT is this pull request doing?

**Changes:**
1. Reports telemetry failures to Bugsnag (marked as `'expected_error'` - infrastructure issue, not a CLI bug)
2. Maintains existing behavior: telemetry failures remain transparent to users (no warnings/errors shown)
3. Adds test coverage for the new error reporting behavior

**Implementation:**
- Imports `sendErrorToBugsnag` in `analytics.ts`
- Calls it in the catch block when telemetry reporting fails (either Monorail or OTLP)

**What this achieves:**
- ✅ Operations team gets visibility into OTLP/Monorail outages
- ✅ Users never see telemetry errors (transparent operation)
- ✅ Protected against flooding Bugsnag (existing rate limits apply)

**Protection against overwhelming error reporting:**
- Bugsnag reporting has built-in rate limiting: **300 reports per day per user**
- This limit is shared across ALL error types (CLI bugs + telemetry failures)
- If OTLP has an extended outage, each user can only send a maximum of 300 reports total
- Rate limit tracking is per-user (stored in local config), not global
- Once the limit is hit, all subsequent errors are silently dropped with a debug log

### How to test your changes?

**Automated tests:**
```bash
pnpm test packages/cli-kit/src/public/node/analytics.test.ts
```

The new test `reports telemetry failures to Bugsnag` verifies:
- `sendErrorToBugsnag` is called when telemetry fails
- Error is categorized as `'expected_error'` (not a CLI bug)
- CLI doesn't crash if Bugsnag reporting fails

**Manual testing:**
Manual testing in development is limited because Bugsnag reporting is disabled in local/development environments (`isLocalEnvironment()` returns true).

The changes have been verified through:
- ✅ Code review of built output (changes are present)
- ✅ Unit tests pass
- ✅ TypeScript compilation succeeds

In production, this will work correctly when:
- `isLocalEnvironment()` returns false
- Bugsnag reporting is enabled
[...]

PR #6854, opened 2026-02-12
Do not crash on metrics errors

dmerand:

Resolves incident #inv-18682 (February 2026 OTLP failures)

When the OTLP telemetry endpoint is unavailable or experiencing failures, the CLI currently crashes with unhandled errors, forcing users to run commands with `SHOPIFY_CLI_NO_ANALYTICS=1` as a workaround. This violates the principle that telemetry should be transparent to users - it should never impact CLI functionality.

**Related to:** PR #6851 (adds Bugsnag reporting for telemetry failures)

**Changes:**

**Always** **resolve** **the** **OTEL** **metrics** **emitter** - There is no need to halt execution/reject the emitting process if OTEL reporting fails. This is a non-critical process, and we should allow things to continue to function normally even if this fails

**HOW** is this fixing things?

- Commands exit with `EXIT_CODE=0` now as we are always resolving. Previously, they were exiting with `1`

```
1. CLI command finishes                                                                                                          
     ↓                                                                                                                             
  2. postrun hook (cli-kit/.../hooks/postrun.ts)                                                                                   
     → calls reportAnalyticsEvent()                                                                                                
     ↓                                                                                                                             
  3. reportAnalyticsEvent (analytics.ts:44)                                                                                        
     → Promise.all([doMonorail(), doOpenTelemetry()])                                                                              
     → has try/catch, BUT...                                                                                                       
     ↓                                                                                                                             
  4. doOpenTelemetry → recordMetrics (otel-metrics.ts:58)                                                                          
     → calls recorder.otel.record() on counters/histograms                                                                         
     ↓                                                                                                                             
  5. BaseOtelService.record() (BaseOtelService.ts:~140)                                                                            
     → instrument.record(value, labels)                                                                                            
     → void meterProvider.forceFlush({}).catch(() => {})   ← FIRE-AND-FORGET                                                       
     ↓                                                                                                                             
  6. MeterProvider.forceFlush()                                                                                                    
     ↓                                                                                                                             
  7. InstantaneousMetricReader.onForceFlush()  [throttled to 1000ms]                                                               
     → this.collect({})  — gathers accumulated metrics                                                                             
     → this._exporter.export(resourceMetrics, callback)                                                                            
     ↓                                                                                                                             
  8. OTLPMetricExporter sends HTTP request to OTEL collector                                                                       
     → callback fires with { code: SUCCESS | FAILED, error? } 
     → previously rejected any errors, now we log/report them, but always resolve
```

### DIFF
--- a/packages/cli-kit/src/public/node/vendor/otel-js/export/InstantaneousMetricReader.test.ts
+++ b/packages/cli-kit/src/public/node/vendor/otel-js/export/InstantaneousMetricReader.test.ts
@@ -1,0 +1,73 @@
+import {InstantaneousMetricReader} from './InstantaneousMetricReader.js'
+import {ExportResultCode} from '@opentelemetry/core'
+import type {PushMetricExporter, ResourceMetrics} from '@opentelemetry/sdk-metrics'
+import {MeterProvider} from '@opentelemetry/sdk-metrics'
+import {describe, expect, test, vi} from 'vitest'
+import {diag} from '@opentelemetry/api'
+
+vi.mock('@opentelemetry/api')
+
+function createMockExporter(resultCode: ExportResultCode, error?: Error): PushMetricExporter {
+  return {
+    export: vi.fn((_metrics: ResourceMetrics, callback: (result: {code: ExportResultCode; error?: Error}) => void) => {
+      callback({code: resultCode, error})
+    }),
+    forceFlush: vi.fn().mockResolvedValue(undefined),
+    shutdown: vi.fn().mockResolvedValue(undefined),
+  } as unknown as PushMetricExporter
+}
+
+function createReaderWithProvider(exporter: PushMetricExporter): {
+  reader: InstantaneousMetricReader
+  provider: MeterProvider
+} {
+  const reader = new InstantaneousMetricReader({exporter, throttleLimit: 0})
+  const provider = new MeterProvider()
+  provider.addMetricReader(reader)
+  return {reader, provider}
+}
+
+describe('InstantaneousMetricReader', () => {
+  test('logs errors when metrics collection fails', async () => {
+    const exporter = createMockExporter(ExportResultCode.SUCCESS)
+    const {reader, provider} = createReaderWithProvider(exporter)
+
+    const collectionError = new Error('Collection failed')
+    vi.spyOn(reader as any, 'collect').mockResolvedValue({
+      resourceMetrics: {resource: {}, scopeMetrics: []},
+      errors: [collectionError],
+    })
+
+    await expect(reader.forceFlush()).resolves.toBeUndefined()
+    expect(diag.error).toHaveBeenCalledWith('InstantaneousMetricReader: metrics collection errors', collectionError)
+    await provider.shutdown()
+  })
+
+  test('resolves on successful export', async () => {
+    const exporter = createMockExporter(ExportResultCode.SUCCESS)
+    const {reader, provider} = createReaderWithProvider(exporter)
+
+    await expect(reader.forceFlush()).resolves.toBeUndefined()
+    expect(diag.error).not.toHaveBeenCalled()
+    await provider.shutdown()
+  })
+
+  test('resolves without rejecting on export failure', async () => {
+    const err = new Error('Export failed with retryable status')
+    const exporter = createMockExporter(ExportResultCode.FAILED, err)
+    const {reader, provider} = createReaderWithProvider(exporter)
+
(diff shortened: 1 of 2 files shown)


dmerand:
* **#6854** <a href="https://app.graphite.com/github/pr/Shopify/cli/6854?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Shopify/cli/6854?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#6851** <a href="https://app.graphite.com/github/pr/Shopify/cli/6851?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

packages/cli-kit/src/public/node/vendor/otel-js/service/FailSafeOtelService.ts
alexanderMontague, line 1: Probably don't need to wrap this service with what is effectively a try/catch. We chatted about this in DMs. I think we can work it into the existing service.

dmerand: Yep, this is way too heavy. I was operating on the assumption we don't want to touch vendored code, it was a false assumption.

alexanderMontague: Didn't ya hear? We like to roll our own stuff here :sunglasses:

dmerand: We were way ahead of LLMs.

packages/cli-kit/src/public/node/vendor/otel-js/export/InstantaneousMetricReader.test.ts
```diff
@@ -0,0 +1,51 @@
...
+    const {reader, provider} = createReaderWithProvider(exporter)
+
+    await expect(reader.forceFlush()).resolves.toBeUndefined()
+    await provider.shutdown()
+  })
+
+  test('resolves without rejecting when export error is undefined', async () => {
+    const exporter = createMockExporter(ExportResultCode.FAILED)
+    const {reader, provider} = createReaderWithProvider(exporter)
+
```
dmerand, line 68: We should perhaps test that the `diag.error` comes through in this and the above case.

packages/cli-kit/src/public/node/vendor/otel-js/export/InstantaneousMetricReader.ts
```diff
@@ -41,13 +41,12 @@ export class InstantaneousMetricReader extends MetricReader {
       diag.error('PeriodicExportingMetricReader: metrics collection errors', ...errors)
```
dmerand: Nit: This isn't anything we did in this PR, but this class name isn't correct any more -- perhaps a miss when vendoring?

packages/cli-kit/src/public/node/vendor/otel-js/export/InstantaneousMetricReader.test.ts
```diff
@@ -0,0 +1,51 @@
...
+describe('InstantaneousMetricReader', () => {
+  test('resolves on successful export', async () => {
+    const exporter = createMockExporter(ExportResultCode.SUCCESS)
+    const {reader, provider} = createReaderWithProvider(exporter)
+
+    await expect(reader.forceFlush()).resolves.toBeUndefined()
+    await provider.shutdown()
+  })
+
+  test('resolves without rejecting on export failure', async () => {
```
dmerand, line 55: Should we also add a test case for `export` returning synchronous exceptions?

alexanderMontague: There was no code testing the possibility of `export` throwing a synchronous error in the first place. I think this would also be caught downstream if it were to ever happen.

packages/cli-kit/src/public/node/vendor/otel-js/export/InstantaneousMetricReader.test.ts
alexanderMontague: I can see if we can mock/observe this. Will give it a shot.

packages/cli-kit/src/public/node/vendor/otel-js/export/InstantaneousMetricReader.ts
alexanderMontague: :+1: will update

PR #6862, opened 2026-02-18
doctor-release command runs on development envs

aswamy:

- When i run `shopify-dev theme-doctor theme` or `pnpm shopify theme-doctor theme` on local environment, i want it to run the development CLI, not the production one

- detects how the CLI was run

### How to test your changes?

- run the doctor-release command locally (try all the ones noted in the test cases)
  - just make sure you set the environment variable first (`export SHOPIFY_CLI_DOCTOR=1`)

### Measuring impact

### Checklist


aswamy:
* **#6862** <a href="https://app.graphite.com/github/pr/Shopify/cli/6862?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Shopify/cli/6862?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/cli/src/cli/services/doctor-release/context.ts
```diff
@@ -27,9 +27,35 @@ export interface ThemeDoctorOptions {
...
+
+  const firstTopic = commandHandle.split(':')[0]
+  if (!firstTopic) return defaultCommand
+
+  const index = argv.findIndex((arg) => arg === firstTopic)
+
+  if (index <= 0) return defaultCommand
+
+  return argv.slice(0, index).join(' ')
+}
```
aswamy, line 53: @EvilGenius13 do you think this might be too fragile? I am not sure if we should just make a new arg passed in from the user.  
  
If i were to steelman this approach, this would work in CI if we'd launch `theme-doctor theme` from a CI environment if it used `npx shopify` instead of `shopify`

EvilGenius13: Hey so I console logged the input of running the command and I didn't get an argument setup of something like `npx`, `shopify`, i got 
```
process.argv: ["/opt/homebrew/Cellar/node/23.9.0/bin/node","/Users/joshuafaigan/bin/shopify","doctor-release","theme","-e","evil"]
```
So I think we could just simplify the function to `return argv.slice(0, 2).join(' ')`, and then we only pass `detectCliCommand` through `createDoctorContext`. Let me know if I'm wrong!

PR #6863, opened 2026-02-18
Amend downloadFiles loop to a Promise.all to ensure writes are complete

EvilGenius13:

The current forEach loop in `downloadFiles()` doesn't await async callbacks. This causes the function to return immediately while file writes are still in progress. The CLI may show 100% progress even though files may still be writing to disk, and if there are any errors, they won't surface since the function has already completed.                                               

Replaces `forEach(async ...)` with `Promise.all()` to properly await all file writes. If there is an error, it will properly propagate up.                                                                                                                                               

### How to test your changes?
- Build the branch
- Run `theme pull` and ensure downloads still work and the CLI terminal output works.

You can also run this handy script to see how `forEach` doesn't await in the loop.

<details>
```
const POKEMON_IDS = [1, 4, 7, 25, 133];

async function fetchPokemon(id) {
  const response = await fetch(`https://pokeapi.co/api/v2/pokemon/${id}`);
  const data = await response.json();
  return data.name;
}

function timestamp() {
  return `[${new Date().toISOString().split("T")[1].slice(0, -1)}]`;
}

async function withForEach() {
  console.log(`\n--- forEach ---`);
  console.log(`${timestamp()} forEach started`);

  POKEMON_IDS.forEach(async (id) => {
    const name = await fetchPokemon(id);
    console.log(`${timestamp()}   fetched: ${name}`);
  });

  console.log(`${timestamp()} forEach complete\n`);
}

async function withPromiseAll() {
  console.log(`\n--- Promise.all ---`);
  console.log(`${timestamp()} Promise.all started`);

  await Promise.all(
    POKEMON_IDS.map(async (id) => {
      const name = await fetchPokemon(id);
      console.log(`${timestamp()}   fetched: ${name}`);
    })
  );

  console.log(`${timestamp()} Promise.all complete\n`);
}

async function main() {
  await withForEach();
  // give the orphaned forEach fetches time to print
  await new Promise((resolve) => setTimeout(resolve, 3000));
  await withPromiseAll();
}

main().catch(console.error);
```
</details>
### Measuring impact

### Checklist

### DIFF
--- a/packages/theme/src/cli/utilities/theme-downloader.ts
+++ b/packages/theme/src/cli/utilities/theme-downloader.ts
@@ -2,7 +2,7 @@ import {batchedTasks, Task} from './batching.js'
 import {MAX_GRAPHQL_THEME_FILES} from '../constants.js'
 import {AdminSession} from '@shopify/cli-kit/node/session'
 import {fetchThemeAssets} from '@shopify/cli-kit/node/themes/api'
-import {ThemeFileSystem, Theme, Checksum, ThemeAsset} from '@shopify/cli-kit/node/themes/types'
+import {ThemeFileSystem, Theme, Checksum} from '@shopify/cli-kit/node/themes/types'
 import {renderTasks} from '@shopify/cli-kit/node/ui'
 import {Writable} from 'stream'
 
@@ -90,8 +90,5 @@ async function downloadFiles(theme: Theme, fileSystem: ThemeFileSystem, filename
   const assets = await fetchThemeAssets(theme.id, filenames, session)
   if (!assets) return
 
-  // eslint-disable-next-line @typescript-eslint/no-misused-promises
-  assets.forEach(async (asset: ThemeAsset) => {
-    await fileSystem.write(asset)
-  })
+  await Promise.all(assets.map((asset) => fileSystem.write(asset)))
 }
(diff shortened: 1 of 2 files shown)


packages/theme/src/cli/utilities/theme-downloader.ts
```diff
@@ -90,8 +90,5 @@ async function downloadFiles(theme: Theme, fileSystem: ThemeFileSystem, filename
   const assets = await fetchThemeAssets(theme.id, filenames, session)
   if (!assets) return
 
-  // eslint-disable-next-line @typescript-eslint/no-misused-promises
-  assets.forEach(async (asset: ThemeAsset) => {
-    await fileSystem.write(asset)
-  })
+  await Promise.all(assets.map((asset: ThemeAsset) => fileSystem.write(asset)))
```
fredericoo: isn’t `asset` already inferred to be ThemeAsset?

aswamy (approved):
agree with freddie's comment, but functionally sound 👍

packages/theme/src/cli/utilities/theme-downloader.ts
EvilGenius13: Yes thank you fo the point out!

PR #6870, opened 2026-02-19
[Feature] Add build.typegen_command support for non-JS Shopify Functions

davejcameron:

Currently, typegen for Shopify Functions is hardcoded to run `npm exec -- graphql-code-generator --config package.json`, which only works for JavaScript functions. For other languages (Rust, Zig, C), users must bundle their typegen step inside a `build.sh` script alongside their actual build command. This makes it impossible to run typegen independently via `shopify app function typegen` for non-JS functions.

Adds a `build.typegen_command` TOML field so users can specify a custom typegen command that runs both via `shopify app function typegen` and automatically before the build command.

**Example TOML:**
```toml
[build]
command = "zig build -Doptimize=ReleaseSmall"
typegen_command = "npx shopify-function-codegen --language zig"
```

Changes:
- Adds `typegen_command` as an optional string field in the function extension schema
- Adds `typegenCommand` getter on `ExtensionInstance`
- Updates `buildGraphqlTypes` to run custom command when set, falling back to the existing JS codegen flow
- Runs typegen automatically before the build command for both JS and non-JS functions when `typegen_command` is configured
- Updates the `typegen` CLI command description to be language-agnostic
- Adds tests across all affected files

### How to test your changes?

1. Run function build tests: `npx vitest run packages/app/src/cli/services/function/build.test.ts`
2. Run extension build tests: `npx vitest run packages/app/src/cli/services/build/extension.test.ts`
3. Run function schema tests: `npx vitest run packages/app/src/cli/models/extensions/specifications/function.test.ts`

To test manually with a non-JS function:
1. Add `typegen_command = "echo typegen running"` under `[build]` in a function's TOML
2. Run `shopify app function typegen` — should execute the custom command
3. Run `shopify app function build` — should run typegen before the build command

### Measuring impact

### Checklist

### DIFF
--- a/packages/app/src/cli/services/function/build.ts
+++ b/packages/app/src/cli/services/function/build.ts
@@ -130,6 +130,7 @@ export async function buildGraphqlTypes(
     return runWithTimer('cmd_all_timing_network_ms')(async () => {
       return exec(commandComponents[0]!, commandComponents.slice(1), {
         cwd: fun.directory,
+        stdout: options.stdout,
         stderr: options.stderr,
         signal: options.signal,
       })
--- a/packages/app/src/cli/services/function/build.test.ts
+++ b/packages/app/src/cli/services/function/build.test.ts
@@ -121,6 +121,7 @@ describe('buildGraphqlTypes', () => {
     await expect(got).resolves.toBeUndefined()
     expect(exec).toHaveBeenCalledWith('npx', ['shopify-function-codegen', '--schema', 'schema.graphql'], {
       cwd: ourFunction.directory,
+      stdout,
       stderr,
       signal,
     })
@@ -150,6 +151,7 @@ describe('buildGraphqlTypes', () => {
     await expect(got).resolves.toBeUndefined()
     expect(exec).toHaveBeenCalledWith('custom-typegen', ['--output', 'types.ts'], {
       cwd: ourFunction.directory,
+      stdout,
       stderr,
       signal,
     })


shayarnett:
🎩 

Before: Screenshot 2026-02-19 at 3 04 50 PM

After: 
Screenshot 2026-02-19 at 3 14 10 PM

packages/app/src/cli/services/function/build.ts
```diff
@@ -123,9 +123,20 @@ async function buildJSFunctionWithTasks(
...
+      return exec(commandComponents[0]!, commandComponents.slice(1), {
+        cwd: fun.directory,
+        stderr: options.stderr,
+        signal: options.signal,
+      })
+    })
+  }
+
   if (!fun.isJavaScript) {
     throw new AbortError('GraphQL types can only be built for JavaScript functions')
```
isaacroldan: Maybe we want to update this error message to let the developer know they can use the new toml property?

davejcameron: Good catch Isaac, I'll update that.

packages/app/src/cli/models/extensions/extension-instance.ts
```diff
@@ -289,6 +289,11 @@ export class ExtensionInstance<TConfiguration extends BaseConfigType = BaseConfi
     return config.build?.command
   }
 
+  get typegenCommand() {
+    const config = this.configuration as unknown as FunctionConfigType
+    return config.build?.typegen_command
+  }
```
isaacroldan, line 293: I'll ping @ryancbahan here for an unrelated project discussion.
This is the kind of thing that shouldn't be in the generic `extension-instance`, but is an existing pattern with no real alternative right now.

The specification for Functions (or appmodule definition, or whatever we end up calling it) should own its build process, including the typegen step.

Issue #1325, opened 2023-02-08
[Bug]: Cannot deploy Theme extension


kinngh:
### Please confirm that you have:

- [X] Searched [existing issues](.) to see if your issue is a duplicate. (If you’ve found a duplicate issue, feel free to add additional information in a comment on it.)
- [X] Reproduced the issue in [the latest CLI version](https://www.npmjs.com/package/@shopify/cli).

### In which of these areas are you experiencing a problem?

App, Extension

### Expected behavior

Create an extension in a new project or use an existing project and push a theme extension to Shopify.

### Actual behavior

The extension shows up in the Partner Dashboard for versioning.

### Reproduction steps

1. Create a new Shopify app and generate a new theme extension
2. Add a `liquid` file to the blocks folder
3. Push to Shopify

Met with error on both existing and new app projects:

![Screenshot 2023-02-08 at 8 51 41 PM](https://user-images.githubusercontent.com/773555/217572746-2aacc624-43a4-4541-b0e7-566b3747fa4b.png)

### Operating System

macOS 13.2 / M1

### Shopify CLI version (check your project's `package.json` if you're not sure)

3.41.1

### Shell

zsh

### Node version (run `node -v` if you're not sure)

19.6.0

### What language and version are you using in your application?

Node 19.6.0

Arkham:
hey @kinngh, thanks for the report. Our team is working on a fix! As a workaround can you try prefixing your command with `SHOPIFY_CLI_NO_THEME_BUNDLING=1`?

kinngh:
Working as expected with `SHOPIFY_CLI_NO_THEME_BUNDLING=1`!

briandesp:
What does this mean - Working as expected with SHOPIFY_CLI_NO_THEME_BUNDLING=1! where is that value or how is it changed?

gonzaloriestra:
@jfisher727 could you take a look at this issue? 🙏

PR #6873, opened 2026-02-20
Gracefully catch unexpected html (non json) response from external services

alfonso-noriega:

[Issue #1325](https://github.com/shop/issues/issues/1325)

[Vault error](https://vault.shopify.io/teams/16949-Dev-Experience/issues/1312-error-unexpected-token-html-h-is-not-valid-json)

[Observe error](https://observe.shopify.io/a/observe/errors/11062327764500656069)

JSON parsing errors in HTTP responses were causing unhandled exceptions and poor error messages when external services return malformed responses or are temporarily unavailable.

Add proper error handling for JSON parsing failures in HTTP response processing across three modules:

- **Token exchange service**: Wrap JSON parsing in try-catch and throw descriptive `AbortError` when identity service returns invalid JSON
- **GitHub API client**: Handle JSON parsing errors when fetching release information and provide clear error messaging
- **Admin authentication**: Add error handling for malformed responses from the admin authentication service

All error messages inform users that the service may be temporarily unavailable and suggest retrying.

### Measuring impact

### Checklist

### DIFF
--- a/packages/cli-kit/src/private/node/session/exchange.ts
+++ b/packages/cli-kit/src/private/node/session/exchange.ts
@@ -229,12 +229,22 @@ async function tokenRequest(params: {
   url.search = new URLSearchParams(Object.entries(params)).toString()
 
   const res = await shopifyFetch(url.href, {method: 'POST'})
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  const payload: any = await res.json()
+  try {
+    const responseText = await res.text()
 
-  if (res.ok) return ok(payload)
+    const payload = JSON.parse(responseText)
+    if (res.ok) return ok(payload)
 
-  return err({error: payload.error, store: params.store})
+    return err({error: payload.error, store: params.store})
+  } catch (error) {
+    if (error instanceof SyntaxError) {
+      throw new AbortError(
+        `Received invalid response from authentication service (HTTP ${res.status}).`,
+        'The response could not be parsed as JSON. The service may be temporarily unavailable. Please try again.',
+      )
+    }
+    throw error
+  }
 }
 
 function buildIdentityToken(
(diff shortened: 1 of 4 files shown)


alfonso-noriega:
* **#6873** <a href="https://app.graphite.com/github/pr/Shopify/cli/6873?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Shopify/cli/6873?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/cli-kit/src/private/node/session/exchange.ts
```diff
@@ -229,12 +229,20 @@ async function tokenRequest(params: {
...
+    const responseText = await res.text()
 
-  if (res.ok) return ok(payload)
+    const payload = JSON.parse(responseText)
+    if (res.ok) return ok(payload)
 
-  return err({error: payload.error, store: params.store})
+    return err({error: payload.error, store: params.store})
+  } catch {
+    throw new AbortError(
```
gonzaloriestra: We could also wrap it in something like this (I think it returns SyntaxError when the Json is not valid) to avoid catching other errors:

```
if (error instanceof SyntaxError)
```

Same for the other catchs.

packages/cli-kit/src/private/node/session/exchange.ts
```diff
@@ -229,12 +229,20 @@ async function tokenRequest(params: {
...
+      `Received invalid response from identity service (HTTP ${res.status}).` +
+        ' The response could not be parsed as JSON. The service may be temporarily unavailable.' +
+        ' Please try again.',
```
gonzaloriestra: You can pass this as the second parameter (tryMessage). Same for the rest.

packages/cli-kit/src/private/node/session/exchange.ts
```diff
@@ -229,12 +229,20 @@ async function tokenRequest(params: {
...
 
-  if (res.ok) return ok(payload)
+    const payload = JSON.parse(responseText)
+    if (res.ok) return ok(payload)
 
-  return err({error: payload.error, store: params.store})
+    return err({error: payload.error, store: params.store})
+  } catch {
+    throw new AbortError(
+      `Received invalid response from identity service (HTTP ${res.status}).` +
```
gonzaloriestra: Users don't know about identity, what about `authentication service`?

PR #6880, opened 2026-02-20
[Feature] emit metafile for ui extensions to allow for bundle analyzation

robin-drexler:

Shopify imposes strict bundle size limits on extensions and functions to ensure great performance. Starting with the 2025-10 version, these limits have been decreased even further to 64kb compressed for ui extensions.

Bundle size can be impacted by different factors, such as large dependencies or inline assets.
 
However, we are not making it easy to understand why a bundle might exceeding these limits.

See https://shopify.slack.com/archives/C05E3BDFDRB/p1771618346307599 for context

When `build` for ui extensions runs, the CLI now emits a [metafile](https://esbuild.github.io/api/#metafile) for each of them in the extension's `dist/` folder.

That metafile can then be analyzed by tools like https://esbuild.github.io/analyze/ or locally via tools like https://github.com/btd/esbuild-visualizer

### How to test your changes?

- have an app with ui extension(s)
- run ` pnpm shopify app build  --path=/path/to/your/app` inside the cli repo
- cd into `/path/to/your/app`
- run `ls extensions/**/dist/*.metafile.json`
- It should print a metafile for each extension
- drop one or more of these into https://esbuild.github.io/analyze to see the analyzer

### Post-release steps
We will mention the metafile in docs like here: https://shopify.dev/docs/api/checkout-ui-extensions/2025-10/upgrading-to-2025-10#file-size-limit

### Checklist

### DIFF
--- a/packages/app/src/cli/services/bundle.test.ts
+++ b/packages/app/src/cli/services/bundle.test.ts
@@ -1,4 +1,4 @@
-import {writeManifestToBundle, compressBundle} from './bundle.js'
+import {writeManifestToBundle, compressBundle, BUNDLE_EXCLUSION_PATTERNS} from './bundle.js'
 import {AppInterface} from '../models/app/app.js'
 import {describe, test, expect, vi} from 'vitest'
 import {joinPath} from '@shopify/cli-kit/node/path'
@@ -53,7 +53,7 @@ describe('compressBundle', () => {
       expect(zip).toHaveBeenCalledWith({
         inputDirectory: inputDir,
         outputZipPath: outputZip,
-        matchFilePattern: ['**/*', '!**/*.js.map'],
+        matchFilePattern: ['**/*', ...BUNDLE_EXCLUSION_PATTERNS],
       })
       expect(brotliCompress).not.toHaveBeenCalled()
     })
@@ -74,7 +74,7 @@ describe('compressBundle', () => {
       // Then
       expect(zip).toHaveBeenCalledWith(
         expect.objectContaining({
-          matchFilePattern: ['**/*', '!**/*.js.map'],
+          matchFilePattern: ['**/*', ...BUNDLE_EXCLUSION_PATTERNS],
         }),
       )
     })
@@ -95,7 +95,7 @@ describe('compressBundle', () => {
       expect(brotliCompress).toHaveBeenCalledWith({
         inputDirectory: inputDir,
         outputPath: outputBr,
-        matchFilePattern: ['**/*', '!**/*.js.map'],
+        matchFilePattern: ['**/*', ...BUNDLE_EXCLUSION_PATTERNS],
       })
       expect(zip).not.toHaveBeenCalled()
     })
@@ -116,7 +116,48 @@ describe('compressBundle', () => {
       // Then
       expect(brotliCompress).toHaveBeenCalledWith(
         expect.objectContaining({
-          matchFilePattern: ['**/*', '!**/*.js.map'],
+          matchFilePattern: ['**/*', ...BUNDLE_EXCLUSION_PATTERNS],
+        }),
+      )
+    })
+  })
+
+  test('excludes .metafile.json files from the zip', async () => {
+    await inTemporaryDirectory(async (tmpDir) => {
+      // Given
+      const inputDir = joinPath(tmpDir, 'input')
+      const outputZip = joinPath(tmpDir, 'output.zip')
+      await mkdir(inputDir)
+      await writeFile(joinPath(inputDir, 'test.txt'), 'test content')
+      await writeFile(joinPath(inputDir, 'main.metafile.json'), '{"inputs":{},"outputs":{}}')
+
+      // When
+      await compressBundle(inputDir, outputZip)
+
+      // Then
+      expect(zip).toHaveBeenCalledWith(
+        expect.objectContaining({
+          matchFilePattern: ['**/*', ...BUNDLE_EXCLUSION_PATTERNS],
+        }),
+      )
+    })
+  })
+
+  test('uses custom file patterns as-is when provided', async () => {
+    await inTemporaryDirectory(async (tmpDir) => {
+      // Given
+      const inputDir = joinPath(tmpDir, 'input')
+      const outputZip = joinPath(tmpDir, 'output.zip')
+      await mkdir(inputDir)
+      await writeFile(joinPath(inputDir, 'test.txt'), 'test content')
+
+      // When
+      await compressBundle(inputDir, outputZip, ['ext1/**', 'manifest.json'])
+
+      // Then
+      expect(zip).toHaveBeenCalledWith(
+        expect.objectContaining({
+          matchFilePattern: ['ext1/**', 'manifest.json'],
         }),
       )
     })
--- a/packages/app/src/cli/services/bundle.ts
+++ b/packages/app/src/cli/services/bundle.ts
@@ -14,8 +14,10 @@ export async function writeManifestToBundle(appManifest: AppManifest, bundlePath
   await writeFile(manifestPath, JSON.stringify(appManifest, null, 2))
 }
 
+export const BUNDLE_EXCLUSION_PATTERNS = ['!**/*.js.map', '!**/*.metafile.json']
+
 export async function compressBundle(inputDirectory: string, outputPath: string, customMatchFilePattern?: string[]) {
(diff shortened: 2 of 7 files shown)


packages/app/src/cli/services/extensions/bundle.ts
```diff
@@ -58,6 +58,13 @@ export async function bundleExtension(options: BundleOptions, processEnv = proce
   const context = await esContext(esbuildOptions)
   const result = await context.rebuild()
   onResult(result, options)
+  if (result.metafile) {
+    const metafilePath = joinPath(
```
robin-drexler: Some extensions have multiple JS bundles. The main one and potentially one for should-render. 

If we only had a single metafile.json, one would override the other.

That's why create a `your-extension-handle.mefafile.json` file.

packages/app/src/cli/services/extensions/bundle.ts
```diff
@@ -154,6 +161,9 @@ export function getESBuildOptions(options: BundleOptions, processEnv = process.e
     esbuildOptions.sourcemap = true
     esbuildOptions.sourceRoot = `${options.stdin.resolveDir}/src`
   }
+  if (options.environment === 'production') {
```
robin-drexler, line 175: Question for reviewers. Do we even ever run this not in production mode? Could we just always emit the metafile?

robin-drexler: Should we actually make this a feature like `generates_source_maps`?

packages/app/src/cli/services/bundle.ts
```diff
@@ -15,7 +15,7 @@ export async function writeManifestToBundle(appManifest: AppManifest, bundlePath
 }
 
 export async function compressBundle(inputDirectory: string, outputPath: string, customMatchFilePattern?: string[]) {
-  const matchFilePattern = customMatchFilePattern ?? ['**/*', '!**/*.js.map']
+  const matchFilePattern = customMatchFilePattern ?? ['**/*', '!**/*.js.map', '!**/*.metafile.json']
```
isaacroldan: check the callers of `compressBundle`, the `dev-session` passes a custom filePattern, do we want the metafile to be included during dev sessions? if not, we probably need to explicitly ignore it there too. (althought it seems like we are duplicating stuff, see if you can simplify that, that'd be great)

packages/app/src/cli/services/extensions/bundle.ts
isaacroldan: Yes to the production question, check the callers of `getESBuildOptions`, when called from `app-watcher-esbuild` it is called in development mode, because that's from the dev session flow.
Do we want to generate metafiles only when deploying? or also when deving? is there any performance hit by doing it?

I don't think we need to convert this into a feature, it makes sense to want to this for every esbuild extension.

packages/app/src/cli/services/extensions/bundle.ts
isaacroldan: This snippet is a bit hard to follow 🤔

I think we can use `parse` from node's path to simplify it:
```ts
const { dir, name } = parse(options.outputPath)
const metafilePath = joinPath(dir, `${name}.metafile.json`)
```
`parse` returns the dir, name, ext... so you can use the components you need

let's add `parse` to `packages/cli-kit/src/public/node/path.ts` and use it here :)

packages/app/src/cli/services/extensions/bundle.ts
```diff
@@ -58,6 +58,13 @@ export async function bundleExtension(options: BundleOptions, processEnv = proce
   const context = await esContext(esbuildOptions)
   const result = await context.rebuild()
   onResult(result, options)
+  if (result.metafile) {
+    const metafilePath = joinPath(
+      dirname(options.outputPath),
+      `${basename(options.outputPath, extname(options.outputPath))}.metafile.json`,
+    )
+    await writeFile(metafilePath, JSON.stringify(result.metafile))
```
dmerand: We may want to wrap this in case it fails. Do we want metafile write failures to abort the main code flow? My opinion is that we do not.

robin-drexler: @dmerand that's fair. Should we just `try{}catch{}` swallow the error or report it somewhere? Are there established patterns for those cases in the repo?

packages/app/src/cli/services/bundle.test.ts
dmerand, line 62: Nit: We may want a test similar to this for exclusion of `metafile.json`.

packages/app/src/cli/services/extensions/bundle.ts
robin-drexler: >Do we want to generate metafiles only when deploying?

I think only doing it for prod builds is ok for now.

packages/app/src/cli/services/extensions/bundle.ts
dmerand: I'm still not a codebase expert :/ But I was thinking just try/catch, yeah. Maybe with an `outputDebug(`Failed to write metafile: ${error}`)` so that `--verbose` would show the error.

packages/app/src/cli/services/extensions/bundle.ts
robin-drexler: changed, thanks!

packages/app/src/cli/services/bundle.ts
robin-drexler: introduced a `BUNDLE_EXCLUSION_PATTERNS` const am re-using this now. also added it to dev-session.

robin-drexler:
@dmerand @isaacroldan thanks for your reviews. I have hopefully addressed all your concerns. Could give this another go, please? 🙏

packages/app/src/cli/services/extensions/bundle.ts
```diff
@@ -58,6 +58,16 @@ export async function bundleExtension(options: BundleOptions, processEnv = proce
...
   const result = await context.rebuild()
   onResult(result, options)
+  if (result.metafile) {
+    const {dir, name} = parsePath(options.outputPath)
+    const metafilePath = joinPath(dir, `${name}.metafile.json`)
+    try {
+      await writeFile(metafilePath, JSON.stringify(result.metafile))
+      // eslint-disable-next-line no-catch-all/no-catch-all
+    } catch (error) {
+      outputDebug(`Failed to write metafile to ${metafilePath}: ${error}`)
```
isaacroldan: outputDebug messages are not shown in the terminal by default (it's only visible in verbose mode).  
  
Is this intentional? or do you want the user to know that this failed?

robin-drexler: that's a good question. I followed @dmerand request here. 

Before I swallowed it entirely.

I'm down to always log it. What do you think?

robin-drexler: Wait, before I actually didn't catch it at all and it would have failed the build. Sorry, I got confused for a sec.

robin-drexler: @isaacroldan changed to `outputWarn`

robin-drexler:
@isaacroldan https://github.com/Shopify/cli/pull/6880#issuecomment-3961018527 seems fine to me since we're only adding a new function. Would you agree?


Also curious why those changes need to be backwards compatible? 🤔

packages/app/src/cli/services/extensions/bundle.ts
```diff
@@ -58,6 +58,16 @@ export async function bundleExtension(options: BundleOptions, processEnv = proce
...
+  if (result.metafile) {
+    const {dir, name} = parsePath(options.outputPath)
+    const metafilePath = joinPath(dir, `${name}.metafile.json`)
+    try {
+      await writeFile(metafilePath, JSON.stringify(result.metafile))
+      // eslint-disable-next-line no-catch-all/no-catch-all
+    } catch (error) {
+      outputWarn(`Failed to write metafile to ${metafilePath}: ${error}`)
+    }
+  }
```
isaacroldan: Just one more ask, can we extract this to its own function?
```ts
function writeMetafile(result) {
..
}
```

The bundleExtension function was very clean, let's keep it that way :) 
(i'll approve the PR, but please do extract the code first)

isaacroldan (approved):

packages/app/src/cli/services/extensions/bundle.ts
dmerand: +1. This extraction may also make it easier to test that this warning actually happens as desired.

PR #1063, opened 2023-01-11
Add missing documentation

pepicrft:
### WHY are these changes introduced?
I'm adding missing documentation to public modules in `@shopify/cli-kit`. I'll be doing it in batches to make PR reviews more digestible.

### WHAT is this pull request doing?

I'm ensuring the documentation is present in modules modules:

- `@shopify/cli-kit/common/object`
- `@shopify/cli-kit/common/array`
- `@shopify/cli-kit/common/url`

### How to test your changes?
Linting should pass

### Measuring impact

How do we know this change was effective? Please choose one:

- [x] n/a - this doesn't need measurement, e.g. a linting rule or a bug-fix
- [ ] Existing analytics will cater for this addition
- [ ] PR includes analytics changes to measure impact

### Checklist

- [x] I've considered possible cross-platform impacts (Mac, Linux, Windows)
- [x] I've considered possible [documentation](https://shopify.dev) changes
- [x] I've made sure that any changes to `dev` or `deploy` have been reflected in the [internal flowchart](https://www.figma.com/file/7vqUp50u6dm48Zfb4JRRn8/CLI3-Internals).

PR #6881, opened 2026-02-21
change hardcoded links in theme pull to use methods to surface urls instead

FKauwe:
The `theme pull` command had hardcoded links for previewing your theme, and launching the theme editor. 

Fixes [#1063 ](https://github.com/Shopify/developer-tools-team/issues/1063)

The hardcoded links were replaced by `themePreviewUrl` and `themeEditorUrl` methods.

The `themePreviewURL` function builds the url following two paths and one path can lead to the url looking slightly different than the original string behavior (it truncates the url if `theme.role === live`, but I tested both urls in my browser and they went to the same place. 
This is the Before + After:
Before: 
```
[1] https://f9e81e-af.myshopify.com/?preview_theme_id=145364484332
[2] https://f9e81e-af.myshopify.com/admin/themes/145364484332/editor
```
After:
```
[1] https://f9e81e-af.myshopify.com
[2] https://f9e81e-af.myshopify.com/admin/themes/145364484332/editor
```

The After is cleaner for this edge case and doesn't add the unnecessary preview part, as the theme is live and the url can point to the live view.

I noticed there were no unittests testing `renderSuccess` but that other sibling tests do test that and the message surfaced to user, so I added that test

### How to test your changes?

- while on main, run `p build`, then `shopify-dev theme pull  -e <local-env-name> `in external terminal
- choose a live theme to pull to see the edge case
- note the older, longer url surfaced to the user after the `renderSuccess` message
- pull my branch down locally: `fix-hardcoded-links-in-theme-pull`
- `p build`
- run `shopify-dev theme pull  -e <local-env-name>` in external terminal
- note the newer url without the unnecessary preview appended (if you chose a live theme). 
- If you choose an unpublished theme to pull the urls remain exactly the same

### Measuring impact

### DIFF
--- a/.changeset/tender-wolves-doubt.md
+++ b/.changeset/tender-wolves-doubt.md
@@ -1,0 +1,5 @@
+---
+'@shopify/cli': patch
+---
+
+change hardcoded links in theme pull to use methods to surface urls instead
--- a/packages/theme/src/cli/services/pull.test.ts
+++ b/packages/theme/src/cli/services/pull.test.ts
@@ -14,6 +14,8 @@ import {fetchChecksums} from '@shopify/cli-kit/node/themes/api'
 import {insideGitDirectory, isClean} from '@shopify/cli-kit/node/git'
 import {test, describe, expect, vi, beforeEach} from 'vitest'
 import {dirname, joinPath} from '@shopify/cli-kit/node/path'
+import {renderSuccess} from '@shopify/cli-kit/node/ui'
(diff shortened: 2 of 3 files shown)


EvilGenius13 (approved):
Thank you for the PR @FKauwe!

EvilGenius13:
Let's add a `patch` level changeset. Since the success message used to show `?preview_theme_id` when it was a live theme, it's an observable change in the CLI output.

PR #6889, opened 2026-02-23
return original zod schema after app module validation

ryancbahan:

See [incident notes](https://docs.google.com/document/d/1RqN-H6_6yhRvo0C20fXr9r1x3gYJ7mE3dxghpePLcqg/edit?tab=t.0). The current validation mutates data in-place twice: once for zod.transform and once during json schema validation. We can keep the json schema validation approach as-is with its stripping behavior (sohuld we long term? IMO no, but different discussion) so long as we return the zod object, which we don't strip. This is a least-we-can-do approach to gain some safety back by reducing the amount of mutation to data happening inside a validation loop. I've also added a check to run transforms on the data pre-json schema validation only if a transform function is provided. This allows us a migration path to get things on contracts.

### Measuring impact

### Checklist

### DIFF
--- a/packages/app/src/cli/utilities/json-schema.test.ts
+++ b/packages/app/src/cli/utilities/json-schema.test.ts
@@ -19,18 +19,15 @@ describe('unifiedConfigurationParserFactory', () => {
   }
 
   test('falls back to zod parser when no JSON schema is provided', async () => {
-    // Given
     const merged = {
       identifier: randomUUID(),
       parseConfigurationObject: mockParseConfigurationObject,
       validationSchema: undefined,
     }
 
-    // When
     const parser = await unifiedConfigurationParserFactory(merged as any)
     const result = parser({type: 'product_subscription'})
 
-    // Then
     expect(result).toEqual({
       state: 'ok',
       data: {type: 'product_subscription'},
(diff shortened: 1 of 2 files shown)


ryancbahan:
* **#6889** <a href="https://app.graphite.com/github/pr/Shopify/cli/6889?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Shopify/cli/6889?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/app/src/cli/utilities/json-schema.test.ts
```diff
@@ -113,11 +100,9 @@ describe('unifiedConfigurationParserFactory', () => {
       },
     }
 
-    // When
```
ryancbahan, line 116: deleting some unneeded llm output as i pass by

isaacroldan: I don't think this is llm :sweat_smile:, is something we used to do as a team to separate the different parts of a test. And llm's have picked it up when adding new tests

gonzaloriestra: Yeah, it's actually part our testing conventions: https://shopify.github.io/cli/cli/testing-strategy.html

PR #6890, opened 2026-02-24
[Feature] Add Theme previews via JSON

dengjeffrey:

Resolves https://github.com/shop/issues-growth-activation/issues/1902

1. Adds  `theme preview` to support temporarily overriding a theme via JSON and being able to preview it

More internal context: https://docs.google.com/document/d/13sa0iRs_DSgyPHbx9IIF64rvhh_KZCDLfJgwpSIxP4A/edit?usp=sharing

`--override` specifies the path to the override JSON to process
`--preview-id` specifies an existing preview session to be updated with the JSON value in source

### How to test your changes?
https://docs.google.com/document/d/13sa0iRs_DSgyPHbx9IIF64rvhh_KZCDLfJgwpSIxP4A/edit?tab=t.0#bookmark=id.rhjcmeq18hn7

### Post-release steps

- [X] Add documentation on shopify.dev
https://github.com/shop/world/pull/483644

### Measuring impact

### Checklist


packages/theme/src/cli/commands/theme/dev.ts
```diff
@@ -143,6 +161,18 @@ You can run this command only in a directory that matches the [default Shopify t
...
     recordEvent('theme-command:dev:single-env:authenticated')
 
+    if (devFlags.source) {
+      const resolvedSource = resolvePath(devFlags.source)
+
+      if (isDirectorySync(resolvedSource)) {
+        devFlags.path = resolvedSource
+      } else {
+        recordEvent('theme-command:dev:override-session')
+        await createOverrideDevSession(resolvedSource, devFlags, adminSession)
```
dengjeffrey: Using this same pattern here to "create" a dev session even if we don't have a long running process yet. This will make it trivial to add hot reloading.

packages/theme/src/cli/services/dev-override.ts
```diff
@@ -0,0 +1,75 @@
...
+ * Reads a JSON overrides file and creates or updates a Storefront preview.
+ * The resulting preview URL is displayed to the user.
+ */
+export async function devWithOverrideFile(options: DevWithOverrideFileOptions) {
+  if (!fileExistsSync(options.overrideJson)) {
+    throw new AbortError(`Override file not found: ${options.overrideJson}`)
+  }
+
+  const fileContent = await readFile(options.overrideJson)
+  const overrides = JSON.parse(fileContent) as ThemeOverrides
```
karreiro: May we handle JSON parser errors here?

packages/theme/src/cli/commands/theme/dev.ts
```diff
@@ -134,6 +142,16 @@ You can run this command only in a directory that matches the [default Shopify t
       env: 'SHOPIFY_FLAG_ALLOW_LIVE',
       default: false,
     }),
+    source: Flags.string({
+      description:
+        'The source for the dev server. Can be a directory or a JSON overrides file. When a directory is provided, it is used as the theme directory. When a JSON file is provided, overrides are applied to the theme specified by --theme.',
+      env: 'SHOPIFY_FLAG_SOURCE',
+    }),
```
karreiro: While reviewing the flags for this command, I noticed we already have `--path`, so there are two options:

1. When `--source` is empty, document that the command uses the file system as the source and relies on `--path`.
2. Remove `--source` in favor of `--path`, allowing you to pass either a `--path` or an override file as a parameter.

I think option 2 is the cleaner approach. What do you think?

packages/theme/src/cli/commands/theme/dev.ts
```diff
@@ -134,6 +142,16 @@ You can run this command only in a directory that matches the [default Shopify t
...
+    'preview-id': Flags.string({
+      description:
+        'An existing preview identifier to update instead of creating a new preview. Used with --source when pointing to a JSON overrides file.',
+      env: 'SHOPIFY_FLAG_PREVIEW_ID',
+    }),
```
karreiro: When reading this as a user, I immediately wondered how I would get a `--preview-id`. I think we could mention that `--preview-id` is an optional field that you may get from previous preview runs.

packages/cli-kit/src/private/node/session.ts
```diff
@@ -80,7 +80,7 @@ interface AppManagementAPIOauthOptions {
 /**
  * A scope supported by the Storefront Renderer API.
  */
-export type StorefrontRendererScope = 'devtools'
```
karreiro, line 83: Deferring review of this line to the @Shopify/app-inner-loop team, as they implemented the storefront renderer scopes and have more context on this area.

packages/theme/src/cli/utilities/theme-previews/preview.ts
```diff
@@ -0,0 +1,112 @@
...
+ * @param options - The options for creating a theme preview.
+ * @returns The preview URL and identifier for the applied overrides.
+ */
+export async function createThemePreview({
+  session,
+  storefrontToken,
+  overridesContent,
+}: CreateThemePreviewOptions): Promise<ThemePreviewResult> {
+  recordTiming('theme-preview:create')
+  const baseUrl = buildBaseStorefrontUrl(session)
```
karreiro, line 40: Did you have a chance to tophat this using the Theme Access app?

I wonder if changes will be required there to support the new API call. I think it should work out of the box, but worth double-checking (specially from scopes perspective).

dengjeffrey: Will do! Now that everything required to test this has been merged, I will give it a go tomorrow.

packages/cli-kit/src/private/node/session/scopes.ts
```diff
@@ -48,7 +48,7 @@ function defaultApiScopes(api: API): string[] {
     case 'admin':
       return ['graphql', 'themes', 'collaborator']
     case 'storefront-renderer':
-      return ['devtools']
+      return ['devtools', 'graphql']
```
isaacroldan, line 51: have you updated the identity app to support this? reach out to me in slack for more details if not.

packages/theme/src/cli/commands/theme/dev.ts
```diff
@@ -32,19 +37,26 @@ This command returns the following information:
...
   static description = this.descriptionWithoutMarkdown()
 
   static flags = {
     ...globalFlags,
     ...themeFlags,
+    path: {
+      ...themeFlags.path,
+      description:
+        'The path for the dev server. It can be a directory or a JSON overrides file. When a directory is provided, it is used as the theme directory. When a JSON file is provided, overrides are applied to the theme specified by --theme. Defaults to the current working directory.',
+    } as typeof themeFlags.path,
```
isaacroldan: path is a globally used flag and should keep its behaviour consistent over all commands (apps, theme...).  
`path` means: execute the command in the given path instead of in the current directory.  
It's not in `globalFlags` because not all commands are executed in a specific path, but maybe it should.  
  
Anyway, I don't think we should change how `path` works in this specific command only. It'd be better to have a new flag

dengjeffrey: I'll move it out to an --overrides flag

dengjeffrey:
@karreiro thanks for taking a look! Yea there were some changes to Identity. I'll DM you

packages/theme/src/cli/utilities/theme-previews/preview.ts
```diff
@@ -0,0 +1,113 @@
...
+ */
+export async function createThemePreview({
+  session,
+  storefrontToken,
+  overridesContent,
+  themeId,
+}: CreateThemePreviewOptions): Promise<ThemePreviewResult> {
+  recordTiming('theme-preview:create')
+  const baseUrl = buildBaseStorefrontUrl(session)
+  const url = `${baseUrl}/theme_preview.json?preview_theme_id=${themeId}`
```
robthornton, line 41: 👍

.changeset/slick-humans-enjoy.md
```diff
@@ -0,0 +1,6 @@
+---
+'@shopify/theme': minor
+'@shopify/cli': minor
+---
+
+Add support for theme previews using an override via `theme dev`. --path now supports an override JSON. A new --preview-id flag is also introduced to handle updates for preview.
```
isaacroldan: I think this changeset is outdated

packages/theme/src/cli/services/dev-override.ts
```diff
@@ -0,0 +1,72 @@
...
+ */
+export async function devWithOverrideFile(options: DevWithOverrideFileOptions) {
+  if (!fileExistsSync(options.overrideJson)) {
+    throw new AbortError(`Override file not found: ${options.overrideJson}`)
+  }
+
+  const fileContent = await readFile(options.overrideJson)
+  let overrides: ThemeOverrides
+  try {
+    overrides = JSON.parse(fileContent) as ThemeOverrides
```
frandiox, line 34: Should we also validate here the size limit and the expected JSON shape? Or are we going to rely solely on server validation?

packages/theme/src/cli/commands/theme/dev.ts
```diff
@@ -144,6 +159,12 @@ You can run this command only in a directory that matches the [default Shopify t
 
     recordEvent('theme-command:dev:single-env:authenticated')
 
+    if (devFlags.overrides) {
+      recordEvent('theme-command:dev:override-session')
+      await createOverrideDevSession(devFlags.overrides, devFlags, adminSession)
+      return
+    }
```
frandiox: This is rather counter-intuitive for me. I would expect a `theme dev` command to be a long running server that updates when I change something in my source files, not a one-off override from a separate file.

In my opinion, this seems more aligned with `theme share` (`shopify theme share --overrides changes.json`). Even with file watching added later, `theme dev` comes with a lot of irrelevant baggage (local proxy server, HMR, port/host flags, etc) that doesn't apply to this flow. A `--watch` flag on a simpler command would achieve the same result without the semantic mismatch 🤔

Is there any strong reason we need this to be in `theme dev` specifically? Are we thinking of adding extra functionality in the future that makes this play well with `theme dev`?

The main mismatch I see with `theme share` is that this command currently creates a new theme entry in Admin, and it wouldn't do so with `--overrides`. But it stills feels closer semantically than `theme dev`. In any case, not a strong opposition if we want to keep going with `theme dev`.

packages/theme/src/cli/services/dev-override.ts
dengjeffrey: Solely rely on server validation so we don't repeat logic

packages/theme/src/cli/services/dev-override.ts
```diff
@@ -0,0 +1,70 @@
...
+  renderSuccess({
+    body: [
+      {
+        list: {
+          title: options.previewIdentifier ? 'Preview updated' : 'Preview is ready',
+          items: [{link: {url: preview.url}}, `Preview ID: ${preview.preview_identifier}`],
+        },
+      },
+    ],
+  })
```
frandiox, line 65: To make this easily usable by agents, perhaps we could add `--json` flag support for structured output instead of a banner? I think we do that in other commands.

frandiox (approved):
LGTM! Tested with normal auth and also with theme access app, it works 🎉  Thanks!

packages/theme/src/cli/services/dev-override.ts
dengjeffrey: That's a good idea, i'll add as a follow up

PR #6905, opened 2026-02-27
fix missing json output for theme info command when dev and theme flags are missing

FKauwe:

Fixes [#1064](https://github.com/Shopify/developer-tools-team/issues/1064) : Theme info not outputting json with the --json flag

I identified that when running `shopify theme info --json `without a `--theme` or `--development` flag, execution goes to the else branch, which had no JSON handling. The `JSON.stringify` logic only existed in the `if` branch (which requires `--theme` or `--development`)

The `else` branch had no `--json` handling at all, so the flag was silently ignored.

I could have chosen to jsonify the existing info message, but this was formulated for the tabular data to output the formatted box in the terminal and looks messy and is more difficult to parse. It would have looked like this:
<details>
<summary>Click to expand</summary>

Output:
```
[
  {
    "title": "Theme Configuration PROOF OF WHERE I AM",
    "body": {
      "tabularData": [
        [
          "Store",
          "f9e81e-af.myshopify.com"
        ],
        [
          "Development Theme ID",
          {
            "subdued": "Not set"
          }
        ]
      ],
      "firstColumnSubdued": true
    }
  },
  {
    "title": "Tooling and System",
    "body": {
      "tabularData": [
        [
          "Shopify CLI",
          "3.91.0"
        ],
        [
          "OS",
          "darwin-arm64"
        ],
        [
          "Shell",
          "/bin/zsh"
        ],
        [
          "Node version",
          "v23.6.1"
        ]
      ],
      "firstColumnSubdued": true
    }
  }
]

```

</details>

I chose to make a new, more easily parse-able json object (following the pattern of the if branch, where a clean object is also created specifically to be consumed by `JSON.stringify`).  I built a new Interface and function to populate the object in `packages/theme/src/cli/services/info.ts`, then consumed it in the `else` branch in `packages/theme/src/cli/commands/theme/info.ts`

I wrote a new unittest to verify that the JSON conversion is working correctly.
All existing tests pass. 

UPDATE: 
during PR conversation with Gray, I discovered there was no command level test file for the info command so I added one with tests to cover all conditional branches

### How to test your changes?
 while on main, run `shopify theme info --json`, should see no json and formulated tabular info like this:
 
Screenshot 2026-02-27 at 10 17 05 AM

then pull down my branch locally,
`p build `,
run `shopify-dev theme info --json`
see the nice JSON formatting!
Screenshot 2026-02-27 at 9 32 06 AM

### Measuring impact

### DIFF
--- a/packages/theme/src/cli/commands/theme/info.test.ts
+++ b/packages/theme/src/cli/commands/theme/info.test.ts
@@ -1,0 +1,103 @@
+import Info from './info.js'
+import {themeEnvironmentInfoJSON, fetchDevInfo, fetchThemeInfo, formatThemeInfo} from '../../services/info.js'
+import {describe, vi, expect, test} from 'vitest'
+import {Config} from '@oclif/core'
+import {ensureAuthenticatedThemes} from '@shopify/cli-kit/node/session'
+import {outputResult} from '@shopify/cli-kit/node/output'
+import {renderInfo} from '@shopify/cli-kit/node/ui'
+
+vi.mock('../../services/info.js')
+vi.mock('@shopify/cli-kit/node/session')
+vi.mock('@shopify/cli-kit/node/output')
+vi.mock('@shopify/cli-kit/node/ui')
+
+const CommandConfig = new Config({root: __dirname})
+
+const session = {
+  token: 'test-token',
+  storeFqdn: 'my-shop.myshopify.com',
+}
+
+describe('Info', () => {
+  async function run(argv: string[]) {
+    await CommandConfig.load()
+    vi.mocked(ensureAuthenticatedThemes).mockResolvedValue(session)
+    const info = new Info(['--store=my-shop.myshopify.com', '--password=test-password', ...argv], CommandConfig)
+    await info.run()
+  }
+
+  describe('when theme or development flag is provided', () => {
+    const mockThemeInfo = {
+      theme: {
+        id: 123,
+        name: 'my theme',
+        role: 'live',
+        shop: 'my-shop.myshopify.com',
+        preview_url: 'https://my-shop.myshopify.com/preview',
+        editor_url: 'https://my-shop.myshopify.com/editor',
+      },
+    }
+
+    test('outputs JSON when --json flag is passed', async () => {
+      vi.mocked(fetchThemeInfo).mockResolvedValue(mockThemeInfo)
+
+      await run(['--theme', '123', '--json'])
+
+      expect(fetchThemeInfo).toHaveBeenCalled()
+      expect(outputResult).toHaveBeenCalledWith(JSON.stringify(mockThemeInfo, null, 2))
+      expect(renderInfo).not.toHaveBeenCalled()
+    })
+
+    test('renders formatted info when no --json flag is passed', async () => {
+      const mockFormatted = {
+        customSections: [{title: 'Theme Details', body: {tabularData: [], firstColumnSubdued: true}}],
+      }
+      vi.mocked(fetchThemeInfo).mockResolvedValue(mockThemeInfo)
+      vi.mocked(formatThemeInfo).mockResolvedValue(mockFormatted)
+
+      await run(['--theme', '123'])
+
+      expect(fetchThemeInfo).toHaveBeenCalled()
+      expect(formatThemeInfo).toHaveBeenCalled()
+      expect(renderInfo).toHaveBeenCalled()
+      expect(outputResult).not.toHaveBeenCalled()
+    })
+
+    test('throws error when theme is not found', async () => {
+      vi.mocked(fetchThemeInfo).mockResolvedValue(undefined)
+
+      await expect(run(['--theme', '999'])).rejects.toThrow()
+    })
+  })
+
+  describe('when no theme or development flag is provided', () => {
+    test('outputs JSON when --json flag is passed', async () => {
+      const mockDevInfo = {
+        store: 'my-shop.myshopify.com',
+        development_theme_id: null,
+        cli_version: '3.91.0',
+        os: 'darwin-arm64',
+        shell: '/bin/zsh',
+        node_version: 'v23.6.1',
+      }
+      vi.mocked(themeEnvironmentInfoJSON).mockReturnValue(mockDevInfo)
+
+      await run(['--json'])
+
+      expect(themeEnvironmentInfoJSON).toHaveBeenCalled()
+      expect(outputResult).toHaveBeenCalledWith(JSON.stringify(mockDevInfo, null, 2))
+      expect(renderInfo).not.toHaveBeenCalled()
+    })
+
+    test('renders info when no --json flag is passed', async () => {
+      const mockSections = [{title: 'Theme Configuration', body: {tabularData: [], firstColumnSubdued: true}}]
+      vi.mocked(fetchDevInfo).mockResolvedValue(mockSections)
+
+      await run([])
+
+      expect(fetchDevInfo).toHaveBeenCalled()
+      expect(renderInfo).toHaveBeenCalled()
+      expect(outputResult).not.toHaveBeenCalled()
+    })
+  })
+})


packages/theme/src/cli/services/info.ts
```diff
@@ -41,6 +50,18 @@ export function themeInfoJSON(theme: Theme, adminSession: AdminSession): ThemeIn
   }
 }
 
+export function devInfoJSON(config: {cliVersion: string}): DevInfo {
+  const {platform, arch} = platformAndArch()
+  return {
+    store: getThemeStore() ?? 'Not configured',
+    development_theme_id: getDevelopmentTheme() ?? null,
```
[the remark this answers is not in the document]
FKauwe: thank you, great catch, I added in a protective check that now looks for store first

packages/theme/src/cli/services/info.ts
```diff
@@ -28,6 +28,15 @@ interface ThemeInfoOptions {
   json?: boolean
 }
 
+interface DevInfo {
```
EvilGenius13: Could we rename this to `ThemeEnvironmentInfo`?

packages/theme/src/cli/services/info.ts
```diff
@@ -41,6 +50,19 @@ export function themeInfoJSON(theme: Theme, adminSession: AdminSession): ThemeIn
   }
 }
 
+export function devInfoJSON(config: {cliVersion: string}): DevInfo {
+  const {platform, arch} = platformAndArch()
+  const store = getThemeStore()
+  return {
+    store: store ?? 'Not configured',
+    development_theme_id: store ? getDevelopmentTheme() ?? null : null,
```
EvilGenius13: This took me a minute to understand. If you don't want to make it more verbose can we wrap the first part in `()` so like `development_theme_id: store ? (getDevelopmentTheme() ?? null) : null,
`? That way you can more easily ready that it's `getDevelopmentTheme` that may return null OR that if store isn't found we don't want to run that method and skip right to null as well.

packages/theme/src/cli/commands/theme/info.ts
```diff
@@ -50,6 +50,9 @@ export default class Info extends ThemeCommand {
       renderInfo(formattedInfo)
     } else {
       const infoMessage = await fetchDevInfo({cliVersion: this.config.version})
+      if (flags.json) {
```
EvilGenius13, line 52: It looks like we're calling `flags.json` twice now. Can you look into merging that logic?

FKauwe: What did you have in mind? I'm not sure I'm following.
The branching is:
if dev or theme flag present, proceed to json flag check, if present, return one converted result
OR if dev and theme flags are not present, proceed to json flag check, if present, return a *different converted result
because the initial check for theme and dev flags means there will be different data to jsonify based on that check, I don't think I can combine the json checks at a higher level further up the chain (if that's what you meant)

packages/theme/src/cli/services/info.ts
FKauwe: absolutely. I chose `devInfo` bc there was an existing function that gathers the same/ similar info called
```
export async function fetchDevInfo(config: {cliVersion: string}): Promise<AlertCustomSection[]> {
  return [devConfigSection(), await systemInfoSection(config)]
}

function devConfigSection(): AlertCustomSection {
  const store = getThemeStore() ?? 'Not configured'
  const developmentTheme = getDevelopmentTheme()

  recordEvent(`theme-command:info:dev-theme-loaded:${developmentTheme}`)

  return tabularSection('Theme Configuration', [
    ['Store', store],
    ['Development Theme ID', developmentTheme ? `#${developmentTheme}` : {subdued: 'Not set'}],
  ])
}
```
But I agree, `ThemeEnvironmentInfo` is more descriptive

FKauwe: and since I'm changing the Interface name I'm going to change the function that returns this type to match as well, to: `themeEnvironmentInfoJSON`

packages/theme/src/cli/commands/theme/info.ts
EvilGenius13: Ah yes that makes sense. It does do two different things so we can't really unify it.

On another note, can you move the new `if (flags.json)` one level up above `const infoMessage = await fetchDevInfo({cliVersion: this.config.version})`?

Right now what happens is that line gets run if we hit the `else` branch, and then if json is enabled we do nothing with that code. We should check if it's going to be `json` first and if not, run the `const infoMessage...`

packages/theme/src/cli/services/info.ts
FKauwe: the linter unfortunately strips the `( ) `as unnecessary.  I will use a multi line conditional, I also agree it is difficult to read

packages/theme/src/cli/commands/theme/info.ts
FKauwe: great catch, don't need to waste an unnecessary call to `fetchDevInfo` since my `themeEnvironmentInfoJSON` builds its own clean object independently

packages/theme/src/cli/services/info.ts
```diff
@@ -41,6 +50,23 @@ export function themeInfoJSON(theme: Theme, adminSession: AdminSession): ThemeIn
...
+  const {platform, arch} = platformAndArch()
+  const store = getThemeStore()
+  let developmentThemeID = null
+  if (store) {
+    developmentThemeID = getDevelopmentTheme() ?? null
+  }
+  return {
+    store: store ?? 'Not configured',
+    development_theme_id: developmentThemeID,
+    cli_version: config.cliVersion,
```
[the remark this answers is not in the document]
FKauwe: I got a string back from getDevelopmentTheme, which I didn't realize comes from local storage which stores everything as a string.  But I checked the type/ interface for theme and the correct type for id is number:
```
interface ThemeInfo {
  theme: {
    id: number
    name: string
    role: string
    shop: string
    editor_url: string
    preview_url: string
  }
}
```
I went ahead and added a number conversion to my function to make sure its returning the correct type according to theme object interface

packages/theme/src/cli/commands/theme/info.ts
```diff
@@ -1,6 +1,6 @@
-import ThemeCommand from '../../utilities/theme-command.js'
-import {fetchThemeInfo, fetchDevInfo, formatThemeInfo} from '../../services/info.js'
 import {themeFlags} from '../../flags.js'
+import {fetchThemeInfo, fetchDevInfo, formatThemeInfo, themeEnvironmentInfoJSON} from '../../services/info.js'
+import ThemeCommand from '../../utilities/theme-command.js'
```
graygilmore, line 1: There might be something off with your editor setup if these moved around like this on their own. Might be worth having a little poke at.

packages/theme/src/cli/services/info.test.ts
```diff
@@ -47,6 +49,23 @@ describe('info', () => {
     expect(output).toHaveProperty('theme.editor_url', expect.stringContaining(session.storeFqdn))
   })
 
+  test('generate dev info JSON when no theme or dev flag provided', () => {
```
graygilmore: This particular file has no concept of flags being passed in:

```suggestion
  test('generates information about the dev environment formatted in JSON', () => {
```

In a more perfect world we'd also wrap it in a describe block so that we know that we're explicitly talking about a certain function:

```ts
describe('themeEnvironmentInfoJSON', () => {
  test('');
});
```

But this entire test file is kind of wrong as it is (it's testing `commands/theme/info` instead of `services/info`).

packages/theme/src/cli/commands/theme/info.ts
```diff
@@ -49,7 +49,11 @@ export default class Info extends ThemeCommand {
       const formattedInfo = await formatThemeInfo(output, flags)
       renderInfo(formattedInfo)
     } else {
+      if (flags.json) {
```
graygilmore, line 52: Should we have a test for this branching logic?

packages/theme/src/cli/commands/theme/info.ts
FKauwe: Thank you, prob a cursor setting.  I've transitioned to VS Code so hopefully it's not an issue with that editor but I'll keep an eye on it for the first few imports I add to in VS Code

packages/theme/src/cli/services/info.test.ts
FKauwe: great callout- you're right this is testing the services path not the commands path. I got confused bc there is no commands path test file so I assumed this was the right one since there was only one :/  I am going to rename this test to reflect that it's Testing that themeEnvironmentInfoJSON returns correct data. and then I'll add a new test file at the commands level that checks the conditional logic of the flags being present. the only command-level test is check.test.ts so I'll use that as a pattern

packages/theme/src/cli/commands/theme/info.ts
FKauwe: yes we should! I covered all conditional branches in the new test file

PR #6903, opened 2026-02-27
Add e2e tests to PR CI workflow

ryancbahan:
## WHY are these changes introduced?

The e2e tests from #6899 and #6900 need to run as part of the PR workflow to catch regressions before merge.

## WHAT is this pull request doing?

- Adds an `e2e-tests` job to `.github/workflows/tests-pr.yml`
- Installs Playwright chromium for browser-based OAuth login
- Runs `npx playwright test` from `packages/e2e/`
- Uploads `playwright-report/` and `test-results/` as artifacts on failure
- Does not block merge on e2e failure (informational)
- Fixes knip: adds `playwright` to `ignoreBinaries`, removes stale `@graphql-typed-document-node/core` from `ignoreDependencies`

## How to test your changes?

CI runs automatically on PR.

## Measuring impact

n/a

PR #6900, opened 2026-02-27
Add e2e tests for app init, deploy, and dev server

ryancbahan:
## WHY are these changes introduced?

Builds on the e2e infrastructure from #6899 to add test coverage for the CLI's core app workflows: scaffolding, deployment, and dev server lifecycle. Also adds OAuth browser login automation so tests can authenticate against real Shopify APIs.

## WHAT is this pull request doing?

**New setup fixtures:**
- `setup/auth.ts` — worker-scoped OAuth login via Playwright chromium + node-pty
- `setup/app.ts` — test-scoped app scaffold fixture (init, build, deploy, generate extension)

**New helpers:**
- `helpers/browser-login.ts` — `completeLogin()` for Shopify OAuth flow (shared by setup and scripts)
- `helpers/wait-for-text.ts` — polling utility for PTY output matching (shared by setup and scripts)

**New tests:**
- `app-scaffold.spec.ts` — init react-router app + build, init extension-only app
- `app-deploy.spec.ts` — deploy with version tag, verify via versions list
- `app-dev-server.spec.ts` — start dev server, wait for ready, press q to quit

**Scripts:**
- `scripts/create-test-apps.ts` — interactive app creation for setting up test org (uses shared `completeLogin` helper)

**Fixture chain:** `env → cli → auth → app`

## How to test your changes?

```bash
cd packages/e2e && npx playwright test
```

## Measuring impact

n/a - test infrastructure

PR #6899, opened 2026-02-26
Add e2e test infrastructure with Playwright + node-pty

ryancbahan:
## WHY are these changes introduced?

The CLI team manually runs a ~30-step QA checklist before every release. This is time-consuming and error-prone. The Cucumber suite tests CLI commands non-interactively but cannot handle interactive CLI prompts, long-running dev servers, auth, or browser verification. We need e2e test infra that automates QA and provides regression testing for CLI features.

## WHAT is this pull request doing?

Introduces a new `packages/e2e/` package with Playwright + node-pty infrastructure for automated end-to-end testing.

**Directory structure:**

```
packages/e2e/
  setup/          → Playwright fixture definitions (DI wiring)
    env.ts          worker-scoped XDG isolation & env var management
    cli.ts          CLI process management (execa + node-pty)
  helpers/        → Pure utility functions
    strip-ansi.ts   ESM wrapper for strip-ansi
  tests/          → Test specs
    smoke.spec.ts       validates execa path (shopify version)
    smoke-pty.spec.ts   validates PTY path (shopify version via node-pty)
```

**Key design decisions:**

- `setup/` to distinguish Playwright fixture definitions from static test data
- `cli.ts` provides both `exec()` (non-interactive via execa) and `spawn()` (interactive via node-pty with output matching, key sending, process lifecycle)
- Worker-scoped env isolation with XDG directory sandboxing
- `helpers/` contains stateless utility functions with no Playwright dependency

Subsequent PRs in this stack add OAuth login, app scaffold/deploy/dev tests, CI integration, and TOML config regression tests.

## How to test your changes?

```bash
pnpm test:e2e
```

## Measuring impact

n/a - this is test infrastructure

PR #6907, opened 2026-02-27
Add cleanup script for e2e test org

ryancbahan:

E2e test runs create apps in the test org. Without cleanup, these accumulate over time.

Adds `scripts/cleanup-test-apps.ts` — browser automation script that:
- Logs in via OAuth (reuses `completeLogin` from `helpers/browser-login.ts`)
- Navigates to the dev dashboard
- Finds and deletes test apps matching a filter pattern
- Supports `--dry-run`, `--filter`, and `--headed` flags

```bash
npx tsx packages/e2e/scripts/cleanup-test-apps.ts --filter "cli-e2e"
```

## How to test your changes?

Run the script manually against the test org.

## Measuring impact

n/a

### DIFF
--- a/packages/e2e/scripts/cleanup-test-apps.ts
+++ b/packages/e2e/scripts/cleanup-test-apps.ts
@@ -1,0 +1,264 @@
+/**
+ * Deletes all test apps from the dev dashboard via browser automation.
+ * Run: npx tsx packages/e2e/scripts/cleanup-test-apps.ts
+ *
+ * Pass --dry-run to list apps without deleting.
+ * Pass --filter <pattern> to only delete apps matching the pattern.
+ */
+
+import * as fs from 'fs'
+import * as path from 'path'
+import {fileURLToPath} from 'url'
+import {execa} from 'execa'
+import {chromium, type Page} from '@playwright/test'
+import stripAnsiModule from 'strip-ansi'
+
+const __dirname = path.dirname(fileURLToPath(import.meta.url))
+const rootDir = path.resolve(__dirname, '../../..')
+const cliPath = path.join(rootDir, 'packages/cli/bin/run.js')
+
+const dryRun = process.argv.includes('--dry-run')
+const filterIdx = process.argv.indexOf('--filter')
+const filterPattern = filterIdx >= 0 ? process.argv[filterIdx + 1] : undefined
+const headed = process.argv.includes('--headed') || !process.env.CI
+
+// Load .env
+const envPath = path.join(__dirname, '../.env')
+if (fs.existsSync(envPath)) {
+  for (const line of fs.readFileSync(envPath, 'utf-8').split('\n')) {
+    const trimmed = line.trim()
+    if (!trimmed || trimmed.startsWith('#')) continue
+    const eqIdx = trimmed.indexOf('=')
+    if (eqIdx === -1) continue
+    const key = trimmed.slice(0, eqIdx).trim()
+    const value = trimmed.slice(eqIdx + 1).trim()
+    if (!process.env[key]) process.env[key] = value
+  }
+}
+
+const email = process.env.E2E_ACCOUNT_EMAIL
+const password = process.env.E2E_ACCOUNT_PASSWORD
+if (!email || !password) {
+  console.error('E2E_ACCOUNT_EMAIL and E2E_ACCOUNT_PASSWORD must be set')
+  process.exit(1)
+}
+
+const baseEnv: {[key: string]: string} = {
+  ...(process.env as {[key: string]: string}),
+  NODE_OPTIONS: '',
+  SHOPIFY_RUN_AS_USER: '0',
+}
+delete baseEnv.SHOPIFY_CLI_PARTNERS_TOKEN
+
+async function main() {
+  // Step 1: OAuth login to get a browser session
+  console.log('--- Logging out ---')
+  await execa('node', [cliPath, 'auth', 'logout'], {env: baseEnv, reject: false})
+
+  console.log('--- Logging in via OAuth ---')
+  const nodePty = await import('node-pty')
+  const spawnEnv = {...baseEnv, CI: '', BROWSER: 'none'}
+  const pty = nodePty.spawn('node', [cliPath, 'auth', 'login'], {
+    name: 'xterm-color',
+    cols: 120,
+    rows: 30,
+    env: spawnEnv,
+  })
+
+  let output = ''
+  pty.onData((data: string) => {
+    output += data
+  })
+
+  await waitForText(() => output, 'Press any key to open the login page', 30_000)
+  pty.write(' ')
+  await waitForText(() => output, 'start the auth process', 10_000)
+
+  const stripped = stripAnsiModule(output)
+  const urlMatch = stripped.match(/https:\/\/accounts\.shopify\.com\S+/)
+  if (!urlMatch) throw new Error(`No login URL found:\n${stripped}`)
+
+  // Launch browser - we'll reuse this session for dashboard navigation
+  const browser = await chromium.launch({headless: !headed})
+  const context = await browser.newContext({
+    extraHTTPHeaders: {
+      'X-Shopify-Loadtest-Bf8d22e7-120e-4b5b-906c-39ca9d5499a9': 'true',
+    },
+  })
+  const page = await context.newPage()
+
+  // Complete OAuth login
+  await page.goto(urlMatch[0])
+  await page.waitForSelector('input[name="account[email]"], input[type="email"]', {timeout: 60_000})
+  await page.locator('input[name="account[email]"], input[type="email"]').first().fill(email)
+  await page.locator('button[type="submit"]').first().click()
+  await page.waitForSelector('input[name="account[password]"], input[type="password"]', {timeout: 60_000})
+  await page.locator('input[name="account[password]"], input[type="password"]').first().fill(password!)
+  await page.locator('button[type="submit"]').first().click()
+
+  // Wait for confirmation page and handle it
+  await page.waitForTimeout(3000)
+  try {
+    const confirmBtn = page.locator('button[type="submit"]').first()
+    if (await confirmBtn.isVisible({timeout: 5000})) {
+      await confirmBtn.click()
+    }
+    // eslint-disable-next-line no-catch-all/no-catch-all
+  } catch (_err) {
+    // No confirmation page
+  }
+
+  await waitForText(() => output, 'Logged in', 60_000)
+  console.log('Logged in successfully!')
+  try {
+    pty.kill()
+    // eslint-disable-next-line no-catch-all/no-catch-all
+  } catch (_err) {
+    // already dead
+  }
+
+  // Step 2: Navigate to dev dashboard
+  console.log('\n--- Navigating to dev dashboard ---')
+  await page.goto('https://dev.shopify.com/dashboard', {waitUntil: 'domcontentloaded'})
+  await page.waitForTimeout(3000)
+
+  // Handle account picker if shown
+  const accountButton = page.locator(`text=${email}`).first()
+  if (await accountButton.isVisible({timeout: 5000}).catch(() => false)) {
+    console.log('Account picker detected, selecting account...')
+    await accountButton.click()
+    await page.waitForTimeout(3000)
+  }
+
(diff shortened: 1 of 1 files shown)


ryancbahan:
* **#6940** <a href="https://app.graphite.com/github/pr/Shopify/cli/6940?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6907** <a href="https://app.graphite.com/github/pr/Shopify/cli/6907?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Shopify/cli/6907?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#6903** <a href="https://app.graphite.com/github/pr/Shopify/cli/6903?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6900** <a href="https://app.graphite.com/github/pr/Shopify/cli/6900?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#6901](https://github.com/Shopify/cli/pull/6901) <a href="https://app.graphite.com/github/pr/Shopify/cli/6901?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#6899** <a href="https://app.graphite.com/github/pr/Shopify/cli/6899?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

packages/e2e/scripts/cleanup-test-apps.ts
dmerand, line 1: This seems like a simple enough change, even if we're not targeting Windows right now:

💬 **Suggestion**: Screenshots are hard-coded to save in `/tmp/` (lines 152, 221, 230, 203) which is Unix-specific and won't work on Windows. Consider using `os.tmpdir()` for cross-platform compatibility. The sibling script `create-test-apps.ts` uses `os.tmpdir()`, so this would align the two scripts.

**Suggestion:** Import os at the top:
```typescript
import * as os from 'os'
```

Then replace `/tmp/` paths throughout with `path.join(os.tmpdir(), 'filename.png')`. For example:
```typescript
await page.screenshot({path: path.join(os.tmpdir(), 'e2e-dashboard.png')})
```

Note: This may not be critical if the script only runs in CI/Linux environments, but it improves portability.

dmerand, line 1: The shared E2E auth fixture already wraps its Playwright browser in `try/finally` ([[file:packages/e2e/setup/auth.ts:63-75]]), but this script only closes the browser on explicit success and early-return paths ([[file:packages/e2e/scripts/cleanup-test-apps.ts:176-208]]). If anything throws after launch during login or initial dashboard navigation ([[file:packages/e2e/scripts/cleanup-test-apps.ts:82-122]]), the browser can be left running. Consider aligning this standalone script with the fixture's cleanup pattern by wrapping the browser lifetime in `try/finally`.

packages/e2e/scripts/cleanup-test-apps.ts
```diff
@@ -0,0 +1,264 @@
...
+ * Pass --dry-run to list apps without deleting.
+ * Pass --filter <pattern> to only delete apps matching the pattern.
+ */
+
+import * as fs from 'fs'
+import * as path from 'path'
+import {fileURLToPath} from 'url'
+import {execa} from 'execa'
+import {chromium, type Page} from '@playwright/test'
+import stripAnsiModule from 'strip-ansi'
```
dmerand: 📐 **Design**: The script imports `strip-ansi` directly as `stripAnsiModule`, but the project has a helper at `helpers/strip-ansi.ts` that provides a cleaner named export. Using the helper improves consistency across the e2e package and is especially important if you adopt the `waitForText` helper, which expects the named export.

**Suggestion:** 
```suggestion
import {stripAnsi} from '../helpers/strip-ansi.js'
```

Then update usages at lines 77 and 248 from `stripAnsiModule(...)` to `stripAnsi(...)`.

packages/e2e/scripts/cleanup-test-apps.ts
```diff
@@ -0,0 +1,264 @@
...
+// Load .env
+const envPath = path.join(__dirname, '../.env')
+if (fs.existsSync(envPath)) {
+  for (const line of fs.readFileSync(envPath, 'utf-8').split('\n')) {
+    const trimmed = line.trim()
+    if (!trimmed || trimmed.startsWith('#')) continue
+    const eqIdx = trimmed.indexOf('=')
+    if (eqIdx === -1) continue
+    const key = trimmed.slice(0, eqIdx).trim()
+    const value = trimmed.slice(eqIdx + 1).trim()
```
dmerand: 🐛 **Bug**: The manual .env parser (lines 27-36) doesn't handle quoted values (e.g., `KEY="value with spaces"`). The quotes will be included in the parsed value, which may cause authentication issues if the .env file uses quoted strings. Additionally, this reimplements functionality available in the `dotenv` package, which handles edge cases like escaped characters and multiline values.

**Suggestion:** Consider replacing manual parsing with the `dotenv` library:
```typescript
import dotenv from 'dotenv'
dotenv.config({ path: path.join(__dirname, '../.env') })
```

Alternatively, if keeping manual parsing, handle quoted values:
```suggestion
let value = trimmed.slice(eqIdx + 1).trim()
// Remove surrounding quotes if present
if ((value.startsWith('"') && value.endsWith('"')) || (value.startsWith("'") && value.endsWith("'"))) {
  value = value.slice(1, -1)
}
process.env[key] = value
```

packages/e2e/scripts/cleanup-test-apps.ts
```diff
@@ -0,0 +1,264 @@
...
+  const browser = await chromium.launch({headless: !headed})
+  const context = await browser.newContext({
+    extraHTTPHeaders: {
+      'X-Shopify-Loadtest-Bf8d22e7-120e-4b5b-906c-39ca9d5499a9': 'true',
+    },
+  })
+  const page = await context.newPage()
+
+  // Complete OAuth login
+  await page.goto(urlMatch[0])
```
dmerand: 📐 **Design**: The OAuth login flow (lines 91-109) duplicates functionality that already exists in `helpers/browser-login.ts` as `completeLogin()`. The PR description states this script reuses `completeLogin`, but the implementation manually recreates the entire flow. This creates maintenance burden - if the login flow needs updating, it must be changed in multiple places. The sibling script `create-test-apps.ts` demonstrates the correct pattern of importing and using this helper.

**Suggestion:** Import and use the existing `completeLogin()` helper:

```suggestion
  // Complete OAuth login
  await completeLogin(page, urlMatch[0], email, password!)
```

Add this import at the top of the file:
```typescript
import {completeLogin} from '../helpers/browser-login.js'
```

Then remove lines 92-109 (the manual OAuth flow implementation).

packages/e2e/scripts/cleanup-test-apps.ts
```diff
@@ -0,0 +1,264 @@
...
+  console.log(`Found ${appCards.length} app links on dashboard`)
+
+  // Collect app names and URLs
+  const apps: {name: string; url: string}[] = []
+  for (const card of appCards) {
+    const href = await card.getAttribute('href')
+    const text = await card.textContent()
+    if (href && text && href.match(/\/apps\/\d+/)) {
+      // Extract just the app name (first line of text, before "installs")
+      const name = text.split(/\d+ install/)[0]?.trim() ?? text.trim()
```
dmerand: 🐛 **Bug**: Extracting app names by splitting on `/\d+ install/` assumes a specific format. If the UI changes or if an app has 0 installs (showing different text like '0 installs' or no install text at all), this will break. The current approach also doesn't validate the extracted name.

**Suggestion:** Add fallback and validation:
```suggestion
const name = text.split(/\d+ install/)[0]?.trim() || text.split('\n')[0]?.trim() || text.trim()
if (!name || name.length > 200) continue  // Skip invalid entries
```

This provides fallbacks for different text formats and validates that the extracted name is reasonable before processing.

packages/e2e/scripts/cleanup-test-apps.ts
```diff
@@ -0,0 +1,264 @@
...
+  let deleted = 0
+  for (const app of apps) {
+    console.log(`\nDeleting "${app.name}"...`)
+    try {
+      await deleteApp(page, app.url, app.name)
+      deleted++
+      console.log(`  Deleted "${app.name}"`)
+    } catch (err) {
+      console.error(`  Failed to delete "${app.name}":`, err)
+      await page.screenshot({path: `/tmp/e2e-delete-error-${deleted}.png`})
```
dmerand: 🐛 **Bug**: When an app deletion fails, the error screenshot is saved to `/tmp/e2e-delete-error-${deleted}.png`. Since the `deleted` counter is only incremented on successful deletions (line 199), multiple failures will overwrite the same file. This loses valuable debugging information for all but the last failure.

**Suggestion:** 
```suggestion
      await page.screenshot({path: `/tmp/e2e-delete-error-${apps.indexOf(app)}.png`})
```

Using the app's index in the array ensures each screenshot is unique. Alternatively, consider using a timestamp or sanitized app name for even better debuggability:
```typescript
await page.screenshot({path: `/tmp/e2e-delete-error-${app.name.replace(/[^a-z0-9]/gi, '_')}-${Date.now()}.png`})
```

packages/e2e/scripts/cleanup-test-apps.ts
```diff
@@ -0,0 +1,264 @@
...
+    }
+  }
+
+  console.log(`\n--- Done: deleted ${deleted}/${apps.length} apps ---`)
+  await browser.close()
+}
+
+async function deleteApp(page: Page, appUrl: string, _appName: string): Promise<void> {
```
dmerand: 🎨 **Code Style**: The `_appName` parameter is never used in the function body (correctly prefixed with underscore to suppress linting warnings). Since the app name isn't needed by the deletion logic, the parameter can be removed entirely to simplify the function signature.

**Suggestion:** 
```suggestion
async function deleteApp(page: Page, appUrl: string): Promise<void> {
```

Also update the call site at line 195 to remove the app.name argument:
```typescript
await deleteApp(page, app.url)
```

packages/e2e/scripts/cleanup-test-apps.ts
```diff
@@ -0,0 +1,264 @@
...
+    await page.waitForTimeout(500)
+  }
+
+  // Click the final delete/confirm button in the dialog
+  const confirmButton = page.locator('button:has-text("Delete app")').last()
+  await confirmButton.click()
+  await page.waitForTimeout(3000)
+}
+
+function waitForText(getOutput: () => string, text: string, timeoutMs: number): Promise<void> {
```
dmerand: 📐 **Design**: The `waitForText()` function (lines 245-259) reimplements functionality that already exists in `helpers/wait-for-text.ts`. The existing helper is used by `setup/auth.ts` and provides better error messages that include the actual output when timeouts occur. This duplication means bug fixes or improvements to the helper won't be reflected here.

**Suggestion:** Remove the local `waitForText` function (lines 245-259) and import the existing helper:

```suggestion
import {waitForText} from '../helpers/wait-for-text.js'
```

Add this import at the top of the file, then delete the function definition at lines 245-259.

dmerand (changes requested):
Looks like there are a few cleanup opportunities here that we should take.

---
_Review assisted by [pair-review](https://github.com/in-the-loop-labs/pair-review)_

packages/e2e/scripts/cleanup-test-apps.ts
```diff
@@ -0,0 +1,243 @@
...
+const filterIdx = process.argv.indexOf('--filter')
+if (filterIdx >= 0 && !process.argv[filterIdx + 1]) {
+  console.error('--filter requires a value')
+  process.exit(1)
+}
+const filterPattern = filterIdx >= 0 ? process.argv[filterIdx + 1] : undefined
+const headed = process.argv.includes('--headed') || !process.env.CI
+
+// Load .env
+const envPath = path.join(__dirname, '../.env')
```
ryancbahan, line 33: going to update this to use dotenv, but as a separate pr since dotenv isn't a dep of this package yet.

dmerand (approved):
Thanks for changes!

PR #6924, opened 2026-03-02
Fix theme dev shortcut to editor tracking non page requests

EvilGenius13:

Fixes https://community.shopify.dev/t/shopify-cli-theme-dev-editor-shortcut-loading-ajax-url/29157/5

When running `theme dev`, non-navigational requests like AJAX/fetch requests could be tracked as the `lastRequestedPath`. This caused the theme editor (E) shortcut to open the editor pointing to an AJAX endpoint (e.g., /cart/add.js or section URL instead of the actual page the user was viewing.

Use `Sec-Fetch-Mode` HTTP [header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Sec-Fetch-Mode) to update the `lastRequestedPath` only if it's a `navigation` type (like clicking links, typing a URL in the address bar)

### How to test your changes?
You can replicate the issue by making a new file `/templates/product.ajax-test.liquid` and add this code

<details>

```
{%- layout none -%}
<div class="ajax-response">
  <h2>{{ product.title }}</h2>
  <p>{{ product.description | strip_html | truncate: 100 }}</p>
  <span class="price">{{ product.price | money }}</span>
</div>
```

Then edit your `layout/theme.liquid` by adding this code

```
<script>
      document.addEventListener('DOMContentLoaded', function() {
        if (window.location.pathname.startsWith('/products/')) {
          fetch(window.location.pathname + '?view=ajax-test')
            .then(response => response.text())
            .then(data => console.log('AJAX response loaded:', data.substring(0, 100) + '...'))
            .catch(err => console.error('AJAX fetch error:', err));
        }
      });
    </script>
```

</details>

Run theme dev, navigate to a product page and then try the shortcut key to the editor. You should get an error.
Afterwards, build and run this branch and try the same thing, you should see the product page as expected.

### Measuring impact

### Checklist

### DIFF
--- a/.changeset/eighty-papers-hunt.md
+++ b/.changeset/eighty-papers-hunt.md
@@ -1,0 +1,5 @@
+---
+'@shopify/theme': patch
+---
+
+Fix theme editor shortcut tracking fetch requests instead of page navigation
--- a/packages/theme/src/cli/utilities/theme-environment/html.test.ts
+++ b/packages/theme/src/cli/utilities/theme-environment/html.test.ts
@@ -39,12 +39,12 @@ describe('getHtmlHandler', async () => {
     lastRequestedPath: '',
   } as unknown as DevServerContext
 
-  test('sets lastRequestedPath to the rendering URL', async () => {
+  test('sets lastRequestedPath when Sec-Fetch-Mode is navigate', async () => {
     const handler = getHtmlHandler(theme, ctx)
 
     expect(ctx.lastRequestedPath).toStrictEqual('')
 
-    const event = createH3Event('GET', '/search?q=foo&options%5Bprefix%5D=last')
+    const event = createH3Event('GET', '/search?q=foo&options%5Bprefix%5D=last', {'sec-fetch-mode': 'navigate'})
 
     vi.mocked(render).mockResolvedValueOnce(
       new Response('', {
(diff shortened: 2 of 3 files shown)


tyleralsbury:
Claude review:

## PR #6924: Fix theme dev shortcut to editor tracking non page requests

**Summary:** Good bug fix. The approach of using `Sec-Fetch-Mode: navigate` to distinguish real page navigations from AJAX/fetch requests is the right solution. Two issues need addressing before merge.

---

### Issue 1: `console.log` left in production code

`packages/theme/src/cli/utilities/theme-environment/html.ts` — the new `isNavigationRequest` function contains a `console.log`:

```ts
function isNavigationRequest(event: H3Event): boolean {
  console.log('Sec-Fetch-Mode:', event.headers.get('sec-fetch-mode'))
  return event.headers.get('sec-fetch-mode') === 'navigate'
}
```

This looks like a debug leftover. It should either be removed entirely or replaced with `outputDebug()` (which is already imported and used elsewhere in this file) if the logging is intentionally desired.

---

### Issue 2: Behavioral regression — `lastRequestedPath` will never be set when the header is missing

The old code unconditionally set `lastRequestedPath` on every request. The new code only sets it when `Sec-Fetch-Mode` is exactly `"navigate"`. This means if the header is absent entirely (e.g., older browsers, curl, non-browser clients, or requests made with `mode: 'no-cors'`), `lastRequestedPath` will never get updated and will stay at its initial empty-string value.

The existing test at line 42 (`'sets lastRequestedPath to the rendering URL'`) is updated in the PR to pass `{'sec-fetch-mode': 'navigate'}`, but there's no coverage for the case where the header is completely absent (vs. present but non-`navigate`). Consider whether the check should be:

```ts
function isNavigationRequest(event: H3Event): boolean {
  const mode = event.headers.get('sec-fetch-mode')
  // If the header is missing (non-browser client), default to treating it as navigation
  return mode === null || mode === 'navigate'
}
```

This would preserve the original behavior for clients that don't send `Sec-Fetch-Mode` while still filtering out AJAX requests (which always send the header with a non-`navigate` value like `cors`, `no-cors`, or `same-origin`). The current implementation is probably fine for the real-world use case (browsers always send this header), but it's a subtle behavioral change worth being intentional about.

---

### Minor notes

- Changeset is present and correctly scoped — good.
- Tests cover both the positive case (navigate) and negative case (cors) — good.
- The `createH3Event` helper was nicely extended to accept headers.

**Verdict:** Fix the `console.log`, consider the missing-header edge case, and this is good to ship.

EvilGenius13:
Thanks for the claude review @tyleralsbury. 🤦 on the `console.log`. That's been fixed. As for the part about the headers. This header should be [available](https://caniuse.com/?search=sec-fetch-mode) on all modern browsers (at least 2023 onwards). Since users are navigating links via browser, I would expect this is okay.  Happy to hear if there are any differing opinions.

PR #6929, opened 2026-03-03
Fix flaky AutocompletePrompt tests under React 19 on CI

tewson:
## Summary
- Fix the `Stdin` mock to re-emit `'readable'` when a listener is added and there's pending data. This mirrors real Node stream behavior and prevents dropped input when data is written before Ink's `useInput` effect registers its listener — the root cause of the "allows searching with pagination" timeout on slow CI runners.
- Replace `sendInputAndWait` fixed delays in the "doesn't submit if there are no choices" and "has a loading state" tests with `sendInputAndWaitForContent`, which deterministically waits for React to confirm the frame still shows expected content after the keypress
- Use a never-resolving search promise in the "has a loading state" test so the loading state persists for the entire test, eliminating races where React 19's batched rendering could resolve the search and transition back to Idle before the ENTER keypress is processed

## Test plan
- [x] `pnpm --filter @shopify/cli-kit vitest run src/private/node/ui/components/AutocompletePrompt.test.tsx` passes locally
- [x] `pnpm --filter @shopify/app vitest run src/cli/services/dev/ui/components/Dev.test.tsx src/cli/services/dev/ui/components/DevSessionUI.test.tsx` passes locally
- [ ] CI passes without flaky failures

### DIFF
--- a/packages/cli-kit/src/private/node/testing/ui.ts
+++ b/packages/cli-kit/src/private/node/testing/ui.ts
@@ -23,6 +23,17 @@ export class Stdin extends EventEmitter {
   constructor(options: {isTTY?: boolean} = {}) {
     super()
     this.isTTY = options.isTTY ?? true
+
+    // When a 'readable' listener is added and there's pending data,
+    // re-emit 'readable' so the new listener can read it. This mirrors
+    // real Node streams where buffered data is available to new readers,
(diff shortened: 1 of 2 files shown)


tewson:
Done with https://github.com/Shopify/cli/pull/6936

PR #6934, opened 2026-03-03
Add `app validate` command

ryancbahan:

Adds a `shopify app validate` command that checks app configuration and extension TOMLs for errors without running a full `dev` or `deploy`. This gives developers (and agents) a fast, non-interactive way to verify config correctness.

- **New command:** **`app validate`** (`packages/app/src/cli/commands/app/validate.ts`)
    - Extends `AppLinkedCommand` so it links to the user's app via `linkedAppContext()` (same flow as `deploy`, `dev`, etc.)
    - Passes `unsafeReportMode: true` to collect all validation errors instead of failing on the first one
    - Renders a success message if config is valid, or a formatted error summary + `AbortError` if not
    - Supports standard app flags: `--path`, `--config`, `--client-id`, `--reset`
- **Service layer** (`packages/app/src/cli/services/validate.ts`)
    - `validateApp()` reads `app.errors`, formats them, and renders success/error output
- **Registered in CLI** — added to `packages/app/src/cli/index.ts` command map and oclif manifest

#### Demo

[Screen Recording 2026-03-03 at 4.23.01 PM.mov <span class="graphite__hidden">(uploaded via Graphite)</span> ](https://app.graphite.com/user-attachments/video/9a306762-b4a0-4bd8-bc07-7a47b8b5d453.mov)

### How to test your changes?

1. In an app directory with a valid `shopify.app.toml`, run `shopify app validate` — should see success message
2. Introduce an error in the TOML (e.g. remove a required field) and run again — should see formatted validation errors and a non-zero exit code
3. Test with `--config` flag pointing to a specific config file
4. Run `npx vitest run packages/app/src/cli/services/validate.test.ts` — 3 tests covering valid, invalid, and empty-errors cases

### Measuring impact

### Checklist

### DIFF
--- a/packages/app/src/cli/commands/app/validate.ts
+++ b/packages/app/src/cli/commands/app/validate.ts
@@ -1,0 +1,34 @@
+import {appFlags} from '../../flags.js'
+import {validateApp} from '../../services/validate.js'
+import AppLinkedCommand, {AppLinkedCommandOutput} from '../../utilities/app-linked-command.js'
+import {linkedAppContext} from '../../services/app-context.js'
+import {globalFlags} from '@shopify/cli-kit/node/cli'
+
+export default class Validate extends AppLinkedCommand {
+  static summary = 'Validate your app configuration and extensions.'
+
+  static descriptionWithMarkdown = `Validates the selected app configuration file and all extension configurations against their schemas and reports any errors found.`
+
+  static description = this.descriptionWithoutMarkdown()
+
+  static flags = {
+    ...globalFlags,
+    ...appFlags,
+  }
+
+  public async run(): Promise<AppLinkedCommandOutput> {
+    const {flags} = await this.parse(Validate)
+
+    const {app} = await linkedAppContext({
+      directory: flags.path,
+      clientId: flags['client-id'],
+      forceRelink: flags.reset,
+      userProvidedConfigName: flags.config,
+      unsafeReportMode: true,
+    })
+
+    await validateApp(app)
+
+    return {app}
+  }
+}
--- a/packages/app/src/cli/services/validate.ts
+++ b/packages/app/src/cli/services/validate.ts
@@ -1,0 +1,22 @@
+import {AppLinkedInterface} from '../models/app/app.js'
+import {stringifyMessage} from '@shopify/cli-kit/node/output'
+import {renderError, renderSuccess} from '@shopify/cli-kit/node/ui'
+import {AbortSilentError} from '@shopify/cli-kit/node/error'
+
+export async function validateApp(app: AppLinkedInterface): Promise<void> {
+  const errors = app.errors
+
+  if (!errors || errors.isEmpty()) {
+    renderSuccess({headline: 'App configuration is valid.'})
+    return
+  }
+
+  const errorMessages = errors.toJSON().map((error) => stringifyMessage(error).trim())
+
+  renderError({
+    headline: 'Validation errors found.',
+    body: errorMessages.join('\n\n'),
+  })
+
+  throw new AbortSilentError()
+}
--- a/docs-shopify.dev/generated/generated_docs_data.json
+++ b/docs-shopify.dev/generated/generated_docs_data.json
@@ -3085,6 +3085,98 @@
     "related": []
   },
   {
+    "name": "app validate",
+    "description": "Validates the selected app configuration file and all extension configurations against their schemas and reports any errors found.",
+    "overviewPreviewDescription": "Validate your app configuration and extensions.",
+    "type": "command",
+    "isVisualComponent": false,
+    "defaultExample": {
+      "codeblock": {
+        "tabs": [
+          {
+            "title": "app validate",
+            "code": "shopify app validate [flags]",
+            "language": "bash"
+          }
+        ],
+        "title": "app validate"
+      }
+    },
+    "definitions": [
+      {
+        "title": "Flags",
+        "description": "The following flags are available for the `app validate` command:",
+        "type": "appvalidate",
+        "typeDefinitions": {
+          "appvalidate": {
+            "filePath": "docs-shopify.dev/commands/interfaces/app-validate.interface.ts",
+            "name": "appvalidate",
+            "description": "",
+            "members": [
+              {
+                "filePath": "docs-shopify.dev/commands/interfaces/app-validate.interface.ts",
+                "syntaxKind": "PropertySignature",
+                "name": "--client-id <value>",
+                "value": "string",
+                "description": "The Client ID of your app.",
+                "isOptional": true,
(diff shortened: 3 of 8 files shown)


ryancbahan:
* **#7040** <a href="https://app.graphite.com/github/pr/Shopify/cli/7040?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6979** <a href="https://app.graphite.com/github/pr/Shopify/cli/6979?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6934** <a href="https://app.graphite.com/github/pr/Shopify/cli/6934?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Shopify/cli/6934?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/app/src/cli/commands/app/validate.ts
dmerand, line 1: `app validate` currently goes through the linked-app path rather than the local-only loader. In [[file:packages/app/src/cli/commands/app/validate.ts:18-27]] the command extends `AppLinkedCommand` and calls `linkedAppContext(...)`. That means validation still inherits the linked/remote behavior in [[file:packages/app/src/cli/services/app-context.ts:64-121]]: it can relink when the app is unlinked (`if (!configState.isLinked || forceRelink) { ... link(...) }`), fetch the remote app via `appFromIdentifiers(...)`, and fetch organization data via `fetchOrgFromId(...)` before validation runs.

The surrounding code suggests this split is intentional. [[file:packages/app/src/cli/services/app-context.ts:147-160]] has a separate `localAppContext()` specifically for loading an app without network calls or linking, and [[file:packages/app/src/cli/services/app-context.test.ts:287-304]] explicitly tests that it does **not** call `appFromIdentifiers`, `fetchOrgFromId`, or `link`. `unsafeReportMode` changes config parsing behavior, but it does not make this command local or non-interactive.

That seems like a product/expectation mismatch. The PR reads like a fast local config validator, but this implementation can still depend on linked state, authentication, and relinking. If the goal is true local validation, this probably wants `AppUnlinkedCommand` plus `localAppContext()` and a test proving no link/network calls occur. If the goal is linked-app validation only, I think the help text and tests should say that explicitly and cover unlinked/auth-required behavior.

packages/app/src/cli/commands/app/validate.ts
```diff
@@ -0,0 +1,34 @@
+import {appFlags} from '../../flags.js'
+import {validateApp} from '../../services/validate.js'
+import AppLinkedCommand, {AppLinkedCommandOutput} from '../../utilities/app-linked-command.js'
+import {linkedAppContext} from '../../services/app-context.js'
+import {globalFlags} from '@shopify/cli-kit/node/cli'
+
+export default class Validate extends AppLinkedCommand {
+  static summary = 'Validate your app configuration and extensions.'
+
+  static descriptionWithMarkdown = `Validates your app's \`shopify.app.toml\` and all extension configurations against their schemas and reports any errors found.`
```
dmerand: 💡 **Improvement**: The help text currently says the command validates `shopify.app.toml`, but the implementation accepts `--config` and forwards that selected config name into app loading. Because this string also feeds generated help artifacts, reviewers may want the wording to match the broader behavior so docs, README output, and manifest metadata stay accurate.

**Suggestion:** Consider broadening the description so it reflects both default and explicit config-file selection:
```suggestion
  static descriptionWithMarkdown = `Validates the selected app configuration file and all extension configurations against their schemas and reports any errors found.`
```

packages/app/src/cli/commands/app/validate.ts
```diff
@@ -0,0 +1,34 @@
...
+
+  static flags = {
+    ...globalFlags,
+    ...appFlags,
+  }
+
+  public async run(): Promise<AppLinkedCommandOutput> {
+    const {flags} = await this.parse(Validate)
+
+    const {app} = await linkedAppContext({
```
dmerand, line 22: 🐛 **Bug**: Worth reviewing because the command appears to opt into report mode, but the first `shopify.app.toml` parse still happens earlier in the load pipeline. If the top-level app config is malformed or missing required fields, the command exits before `validateApp(app)` runs, so reviewers may want to verify that the new command really can report the full validation set it advertises.

**Suggestion:** Move the relaxed/report-mode boundary earlier in the app-loading flow so invalid root app config can still be materialized into `app.errors` instead of throwing during initial configuration-state loading. Reusing the existing `loadOpaqueApp(..., {mode: 'report'})` pattern, or threading report mode into `getAppConfigurationState()`, would align the command with its intended behavior.

packages/app/src/cli/services/validate.ts
```diff
@@ -0,0 +1,22 @@
...
+  }
+
+  const errorMessages = errors.toJSON().map((error) => stringifyMessage(error).trim())
+
+  renderError({
+    headline: 'Validation errors found.',
+    body: errorMessages.join('\n\n'),
+  })
+
+  throw new AbortError('App configuration is invalid.')
```
dmerand: This service already renders the full validation report itself in [[file:packages/app/src/cli/services/validate.ts:16-19]], so throwing a normal `AbortError` afterward likely sends execution through the standard CLI fatal-error path as well. That would leave users with the detailed validation banner followed by a second generic failure banner, which is noisy and less informative.

There is already a matching pattern elsewhere in the app codebase for this case: when code manually renders its own error banner, it throws `AbortSilentError` instead of `AbortError` so the process still exits non-zero without duplicate output. See [[file:packages/app/src/cli/services/release/version-diff.ts:31-39]].

I think this should follow that same pattern here:

```suggestion
  throw new AbortSilentError()
```

The test should move with it too. [[file:packages/app/src/cli/services/validate.test.ts:27-33]] currently expects `AbortError`, which appears to lock in the wrong behavior rather than catching the duplicate-render case.

packages/app/src/cli/commands/app/validate.ts
ryancbahan: Intentional. It's still "local" as in doesn't run dev/deploy, I can update the pr desc. But we need extension validation which requires some network hop (for now).

packages/app/src/cli/commands/app/validate.ts
ryancbahan: Loading flow changes in https://app.graphite.com/github/pr/Shopify/cli/7022 will address this.

packages/app/src/cli/commands/app/validate.ts
dmerand: Makes sense + that's what I figured. Let's make sure the PR is accurate so later context retrieval isn't confused.

PR #6933, opened 2026-03-03
Fix --alias to also match by userId for UUID-aliased sessions

GeorgiZhelev:
## WHY are these changes introduced?

When a session's email alias fails to resolve during login, the `alias` field ends up as `undefined`. The session list display falls back to showing `userId` (UUID) via `session.identity.alias ?? userId`. But `findSessionByAlias` only matches against `session.identity.alias`, so `--alias <UUID>` doesn't find anything and falls through to the interactive prompt.

(The personal context is that I have a few different staff/partner accounts I work through so in every separate theme folder on my machine I automatically run `shopify auth login --alias` with the appropriate account - around 15 times a day).

## WHAT is this pull request doing?

`findSessionByAlias` now also matches on the session's `userId` key as a fallback, so `--alias <UUID>` works for sessions where the alias was never set.

One-line change in `store.ts`, one new test in `store.test.ts`.

Note: this does not fix the display issue — sessions with undefined aliases will still show as UUIDs in the session picker. That would require re-fetching the email at some point (e.g. during token refresh).

## How to test your changes?

- `pnpm -F @shopify/cli-kit test -- src/private/node/session/store.test.ts`
- Manual: on the live build, `shopify auth login --alias "<UUID>"` falls through to the interactive prompt instead of selecting the session. On the patched build, it selects the session correctly. 
image

## Measuring impact

- [x] n/a - this doesn't need measurement, e.g. a linting rule or a bug-fix
- [ ] Existing analytics will cater for this addition
- [ ] PR includes analytics changes to measure impact

## Checklist

- [x] I've considered possible cross-platform impacts (Mac, Linux, Windows)
- [x] I've considered possible [documentation](https://shopify.dev) changes

PR #6935, opened 2026-03-04
Ask for an alias on login when the email is not found

gonzaloriestra:

Apparently, sometimes the sessions are stored with the UUID (the fallback) instead of the email, or overwritten after some time (when refreshing).

See discussion in https://github.com/Shopify/cli/pull/6933

- Avoid overwriting the session alias on refresh, but only when running `shopify auth login`
- If the email can't be fetched, ask for a manual alias instead of using the UUID:
04-50-022h2-44gy1

### How to test your changes?

`shopify auth login`

### Measuring impact

### Checklist

### DIFF
--- a/packages/cli-kit/src/private/node/session.ts
+++ b/packages/cli-kit/src/private/node/session.ts
@@ -320,7 +320,7 @@ async function executeCompleteFlow(applications: OAuthApplications, existingAlia
 
   // Preserve existing alias if available, otherwise try fetching email
   const businessPlatformToken = result[applicationId('business-platform')]?.accessToken
-  const alias = existingAlias ?? (await fetchEmail(businessPlatformToken))
+  const alias = existingAlias ?? (await fetchEmail(businessPlatformToken)) ?? identityToken.userId
 
   const session: Session = {
     identity: {
@@ -352,7 +352,7 @@ async function refreshTokens(session: Session, applications: OAuthApplications):
   )
 
   return {
-    identity: identityToken,
+    identity: {...identityToken, alias: session.identity.alias},
     applications: applicationTokens,
   }
 }
--- a/packages/cli-kit/src/private/node/session.test.ts
+++ b/packages/cli-kit/src/private/node/session.test.ts
@@ -228,7 +228,7 @@ The CLI is currently unable to prompt for reauthentication.`,
     expect(fetchSessions).toHaveBeenCalledOnce()
   })
 
-  test('leaves alias undefined when email fetch fails', async () => {
+  test('falls back to userId as alias when email fetch fails', async () => {
     // Given
     vi.mocked(validateSession).mockResolvedValueOnce('needs_full_auth')
     vi.mocked(fetchSessions).mockResolvedValue(undefined)
@@ -243,7 +243,7 @@ The CLI is currently unable to prompt for reauthentication.`,
     expect(storeSessions).toHaveBeenCalledOnce()
 
     const storedSession = vi.mocked(storeSessions).mock.calls[0]![0]
-    expect(storedSession[fqdn]![userId]!.identity.alias).toBeUndefined()
+    expect(storedSession[fqdn]![userId]!.identity.alias).toBe(userId)
 
     expect(got).toEqual(validTokens)
   })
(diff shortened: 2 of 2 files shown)


packages/cli-kit/src/private/node/session.ts
```diff
@@ -35,7 +35,7 @@ async function fetchEmail(businessPlatformToken: string | undefined): Promise<st
 
   try {
     const userEmailResult = await businessPlatformRequest<UserEmailQuery>(UserEmailQueryString, businessPlatformToken)
-    return userEmailResult.currentUserAccount?.email
+    return userEmailResult.currentUserAccount?.email ?? undefined
```
alfonso-noriega, line 38: ⭐ **Praise**: Adding `?? undefined` correctly handles the case where the GraphQL API returns `email: null`. Without this, a `null` value would propagate despite the function's `Promise<string | undefined>` return type. This makes the function's contract cleaner and prevents downstream issues where code checks for `undefined` but receives `null`.

packages/cli-kit/src/private/node/session.ts
```diff
@@ -230,15 +230,15 @@ ${outputToken.json(applications)}
   if (validationResult === 'needs_full_auth') {
     await throwOnNoPrompt(noPrompt)
     outputDebug(outputContent`Initiating the full authentication flow...`)
-    newSession = await executeCompleteFlow(applications)
+    newSession = await executeCompleteFlow(applications, currentSession?.identity.alias)
```
alfonso-noriega, line 233: 💬 **Suggestion**: Even after fixing the userId fallback, there's a secondary concern: when `existingAlias` is a UUID (from a previous login where email fetch failed), re-authentication via `needs_full_auth` will short-circuit the email fetch entirely.

1. Previous login: email fetch fails → alias stored as userId UUID (or `undefined` after the fix above, then user provides one via prompt)
2. Session expires → `needs_full_auth`
3. `executeCompleteFlow(applications, currentSession?.identity.alias)` — `existingAlias` is the stored alias
4. `existingAlias ?? (await fetchEmail(...))` — short-circuits at `existingAlias`
5. Email is never fetched, even if the API is now available

The docstring on line 292 says `existingAlias` is "to preserve if the email fetch fails", but the code gives it unconditional priority over a fresh fetch. This means sessions with stale aliases can never self-heal.

This may be intentional if user-provided aliases should never be overwritten by email, but it's worth considering.

**Suggestion:** If you want sessions to self-heal when the email API becomes available, consider prioritizing the email fetch:
```
const alias = (await fetchEmail(businessPlatformToken)) ?? existingAlias
```

If the current behavior is intentional (always preserve existing aliases), update the docstring on line 292 to clarify:
```
 * @param existingAlias - Optional alias from a previous session. Takes priority over email fetch to preserve user-chosen aliases.
```

packages/cli-kit/src/public/node/session-prompt.ts
```diff
@@ -37,13 +37,23 @@ function buildSessionChoices(sessions: Sessions, fqdn: string): SessionChoice[]
...
  *
  * @returns The alias of the authenticated user.
  */
 async function handleNewLogin(): Promise<string> {
   const result = await ensureAuthenticatedUser({}, {forceNewSession: true})
   const alias = await sessionStore.getSessionAlias(result.userId)
-  return alias ?? result.userId
+
+  if (!alias) {
+    const userAlias = await renderTextPrompt({
```
alfonso-noriega, line 49: 💡 **Improvement**: `renderTextPrompt` is called without validation. If the user submits an empty string, it would be stored as the alias. Since `buildSessionChoices` uses `session.identity.alias ?? userId`, and `??` does not catch empty strings, this would result in a blank label in the session picker. The impact is minor since the next `handleNewLogin` call would re-prompt (`!alias` is true for empty string), but the intermediate state is messy.

**Suggestion:** Consider adding validation to reject empty input (check what options `renderTextPrompt` supports in this codebase):
```javascript
const userAlias = await renderTextPrompt({
  message: 'Enter an alias for this account (e.g. your email or a nickname)',
  allowEmpty: false,
})
```

alfonso-noriega (commented):
---
_Review assisted by [pair-review](https://vault.shopify.io/posts/346224-Human-in-the-loop-code-review-with-coding-agents)_

packages/cli-kit/src/public/node/session-prompt.ts
gonzaloriestra: RenderTextPrompt doesn't allow empty values

packages/cli-kit/src/private/node/session.ts
gonzaloriestra: This is intentional, we always want to keep the original alias. If it's not found, the user will add it manually.

alfonso-noriega (approved):

gonzaloriestra:
### Merge activity

* **Mar 26, 12:04 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/Shopify/cli/6935).
* **Mar 26, 12:04 PM UTC**: @gonzaloriestra added this pull request to the GitHub merge queue with [Graphite](https://app.graphite.com/github/pr/Shopify/cli/6935).

PR #6939, opened 2026-03-05
fix(theme): remove browser user-agent from proxied requests

stephanie-shopify:
## Summary

- Fixes 429 errors merchants were seeing on `POST /cart/add.js` during `shopify theme dev`
- In HTTP/2, browsers send headers lowercase (e.g. `user-agent`). The CLI sets `User-Agent` (title-case) via `defaultHeaders()`. Without explicit deletion, both coexist as separate JS object keys and Node's `fetch` normalizes them so the **browser UA wins**
- SFR then sees a real browser UA and applies bot-mitigation/rate-limiting rather than CLI allowances
- Fix: delete `proxyRequestHeaders['user-agent']` in `getProxyStorefrontHeaders()` before the merge with `defaultHeaders()`, so CLI always controls the outgoing `User-Agent`
- Cart/checkout/account endpoints are most affected since they intentionally omit Bearer token auth

Closes https://github.com/Shopify/theme-tools/issues/1142

## Test plan

- [ ] Updated existing `getProxyStorefrontHeaders` snapshot test — `user-agent` is now removed alongside other hop-by-hop/safari headers
- [ ] Added explicit test asserting `user-agent` is undefined after `getProxyStorefrontHeaders`
- [ ] Run `pnpm --filter @shopify/theme test` locally

### DIFF
--- a/packages/theme/src/cli/utilities/theme-environment/proxy.test.ts
+++ b/packages/theme/src/cli/utilities/theme-environment/proxy.test.ts
@@ -279,10 +279,10 @@ describe('dev proxy', () => {
       event.node.req.headers.connection = '...'
       event.node.req.headers['proxy-authenticate'] = '...'
       event.node.req.headers.host = 'abnb'
+      event.node.req.headers['user-agent'] = 'Mozilla/5.0 (browser UA)'
       // Kept:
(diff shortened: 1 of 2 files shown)


stephanie-shopify:
nope. not correct

PR #6943, opened 2026-03-06
Introduce TomlFile abstraction for TOML file I/O

ryancbahan:
## Summary

Adds a general-purpose `TomlFile` class in `cli-kit` with `read`/`patch`/`remove`/`replace`/`transformRaw` methods. Restructures TOML utilities into a `toml/` directory with internal codec module (`encodeToml`/`decodeToml`).

### Why this change exists

Today, the concerns of "parse a file", "validate a file", "validate Shopify domain config", and "transform file contents to domain config" are interwoven and dispersed across callsites. This makes it very difficult to instrument changes or drive consistency in the codebase. This pr introduces a TOML file abstraction that owns a specific slice of jobs: read/write TOML, and check whether that TOML is valid syntax. It _does_ _not_ care about Shopify domain knowledge or validation. This is the first step in providing a clearer lifecycle of config read/writes, and separation between transform/validate/IO jobs. 

#### Current problems

- Each callsite making its own choices about error handling, comment preservation, and I/O.
- Serialization and I/O mixed across file boundaries. Extension builders called encodeToml() and returned TOML strings. Their caller (import-extensions.ts) called writeFile() to put those strings on disk.
The responsibility for "write a TOML file" is split between the function that knew the content and the function that knew the path.
- Multiple approaches to reads/writes: some callsites can preserve comments, others use encode/decode functions directly and cannot.
- Parse errors lost context. decodeToml throws raw @iarna/toml errors with line/column but no file path. Each callsite had to catch and re-wrap with path context independently — or didn't, producing confusing errors.

#### With this PR:

- One abstraction for all TOML file operations. read/patch/remove/replace/transformRaw. Every TOML file operation in the codebase goes through it. 
- Comment preservation is built into the model. patch and remove use toml_edit (via WASM) to preserve comments and formatting. replace + transformRaw handles the comment injection case for config link. Subsequent edits via dev/deploy/urls preserve whatever comments exist. This was previously implicit knowledge spread across callsites.
- Parse errors include file path automatically. TomlParseError wraps parse failures with the file path. Every consumer of TomlFile.read() gets good error messages without extra catch/re-throw boilerplate.
- encodeToml/decodeToml become internal. The public API surface shrinks. Consumers use TomlFile for file I/O. The codec functions are implementation details in cli-kit/node/toml/codec.ts, not something
every file in the app package imports.
- Foundation for schema validation; beginning of separation between file, validation, and transform concerns. 

## Test plan

- [x] 20 TomlFile unit tests (read/patch/remove/replace/transformRaw + error handling)
- [x] All existing cli-kit tests pass

### DIFF
--- a/packages/cli-kit/src/public/node/toml/toml-file.ts
+++ b/packages/cli-kit/src/public/node/toml/toml-file.ts
@@ -69,8 +69,9 @@ export class TomlFile {
     const patches = flattenToPatchEntries(changes)
     const raw = await readFile(this.path)
     const updated = updateTomlValues(raw, patches)
+    const parsed = this.decode(updated)
     await writeFile(this.path, updated)
-    this.content = this.decode(updated)
+    this.content = parsed
   }
 
   /**
@@ -86,8 +87,9 @@ export class TomlFile {
     const keys = keyPath.split('.')
     const raw = await readFile(this.path)
     const updated = updateTomlValues(raw, [[keys, undefined]])
+    const parsed = this.decode(updated)
     await writeFile(this.path, updated)
-    this.content = this.decode(updated)
+    this.content = parsed
   }
 
   /**
@@ -102,6 +104,7 @@ export class TomlFile {
    */
   async replace(content: JsonMapType): Promise<void> {
     const encoded = encodeToml(content)
+    this.decode(encoded)
     await writeFile(this.path, encoded)
     this.content = content
   }


ryancbahan:
* **#6980** <a href="https://app.graphite.com/github/pr/Shopify/cli/6980?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6953** <a href="https://app.graphite.com/github/pr/Shopify/cli/6953?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6952** <a href="https://app.graphite.com/github/pr/Shopify/cli/6952?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6951** <a href="https://app.graphite.com/github/pr/Shopify/cli/6951?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6950** <a href="https://app.graphite.com/github/pr/Shopify/cli/6950?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6949** <a href="https://app.graphite.com/github/pr/Shopify/cli/6949?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6947** <a href="https://app.graphite.com/github/pr/Shopify/cli/6947?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6945** <a href="https://app.graphite.com/github/pr/Shopify/cli/6945?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6944** <a href="https://app.graphite.com/github/pr/Shopify/cli/6944?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6943** <a href="https://app.graphite.com/github/pr/Shopify/cli/6943?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Shopify/cli/6943?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

phyllis-sy-wu:
Tophatted, everything looked good with no errors or weird behaviour 

**Scenarios tested** (in cli repo: pnpm shopify app xxxxx --path <app-path>):
- App without extensions → app dev → app deploy ✅
- App with extensions (checkout-ui, product-configuration, theme-extension, discount-function) → app dev → app deploy ✅
- Compared with prod CLI (shopify app <command>)

**Result:** 
- Behaviour aligned with apps created/developed from prod CLI
- Config was read and updated as expected
- TOML stayed valid and comments were preserved
- Deploy completed with no validation errors (payload accepted).

packages/cli-kit/src/public/node/toml/toml-file.ts
```diff
@@ -0,0 +1,157 @@
...
+   * @param transform - A function that receives the raw TOML string and returns the modified string.
+   * @example
+   * ```ts
+   * await file.transformRaw((raw) => `# Header comment\n${raw}`)
+   * ```
+   */
+  async transformRaw(transform: (raw: string) => string): Promise<void> {
+    const raw = await readFile(this.path)
+    const transformed = transform(raw)
+    await writeFile(this.path, transformed)
```
isaacroldan, line 130: shouldn't we validate that the transformed string is a valid TOML?

ryancbahan: good catch, will update

packages/cli-kit/package.json
```diff
@@ -24,6 +24,10 @@
       "import": "./dist/index.js",
       "types": "./dist/index.d.ts"
     },
+    "./node/toml": {
+      "node": "./dist/public/node/toml/index.js",
+      "types": "./dist/public/node/toml/index.d.ts"
+    },
```
isaacroldan, line 30: you shouldn't need this, is already covered by the wildcard below :thinking_face:

ryancbahan: hmm, IIRC imports from the new toml dir didn't work without this

packages/cli-kit/src/public/node/toml/toml-file.ts
```diff
@@ -0,0 +1,162 @@
...
+    try {
+      return decodeToml(raw)
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    } catch (err: any) {
+      if (err.line !== undefined && err.col !== undefined) {
+        throw new TomlParseError(path, err)
+      }
+      throw err
+    }
+  }
```
isaacroldan: Are we passing the path here just for the error string? I think this is a bit confusing, I was reading the code, and a call to this function is like `decode(path, content)`, which at first sight for me it meant decode the whatever is at `path`but then I saw we were passing a content too. Do we need the `path` here?   
  
I guess it's fine since it's a private method, but still a bit confusing the first time you see a call to decode

packages/cli-kit/package.json
ryancbahan: yea this is necessary to import from `@shopify/cli-kit/node/toml which the next pr does

packages/cli-kit/src/public/node/toml/toml-file.ts
ryancbahan: Yeah that's fair feedback, I'll clean this up

packages/cli-kit/package.json
isaacroldan: all other modules work with the wildcard export, for instance `/node/crypto` or `/node/path`  
I guess the issue is having an index file, can we do this without an index file and expose a `toml.ts` directly?

ryancbahan: node/crypto and node/path both point to files in the public/node root, not subfolders. i put the toml files in their own subdir since it's a domain with multiple file instances and i didn't want to keep expanding the root, which is already very large.

ryancbahan: i guess we could do re-exports from the root, but i don't see much value in it

packages/cli-kit/src/public/node/toml/toml-file.ts
```diff
@@ -0,0 +1,163 @@
...
+   * ```ts
+   * await file.remove('build.include_config_on_deploy')
+   * ```
+   */
+  async remove(keyPath: string): Promise<void> {
+    const keys = keyPath.split('.')
+    const raw = await readFile(this.path)
+    const updated = updateTomlValues(raw, [[keys, undefined]])
+    await writeFile(this.path, updated)
+    this.content = this.decode(updated)
```
dmerand: Should we decode before writing to disk to ensure validity?

packages/cli-kit/src/public/node/toml/toml-file.ts
```diff
@@ -0,0 +1,163 @@
...
+   * ```ts
+   * await file.patch({build: {dev_store_url: 'my-store.myshopify.com'}})
+   * await file.patch({application_url: 'https://example.com', auth: {redirect_urls: ['...']}})
+   * ```
+   */
+  async patch(changes: {[key: string]: unknown}): Promise<void> {
+    const patches = flattenToPatchEntries(changes)
+    const raw = await readFile(this.path)
+    const updated = updateTomlValues(raw, patches)
+    await writeFile(this.path, updated)
```
dmerand, line 73: Same comment here about decoding before writing to disk.

packages/cli-kit/src/public/node/toml/toml-file.ts
ryancbahan: I thought about this and didn't because we're effectively testing `@shopify/toml-patch` for that. but on reflection I think the hardening is better in any case. updating

packages/cli-kit/package.json
isaacroldan: Just worried of having to manually export things here, can grow a lot very quickly, let's leave it for now and we can revisit if this grows too much.

ryancbahan: Just curious -- why? The cost of change is basically zero to update an import alias/etc. I personally prefer more structured directories, as it's easier to visually navigate.

isaacroldan: just a preference of having automatic exports (or maybe an index file) and not needing to update the package.json for every folder/file we add to the project

PR #6944, opened 2026-03-06
Migrate all TOML I/O callsites to TomlFile

ryancbahan:
## Summary

Migrates every TOML file I/O callsite to use `TomlFile`:

- `dev.ts`, `context.ts`, `urls.ts`: `setAppConfigValue`/`unsetAppConfigValue` → `TomlFile.patch()`/`remove()`
- `write-app-configuration-file.ts`: `encodeToml` + `writeFileSync` → `TomlFile.replace()` + `transformRaw()`
- `add-uid-to-extension-toml.ts`: raw `readFile`/`writeFile` → `TomlFile.read()` + `patch()`/`transformRaw()`
- `breakdown-extensions.ts`: `encodeToml` → regex → field names replaced with `Object.keys()` on diff objects
- Extension builders: return objects instead of TOML strings
- `loader.ts`: removed unused `decode` parameter from `loadConfigurationFileContent`/`parseConfigurationFile`
- `environments.ts`: `decodeToml` → `TomlFile.read()`
- `patch-app-configuration-file.ts`: removed all TOML functions (only `patchAppHiddenConfigFile` remains)

`encodeToml`/`decodeToml` are no longer imported outside `cli-kit`.

## Test plan

- [x] All 1688 app tests pass (services + models)
- [x] All cli-kit tests pass
- [x] TypeScript type checks pass for both packages


ryancbahan:
* **#6980** <a href="https://app.graphite.com/github/pr/Shopify/cli/6980?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6953** <a href="https://app.graphite.com/github/pr/Shopify/cli/6953?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6952** <a href="https://app.graphite.com/github/pr/Shopify/cli/6952?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6951** <a href="https://app.graphite.com/github/pr/Shopify/cli/6951?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6950** <a href="https://app.graphite.com/github/pr/Shopify/cli/6950?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6949** <a href="https://app.graphite.com/github/pr/Shopify/cli/6949?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6947** <a href="https://app.graphite.com/github/pr/Shopify/cli/6947?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6945** <a href="https://app.graphite.com/github/pr/Shopify/cli/6945?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6944** <a href="https://app.graphite.com/github/pr/Shopify/cli/6944?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Shopify/cli/6944?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#6943** <a href="https://app.graphite.com/github/pr/Shopify/cli/6943?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

packages/app/src/cli/services/subscription_link/extension-to-toml.ts
```diff
@@ -25,6 +24,5 @@ export function buildTomlObject(extension: ExtensionRegistration): string {
       },
     ],
   }
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  return encodeToml(localExtensionRepresentation as any)
+  return localExtensionRepresentation
```
isaacroldan, line 27: Wait, so then `buildTomlObject` doesn't build a toml object anymore no? it builds a JSON that might or might not be converted to toml. Should this `buildTomlObject` return a TomlFile? should we rename it?

isaacroldan: same for all the other `buildTomlObject` that exist in other files, seems like these are the main changes in this PR (by number of lines), should this function just use TomlFile and in the tests, instead of validating the JSON, validate the TomlFile raw content?

ryancbahan: The nex pr in the stack renames the functions to indicate that it's not specifically toml - I just didn't want to include here as the function reference change would make this pr more indirect. I went with a name change instead of consuming TOML directly but am happy to talk about it. I think it makes sense to pass around object reference and convert to/from toml in narrower pipelines.

packages/app/src/cli/services/context/breakdown-extensions.ts
dmerand, line 1: 🧑‍🍳 💋 to how much simpler this is.

packages/app/src/cli/services/dev/urls.ts
```diff
@@ -187,20 +187,21 @@ export async function updateURLs(
   }
 
   if (localApp && localApp.configuration.client_id === apiKey) {
-    const configValues = [
-      {keyPath: 'application_url', value: urls.applicationUrl},
-      {keyPath: 'auth.redirect_urls', value: urls.redirectUrlWhitelist},
-    ]
+    const configFile = await TomlFile.read(localApp.configuration.path)
```
dmerand, line 190: 💡 **Improvement**: The current implementation builds the patch object with conditional property assignment—creating a mutable `patch` variable that's modified across different code paths. While clear, this could be more concise using object spread to make the conditional inclusion of `app_proxy` more explicit and avoid mutating state.

**Suggestion:** Consider using a single object spread:
```typescript
const configFile = await TomlFile.read(localApp.configuration.path)
await configFile.patch({
  application_url: urls.applicationUrl,
  auth: {redirect_urls: urls.redirectUrlWhitelist},
  ...(urls.appProxy && {
    app_proxy: {
      url: urls.appProxy.proxyUrl,
      subpath: urls.appProxy.proxySubPath,
      prefix: urls.appProxy.proxySubPathPrefix,
    },
  }),
})
```
This makes it clearer that app_proxy is conditionally included and avoids mutating the patch object.

packages/app/src/cli/services/import-extensions.ts
```diff
@@ -109,9 +111,10 @@ export async function importExtensions(options: ImportOptions) {
     extensionUuids[handle] = ext.uuid
 
     if (action === DirectoryAction.Write) {
-      const tomlObject = buildTomlObject(ext, extensions, app.configuration)
-      const path = joinPath(directory, 'shopify.extension.toml')
-      await writeFile(path, tomlObject)
+      const tomlContent = buildTomlObject(ext, extensions, app.configuration)
+      const tomlPath = joinPath(directory, 'shopify.extension.toml')
+      const file = new TomlFile(tomlPath, tomlContent as JsonMapType)
```
dmerand, line 116: 🐛 **Bug**: Line 116 creates a TomlFile with `tomlContent` as the second argument, but line 117 immediately calls `replace()` with the same content. The TomlFile constructor sets `this.content = content`, so passing `tomlContent` to the constructor is redundant since it's overwritten by `replace()`. This pattern differs from other parts of the codebase where TomlFile is initialized with an empty object before replacement.

**Suggestion:** 
```suggestion
      const file = new TomlFile(tomlPath, {})
```

packages/app/src/cli/services/dev/urls.ts
ryancbahan: yeah, there's improvement to be done here. trying to mirror pre-existing behavior 1:1 first and then keep addressing this stuff, rather than mixing concerns.

PR #6948, opened 2026-03-06
Gate tools and instructions behind Remote DOM api_version check

vividviolet:
### What:
Only add tools and instructions to the build_manifest for ui_extension when isRemoteDomExtension is true (`api_version >= 2025-10`). Previously these were unconditionally included regardless of api_version even though they are not supported.

Also fixes the isRemoteDomExtension type guard return type from `ExtensionInstance<{api_version: string}>['configuration']` (which lost all other config properties) to `ExtensionInstance['configuration'] & {api_version: string}` (which preserves them while narrowing api_version).

### Changes:

ui_extension.ts:
- In the UIExtensionSchema transform, compute whether the config qualifies as Remote DOM and only spread tools/instructions into build_manifest assets when true
- tools and instructions on the extension point object are set to undefined for non-Remote-DOM versions
- Fixed isRemoteDomExtension type guard return type

ui_extension.test.ts:
- Updated existing tools/instructions build_manifest tests to use api_version: '2025-10'
- Added 2 new negative tests confirming tools/instructions are excluded for non-Remote-DOM api versions
- Updated setupToolsExtension helper to default to '2025-10' with an apiVersion option
- Renamed test descriptions to clarify Remote DOM version dependency

### Testing
All unit tests should pass

### DIFF
--- a/.changeset/smart-teams-win.md
+++ b/.changeset/smart-teams-win.md
@@ -1,0 +1,5 @@
+---
+'@shopify/app': patch
+---
+
+Exclude unsupported assets from the build manifest for Remote UI extensions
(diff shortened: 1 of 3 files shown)


vividviolet:
We will no longer need this

PR #6957, opened 2026-03-09
Generate types for shared folders

justinhenricks:

  UI extensions can import files from shared folders outside their own extension directory (e.g. a top-level `shared/`
  folder in the app workspace). Previously, type generation was scoped per-extension - so shared files either didn't
  get types at all, or got incomplete ones because each extension only knew about its own targets.

  Changes the internal data structure for type generation from storing pre-rendered strings (`Map<string,
  Set<string>>`) to storing structured definitions (`Map<string, Map<string, SharedTypeDefinition>>`). Rendering is
  deferred to a new `renderTypeDefinitions` step that runs after all extensions have contributed their types.

  This means when two extensions both import `shared/utils.ts` with different targets, the targets get merged into a
  single union type declaration instead of one overwriting the other.

  Other changes:
  - `findNearestTsConfigDir` now walks up to `appDirectory` instead of `extensionDirectory`, so it can find tsconfigs for shared folders
  - Extracted `addTypeDefinition` (handles target merging + API version tracking), `assertTargetsResolvable`, and
  `renderTypeDefinitions` as shared helpers
  - Cleaned up exports that are no longer needed outside the module

  ### How to test your changes?
- Clone https://github.com/shopify-playground/robin-cli-shared-types
- This app has two shared folders that extensions import from. One inside extension/ and one outside. One of them is imported by both extneions, the other one only by one. I did this to test a more complex case
- Delete any shopify.d.ts files that might exist `rm **/shopify.d.ts`
- While in the cli folder
- run `pnpm shopify app build --path=/path/to/your-app` (you might need to run this with --reset once, I'm not entirely sure. Make sure you have this setup as an app for you so you can run build)
- Go your app
- There should now be a `shopify.d.ts` in the root with this content
```ts
import '@shopify/ui-extensions';

//@ts-ignore
declare module './extensions/shared/math.js' {
  const shopify: import('@shopify/ui-extensions/purchase.checkout.block.render').Api;
  const globalThis: { shopify: typeof shopify };
}

//@ts-ignore
declare module './shared/math.js' {
  const shopify:
    | import('@shopify/ui-extensions/customer-account.order-status.block.render').Api
    | import('@shopify/ui-extensions/purchase.checkout.block.render').Api;
  const globalThis: { shopify: typeof shopify };
}

```

  ### Measuring impact

  ### Checklist

### DIFF
--- a/packages/app/src/cli/models/extensions/specifications/type-generation.ts
+++ b/packages/app/src/cli/models/extensions/specifications/type-generation.ts
@@ -166,7 +166,7 @@ interface CreateTypeDefinitionOptions {
   toolsTypeDefinition?: string
 }
 
-export interface SharedTypeDefinition {
+interface SharedTypeDefinition {
   apiVersion: string
   targets: Set<string>
   toolsTypeDefinition?: string
@@ -194,7 +194,7 @@ function buildShopifyType(targets: string[], toolsTypeDefinition?: string): stri
   return null
 }
 
-export function createTypeDefinition({
+function createTypeDefinition({
   fullPath,
   typeFilePath,
   targets,
--- a/packages/app/src/cli/models/extensions/specifications/type-generation.ts
+++ b/packages/app/src/cli/models/extensions/specifications/type-generation.ts
@@ -276,7 +276,7 @@ export function addTypeDefinition(
   typeDefinitionsByFile.set(typeFilePath, currentTypeDefinitions)
 }
 
-export async function renderTypeDefinitions(
+export async function getTypeDefinitions(
   typeDefinitionsByPath: Map<string, SharedTypeDefinition>,
   typeFilePath: string,
 ): Promise<Set<string>> {
--- a/packages/app/src/cli/models/extensions/specifications/type-generation.ts
+++ b/packages/app/src/cli/models/extensions/specifications/type-generation.ts
@@ -1,5 +1,6 @@
+import {formatContent} from '../../../utilities/file-formatter.js'
 import {fileExists, findPathUp, readFileSync} from '@shopify/cli-kit/node/fs'
-import {dirname, joinPath, relativizePath, resolvePath} from '@shopify/cli-kit/node/path'
+import {dirname, isSubpath, joinPath, relativizePath, resolvePath} from '@shopify/cli-kit/node/path'
 import {AbortError} from '@shopify/cli-kit/node/error'
 import ts from 'typescript'
 import {compile} from 'json-schema-to-typescript'
@@ -165,6 +166,14 @@ interface CreateTypeDefinitionOptions {
   toolsTypeDefinition?: string
 }
 
+export interface SharedTypeDefinition {
+  apiVersion: string
+  targets: Set<string>
+  toolsTypeDefinition?: string
+}
+
+export type TypeDefinitionsByFile = Map<string, Map<string, SharedTypeDefinition>>
+
 /**
  * Builds the shopify API type based on targets and optional tools type.
  * Returns null if no targets are provided.
@@ -193,19 +202,7 @@ export function createTypeDefinition({
   toolsTypeDefinition,
 }: CreateTypeDefinitionOptions): string | null {
   try {
-    // Validate that all targets can be resolved
-    for (const target of targets) {
-      try {
-        require.resolve(`@shopify/ui-extensions/${target}`, {paths: [fullPath, typeFilePath]})
-      } catch (_) {
-        const {year, month} = parseApiVersion(apiVersion) ?? {year: 2025, month: 10}
-        // Throw specific error for the target that failed, matching the original getSharedTypeDefinition behavior
-        throw new AbortError(
-          `Type reference for ${target} could not be found. You might be using the wrong @shopify/ui-extensions version.`,
-          `Fix the error by ensuring you have the correct version of @shopify/ui-extensions, for example ~${year}.${month}.0, in your dependencies.`,
-        )
-      }
-    }
+    assertTargetsResolvable({fullPath, typeFilePath, targets, apiVersion})
 
     const relativePath = relativizePath(fullPath, dirname(typeFilePath))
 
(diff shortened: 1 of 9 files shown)


packages/app/src/cli/models/extensions/specifications/type-generation.ts
```diff
@@ -236,24 +233,103 @@ export function createTypeDefinition({
...
+  typeDefinitionsByFile: TypeDefinitionsByFile,
+  {fullPath, typeFilePath, targets, apiVersion, toolsTypeDefinition}: CreateTypeDefinitionOptions,
+) {
+  const currentTypeDefinitions = typeDefinitionsByFile.get(typeFilePath) ?? new Map<string, SharedTypeDefinition>()
+  const existingTypeDefinition = currentTypeDefinitions.get(fullPath)
+
+  if (existingTypeDefinition) {
+    targets.forEach((target) => existingTypeDefinition.targets.add(target))
+    existingTypeDefinition.toolsTypeDefinition ??= toolsTypeDefinition
+    if (isApiVersionNewer(apiVersion, existingTypeDefinition.apiVersion)) {
```
vividviolet, line 265: What is this for? We should just fail early because it's not expected for extensions on different api versions to share files

packages/app/src/cli/models/extensions/specifications/type-generation.ts
```diff
@@ -236,24 +233,103 @@ export function createTypeDefinition({
...
+      apiVersion,
+      targets: new Set(targets),
+      toolsTypeDefinition,
+    })
+  }
+
+  typeDefinitionsByFile.set(typeFilePath, currentTypeDefinitions)
+}
+
+export async function renderTypeDefinitions(
```
vividviolet: nit: can we rename this to something else? we're not rendering but returning content right?

packages/app/src/cli/models/extensions/specifications/type-generation.ts
robin-drexler: @vividviolet what do you mean by "fail early" here btw? Actually throw or just silently do nothing?

packages/app/src/cli/models/extensions/specifications/type-generation.ts
robin-drexler: `getTypeDefinitions`?

packages/app/src/cli/models/extensions/specifications/type-generation.ts
robin-drexler: Wouldn't it be fine if the versions are different as long as the target exists? (Which is currently checked)

Otherwise, you'd force partners to always use the same version for all extensions, right? As it'd otherwise always abort(?)

robin-drexler:
* **#6957** <a href="https://app.graphite.com/github/pr/Shopify/cli/6957?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Shopify/cli/6957?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/app/src/cli/models/extensions/specifications/type-generation.ts
vividviolet: The types are generated per extension though. This code should only hit if they have shared imports. We shouldn't encourage sharing code between different major versions. When the types are not compatible they'd have to do a custom type guard to make it work. Even when the types are compatible it just seems risky to allow this. The toml's api version also sets the default GraphQL version which might not be compatible either.

robin-drexler: @vividviolet how do you suggest should the CLI behave if we encounter different versions?

robin-drexler: >This code should only hit if they have shared imports. We shouldn't encourage sharing code between different major versions. 

Are you saying, we should check if all importing extensions of a shared package use the same extension version and if not, not generate any shared types? 

Should  this log a warning?

Are you suggesting something else entirely? 😅

vividviolet: I'm suggesting we throw an error if `existingTypeDefinition.apiVersion !== apiVersion`

robin-drexler: Wouldn't that lead to a situation where you as a partner aren't able to use extensions in different versions *at all* if you used shared code? Unless you're aware and don't access `shopify.` apis I suppose(?)

Do we really want to go that far?

robin-drexler: @vividviolet I wanna avoid a situation where someone updates the cli and suddenly they can't build anymore because they have a shared package and extensions in different versions that import from it. Probably unlikely to be the case at the moment given that global `shopify` is kinda new, but that still seems weird, don't you think?

robin-drexler: @vividviolet  The more I think about this, the less I love breaking the build in that situation. It would even break if you use shared code and don't even rely on global APIs at all.

Unless I'm misunderstanding your suggestion.

vividviolet: Ok fair. I will settle for a warning then and clearly tell them we're using the newer apiVersion since there are multiple

robin-drexler: @vividviolet 

> we're using the newer apiVersion

are we really **using** a newer version, tho?

In the end, we create a shopify.d.ts file with something like: 

 ```typescript
   import '@shopify/ui-extensions';

   //@ts-ignore
   declare module './shared/utils.ts' {
     const shopify: (import('@shopify/ui-extensions/admin.orders-details.block.render').Api
       | import('@shopify/ui-extensions/admin.product-details.action.render').Api);
     const globalThis: { shopify: typeof shopify };
   }
 ```

What '@shopify/ui-extensions` resolves to is entirely up to where the closest tsconfig is and the node module resolution process, no?

So if shared-package has a `package.json` with `'@shopify/ui-extensions` in it (and a tsconfig), that version will be used. 
If not, it'll go up to the root and use whatever version is in there. 

Which versions the extensions declare for `'@shopify/ui-extensions` is almost irrelevant 😬

vividviolet: Wait, what is `isApiVersionNewer` checking then? Is just checking eligibility for type generation?

robin-drexler: I think that's only used for the error message generated here: https://github.com/Shopify/cli/blob/generate-for-shared-folders/packages/app/src/cli/models/extensions/specifications/type-generation.ts#L247-L250

@JoviDeCroock can you maybe chime in on this thread too? Just wanna make sure I'm getting this all right.

JoviDeCroock: Your understanding is correct, that does mean that depending on from what perspective you look at an extension it could be using the wrong types. i.e. an extension that uses version x could have property y but another extension using version z could not have that property. We should be careful with this as there is no way to easily disambiguate this in TypeScript hence the bail to just not generate, a warning might be well suited though

PR #6965, opened 2026-03-10
Create abstract client steps infrastracture to externalize the ap modules client configurations

alfonso-noriega:
### WHY are these changes introduced?

This introduces a new client-side build pipeline system to replace the existing buildConfig approach for extensions. The current system is limited and doesn't provide enough flexibility for complex build scenarios.

### WHAT is this pull request doing?

- Adds a new `ClientSteps` interface that defines build steps as configuration objects with execution handled separately
- Introduces `BuildContext` to pass data through the step pipeline and store results
- Implements step executors for `build_function`, `build_theme`, `bundle_theme`, and `bundle_ui` operations
- Adds an `include_assets` step type with support for pattern-based and config-key-based file inclusions
- Updates `ExtensionSpecification` to include optional `clientSteps` field alongside existing `buildConfig`
- Modifies `buildForBundle` method to handle app config extensions differently by using the extension UID as the output path
- Adds null coalescing operators to prevent undefined errors when `filePatterns` and `ignoredFilePatterns` are not defined
- Includes comprehensive test coverage for the new client steps system

### How to test your changes?

1. Create an extension that uses the new `clientSteps` configuration
2. Run the build process and verify that steps execute in sequence
3. Test the `include_assets` step with both pattern and configKey inclusion types
4. Verify that error handling works correctly with `continueOnError` flag
5. Ensure existing extensions using `buildConfig` continue to work unchanged

### Measuring impact

How do we know this change was effective? Please choose one:

- [x] n/a - this doesn't need measurement, e.g. a linting rule or a bug-fix

### Checklist

- [x] I've considered possible cross-platform impacts (Mac, Linux, Windows)
- [x] I've considered possible [documentation](https://shopify.dev) changes

PR #6941, opened 2026-03-06
Read client steps from remote config

alfonso-noriega:
### WHY are these changes introduced?

Extension specifications need to include client steps information to support build processes that require step-by-step client-side operations.

### WHAT is this pull request doing?

- Adds `clientSteps` field to the GraphQL query for fetching extension specifications
- Updates the `RemoteSpecification` interface to include a `clientSteps` property of type `ClientSteps`
- Implements parsing logic in `AppManagementClient` to convert the string-based `clientSteps` field from the API response into a structured `ClientSteps` array
- Imports the `ClientSteps` type from the build services module

### How to test your changes?

1. Run the application and trigger a fetch of extension specifications
2. Verify that the `clientSteps` field is properly retrieved from the GraphQL API
3. Confirm that the client steps are correctly parsed and available in the specification objects

### Measuring impact

How do we know this change was effective? Please choose one:

- [x] n/a - this doesn't need measurement, e.g. a linting rule or a bug-fix
- [ ] Existing analytics will cater for this addition
- [ ] PR includes analytics changes to measure impact

### Checklist

- [ ] I've considered possible cross-platform impacts (Mac, Linux, Windows)
- [ ] I've considered possible [documentation](https://shopify.dev) changes

PR #6867, opened 2026-02-19
Refactor extension build process to use modular build steps

alfonso-noriega:
# Wire extension specifications to use client steps

## Summary

This PR wires all extension specifications to use the `clientSteps` contract introduced in [#01-build-steps-infrastructure](#). The `build()` method in `ExtensionInstance` is refactored from a hard-coded mode switch to a generic step execution loop, and every spec now declares its own `clientSteps` array.

**Depends on:** `01-build-steps-infrastructure`

---

### `ExtensionInstance.build()` — mode switch replaced by step loop

Before:

```typescript
async build(options: ExtensionBuildOptions): Promise<void> {
  const mode = this.specification.buildConfig.mode
  switch (mode) {
    case 'theme':
      await buildThemeExtension(this, options)
      return bundleThemeExtension(this, options)
    case 'function':
      return buildFunctionExtension(this, options)
    case 'ui':
      return buildUIExtension(this, options)
    case 'copy_files':
      return copyFilesForExtension(this, options)
    // ...
  }
}
```

After:

```typescript
async build(options: ExtensionBuildOptions): Promise<void> {
  const {clientSteps} = this.specification

  const context: BuildContext = {
    extension: this,
    options,
    stepResults: new Map(),
  }

  const steps = clientSteps
    .filter((group) => group.lifecycle === 'deploy')
    .flatMap((group) => group.steps)

  for (const step of steps) {
    const result = await executeStep(step, context)
    context.stepResults.set(step.id, result)
  }
}
```

The mode switch and all its direct function call imports (`buildThemeExtension`, `buildUIExtension`, `buildFunctionExtension`, `copyFilesForExtension`) are removed. The step loop is uniform across all extension types.

### `extension.ts` — `buildThemeExtension` removed

`buildThemeExtension()` and its `runThemeCheck` import are removed from `extension.ts` — the logic now lives exclusively in `build-theme-step.ts`.

---

## Specs wired

All specs now declare `clientSteps`. The `buildConfig.mode` is kept as a transitional field.

### UI extensions

```typescript
// checkout_post_purchase, checkout_ui_extension, pos_ui_extension,
// product_subscription, web_pixel_extension, ui_extension
clientSteps: [{
  lifecycle: 'deploy',
  steps: [
    {id: 'bundle-ui',          name: 'Bundle UI Extension', type: 'bundle_ui',          config: {}},
    {id: 'copy-static-assets', name: 'Copy Static Assets',  type: 'copy_static_assets', config: {}},
  ],
}]
```

[...]

PR #6966, opened 2026-03-10
Add client steps implementation to support existing functionalities

alfonso-noriega:
# Add build step infrastructure for extension asset management

This introduces a flexible build step system to handle various asset copying and processing needs for different extension types, replacing hardcoded build logic with configurable steps.

- Adds \`'admin'\` to the \`CONFIG_EXTENSION_IDS\` array for admin extensions
- Creates new build step executors:
  - \`executeIncludeAssetsStep\` — handles asset inclusion via three entry types (see below), with sequential \`configKey\` execution to avoid filesystem races and parallel \`pattern\`/\`static\` execution
  - \`executeCopyStaticAssetsStep\` — copies static assets defined in extension build manifests
  - \`executeCreateTaxStubStep\` — generates minimal JavaScript stub files for tax calculation extensions
- Implements a centralized step router (\`executeStepByType\`) that dispatches to appropriate handlers based on step type
- Supports three inclusion strategies in \`include_assets\`:
  - **\`pattern\`** — glob-based file selection from a source directory (\`baseDir\` defaults to extension root); selected files are copied preserving their relative paths
  - **\`static\`** — explicit source path; files copy to their basename in the output, directories copy under their own name (or to an explicit \`destination\`)
  - **\`configKey\`** — resolves a path (or array of paths) from the extension configuration; directory contents are merged into the output root (or into \`destination\` when given); silently skipped when the key is absent

### How to test your changes?

1. Create an extension with various asset types (static files, pattern-matched files, config-driven paths)
2. Configure build steps using the new \`include_assets\` step type with different inclusion strategies
3. Run the build process and verify assets are copied correctly to output directories
4. Test with admin extensions to ensure they're recognized as config extensions
5. Run the test suite: \`vitest run "include-assets"\`

### Measuring impact

### Checklist

### DIFF
--- a/packages/app/src/cli/services/build/steps/include-assets-step.test.ts
+++ b/packages/app/src/cli/services/build/steps/include-assets-step.test.ts
@@ -535,7 +535,6 @@ describe('executeIncludeAssetsStep', () => {
 
       // Then
       expect(result.filesCopied).toBe(0)
-      expect(mockStdout.write).toHaveBeenCalledWith(expect.stringContaining('no files matched'))
     })
   })
 
--- a/packages/app/src/cli/services/build/steps/include-assets/copy-by-pattern.ts
+++ b/packages/app/src/cli/services/build/steps/include-assets/copy-by-pattern.ts
@@ -21,7 +21,6 @@ export async function copyByPattern(
   })
 
   if (files.length === 0) {
-    options.stdout.write(`Warning: no files matched, skipping\n`)
     return 0
   }
 
--- a/packages/app/src/cli/services/build/steps/include-assets/copy-by-pattern.test.ts
+++ b/packages/app/src/cli/services/build/steps/include-assets/copy-by-pattern.test.ts
@@ -54,7 +54,6 @@ describe('copyByPattern', () => {
     expect(result).toBe(0)
     expect(fs.mkdir).not.toHaveBeenCalled()
     expect(fs.copyFile).not.toHaveBeenCalled()
-    expect(mockStdout.write).toHaveBeenCalledWith(expect.stringContaining('no files matched'))
   })
 
   test('skips file and warns when resolved destination escapes the output directory', async () => {


alfonso-noriega:
* **#6966** <a href="https://app.graphite.com/github/pr/Shopify/cli/6966?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#6867](https://github.com/Shopify/cli/pull/6867) <a href="https://app.graphite.com/github/pr/Shopify/cli/6867?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#6941](https://github.com/Shopify/cli/pull/6941) <a href="https://app.graphite.com/github/pr/Shopify/cli/6941?utm_source=stack-comment-icon" target="_blank">Graphite</a>) 👈 <a href="https://app.graphite.com/github/pr/Shopify/cli/6966?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/app/src/cli/services/build/steps/copy-static-assets-step.ts
```diff
@@ -0,0 +1,11 @@
+import type {LifecycleStep, BuildContext} from '../client-steps.js'
+
+/**
+ * Executes a copy_static_assets build step.
+ *
+ * Copies static assets defined in the extension's build_manifest to the output directory.
+ * This is a no-op for extensions that do not define static assets.
+ */
+export async function executeCopyStaticAssetsStep(_step: LifecycleStep, context: BuildContext): Promise<void> {
```
ryancbahan, line 9: are we ever using `_step`? why is it included?

packages/app/src/cli/models/extensions/extension-instance.ts
```diff
@@ -44,6 +44,7 @@ export const CONFIG_EXTENSION_IDS: string[] = [
   WebhookSubscriptionSpecIdentifier,
   WebhooksSpecIdentifier,
   EventsSpecIdentifier,
+  'admin',
```
isaacroldan, line 51: Better to use a constant from the spec file instead of hardcoding here

alfonso-noriega: there is no spec. file yet here, it is updated later in the stack

packages/app/src/cli/services/build/steps/copy-static-assets-step.ts
```diff
@@ -0,0 +1,12 @@
+import type {LifecycleStep, BuildContext} from '../client-steps.js'
+
+/**
+ * Executes a copy_static_assets build step.
+ *
+ * Copies static assets defined in the extension's build_manifest to the output directory.
+ * This is a no-op for extensions that do not define static assets.
+ */
+// _step is intentionally unused — required to match the LifecycleStep executor signature
```
isaacroldan: 🎨 **Code Style**: The `_` prefix is a standard TypeScript convention for unused parameters that already communicates intent. None of the other step files (`build-function-step.ts`, `bundle-ui-step.ts`, `build-theme-step.ts`, `create-tax-stub-step.ts`) include this comment despite using the same `_step` pattern. The comment adds noise and creates an inconsistency across the steps directory.

**Suggestion:** Remove the comment on line 9 — the `_step` convention is self-documenting and consistent with all other step files.

packages/app/src/cli/services/build/steps/include-assets-step.test.ts
isaacroldan, line 1: 💡 **Improvement**: The `isDirectory` check added in `copySourceEntry` (lines 185-191) — where a source is a file with no destination — is completely untested. The existing test at line 144 mocks `isDirectory` returning `false` for a second entry, but that entry has a `destination` and takes the early return at line 177, never reaching the `isDirectory` branch. A test exercising this path would validate the new security-hardening code.

**Suggestion:** Add a test case with a static entry that has a source file (not directory), no destination, and verify it uses `copyFile` to copy the single file. Mock `isDirectory` to return `false` and `fileExists` to return `true`.

isaacroldan, line 1: 💡 **Improvement**: The test suite covers file+destination (lines 103-125) and directory-without-destination (lines 36-59), but no test exercises the case where `source` is a directory AND `destination` is provided. This untested code path (lines 177-183) calls `copyFile` which recursively copies via fs-extra but returns a hardcoded count of `1`. A test here would surface the count bug and verify the directory-copy-to-destination behavior is intentional.

**Suggestion:** Add a test case like `{type: 'static', source: 'dist', destination: 'assets/dist'}` where `dist` is a directory, and verify both the copy behavior and the accuracy of the returned `filesCopied` count.

packages/app/src/cli/services/build/steps/include-assets-step.ts
isaacroldan, line 1: 🎨 **Code Style**: This file is named `include_assets_step.ts` (underscores) while every other step file in the directory uses kebab-case: `build-function-step.ts`, `build-theme-step.ts`, `bundle-theme-step.ts`, `bundle-ui-step.ts`, `copy-static-assets-step.ts`, `create-tax-stub-step.ts`. The test file itself already uses kebab-case: `include-assets-step.test.ts`.

**Suggestion:** Rename to `include-assets-step.ts` and update the import in `index.ts` (line 1: `'./include_assets_step.js'`) accordingly.

packages/app/src/cli/services/build/steps/include-assets-step.ts
```diff
@@ -0,0 +1,378 @@
...
+  destination: z.string().optional(),
+  preserveStructure: z.boolean().default(false),
+})
+
+const InclusionEntrySchema = z.discriminatedUnion('type', [PatternEntrySchema, StaticEntrySchema, ConfigKeyEntrySchema])
+
+/**
+ * Configuration schema for include_assets step.
+ *
+ * `inclusions` is a flat array of entries, each with a `type` discriminant
```
isaacroldan, line 51: 💡 **Improvement**: The JSDoc for `IncludeAssetsConfigSchema` (lines 58-59) says the type discriminants are `'files'` and `'pattern'`, but the actual discriminated union defines `'static'`, `'configKey'`, and `'pattern'`. The same stale reference appears in the `executeIncludeAssetsStep` JSDoc at lines 95-97 which references `type: 'files'`. This is a holdover from an earlier design iteration and will mislead future maintainers.

**Suggestion:** Update both JSDoc blocks. For lines 58-59:

```suggestion
 * `inclusions` is a flat array of entries, each with a `type` discriminant
 * (`'static'`, `'configKey'`, or `'pattern'`). All entries are processed in parallel.
```

Also update lines 95-97 to reference `'static'`, `'configKey'`, and `'pattern'` instead of `'files'`.

packages/app/src/cli/services/build/steps/include_assets_step.ts
```diff
@@ -0,0 +1,378 @@
...
+})
+
+/**
+ * Removes any '..' traversal segments from a relative destination path and
+ * emits a warning if any were found. Preserves normal '..' that only navigate
+ * within the path (e.g. 'foo/../bar' → 'bar') but never allows the result to
+ * escape the output root.
+ */
+function sanitizeDestination(input: string, warn: (msg: string) => void): string {
+  const segments = input.split('/')
```
isaacroldan: 🔒 **Security**: `sanitizeDestination` splits on `/` only (line 72). On Windows, `\` is also a valid path separator. A destination like `..\\..\\etc` wouldn't be split into segments, so the `..` detection would miss it. The subsequent `joinPath` (wrapping Node's `path.join`) WOULD resolve backslash separators on Windows, allowing the path to escape the output root. The defense-in-depth check at line 306 in `copyByPattern` would catch this for pattern entries, but static and configKey entries pass through `sanitizeDestination` without a secondary bounds check.

**Suggestion:** Normalize path separators before splitting:

```suggestion
  const segments = input.replace(/\\/g, '/').split('/')
```

packages/app/src/cli/services/build/steps/include-assets-step.ts
```diff
@@ -0,0 +1,378 @@
...
+  if (!exists) {
+    throw new Error(`Source does not exist: ${sourcePath}`)
+  }
+
+  if (destination !== undefined) {
+    const destPath = joinPath(outputDir, destination)
+    await mkdir(dirname(destPath))
+    await copyFile(sourcePath, destPath)
+    options.stdout.write(`Copied ${source} to ${destination}\n`)
+    return 1
```
isaacroldan: Human: I guess the AI is right, but not sure if the fix is necessary? how important is this count?

🐛 **Bug**: When a `static` entry has an explicit `destination` (lines 177-183), `copyFile` is called and the count is hardcoded to `1`. Since `copyFile` wraps fs-extra's `copy` which handles directories recursively, this works functionally — but if the source is a directory containing many files, the reported count is still 1. This is inconsistent with the no-destination path (lines 193-200) which correctly counts files via `glob`.

**Suggestion:** Add an `isDirectory` branch in the destination-provided path: if the source is a directory, count the actual files after copy (via glob on the destination), otherwise return 1. This mirrors the logic in the no-destination path below.

packages/app/src/cli/services/build/steps/include-assets-step.ts
```diff
@@ -0,0 +1,378 @@
...
+  const counts = await Promise.all(
+    paths.map(async (sourcePath) => {
+      const fullPath = joinPath(baseDir, sourcePath)
+      const exists = await fileExists(fullPath)
+      if (!exists) {
+        options.stdout.write(`Warning: path '${sourcePath}' does not exist, skipping\n`)
+        return 0
+      }
+      const destDir = preserveStructure ? joinPath(effectiveOutputDir, basename(fullPath)) : effectiveOutputDir
+      await copyDirectoryContents(fullPath, destDir)
```
isaacroldan: 🐛 **Bug**: `copyConfigKeyEntry` unconditionally calls `copyDirectoryContents(fullPath, destDir)` at line 246 without checking whether the resolved path is actually a directory. If a config key resolves to a file (e.g., `tools: 'tools-a.js'`), `copyDirectoryContents` internally globs `srcDir/**/*` which matches nothing for a file — the copy silently becomes a no-op. The `copySourceEntry` function in this same PR correctly handles this with an `isDirectory` check (line 185), but the same fix wasn't applied here. Tests pass only because `copyDirectoryContents` is mocked.

**Suggestion:** Add the same `isDirectory` check that exists in `copySourceEntry`. When the resolved path is a file, use `copyFile` instead:

```suggestion
      if (!(await isDirectory(fullPath))) {
        const destPath = joinPath(effectiveOutputDir, basename(fullPath))
        await mkdir(effectiveOutputDir)
        await copyFile(fullPath, destPath)
        options.stdout.write(`Copied '${sourcePath}' to ${basename(fullPath)}\n`)
        return 1
      }
      const destDir = preserveStructure ? joinPath(effectiveOutputDir, basename(fullPath)) : effectiveOutputDir
```

packages/app/src/cli/services/build/steps/include-assets-step.ts
```diff
@@ -0,0 +1,378 @@
...
+      if (seen.has(name)) {
+        duplicates.add(name)
+      } else {
+        seen.add(name)
+      }
+    }
+    if (duplicates.size > 0) {
+      const colliding = files.filter((fp) => duplicates.has(basename(fp)))
+      options.stdout.write(
+        `Warning: filename collision detected when flattening — the following files share a basename and will overwrite each other: ${colliding.join(', ')}\n`,
```
isaacroldan: 🐛 **Bug**: When `preserveStructure` is false and duplicate basenames are detected (lines 282-299), a warning is emitted but `Promise.all` at line 301 proceeds to copy all files concurrently. Since colliding files target the same destination path, the final file content depends on which `copyFile` promise resolves last — this is a race condition. The warning correctly identifies the problem but the behavior should be deterministic.

**Suggestion:** Consider either: (1) processing files sequentially when collisions are detected so the last entry in the array deterministically wins, or (2) deduplicating the file list before copying so only one file per basename is copied (e.g., last-in-array wins), adjusting the count accordingly.

packages/app/src/cli/services/build/steps/include_assets_step.ts
```diff
@@ -0,0 +1,378 @@
...
+
+      if (filepath === destPath) return
+
+      await mkdir(dirname(destPath))
+      await copyFile(filepath, destPath)
+    }),
+  )
+
+  options.stdout.write(`Copied ${files.length} file(s) from ${sourceDir} to ${outputDir}\n`)
+  return {filesCopied: files.length}
```
isaacroldan: 🐛 **Bug**: `copyByPattern` returns `files.length` as the copy count, but files can be skipped in two places: the path-traversal guard (lines 306-308, early `return`) and the same-path check (line 311, early `return`). The returned count overstates actual copies, and this value propagates to `stepResults` via `BuildContext`, affecting downstream logging or steps that consume it.

**Suggestion:** Track actual copies by having each `map` callback return 1 or 0, then sum:

```suggestion
  const copyResults = await Promise.all(
    files.map(async (filepath) => {
      const relPath = preserveStructure ? relativePath(sourceDir, filepath) : basename(filepath)
      const destPath = joinPath(outputDir, relPath)

      if (relativePath(outputDir, destPath).startsWith('..')) {
        options.stdout.write(`Warning: skipping '${filepath}' - resolved destination is outside the output directory\n`)
        return 0
      }

      if (filepath === destPath) return 0

      await mkdir(dirname(destPath))
      await copyFile(filepath, destPath)
      return 1
    }),
  )

  const copiedCount = copyResults.reduce((sum, n) => sum + n, 0)
  options.stdout.write(`Copied ${copiedCount} file(s) from ${sourceDir} to ${outputDir}\n`)
  return {filesCopied: copiedCount}
```

packages/app/src/cli/services/build/steps/include-assets-step.ts
```diff
@@ -0,0 +1,398 @@
...
+ * - Without `destination`, `preserveStructure` false (default): merges
+ *   directory contents into the output root.
+ * - Without `destination`, `preserveStructure` true: places the directory
+ *   under its own name in the output.
+ */
+const StaticEntrySchema = z.object({
+  type: z.literal('static'),
+  source: z.string(),
+  destination: z.string().optional(),
+  preserveStructure: z.boolean().default(false),
```
isaacroldan: Why is this not defaulting to true?

packages/app/src/cli/services/build/steps/include-assets-step.ts
```diff
@@ -0,0 +1,398 @@
...
+  inclusions: z.array(InclusionEntrySchema),
+})
+
+/**
+ * Removes any '..' traversal segments from a relative destination path and
+ * emits a warning if any were found. Preserves normal '..' that only navigate
+ * within the path (e.g. 'foo/../bar' → 'bar') but never allows the result to
+ * escape the output root.
+ */
+function sanitizeDestination(input: string, warn: (msg: string) => void): string {
```
isaacroldan: I feel like this could be in `path.ts` as a shared tool for anyone that wants to sanitize a path?

packages/app/src/cli/services/build/steps/include-assets-step.ts
```diff
@@ -0,0 +1,398 @@
...
+ */
+export async function executeIncludeAssetsStep(
+  step: LifecycleStep,
+  context: BuildContext,
+): Promise<{filesCopied: number}> {
+  const config = IncludeAssetsConfigSchema.parse(step.config)
+  const {extension, options} = context
+  // When outputPath is a file (e.g. index.js, index.wasm), the output directory is its
+  // parent. When outputPath has no extension, it IS the output directory.
+  const outputDir = extname(extension.outputPath) ? dirname(extension.outputPath) : extension.outputPath
```
isaacroldan, line 77: We are assuming that having an extension === is a file, which in 99% of the cases is true... but is not a univeral truth :upside_down_face: I guess it's ok

packages/app/src/cli/services/build/steps/include-assets-step.ts
```diff
@@ -0,0 +1,398 @@
...
+
+  const counts = await Promise.all(
+    config.inclusions.map(async (entry) => {
+      const warn = (msg: string) => options.stdout.write(msg)
+      const sanitizedDest = entry.destination !== undefined ? sanitizeDestination(entry.destination, warn) : undefined
+
+      if (entry.type === 'pattern') {
+        const sourceDir = entry.baseDir ? joinPath(extension.directory, entry.baseDir) : extension.directory
+        const destinationDir = sanitizedDest ? joinPath(outputDir, sanitizedDest) : outputDir
```
isaacroldan, line 86: for dev/deploy are we relaying on those flow to update the `extension.outputPath` ?  
Or how do we tell the `steps-runner` where the build/deploy/dev output should be?

packages/app/src/cli/services/build/steps/include-assets-step.ts
```diff
@@ -0,0 +1,398 @@
...
+
+      return copySourceEntry(
+        entry.source,
+        sanitizedDest,
+        extension.directory,
+        outputDir,
+        options,
+        entry.preserveStructure,
+      )
+    }),
```
isaacroldan: all these should receive a config object as input instead of positional arguments. The way it is now there is no validation that you haven't passed the values in the wront order (for instance: mixing the include/ignore patterns)

packages/app/src/cli/services/build/steps/include-assets-step.ts
```diff
@@ -0,0 +1,398 @@
...
+          extension.directory,
+          outputDir,
+          context,
+          options,
+          entry.preserveStructure,
+          sanitizedDest,
+        )
+      }
+
+      return copySourceEntry(
```
isaacroldan: is this case for `static`? let's add it explicitly or maybe a switch with both `static` and `default` cases

packages/app/src/cli/services/build/steps/include-assets-step.ts
```diff
@@ -0,0 +1,398 @@
...
+  destination: string | undefined,
+  baseDir: string,
+  outputDir: string,
+  options: {stdout: NodeJS.WritableStream},
+  preserveStructure: boolean,
+): Promise<number> {
+  const sourcePath = joinPath(baseDir, source)
+  const exists = await fileExists(sourcePath)
+  if (!exists) {
+    throw new Error(`Source does not exist: ${sourcePath}`)
```
isaacroldan: is it correct to throw an error here? can a file be included "optionally"? :thinking_face:

packages/app/src/cli/services/build/steps/include-assets-step.ts
```diff
@@ -0,0 +1,398 @@
...
+    return 1
+  }
+
+  if (!(await isDirectory(sourcePath))) {
+    const destPath = joinPath(outputDir, basename(sourcePath))
+    await mkdir(outputDir)
+    await copyFile(sourcePath, destPath)
+    options.stdout.write(`Copied ${source} to ${basename(sourcePath)}\n`)
+    return 1
+  }
```
isaacroldan: not sure I understand the order there.  
First you check `destination !== undefined`, so if there is a destination, the first `if` is executed.  
If there is no destination AND is not a directory, then the second `if` is executed.   
And if there is no destination AND is a directory, the rest which handles the `preserveStructure`  
  
Correct?  
But it could enter in the first IF being a directory too no? so the `preserveStructure`is ignored in that case?

packages/app/src/cli/services/build/steps/include-assets-step.ts
```diff
@@ -0,0 +1,398 @@
...
+ * Pattern strategy: glob-based file selection.
+ */
+async function copyByPattern(
+  sourceDir: string,
+  outputDir: string,
+  patterns: string[],
+  ignore: string[],
+  preserveStructure: boolean,
+  options: {stdout: NodeJS.WritableStream},
+): Promise<{filesCopied: number}> {
```
isaacroldan: why this one returns an object and the others just an integer?

packages/app/src/cli/services/build/steps/include-assets-step.ts
```diff
@@ -0,0 +1,398 @@
...
+    }
+    if (duplicates.size > 0) {
+      const colliding = files.filter((fp) => duplicates.has(basename(fp)))
+      options.stdout.write(
+        `Warning: filename collision detected when flattening — the following files share a basename and will overwrite each other: ${colliding.join(', ')}\n`,
+      )
+    }
+  }
+
+  // When flattening and collisions exist, deduplicate so last-in-array deterministically wins
```
isaacroldan: why not add an index to the file name at the destination? isn't the manifest in part for this reason?

packages/app/src/cli/services/build/steps/include-assets-step.ts
```diff
@@ -0,0 +1,421 @@
...
+  }
+
+  if (destination !== undefined) {
+    const destPath = joinPath(outputDir, destination)
+    if (await isDirectory(sourcePath)) {
+      await copyDirectoryContents(sourcePath, destPath)
+      const copied = await glob(['**/*'], {cwd: destPath, absolute: false})
+      options.stdout.write(`Copied ${source} to ${destination}\n`)
+      return copied.length
+    }
```
isaacroldan: You injected this here, but I think there is a lot of duplication that can be simplified within the `copySourceEntry` code

packages/app/src/cli/services/build/steps/include-assets-step.ts
alfonso-noriega: The race condition is already eliminated: lines 332–340 deduplicate `filesToCopy` before `Promise.all`, so at most one copy of each colliding basename is ever scheduled. No two promises write to the same destination path concurrently.

packages/app/src/cli/services/build/steps/include-assets-step.ts
alfonso-noriega: Yes — `static` entries have hardcoded source paths baked into the extension config, so a missing path is always a configuration bug that should surface immediately. `configKey` entries are different: their values come from user TOML config at runtime, so silently skipping a missing key is intentional. The asymmetry is deliberate.

packages/app/src/cli/services/build/steps/include-assets-step.ts
alfonso-noriega: Why do you think so? we are expecting that an extension with a include assets step defined is going to have an output path, this may be an file or a folder as we keep updating that value  in the live of the extension instance.

alfonso-noriega: Good catch — `outputPath` is typed as `string` (never `undefined`), so the real question is about the `extname()` heuristic: it assumes `has extension → is a file`. This holds for all current extension types (UI extensions → `.js`, functions → `.wasm`, themes → directories). The edge case would be an extensionless output file (e.g. a Wasm binary named `index` with no `.wasm` suffix), where `extname()` returns `""` and the code would treat the path itself as the output directory — incorrect, but not a scenario any current extension type produces. Happy to document this assumption in a comment if you'd like, but leaving as-is for now per your 🙃.

packages/app/src/cli/services/build/steps/include-assets-step.ts
alfonso-noriega: The client steps are infrastructure being introduced here — not yet wired into the dev/deploy flows. When they are, the runner will call `executeStep` after `buildForBundle()` has already mutated `extension.outputPath` to point at the bundle directory. Since `BuildContext` holds a reference to the same `ExtensionInstance`, any step that runs after the build will automatically see the correct, up-to-date `outputPath`.

packages/app/src/cli/services/build/steps/include-assets-step.ts
alfonso-noriega: The original `copyFilesForExtension` (and `bundleThemeExtension`) always preserve structure with no flatten option at all — they do `relativePath(extension.directory, filepath)` and never touch the basename. There is no collision surface in the original implementation.

Our `pattern` entry replicates that exactly when `preserveStructure: true` (the default). The collision case only arises when the consumer explicitly opts into `preserveStructure: false` to flatten. In that case, silently renaming files to `foo_1.js` etc. would break any consumer code that references `foo.js` by name — and the manifest would need to map keys to renamed paths it never had a concept for. Warning + deterministic last-wins is safer: the user gets a clear signal to fix their config (or switch back to the default `preserveStructure: true` which eliminates the problem entirely).

isaacroldan: There is no use case for `preserveStructure: false` in the pattern case right? is there a use case for `preserveStructure: false` in any of the copy files options? if not, let's assume true for everything and simplify the code, no need to add unnecessary complexity now

packages/app/src/cli/services/build/steps/include-assets/copy-config-key-entry.test.ts
```diff
@@ -0,0 +1,184 @@
...
+    mockStdout = {write: vi.fn()}
+  })
+
+  test('merges directory contents into output root', async () => {
+    // Given
+    const context = makeContext({static_root: 'public'})
+    vi.mocked(fs.fileExists).mockResolvedValue(true)
+    vi.mocked(fs.isDirectory).mockResolvedValue(true)
+    vi.mocked(fs.copyDirectoryContents).mockResolvedValue()
+    vi.mocked(fs.glob).mockResolvedValue(['index.html', 'logo.png'])
```
isaacroldan: i think these tests would be more useful if we don't mock all this stuff.  
We can use `inTemporaryDirectory` , create a `src` folder with some files and an empty `dist` folder. Then validate that the correct files are included in `dist`

packages/app/src/cli/services/build/steps/include-assets/copy-source-entry.ts
```diff
@@ -0,0 +1,49 @@
...
+
+  // Resolve destination path and log message up front, then dispatch on file vs directory.
+  let destPath: string
+  let logMsg: string
+  if (destination !== undefined) {
+    destPath = joinPath(outputDir, destination)
+    logMsg = `Copied ${source} to ${destination}\n`
+  } else {
+    destPath = joinPath(outputDir, basename(sourcePath))
+    logMsg = `Copied ${source} to ${basename(sourcePath)}\n`
```
isaacroldan: I don't think we should show the destination to the user, that's a hidden location.   
Just `Included <file>`could be enough imho

packages/app/src/cli/services/build/steps/include-assets/copy-config-key-entry.ts
```diff
@@ -0,0 +1,121 @@
...
+      if (!(await isDirectory(fullPath))) {
+        const destPath = joinPath(effectiveOutputDir, basename(fullPath))
+        await mkdir(effectiveOutputDir)
+        await copyFile(fullPath, destPath)
+        options.stdout.write(`Copied '${sourcePath}' to ${basename(fullPath)}\n`)
+        return 1
+      }
+      await copyDirectoryContents(fullPath, effectiveOutputDir)
+      const copied = await glob(['**/*'], {cwd: effectiveOutputDir, absolute: false})
+      options.stdout.write(`Copied contents of '${sourcePath}' to output root\n`)
```
isaacroldan: The user doesn't know what `output root`means. We need to be more clear.   
`Included <sourcePath>` or `Included <sourcePath> in bundle`  
  
I wonder if we should say anything at all :thinking_face:  
It might be too verbose for `app dev`

packages/app/src/cli/services/build/steps/include-assets/copy-by-pattern.ts
```diff
@@ -0,0 +1,53 @@
...
+      if (filepath === destPath) return 0
+
+      await mkdir(dirname(destPath))
+      await copyFile(filepath, destPath)
+      return 1
+    }),
+  )
+
+  const copiedCount = copyResults.reduce((sum, count) => sum + count, 0)
+  options.stdout.write(`Copied ${copiedCount} file(s) from ${sourceDir} to ${outputDir}\n`)
```
isaacroldan: same here, no need to show the outputDir

packages/app/src/cli/services/build/steps/include-assets/copy-by-pattern.ts
```diff
@@ -0,0 +1,53 @@
...
+): Promise<number> {
+  const {sourceDir, outputDir, patterns, ignore} = config
+  const files = await glob(patterns, {
+    absolute: true,
+    cwd: sourceDir,
+    ignore,
+  })
+
+  if (files.length === 0) {
+    options.stdout.write(`Warning: No files matched patterns in ${sourceDir}\n`)
```
isaacroldan: not sure this is useful to the user. What patterns? the user won't know anything about that

packages/app/src/cli/services/build/steps/include-assets/copy-config-key-entry.test.ts
alfonso-noriega: Done — rewrote all tests in `copy-config-key-entry.test.ts` using `inTemporaryDirectory` with real file operations. Tests now assert on actual copied file existence and content rather than spy call arguments, giving real I/O coverage.

packages/app/src/cli/services/build/steps/include-assets/copy-source-entry.ts
alfonso-noriega: Done — simplified to `Included ${source}\n` across all copy paths in `copy-source-entry.ts`, dropping destination from the message.

packages/app/src/cli/services/build/steps/include-assets/copy-config-key-entry.ts
alfonso-noriega: Done — changed to `Included '${sourcePath}'\n` for both the file and directory copy cases in `copy-config-key-entry.ts`.

packages/app/src/cli/services/build/steps/include-assets/copy-by-pattern.ts
alfonso-noriega: Done — changed to `Included ${filesCopied} file(s)\n` in `copy-by-pattern.ts`, removing both `sourceDir` and `outputDir` from the message.

packages/app/src/cli/services/build/steps/include-assets/copy-by-pattern.ts
alfonso-noriega: Done — softened to `Warning: no files matched, skipping\n`, removing the `sourceDir` reference the user wouldn't recognise.

isaacroldan: I wouldn't even show this. Why is it a warning? maybe the files are optional?

alfonso-noriega: Agreed — removed the warning entirely. Empty matches are not an error condition; the patterns may be intentionally optional.

PR #6967, opened 2026-03-10
Clean up module outputFile calculation and move it to the specifications

alfonso-noriega:

The `outputFileName` property was previously calculated dynamically using a getter method that relied on the extension's build configuration mode. This approach was inflexible and didn't allow for extension-specific customization of output file names.

- Converts `outputFileName` from a computed getter to a direct property on `ExtensionInstance`
- Adds an optional `getOutputFileName` method to the `ExtensionSpecification` interface
- Implements `getOutputFileName` in all UI extension specifications to return `${extension.handle}.js`
- Implements `getOutputFileName` in function extensions to return `index.wasm`
- Refactors the `BuildConfig` type to use a discriminated union for better type safety
- Initializes `outputFileName` in the constructor using the specification's method or defaults to empty string

### How to test your changes?

1. Create extensions of different types (UI extensions, functions, etc.)
2. Verify that the output file names are generated correctly for each extension type
3. Test the build process to ensure extensions are built with the expected output file names
4. Confirm that existing functionality remains unchanged

### Measuring impact

### Checklist

### DIFF
--- a/packages/app/src/cli/models/extensions/extension-instance.ts
+++ b/packages/app/src/cli/models/extensions/extension-instance.ts
@@ -27,7 +27,7 @@ import {ok} from '@shopify/cli-kit/node/result'
 import {constantize, slugify} from '@shopify/cli-kit/common/string'
 import {hashString, nonRandomUUID} from '@shopify/cli-kit/node/crypto'
 import {partnersFqdn} from '@shopify/cli-kit/node/context/fqdn'
-import {joinPath, basename, normalizePath, resolvePath} from '@shopify/cli-kit/node/path'
+import {joinPath, normalizePath, resolvePath, relativePath, basename} from '@shopify/cli-kit/node/path'
 import {fileExists, touchFile, moveFile, writeFile, glob, copyFile, globSync} from '@shopify/cli-kit/node/fs'
 import {getPathValue} from '@shopify/cli-kit/common/object'
 import {outputDebug} from '@shopify/cli-kit/node/output'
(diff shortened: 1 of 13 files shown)


alfonso-noriega:
* **#6967** <a href="https://app.graphite.com/github/pr/Shopify/cli/6967?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Shopify/cli/6967?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/app/src/cli/models/extensions/extension-instance.ts
```diff
@@ -137,14 +137,7 @@ export class ExtensionInstance<TConfiguration extends BaseConfigType = BaseConfi
...
   get outputFileName() {
-    const mode = this.specification.buildConfig.mode
-    if (mode === 'copy_files' || mode === 'theme') {
-      return ''
-    } else if (mode === 'function') {
-      return 'index.wasm'
-    } else {
-      return `${this.handle}.js`
-    }
+    return basename(this.specification.getOutputFileName?.(this) ?? '')
```
ryancbahan: Do _all_ extension instances always have a build/output file name? `extension-instance` feels very overloaded today, and I worry we're overloading it further with things that might be optional properties.

isaacroldan: I think the change in this PR actually cleans this. Move the optional implementation to the specification, so each module can decide if it has an outputFileName or not.   
  
This function already existed in `extension-instance`, but isntead of hardcoded if/else we move the ownership to each specification

packages/app/src/cli/models/extensions/extension-instance.ts
```diff
@@ -162,19 +155,13 @@ export class ExtensionInstance<TConfiguration extends BaseConfigType = BaseConfi
     this.handle = this.buildHandle()
     this.localIdentifier = this.handle
     this.idEnvironmentVariableName = `SHOPIFY_${constantize(this.localIdentifier)}_ID`
-    this.outputPath = this.directory
+    this.outputPath = joinPath(
+      this.directory,
+      this.specification.getOutputRelativePath?.(this) ?? '',
+      this.outputFileName,
```
isaacroldan: If we have a helper getter for `outputFileName`, we can have one for `outputRelativeFolder` that defaults to `''`

isaacroldan: In fact, getOutputRelativePath doesn't need to be optional no? it could have a default implementation returning `''` in the specifiaction factory method.

PR #6972, opened 2026-03-10
fix(theme): debounce file watch events to prevent GCS rate limit errors

stephanie-shopify:
## Problem

When using \`theme dev --watch\`, chokidar fires one event per file save with no coalescing. Rapid successive saves to the same file (common with CLI watch mode or editor auto-save) trigger one GCS write per event, hitting GCS's per-object mutation rate limit (HTTP 429). This surfaces as unhandled \`Google::Cloud::ResourceExhaustedError\` errors on the server side.

Fixes shop/issues#15916.

## Solution

Add a 300ms trailing debounce keyed by file path for \`add\`/\`change\` events in \`handleFsEvent\`. If the same file fires multiple events within the window, only the last event triggers an upload. \`unlink\` events remain immediate.

Also add exponential backoff (1s for first retry, 2s for second) in \`handleFailedUploads\` so retry attempts don't immediately compound rate limit pressure.

## Expected behavior

A developer saving a file rapidly (e.g. editor auto-save firing multiple times, or CLI watch mode) triggers at most one upload per 300ms per file instead of one per save event. GCS 429 errors are no longer triggered by normal dev workflows.

## Testing

New tests added in \`theme-fs.test.ts\` (3 tests: debounces rapid same-file events, fires separate uploads for different files, does not debounce unlink) and \`theme-uploader.test.ts\` (1 test: verifies delay before retry). All existing tests pass.

### DIFF
--- a/packages/theme/src/cli/utilities/theme-fs.test.ts
+++ b/packages/theme/src/cli/utilities/theme-fs.test.ts
@@ -949,6 +949,120 @@ describe('theme-fs', () => {
     })
   })
 
+  describe('file event debouncing', () => {
+    const themeId = '1'
+    const adminSession = {token: 'token', storeFqdn: 'store.myshopify.com'} as AdminSession
+    const root = joinPath(locationOfThisFile, 'fixtures/theme')
+
+    beforeEach(() => {
+      const mockWatcher = new EventEmitter()
+      vi.spyOn(chokidar, 'watch').mockImplementation((_) => {
+        return mockWatcher as any
+      })
+    })
+
+    test('debounces rapid successive change events for the same file to a single upload', async () => {
+      // Given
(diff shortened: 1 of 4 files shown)


packages/theme/src/cli/utilities/theme-environment/proxy.test.ts
```diff
@@ -279,22 +279,34 @@ describe('dev proxy', () => {
...
           "X-Forwarded-For": "42",
           "accept": "text/html",
           "cookie": "oreo",
-          "user-agent": "vitest",
           "x-custom": "true",
         }
       `)
     })
+
+    test('removes browser user-agent so CLI user-agent is not overridden', () => {
```
stephanie-shopify, line 298: this does not seem right

PR #6976, opened 2026-03-11
remove simple git

ryancbahan:

This change removes the dependency on the `simple-git` library and replaces it with direct `git` command execution using `execa`. This PR is part of an assertion I believe to be true: with the rise of agentic development and parallel rise of supply chain attacks, external dependencies have become less valuable and higher risk. We should be looking at our dependencies with a critical eye and removing ones that don't have a very clear value proposition.

### Testing

- [x] shopify app init
- [x] shopify theme init
- [x] shopify app generate extension

### DIFF
--- a/packages/cli-kit/src/public/node/git.test.ts
+++ b/packages/cli-kit/src/public/node/git.test.ts
@@ -109,7 +109,7 @@ describe('downloadRepository()', async () => {
       const latestTag = true
 
       await git.downloadGitRepository({repoUrl, destination, latestTag})
-    }).rejects.toThrowError()
+    }).rejects.toThrowError(/fatal: No names found/)
   })
 
   test('clones and checks out the latest tag', async () => {
@@ -220,7 +220,7 @@ describe('createGitIgnore()', () => {
 
 describe('getLatestCommit()', () => {
   test('gets the latest commit through git log', async () => {
-    mockGitCommand('abc123\n2024-01-01\ncommit message\nHEAD -> main\n\nJohn\njohn@test.com')
+    mockGitCommand('abc123\x002024-01-01\x00commit message\x00HEAD -> main\x00\x00John\x00john@test.com')
 
     const result = await git.getLatestGitCommit()
     expect(result.hash).toBe('abc123')
@@ -236,7 +236,7 @@ describe('getLatestCommit()', () => {
 
   test('passes the directory option', async () => {
     const directory = '/test/directory'
-    mockGitCommand('abc123\n2024-01-01\nmsg\nrefs\n\nJohn\njohn@test.com')
+    mockGitCommand('abc123\x002024-01-01\x00msg\x00refs\x00\x00John\x00john@test.com')
 
     await git.getLatestGitCommit(directory)
 
--- a/packages/cli-kit/src/public/node/git.ts
+++ b/packages/cli-kit/src/public/node/git.ts
@@ -35,6 +35,9 @@ async function gitCommand(args: string[], directory?: string): Promise<string> {
     if (err instanceof Error) {
       const abortError = new AbortError(err.message)
       abortError.stack = err.stack
+      if ('exitCode' in err) {
+        Object.assign(abortError, {exitCode: err.exitCode})
+      }
       throw abortError
     }
     throw err
@@ -70,7 +73,7 @@ export async function checkIfIgnoredInGitRepository(directory: string, files: st
     return stdout.split('\n').filter(Boolean)
   } catch (error) {
     // git check-ignore exits with code 1 when no files are ignored
-    if (error instanceof AbortError) return []
+    if (error instanceof AbortError && 'exitCode' in error && error.exitCode === 1) return []
     throw error
   }
 }
@@ -138,14 +141,12 @@ export function addToGitIgnore(root: string, entry: string): void {
  *
  * @param repoUrl - The URL of the repository to clone.
  * @param destination - The directory where the repository will be cloned.
- * @param progressUpdater - A function that will be called with the progress of the clone.
  * @param shallow - Whether to clone the repository shallowly.
  * @param latestTag - Whether to clone the latest tag instead of the default branch.
  */
 export interface GitCloneOptions {
   repoUrl: string
   destination: string
-  progressUpdater?: (statusString: string) => void
   shallow?: boolean
   latestTag?: boolean
 }
@@ -249,7 +250,7 @@ async function getLatestTagFromDirectory(directory: string, repoUrl: string): Pr
  * @returns The latest commit of the repository.
  */
 export async function getLatestGitCommit(directory?: string): Promise<GitLogEntry> {
-  const format = '%H%n%ai%n%s%n%D%n%b%n%an%n%ae'
+  const format = '%H%x00%ai%x00%s%x00%D%x00%b%x00%an%x00%ae'
   const stdout = await gitCommand(['log', '-1', `--format=${format}`], directory)
   if (!stdout.trim()) {
     throw new AbortError(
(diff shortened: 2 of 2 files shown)


ryancbahan:
* **#6976** <a href="https://app.graphite.com/github/pr/Shopify/cli/6976?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Shopify/cli/6976?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/cli-kit/src/public/node/git.test.ts
```diff
@@ -53,210 +26,178 @@ vi.mock('./fs.js', async () => {
...
-        ...simpleGitProperties,
-        tags: mockedTags,
-      })
-
-      // When
       await git.downloadGitRepository({repoUrl, destination, latestTag})
-
-      // Then
-    }).rejects.toThrowError(/Couldn't obtain the most recent tag of the repository http:\/\/repoUrl/)
+    }).rejects.toThrowError()
```
dmerand: Nit: Is there a reason we can't add an error message to this check?

packages/cli-kit/src/public/node/git.ts
```diff
@@ -13,10 +13,34 @@ import {
...
+}
+
+async function gitCommand(args: string[], directory?: string): Promise<string> {
+  try {
+    const result = await execa('git', args, {cwd: directory})
+    return result.stdout
+  } catch (err) {
+    if (err instanceof Error) {
+      const abortError = new AbortError(err.message)
+      abortError.stack = err.stack
```
dmerand, line 37: 🐛 **Bug**: The `gitCommand` helper wraps all errors in `AbortError` but only preserves the `message` and `stack` properties. The `exitCode` property from execa errors is lost, making it impossible for calling code to distinguish between different types of git failures. This is the root cause of several error handling issues throughout the file: `getLatestTag`, `checkIfIgnoredInGitRepository`, and other functions need to check specific exit codes to differentiate expected failures (like 'no tags found') from actual errors (permission denied, git not installed).

**Suggestion:** Preserve the exitCode property when wrapping errors:

```suggestion
      const abortError = new AbortError(err.message)
      abortError.stack = err.stack
      if ('exitCode' in err) {
        Object.assign(abortError, {exitCode: err.exitCode})
      }
      throw abortError
```

packages/cli-kit/src/public/node/git.ts
```diff
@@ -43,7 +65,14 @@ export async function initializeGitRepository(directory: string, initialBranch =
  * @returns Files ignored by the lockfile.
  */
 export async function checkIfIgnoredInGitRepository(directory: string, files: string[]): Promise<string[]> {
-  return withGit({directory}, (repo) => repo.checkIgnore(files))
+  try {
+    const stdout = await gitCommand(['check-ignore', ...files], directory)
+    return stdout.split('\n').filter(Boolean)
+  } catch (error) {
+    // git check-ignore exits with code 1 when no files are ignored
+    if (error instanceof AbortError) return []
```
dmerand: 🐛 **Bug**: The function catches any `AbortError` and returns an empty array, assuming it means 'no files are ignored'. However, since `gitCommand` wraps ALL errors in `AbortError` (including permission errors, invalid directories, or git not being installed), this will silently hide real failures. The comment at line 72 indicates the intent is to handle exit code 1 specifically, but the implementation doesn't check for this.

**Suggestion:** Check for the specific exit code to distinguish 'no files ignored' from actual errors. Note: This requires fixing the gitCommand helper to preserve exitCode first:

```suggestion
    if (error instanceof AbortError && 'exitCode' in error && error.exitCode === 1) return []
```

packages/cli-kit/src/public/node/git.ts
dmerand, line 1: (Ref Line 152) 🐛 **Bug**: The `progressUpdater` parameter is still declared in the `GitCloneOptions` interface but is no longer used in the implementation (removed at line 160). This is a silent breaking change - callers can pass this parameter expecting progress updates, but it will be ignored without any error or warning. While no code in the codebase currently uses this parameter, it's still part of the public API contract.

**Suggestion:** Either remove the parameter from the interface to accurately reflect the implementation, or add a deprecation notice:

```suggestion
  /** @deprecated Progress updates are no longer supported after removing simple-git dependency */
  progressUpdater?: (statusString: string) => void
```

Alternatively, implement progress tracking using execa's stdio events, or remove it entirely with a migration note.

packages/cli-kit/src/public/node/git.ts
```diff
@@ -231,17 +248,27 @@ async function getLocalLatestTag(repository: SimpleGit, repoUrl: string): Promis
...
+  if (!stdout.trim()) {
     throw new AbortError(
       'Must have at least one commit to run command',
       outputContent`Run ${outputToken.genericShellCommand(
         "git commit -m 'Initial commit'",
       )} to create your first commit.`,
     )
   }
-  return logs.latest
+  const lines = stdout.split('\n')
```
dmerand: 🐛 **Bug**: The `getLatestGitCommit` function splits git log output on newlines, but the commit body (`%b` in the format string) can contain multiple lines. This will cause the parser to incorrectly assign array indices, breaking extraction of `author_name` and `author_email`. For example, if a commit body has 3 lines, `lines[5]` will be the second line of the body instead of the author name.

**Suggestion:** Use a null byte delimiter that won't appear in commit messages:

```suggestion
  const format = '%H%x00%ai%x00%s%x00%D%x00%b%x00%an%x00%ae'
  const stdout = await gitCommand(['log', '-1', `--format=${format}`], directory)
  if (!stdout.trim()) {
    throw new AbortError(
      'Must have at least one commit to run command',
      outputContent`Run ${outputToken.genericShellCommand(
        "git commit -m 'Initial commit'",
      )} to create your first commit.`,
    )
  }
  const lines = stdout.split('\x00')
```

packages/cli-kit/src/public/node/git.ts
```diff
@@ -332,7 +361,13 @@ export async function ensureInsideGitDirectory(directory?: string): Promise<void
  * @returns True if the directory is inside a .git directory tree.
  */
 export async function insideGitDirectory(directory?: string): Promise<boolean> {
-  return withGit({directory}, (repo) => repo.checkIsRepo())
+  try {
+    await execa('git', ['rev-parse', '--git-dir'], {cwd: directory})
+    return true
+  } catch (error) {
+    if (error instanceof Error && 'exitCode' in error) return false
```
dmerand: 🐛 **Bug**: The function returns `false` for any error with an `exitCode` property, but should specifically check for exit code 128 which indicates 'not a git directory'. Other exit codes might indicate actual errors that should be thrown rather than silently treated as 'not a git repo'.

**Suggestion:** Check specifically for exit code 128:

```suggestion
    if (error instanceof Error && 'exitCode' in error && error.exitCode === 128) return false
```

packages/cli-kit/src/public/node/git.ts
```diff
@@ -365,8 +401,13 @@ export async function isClean(directory?: string): Promise<boolean> {
  * @returns String with the latest tag or undefined if no tags are found.
  */
 export async function getLatestTag(directory?: string): Promise<string | undefined> {
-  const tags = await withGit({directory}, (repo) => repo.tags())
-  return tags.latest
+  try {
+    const stdout = await gitCommand(['describe', '--tags', '--abbrev=0'], directory)
+    return stdout.trim() || undefined
+  } catch (error) {
+    if (error instanceof AbortError) return undefined
```
dmerand: 🐛 **Bug**: This function catches any `AbortError` and returns `undefined`, assuming it means 'no tags exist'. However, this will also hide real errors like permission issues, invalid directories, or git not being installed. The test at line 391 specifically creates an error with exitCode 128 and message 'fatal: No names found', but the implementation doesn't check for these specific conditions - it returns undefined for ANY AbortError.

**Suggestion:** Check for the specific error condition. Note: This requires fixing the gitCommand helper to preserve exitCode first:

```suggestion
    if (error instanceof AbortError && ('exitCode' in error && error.exitCode === 128 || error.message.includes('No names found'))) return undefined
```

PR #6996, opened 2026-03-12
Remove Cucumber test infrastructure

ryancbahan:
NOTE: I will update the branch protection rules to remove cucumber tests once the pr is approved, to reduce diff in time where rules are disabled for actively running infra.

## Summary

- Remove the entire `packages/features` Cucumber test suite, which has been superseded by the Playwright-based e2e tests in `packages/e2e`
- Remove all CI jobs/steps that ran Cucumber acceptance tests (PR, main, and manual workflows)
- Clean up root `package.json` scripts (`test:features`, `test:regenerate-snapshots`), knip config, and ESLint overrides
- Update testing strategy docs to reference e2e tests instead of Cucumber

## Cucumber scenarios to migrate to e2e (followup)

The following active scenarios should be considered for migration to the Playwright e2e suite:

- **App scaffold + build** — creates react-router and extension-only apps, then builds them (will add to e2e)
- **CLI command snapshot** — lists all commands and validates against a snapshot (do we want snapshots?)
- **GitHub Actions pinning** — validates non-official actions are pinned (could also be a lint rule)
- **Shared node deps sync** — checks shared dependencies are on the same version (could also be a lint rule)
- **Theme** — was entirely commented out, nothing to migrate

## Test plan

- [x] `pnpm-lock.yaml` regenerated cleanly
- [x] `package.json` validates as valid JSON
- [x] No remaining references to `@shopify/features`, `cucumber`, or `test:features` in code
- [x] CI passes without the removed acceptance-tests job

### DIFF
--- a/docs/cli/testing-strategy.md
+++ b/docs/cli/testing-strategy.md
@@ -30,7 +30,7 @@ test("loads the app", async () => {
 })
 ```
 
-Tests can be run with `pnpm test` or `pnpm test:watch` for the entire workspace or individual packages. `test:watch` keeps the process alive and runs tests as files are changed. If you want to run a single test, pass the path to the file as argument:
+Tests can be run with `pnpm test` for the Vitest suite, `pnpm test:watch` for watch mode, or `pnpm test:e2e` for the Playwright end-to-end suite. If you want to run a single unit test, pass the path to the file as argument:
 
 ```
 pnpm test path/to/my.test.ts


ryancbahan:
* **#6997** <a href="https://app.graphite.com/github/pr/Shopify/cli/6997?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6996** <a href="https://app.graphite.com/github/pr/Shopify/cli/6996?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Shopify/cli/6996?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

.github/workflows/tests-pr.yml
dmerand, line 1: (Ref Line 204) Note that we should make a plan for activating the E2E tests as blocking.

docs/cli/testing-strategy.md
dmerand, line 1: (Ref Line 33) One thing I appreciate about the AI reviewer is that it checks the whole code base + finds things that would be easy to miss:

💡 **Improvement**: The documentation now appears out of sync with the root scripts. `package.json` maps `pnpm test` to `pnpm vitest run`, so contributors following this page could reasonably think they exercised the full test suite when they actually skipped Playwright.

**Suggestion:** Consider updating the sentence to distinguish unit and end-to-end entry points explicitly:
```suggestion
Tests can be run with `pnpm test` for the Vitest suite, `pnpm test:watch` for watch mode, or `pnpm test:e2e` for the Playwright end-to-end suite. If you want to run a single unit test, pass the path to the file as argument:
```

packages/features/features/commands.feature
```diff
@@ -1,5 +0,0 @@
-Feature: Loading all commands
-
-Scenario: No runtime errors happen when loading the CLI modules graph
```
dmerand, line 3: 🐛 **Bug**: This small feature was doing more than snapshot maintenance: it exercised `shopify commands --tree` and verified that the full command graph could load successfully. That catches import-time or registration failures in commands that targeted Playwright journeys may never touch. If it is removed without a replacement, broken command wiring may only surface when a user hits the affected command.

**Suggestion:** Consider porting this to a lightweight smoke test or CI step before deleting it—for example, a small spec that runs `shopify commands --tree` and asserts successful loading, with or without the snapshot comparison.

packages/features/features/github-actions.feature
```diff
@@ -1,5 +0,0 @@
-Feature: Github Actions
-
-Scenario: All non-official github actions are pinned
```
dmerand, line 3: 🔒 **Security**: Worth reviewing whether this Cucumber scenario is being replaced elsewhere before it is removed. The deleted feature and step implementation enforced a repo-wide policy that non-official GitHub actions under `.github` must be pinned to commit SHAs, and there does not appear to be a corresponding lint rule or CI script added in this PR. Without a replacement, future action references can drift back to mutable tags without automation catching it.

**Suggestion:** Consider moving this policy into a lightweight standalone script or lint-style CI check before deleting the Cucumber version, so unpinned third-party actions still fail automation.

docs/cli/testing-strategy.md
ryancbahan: updated

packages/features/features/commands.feature
ryancbahan: this was noted in pr desc and a lighter vitest replacement is coming on top of this pr

ryancbahan:
@dmerand I've got a pr incoming on top of this to add tests for action SHA and command graph snapshot.

PR #7005, opened 2026-03-13
fix: infer uidStrategy from backend typename

alfonso-noriega:
## Problem

Channel extensions (`channel_config`) were always shown as **deleted + new** on every deploy, even when no changes were made to them.

### Root cause

The `channel_config` local spec is defined via `createContractBasedModuleSpecification`, which defaults to `uidStrategy: 'uuid'` (the default for non-configuration specs). During the merge of local and remote specs in `mergeLocalAndRemoteSpecs`, the local value wins because `FlattenedRemoteSpecification` has no `uidStrategy` field — only the binary `options.uidIsClientProvided` flag.

The backend knows `channel_config` should use `'single'` strategy, but the CLI was unable to receive that signal. The `uidIsClientProvided` flag cannot distinguish between `'single'` and `'dynamic'` (both map to `false`), so the merge always kept the local default of `'uuid'`.

With `uidStrategy: 'uuid'`, the extension goes through UUID-based matching. The remote registration lives in the single/config group and can never match → it appears as deleted, and the local copy appears as new, on every deploy.

## Solution

The backend `uidStrategy` GraphQL field is a union type whose `__typename` encodes the concrete strategy (`single`, `dynamic`, or client-provided/`uuid`). The GraphQL document already requests `__typename` — we just weren't using it.

This PR:

1. **`specifications.graphql`** — explicitly adds `__typename` to the `uidStrategy` selection so it becomes part of the source-of-truth query (and will be properly typed after the next codegen run).

2. **`extension_specifications.ts`** — adds `uidStrategy?: 'single' | 'dynamic' | 'uuid'` to `RemoteSpecification.options` to carry the backend-resolved value through the pipeline.

3. **`app-management-client.ts`** — adds `uidStrategyFromTypename()` that maps `__typename + isClientProvided` → CLI `UidStrategy`, and populates `options.uidStrategy` when building `RemoteSpecification` objects.

4. **`fetch-extension-specifications.ts`** — applies `options.uidStrategy` as an override to `merged.uidStrategy` for **all** specs when it is present (app-management API path). The Partners API path is unaffected since it never sets `options.uidStrategy`.

### ⚠️ One thing to verify

The dynamic typename used in `uidStrategyFromTypename` is currently `'AppModuleDynamicUidStrategy'`. Please confirm this matches what the app-management backend actually returns and update if needed.

## Testing steps

1. **Manual — channel deploy regression**
   - Have an app with a `channel_config` extension already deployed.
   - Run `shopify app deploy` twice in a row with no local changes.
   - **Before fix:** second deploy shows the channel extension as _removed_ + _new_.
   - **After fix:** second deploy shows the channel extension as _unchanged_ (no prompt required).

2. **Manual — other extension types unaffected**
   - Verify that UI extensions (uuid strategy), app config extensions (single strategy), and webhook subscriptions (dynamic strategy) continue to match correctly across deploys.

3. **Unit tests**
   ```
   pnpm nx run app:vitest -- src/cli/services/generate/fetch-extension-specifications.test.ts
   pnpm nx run app:vitest -- src/cli/utilities/developer-platform-client/app-management-client.test.ts
   ```

4. **Confirm `__typename` value** — after deploying to a test environment, add a temporary `console.log` in `uidStrategyFromTypename` to confirm the actual typename the backend returns for `channel_config` and for webhook subscriptions.

### DIFF
--- a/packages/app/src/cli/services/generate/fetch-extension-specifications.ts
+++ b/packages/app/src/cli/services/generate/fetch-extension-specifications.ts
@@ -90,6 +90,8 @@ async function mergeLocalAndRemoteSpecs(
     merged.uidStrategy = merged.options.uidStrategy ?? localSpec.uidStrategy ?? 'single'
 
     // If configuration is inside an app.toml -- i.e. single UID mode -- we must be able to parse a partial slice.
+    // DEPRECATED: not all single specs are config specs.
+    // Should be removed once we can get the experience from the API.
     let handleInvalidAdditionalProperties: HandleInvalidAdditionalProperties
     switch (merged.uidStrategy) {
       case 'uuid':
@@ -101,6 +103,11 @@ async function mergeLocalAndRemoteSpecs(
       case 'dynamic':
         handleInvalidAdditionalProperties = 'fail'
         break
+    }
+
+    // If the experience is 'configuration', force strip.
+    if (merged.experience === 'configuration') {
+      handleInvalidAdditionalProperties = 'strip'
     }
 
     const parseConfigurationObject = await unifiedConfigurationParserFactory(merged, handleInvalidAdditionalProperties)


alfonso-noriega:
* **#7005** <a href="https://app.graphite.com/github/pr/Shopify/cli/7005?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Shopify/cli/7005?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/app/src/cli/utilities/developer-platform-client/app-management-client.ts
```diff
@@ -1356,6 +1357,26 @@ function experience(identifier: string): 'configuration' | 'extension' {
...
 }
 
+/**
+ * Maps the backend uidStrategy __typename to the CLI UidStrategy value.
+ * The __typename is always present in the response because the GraphQL document
+ * selects it, even though the generated TypeScript type doesn't expose it directly.
+ *
+ * Type names come from the app-management GraphQL schema's UidStrategy union.
+ */
+function uidStrategyFromTypename(typename: string): 'single' | 'dynamic' | 'uuid' | undefined {
```
dmerand: We should have tests for this new method.

packages/app/src/cli/services/generate/fetch-extension-specifications.ts
```diff
@@ -72,13 +72,22 @@ async function mergeLocalAndRemoteSpecs(
...
+      localSpec.uidStrategy =
+        remoteSpec.options.uidStrategy ?? (remoteSpec.options.uidIsClientProvided ? 'uuid' : 'single')
     }
     if (!localSpec) return undefined
 
     const merged = {...localSpec, ...remoteSpec, loadedRemoteSpecs: true} as RemoteAwareExtensionSpecification &
       FlattenedRemoteSpecification
 
+    // Always prefer the backend-derived uidStrategy (from __typename) when available.
+    // This correctly overrides the local spec's default (e.g. channel_config defaults to 'uuid'
```
isaacroldan, line 87: we should also force `channel_config` to be `single` locally.

packages/app/src/cli/services/context/identifiers-extensions.ts
```diff
@@ -40,7 +40,7 @@ export async function ensureExtensionsIds(
 ) {
   let remoteExtensions = initialRemoteExtensions
   const identifiers = options.envIdentifiers.extensions ?? {}
-  const localExtensions = options.app.allExtensions.filter((ext) => ext.isUUIDStrategyExtension)
+  const localExtensions = options.app.allExtensions.filter((ext) => ext.specification.experience === 'extension')
```
isaacroldan: Maybe we can make this more explicit by using `!ext.isAppConfigExtension`? Not a strong opinion though

PR #7008, opened 2026-03-13
fix(cli-kit): batch large output chunks to prevent event loop blocking

alfonso-noriega:
## Problem

When a POS UI extension throws a large stack trace, the entire output arrives as a single `write()` call to `ConcurrentOutput`'s `Writable` stream. The synchronous processing of large chunks — `stripAnsi()`, `split(/\n/)`, and `setProcessOutput` with a spread of the full accumulated array — causes a render cycle expensive enough to block the Node.js event loop. During that time, Ink's `useInput` hook cannot process keypresses, so `q` (quit) and `p` (preview) become unresponsive.

## Fix

The `writableStream` write handler now uses a two-stage pipeline: a `Transform` stream that splits large chunks into batches of `MAX_LINES_PER_BATCH` (20) lines, followed by a `Writable` sink that renders each batch. For batches from large chunks, `setImmediate(next)` yields to the event loop between renders so Ink's input handling can run.

- Chunks ≤ 20 lines: unchanged behavior, `next()` called synchronously
- Chunks > 20 lines: split into 20-line batches with `setImmediate` scheduling between each batch

The Transform stream preserves the existing `outputContextStore` context (prefix and stripAnsi overrides) and maintains proper Node.js backpressure by deferring `next()` until each batch is processed.

## Test

Added a test that writes 250 lines as a single chunk and verifies all lines appear in the rendered output without being dropped during batch processing.

### DIFF
--- a/packages/cli-kit/src/private/node/ui/components/ConcurrentOutput.test.tsx
+++ b/packages/cli-kit/src/private/node/ui/components/ConcurrentOutput.test.tsx
@@ -262,6 +262,41 @@ describe('ConcurrentOutput', () => {
     expect(logColumns[1]?.length).toBe(25 + 2)
   })
 
+  test('renders large chunks split into batches without dropping lines', async () => {
+    // Given - simulate a large stack trace (>100 lines) arriving as a single write
+    const processSync = new Synchronizer()
+    const lineCount = 250
+    const largeOutput = Array.from({length: lineCount}, (_, i) => `line ${i + 1}`).join('\n')
+
+    const processes = [
+      {
+        prefix: 'pos-ext',
+        action: async (stdout: Writable, _stderr: Writable, _signal: AbortSignal) => {
+          stdout.write(largeOutput)
+          processSync.resolve()
+        },
+      },
+    ]
+
+    // When - keepRunningAfterProcessesResolve prevents the component from unmounting
+    // before all setImmediate-batched state updates have been applied
+    const renderInstance = render(
+      <ConcurrentOutput
+        processes={processes}
+        abortSignal={new AbortController().signal}
+        keepRunningAfterProcessesResolve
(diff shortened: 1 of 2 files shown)


alfonso-noriega:
* **#7008** <a href="https://app.graphite.com/github/pr/Shopify/cli/7008?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Shopify/cli/7008?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

ryancbahan (changes requested):
nice job! the current approach is in the right direction, but using `setImmediate` with recursion isn't going to give you backpressure (something node is good at managing), it's just spreading work across event loop ticks. i'm, pretty sure you can leverage node's streaming to split work + auto-support backpressure with something like this:

```js
 const splitter = new Transform({
    transform(chunk, _encoding, callback) {
      const lines = chunk.toString('utf8').replace(/\n$/, '').split('\n')
      for (let i = 0; i < lines.length; i += MAX_LINES_PER_BATCH) {
        this.push(lines.slice(i, i + MAX_LINES_PER_BATCH).join('\n'))
      }
      callback()
    },
  })

  const writable = new Writable({
    write(chunk, _encoding, next) {
      // existing logic, unchanged — always gets small chunks now
      const log = chunk.toString('utf8')
      // ...
      next()
    },
  })

  splitter.pipe(writable)
  // hand `splitter` to the process as its stdout/stderr
```

that way, when writer hasn't called `next()` yet, the internal buffer fills up, and node automatically pauses the producer. that will allow us to be a bit more blind to the chunking management in the main flow and allow node to stream accordingly to the local machine's memory limits.

PR #7013, opened 2026-03-13
Smaller bundle

gonzaloriestra:

The CLI bundle is larger than necessary.

Reduces CLI bundle size by ~55% ([62 MB](https://www.npmjs.com/package/@shopify/cli/v/3.92.1) → [28 MB](https://www.npmjs.com/package/@shopify/cli/v/0.0.0-snapshot-20260319110410))

- Enable `minifyWhitespace` and `minifyIdentifiers` in esbuild (previously only `minifySyntax` was on)
- Exclude source map files from the npm package via `!/dist/**/*.map` in `files` (maps are still generated for Observe)
- Replace `brotli` JS package with native `zlib.brotliCompressSync`
- Simplify entry points instead of globbing all `.ts` files

### How to test your changes?

- `pnpm bundle-for-release` and check `dist` folder
- `npm i -g --@shopify:registry=https://registry.npmjs.org ￼@0.0.0-snapshot-20260318104923`
- `shopify app deploy` (exercises brotli compression)

### Measuring impact

### Checklist

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -351,9 +351,6 @@ importers:
       bottleneck:
         specifier: 2.19.5
         version: 2.19.5
-      brotli:
-        specifier: 1.3.3
-        version: 1.3.3
       chalk:
         specifier: 5.4.1
         version: 5.4.1
@@ -490,9 +487,6 @@ importers:
         specifier: 3.24.4
         version: 3.24.4
     devDependencies:
-      '@types/brotli':
-        specifier: ^1.3.4
-        version: 1.3.4
       '@types/commondir':
         specifier: ^1.0.0
         version: 1.0.2
@@ -3895,9 +3889,6 @@ packages:
   '@types/babel__traverse@7.28.0':
     resolution: {integrity: sha512-8PvcXf70gTDZBgt9ptxJ8elBeBjcLOAcOtoO/mPJjtji1+CdGbHgm77om1GrsPxsiE+uXIpNSK64UYaIwQXd4Q==}
 
-  '@types/brotli@1.3.4':
-    resolution: {integrity: sha512-cKYjgaS2DMdCKF7R0F5cgx1nfBYObN2ihIuPGQ4/dlIY6RpV7OWNwe9L8V4tTVKL2eZqOkNM9FM/rgTvLf4oXw==}
-
   '@types/chai@5.2.3':
     resolution: {integrity: sha512-Mw558oeA9fFbv65/y4mHtXDs9bPnFMZAL/jxdPFUpOHHIXX91mcgEHbS5Lahr+pwZFR8A7GQleRWeI6cGFC2UA==}
 
@@ -4578,9 +4569,6 @@ packages:
   braces@3.0.3:
     resolution: {integrity: sha512-yQbXgO/OSZVD2IsiLlro+7Hf6Q18EJrKSEsdoMzKePKXct3gvD8oLcOQdIzGupr5Fj+EDe8gO/lxc1BzfMpxvA==}
     engines: {node: '>=8'}
-
-  brotli@1.3.3:
-    resolution: {integrity: sha512-oTKjJdShmDuGW94SyyaoQvAjf30dZaHnjJ8uAF+u2/vGJkJbJPJAT1gDiOJP5v1Zb6f9KEyW/1HpuaWIXtGHPg==}
 
   browserify-zlib@0.1.4:
     resolution: {integrity: sha512-19OEpq7vWgsH6WkvkBJQDFvJS1uPcbFOQ4v9CU839dO+ZZXUZO6XpE6hNCqvlIIj+4fZvRiJ6DsAQ382GwiyTQ==}
@@ -13188,10 +13176,6 @@ snapshots:
     dependencies:
       '@babel/types': 7.29.0
 
-  '@types/brotli@1.3.4':
-    dependencies:
-      '@types/node': 18.19.70
-
   '@types/chai@5.2.3':
     dependencies:
       '@types/deep-eql': 4.0.2
@@ -13964,10 +13948,6 @@ snapshots:
   braces@3.0.3:
     dependencies:
       fill-range: 7.1.1
-
-  brotli@1.3.3:
-    dependencies:
-      base64-js: 1.5.1
 
   browserify-zlib@0.1.4:
     dependencies:


packages/cli/bin/bundle.js
```diff
@@ -21,6 +21,128 @@ const external = [
...
+  'proper-lockfile',
+  'is-interactive',
+  'latest-version',
+  'get-port-please',
+  'mrmime',
+  'gradient-string',
+  'is-executable',
+  'macaddress',
+  'network-interfaces',
+  'diff',
```
isaacroldan: I mean, putting all dependencies as external defeats the purpose of a bundle 😅

packages/cli/bin/bundle.js
```diff
@@ -53,14 +175,14 @@ esBuild({
   },
   inject: ['../../bin/bundling/cjs-shims.js'],
   external,
-  sourcemap: true,
+  sourcemap: false,
```
isaacroldan: We need sourcemaps for Observe. They are not included in the final bundle though (they shouldn't at least)

packages/cli/bin/bundle.js
```diff
@@ -40,7 +49,7 @@ const hydrogenAssets = joinPath(hydrogenPath, 'dist/assets/hydrogen/**/*')
 
 esBuild({
   bundle: true,
-  entryPoints: ['./src/**/*.ts'],
+  entryPoints: (await glob('./src/**/*.ts')).filter(f => !f.includes('.test.')),
```
isaacroldan: entry point can be simplified to `['./src/index.ts']`

packages/cli/bin/bundle.js
```diff
@@ -53,14 +62,14 @@ esBuild({
   },
   inject: ['../../bin/bundling/cjs-shims.js'],
   external,
-  sourcemap: true,
+  sourcemap: 'external',
```
isaacroldan, line 56: what does this do?

packages/cli/bin/bundle.js
```diff
@@ -19,8 +19,17 @@ const external = [
...
   '@ast-grep/napi',
+  // prettier is ~4MB and only used for formatting generated types.
+  // Lazily loaded via dynamic import, safe to externalize.
+  'prettier',
+  // ts-morph + typescript are ~19MB. Used by Hydrogen for JS/TS transpilation.
+  // Already lazily loaded via dynamic import, safe to externalize.
+  'ts-morph',
+  // typescript compiler (~9MB) is pulled in by ts-morph and json-schema-to-typescript.
+  // Available at runtime as a project dependency.
+  'typescript',
```
isaacroldan: making this external means that they are need to be installed by npm, so they don't count for the "bundle size", but this is actually slower because npm needs to find them, resolve them, and download them.

We should only mark as external dependencies that are incompatible with bundling (like esbuild)

packages/cli/bin/bundle.js
```diff
@@ -40,7 +40,7 @@ const hydrogenAssets = joinPath(hydrogenPath, 'dist/assets/hydrogen/**/*')
 
 esBuild({
   bundle: true,
-  entryPoints: ['./src/**/*.ts'],
+  entryPoints: ['./src/index.ts'],
```
dmerand: The change from globbing all .ts files to a single `"./src/index.ts"` entry point breaks the oclif hooks configuration.

## The Problem

The oclif configuration in packages/cli/package.json expects these hook files:
```json
"hooks": {
  "prerun": "./dist/hooks/prerun.js",
  "postrun": "./dist/hooks/postrun.js"
}
```

These source files exist:
- packages/cli/src/hooks/prerun.ts
- packages/cli/src/hooks/postrun.ts

However, packages/cli/src/index.ts does NOT import or export these hooks. They are meant to be standalone files per oclif convention.

## Impact

**With the old config** (`entryPoints: ["./src/**/*.ts"]`):
- esbuild finds all .ts files via glob
- Creates corresponding .js files maintaining directory structure
- Outputs `dist/hooks/prerun.js` and `dist/hooks/postrun.js` ✅

**With the new config** (`entryPoints: ["./src/index.ts"]`):
- esbuild only bundles what's reachable from index.ts
- Since prerun/postrun aren't imported by index.ts, they're unreachable
- **No `dist/hooks/prerun.js` or `dist/hooks/postrun.js` files are created** ❌
- oclif will fail when trying to load these hooks at runtime

## Fix Options

1. **Keep the glob pattern**: `entryPoints: ["./src/**/*.ts"]`
2. **Add explicit entry points**: `entryPoints: ["./src/index.ts", "./src/hooks/prerun.ts", "./src/hooks/postrun.ts"]`
3. Import/export the hooks in index.ts (not recommended - they're meant to be separate files per oclif convention)

packages/cli/bin/bundle.js
dmerand, line 1: (Ref Line 56) 🐛 **Bug**: The configuration uses `sourcemap: true`, which generates inline sourcemaps embedded directly in the .js files rather than separate .map files. This means the `!/dist/**/*.map` exclusion in package.json has no effect, and sourcemaps are shipped to npm users inside the bundles, directly contradicting the PR's goal of reducing bundle size by ~55%. An earlier commit noted that disabling sourcemaps saved 35,752 KB.

**Suggestion:** 
```suggestion
  sourcemap: 'external',
```

This will generate separate .map files that can be uploaded to Observe while being excluded from the npm package via the existing `!/dist/**/*.map` pattern.

gonzaloriestra: Well, the configuration didn't change, so definitely not a bug. Anyway, I changed it, although the size is the same.

packages/cli/bin/bundle.js
gonzaloriestra: Good catch, thanks! It was indeed skipping the hooks (analytics, for example). I added it and now it's working again.

packages/cli/bin/bundle.js
gonzaloriestra: The esbuild sourcemap options are documented here: https://esbuild.github.io/api/#sourcemap

The four modes:

- **`true`** — separate `.map` file + `//# sourceMappingURL=` comment in JS
- **`'external'`** — separate `.map` file, no comment in JS
- **`'inline'`** — base64 source map embedded in JS, no `.map` file
- **`'both'`** — separate `.map` file + inline base64 in JS

packages/cli/bin/bundle.js
gonzaloriestra: > The configuration uses sourcemap: true, which generates inline sourcemaps embedded directly in the .js files rather than separate .map files

This is wrong, btw. This is the difference from [the docs](https://esbuild.github.io/api/#sourcemap):

- true (linked) — separate .map file + //# sourceMappingURL= comment in JS
- 'external' — separate .map file, no comment in JS
- 'inline' — base64 source map embedded in JS, no .map file
- 'both' — separate .map file + inline base64 in JS

We upload the .map files to Observe, so we don't need the inline comments.

PR #6869, opened 2026-02-19
Add a build manifest build step

alfonso-noriega:
# Add `build_manifest` build step

## Summary

This PR adds the `build_manifest` step type to the client steps pipeline introduced in [#01-build-steps-infrastructure](#) and wired in [#02-wire-build-config-into-extension-specs](#).

The step generates a JSON manifest that maps asset keys to their resolved filepaths and source modules. It supports two modes: a flat single-extension manifest and a forEach mode that iterates a config array and writes one entry per item. A `manifest_result` inclusion type is also added to `include_assets` so downstream steps can consume the manifest output to copy static assets.

**Depends on:** `02-wire-build-config-into-extension-specs`

---

### Config schema

```typescript
{
  outputFile?: string              // default: 'build-manifest.json'
  forEach?: {
    tomlKey: string                // config key pointing to an array to iterate
    keyBy: string                  // field used to identify each item (e.g. 'target')
  }
  assets: {
    [assetKey: string]: {
      moduleKey: string            // dot-notation path into extension config
      static?: boolean             // true → preserve source extension; false → .js
      optional?: boolean           // true → silently omit if unresolved
    }
  }
}
```

### Single mode

One manifest file for the whole extension:

```typescript
{
  id: 'build-manifest',
  type: 'build_manifest',
  config: {
    assets: {
      main:         {moduleKey: 'module'},
      tools:        {moduleKey: 'tools', static: true},
      instructions: {moduleKey: 'instructions.file', optional: true},
    },
  },
}
```

Output written to `build-manifest.json`:

```json
{
  "assets": {
    "main":  {"filepath": "my-ext-main.js",    "module": "src/index.tsx"},
    "tools": {"filepath": "my-ext-tools.json", "module": "src/tools.json", "static": true}
  }
}
```

### forEach mode

One entry per item in a config array:

```typescript
{
  id: 'build-manifest',
  type: 'build_manifest',
  config: {
    forEach: {tomlKey: 'targeting', keyBy: 'target'},
    assets: {
      main:          {moduleKey: 'module'},
      should_render: {moduleKey: 'should_render.module', optional: true},
    },
  },
}
```

Output:

```json
[
  {
    "target": "purchase.checkout.block.render",
    "build_manifest": {
      "assets": {
        "main": {"filepath": "my-ext-purchase.checkout.block.render-main.js", "module": "src/checkout.tsx"}
      }
    }
  }
]
```

[...]

PR #6868, opened 2026-02-19
Replace hosted_app_home build mode with copy_files mode

alfonso-noriega:
### WHY are these changes introduced?

Fixes https://github.com/shop/issues-admin-extensibility/issues/2240

### Measuring impact

How do we know this change was effective? Please choose one:

- [ ] n/a - this doesn't need measurement, e.g. a linting rule or a bug-fix
- [ ] Existing analytics will cater for this addition
- [ ] PR includes analytics changes to measure impact

### Checklist

- [ ] I've considered possible cross-platform impacts (Mac, Linux, Windows)
- [ ] I've considered possible [documentation](https://shopify.dev) changes

PR #7016, opened 2026-03-16
Add manifest generation to include_assets build step

alfonso-noriega:

Fixes #0000 

### Measuring impact

### Checklist

### DIFF
--- a/packages/app/src/cli/services/build/steps/include-assets-step.ts
+++ b/packages/app/src/cli/services/build/steps/include-assets-step.ts
@@ -6,6 +6,8 @@ import {joinPath, dirname, extname, sanitizeRelativePath} from '@shopify/cli-kit
 import {z} from 'zod'
 import type {LifecycleStep, BuildContext} from '../client-steps.js'
 
+export type {IncludeAssetsConfig}
+
 /**
  * Pattern inclusion entry.
  *
@@ -36,8 +38,7 @@ const StaticEntrySchema = z.object({
  *
  * Resolves a path (or array of paths) from the extension configuration and
  * copies the directory contents into the output. Silently skipped when the
- * key is absent. Respects `preserveStructure` and `destination` the same way
- * as the static entry.
+ * key is absent.
  *
  * `anchor` — the config key path whose array value provides the grouping
  * dimension. Each array item becomes one top-level manifest entry, keyed by
@@ -52,7 +53,6 @@ const ConfigKeyEntrySchema = z.object({
   type: z.literal('configKey'),
   key: z.string(),
   destination: z.string().optional(),
-  preserveStructure: z.boolean().default(false),
   anchor: z.string().optional(),
   groupBy: z.string().optional(),
 })
@@ -78,6 +78,7 @@ const IncludeAssetsConfigSchema = z
     inclusions: z.array(InclusionEntrySchema),
     generatesAssetsManifest: z.boolean().default(false),
   })
+  .strict()
   .superRefine((data, ctx) => {
     for (const [i, entry] of data.inclusions.entries()) {
       if (entry.type === 'configKey') {
@@ -93,6 +94,13 @@ const IncludeAssetsConfigSchema = z
       }
     }
   })
+
+/**
+ * TypeScript type for the include_assets step config.
+ * Derived from the Zod schema so that static type-checking catches
+ * mismatched property names at compile time.
+ */
+type IncludeAssetsConfig = z.input<typeof IncludeAssetsConfigSchema>
 
 /**
  * Executes an include_assets build step.
--- a/packages/app/src/cli/services/build/steps/include-assets/copy-config-key-entry.ts
+++ b/packages/app/src/cli/services/build/steps/include-assets/copy-config-key-entry.ts
@@ -29,7 +29,7 @@ export async function copyConfigKeyEntry(
   },
   options: {stdout: NodeJS.WritableStream},
 ): Promise<{filesCopied: number; pathMap: Map<string, string | string[]>}> {
-  const {key, baseDir, outputDir, context, preserveStructure, destination} = config
+  const {key, baseDir, outputDir, context, destination} = config
   const value = getNestedValue(context.extension.configuration, key)
   let paths: string[]
   if (typeof value === 'string') {
@@ -67,16 +67,12 @@ export async function copyConfigKeyEntry(
 
     const sourceIsDir = await isDirectory(fullPath)
 
-    const destDir =
-      sourceIsDir && preserveStructure ? joinPath(effectiveOutputDir, basename(fullPath)) : effectiveOutputDir
+    const destDir = effectiveOutputDir
 
     if (sourceIsDir) {
       await copyDirectoryContents(fullPath, destDir)
       const copied = await glob(['**/*'], {cwd: destDir, absolute: false})
-      const msg = preserveStructure
-        ? `Copied '${sourcePath}' to ${basename(fullPath)}\n`
-        : `Copied contents of '${sourcePath}' to output root\n`
-      options.stdout.write(msg)
+      options.stdout.write(`Included '${sourcePath}'\n`)
       const relFiles = copied.map((file) => relativePath(outputDir, joinPath(destDir, file)))
       pathMap.set(sourcePath, relFiles)
       filesCopied += copied.length
@@ -85,7 +81,7 @@ export async function copyConfigKeyEntry(
       const uniqueDestPath = await findUniqueDestPath(destDir, basename(fullPath))
       await copyFile(fullPath, uniqueDestPath)
       const outputRelative = relativePath(outputDir, uniqueDestPath)
-      options.stdout.write(`Copied '${sourcePath}' to ${outputRelative}\n`)
+      options.stdout.write(`Included '${sourcePath}'\n`)
       pathMap.set(sourcePath, outputRelative)
       filesCopied += 1
     }
--- a/packages/app/src/cli/models/extensions/specification.ts
+++ b/packages/app/src/cli/models/extensions/specification.ts
@@ -302,6 +302,7 @@ export function createContractBasedModuleSpecification<TConfiguration extends Ba
     appModuleFeatures: spec.appModuleFeatures,
     experience: spec.experience,
     buildConfig: spec.buildConfig ?? {mode: 'none'},
+    clientSteps: spec.clientSteps,
     uidStrategy: spec.uidStrategy,
     transformRemoteToLocal: spec.transformRemoteToLocal,
     deployConfig: async (config, directory) => {
--- a/packages/app/src/cli/models/extensions/specifications/admin.ts
+++ b/packages/app/src/cli/models/extensions/specifications/admin.ts
@@ -24,7 +24,7 @@ const adminSpecificationSpec = createContractBasedModuleSpecification({
           name: 'Hosted App Copy Files',
           type: 'include_assets',
           config: {
-            generateManifest: true,
+            generatesAssetsManifest: true,
             inclusions: [
               {
                 type: 'configKey',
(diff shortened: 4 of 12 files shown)


alfonso-noriega:
* **#7016** <a href="https://app.graphite.com/github/pr/Shopify/cli/7016?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#7017](https://github.com/Shopify/cli/pull/7017) <a href="https://app.graphite.com/github/pr/Shopify/cli/7017?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#7134](https://github.com/Shopify/cli/pull/7134) <a href="https://app.graphite.com/github/pr/Shopify/cli/7134?utm_source=stack-comment-icon" target="_blank">Graphite</a>) 👈 <a href="https://app.graphite.com/github/pr/Shopify/cli/7016?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#6867** <a href="https://app.graphite.com/github/pr/Shopify/cli/6867?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 3 other dependent PRs ([#6868](https://github.com/Shopify/cli/pull/6868) <a href="https://app.graphite.com/github/pr/Shopify/cli/6868?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#6869](https://github.com/Shopify/cli/pull/6869) <a href="https://app.graphite.com/github/pr/Shopify/cli/6869?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#7027](https://github.com/Shopify/cli/pull/7027) <a href="https://app.graphite.com/github/pr/Shopify/cli/7027?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#6966** <a href="https://app.graphite.com/github/pr/Shopify/cli/6966?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#6941](https://github.com/Shopify/cli/pull/6941) <a href="https://app.graphite.com/github/pr/Shopify/cli/6941?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* `main`


<h2></h2>

packages/app/src/cli/services/build/steps/include-assets-step.ts
```diff
@@ -42,12 +51,17 @@ const StaticEntrySchema = z.object({
  * copies the directory contents into the output. Silently skipped when the
  * key is absent. Respects `preserveStructure` and `destination` the same way
  * as the static entry.
+ *
+ * `anchor` and `groupBy` are optional fields used for manifest generation.
+ * When both are present, this entry participates in `generateManifestFile`.
```
isaacroldan: I think that manifest generation should be more explicit, an extension might want to generate an assets manifest without an anchor or groupBy properties.  
Maybe add an optional: `generatesAssetsManifest: z.boolean().default(false)`?

packages/app/src/cli/services/build/steps/include-assets-step.ts
```diff
@@ -56,11 +70,34 @@ const InclusionEntrySchema = z.discriminatedUnion('type', [PatternEntrySchema, S
...
+  })
+  .superRefine((data, ctx) => {
+    for (const [i, entry] of data.inclusions.entries()) {
+      if (entry.type === 'configKey') {
+        const hasAnchor = entry.anchor !== undefined
+        const hasGroupBy = entry.groupBy !== undefined
+        if (hasAnchor !== hasGroupBy) {
+          ctx.addIssue({
+            code: z.ZodIssueCode.custom,
+            message: '`anchor` and `groupBy` must both be set or both be omitted',
```
isaacroldan, line 90: is this true though? if the anchor is ommited, you could treat it the "root" as anchor no?

isaacroldan: You should probably add some comment on what `anchor` does

packages/app/src/cli/services/build/steps/include-assets-step.ts
```diff
@@ -76,51 +117,82 @@ export async function executeIncludeAssetsStep(
...
+      {
+        key: entry.key,
+        baseDir: extension.directory,
+        outputDir,
+        context,
+        destination: sanitizedDest,
+      },
+      options,
+    )
+    result.pathMap.forEach((val, key) => aggregatedPathMap.set(key, val))
```
isaacroldan, line 153: 🐛 **Bug**: When two different configKey entries resolve to the same source path (e.g., `targeting.tools` → `./tools.json` and `settings.tools` → `./tools.json`), the second `copyConfigKeyEntry` call uses `findUniqueDestPath` to rename its copy (e.g., `tools-1.json`), then `result.pathMap.forEach((val, key) => aggregatedPathMap.set(key, val))` overwrites the first entry's mapping. Result: the manifest references `tools-1.json` for ALL occurrences of `./tools.json`, while the first copy at `tools.json` becomes an orphan.

**Suggestion:** Preserve the first (canonical) mapping by skipping already-aggregated keys:

```suggestion
    result.pathMap.forEach((val, key) => {
      if (!aggregatedPathMap.has(key)) {
        aggregatedPathMap.set(key, val)
      }
    })
```

packages/app/src/cli/services/build/steps/include-assets/copy-config-key-entry.ts
```diff
@@ -33,33 +42,75 @@ export async function copyConfigKeyEntry(
...
+      continue
+    }
+
+    const sourceIsDir = await isDirectory(fullPath)
+
+    const destDir = effectiveOutputDir
+
+    if (sourceIsDir) {
+      await copyDirectoryContents(fullPath, destDir)
+      const copied = await glob(['**/*'], {cwd: destDir, absolute: false})
```
isaacroldan, line 74: 🐛 **Bug**: When multiple directory sources are copied sequentially to the same `effectiveOutputDir`, the `glob(['**/*'], {cwd: destDir})` scans the entire destination — including files deposited by earlier iterations. This produces two problems: (1) `filesCopied` is inflated (the integration test already acknowledges this with `toBeGreaterThanOrEqual`), and (2) `pathMap.set(sourcePath, relFiles)` associates the wrong file list with this source path. Since `pathMap` is consumed by `generateManifestFile` to resolve config paths into output-relative paths, a corrupted pathMap can produce an incorrect manifest.

Example: entry A copies dir `{a.js}`, entry B copies dir `{b.js}` — glob after B returns `{a.js, b.js}`, so pathMap for source B contains both files.

**Suggestion:** Glob the source directory before copying to get the precise list of files contributed by this source:

```suggestion
      const sourceFiles = await glob(['**/*'], {cwd: fullPath, absolute: false})
      await copyDirectoryContents(fullPath, destDir)
      options.stdout.write(`Included '${sourcePath}'\n`)
      const relFiles = sourceFiles.map((file) => relativePath(outputDir, joinPath(destDir, file)))
      pathMap.set(sourcePath, relFiles)
      filesCopied += sourceFiles.length
```

packages/app/src/cli/services/build/steps/include-assets/generate-manifest.ts
isaacroldan, line 1: 💬 **Suggestion**: This module contains several non-trivial private functions (`stripAnchorPrefix`, `buildRelativeEntry`, `resolveManifestPaths`, `lastKeySegment`, `containsUnresolvedPath`, `shallowMerge`) tested only indirectly through integration tests. Other modules in this directory (`copy-by-pattern`, `copy-source-entry`, `copy-config-key-entry`) each have their own test files. Edge cases in the helpers (e.g., `stripAnchorPrefix` when key doesn't start with anchor prefix, `buildRelativeEntry` with deeply nested paths, `resolveManifestPaths` with mixed resolved/unresolved values) are harder to isolate when failures surface only as cryptic manifest content mismatches.

**Suggestion:** Consider adding a `generate-manifest.test.ts` with focused unit tests for the key helpers. Either export them (like was done for `tokenizePath` and `getNestedValue`) or test them through `generateManifestFile` with targeted inputs that exercise specific branches.

packages/app/src/cli/services/build/steps/include-assets/generate-manifest.ts
```diff
@@ -0,0 +1,227 @@
...
+  if (anchoredIncs.length === 0 && rootIncs.length === 0 && otherFiles.length === 0) return
+
+  const manifest: {[key: string]: unknown} = {}
+
+  // Root-level entries
+  for (const inc of rootIncs) {
+    const key = lastKeySegment(inc.key)
+    const rawValue = getNestedValue(extension.configuration, inc.key)
+    if (rawValue === null || rawValue === undefined) continue
+    manifest[key] = resolveManifestPaths(rawValue, pathMap)
```
isaacroldan, line 58: 🐛 **Bug**: When two root-level configKey entries have different `key` paths but the same `lastKeySegment` (e.g., `targeting.tools` and `settings.tools`), the second entry silently overwrites the first in the manifest. The anchored entry path already has a duplicate-key warning (lines 98-100), but root-level entries have no equivalent check. This inconsistency means data loss from key collisions goes undetected for root-level entries.

**Suggestion:** Add a duplicate key check before setting the root-level entry, mirroring the warning already present for anchored entries:

```suggestion
    if (Object.prototype.hasOwnProperty.call(manifest, key)) {
      options.stdout.write(`Warning: duplicate manifest key '${key}' — later entry overwrites earlier one\n`)
    }
    manifest[key] = resolveManifestPaths(rawValue, pathMap)
```

packages/app/src/cli/services/build/steps/include-assets/generate-manifest.ts
```diff
@@ -0,0 +1,227 @@
...
+
+      if (Object.prototype.hasOwnProperty.call(manifest, manifestKey)) {
+        options.stdout.write(`Warning: duplicate manifest key '${manifestKey}' — later entry overwrites earlier one\n`)
+      }
+      manifest[manifestKey] = shallowMerge(partials)
+    }
+  }
+
+  if (otherFiles.length > 0) {
+    manifest.files = otherFiles
```
isaacroldan, line 106: 🐛 **Bug**: If a root-level configKey entry resolves to manifest key `"files"` (via `lastKeySegment` — e.g., key `"build.files"`), line 58 writes `manifest['files'] = ...`, then line 106 unconditionally overwrites it with `manifest.files = otherFiles`. Unlike the anchored entry duplicate check at lines 98-100, no collision warning is emitted here. This is the same class of issue as missing root-level duplicate detection, but specifically for the reserved `"files"` key.

**Suggestion:** Check for the collision and warn:

```suggestion
  if (otherFiles.length > 0) {
    if (Object.prototype.hasOwnProperty.call(manifest, 'files')) {
      options.stdout.write(`Warning: manifest key 'files' from configKey entry will be overwritten by pattern/static file paths\n`)
    }
    manifest.files = otherFiles
  }
```

packages/app/src/cli/services/build/steps/include-assets/generate-manifest.ts
```diff
@@ -0,0 +1,227 @@
...
+  const last = key.split('.').at(-1) ?? key
+  return last.endsWith('[]') ? last.slice(0, -2) : last
+}
+
+/**
+ * Returns `true` if `value` contains any string — at any depth — that looks
+ * like an unresolved config path (starts with `./` or `../`).
+ */
+function containsUnresolvedPath(value: unknown): boolean {
+  if (typeof value === 'string') return value.startsWith('./') || value.startsWith('../')
```
isaacroldan, line 221: 🐛 **Bug**: `containsUnresolvedPath` only detects strings starting with `./` or `../`. Bare filename config values (e.g., `tools.json` without a `./` prefix) whose source file was missing will appear in the manifest without any warning.

Trace: config has `tools: 'tools.json'`, source file doesn't exist → `copyConfigKeyEntry` skips it → `pathMap` has no entry → `resolveManifestPaths` sees `looksLikePath = false` (no `.`/`/`/`\` prefix and not in pathMap) → returns `'tools.json'` as-is → `containsUnresolvedPath('tools.json')` returns `false` → no warning. Contrast with `'./tools.json'` which is correctly caught.

**Suggestion:** Consider broadening the detection or, more robustly, comparing pre- and post-resolution values to detect when resolution was a no-op for values that were expected to be paths. A targeted approach would be to pass `pathMap` into this function and flag any leaf string that exists as a key in the config but not in the pathMap.

PR #7022, opened 2026-03-16
Add Project domain model for filesystem abstraction

ryancbahan:

The CLI currently discovers filesystem state (config files, extensions, webs, dotenv, hidden config, dependencies) across many scattered code paths — the loader, app-context, link service, and individual commands all independently scan the filesystem. This makes it hard to reason about what files the system knows about, leads to redundant disk reads, and tightly couples Shopify domain logic to OS/filesystem concerns (the latter of which is the highest concern as we start to consider isomorphic CLI operations.)

Introducing a `Project` model separates these concerns. Project is a pure filesystem abstraction — it discovers what's on disk and provides access to it. It doesn't interpret configs as modules, select which config is active, or know about the Shopify platform. This separation is a foundation for making the CLI codebase more isomorphic by abstracting away OS/filesystem operations behind a clean interface.

Adds three new modules under `models/project/`:

- **`Project`** — discovers and holds all filesystem state for a Shopify app project. Finds the project root by walking up from the working directory, discovers all `shopify.app*.toml` files (supporting multi-config projects), extension TOMLs, web TOMLs, dotenv files, hidden config (`.shopify/project.json`), and package metadata. Single scan, no duplicate reads.
- **`ActiveConfig`** — represents the selected app configuration (one of potentially many in the project). Derived from Project via `selectActiveConfig()`, which resolves flag > cache > default, handles stale cache, and derives config-specific dotenv and hidden config. A sibling to Project, never nested inside it.
- **Config selection functions** — `resolveDotEnv`, `resolveHiddenConfig`, `extensionFilesForConfig`, `webFilesForConfig`. Pure functions that derive config-specific state from the Project's raw data.

### How to test your changes?

```
npx vitest run src/cli/models/project/
```

### Measuring impact

- [x] n/a

### Checklist

### DIFF
--- a/packages/app/src/cli/models/project/config-selection.ts
+++ b/packages/app/src/cli/models/project/config-selection.ts
@@ -86,7 +86,7 @@ export function extensionFilesForConfig(project: Project, activeConfig: TomlFile
   // This is a simplification — complex globs like "foo/*/bar" will over-match.
   // In practice, only simple directory patterns are used in app configs.
   const dirPrefixes = (configDirs as string[]).map((dir) => {
-    return dir.replace(/\*.*$/, '')
+    return dir.replace(/\*.*$/, '').replace(/\/?$/, '/')
   })
 
   return project.extensionConfigFiles.filter((file) => {
@@ -106,7 +106,7 @@ export function webFilesForConfig(project: Project, activeConfig: TomlFile): Tom
     return project.webConfigFiles
   }
 
-  const dirPrefixes = (configDirs as string[]).map((dir) => dir.replace(/\*.*$/, ''))
+  const dirPrefixes = (configDirs as string[]).map((dir) => dir.replace(/\*.*$/, '').replace(/\/?$/, '/'))
 
   return project.webConfigFiles.filter((file) => {
     const relPath = relativePath(project.directory, file.path).replace(/\\/g, '/')


ryancbahan:
* **#7035** <a href="https://app.graphite.com/github/pr/Shopify/cli/7035?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#7023** <a href="https://app.graphite.com/github/pr/Shopify/cli/7023?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#7022** <a href="https://app.graphite.com/github/pr/Shopify/cli/7022?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Shopify/cli/7022?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/app/src/cli/models/project/active-config.ts
```diff
@@ -0,0 +1,124 @@
...
+ * If the cached config file no longer exists on disk, prompts the user
+ * to select a new one via `app config use`.
+ *
+ * Derives config-specific state: dotenv and hidden config for the selected
+ * config's client_id.
+ * @public
+ */
+export async function selectActiveConfig(project: Project, userProvidedConfigName?: string): Promise<ActiveConfig> {
+  let configName = userProvidedConfigName
+  const source: ConfigSource = configName ? 'flag' : 'cached'
```
dmerand: 🐛 **Bug**: When no `userProvidedConfigName` is given and no cached config exists, the code falls through to use the default config. However, the `source` is set as 'cached' on line 56 even though it should indicate 'default'. This means the source will incorrectly report 'cached' for the default fallback case, which could confuse debugging and telemetry.

**Suggestion:** Consider determining the source after the actual config selection is finalized, or expand the ConfigSource type to include 'default':

```typescript
export type ConfigSource = 'flag' | 'cached' | 'default'
```

Then update the logic:
```typescript
  let source: ConfigSource
  if (userProvidedConfigName) {
    source = 'flag'
  } else if (cachedConfigName) {
    source = 'cached'
  } else {
    source = 'default'
  }
```

packages/app/src/cli/models/project/config-selection.ts
```diff
@@ -0,0 +1,112 @@
...
+ * Falls back to the default .env if the config-specific one doesn't exist.
+ * @public
+ */
+export function resolveDotEnv(project: Project, activeConfigPath: string): DotEnvFile | undefined {
+  const shorthand = getAppConfigurationShorthand(activeConfigPath)
+  if (shorthand) {
+    const specificName = `${dotEnvFileNames.production}.${shorthand}`
+    const specific = project.dotenvFiles.get(specificName)
+    if (specific) return specific
+  }
```
dmerand, line 26: `resolveDotEnv()` changes the existing selection semantics for non-default configs. Historically `loader.ts` loaded the config-specific dotenv file only (`shopify.app.staging.toml` → `.env.staging`) and did not fall back to `.env`; there is prior history here from the earlier dotenv-selection bugfix work.

With this implementation we now try `.env.<shorthand>` and then silently fall back to `.env`. That is not behavior-neutral: a non-default config can now inherit base env values it previously would not see, which can mask a missing config-specific dotenv file or leak values across configs.

I think we should either preserve the old semantics here, or call this out as an intentional behavior change and add regression tests proving the fallback is desired.

packages/app/src/cli/models/project/config-selection.ts
```diff
@@ -0,0 +1,112 @@
...
+ * extension_directories. If the active config doesn't specify extension_directories,
+ * uses the default (extensions/*).
+ * @public
+ */
+export function extensionFilesForConfig(project: Project, activeConfig: TomlFile): TomlFile[] {
+  const configDirs = activeConfig.content.extension_directories
+  if (!Array.isArray(configDirs) || configDirs.length === 0) {
+    // Default: extensions/* — filter project files by path prefix
+    return project.extensionConfigFiles.filter((file) => {
+      const relPath = file.path.replace(project.directory, '').replace(/^\//, '')
```
dmerand: 🐛 **Bug**: The path manipulation uses string replacement with hardcoded forward slashes which will fail on Windows. The pattern `.replace(project.directory, '').replace(/^\//, '')` only removes leading forward slashes, not backslashes used by Windows. This same pattern appears on lines 78, 90, and 109. Your existing codebase in loader.ts uses the `relativePath` utility from `@shopify/cli-kit/node/path` for cross-platform compatibility.

**Suggestion:** 
```suggestion
      const relPath = relativePath(project.directory, file.path)
```

You'll also need to add the import:
```typescript
import {relativePath} from '@shopify/cli-kit/node/path'
```

packages/app/src/cli/models/project/config-selection.ts
```diff
@@ -0,0 +1,112 @@
...
+  const configDirs = activeConfig.content.web_directories
+  if (!Array.isArray(configDirs) || configDirs.length === 0) {
+    return project.webConfigFiles
+  }
+
+  const dirPrefixes = (configDirs as string[]).map((dir) => dir.replace(/\*.*$/, ''))
+
+  return project.webConfigFiles.filter((file) => {
+    const relPath = file.path.replace(project.directory, '').replace(/^\//, '')
+    return dirPrefixes.some((prefix) => relPath.startsWith(prefix))
```
dmerand: The old loader used the configured `extension_directories` / `web_directories` globs directly. Here we reduce each configured pattern to a simple prefix by stripping from the first `*` and then doing `startsWith(...)`. That is a semantic weakening: more complex globs can now over-match.

For example, a pattern shaped like `foo/*/extensions` effectively becomes `foo/`, so files outside the intended glob but under that prefix will still be treated as belonging to the config. That makes this refactor less obviously behavior-preserving than it first appears.

I think we should add tests for nontrivial glob shapes before treating this as a safe equivalence refactor. If only simple prefix-like patterns are meant to be supported, it would be better to codify that in the schema/docs rather than silently weakening the matching behavior here.

packages/app/src/cli/models/project/config-selection.ts
```diff
@@ -0,0 +1,112 @@
...
+  }
+
+  // Filter to files that are within the active config's declared directories
+  const dirPrefixes = (configDirs as string[]).map((dir) => {
+    // Remove trailing glob (e.g., "custom/*" → "custom/")
+    return dir.replace(/\*.*$/, '')
+  })
+
+  return project.extensionConfigFiles.filter((file) => {
+    const relPath = file.path.replace(project.directory, '').replace(/^\//, '')
```
dmerand: 🐛 **Bug**: Same hardcoded forward slash issue as line 78. This will cause extension filtering to fail on Windows when using custom `extension_directories`.

**Suggestion:** 
```suggestion
    const relPath = relativePath(project.directory, file.path)
```

packages/app/src/cli/models/project/config-selection.ts
```diff
@@ -0,0 +1,112 @@
...
+export function webFilesForConfig(project: Project, activeConfig: TomlFile): TomlFile[] {
+  const configDirs = activeConfig.content.web_directories
+  if (!Array.isArray(configDirs) || configDirs.length === 0) {
+    return project.webConfigFiles
+  }
+
+  const dirPrefixes = (configDirs as string[]).map((dir) => dir.replace(/\*.*$/, ''))
+
+  return project.webConfigFiles.filter((file) => {
+    const relPath = file.path.replace(project.directory, '').replace(/^\//, '')
```
dmerand: 🐛 **Bug**: Same hardcoded forward slash issue in web config filtering. Consider extracting this to a helper function since the pattern appears three times.

**Suggestion:** 
```suggestion
    const relPath = relativePath(project.directory, file.path)
```

packages/app/src/cli/models/project/project.ts
```diff
@@ -0,0 +1,240 @@
...
+    const dotenvFiles = await discoverDotEnvFiles(directory)
+
+    // Hidden config: store the raw .shopify/project.json content
+    const hiddenConfigRaw = await loadRawHiddenConfig(directory)
+
+    return new Project({
+      directory,
+      packageManager,
+      nodeDependencies,
+      usesWorkspaces,
```
dmerand, line 93: The PR description says `Project.load()` replaces the old repeated scans, but the current load path still appears to do another project-wide config discovery pass afterwards. `loadAppConfigurationFromState()` still calls `getAllLinkedConfigClientIds(...)`, and that helper re-globs/parses config files instead of reusing the already-loaded `Project` data.

So this is not yet a strict "one scan replaces 6+ scans" path in practice: the new project-wide discovery happens here, and then we still pay for at least one extra config scan for metadata. If the performance reduction is a core claim, I think we should either thread `project` through `loadAppConfigurationFromState()` / `getAllLinkedConfigClientIds()`, or soften the claim until the full call path actually reuses the same discovery result end-to-end.

If you want to keep the claim, it would be good to benchmark the real before/after load path rather than asserting a single-scan architecture that the current implementation does not fully achieve yet.

packages/app/src/cli/models/project/project.ts
```diff
@@ -0,0 +1,240 @@
...
+      }
+    }
+    const extensionConfigFiles = await discoverExtensionFiles(directory, [...allExtensionDirs])
+
+    // Discover web files from all app configs' web_directories (union)
+    const allWebDirs = new Set<string>()
+    for (const appConfig of appConfigFiles) {
+      const dirs = appConfig.content.web_directories
+      if (Array.isArray(dirs)) {
+        for (const dir of dirs) allWebDirs.add(dir as string)
```
dmerand, line 71: `Project.load()` now eagerly parses every `shopify.app*.toml`, plus every extension/web TOML reachable from the union of all configs' `extension_directories` / `web_directories`, before active-config selection happens. That is a behavior change from `loader.ts`, which historically treated the selected config as the unit of loading: it only parsed the chosen app config and only globbed module TOMLs from that config's directories, and even `getAllLinkedConfigClientIds` explicitly ignored parse errors in sibling configs.

The regression is that a malformed or half-migrated *inactive* config/file can now prevent the active config from loading. Concretely, an unrelated broken `shopify.app.other.toml`, inactive extension TOML, or inactive web TOML will now fail `Project.load()` before `selectActiveConfig()` gets a chance to filter.

Can we keep discovery shallow for non-active files, or defer TOML parsing until after active-config selection? At minimum, perhaps add regression coverage proving the active config still loads when an inactive app/extension/web TOML is malformed.

packages/app/src/cli/models/project/project.ts
ryancbahan: This is addressed in followup pr.

PR #7023, opened 2026-03-16
Integrate Project model into loader and decompose into composable stages

ryancbahan:

PR #7022 introduced the `Project` and `ActiveConfig` domain models, but nothing consumed them  (the loader still ran its own filesystem discovery, constructed `AppConfigurationState` as an intermediate, and threaded the entire previous `AppInterface` through reloads). This PR wires the new models into the loading pipeline and uses that as leverage to decompose the monolithic `loadApp` into composable stages with narrow interfaces.

**Decomposes the loader into three explicit stages:**

- `getAppConfigurationContext(dir, configName)` → `{project, activeConfig}` — discovery only, no parsing or state construction
- `loadAppFromContext({project, activeConfig, specifications, …})` → `AppInterface` — validation and assembly, for callers that already hold a Project
- `loadApp({directory, configName, …})` — thin wrapper composing the two above

**[Interim state] Replaces** **`previousApp`** **with narrow** **`ReloadState`:**

Instead of threading the entire `AppLinkedInterface` through reloads (just to read 2 fields), `reloadApp` now constructs a `ReloadState` with only `extensionDevUUIDs: Map<string, string>` and `previousDevURLs`. This is a step change away from broad state passing/mutation, but needs more consideration on "permanent home" as we continue to decompose functionality. 

**Updates all consumers:**

- `linkedAppContext` uses `activeConfig.isLinked` and `activeConfig.file.content` directly instead of going through `AppConfigurationState`
- `link()` returns `{remoteApp, configFileName, configuration}` — drops `state` from its return type
- `use()` reads `activeConfig.file.content` instead of calling `loadAppConfiguration`
- `loadConfigForAppCreation` uses `activeConfig` and `project.directory` directly

**Removes dead code (-400 lines):**

`AppConfigurationState`, `AppConfigurationStateBasics`, `toAppConfigurationState`, `loadAppConfigurationFromState`, `loadAppUsingConfigurationState`, `loadAppConfiguration`, `getAppConfigurationState`, `getAppDirectory`, `loadDotEnv`, `loadHiddenConfig`, `findWebConfigPaths`, `loadWebsForAppCreation`, `getConfigurationPath` (de-exported)

### How to test your changes?

```bash
npx vitest run packages/app/src/cli/models/app/loader.test.ts
npx vitest run packages/app/src/cli/services/app-context.test.ts
npx vitest run packages/app/src/cli/services/app/config/link.test.ts
npx vitest run packages/app/src/cli/services/app/config/use.test.ts
npx vitest run packages/app/src/cli/services/context.test.ts
npx vitest run packages/app/src/cli/models/project/
```

### Measuring impact

- [x] n/a

### Checklist

### DIFF
--- a/packages/app/src/cli/models/app/config-file-naming.ts
+++ b/packages/app/src/cli/models/app/config-file-naming.ts
@@ -1,0 +1,42 @@
+import {configurationFileNames} from '../../constants.js'
+import {slugify} from '@shopify/cli-kit/common/string'
+import {basename} from '@shopify/cli-kit/node/path'
+
+const appConfigurationFileNameRegex = /^shopify\.app(\.[-\w]+)?\.toml$/
+export type AppConfigurationFileName = 'shopify.app.toml' | `shopify.app.${string}.toml`
+
+/**
+ * Gets the name of the app configuration file (e.g. `shopify.app.production.toml`) based on a provided config name.
+ *
+ * @param configName - Optional config name to base the file name upon
+ * @returns Either the default app configuration file name (`shopify.app.toml`), the given config name (if it matched the valid format), or `shopify.app.<config name>.toml` if it was an arbitrary string
+ */
+export function getAppConfigurationFileName(configName?: string): AppConfigurationFileName {
+  if (!configName) {
+    return configurationFileNames.app
+  }
+
+  if (isValidFormatAppConfigurationFileName(configName)) {
+    return configName
+  } else {
+    return `shopify.app.${slugify(configName)}.toml`
+  }
+}
+
+/**
+ * Given a path to an app configuration file, extract the shorthand section from the file name.
+ *
+ * This is undefined for `shopify.app.toml` files, or returns e.g. `production` for `shopify.app.production.toml`.
+ */
+export function getAppConfigurationShorthand(path: string) {
+  const match = basename(path).match(appConfigurationFileNameRegex)
+  return match?.[1]?.slice(1)
+}
+
+/** Checks if configName is a valid one (`shopify.app.toml`, or `shopify.app.<something>.toml`) */
+export function isValidFormatAppConfigurationFileName(configName: string): configName is AppConfigurationFileName {
+  if (appConfigurationFileNameRegex.test(configName)) {
+    return true
+  }
+  return false
+}
--- a/packages/app/src/cli/models/app/loader.ts
+++ b/packages/app/src/cli/models/app/loader.ts
@@ -6,58 +6,52 @@ import {
   WebType,
   getAppScopesArray,
   AppConfigurationInterface,
-  AppConfiguration,
   CurrentAppConfiguration,
   getAppVersionedSchema,
   AppSchema,
-  BasicAppConfigurationWithoutModules,
   SchemaForConfig,
   AppLinkedInterface,
-  AppHiddenConfig,
 } from './app.js'
 import {parseHumanReadableError} from './error-parsing.js'
+import {
+  getAppConfigurationFileName,
+  getAppConfigurationShorthand,
+  type AppConfigurationFileName,
+} from './config-file-naming.js'
 import {configurationFileNames, dotEnvFileNames} from '../../constants.js'
 import metadata from '../../metadata.js'
 import {ExtensionInstance} from '../extensions/extension-instance.js'
 import {ExtensionsArraySchema, UnifiedSchema} from '../extensions/schemas.js'
-import {
-  ExtensionSpecification,
-  RemoteAwareExtensionSpecification,
-  isAppConfigSpecification,
-} from '../extensions/specification.js'
-import {getCachedAppInfo} from '../../services/local-storage.js'
-import use from '../../services/app/config/use.js'
+import {ExtensionSpecification, isAppConfigSpecification} from '../extensions/specification.js'
 import {CreateAppOptions, Flag} from '../../utilities/developer-platform-client.js'
 import {findConfigFiles} from '../../prompts/config.js'
 import {WebhookSubscriptionSpecIdentifier} from '../extensions/specifications/app_config_webhook_subscription.js'
 import {WebhooksSchema} from '../extensions/specifications/app_config_webhook_schemas/webhooks_schema.js'
-import {loadLocalExtensionsSpecifications} from '../extensions/load-specifications.js'
-import {patchAppHiddenConfigFile} from '../../services/app/patch-app-configuration-file.js'
-import {getOrCreateAppConfigHiddenPath} from '../../utilities/app/config/hidden-app-config.js'
 import {ApplicationURLs, generateApplicationURLs} from '../../services/dev/urls.js'
+import {Project} from '../project/project.js'
+import {selectActiveConfig} from '../project/active-config.js'
+import {
+  resolveDotEnv,
+  resolveHiddenConfig,
+  extensionFilesForConfig,
+  webFilesForConfig,
+} from '../project/config-selection.js'
 import {showMultipleCLIWarningIfNeeded} from '@shopify/cli-kit/node/multiple-installation-warning'
-import {fileExists, readFile, glob, findPathUp, fileExistsSync} from '@shopify/cli-kit/node/fs'
+import {fileExists, readFile, fileExistsSync} from '@shopify/cli-kit/node/fs'
 import {TomlFile, TomlParseError} from '@shopify/cli-kit/node/toml/toml-file'
 import {zod} from '@shopify/cli-kit/node/schema'
-import {readAndParseDotEnv, DotEnvFile} from '@shopify/cli-kit/node/dot-env'
-import {
-  getDependencies,
-  getPackageManager,
-  PackageManager,
-  usesWorkspaces as appUsesWorkspaces,
-} from '@shopify/cli-kit/node/node-package-manager'
+import {PackageManager} from '@shopify/cli-kit/node/node-package-manager'
 import {resolveFramework} from '@shopify/cli-kit/node/framework'
 import {hashString} from '@shopify/cli-kit/node/crypto'
 import {JsonMapType} from '@shopify/cli-kit/node/toml'
 import {joinPath, dirname, basename, relativePath, relativizePath} from '@shopify/cli-kit/node/path'
 import {AbortError} from '@shopify/cli-kit/node/error'
 import {outputContent, outputDebug, OutputMessage, outputToken} from '@shopify/cli-kit/node/output'
-import {joinWithAnd, slugify} from '@shopify/cli-kit/common/string'
+import {joinWithAnd} from '@shopify/cli-kit/common/string'
 import {getArrayRejectingUndefined} from '@shopify/cli-kit/common/array'
 import {showNotificationsIfNeeded} from '@shopify/cli-kit/node/notifications-system'
 import ignore from 'ignore'
-
-const defaultExtensionDirectory = 'extensions/*'
+import type {ActiveConfig} from '../project/active-config.js'
 
 /**
  * The mode in which the app is loaded, this affects how errors are handled:
@@ -67,13 +61,24 @@ const defaultExtensionDirectory = 'extensions/*'
  */
 export type AppLoaderMode = 'strict' | 'report' | 'local'
 
+/**
+ * Narrow runtime state carried forward across app reloads.
+ *
+ * Replaces passing the entire previous AppInterface — only genuine runtime
+ * state (devUUIDs and tunnel URLs) needs to survive a reload.
+ */
+export interface ReloadState {
+  /** Extension handle → devUUID, preserved for dev-console stability across reloads */
+  extensionDevUUIDs: Map<string, string>
+  /** Previous dev tunnel URL, kept stable across reloads */
+  previousDevURLs?: ApplicationURLs
+}
+
 type AbortOrReport = <T>(errorMessage: OutputMessage, fallback: T, configurationPath: string) => T
 
 const abort: AbortOrReport = (errorMessage) => {
   throw new AbortError(errorMessage)
 }
-
-const noopAbortOrReport: AbortOrReport = (_errorMessage, fallback, _configurationPath) => fallback
 
 /**
  * Loads a configuration file, and returns its content as an unvalidated object.
@@ -196,8 +201,10 @@ interface AppLoaderConstructorArgs<
 > {
   mode?: AppLoaderMode
   loadedConfiguration: ConfigurationLoaderResult<TConfig, TModuleSpec>
-  // Used when reloading an app, to avoid some expensive steps during loading.
-  previousApp?: AppLinkedInterface
+  // Pre-discovered project data — avoids re-scanning the filesystem for dependencies, package manager, etc.
+  project: Project
+  // Narrow runtime state from a previous app load, used during reloads
+  reloadState?: ReloadState
 }
 
 export async function checkFolderIsValidApp(directory: string) {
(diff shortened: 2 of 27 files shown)


ryancbahan:
* **#7035** <a href="https://app.graphite.com/github/pr/Shopify/cli/7035?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#7023** <a href="https://app.graphite.com/github/pr/Shopify/cli/7023?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Shopify/cli/7023?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#7022** <a href="https://app.graphite.com/github/pr/Shopify/cli/7022?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

packages/app/src/cli/models/app/loader.ts
```diff
@@ -67,14 +56,25 @@ const defaultExtensionDirectory = 'extensions/*'
  */
 export type AppLoaderMode = 'strict' | 'report' | 'local'
 
+/**
+ * Narrow runtime state carried forward across app reloads.
+ *
+ * Replaces passing the entire previous AppInterface — only genuine runtime
+ * state (devUUIDs and tunnel URLs) needs to survive a reload.
+ */
+export interface ReloadState {
```
isaacroldan, line 70: ⭐ **Praise**: Replacing the entire `AppLinkedInterface` passthrough with `ReloadState { extensionDevUUIDs, previousDevURLs }` is a significant improvement. It makes the reload contract explicit — only two pieces of runtime state need to survive a reload. The `Map<string, string>` for devUUIDs is more appropriate than extracting from an array on each access. This eliminates a class of bugs where reloads accidentally depended on stale values from the previous app instance.

packages/app/src/cli/models/app/loader.ts
```diff
@@ -624,16 +668,18 @@ class AppLoader<TConfig extends CurrentAppConfiguration, TModuleSpec extends Ext
     return allExtensions
   }
 
-  private async createExtensionInstances(appDirectory: string, extensionDirectories?: string[]) {
-    const extensionConfigPaths = [...(extensionDirectories ?? [defaultExtensionDirectory])].map((extensionPath) => {
-      return joinPath(appDirectory, extensionPath, '*.extension.toml')
-    })
-    extensionConfigPaths.push(`!${joinPath(appDirectory, '**/node_modules/**')}`)
-    const configPaths = await glob(extensionConfigPaths)
+  private async createExtensionInstances(appDirectory: string, _extensionDirectories?: string[]) {
```
isaacroldan: 💡 **Improvement**: `createExtensionInstances` takes `_extensionDirectories` (underscore-prefixed to suppress unused-var linting) but completely ignores it. Extension discovery comes from `this.activeConfigFile` and `extensionFilesForConfig`, which reads `extension_directories` from `activeConfig.content` internally. The parameter is passed from the call site at line 641 as `appConfiguration.extension_directories`, but it's never consumed.

**Suggestion:** Remove the `_extensionDirectories` parameter from the method signature and the corresponding argument from the call site at line 641.

packages/app/src/cli/models/app/loader.ts
```diff
@@ -624,16 +668,18 @@ class AppLoader<TConfig extends CurrentAppConfiguration, TModuleSpec extends Ext
...
+    const activeConfig = this.activeConfigFile
+    const extensionFiles = activeConfig
+      ? extensionFilesForConfig(this.project, activeConfig)
+      : this.project.extensionConfigFiles
+    const configPaths = extensionFiles.map((file) => file.path)
 
     return configPaths.map(async (configurationPath) => {
       const directory = dirname(configurationPath)
-      const obj = await loadConfigurationFileContent(configurationPath)
+      const projectFile = extensionFiles.find((file) => file.path === configurationPath)
```
isaacroldan: 💡 **Improvement**: `configPaths` at line 677 is derived from `extensionFiles.map(file => file.path)`, then `projectFile` at line 681 does `.find(file => file.path === configurationPath)` over the same `extensionFiles` array. Since `configurationPath` came from that same array, `projectFile` is guaranteed to be found. The fallback `?? (await loadConfigurationFileContent(configurationPath))` is dead code that misleadingly suggests a filesystem read could happen.

**Suggestion:** Iterate `extensionFiles` directly instead of round-tripping through `configPaths`:
```suggestion
      const obj = extensionFiles.find((file) => file.path === configurationPath)!.content
```
Or restructure the loop to use `extensionFiles.map` and access `file.content` directly.

packages/app/src/cli/models/project/active-config.ts
```diff
@@ -1,13 +1,15 @@
 import {Project} from './project.js'
 import {resolveDotEnv, resolveHiddenConfig} from './config-selection.js'
-import {AppHiddenConfig, BasicAppConfigurationWithoutModules} from '../app/app.js'
-import {AppConfigurationFileName, AppConfigurationState, getConfigurationPath} from '../app/loader.js'
+import {AppHiddenConfig} from '../app/app.js'
+import {getAppConfigurationFileName} from '../app/loader.js'
```
isaacroldan: 📐 **Design**: `active-config.ts` imports `getAppConfigurationFileName` from `loader.js` (line 4), and `loader.ts` imports `selectActiveConfig` from `active-config.js`. Additionally, `active-config.ts` imports `use.js` (line 6) which imports `getAppConfigurationContext` from `loader.js`, creating a deeper cycle: `loader → active-config → use → loader`. Node ESM handles this without crashing at startup thanks to live bindings, but it's fragile — reordering imports or adding a top-level side effect in any of these files can produce `undefined` at import time. It also makes isolated testing harder.

**Suggestion:** Extract `getAppConfigurationFileName` into a small, leaf-level utility module (e.g., `config-naming.ts`) that both `loader.ts` and `active-config.ts` can import without forming a cycle. This function is a pure string utility with no dependencies on loader internals.

packages/app/src/cli/services/app/config/use.ts
```diff
@@ -47,11 +47,8 @@ export default async function use({
 
   const configFileName = (await getConfigFileName(directory, configName)).valueOrAbort()
 
-  const {configuration} = await loadAppConfiguration({
-    userProvidedConfigName: configFileName,
-    directory,
-  })
-  setCurrentConfigPreference(configuration, {configFileName, directory})
+  const {activeConfig} = await getAppConfigurationContext(directory, configFileName)
+  setCurrentConfigPreference(activeConfig.file.content as AppConfiguration, {configFileName, directory})
```
isaacroldan: Human comment: Are we planning to validate the active config content using the modules schemas?
AI:
🐛 **Bug**: The old code called `loadAppConfiguration` which parsed and validated the config against `AppSchema` before passing it to `setCurrentConfigPreference`. The new code reads raw TOML via `getAppConfigurationContext` and casts it with `as AppConfiguration`, bypassing all schema validation. A malformed config file (invalid field types, missing nested structure, wrong shapes) will now be silently accepted as the current preference. Previously, such files would produce a validation error. This is a behavioral regression — the `app config use` command no longer validates configs before adopting them.

**Suggestion:** Either validate the content against AppSchema before setting the preference, or narrow the type to only what `setCurrentConfigPreference` actually needs. If full validation isn't feasible here, at minimum validate the fields that `setCurrentConfigPreference` relies on (like `client_id`).

packages/app/src/cli/models/app/loader.ts
ryancbahan: This is implemented in the next pr in stack `03-17-remove-project-fields-from-app`

packages/app/src/cli/services/app/config/use.ts
ryancbahan: I believe this is the correct behavior. "Use" is just switching active TOML -- it should only validate whether the toml is syntactically correct (which it does). It's not running anything, and therefore should not concern itself with schema validation. Other commands handle that.

packages/app/src/cli/models/project/active-config.ts
ryancbahan: will extract.

packages/app/src/cli/models/app/loader.ts
ryancbahan: good point. will address.

packages/app/src/cli/services/app/config/use.ts
isaacroldan: So the validation would happen on the next command run?   
The casting is weird though, because you can't be sure that content conforms to `AppConfiguration` :/

packages/app/src/cli/models/app/config-file-naming.ts
isaacroldan, line 1: Let's add tests for this file :)

packages/app/src/cli/models/app/loader.ts
```diff
@@ -196,8 +201,10 @@ interface AppLoaderConstructorArgs<
 > {
   mode?: AppLoaderMode
   loadedConfiguration: ConfigurationLoaderResult<TConfig, TModuleSpec>
-  // Used when reloading an app, to avoid some expensive steps during loading.
-  previousApp?: AppLinkedInterface
+  // Pre-discovered project data — avoids re-scanning the filesystem for dependencies, package manager, etc.
+  project: Project
```
isaacroldan, line 205: Umm, during dev you might add a new extension, that forces an app reload. But this `project` won't have the new files for the new extension right? :thinking_face:

isaacroldan: Yeah I confirmed this:  
`app dev` appears to work when generating an extension mid-dev, because the watcher detects the changes. And it triggers the `reloadApp` so that the in-memory App representation is updated. But Since this reload is not re-scanning the file-system is not loading the new extensions and they are not included in the deployed manifest.  
  
We need to re-scan when reloading the app :)

ryancbahan:
Pulling context from Slack -- we discovered the reload issue with manifest writes is already on main, not introduced in this pr. I've added more tests here and will tophat thoroughly + document.

packages/app/src/cli/models/app/loader.ts
ryancbahan: Context from Slack -- this is a pre-existing issue on main branch. Will make sure it's fixed here though.

packages/app/src/cli/services/app/config/use.ts
ryancbahan: that's fair re: casting

packages/app/src/cli/models/app/loader.ts
ryancbahan: The reload is called as expected in this pr. I added a test to demonstrate this

packages/app/src/cli/services/app/config/use.ts
ryancbahan: That's fair, though I'm not sure it's rewriting at the moment since it's just passthrough and any "real" command will catch the errors.

packages/app/src/cli/services/function/ui/components/Replay/hooks/useFunctionWatcher.test.tsx
```diff
@@ -65,6 +65,9 @@ const SECOND_EXEC_RESPONSE = {
 describe('useFunctionWatcher', () => {
   beforeEach(() => {
     vi.useFakeTimers()
+    // Prevent real chokidar/fsevents watchers from being created in tests.
+    // The hook only needs the AppEventWatcher as an EventEmitter, not real file watching.
+    vi.spyOn(AppEventWatcher.prototype, 'start').mockResolvedValue()
```
ryancbahan, line 70: this was hogging up memory now that tests are sharded

ryancbahan:
Added a bunch more tests around hot reload, file watching, and config switching. Doing a bunch of local tophatting and good so far.

packages/app/src/cli/models/app/loader.ts
```diff
@@ -492,12 +529,19 @@ class AppLoader<TConfig extends CurrentAppConfiguration, TModuleSpec extends Ext
   }
 
   async loadWebs(appDirectory: string, webDirectories?: string[]): Promise<{webs: Web[]; usedCustomLayout: boolean}> {
-    const webTomlPaths = await findWebConfigPaths(appDirectory, webDirectories)
-    const webs = await Promise.all(webTomlPaths.map((path) => loadSingleWeb(path, this.abortOrReport.bind(this))))
+    const activeConfig = this.activeConfigFile
+    const webFiles = activeConfig ? webFilesForConfig(this.project, activeConfig) : this.project.webConfigFiles
+    const webTomlPaths = webFiles.map((file) => file.path)
+    const webs = await Promise.all(
+      webFiles.map((webFile) => loadSingleWeb(webFile.path, this.abortOrReport.bind(this), webFile.content)),
```
dmerand, line 536: `extensionFilesForConfig()` / `webFilesForConfig()` intentionally reduce configured globs to simple path prefixes, so patterns like `foo/*/extensions` over-match. This PR is where that simplification becomes production loader behavior: `loadConfigForAppCreation()`, `loadWebs()`, and `createExtensionInstances()` now route through those helpers instead of the old direct glob discovery. The new integration tests cover rescans and reloads, but they do not protect nontrivial glob shapes. Can we either preserve true glob semantics here or add regression coverage for complex `extension_directories` and `web_directories` patterns before merging?

packages/app/src/cli/services/app/config/use.test.ts
```diff
@@ -120,7 +121,7 @@ describe('use', () => {
     "message": "Expected array, received string"
   }
 ]'`)
-      vi.mocked(loadAppConfiguration).mockRejectedValue(error)
+      vi.mocked(getAppConfigurationContext).mockRejectedValue(error)
```
dmerand, line 124: This test now only proves that discovery errors propagate out of `use()`. If the intended contract is still "don’t cache a config unless it is valid enough to be used as the active app config," then I think we need at least one fixture-based test with a parseable-but-schema-invalid TOML and an assertion that the cached preference is not updated.

If the new contract is intentionally "syntactically valid + has `client_id` is enough for `use()`," then I’d suggest updating the test name/comments to make that boundary explicit, because the old test shape implied a stronger guarantee.

packages/app/src/cli/services/app/config/use.ts
```diff
@@ -47,11 +46,8 @@ export default async function use({
 
   const configFileName = (await getConfigFileName(directory, configName)).valueOrAbort()
 
-  const {configuration} = await loadAppConfiguration({
-    userProvidedConfigName: configFileName,
-    directory,
-  })
-  setCurrentConfigPreference(configuration, {configFileName, directory})
+  const {activeConfig} = await getAppConfigurationContext(directory, configFileName)
+  setCurrentConfigPreference(activeConfig.file.content, {configFileName, directory})
```
dmerand, line 50: I see the intent here is to make `app config use` only switch the active TOML and defer full schema validation to later commands. I’m still worried about the UX regression relative to the old path: this command used to reject configs that were parseable TOML but no longer valid app config, whereas now it can cache them as the active selection as long as `client_id` is present. That shifts the failure away from the point where the user made the choice, which makes diagnosis harder.

If that behavior change is intentional, could we call it out explicitly in the command contract/tests? Otherwise I think we should keep a narrow validated load here before writing the cached preference.

dmerand (commented):
I just have a couple of extra comments.

packages/app/src/cli/models/app/config-file-naming.ts
ryancbahan: there's a test file included

packages/app/src/cli/models/app/loader.ts
ryancbahan: I'll update to preserve original behavior and add test.

packages/app/src/cli/services/app/config/use.test.ts
ryancbahan: I'll make the test more explicit that syntactic validation is the aim here.

packages/app/src/cli/services/dev/app-events/file-watcher.ts
```diff
@@ -87,6 +87,23 @@ export class FileWatcher {
...
+    await Promise.all(
+      fullExtensionDirectories.map(async (dir) => {
+        try {
+          await mkdir(dir.replace(/\/\*+$/, ''))
+          // eslint-disable-next-line no-catch-all/no-catch-all
+        } catch {
+          // Non-fatal: directory may be unwritable (e.g. test fixtures)
+        }
+      }),
+    )
```
isaacroldan, line 105: This makes sense to do, but I don't think it should be the file-watcher responsibility :thinking_face:  
Maybe something to do at the start of `dev`, before the dev session starts?

ryancbahan:
@dmerand @isaacroldan I'd love to get alignment on your full feedback of what it will take to get this merged. Does the current state of your comments reflect that? I just want to make sure there's clarity in next steps to getting this over the line, if we're in agreement that it's the right abstraction.

packages/app/src/cli/services/dev/app-events/file-watcher.ts
ryancbahan: I'm not sure I agree -- the file watcher is what breaks if chokidar can't resolve dirs properly.   
  
in any case, is there any way to decouple this feedback to unblock this pr? I think there are improvements we can continue to work on. This was addressing a pre-existing bug and I'd love to see us ship the project model if we agree it's the right abstraction.

packages/app/src/cli/services/app/config/use.ts
ryancbahan: Can you elaborate on which diagnoses are specifically harder? The tests and jsdoc call out the update behavior pretty clearly already.

packages/app/src/cli/services/dev/app-events/file-watcher.ts
isaacroldan: We can problably do this in a separate PR yeah. You are right this is fixing a pre-existing bug and is not related to the main goal of the PR.

packages/app/src/cli/services/app-context.ts
```diff
@@ -68,26 +73,28 @@ export async function linkedAppContext({
...
   }
 
-  // If the clientId is provided, update the configuration state with the new clientId
-  if (clientId && clientId !== configState.basicConfiguration.client_id) {
-    configState.basicConfiguration.client_id = clientId
-  }
+  // Determine the effective client ID
+  // Note: activeConfig.file.content is JsonMapType — client_id is untyped.
+  // The cast is safe here because activeConfig comes from a linked app (has client_id).
+  const configClientId = activeConfig.file.content.client_id as string
```
isaacroldan: if we know the activeConfig is linked, why not store the client_id in a way that can be accesed safely?  
I know that when building the active config we are already validating this, but if that ever changes this would silently start failing.  
  
A couple options:  
\- store the client_id as optional directly in the "ActiveConfig"  
\- Have something like this in ActiveConfig:

```typescript
{
...
linkedStatus: {isLinked: false} | {isLinked: true, clientId: string}
...
}
```

ryancbahan: To me, this goes against the principle of what we are trying to do. We should be moving away from storing so much state and passing it around, and moving towards more derived state expressly at the callsite that needs it. This is the callsite at which we derive the state.

isaacroldan: I can see that point, but being linked is tied to having a clientId, so I don't think is that bad.  
  
If not, I'd say we'll need to validate here that `activeConfig.file.content.client_id` is actually a non-empty string, not just casting.

PR #7024, opened 2026-03-16
fix: wire --poll flag through to chokidar in theme dev

graygilmore:
## Summary

- The `--poll` flag was defined on `theme dev` but never wired through to chokidar's `usePolling` option, making it a no-op
- On Linux/Windows, chokidar's `fs.watch` backend compares `mtimeMs` and suppresses change events when the timestamp hasn't changed. Build tools (like Gulp + Sass) that preserve the original source file's timestamp on compiled output cause file change events to be silently dropped
- This wires `poll` through the full chain: command → service → `mountThemeFileSystem` → `chokidar.watch`. When `--poll` is set, both `usePolling: true` and `useFsEvents: false` are passed to chokidar (`useFsEvents: false` is required because on macOS chokidar prefers FSEvents over polling even when `usePolling: true` is set)
- Also unhides the `--poll` flag and improves its description to tell users when to use it

Related: https://community.shopify.dev/t/theme-dev-sometimes-misses-file-updates/29269/4

## Test plan

- [x] New tests verify `usePolling: true` + `useFsEvents: false` are passed to chokidar when `poll: true`
- [x] New tests verify `usePolling` and `useFsEvents` are absent from chokidar options when `poll` is not set
- [x] All 44 existing tests in `theme-fs.test.ts` continue to pass
- [ ] Manual: `shopify theme dev --poll` enables chokidar polling mode
- [ ] Manual: `shopify theme dev --help` shows the `--poll` flag with improved description

### DIFF
--- a/packages/cli-kit/src/public/node/themes/types.ts
+++ b/packages/cli-kit/src/public/node/themes/types.ts
@@ -30,6 +30,7 @@ export interface ThemeFileSystemOptions {
   listing?: string
   noDelete?: boolean
   notify?: string
+  poll?: boolean
 }
 
 /**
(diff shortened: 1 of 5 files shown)


graygilmore:
Closing in favor of new PR with corrected branch name (gg-wire-poll-flag) and regenerated OCLIF manifests.

PR #7009, opened 2026-03-13
Faster CI

gonzaloriestra:
### WHY are these changes introduced?

The CI is a bit slow

### WHAT is this pull request doing?

Reduces PR check wall-clock time from 8.7m to 5.1m (47% faster)

- Shard Windows unit tests 2-way — slowest job drops from 8.7m → 5.1m
- Shard test coverage 2-way + merge — coverage drops from 7.4m → 4.1m
- Use `basic` reporter in CI — less output overhead (local stays verbose)
- Add `--prefer-offline` to pnpm install — skip network for cached packages
- Add `--coverage.all=false` — skip instrumenting untested files, removes need for full build
- Remove `ui-extensions-test-utils` from vitest workspace — no tests in package
- Add `unit-tests-gate` job — single required check for all matrix results

### How to test your changes?

CI

### Measuring impact

How do we know this change was effective? Please choose one:

- [x] n/a - this doesn't need measurement, e.g. a linting rule or a bug-fix
- [ ] Existing analytics will cater for this addition
- [ ] PR includes analytics changes to measure impact

### Checklist

- [x] I've considered possible cross-platform impacts (Mac, Linux, Windows)
- [x] I've considered possible [documentation](https://shopify.dev) changes

PR #7029, opened 2026-03-17
Faster unit tests

gonzaloriestra:

Local unit test wall-clock time is ~67s, largely due to vitest overhead (fork pool, eager `includeSource` scanning, heavy setup imports) and unnecessary `sleep`/`setTimeout` delays scattered across tests.

Reduces local unit test wall-clock time from 67s → 53s (21% faster).

With the improvements from both branches in the stack, CI time is reduced from ~10 min to less than 5 min (+50% faster).

Changes:

- Switch vitest pool from `forks` to `threads` — saves ~12s by avoiding per-file process spawning
- Remove `includeSource` config and extract the one in-source test into its own file — cuts ~90s off aggregate collect time
- Replace eager `zod` import in vitest setup with lazy `constructor.name` check — faster test bootstrap
- Use `define` in vite config for compile-time env vars (`SHOPIFY_UNIT_TEST`, `FORCE_HYPERLINK`, `FORCE_COLOR`)
- Replace `vitest.workspace.json` with `test.projects` in root `vite.config.ts`
- Reduce excessive `sleep`/`setTimeout` delays in tests (100ms → 10ms for `waitForInputsToBeReady`, `sendInputAndWait`, and various test timers)
- Remove `setImmediate`/`clearImmediate` from `fakeTimers` (not available in threads pool)

### How to test your changes?

`pnpm test:unit`

### Measuring impact

### Checklist

### DIFF
--- a/configurations/vitest/setup.js
+++ b/configurations/vitest/setup.js
@@ -1,4 +1,5 @@
 import {expect} from 'vitest'
+import {z} from 'zod'
 
 process.env.SHOPIFY_UNIT_TEST = '1'
 process.removeAllListeners('warning')
@@ -7,9 +8,8 @@ process.removeAllListeners('warning')
 // https://github.com/vitest-dev/vitest/issues/7315#issuecomment-2606572923
 expect.addEqualityTesters([
   function (a, b) {
-    // Lazy check for ZodError to avoid importing zod eagerly
-    const aOk = a?.constructor?.name === 'ZodError' && 'issues' in a
-    const bOk = b?.constructor?.name === 'ZodError' && 'issues' in b
+    const aOk = a instanceof z.ZodError
+    const bOk = b instanceof z.ZodError
     if (aOk && bOk) {
       return this.equals(a.message, b.message) && this.equals(a.issues, b.issues)
     }
--- a/package.json
+++ b/package.json
@@ -77,7 +77,8 @@
     "tmp": "^0.2.5",
     "ts-node": "^10.9.1",
     "typescript": "5.9.3",
-    "vitest": "^3.1.4"
+    "vitest": "^3.1.4",
+    "zod": "3.24.4"
   },
   "workspaces": {
     "packages": [
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -145,6 +145,9 @@ importers:
       vitest:
         specifier: ^3.1.4
         version: 3.2.4(@types/node@18.19.70)(jiti@2.6.1)(jsdom@28.1.0)(msw@2.12.10(@types/node@18.19.70)(typescript@5.9.3))(sass@1.97.3)(yaml@2.8.2)
+      zod:
+        specifier: 3.24.4
+        version: 3.24.4
 
   packages/app:
     dependencies:


gonzaloriestra:
* **#7029** <a href="https://app.graphite.com/github/pr/Shopify/cli/7029?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Shopify/cli/7029?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#7009** <a href="https://app.graphite.com/github/pr/Shopify/cli/7009?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

configurations/vite.config.ts
```diff
@@ -58,8 +59,6 @@ export default function config(packagePath: string, {poolStrategy}: ConfigOption
           'clearTimeout',
           'setInterval',
           'clearInterval',
-          'setImmediate',
-          'clearImmediate',
```
ryancbahan, line 62: how does this affect shared memory between runs? any risk of removal?

configurations/vitest/setup.js
```diff
@@ -8,8 +7,9 @@ process.removeAllListeners('warning')
 // https://github.com/vitest-dev/vitest/issues/7315#issuecomment-2606572923
 expect.addEqualityTesters([
   function (a, b) {
-    const aOk = a instanceof z.ZodError
-    const bOk = b instanceof z.ZodError
+    // Lazy check for ZodError to avoid importing zod eagerly
```
ryancbahan: this makes us rely on zod's internal api, which is not considered public and stable. probably will be fine, but worth noting that a version bump could break this

packages/app/src/cli/services/dev/app-events/app-event-watcher.test.ts
```diff
@@ -474,7 +474,7 @@ describe('app-event-watcher', () => {
         await flushPromises()
 
         // Wait for event processing
-        await new Promise((resolve) => setTimeout(resolve, 100))
+        await new Promise((resolve) => setTimeout(resolve, 10))
```
ryancbahan, line 477: is this safe to drop 10x in one go? should we start with 50?

configurations/vite.config.ts
gonzaloriestra: According to Claude, fake timers are per-test (reset by `mockReset: true`), and real `setImmediate` callbacks are per-event-loop-tick, so they don't leak across tests more than they would in forks mode

configurations/vitest/setup.js
gonzaloriestra: I think it's not a big deal, but I just checked that this change saves less than a second, so I've reverted it

packages/app/src/cli/services/dev/app-events/app-event-watcher.test.ts
gonzaloriestra: I'll re-run the tests a few times before merging to ensure there's no flakiness instead, makes sense?

PR #7031, opened 2026-03-17
Redirect to app redirect landing page in admin

dmerand:
## What

Route the CLI preview URL through the admin-web preflight entrypoint instead of linking directly into the app redirect flow.

Use the same preview entry URL shape for local stores, with `admin.shop.dev` as the unified admin host.

## Why

Pressing `p` from `shopify app dev` should reopen Dev Console through the UI layer that owns that state.

Sending the browser directly to Core's redirect flow gets the app installed or redirected, but it does not reliably restore Dev Console for all app shapes. Moving the entrypoint to admin-web keeps the Dev Console concern in admin-web instead of threading it through OAuth and callback URLs.

## How

- update `buildAppURLForWeb` to generate `/store/<store>/extensions-dev/preview?client_id=...`
- keep the production host on `admin.shopify.com`
- map local stores to the same admin-web path on `admin.shop.dev`
- use that builder for local store preview URLs in `setup-dev-processes`
- add coverage for the local payload path that also surfaces `app.url`

## Testing

Manual verification:
- Verified with the `apps-platform` rig in local development
- Confirmed the preview flow restores Dev Console for embedded app flows after entering through admin-web preflight

Additional checks:
- Run `shopify app dev` against a production-like store and press `p`
- Run `SHOPIFY_SERVICE_ENV=local shopify app dev` against a local store and confirm the preview URL uses `https://admin.shop.dev/store/<store>/extensions-dev/preview?...`
- Open the preview URL and verify the flow enters admin-web before continuing to the existing app redirect flow

### DIFF
--- a/packages/app/src/cli/services/dev/extension/payload/store.test.ts
+++ b/packages/app/src/cli/services/dev/extension/payload/store.test.ts
@@ -7,7 +7,17 @@ import {
 import {UIExtensionPayload, ExtensionsEndpointPayload} from './models.js'
 import * as payload from '../payload.js'
 import {ExtensionInstance} from '../../../../models/extensions/extension-instance.js'
+import {normalizeStoreFqdn, storeAdminUrl} from '@shopify/cli-kit/node/context/fqdn'
 import {beforeEach, describe, expect, test, vi} from 'vitest'
+
+vi.mock('@shopify/cli-kit/node/context/fqdn', async (importOriginal) => {
+  const original = await importOriginal<typeof import('@shopify/cli-kit/node/context/fqdn')>()
+  return {
+    ...original,
+    normalizeStoreFqdn: vi.fn(original.normalizeStoreFqdn),
+    storeAdminUrl: vi.fn(original.storeAdminUrl),
+  }
+})
 
 describe('getExtensionsPayloadStoreRawPayload()', () => {
   test('returns the raw payload', async () => {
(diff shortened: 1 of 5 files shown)


dmerand:
* **#7031** <a href="https://app.graphite.com/github/pr/Shopify/cli/7031?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Shopify/cli/7031?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

dmerand:
One implementation note: this PR intentionally does **not** change  globally.\n\nThe bug here is specifically about the preview-entry URL for , so the change is scoped to store-aware preview URL building and local-store preview wiring. That fixes the user-facing flow without changing the broader local admin host convention repo-wide.\n\nThere is still a separate local-dev consistency question around  vs  for generic admin host helpers. If we want to clean that up, I think it should be a follow-up that audits current  callers and decides whether they want a generic admin host or a store-scoped unified admin URL.

dmerand:
Follow-up on the note above — reposting with formatting fixed:

This PR intentionally does **not** change `adminFqdn()` globally.

The bug here is specifically about the preview-entry URL for `shopify app dev`, so the change is scoped to store-aware preview URL building and local-store preview wiring. That fixes the user-facing flow without changing the broader local admin host convention repo-wide.

There is still a separate local-dev consistency question around `admin.my.shop.dev` vs `admin.shop.dev` for generic admin host helpers. If we want to clean that up, I think it should be a follow-up that audits current `adminFqdn()` callers and decides whether they want a generic admin host or a store-scoped unified admin URL.

PR #7032, opened 2026-03-17
create allowlist for localhost variants and validate host on requests

EvilGenius13:

Fixes https://github.com/Shopify/developer-tools-team/issues/1092

When running `theme dev` a potential security vulnerability exists where a developer could visit a malicious site while the CLI is running that uses [DNS rebinding](https://en.wikipedia.org/wiki/DNS_rebinding) to extract data.

On `theme dev` boot up, create an allow list of localhost forms `'localhost', '127.0.0.1', '::1', '0.0.0.0'`. When requests come in, we deny those that aren't from the allow list.

> **Update:** This Host-header validation also covers the **theme app extension** dev server started by `shopify app dev` (default port `9293`), not just `theme dev` (`9292`). The same `createHostValidationHandler` middleware is wired into `theme-ext-server.ts` as well. Verified live on both servers: allowlisted localhost variants (`localhost`, `127.0.0.1`, `[::1]`, case-insensitive, trailing-dot) pass through, while bad or missing `Host` headers return `400`.

### How to test your changes?

**Note** This vulnerability does not exist when using an environment. You can test this by running the commands below using an environment and they should return a `401`.

On any current version of the CLI:
- run `theme dev` and in a separate terminal, try these curl commands:
    - `curl -s -I -H "Host: localhost:9292" http://localhost:9292/ | head -1`
    - `curl -s -I -H "Host: attacker.com:9292" http://localhost:9292/ | head -1`
    - `curl -s -I -H "Host: evil.local:9292" http://localhost:9292/ | head -1`
    - `curl -s -I -H "Host: localhost:1234" http://localhost:9292/ | head -1`
These should return a `200`

On the current branch:
- build the branch
- run `theme dev` and in a separate terminal, try the curl commands again:
    - `curl -s -I -H "Host: localhost:9292" http://localhost:9292/ | head -1` **This will return a `200` as it's on the allowlist**
   These commands should return a `400`
    - `curl -s -I -H "Host: attacker.com:9292" http://localhost:9292/ | head -1` **This will return
    - `curl -s -I -H "Host: evil.local:9292" http://localhost:9292/ | head -1`
    - `curl -s -I -H "Host: localhost:1234" http://localhost:9292/ | head -1`

To verify the theme app extension server is also protected, repeat the bad-host curls against port `9293` while running `shopify app dev` — they should likewise return `400`.

### Measuring impact

### Checklist

### DIFF
--- a/packages/theme/src/cli/utilities/theme-environment/host-validation.ts
+++ b/packages/theme/src/cli/utilities/theme-environment/host-validation.ts
@@ -1,0 +1,54 @@
+import {createError, defineEventHandler, sendError} from 'h3'
+
+function createAllowedHostsSet(host: string, port: number): Set<string> {
+  const allowedHosts = new Set<string>()
+  const portSuffix = `:${port}`
+  const normalizedHost = host.toLowerCase().trim()
+
+  allowedHosts.add(`${normalizedHost}${portSuffix}`)
+
+  // When binding to localhost variants or 0.0.0.0, allow all localhost forms
+  const localhostVariants = ['localhost', '127.0.0.1', '::1', '0.0.0.0']
+  if (localhostVariants.includes(normalizedHost)) {
+    allowedHosts.add(`localhost${portSuffix}`)
+    allowedHosts.add(`127.0.0.1${portSuffix}`)
+    allowedHosts.add(`[::1]${portSuffix}`)
+  }
+
+  return allowedHosts
+}
+
+function normalizeHostHeader(hostHeader: string | undefined): string | undefined {
+  if (!hostHeader) return undefined
+  // Lowercase, then strip trailing dot before port (or at end for plain hostname).
+  // IPv6 brackets: trailing dot would be after `]`, e.g. [::1].:9292
+  return hostHeader.toLowerCase().replace(/\.(?=:\d|$)/, '')
+}
+
+/**
+ * Creates an h3 event handler that validates the request's Host header
+ * against an allowlist of configured host/port and localhost variants.
+ *
+ * Used to mitigate DNS rebinding attacks on local dev servers.
+ *
+ * Returns a 400 Bad Request when the Host header is missing or not in the allowlist.
+ */
+export function createHostValidationHandler(host: string, port: number) {
+  const allowedHosts = createAllowedHostsSet(host, port)
+
+  return defineEventHandler((event) => {
+    const hostHeader = event.node.req.headers.host
+    const normalizedHost = normalizeHostHeader(hostHeader)
+
+    if (!normalizedHost || !allowedHosts.has(normalizedHost)) {
+      return sendError(
+        event,
+        createError({
+          statusCode: 400,
+          statusMessage: 'Bad Request',
+          message: 'Invalid Host header',
+        }),
+      )
+    }
+  })
+}
--- a/packages/theme/src/cli/utilities/theme-environment/host-validation.ts
+++ b/packages/theme/src/cli/utilities/theme-environment/host-validation.ts
@@ -1,4 +1,5 @@
 import {createError, defineEventHandler, sendError} from 'h3'
+import * as os from 'node:os'
 
 function createAllowedHostsSet(host: string, port: number): Set<string> {
   const allowedHosts = new Set<string>()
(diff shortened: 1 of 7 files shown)


packages/theme/src/cli/utilities/theme-environment/theme-environment.ts
```diff
@@ -123,9 +131,54 @@ interface DevelopmentServerInstance {
...
 }
 
+function createAllowedHostsSet(host: string, port: string): Set<string> {
+  const allowedHosts = new Set<string>()
+  const portSuffix = `:${port}`
+  const normalizedHost = host.toLowerCase().trim()
+
+  allowedHosts.add(`${normalizedHost}${portSuffix}`)
+
+  // When binding to localhost variants or 0.0.0.0, allow all localhost forms
```
graygilmore: Either explain more of a why if it's confusing or drop this comment

```suggestion
```

packages/theme/src/cli/utilities/theme-environment/theme-environment.ts
graygilmore, line 1: There's a chance I'm blind here: does this work if a user passes in a `--host` flag?

EvilGenius13: Yes it does. In the `services/dev.ts` file it runs `setupDevServer` and passes a context in that would include flags like `host` and `port` so we would be creating the custom host/port if passed in.

packages/theme/src/cli/utilities/theme-environment/theme-environment.ts
EvilGenius13: If you think most people would understand I can drop the comment. Otherwise I can add something like `browsers will send whatever localhost you typed such as localhost or 127.0.0.1 so we need to cover the different formats.`

graygilmore (approved):

packages/theme/src/cli/utilities/theme-environment/theme-environment.ts
aswamy: I didn't think people would actually type `0.0.0.0` - that's usually an address you bind to accept _any_ request.

However, I can imagine people visiting their local private IP address in their home (like `192.168.2.100`). Should we also be white listing private IPs? 🤔

aswamy: Wait, i think you added the `host` you ran the `theme dev` command with to the allowlist. You should be fine here (ignore previous comment).

aswamy (approved):

packages/theme/src/cli/utilities/theme-environment/host-validation.ts
```diff
@@ -0,0 +1,54 @@
...
+
+function createAllowedHostsSet(host: string, port: number): Set<string> {
+  const allowedHosts = new Set<string>()
+  const portSuffix = `:${port}`
+  const normalizedHost = host.toLowerCase().trim()
+
+  allowedHosts.add(`${normalizedHost}${portSuffix}`)
+
+  // When binding to localhost variants or 0.0.0.0, allow all localhost forms
+  const localhostVariants = ['localhost', '127.0.0.1', '::1', '0.0.0.0']
```
gonzaloriestra, line 11: `0.0.0.0` is a bind address, not usually the Host header clients send. If someone runs `theme dev --host 0.0.0.0`, they’ll typically access it through `127.0.0.1` locally or a LAN IP like `192.168.x.x` from another device, and that would throw an error now, no? I think we could add the missing variants with `os.networkInterfaces()`.

The localhost cases are mostly covered, but there are a couple of edge cases: `127.0.0.0/8` is all loopback, not just `127.0.0.1`, and IPv6 loopback can also appear expanded as `[0:0:0:0:0:0:0:1]`.

EvilGenius13: Great catches thanks @gonzaloriestra!

Fixed the `0.0.0.0 / LAN` work. 

As for the loopback and IPV6. I didn't add the extra coverage of all `127.0.0.0/8` and forms like `[0:0:0:0:0:0:0:1] ` because it meant adding a lot more surface area and creating parsing to ensure we catch everything.

If you think it's worth adding the remaining cases let me know and I can make a follow up PR if that's okay!

gonzaloriestra: Yeah, I agree the extra cases are probably not worth it. Thanks for fixing it!

gonzaloriestra (approved):

PR #7033, opened 2026-03-17
fix: require authentication key for GraphiQL dev server

isaacroldan:

GraphiQL authentication keys were previously optional, requiring manual specification for security. This created a poor developer experience where keys had to be manually managed and could be lost across dev server restarts.

- Automatically derives GraphiQL authentication keys from the app secret and store FQDN when no explicit key is provided
- Updates the `--graphiql-key` flag description to reflect that keys are now automatically generated by default
- Adds authentication requirement to the `/graphiql/status` endpoint
- Ensures GraphiQL URLs always include the authentication key parameter
- Adds comprehensive tests for the key derivation function to verify deterministic behavior and uniqueness across different inputs

The derived keys are deterministic (same across dev server restarts) but not guessable without the app secret, providing both security and a seamless developer experience.

### How to test your changes?

1. Start a dev server without specifying `--graphiql-key`
2. Verify that GraphiQL loads successfully with an auto-generated key
3. Restart the dev server and confirm the same key is used (browser tabs should remain functional)
4. Test with different stores/apps to ensure unique keys are generated
5. Verify that providing an explicit `--graphiql-key` still works as expected

### Measuring impact

### Checklist

### DIFF
--- a/packages/app/src/cli/services/dev/graphiql/server.ts
+++ b/packages/app/src/cli/services/dev/graphiql/server.ts
@@ -10,9 +10,27 @@ import {adminUrl, supportedApiVersions} from '@shopify/cli-kit/node/api/admin'
 import {fetch} from '@shopify/cli-kit/node/http'
 import {renderLiquidTemplate} from '@shopify/cli-kit/node/liquid'
 import {outputDebug} from '@shopify/cli-kit/node/output'
+import {createHmac, timingSafeEqual} from 'crypto'
 import {Server} from 'http'
 import {Writable} from 'stream'
 import {createRequire} from 'module'
+
+/**
+ * Derives a deterministic GraphiQL authentication key from the app's API secret and store FQDN.
+ * The key is stable across dev server restarts (so browser tabs survive restarts)
+ * but is not guessable without the app secret.
+ */
+export function deriveGraphiQLKey(apiSecret: string, storeFqdn: string): string {
+  return createHmac('sha256', apiSecret).update(`graphiql:${storeFqdn}`).digest('hex')
+}
+
+/**
+ * Resolves the GraphiQL authentication key. Uses the explicitly provided key
+ * if non-empty, otherwise derives one deterministically from the app secret.
+ */
+export function resolveGraphiQLKey(providedKey: string | undefined, apiSecret: string, storeFqdn: string): string {
+  return providedKey?.trim() || deriveGraphiQLKey(apiSecret, storeFqdn)
+}
 
 const require = createRequire(import.meta.url)
 
@@ -50,15 +68,21 @@ export function setupGraphiQLServer({
   appUrl,
   apiKey,
   apiSecret,
-  key,
+  key: providedKey,
   storeFqdn,
 }: SetupGraphiQLServerOptions): Server {
+  // Always require an authentication key. If not explicitly provided, derive one
+  // deterministically from apiSecret + storeFqdn so the key is stable across restarts
+  // (browser tabs survive dev server restarts) but not guessable without the app secret.
+  const key = resolveGraphiQLKey(providedKey, apiSecret, storeFqdn)
   outputDebug(`Setting up GraphiQL HTTP server on port ${port}...`, stdout)
   const app = express()
 
   function failIfUnmatchedKey(str: string, res: express.Response): boolean {
-    if (!key || str === key) return false
-    res.status(404).send(`Invalid path ${res.req.originalUrl}`)
+    const strBuffer = Buffer.from(str ?? '')
+    const keyBuffer = Buffer.from(key)
+    if (strBuffer.length === keyBuffer.length && timingSafeEqual(strBuffer, keyBuffer)) return false
+    res.status(404).type('text/plain').send(`Invalid path ${res.req.originalUrl}`)
     return true
   }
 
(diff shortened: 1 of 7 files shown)


isaacroldan:
* **#7033** <a href="https://app.graphite.com/github/pr/Shopify/cli/7033?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Shopify/cli/7033?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/app/src/cli/services/dev/processes/setup-dev-processes.ts
```diff
@@ -119,8 +120,11 @@ export async function setupDevProcesses({
   const useDevConsole = is1PDev && anyPreviewableExtensions
   const previewURL = useDevConsole ? devConsoleURL : appPreviewUrl
 
+  // Always resolve a GraphiQL auth key. Use the explicit flag if provided,
+  // otherwise derive one deterministically from apiSecret + storeFqdn.
+  const resolvedGraphiqlKey = graphiqlKey ?? deriveGraphiQLKey(apiSecret, storeFqdn)
```
ryancbahan: it would be nice to extract graphiql concerns into its own layer and make the `dev` processes more explicitly composed

isaacroldan: For now I extracted the key generation logic to the graphiql server file.   
But the GrapiQL URL is used for other processes, we need a mini refactor to not have that here. Something to improve in a new PR problably!

packages/app/src/cli/services/dev/graphiql/server.ts
```diff
@@ -50,14 +68,18 @@ export function setupGraphiQLServer({
...
+  // Always require an authentication key. If not explicitly provided, derive one
+  // deterministically from apiSecret + storeFqdn so the key is stable across restarts
+  // (browser tabs survive dev server restarts) but not guessable without the app secret.
+  const key = resolveGraphiQLKey(providedKey, apiSecret, storeFqdn)
   outputDebug(`Setting up GraphiQL HTTP server on port ${port}...`, stdout)
   const app = express()
 
   function failIfUnmatchedKey(str: string, res: express.Response): boolean {
-    if (!key || str === key) return false
+    if (str === key) return false
```
dmerand: 🔒 **Security**: Using `===` for secret comparison is susceptible to timing side-channel attacks, as the comparison fails immediately upon the first mismatched character. While the risk is lower in a local development environment, using `crypto.timingSafeEqual` is a best practice for authentication tokens.

**Suggestion:**     // import { timingSafeEqual } from 'crypto'
    const strBuffer = Buffer.from(str ?? '')
    const keyBuffer = Buffer.from(key)
    if (strBuffer.length !== keyBuffer.length || !timingSafeEqual(strBuffer, keyBuffer)) return false

packages/app/src/cli/services/dev/graphiql/server.ts
```diff
@@ -50,14 +68,18 @@ export function setupGraphiQLServer({
...
     res.status(404).send(`Invalid path ${res.req.originalUrl}`)
```
dmerand: 🔒 **Security**: Echoing `res.req.originalUrl` directly in the response without sanitization or setting the `Content-Type` to `text/plain` creates a Reflected XSS vulnerability. An attacker could construct a malicious URL (e.g., `?key=<script>...`) that executes JavaScript in the victim's browser context.

**Suggestion:**     res.status(404).type('text/plain').send(`Invalid path ${res.req.originalUrl}`)

packages/app/src/cli/services/dev/graphiql/templates/graphiql.tsx
dmerand, line 1: (Ref Line 54) 🎨 **Code Style**: Since the setup logic now guarantees a `key` is always derived or provided, this property should be marked as required in the interface to reflect the stricter contract.

**Suggestion:**   key: string

packages/app/src/cli/services/dev/processes/setup-dev-processes.ts
```diff
@@ -119,8 +120,9 @@ export async function setupDevProcesses({
   const useDevConsole = is1PDev && anyPreviewableExtensions
   const previewURL = useDevConsole ? devConsoleURL : appPreviewUrl
 
+  const resolvedGraphiqlKey = resolveGraphiQLKey(graphiqlKey, apiSecret, storeFqdn)
   const graphiqlURL = shouldRenderGraphiQL
-    ? `http://localhost:${graphiqlPort}/graphiql${graphiqlKey ? `?key=${graphiqlKey}` : ''}`
+    ? `http://localhost:${graphiqlPort}/graphiql?key=${resolvedGraphiqlKey}`
```
dmerand: 🐛 **Bug**: The `resolvedGraphiqlKey` may contain special characters if provided by the user via flags. Injecting it directly into the query string without encoding can result in a malformed URL or unexpected behavior.

**Suggestion:**     ? `http://localhost:${graphiqlPort}/graphiql?key=${encodeURIComponent(resolvedGraphiqlKey)}`

dmerand (commented):
I'm aligned with the approach. The agents had some good security suggestions worth considering, I've left notes here.

packages/app/src/cli/services/dev/graphiql/server.ts
isaacroldan: I mean, i guess its true, but seems like an overkill for this use case :sweat_smile:

packages/app/src/cli/services/dev/processes/setup-dev-processes.ts
isaacroldan: Fixed!

isaacroldan:
### Merge activity

* **Mar 18, 4:12 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/Shopify/cli/7033).
* **Mar 18, 4:12 PM UTC**: @isaacroldan added this pull request to the GitHub merge queue with [Graphite](https://app.graphite.com/github/pr/Shopify/cli/7033).

ryancbahan (approved):

PR #7047, opened 2026-03-19
QA to E2E: app basic flow (no extensions)

phyllis-sy-wu:

To replace manual CLI QA with automated end-to-end tests that validate the complete app development workflow in a real environment.

This is part of [Vault project #gsd:49408](https://vault.shopify.io/gsd/projects/49408-Developer-Platform-Agentic-Validation-of-app-configuraiton) to automate CLI QA coverage.

Adds `tests/app-basic.spec.ts`: a full lifecycle e2e test covering 9 steps:

| Step | Command | Description |
|------|---------|-------------|
| 1 | `app init` | React Router, JS, npm |
| 2 | `app dev` | Waits for "Ready" with keyboard shortcuts enabled (`CI=''`) |
| 3 | `app execute` | Runs a GraphQL query against the live dev server |
| 4 | `q` key | Sends `q` keystroke to terminate the dev server |
| 5 | `app dev clean` | Tears down the dev preview |
| 6 | `app deploy` | Deploys with a unique version tag |
| 7 | `app versions list --json` | Asserts the version tag appears |
| 8 | `app config link --client-id <secondary>` | Pre-creates a minimal `shopify.app.secondary.toml` stub (`client_id = "..."`) before spawning, so `getTomls()` finds the secondary client ID and `loadConfigurationFileName` returns the filename immediately without triggering the interactive "Configuration file name:" prompt |
| 9 | `app deploy --config secondary` | Deploys to the secondary app using the linked config file, with `SHOPIFY_FLAG_CLIENT_ID` unset to avoid the `--config`/`--client-id` mutual exclusion enforced by oclif |

Supporting changes:
- `setup/cli.ts`: adds `extendEnv: false` to `exec` and `execCreateApp` so the test's explicit `processEnv` is the complete environment — no leakage from the parent shell
- `setup/auth.ts`: logs "Authenticating automatically" for both the browser OAuth path and the partners-token path (was missing from the token path)
- `package.json`: adds `@shopify/cli-kit: 3.92.0` as a devDependency (required for the `joinPath` import, pinned to match all other packages per the repo-health version sync check)

#### Divergence from the manual flow

In the manual flow:
1. `app init` creates a **new** app (you pick org, name a new app)
2. `app dev` links to that new app
3. `app deploy` deploys to that new app
4. At the end you clean up the app from the Partners dashboard

In the e2e test:
1. `app init` scaffolds the local directory but links to a **pre-existing app** via `SHOPIFY_FLAG_CLIENT_ID`
2. Everything else runs against that pre-existing app

Why the test does it this way:
- Creating a new Partners app in CI requires interactive prompts (pick org, name the app) that are hard to automate reliably
- Using a pre-existing app with a known `client_id` skips all that interactivity
- No cleanup needed — no risk of leaving orphaned apps on the platform

The tradeoff:
- The test doesn't cover the "create new app" path of `app init`
- It does cover everything else faithfully (dev, execute, deploy, versions, config link, secondary deploy)

#### Running the test
- Requires `.env` with `SHOPIFY_FLAG_CLIENT_ID` (primary test app), `E2E_STORE_FQDN`, `E2E_SECONDARY_CLIENT_ID` (secondary test app). 
- Primary test needs access scope: `write_product`

```bash
DEBUG=1 pnpm --filter e2e exec playwright test app-basic
```

#### Verifying each step catches failures
1. Break — edit source, then 
   ```bash
   pnpm nx build app --skip-nx-cache && pnpm nx build cli --skip-nx-cache
   ```
2. Run test — confirm it fails with `‼️ Step N - <command> failed`
3. Fix — revert source, then
   ```bash
   pnpm nx build app --skip-nx-cache && pnpm nx build cli --skip-nx-cache
   ```
4. Run test — confirm it passes again

| Step | Command | Source |
|------|---------|-------------|
| 1 | `app init` | [init.ts](https://github.com/Shopify/cli/blob/8941db3500723818450cea2158b39add67256373/packages/app/src/cli/commands/app/init.ts) |
| 2 | `app dev` | [dev.ts](https://github.com/Shopify/cli/blob/04702b0a1f5d127edd867ada0d8669d266a50624/packages/app/src/cli/commands/app/dev.ts) |
| 3 | `app execute` | [execute.ts](https://github.com/Shopify/cli/blob/534aadcbe83272f016587a58ef14b169ca1a12f8/packages/app/src/cli/commands/app/execute.ts) |
| 4 | `q` key | [DevSessionUI.tsx](https://github.com/Shopify/cli/blob/6cd5658aba4307e839734021a499daecff5394c3/packages/app/src/cli/services/dev/ui/components/DevSessionUI.tsx#L228-L233) |
| 5 | `app dev clean` | [dev-clean.ts](https://github.com/Shopify/cli/blob/a020b29f2164ed77b61b65611eaa58a0b7cb9bac/packages/app/src/cli/services/dev-clean.ts) |
| 6 | `shopify app deploy --version v1` | [deploy.ts](https://github.com/Shopify/cli/blob/13470559e5c3bbd1575acd7f01cde1082a616cf1/packages/app/src/cli/commands/app/deploy.ts), [version-name.ts](https://github.com/Shopify/cli/blob/144acba386fedfcf6bdfeb58b009d5d8d59cf723/packages/app/src/cli/validations/version-name.ts) |
| 7 | `app versions list --json` | [versions-list.ts](https://github.com/Shopify/cli/blob/cb06c5d3cead807b9659685cb70523f345810fb7/packages/app/src/cli/services/versions-list.ts) |
| 8 | `app config link --client-id <secondary>` | [link.ts](https://github.com/Shopify/cli/blob/c1b9aa7cedbbff2bd97f1db4ea36012df9a3aed9/packages/app/src/cli/commands/app/config/link.ts) |
| 9 | `app deploy --config secondary` | [deploy.ts](https://github.com/Shopify/cli/blob/13470559e5c3bbd1575acd7f01cde1082a616cf1/packages/app/src/cli/commands/app/deploy.ts) |

### Measuring impact

### Checklist

### Screenshots

<details>
<summary>Expand to see comparison of manual vs. automated flow</summary>

**Step 1: `shopify app init`**
| Flow | Process/Actions |
| ------------ | ----------------|
| Manual | image |
| Automated | image |

**Step 2: `shopify app dev`**
| Flow | Process/Actions |
| ------------ | ----------------|
| Manual | image image |
| Automated | image image image |

**Step 3: `shopify app execute`**
| Flow | Process/Actions |
| ------------ | ----------------|
| Manual | image |
| Automated | image |

**Step 4: Quit - `q`**
| Flow | Process/Actions |
| ------------ | ----------------|
| Manual | image |
| Automated | image |

**Step 5: `shopify app dev clean`**
| Flow | Process/Actions |
| ------------ | ----------------|
| Manual | image |
| Automated | image |

**Step 6: `shopify app deploy --version v1`**
| Flow | Process/Actions |
| ------------ | ----------------|
| Manual | image |
| Automated | image |

**Step 7: `shopify app versions list --json`**
| Flow | Process/Actions |
| ------------ | ----------------|
| Manual | First version is from manual flow, second version is from automated flow: image |
| Automated | image |

**Step 8: `app config link`**
| Flow | Process/Actions |
| ------------ | ----------------|
| Manual | image |
| Automated | image |

**Step 9: `shopify app deploy`**
| Flow | Process/Actions |
| ------------ | ----------------|
| Manual | image |
| Automated | image |

</details>

### DIFF
--- a/packages/e2e/package.json
+++ b/packages/e2e/package.json
@@ -30,8 +30,7 @@
   },
   "devDependencies": {
     "@playwright/test": "^1.50.0",
-    "@shopify/cli-kit": "3.92.0",
-    "@types/node": "18.19.70",
+"@types/node": "18.19.70",
     "execa": "^7.2.0",
     "node-pty": "^1.0.0",
     "strip-ansi": "^7.1.0",
--- a/packages/e2e/package.json
+++ b/packages/e2e/package.json
@@ -30,6 +30,7 @@
   },
   "devDependencies": {
     "@playwright/test": "^1.50.0",
+    "@shopify/cli-kit": "3.92.0",
     "@types/node": "18.19.70",
     "execa": "^7.2.0",
     "node-pty": "^1.0.0",
--- a/packages/e2e/setup/app.ts
+++ b/packages/e2e/setup/app.ts
@@ -78,10 +78,6 @@ export const appScaffoldFixture = authFixture.extend<{appScaffold: AppScaffold}>
           timeout: 5 * 60 * 1000,
         })
 
-        if (result.exitCode !== 0) {
-          return result
-        }
-
         const allOutput = `${result.stdout}\n${result.stderr}`
         const match = allOutput.match(/([\w-]+) is ready for you to build!/)
 
--- a/packages/e2e/setup/app.ts
+++ b/packages/e2e/setup/app.ts
@@ -78,6 +78,10 @@ export const appScaffoldFixture = authFixture.extend<{appScaffold: AppScaffold}>
           timeout: 5 * 60 * 1000,
         })
 
+        if (result.exitCode !== 0) {
+          return result
+        }
+
         const allOutput = `${result.stdout}\n${result.stderr}`
         const match = allOutput.match(/([\w-]+) is ready for you to build!/)
 
--- a/packages/e2e/setup/auth.ts
+++ b/packages/e2e/setup/auth.ts
@@ -21,6 +21,8 @@ export const authFixture = cliFixture.extend<{}, {authLogin: void}>({
         return
       }
 
+      process.stdout.write('[e2e] Authenticating automatically — no action required.\n')
+
       // Clear any existing session
       await execa('node', [executables.cli, 'auth', 'logout'], {
         env: env.processEnv,
--- a/packages/e2e/setup/cli.ts
+++ b/packages/e2e/setup/cli.ts
@@ -52,6 +52,7 @@ export const cliFixture = envFixture.extend<{cli: CLIProcess}>({
         const execaOpts: ExecaOptions = {
           cwd: opts.cwd,
           env: {...env.processEnv, ...opts.env},
+          extendEnv: false,
           timeout,
           reject: false,
         }
@@ -75,6 +76,7 @@ export const cliFixture = envFixture.extend<{cli: CLIProcess}>({
         const execaOpts: ExecaOptions = {
           cwd: opts.cwd,
           env: {...env.processEnv, ...opts.env},
+          extendEnv: false,
           timeout,
           reject: false,
         }
--- a/packages/e2e/setup/cli.ts
+++ b/packages/e2e/setup/cli.ts
@@ -49,9 +49,15 @@ export const cliFixture = envFixture.extend<{cli: CLIProcess}>({
       async exec(args, opts = {}) {
         // 3 min default
         const timeout = opts.timeout ?? 3 * 60 * 1000
+        const execEnv: {[key: string]: string} = {}
+        for (const [key, value] of Object.entries({...env.processEnv, ...opts.env})) {
+          if (value !== undefined) {
+            execEnv[key] = value
+          }
+        }
         const execaOpts: ExecaOptions = {
           cwd: opts.cwd,
-          env: {...env.processEnv, ...opts.env},
+          env: execEnv,
           extendEnv: false,
           timeout,
           reject: false,
@@ -73,9 +79,15 @@ export const cliFixture = envFixture.extend<{cli: CLIProcess}>({
       async execCreateApp(args, opts = {}) {
         // 5 min default for scaffolding
         const timeout = opts.timeout ?? 5 * 60 * 1000
+        const execEnv: {[key: string]: string} = {}
+        for (const [key, value] of Object.entries({...env.processEnv, ...opts.env})) {
+          if (value !== undefined) {
+            execEnv[key] = value
+          }
+        }
         const execaOpts: ExecaOptions = {
           cwd: opts.cwd,
-          env: {...env.processEnv, ...opts.env},
+          env: execEnv,
           extendEnv: false,
           timeout,
           reject: false,
--- a/packages/e2e/tests/app-basic.spec.ts
+++ b/packages/e2e/tests/app-basic.spec.ts
@@ -1,0 +1,122 @@
+import {appScaffoldFixture as test} from '../setup/app.js'
+import {requireEnv} from '../setup/env.js'
+import {expect} from '@playwright/test'
+import {writeFileSync} from 'node:fs'
+import {join} from 'node:path'
+
+test.describe('App basic flow (no extensions)', () => {
+  test('init, dev, execute, quit, clean, deploy, versions, config link, deploy to secondary', async ({
+    appScaffold,
+    cli,
+    env,
+  }) => {
+    // Full flow: init + dev (3 min) + deploy + config link + secondary deploy — needs 10 min
+    test.setTimeout(10 * 60 * 1000)
+
+    requireEnv(env, 'clientId', 'storeFqdn', 'secondaryClientId')
+
+    // Step 1: Create a React Router app
+    const initResult = await appScaffold.init({
+      template: 'reactRouter',
+      flavor: 'javascript',
+      packageManager: 'npm',
+    })
+    expect(initResult.exitCode, `app init failed:\n${initResult.stdout}\n${initResult.stderr}`).toBe(0)
+
+    // Step 2: Start dev server via PTY
+    // Unset CI so keyboard shortcuts are enabled in the Dev UI
+    const dev = await cli.spawn(['app', 'dev', '--path', appScaffold.appDir], {env: {CI: ''}})
+    try {
+      await dev.waitForOutput('Ready, watching for changes in your app', 3 * 60 * 1000)
+
+      // Step 3: Run a GraphQL query while the dev server is running
+      const executeResult = await cli.exec(
+        ['app', 'execute', '--query', 'query { shop { name } }', '--path', appScaffold.appDir],
+        {timeout: 60 * 1000},
+      )
+      const executeOutput = executeResult.stdout + executeResult.stderr
+      expect(executeResult.exitCode, `app execute failed:\n${executeOutput}`).toBe(0)
+      expect(executeOutput).toContain('shop')
+
+      // Step 4: Press q to quit the dev server
+      dev.sendKey('q')
+      const devExitCode = await dev.waitForExit(30_000)
+      expect(devExitCode).toBe(0)
+    } finally {
+      // Step 5: Always clean up the dev preview, even if the test fails
+      dev.kill()
+      const cleanResult = await cli.exec(['app', 'dev', 'clean', '--path', appScaffold.appDir])
+      const cleanOutput = cleanResult.stdout + cleanResult.stderr
+      expect(cleanResult.exitCode, `dev clean failed:\n${cleanOutput}`).toBe(0)
+      expect(cleanOutput).toContain('Dev preview stopped')
+    }
+
+    // Step 6: Deploy the primary app
+    const versionTag = `e2e-v-${Date.now()}`
+    const deployResult = await cli.exec(
+      [
+        'app',
+        'deploy',
+        '--path',
+        appScaffold.appDir,
+        '--force',
+        '--version',
+        versionTag,
+        '--message',
+        'E2E basic flow deployment',
+      ],
+      {timeout: 5 * 60 * 1000},
+    )
+    const deployOutput = deployResult.stdout + deployResult.stderr
+    expect(deployResult.exitCode, `deploy failed:\n${deployOutput}`).toBe(0)
+
+    // Step 7: List versions and verify our tag appears
+    const listResult = await cli.exec(['app', 'versions', 'list', '--path', appScaffold.appDir, '--json'], {
+      timeout: 60 * 1000,
+    })
+    const listOutput = listResult.stdout + listResult.stderr
+    expect(listResult.exitCode, `versions list failed:\n${listOutput}`).toBe(0)
+    expect(listOutput).toContain(versionTag)
+
+    // Step 8: Config link to the secondary app
+    // Pre-create a minimal TOML stub so getTomls() finds the secondary client ID and skips
+    // the interactive "Configuration file name" prompt entirely. This avoids PTY timing races
+    // where the Enter key arrives before ink has fully initialized the text prompt, which
+    // causes renderTextPrompt to return '' → filenameFromName('') = 'shopify.app.toml' →
+    // that file already exists → overwrite confirmation prompt hangs.
+    // (--config and --client-id are mutually exclusive flags, so we can't pass both directly.)
+    writeFileSync(
+      join(appScaffold.appDir, 'shopify.app.secondary.toml'),
+      `client_id = "${env.secondaryClientId}"\n`,
+    )
+
+    const configLink = await cli.spawn(
+      ['app', 'config', 'link', '--path', appScaffold.appDir, '--client-id', env.secondaryClientId],
+      {env: {CI: '', SHOPIFY_FLAG_CLIENT_ID: undefined}},
+    )
+    await configLink.waitForOutput('is now linked to', 2 * 60 * 1000)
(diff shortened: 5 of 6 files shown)


phyllis-sy-wu:
* **#7133** <a href="https://app.graphite.com/github/pr/Shopify/cli/7133?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#7080** <a href="https://app.graphite.com/github/pr/Shopify/cli/7080?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#7047** <a href="https://app.graphite.com/github/pr/Shopify/cli/7047?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Shopify/cli/7047?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/e2e/package.json
```diff
@@ -30,6 +30,7 @@
   },
   "devDependencies": {
     "@playwright/test": "^1.50.0",
+    "@shopify/cli-kit": "3.92.0",
```
ryancbahan: I don't think we want to include cli-kit in e2e. e2e should test the public exposed behavior via the cli binary itself, it shouldn't need to be aware of any cli internals.

packages/e2e/tests/app-basic.spec.ts
```diff
@@ -0,0 +1,128 @@
+import {appScaffoldFixture as test} from '../setup/app.js'
+import {requireEnv} from '../setup/env.js'
+import {expect} from '@playwright/test'
+import {joinPath} from '@shopify/cli-kit/node/path'
```
ryancbahan: you can just use node directly in e2e

packages/e2e/tests/app-basic.spec.ts
```diff
@@ -0,0 +1,128 @@
...
+        throw new Error(`‼️ Step 2 - app dev failed\n${err.message}`)
+      })
+
+      // Step 3: Run a GraphQL query while the dev server is running
+      const executeResult = await cli.exec(
+        ['app', 'execute', '--query', 'query { shop { name } }', '--path', appScaffold.appDir],
+        {timeout: 60 * 1000},
+      )
+      const executeOutput = executeResult.stdout + executeResult.stderr
+      expect(executeResult.exitCode, '‼️ Step 3 - app execute failed').toBe(0)
```
ryancbahan, line 41: a little unclear -- we expect exit code to be 0, but the text is asserting the operation failed?

phyllis-sy-wu: This argument in expect(value, message) is Playwright's custom expect message. It only appears when the assertion fails. So '‼️ Step 3 - app execute failed' is describing what went wrong, not what we're asserting. Since this is a long multi-step test, I thought these labels help quickly identify which step broke in CI output.

So in this case, if there's bug when running `app execute`, the whole E2E test will fail with this message displayed.

packages/e2e/setup/app.ts
```diff
@@ -78,6 +78,10 @@ export const appScaffoldFixture = authFixture.extend<{appScaffold: AppScaffold}>
           timeout: 5 * 60 * 1000,
         })
 
+        if (result.exitCode !== 0) {
+          return result
```
dmerand: 🐛 **Bug**: The early return when exitCode !== 0 prevents the appDir variable from being set. However, tests immediately access appScaffold.appDir after calling init(), which triggers the getter that throws 'App has not been initialized yet. Call init() first.' This obscures the actual init failure with a confusing error message. The previous implementation would attempt to parse output and provide detailed diagnostics even on failure.

**Suggestion:** Remove the early return and let the existing error handling in lines 91-102 provide detailed diagnostics. The test will still fail via the expect assertion, but with clearer error messages.

Alternatively, if keeping the early return for performance, ensure tests check the result before accessing appDir:

```typescript
const initResult = await appScaffold.init({...})
if (initResult.exitCode !== 0) {
  throw new Error(`Step 1 failed: ${initResult.stdout}\n${initResult.stderr}`)
}
```

packages/e2e/setup/cli.ts
```diff
@@ -52,6 +52,7 @@ export const cliFixture = envFixture.extend<{cli: CLIProcess}>({
         const execaOpts: ExecaOptions = {
           cwd: opts.cwd,
           env: {...env.processEnv, ...opts.env},
+          extendEnv: false,
```
dmerand, line 61: 🐛 **Bug**: Setting extendEnv: false means child processes ONLY receive variables explicitly provided in env.processEnv and opts.env. This prevents inheritance of system variables like PATH, HOME, TMPDIR that system commands (npm, git, node) may require. This change affects ALL e2e tests, not just app-basic.spec.ts, but wasn't flagged as a breaking change in the PR description. While it provides strict isolation, it could break commands that rely on system-level environment variables.

**Suggestion:** Verify this change doesn't break existing e2e tests:
1. Run the entire e2e test suite to check for regressions
2. Confirm env.processEnv includes all required system variables (PATH, HOME, TMPDIR, NODE_OPTIONS, etc.)
3. Consider whether this strict isolation should apply to all tests or only specific ones
4. If this was specifically to prevent SHOPIFY_FLAG_CLIENT_ID leakage, consider explicitly deleting that variable instead of changing behavior for all tests

Document this as a breaking change if proceeding.

packages/e2e/tests/app-basic.spec.ts
```diff
@@ -0,0 +1,128 @@
...
+      expect(executeOutput, '‼️ Step 3 - app execute: response missing "shop" field').toContain('shop')
```
dmerand: 🐛 **Bug**: The test only checks if output contains the string 'shop', which could match error messages like 'Error: shop not found' or debug logs. This doesn't verify that the GraphQL query 'query { shop { name } }' actually succeeded and returned valid JSON.

**Suggestion:** Check for the complete JSON structure or a more specific success indicator:

```suggestion
expect(executeOutput).toContain('"shop":{') // Verify JSON structure
```

packages/e2e/tests/app-basic.spec.ts
```diff
@@ -0,0 +1,128 @@
...
+
+      // Step 4: Press q to quit the dev server
+      dev.sendKey('q')
+      const devExitCode = await dev.waitForExit(30_000).catch((err: Error) => {
+        throw new Error(`‼️ Step 4 - app dev did not exit after pressing q\n${err.message}`)
+      })
+      expect(devExitCode, '‼️ Step 4 - app dev quit failed').toBe(0)
+    } finally {
+      // Step 5: Always clean up the dev preview, even if the test fails
+      dev.kill()
```
dmerand, line 52: 🐛 **Bug**: In the finally block, dev.kill() is always called even after dev.waitForExit() has confirmed the process exited. Calling kill() on an already-exited process might throw an error or cause unexpected behavior depending on the PTY implementation.

**Suggestion:** Only call dev.kill() if the process hasn't exited:

```suggestion
if (!dev.hasExited()) dev.kill()
```

Or wrap in try-catch to handle the case gracefully.

packages/e2e/tests/app-basic.spec.ts
```diff
@@ -0,0 +1,128 @@
...
+      {env: {CI: '', SHOPIFY_FLAG_CLIENT_ID: undefined}},
+    )
+    await configLink.waitForOutput('is now linked to', 2 * 60 * 1000).catch((err: Error) => {
+      throw new Error(`‼️ Step 8 - app config link failed\n${err.message}`)
+    })
+    const configLinkExitCode = await configLink.waitForExit(30_000)
+    expect(configLinkExitCode, '‼️ Step 8 - app config link failed').toBe(0)
+
+    // Step 9: Deploy to the secondary app using the linked config file
+    const secondaryVersionTag = `e2e-secondary-v-${Date.now()}`
```
dmerand: 🐛 **Bug**: Both versionTag (line 61) and secondaryVersionTag (line 109) use Date.now() for uniqueness. While unlikely, if async operations between these lines complete quickly (e.g., with mocked/cached responses), both could generate the same timestamp and collide, especially if version tags have a global namespace.

**Suggestion:** Add a distinct prefix or random component to ensure uniqueness:

```suggestion
const secondaryVersionTag = `e2e-secondary-v-${Date.now()}-${Math.random().toString(36).slice(2, 8)}`
```

dmerand (changes requested):
I'm aligned with the overall approach. Agent review found a few bugs worth checking.

---
_Review assisted by [pair-review](https://vault.shopify.io/posts/346224-Human-in-the-loop-code-review-with-coding-agents)_

packages/e2e/package.json
phyllis-sy-wu: Addressed in [the latest commit](https://github.com/Shopify/cli/pull/7047/changes/eaad57bbb1f30039399bad2406e355d99b5ab32a).

packages/e2e/tests/app-basic.spec.ts
phyllis-sy-wu: Addressed in [the latest commit](https://github.com/Shopify/cli/pull/7047/changes/eaad57bbb1f30039399bad2406e355d99b5ab32a).

packages/e2e/setup/auth.ts
```diff
@@ -22,6 +21,8 @@ export const authFixture = cliFixture.extend<{}, {authLogin: void}>({
         return
       }
 
+      process.stdout.write('[e2e] Authenticating automatically — no action required.\n')
+
```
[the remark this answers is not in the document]
phyllis-sy-wu: This is intentional. The partners-token path is a silent early return (no browser interaction), so there's nothing to reassure the user about. This log exists specifically for the OAuth flow where the device code URL appears and you might wonder if manual action is needed.

packages/e2e/setup/cli.ts
```diff
@@ -52,6 +52,7 @@ export const cliFixture = envFixture.extend<{cli: CLIProcess}>({
...
           timeout,
           reject: false,
         }
```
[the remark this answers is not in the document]
phyllis-sy-wu: Addressed in [the latest commit](https://github.com/Shopify/cli/pull/7047/changes/eaad57bbb1f30039399bad2406e355d99b5ab32a).

packages/e2e/tests/app-basic.spec.ts
```diff
@@ -0,0 +1,128 @@
...
+        appScaffold.appDir,
+        '--config',
+        'secondary',
+        '--force',
+        '--version',
+        secondaryVersionTag,
+        '--message',
+        'E2E secondary app deployment',
+      ],
+      {timeout: 5 * 60 * 1000, env: {SHOPIFY_FLAG_CLIENT_ID: undefined}},
```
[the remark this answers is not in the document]
phyllis-sy-wu: Addressed in [the latest commit](https://github.com/Shopify/cli/pull/7047/changes/eaad57bbb1f30039399bad2406e355d99b5ab32a).

packages/e2e/tests/app-basic.spec.ts
```diff
@@ -0,0 +1,128 @@
+import {appScaffoldFixture as test} from '../setup/app.js'
+import {requireEnv} from '../setup/env.js'
+import {expect} from '@playwright/test'
+import {joinPath} from '@shopify/cli-kit/node/path'
+import * as fs from 'fs'
+
```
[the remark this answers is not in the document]
phyllis-sy-wu: Addressed in [the latest commit](https://github.com/Shopify/cli/pull/7047/changes/eaad57bbb1f30039399bad2406e355d99b5ab32a).

packages/e2e/setup/app.ts
phyllis-sy-wu: Addressed in [the latest commit](https://github.com/Shopify/cli/pull/7047/changes/eaad57bbb1f30039399bad2406e355d99b5ab32a).

packages/e2e/setup/cli.ts
phyllis-sy-wu: Addressed in [the latest commit](https://github.com/Shopify/cli/pull/7047/changes/eaad57bbb1f30039399bad2406e355d99b5ab32a).

packages/e2e/tests/app-basic.spec.ts
phyllis-sy-wu: Addressed in [the latest commit](https://github.com/Shopify/cli/pull/7047/changes/eaad57bbb1f30039399bad2406e355d99b5ab32a).
Tightened the match to `'"shop"'` instead of `'"shop":{'`. This verifies the JSON key without being sensitive to whitespace formatting (pretty-printed output would have "shop": { with a space).

packages/e2e/tests/app-basic.spec.ts
phyllis-sy-wu: Addressed in [the latest commit](https://github.com/Shopify/cli/pull/7047/changes/eaad57bbb1f30039399bad2406e355d99b5ab32a).

packages/e2e/tests/app-basic.spec.ts
phyllis-sy-wu: The two tags have different prefixes (`QA-E2E-1st-` vs now `QA-E2E-2nd-`, previously `e2e-secondary-v-`), and there are multiple async operations (deploy, versions list, config link) between the two Date.now() calls. No collision risk.

phyllis-sy-wu:
Replaced with https://github.com/Shopify/cli/pull/7367

PR #7062, opened 2026-03-20
Fix progress bars not clearing on completion after React 19 upgrade

nickwesselman:
## Summary
- The Ink 6 / React 19 upgrade (#6831, `269f3aa6`) deferred `unmountInk()` in `ConcurrentOutput` with `setImmediate()` to let React 19 flush batched state updates, but missed the same pattern in `useAsyncAndUnmount` (used by `Tasks`) and `SingleTask`
- Without the deferral, `unmountInk()` fires before the `setState` that triggers `return null` is flushed, so the final render still contains the `LoadingBar` and it is never erased from the terminal
- Wraps `unmountInk()` in `setImmediate()` in both places, matching the existing fix in `ConcurrentOutput`
- **Fix error swallowing on sequential `renderSingleTask` calls:** The `setImmediate` deferral introduced a race condition — when `renderSingleTask` is called sequentially, the first instance's deferred `unmountInk()` can fire after the second instance starts, causing the second `waitUntilExit()` to resolve prematurely. If the second task throws, the error is silently swallowed (the promise never settles). Fix: add `onError` callback to `SingleTask` (mirroring `onComplete`) so errors propagate directly rather than relying on `waitUntilExit()` rejection.

Repro: `shopify kitchen-sink async` — progress bars remain on screen after completing.

See the equivalent change in `ConcurrentOutput.tsx` from #6831 ([`269f3aa6`](https://github.com/Shopify/cli/commit/269f3aa62b831f563e34248b7e2a0ff6746f8df0#diff-c868899c4e)).

## Test plan
- [x] `Tasks.test.tsx` — 13 tests pass
- [x] `SingleTask.test.tsx` — 11 tests pass
- [x] `ui.test.ts` — 16 tests pass
- [x] Verified sequential `renderSingleTask` error path works (previously hung indefinitely)
- [ ] Manual: run `shopify kitchen-sink async` and verify progress bars clear on completion

### DIFF
--- a/.changeset/cool-papers-behave.md
+++ b/.changeset/cool-papers-behave.md
@@ -1,0 +1,5 @@
+---
+'@shopify/app': patch
+---
+
+[fix] Task progress bars once again clear when complete
--- a/.changeset/cool-papers-behave.md
+++ b/.changeset/cool-papers-behave.md
@@ -2,4 +2,4 @@
 '@shopify/app': patch
 ---
 
-[fix] Task progress bars once again clear when complete
+Task progress bars once again clear when complete
--- a/packages/cli-kit/src/public/node/ui.tsx
+++ b/packages/cli-kit/src/public/node/ui.tsx
@@ -522,7 +522,7 @@ export async function renderSingleTask<T>({
   renderOptions,
 }: RenderSingleTaskOptions<T>): Promise<T> {
   return new Promise<T>((resolve, reject) => {
-    render(<SingleTask title={title} task={task} onComplete={resolve} onAbort={onAbort} />, {
+    render(<SingleTask title={title} task={task} onComplete={resolve} onError={reject} onAbort={onAbort} />, {
       ...renderOptions,
       exitOnCtrlC: false,
     }).catch(reject)
--- a/packages/cli-kit/src/public/node/ui.tsx
+++ b/packages/cli-kit/src/public/node/ui.tsx
@@ -486,15 +486,26 @@ interface RenderTasksOptions {
 export async function renderTasks<TContext>(
   tasks: Task<TContext>[],
   {renderOptions, noProgressBar}: RenderTasksOptions = {},
-) {
-  return new Promise<TContext>((resolve, reject) => {
-    render(<Tasks tasks={tasks} onComplete={resolve} noProgressBar={noProgressBar} />, {
+): Promise<TContext> {
+  let result: TContext | undefined
+  let taskError: Error | undefined
+  await render(
+    <Tasks
+      tasks={tasks}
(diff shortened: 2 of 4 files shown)


ryancbahan (approved):
Let a note in Slack -- we should merge to fix regression, since this matches code we shipped elsewhere. but this manual management of Node event loop within a React side effect to handle mounting/unmounting feels...very cursed. I'd like us to come back to this.

nickwesselman:
* **#7062** <a href="https://app.graphite.com/github/pr/Shopify/cli/7062?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#7067](https://github.com/Shopify/cli/pull/7067) <a href="https://app.graphite.com/github/pr/Shopify/cli/7067?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#7118](https://github.com/Shopify/cli/pull/7118) <a href="https://app.graphite.com/github/pr/Shopify/cli/7118?utm_source=stack-comment-icon" target="_blank">Graphite</a>) 👈 <a href="https://app.graphite.com/github/pr/Shopify/cli/7062?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

isaacroldan (approved):

.changeset/cool-papers-behave.md
```diff
@@ -0,0 +1,5 @@
+---
+'@shopify/app': patch
+---
+
+[fix] Task progress bars once again clear when complete
```
graygilmore: nit: we generally don't include things like `[fix]` in the changesets. I'm guessing this was our pal Claude?

graygilmore (approved):
approving changeset

packages/cli-kit/src/public/node/ui.tsx
```diff
@@ -487,14 +487,25 @@ export async function renderTasks<TContext>(
   tasks: Task<TContext>[],
   {renderOptions, noProgressBar}: RenderTasksOptions = {},
 ) {
-  return new Promise<TContext>((resolve, reject) => {
-    render(<Tasks tasks={tasks} onComplete={resolve} noProgressBar={noProgressBar} />, {
+  let result: TContext
+  let taskError: Error | undefined
+  await render(
```
nickwesselman, line 492: Without this, there were unresolved promises in our tests that were causing timeouts

nickwesselman:
@ryancbahan  @isaacroldan 

I'm less confident now about this given the additional changes required to get all unit tests passing. Do we need to take a deeper look at this?​

PR #7087, opened 2026-03-24
Add install count warning for safer deploys

alfonso-noriega:
## What

Surfaces the number of affected app installations in the `app deploy` / `app release` confirmation prompt when a release removes extensions. This gives developers a clear picture of the blast radius before they confirm a destructive operation.

Closes https://github.com/shop/issues-develop/issues/22421

## How

**New GraphQL query** (`app-install-count.graphql`) — queries `installCount` on the `App` type, exposed by the Management API in shop/world#532386.

**New** **`appInstallCount()`** **method** on `DeveloperPlatformClient` + `AppManagementClient` implementation. `PartnersClient` stubs it as unsupported.

**Fetched lazily in** **`ensureDeploymentIdsPresence()`** — only when the extension breakdown contains removals (`onlyRemote.length > 0`), right before the confirmation prompt. Wrapped in try/catch so a failure never blocks the deploy.

**Warning rendered in** **`deploy-release.ts`** — added to the dangerous confirmation prompt (the one requiring the user to type the app name) when `hasDeletedExtensions && installCount > 0`.

## Behaviour unchanged in these paths

- Non-interactive (`--no-interactive` / piped input) — prompt is skipped entirely
- `--force` / `--allow-deletes` — prompt is skipped entirely
- API call failure — warning is silently omitted, deploy proceeds normally

### DIFF
--- a/packages/app/src/cli/prompts/deploy-release.ts
+++ b/packages/app/src/cli/prompts/deploy-release.ts
@@ -162,7 +162,11 @@ async function deployConfirmationPrompt({
       confirmation: appTitle,
       ...(showInstallCountWarning
         ? {
-            body: `This ${release ? 'release' : 'version'} removes extensions and related data from ${installCount} app installations.\nUse caution as this may include production data on live stores.`,
+            warningItem: [
+              'This release removes extensions and related data from',
+              {error: installCount.toString()},
+              'app installations.\nUse caution as this may include production data on live stores.',
+            ],
           }
         : {}),
     })
--- a/packages/app/src/cli/prompts/deploy-release.ts
+++ b/packages/app/src/cli/prompts/deploy-release.ts
@@ -162,7 +162,7 @@ async function deployConfirmationPrompt({
       confirmation: appTitle,
       ...(showInstallCountWarning
         ? {
-            body: `This release removes extensions and related data from ${installCount} app installations.\nUse caution as this may include production data on live stores.`,
+            body: `This ${release ? 'release' : 'version'} removes extensions and related data from ${installCount} app installations.\nUse caution as this may include production data on live stores.`,
           }
         : {}),
     })
--- a/packages/app/src/cli/prompts/deploy-release.ts
+++ b/packages/app/src/cli/prompts/deploy-release.ts
@@ -24,6 +24,7 @@ interface DeployOrReleaseConfirmationPromptOptions {
   /** If true, allow removing extensions and configuration without user confirmation */
   allowDeletes?: boolean
   showConfig?: boolean
+  installCount?: number
 }
 
 interface DeployConfirmationPromptOptions {
@@ -36,6 +37,7 @@ interface DeployConfirmationPromptOptions {
     configInfoTable: InfoTableSection
   }
   release: boolean
+  installCount?: number
 }
 
 /**
@@ -97,6 +99,7 @@ export async function deployOrReleaseConfirmationPrompt({
   configExtensionIdentifiersBreakdown,
   appTitle,
   release,
+  installCount,
 }: DeployOrReleaseConfirmationPromptOptions): Promise<boolean> {
   await metadata.addPublicMetadata(() => buildConfigurationBreakdownMetadata(configExtensionIdentifiersBreakdown))
 
@@ -117,6 +120,7 @@ export async function deployOrReleaseConfirmationPrompt({
     extensionsContentPrompt,
     configContentPrompt,
     release,
+    installCount,
   })
 }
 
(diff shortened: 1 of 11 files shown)


alfonso-noriega:
* **#7087** <a href="https://app.graphite.com/github/pr/Shopify/cli/7087?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Shopify/cli/7087?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/app/src/cli/prompts/deploy-release.ts
```diff
@@ -149,11 +154,17 @@ async function deployConfirmationPrompt({
...
   const question = `${release ? 'Release' : 'Create'} a new version${appTitle ? ` of ${appTitle}` : ''}?`
+  const showInstallCountWarning = hasDeletedExtensions && installCount !== undefined && installCount > 0
   if (isDangerous) {
     confirmationResponse = await renderDangerousConfirmationPrompt({
       message: question,
       infoTable,
       confirmation: appTitle,
+      ...(showInstallCountWarning
+        ? {
+            body: `This release removes extensions and related data from ${installCount} app installations.\nUse caution as this may include production data on live stores.`,
```
dmerand: 🐛 **Bug**: This shared prompt already switches other user-facing text based on whether the flow is a release or a deploy, but the new warning body is hard-coded to say "This release removes...". That makes the deploy-without-release path show contradictory messaging in the same prompt, which could confuse users during a destructive action.

**Suggestion:** Make the warning copy follow the existing `release` mode switch.
```suggestion
            body: `This ${release ? 'release' : 'version'} removes extensions and related data from ${installCount} app installations.\nUse caution as this may include production data on live stores.`,
```

packages/app/src/cli/services/context/identifiers.ts
```diff
@@ -58,6 +58,20 @@ export async function ensureDeploymentIdsPresence(options: EnsureDeploymentIdsPr
...
+      installCount = await options.developerPlatformClient.appInstallCount({
+        id: options.appId,
+        apiKey: options.remoteApp.apiKey,
+        organizationId: options.remoteApp.organizationId,
+      })
+      // eslint-disable-next-line no-catch-all/no-catch-all
+    } catch (_error) {
+      installCount = undefined
+    }
+  }
```
dmerand, line 78: `deployOrReleaseConfirmationPrompt()` still short-circuits early for `--force`, `--allow-updates`, `--allow-deletes`, and some non-interactive flows before any prompt work happens in [[file:packages/app/src/cli/prompts/deploy-release.ts:94-124]]. But this new `appInstallCount()` lookup now runs earlier, here, as soon as `onlyRemote.length > 0`, before that skip decision is made.

That means the confirmation-skipping paths are no longer actually skipping all confirmation-related work: they now pay for an extra API request, and on `PartnersClient` they also trigger an exception that is immediately swallowed. Since those flows never render the warning body, this lookup has no user-visible effect there.

Can we move the install-count fetch behind the same skip predicate, or otherwise reuse that decision here so we only do the lookup when the dangerous prompt will actually be shown?

packages/app/src/cli/services/context/identifiers.ts
```diff
@@ -58,6 +58,20 @@ export async function ensureDeploymentIdsPresence(options: EnsureDeploymentIdsPr
     activeAppVersion: options.activeAppVersion,
   })
 
+  let installCount: number | undefined
+  if (extensionIdentifiersBreakdown.onlyRemote.length > 0) {
+    try {
+      installCount = await options.developerPlatformClient.appInstallCount({
+        id: options.appId,
```
dmerand: `options.appId` is still the deploy-time API key in this flow, not the Management API app ID. The current callers pass `remoteApp.apiKey` into `ensureDeploymentIdsPresence()` ([[file:packages/app/src/cli/services/context.ts:166-176]], [[file:packages/app/src/cli/services/dev/processes/draftable-extension.ts:82-91]]), and this function already treats that same value as an API key a few lines above ([[file:packages/app/src/cli/services/context/identifiers.ts:52-55]]). But `appInstallCount()` forwards its `id` field as GraphQL `appId` ([[file:packages/app/src/cli/utilities/developer-platform-client/app-management-client.ts:656-660]]), which matches the real app ID returned separately from `apiKey` by `appFromIdentifiers()` ([[file:packages/app/src/cli/utilities/developer-platform-client/app-management-client.ts:361-373]]).

Passing `id: options.appId` here therefore looks incorrect: it sends the API key where the Management API expects the app ID. Because the lookup is wrapped in a broad catch, this can silently suppress the new install-count warning in the App Management flow instead of failing loudly. Can we pass `options.remoteApp.id` here instead?

dmerand (commented):
Found a few things worth considering

gonzaloriestra (approved):
Working as expected 👌

gonzaloriestra:
@alfonso-noriega I think this feature does deserve a changeset

PR #6751, opened 2026-01-09
Support flexible templates for `app init`

shauns:
### WHY are these changes introduced?

Fixes an issue where app templates using certain module types cause `app init` to fail. The issue is due to the app template using module types that aren't defined within the CLI, but more generally due to `app init` validating/understanding app configuration.

### WHAT is this pull request doing?

Instead, we process app config during app instantiation as opaque blobs -- we don't rely on being able to validate them or a particular structure being in place.

In `app init` the app is loaded twice -- once as a template, and then again as the local copy of the app is linked to the remote version of it.

### How to test your changes?

Locally: `pnpm shopify app init --path="<where I keep my apps>" --template="https://github.com/Shopify/shopify-app-template-react-router#include-declarative-definition"`

### Measuring impact

How do we know this change was effective? Please choose one:

- [ ] n/a - this doesn't need measurement, e.g. a linting rule or a bug-fix
- [x] Existing analytics will cater for this addition
- [ ] PR includes analytics changes to measure impact

### Checklist

- [x] I've considered possible cross-platform impacts (Mac, Linux, Windows)
- [x] I've considered possible [documentation](https://shopify.dev) changes

PR #7131, opened 2026-03-30
Remove rewriteConfiguration from write path to fix config2.map crash

lizkenyon:
## Summary

Fixes [shop/issues#8553](https://github.com/shop/issues/issues/8553) — `config2.map is not a function`
Vault issue: https://vault.shopify.io/issues/8435

Stops using `rewriteConfiguration` in `writeAppConfigurationFile` and replaces it with a simple `stripEmptyObjects` utility. The Zod schema walk was the root cause of the crash. `rewriteConfiguration` is kept exported for its other consumer (`breakdown-extensions.ts`), which is untouched in this PR.

## Root Cause

When `loadOpaqueApp` ([#6751](https://github.com/Shopify/cli/pull/6751)) can't Zod-validate a local TOML config (e.g., a third-party template without `client_id`), it falls back to raw, unvalidated TOML data. This raw data is deep-merged with the remote config and passed to `writeAppConfigurationFile`, where `rewriteConfiguration` walked the Zod schema assuming config values matched schema types — calling `.map()` on non-arrays and `Object.entries()` on non-objects.

Triggered by `app dev --reset` (or any link flow) on apps using third-party templates.

## Why remove instead of guard

- Defensive guards are an infinite game of catching arbitrary type mismatches
- Pass-through of invalid data causes downstream bugs (e.g., writing a config without `client_id` makes the app appear "unlinked")
- `rewriteConfiguration`'s only meaningful behaviors were **key reordering** (cosmetic) and **empty-object stripping** — only the latter is needed
- A longer-term improvement would be to use `TomlFile.patch()` for surgical edits that preserve comments/formatting

## Changes

- **`write-app-configuration-file.ts`**: Stop calling `rewriteConfiguration` in `writeAppConfigurationFile`. Add `stripEmptyObjects` utility. Add `structuredClone` to prevent mutating the caller's config (previously masked by `rewriteConfiguration` creating fresh objects). Add `Array.isArray` guard in `condenseComplianceAndNonComplianceWebhooks`. Keep `rewriteConfiguration` exported for `breakdown-extensions.ts`.
- **`link.ts`**: Remove `schema` from `writeAppConfigurationFile` call and `getAppVersionedSchema` import.
- **Tests**: Update all callers. Add `stripEmptyObjects` tests. Add type-mismatched data crash test. Convert `link.test.ts` TOML comparisons to snapshots. Update round-trip stability tests to 3-write pattern. Update 2 `configuration` object expectations to reflect that write-time mutations (webhook condensation, URI relativization) no longer leak back to the caller.

## Behavior change

Key ordering in the generated TOML now follows JS object insertion order rather than schema declaration order. This may cause a **one-time key reorder** in `shopify.app.toml` on the next `app link` or `app config pull`. Subsequent writes are stable.

## Not changed

- `breakdown-extensions.ts` — still uses `rewriteConfiguration` for config diffing. Its inputs come from the platform API and Zod-parsed config (not the raw TOML fallback), so it does not have the same crash risk.

## Test plan

- [x] `stripEmptyObjects` unit tests
- [x] `writeAppConfigurationFile` tests (comments, empty entries, empty arrays, type-mismatched data)
- [x] Config pipeline snapshot tests regenerated
- [x] Link tests pass (23/23)
- [x] Breakdown extensions tests pass without changes (25/25)
- [ ] Manual: `shopify app init` with a template lacking `client_id`, then `shopify app dev --reset` — should not crash


lizkenyon:
* **#7131** <a href="https://app.graphite.com/github/pr/Shopify/cli/7131?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Shopify/cli/7131?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

.changeset/chilly-glasses-fetch.md
```diff
@@ -0,0 +1,5 @@
+---
+'@shopify/app': patch
```
alfonso-noriega: Normally we do not add changesets for non user facing changes, but if it is meant to be trigger a patch, we are only generating changesets on the PR against the stable branch to force the new version generation

alfonso-noriega (approved):
Minor comment that needs addressing, everything else looks great!

PR #7132, opened 2026-03-31
Support assets for admin links app intents

vividviolet:
### WHY are these changes introduced?

Enables assets uploading for `admin_link` to support intents

Related to https://github.com/shop/issues-admin-extensibility/issues/2205

### WHAT is this pull request doing?

- Adds a new `admin_link` extension specification with localization support
- Implements asset inclusion for targeting configuration including tools, instructions, and intent schemas
- Configures copy-based build mode for admin link extensions
- Adds comprehensive test coverage for the new specification including configuration parsing, deploy config generation, and localization handling

### How to test your changes?

1. Create a new extension with type `admin_link`
2. Verify the extension loads correctly with the new specification
3. Test that localization features work as expected
4. Confirm that asset inclusion works for targeting configuration fields

### Measuring impact

How do we know this change was effective? Please choose one:

- [ ] n/a - this doesn't need measurement, e.g. a linting rule or a bug-fix
- [x] Existing analytics will cater for this addition
- [ ] PR includes analytics changes to measure impact

### Checklist

- [ ] I've considered possible cross-platform impacts (Mac, Linux, Windows)
- [ ] I've considered possible [documentation](https://shopify.dev) changes

PR #7027, opened 2026-03-17
Add admin hoted app local spec

alfonso-noriega:
### WHY are these changes introduced?

The admin extension type was previously hardcoded as a string literal, which prevented proper specification handling and caused inconsistencies in the extension system.

### WHAT is this pull request doing?

- Creates a new `AdminSpecIdentifier` constant and corresponding specification file for admin extensions
- Replaces the hardcoded 'admin' string with the proper `AdminSpecIdentifier` in the configuration extension IDs array
- Adds the admin specification to the loaded specifications list
- Implements admin extension as a contract-based module with single UID strategy and copy files build mode
- Adds a default case to the extension instance UID generation logic
- Updates the app configuration schema contribution logic to handle specifications without shape definitions
- Adds `uidStrategy` parameter support to contract-based module specifications

### How to test your changes?

1. Create an app with an admin section in the toml
2. Verify the extension builds and deploys correctly
3. Test that the hosted app copy files step executes during deployment

### Measuring impact

How do we know this change was effective? Please choose one:

- [x] n/a - this doesn't need measurement, e.g. a linting rule or a bug-fix
- [ ] Existing analytics will cater for this addition
- [ ] PR includes analytics changes to measure impact

### Checklist

- [ ] I've considered possible cross-platform impacts (Mac, Linux, Windows)
- [ ] I've considered possible documentation changes

PR #7134, opened 2026-03-31
Support intents for ui extensions

vividviolet:

This change modernizes the UI extension asset handling system by replacing the legacy `build_manifest` approach with a new manifest.json-based system that supports additional asset types including intents.

Related https://github.com/shop/issues-admin-extensibility/issues/2274

- Adds support for `intents` field in extension point schemas with type, action, schema, name, and description properties
- Adds `Intents` to the `AssetIdentifier` enum for asset handling
- Updates UI extension build pipeline to use `include_assets` step instead of `copy_static_assets`, with support for tools, instructions, and intent schemas
- Refactors extension payload generation to read from manifest.json instead of build_manifest, enabling better asset tracking with source file timestamps
- Implements specialized handling for intents assets that transforms them into extension point-level properties with schema asset references
- Updates tests to use the new manifest.json approach and adds comprehensive test coverage for intents, tools, and instructions assets

### How to test your changes?

Make sure you have the proper betas applied (see [canvas](https://shopify.enterprise.slack.com/docs/T024G4BGQ/F0A8DLSG1CL))

1. Create a conditional action extension `shopify app generate --template conditional_admin_action`
2. Run `pnpm shopify app dev` and verify the the UI extension has assets (main.js, should_render.js, tools schema url and instructions.md) that are fetchable through the Dev Server. You should be able to preview it in Admin Web
3. Make a change to the action extension's code a verify that live reload works in Admin Web
4. Create an Admin Link intent extension `shopify app generate --template admin_intent_link`
5. Run `pnpm shopify app dev` and verify the the Admin link extension has assets (main.js, tools schema url and instructions.md and intents schema URLs) fetchable through the Dev Server
6. Make a change to one of the static files like the intent-schema.json or tools.json and verify that Admin Web receives an update even through the websocket

### Measuring impact

### Checklist


vividviolet:
* **#7132** <a href="https://app.graphite.com/github/pr/Shopify/cli/7132?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#7134** <a href="https://app.graphite.com/github/pr/Shopify/cli/7134?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Shopify/cli/7134?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#7016** <a href="https://app.graphite.com/github/pr/Shopify/cli/7016?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#7017](https://github.com/Shopify/cli/pull/7017) <a href="https://app.graphite.com/github/pr/Shopify/cli/7017?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#6867** <a href="https://app.graphite.com/github/pr/Shopify/cli/6867?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 3 other dependent PRs ([#6868](https://github.com/Shopify/cli/pull/6868) <a href="https://app.graphite.com/github/pr/Shopify/cli/6868?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#6869](https://github.com/Shopify/cli/pull/6869) <a href="https://app.graphite.com/github/pr/Shopify/cli/6869?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#7027](https://github.com/Shopify/cli/pull/7027) <a href="https://app.graphite.com/github/pr/Shopify/cli/7027?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#6966** <a href="https://app.graphite.com/github/pr/Shopify/cli/6966?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#6941](https://github.com/Shopify/cli/pull/6941) <a href="https://app.graphite.com/github/pr/Shopify/cli/6941?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* `main`


<h2></h2>

packages/app/src/cli/services/dev/app-events/app-event-watcher.ts
```diff
@@ -322,6 +324,48 @@ export class AppEventWatcher extends EventEmitter {
     await extension.buildForBundle(buildOptions, this.buildOutputPath)
   }
 
+  /**
+   * Run client steps that aren't handled by esbuild (e.g. include_assets).
+   * Skips bundle_ui since esbuild already handled it.
+   */
+  private async runPostEsbuildSteps(extension: ExtensionInstance): Promise<void> {
```
isaacroldan: we need to find a better way to do this, @alfonso-noriega it makes sense for `dev` to execute this missing steps. Wonder how much we gain from having a custom esbuild for dev (in theory is faster because it uses contexts and only rebuilds what changed, not the whole extension)

isaacroldan: I have an agent working on this, will have an independent PR solving this on monday morning

vividviolet: Thank you so much Isaac!

isaacroldan: This PR should handle this: https://app.graphite.com/github/pr/Shopify/cli/7252

packages/app/src/cli/models/extensions/specifications/ui_extension.ts
```diff
@@ -388,28 +399,23 @@ async function validateUIExtensionPointConfig(
...
 
     const missingInstructionsError = await checkForMissingPath(
       directory,
-      buildManifest?.assets[AssetIdentifier.Instructions]?.module,
+      extensionPoint.instructions,
       target,
-      AssetIdentifier.Instructions,
+      'instructions',
     )
     if (missingInstructionsError) {
```
alfonso-noriega: should we get something similar for schema property in the intents definition? validate that the path is there and it is right?

packages/app/src/cli/services/dev/extension/server/middlewares.ts
```diff
@@ -80,12 +80,15 @@ export function getExtensionAssetMiddleware({devOptions, getExtensions}: GetExte
 
     const bundlePath = devOptions.appWatcher.buildOutputPath
     const extensionOutputPath = extension.getOutputPathForDirectory(bundlePath)
-
     const buildDirectory = dirname(extensionOutputPath)
 
-    return fileServerMiddleware(event, {
-      filePath: joinPath(buildDirectory, assetPath),
-    })
+    // Try the build output directory first (for compiled bundles), then fall back
```
alfonso-noriega: if everything is working fine, the dev bundle should have the static assets and therfore this won't be needed, right?

packages/app/src/cli/models/extensions/specifications/ui_extension.ts
```diff
@@ -109,7 +92,35 @@ const uiExtensionSpec = createExtensionSpecification({
...
       steps: [
         {id: 'bundle-ui', name: 'Bundle UI Extension', type: 'bundle_ui', config: {}},
-        {id: 'copy-static-assets', name: 'Copy Static Assets', type: 'copy_static_assets', config: {}},
+        {
+          id: 'include-ui-extension-assets',
+          name: 'Include UI Extension Assets',
+          type: 'include_assets',
+          config: {
+            generatesAssetsManifest: true,
+            builtAssets: {anchor: 'extension_points[]', groupBy: 'target', key: 'build_manifest'},
```
alfonso-noriega: I think this should be also part of the inclusions, not on the top level. actually, i am wondering if it wouldn't work as a normal type `configKey?`

alfonso-noriega (dismissed):

packages/app/src/cli/models/extensions/specifications/ui_extension.ts
vividviolet: It doesn't work because the toml points to the raw module which we don't want to copy over. We need to the compiled asset copied over and there's no toml reference for that. The `bundle_ui` step already handles the building and copying. What we're missing is putting an entry for it inside the manifest.json for the `generatesAssetsManifest` step. If we can update `bundle_ui` to take in`generatesAssetsManifest`that would work but right now the manifest.json does a replace only instead of each step adding to it.

packages/app/src/cli/models/extensions/specifications/ui_extension.ts
vividviolet: Oops, this is left over code from when we were not using client steps. I'm thinking this should be moved over to the client steps instead? Maybe we can support a `required` prop so the steps will throw if the asset is missing?

packages/app/src/cli/services/dev/extension/server/middlewares.ts
vividviolet: We still will need this. The build directory is a local directory not the dev bundle directory. Assets are compiled locally and then copied over to the dev bundle. The Dev Server serves local assets and we create a timestamp for when the files are updated. If we switch to using the dev bundle then I believe the timestamp will be wrong since it re-creates the bundle every time. This will cause unnecessary updates for consumers.

packages/app/src/cli/models/extensions/specifications/ui_extension.ts
vividviolet: I saw that there was existing code to warn for missing files. I updated to throw instead so we can delete the old custom code to handle this. I think this should error out because the toml config is incorrect if the file doesn't exist. I think it's also simpler this way 

cc @isaacroldan

packages/app/src/cli/services/dev/app-events/app-event-watcher.ts
vividviolet: Rebased on main and removed the changes

packages/app/src/cli/models/extensions/specifications/ui_extension.ts
vividviolet: Fixed! `build_ui` now works with `generatesAssetsManifest`

packages/app/src/cli/services/dev/extension/server/middlewares.ts
```diff
@@ -78,14 +78,13 @@ export function getExtensionAssetMiddleware({devOptions, getExtensions}: GetExte
...
-    const extensionOutputPath = extension.getOutputPathForDirectory(bundlePath)
-
-    const buildDirectory = dirname(extensionOutputPath)
-
-    return fileServerMiddleware(event, {
-      filePath: joinPath(buildDirectory, assetPath),
-    })
+    // Try the extension's build output first (for compiled bundles), then fall back
+    // to the extension's source directory (for static assets like tools, instructions).
+    const buildPath = joinPath(extension.directory, extension.outputRelativePath)
```
vividviolet: @isaacroldan this fixes the issue mentioned [here](https://app.graphite.com/github/pr/Shopify/cli/7252#review-PRR_kwDOHibhHs70jd7n). There was a regression where we were getting updates even when the assets fail to build because we were serving them from the `.dev-bundle` which gets freshly created even when the build fails but it should be using the local assets. When the build fails the old local assets should not be deleted and we should not get a new update event in the websocket

packages/app/src/cli/services/build/steps/include-assets/copy-config-key-entry.ts
```diff
@@ -60,26 +58,36 @@ export async function copyConfigKeyEntry(config: {
...
 
     const destDir = effectiveOutputDir
 
     if (sourceIsDir) {
+      // Glob the source directory (not the destination) to get the accurate file list.
+      // During dev, the include_assets step runs on every rebuild. If we glob the
+      // destination instead, it would pick up files accumulated from previous builds
+      // that may no longer exist in the source, inflating the file count and producing
+      // stale entries in the manifest's pathMap.
+      const sourceFiles = await glob(['**/*'], {cwd: fullPath, absolute: false})
```
vividviolet, line 77: Previously this was looking at copied files which broke when including assets for  the `admin`specificaiton after I fixed the issue where files were getting a new name on every build in dev even when there was no collision. 

Fixed this to match `copy-by-pattern.ts` by looking at the source files, not the files in the destination folder.

packages/app/src/cli/services/build/steps/include-assets/copy-source-entry.ts
```diff
@@ -36,12 +36,17 @@ export async function copySourceEntry(
   }
 
   if (sourceIsDir) {
+    // Glob the source directory (not the destination) to get the accurate file list.
+    // During dev, the include_assets step runs on every rebuild. If we glob the
+    // destination instead, it would pick up files accumulated from previous builds
+    // that may no longer exist in the source, inflating the file count and producing
+    // stale entries in the manifest.
+    const sourceFiles = await glob(['**/*'], {cwd: sourcePath, absolute: false})
```
vividviolet, line 44: Same here, fixed this to match `copy-by-pattern.ts` by looking at the source files, not the files in the destination folder.

packages/app/src/cli/services/dev/extension/server/middlewares.ts
vividviolet: Nope, never mind. I double checked and it still sends an update event, something else is going on

packages/app/src/cli/services/build/client-steps.ts
```diff
@@ -21,15 +21,16 @@ interface IncludeAssetsStep extends BaseStep {
   readonly config: IncludeAssetsConfig
 }
 
+/** Step with typed config specific to bundle_ui. */
+interface BundleUIStep extends BaseStep {
+  readonly type: 'bundle_ui'
+  readonly generatesAssetsManifest?: boolean
+  readonly config?: Record<string, never>
+}
```
isaacroldan, line 30: Shouldn't `generatesAssetsManifest` be part of the `config` ?

packages/app/src/cli/services/build/steps/bundle-ui-step.ts
```diff
@@ -1,11 +1,56 @@
...
 /**
  * Executes a bundle_ui build step.
  *
- * Bundles the UI extension using esbuild, writing output to extension.outputPath.
+ * Bundles the UI extension using esbuild into the extension's local directory
+ * and copies the output to the bundle. When `generatesAssetsManifest` is true,
+ * writes built asset entries (from build_manifest) into manifest.json so
+ * downstream steps can merge on top.
+ */
+export async function executeBundleUIStep(step: LifecycleStep, context: BuildContext): Promise<void> {
```
isaacroldan: you can safely change this from `LifecycleStep` to `BundleUIStep`  
So you don't need to check at runtime for the config fields.

packages/app/src/cli/services/build/steps/include-assets/generate-manifest.ts
```diff
@@ -112,10 +109,51 @@ export async function generateManifestFile(
...
 
+  await mergeManifestEntries(context, manifest)
+}
+
+/**
+ * Merges new entries into the manifest.json for an extension.
+ * Reads the existing manifest if present and deep merges the new entries.
+ * This allows multiple build steps to contribute to the same manifest.
+ */
+export async function mergeManifestEntries(context: BuildContext, entries: {[key: string]: unknown}): Promise<void> {
```
isaacroldan: I think this should have a more explicit name?  
The name suggests an in memory merge, but it is updating the existing file (or creating it if doesn't exist).  
  
`createOrUpdateManifestFile`maybe?

packages/app/src/cli/services/build/extension.ts
isaacroldan, line 1: (Ref Line 96) 🐛 **Bug**: All three analysis passes flagged this: `buildUIExtension` now writes the main bundle to `localOutputPath` (the extension's own directory), but secondary assets at line 96 still write to `dirname(extension.outputPath)`. When called through `buildForBundle` (which both dev and deploy use), `extension.outputPath` is mutated to the bundle directory. This creates a split:

1. Main bundle → written to local dir
2. should_render → written to bundle dir
3. `executeBundleUIStep` copies local dir → bundle dir (only main transfers)
4. Dev middleware checks local dir then source dir — should_render is in neither

This is entangled with the middleware change at `middlewares.ts:83`, which no longer falls back to the bundle directory. Fixing this line so all assets go to the local directory would make the middleware change correct. Without this fix, secondary compiled assets will 404 from the dev server.

**Suggestion:** Build secondary assets to the local directory, consistent with main:

```suggestion
            outputPath: joinPath(dirname(localOutputPath), asset.outputFileName),
```

packages/app/src/cli/services/build/extension.ts
```diff
@@ -122,6 +126,7 @@ export async function buildUIExtension(extension: ExtensionInstance, options: Ex
   const duration = Math.round(performance.now() - startTime)
   const sizeInfo = await formatBundleSize(extension.outputPath)
```
isaacroldan: 🐛 **Bug**: All three analysis passes agree (confidence 0.90): `formatBundleSize(extension.outputPath)` reads from the bundle directory, but the file was just written to `localOutputPath`. The copy to the bundle directory happens later in `executeBundleUIStep`. On the first build, the file doesn't exist at the bundle path yet, so `formatBundleSize` silently returns empty string (it catches errors). On rebuilds, it reads the previous build's file — reporting stale size data. No crash, but the build success message loses bundle size info.

**Suggestion:** 
```suggestion
  const sizeInfo = await formatBundleSize(localOutputPath)
```

packages/app/src/cli/services/build/steps/bundle-ui-step.ts
```diff
@@ -1,11 +1,56 @@
...
+  const localOutputPath = await buildUIExtension(context.extension, context.options)
+  // Copy the locally built files into the bundle
+  await copyFile(dirname(localOutputPath), dirname(context.extension.outputPath))
+
+  if (!('generatesAssetsManifest' in step) || !step.generatesAssetsManifest) return
+
+  const config = context.extension.configuration as Record<string, unknown>
+  const extensionPoints = config.extension_points
+  if (!Array.isArray(extensionPoints) || !extensionPoints.every((ep) => typeof ep === 'object' && ep?.build_manifest))
```
isaacroldan: 💡 **Improvement**: The check `extensionPoints.every(ep => ep?.build_manifest)` is all-or-nothing: if even one extension point doesn't have `build_manifest`, manifest generation is skipped for ALL extension points — including those that do have it. Currently this works because `UIExtensionSchema` always generates `build_manifest` for every extension point. But the guard would silently drop manifest entries if a future configuration omits `build_manifest` for some targets.

**Suggestion:** A filter would be more robust than an all-or-nothing check:

```suggestion
  if (!Array.isArray(extensionPoints)) return

  const pointsWithManifest = extensionPoints.filter(
    (ep): ep is ExtensionPointWithBuildManifest => typeof ep === 'object' && !!ep?.build_manifest,
  )

  const entries = extractBuiltAssetEntries(pointsWithManifest)
```

packages/app/src/cli/services/build/steps/include-assets/generate-manifest.ts
```diff
@@ -112,10 +109,51 @@ export async function generateManifestFile(
...
+
   const manifestPath = joinPath(outputDir, 'manifest.json')
-  await mkdir(outputDir)
-  await writeFile(manifestPath, JSON.stringify(manifest, null, 2))
-  outputDebug(`Generated manifest.json in ${outputDir}\n`, options.stdout)
+
+  let existing: {[key: string]: unknown} = {}
+  if (await fileExists(manifestPath)) {
+    const content = await readFile(manifestPath)
+    existing = JSON.parse(content)
```
isaacroldan: 🐛 **Bug**: Same pattern as `readBundleManifest` above but with the opposite failure mode: `readFile` + `JSON.parse` at line 132-133 run without any error handling. If manifest.json is corrupted from a previous failed build, `JSON.parse` throws an unhandled error that crashes the build. Unlike `readBundleManifest` which silently swallows errors, this one lets them propagate — blocking all subsequent builds until the developer manually deletes the corrupted file.

**Suggestion:** Wrap in a try/catch that falls back to an empty manifest:

```suggestion
  if (await fileExists(manifestPath)) {
    try {
      const content = await readFile(manifestPath)
      existing = JSON.parse(content)
    } catch {
      outputDebug(`Warning: could not parse existing manifest.json, starting fresh\n`, context.options.stdout)
    }
  } else {
```

packages/app/src/cli/services/dev/extension/payload.ts
```diff
@@ -126,15 +128,16 @@ async function getExtensionPoints(extension: ExtensionInstance, url: string) {
...
         }
 
         return {
           ...payload,
-          ...(await mapBuildManifestToPayload(
-            extensionPoint.build_manifest,
-            extensionPoint as NewExtensionPointSchemaType & {build_manifest: BuildManifest},
+          ...(await mapManifestAssetsToPayload(
+            manifestEntry,
+            extensionPoint as unknown as NewExtensionPointSchemaType,
```
isaacroldan: 🎨 **Code Style**: The cast `extensionPoint as unknown as NewExtensionPointSchemaType` uses `as unknown` to bypass TypeScript's type checking entirely. Since `DevNewExtensionPointSchema` extends `Omit<NewExtensionPointSchemaType, 'intents'>`, the core fields already match. The `as unknown` escape hatch suppresses potentially useful type errors that would catch real incompatibilities if these types diverge.

**Suggestion:** Consider widening `mapManifestAssetsToPayload`'s parameter type to accept `DevNewExtensionPointSchema` directly, or use a direct cast without the `unknown` intermediary.

packages/app/src/cli/services/dev/extension/payload.ts
```diff
@@ -147,35 +150,96 @@ async function getExtensionPoints(extension: ExtensionInstance, url: string) {
...
+ */
+async function readBundleManifest(
+  buildDirectory: string,
+): Promise<{[target: string]: {[assetName: string]: unknown}} | null> {
+  try {
+    const manifestPath = joinPath(buildDirectory, 'manifest.json')
+    const content = await readFile(manifestPath)
+    return JSON.parse(content)
+    // eslint-disable-next-line no-catch-all/no-catch-all
+  } catch {
```
isaacroldan: 💡 **Improvement**: The `catch` block catches all errors including `JSON.parse` failures from corrupted or partially-written manifest.json files. A `readFile` ENOENT (file doesn't exist) should return null — that's expected when no manifest has been generated yet. But a `SyntaxError` from `JSON.parse` means the file exists but is corrupted (e.g., partial write during a concurrent rebuild race), and silently returning null hides that problem. The developer would see missing assets with no explanation of why.

**Suggestion:** Distinguish file-not-found from parse errors:

```suggestion
  try {
    const manifestPath = joinPath(buildDirectory, 'manifest.json')
    const content = await readFile(manifestPath)
    return JSON.parse(content)
    // eslint-disable-next-line no-catch-all/no-catch-all
  } catch (error: unknown) {
    if (error instanceof SyntaxError) {
      throw new Error(`Invalid manifest.json in ${buildDirectory}: ${error.message}`)
    }
    return null
```

packages/app/src/cli/services/dev/extension/server/middlewares.ts
```diff
@@ -78,14 +78,13 @@ export function getExtensionAssetMiddleware({devOptions, getExtensions}: GetExte
...
-    return fileServerMiddleware(event, {
-      filePath: joinPath(buildDirectory, assetPath),
-    })
+    // Try the build output directory first (for compiled assets like dist/handle.js),
+    // then fall back to the extension's source directory (for static assets like tools, instructions).
+    const builtAssetPath = joinPath(dirname(joinPath(extension.directory, extension.outputRelativePath)), assetPath)
+    if (await fileExists(builtAssetPath)) {
+      return fileServerMiddleware(event, {filePath: builtAssetPath})
+    }
+    return fileServerMiddleware(event, {filePath: joinPath(extension.directory, assetPath)})
```
isaacroldan: 🔒 **Security**: The new middleware fallback serves files from `extension.directory` (the developer's source) using `assetPath` from the router. A crafted request like `/extensions/{id}/assets/../../.env` could read arbitrary files from the project directory. `getAppAssetsMiddleware` in the same file (lines 144-147) already has proper path traversal validation using `resolvePath` and `startsWith`. The inconsistency is worth addressing — the previous middleware only served from the build output temp directory, which had narrower scope. Now that the source directory is exposed, the same protection pattern should apply. The blast radius is limited to the developer's own machine (local dev server), but it's a gap compared to the established pattern.

**Suggestion:** Add path traversal validation matching the existing pattern from `getAppAssetsMiddleware`:

```suggestion
    // Try the build output directory first (for compiled assets like dist/handle.js),
    // then fall back to the extension's source directory (for static assets like tools, instructions).
    const builtAssetPath = joinPath(dirname(joinPath(extension.directory, extension.outputRelativePath)), assetPath)
    if (await fileExists(builtAssetPath)) {
      return fileServerMiddleware(event, {filePath: builtAssetPath})
    }
    const resolvedDir = resolvePath(extension.directory)
    const resolvedAssetPath = resolvePath(extension.directory, assetPath)
    if (!resolvedAssetPath.startsWith(resolvedDir)) {
      return sendError(event, {statusCode: 403, statusMessage: 'Path traversal is not allowed'})
    }
    return fileServerMiddleware(event, {filePath: resolvedAssetPath})
```

isaacroldan (commented):
---

packages/app/src/cli/services/build/extension.ts
vividviolet: Good catch!

packages/app/src/cli/services/dev/extension/server/middlewares.ts
vividviolet: The suggestion is not quite right but I will fix it

isaacroldan (approved):

PR #7152, opened 2026-04-01
Watch admin static_root and serve assets via dev server WebSocket

melissaluu:
## Summary

Adds end-to-end support for watching admin extension `static_root` directories during `shopify app dev`. When files change in the configured `static_root` directory, the dev server detects the change, updates asset timestamps in the payload store, and broadcasts the update over WebSocket so connected clients can react.

This PR does not include exposing CSP content for the app.

## What changed

**File watching pipeline** — Extended the existing `FileWatcher` → `AppEventWatcher` → dev server pipeline with a new `app_asset_updated` event type. The `FileWatcher` resolves `static_root` from the admin extension config, watches that directory, and emits `app_asset_updated` events on file changes. A new `AppAssetUpdatedHandler` propagates these as `AppEvent`s with an `appAssetsUpdated` flag.

**`dist/` ignore exemption** — Replaced the static `**/dist/**` chokidar ignore pattern with a function that allows `dist/` paths through when they fall under a watched app asset directory (since `static_root` may point to a `dist/` folder).

**Payload store** — Added `app.assets` to the WebSocket payload model: a `Record<string, { url, lastUpdated }>`. New methods `updateAppAssetTimestamp()` and `updateAppAssets()` manage timestamps and emit updates to WebSocket clients.

**HTTP serving** — Added a parameterized route at `/extensions/assets/:assetKey/:filePath` that serves files from the resolved asset directory.

**Admin spec** — Set `experience: 'configuration'` so the admin extension loads from `shopify.app.toml`. Added a proper zod schema (`{ admin: { static_root? } }`) so the extension is only instantiated when `[admin]` is present in the config. Also added optional `schema` support to `createContractBasedModuleSpecification` to enable this.

## Test plan

- [x] `loader.test.ts` — 129 tests pass, including the 3 that were failing due to the admin spec change
- [x] `extension.test.ts` — `resolveAppAssets()` returns correct map; `appAssets` passed through to HTTP server; `getAppAssets` callback works
- [x] `store.test.ts` — `updateAppAssetTimestamp` updates timestamp and emits; `updateAppAssets` rebuilds or removes assets; raw payload populates `app.assets` when configured
- [x] `middlewares.test.ts` — `getAppAssetsMiddleware` serves files for valid keys, returns 404 for unknown keys
- [x] `file-watcher.test.ts` — Updated chokidar ignored pattern assertion for the new `dist/` filter function
- [x] `app-event-watcher.test.ts` — 19 tests pass


packages/app/src/cli/services/dev/extension/server/middlewares.ts
```diff
@@ -134,6 +134,22 @@ export const devConsoleAssetsMiddleware = defineEventHandler(async (event) => {
...
+
+    const directory = appAssets[assetKey]
+
+    if (!directory) {
+      return sendError(event, {statusCode: 404, statusMessage: `No app assets configured for key: ${assetKey}`})
+    }
+
+    return fileServerMiddleware(event, {
+      filePath: joinPath(directory, filePath),
+    })
```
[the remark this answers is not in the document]
melissaluu: Will not happen with local dev server

packages/app/src/cli/models/extensions/specifications/admin.ts
```diff
@@ -1,8 +1,20 @@
...
+      static_root: zod.string().optional(),
+    })
+    .optional(),
+})
 
 const adminSpecificationSpec = createContractBasedModuleSpecification({
   identifier: 'admin',
   uidStrategy: 'single',
+  experience: 'configuration',
+  schema: AdminSchema as unknown as ZodSchemaType<BaseConfigType>,
```
isaacroldan, line 17: If you are providing a schema here, then you shouldn't use `createContractbasedModuleSpecification`, because it's not using contracts anymore.   
Use the generic `createExtensionSpecification`

packages/app/src/cli/models/extensions/specification.ts
```diff
@@ -294,11 +294,12 @@ export function createContractBasedModuleSpecification<TConfiguration extends Ba
     | 'clientSteps'
     | 'experience'
     | 'transformRemoteToLocal'
+    | 'schema'
   >,
 ) {
   return createExtensionSpecification({
     identifier: spec.identifier,
-    schema: zod.any({}) as unknown as ZodSchemaType<TConfiguration>,
+    schema: spec.schema ?? (zod.any({}) as unknown as ZodSchemaType<TConfiguration>),
```
isaacroldan, line 302: by definition, contract based modules can't provide a local schema.

isaacroldan (changes requested):
Let's refactor and treat it as is `identifier === 'admin'` is a forbidden pattern.  The file-watcher shouldn't know about specific extensions in any case.

isaacroldan:
Why do we need to treat app assets as a different event/output/property instead of just the output of another extension? This PR introduces the concept of App Assets everywhere, is it necessary? are App Assets just the output of an admin extension? can they come from a different extension? Seems like we are making the output of this specific extension a first-class element of the app, instead of treating it like any other extension.

vividviolet:
> This PR introduces the concept of App Assets everywhere, is it necessary? are App Assets just the output of an admin extension? can they come from a different extension?

@isaacroldan I'm not sure what's the best way to handle this would be but this is different than other extensions because the settings comes from the `shopify.app.toml`. The way we fetch the data too will be from the app itself rather than an extension so it would be nice to keep the assets on the app's payload if possible.

PR #7147, opened 2026-04-01
E2E: infra refactor + app tests update

phyllis-sy-wu:
### WHY are these changes introduced?

The main goal of this PR is to strengthen the E2E infrastructure by addressing four areas:

1. **Reusable helpers** — Added composable CLI helpers (`createApp`, `buildApp`, `deployApp`, etc.) and browser helpers (`navigateToDashboard`, `uninstallApp`, `deleteApp`, etc.) that future tests can build on.

2. **Self-contained app tests** — The existing E2E app tests depend on pre-provisioned apps via `SHOPIFY_FLAG_CLIENT_ID`. This makes tests fragile — they break if the pre-provisioned app is deleted, modified, or its org permissions change. This PR migrates app-relevant tests to create their own apps during the test run using `--organization-id`, removing the dependency on fixed `SHOPIFY_FLAG_CLIENT_ID` values.

3. **Automatic cleanup** — Tests that create apps pollute the organization if not cleaned up. Each test now cleans up after itself in a `finally` block — deleting local files and removing the app from the partner dashboard (uninstall from stores + delete). This is best-effort (with DEBUG logging) so cleanup failures never fail the test.

4. **Auth robustness** — The `SHOPIFY_CLI_PARTNERS_TOKEN` is not passed to the E2E test step in CI. When present, the CLI prioritizes it over OAuth and exchanges it for an App Management API token that lacks permission to create apps (403). By keeping it out of the E2E environment, all tests use the OAuth session consistently.

### WHAT is this pull request doing?

**New fixture chain:** `envFixture → cliFixture → browserFixture → authFixture → appTestFixture`

- **`setup/browser.ts`** (new) — Worker-scoped Chromium browser page, persistent across tests for OAuth login and dashboard cleanup. Exports `BrowserContext` type and `navigateToDashboard` helper (generic dashboard navigation with account picker handling and 500 retry).
- **`setup/auth.ts`** (modified) — Now extends `browserFixture`. Performs OAuth login using the shared browser page.
- **`setup/cli.ts`** (modified) — Exports `CLIContext` type (used by app helpers).
- **`setup/app.ts`** (rewritten) — Replaces the old `AppScaffold` class with composable helper functions:
  - **CLI helpers:** `createApp`, `generateExtension`, `buildApp`, `deployApp`, `appInfo`, `functionBuild`, `functionRun`, `versionsList`, `configLink` — all take a single object arg, all use `FRESH_APP_ENV` to strip `CLIENT_ID`
[...]

PR #7162, opened 2026-04-02
E2E: migrate tests to use new infra

phyllis-sy-wu:

This is a follow-up to #7147 (E2E infra refactor). That PR migrated the app tests (scaffold, deploy, dev-server) to create fresh apps from scratch. This PR completes the migration by converting the remaining 4 test files that still depended on pre-provisioned apps via `SHOPIFY_FLAG_CLIENT_ID`.

After this PR, **all E2E tests are self-contained** — each test creates its own app, runs its assertions, and cleans up. No test depends on a pre-provisioned app or a fixed `SHOPIFY_FLAG_CLIENT_ID`.

**TOML helpers in `setup/app.ts`:**
- `extractClientId(appDir)` — reads `client_id` from `shopify.app.toml` using `@iarna/toml` parser (same parser cli-kit uses internally via `decodeToml`)
- `injectFixtureToml(appDir, template, name)` — patches `client_id` and `name` into fixture TOML using `@shopify/toml-patch` (`updateTomlValues`), preserving comments and formatting

**TOML editing in tests:**
- `dev-hot-reload.spec.ts` — scopes edit now uses `updateTomlValues` from `@shopify/toml-patch` instead of string replace

**Shared timeout constants (`setup/constants.ts`):**
- `TEST_TIMEOUT` — per-test Playwright timeouts: `default` (3 min), `long` (10 min)
- `CLI_TIMEOUT` — CLI command execution: `short` (1 min), `medium` (3 min), `long` (5 min)
- `BROWSER_TIMEOUT` — browser interactions: `short` (1s), `medium` (3s), `long` (5s), `max` (60s)
- All hardcoded timeouts across setup/, helpers/, and test files replaced with named constants

**Migrated tests:**
- **`dev-hot-reload.spec.ts`** — creates fresh app, injects fully populated TOML fixture, tests hot reload (edit scopes, create/delete extensions mid-dev)
- **`multi-config-dev.spec.ts`** — creates fresh app, injects fixture TOML, creates staging config with same `client_id`, tests `-c` flag config selection
- **`toml-config.spec.ts`** — creates fresh app, injects fixture TOML, tests deploy and dev with fully populated config
- **`toml-config-invalid.spec.ts`** — uses hardcoded dummy `client_id` (CLI rejects invalid TOMLs at validation before any API call, so no real app needed)

**Cleaned up:**
- Deleted `setup/toml-app.ts` — the `tomlAppFixture` is no longer used
- Removed `SHOPIFY_FLAG_CLIENT_ID` and `E2E_SECONDARY_CLIENT_ID` from CI workflow
- Stripped `SHOPIFY_FLAG_CLIENT_ID` globally in `env.ts` so it can't leak from local `.env` files
- Removed `clientId`, `secondaryClientId`, `partnersToken` from `E2EEnv` interface
- Removed `FRESH_APP_ENV` from `setup/app.ts`
- Updated TOML fixtures — `data/valid-app/shopify.app.toml` uses valid placeholder values patched at runtime by `updateTomlValues`
- Global timeout tuned to 20 min (CI runs at ~16-19 min)

### How to test your changes?

```bash
DEBUG=1 pnpm --filter e2e exec playwright test
```

Ensure `SHOPIFY_FLAG_CLIENT_ID` is NOT in `.env` (no longer used).

### Measuring impact

### Checklist


phyllis-sy-wu:
* **#7162** <a href="https://app.graphite.com/github/pr/Shopify/cli/7162?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#7171](https://github.com/Shopify/cli/pull/7171) <a href="https://app.graphite.com/github/pr/Shopify/cli/7171?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#7255](https://github.com/Shopify/cli/pull/7255) <a href="https://app.graphite.com/github/pr/Shopify/cli/7255?utm_source=stack-comment-icon" target="_blank">Graphite</a>) 👈 <a href="https://app.graphite.com/github/pr/Shopify/cli/7162?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#7147** <a href="https://app.graphite.com/github/pr/Shopify/cli/7147?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

packages/e2e/setup/app.ts
```diff
@@ -81,6 +75,29 @@ export async function createApp(ctx: {
...
+}
+
+/**
+ * Overwrite a created app's shopify.app.toml with a fixture TOML template.
+ * The template should contain `__CLIENT_ID__` and `__NAME__` placeholders which get
+ * replaced with the app's real client_id and the provided name.
+ */
+export function injectFixtureToml(appDir: string, fixtureTomlContent: string, name: string): void {
+  const clientId = extractClientId(appDir)
+  const toml = fixtureTomlContent.replace('__CLIENT_ID__', clientId).replace('__NAME__', name)
```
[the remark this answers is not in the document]
phyllis-sy-wu: Addressed in latest update.

packages/e2e/setup/app.ts
```diff
@@ -46,7 +40,7 @@ export async function createApp(ctx: {
 
   const result = await cli.execCreateApp(args, {
     // Strip CLIENT_ID so the CLI creates a new app instead of linking to a pre-existing one
-    env: {FORCE_COLOR: '0', ...FRESH_APP_ENV},
+    env: {FORCE_COLOR: '0'},
```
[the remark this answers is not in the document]
phyllis-sy-wu: Addressed in latest update.

.github/workflows/tests-pr.yml
```diff
@@ -219,7 +219,7 @@ jobs:
     name: 'E2E tests'
     if: github.event.pull_request.head.repo.full_name == github.repository
     runs-on: ubuntu-latest
-    timeout-minutes: 15
+    timeout-minutes: 30
```
ryancbahan, line 222: why double?

packages/e2e/setup/app.ts
```diff
@@ -46,8 +40,8 @@ export async function createApp(ctx: {
   if (ctx.flavor) args.push('--flavor', ctx.flavor)
 
   const result = await cli.execCreateApp(args, {
-    // Strip CLIENT_ID so the CLI creates a new app instead of linking to a pre-existing one
-    env: {FORCE_COLOR: '0', ...FRESH_APP_ENV},
+    // Disable color output and strip CLIENT_ID to prevent leaking from parent process.env
+    env: {FORCE_COLOR: '0', SHOPIFY_FLAG_CLIENT_ID: undefined},
```
ryancbahan: let's just remove SHOPIFY_FLAG_CLIENT_ID

packages/e2e/setup/app.ts
```diff
@@ -82,6 +76,33 @@ export async function createApp(ctx: {
...
 }
 
+// ---------------------------------------------------------------------------
+// Fixture helpers — TOML manipulation for test setup
+// ---------------------------------------------------------------------------
+
+/**
+ * Read the client_id from a shopify.app.toml file.
+ */
+export function extractClientId(appDir: string): string {
```
ryancbahan, line 88: won't block on this, but we can use `@shopify/toml-patch` for this in the same manner that the cli does to extract toml fields. as much as possible we should avoid writing our own parsers

packages/e2e/setup/app.ts
```diff
@@ -82,6 +76,33 @@ export async function createApp(ctx: {
...
+  const toml = fs.readFileSync(path.join(appDir, 'shopify.app.toml'), 'utf8')
+  const match = toml.match(/client_id\s*=\s*"([^"]+)"/)
+  if (!match?.[1]) {
+    throw new Error(`Could not find client_id in ${path.join(appDir, 'shopify.app.toml')}`)
+  }
+  return match[1]
+}
+
+/**
+ * Overwrite a created app's shopify.app.toml with a fixture TOML template.
```
ryancbahan, line 99: same comment re: @shopify/toml-patch

packages/e2e/setup/app.ts
```diff
@@ -113,7 +134,7 @@ export async function deployApp(
   if (ctx.version) args.push('--version', ctx.version)
   if (ctx.message) args.push('--message', ctx.message)
   if (ctx.config) args.push('--config', ctx.config)
-  return ctx.cli.exec(args, {env: FRESH_APP_ENV, timeout: 5 * 60 * 1000})
+  return ctx.cli.exec(args, {timeout: 5 * 60 * 1000})
```
ryancbahan: it's probably a good time to register a default timeout and add constants for areas that need more time

packages/e2e/tests/dev-hot-reload.spec.ts
```diff
@@ -60,139 +35,149 @@ description = "E2E test trigger"
...
+        env: {CI: ''},
+      })
+
+      try {
+        await proc.waitForOutput(READY_MESSAGE, 3 * 60 * 1000)
+
+        // Edit scopes in the TOML to trigger a reload
+        const tomlPath = path.join(appDir, 'shopify.app.toml')
+        const original = fs.readFileSync(tomlPath, 'utf8')
+        fs.writeFileSync(
```
ryancbahan: same comment re: toml patch

packages/e2e/tests/dev-hot-reload.spec.ts
```diff
@@ -60,139 +35,149 @@ description = "E2E test trigger"
...
+        expect(exitCode, `dev exited with non-zero code. Output:\n${output}`).toBe(0)
+      } catch (error) {
+        console.error(`[hot-reload app-config] Captured PTY output:\n${proc.getOutput()}`)
+        throw error
+      } finally {
+        proc.kill()
+      }
     } finally {
-      proc.kill()
+      fs.rmSync(parentDir, {recursive: true, force: true})
```
ryancbahan, line 84: i think cli also has tmp dir utils for setting up ephemeral temporary directories that we can leverage here

packages/e2e/playwright.config.ts
```diff
@@ -15,7 +15,7 @@ export default defineConfig({
   maxFailures: isCI ? 3 : 0, // Stop early in CI after 3 failures
   reporter: isCI ? [['html', {open: 'never'}], ['list']] : [['list']],
   timeout: 3 * 60 * 1000, // 3 minutes per test
-  globalTimeout: 15 * 60 * 1000, // 15 minutes total
+  globalTimeout: 30 * 60 * 1000, // 30 minutes total
```
ryancbahan: i'd encourage us to increment these in smaller portions, and try to use data when available to inform the decision

ryancbahan (commented):
Looking good overall! i think there are a few quality of life things that would be good to take a pass on, then ship. lmk what you think.

.github/workflows/tests-pr.yml
phyllis-sy-wu: Locally, the full suite takes ~12-13 min. In CI it went over the original 15 min limit, 19 min on this PR, mainly due to per-test teardown. To prevent CI from killing the run mid-suite, I increased the timeout. But I agree 30 min might be too much. Will set it to 20 min now.

The follow-up stacked PR #7171 optimizes the teardown process a bit, which should bring CI times down further, will evaluate this again on that PR!

packages/e2e/playwright.config.ts
phyllis-sy-wu: Same reply [here](https://github.com/Shopify/cli/pull/7162#discussion_r3052610333)

packages/e2e/setup/app.ts
phyllis-sy-wu: Thanks for spottig this! Removed, but added `SHOPIFY_FLAG_CLIENT_ID: undefined` in `pe2e/setup/env.ts` to avoid errors if `SHOPIFY_FLAG_CLIENT_ID` is set in local `.env`

packages/e2e/setup/app.ts
phyllis-sy-wu: Thanks for the suggestions! Now added `setup/constants.ts`. All magic numbers in e2e are replaced. Few older files are not processed because the files are planned to be removed completely later.

packages/e2e/setup/app.ts
phyllis-sy-wu: Thanks for the suggestions. Switched `extractClientId` from regex to `@iarna/toml`, which parses TOML into a JS object so we can access `client_id` directly. `@shopify/toml-patch` only works with TOML strings in/out — it doesn't expose parsed values as JS objects, so it can't be used for reading fields.

packages/e2e/setup/app.ts
phyllis-sy-wu: Thanks for the suggestions. Now used `toml-patch` for `injectFixtureToml`. The fixture TOML now uses valid placeholder values instead of `__CLIENT_ID__`/`__NAME__` — toml-patch patches them surgically at runtime, preserving comments and formatting.

packages/e2e/tests/dev-hot-reload.spec.ts
phyllis-sy-wu: Thanks for the suggestions. Updated!

packages/e2e/tests/dev-hot-reload.spec.ts
phyllis-sy-wu: Thanks for the suggestions. Looked into this — cli-kit has `inTemporaryDirectory` and `tempDirectory` in `@shopify/cli-kit/node/fs`. Our env fixture already creates a single worker-scoped temp root (`env.tempDir`) that holds both XDG dirs and per-test app dirs. This gives us one place to inspect during debugging and one cleanup path on worker teardown. cli-kit's `inTemporaryDirectory` would scatter dirs across `/tmp` and always cleans up in its `finally` block — no way to preserve files for debugging. So keeping the current approach for now.

ryancbahan (approved):

PR #7173, opened 2026-04-02
Add glob pattern support for --environment flag

### LINKED ISSUE #6481: [Feature] Support glob patterns in theme environment flags

graygilmore:
## Summary

- Adds glob pattern support (e.g. `--environment "*-production"`) so users can target multiple environments at once instead of specifying each individually
- Expands patterns at a single point in `base-command.ts` before single/multi environment routing, keeping downstream consumers (like theme-command) unchanged
- Uses `minimatch` (already a dependency) for pattern matching with support for `*`, `?`, `[abc]`, and `{a,b}` syntax

Closes https://github.com/Shopify/cli/issues/6481

Supersedes https://github.com/Shopify/cli/pull/6493 with a cleaner single-expansion-point architecture.

### Setup

Create a `shopify.theme.toml` with multiple environments:

```toml
[environments.us-production]
store = "us-store.myshopify.com"

[environments.eu-production]
store = "eu-store.myshopify.com"

[environments.staging]
store = "staging-store.myshopify.com"
```

### Test cases

1. **Glob matching multiple envs**: `shopify theme pull -e "*-production"` — should list both `us-production` and `eu-production` as matched environments
2. **Single glob match**: `shopify theme pull -e "staging"` — should apply the staging environment flags directly
3. **No match warning**: `shopify theme pull -e "*-sandbox"` — should warn that no environments matched
4. **Quote requirement**: Patterns with `*` must be wrapped in quotes to prevent shell expansion

## Test plan

- [x] `environments.test.ts` — 19 tests covering `getEnvironmentNames` and `expandEnvironmentPatterns`
- [x] `base-command.test.ts` — 34 tests including 4 new glob expansion tests
- [x] `theme-command.test.ts` — 24 tests (regression, no changes needed)


packages/theme/src/cli/flags.ts
```diff
@@ -44,14 +46,12 @@ export const themeFlags = {
   }),
   environment: Flags.string({
     char: 'e',
-    description: 'The environment to apply to the current command.',
+    description: `The environment to apply to the current command. Supports glob patterns (e.g. "*-production"). ${globQuotesDescription}`,
```
graygilmore: 💡 **Improvement**: The updated `--environment` description is shared across all theme commands and propagates into generated docs. Several theme commands still reject multiple resolved environments, so the current wording can read as if glob-based fan-out is uniformly supported when it is only available on commands with multi-environment routing.

**Suggestion:** Qualify the description so the generated help matches actual command behavior:
```suggestion
    description: `The environment to apply to the current command. Supports glob patterns (e.g. "*-production"). Patterns that resolve to multiple environments only work on commands that support multiple environments. ${globQuotesDescription}`,
```

PR #7208, opened 2026-04-06
Run web build commands during deploy for hosted apps

MitchLillie:
Closes https://github.com/shop/issues-admin-extensibility/issues/2398

## Summary

- `bundleAndBuildExtensions` now runs `web.toml` `build` commands concurrently alongside extension bundling during `shopify app deploy`
- Scoped to webs that explicitly define a `build` command — extension-only apps and dev-only webs are unaffected
- Respects `--no-build` / `skipBuild` flag

**Root cause:** `deploy` called `bundleAndBuildExtensions` (extensions only) but never invoked `buildWeb('build', ...)`, unlike `shopify app build` which does. Users had to manually run `build` before `deploy`.

Closes shop/issues-admin-extensibility#2398

## Manual test instructions

**Setup:** You need an app with a `web/` directory and a `web.toml` that defines a `build` command:

```toml
# web/web.toml
[commands]
dev = "npm run dev"
build = "npm run build"
```

**Test: build command runs on deploy**
1. Make a change to your web app source that only reflects after a build
2. Run `shopify app deploy`
3. Confirm your `build` command output appears in the concurrent output alongside extensions
4. Confirm the deploy succeeds

**Test: `--no-build` skips web build**
1. Run `shopify app deploy --no-build`
2. Confirm the `build` command does NOT run (no web-prefixed process in output)

**Test: extension-only apps unaffected**
1. Use an app whose `web.toml` has no `build` command (or no webs at all)
2. Run `shopify app deploy` — confirm no new web-prefixed process appears in output

## Test plan

- [x] `runs web build command concurrently with extensions when build command is defined`
- [x] `skips web build for webs without a build command defined`
- [x] `skips web build command when skipBuild is true`
- [ ] Manual test with a hosted app
- [ ] Manual test with `--no-build`

### DIFF
--- a/packages/app/src/cli/services/deploy/bundle.ts
+++ b/packages/app/src/cli/services/deploy/bundle.ts
@@ -1,6 +1,7 @@
 import {AppInterface, AppManifest} from '../../models/app/app.js'
 import {Identifiers} from '../../models/app/identifiers.js'
 import {installJavy} from '../function/build.js'
+import buildWeb from '../web.js'
 import {compressBundle, writeManifestToBundle} from '../bundle.js'
 import {AbortSignal} from '@shopify/cli-kit/node/abort'
 import {mkdir, rmdir} from '@shopify/cli-kit/node/fs'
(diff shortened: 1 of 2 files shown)


packages/app/src/cli/services/deploy/bundle.ts
```diff
@@ -30,9 +31,21 @@ export async function bundleAndBuildExtensions(options: BundleOptions) {
...
   }
 
+  const webBuildProcesses = options.skipBuild
+    ? []
+    : options.app.webs
+        .filter((web) => web.configuration.commands.build)
+        .map((web) => ({
+          prefix: ['web', ...web.configuration.roles].join('-'),
+          action: async (stdout: Writable, stderr: Writable, signal: AbortSignal) => {
+            await buildWeb('build', {web, stdout, stderr, signal})
```
isaacroldan, line 42: for consistency with how extensionProcesses work and to make it easier to read, let's check the `skipBuild` here:  


```typescript
      action: async (stdout: Writable, stderr: Writable, signal: AbortSignal) => {
        if (options.skipBuild) return
        await buildWeb('build', {web, stdout, stderr, signal})
      },
```

packages/app/src/cli/services/deploy/bundle.ts
```diff
@@ -30,9 +31,21 @@ export async function bundleAndBuildExtensions(options: BundleOptions) {
...
+          },
+        }))
+
   await renderConcurrent({
-    processes: options.app.allExtensions.map((extension) => {
-      return {
+    processes: [
+      ...webBuildProcesses,
+      ...options.app.allExtensions.map((extension) => ({
```
isaacroldan: and for consistency also, let's extract the `extensionProcesses` generation outside of the `renderConcurrent` call so we end up with:  


```typescript
  await renderConcurrent({
    processes: [webBuildProcesses, extensionProcesses].flat(),
    showTimestamps: false,
  })
```

isaacroldan (approved):
Approving but please apply the suggested formatting changes :)

PR #7217, opened 2026-04-07
Reapply: central unmount for ink render tree + fix render lifecycle race

ryancbahan:
## Summary

Reapplies #7153 (reverted in #7157) with a fix for the two bugs that caused the revert:

1. **Loading bar stayed visible** after `deploy` / `store execute` completed
2. **Last line of output was cut off** in some commands

### Root cause

`renderTasks` and `renderSingleTask` resolved their promises via `onComplete={resolve}` — a callback that fires inside the React tree **before** ink unmounts. The caller resumed and wrote to stdout while ink was still tearing down, causing ink's cleanup to overwrite the caller's output.

### Fix

Changed `renderTasks` and `renderSingleTask` to use the same pattern the prompt functions (`renderSelectPrompt`, `renderTextPrompt`, etc.) already use:

```ts
// Before (broken): promise resolves before ink unmounts
return new Promise((resolve, reject) => {
  render(<Component onComplete={resolve} />, options).catch(reject)
})

// After (correct): awaits full ink lifecycle before returning
let result: T
await render(<Component onComplete={(v) => { result = v }} />, options)
return result!
```

This guarantees the caller only resumes after ink has fully unmounted and flushed its final frame to stdout. This aligns with Ink's rendering approach when concurrent mode is disabled.

## Tophatting

- [x] All existing UI tests pass (17/17)
- [x] `shopify app deploy` — loading bar disappears after completion
- [x] `pnpm shopify store execute` — loading bars disappear after completion
- [x] Last line of output is not cut off in any command
- [x] `shopify app init` — no render issues
- [x] `shopify app generate extension` — no render issues
- [x] `shopify kitchen-sink async` — loading bar clears properly
- [x] Select/text prompts still work (e.g. `shopify app dev` org selection)

### DIFF
--- a/packages/cli-kit/src/private/node/ui/hooks/use-abort-signal.ts
+++ b/packages/cli-kit/src/private/node/ui/hooks/use-abort-signal.ts
@@ -1,32 +1,28 @@
 import {AbortSignal} from '../../../../public/node/abort.js'
-import {useApp} from 'ink'
-import {useLayoutEffect, useState} from 'react'
+import {useComplete} from '../../ui.js'
+import {useEffect, useLayoutEffect, useState} from 'react'
 
 const noop = () => Promise.resolve()
 
 export default function useAbortSignal(abortSignal?: AbortSignal, onAbort: (error?: unknown) => Promise<void> = noop) {
-  const {exit: unmountInk} = useApp()
+  const complete = useComplete()
   const [isAborted, setIsAborted] = useState(false)
 
   useLayoutEffect(() => {
     abortSignal?.addEventListener('abort', () => {
       const abortWithError = abortSignal.reason.message === 'AbortError' ? undefined : abortSignal.reason
       onAbort(abortWithError)
-        .then(() => {
-          setIsAborted(true)
-          // Defer unmounting to the next setImmediate so React 19 can flush
-          // batched state updates before the tree is torn down.  React 19's
-          // scheduler also uses setImmediate in Node.js (check phase), and
-          // since it was queued first (by setIsAborted above), it renders
-          // before this callback fires (FIFO within the check phase).
-          // NOTE: setTimeout(fn, 0) is NOT safe here because its timers-phase
-          // fires BEFORE the check phase on slow CI machines where >1 ms has
-          // elapsed, causing unmount to race ahead of the render.
-          setImmediate(() => unmountInk(abortWithError))
-        })
+        .then(() => setIsAborted(true))
         .catch(() => {})
     })
   }, [])
+
+  useEffect(() => {
+    if (isAborted) {
+      const abortWithError = abortSignal?.reason?.message === 'AbortError' ? undefined : abortSignal?.reason
+      complete(abortWithError)
+    }
+  }, [isAborted])
 
   return {isAborted}
 }
--- a/packages/cli-kit/src/private/node/ui/hooks/use-async-and-unmount.ts
+++ b/packages/cli-kit/src/private/node/ui/hooks/use-async-and-unmount.ts
@@ -1,5 +1,5 @@
-import {useApp} from 'ink'
-import {useEffect} from 'react'
+import {useComplete} from '../../ui.js'
+import {useEffect, useState} from 'react'
 
 interface Options {
   onFulfilled?: () => unknown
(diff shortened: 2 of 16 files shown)


ryancbahan:
* **#7217** <a href="https://app.graphite.com/github/pr/Shopify/cli/7217?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Shopify/cli/7217?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/cli-kit/src/private/node/ui/hooks/use-abort-signal.ts
```diff
@@ -1,32 +1,28 @@
...
-          setImmediate(() => unmountInk(abortWithError))
-        })
+        .then(() => setIsAborted(true))
         .catch(() => {})
     })
   }, [])
 
+  useEffect(() => {
+    if (isAborted) {
+      const abortWithError = abortSignal?.reason?.message === 'AbortError' ? undefined : abortSignal?.reason
```
gonzaloriestra, line 22: 💡 **Improvement**: The `abortWithError` value is computed twice: once in the abort listener (line 13, `abortSignal.reason.message`) and again in the `useEffect` (line 22, `abortSignal?.reason?.message`). The two derivations use subtly different optional chaining, creating a maintenance hazard if someone edits one without the other. Every other completion hook in this PR (`SingleTask`, `ConcurrentOutput`, `use-async-and-unmount`) avoids this by storing the error in a `{error?: Error}` state variable and reading from state in the completion effect. This hook is the sole outlier.

**Suggestion:** Consider storing the abort error in state, matching the pattern used throughout the PR:

```typescript
const [abortResult, setAbortResult] = useState<{error?: Error} | null>(null)

useLayoutEffect(() => {
  abortSignal?.addEventListener('abort', () => {
    const abortWithError = abortSignal.reason.message === 'AbortError' ? undefined : abortSignal.reason
    onAbort(abortWithError)
      .then(() => setAbortResult({error: abortWithError}))
      .catch(() => {})
  })
}, [])

useEffect(() => {
  if (abortResult !== null) {
    complete(abortResult.error)
  }
}, [abortResult, complete])

return {isAborted: abortResult !== null}
```

This eliminates the duplicate derivation and aligns with SingleTask/ConcurrentOutput/use-async-and-unmount.

packages/cli-kit/src/private/node/ui/hooks/use-abort-signal.ts
```diff
@@ -1,32 +1,28 @@
...
+      complete(abortWithError)
+    }
+  }, [isAborted])
```
gonzaloriestra, line 25: 🎨 **Code Style**: The `useEffect` at line 20 references `complete` but the dependency array at line 25 only lists `[isAborted]`. Compare with `ConcurrentOutput.tsx` (`[completionResult, complete]`), `SingleTask.tsx` (`[taskResult, complete]`), and `InkLifecycleRoot` (`[exitResult, exit]`) — all include the callback in their deps. While `complete` is stable via `useCallback`, the incomplete array violates the exhaustive-deps rule and is inconsistent with the rest of this PR. Note: this is superseded if the state-based refactor from the previous suggestion is adopted.

**Suggestion:** 
```suggestion
  }, [isAborted, complete])
```

packages/cli-kit/src/private/node/ui/hooks/use-async-and-unmount.ts
```diff
@@ -10,17 +10,24 @@ export default function useAsyncAndUnmount(
...
-        unmountInk(error)
+        setResult({error})
       })
   }, [])
+
+  useEffect(() => {
+    if (result !== null) {
+      complete(result.error)
+    }
+  }, [result])
```
gonzaloriestra, line 32: 🎨 **Code Style**: Same inconsistency as `use-abort-signal.ts`: the `useEffect` at line 28 calls `complete()` but the dependency array at line 32 is `[result]` instead of `[result, complete]`. The component-level implementations (`ConcurrentOutput.tsx`, `SingleTask.tsx`) both include `complete` in their deps for the equivalent pattern.

**Suggestion:** 
```suggestion
  }, [result, complete])
```

gonzaloriestra (commented):
---

packages/cli-kit/src/private/node/ui/hooks/use-abort-signal.ts
ryancbahan: This won't work without wrapping the app in a context to keep the state actively stored in memory, which is going to add even more side effects and complexity. I think this implementation is fine as-is. Pre-existing tests passing and thorough tophatting demonstrate parity behavior, and this code almost never gets touched. The computation itself is very cheap and doesn't need memoization.

packages/cli-kit/src/private/node/ui/hooks/use-abort-signal.ts
ryancbahan: The comparisons made here aren't really relevant -- this is a hook and not ui. the dep array here is correct.

packages/cli-kit/src/private/node/ui/hooks/use-async-and-unmount.ts
ryancbahan: same comment as above, the comparison between component and hook isn't 1:1

isaacroldan (approved):
:clap:

PR #7220, opened 2026-04-08
Extract custom watch paths to specifications

isaacroldan:

The current file watching system for dev sessions uses a hardcoded approach where function extensions have custom watch paths while other extensions watch everything. This creates inconsistent behavior and makes it difficult for different extension types to define their own watching requirements.

Refactors the dev session file watching system to use a specification-based approach:

- Replaces `devSessionCustomWatchPaths` with a new `devSessionWatchConfig` method on extension specifications
- Introduces a `DevSessionWatchConfig` interface that allows extensions to define both paths to watch and patterns to ignore
- Moves function-specific watch logic from the base `ExtensionInstance` class to the function specification
- Adds watch configuration for admin extensions that only watches the `static_root` directory when configured
- Sets default behavior where 'configuration' experience extensions watch nothing and 'extension' experience extensions watch everything
- Moves default ignore patterns from the file watcher service to the extension instance level

### How to test your changes?

1. Create a function extension with custom `build.watch` configuration and verify it only watches the specified paths plus default files (locales, GraphQL, TOML)
2. Create an admin extension with `static_root` configured and verify it only watches files in that directory
3. Create other extension types and verify they follow the default watching behavior based on their experience type
4. Run dev sessions and confirm file changes trigger rebuilds appropriately for each extension type

### Checklist

### DIFF
--- a/packages/app/src/cli/models/app/loader.test.ts
+++ b/packages/app/src/cli/models/app/loader.test.ts
@@ -2070,12 +2070,9 @@ describe('load', () => {
     const app = await loadTestingApp()
 
     // Then
-    expect(app.allExtensions).toHaveLength(7)
+    expect(app.allExtensions).toHaveLength(6)
     const extensionsConfig = app.allExtensions.map((ext) => ext.configuration)
     expect(extensionsConfig).toEqual([
-      expect.objectContaining({
-        name: 'for-testing',
-      }),
       expect.objectContaining({
         name: 'for-testing',
       }),
@@ -2131,12 +2128,9 @@ describe('load', () => {
     const app = await loadTestingApp({remoteFlags: []})
 
     // Then
-    expect(app.allExtensions).toHaveLength(8)
+    expect(app.allExtensions).toHaveLength(7)
     const extensionsConfig = app.allExtensions.map((ext) => ext.configuration)
     expect(extensionsConfig).toEqual([
-      {
-        name: 'for-testing-webhooks',
-      },
       {
         name: 'for-testing-webhooks',
       },
--- a/packages/app/src/cli/models/app/loader.test.ts
+++ b/packages/app/src/cli/models/app/loader.test.ts
@@ -2070,9 +2070,12 @@ describe('load', () => {
     const app = await loadTestingApp()
 
     // Then
-    expect(app.allExtensions).toHaveLength(6)
+    expect(app.allExtensions).toHaveLength(7)
     const extensionsConfig = app.allExtensions.map((ext) => ext.configuration)
     expect(extensionsConfig).toEqual([
+      expect.objectContaining({
+        name: 'for-testing',
+      }),
       expect.objectContaining({
         name: 'for-testing',
       }),
@@ -2128,9 +2131,12 @@ describe('load', () => {
     const app = await loadTestingApp({remoteFlags: []})
 
     // Then
-    expect(app.allExtensions).toHaveLength(7)
+    expect(app.allExtensions).toHaveLength(8)
     const extensionsConfig = app.allExtensions.map((ext) => ext.configuration)
     expect(extensionsConfig).toEqual([
+      {
+        name: 'for-testing-webhooks',
+      },
       {
         name: 'for-testing-webhooks',
       },
--- a/packages/app/src/cli/models/extensions/extension-instance.ts
+++ b/packages/app/src/cli/models/extensions/extension-instance.ts
@@ -1,6 +1,6 @@
 import {BaseConfigType, MAX_EXTENSION_HANDLE_LENGTH, MAX_UID_LENGTH} from './schemas.js'
 import {FunctionConfigType} from './specifications/function.js'
-import {ExtensionFeature, ExtensionSpecification} from './specification.js'
+import {DevSessionWatchConfig, ExtensionFeature, ExtensionSpecification} from './specification.js'
 import {SingleWebhookSubscriptionType} from './specifications/app_config_webhook_schemas/webhooks_schema.js'
 import {ExtensionBuildOptions, bundleFunctionExtension} from '../../services/build/extension.js'
 import {bundleThemeExtension} from '../../services/extensions/bundle.js'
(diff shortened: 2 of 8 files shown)


isaacroldan:
* **#7220** <a href="https://app.graphite.com/github/pr/Shopify/cli/7220?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#7226](https://github.com/Shopify/cli/pull/7226) <a href="https://app.graphite.com/github/pr/Shopify/cli/7226?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#7227](https://github.com/Shopify/cli/pull/7227) <a href="https://app.graphite.com/github/pr/Shopify/cli/7227?utm_source=stack-comment-icon" target="_blank">Graphite</a>) 👈 <a href="https://app.graphite.com/github/pr/Shopify/cli/7220?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#7224** <a href="https://app.graphite.com/github/pr/Shopify/cli/7224?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

packages/app/src/cli/models/app/loader.test.ts
```diff
@@ -2070,9 +2070,12 @@ describe('load', () => {
     const app = await loadTestingApp()
 
     // Then
-    expect(app.allExtensions).toHaveLength(6)
+    expect(app.allExtensions).toHaveLength(7)
```
craigmichaelmartin: 📐 **Design**: The extension count increased from 6 to 7 (and 7 to 8 in the other test). The new entries are admin extensions created even though the test TOML configs don't have an `[admin]` section. This is a side effect of switching admin from `createContractBasedModuleSpecification` to `createExtensionSpecification` — the admin spec now participates in extension loading unconditionally because `AdminSchema` makes the `admin` field optional.

This means every app now gets an admin extension instance with `admin: undefined`, which could affect extension counts, deploy payloads (`deployConfig` returns `{admin: undefined}`), and any code iterating over `allExtensions`.

**Suggestion:** Worth verifying this is intentional. If the admin extension should only be created when `[admin]` is present in the TOML, consider adding a guard in the spec or loader.

packages/app/src/cli/models/extensions/extension-instance.ts
```diff
@@ -436,20 +431,31 @@ export class ExtensionInstance<TConfiguration extends BaseConfigType = BaseConfi
...
+      '**/*.test.*',
+      '**/dist/**',
+      '**/*.swp',
+      '**/generated/**',
+      '**/.gitignore',
+    ]
+    const watchConfig = this.devSessionWatchConfig
+
+    const patterns = watchConfig?.paths ?? ['**/*']
+    const ignore = watchConfig?.ignore ?? defaultIgnore
```
craigmichaelmartin, line 446: 🐛 **Bug**: The `??` operator only falls through on `null`/`undefined`, not on empty arrays. When a specification's `devSessionWatchConfig` returns `{ignore: []}` (as admin.ts does on line 27), the `ignore` variable becomes `[]` instead of `defaultIgnore`, so `globSync` runs with NO ignore patterns.

The `DevSessionWatchConfig` interface at `specification.ts:151` documents `ignore` as "Additional glob patterns to ignore (on top of the default ignore list)" — but the implementation here replaces the defaults rather than augmenting them.

Currently only the admin spec with `static_root` is affected, and static asset directories are unlikely to contain `node_modules`, so the practical impact is low today. But this contract mismatch will bite future spec authors who rely on the documented additive behavior and accidentally lose all default ignores (node_modules, .git, dist, etc.).

**Suggestion:** Either change the merge logic so that a spec's `ignore` augments the defaults:

```suggestion
    const ignore = [...defaultIgnore, ...(watchConfig?.ignore ?? [])]
```

Or, if replacement semantics are intended, update the JSDoc on `DevSessionWatchConfig.ignore` to say "Replaces the default ignore list when provided" and ensure admin.ts passes defaults explicitly if needed.

packages/app/src/cli/models/extensions/specifications/function.ts
```diff
@@ -90,6 +90,18 @@ const functionSpec = createExtensionSpecification({
   appModuleFeatures: (_) => ['function'],
   buildConfig: {mode: 'function'},
   getOutputRelativePath: (_extension: ExtensionInstance<FunctionConfigType>) => joinPath('dist', 'index.wasm'),
+  devSessionWatchConfig: (extension: ExtensionInstance<FunctionConfigType>) => {
```
craigmichaelmartin, line 93: ⭐ **Praise**: Moving the function watch path logic from `ExtensionInstance` into the function specification is a solid architectural improvement. The old code used `this.configuration as unknown as FunctionConfigType` — a type escape hatch. Now the function spec has proper access to its typed configuration and owns its watch behavior. This is the right level of abstraction.

packages/app/src/cli/services/dev/app-events/file-watcher.ts
```diff
@@ -118,15 +118,6 @@ export class FileWatcher {
     // Create new watcher
     const {default: chokidar} = await import('chokidar')
     this.watcher = chokidar.watch(watchPaths, {
```
craigmichaelmartin, line 120: ⚡ **Performance**: The `ignored` option was removed from `chokidar.watch()`. Chokidar still recursively watches `fullExtensionDirectories` (line 110), so it now sets up OS-level file watchers (inotify/FSEvents) for `node_modules/`, `.git/`, `dist/`, `generated/`, and test files inside every extension directory.

Events for these files ARE filtered downstream via `shouldIgnoreEvent()` and `handleFileEvent()`, so correctness is not affected. However:

1. OS watcher resources are consumed for potentially thousands of files in node_modules
2. Each change fires a JS callback + path normalization + map lookup before being discarded
3. During `npm install` in an extension directory, this could cause a storm of spurious events
4. On Linux, this can push toward the `inotify` watch limit

The per-extension `watchedFiles()` filtering controls which files appear in `extensionWatchedFiles` — it doesn't prevent chokidar from watching everything in the watched directories.

**Suggestion:** Consider keeping a minimal set of universally-ignored patterns on chokidar for performance. These directories are never useful to watch, and the extension-level filtering remains the source of truth for everything else:

```suggestion
    this.watcher = chokidar.watch(watchPaths, {
      ignored: ['**/node_modules/**', '**/.git/**'],
      persistent: true,
      ignoreInitial: true,
    })
```

craigmichaelmartin (changes requested):
Love these changes! Just a few comments I want to make sure we evaluate!

packages/app/src/cli/models/app/loader.test.ts
isaacroldan: Good catch, admin should not be optional for the admin extension

isaacroldan: Fixed!

packages/app/src/cli/models/extensions/extension-instance.ts
isaacroldan: This is intended, updated the JSDoc

craigmichaelmartin (approved):
Looks great!

craigmichaelmartin (approved):

PR #7224, opened 2026-04-08
Use extension handle to identify extensions in file-watcher

isaacroldan:
### WHY are these changes introduced?

The file watcher was using extension directory paths to identify which extensions are affected by file changes. This approach has limitations when dealing with shared files or when extensions need to be uniquely identified beyond their filesystem location.

### WHAT is this pull request doing?

Refactors the file watcher system to use extension handles instead of directory paths for tracking file ownership and identifying affected extensions. The changes include:

- Added `extensionHandle` field to `WatcherEvent` interface to uniquely identify extensions
- Modified file tracking to map file paths to extension handles rather than directory paths
- Updated event filtering logic to use extension handles when available, falling back to directory paths for folder-level events
- Enhanced test cases to verify the new event structure and handle multiple events correctly
- Improved event deduplication to consider extension handles alongside path and type

The `extensionPath` field is retained for backward compatibility and folder-level events where no extension handle exists yet.

### How to test your changes?

1. Run the existing file watcher tests to ensure all functionality works as expected
2. Test file changes in extensions to verify events are properly attributed to the correct extension handles
3. Test extension creation/deletion scenarios to ensure folder-level events still work correctly
4. Verify that shared files between extensions generate separate events for each affected extension

### Checklist

- [ ] I've considered possible cross-platform impacts (Mac, Linux, Windows)
- [ ] I've considered possible [documentation](https://shopify.dev) changes
- [ ] I've considered analytics changes to measure impact
- [ ] The change is user-facing, so I've added a changelog entry with `pnpm changeset add`

PR #7226, opened 2026-04-08
Add appAssets to the websocket app payload

melissaluu:

Fixes #0000 

### Checklist

- [ ] I've considered analytics changes to measure impact
- [ ] The change is user-facing, so I've added a changelog entry with `pnpm changeset add`

### DIFF
--- a/packages/app/src/cli/models/extensions/specifications/admin.ts
+++ b/packages/app/src/cli/models/extensions/specifications/admin.ts
@@ -24,7 +24,7 @@ const adminSpecificationSpec = createExtensionSpecification({
     if (!staticRoot) return {paths: []}
 
     const path = joinPath(extension.directory, staticRoot, '**/*')
-    return {paths: [path], ignore: []}
+    return {paths: [path], ignore: [], assetKey: 'staticRoot'}
   },
   transformRemoteToLocal: (remoteContent) => {
     return {
--- a/packages/app/src/cli/services/dev/extension.ts
+++ b/packages/app/src/cli/services/dev/extension.ts
@@ -9,10 +9,16 @@ import {
 import {AppEvent, AppEventWatcher, EventType} from './app-events/app-event-watcher.js'
 import {buildCartURLIfNeeded} from './extension/utilities.js'
 import {ExtensionInstance} from '../../models/extensions/extension-instance.js'
+import {globPatternBaseDir} from '../../models/extensions/specification.js'
 import {AbortSignal} from '@shopify/cli-kit/node/abort'
 import {outputDebug} from '@shopify/cli-kit/node/output'
+import {normalizePath} from '@shopify/cli-kit/node/path'
 import {DotEnvFile} from '@shopify/cli-kit/node/dot-env'
 import {Writable} from 'stream'
+
+interface AppAssets {
+  [key: string]: string
+}
 
 export interface ExtensionDevOptions {
   /**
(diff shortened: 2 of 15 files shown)


melissaluu:
* **#7226** <a href="https://app.graphite.com/github/pr/Shopify/cli/7226?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Shopify/cli/7226?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#7220** <a href="https://app.graphite.com/github/pr/Shopify/cli/7220?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#7227](https://github.com/Shopify/cli/pull/7227) <a href="https://app.graphite.com/github/pr/Shopify/cli/7227?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#7224** <a href="https://app.graphite.com/github/pr/Shopify/cli/7224?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

packages/app/src/cli/models/extensions/specifications/admin.ts
```diff
@@ -24,7 +24,7 @@ const adminSpecificationSpec = createExtensionSpecification({
     if (!staticRoot) return {paths: []}
 
     const path = joinPath(extension.directory, staticRoot, '**/*')
-    return {paths: [path], ignore: []}
+    return {paths: [path], ignore: [], assetKey: 'staticRoot'}
```
melissaluu, line 27: This was added so that we can use this config in the `resolveAppAssets()`

We don't have to do this if we're okay with directly using `App` object to find the `appAsset` directory

packages/app/src/cli/services/dev/extension.ts
```diff
@@ -112,6 +118,28 @@ export interface ExtensionDevOptions {
...
+   * Map of asset key to absolute directory path for app-level assets (e.g., admin static_root)
+   */
+  appAssets?: AppAssets
+}
+
+/**
+ * Derives app-level asset directories from config extensions that define devSessionWatchConfig
+ * with an assetKey. Returns a map of asset key (e.g. 'static_root') to absolute directory path.
+ */
+export function resolveAppAssets(allExtensions: ExtensionInstance[]): Record<string, string> {
```
melissaluu, line 132: We can switch this back to using the `App` object to find the `appAssets` directory path  (i.e. `app.configuration.assets.static_root`) but this implementation removes hardcoding and specifying the path directly and we grab it from the `devSessionWatchConfig` instead

packages/app/src/cli/services/dev/processes/setup-dev-processes.ts
```diff
@@ -149,7 +149,7 @@ export async function setupDevProcesses({
         })
       : undefined,
     await setupPreviewableExtensionsProcess({
-      allExtensions: reloadedApp.allExtensions,
+      allExtensions: reloadedApp.realExtensions,
```
melissaluu, line 152: @isaacroldan is this safe? I asked Claude multiple times to check this and it said that this should be okay since technically `reloadedApp.allExtensions` can omit config extensions with when `includeConfigOnDeploy` is set to `false`

isaacroldan: why do you need this change though?

packages/app/src/cli/services/dev/extension.ts
```diff
@@ -133,17 +161,29 @@ export async function devUIExtensions(options: ExtensionDevOptions): Promise<voi
...
+    getExtensions,
+    getAppAssets,
+  })
 
   outputDebug(`Setting up the UI extensions Websocket server...`, payloadOptions.stdout)
   const websocketConnection = setupWebsocketConnection({...payloadOptions, httpServer, payloadStore})
   outputDebug(`Setting up the UI extensions bundler and file watching...`, payloadOptions.stdout)
 
-  const eventHandler = async ({appWasReloaded, app, extensionEvents}: AppEvent) => {
+  const eventHandler = async ({appWasReloaded, app, extensionEvents, appAssetsUpdated}: AppEvent) => {
```
isaacroldan, line 176: you also have the app here from the AppEvent, this one is better because is the updated app (in case anything changed)

packages/app/src/cli/services/dev/extension.ts
```diff
@@ -112,6 +118,28 @@ export interface ExtensionDevOptions {
...
+  const appAssets: Record<string, string> = {}
+  for (const ext of allExtensions) {
+    if (!ext.isAppConfigExtension) continue
+    const watchConfig = ext.devSessionWatchConfig
+    if (!watchConfig || watchConfig.paths.length === 0 || !watchConfig.assetKey) continue
+
+    const baseDir = normalizePath(globPatternBaseDir(watchConfig.paths[0]!))
+    appAssets[watchConfig.assetKey] = baseDir
+  }
+  return appAssets
```
isaacroldan, line 142: This solution is very hacky. We are adding the `assetKey` as a way of generization, but in the way we are polluting the `devSessionWatchConfig` with a property that doesn't have anything to do with watching, i'll think on a better way to extract this.

isaacroldan, line 142: This solution is very hacky. We are adding the `assetKey` as a way of generization, but in the way we are polluting the `devSessionWatchConfig` with a property that doesn't have anything to do with watching, i'll think on a better way to extract this.

packages/app/src/cli/services/dev/processes/setup-dev-processes.ts
melissaluu: This came from a Claude review where it said that `allExtensions` can technically omit config extensions which we need.

PR #7228, opened 2026-04-09
Use findUp instead of npm prefix to find the project folder

gonzaloriestra:

Related to: https://shopify.slack.com/archives/C07UJ7UNMTK/p1775663292357249

`getPackageManager()` spawns `npm prefix` as a subprocess to locate the project root. If `npm` is unavailable, the subprocess fails and the function falls back to the user agent string. When the CLI is invoked directly (not via `pnpm shopify`), the user agent is unset, so the package manager resolves to `"unknown"` — causing commands like `shopify app generate extension` to fail with `spawn unknown ENOENT`.

Replaces the `npm prefix` subprocess call with `findPathUp('package.json')`, a pure-filesystem walk that's already available in cli-kit. This eliminates the dependency on `npm`, making package manager detection reliable regardless of how the CLI is invoked.

### How to test your changes?

- `pnpm i -g --@shopify:registry=https://registry.npmjs.org @shopify/cli@0.0.0-snapshot-20260409112001`
- `shopify app init`
- `shopify app generate extension --template checkout_ui --name test-ext`

### Checklist

### DIFF
--- a/packages/cli-kit/src/public/node/node-package-manager.ts
+++ b/packages/cli-kit/src/public/node/node-package-manager.ts
@@ -1,6 +1,6 @@
 import {AbortError, BugError} from './error.js'
 import {AbortController, AbortSignal} from './abort.js'
-import {captureOutput, exec} from './system.js'
+import {exec} from './system.js'
 import {fileExists, readFile, writeFile, findPathUp, glob} from './fs.js'
 import {dirname, joinPath} from './path.js'
 import {runWithTimer} from './metadata.js'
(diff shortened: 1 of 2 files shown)


craigmichaelmartin (approved):

packages/cli-kit/src/public/node/node-package-manager.ts
```diff
@@ -115,32 +115,18 @@ export function packageManagerFromUserAgent(env = process.env): PackageManager {
...
-    directory = await captureOutput('npm', ['prefix'], {cwd: fromDirectory})
-    outputDebug(outputContent`Obtaining the dependency manager in directory ${outputToken.path(directory)}...`)
-    packageJson = joinPath(directory, 'package.json')
-    // eslint-disable-next-line no-catch-all/no-catch-all
-  } catch {
-    // if problems locating directoy/package file, we use user agent instead
-  }
-
-  if (!directory || !packageJson || !(await fileExists(packageJson))) {
+  const packageJsonPath = await findPathUp('package.json', {cwd: fromDirectory, type: 'file'})
```
isaacroldan, line 118: Do we want to use `pnpm prefix` instead? or is it better to do this?

gonzaloriestra: We were running this in the machine of the users, so that would require them to have PNPM installed. I think this approach is simpler, faster and more reliable.

isaacroldan (approved):

gonzaloriestra:
### Merge activity

* **Apr 9, 1:41 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/Shopify/cli/7228).
* **Apr 9, 1:41 PM UTC**: @gonzaloriestra added this pull request to the GitHub merge queue with [Graphite](https://app.graphite.com/github/pr/Shopify/cli/7228).

dmerand:
I can see this is merging, and it may be the right fix for the immediate problem. I just want to make sure we’re explicitly aligned on the architectural direction here.

As I understand it, this changes `getPackageManager(...)` from:

- `npm prefix` → project/workspace root

to:

- `findPathUp('package.json')` → nearest package root

I’ve been working on a follow-up stack under a different assumption: keep the root/workspace-oriented behavior for the broad helper, then add narrower helper boundaries for cases that need different intent:

- directory-local execution: https://github.com/Shopify/cli/pull/7239
- root-known project resolution: https://github.com/Shopify/cli/pull/7241
- explicit project-operation boundary: https://github.com/Shopify/cli/pull/7242
- display/follow-up-command helper: https://github.com/Shopify/cli/pull/7243

So this PR seems like more than an implementation detail — it changes the default resolution model. If that shift is intentional, I should probably rework parts of my stack accordingly, especially `#7241`. I mostly want to confirm we’re aligned that the architectural direction is changing here, rather than just fixing the `npm prefix` subprocess dependency.

gonzaloriestra:
@dmerand `npm prefix` is actually returning `the closest parent directory to contain a package.json file or node_modules directory`, according to [their documentation](https://docs.npmjs.com/cli/v8/commands/npm-prefix#description). And that's exactly what we are doing now with `findPathUp`, so the behavior shouldn't change.

dmerand:
> @dmerand npm prefix is actually returning the closest parent directory to contain a package.json file or node_modules directory, according to [their documentation](https://docs.npmjs.com/cli/v8/commands/npm-prefix#description). And that's exactly what we are doing now with findPathUp, so the behavior shouldn't change.

I re-checked the npm docs and implementation after your note, and I think the correction is:

- my earlier “`npm prefix` == workspace root” framing was too strong
- but `findPathUp('package.json')` still isn’t quite the same as npm’s local-prefix behavior either

The docs describe `npm prefix` as the closest parent with `package.json` **or** `node_modules`, and the implementation also has some extra upward-walk/workspace behavior beyond a plain nearest-`package.json` lookup.

Not holding up this PR, you should continue to ship. I think this updates the rationale for my stack more than it invalidates it. I'll reframe the follow-up work around explicit caller intent rather than around matching `npm prefix` exactly.

gonzaloriestra:
@dmerand you were right, the new implementation is not equivalent.

Apparently `npx prefix` is returning the workspace root, even if the current folder has a package.json (the documentation is not very clear...):
13-23-4htmn-cqv07

So this code is not working correctly when trying to find the package manager from an extension, where we have a package.json, but not a lock file.

I'm fixing that in https://github.com/Shopify/cli/pull/7253

PR #7229, opened 2026-04-09
Use detected package manager for function typegen

saga-dasgupta:
## Problem

Function builds fail in environments where npm is blocked:

```
Failed to build function.

Command failed with exit code 1: npm exec -- graphql-code-generator --config package.json
npm is no longer supported in a global context.
```

This affects all JS function extensions because `buildGraphqlTypes` hardcoded `npm exec` for the typegen step.

## Root cause

Two issues in `getPackageManager()` in cli-kit:

1. **Shelled out to `npm prefix`** to find the project root — which fails when npm itself is unavailable
2. **Anchored lockfile search to the nearest `package.json`** — function extensions have their own `package.json` in subdirectories (e.g., `extensions/my-function/package.json`), but the lockfile (`pnpm-lock.yaml`) lives at the app root. The old approach found the function's `package.json` first, saw no lockfile next to it, and defaulted to `npm`

## Fix

- Replaced `npm prefix` subprocess with `findPathUp` for pure filesystem-based detection — no subprocess needed
- Lockfiles are now searched for directly by walking up the directory tree, so `pnpm-lock.yaml` at the app root is found even when called from a nested function directory
- `buildGraphqlTypes` now uses the fixed `getPackageManager(fun.directory)` to detect the project's package manager

## Test plan
- [x] All 58 `node-package-manager.test.ts` tests pass, including new test for nested subdirectory with parent lockfile
- [x] All 25 `build.test.ts` tests pass
- [x] Manual test: Verified building a function using 0.0.0-snapshot-20260409154217

### DIFF
--- a/packages/app/src/cli/services/function/binaries.ts
+++ b/packages/app/src/cli/services/function/binaries.ts
@@ -14,7 +14,7 @@ import {fileURLToPath} from 'node:url'
 export const PREFERRED_FUNCTION_RUNNER_VERSION = '9.1.2'
 
 // Javy dependencies.
-export const PREFERRED_JAVY_VERSION = '8.1.1'
+export const PREFERRED_JAVY_VERSION = '7.0.1'
 // The Javy plugin version should match the plugin version used in the
 // function-runner version specified above.
 export const PREFERRED_JAVY_PLUGIN_VERSION = '3'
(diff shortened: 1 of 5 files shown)


saga-dasgupta:
snapit

PR #7230, opened 2026-04-09
Add directory-based package manager inference helper

dmerand:
## What

Add `inferPackageManagerForDirectory` to `cli-kit` and expand package manager detection to walk ancestor directories for lockfiles and `pnpm-workspace.yaml`.

Also update `getPackageManager(fromDirectory)` to reuse the same ancestor-aware directory detection once `npm prefix` has resolved a package root, and add coverage for ancestor workspace roots, nested directories, and lockfile precedence.

## Why

This is groundwork for reducing reliance on `npm prefix` in project-directory package-manager detection, without changing call sites yet.

The CLI still needs a way to infer the package manager for a project directory when `npm` subprocesses are unavailable or are the wrong source of truth. This PR introduces that directory-first path first, so follow-up PRs can adopt it in specific call sites starting with function typegen.

## How

This PR keeps both entrypoints, but makes their difference explicit:

- `getPackageManager(fromDirectory)` remains the existing **npm-prefix-based** API. It first asks `npm prefix` for the package root, then detects lockfiles/workspace markers from that resolved root upward. If `npm prefix` or package discovery fails, it falls back to the current user agent.
- `inferPackageManagerForDirectory(fromDirectory)` is the new **directory-first** API. It never shells out to `npm` and never consults global CLI installation detection. It walks upward from the provided directory itself, then falls back to the current user agent and finally `npm`.

Shared directory detection now:
- prefers the closest matching lockfile or workspace marker
- covers ancestor `pnpm-workspace.yaml` / lockfile cases
- keeps the existing fallback policy of each public API intact

### DIFF
--- a/packages/cli-kit/src/public/node/node-package-manager.test.ts
+++ b/packages/cli-kit/src/public/node/node-package-manager.test.ts
@@ -1050,7 +1050,9 @@ describe('inferPackageManagerForDirectory', () => {
       const nestedDirectory = joinPath(tmpDir, 'extensions', 'discount-function')
       await mkdir(nestedDirectory)
 
-      const packageManager = await inferPackageManagerForDirectory(nestedDirectory, {npm_config_user_agent: 'yarn/1.22.0'})
+      const packageManager = await inferPackageManagerForDirectory(nestedDirectory, {
+        npm_config_user_agent: 'yarn/1.22.0',
+      })
 
       expect(packageManager).toEqual('yarn')
       expect(mockedCaptureOutput).not.toHaveBeenCalled()
--- a/packages/cli-kit/src/public/node/node-package-manager.test.ts
+++ b/packages/cli-kit/src/public/node/node-package-manager.test.ts
@@ -12,6 +12,7 @@ import {
   addResolutionOrOverride,
   writePackageJSON,
   getPackageManager,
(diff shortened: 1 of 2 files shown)


dmerand:
* **#7233** <a href="https://app.graphite.com/github/pr/Shopify/cli/7233?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#7231** <a href="https://app.graphite.com/github/pr/Shopify/cli/7231?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#7230** <a href="https://app.graphite.com/github/pr/Shopify/cli/7230?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Shopify/cli/7230?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

dmerand:
Closing this as exploratory groundwork.

This branch helped clarify two things:
- the immediate fix should use the package manager for the project/function directory, not the CLI install context
- there is a real higher-level design question around package-manager detection APIs and workspace semantics

I don’t think we should land that broader API split in this stack, though. I’m going to replace this series with a single narrower PR for the reported function typegen issue, and then step back on the larger package-manager design separately.

PR #7231, opened 2026-04-09
Use directory-based package manager inference for function typegen

dmerand:
## What

Use directory-based package manager inference for the default function GraphQL typegen path.

Move the package-manager-specific binary invocation logic into `cli-kit`, and update function typegen to execute `graphql-code-generator` through that shared helper instead of hardcoding `npm exec`.

## Why

The default JS function typegen path still assumes `npm`, which breaks in projects that are managed with pnpm, Yarn, or Bun.

PR #7230 adds the shared inference groundwork. This PR is the first call-site adoption in the stack: fix function typegen so it uses the package manager for the function directory rather than the CLI's old npm-shaped default.

## How

- infer the package manager from `fun.directory` with `inferPackageManagerForDirectory(...)`
- build the binary invocation through shared `cli-kit` helper `packageManagerBinaryCommand(...)`
- keep `typegen_command` behavior unchanged, including the existing custom-command override path

This keeps the seam explicit:
- `cli-kit` owns package-manager detection and command-shape differences
- `function/build.ts` just chooses the default codegen binary to run

This helper is intentionally narrow. If we need broader package-manager handling later, we should evaluate a dedicated dependency instead of continuing to grow this surface ad hoc.

### DIFF
--- a/packages/app/src/cli/services/function/build.test.ts
+++ b/packages/app/src/cli/services/function/build.test.ts
@@ -22,12 +22,20 @@ import {
 import {testApp, testFunctionExtension} from '../../models/app/app.test-data.js'
 import {beforeEach, describe, expect, test, vi} from 'vitest'
 import {exec} from '@shopify/cli-kit/node/system'
+import {inferPackageManagerForDirectory} from '@shopify/cli-kit/node/node-package-manager'
 import {dirname, joinPath} from '@shopify/cli-kit/node/path'
 import {inTemporaryDirectory, mkdir, readFileSync, writeFile, removeFile} from '@shopify/cli-kit/node/fs'
 import {build as esBuild} from 'esbuild'
 
 vi.mock('@shopify/cli-kit/node/fs')
 vi.mock('@shopify/cli-kit/node/system')
+vi.mock('@shopify/cli-kit/node/node-package-manager', async () => {
+  const actual: any = await vi.importActual('@shopify/cli-kit/node/node-package-manager')
+  return {
+    ...actual,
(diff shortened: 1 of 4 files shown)


dmerand:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/Shopify/cli/7231?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#7233** <a href="https://app.graphite.com/github/pr/Shopify/cli/7233?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#7231** <a href="https://app.graphite.com/github/pr/Shopify/cli/7231?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Shopify/cli/7231?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#7230** <a href="https://app.graphite.com/github/pr/Shopify/cli/7230?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

dmerand:
Closing this as exploratory follow-through from the helper branch.

I still think the project-directory behavior here is better for the function typegen target itself, but this PR depends on a broader helper split that I don’t think we want to carry forward as the long-term direction.

I’m going to replace this stack with a single narrower PR that fixes the reported function typegen issue using the best parts of what we learned here, without committing to the larger package-manager architecture yet.

PR #7239, opened 2026-04-09
Use function directory package manager for GraphQL typegen

dmerand:
## Stack context

This stack is not trying to reduce Shopify CLI's supported package managers or preserve every detail of npm's local-prefix behavior.

The real problem is that Shopify CLI answers several different package-manager questions through helpers that are too broad for the caller's intent.

The end state is to make callers choose intent, not implementation. Concretely, the codebase needs clearer helper boundaries for:
- launcher/global-install resolution
- broad upward-walk package-manager detection
- known project-root resolution
- directory-local execution

This PR establishes the directory-local execution boundary for function GraphQL typegen.

## What

Use the function directory to choose the package manager for the default JavaScript GraphQL typegen path instead of always running `npm exec`.

Add a `cli-kit` helper that builds the right command for running a project binary from a directory, and use it from `buildGraphqlTypes(...)`.

## Why

The current typegen path hardcodes `npm`, which breaks apps whose function directories are managed by `pnpm`, `yarn`, or `bun`.

This keeps the fix scoped to the function typegen problem. It does not change the broader `getPackageManager(...)` contract for other callers.

## How

Build `packageManagerBinaryCommandForDirectory(...)` on top of the current `main` upward-walk package-manager detection instead of introducing a separate detection path.

That helper resolves the package manager for the function directory and returns the right binary invocation shape for `npm`, `pnpm`, `yarn`, or `bun`.

While touching this path, restore missing marker coverage for directory-local resolution:
- `pnpm-workspace.yaml`
- `bun.lock`

Existing `bun.lockb` support remains in place.

## Testing

- `pnpm exec vitest run packages/cli-kit/src/public/node/node-package-manager.test.ts packages/app/src/cli/services/function/build.test.ts`

To test manually:
1. Create or use an app with a JavaScript function managed by `pnpm`, `yarn`, or `bun`
2. Run `shopify app function build`
3. Confirm GraphQL type generation runs through that package manager instead of `npm`

### DIFF
--- a/packages/app/src/cli/services/function/build.ts
+++ b/packages/app/src/cli/services/function/build.ts
@@ -24,6 +24,7 @@ import {renderTasks} from '@shopify/cli-kit/node/ui'
 import {pickBy} from '@shopify/cli-kit/common/object'
 import {runWithTimer} from '@shopify/cli-kit/node/metadata'
 import {AbortError} from '@shopify/cli-kit/node/error'
+import {packageManagerBinaryCommandForDirectory} from '@shopify/cli-kit/node/node-package-manager'
 import {Writable} from 'stream'
 
 export const PREFERRED_FUNCTION_NPM_PACKAGE_MAJOR_VERSION = '2'
@@ -143,8 +144,15 @@ export async function buildGraphqlTypes(
     )
   }
 
+  const command = await packageManagerBinaryCommandForDirectory(
+    fun.directory,
+    'graphql-code-generator',
+    '--config',
+    'package.json',
+  )
+
   return runWithTimer('cmd_all_timing_network_ms')(async () => {
-    return exec('npm', ['exec', '--', 'graphql-code-generator', '--config', 'package.json'], {
+    return exec(command.command, command.args, {
       cwd: fun.directory,
       stderr: options.stderr,
       signal: options.signal,
--- a/packages/cli-kit/src/public/node/node-package-manager.ts
+++ b/packages/cli-kit/src/public/node/node-package-manager.ts
@@ -27,6 +27,7 @@ export const pnpmLockfile = 'pnpm-lock.yaml'
 
 /** The name of the bun lock file */
 export const bunLockfile = 'bun.lockb'
+const modernBunLockfile = 'bun.lock'
 
 /** The name of the pnpm workspace file */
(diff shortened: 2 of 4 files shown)


dmerand:
* **#7239** <a href="https://app.graphite.com/github/pr/Shopify/cli/7239?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Shopify/cli/7239?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/cli-kit/src/public/node/node-package-manager.ts
```diff
@@ -27,6 +27,7 @@ export const pnpmLockfile = 'pnpm-lock.yaml'
 
 /** The name of the bun lock file */
 export const bunLockfile = 'bun.lockb'
+const modernBunLockfile = 'bun.lock'
```
gonzaloriestra, line 30: TIL

packages/app/src/cli/services/function/build.ts
```diff
@@ -143,8 +144,15 @@ export async function buildGraphqlTypes(
     )
   }
 
+  const command = await packageManagerBinaryCommandForDirectory(
+    fun.directory,
```
ryancbahan, line 148: fun!

ryancbahan (approved):

packages/cli-kit/src/public/node/node-package-manager.ts
```diff
@@ -137,6 +175,19 @@ export async function getPackageManager(fromDirectory: string): Promise<PackageM
   return 'npm'
 }
 
+/**
+ * Builds the command and argv needed to execute a local binary using the package manager
+ * detected from the provided directory or its ancestors.
+ */
+export async function packageManagerBinaryCommandForDirectory(
```
gonzaloriestra, line 182: Much simpler with this helper, thanks for merging it with my PR!

gonzaloriestra (approved):

packages/cli-kit/src/public/node/node-package-manager.ts
dmerand: I, as well, learned this in this process.

PR #7241, opened 2026-04-09
Add root-known package manager helper

dmerand:
## What

This follow-up continues the same package-manager stack as #7239.

Add a helper for callers that already know the project root, and use it for project metadata and root dependency update paths instead of depending on the broader upward-walk package-manager helper.

## Why

This stack is not about reducing Shopify CLI's supported package managers.

`#7239` handles directory-local execution for function typegen. This PR handles callers that already know the project root.

Even if `getPackageManager(...)` remains a broad upward-walk helper, callers that already know the project root should not depend on its broader root-finding and fallback behavior.

## How

Add `getPackageManagerForProjectRoot(...)` to inspect a known project root directly.

Use it from:
- `Project.load(...)`
- `addResolutionOrOverride(...)`
- local upgrade dependency updates

Also narrow `Project.packageManager` to project-valid values plus `'unknown'` when the app root has no `package.json`.

### DIFF
--- a/packages/app/src/cli/models/project/project.ts
+++ b/packages/app/src/cli/models/project/project.ts
@@ -4,8 +4,8 @@ import {readAndParseDotEnv, DotEnvFile} from '@shopify/cli-kit/node/dot-env'
 import {fileExists, glob, findPathUp, readFile} from '@shopify/cli-kit/node/fs'
 import {
   getDependencies,
-  getPackageManager,
-  PackageManager,
+  getPackageManagerForProjectRoot,
+  ProjectPackageManager,
   usesWorkspaces as detectUsesWorkspaces,
 } from '@shopify/cli-kit/node/node-package-manager'
 import {joinPath, basename} from '@shopify/cli-kit/node/path'
@@ -76,7 +76,7 @@ export class Project {
     // Project metadata
     const packageJSONPath = joinPath(directory, 'package.json')
     const hasPackageJson = await fileExists(packageJSONPath)
-    const packageManager = hasPackageJson ? await getPackageManager(directory) : 'unknown'
+    const packageManager = hasPackageJson ? await getPackageManagerForProjectRoot(directory) : 'unknown'
     const nodeDependencies = hasPackageJson ? await getDependencies(packageJSONPath) : {}
     const usesWorkspaces = hasPackageJson ? await detectUsesWorkspaces(directory) : false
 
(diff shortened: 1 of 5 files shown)


dmerand:
* **#7241** <a href="https://app.graphite.com/github/pr/Shopify/cli/7241?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Shopify/cli/7241?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/app/src/cli/models/project/project.ts
```diff
@@ -101,7 +101,7 @@ export class Project {
   }
 
   readonly directory: string
-  readonly packageManager: PackageManager
+  readonly packageManager: ProjectPackageManager | 'unknown'
```
ryancbahan, line 104: can you remove unknown and either make the operation sync or force the caller to narrow?

dmerand: Yeah, I looked at that. I kept `'unknown'` here because `Project` is still just modeling what’s on disk, and “app root with no package.json” is a real state.

The sync narrowing is in the next PR which adds `requirePackageManagerForOperations()`, so install/mutation callers have to narrow before using it.

Happy to fold that in here instead if you’d prefer, but I split it to keep metadata vs operation-owned use separate.

packages/cli-kit/src/public/node/node-package-manager.ts
```diff
@@ -175,6 +183,25 @@ export async function getPackageManager(fromDirectory: string): Promise<PackageM
...
+ * Resolves the package manager for a directory already known to be the project root.
+ *
+ * Use this when the caller already knows the root directory and wants to inspect that root
+ * directly rather than walking upward from a child directory.
+ *
+ * @param rootDirectory - The known project root directory.
+ * @returns The package manager detected from root markers, defaulting to `npm` when no marker exists.
+ * @throws PackageJsonNotFoundError if the provided directory does not contain a package.json.
+ */
+export async function getPackageManagerForProjectRoot(rootDirectory: string): Promise<ProjectPackageManager> {
```
gonzaloriestra, line 196: When you work with the CLI, I think you typically run it from the project root or from an extension. But both of them already have a package.json, so why using this instead of getPackageManager? Isn't it the same?

packages/cli-kit/src/public/node/node-package-manager.ts
```diff
@@ -733,7 +760,7 @@ export async function findUpAndReadPackageJson(fromDirectory: string): Promise<{
 }
 
 export async function addResolutionOrOverride(directory: string, dependencies: Record<string, string>): Promise<void> {
-  const packageManager = await getPackageManager(directory)
+  const packageManager = await getPackageManagerForProjectRoot(directory)
```
gonzaloriestra, line 763: How do you know that this directory is the root or not? If it must be, I'd make it explicit with the param name or documentation.

PR #7242, opened 2026-04-10
Require explicit package manager for project operations

dmerand:
## What

This follow-up continues the same package-manager stack as #7239 and #7241.

Add an explicit project operation boundary for package-manager use, and use it in dependency install and extension generation paths instead of reading `project.packageManager` directly.

## Why

`#7241` gives callers that already know the project root a more explicit path. This PR applies the same idea to operation-owned callers: install and mutation code should require a usable project package manager instead of letting `'unknown'` flow into dependency helpers.

## How

Add `requireProjectPackageManagerForOperations(project)` in `packages/app/src/cli/utilities/project-package-manager.ts`.

Use it from:
- `installAppDependencies(...)`
- extension generation dependency install paths

When the app root has no `package.json`, these operations now fail early with a clearer error instead of passing ambiguous package-manager state deeper into install code.

### DIFF
--- a/packages/app/src/cli/models/app/app.test-data.ts
+++ b/packages/app/src/cli/models/app/app.test-data.ts
@@ -82,7 +82,7 @@ import {Project} from '../project/project.js'
 import {Session} from '@shopify/cli-kit/node/session'
 import {vi} from 'vitest'
 import {joinPath} from '@shopify/cli-kit/node/path'
-import {PackageManager} from '@shopify/cli-kit/node/node-package-manager'
+import {ProjectPackageManager} from '@shopify/cli-kit/node/node-package-manager'
 
 export const DEFAULT_CONFIG = {
   application_url: 'https://myapp.com',
@@ -154,7 +154,7 @@ export function testAppWithConfig(options?: TestAppWithConfigOptions): AppLinked
 
 interface TestProjectOptions {
   directory?: string
-  packageManager?: PackageManager
+  packageManager?: ProjectPackageManager | 'unknown'
   nodeDependencies?: Record<string, string>
   usesWorkspaces?: boolean
 }
--- a/packages/app/src/cli/models/project/project-integration.test.ts
+++ b/packages/app/src/cli/models/project/project-integration.test.ts
@@ -1,5 +1,6 @@
 import {Project} from './project.js'
 import {resolveDotEnv, resolveHiddenConfig, extensionFilesForConfig, webFilesForConfig} from './config-selection.js'
+import {requireProjectPackageManagerForOperations} from '../../utilities/project-package-manager.js'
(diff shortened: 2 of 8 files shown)


dmerand:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/Shopify/cli/7242?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#7242** <a href="https://app.graphite.com/github/pr/Shopify/cli/7242?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Shopify/cli/7242?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

afurm:
Good addition — enforcing an explicit boundary for operations that require a known package manager is the right pattern. A few things to consider:

**1. The `'unknown'` sentinel type safety**

The test file changes use `ProjectPackageManager | 'unknown'` as the type, but does `ProjectPackageManager` (the actual runtime type) include `'unknown'` as a valid variant? If `packageManager` is typed as `PackageManager` (without `'unknown'`), there's a type mismatch — the test is lying about the type to accommodate the new behavior. Worth verifying the actual type definition in `node-package-manager.ts`.

**2. Should `requirePackageManagerForOperations` cache the result?**

Each call to `requirePackageManagerForOperations` re-validates and returns the package manager. If this is called multiple times in a hot path (e.g., during `installAppDependencies` and then extension generation), is there any cost to re-checking? A cached getter might be cleaner, or at least a comment noting it's idempotent.

**3. The test case — what is "no package.json" in CI/test environments?**

The new test removes `package.json` and expects an `AbortError`. But in containerized or generated test fixtures, is there any scenario where a project legitimately has no `package.json` but should still be operable? The error message says "Could not determine" — is that always the right failure mode, or should there be a way to fall back to a CLI-wide default package manager?

ryancbahan (changes requested):
This feels like the control hould be inverted. Rather than the project itself having a method that throws when its property is missing, the caller should check `project.packageManager` when it needs to and ignore it when it doesn't. If you really need a dedicated helper, you could just have a plain function that takes the project and does the check/narrowing.

dmerand:
> This feels like the control hould be inverted. Rather than the project itself having a method that throws when its property is missing, the caller should check `project.packageManager` when it needs to and ignore it when it doesn't. If you really need a dedicated helper, you could just have a plain function that takes the project and does the check/narrowing.

​That works for me. I took another pass, wanna give it a look?

PR #7243, opened 2026-04-10
Use Project follow-up command helper for display-only paths

dmerand:
## What

This follow-up continues the same package-manager stack as #7239, #7241, and #7242.

Add a `formatProjectFollowUpCommand(...)` helper for Project-backed follow-up commands, and use it in the display-only callers in deploy, dev, info, and generate instead of formatting commands from `project.packageManager` directly.

## Why

`#7242` added an explicit package-manager boundary for project operations. This PR applies the same idea to display-only paths: when a caller only needs the right follow-up command for a project, it should ask for that command shape instead of reaching into package-manager state and formatting it itself.

This keeps the slice narrow. It does not change the package-manager values still surfaced for project info output, and it does not reopen the loader/config-link paths.

## How

Add `formatProjectFollowUpCommand(...)` in `packages/app/src/cli/utilities/project-command.ts`.

Use it from:
- deploy completion messaging
- dev scope mismatch warnings
- info follow-up messaging
- generate success messaging

### DIFF
--- a/packages/app/src/cli/services/deploy.ts
+++ b/packages/app/src/cli/services/deploy.ts
@@ -8,6 +8,7 @@ import {AppLinkedInterface} from '../models/app/app.js'
 import {Project} from '../models/project/project.js'
 import {updateAppIdentifiers} from '../models/app/identifiers.js'
 import {DeveloperPlatformClient} from '../utilities/developer-platform-client.js'
+import {formatProjectFollowUpCommand} from '../utilities/project-command.js'
 import {Organization, OrganizationApp} from '../models/organization.js'
 import {reloadApp} from '../models/app/loader.js'
 import {ExtensionRegistration} from '../api/graphql/all_app_extension_registrations.js'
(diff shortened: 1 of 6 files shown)


dmerand:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/Shopify/cli/7243?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#7243** <a href="https://app.graphite.com/github/pr/Shopify/cli/7243?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Shopify/cli/7243?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

afurm:
Solid refactoring — introducing `formatProjectFollowUpCommand` as an explicit boundary between callers and `packageManager` state is the right abstraction. A few questions:

**1. Does the helper guard against invalid/unknown package managers at display time?**

The PR description says this is the "display-only" sibling to #7242's operation-gating. I notice `formatProjectFollowUpCommand` internally calls `project.packageManager` — which may still return `'unknown'` for projects without a `package.json`. Does the helper degrade gracefully in that case (e.g., show a plain `shopify` command), or could users see `unknown shopify app deploy` in the output? If the latter, it might be worth adding a fast-path guard here too, even if display-only.

**2. Is there a race between `Project.load()` and `project.packageManager` access?**

In the `deploy.ts` and `dev.ts` callers, `project` is already loaded. But if these helpers are ever called before the `Project` is fully resolved, does `project.packageManager` always return a valid value, or could it throw? Worth adding a comment or a defensive fallback.

**3. Test coverage for the new helper**

The PR touches 5 files but I don't see a dedicated unit test for `formatProjectFollowUpCommand` itself. Are there integration tests that cover the display paths, or should there be a focused unit test for the helper with various `packageManager` values (`'npm'`, `'yarn'`, `'unknown'`, `''`)?

PR #7244, opened 2026-04-10
Fix function extension build path and info output path resolution

isaacroldan:

Fixes https://community.shopify.dev/t/cli-3-93-0-regression-vitest-wasm-tests-fail-for-rust-function-extensions-base64-written-to-dist-index-wasm-instead-of-raw-binary/33061

The `function info` command was printing the wrong `wasmPath`.   
It should be read from the toml config and not default to `extension.outputPath`  
  
Also, during build we need to make sure to only do the base64 conversion when the `bundlePath` is really different from the outputPath.

- Add support for `build.path` configuration in function configurations by extending the `FunctionConfiguration` interface
- Update the function build logic to avoid unnecessary bundling when the bundle path matches the project output path
- Modify function info display to use the configured build path instead of the default output path
- Add logic to determine the correct function output path by checking for a custom `build.path` or falling back to the default `outputRelativePath`

### How to test your changes?

1. Create a function with a custom `build.path` in its configuration
2. Run the build command and verify the function builds to the correct location
3. Run the function info command and verify it displays the correct output path
4. Test with functions that don't specify a custom build path to ensure backward compatibility

### Checklist

### DIFF
--- a/packages/app/src/cli/services/build/extension.ts
+++ b/packages/app/src/cli/services/build/extension.ts
@@ -161,13 +161,18 @@ export async function buildFunctionExtension(
       await runTrampoline(extension.outputPath)
     }
 
+    const projectOutputPath = joinPath(extension.directory, extension.outputRelativePath)
+
     if (
       fileExistsSync(extension.outputPath) &&
       bundlePath !== extension.outputPath &&
+      bundlePath !== projectOutputPath &&
       dirname(bundlePath) !== dirname(extension.outputPath)
     ) {
+      // Bundle build for deploy: base64-encode into the bundle directory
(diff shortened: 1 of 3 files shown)


isaacroldan:
* **#7244** <a href="https://app.graphite.com/github/pr/Shopify/cli/7244?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Shopify/cli/7244?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

alfonso-noriega (approved):

gonzaloriestra (approved):
Tested that it works as expected 👌

PR #7252, opened 2026-04-13
refactor: remove ESBuildContextManager from app-event-watcher

isaacroldan:
We originally added `ESBuildContextManager`as a faster way to re-build UI extensions -> by keeping in memory esbuild contexts, re-builds are faster.  
  
This was at the expense of having a custom build pipeline in the app watcher just for ui extensions. This is now causing issues because ui-extensions are not following the standard build flow and therefore not running all "build steps".  
  
This PR removes `ESBuildContextManager`and brings back ui-extensions to be compiled like the rest of extensions:

- **Simplifies the dev watcher pipeline** — one build path instead of two.
- **Reduces surface area** — ~375 lines of code and tests deleted.
- **Makes the watcher easier to reason about** — no need to track which extensions use esbuild contexts vs. the default build path.  

The benefit of this manager is negligible for small extensions (we might save single digit miliseconds)


isaacroldan:
* **#7252** <a href="https://app.graphite.com/github/pr/Shopify/cli/7252?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Shopify/cli/7252?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

vividviolet (commented):
This is working but we lose the stack trace for the errorThe new build process only shows generic errors. 

There's also a slight difference in how the extension gets reloaded. I'm pretty sure this is a regression (not related to this PR). When there's a build error the extension isn't supposed to reloaded at all. With this PR we get a reload of the extension with the old code. On main we get a reload with a 500 because the script is missing. This PR improves this behaviour but we're not supposed update the assets at all if the build has failed.



See videos for comparison.

**Before**

[Screen Recording 2026-04-13 at 9.57.33 PM.mov <span class="graphite__hidden">(uploaded via Graphite)</span> ](https://app.graphite.com/user-attachments/video/41bb0228-d2c5-45f2-8240-eaed2701211c.mov)

**After**[Screen Recording 2026-04-13 at 9.56.20 PM.mov <span class="graphite__hidden">(uploaded via Graphite)</span> ](https://app.graphite.com/user-attachments/video/6a19518d-e8ca-451a-9010-70b16cd40dec.mov)

isaacroldan:
Fixed the missing stacktrace, we should look into the extra updates in a new PR

alfonso-noriega (approved):

isaacroldan:
### Merge activity

* **Apr 14, 11:12 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/Shopify/cli/7252).
* **Apr 14, 11:12 AM UTC**: @isaacroldan added this pull request to the GitHub merge queue with [Graphite](https://app.graphite.com/github/pr/Shopify/cli/7252).

PR #7295, opened 2026-04-14
Render task progress bars to stderr by default

amcaplan:
## Problem

The animated loading bar renders to stdout, which causes output flooding in non-TTY environments (AI coding agents, CI) where Ink's ANSI escape codes can't overwrite previous frames. Each animation frame (~every 35ms) gets appended as a new line, producing hundreds of lines of noise for commands like `app generate extension`.

## Solution

Progress bars are status feedback, not program output — they belong on stderr per Unix convention. This change makes `renderTasks` and `renderSingleTask` default to rendering on stderr.

The `LoadingBar` component also now detects whether Ink's output stream is a TTY (via `useStdout`). When it's not — e.g. AI agents that merge stderr via `2>&1` — it shows only the static task title instead of the animated progress bar.

Only 3 files changed — no per-package wrappers needed. Every command using `renderTasks` or `renderSingleTask` gets the fix automatically.

This handles all cases:
- **Interactive terminal**: animated progress bar on stderr ✅
- **`| cat`**: stderr is still a TTY → animated bar visible; stdout is clean ✅
- **AI agent (stdout capture)**: stderr animation, stdout clean ✅
- **AI agent (`2>&1`)**: stderr is non-TTY → static title only, no flooding ✅
- **CI**: Ink's built-in `isCi` already suppresses dynamic rendering ✅

### Before (AI agent)
```
Generating extension ...
▂▁▁▁▁▂▂▃▃▄▄▅▅▆▆▇▇██▇▇▆▆▅▅▄▄▃▃▂▂▁▁▁▁▂▂▃▃▄▄▅▅▆▆▇▇██▇▇▆▆▅▅▄▄▃▃▂▂▁▁
Generating extension ...
▂▂▁▁▁▁▂▂▃▃▄▄▅▅▆▆▇▇██▇▇▆▆▅▅▄▄▃▃▂▂▁▁▁▁▂▂▃▃▄▄▅▅▆▆▇▇██▇▇▆▆▅▅▄▄▃▃▂▂▁
... (hundreds more lines)
```

### After (AI agent)
```
Generating extension ...
Installing dependencies ...
Update shared type definition ...

╭─ success ──────────────────────────────────────────────────────────────╮
│  Your extension was created in extensions/checkout-ui-final-test.     │
╰────────────────────────────────────────────────────────────────────────╯
```

### Note on the theme package
The theme package already has its own `renderTasksToStdErr` helper that passes a custom stderr stream. That helper can be simplified in a follow-up since `renderTasks` now defaults to stderr, but the theme callers sometimes pass `context.stderr` (a custom writable), so they're left as-is for now.

## Type of change
- [x] Bug fix (non-breaking change which fixes an issue)

### DIFF
--- a/packages/cli-kit/src/private/node/ui/components/LoadingBar.tsx
+++ b/packages/cli-kit/src/private/node/ui/components/LoadingBar.tsx
@@ -3,7 +3,7 @@ import useLayout from '../hooks/use-layout.js'
 import {shouldDisplayColors} from '../../../../public/node/output.js'
 import React from 'react'
 
-import {Box, Text} from 'ink'
+import {Box, Text, useStdout} from 'ink'
 
 const loadingBarChar = '▀'
 const hillString = '▁▁▂▂▃▃▄▄▅▅▆▆▇▇██▇▇▆▆▅▅▄▄▃▃▂▂▁▁'
@@ -14,8 +14,29 @@ interface LoadingBarProps {
   noProgressBar?: boolean
 }
 
+/**
+ * Checks whether the stream Ink is rendering to supports cursor movement.
+ * When it doesn't (piped stdout, dumb terminal, non-TTY CI runner, AI coding
+ * agents), every animation frame would be appended as a new line instead of
+ * overwriting the previous one.
+ *
+ * We inspect the stdout object Ink is actually using (via `useStdout`) so the
+ * check stays accurate even when a custom stream is provided through
+ * `renderOptions` (e.g. `renderTasksToStdErr` passes `process.stderr`).
+ *
+ * On real Node streams, `isTTY` is only defined as an own property when the
+ * stream IS a TTY — it's completely absent otherwise, not set to `false`.
+ * So we check the value directly: truthy means TTY, falsy/missing means not.
+ */
+function useOutputSupportsCursor(stdout: NodeJS.WriteStream | Record<string, unknown>): boolean {
+  return Boolean((stdout as Record<string, unknown>).isTTY)
+}
+
 const LoadingBar = ({title, noColor, noProgressBar}: React.PropsWithChildren<LoadingBarProps>) => {
   const {twoThirds} = useLayout()
+  const {stdout} = useStdout()
+  const supportsCursor = useOutputSupportsCursor(stdout)
+
   let loadingBar = new Array(twoThirds).fill(loadingBarChar).join('')
   if (noColor ?? !shouldDisplayColors()) {
     loadingBar = hillString.repeat(Math.ceil(twoThirds / hillString.length))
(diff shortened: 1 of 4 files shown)


packages/cli-kit/src/private/node/ui/components/LoadingBar.tsx
craigmichaelmartin: I think it'd be cleaner like this, rather than having an early return.

```suggestion
      {(!interactive || !noProgressBar) && <TextAnimation text={loadingBar} maxWidth={twoThirds} />}
```

craigmichaelmartin (commented):
Suggestion, but thanks for knocking this out!

packages/cli-kit/src/private/node/ui/components/LoadingBar.tsx
```diff
@@ -16,6 +17,18 @@ interface LoadingBarProps {
...
+
+  // In non-TTY environments (CI, piped output, AI agents), skip the animated
+  // progress bar entirely to avoid flooding the output with animation frames.
+  if (!interactive) {
+    return (
+      <Box flexDirection="column">
+        <Text>{title} ...</Text>
+      </Box>
+    )
+  }
```
isaacroldan: can we animate safely here at least a simple spinner or dot animation? Wouldelp when piping a command :thinking_face:

nickwesselman (commented):
Can we try with just a less intrusive animation? There are "interactive" no-TTY cases (like piping output to `jq`) where the animation is still useful for letting a user know that something is happening  
  
or maybe we should consider just going to a simpler, no-color animation for everyone?

amcaplan:
@nickwesselman I'm trying to make it that it checks whether the output stream _actually used by Ink_ (in this case, should be stderr) is TTY or not. If you pipe to jq, that only should pipe stdout. I don't think this PR works yet, so I'm currently debugging, I think the wrong approach was chosen.

amcaplan:
OK, I think I have this in order:

* `renderTasks` now renders the dynamic output to stderr by default (which it should have done before)
* The component checks the utilized output stream and determines whether or not to display animation based on that

Hence:

```sh
# displays animation
shopify app generate extension --path ~/dev/experiments/lean-supply-app/ --template checkout_ui --name checkout_ui_final_test

# displays animation
shopify app generate extension --path ~/dev/experiments/lean-supply-app/ --template checkout_ui --name checkout_ui_final_test | cat

# doesn't display animation because stderr redirected to stdout which is piped - this is the common AI use case
shopify app generate extension --path ~/dev/experiments/lean-supply-app/ --template checkout_ui --name checkout_ui_final_test 2>&1 | cat
```

graygilmore (approved):
Thanks for the note about renderTasksToStdErr

packages/cli-kit/src/private/node/ui/components/LoadingBar.tsx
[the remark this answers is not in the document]
amcaplan: Switched to useStdout

isaacroldan (approved):

amcaplan:
### Merge activity

* **Apr 16, 11:15 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/Shopify/cli/7295).
* **Apr 16, 11:15 AM UTC**: @amcaplan added this pull request to the GitHub merge queue with [Graphite](https://app.graphite.com/github/pr/Shopify/cli/7295).

PR #7313, opened 2026-04-15
Fix autoupgrade incorrectly skipping in development mode

alfonso-noriega:

When running \`pnpm shopify\` from the CLI repo itself, autoupgrade incorrectly fires as if the CLI were globally installed. This causes contributors to see unexpected upgrade attempts during local development.

**Root cause:** The previous guard was \`!isGlobal && isDevelopment()\`, which only skipped the upgrade when both conditions were true. But \`currentProcessIsGlobal()\` misidentified a pnpm run from the CLI repo as global (since there's no \`shopify.app.toml\` to detect a local project), so the guard never triggered.

Simplifies the guard to just \`isDevelopment()\`. \`isDevelopment()\` checks \`SHOPIFY_CLI_ENV === development\`, which is only set by the CLI repo's pnpm dev scripts — not by a globally-installed binary, even if run from the CLI folder. So contributors are protected and regular users are unaffected.

### How to test your changes?

1. From the root of the CLI repo, run any command: \`pnpm shopify version\`
2. Enable autoupgrade first if needed: \`pnpm shopify upgrade\` → opt in
3. Confirm no upgrade attempt is made after the command completes

You can also force the autoupgrade check with:

\`\`\`
SHOPIFY_CLI_FORCE_AUTO_UPGRADE=1 pnpm shopify version
\`\`\`

Before this fix: upgrade runs (SHOPIFY_CLI_ENV=development is set but \`!isGlobal\` was false, so the guard was skipped).
After this fix: upgrade is skipped (isDevelopment() returns true).

### Checklist

- [ ] The change is user-facing, so I've added a changelog entry with \`pnpm changeset add\`

### DIFF
--- a/packages/cli-kit/src/public/node/upgrade.ts
+++ b/packages/cli-kit/src/public/node/upgrade.ts
@@ -55,7 +55,7 @@ export async function runCLIUpgrade(): Promise<void> {
   const isGlobal = currentProcessIsGlobal()
 
   // Don't auto-upgrade for development mode
-  if (!isGlobal && isDevelopment()) {
+  if (isDevelopment()) {
     outputInfo('Skipping upgrade in development mode.')
     return
   }


alfonso-noriega:
* **#7313** <a href="https://app.graphite.com/github/pr/Shopify/cli/7313?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Shopify/cli/7313?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

gonzaloriestra (commented):
What about just adding another guard clause before to check if `isDevelopment()`? In that case, just skip the auto-upgrade before checking if it's global or not.  
  
This code seems a bit complex and our dev environment is a bit of a special case...

alfonso-noriega:
> What about just adding another guard clause before to check if `isDevelopment()`? In that case, just skip the auto-upgrade before checking if it's global or not.
>
> This code seems a bit complex and our dev environment is a bit of a special case...

​we are currently checking `isDevelopment && !isGlobal` to decide that we are running the local sources of the CLI. The problem with just checking `isDevelopment()` is that we may run global from the CLI folder. 

This fix is actually addressing a problem with `isGlobal` assuming that if we are not in a project dir then we are running a global instance.

But yes, we could actually just check `isDevelopment` and there would be a small amount of times that we fail to run upgrade for CLI contributors only

gonzaloriestra:
`isDevelopment()` simply checks if `SHOPIFY_CLI_ENV === development,`so if you just run a global CLI in the CLI folder, it should be false.

gonzaloriestra (approved):

packages/cli-kit/src/public/node/upgrade.ts
```diff
@@ -55,7 +55,7 @@ export async function runCLIUpgrade(): Promise<void> {
   const isGlobal = currentProcessIsGlobal()
 
   // Don't auto-upgrade for development mode
-  if (!isGlobal && isDevelopment()) {
+  if (isDevelopment()) {
```
craigmichaelmartin, line 58: Is there no test for this which needs updated?

craigmichaelmartin (approved):

PR #7328, opened 2026-04-16
feat(cli): add config autoupgrade [on|off|status] commands

alfonso-noriega:
## What

Adds `shopify config autoupgrade [on|off|status]` commands so users can manage the auto-upgrade preference directly, without having to first run `shopify upgrade`.

## Why

Previously, the only way to configure auto-upgrade was through the interactive prompt that appears during `shopify upgrade`. This is now a first-class config command, consistent with how `config autocorrect` works.

## Changes

- **`shopify config autoupgrade on`** — enables auto-upgrade
- **`shopify config autoupgrade off`** — disables auto-upgrade
- **`shopify config autoupgrade status`** — shows current state (on / off / not yet configured)

Re-exports `getAutoUpgradeEnabled` and `setAutoUpgradeEnabled` from the public `@shopify/cli-kit/node/upgrade` module so the new commands can consume them.

Updates `promptAutoUpgrade()` to skip the prompt when the preference is already set, so `shopify upgrade` won't re-prompt after the user configures it via these commands (or from a previous upgrade run).

### DIFF
--- a/packages/cli/src/cli/commands/config/autoupgrade/constants.ts
+++ b/packages/cli/src/cli/commands/config/autoupgrade/constants.ts
@@ -1,0 +1,5 @@
+export const autoUpgradeStatus = {
+  on: 'Auto-upgrade on. Shopify CLI will update automatically after each command.',
+  off: "Auto-upgrade off. You'll need to run `shopify upgrade` to update manually.",
+  notConfigured: "Auto-upgrade not configured. Run `shopify upgrade` to set your preference.",
+} as const
--- a/packages/cli/src/cli/commands/config/autoupgrade/constants.ts
+++ b/packages/cli/src/cli/commands/config/autoupgrade/constants.ts
@@ -1,5 +1,5 @@
 export const autoUpgradeStatus = {
   on: 'Auto-upgrade on. Shopify CLI will update automatically after each command.',
   off: "Auto-upgrade off. You'll need to run `shopify upgrade` to update manually.",
-  notConfigured: 'Auto-upgrade not configured. Run `shopify upgrade` to set your preference.',
+  notConfigured: 'Auto-upgrade not configured. Run `shopify config autoupgrade on` to enable it.',
 } as const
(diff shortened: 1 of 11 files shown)


alfonso-noriega:
* **#7328** <a href="https://app.graphite.com/github/pr/Shopify/cli/7328?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Shopify/cli/7328?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/cli/src/cli/commands/config/autoupgrade/constants.ts
```diff
@@ -0,0 +1,5 @@
+export const autoUpgradeStatus = {
+  on: 'Auto-upgrade on. Shopify CLI will update automatically after each command.',
+  off: "Auto-upgrade off. You'll need to run `shopify upgrade` to update manually.",
+  notConfigured: 'Auto-upgrade not configured. Run `shopify upgrade` to set your preference.',
```
gonzaloriestra: I'd suggest running `shopify config autoupgrade on` instead

packages/cli/src/cli/commands/config/autoupgrade/on.ts
```diff
@@ -0,0 +1,22 @@
...
+import {setAutoUpgradeEnabled} from '@shopify/cli-kit/node/upgrade'
+import Command from '@shopify/cli-kit/node/base-command'
+import {renderInfo} from '@shopify/cli-kit/node/ui'
+
+export default class AutoupgradeOn extends Command {
+  static summary = 'Enable automatic upgrades for Shopify CLI.'
+
+  static descriptionWithMarkdown = `Enable automatic upgrades for Shopify CLI.
+
+  When auto-upgrade is enabled, Shopify CLI automatically updates to the latest version after each command.
```
gonzaloriestra: What about explaining that it only happens once per day and that it skips major versions?

gonzaloriestra (approved):
Working great!

PR #7331, opened 2026-04-16
Extract @shopify/organizations package

amcaplan:
## Summary

- Creates a new `@shopify/organizations` package with org listing and selection utilities
- Calls the Business Platform Destinations API directly for organization fetching
- Includes `fetchOrganizations` (with base64 GID decoding), `selectOrganizationPrompt` (auto-selects single org), and `selectOrg` helper
- Sets up GraphQL codegen pipeline for the package
- Ready to wire as dependency of `@shopify/store`

**Stack**: PR 2 of 3 — depends on #7330, #7332 adds `store create dev`

## Test plan

- [x] `pnpm nx build organizations` passes
- [x] `pnpm nx type-check organizations` passes
- [x] `cd packages/organizations && pnpm vitest run` — 10/10 tests pass
- [x] `cd packages/store && pnpm vitest run` — no regressions

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #7330, opened 2026-04-16
Extract @shopify/store package from @shopify/cli

amcaplan:
## Summary
- Extracts existing `store:auth` and `store:execute` commands from `packages/cli` into a dedicated `@shopify/store` package
- Follows the same pattern as `@shopify/theme` for package scaffolding, Nx config, and command registration
- Adds `@shopify/store` to the changeset fixed version group

**Stack**: PR 1 of 3 — #7331 extracts `@shopify/organizations`, #7332 adds `store create dev`

## Test plan
- [x] `pnpm nx build store` passes
- [x] `pnpm nx type-check store` passes
- [x] `cd packages/store && pnpm vitest run` — all tests pass
- [x] `cd packages/cli && pnpm vitest run` — no regressions

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #7332, opened 2026-04-16
Add `store create dev` command

amcaplan:
## Summary
- Adds `shopify store create dev` command for creating app development stores via CLI
- Flags: `--name` (required), `--organization` (numeric ID, auto-selects single org), `--json`
- Calls `createAppDevelopmentStore` mutation via Business Platform Organizations API
- Polls for completion with `renderSingleTask` (2s interval, 5-minute timeout)
- Structured JSON output for both success and error paths (supports agentic/automation use cases)
- Sets up GraphQL codegen pipeline for the store package

**Stack**: PR 3 of 3 — depends on [PR 1](https://github.com/Shopify/cli/pull/7330) and [PR 2](https://github.com/Shopify/cli/pull/7331)

## Test plan
- [ ] `pnpm nx build store` passes
- [ ] `pnpm nx type-check store` passes
- [ ] `cd packages/store && pnpm vitest run` — 128/128 tests pass (15 new)
- [ ] `cd packages/cli && pnpm vitest run` — no regressions
- [ ] Manual: `shopify store create dev --name test-store --json` creates store and returns JSON

### DIFF
--- a/packages/store/src/cli/commands/store/create/dev.ts
+++ b/packages/store/src/cli/commands/store/create/dev.ts
@@ -1,0 +1,57 @@
+import {createDevStore} from '../../../services/store/create/dev.js'
+import Command from '@shopify/cli-kit/node/base-command'
+import {globalFlags, jsonFlag} from '@shopify/cli-kit/node/cli'
+import {AbortError} from '@shopify/cli-kit/node/error'
+import {outputResult} from '@shopify/cli-kit/node/output'
+import {Flags} from '@oclif/core'
+
+export default class StoreCreateDev extends Command {
+  static summary = 'Create a new development store.'
+
+  static descriptionWithMarkdown = 'Creates a new app development store in your organization.'
+
+  static description = this.descriptionWithoutMarkdown()
+
+  static flags = {
+    ...globalFlags,
+    ...jsonFlag,
+    name: Flags.string({
+      description: 'Name for the new development store.',
+      required: true,
+      env: 'SHOPIFY_FLAG_STORE_NAME',
+    }),
+    organization: Flags.string({
+      description:
+        'The organization to create the store in (numeric ID). Auto-selects if you belong to a single org.',
+      env: 'SHOPIFY_FLAG_ORGANIZATION',
+    }),
+  }
+
+  async run(): Promise<void> {
+    const {flags} = await this.parse(StoreCreateDev)
+    try {
+      await createDevStore({
+        name: flags.name,
+        organization: flags.organization,
+        json: flags.json,
+      })
+    } catch (error) {
+      if (flags.json && error instanceof AbortError) {
+        outputResult(
+          JSON.stringify(
+            {
+              error: true,
+              message: error.message,
+              nextSteps: error.nextSteps ?? [],
+              exitCode: 1,
+            },
+            null,
+            2,
+          ),
+        )
+        process.exit(1)
+      }
+      throw error
+    }
+  }
+}
--- a/packages/store/src/cli/commands/store/create/dev.ts
+++ b/packages/store/src/cli/commands/store/create/dev.ts
@@ -6,6 +6,8 @@ import {outputResult} from '@shopify/cli-kit/node/output'
 import {Flags} from '@oclif/core'
 
 export default class StoreCreateDev extends Command {
+  static hidden = true
+
   static summary = 'Create a new development store.'
 
   static descriptionWithMarkdown = 'Creates a new app development store in your organization.'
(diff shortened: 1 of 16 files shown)


amcaplan:
* **#7471** <a href="https://app.graphite.com/github/pr/Shopify/cli/7471?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#7332** <a href="https://app.graphite.com/github/pr/Shopify/cli/7332?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Shopify/cli/7332?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#7331** <a href="https://app.graphite.com/github/pr/Shopify/cli/7331?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#7330** <a href="https://app.graphite.com/github/pr/Shopify/cli/7330?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

packages/store/src/cli/commands/store/create/dev.ts
```diff
@@ -0,0 +1,58 @@
...
+
+  static flags = {
+    ...globalFlags,
+    ...jsonFlag,
+    name: Flags.string({
+      description: 'Name for the new development store.',
+      required: true,
+      env: 'SHOPIFY_FLAG_STORE_NAME',
+    }),
+    organization: Flags.string({
```
dmerand: Nit: `organization-id`?

packages/store/src/cli/commands/store/create/dev.ts
```diff
@@ -0,0 +1,58 @@
...
+
+  static summary = 'Create a new development store.'
+
+  static descriptionWithMarkdown = 'Creates a new app development store in your organization.'
+
+  static description = this.descriptionWithoutMarkdown()
+
+  static flags = {
+    ...globalFlags,
+    ...jsonFlag,
```
dmerand, line 20: I'm sure we discussed being able to pass plan? Obviously not for this PR.

alfonso-noriega (approved):
🎩 LGTM

packages/store/src/cli/services/store/create/dev.ts
```diff
@@ -0,0 +1,149 @@
...
+import {selectOrg} from '@shopify/organizations'
+import {businessPlatformOrganizationsRequestDoc} from '@shopify/cli-kit/node/api/business-platform'
+import {ensureAuthenticatedBusinessPlatform} from '@shopify/cli-kit/node/session'
+import {renderSingleTask, renderSuccess} from '@shopify/cli-kit/node/ui'
+import {outputContent, outputResult} from '@shopify/cli-kit/node/output'
+import {AbortError} from '@shopify/cli-kit/node/error'
+import {sleep} from '@shopify/cli-kit/node/system'
+
+const POLL_INTERVAL_SECONDS = 2
+const POLL_TIMEOUT_MS = 5 * 60 * 1000
```
dmerand, line 15: 5 minutes is a long time for a CLI command to hang, is this number vetted?

dmerand (approved):

packages/store/src/cli/commands/store/create/dev.ts
amcaplan: Correct. This initial PR adds the end-to-end flow, but it needs some follow-ups to add the (fairly large) full suite of command options, each of which has some trickiness to it.

packages/store/src/cli/commands/store/create/dev.ts
amcaplan: Yeah, I think that's correct. We should have a single `organization-id` flag that's shared among the various commands that use it.

packages/store/src/cli/services/store/create/dev.ts
amcaplan: Per [stats](https://shopify.slack.com/archives/C0B0186NXUM/p1781471839482089?thread_ts=1781469948.907299&cid=C0B0186NXUM), the highest recorded number in recent history is 215 seconds. But under normal circumstances, it's done by about 15s.

It might be an option just to stop after 2 minutes, and say something like:

> Your store is taking a while to set up. Check back later at <store domain> to see if it's ready.

However, the question is - what is gained by that approach? If it finishes quickly, and polling succeeds - we haven't lost anything. If it finishes slowly, it's probably a better experience to wait.

So I think of the timeout as an indication that something has probably gone terribly wrong. And for that, I think 5 minutes is reasonable.

Also note it doesn't hang for 5 minutes, it shows the rainbow bar with a message indicating the stage of store setup. Eventually it will just say `Finalizing store setup...` for a while. But at least you can observe progress over time.

PR #7341, opened 2026-04-17
Make auto-upgrade enabled by default, remove activation prompt

alfonso-noriega:
🚨 NOTE: DON'T MERGE UNTIL VERSION 4.0.0 🚨

Auto-upgrade was opt-in, requiring users to run `shopify upgrade` and answer a prompt before it would ever activate automatically. This made it effectively a no-op for most users who never ran that command. Auto-upgrade should be on by default so everyone benefits from staying up to date.

- **`getAutoUpgradeEnabled`** **now defaults to** **`true`** — returns `true` when the preference has never been explicitly set (previously returned `undefined`, which disabled auto-upgrade)
- **Removed** **`promptAutoUpgrade`** — the interactive "Enable automatic updates?" prompt that was the only way to activate auto-upgrade; no longer needed
- **`shopify upgrade`** **runs directly** without prompting the user first
- **`shopify config autoupgrade status`** no longer has a "not configured" state — users see either "on" (the default) or "off"
- **`notConfigured`** **constant removed** from `autoUpgradeStatus`
- Users who previously ran `shopify config autoupgrade off` are unaffected — an explicit `false` stored in their config is respected

### How to test your changes?

1. Fresh install (no prior `autoUpgradeEnabled` in config): run any CLI command — it should auto-upgrade after completion without any setup
2. `shopify config autoupgrade status` → should show "on" by default
3. `shopify config autoupgrade off` → should show "off"; subsequent `status` should show "off"
4. `shopify config autoupgrade on` → re-enables; `status` shows "on"
5. `shopify upgrade` → should run the upgrade directly without any prompt

### Checklist

- [x] The change is user-facing, so I've added a changelog entry with `pnpm changeset add`

### DIFF
--- a/packages/cli-kit/src/private/node/conf-store.ts
+++ b/packages/cli-kit/src/private/node/conf-store.ts
@@ -268,11 +268,12 @@ export async function runWithRateLimit(options: RunWithRateLimitOptions, config 
 
 /**
  * Get auto-upgrade preference.
+ * Defaults to true if the preference has never been explicitly set.
  *
- * @returns Whether auto-upgrade is enabled, or undefined if never set.
+ * @returns Whether auto-upgrade is enabled.
  */
-export function getAutoUpgradeEnabled(config: LocalStorage<ConfSchema> = cliKitStore()): boolean | undefined {
-  return config.get('autoUpgradeEnabled')
+export function getAutoUpgradeEnabled(config: LocalStorage<ConfSchema> = cliKitStore()): boolean {
+  return config.get('autoUpgradeEnabled') ?? true
 }
 
 /**
--- a/packages/cli-kit/src/public/node/upgrade.test.ts
+++ b/packages/cli-kit/src/public/node/upgrade.test.ts
@@ -230,11 +230,11 @@ describe('versionToAutoUpgrade', () => {
     expect(versionToAutoUpgrade()).toBeUndefined()
   })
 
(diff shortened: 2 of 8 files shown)


alfonso-noriega:
* **#7341** <a href="https://app.graphite.com/github/pr/Shopify/cli/7341?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Shopify/cli/7341?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

craigmichaelmartin (approved):
Yes!!!!

isaacroldan (approved):

amcaplan (approved):
Tophatting looks good. Glad to see we are finally going fully in this direction!  
  
As this work is news to me, I'd like to know if we have built in a system to shut off auto-upgrades remotely if we release a version we realize is broken (to contain the damage until we can ship a fix).

gonzaloriestra:
@amcaplan I also proposed that [here](https://shopify.slack.com/archives/C0AL2BSRZLL/p1774524535948749?thread_ts=1774456073.076539&cid=C0AL2BSRZLL) and @nickwesselman liked the idea.  
  
I just realized there's an easy way to block it remotely: on auto-upgrade we could run `shopify upgrade` internally (instead of directly running the NPM command or whatever). And then just create a CLI notification with "error" type for that command.

PR #7342, opened 2026-04-17
Generate types for UI extension intents

JoviDeCroock:

UI extensions can now declare **intents** — a
declarative way of saying "this target handles the `create:application/email`
intent" and shipping an input/output/value JSON Schema alongside it.

At runtime the host passes the intent payload to the extension via
`shopify.intents.request`, and the extension acknowledges it with
`shopify.intents.response.ok(data)`. Without type generation, developers have no
way of knowing the shape of `request.data`, what `request.value` contains, or
what `response.ok` expects — they're back to `any` territory and have to cross-
reference the schema file by hand.

We already generate types for the related `tools` concept in the same
`shopify.d.ts` file. This PR brings intents to feature parity.

For any UI-extension target that declares `[[extensions.targeting.intents]]`
entries with a `schema` file, the CLI now:

1. Reads each referenced intent schema JSON file.
2. Validates it against a zod schema (`inputSchema` required, `value` and
   `outputSchema` optional).
3. Compiles each schema into a named TypeScript interface via
   `json-schema-to-typescript`, yielding (for e.g. `create` /
   `application/email`):
   - `CreateApplicationEmailIntentInput`
   - `CreateApplicationEmailIntentValue`
   - `CreateApplicationEmailIntentOutput`
   - `CreateApplicationEmailIntentRequest` — the discriminated request shape
     with literal `action` / `type` and a typed `data` / `value`.
4. Emits two reusable helpers:
   - `ShopifyGeneratedIntentResponse<Data>` — the `ok(data?: Data)` handle.
   - `ShopifyGeneratedIntentsApi<Request, ResponseData>` — request/response
     pair, plus a `ShopifyGeneratedIntentVariants` union of every declared
     intent.
5. Wires the generated types into the module's `shopify` declaration through a
   new `WithGeneratedIntents<T>` wrapper that merges them with whatever the
   target's own `Api['intents']` already exposes (so base host properties like
   `cancel`, etc. survive augmentation).

The wrapper composes with the existing `WithGeneratedTools<T>`:

```ts
const shopify: WithGeneratedTools<
  WithGeneratedIntents<
    import('@shopify/ui-extensions/admin.app.intent.render').Api
  >
>;
```

Duplicate `action:type` pairs within a target throw an `AbortError`, matching
the existing behaviour for duplicate tool names. Missing / invalid schema files
emit a warning and skip generation rather than aborting, consistent with how
tools handles bad input.

Intent types are only emitted for the extension's entry-point file, not for any
file it imports (same behaviour as tools).

Edge cases to spot-check:
- Removing the `schema` file: the CLI should `outputWarn` and regenerate
  `shopify.d.ts` without the intent types (no crash).
- Duplicating an `action:type` pair within a single target: the CLI should
  abort with a clear error.
- Intent schema with a root-level `value` key: the compiled
  `*IntentValue` interface should reflect that schema and be wired into
  `request.value`.

## Tophatting

- Create an intent action extension with `shopify app generate extension`
- Run `shopify app dev`
- Verify that types are generated for `shopify.intents.response.ok()` and `shopify.intents.request.value`

### DIFF
--- a/packages/app/src/cli/models/extensions/specifications/type-generation.ts
+++ b/packages/app/src/cli/models/extensions/specifications/type-generation.ts
@@ -15,6 +15,27 @@ async function loadTypeScript(): Promise<typeof ts> {
 }
 
 const require = createRequire(import.meta.url)
+const uiExtensionsPackage = '@shopify/ui-extensions'
+
+function getGeneratedTypesHelperSurface(target: string): string {
+  const domain = target.toLowerCase().replace(/(::|\.).+$/, '')
+
+  switch (domain) {
+    case 'purchase':
+      return 'checkout'
+    case 'pos':
+      return 'point-of-sale'
+    default:
+      return domain
+  }
+}
+
+export function getGeneratedTypesHelperImportPath(targets: string[]): string {
+  const target = targets[0]
+  if (!target) return uiExtensionsPackage
+
+  return `${uiExtensionsPackage}/${getGeneratedTypesHelperSurface(target)}`
+}
 
 export function parseApiVersion(apiVersion: string): {year: number; month: number} | null {
   const [year, month] = apiVersion.split('-')
@@ -270,6 +291,7 @@ async function buildShopifyType(
   targets: string[],
   resolvedTargetPaths: Map<string, string>,
   {includesTools, includesIntents}: ShopifyTypeOptions,
+  generatedTypesHelperImportPath: string,
 ): Promise<string | null> {
   const baseShopifyType = await buildBaseShopifyType(targets, resolvedTargetPaths)
   if (!baseShopifyType) return null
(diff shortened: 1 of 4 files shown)


olavoasantos (approved):

packages/app/src/cli/models/extensions/specifications/type-generation.ts
```diff
@@ -163,34 +163,142 @@ interface CreateTypeDefinitionOptions {
...
+
+interface GeneratedIntentResponseFallback<GeneratedResponse> {
+  response?: NonNullable<GeneratedResponse>;
+}
+
+interface GeneratedIntentRequestConstraint<Request> {
+  request: Request;
+}
+
+type ReplaceSubscribableValue<Base, Value> = Base extends {value: unknown; subscribe: (callback: (value: infer _) => void) => () => void}
```
vividviolet: Is there a way to put some of these as generics in the library so we can update them outside of the CLI?

packages/app/src/cli/models/extensions/specifications/type-generation.ts
```diff
@@ -15,6 +15,7 @@ async function loadTypeScript(): Promise<typeof ts> {
 }
 
 const require = createRequire(import.meta.url)
+const generatedTypesHelperImportPath = '@shopify/ui-extensions/admin'
```
gonzaloriestra: Why admin? This should work for any kind of UI extension, no?

packages/app/src/cli/models/extensions/specifications/ui_extension.test.ts
```diff
@@ -2491,12 +2496,191 @@ Please check the configuration in ${uiExtension.configurationPath}`),
...
+          },
+          outputSchema: {
+            type: 'object',
+            properties: {
+              success: {type: 'boolean'},
+            },
+          },
+        })
+        await writeFile(joinPath(tmpDir, 'intent-schema.json'), intentSchemaContent)
+        ;(extension.configuration.extension_points[0] as any).intents = [
```
gonzaloriestra: Can we avoid `as any` and use proper types? There are a few appearances

packages/app/src/cli/models/extensions/specifications/type-generation.ts
```diff
@@ -392,8 +530,15 @@ export async function createToolsTypeDefinition(tools: ToolDefinition[]): Promis
...
 async function formatJsonSchemaType(name: string, schema?: object): Promise<string> {
-  const outputType = schema ? await compile(schema, name, {bannerComment: ''}) : `type ${name} = unknown`
-  // The json-schema-to-typescript library adds an export keyword to the type definition, we need to remove it
-  return outputType.startsWith('export ') ? outputType.slice(7) : outputType
+  if (!schema) return `type ${name} = unknown`
+
+  const outputType = await compile(schema, name, {bannerComment: ''})
+  const normalizedOutputType = outputType.startsWith('export ') ? outputType.slice(7) : outputType
+
+  return renameGeneratedType(normalizedOutputType, name)
```
gonzaloriestra, line 574: Could you add a test for this change?

gonzaloriestra (commented):
I added some minor comments, but Copilot also left some good feedback.

Could you also add tophatting instructions? With AI the default PR template is now overwritten sometimes and it has some good points...

gonzaloriestra (approved):

PR #7351, opened 2026-04-20
Add breaking change detection CI check

alfonso-noriega:

PRs that introduce breaking changes (major version bumps, removed CLI commands/flags/env vars, or removed Zod schema fields) currently have no automated gate to ensure the dev_experience team is aware and can coordinate the release timing.

Adds an automated breaking change detection workflow to the PR CI pipeline with two new jobs:

- **`major-change-check`**: Runs a new `workspace/src/major-change-check.js` script that compares the current branch against the `main` baseline and detects:

\-Changesets requesting a major version bump

\-Removed commands, flags, or flag environment variables in the OCLIF manifest (packages/cli/oclif.manifest.json)

\-Removed or deleted Zod schema fields in app/extension configuration models

- When breaking changes are found, the script posts a detailed sticky PR comment via `marocchino/sticky-pull-request-comment` and sets a `has_breaking_changes` output.
- **`major-change-approval`**: A dependent job that only runs when `has_breaking_changes` is `true`. It gates on the `breaking-change-approval` GitHub environment, requiring a member of the `@shopify/dev_experience` team to manually approve the workflow run before the PR can merge.

### How to test your changes?

1. Open a PR that includes a changeset with a `major` bump, removes a CLI command or flag, or removes a Zod schema field.
2. Verify the `major-change-check` job posts a comment to the PR detailing the detected breaking changes.
3. Verify the `major-change-approval` job blocks and requires a `breaking-change-approval` environment approval.
4. Approve the environment and confirm the job passes.
5. Open a PR with no breaking changes and verify both jobs pass without requiring approval.

### Checklist

### DIFF
--- a/workspace/src/major-change-check.js
+++ b/workspace/src/major-change-check.js
@@ -1,0 +1,375 @@
+#!/usr/bin/env node
+
+/**
+ * Detects potential breaking changes in a PR by checking:
+ *
+ * 1. Changesets requesting a major version bump
+ * 2. OCLIF manifest changes: removed commands, removed flags, or removed flag env vars
+ * 3. Zod schema changes: removed or renamed fields in app/extension config schemas
+ *
+ * Compares the current branch against the main branch baseline.
+ * Outputs a GitHub Actions summary and sets a `has_breaking_changes` output.
+ */
+
+import * as path from 'pathe'
+import * as url from 'url'
+import {mkdtemp, rm} from 'fs/promises'
+import os from 'os'
+import {promises as fs} from 'fs'
+import {setOutput} from '@actions/core'
+import {cloneCLIRepository} from './utils/git.js'
+import {logMessage, logSection} from './utils/log.js'
+import fg from 'fast-glob'
+
+const currentDirectory = path.join(url.fileURLToPath(new URL('.', import.meta.url)), '../..')
+
+// ---------------------------------------------------------------------------
+// 1. Check changesets for major bumps
+// ---------------------------------------------------------------------------
+
+async function checkChangesets() {
+  logSection('Checking changesets for major bumps')
+
+  const changesetFiles = await fg('.changeset/*.md', {
+    cwd: currentDirectory,
+    absolute: true,
+    ignore: ['**/README.md'],
+  })
+
+  const majorChangesets = []
+
+  for (const file of changesetFiles) {
+    const content = (await fs.readFile(file, 'utf-8')).trim()
+    // Changeset format: YAML frontmatter between --- markers, with 'package: major'
+    const frontmatterMatch = content.match(/^---\n([\s\S]*?)\n---/)
+    if (!frontmatterMatch) continue
+
+    const frontmatter = frontmatterMatch[1]
+    const lines = frontmatter.split('\n')
+
+    for (const line of lines) {
+      if (line.match(/:\s*major\s*$/)) {
+        majorChangesets.push({
+          file: path.basename(file),
+          line: line.trim(),
+        })
+      }
+    }
+  }
+
+  if (majorChangesets.length > 0) {
+    logMessage(`Found ${majorChangesets.length} changeset(s) with major bump`)
+    for (const cs of majorChangesets) {
+      logMessage(`  ${cs.file}: ${cs.line}`)
+    }
+  } else {
+    logMessage('No major changesets found')
+  }
+
+  return majorChangesets
+}
+
+// ---------------------------------------------------------------------------
+// 2. Check OCLIF manifest for removed commands, flags, or flag env vars
+// ---------------------------------------------------------------------------
+
+async function parseManifest(directory) {
+  const manifestPath = path.join(directory, 'packages/cli/oclif.manifest.json')
+  try {
+    const content = await fs.readFile(manifestPath, 'utf-8')
+    return JSON.parse(content)
+  } catch {
+    return null
+  }
+}
+
+function extractManifestSurface(manifest) {
+  const surface = {commands: {}, envVars: {}}
+  if (!manifest?.commands) return surface
+
+  for (const [cmdName, cmd] of Object.entries(manifest.commands)) {
+    const flags = {}
+    if (cmd.flags) {
+      for (const [flagName, flag] of Object.entries(cmd.flags)) {
+        flags[flagName] = {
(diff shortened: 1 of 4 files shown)


alfonso-noriega:
* **#7351** <a href="https://app.graphite.com/github/pr/Shopify/cli/7351?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Shopify/cli/7351?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

workspace/src/major-change-check.js
```diff
@@ -0,0 +1,375 @@
...
+ * Extracts a simplified map of schema field names from Zod schema source files.
+ * Looks for patterns like: `fieldName: zod.` or `field_name: zod.` in .object({...}) blocks.
+ * This is an intentionally conservative heuristic — AST parsing would be heavier.
+ */
+function extractSchemaFields(content) {
+  const fields = new Set()
+
+  // Match zod.object({ ... }) blocks and extract field names
+  // Handles multi-line objects
+  const objectBlockRegex = /\.object\(\s*\{([\s\S]*?)\}\s*\)/g
```
isaacroldan: 🐛 **Bug**: The regex `/\.object\(\s*\{([\s\S]*?)\}\s*\)/g` is non-greedy, so it terminates at the first `}` followed by `)`. Whenever a Zod schema contains a nested `.object({...})` (or any inline object literal like `{ type: 'app' }`), the outer match closes at the inner brace, and any sibling fields appearing after the nested block are never captured. Because the same regex runs against both the baseline and current files, removing one of those sibling fields produces no diff at all — the breaking change passes silently. This is the worst direction for a heuristic on a gate of this kind: the comment on line 171 calls it 'conservative,' but conservative-on-the-side-of-missing-fields means false negatives, not false positives. Nested `.object()` composition is idiomatic in `packages/app/src/cli/models/extensions/specifications/**`, so this isn't theoretical — it will fire in real schemas.

**Suggestion:** Replace the regex with a brace-counting walker that tracks `{`/`}` depth to find the matched closing brace, or parse the file with the TypeScript compiler (already in the toolchain) / `@typescript-eslint/parser`. At minimum, add a fixture test mirroring `app_config_app_access.ts` and verify all top-level field names round-trip through `extractSchemaFields`.

workspace/src/major-change-check.js
```diff
@@ -0,0 +1,375 @@
...
+}
+
+async function checkSchemas(baselineDirectory) {
+  logSection('Checking Zod schemas for removed fields')
+
+  const schemaGlob = 'packages/app/src/cli/models/**/specifications/**/*.ts'
+  const appModelGlob = 'packages/app/src/cli/models/app/app.ts'
+  const ignorePatterns = ['**/*.test.ts', '**/*.test-data.ts']
+
+  const currentSchemaFiles = await fg([schemaGlob, appModelGlob], {
```
isaacroldan: 🐛 **Bug**: Line 201 fetches `currentSchemaFiles` via `fg(...)`, but the function never references the variable — the loop on line 212 iterates only `baselineSchemaFiles`, and per-file current contents are read directly via `fs.readFile(currentPath, ...)` inside the loop. This costs an extra glob walk per CI run and reads as a leftover from a refactor where added-schema-file detection was dropped. If detecting newly added schema files was the intent, the wiring is missing; otherwise the block should be deleted.

**Suggestion:** Remove the dead `currentSchemaFiles` block (lines 201–204), or wire it back into the loop if added-file detection was the original intent.

workspace/src/major-change-check.js
```diff
@@ -0,0 +1,375 @@
...
+      // File was deleted — all fields are removed
+      const fields = extractSchemaFields(baselineContent)
+      if (fields.size > 0) {
+        removedFields.push({file, fields: [...fields], type: 'file_deleted'})
+      }
+      continue
+    }
+
+    // Only check files that actually contain zod schemas
+    if (!baselineContent.includes('zod.object(') && !baselineContent.includes('.object(')) continue
```
isaacroldan: 🐛 **Bug**: The substring guard on line 235 only proceeds if the source contains `zod.object(` or `.object(`. A large portion of spec files in `packages/app/src/cli/models/extensions/specifications/` introduce their top-level fields via `BaseSchemaWithoutHandle.extend({...})` and only use `.object()` for sub-shapes. In `app_config_app_home.ts`, for example, `application_url` and `embedded` are added through `.extend()` and never appear inside a `.object()` call — the heuristic will never see them, so removing them is undetectable. Combined with the nested-object regex bug, this produces compounding false negatives in exactly the spec files this gate is supposed to protect.

**Suggestion:** Extend `extractSchemaFields` to also match `.extend(\s*\{...\}\s*)` (with a brace-balanced walker) and update the substring filter at line 235 accordingly. Long-term, prefer an AST-based extraction over regex.

workspace/src/major-change-check.js
```diff
@@ -0,0 +1,375 @@
...
+}
+
+// ---------------------------------------------------------------------------
+// Main
+// ---------------------------------------------------------------------------
+
+const tmpDir = await mkdtemp(path.join(os.tmpdir(), 'major-change-check-'))
+
+try {
+  const baselineDirectory = await cloneCLIRepository(tmpDir)
```
isaacroldan: ⚡ **Performance**: `cloneCLIRepository` (in `workspace/src/utils/git.js`) unconditionally runs `pnpm install` and `pnpm build` on the cloned `main`. `type-diff.js` needs that build because it diffs `dist/**/*.d.ts`, but this script consumes only the committed OCLIF manifest (`packages/cli/oclif.manifest.json` is git-tracked) and `.ts` source files for schemas — none of which are produced by the build. Reusing the same utility burns ~5–10 minutes of CI per PR for no benefit. Because the function is shared, the clean fix is to parameterize it rather than fork.

**Suggestion:** Add `{install?: boolean, build?: boolean}` options to `cloneCLIRepository` and pass `{install: false, build: false}` from this script. Alternatively introduce a sibling `cloneCLIRepositoryShallow` for the source-only case.

isaacroldan (commented):
---

isaacroldan (approved):

PR #7377, opened 2026-04-22
Pass Storefront API requests through the dev proxy as-is

frandiox:

Requests to /api/{version|unstable}/graphql.json were returning 401 because the dev proxy injected the SFR devtools bearer via the Authorization header. The public Storefront API rejects that token — it expects X-Shopify-Storefront-Access-Token from the caller.

Forward these requests untouched (no Authorization, Cookie, referer, or _fd/pb params) so the caller's own auth passes through. Introduces isPassthroughRequest(event) as the extension point for future pass-through rules.

### How to test your changes?

Make a request to SFAPI from the Chrome dev tools. Without this PR it will return 401.

### Checklist


bashu-shopify (approved):

packages/theme/src/cli/utilities/theme-environment/proxy.ts
graygilmore, line 1: Before this gets merged let's get @EvilGenius13 to look at it who has done a bunch of header stuff recently to prevent 401 and 429s.

EvilGenius13: This should be okay. What matters most is that regular proxy requests to storefront maintain having user agent and bearer token on all requests. This bypass looks safe.

EvilGenius13 (approved):

PR #7384, opened 2026-04-23
Auto-detect and linkify URLs in plain-string TokenizedText items

ryancbahan:

Long URLs embedded as plain strings in CLI error messages wrap across multiple lines inside the fixed-width error banner (~53 chars on an 80-col terminal). The wrap splits the URL with the border's `│` characters, producing output that is neither clickable nor copy-pasteable as a single URL.

Three small changes in `packages/cli-kit`:

1. **`TokenizedText.tsx`** — New `renderStringWithLinks` helper auto-detects `http(s)://…` URLs in plain-string tokens, strips common trailing sentence punctuation, and routes each URL through the existing `<Link>` component. This is the chokepoint through which all banner/alert bodies render, so server-returned error strings benefit automatically.
2. **`FatalError.tsx`** — The `error.message` fallback (used when a `FatalError` is constructed with a plain `OutputMessage` instead of a `TokenItem`) now flows through `TokenizedText` instead of a bare `<Text>`, picking up the same URL treatment.
3. **`Link.tsx`** — Two fixes to the non-OSC-8 fallback path, both limited to behavior inside a `LinksContext` (i.e. inside a Banner):
    - When a link has no label, render just `[N]` as the in-body anchor instead of `url [N]` (prevents the URL from being visible in-body where it would wrap).
    - Always use the footnote mechanism when a `LinksContext` is present, regardless of whether a label was provided (previously the method short-circuited for label-less links and emitted the raw URL inline).

**Net effect**

Before:

```
│ See specification requirements: https://shopify.dev/docs/apps/build/sales- │
│ channels/channel-config-extension#specification-properties                 │
```

After:

```
│ See specification requirements: [1]                                        │
╰────────────────────────────────────────────────────────────────────────────╯
[1] https://shopify.dev/docs/apps/build/sales-channels/channel-config-extension#specification-properties
```

On OSC 8 terminals (iTerm2, modern Terminal.app, VS Code, Warp, Kitty, WezTerm, Ghostty) the `[1]` anchor is a clickable hyperlink to the full URL. On terminals without OSC 8 support, the URL appears outside the bordered box in the footnote block, where it can still wrap at terminal width but is not interleaved with `│` characters, so copy-paste recovers the full URL.

### How to test your changes?

**Unit coverage added**

- `TokenizedText.test.tsx`: 5 tests — no-URL passthrough, URL preserved without OSC 8, URL wrapped in OSC 8 when supported, multiple URLs, trailing punctuation stripped.
- `Alert.test.tsx` and `FatalError.test.tsx`: inline snapshot integration tests that render a long-URL message inside a real `Banner` / `FatalError` and lock in the `[1]` in-body + URL-in-footnote layout.

**Visual QA in a real terminal**

1. `pnpm --filter @shopify/cli-kit build`
2. Create a small script that calls `renderError` / `renderFatalError` with a plain-string body containing a long URL, then run it with `node`.
3. Verify the URL is no longer interleaved with `│`, the `[1]` anchor is clickable (OSC 8 terminals), and copy-pasting the footnote URL yields a single clean URL.
4. Force the non-OSC-8 path with `FORCE_HYPERLINK=0` and confirm the `[N]` + footnote layout still renders cleanly.

### Post-release steps

None.

### Checklist

- [x] I've considered possible cross-platform impacts (Mac, Linux, Windows) — change is pure text rendering; OSC 8 detection is delegated to the existing `supports-hyperlinks` dependency.
- [x] I've considered possible [documentation](https://shopify.dev) changes — none needed; the fix is transparent to callers.
- [x] I've considered analytics changes to measure impact — not applicable.
- [x] The change is user-facing — added a `patch` changeset (bug fix).

### DIFF
--- a/packages/cli-kit/src/private/node/ui/components/Link.tsx
+++ b/packages/cli-kit/src/private/node/ui/components/Link.tsx
@@ -12,16 +12,15 @@ interface LinkProps {
 
 function link(label: string | undefined, url: string, linksContext: LinksContextValue | null) {
   if (!supportsHyperlinks.stdout) {
-    if (url === (label ?? url)) {
-      return url
-    }
-
     if (linksContext === null) {
+      if (url === (label ?? url)) {
+        return url
+      }
       return label ? `${label} ${chalk.dim(`( ${url} )`)}` : url
     }
 
     const linkId = linksContext.addLink(label, url)
-    return `${label ?? url} [${linkId}]`
+    return label ? `${label} [${linkId}]` : `[${linkId}]`
   }
 
   return ansiEscapes.link(label ?? url, url)
--- a/packages/cli-kit/src/private/node/ui/components/TokenizedText.test.tsx
+++ b/packages/cli-kit/src/private/node/ui/components/TokenizedText.test.tsx
@@ -1,12 +1,31 @@
 import {tokenItemToString, TokenizedText} from './TokenizedText.js'
+import {LinksContext, Link} from '../contexts/LinksContext.js'
 import {unstyled} from '../../../../public/node/output.js'
 import {render} from '../../testing/ui.js'
 import {describe, expect, test, vi} from 'vitest'
 import supportsHyperlinks from 'supports-hyperlinks'
 
-import React from 'react'
+import React, {FunctionComponent, useRef} from 'react'
 
 vi.mock('supports-hyperlinks')
+
+const WithLinksContext: FunctionComponent<{children: React.ReactNode}> = ({children}) => {
+  const links = useRef<Record<string, Link>>({})
+  return (
+    <LinksContext.Provider
+      value={{
+        links,
+        addLink: (label, url) => {
+          const newId = (Object.keys(links.current).length + 1).toString()
+          links.current = {...links.current, [newId]: {label, url}}
+          return newId
+        },
+      }}
+    >
+      {children}
+    </LinksContext.Provider>
+  )
+}
 
 describe('TokenizedText', async () => {
   test('renders arrays of items separated by spaces', async () => {
@@ -64,25 +83,38 @@ describe('TokenizedText', async () => {
     test('renders strings without URLs unchanged', async () => {
       vi.mocked(supportsHyperlinks).stdout = false
 
-      const {lastFrame} = render(<TokenizedText item="no link here, just text" />)
+      const {lastFrame} = render(
+        <WithLinksContext>
+          <TokenizedText item="no link here, just text" />
+        </WithLinksContext>,
+      )
 
       expect(lastFrame()).toBe('no link here, just text')
     })
 
-    test('preserves a URL intact when the terminal does not support hyperlinks', async () => {
+    test('replaces a URL with a `[N]` footnote anchor in the body when the terminal does not support hyperlinks', async () => {
       vi.mocked(supportsHyperlinks).stdout = false
       const url = 'https://shopify.dev/docs/apps/build/sales-channels/channel-config-extension#specification-properties'
 
-      const {lastFrame} = render(<TokenizedText item={`See specification requirements: ${url}`} />)
+      const {lastFrame} = render(
+        <WithLinksContext>
+          <TokenizedText item={`See specification requirements: ${url}`} />
+        </WithLinksContext>,
+      )
 
-      expect(lastFrame()).toContain(url)
+      expect(lastFrame()).toBe('See specification requirements: [1]')
+      expect(lastFrame()).not.toContain(url)
     })
 
     test('wraps detected URLs in OSC 8 escapes when the terminal supports hyperlinks', async () => {
       vi.mocked(supportsHyperlinks).stdout = true
       const url = 'https://example.com/docs'
 
-      const {lastFrame} = render(<TokenizedText item={`visit ${url} now`} />)
+      const {lastFrame} = render(
+        <WithLinksContext>
+          <TokenizedText item={`visit ${url} now`} />
+        </WithLinksContext>,
+      )
 
       expect(lastFrame()).toContain(`]8;;${url}${url}]8;;`)
     })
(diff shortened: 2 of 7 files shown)


ryancbahan:
* **#7384** <a href="https://app.graphite.com/github/pr/Shopify/cli/7384?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Shopify/cli/7384?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/cli-kit/src/private/node/ui/components/TokenizedText.tsx
```diff
@@ -152,9 +152,40 @@ interface TokenizedTextProps {
  * `TokenizedText` renders a text string with tokens that can be either strings,
  * links, and commands.
  */
+const URL_REGEX = /https?:\/\/\S+/g
+const URL_TRAILING_PUNCTUATION = /[.,;:!?)\]}>'"]+$/
```
isaacroldan: 🐛 **Bug**: `URL_TRAILING_PUNCTUATION = /[.,;:!?)\]}>'"]+$/` unconditionally strips `)`, `]`, `}`, and `>` from the end of a matched URL, but these characters are legal URL suffixes. Real-world examples that would break: Wikipedia disambiguation links like `https://en.wikipedia.org/wiki/Ruby_(programming_language)`, MDN `Function()` pages, Microsoft Learn URLs using parens in paths, and query strings ending in JSON-array-style values. Because this helper is now the chokepoint for every plain string passed through `TokenizedText`, any such URL in a server-returned error message will have its closing bracket stripped, and the `<Link>` component will emit an OSC 8 escape pointing at a broken target (with the trailing `)` rendered outside the link). None of the new tests exercise URLs containing `()`, so this silent corruption is uncovered.

**Suggestion:** Narrow the character class to unambiguous sentence punctuation, or balance-count parens before stripping. Minimal fix:

```suggestion
const URL_TRAILING_PUNCTUATION = /[.,;:!?'"]+$/
```

And add a test for `https://en.wikipedia.org/wiki/Foo_(bar)` to lock the behavior in.

packages/cli-kit/src/private/node/ui/components/TokenizedText.tsx
```diff
@@ -152,9 +152,40 @@ interface TokenizedTextProps {
...
+  if (cursor < str.length) {
+    parts.push(<Text key="tail">{str.slice(cursor)}</Text>)
+  }
+  return <Text>{parts}</Text>
+}
+
 const TokenizedText: FunctionComponent<TokenizedTextProps> = ({item}) => {
   if (typeof item === 'string') {
-    return <Text>{item}</Text>
+    return renderStringWithLinks(item)
```
isaacroldan: 📐 **Design**: The PR is framed as a Banner/alert fix, but `renderStringWithLinks` runs on every plain-string token passing through `TokenizedText`. That includes `Alert`, `FatalError`, `List`, `TabularData`, `DangerousConfirmationPrompt`, `Prompts/PromptLayout`, `Prompts/InfoMessage`, and `TextPrompt`. In non-Banner contexts there's no `LinksContext`, so the `Link` component falls back to raw-URL output for non-OSC-8 terminals (preserved) but emits OSC 8 escapes on OSC-8-capable terminals (behavior change). A concrete example: the placeholder URL in `packages/app/src/cli/services/dev/urls.ts:86` (`Valid format: "https://my-tunnel-url:port"`) will now be rendered as a clickable escape to a non-existent host. Reviewers of downstream packages won't catch this from the PR description alone.

**Suggestion:** Either (a) scope the auto-detection to contexts that own a `LinksContext` by threading an explicit flag through (e.g. an `autoLinkify` prop or opt-in from Banner/Alert/FatalError), or (b) keep the current broad behavior but update the PR description/changeset and add snapshot coverage for non-Banner callers (`TabularData`, `List`, prompts) so the behavior change is visible at review time.

isaacroldan (commented):
---

packages/cli-kit/src/private/node/ui/components/TokenizedText.tsx
isaacroldan: I tested this and does indeed break other alerts:

isaacroldan: You can trigger this one by running:

```
pnpm shopify app dev --tunnel-url https://wrong
```

packages/cli-kit/src/private/node/ui/components/TokenizedText.tsx
johnnie-chen: +1 on this

packages/cli-kit/src/private/node/ui/components/TokenizedText.test.tsx
```diff
@@ -57,6 +60,55 @@ describe('TokenizedText', async () => {
...
+    test('wraps detected URLs in OSC 8 escapes when the terminal supports hyperlinks', async () => {
+      vi.mocked(supportsHyperlinks).stdout = true
+      const url = 'https://example.com/docs'
+
+      const {lastFrame} = render(<TokenizedText item={`visit ${url} now`} />)
+
+      expect(lastFrame()).toContain(`]8;;${url}${url}]8;;`)
+    })
+
+    test('detects multiple URLs in the same string', async () => {
```
johnnie-chen, line 122: Might be worth adding edge test cases where there's back to back URLs

packages/cli-kit/src/private/node/ui/components/Link.tsx
```diff
@@ -12,16 +12,15 @@ interface LinkProps {
 
 function link(label: string | undefined, url: string, linksContext: LinksContextValue | null) {
   if (!supportsHyperlinks.stdout) {
-    if (url === (label ?? url)) {
-      return url
-    }
-
     if (linksContext === null) {
```
Catherine-Fu, line 19: can we add unit tests for changes in this file?

packages/cli-kit/src/private/node/ui/components/TokenizedText.tsx
```diff
@@ -152,9 +152,40 @@ interface TokenizedTextProps {
  * `TokenizedText` renders a text string with tokens that can be either strings,
  * links, and commands.
  */
+const URL_REGEX = /https?:\/\/\S+/g
```
Catherine-Fu, line 156: Are there any existing libraries we can use to validate URLs?

ryancbahan: we're actively trying to reduce lib usage given the amount of supply chain attacks.

Catherine-Fu (approved):
The expected behaviour is that the link in CLI validation messages should be easily clickable or copyable as a complete, valid URL. 

I don’t have much context on the CLI library, so I’d prefer to have another reviewer take a look and approve this.

PR #7378, opened 2026-04-22
address png encoding bug, path changes

JoshuaWhite1:
Closes https://github.com/shop/issues-admin-extensibility/issues/2452

### WHY are these changes introduced?

When serving static assets for UI extensions during development, the dev server only looked in the build output directory and the extension's root source directory. Extensions that declare a per-extension-point `assets` folder (via the `assets` config key) had no way to serve those static files through the dev server, leaving assets like images inaccessible at runtime.

### WHAT is this pull request doing?

Adds a third candidate lookup path when resolving extension assets: the per-extension-point `assets` directory declared in the extension configuration. The resolution order is now:

1. Build output directory (e.g. `dist/handle.js`)
2. Extension source root (e.g. `tools`, `instructions`)
3. Per-extension-point static asset folders declared via the `assets` config key (e.g. `./assets/logo.png`)

This mirrors how the admin spec's `static_root` serves a whole directory of static files.

Additionally, `readFile` is called with an explicit options argument (`{}`) so that it returns a `Buffer` rather than a `string`, and test assertions are updated to use `.toString()` accordingly.

A new test covers the fallback to the extension-point `assets` folder, verifying that a static asset (e.g. `logo.png`) is correctly resolved and served with the appropriate `Content-Type`.

### How to test your changes?

1. Create a UI extension with an `extension_points` entry that includes an `assets` key pointing to a local directory (e.g. `assets: './assets'`).
2. Place a static file (e.g. `logo.png`) in that directory.
3. Run `shopify app dev` and request the asset via the extension dev server URL.
4. Confirm the asset is served correctly with the right `Content-Type` header.

Alternatively, run the existing test suite:
```
pnpm test packages/app/src/cli/services/dev/extension/server/middlewares.test.ts
```

### Checklist

- [ ] I've considered possible cross-platform impacts (Mac, Linux, Windows)
- [ ] I've considered possible [documentation](https://shopify.dev) changes
- [ ] I've considered analytics changes to measure impact
- [ ] The change is user-facing — I've identified the correct bump type (`patch` for bug fixes · `minor` for new features · `major` for [breaking changes](../CONTRIBUTING.md#what-counts-as-a-breaking-change)) and added a changeset with `pnpm changeset add`

PR #7363, opened 2026-04-21
Add assets config key support for UI extension points

JoshuaWhite1:
Part of https://github.com/shop/issues-admin-extensibility/issues/2439

### WHY are these changes introduced?

Adds support for an `assets` field on UI extension points, allowing extension point configurations to declare associated assets.

### WHAT is this pull request doing?

- Adds an optional `assets` field to `NewExtensionPointSchema` in the extension schemas
- Passes the `assets` property through when transforming extension point targeting configurations in `UIExtensionSchema`
- Registers `extension_points[].assets` as a `configKey` in the UI extension specification so it is properly grouped and tracked per extension point target
- Updates `DevNewExtensionPointSchema` to omit `assets` from the base type and redefine it as a map of named `Asset` objects (with `lastUpdated`), preserving the existing dev payload structure while accommodating the new schema field

### How to test your changes?

1. Create a UI extension with extension points that include an `assets` field in the configuration
2. Run `shopify app dev` and verify the extension point payload includes the `assets` field correctly
3. Verify that the dev server payload maps `assets` to the expected `{ [name: string]: Asset }` structure

### Checklist

- [ ] I've considered possible cross-platform impacts (Mac, Linux, Windows)
- [ ] I've considered possible [documentation](https://shopify.dev) changes
- [ ] I've considered analytics changes to measure impact
- [ ] The change is user-facing — I've identified the correct bump type (`patch` for bug fixes · `minor` for new features · `major` for [breaking changes](../CONTRIBUTING.md#what-counts-as-a-breaking-change)) and added a changeset with `pnpm changeset add`

PR #7340, opened 2026-04-17
support bundle folder in the bundle ui step config

alfonso-noriega:
### WHY are these changes introduced?

Fixes #0000 

### Checklist

- [ ] I've considered possible cross-platform impacts (Mac, Linux, Windows)
- [ ] I've considered possible [documentation](https://shopify.dev) changes
- [ ] I've considered analytics changes to measure impact
- [ ] The change is user-facing — I've identified the correct bump type (`patch` for bug fixes · `minor` for new features · `major` for [breaking changes](../CONTRIBUTING.md#what-counts-as-a-breaking-change)) and added a changeset with `pnpm changeset add`

PR #7386, opened 2026-04-23
Fix Dev Server to allow serving static assets from a shared folder outside of the extension directory

vividviolet:

When two extension points within the same extension declare assets whose source basenames collide (e.g. both reference a `tools.json`, or both compile to `main.js`), the dev server previously served whichever file happened to win the filesystem race. The URL shape (`/assets/<basename>`) gave no way to distinguish which extension point a request belonged to, so one target's asset would silently shadow another's.

Additionally, the middleware served files directly from the extension's source directory, which allowed path traversal to files outside the extension's bundle and made it impossible to serve `include_assets`-copied files that had been renamed by `uniqueBasename` during the build step.

**Opaque, target-scoped asset URLs**

Asset URLs are now emitted as `/assets/<target>/<assetKey>[.ext]` instead of `/assets/<basename>`. This makes every URL unique per extension point, even when two targets share a source file.

**Asset resolver**

A new `AssetResolver` (`Map<string, string>`) is introduced per extension. During payload generation, each emitted URL subpath is registered in the resolver alongside the output-relative filesystem path the middleware should read. The resolver is cleared and rebuilt on every payload regeneration so stale entries from removed targets or renamed assets don't persist.

The resolver is stored in `ExtensionsPayloadStore` (keyed by `devUUID`) and exposed via `getAssetResolver(devUUID)`. It is cleaned up when an extension is deleted.

**Middleware rewrite**

`getExtensionAssetMiddleware` now:
1. Looks up the request's `assetPath` in the extension's resolver to find the actual output-relative filename (falling back to the raw path for direct compiled-artefact fetches).
2. Resolves the candidate path against the extension's `outputDir` (the bundle directory that `include_assets` populates).
3. Rejects any path — whether from the URL or from a resolver value — that escapes `outputDir` via traversal, returning 404 rather than 403 to avoid leaking directory structure.

**Static directory assets**

A new `staticAssetsMapper` handles directory-valued config entries (e.g. `assets = "./assets"`). It emits a directory-prefix URL with a trailing slash and registers one resolver entry per copied file, including nested subdirectories.

**`resolveOutputDir` extracted**

The logic for deriving an output directory from `outputPath` (file with extension → `dirname`; no extension → path itself) is extracted into a shared `resolveOutputDir` helper used by both the build step and the middleware.

### How to test your changes?

1. Create a UI extension with two extension points that both declare a `tools` asset pointing at different source files.
2. Run `shopify app dev` and observe that each extension point's payload contains a distinct URL (`/<target-a>/tools.json` vs `/<target-b>/tools.json`).
3. Fetch each URL from the dev server and confirm the correct file content is returned for each target.
4. Verify that requesting a file path not present in the resolver (e.g. a raw `../secret.txt`) returns 404.
5. Create an extension point with `assets = "./assets"` containing nested files; confirm each file is individually servable via its resolver-mapped URL.

### Checklist

### DIFF
--- a/packages/app/src/cli/services/build/steps/include-assets-step.ts
+++ b/packages/app/src/cli/services/build/steps/include-assets-step.ts
@@ -1,8 +1,8 @@
-import {generateManifestFile} from './include-assets/generate-manifest.js'
+import {generateManifestFile, resolveOutputDir} from './include-assets/generate-manifest.js'
 import {copyByPattern} from './include-assets/copy-by-pattern.js'
 import {copySourceEntry} from './include-assets/copy-source-entry.js'
 import {copyConfigKeyEntry} from './include-assets/copy-config-key-entry.js'
-import {joinPath, dirname, extname, sanitizeRelativePath} from '@shopify/cli-kit/node/path'
+import {joinPath, sanitizeRelativePath} from '@shopify/cli-kit/node/path'
 import {z} from 'zod'
 import type {LifecycleStep, BuildContext} from '../client-steps.js'
 
@@ -123,9 +123,7 @@ export async function executeIncludeAssetsStep(
 ): Promise<{filesCopied: number}> {
   const config = IncludeAssetsConfigSchema.parse(step.config)
   const {extension, options} = context
-  // When outputPath is a file (e.g. index.js, index.wasm), the output directory is its
-  // parent. When outputPath has no extension, it IS the output directory.
-  const outputDir = extname(extension.outputPath) ? dirname(extension.outputPath) : extension.outputPath
+  const outputDir = resolveOutputDir(extension.outputPath)
 
   const aggregatedPathMap = new Map<string, string | string[]>()
   // Track basenames written across all configKey entries in this build to detect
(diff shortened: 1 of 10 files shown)


vividviolet:
* **#7386** <a href="https://app.graphite.com/github/pr/Shopify/cli/7386?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Shopify/cli/7386?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#7378** <a href="https://app.graphite.com/github/pr/Shopify/cli/7378?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#7363** <a href="https://app.graphite.com/github/pr/Shopify/cli/7363?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#7340** <a href="https://app.graphite.com/github/pr/Shopify/cli/7340?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

alfonso-noriega (approved):

isaacroldan (approved):

PR #7389, opened 2026-04-24
Allow preserving file paths when including assets from config key entry

vividviolet:

When multiple `configKey` entries reference different source directories or files that produce the same output-relative path, the last writer silently wins. This makes it impossible to detect accidental file collisions when directory structures need to be preserved in the output.

Adds a `preserveFilePaths` option (default `false`) to `configKey` inclusion entries. When enabled:

- Files are written to the output directory preserving their relative paths from the source directory, rather than being flattened to unique basenames.
- A collision check is enforced across all entries in the same build: if two different sources would produce the same output-relative file path, an error is thrown immediately with a descriptive message indicating which file caused the conflict.
- A shared `usedFiles` set is threaded through all `configKey` entries so cross-entry collisions are detected, not just within-entry ones.
- The existing `usedBasenames` deduplication logic is preserved for entries where `preserveFilePaths` is `false`.

The collision check is extracted into a single `assertNoCollision` helper used by both the directory and single-file copy branches.

### How to test your changes?

Unit tests should be sufficient

### Checklist

### DIFF
--- a/packages/app/src/cli/services/build/steps/include-assets/copy-config-key-entry.ts
+++ b/packages/app/src/cli/services/build/steps/include-assets/copy-config-key-entry.ts
@@ -1,6 +1,7 @@
 import {joinPath, basename, relativePath, extname} from '@shopify/cli-kit/node/path'
 import {glob, copyFile, copyDirectoryContents, fileExists, mkdir, isDirectory} from '@shopify/cli-kit/node/fs'
 import {outputContent, outputDebug, outputToken} from '@shopify/cli-kit/node/output'
+import {AbortError} from '@shopify/cli-kit/node/error'
 import type {BuildContext} from '../../client-steps.js'
 
 /**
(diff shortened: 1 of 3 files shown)


vividviolet:
* **#7389** <a href="https://app.graphite.com/github/pr/Shopify/cli/7389?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Shopify/cli/7389?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/app/src/cli/services/build/steps/include-assets/copy-config-key-entry.ts
```diff
@@ -98,6 +119,18 @@ export async function copyConfigKeyEntry(config: {
...
 }
 
+/**
+ * Throws if `path` is already present in `used`. Shared between directory and
+ * single-file branches so the error message and check shape stay in one place.
+ */
+function assertNoCollision(path: string, sourcePath: string, used: Set<string>): void {
+  if (used.has(path)) {
+    throw new Error(
+      `File collision with preserveFilePaths enabled: '${path}' from '${sourcePath}' would overwrite a file copied from a different source. Rename or relocate the conflicting file.`,
```
isaacroldan: This message leaks implementation details to the user.  
It should also be an `AbortError` and not a plain `Error`  
  
I would just remove the mention to "preserveFilePaths"

vividviolet: Fixed!

isaacroldan (approved):

PR #7391, opened 2026-04-24
Respect user .theme-check.yml in theme app extension builds

erikmay:
## Summary

`runThemeCheck` in `packages/app/src/cli/services/build/theme-check.ts` hardcodes the config identifier `'theme-check:theme-app-extension'`, which bypasses theme-check-node's user-config discovery (`findConfigPath`). As a result, a `.theme-check.yml` placed in the theme app extension root is ignored during `shopify app dev` and `shopify app deploy`, even though the standalone `shopify theme check` command respects it.

This PR makes `runThemeCheck` check for `.theme-check.yml` in the extension directory and pass `undefined` to `check()` when present, letting theme-check-node auto-discover the user config. When absent, the existing bundled `theme-check:theme-app-extension` defaults are applied exactly as before — no behavior change for extensions without a user config.

## Motivation

Theme app extensions only allow a fixed set of top-level directories (`assets/`, `blocks/`, `snippets/`, `locales/`, `config/`). Authors who want a separate source directory (e.g. uncompiled Liquid / TS / Tailwind that gets built into `blocks/` + `assets/`) currently have two options:

1. Dot-prefix the source dir (`.src/`) so theme-check's `**/*.{liquid,json}` glob skips it.
2. Fork the CLI.

Option 1 works but is surprising and fragile. Option 2 is heavy. With this change, a plain `src/` source directory works as expected:

```yaml
# extensions/my-extension/.theme-check.yml
extends: theme-check:theme-app-extension
ignore:
  - "src/**"
  - "node_modules/**"
```

## Changes

- `packages/app/src/cli/services/build/theme-check.ts`: check for `.theme-check.yml` and fall through to user-config discovery when present.
- Added a changeset (`@shopify/app` patch).

## Test plan

- [ ] Theme app extension without `.theme-check.yml` — `shopify app dev` and `shopify app deploy` produce identical theme-check output to before (uses bundled `theme-app-extension` defaults).
- [ ] Theme app extension with `.theme-check.yml` extending `theme-check:theme-app-extension` and adding `ignore: ['src/**']` — `shopify app dev` skips `src/**` offenses.
- [ ] Same `.theme-check.yml` with a custom check severity override — `shopify app dev` reflects the override.

### DIFF
--- a/.changeset/theme-check-respect-user-config.md
+++ b/.changeset/theme-check-respect-user-config.md
@@ -1,0 +1,5 @@
+---
+'@shopify/app': patch
+---
+
+`runThemeCheck` now respects a user's `.theme-check.yml` in the theme app extension root. When present, theme-check-node auto-discovery is used; otherwise the bundled `theme-check:theme-app-extension` config is applied as before. This lets extensions ignore source directories (e.g. a `src/` folder containing uncompiled Liquid templates) without forking the CLI.
--- a/packages/app/src/cli/services/build/theme-check.ts
+++ b/packages/app/src/cli/services/build/theme-check.ts
@@ -1,4 +1,5 @@
-import {readFileSync} from '@shopify/cli-kit/node/fs'
+import {fileExists, readFileSync} from '@shopify/cli-kit/node/fs'
+import {joinPath} from '@shopify/cli-kit/node/path'
 import {itemToString} from '@shopify/cli-kit/node/output'
 import {TokenItem} from '@shopify/cli-kit/node/ui'
 import {Severity, type Offense, check, path as pathUtils} from '@shopify/theme-check-node'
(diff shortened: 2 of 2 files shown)


erikmay:
still active

graygilmore (commented):
Thanks, Erik! This makes sense to me. Would you mind rebasing with `main`? Looks like we're seeing some CI failures at the current HEAD.

erikmay:
Thanks @graygilmore! Rebased onto the latest `main` (clean — no conflicts, the diff is unchanged). CI should be green now. Let me know if anything else is needed for merge. 🙏

graygilmore (approved):

gonzaloriestra (approved):

graygilmore:
Thanks @erikmay!

PR #7408, opened 2026-04-27
Normalize paths in bundle-ui-step same-path guard

isaacroldan:
The same-path guard at bundle-ui-step.ts compared `localOutputDir` and `bundleOutputDir` as raw strings. In practice these two values are computed along different code paths (`dirname(buildUIExtension())` vs `dirname(extension.outputPath)` or `joinPath(..., bundleFolder)`) and can resolve to the same filesystem directory while differing as strings — e.g. when one path has a `.` segment, a trailing slash, or otherwise non-canonical shape. When that happens, the guard slips through and fs-extra rejects the copy with "Source and destination must not be the same".

Normalize both sides via `resolvePath` before comparing so the guard catches any string variant that maps to the same directory.

Fixes #0000 

### Checklist


packages/app/src/cli/services/build/steps/bundle-ui-step.ts
robin-drexler, line 1: does this merit a test?

vividviolet (commented):
`shopify app build` ❌ still not working

Image

PR #7422, opened 2026-04-29
Add notification-driven kill switch for auto-upgrade

alfonso-noriega:

We need a remote kill switch for CLI auto-upgrade so we can stop a bad release from being silently rolled out further while a fix is in flight. Today, once a user has opted into auto-upgrade, every postrun will attempt to upgrade to the latest cached version with no way for us to halt that flow short of unpublishing.

This PR reuses the existing notifications system as the transport for that kill switch: an `error`-type notification on a new `autoupgrade` surface silently disables the auto-upgrade for matching CLI versions / date windows.

- Adds `hasBlockingAutoUpgradeNotification()` in `packages/cli-kit/src/public/node/upgrade.ts`. It freshly fetches the notifications feed (no cache) and returns `true` iff there is at least one notification with `surface === 'autoupgrade'`, `type === 'error'`, and matching `min/maxVersion` + `min/maxDate`. Lives in the upgrade module rather than the notifications module because the policy (which surface, which type counts as blocking) is auto-upgrade specific.
- Wires the helper into `autoUpgradeIfNeeded`'s `performAutoUpgrade` path. The check runs **after** every existing gate — cached newer version, opt-in, not CI, not pre-release, daily rate limit / `SHOPIFY_CLI_FORCE_AUTO_UPGRADE` — so:
  - the feed is fetched at most once per day per user, on the day an upgrade would actually happen,
  - users that have auto-upgrade off, are on CI, on a pre-release, or already ran today incur zero overhead.
- When a blocking notification is found, the upgrade is silently skipped (no `outputWarn` / `outputInfo`) and `env_auto_upgrade_skipped_reason: 'blocked_by_notification'` is recorded for analytics.
- Fail-open: any error fetching or parsing the feed returns `false`, with no logging, so a broken notifications endpoint never prevents upgrades.
- Tests cover the new helper (matching surface/type/version/date, fail-open path) and the postrun branch (silent skip + metadata, and that we don't query notifications when there's nothing to upgrade to).

### How to test your changes?

Locally simulate a kill-switch notification and verify that auto-upgrade is silently skipped:

1. Host a notifications JSON containing an entry like:
   ```json
   {
     "id": "block-autoupgrade",
     "message": "Auto-upgrade temporarily disabled",
     "type": "error",
     "frequency": "always",
     "ownerChannel": "#cli",
     "surface": "autoupgrade"
   }
   ```
2. Point the CLI at it: `export SHOPIFY_CLI_NOTIFICATIONS_URL=...`.
3. Force auto-upgrade to attempt: `export SHOPIFY_CLI_FORCE_AUTO_UPGRADE=1`, with auto-upgrade enabled, on a version older than the latest published `@shopify/cli`.
4. Run any non-`upgrade` / non-`notifications` Shopify CLI command.

Expected: no upgrade is performed, no warning/info output is shown, and analytics show `env_auto_upgrade_skipped_reason=blocked_by_notification`. Removing the notification (or changing `type` to `warning`) should let the upgrade proceed as before.

Unit tests:

```
pnpm -F @shopify/cli-kit test src/public/node/upgrade.test.ts src/public/node/hooks/postrun.test.ts
```

### Post-release steps

None.

### Checklist

- [x] I've considered possible cross-platform impacts (Mac, Linux, Windows) — pure logic on top of existing notifications fetch.
- [x] I've considered possible [documentation](https://shopify.dev) changes — no public-facing docs change; the kill-switch contract is operational.
- [x] I've considered analytics changes to measure impact — added `env_auto_upgrade_skipped_reason: 'blocked_by_notification'`.
- [x] This change is internal-only (operational kill switch); no user-facing behavior change and no changeset required.

### DIFF
--- a/packages/cli-kit/src/public/node/hooks/postrun.test.ts
+++ b/packages/cli-kit/src/public/node/hooks/postrun.test.ts
@@ -1,6 +1,11 @@
 import {autoUpgradeIfNeeded} from './postrun.js'
 import {mockAndCaptureOutput} from '../testing/output.js'
-import {getOutputUpdateCLIReminder, runCLIUpgrade, versionToAutoUpgrade} from '../upgrade.js'
+import {
+  getOutputUpdateCLIReminder,
+  hasBlockingAutoUpgradeNotification,
+  runCLIUpgrade,
+  versionToAutoUpgrade,
+} from '../upgrade.js'
 import {isMajorVersionChange} from '../version.js'
 import {inferPackageManagerForGlobalCLI} from '../is-global.js'
 import {addPublicMetadata} from '../metadata.js'
@@ -13,6 +18,7 @@ vi.mock('../upgrade.js', async (importOriginal) => {
     runCLIUpgrade: vi.fn(),
     getOutputUpdateCLIReminder: vi.fn(),
     versionToAutoUpgrade: vi.fn(),
+    hasBlockingAutoUpgradeNotification: vi.fn().mockResolvedValue(false),
   }
 })
 
@@ -55,6 +61,7 @@ vi.mock('../../../private/node/conf-store.js', async (importOriginal) => {
 afterEach(() => {
   mockAndCaptureOutput().clear()
   vi.mocked(addPublicMetadata).mockClear()
+  vi.mocked(hasBlockingAutoUpgradeNotification).mockResolvedValue(false)
 })
 
 describe('autoUpgradeIfNeeded', () => {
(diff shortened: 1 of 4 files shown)


alfonso-noriega:
* **#7422** <a href="https://app.graphite.com/github/pr/Shopify/cli/7422?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Shopify/cli/7422?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

isaacroldan (approved):
Nice!

PR #7430, opened 2026-04-30
[Security] Fix command injection in treeKill utility

gonzaloriestra:

`treeKill` in `@shopify/cli-kit` had a critical command injection vulnerability on Windows. The PID validation used `Number.isNaN(pid)` against a string, which always returns `false` for non-`NaN` string values, so malicious inputs like `"1234; calc.exe"` slipped through. The Windows branch then forwarded the value to `child_process.exec` via string concatenation, which spawns a shell and interprets the injected command.

- Replaces the broken `Number.isNaN` PID check with a strict regex (`/^\d+$/`) so only digit-only strings are accepted.
- Switches the Windows `taskkill` invocation from `child_process.exec` (shell-based) to `child_process.spawn` with arguments passed as an array, eliminating shell interpolation.
- Adds `packages/cli-kit/src/public/node/tree-kill.test.ts` with regression coverage for malicious and valid PID inputs.
- Adds short security-focused comments at the validation and spawn sites.

### How to test your changes?

`treeKill` is invoked when you quit `shopify app dev` (it tears down the dev process tree). Build the CLI, run a dev session against any sample app, and confirm a graceful shutdown:

```bash
pnpm install
pnpm shopify app dev --path <path-to-app>
# Inside the dev session, press 'q' (or Ctrl+C) and confirm the CLI exits cleanly.
# Then confirm no orphans were left behind (matches against full argv):
pgrep -fl 'app dev --path .*<path-to-app>' || echo "no orphan dev processes"
```

Then sanity-check the new PID validation directly. With a malicious PID it should refuse to spawn anything and surface the validation error. Run from the repo root against the built bundle:

```bash
pnpm nx build cli-kit
node -e "import('./packages/cli-kit/dist/public/node/tree-kill.js').then(({treeKill}) => treeKill('1234; calc.exe', 'SIGTERM', true, (err) => console.log(err?.message)))"
# Expect: pid must be a number
```

On Windows specifically, the `taskkill` invocation now uses `spawn` with array arguments instead of `exec` with string concatenation. Reviewers on Windows can confirm by running `shopify app dev` and quitting — the child tree should still terminate.

### Checklist

### DIFF
--- a/packages/cli-kit/src/public/node/tree-kill.test.ts
+++ b/packages/cli-kit/src/public/node/tree-kill.test.ts
@@ -1,0 +1,58 @@
+/* eslint-disable no-restricted-imports */
+import {treeKill} from './tree-kill.js'
+import {vi, describe, test, expect, afterEach} from 'vitest'
+import {spawn} from 'child_process'
+
+vi.mock('child_process', async () => {
+  const actual: any = await vi.importActual('child_process')
+  return {
+    ...actual,
+    spawn: vi.fn(),
+  }
+})
+
+describe('treeKill', () => {
+  afterEach(() => {
+    vi.unstubAllGlobals()
+  })
+
+  test('calls the callback with an error if the PID is not a number (string with digits)', async () => {
+    const maliciousPid = '1234; calc.exe'
+
+    await new Promise<void>((resolve) => {
+      // Correct signature for treeKill is (pid, signal, killRoot, callback)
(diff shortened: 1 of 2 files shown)


isaacroldan (approved):
Code looks good, but maybe we should test in windows?

gonzaloriestra:
@isaacroldan checked that it works as expected in Windows 👍

isaacroldan:
Noice! :white_check_mark:

PR #7434, opened 2026-04-30
Skip 400ms search throttle for in-memory autocomplete filtering

GeorgiZhelev:

Filtering inside `renderAutocompletePrompt` (e.g. the theme selector that appears for `shopify theme push`) feels noticeably laggy: there's a ~400ms gap between typing and seeing matching results update.

The cause is the throttle on the search callback in `AutocompletePrompt` (`packages/cli-kit/src/private/node/ui/components/AutocompletePrompt.tsx`). The 400ms window makes sense for callers that supply a remote/paginated `search` (we don't want to hammer an API on every keystroke), but it also gates the default in-memory filter that `renderAutocompletePrompt` injects when the caller doesn't pass one — and that filter is fully synchronous (`Promise.resolve(choices.filter(...))`), so there's no I/O to throttle.

- Adds an optional `searchDebounceMs` prop to `AutocompletePrompt`, default `400` (backward-compatible).
- `renderAutocompletePrompt` sets `searchDebounceMs: 0` only when it's injecting its own default in-memory filter; if the caller supplied their own `search`, the default 400ms throttle still applies.
- Adds a test asserting that `searchDebounceMs={0}` invokes the search callback on every keystroke (vs. coalescing to leading + trailing edges).

Net effect: prompts that filter against in-memory choices feel instant; remote/paginated consumers see no behavior change unless they opt in.

### How to test your changes?

1. Build the CLI: `pnpm nx build cli`
2. From any theme directory, run a command that opens the theme selector — e.g. `node packages/cli/bin/dev.js theme push`
3. Type into the search box; results should update on each keystroke instead of in ~400ms bursts.

Unit tests:
```
pnpm -F @shopify/cli-kit vitest run src/private/node/ui/components/AutocompletePrompt.test.tsx
```

### Post-release steps

None.

### Checklist

- [x] I've considered possible cross-platform impacts (Mac, Linux, Windows) — pure JS timing change, platform-agnostic.
- [x] I've considered possible [documentation](https://shopify.dev) changes — JSDoc on the new prop covers it.
- [x] I've considered analytics changes to measure impact — none required.
- [x] The change is user-facing — I've added a `patch` changeset for `@shopify/cli-kit`.

### DIFF
--- a/.changeset/autocomplete-instant-local-search.md
+++ b/.changeset/autocomplete-instant-local-search.md
@@ -1,0 +1,5 @@
+---
+'@shopify/cli-kit': patch
+---
+
+Make the autocomplete prompt feel instant when filtering against in-memory choices. The 400ms throttle on the search callback was designed for remote/paginated backends, but it also gated the default in-memory filter used by callers that don't supply their own `search` (e.g. the theme selector), producing a noticeable lag while typing. The prompt now exposes a `searchDebounceMs` prop, and `renderAutocompletePrompt` sets it to `0` when it injects its own synchronous filter. Custom remote-search consumers keep the existing 400ms throttle unless they opt out.
--- a/.changeset/autocomplete-instant-local-search.md
+++ b/.changeset/autocomplete-instant-local-search.md
@@ -2,4 +2,4 @@
 '@shopify/cli-kit': patch
 ---
 
-Make the autocomplete prompt feel instant when filtering against in-memory choices. The 400ms throttle on the search callback was designed for remote/paginated backends, but it also gated the default in-memory filter used by callers that don't supply their own `search` (e.g. the theme selector), producing a noticeable lag while typing. The prompt now exposes a `searchDebounceMs` prop, and `renderAutocompletePrompt` sets it to `0` when it injects its own synchronous filter. Custom remote-search consumers keep the existing 400ms throttle unless they opt out.
+Make the autocomplete prompt feel instant when filtering against in-memory choices.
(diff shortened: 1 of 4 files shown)


gonzaloriestra:
Thanks for your contribution! Could you please rebase? The failing check should pass after that

GeorgiZhelev:
Hey, done!

.changeset/autocomplete-instant-local-search.md
```diff
@@ -0,0 +1,5 @@
+---
+'@shopify/cli-kit': patch
+---
+
+Make the autocomplete prompt feel instant when filtering against in-memory choices. The 400ms throttle on the search callback was designed for remote/paginated backends, but it also gated the default in-memory filter used by callers that don't supply their own `search` (e.g. the theme selector), producing a noticeable lag while typing. The prompt now exposes a `searchDebounceMs` prop, and `renderAutocompletePrompt` sets it to `0` when it injects its own synchronous filter. Custom remote-search consumers keep the existing 400ms throttle unless they opt out.
```
gonzaloriestra: Would you mind simplifying here a bit?

```suggestion
Make the autocomplete prompt feel instant when filtering against in-memory choices.

gonzaloriestra (approved):
Nice work, thanks!! It really feels much faster ⚡

.changeset/autocomplete-instant-local-search.md
gonzaloriestra: I've applied this and I'll merge the PR, I hope you don't mind

Issue #2877, opened 2023-09-21
[Bug]: Issue running deploy with Functions


christiecc8:
### Please confirm that you have:

- [X] Searched [existing issues](.) to see if your issue is a duplicate. (If you’ve found a duplicate issue, feel free to add additional information in a comment on it.)
- [X] Reproduced the issue in [the latest CLI version](https://www.npmjs.com/package/@shopify/cli).

### In which of these areas are you experiencing a problem?

Extension

### Expected behavior

Hi, I have a project with one theme app extension and two different functions, one for Product discounts and the other for Cart checkout validation. I expect to be able to deploy my app with all three extensions from my CI environment (Github actions) without errors. This is a snippet of the output when I deploy from localhost on macOS:

<details>
  <summary>Log</summary>

```
2023-09-21T21:32:50.139Z: Sending Identity Introspection request to URL: https://accounts.shopify.com/oauth/introspection
2023-09-21T21:32:50.139Z: Sending POST request to URL https://accounts.shopify.com/oauth/introspection
With request headers:
 - User-Agent: Shopify CLI; v=3.49.2
 - Keep-Alive: timeout=30
 - Sec-CH-UA-PLATFORM: darwin
 - Content-Type: application/json

2023-09-21T21:32:50.286Z: Request to https://accounts.shopify.com/oauth/introspection completed in 139 ms
With response headers:
 - cache-control: no-cache, no-store, private, must-revalidate, max-age=0
 - content-type: application/json; charset=utf-8
 - etag: W/"cff832d64a045869304ce6d9f3c20690"
 - x-request-id: 564d6f38-800a-40c5-a8d9-7d4a4ecbce25
    
2023-09-21T21:32:50.288Z: The identity token is valid: true
2023-09-21T21:32:50.288Z: 
The validation of the token for application/identity completed with the following results:
- It's expired: false
- It's invalid in identity: false
  
2023-09-21T21:32:50.289Z: Verifying that the user has a Partner organization
2023-09-21T21:32:50.294Z: Sending "Partners" GraphQL request:
  {
    organizations(first: 1) {
      nodes {
        id
      }
    }
  }

With request headers:
 - User-Agent: Shopify CLI; v=3.49.2
 - Keep-Alive: timeout=30
 - Sec-CH-UA-PLATFORM: darwin
 - Content-Type: application/json

2023-09-21T21:32:50.506Z: Request to https://partners.shopify.com/api/cli/graphql completed in 198 ms
With response headers:
[...]

PR #7438, opened 2026-04-30
Preload trampoline binaries to avoid race conditions when lazily downloading during parallel builds

dnagoda:

`shopify app build` fails intermittently on CI with `ETXTBSY` ("text file busy") when building apps with multiple Shopify Function extensions:

```
spawn ETXTBSY
Command failed with ETXTBSY: .../node_modules/@shopify/cli/bin/shopify-function-trampoline-1.0.2 -i <wasm> -o <wasm>
```

The trampoline binaries (`shopify-function-trampoline-1.0.2`, `shopify-function-trampoline-2.0.1`) are not shipped with the npm package — they are lazily downloaded from GitHub on first use by `downloadBinary()`. When `renderConcurrent` runs parallel extension builds, multiple concurrent tasks race to download the same binary. One task opens the file for writing (`moveFile` with `{overwrite: true}`) while another tries to execute it, and the Linux kernel returns `ETXTBSY`.

The in-process dedup map (`downloadsInProgress`) prevents redundant downloads within a single event loop tick, but the race window between the download completing and the file being executed by a concurrent build is still open.

This is the **same class of bug** that was previously fixed for Javy binaries in https://github.com/Shopify/cli/issues/2877 — which introduced `installJavy()` to pre-download Javy + plugin binaries before parallel builds start. The trampoline was simply missed because it was added later.

Adds `installTrampolines(app)` — a new pre-download function that mirrors the existing `installJavy(app)` pattern. It eagerly downloads both trampoline versions (v1 and v2) before the parallel build fan-out, so that when individual `runTrampoline()` calls run concurrently, the binaries are already on disk and `downloadBinary()` is a no-op.

**Key design choice:** Unlike Javy (where the version depends on each extension's `@shopify/shopify_function` package version), the trampoline version is determined _post-build_ from WASM imports (`shopify_function_v1` vs `shopify_function_v2`). Since there are only two possible versions and the binaries are small, we simply pre-download both for any app with function extensions.

**Changes:**

- **`packages/app/src/cli/services/function/build.ts`** — Added `installTrampolines(app)`: skips if no function extensions, otherwise downloads both V1 and V2 trampoline binaries in parallel
- **`packages/app/src/cli/services/build.ts`** — Calls `installTrampolines` alongside `installJavy` (via `Promise.all`) before `renderConcurrent`
- **`packages/app/src/cli/services/deploy/bundle.ts`** — Same pre-download before parallel deploy builds
- **`packages/app/src/cli/services/dev/processes/draftable-extension.ts`** — Same pre-download before dev draft updates
- **`packages/app/src/cli/services/function/build.test.ts`** — Added 2 tests: verifies both versions are downloaded when function extensions exist, verifies no downloads when they don't

### How to test your changes?

1. **Unit tests:** `pnpm vitest run packages/app/src/cli/services/function/build.test.ts` (28 tests pass including 2 new)
2. **Manual repro (requires Linux CI or Docker):**
    - Create an app with 2+ function extensions
    - Clear any cached trampoline binaries from `node_modules/@shopify/cli/bin/`
    - Run `shopify app build` — should succeed consistently without `ETXTBSY`
        - This is hard to reproduce manually and is flakey on CI, but an example of the failure can be found here: https://buildkite.com/shopify/checkout-blocks-tests/builds/5475#019de03e-9d05-470f-9864-765465ea71a7
3. **Regression check:** existing Javy pre-download behavior unchanged — `installJavy` and `installTrampolines` run in parallel via `Promise.all`

### Checklist

### DIFF
--- a/packages/app/src/cli/services/dev/processes/draftable-extension.ts
+++ b/packages/app/src/cli/services/dev/processes/draftable-extension.ts
@@ -4,7 +4,6 @@ import {ExtensionInstance} from '../../../models/extensions/extension-instance.j
 import {AppInterface} from '../../../models/app/app.js'
 import {PartnersAppForIdentifierMatching, ensureDeploymentIdsPresence} from '../../context/identifiers.js'
 import {getAppIdentifiers} from '../../../models/app/identifiers.js'
-import {installJavy} from '../../function/build.js'
 import {DeveloperPlatformClient} from '../../../utilities/developer-platform-client.js'
 import {AppEvent, AppEventWatcher, EventType} from '../app-events/app-event-watcher.js'
 import {AbortError} from '@shopify/cli-kit/node/error'
@@ -28,10 +27,6 @@ export const pushUpdatesForDraftableExtensions: DevProcessFunction<DraftableExte
   {stderr, stdout},
   {developerPlatformClient, apiKey, remoteExtensionIds: remoteExtensions, localApp: app, appWatcher},
 ) => {
-  // Force the download of the javy binary in advance to avoid later problems,
-  // as it might be done multiple times in parallel. https://github.com/Shopify/cli/issues/2877
-  await installJavy(app)
-
   const draftableExtensions = app.draftableExtensions.map((ext) => ext.handle)
 
   const handleAppEvent = async (event: AppEvent) => {
--- a/packages/app/src/cli/services/function/build.ts
+++ b/packages/app/src/cli/services/function/build.ts
@@ -379,6 +379,22 @@ export async function installJavy(app: AppInterface) {
   await Promise.all(downloadPromises)
 }
 
+export async function installTrampolines(app: AppInterface) {
+  const hasFunctionExtensions = app.allExtensions.some((ext) => ext.features.includes('function'))
+  if (!hasFunctionExtensions) {
+    return
+  }
+
+  await Promise.all([
+    downloadBinary(trampolineBinary(V1_TRAMPOLINE_VERSION)).catch((err) => {
+      outputDebug(`Failed to preload trampoline v${V1_TRAMPOLINE_VERSION}: ${err.message}`)
+    }),
+    downloadBinary(trampolineBinary(V2_TRAMPOLINE_VERSION)).catch((err) => {
+      outputDebug(`Failed to preload trampoline v${V2_TRAMPOLINE_VERSION}: ${err.message}`)
+    }),
+  ])
+}
+
 interface JavyBuilder {
   bundle(fun: ExtensionInstance<FunctionConfigType>, options: JSFunctionBuildOptions): Promise<BuildResult>
   compile(
--- a/.changeset/preload-trampoline-binaries.md
+++ b/.changeset/preload-trampoline-binaries.md
@@ -1,0 +1,5 @@
+---
+'@shopify/app': patch
+---
+
+Fix `shopify app build` intermittently failing with `ETXTBSY` ("text file busy") when building apps with multiple function extensions on a fresh environment
--- a/packages/app/src/cli/services/build.ts
+++ b/packages/app/src/cli/services/build.ts
@@ -1,6 +1,6 @@
 import buildWeb from './web.js'
 import {installAppDependencies} from './dependencies.js'
-import {installJavy} from './function/build.js'
+import {installJavy, installTrampolines} from './function/build.js'
 import {AppInterface, Web} from '../models/app/app.js'
 import {Project} from '../models/project/project.js'
 import {renderConcurrent, renderSuccess} from '@shopify/cli-kit/node/ui'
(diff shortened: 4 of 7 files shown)


dnagoda:
* **#7438** <a href="https://app.graphite.com/github/pr/Shopify/cli/7438?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Shopify/cli/7438?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/app/src/cli/services/dev/processes/draftable-extension.ts
```diff
@@ -28,9 +28,11 @@ export const pushUpdatesForDraftableExtensions: DevProcessFunction<DraftableExte
...
   {developerPlatformClient, apiKey, remoteExtensionIds: remoteExtensions, localApp: app, appWatcher},
 ) => {
-  // Force the download of the javy binary in advance to avoid later problems,
-  // as it might be done multiple times in parallel. https://github.com/Shopify/cli/issues/2877
-  await installJavy(app)
+  // Force the download of binaries in advance to avoid later problems,
+  // as it might be done multiple times in parallel.
+  // javy -> https://github.com/Shopify/cli/issues/2877
+  // trampoline -> ETXTBSY race conditions during parallel Rust builds
+  await Promise.all([installJavy(app), installTrampolines(app)])
```
dmerand: Have you tested `dev` with this change? I wonder if the dev-side preload is attached to the wrong lifecycle point, meaning this likely still misses `app dev` startup in some cases.

Right now the new preload runs in `pushUpdatesForDraftableExtensions()` in `packages/app/src/cli/services/dev/processes/draftable-extension.ts`, but the actual initial concurrent extension builds happen in `AppEventWatcher.start()` via `buildExtensions(...)`.

Two consequences:
- `setupDevProcesses()` uses `setupDevSessionProcess(...)` instead of `setupDraftableExtensionsProcess(...)` when `developerPlatformClient.supportsDevSessions` is true, so that path appears to get **no** trampoline preload at all.
- More generally, the watcher startup is the place that owns the initial build fan-out, so putting the preload in a side process feels racy / easy to bypass.

Would it be safer to move this into the shared watcher startup path (e.g. before `AppEventWatcher.start()` begins `buildExtensions(...)`), so all `app dev` modes get the same protection?

packages/app/src/cli/services/function/build.ts
```diff
@@ -379,6 +379,18 @@ export async function installJavy(app: AppInterface) {
   await Promise.all(downloadPromises)
 }
 
+export async function installTrampolines(app: AppInterface) {
```
dmerand, line 382: One compatibility concern here that my 'bot found: `installTrampolines()` now eagerly downloads both trampoline versions, but `trampolineBinary(version)` only marks Windows-on-ARM support for versions `>= 2.0.1`.

That means on `win32/arm64`, a function app that only needs the v2 trampoline may have worked before (because `runTrampoline()` lazily selected/downloaded just v2), but could now fail up front trying to predownload the unused v1 binary.

Would it be safer to skip pre-downloading versions that are unsupported on the current platform/arch, or otherwise avoid failing the whole command on an unused trampoline version?

packages/app/src/cli/services/dev/processes/draftable-extension.ts
dnagoda: I'm very low context on CLI architecture, but I can definitely look into how we could do this at a higher level in the flow. Would we want to move the javy preload at the same time?

packages/app/src/cli/services/function/build.ts
dnagoda: Yes, that makes sense. Let me see how we check for the needed version and look into adding that logic here.

packages/app/src/cli/services/dev/processes/draftable-extension.ts
dmerand: I'm also low-context on the functions side of things. Perhaps it works for Javy at this layer, it should work for Trampoline. Hopefully somebody with more Functions experience can weigh in.

packages/app/src/cli/services/function/build.ts
dnagoda: We can't determine which trampolines we need during the preload because we don't have WASMs to introspect, but we can protect from download errors and log. This will protect us from having issues during the preload. Thanks for catching this. :+1:

packages/app/src/cli/services/dev/processes/draftable-extension.ts
dnagoda: I think it makes sense to move both preloads out of the watcher. There's probably also an argument that we don't need the preload for dev contexts as these are likely to run in long lived environments. The issue with the build and bundle flows is that they would often be run in CI on fresh env where the resource has to be downloaded fresh.

dnagoda: Moved to SetupDevProcesses.

jeffcharles:
I've opened #7443 as an alternative fix. The underlying problem IMO is `downloadBinary` returning too early. That is, when it sees the file on the file system, not when the file is ready to be executed. What I prefer with my fix is it'll cover all Functions binaries, we don't have to download more than we need to, and we don't have to swallow errors during the download process.

If we do opt to go with this PR, I would ask we don't swallow all errors that happen with downloading but much more selective. The error messages would be misleading if someone tries building a Function for the first time and GitHub is having an outage or if they're using the v1 trampoline on Windows for ARM.

jeffcharles:
Thinking about this a little more, if the Shopify CLI, or the Functions binaries, were to drop support for Windows 10, we might be able to use Prism on Windows 11 to run the Intel binaries on ARM machines. I haven't tested it though as I don't have access to a Windows 11 machine.

dnagoda:
@jeffcharles Thanks for taking a look. Yes, your approach is much better. I completely missed the `downloadsInProgress` behavior. Closing this PR is favor of your approach.

PR #7443, opened 2026-05-01
Fix Spawn ETXTBSY bug

jeffcharles:

#7438 mentioned seeing the following in Linux CI environments:

```
spawn ETXTBSY
Command failed with ETXTBSY: .../node_modules/@shopify/cli/bin/shopify-function-trampoline-1.0.2 -i <wasm> -o <wasm>
```

I think the source of the bug is that `moveFile` is making a directory entry visible to `fs` but with a kernel lock so we need to account for that in `downloadBinary`.

I've updated the `isDownloaded` check to ensure the download process for that file isn't still in progress before we return that the download process is complete. If the file exists but the download process is still in progress, we immediately get the promise for the download in progress and await on it.

When `moveFile` is resolved in the promise performing the download, the kernel lock that was triggering the `ETXTBSY` on spawn should be released, and then that will resolve the `downloadPromise` for anyone awaiting on it. So that _should_ clear up any `ETXTBSY` errors.

I've also added a test case. 

### How to test your changes?

1. Clear out any cached binaries in `node_modules/@shopify/cli/bin/`.
1. `pnpm shopify app build --path <path>` where `<path>` is a path to a project with multiple Function extensions using one of the trampolines.

### Checklist

### DIFF
--- a/.changeset/common-jokes-hide.md
+++ b/.changeset/common-jokes-hide.md
@@ -1,0 +1,5 @@
+---
+'@shopify/app': patch
+---
+
+Fix Spawn ETXTBSY bug with multiple Functions using trampoline binaries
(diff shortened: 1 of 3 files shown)


jeffcharles:
@isaacroldan any chance you could take a look at this fix given you reviewed #6395?

isaacroldan (approved):

PR #7448, opened 2026-05-03
[Security] Add command safety check to execCommand

gonzaloriestra:
Related to https://github.com/Shopify/develop-app-inner-loop/issues/120

### What was improved
Added a security check to `execCommand` to prevent execution of unsecure binaries in the current directory.

### How to test your changes?
Run the tests for the `@shopify/cli-kit` package:
`pnpm --filter @shopify/cli-kit vitest run src/public/node/system.test.ts`

The new test case "raises an error if the command to run is found in the current directory" should pass.

---
*PR created automatically by Jules for task [17378133700905160634](https://jules.google.com/task/17378133700905160634) started by @gonzaloriestra*

### DIFF
--- a/packages/cli-kit/src/public/node/system.ts
+++ b/packages/cli-kit/src/public/node/system.ts
@@ -206,6 +206,10 @@ export async function execCommand(command: string, options?: ExecOptions): Promi
     env.FORCE_COLOR = '1'
   }
   const executionCwd = options?.cwd ?? cwd()
+  const [cmd] = parseCommand(command)
+  if (cmd) {
+    checkCommandSafety(cmd, {cwd: executionCwd})
+  }
   try {
     await execaCommand(command, {
       env,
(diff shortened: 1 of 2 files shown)


packages/cli-kit/src/public/node/system.ts
```diff
@@ -206,6 +206,10 @@ export async function execCommand(command: string, options?: ExecOptions): Promi
     env.FORCE_COLOR = '1'
   }
   const executionCwd = options?.cwd ?? cwd()
+  const [cmd] = parseCommand(command)
```
dmerand: 🔒 **Security**: The new guard in `execCommand()` improves coverage, but it currently derives the executable name with the file-local `parseCommand()` helper while execution still goes through `execaCommand()`. Those parsers do not interpret command strings identically: `parseCommand()` only handles quoting, while Execa also accepts forms like backslash-escaped spaces. That mismatch means the preflight can approve one token and `execaCommand()` can still launch a different binary, which undermines the current-directory binary protection this change is trying to add.

**Suggestion:** Consider extracting the executable with the same parser Execa uses for command strings, or otherwise sharing one parsing path between validation and execution so escaped spaces and similar syntax cannot bypass `checkCommandSafety()`.

dmerand (commented):
Agent found a helpful note here.

dmerand (approved):

packages/cli-kit/src/public/node/system.ts
gonzaloriestra: Good catch! Fixed

isaacroldan (approved):

PR #7465, opened 2026-05-05
Check JSON response for knip to avoid false positives

gonzaloriestra:

The `Knip (unused code check)` job can pass silently in CI even when knip never actually analyzes the codebase. We hit this in [PR #7463](https://github.com/Shopify/cli/pull/7463): the job exited 0 with no output and was reported as green, even though running `pnpm knip` locally on the exact same SHA reproducibly fails with two unused dependencies. A false-green knip job lets real findings slip into `main`.

- Adds `bin/run-knip-ci.js`, a small Node script that runs knip with `--reporter json` and validates the output before trusting the exit code.
  - Knip with `--reporter json` always emits a structured object on success (`{"files":[],"issues":[]}`), so an empty / unparseable stdout is a reliable signal that knip didn't actually run — in which case the script fails the job with a `::error::` annotation.
  - When knip reports findings, the script re-runs it with the default reporter so the human-readable report still appears in the build log.
- Simplifies the `Run knip` step in `.github/workflows/tests-pr.yml` to a single line: `node bin/run-knip-ci.js`.

### How to test your changes?

I verified all three relevant scenarios locally against the real workspace:

1. **Clean run** (current `main`): `node bin/run-knip-ci.js` exits 0 with no output.
2. **Findings**: temporarily remove `@graphql-typed-document-node/core` from `knip.workspaces["packages/app"].ignoreDependencies` in `package.json` → script re-runs knip in default reporter mode, prints `Unused dependencies (1) ...`, and exits 1.
3. **Silent-pass bug** (the regression we're guarding against): temporarily replace the root `scripts.knip` with `node -e ""` so `pnpm knip` exits 0 with empty stdout → script prints `::error::knip exited N but produced no parseable JSON output. Failing CI to avoid a false green.` and exits 1.

You can reproduce 2 and 3 by editing `package.json` as described above, running `node bin/run-knip-ci.js`, and reverting the edit.

### Checklist

### DIFF
--- a/.github/workflows/tests-pr.yml
+++ b/.github/workflows/tests-pr.yml
@@ -87,7 +87,7 @@ jobs:
         with:
           node-version: ${{ env.DEFAULT_NODE_VERSION }}
       - name: Run knip
-        run: pnpm knip
+        run: node bin/run-knip-ci.js
 
   graphql-schema:
     if: ${{ github.event.pull_request.user.login != 'dependabot[bot]' }}
--- a/bin/run-knip-ci.js
+++ b/bin/run-knip-ci.js
@@ -1,0 +1,41 @@
+import {spawnSync} from 'node:child_process'
+
+// Run knip with the JSON reporter so a successful run still emits structured
+// output (e.g. {"files":[],"issues":[]}). An empty stdout therefore indicates
+// knip did not actually analyze the codebase — a silent-pass mode previously
+// observed — and must be treated as a failure to avoid false-green builds.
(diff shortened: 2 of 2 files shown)


alfonso-noriega (approved):

gonzaloriestra:
### Merge activity

* **May 5, 10:34 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/Shopify/cli/7465).
* **May 5, 10:34 AM UTC**: @gonzaloriestra added this pull request to the GitHub merge queue with [Graphite](https://app.graphite.com/github/pr/Shopify/cli/7465).

PR #7468, opened 2026-05-05
Remove unused 'Breaking change approval' job

alfonso-noriega:
### WHY are these changes introduced?

`tests-pr.yml` has two jobs for breaking changes:

- **`major-change-check`** ("Breaking change detection") — runs the script and `exit 1`s on any detected breaking change.
- **`major-change-approval`** ("Breaking change approval") — gated on the `breaking-change-approval` GitHub Environment, intended to require a `@shopify/dev_experience` member to approve before passing.

The second job's environment has **no protection rules** (`gh api repos/Shopify/cli/environments → "protection_rules": []`), so the "approval" job auto-passes in seconds with nobody actually approving anything. On PR #7466 the failure step ran at 09:46:20 and "Breaking change approval" turned green at 09:46:32 — 12s of "review", with zero human input. It's a placebo.

Worse, it's actively misleading: the script's failure message says "A member of @shopify/dev_experience must approve the breaking-change-approval environment", which sends people looking for a Settings → Environments → Required reviewers control that doesn't exist for them to use. PR #7469 (next in the stack) adds a real, working override — this PR removes the fake one so the two don't coexist.

#### `.github/workflows/tests-pr.yml`

- Removes the entire `major-change-approval` job.
- Updates the failure-step error message to drop the dead reference to the environment.

#### `workspace/src/major-change-check.js`

- Updates the sticky-comment report to remove the "this check will remain failed until a member of the team approves the workflow run" sentence — there is no workflow run to approve.

The `breaking-change-approval` GitHub Environment is **left alone in repo settings**. Other workflows might reference it, and deleting environments requires a separate API call. It's harmless to leave behind.

### How to test your changes?

```sh
cd workspace
node --test src/major-change-check.test.js
```

All tests still pass — this PR is workflow + comment-text only, no logic changes.

In CI, the "Breaking change approval" check name will simply stop appearing on new PR runs. The "Breaking change detection" check still runs (and after PR #7469 will be unblockable via a code-owner approval).

### Post-release steps

None.

### Checklist

- [x] I've considered possible cross-platform impacts (Mac, Linux, Windows) — N/A
- [x] I've considered possible documentation changes
- [x] I've considered analytics changes to measure impact — N/A
- [ ] The change is user-facing — N/A (internal CI tooling, no changeset needed)

PR #7467, opened 2026-05-05
Fix breaking-change check: compare against merge-base, scope to PR diff

alfonso-noriega:
### WHY are these changes introduced?

The "Breaking change detection" check in `tests-pr.yml` was reporting false-positive breaking changes whenever `main` had progressed past a PR's branch point. Concrete repro: PR #7466 only edits `.github/workflows/tests-pr.yml`, yet the check reports `packages/app/src/cli/models/extensions/specifications/type-generation.ts` removed a Zod field `value`. The field wasn't removed by the PR — it was *added on `main`* in commit `bc9ca2e` after the PR branched, and the script was diffing the PR (which is 7 commits behind) against `main`'s current tip.

This pattern silently inflates over time: any PR that sits unrebased while schema or manifest edits land on `main` accumulates phantom breaking-change reports.

### WHAT is this pull request doing?

Replaces "compare against `main` HEAD" with "compare against the merge-base of the PR head and the base branch", and scopes the schema/manifest scan to files actually present in the PR's diff.

#### `workspace/src/major-change-check.js`

- New `resolveContext()` reads `GITHUB_EVENT_PATH` and resolves:
  - `pull_request` / `pull_request_review` → `merge_base_commit.sha` from the GitHub compare API; `changedFiles` = the compare's `files[]`
  - `merge_group` → `merge_group.base_sha` (already the right baseline); `changedFiles` from the same compare API
  - Anything else (e.g. local invocation) → falls back to `main` HEAD with `changedFiles=null` so legacy behavior is preserved
- `checkSchemas` and `checkManifest` now accept `{changedFiles}`. When provided, they skip files outside the diff. The OCLIF manifest check short-circuits entirely if `packages/cli/oclif.manifest.json` isn't in the diff (a removed command always shows up there by definition).
- Compare-API failures degrade *open*, not closed: `changedFiles=null` so the scan widens rather than silently skipping potential removals. We never want to mask a real breaking change because of a transient API blip.

#### `workspace/src/utils/git.js`

- `cloneCLIRepository(tmpDir, {ref})` now takes a `ref` so callers can pin the baseline to the merge-base SHA (default still `main`).

#### `workspace/src/major-change-check.test.js`

- 4 new tests covering the new context resolution: `pull_request` uses merge-base, `merge_group` uses `base_sha`, compare-API failure widens the scan, missing event payload falls through to `main`.

### How to test your changes?

```sh
cd workspace
node --test src/major-change-check.test.js
```

[...]

PR #7469, opened 2026-05-05
Allow code-owner approval to override breaking-change check

alfonso-noriega:

The breaking-change check fails hard (`exit 1`) whenever the script detects a removed Zod field, command, flag, or major changeset. There's currently no way to override it short of force-merging or admin-bypassing branch protection — and after PR #7468 in this stack, the previous (broken) "approval" path is gone too.

In practice, breaking changes are sometimes intentional: removing a deprecated field at a major-version boundary, restructuring a schema after a migration, etc. The team needs a way to acknowledge "yes, we know this is breaking, ship it" that's auditable and lightweight.

The simplest, most natural mechanism is the one already in place for everything else: a code-owner approval on the PR.

Adds a code-owner-approval override to the breaking-change check. When a code owner of the changed files approves the PR, the check re-runs automatically (via a new `pull_request_review` trigger) and turns green.

#### `.github/workflows/breaking-change-check.yml` (new)

The breaking-change job moved into its own workflow file so we can trigger it on `pull_request_review` events without re-running the entire test suite (unit tests, e2e, type-diff, etc.) every time someone clicks Approve. Triggers:

- `pull_request` — same as before
- `pull_request_review: types: [submitted, dismissed, edited]` — re-runs when an approval is added or withdrawn so the check turns green/red automatically
- `merge_group` — same as before

Concurrency is keyed per PR (or per merge-queue head) so the latest review event always wins.

#### `.github/workflows/tests-pr.yml`

- Removes the now-relocated `major-change-check` job.

#### `workspace/src/major-change-check.js`

When breaking changes are detected, the script now:

1. Fetches the PR's reviews via `GET /repos/{owner}/{repo}/pulls/{N}/reviews`.
2. Computes the *latest review per author* (matches GitHub's "latest review wins" semantics — a later `CHANGES_REQUESTED` cancels an earlier `APPROVED`, exactly like `dismiss_stale_reviews`).
3. For each approver, checks `GET /repos/{owner}/{repo}/collaborators/{user}/permission`. If the permission is `admin`, `maintain`, or `write`, treat the approval as a code-owner override.
4. If override is granted: emit `has_breaking_changes=false`, log the approver, append an "Override: approved by @user" footer to the sticky comment, and exit cleanly.
5. Otherwise: behave exactly as before — sticky comment + `exit 1`.

The script also reads CODEOWNERS (`.github/CODEOWNERS`) and logs which teams own the changed paths for transparency in the run output. The override decision uses the repo-permission check (not direct team-membership lookup) because the default `GITHUB_TOKEN` can't list arbitrary org-team memberships across Shopify, but it can read repo-level collaborator permissions. Anyone with `write` on this repo got that access via a code-owning team grant — same security posture as branch protection's "require code-owner review", evaluated earlier so the check can flip without manual CI re-runs.

The override does **not** auto-approve when the GitHub API fails — every degradation path returns `approved: false`. We'd rather over-flag than silently grant.

#### `workspace/src/major-change-check.test.js`

10 new tests:

- `parseCodeowners` — comments and blank lines stripped, owners preserved
- `codeownersPatternToRegExp` — `*`, `**`, anchored, and unanchored patterns
- `ownersForFiles` — last-matching-rule-wins (matches CODEOWNERS semantics, including theme overriding the catch-all)
- `findCodeownerApproval`:
  - Approver with write access overrides (✓)
  - Latest review per author wins (a later `CHANGES_REQUESTED` cancels an earlier `APPROVED`, sticky-as-`dismiss_stale_reviews`)
  - No approving reviewer with write access → not approved
  - Missing PR number bails immediately
  - Reviews API failure bails (does NOT silently grant)

### Stickiness behaviour

Per the requirement: matches `dismiss_stale_reviews`. Concretely:

- If the same author submits a later `CHANGES_REQUESTED` or `DISMISSED`, the override is gone (we always look at *latest review per author*).
- After a `git push`, the PR's existing approving reviews remain (since `dismiss_stale_reviews: false` on `main`'s branch protection), so the override survives a force-push. If branch-protection ever flips `dismiss_stale_reviews` to true, GitHub will dismiss the reviews and our check will correctly go red again on the next event.

### How to test your changes?

```sh
cd workspace
node --test src/major-change-check.test.js
```

All 21 tests pass locally (11 pre-existing + 10 new). End-to-end testing requires a PR that introduces a real or fake breaking change; smoke-test plan:

1. Open a PR that removes a Zod field. The check should be red with the existing sticky-comment.
2. A code owner clicks Approve.
3. The workflow re-triggers (visible in Actions → Breaking change check). The check turns green and the sticky-comment gains the "Override: approved by @user" footer.
4. The reviewer requests changes. The check turns red again.

### Post-release steps

None — the workflow takes effect immediately on first run on `main`.

### Checklist

- [x] I've considered possible cross-platform impacts (Mac, Linux, Windows) — pure Node + workflow YAML, no platform calls
- [ ] I've considered analytics changes to measure impact — N/A (CI tooling)
- [ ] The change is user-facing — N/A (internal CI tooling, no changeset needed)

### DIFF
--- a/.github/workflows/breaking-change-check.yml
+++ b/.github/workflows/breaking-change-check.yml
@@ -1,0 +1,75 @@
+name: Breaking change check
+
+# Runs the breaking-change detector on PR opens, updates, AND on review
+# submissions. Re-running on review submissions lets the script flip from
+# red to green when a code-owner approves the PR — see the override logic
+# in workspace/src/major-change-check.js.
+#
+# Lives in a dedicated workflow so review events don't re-trigger the full
+# unit/e2e/type-diff suite.
+on:
+  pull_request:
+  pull_request_review:
+    types: [submitted, dismissed, edited]
+  merge_group:
+
+concurrency:
+  group: shopify-cli-breaking-change-${{ github.event.pull_request.number || github.event.merge_group.head_sha || github.run_id }}
+  cancel-in-progress: true
+
+env:
+  DEBUG: '1'
+  SHOPIFY_CLI_ENV: development
+  SHOPIFY_CONFIG: debug
+  PNPM_VERSION: '10.11.1'
+  BUNDLE_WITHOUT: 'test:development'
+  GH_TOKEN: ${{ secrets.SHOPIFY_GH_READ_CONTENT_TOKEN }}
+  GH_TOKEN_SHOP: ${{ secrets.SHOP_GH_READ_CONTENT_TOKEN }}
+  DEFAULT_NODE_VERSION: '24.1.0'
+
+jobs:
+  major-change-check:
+    # Skip fork PRs — fork tokens are read-only and the codeowner-override
+    # API calls would fail.
+    if: github.event.pull_request.head.repo.full_name == github.repository || github.event_name == 'merge_group'
+    name: 'Breaking change detection'
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
+    permissions:
+      contents: read
+      pull-requests: write
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          repository: ${{ github.event.pull_request.head.repo.full_name || github.event.repository.full_name }}
+          ref: ${{ github.event.pull_request.head.sha || github.event.merge_group.head_sha }}
+          fetch-depth: 1
+      - name: Setup deps
+        uses: ./.github/actions/setup-cli-deps
+        with:
+          node-version: ${{ env.DEFAULT_NODE_VERSION }}
+      - name: Build
+        run: pnpm nx run-many --all --skip-nx-cache --target=build --output-style=stream
+      - name: Check for breaking changes
+        id: check
+        env:
+          # The default GITHUB_TOKEN can read PR reviews and the repo's
+          # CODEOWNERS file, which is all the override needs.
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: node workspace/src/major-change-check.js
+      - uses: marocchino/sticky-pull-request-comment@fcf6fe9e4a0409cd9316a5011435be0f3327f1e1 # v2.3.1
+        if: steps.check.outputs.has_breaking_changes == 'true'
+        with:
+          header: Breaking-change-detection
+          message: ${{ steps.check.outputs.report }}
+          recreate: true
+      - uses: marocchino/sticky-pull-request-comment@fcf6fe9e4a0409cd9316a5011435be0f3327f1e1 # v2.3.1
+        if: steps.check.outputs.has_breaking_changes != 'true'
+        with:
+          header: Breaking-change-detection
+          delete: true
+      - name: Fail if breaking changes detected
+        if: steps.check.outputs.has_breaking_changes == 'true'
+        run: |
+          echo '::error::Breaking changes detected. See the sticky comment on the PR for details. A code-owner approval on this PR will turn this check green.'
+          exit 1
--- a/.github/workflows/breaking-change-check.yml
+++ b/.github/workflows/breaking-change-check.yml
@@ -10,7 +10,9 @@ name: Breaking change check
 on:
   pull_request:
   pull_request_review:
-    types: [submitted, dismissed, edited]
+    # Only state-changing actions are interesting: `edited` fires when a
+    # review's body text is mutated, which doesn't affect approval state.
+    types: [submitted, dismissed]
   merge_group:
 
 concurrency:
@@ -25,7 +27,7 @@ env:
   BUNDLE_WITHOUT: 'test:development'
   GH_TOKEN: ${{ secrets.SHOPIFY_GH_READ_CONTENT_TOKEN }}
   GH_TOKEN_SHOP: ${{ secrets.SHOP_GH_READ_CONTENT_TOKEN }}
-  DEFAULT_NODE_VERSION: '24.1.0'
+  DEFAULT_NODE_VERSION: '26.1.0'
 
 jobs:
   major-change-check:
--- a/.github/workflows/breaking-change-check.yml
+++ b/.github/workflows/breaking-change-check.yml
@@ -43,7 +43,12 @@ jobs:
         with:
           repository: ${{ github.event.pull_request.head.repo.full_name || github.event.repository.full_name }}
           ref: ${{ github.event.pull_request.head.sha || github.event.merge_group.head_sha }}
-          fetch-depth: 1
+          # Full history so `git merge-base origin/<base> HEAD` (used by
+          # major-change-check.js to scope the diff to PR-only changes) can
+          # reach the divergence point. With fetch-depth: 1 the merge-base
+          # call fails and the script falls back to scanning everything,
+          # surfacing every pre-existing major changeset as "new".
+          fetch-depth: 0
       - name: Setup deps
         uses: ./.github/actions/setup-cli-deps
         with:
--- a/.github/workflows/breaking-change-check.yml
+++ b/.github/workflows/breaking-change-check.yml
@@ -53,8 +53,14 @@ jobs:
         uses: ./.github/actions/setup-cli-deps
         with:
           node-version: ${{ env.DEFAULT_NODE_VERSION }}
-      - name: Build
-        run: pnpm nx run-many --all --skip-nx-cache --target=build --output-style=stream
+      # No build step: major-change-check.js operates on git-tracked
+      # source files (.ts, oclif.manifest.json, .changeset/*.md) and
+      # `git diff` output only — it never imports built artefacts. The
+      # script itself explicitly opts out of installing/building the
+      # baseline checkout for the same reason (see runMain in
+      # workspace/src/major-change-check.js). Skipping the build keeps
+      # approval-triggered re-runs fast (~30s instead of ~5–8min) so the
+      # check can flip green quickly when a code owner approves.
       - name: Check for breaking changes
         id: check
         env:
--- a/.github/workflows/breaking-change-check.yml
+++ b/.github/workflows/breaking-change-check.yml
@@ -68,8 +68,19 @@ jobs:
           header: Breaking-change-detection
           message: ${{ steps.check.outputs.report }}
           recreate: true
+      # Override granted: breaking changes were detected but a code-owner
+      # approved, so we keep the sticky comment up to record the override
+      # footer ("✅ Override: approved by @user") emitted by the script.
       - uses: marocchino/sticky-pull-request-comment@fcf6fe9e4a0409cd9316a5011435be0f3327f1e1 # v2.3.1
-        if: steps.check.outputs.has_breaking_changes != 'true'
+        if: steps.check.outputs.has_breaking_changes != 'true' && steps.check.outputs.report != ''
+        with:
+          header: Breaking-change-detection
+          message: ${{ steps.check.outputs.report }}
+          recreate: true
+      # No breaking changes at all: delete any stale sticky comment from a
+      # previous push.
+      - uses: marocchino/sticky-pull-request-comment@fcf6fe9e4a0409cd9316a5011435be0f3327f1e1 # v2.3.1
+        if: steps.check.outputs.has_breaking_changes != 'true' && steps.check.outputs.report == ''
         with:
           header: Breaking-change-detection
           delete: true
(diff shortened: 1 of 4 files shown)


alfonso-noriega:
* **#7469** <a href="https://app.graphite.com/github/pr/Shopify/cli/7469?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Shopify/cli/7469?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#7468** <a href="https://app.graphite.com/github/pr/Shopify/cli/7468?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#7467** <a href="https://app.graphite.com/github/pr/Shopify/cli/7467?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

isaacroldan (dismissed):

.github/workflows/breaking-change-check.yml
```diff
@@ -0,0 +1,80 @@
...
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: node workspace/src/major-change-check.js
+      - uses: marocchino/sticky-pull-request-comment@fcf6fe9e4a0409cd9316a5011435be0f3327f1e1 # v2.3.1
+        if: steps.check.outputs.has_breaking_changes == 'true'
+        with:
+          header: Breaking-change-detection
+          message: ${{ steps.check.outputs.report }}
+          recreate: true
+      - uses: marocchino/sticky-pull-request-comment@fcf6fe9e4a0409cd9316a5011435be0f3327f1e1 # v2.3.1
+        if: steps.check.outputs.has_breaking_changes != 'true'
```
isaacroldan: 🐛 **Bug**: When the override is granted, the script sets `has_breaking_changes=false` (line 889) and sets `report` to include the `✅ Override: approved by @user` footer (line 890). But the workflow at line 72 checks `has_breaking_changes != 'true'` and runs the sticky-comment action with `delete: true` — the carefully constructed override report is discarded. The PR description states the sticky comment should gain the override footer, but that won't happen.

**Suggestion:** Split the workflow logic to post the override report when present, and only delete when there's nothing to show:

```yaml
      - uses: marocchino/sticky-pull-request-comment@fcf6fe9e4a0409cd9316a5011435be0f3327f1e1
        if: steps.check.outputs.has_breaking_changes != 'true' && steps.check.outputs.report != ''
        with:
          header: Breaking-change-detection
          message: ${{ steps.check.outputs.report }}
          recreate: true
      - uses: marocchino/sticky-pull-request-comment@fcf6fe9e4a0409cd9316a5011435be0f3327f1e1
        if: steps.check.outputs.has_breaking_changes != 'true' && steps.check.outputs.report == ''
        with:
          header: Breaking-change-detection
          delete: true
```

Alternatively, if deleting the comment on override is the intended behavior, remove the `setOutput('report', ...)` call at line 890 and update the PR description to match.

workspace/src/major-change-check.js
```diff
@@ -89,6 +89,195 @@ async function defaultRunGit(args) {
...
+  if (!prNumber) return {approved: false, approver: null, reason: 'no PR number in event'}
+
+  const reviews = await fetchReviews(repo, prNumber)
+  if (!reviews) return {approved: false, approver: null, reason: 'reviews API failed'}
+
+  // Last review per author wins (matches GitHub's own "latest review" semantics).
+  const latestByUser = new Map()
+  for (const review of reviews) {
+    if (!review.user?.login) continue
+    latestByUser.set(review.user.login, review)
```
isaacroldan, line 288: 🐛 **Bug**: GitHub's PR reviews API returns `COMMENTED` entries for every review comment submission. GitHub's own branch protection semantics only consider `APPROVED`, `CHANGES_REQUESTED`, and `DISMISSED` when determining the latest review per author — `COMMENTED` is not an actionable review state. This code treats every review entry equally, so if a reviewer approves and then leaves a regular review comment, the `COMMENTED` entry overwrites their `APPROVED` entry in the `latestByUser` map. The approval is silently lost, and the override won't be granted even when it should be. This is a common workflow: approve a PR, then comment on a specific line.

**Suggestion:** Filter out non-actionable review states before building the latest-per-user map:

```suggestion
    const actionableStates = new Set(['APPROVED', 'CHANGES_REQUESTED', 'DISMISSED'])
    if (!review.user?.login || !actionableStates.has(review.state)) continue
    latestByUser.set(review.user.login, review)
```

workspace/src/major-change-check.js
```diff
@@ -680,14 +869,32 @@ async function runMain() {
...
       setOutput('has_breaking_changes', 'false')
+      return
+    }
+
+    // Breaking changes were detected. If a code-owner has already approved
+    // the PR, treat that as sign-off and turn the check green. The script
+    // re-runs on `pull_request_review` events (see breaking-change-check.yml)
+    // so a fresh approval will flip this check without manual CI re-runs.
+    const override = await findCodeownerApproval({
+      repo: context.repo,
```
isaacroldan, line 941: 🐛 **Bug**: `resolveContext()` (lines 63–85) returns `{baselineRef, changedFiles}` in every code path — it never sets `repo` or `prNumber`. At lines 883–884, `context.repo` and `context.prNumber` are both `undefined`. This means `findCodeownerApproval` always bails at line 226 with `'no PR number in event'`, and even if it didn't, the API URLs would resolve to `https://api.github.com/repos/undefined/undefined/...`. The unit tests all pass because they inject `repo` and `prNumber` directly — they never exercise the integration path through `runMain`. The entire code-owner approval override feature will silently do nothing in CI.

**Suggestion:** `resolveContext` must parse the GitHub Actions environment and return `repo` and `prNumber` alongside the existing fields. For example:

```js
const [owner, name] = (process.env.GITHUB_REPOSITORY || '').split('/')
const repo = owner && name ? {owner, name} : null

const eventPath = process.env.GITHUB_EVENT_PATH
let prNumber = null
if (eventPath) {
  const event = JSON.parse(await readFile(eventPath, 'utf-8'))
  prNumber = event.pull_request?.number ?? event.number ?? null
}
```

Include `repo` and `prNumber` in the returned context object from every code path.

isaacroldan (commented):
---

.github/workflows/breaking-change-check.yml
```diff
@@ -0,0 +1,80 @@
...
+          # reach the divergence point. With fetch-depth: 1 the merge-base
+          # call fails and the script falls back to scanning everything,
+          # surfacing every pre-existing major changeset as "new".
+          fetch-depth: 0
+      - name: Setup deps
+        uses: ./.github/actions/setup-cli-deps
+        with:
+          node-version: ${{ env.DEFAULT_NODE_VERSION }}
+      - name: Build
+        run: pnpm nx run-many --all --skip-nx-cache --target=build --output-style=stream
```
isaacroldan: why do you need to build to find breaking changes?

.github/workflows/breaking-change-check.yml
```diff
@@ -0,0 +1,97 @@
...
+# Runs the breaking-change detector on PR opens, updates, AND on review
+# submissions. Re-running on review submissions lets the script flip from
+# red to green when a code-owner approves the PR — see the override logic
+# in workspace/src/major-change-check.js.
+#
+# Lives in a dedicated workflow so review events don't re-trigger the full
+# unit/e2e/type-diff suite.
+on:
+  pull_request:
+  pull_request_review:
```
isaacroldan, line 12: 💬 **Suggestion**: The `edited` type fires when a review's body text is edited, not when its state changes. Editing a review comment doesn't change whether the PR is approved, so this trigger re-runs the full breaking-change check (clone, diff, scan) for zero functional effect. Each unnecessary run costs a full repo clone + schema diff scan. The `submitted` and `dismissed` types already cover all state-changing review actions.

**Suggestion:** 
```suggestion
    types: [submitted, dismissed]
```

.github/workflows/breaking-change-check.yml
```diff
@@ -0,0 +1,97 @@
...
+
+env:
+  DEBUG: '1'
+  SHOPIFY_CLI_ENV: development
+  SHOPIFY_CONFIG: debug
+  PNPM_VERSION: '10.11.1'
+  BUNDLE_WITHOUT: 'test:development'
+  GH_TOKEN: ${{ secrets.SHOPIFY_GH_READ_CONTENT_TOKEN }}
+  GH_TOKEN_SHOP: ${{ secrets.SHOP_GH_READ_CONTENT_TOKEN }}
+  DEFAULT_NODE_VERSION: '24.1.0'
```
isaacroldan: 💡 **Improvement**: The breaking-change-check workflow pins `DEFAULT_NODE_VERSION: '24.1.0'` while tests-pr.yml, tests-main.yml, and tests-manual.yml all use `'26.1.0'`. This means the script runs on a different Node version than the rest of the CI suite. Likely an oversight from when the job was extracted from tests-pr.yml.

**Suggestion:** 
```suggestion
  DEFAULT_NODE_VERSION: '26.1.0'
```

workspace/src/major-change-check.js
isaacroldan, line 1: 📐 **Design**: The `parseCodeowners`, `codeownersPatternToRegExp`, and `ownersForFiles` functions (~80 lines of exported, tested code) are only consumed at line 295 for a `logMessage` call — they never gate the approval decision. The doc-comment at line 248–258 explains why (token limitations prevent resolving team membership), and the transparency logging is a reasonable design choice.

However, the exported/tested surface area creates an expectation that ownership is being enforced. The PR description says 'code-owner-approval override' but the implementation is really 'write-access-approval override'. This isn't wrong — just worth noting that future readers may be surprised by the gap between the CODEOWNERS machinery and its actual role.

**Suggestion:** Consider either (a) making the CODEOWNERS functions private (unexport) and reducing test coverage to match their log-only role, or (b) adding a brief note in the function doc-comment clarifying that CODEOWNERS matching is informational-only. The current state is correct but potentially misleading.

workspace/src/major-change-check.js
```diff
@@ -54,41 +54,278 @@ const currentDirectory = path.join(url.fileURLToPath(new URL('.', import.meta.ur
...
 }
 
+/**
+ * Default GitHub compare-API fetcher. Uses the standard `GITHUB_TOKEN` so
+ * it works on first-party PRs; falls back to unauthenticated for forks
+ * (still works for public-repo compare). On any failure we return null and
+ * the caller falls back to scanning everything — i.e. degrades to legacy
+ * behavior, never silently ignores potential breaking changes.
+ */
+async function defaultFetchCompare(repo, base, head) {
```
isaacroldan: 🎨 **Code Style**: `defaultFetchCompare` (lines 134–145) has zero callers anywhere in the codebase. It was likely scaffolded for a compare-API-based approach that was replaced by the git-based diff in `resolveContext`. All three analysis passes confirmed no references exist.

**Suggestion:** Delete the function and its doc-comment (lines 134–145).

```suggestion
```

workspace/src/major-change-check.js
```diff
@@ -54,41 +54,278 @@ const currentDirectory = path.join(url.fileURLToPath(new URL('.', import.meta.ur
...
+ */
+export async function findCodeownerApproval({
+  repo,
+  prNumber,
+  changedFiles,
+  fetchReviews = defaultFetchReviews,
+  fetchCodeowners = defaultFetchCodeowners,
+  fetchPermission = defaultFetchPermission,
+} = {}) {
+  if (!prNumber) return {approved: false, approver: null, reason: 'no PR number in event'}
```
isaacroldan, line 273: 🐛 **Bug**: `findCodeownerApproval` guards on `!prNumber` but not on `!repo`. If `repo` is null and `prNumber` is truthy, `defaultFetchReviews(null, prNumber)` constructs `${repo.owner}` which throws `TypeError: Cannot read properties of null (reading 'owner')`. In CI this is unlikely (GITHUB_REPOSITORY is always set), but `runMain` at line 935 passes `context.repo` through without a guard, and the script can be invoked locally. The crash is accidentally fail-safe (non-zero exit keeps the check red), but the error message would be baffling.

**Suggestion:** Add a `repo` guard alongside the `prNumber` one:

```suggestion
  if (!repo) return {approved: false, approver: null, reason: 'no repo context'}
  if (!prNumber) return {approved: false, approver: null, reason: 'no PR number in event'}
```

isaacroldan (commented):
---

PR #7470, opened 2026-05-05
Watch asset directories for changes during `dev`

melissaluu:

Closes  https://github.com/shop/issues-admin-extensibility/issues/2461

When developers add new files to the watched `assets` directories during a `dev` sesstion, they were not being picked up by the file watcher and meant that adding an asset image would not trigger a rebuild or hot-reload.

Introduces a mechanism for `ExtensionInstance` to track additional watched paths beyond its source imports, and wires it into the file watcher and build pipeline:

- Adds `addWatchedPath`, `getWatchedPaths`, and `clearWatchedPaths` methods to `ExtensionInstance`. Paths are stored in a deduplicated `Set` and are included in the output of `watchedFiles()`.
- In `executeIncludeAssetsStep`, directory-type asset sources are registered as watched paths on the extension after being copied, so the watcher knows to monitor them.
- In `FileWatcher`, the `extensionWatchedFiles` map now tracks both files and directories. A new `getExtensionHandlesForFilePath` method resolves the owning extension for any file path by checking both exact matches and prefix matches against watched directories. Directories themselves are excluded from the paths passed to chokidar (since chokidar already watches the extension directory), but files inside them are still attributed to the correct extension.
- In `AppEventWatcher`, after building extensions, the set of watched paths is compared before and after the build. If new paths were registered (e.g., a new asset directory was discovered), the file watcher is restarted to pick them up.

### How to test your changes?

[Video](https://drive.google.com/file/d/1MArEMuYI3uLuvSWE9p_5OUYyUbZDIFuy/view?usp=drive_link)

1. Create an `app-home` extension that uses `include_assets` with a `asset` config key pointing to a directory.
2. Run `shopify app dev`.
3. Add a file inside the watched asset directory.
4. Confirm that the change is detected as a new websocket message and triggers the expected rebuild/update.
5. Update the app to show the new image and verify the image is displayed correctly
6. Rename the new file and ensure the app now shows a broken image
7. Rename the file back to the original name and check if the app now shows the correct image again
8. Delete the file and check if the app shows the broken image

### Checklist

- [ ] The change is user-facing — I've identified the correct [bump type](https://github.com/Shopify/cli/blob/main/CONTRIBUTING.md#choosing-the-right-bump-type) (`patch` for bug fixes · `minor` for new features · `major` for [breaking changes](https://github.com/Shopify/cli/blob/main/CONTRIBUTING.md#what-counts-as-a-breaking-change)) and added a changeset with `pnpm changeset addgeeck`

### DIFF
--- a/packages/app/src/cli/services/dev/app-events/file-watcher.ts
+++ b/packages/app/src/cli/services/dev/app-events/file-watcher.ts
@@ -1,7 +1,7 @@
 /* eslint-disable no-case-declarations */
 import {AppLinkedInterface} from '../../../models/app/app.js'
 import {configurationFileNames} from '../../../constants.js'
-import {dirname, joinPath, normalizePath, relativePath} from '@shopify/cli-kit/node/path'
+import {basename, dirname, joinPath, normalizePath, relativePath} from '@shopify/cli-kit/node/path'
 import {FSWatcher} from 'chokidar'
 import {outputDebug} from '@shopify/cli-kit/node/output'
 import {AbortSignal} from '@shopify/cli-kit/node/abort'
@@ -59,7 +59,7 @@ export class FileWatcher {
   private watcher?: FSWatcher
   private readonly debouncedEmit: () => void
   private readonly ignored: {[key: string]: ignore.Ignore | undefined} = {}
-  // Map of file paths to the extension handles that watch them
+  // Map of watched paths (files and directories) to the extension handles that own them
   private readonly extensionWatchedFiles = new Map<string, Set<string>>()
 
   constructor(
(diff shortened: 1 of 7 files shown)


melissaluu:
* **#7470** <a href="https://app.graphite.com/github/pr/Shopify/cli/7470?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Shopify/cli/7470?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/app/src/cli/services/dev/app-events/file-watcher.ts
```diff
@@ -223,6 +239,11 @@ export class FileWatcher {
   private shouldIgnoreEvent(event: WatcherEvent) {
     if (event.type === 'extension_folder_deleted' || event.type === 'extension_folder_created') return false
 
+    // For deleted files, skip the live watchPaths check — the file is already gone from disk
+    // so globSync won't find it. The file was already validated as belonging to the extension
+    // via the cached extensionWatchedFiles map earlier in the flow.
+    if (event.type === 'file_deleted') return false
```
melissaluu, line 245: Added this to enable deleting an asset will emit a websocket update event.   
  
Without it, since the `shouldIgnoreEvent()` calls `extension.watchedFiles()` which runs `globSync` live against the filesystem. For `delete` events, by the time this check runs, the file is already gone from disk — so `globSync` never returns it, `isAValidWatchedPath` is `false`, and the event gets silently dropped.

alfonso-noriega (changes requested):
We should keep the changes out of include-asset-steps and app-event-watcher. The problem here (if i understood properly) is that the file watcher is not prepared to watch folders. To focus on that the green-path would be to:
1 -  in extension-instance.ts:431 watchedFiles support a folders property that is not included in the glob exploration
2 - define/extend in the ui-extension or the desired specification the `devSessionWatchConfig` which defines this folders property
3 - in the file watcher we would already have the proper folder path passed to chokidar, but your changes to avoid deleted files make sense to me.

melissaluu:
We're pivoting to a more generic solution https://github.com/Shopify/cli/pull/7491

PR #7473, opened 2026-05-06
[Security] Fix command injection risk in cloudflared installer

gonzaloriestra:

Fixes a potential command injection vulnerability in the Cloudflare installer. The previous implementation used `execSync` with a template literal containing a variable derived from an environment variable (`SHOPIFY_CLI_CLOUDFLARED_PATH`), which could be exploited to execute arbitrary commands if the environment was compromised.

- Replaces direct `child_process.execSync` and `child_process.execFileSync` calls with secure wrappers `exec` and `captureOutputWithExitCode` from `@shopify/cli-kit/node/system`.
- Switched to argument-list-based command execution for `tar` to prevent shell injection.
- Removed direct `child_process` imports to ensure all command executions go through `@shopify/cli-kit` which implements `checkCommandSafety`.
- Updated unit tests in `install-cloudflared.test.ts` to match the new asynchronous implementation and use the correct mocks.

### How to test your changes?

Run the unit tests for the affected package:
```bash
pnpm --filter @shopify/plugin-cloudflare vitest run
```

### Checklist

---
*PR created automatically by Jules for task [17491907231653004163](https://jules.google.com/task/17491907231653004163) started by @gonzaloriestra*

### DIFF
--- a/packages/plugin-cloudflare/src/install-cloudflared.test.ts
+++ b/packages/plugin-cloudflare/src/install-cloudflared.test.ts
@@ -1,18 +1,19 @@
 import install, {CURRENT_CLOUDFLARE_VERSION, versionIsGreaterThan} from './install-cloudflared.js'
 import * as fsActions from '@shopify/cli-kit/node/fs'
 import * as http from '@shopify/cli-kit/node/http'
+import * as system from '@shopify/cli-kit/node/system'
 import {beforeEach, describe, expect, test, vi} from 'vitest'
 import util from 'util'
 
 import {WriteStream} from 'fs'
-// eslint-disable-next-line no-restricted-imports
-import * as childProcess from 'child_process'
 
-vi.mock('child_process')
+vi.mock('@shopify/cli-kit/node/system')
 vi.mock('stream')
 
 describe('install-cloudflare', () => {
   beforeEach(() => {
+    vi.mocked(system.captureOutputWithExitCode).mockResolvedValue({stdout: '', stderr: '', exitCode: 0})
+    vi.mocked(system.exec).mockResolvedValue(undefined)
     vi.spyOn(util, 'promisify').mockReturnValue(vi.fn().mockReturnValue(Promise.resolve()))
     vi.spyOn(http, 'fetch').mockReturnValue(Promise.resolve({ok: true, body: {pipe: vi.fn()}} as any))
     vi.spyOn(fsActions, 'fileExistsSync').mockReturnValueOnce(false)
@@ -46,6 +47,11 @@ describe('install-cloudflare', () => {
       'https://github.com/cloudflare/cloudflared/releases/download/2024.8.2/cloudflared-darwin-amd64.tgz',
       expect.anything(),
       'slow-request',
+    )
+    expect(system.exec).toHaveBeenCalledWith(
+      'tar',
+      ['-xzf', expect.stringMatching(/\.tgz$/)],
+      expect.objectContaining({cwd: expect.any(String)}),
     )
   })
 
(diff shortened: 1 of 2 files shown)


packages/plugin-cloudflare/src/install-cloudflared.test.ts
craigmichaelmartin, line 1: 💡 **Improvement**: The catch block at line 93 of `install-cloudflared.ts` handles the case where `captureOutputWithExitCode` throws (e.g., binary not found / ENOENT). Since `captureOutputWithExitCode` does NOT throw on non-zero exits (unlike the old `execFileSync`), this catch block now only fires for truly exceptional scenarios. There are no tests exercising this failure path with the new async implementation. Adding a test where `captureOutputWithExitCode` rejects would verify the fallback reinstall behavior and document the changed semantics.

**Suggestion:** Add a test case:
```typescript
test('install reinstalls if version check throws', async () => {
  const env = {}
  vi.spyOn(fsActions, 'fileExistsSync').mockReturnValueOnce(true)
  vi.mocked(system.captureOutputWithExitCode).mockRejectedValue(new Error('ENOENT'))
  await install(env, 'linux', 'x64')
  expect(http.fetch).toHaveBeenCalled()
})
```

craigmichaelmartin, line 1: (Ref Line 43) 💡 **Improvement**: The tests mock `system.exec` and `system.captureOutputWithExitCode` but never assert they were called with the correct arguments. Since this PR's primary goal is eliminating command injection by switching from string-interpolated shell commands to array-based arguments, the tests should verify the exact argument format — especially that the `tar` command on macOS uses `exec('tar', ['-xzf', filename])` rather than a concatenated string. Without these assertions, a regression that re-introduces string-based execution would pass all tests silently.

**Suggestion:** Add assertions in the macOS install test to verify the tar extraction call, and in version-check tests to verify `captureOutputWithExitCode` arguments. For example, in the macOS test:

```suggestion
    expect(system.exec).toHaveBeenCalledWith('tar', ['-xzf', 'cloudflared-darwin-amd64.tgz'], expect.objectContaining({cwd: expect.any(String)}))
```

And in the version-check tests:
```javascript
expect(system.captureOutputWithExitCode).toHaveBeenCalledWith(expect.any(String), ['--version'])
```

packages/plugin-cloudflare/src/install-cloudflared.ts
```diff
@@ -82,7 +81,8 @@ export default async function install(env = process.env, platform = process.plat
   if (fileExistsSync(binTarget)) {
     // --version returns an string like "cloudflared version 2023.3.1 (built 2023-03-13-1444 UTC)"
     try {
-      const versionArray = execFileSync(binTarget, ['--version'], {encoding: 'utf8'}).split(' ')
+      const {stdout} = await captureOutputWithExitCode(binTarget, ['--version'])
```
craigmichaelmartin: 🔒 **Security**: The old `execFileSync` would throw on a non-zero exit code, hitting the catch block at line 93 which logs a debug message and triggers reinstallation. The new `captureOutputWithExitCode` passes `reject: false` to execa internally, so a non-zero exit code returns normally instead of throwing. This means the catch block is now dead code for the most common failure mode (e.g., corrupted binary, permission denied). The current behavior accidentally works because bad stdout produces `'0.0.0'` via the fallback, which triggers reinstall — but this is fragile. If version parsing logic changes, a failed version check could silently skip reinstallation.

**Suggestion:** Consider checking `exitCode` explicitly to preserve the original error-handling semantics:

```suggestion
      const {stdout, exitCode} = await captureOutputWithExitCode(binTarget, ['--version'])
      if (exitCode !== 0) throw new Error(`cloudflared --version exited with code ${exitCode}`)
      const versionArray = stdout.split(' ')
```

craigmichaelmartin (changes requested):

packages/plugin-cloudflare/src/install-cloudflared.test.ts
gonzaloriestra: Added coverage for rejected version checks reinstalling.

gonzaloriestra: Asserted the array-based tar and version-check command arguments.

packages/plugin-cloudflare/src/install-cloudflared.ts
gonzaloriestra: Added explicit non-zero exit handling for the version check.

craigmichaelmartin (approved):

PR #7478, opened 2026-05-06
Avoid metrics-only store analytics lookups

dmerand:
## Summary
- align the CLI Monorail mirror to `app_cli3_command/1.24`
- remove `shop_id` from the store command analytics payload
- add `store_fqdn_validated` so requested store attribution is distinct from validated store attribution
- rename the store analytics helper to `attribution.ts`

## Why
- follow the Slack-thread direction to avoid extra CLI network calls purely for analytics enrichment
- keep store attribution owned by the store auth / execute flows instead of shared base parsing
- preserve requested-store attribution even when `store auth` or `store execute` fail before an authenticated store context is established

## Behavior changes
- `store auth` records the requested store early and marks it validated only after token exchange succeeds
- `store execute` records the requested store before request validation and records the validated store after loading the stored session
- `shop_id` is no longer emitted, including from the `publicApiVersions` request path

| Command/path | Requested store recorded | Validated store recorded | Final `store_fqdn_validated` |
|---|---|---|---|
| `store auth` before OAuth completes | yes | no | `false` |
| `store auth` after token exchange succeeds | yes | yes | `true` |
| `store execute` before stored session load | yes | no | `false` |
| `store execute` after stored session loads | yes | yes | `true` |
| `store execute` with no stored auth / early failure | yes | no | `false` |

`store_fqdn_validated=true` means the FQDN was validated against an authenticated store context. It does not imply the command later completed successfully.

Depends on Shopify/monorail#23622.


packages/store/src/cli/utilities/store-command.ts
gonzaloriestra, line 1: I prefer to have this kind of things at this level, so that we don't forget about it if we add new store commands. But I know both commands work differently now. We can explore it later if we add more commands.

gonzaloriestra (commented):
Code LGTM, but I'd like to test it after merging the new Monorail schema

packages/store/src/cli/utilities/store-command.ts
dmerand: Can you please say a bit more about what we might forget to do here?

dmerand:
FYI @gonzaloriestra the monorail schema has merged + deployed.

packages/store/src/cli/utilities/store-command.ts
gonzaloriestra: I meant adding the metrics about the store

gonzaloriestra (approved):
It works great!  
  
I found a small issue: if the command fails too early, for example when you don't add the `--query` flag, the store is not sent. But I think it's not important, I'd just ship it.

PR #7480, opened 2026-05-06
prepend api_version to schema

lopert:

Part of https://github.com/shop/issues-shopifyvm/issues/945

When a developer bumps `api_version` in a function's `shopify.extension.toml`, the on-disk `schema.graphql` becomes stale. The next `shopify app function build` then fails with cryptic `graphql-code-generator` errors like:

> Cannot query field "parentRelationship" on type "CartLine"

…with no indication that the underlying problem is a stale schema that just needs to be re-fetched. Today the build path is fully local — it never consults `api_version` and `schema.graphql` carries no version metadata, so staleness is undetectable.

The error surfaced gives them no actionable hint pointing at `shopify app function schema`.

Stamps the `api_version` into `schema.graphql` when `shopify app function schema` writes it, then validates that marker against the extension's TOML at the start of every `shopify app function build`. When they disagree, the build aborts up-front with an actionable message instead of letting codegen fail mysteriously.

**New file**

- `packages/app/src/cli/services/function/schema-version.ts` — self-contained helpers:
  - `prependSchemaVersionHeader(definition, version)` — adds a `# api_version: <ver>` marker line to the top of the SDL.
  - `readSchemaApiVersion(path)` — scans the leading comment block for the marker; returns `undefined` for missing files or legacy schemas (no marker).
  - `validateSchemaApiVersion({directory, localIdentifier, apiVersion})` — throws an `AbortError` with remediation when the on-disk marker disagrees with the configured `api_version`.

**Modified files**

- `services/generate-schema.ts` — `shopify app function schema` now writes the marker as the first line of the SDL (both file and stdout output).
- `services/build/extension.ts` — `buildFunctionExtension` invokes `validateSchemaApiVersion` once, up-front, before any function build work (typegen, esbuild, javy, wasm-opt, trampoline).

**Behavior**

- **Mismatched marker** → fails fast with:
  > The `schema.graphql` file for `<extension>` was generated for api_version `2025-07` but your function is now on api_version `2025-10`. Run `shopify app function schema` to refresh it.
- **Missing schema file** → no-op (out of scope for this change; codegen handles its own missing-file error).
- **Legacy schema with no marker** → no-op (don't break existing setups; the marker self-heals on the next `shopify app function schema`).

The schema header is a single line:
```
# api_version: 2025-10

<SDL>
```

### How to test your changes?

- Checkout this branch
- In a Shopify app with a function (e.g. discount function), set `api_version = "2025-07"` in the function's `shopify.extension.toml`.
- Run the schema command locally, and confirm `schema.graphql` now starts with `# api_version: 2025-07`.
```
pnpm shopify app function schema --path /Users/lopert/code/functions/schema-sync
```
- Bump `api_version` to `"2025-10"` in the TOML.
- Run the build command locally, it should abort immediately with a message naming both versions and pointing at `shopify app function schema`.
```
pnpm shopify app function build --path /Users/lopert/code/functions/schema-sync
```
- Run the schema command again to refresh, then the build command — it should now succeed.
- Sanity check the legacy path: delete the marker line from `schema.graphql` (simulate an older fetched file). The build command should behave exactly as it did before this change (no false-positive abort).

### Post-release steps

N/A.

### Checklist

### DIFF
--- a/packages/app/src/cli/services/build/extension.test.ts
+++ b/packages/app/src/cli/services/build/extension.test.ts
@@ -1,6 +1,7 @@
 import {buildFunctionExtension} from './extension.js'
 import {testFunctionExtension} from '../../models/app/app.test-data.js'
 import {buildGraphqlTypes, buildJSFunction, runWasmOpt, runTrampoline} from '../function/build.js'
+import {validateSchemaApiVersion} from '../function/schema-version.js'
 import {ExtensionInstance} from '../../models/extensions/extension-instance.js'
 import {FunctionConfigType} from '../../models/extensions/specifications/function.js'
 import {beforeEach, describe, expect, test, vi} from 'vitest'
@@ -12,6 +13,7 @@ import {joinPath} from '@shopify/cli-kit/node/path'
 
 vi.mock('@shopify/cli-kit/node/system')
 vi.mock('../function/build.js')
+vi.mock('../function/schema-version.js')
 vi.mock('proper-lockfile')
 vi.mock('@shopify/cli-kit/node/fs')
 
@@ -416,6 +418,26 @@ describe('buildFunctionExtension', () => {
     })
     expect(releaseLock).toHaveBeenCalled()
     expect(runWasmOpt).toHaveBeenCalled()
+  })
+
+  test('calls validateSchemaApiVersion with the values from the extension config', async () => {
+    // When
+    await expect(
+      buildFunctionExtension(extension, {
+        stdout,
+        stderr,
+        signal,
+        app,
+        environment: 'production',
+      }),
+    ).resolves.toBeUndefined()
+
+    // Then
+    expect(validateSchemaApiVersion).toHaveBeenCalledWith({
+      directory: extension.directory,
+      localIdentifier: extension.localIdentifier,
+      apiVersion: extension.configuration.api_version,
+    })
   })
 
   test('does not rebundle when build.path stays in the default output directory', async () => {
--- a/packages/app/src/cli/services/build/extension.ts
+++ b/packages/app/src/cli/services/build/extension.ts
@@ -2,6 +2,7 @@ import {formatBundleSize} from './bundle-size.js'
 import {AppInterface} from '../../models/app/app.js'
 import {bundleExtension} from '../extensions/bundle.js'
 import {buildGraphqlTypes, buildJSFunction, runTrampoline, runWasmOpt} from '../function/build.js'
+import {validateSchemaApiVersion} from '../function/schema-version.js'
 import {ExtensionInstance} from '../../models/extensions/extension-instance.js'
 import {FunctionConfigType} from '../../models/extensions/specifications/function.js'
 import {exec} from '@shopify/cli-kit/node/system'
@@ -156,11 +157,17 @@ export async function buildFunctionExtension(
   }
 
   try {
+    const functionConfiguration = (extension as ExtensionInstance<FunctionConfigType>).configuration
     const bundlePath = extension.outputPath
-    const relativeBuildPath =
-      (extension as ExtensionInstance<FunctionConfigType>).configuration.build?.path ?? extension.outputRelativePath
+    const relativeBuildPath = functionConfiguration.build?.path ?? extension.outputRelativePath
 
     extension.outputPath = joinPath(extension.directory, relativeBuildPath)
+
+    await validateSchemaApiVersion({
+      directory: extension.directory,
+      localIdentifier: extension.localIdentifier,
+      apiVersion: functionConfiguration.api_version,
+    })
 
     if (extension.isJavaScript) {
       await runCommandOrBuildJSFunction(extension, options)
--- a/packages/app/src/cli/services/function/schema-version.test.ts
+++ b/packages/app/src/cli/services/function/schema-version.test.ts
@@ -1,0 +1,109 @@
+import {
+  prependSchemaVersionHeader,
+  readSchemaApiVersion,
+  validateSchemaApiVersion,
+  SCHEMA_VERSION_MARKER_PREFIX,
+} from './schema-version.js'
+import {describe, expect, test} from 'vitest'
+import {AbortError} from '@shopify/cli-kit/node/error'
+import {inTemporaryDirectory, writeFile} from '@shopify/cli-kit/node/fs'
+import {joinPath} from '@shopify/cli-kit/node/path'
+
+function options(directory: string, apiVersion: string) {
+  return {directory, localIdentifier: 'my-function', apiVersion}
+}
+
+describe('prependSchemaVersionHeader', () => {
+  test('prepends a comment block with the version marker', () => {
+    const result = prependSchemaVersionHeader('type Query { id: ID }', '2025-10')
+
+    expect(result.startsWith(`${SCHEMA_VERSION_MARKER_PREFIX}2025-10\n`)).toBe(true)
+    expect(result.endsWith('type Query { id: ID }')).toBe(true)
+  })
+})
+
+describe('readSchemaApiVersion', () => {
+  test('returns the version when the marker is present', async () => {
+    await inTemporaryDirectory(async (tmpDir) => {
+      const path = joinPath(tmpDir, 'schema.graphql')
+      await writeFile(path, prependSchemaVersionHeader('type Query { id: ID }', '2025-10'))
+
+      await expect(readSchemaApiVersion(path)).resolves.toEqual('2025-10')
+    })
+  })
+
+  test('returns undefined when the file has no marker', async () => {
+    await inTemporaryDirectory(async (tmpDir) => {
+      const path = joinPath(tmpDir, 'schema.graphql')
+      await writeFile(path, '# some other comment\ntype Query { id: ID }')
+
+      await expect(readSchemaApiVersion(path)).resolves.toBeUndefined()
+    })
+  })
+
+  test('returns undefined when the file does not exist', async () => {
+    await inTemporaryDirectory(async (tmpDir) => {
+      await expect(readSchemaApiVersion(joinPath(tmpDir, 'missing.graphql'))).resolves.toBeUndefined()
+    })
+  })
+
+  test('does not match the marker once SDL content has started', async () => {
+    await inTemporaryDirectory(async (tmpDir) => {
+      const path = joinPath(tmpDir, 'schema.graphql')
+      // Marker buried after SDL content should be ignored.
+      await writeFile(path, `type Query { id: ID }\n${SCHEMA_VERSION_MARKER_PREFIX}2025-10\n`)
+
+      await expect(readSchemaApiVersion(path)).resolves.toBeUndefined()
+    })
+  })
+
+  test('trims surrounding whitespace from the marker value', async () => {
+    await inTemporaryDirectory(async (tmpDir) => {
+      const path = joinPath(tmpDir, 'schema.graphql')
+      await writeFile(path, `${SCHEMA_VERSION_MARKER_PREFIX}  2025-10  \ntype Query { id: ID }`)
+
+      await expect(readSchemaApiVersion(path)).resolves.toEqual('2025-10')
+    })
+  })
+})
+
+describe('validateSchemaApiVersion', () => {
+  test('no-ops when the schema file does not exist', async () => {
+    await inTemporaryDirectory(async (tmpDir) => {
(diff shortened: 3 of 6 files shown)


packages/app/src/cli/services/function/schema-version.test.ts
```diff
@@ -0,0 +1,110 @@
...
+
+function options(directory: string, apiVersion: string) {
+  return {directory, localIdentifier: 'my-function', apiVersion}
+}
+
+describe('prependSchemaVersionHeader', () => {
+  test('prepends a comment block with the version marker', () => {
+    const result = prependSchemaVersionHeader('type Query { id: ID }', '2025-10')
+
+    expect(result.startsWith(`${SCHEMA_VERSION_MARKER_PREFIX}2025-10`)).toBe(true)
```
adampetro: nit: can we assert there's a newline at the end here?
```suggestion
    expect(result.startsWith(`${SCHEMA_VERSION_MARKER_PREFIX}2025-10\n`)).toBe(true)
```

packages/app/src/cli/services/function/schema-version.test.ts
```diff
@@ -0,0 +1,110 @@
...
+})
+
+describe('validateSchemaApiVersion', () => {
+  test('no-ops when the schema file does not exist', async () => {
+    await inTemporaryDirectory(async (tmpDir) => {
+      await expect(validateSchemaApiVersion(options(tmpDir, '2025-10'))).resolves.toBeUndefined()
+    })
+  })
+
+  test('no-ops when the schema file has no version marker (legacy)', async () => {
```
adampetro: nit: can we remove legacy, I think it's something good to have in perpetuity, but legacy implies to me that at some point this case might go away

packages/app/src/cli/services/function/schema-version.test.ts
```diff
@@ -0,0 +1,110 @@
...
+      await writeFile(
+        joinPath(tmpDir, 'schema.graphql'),
+        prependSchemaVersionHeader('type Query { id: ID }', '2025-07'),
+      )
+
+      const result = validateSchemaApiVersion(options(tmpDir, '2025-10'))
+
+      await expect(result).rejects.toThrow(AbortError)
+      await expect(result).rejects.toThrow(/2025-07/)
+      await expect(result).rejects.toThrow(/2025-10/)
```
adampetro: can we combine these into one regex so we know that the ordering is correct?

packages/app/src/cli/services/function/schema-version.ts
```diff
@@ -0,0 +1,77 @@
...
+
+  const contents = await readFile(filePath)
+  // Only inspect the leading comment block — bail out as soon as we see a
+  // non-comment, non-empty line so we don't scan the whole SDL.
+  for (const line of contents.split(/\r?\n/)) {
+    if (line.startsWith(SCHEMA_VERSION_MARKER_PREFIX)) {
+      return line.slice(SCHEMA_VERSION_MARKER_PREFIX.length)
+    }
+    if (!line.startsWith('#')) break
+  }
```
adampetro: do we really need to inspect the whole comment block and not just the first line?

packages/app/src/cli/services/function/schema-version.ts
```diff
@@ -0,0 +1,77 @@
...
+  localIdentifier: string
+  apiVersion: string
+}
+
+export async function validateSchemaApiVersion({
+  directory,
+  localIdentifier,
+  apiVersion,
+}: ValidateSchemaApiVersionOptions): Promise<void> {
+  const schemaPath = joinPath(directory, 'schema.graphql')
```
adampetro, line 66: hmm, not ideal that we're hardcoding the file name but I don't think we have this in the config anywhere 🤔

packages/app/src/cli/services/function/schema-version.ts
```diff
@@ -0,0 +1,77 @@
...
+  if (!(await fileExists(filePath))) {
+    return undefined
+  }
+
+  const contents = await readFile(filePath)
+  // Only inspect the leading comment block — bail out as soon as we see a
+  // non-comment, non-empty line so we don't scan the whole SDL.
+  for (const line of contents.split(/\r?\n/)) {
+    if (line.startsWith(SCHEMA_VERSION_MARKER_PREFIX)) {
+      return line.slice(SCHEMA_VERSION_MARKER_PREFIX.length)
```
adampetro: should we trim too in case there is leading or trailing whitespace?

packages/app/src/cli/services/build/extension.test.ts
```diff
@@ -418,6 +420,26 @@ describe('buildFunctionExtension', () => {
     expect(runWasmOpt).toHaveBeenCalled()
   })
 
+  test('validates the schema api_version with the values from the extension config', async () => {
```
saga-dasgupta: The title of this test makes me think that there will be some validation assertion but only thing we do is assert that we called the validate function.

packages/app/src/cli/services/build/extension.ts
```diff
@@ -156,12 +157,20 @@ export async function buildFunctionExtension(
...
   try {
+    const functionConfiguration = (extension as ExtensionInstance<FunctionConfigType>).configuration
     const bundlePath = extension.outputPath
-    const relativeBuildPath =
-      (extension as ExtensionInstance<FunctionConfigType>).configuration.build?.path ?? extension.outputRelativePath
+    const relativeBuildPath = functionConfiguration.build?.path ?? extension.outputRelativePath
 
     extension.outputPath = joinPath(extension.directory, relativeBuildPath)
 
+    if (functionConfiguration.api_version) {
```
saga-dasgupta: Why do we need this condition, why wouldn't there be an api_version in the function_configuration?

packages/app/src/cli/services/function/schema-version.ts
```diff
@@ -0,0 +1,77 @@
...
+  directory: string
+  localIdentifier: string
+  apiVersion: string
+}
+
+export async function validateSchemaApiVersion({
+  directory,
+  localIdentifier,
+  apiVersion,
+}: ValidateSchemaApiVersionOptions): Promise<void> {
```
saga-dasgupta, line 65: Should the validate function just return a pure True or False type validation (with it defaulting to True if it cant verify because the schema file doesn't have the version) and then the extension file is responsible for throwing the `AbortError` like it does for other cases?  So this method is pure validation, not responsible for responding to the user, that happens at the caller?

packages/app/src/cli/services/generate-schema.ts
```diff
@@ -22,7 +23,7 @@ export async function generateSchemaService(options: GenerateSchemaOptions) {
   const apiKey = app.configuration.client_id
   const {api_version: version, type, targeting} = extension.configuration
   const usingTargets = Boolean(targeting?.length)
-  const definition = await (usingTargets
+  const fetchedDefinition = await (usingTargets
```
saga-dasgupta, line 26: I dont understand the difference between `definition` and `fetched_definition`, what if we keep this definition and call the other `definition_with_version` or something?

packages/app/src/cli/services/function/schema-version.ts
lopert: My concern here was trying to not rely on being the first line. Stuff like:
 - Someone adds a # DO NOT EDIT / license / TODO comment above it
 - A future change prepends another metadata line (e.g. # generated_at: …)
 - Some tool reformats the file

I guess for now, since I'm the only one adding a special metadata comment, it's fine to just look at the first line.

packages/app/src/cli/services/function/schema-version.ts
lopert: Like we discussed on slack, we've got this hardcoded in a few places.
It's actually just the `function build` command that ends up looking at the `package.json/codegen/schema` property, so there's already a disconnect today if someone changes that file.

I've [created an issue](https://github.com/shop/issues-shopifyvm/issues/1020) for this, but for this PR, I'll keep things consistent with the rest of the code.

packages/app/src/cli/services/build/extension.test.ts
lopert: Fair enough. The actual logic is covered in the schema-version code so this is in fact just making sure we're invoking it with the right params

packages/app/src/cli/services/function/schema-version.ts
lopert: I don't think so? I'm not sure what the `other cases` are here.
In extensions.ts, we're catching errors from other things (such as trying to acquire the lock) and re-throwing an AbortError. I think this is different than validators, such as [validateShopifyFunctionPackageVersion](https://github.com/Shopify/cli/blob/9bdcacce222ca901aa39fcc6c150dc260ea73d72/packages/app/src/cli/services/function/build.ts#L185-L205), which do throw their own errors.

The validator also knows what it's looking for, and the error provided can include more detail (i.e. the different versions).

packages/app/src/cli/services/generate-schema.ts
lopert: `definition` is essentially what was here before, the contents we're about to write to the local schema. 

I introduced `fetchedDefinition` to name the intermediate value: the raw schema we just grabbed from the server, before the prepending step. 

Reading top-down: fetch -> add version header -> write. 
Whereas before it was just fetch -> write.
Open to renaming if it still reads confusingly, but it felt clearer to me separated.

packages/app/src/cli/services/function/schema-version.ts
```diff
@@ -0,0 +1,79 @@
...
+ */
+export async function readSchemaApiVersion(filePath: string): Promise<string | undefined> {
+  if (!(await fileExists(filePath))) {
+    outputDebug(`Could not determine api_version: schema file not found at ${filePath}.`)
+    return undefined
+  }
+
+  const contents = await readFile(filePath)
+  // The marker is always written as the first line by `prependSchemaVersionHeader`.
+  const firstLine = contents.split('\n', 1)[0]!
```
adampetro, line 34: is this resilient to a file without newlines?

adampetro: nevermind, it should always include the string as the first element

adampetro (approved):

packages/app/src/cli/services/generate-schema.ts
saga-dasgupta: Yeah I guess fetched_definition doesn't immediately make sense to me when I read the code but its okay, its just a variable name, I won't let it block you from shipping the PR.

packages/app/src/cli/services/function/schema-version.ts
saga-dasgupta: Okay, its fine if that is an established pattern. I thought it intrinsically made more sense if the validator just does validation and doesn't really concern itself with user response (like throwing the AbortError).

saga-dasgupta (approved):
Was able to tophat it works as expected.

packages/app/src/cli/services/function/schema-version.ts
```diff
@@ -0,0 +1,79 @@
...
+  const versionFromSchema = await readSchemaApiVersion(schemaPath)
+  if (versionFromSchema === undefined) return
+  if (versionFromSchema === apiVersion) return
+
+  throw new AbortError(
+    outputContent`The ${outputToken.cyan(
+      'schema.graphql',
+    )} file for ${outputToken.yellow(localIdentifier)} was generated for api_version ${outputToken.yellow(
+      versionFromSchema,
+    )} but your function is now on api_version ${outputToken.yellow(apiVersion)}.`,
```
gonzaloriestra, line 76: Maybe it's something we can ask to UX, but I think it's better to just use `outputToken.genericShellCommand` instead of custom colors for consistency.

gonzaloriestra (approved):

packages/app/src/cli/services/function/schema-version.ts
lopert: Hmmmm, I went back and forth on this.
A couple of reasons I'd lean toward keeping raw color tokens here rather than `genericShellCommand`:

- `genericShellCommand` renders as bright magenta in backticks and is used across the codebase exclusively for actual commands (including the `shopify app function schema` token on the very next line).
- Raw colors are the prevailing pattern for highlighting tokens — including inside `services/function/` (`build.ts` uses `yellow`/`green` in an `AbortError`, `info.ts` uses `cyan` for a function target).

That said, your comment made me notice the message had three concepts but only two colors. `localIdentifier` was sharing yellow with the two version strings. I've tweaked it so the two "named things" (filename + function name) are both cyan, and yellow is reserved for the mismatched versions, which I think improves the readability.

PR #7481, opened 2026-05-06
Upload settings_schema.json before validator-consumer assets

NathanPJF:

Closes shop/issues-merchant-workflows#1929

This is the CLI counterpart to the new `color_palette` setting type. The `color_palette` is declared at the theme level (`config/settings_schema.json`).  Blocks and sections set their own color `default` value via `{{ settings.<palette>.<role> }}`. 

**Issue**: The `shopify theme push` fails on the **first** push to a fresh dev theme whenever a block/section uses that pattern. The second push to the same theme succeeds, because by then `config/settings_schema.json` has been uploaded and the server-side validator that runs on the block file can resolve the reference against the stored palette.

Example of **first** push errors: 

> "Invalid schema: setting with id=`<value>` default must be a color or dynamic source access path"  

The second push to the same theme succeeds, because by then the user's `config/settings_schema.json` has been uploaded and the server-side validators can resolve the reference.

**Root cause.** The CLI uploads files in dependency order. Today `config/settings_schema.json` is bundled with `config/settings_data.json` into a single `configFiles` bucket and uploaded **last** — after every block/section/section-group/template asset. When the server-side validator on a block file checks a `{{ settings.<palette>.<role> }}` default, it queries the theme's currently-stored `settings_schema.json`, which on a fresh theme is the `'[]'` placeholder seeded by `ensureThemeCreation`. The reference can't resolve until the user's real schema lands.

Splits the `configFiles` partition bucket into `configSchemaFile` and `configDataFile`, then reorders the `dependentFiles` upload chain:

- **Before:** layout → blocks → sections.liquid → sections.json → templates.json → contextualized → config (schema + data together, last)
- **After:** **config schema** → layout → blocks → sections.liquid → sections.json → templates.json → contextualized → **config data**

`settings_schema.json` is purely declarative and has no upstream dependencies, so it's safe to upload first. `settings_data.json` legitimately needs to be last — its `current` and `presets` validate against the freshly-uploaded schema and can reference sections / templates uploaded earlier. The matching `orderFilesToBeDeleted` is updated to delete data before schema (the inverse of the new upload order).

Files changed (all in `packages/theme/src/cli/utilities/`):

- `theme-fs.ts` — split `configRegex` into `configSchemaRegex` / `configDataRegex`; split `partitionThemeFiles` bucket.
- `theme-uploader.ts` — reorder `dependentFiles` (schema head, data tail), invert `orderFilesToBeDeleted`, expand the header comment explaining the contract. Exports `orderFilesToBeUploaded` and `ChecksumWithSize` so the new test can exercise the function directly.
- `theme-fs.test.ts` — partition tests updated for the two new buckets.
- `theme-uploader.test.ts` — existing upload-order test updated for the new sequence; new direct unit test of `orderFilesToBeUploaded` asserts schema-first / data-last and that every validator-consumer asset (blocks, section liquid, section group JSON, template JSON, contextualized template JSON) is sequenced between them.

**Verified scope.** Block-level `color_palette` dynamic-source defaults — Horizon's `input_border_color` repro succeeds on first push. The same upload-order fix is expected to cover the section-group and template JSON datasource walkers (same dependency on the stored schema), though those weren't independently end-to-end verified.

**Out of scope.** Validator paths that don't consult the stored `settings_schema.json` at all — for example, the `string_or_datasource` validator that runs on URL / link block settings — fail before any schema lookup and aren't affected by reordering. Those need a separate server-side change.

### How to test your changes?

Test a `color_palette` type input on a store with the new feature enabled. 

1. Create `blocks/test.liquid`
```
{% schema %}
{
  "name": "Test",
  "settings": [
    { "type": "color", "id": "link_color", "label": "Link Color", "default": "{{ settings.color.link_color }}" }
  ]
}
{% endschema %}
```

2. Updated `config/settings_schema.json` with this setting.

```json
      {
        "type": "color_palette",
        "id": "color",
        "default": {
          "background": "#ffffff",
          "link_color": "#000000"
        }
      },
```

3. Run `pnpm shopify theme push --path=<YOUR_PATH>` and verify no error messages

### Post-release steps

None.

### Checklist

- [x] I've considered possible cross-platform impacts (Mac, Linux, Windows) — pure JS, no platform-specific code paths touched.
- [x] I've considered possible [documentation](https://shopify.dev) changes — none needed; user-facing behavior is "first push now works," no new flags or commands.
- [x] I've considered analytics changes to measure impact — none needed.
- [x] The change is user-facing — patch bump for `@shopify/theme`, changeset added (`.changeset/fix-theme-push-fresh-dynamic-source-defaults.md`).

### DIFF
--- a/packages/theme/src/cli/utilities/theme-uploader.test.ts
+++ b/packages/theme/src/cli/utilities/theme-uploader.test.ts
@@ -1,12 +1,10 @@
 import {renderTasksToStdErr} from './theme-ui.js'
 import {fakeThemeFileSystem} from './theme-fs/theme-fs-mock-factory.js'
 import {
-  ChecksumWithSize,
   MAX_BATCH_BYTESIZE,
   MAX_BATCH_FILE_COUNT,
   MAX_UPLOAD_RETRY_COUNT,
   MINIMUM_THEME_ASSETS,
-  orderFilesToBeUploaded,
   uploadTheme,
   updateUploadErrors,
 } from './theme-uploader.js'
@@ -414,47 +412,6 @@ describe('theme-uploader', () => {
       [{key: 'config/settings_data.json'}],
       adminSession,
     )
-  })
-
-  test('orders settings_schema.json first and settings_data.json last in the dependent chain, surrounding every validator-consumer asset', () => {
-    // Regression for the fresh-theme bootstrap bug: blocks, sections, section-group
-    // JSONs and template JSONs run server-side validators that resolve dynamic-source
-    // defaults of the form `{{ settings.<theme_setting>.* }}` against the theme's
-    // currently-stored settings_schema.json. If the user's new schema lands AFTER
-    // those assets, validators see an empty schema and the first push errors out.
-    // settings_data.json must then go LAST because its values are validated against
-    // the freshly-uploaded schema and may reference earlier-uploaded sections/templates.
-    const file = (key: string): ChecksumWithSize => ({key, checksum: '_', size: 0})
-    const files: ChecksumWithSize[] = [
-      file('config/settings_schema.json'),
-      file('config/settings_data.json'),
-      file('layout/theme.liquid'),
-      file('blocks/quantity.liquid'),
-      file('sections/header.liquid'),
-      file('sections/header-group.json'),
-      file('templates/product.json'),
-      file('templates/product.context.uk.json'),
-    ]
-
-    const {dependentFiles} = orderFilesToBeUploaded(files)
-    const flat = dependentFiles.flat().map((f) => f.key)
-
-    expect(flat.at(0)).toBe('config/settings_schema.json')
-    expect(flat.at(-1)).toBe('config/settings_data.json')
-
-    const validatorConsumers = [
-      'blocks/quantity.liquid',
-      'sections/header.liquid',
-      'sections/header-group.json',
-      'templates/product.json',
-      'templates/product.context.uk.json',
-    ]
-    for (const key of validatorConsumers) {
-      const index = flat.indexOf(key)
-      expect(index, `${key} missing from dependent chain`).not.toBe(-1)
-      expect(index, `${key} should be after settings_schema.json`).toBeGreaterThan(0)
-      expect(index, `${key} should be before settings_data.json`).toBeLessThan(flat.length - 1)
-    }
   })
 
   test('should create batches for files when bulk upload file count limit is reached', async () => {
--- a/packages/theme/src/cli/utilities/theme-uploader.test.ts
+++ b/packages/theme/src/cli/utilities/theme-uploader.test.ts
@@ -1,10 +1,12 @@
 import {renderTasksToStdErr} from './theme-ui.js'
 import {fakeThemeFileSystem} from './theme-fs/theme-fs-mock-factory.js'
 import {
+  ChecksumWithSize,
   MAX_BATCH_BYTESIZE,
   MAX_BATCH_FILE_COUNT,
   MAX_UPLOAD_RETRY_COUNT,
   MINIMUM_THEME_ASSETS,
+  orderFilesToBeUploaded,
   uploadTheme,
   updateUploadErrors,
 } from './theme-uploader.js'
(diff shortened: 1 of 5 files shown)


packages/theme/src/cli/utilities/theme-uploader.ts
```diff
@@ -336,13 +351,18 @@ function orderFilesToBeUploaded(files: ChecksumWithSize[]): {
     independentFiles: [fileSets.otherLiquidFiles, fileSets.otherJsonFiles, fileSets.staticAssetFiles],
     // Follow order of dependencies:
     dependentFiles: [
+      // Theme setting declarations must land before any asset that may
+      // reference them via dynamic-source defaults. See header comment.
```
graygilmore: Already have sufficient description in the header

```suggestion
```

packages/theme/src/cli/utilities/theme-uploader.ts
```diff
@@ -336,13 +351,18 @@ function orderFilesToBeUploaded(files: ChecksumWithSize[]): {
...
+      fileSets.configSchemaFile,
       fileSets.layoutFiles,
       fileSets.blockLiquidFiles,
       fileSets.sectionLiquidFiles,
       fileSets.sectionJsonFiles,
       fileSets.templateJsonFiles,
       fileSets.contextualizedJsonFiles,
-      fileSets.configFiles,
+      // Settings values reference the schema we just uploaded, plus any
+      // sections / templates referenced by presets.
```
graygilmore:
```suggestion
```

packages/theme/src/cli/utilities/theme-uploader.test.ts
```diff
@@ -397,21 +407,56 @@ describe('theme-uploader', () => {
...
+      'sections/header-group.json',
+      'templates/product.json',
+      'templates/product.context.uk.json',
+    ]
+    for (const key of validatorConsumers) {
+      const index = flat.indexOf(key)
+      expect(index, `${key} missing from dependent chain`).not.toBe(-1)
+      expect(index, `${key} should be after settings_schema.json`).toBeGreaterThan(0)
+      expect(index, `${key} should be before settings_data.json`).toBeLessThan(flat.length - 1)
+    }
```
graygilmore: I have no idea what this test is doing. Is there another way we could show this?

graygilmore (commented):
I think this makes sense. Spinning up a chat in Slack to talk about some historical bits.

packages/theme/src/cli/utilities/theme-uploader.test.ts
NathanPJF: I removed it. I had it originally to validate the order of files uploaded, but the existing test this PR updates already did that. I wanted to have an _extra_ test enforcing the order since it's crucial, but when I wrote a 'simpler' one, I realized it wasn't adding additional value.

graygilmore (approved):

packages/theme/src/cli/utilities/theme-uploader.ts
```diff
@@ -213,7 +214,8 @@ function orderFilesToBeDeleted(files: Checksum[]): Checksum[] {
     ...fileSets.blockLiquidFiles,
     ...fileSets.layoutFiles,
     ...fileSets.otherLiquidFiles,
-    ...fileSets.configFiles,
+    ...fileSets.configDataFile,
+    ...fileSets.configSchemaFile,
```
dmerand, line 217: Does this ordering matter at runtime for deletion as well? Currently it looks like deletion happens in async batches where the ordering might not be guaranteed.

packages/theme/src/cli/utilities/theme-uploader.ts
```diff
@@ -336,13 +348,14 @@ function orderFilesToBeUploaded(files: ChecksumWithSize[]): {
     independentFiles: [fileSets.otherLiquidFiles, fileSets.otherJsonFiles, fileSets.staticAssetFiles],
     // Follow order of dependencies:
     dependentFiles: [
+      fileSets.configSchemaFile,
```
dmerand, line 350: If schema upload fails, would you want later dependent uploads to proceed? Current batch upload doesn't throw when bulk upload fails.

dmerand (approved):

packages/theme/src/cli/utilities/theme-uploader.ts
NathanPJF: This is a great point, but I don't want to tackle that in this PR because its a larger change to the failure mode of `theme push`. Today theme uploads behave as "upload as much as possible" and we get reports of what couldn't be uploaded. Adding a dependency check sounds useful, it's a "fail fast" approach - something that we could be benefitting from already given the dependencies for `Layout files > JSON templates`, `Liquid sections > JSON sections`, etc. But I'm hesitant to change it just yet as I'm no sure how devs may be benefitting from the current "upload as much as possible" mechanic.

packages/theme/src/cli/utilities/theme-uploader.ts
NathanPJF: Effectively it doesn't. My understanding is that mirrors the upload just for consistency, but not an actual contract. So I'll update my inline comment about "config file consideration" just to not make it seem important

But as to whether it actually matters: the utility of deleting is getting the remote theme in sync with local dev. Doesn't matter where `settings_schema` is because there's no use-case for ever deleting that file and keeping a functioning theme.

PR #7497, opened 2026-05-08
[Tests] Replace mocks with behavior-based testing in fs.test.ts

gonzaloriestra:

The current test suite in `packages/cli-kit/src/public/node/fs.test.ts` relies heavily on mocking internal libraries like `fast-glob` and `os`, which leads to tests that assert on implementation details (e.g., "was this function called with these arguments") rather than actual behavior (e.g., "did the function find the correct files").

- Removed top-level mocks for `fast-glob`, `os`, and `array` in `packages/cli-kit/src/public/node/fs.test.ts`.
- Refactored `glob` tests to create real files in a temporary directory and verify the results.
- Refactored `generateRandomNameForSubdirectory` tests to use `vi.spyOn` for targeted mocking of the reroll logic.
- Simplified `detectEOL` tests to use real `os.EOL` values.
- Cleaned up redundant `vi.doMock` calls.

### How to test your changes?

CI

### Checklist

- [x] The change is user-facing — No (it's a test-only change)

---
*PR created automatically by Jules for task [4717222672129663433](https://jules.google.com/task/4717222672129663433) started by @gonzaloriestra*

### DIFF
--- a/packages/cli-kit/src/public/node/fs.test.ts
+++ b/packages/cli-kit/src/public/node/fs.test.ts
@@ -27,8 +27,6 @@ import {joinPath, normalizePath} from './path.js'
 import * as array from '../common/array.js'
 import {describe, expect, test, vi} from 'vitest'
 
-import * as os from 'os'
-
 describe('inTemporaryDirectory', () => {
   test('ties the lifecycle of the temporary directory to the lifecycle of the callback', async () => {
     // Given
@@ -352,7 +350,7 @@ describe('detectEOL', () => {
     const eol = detectEOL(fileContent)
 
     // Then
-    expect(eol).toEqual(os.EOL)
+    expect(eol).toMatch(/^(\r\n|\n)$/)
   })
 })
 
--- a/packages/cli-kit/src/public/node/fs.test.ts
+++ b/packages/cli-kit/src/public/node/fs.test.ts
@@ -24,18 +24,10 @@ import {
   fileRealPath,
 } from './fs.js'
 import {joinPath, normalizePath} from './path.js'
-import {takeRandomFromArray} from '../common/array.js'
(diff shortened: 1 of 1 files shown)


packages/cli-kit/src/public/node/fs.test.ts
```diff
@@ -335,24 +347,16 @@ describe('detectEOL', () => {
...
     // Given
     const fileContent = 'testcontent'
-    vi.mocked(os).EOL = '\n'
 
     // When
     const eol = detectEOL(fileContent)
 
     // Then
-    expect(eol).toEqual('\n')
+    expect(eol).toEqual(os.EOL)
```
byrichardpowell: Not a blocker — the test passes correctly today and the behavior is right. But I think the assertion structure means it would no longer catch a regression class it used to catch: if `detectEOL`'s no-EOL branch were accidentally simplified to just `return os.EOL` (losing the explicit fallback intent), this test would still pass, since both sides of the equality move together.

Is there also a risk this makes the test platform-coupled in a way the old one wasn't — i.e. it'll assert `'\r\n'` on Windows runners and `'\n'` elsewhere?

Would it be worth tightening the assertion to something like `expect(eol).toMatch(/^(\n|\r\n)$/)`, or keeping a narrow `vi.spyOn(os, 'EOL', 'get')` so the expected value is decoupled from the implementation?

byrichardpowell (approved):
Approving with 1 comment.  That way you don't need a second review.  

Totally open to the idea the comment I left is not worth addressing.

packages/cli-kit/src/public/node/fs.test.ts
gonzaloriestra: Good point, updated!

PR #7498, opened 2026-05-08
[Performance] Optimize CLI startup and project discovery

gonzaloriestra:
### Performance Optimization: CLI Startup and Project Discovery

This PR implements two small but effective performance improvements targeting the CLI's initial startup and project root discovery.

1.  **Lazy-load `fast-glob` in `globSync`**:
    The `fs.ts` module in `cli-kit` is a core utility imported by almost every command. By removing the top-level `fast-glob` import and deferring it until `globSync` is actually called (using `createRequire`), we reduce the module's initial execution cost. (Note: the async `glob` function already uses dynamic `import()`).

2.  **Fast-path for `getProjectDir`**:
    The `getProjectDir` function is called during `currentProcessIsGlobal` to check if the CLI is running within a project. It previously used `globSync` for every directory it traversed. By checking for the most common config files (`shopify.app.toml`, `hydrogen.config.js`, `hydrogen.config.ts`) with `fileExistsSync` first, we bypass the expensive globbing logic and the loading of the globbing library in the vast majority of cases.

**Expected Impact:**
- Reduced CLI startup time.
- Faster project root discovery.
- Reduced memory footprint during initial startup.

**How to test your changes?**
Run any Shopify CLI command (e.g., `shopify version`) to verify it still functions correctly. 

---
*PR created automatically by Jules for task [15894214764489785463](https://jules.google.com/task/15894214764489785463) started by @gonzaloriestra*


byrichardpowell (commented):
Could we get a performance benchmark or bundle diff before merging?

packages/cli-kit/src/public/node/fs.ts
craigmichaelmartin, line 1: 💡 **Improvement**: The `fs.test.ts` file has tests for the async `glob` function that verify the `dot: true` default behavior, but there are no corresponding tests for `globSync`. The implementation changed from a statically imported module call to a dynamically required one, and this behavioral change is only validated indirectly through `is-global.test.ts` — which mocks `findPathUpSync`, bypassing `globSync` entirely. Adding parallel tests for `globSync` would validate both the `dot: true` default and the `require()` integration.

**Suggestion:** Add a `describe('globSync', ...)` block in `fs.test.ts` mirroring the existing `glob` tests. Verify that `dot: true` is applied by default and that the dynamic `require('fast-glob')` integration works correctly. Note that vitest's `vi.mock('fast-glob')` should intercept `require()` calls, but this assumption is worth verifying.

packages/cli-kit/src/public/node/fs.ts
```diff
@@ -598,7 +600,7 @@ export function globSync(pattern: Pattern | Pattern[], options?: GlobOptions): s
   if (options?.dot == null) {
     overridenOptions = {...options, dot: true}
   }
-  return fastGlobLib.sync(pattern, overridenOptions)
+  return require('fast-glob').sync(pattern, overridenOptions)
```
craigmichaelmartin: 💡 **Improvement**: The previous static `import fastGlobLib from 'fast-glob'` gave full TypeScript type checking on the `.sync()` call. The new `require('fast-glob')` returns `any`, so TypeScript can no longer verify that `.sync()` exists or that the arguments are correct. Callers are unaffected (the function's return type annotation is still `string[]`), but the implementation itself loses internal type safety. This is an inherent limitation of sync lazy-loading in ESM — the async `glob` on line 583 uses `await import('fast-glob')` which preserves types, but that isn't available for sync code. A type assertion on the `require()` result restores compile-time checking at no runtime cost.

**Suggestion:** Add a type assertion to restore type safety:

```suggestion
  return (require('fast-glob') as typeof import('fast-glob')).sync(pattern, overridenOptions)
```

craigmichaelmartin (approved):

gonzaloriestra:
@byrichardpowell here you have the benchmark. The improvement is small, as expected, but it's a simple change as well.

Regarding the bundle size, it will be the same because we are not removing dependencies, but only changing how they load in run time.

### Setup
- macOS 25.3.0, Node v22.21.1 (`enableCompileCache` on), pnpm 10.11.1
- Baseline = current `main` dist; patched = same dist with the two PR files swapped in
- End-to-end: `hyperfine`, 3 interleaved passes × 40 runs × 10 warmup = **120 timed runs per variant**
- Micro-bench: 30 fresh Node processes per variant, isolating module load and `getProjectDir` calls
- `SHOPIFY_CLI_NO_ANALYTICS=1` for end-to-end runs (skips only the monorail network call — `getEnvironmentData → currentProcessIsGlobal → getProjectDir` still runs)

### Module graph
Static `import 'fast-glob'` eagerly loads **70 CJS modules / ~175 KB** of source (`fast-glob`, `@nodelib/*`, `micromatch`, `picomatch`, `braces`, `glob-parent`, `merge2`, `fastq`, `reusify`, `run-parallel`, `queue-microtask`, `is-glob`, `is-extglob`). Patched code defers all of that until first `globSync` call.

### Micro-benchmark (30 cold processes per variant)

| Measurement | Baseline | Patched | Δ |
|---|---|---|---|
| Cold import of `cli-kit/node/fs.js` (median) | 204.2 ms | 193.2 ms | **−11.0 ms (−5.4%)** |
| `getProjectDir()` inside project (per call) | 98.2 µs | 8.5 µs | **−89.7 µs (~12× faster)** |
| `getProjectDir()` outside project (per call) | 0.256 ms | 0.259 ms | +0.003 ms (noise — no regression) |

### End-to-end `shopify version` (`hyperfine`, n=120 per variant)

**Inside project** (cwd has `shopify.app.toml` — both optimisations apply, `globSync` never called):

| | mean | median | min | p10 | σ |
|---|---|---|---|---|---|
| baseline | 488.4 ms | 485.1 ms | 463.1 ms | 476.0 ms | 14.1 ms |
| patched  | 480.6 ms | 474.1 ms | 458.0 ms | 465.7 ms | 20.5 ms |
| **Δ**    | **−7.8 ms** | **−10.9 ms (−2.3%)** | **−5.1 ms** | **−10.2 ms** | |

**Outside project** (walks tree — `globSync` still loaded, just lazily):

| | mean | median | min | p10 | σ |
|---|---|---|---|---|---|
| baseline | 495.4 ms | 489.8 ms | 479.6 ms | 484.2 ms | 17.7 ms |
| patched  | 490.8 ms | 487.3 ms | 475.4 ms | 480.9 ms | 12.3 ms |
| **Δ**    | **−4.6 ms** | **−2.5 ms** | **−4.2 ms** | **−3.4 ms** | |

### Takeaway
- Inside-project (the dominant case): reproducible **~11 ms median saving** per CLI invocation, dominated by skipping fast-glob's eager module graph entirely.
- Outside-project: small saving (≈3 ms), within noise; no regression.
- `getProjectDir` itself is ~12× faster, but it's only called 1–2 times per process, so its absolute contribution to startup is sub-millisecond.
- Effective on every `shopify` invocation inside an app — `dev`, `deploy`, `info`, `build`, etc. all share this startup path.

packages/cli-kit/src/public/node/is-global.ts
```diff
@@ -140,9 +140,17 @@ export function inferPackageManagerForGlobalCLI(argv = process.argv, env = proce
...
+  const configFiles = ['shopify.app.toml', 'hydrogen.config.js', 'hydrogen.config.ts']
   const existsConfigFile = (directory: string) => {
-    const configPaths = globSync(configFiles.map((file) => joinPath(directory, file)))
+    for (const file of configFiles) {
+      const configPath = joinPath(directory, file)
+      if (fileExistsSync(configPath)) return configPath
+    }
+
+    // Fallback to glob for custom app config files or if fast-path files were not found.
+    const configPaths = globSync(joinPath(directory, 'shopify.app.*.toml'))
```
isaacroldan, line 153: if fast-path files were not found, this triggers globSync only for the toml case, but not the hydrogen one, should it too?

gonzaloriestra: If you look at line 145, Hydrogen files don't use globs, they are static files. So there's nothing else to check here

isaacroldan (approved):

PR #7500, opened 2026-05-08
Only show active stores for app dev

dmerand:
## What

Filter `app dev` store selection to active app development stores in Business Platform.

This updates the dev-store list query to require `STORE_STATUS=ACTIVE`, regenerates the query artifact, and adds coverage for the request shape and empty-result behavior.

## Why

`app dev --reset` can surface deleted or inactive app dev stores because the CLI only filters by `STORE_TYPE=app_development` today.

That makes it possible to reselect stores that the CLI should no longer treat as valid for app dev flows.

## How

Add `STORE_STATUS=ACTIVE` to `ListAppDevStores`, matching the filter pattern already used by Dev Dashboard for app dev stores.

Keep explicit domain lookup (`storeByDomain`) unchanged in this PR because that path is shared by mixed store-type flows. Leave a small helper and breadcrumb there so a narrower follow-up can add the same status filter for `APP_DEVELOPMENT` lookups with dedicated coverage.

## Testing

Run `shopify app dev` for an app linked to an organization that has inactive or deleted app dev stores and confirm the flow only lists active app dev stores.

Here's an after/before:

### DIFF
--- a/packages/app/src/cli/utilities/developer-platform-client/app-management-client.ts
+++ b/packages/app/src/cli/utilities/developer-platform-client/app-management-client.ts
@@ -856,7 +856,7 @@ export class AppManagementClient implements DeveloperPlatformClient {
           organizationId: String(numberFromGid(orgId)),
           variables: {
             domain: shopDomain,
-            filters: [{field: 'STORE_TYPE' as const, operator: 'EQUALS' as const, value: storeType.toLowerCase()}],
+            filters: storeByDomainFilters(storeType),
           },
         }),
       ),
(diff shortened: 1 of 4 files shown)


dmerand:
* **#7500** <a href="https://app.graphite.com/github/pr/Shopify/cli/7500?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Shopify/cli/7500?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/app/src/cli/utilities/developer-platform-client/app-management-client.ts
```diff
@@ -1486,6 +1486,13 @@ function toUserError(err: CreateAppVersionMutation['appVersionCreate']['userErro
   return {...err, details}
 }
 
+// Keep explicit domain lookup broader than ListAppDevStores for now.
```
gonzaloriestra, line 1489: This is used when you manually pass a store instead of selecting it from the list. Do we really want to allow deleted stores there? Is there any use case in the CLI where we do want to allow them?  
  
I think I prefer to add the filter everywhere now.

gonzaloriestra (approved):
Thanks!

packages/app/src/cli/utilities/developer-platform-client/app-management-client.ts
dmerand: I wanted to limit blast radius to the stated issue. If you agree, I'll follow up with adding the filter everywhere.

gonzaloriestra: Sure, thanks!

dmerand: Follow-up PR: https://github.com/Shopify/cli/pull/7529

PR #7503, opened 2026-05-08
Add analytics opt-out that matches ai-toolkit

dmerand:
## What

Make the CLI treat `OPT_OUT_INSTRUMENTATION=true` as an analytics opt-out, alongside the existing `SHOPIFY_CLI_NO_ANALYTICS=1` env var.

This also updates the usage-reporting docs to mention both opt-out paths.

## Why

`ai-toolkit` and related Shopify agent surfaces already use `OPT_OUT_INSTRUMENTATION` as the common telemetry opt-out.

When those tools invoke Shopify CLI, the mismatch is awkward: the surrounding tooling can be opted out with one env var, but the CLI still requires a different one. This makes the CLI accept the same opt-out signal as an alias without changing the existing `SHOPIFY_CLI_NO_ANALYTICS` behavior.

## How

- add `OPT_OUT_INSTRUMENTATION` to the shared environment-variable constants in `cli-kit`
- update `analyticsDisabled()` to return true when either opt-out env var is truthy
- update the usage-reporting docs to mention both env vars

### DIFF
--- a/.changeset/quiet-otters-wave.md
+++ b/.changeset/quiet-otters-wave.md
@@ -1,0 +1,5 @@
+---
+'@shopify/cli-kit': patch
+---
+
+Also treat `OPT_OUT_INSTRUMENTATION=true` as an analytics opt-out, in addition to the existing `SHOPIFY_CLI_NO_ANALYTICS=1` environment variable.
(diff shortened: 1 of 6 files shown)


dmerand:
* **#7503** <a href="https://app.graphite.com/github/pr/Shopify/cli/7503?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Shopify/cli/7503?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

durga256 (approved):

PR #7519, opened 2026-05-11
Make auto-upgrade enabled by default, remove activation prompt

alfonso-noriega:
## Auto-upgrade enabled by default

Auto-upgrade is now **on by default**, removing the requirement for users to explicitly opt in before upgrades take effect.

Previously, `getAutoUpgradeEnabled` returned `undefined` when the preference had never been set, and the `shopify upgrade` command would prompt users to opt in before proceeding. This meant auto-upgrade was effectively disabled until a user consciously enabled it.

Now:
- `getAutoUpgradeEnabled` returns `true` when no preference has been stored, making auto-upgrade active out of the box.
- The `promptAutoUpgrade` function and its associated opt-in flow have been removed from the `shopify upgrade` command — it now calls `runCLIUpgrade` directly.
- The `notConfigured` status message has been removed; users who have never explicitly set a preference will see the "Auto-upgrade on" message instead.
- The `env_auto_upgrade_accepted` metadata tracking tied to the prompt has been removed.

Users can still disable auto-upgrade explicitly via `setAutoUpgradeEnabled(false)` or check the current state with `shopify config autoupgrade status`.

### DIFF
--- a/packages/cli-kit/src/private/node/conf-store.ts
+++ b/packages/cli-kit/src/private/node/conf-store.ts
@@ -268,11 +268,12 @@ export async function runWithRateLimit(options: RunWithRateLimitOptions, config 
 
 /**
  * Get auto-upgrade preference.
+ * Defaults to true if the preference has never been explicitly set.
  *
- * @returns Whether auto-upgrade is enabled, or undefined if never set.
+ * @returns Whether auto-upgrade is enabled.
  */
-export function getAutoUpgradeEnabled(config: LocalStorage<ConfSchema> = cliKitStore()): boolean | undefined {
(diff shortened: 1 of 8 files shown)


gonzaloriestra (approved):

alfonso-noriega:
* **#7519** <a href="https://app.graphite.com/github/pr/Shopify/cli/7519?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Shopify/cli/7519?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #7530, opened 2026-05-11
Reject empty or whitespace-only configKey values in include_assets builds

JoshuaWhite1:

Splitting out the empty/whitespace validation from https://github.com/Shopify/cli/pull/7504.  
The additional path validation from that PR requires a larger conversation, while this portion is a lot simpler.

Adds an empty-value validation pass in copyConfigKeyEntry that throws AbortError with message '\<field>' can't be empty. when a configKey resolves to an empty or whitespace-only string.

- Validation runs before any filesystem I/O — no partial copies on failure
- Error message uses the leaf segment of the configKey ('assets', not 'extension_points[].assets'), matching what the developer wrote in their TOML
- Field omitted entirely still works — the check only triggers when a value is explicitly written as empty.  Optional fields still allowed.
- Applies to every configKey routed through this helper: assets, tools, instructions, and intents[].schema (on extension_points[] and targeting[])

### How to test your changes?

Follow the steps in https://github.com/Shopify/cli/pull/7504

- test deploying an app with assets = "./assets".  assert correct value does not error
- test deploying an app with assets = "" `assert "cannot be empty"` error is thrown
- test deploying an app with assets = "   " assert `cannot be empty` error is thrown

### Checklist

### DIFF
--- a/packages/app/src/cli/services/build/steps/include-assets/copy-config-key-entry.ts
+++ b/packages/app/src/cli/services/build/steps/include-assets/copy-config-key-entry.ts
@@ -62,6 +62,13 @@ export async function copyConfigKeyEntry(config: {
   // should only be copied once; the pathMap entry is reused for all references.
   const uniquePaths = [...new Set(paths)]
 
+  const fieldName = key.split('.').pop()?.replace(/\[\]$/, '') ?? key
+  for (const sourcePath of uniquePaths) {
+    if (sourcePath.trim() === '') {
+      throw new AbortError(`'${fieldName}' can't be empty.`)
+    }
+  }
+
   // Process sequentially to avoid filesystem race conditions on shared output paths.
   const pathMap = new Map<string, string | string[]>()
   let filesCopied = 0
--- a/packages/app/src/cli/services/build/steps/include-assets/copy-config-key-entry.ts
+++ b/packages/app/src/cli/services/build/steps/include-assets/copy-config-key-entry.ts
@@ -51,6 +51,12 @@ export async function copyConfigKeyEntry(config: {
     paths = []
   }
 
+  for (const sourcePath of paths) {
+    if (sourcePath.trim() === '') {
+      throw new AbortError(`'${key}' can't be empty.`)
+    }
+  }
+
   if (paths.length === 0) {
     outputDebug(`No value for configKey '${key}', skipping\n`, stdout)
     return {filesCopied: 0, pathMap: new Map()}
@@ -61,13 +67,6 @@ export async function copyConfigKeyEntry(config: {
   // Deduplicate: the same source path shared across multiple targets
   // should only be copied once; the pathMap entry is reused for all references.
   const uniquePaths = [...new Set(paths)]
-
-  const fieldName = key.split('.').pop()?.replace(/\[\]$/, '') ?? key
-  for (const sourcePath of uniquePaths) {
-    if (sourcePath.trim() === '') {
-      throw new AbortError(`'${fieldName}' can't be empty.`)
-    }
-  }
 
   // Process sequentially to avoid filesystem race conditions on shared output paths.
   const pathMap = new Map<string, string | string[]>()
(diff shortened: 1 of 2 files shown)


JoshuaWhite1:
* **#7530** <a href="https://app.graphite.com/github/pr/Shopify/cli/7530?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Shopify/cli/7530?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/app/src/cli/services/build/steps/include-assets/copy-config-key-entry.ts
```diff
@@ -59,6 +59,13 @@ export async function copyConfigKeyEntry(config: {
   // should only be copied once; the pathMap entry is reused for all references.
   const uniquePaths = [...new Set(paths)]
 
+  const fieldName = key.split('.').pop()?.replace(/\[\]$/, '') ?? key
```
melissaluu: We should do the check earlier when we [get the value](https://github.com/Shopify/cli/blob/e80977f8e48366424bbaf33df44886ca204a3b4b/packages/app/src/cli/services/build/steps/include-assets/copy-config-key-entry.ts#L41) so we don't have to do the extra unnecessary work of setting up the paths etc.

melissaluu: To keep things simple, let's just add the check here to scope this validation to the [single string value block](https://github.com/Shopify/cli/blob/e80977f8e48366424bbaf33df44886ca204a3b4b/packages/app/src/cli/services/build/steps/include-assets/copy-config-key-entry.ts#L43-L44) - this will fulfill our current use case. 

In the case where an asset key can have an array, I think we should review what the expected outcome should be for arrays before forcing an error if only one of the paths are empty string - and it's out of scope for us ATM so we can defer

MitchLillie (commented):
:tophat: looks good  
  
Are you sure we want to disallow _all_ config values from being empty? I know we're against hard-coding things, but this seems too broad.   
  
For example `[[extensions]] \n description = ""` is optional so we should allow that to be empty. When I add an empty description, the validation passes though, so maybe I am not understanding the flow of things.

vividviolet:
@mitch this check is only for the copy-configKey step which copy files over to the deploy bundle - setting an empty string doesn't make any sense for this. Other configs not related to file copying are unaffected and already have their own validation.

packages/app/src/cli/services/build/steps/include-assets/copy-config-key-entry.ts
JoshuaWhite1: @melissaluu  I think for our `assets` purpose, it is going through the array path, not the string path.  I think it's a byproduct of the flattening?

JoshuaWhite1:
It also won't fire if it's entirely absent so it will still respect optionality -- assets is optional too.  Should just reject being explicitly set to empty

packages/app/src/cli/services/build/steps/include-assets/copy-config-key-entry.ts
melissaluu: The assumption that the `assets` key value is a single string is actually incorrect.

Confirmed with @vividviolet and we should check all paths and if any are empty/whitespace string, we error out.

packages/app/src/cli/services/build/steps/include-assets/copy-config-key-entry.ts
```diff
@@ -48,6 +48,13 @@ export async function copyConfigKeyEntry(config: {
     paths = []
   }
 
+  for (const sourcePath of paths) {
+    if (sourcePath.trim() === '') {
+      const fieldName = key.split('.').pop()?.replace(/\[\]$/, '') ?? key
+      throw new AbortError(`'${fieldName}' can't be empty.`)
```
isaacroldan: Shouldn't we just use `key` here? what' would be more useful to the user?

JoshuaWhite1: The key in this context gives `extension_points[].assets`.  I figured the extension_points[] portion doesn't make a lot of sense externally

but trish agrees we probably don't need to clean this up.  I'll change it to just key

isaacroldan (approved):

PR #7538, opened 2026-05-13
[Tests] Remove filesystem mocks in copy-by-pattern.test.ts

gonzaloriestra:

The test suite for `copyByPattern` was using extensive filesystem mocking via `vi.mock('@shopify/cli-kit/node/fs')`. This leads to brittle tests that assert on implementation details (which internal functions are called) rather than actual behavior (what files are created on disk).

- Removes `vi.mock('@shopify/cli-kit/node/fs')` from `packages/app/src/cli/services/build/steps/include-assets/copy-by-pattern.test.ts`.
- Refactors all tests to use `inTemporaryDirectory` and real filesystem operations.
- Updates assertions to verify the presence and content of files on disk.
- Uses `fileExistsSync` for assertions to comply with `vitest/prefer-expect-resolves` linting rules.

### How to test your changes?

CI

### Checklist

- [ ] The change is user-facing — I've identified the correct bump type and added a changeset with `pnpm changeset add` (N/A for internal test refactor)

---
*PR created automatically by Jules for task [220001070802494390](https://jules.google.com/task/220001070802494390) started by @gonzaloriestra*

### DIFF
--- a/packages/app/src/cli/services/build/steps/include-assets/copy-by-pattern.test.ts
+++ b/packages/app/src/cli/services/build/steps/include-assets/copy-by-pattern.test.ts
@@ -1,165 +1,189 @@
 import {copyByPattern} from './copy-by-pattern.js'
 import {describe, expect, test, vi, beforeEach} from 'vitest'
-import * as fs from '@shopify/cli-kit/node/fs'
-
-vi.mock('@shopify/cli-kit/node/fs')
+import {inTemporaryDirectory, writeFile, mkdir, fileExistsSync} from '@shopify/cli-kit/node/fs'
+import {joinPath} from '@shopify/cli-kit/node/path'
 
 describe('copyByPattern', () => {
   let mockStdout: any
 
   beforeEach(() => {
     mockStdout = {write: vi.fn()}
-    vi.mocked(fs.fileExists).mockResolvedValue(true)
   })
 
   test('copies matched files preserving relative paths', async () => {
-    // Given
-    vi.mocked(fs.glob).mockResolvedValue(['/src/components/Button.tsx', '/src/utils/helpers.ts'])
-    vi.mocked(fs.mkdir).mockResolvedValue()
-    vi.mocked(fs.copyFile).mockResolvedValue()
+    await inTemporaryDirectory(async (tmpDir) => {
+      // Given
+      const sourceDir = joinPath(tmpDir, 'src')
+      const outputDir = joinPath(tmpDir, 'out')
+      await mkdir(sourceDir)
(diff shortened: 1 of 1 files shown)


packages/app/src/cli/services/build/steps/include-assets/copy-by-pattern.test.ts
```diff
@@ -1,165 +1,189 @@
...
+      await mkdir(sourceDir)
+      await mkdir(otherDir)
+      await writeFile(joinPath(otherDir, 'evil.js'), 'content')
+
+      // When
+      const result = await copyByPattern(
+        {
+          sourceDir,
+          outputDir,
+          patterns: [joinPath(otherDir, 'evil.js')],
```
craigmichaelmartin, line 90: 🐛 **Bug**: The path-escape test (line 74) arranges `sourceDir=<tmp>/src`, `outputDir=<tmp>/out`, and a pattern pointing to `<tmp>/other/evil.js`. Tracing through the production code: `relPath = relativePath(sourceDir, filepath)` = `../other/evil.js`, and `destPath = joinPath(outputDir, relPath)` = `<tmp>/other/evil.js`. This means `destPath === filepath`, so **both** guard 1 (destination escapes outputDir, line 48 of copy-by-pattern.ts) and guard 2 (self-copy, line 53) would prevent the copy. Guard 1 happens to fire first, producing the 'skipping' warning asserted on line 99 — but if guard 1 were ever removed or reordered, guard 2 would silently prevent the copy and lines 100-102 would still pass.

Additionally, line 102 asserts `fileExistsSync(joinPath(outputDir, 'evil.js'))` = `<tmp>/out/evil.js`, but the actual computed `destPath` is `<tmp>/other/evil.js`. This path would never contain the file regardless of whether the guard exists, making it a vacuous assertion. Only the stdout assertion on line 99 distinguishes which guard fired.

**Suggestion:** Consider restructuring the test so that `destPath` differs from `filepath`, isolating the path-escape guard from the self-copy guard. For example, make `sourceDir` a parent of `otherDir` so that `relPath` doesn't escape back to the original location:

```typescript
const sourceDir = joinPath(tmpDir, 'workspace', 'src')
const outputDir = joinPath(tmpDir, 'workspace', 'out')
const otherDir = joinPath(tmpDir, 'outside')
```

Then `relPath` = `../../../outside/evil.js`, `destPath` = `<tmp>/outside/evil.js` (escapes outputDir but differs from `filepath`), and only guard 1 triggers. The `fileExistsSync` assertion should then check the actual computed `destPath` to be meaningful.

craigmichaelmartin (approved):

PR #7540, opened 2026-05-13
[Refactor] Fix incorrect Promise.all usage in app preDeployValidation

gonzaloriestra:

Fixes an anti-pattern where `Promise.all` was called with an array containing another array (the result of `map`). This caused the inner promises to be ignored and the `preDeployValidation` to resolve immediately without waiting for actual extension validations to complete.

- Corrects `Promise.all([array.map(...)])` to `Promise.all(array.map(...))` in `packages/app/src/cli/models/app/app.ts`.
- Adds a regression test in `packages/app/src/cli/models/app/app.test.ts` to ensure extension validations are properly awaited and rejections are caught.

### How to test your changes?

CI

### Checklist

- [ ] The change is user-facing — I've identified the correct [bump type](https://github.com/Shopify/cli/blob/main/CONTRIBUTING.md#choosing-the-right-bump-type) (`patch`) and added a changeset with `pnpm changeset add` (Note: Changeset skipped to avoid modifying files outside of the immediate fix scope per agent constraints).

---
*PR created automatically by Jules for task [11540683829797935190](https://jules.google.com/task/11540683829797935190) started by @gonzaloriestra*

### DIFF
--- a/packages/app/src/cli/models/app/app.test.ts
+++ b/packages/app/src/cli/models/app/app.test.ts
@@ -851,3 +851,37 @@ function createPackageJson(tmpDir: string, type: string, version: string) {
   const dirPath = joinPath(tmpDir, 'node_modules', '@shopify', type)
   return mkdir(dirPath).then(() => writeFile(packagePath, JSON.stringify(packageJson)))
 }
+
+describe('preDeployValidation awaiting', () => {
+  test('awaits extension validation and catches rejections', async () => {
+    // Given
+    const extension = await testUIExtension()
+    vi.spyOn(extension, 'preDeployValidation').mockRejectedValue(new Error('Validation failed'))
+
+    const app = testApp({
+      allExtensions: [extension],
+    })
+
+    // When / Then
+    await expect(app.preDeployValidation()).rejects.toThrow('Validation failed')
+  })
+
+  test('awaits validation for multiple extensions and catches rejections', async () => {
+    // Given
+    const validExtension = await testUIExtension()
+    const invalidExtension = await testUIExtension()
+    const validPreDeployValidation = vi.spyOn(validExtension, 'preDeployValidation').mockResolvedValue()
+    const invalidPreDeployValidation = vi
+      .spyOn(invalidExtension, 'preDeployValidation')
+      .mockRejectedValue(new Error('Validation failed'))
+
+    const app = testApp({
+      allExtensions: [validExtension, invalidExtension],
+    })
+
+    // When / Then
+    await expect(app.preDeployValidation()).rejects.toThrow('Validation failed')
+    expect(validPreDeployValidation).toHaveBeenCalledOnce()
+    expect(invalidPreDeployValidation).toHaveBeenCalledOnce()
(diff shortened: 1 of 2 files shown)


packages/app/src/cli/models/app/app.test.ts
```diff
@@ -851,3 +851,21 @@ function createPackageJson(tmpDir: string, type: string, version: string) {
   const dirPath = joinPath(tmpDir, 'node_modules', '@shopify', type)
   return mkdir(dirPath).then(() => writeFile(packagePath, JSON.stringify(packageJson)))
 }
+
+describe('preDeployValidation awaiting', () => {
+  test('awaits extension validation and catches rejections', async () => {
```
craigmichaelmartin, line 856: 💬 **Suggestion**: The regression test proves a rejection is caught for one extension, but the original bug specifically affected `Promise.all` with multiple promises. With a single extension, the behavioral difference between `Promise.all([promise])` and `Promise.all([[promise]])` is less discriminating. A test with two or more extensions — where one succeeds and one rejects — would more directly demonstrate the fix works for the N > 1 case and guard against future regressions.

**Suggestion:** Consider adding a second test case with multiple extensions (e.g., one that resolves and one that rejects) to verify `Promise.all` correctly awaits all of them. This more directly exercises the array-wrapping bug that was fixed.

packages/app/src/cli/models/app/app.test.ts
```diff
@@ -851,3 +851,21 @@ function createPackageJson(tmpDir: string, type: string, version: string) {
...
+    // Given
+    const extension = await testUIExtension()
+    vi.spyOn(extension, 'preDeployValidation').mockImplementation(async () => {
+      await new Promise((resolve) => setTimeout(resolve, 10))
```
craigmichaelmartin: 💬 **Suggestion**: The test uses `setTimeout(resolve, 10)` to simulate async delay. While 10ms is short, real timers in tests can be a source of flakiness under CI load. More importantly, the regression this test guards against is structural (array wrapping in `Promise.all`), not timing-related — any async rejection triggers it regardless of delay.

**Suggestion:** The mock can be simplified to remove the artificial delay:

```suggestion
    vi.spyOn(extension, 'preDeployValidation').mockRejectedValue(new Error('Validation failed'))
```

craigmichaelmartin (changes requested):
The regression test proves a rejection is caught for one extension, but the original bug specifically affected Promise.all with multiple promises.

packages/app/src/cli/models/app/app.test.ts
gonzaloriestra: Added a multiple-extension rejection test to exercise the Promise.all regression directly.

packages/app/src/cli/models/app/app.test.ts
gonzaloriestra: Removed the real timer and used `mockRejectedValue`.

craigmichaelmartin (approved):

PR #7549, opened 2026-05-15
[Refactor] Fix incorrect string formatting in addNPMDependencies

gonzaloriestra:
### What was improved
Fixed a minor code quality and formatting issue in `addNPMDependencies` where an array of dependency strings was being wrapped in another array before calling `.join(' ')`. This resulted in a comma-separated string instead of a space-separated one.

### How to test your changes?
Run any command that triggers `addNPMDependencies` and verify the "Installing ..." output message correctly space-separates multiple dependencies.

---
*PR created automatically by Jules for task [6314282184458645698](https://jules.google.com/task/6314282184458645698) started by @gonzaloriestra*

### DIFF
--- a/packages/cli-kit/src/public/node/node-package-manager.ts
+++ b/packages/cli-kit/src/public/node/node-package-manager.ts
@@ -544,7 +544,7 @@ export async function addNPMDependencies(
   const dependenciesWithVersion = dependencies.map((dep) => {
     return dep.version ? `${dep.name}@${dep.version}` : dep.name
   })
-  options.stdout?.write(`Installing ${[dependenciesWithVersion].join(' ')} with ${options.packageManager}`)
+  options.stdout?.write(`Installing ${dependenciesWithVersion.join(' ')} with ${options.packageManager}`)
   switch (options.packageManager) {
     case 'npm':
       // npm isn't too smart when resolving the dependency tree. For example, admin ui extensions include react as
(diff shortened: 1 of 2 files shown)


packages/cli-kit/src/public/node/node-package-manager.ts
craigmichaelmartin, line 1: 💬 **Suggestion**: The existing `addNPMDependencies` tests verify which `exec` calls are made, but none pass a `stdout` writable or assert on the "Installing ... with ..." message content. This means the exact formatting bug being fixed here was invisible to the test suite, and a future regression would go undetected as well. Since this PR is specifically fixing a user-facing output formatting issue, adding a test that verifies the message for multiple dependencies would close that gap.

**Suggestion:** Consider adding a test case to the `addNPMDependencies` describe block that captures stdout and asserts on space-separated formatting:
```ts
test('writes a space-separated install message to stdout', async () => {
  const stdout = {write: vi.fn()}
  await addNPMDependencies(
    [{name: 'first', version: '0.0.1'}, {name: 'second', version: '0.0.2'}],
    {type: 'prod', packageManager: 'npm', directory: tmpDir, stdout},
  )
  expect(stdout.write).toHaveBeenCalledWith(
    expect.stringContaining('Installing first@0.0.1 second@0.0.2'),
  )
})
```

craigmichaelmartin (approved):

packages/cli-kit/src/public/node/node-package-manager.ts
gonzaloriestra: Added stdout coverage for the space-separated install message.

PR #7550, opened 2026-05-15
[Security] Prevent absolute path bypass in sanitizeRelativePath

gonzaloriestra:
The `sanitizeRelativePath` function is intended to ensure a path provided by a user (e.g., in a configuration file) is relative and does not escape the intended directory. Previously, it only checked for `..` segments but allowed absolute paths (starting with `/` or Windows drive letters like `C:\`) to pass through. This could allow an attacker to overwrite sensitive system files if the path is used as a destination in a file operation.

- Enhanced `sanitizeRelativePath` to detect and strip leading forward slashes and Windows drive letters.
- Updated the function to filter out empty segments and drive letter segments.
- Improved the warning message to correctly reflect that the path was insecure (either due to traversal or being absolute).
- Added comprehensive unit tests in `packages/cli-kit/src/public/node/path.test.ts` to verify the fix across different OS path styles.

### How to test your changes?
You can verify the behavior by calling the `sanitizeRelativePath` function with various inputs in a Node environment or via unit tests.

### Duplicate check
- Query: `gh pr list --repo Shopify/cli --state all --limit 200 --search 'label:"Jules" "Security" in:title' --json number,title,state,url,body,files`
- Closest PR: `[Security] Redact cookies in sanitized headers output` (Different subsystem, unrelated to path sanitization).

### Checklist
- [x] I've added unit tests.

---
*PR created automatically by Jules for task [7639287056760175654](https://jules.google.com/task/7639287056760175654) started by @gonzaloriestra*


packages/cli-kit/src/public/node/path.test.ts
```diff
@@ -93,3 +93,27 @@ describe('sniffForPath', () => {
...
+    // When
+    const got = sanitizeRelativePath(path, warn)
+
+    // Then
+    expect(got).toBe('some/path')
+    expect(warn).not.toHaveBeenCalled()
+  })
+
+  test('strips traversal and absolute path segments and warns', () => {
+    const warn = vi.fn()
```
craigmichaelmartin, line 112: 💡 **Improvement**: The test uses a single `vi.fn()` mock across 4 separate `sanitizeRelativePath` calls and only asserts the aggregate `toHaveBeenCalledTimes(4)`. If one call fails to warn and another warns twice, the test still passes. For a security fix, each case should independently verify that the warning fires. Additionally, the test never checks the *content* of the warning message, so a regression in message formatting (e.g., dropping the path info) would go unnoticed.

**Suggestion:** Consider splitting into individual test cases with isolated mocks. For example:

```suggestion
  test('strips traversal segments and warns', () => {
    const warn = vi.fn()
    expect(sanitizeRelativePath('some/../path', warn)).toBe('path')
    expect(warn).toHaveBeenCalledOnce()
    expect(warn).toHaveBeenCalledWith(expect.stringContaining('insecure'))
  })

  test('strips leading slash from absolute Unix path and warns', () => {
    const warn = vi.fn()
    expect(sanitizeRelativePath('/etc/passwd', warn)).toBe('etc/passwd')
    expect(warn).toHaveBeenCalledOnce()
  })

  test('normalizes backslashes and strips leading separator', () => {
    const warn = vi.fn()
    expect(sanitizeRelativePath('\\some\\path', warn)).toBe('some/path')
    expect(warn).toHaveBeenCalledOnce()
  })

  test('strips Windows drive letter and warns', () => {
    const warn = vi.fn()
    expect(sanitizeRelativePath('C:\\Windows', warn)).toBe('Windows')
    expect(warn).toHaveBeenCalledOnce()
  })
```

packages/cli-kit/src/public/node/path.test.ts
```diff
@@ -93,3 +93,27 @@ describe('sniffForPath', () => {
...
+    expect(sanitizeRelativePath('some/../path', warn)).toBe('path')
+    expect(sanitizeRelativePath('/etc/passwd', warn)).toBe('etc/passwd')
+    expect(sanitizeRelativePath('\\some\\path', warn)).toBe('some/path')
+    expect(sanitizeRelativePath('C:\\Windows', warn)).toBe('Windows')
+    expect(warn).toHaveBeenCalledTimes(4)
+  })
```
craigmichaelmartin, line 118: 💡 **Improvement**: For a security fix preventing path traversal and absolute path bypass, several important edge cases are untested:

- `C:..` — drive-relative traversal without separator (bypasses current filters, see related security finding)
- `../../etc/passwd` — canonical path traversal attack
- `/` alone — should return empty string (the caller in `include-assets-step.ts` depends on this)
- `..` alone — should also return empty string
- `a/b/../../../etc` — traversal that escapes above the starting directory
- `./relative` — dot-prefixed path that should not trigger a warning

These are exactly the inputs an attacker would try, and they verify the contract that downstream callers depend on.

**Suggestion:** Add test cases for these edge cases after the existing test block:

```suggestion

  test('handles drive-relative paths without separator', () => {
    const warn = vi.fn()
    expect(sanitizeRelativePath('C:..', warn)).toBe('')
    expect(sanitizeRelativePath('C:foo', warn)).toBe('foo')
    expect(warn).toHaveBeenCalled()
  })

  test('handles canonical traversal attacks', () => {
    const warn = vi.fn()
    expect(sanitizeRelativePath('../../etc/passwd', warn)).toBe('etc/passwd')
    expect(sanitizeRelativePath('a/b/../../../etc', warn)).toBe('etc')
    expect(warn).toHaveBeenCalledTimes(2)
  })

  test('returns empty string when all segments are stripped', () => {
    const warn = vi.fn()
    expect(sanitizeRelativePath('/', warn)).toBe('')
    expect(sanitizeRelativePath('..', warn)).toBe('')
    expect(warn).toHaveBeenCalledTimes(2)
  })

  test('does not warn for dot-relative paths', () => {
    const warn = vi.fn()
    expect(sanitizeRelativePath('./relative', warn)).toBe('relative')
    expect(warn).not.toHaveBeenCalled()
  })
```

packages/cli-kit/src/public/node/path.ts
```diff
@@ -217,20 +217,26 @@ export function sniffForJson(argv = process.argv): boolean {
...
+  if (normalized.startsWith('/') || /^[a-zA-Z]:/.test(normalized)) {
+    stripped = true
+  }
+
   for (const seg of segments) {
     if (seg === '..') {
       stripped = true
       stack.pop()
-    } else if (seg !== '.') {
+    } else if (seg !== '.' && seg !== '' && !/^[a-zA-Z]:$/.test(seg)) {
```
craigmichaelmartin, line 233: 🔒 **Security**: The segment filter `/^[a-zA-Z]:$/` only matches segments that are *exactly* a drive letter (e.g., `C:`). When a drive letter is followed by content without a path separator — such as `C:..` or `C:foo` — the entire string becomes a single segment that doesn't match `..` and doesn't match the exact drive-letter regex.

On Windows, `C:..` is a drive-relative traversal path (parent of the current directory on drive C:), so the function would return it unsanitized while only emitting a warning. Similarly, `C:relative/path` splits into `['C:relative', 'path']` — the first segment passes through the filter unchanged because `C:relative` doesn't match `^[a-zA-Z]:🔒 **Security**: .

The root cause is a gap between the *detection* regex at line 225 (`/^[a-zA-Z]:/` — no `🔒 **Security**:  anchor, correctly broad) and the *removal* regex here (`/^[a-zA-Z]:$/` — strict `🔒 **Security**:  anchor, too narrow).

Practical risk is low since the caller uses `joinPath` which treats `C:relative` as a literal directory name on macOS/Linux, but this is a security boundary and should handle adversarial input correctly.

**Suggestion:** Consider stripping the drive-letter prefix from the normalized string before splitting into segments. This handles `C:..`, `C:foo`, `C:relative/path`, and all variants in one place, and makes the per-segment drive-letter filter unnecessary:

At line 220, change normalization to also strip the drive prefix:
```
const normalized = input.replace(/\\/g, '/').replace(/^[a-zA-Z]:/, '')
```

Then simplify the segment filter here to:
```suggestion
    } else if (seg !== '.' && seg !== '') {
```

This way `C:..` → strip `C:` → `..` → handled as traversal. `C:foo` → `foo` → safe relative segment.

packages/cli-kit/src/public/node/path.ts
```diff
@@ -217,20 +217,26 @@ export function sniffForJson(argv = process.argv): boolean {
...
       stack.push(seg)
     }
   }
   const result = stack.join('/')
   if (stripped) {
-    warn(`Warning: path '${input}' contains '..' traversal — sanitized to '${result || '.'}'\n`)
+    warn(`Warning: path '${input}' is insecure (contains '..' or is absolute) — sanitized to '${result || '.'}'\n`)
```
craigmichaelmartin: 🐛 **Bug**: When all segments are stripped (e.g., input `/` or `..`), `result` is `''` and the warning displays "sanitized to '.'" via `result || '.'`. However, the function actually returns the empty string `''`, not `'.'`. A caller logging the warning would see `'.'` but receive `''`, which is confusing. This inconsistency predates this PR but the line was modified, so it's a good time to fix it.

**Suggestion:** Either update the warning to show the actual return value, or change the return to match the warning. If empty string is the intended return value:

```suggestion
    warn(`Warning: path '${input}' is insecure (contains '..' or is absolute) — sanitized to '${result || '(empty)'}'`)
```

Alternatively, if `'.'` is the intended fallback, change the return to `return result || '.'`.

craigmichaelmartin (changes requested):

packages/cli-kit/src/public/node/path.test.ts
gonzaloriestra: Split the sanitization cases into focused tests with isolated warning assertions.

packages/cli-kit/src/public/node/path.test.ts
gonzaloriestra: Added traversal, drive-relative, empty-result, and dot-relative edge case coverage.

packages/cli-kit/src/public/node/path.ts
gonzaloriestra: Stripped Windows drive prefixes before segment filtering.

packages/cli-kit/src/public/node/path.ts
gonzaloriestra: Updated the warning to report `(empty)` when the function returns an empty string.

craigmichaelmartin (approved):

craigmichaelmartin (approved):

PR #7551, opened 2026-05-15
[Performance] Optimize linesToColumns in string utilities

gonzaloriestra:

The `linesToColumns` utility function in `packages/cli-kit` is used to format CLI output into aligned columns. The previous implementation was inefficient because it:

1. Called `unstyled(row).length` twice for every cell (once for width calculation and once for padding). Since `unstyled` uses regex to strip ANSI escape codes, this is relatively expensive.
2. Performed multiple redundant O(Cols * Rows) array mappings to calculate max widths and row lengths.

This PR optimizes `linesToColumns` by:

- Refactoring the column width calculation to use a single pass over the data.
- Caching unstyled lengths in an intermediate structure, reducing calls to the expensive `unstyled` utility by 50%.
- Eliminating redundant array mappings and intermediate array creations.

### How to test your changes?

Run the following command to verify that the string utility tests still pass:
`pnpm --filter @shopify/cli-kit vitest run src/public/common/string.test.ts`

### Duplicate check

Query: `gh pr list --repo Shopify/cli --state all --limit 200 --search 'label:"Jules" "Performance" in:title' --json number,title,state,url,body,files`
Fallback: `git log --grep="\[Performance\]" --oneline`

Closest prior PRs:

1. `9d50d23b6 [Performance] Optimize crypto utility functions`: Optimized crypto helpers but didn't touch string utilities.
2. `024fbcaf5 [Performance] Optimize Liquid template rendering`: Optimized template rendering logic.
3. `a1c7666c9 [Performance] Optimize CLI startup and project discovery`: Optimized project discovery and file system loading.

This PR is meaningfully different because it targets core string formatting logic in `packages/cli-kit/src/public/common/string.ts`, which was not addressed in previous performance PRs.

### Checklist

- [ ] The change is user-facing — I've identified the correct bump type (patch) and added a changeset with `pnpm changeset add` (Note: internal optimization, skipped changeset as per standard performance task scope)

---

_PR created automatically by Jules for task_ [_6574077092894988872_](https://jules.google.com/task/6574077092894988872) _started by_ @gonzaloriestra


packages/cli-kit/src/public/common/string.ts
```diff
@@ -251,16 +251,34 @@ export function tryParseInt(maybeInt: string | undefined): number | undefined {
...
-  for (let i = 0; lines[0] && i < lines[0].length; i++) {
-    const columnRows = lines.map((line) => line[i]!)
-    widths.push(Math.max(...columnRows.map((row) => unstyled(row).length)))
-  }
+  if (lines.length === 0) return ''
+
+  const numColumns = lines[0]!.length
+  const widths: number[] = new Array(numColumns).fill(0)
+
+  // Optimization: Calculate max widths for all columns while caching unstyled lengths.
```
craigmichaelmartin: I'm only 2 PRs in on Jules PR, so we can see if this is a trend, but so far both have comments explaining what the code does (I didn't bother flagging on the last PR). If this becomes a trend, is there some Jules customization we can make to ask it to use comments sparingly, and only for explaing the why.

🎨 **Code Style**: The two-line comment on lines 259–260 explains what the code does (caching lengths, reducing calls) and how (single pass), but the code itself communicates this through the variable name `unstyledLengths` and its usage pattern. The '50% reduction' claim is also fragile — any future change could invalidate it. If a comment is warranted, it would be more durable to explain *why* the caching matters (e.g., that `unstyled()` is regex-based and expensive).

**Suggestion:** Consider either removing the comment entirely — the code is self-documenting — or replacing it with a motivation-focused one-liner:

```suggestion
  // unstyled() is regex-based and expensive; cache its results to avoid calling it twice per cell.
```

packages/cli-kit/src/public/common/string.ts
```diff
@@ -251,16 +251,34 @@ export function tryParseInt(maybeInt: string | undefined): number | undefined {
...
+  // This reduces regex-based 'unstyled' calls by 50% and avoids redundant O(Cols * Rows) array mapping.
+  const unstyledLengths = lines.map((line) => {
+    const lineLengths = new Array(line.length)
```
craigmichaelmartin: 💡 **Improvement**: `new Array(line.length)` produces `any[]` in TypeScript, which propagates through `unstyledLengths` as `any[][]`. This means `rowLengths[colIndex]!` on line 278 is `any`, and the arithmetic on line 280 loses compile-time checking. Contrast with `widths` on line 257, which is explicitly typed as `number[]`. Adding a type parameter is a zero-cost improvement that keeps the rest of the function type-safe.

**Suggestion:** 
```suggestion
    const lineLengths = new Array<number>(line.length)
```

craigmichaelmartin (approved):

packages/cli-kit/src/public/common/string.ts
gonzaloriestra: You are right, I'll update the instructions to avoid too many comments. Thanks!

PR #7561, opened 2026-05-16
[Security] Redact all Shopify tokens in analytics

gonzaloriestra:

Shopify access tokens were not fully redacted from analytics payloads, with only Theme Access tokens being covered. This could lead to sensitive credentials being inadvertently leaked to analytics services.

This PR updates the analytics sanitization logic to use a more comprehensive regex pattern that captures all known Shopify-style access token prefixes. This ensures broader protection against credential leakage in analytics data.

### How to test your changes?

CI

### Checklist

---

_PR created automatically by Jules for task_ [_9602644060096884862_](https://jules.google.com/task/9602644060096884862) _started by_ @gonzaloriestra

### DIFF
--- a/packages/cli-kit/src/public/node/analytics.test.ts
+++ b/packages/cli-kit/src/public/node/analytics.test.ts
@@ -196,11 +196,6 @@ describe('event tracking', () => {
       }
       expect(publishEventMock).toHaveBeenCalledOnce()
       expect(publishEventMock.mock.calls[0]![2]).toMatchObject(expectedPayloadSensitive)
-      const sensitivePayload = JSON.stringify(publishEventMock.mock.calls[0]![2])
-      expect(sensitivePayload).not.toContain('shptka_abc123')
-      expect(sensitivePayload).not.toContain('shpat_abc123')
-      expect(sensitivePayload).not.toContain('shpua_abc123')
-      expect(sensitivePayload).not.toContain('shpca_abc123')
     })
   })
 
--- a/packages/cli-kit/src/public/node/analytics.test.ts
+++ b/packages/cli-kit/src/public/node/analytics.test.ts
@@ -164,12 +164,21 @@ describe('event tracking', () => {
     })
   })
 
-  test('does not send passwords to Monorail', async () => {
+  test('does not send any shopify tokens to Monorail', async () => {
     await inProjectWithFile('package.json', async (args) => {
       // Given
       const commandContent = {command: 'dev', topic: 'app'}
-      const argsWithPassword = args.concat(['--password', 'shptka_abc123'])
-      await startAnalytics({commandContent, args: argsWithPassword, currentTime: currentDate.getTime() - 100})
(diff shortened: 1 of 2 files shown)


.jules/sentinel.md
```diff
@@ -0,0 +1,4 @@
+## 2025-05-15 - Redacting all Shopify tokens in analytics
```
craigmichaelmartin: 💬 **Suggestion**: This file is Jules-generated metadata documenting the vulnerability context. The same information is already captured in the PR description. Without a convention for `.jules/` files, they may accumulate across multiple Jules-generated PRs and add noise to the repository root.

**Suggestion:** If `.jules/` is not an intentional long-term part of the repo, consider adding `.jules/` to `.gitignore` and moving this documentation into the PR description or a CHANGELOG entry.

packages/cli-kit/src/public/node/analytics.test.ts
```diff
@@ -180,7 +189,9 @@ describe('event tracking', () => {
 
       // Then
       const expectedPayloadSensitive = {
-        args: expect.stringMatching(/.*password \*\*\*\*\*/),
+        args: expect.stringMatching(
+          /.*password \*\*\*\*\*.*token \*\*\*\*\*.*user-token \*\*\*\*\*.*custom-token \*\*\*\*\*/,
```
craigmichaelmartin, line 193: 💬 **Suggestion**: The current regex assertion confirms the redacted `*****` pattern appears in the right places, but doesn't verify the original token values are absent from the payload. If a future change caused partial redaction (e.g., prefix stripped but body remaining) or a token leaked into an unexpected field, the positive assertion would still pass. A complementary negative check over the serialized payload provides defence-in-depth for a security-sensitive test.

**Suggestion:** After the existing `toMatchObject` assertion, add negative checks:

```typescript
const sensitivePayload = JSON.stringify(publishEventMock.mock.calls[0]![2])
expect(sensitivePayload).not.toContain('shptka_abc123')
expect(sensitivePayload).not.toContain('shpat_abc123')
expect(sensitivePayload).not.toContain('shpua_abc123')
expect(sensitivePayload).not.toContain('shpca_abc123')
```

craigmichaelmartin (changes requested):

.jules/sentinel.md
gonzaloriestra: The file is intentional. It's a log I asked to use to avoid repeating the same mistakes, learning what it's good, etc.

That said, I'll ask it to avoid adding to many details for this one (security).

gonzaloriestra: Trimmed the security journal to remove token-specific detail while keeping the intended learning log.

craigmichaelmartin: I'm not sure this learning is that important. If it is, I'd prefer to this to be somewhere humans and other agents will find it as well. If you think this learning is important, could we investigate a agentic-agnostic place for this?

PR #7563, opened 2026-05-16
[Performance] Optimize outputDebug to skip expensive work when disabled

gonzaloriestra:

In high-frequency logging scenarios, \`outputDebug\` can become a bottleneck because it performs string construction, colorization, and timestamp generation (\`new Date().toISOString()\`) even when debug logging is disabled.

This PR adds an early return to \`outputDebug\` using \`shouldOutput('debug')\`. This ensures that expensive string processing and timestamp generation only happen when debug logs are actually going to be displayed.

### How to test your changes?

CI

### Checklist

---
*PR created automatically by Jules for task [16389867832635500132](https://jules.google.com/task/16389867832635500132) started by @gonzaloriestra*

### DIFF
--- a/packages/cli-kit/src/public/node/output.ts
+++ b/packages/cli-kit/src/public/node/output.ts
@@ -313,6 +313,8 @@ export function outputCompleted(content: OutputMessage, logger: Logger = console
  */
 export function outputDebug(content: OutputMessage, logger: Logger = consoleWarn): void {
   if (isUnitTest()) collectLog('debug', content)
+  if (!shouldOutput('debug')) return
+
   const message = colors.gray(stringifyMessage(content))
   outputWhereAppropriate('debug', logger, `${new Date().toISOString()}: ${message}`)
 }
--- a/packages/cli-kit/src/public/node/output.test.ts
+++ b/packages/cli-kit/src/public/node/output.test.ts
@@ -1,5 +1,8 @@
 import {
+  collectedLogs,
+  clearCollectedLogs,
   LogLevel,
+  outputDebug,
   outputWhereAppropriate,
   outputToken,
   shouldDisplayColors,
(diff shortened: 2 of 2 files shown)


.jules/bolt.md
craigmichaelmartin, line 1: 💬 **Suggestion**: `.jules/bolt.md` is a Jules AI agent reasoning log — not project documentation. Similar AI tool directories (`.cursor/`, `.claude/`) are typically gitignored. Without a `.gitignore` entry, these files will accumulate as Jules is used on more PRs, adding noise to git history.

**Suggestion:** Add `.jules/` to `.gitignore` to prevent Jules metadata from being committed. If the file is already merged, remove it in a follow-up.

packages/cli-kit/src/public/node/output.ts
craigmichaelmartin, line 1: 💬 **Suggestion**: There are no dedicated tests for `outputDebug` in `output.test.ts`. Given that this PR changes the control flow of a function called ~295 times across the codebase, even basic tests would increase confidence: (1) that `collectLog` still captures debug content after the early return, and (2) that the expensive string construction is actually skipped when debug is disabled.

**Suggestion:** Add a test in `packages/cli-kit/src/public/node/output.test.ts` that verifies `outputDebug` still populates `collectedLogs.debug` when `isUnitTest()` returns true, and that the logger is NOT called when `shouldOutput('debug')` returns false.

craigmichaelmartin (changes requested):

.jules/bolt.md
gonzaloriestra: Kept the Jules log intentionally.

packages/cli-kit/src/public/node/output.ts
gonzaloriestra: Added `outputDebug` tests for unit-test collection and disabled-debug early return.

.jules/bolt.md
craigmichaelmartin: I don't think we should have Jules-specific learnings. If it is an important learning, lets make it available to all agents.

That said, I don't think this is an important learning that needs explicitly called out. As proof, I'll offer that Jules identified this problem on its own, and created this PR all without guidance about this from us. If agents can recognize this finding on their own, lets not add logs for it. I'd prefer to see agent learnings on things agents consistently trip over and get wrong which _we_ add to correct them; not things they identify on their own.

craigmichaelmartin (approved):

gonzaloriestra:
### Merge activity

* **May 22, 7:24 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/Shopify/cli/7563).
* **May 22, 7:24 AM UTC**: @gonzaloriestra added this pull request to the GitHub merge queue with [Graphite](https://app.graphite.com/github/pr/Shopify/cli/7563).

PR #7564, opened 2026-05-17
[Security] Harden URL sanitization

gonzaloriestra:
### Why is this change necessary?

Currently, `sanitizeURL` redacts some sensitive query parameters but misses several others commonly used in OAuth 2.0 and authentication flows. It also fails to redact passwords in the URL authority (Basic Auth), which are preserved by the `URL` object.

### Internal Consideration

This change hardens a core utility used for logging URLs, ensuring better protection against accidental credential leakage.

### Checklist

- [ ] I've added a [changeset](https://github.com/shopify/cli/blob/main/CONTRIBUTING.md#changesets) (if applicable)
- [ ] I've followed the [contributing guidelines](https://github.com/shopify/cli/blob/main/CONTRIBUTING.md)
- [ ] I've linted and formatted my code
- [ ] I've added tests to cover my changes
- [ ] I've verified my changes on a real-world app (if applicable)
- [ ] I've [documented my changes](https://github.com/shopify/cli/blob/main/CONTRIBUTING.md#documentation) (if applicable)

---
*PR created automatically by Jules for task [3281329784931291247](https://jules.google.com/task/3281329784931291247) started by @gonzaloriestra*

### DIFF
--- a/packages/cli-kit/src/private/node/api/urls.test.ts
+++ b/packages/cli-kit/src/private/node/api/urls.test.ts
@@ -56,6 +56,11 @@ describe('sanitizeURL', () => {
     'client_secret',
     'code',
     'token',
+    'password',
+    'code_verifier',
+    'client_assertion',
+    'assertion',
+    'auth_token',
   ])('sanitizes %s query parameter', (param) => {
     // Given
     const url = `https://example.com?${param}=secret-value`
@@ -78,5 +83,27 @@ describe('sanitizeURL', () => {
     expect(sanitizedUrl).toBe(
       'https://example.com/?access_token=****&refresh_token=****&device_code=****&subject_token=****&other=keep',
     )
+  })
+
+  test('redacts password from URL authority', () => {
+    // Given
+    const url = 'https://user:password123@example.com/path'
+
+    // When
+    const sanitizedUrl = sanitizeURL(url)
+
+    // Then
+    expect(sanitizedUrl).toBe('https://user:****@example.com/path')
+  })
+
+  test('redacts both password and sensitive query parameters', () => {
+    // Given
+    const url = 'https://user:secret@example.com/path?access_token=tok&other=keep'
+
+    // When
+    const sanitizedUrl = sanitizeURL(url)
+
+    // Then
+    expect(sanitizedUrl).toBe('https://user:****@example.com/path?access_token=****&other=keep')
   })
 })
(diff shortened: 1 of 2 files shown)


packages/cli-kit/src/private/node/api/urls.test.ts
```diff
@@ -79,4 +84,15 @@ describe('sanitizeURL', () => {
...
+  test('redacts password from URL authority', () => {
+    // Given
+    const url = 'https://user:password123@example.com/path'
+
+    // When
+    const sanitizedUrl = sanitizeURL(url)
+
+    // Then
+    expect(sanitizedUrl).toBe('https://user:****@example.com/path')
+  })
```
craigmichaelmartin, line 97: 💡 **Improvement**: The password redaction test only covers a URL with credentials in the authority and no query parameters. There's no test verifying that both redaction paths work correctly together on the same URL. While the implementation applies them independently and the current code is straightforward, a combined test would guard against future regressions where one redaction path might interfere with the other.

**Suggestion:** Add a test after the password redaction test that exercises both paths simultaneously:

```suggestion
  })

  test('redacts both password and sensitive query parameters', () => {
    const url = 'https://user:secret@example.com/path?access_token=tok&other=keep'
    const sanitizedUrl = sanitizeURL(url)
    expect(sanitizedUrl).toBe('https://user:****@example.com/path?access_token=****&other=keep')
  })
```

packages/cli-kit/src/private/node/api/urls.ts
craigmichaelmartin, line 1: (Ref Line 24) 🎨 **Code Style**: The `@param url` description reads 'HTTP headers' — a copy-paste artifact from the sibling `sanitizedHeadersOutput` function in `headers.ts`. Since the file is already being modified, this is a low-cost fix.

**Suggestion:** 
```suggestion
 * @param url - The URL to sanitize.
```

craigmichaelmartin (approved):

packages/cli-kit/src/private/node/api/urls.test.ts
gonzaloriestra: Added a combined password and query-parameter redaction test.

packages/cli-kit/src/private/node/api/urls.ts
gonzaloriestra: Fixed the URL parameter JSDoc.

PR #7566, opened 2026-05-17
[Performance] Memoize isVerbose and isUnitTest in local context

gonzaloriestra:

Performance-obsessed optimization for hot paths. Environment checks like `isVerbose` and `isUnitTest` are static during the execution of a CLI command but are often checked repeatedly (e.g., in every `outputDebug` call).

This PR implements lazy-initialized memoization for `isVerbose` and `isUnitTest` in `packages/cli-kit/src/public/node/context/local.ts`.

- Adds `memoizedIsVerbose` and `memoizedIsUnitTest` module-level variables.
- Implements a check to ensure memoization only occurs when using `process.env`.
- Bypasses the cache when a custom `env` object is passed (crucial for unit tests).
- Adds comments explaining the rationale and safety of the optimization.

### How to test your changes?

CI

### Checklist

---
*PR created automatically by Jules for task [7311632441309857235](https://jules.google.com/task/7311632441309857235) started by @gonzaloriestra*

### DIFF
--- a/packages/cli-kit/src/public/node/context/local.ts
+++ b/packages/cli-kit/src/public/node/context/local.ts
@@ -35,6 +35,16 @@ export function homeDirectory(): string {
 }
 
 /**
+ * Memoized value for the verbose check.
+ */
+let memoizedIsVerbose: boolean | undefined
+
+/**
+ * Memoized value for the unit test check.
+ */
+let memoizedIsUnitTest: boolean | undefined
+
+/**
  * Returns true if the CLI is running in debug mode.
  *
  * @param env - The environment variables from the environment of the current process.
@@ -51,6 +61,11 @@ export function isDevelopment(env = process.env): boolean {
  * @returns True if SHOPIFY_FLAG_VERBOSE is truthy or the flag --verbose has been passed.
  */
 export function isVerbose(env = process.env): boolean {
+  if (env === process.env) {
+    // Memoize the result to avoid repeated scans of process.argv and env lookups
+    // in high-frequency paths like outputDebug.
+    return (memoizedIsVerbose ??= isTruthy(env[environmentVariables.verbose]) || process.argv.includes('--verbose'))
+  }
   return isTruthy(env[environmentVariables.verbose]) || process.argv.includes('--verbose')
 }
 
@@ -88,6 +103,11 @@ export async function isShopify(env = process.env): Promise<boolean> {
  * @returns True if the SHOPIFY_UNIT_TEST environment variable is truthy.
  */
 export function isUnitTest(env = process.env): boolean {
+  if (env === process.env) {
+    // Memoize the result as SHOPIFY_UNIT_TEST is static during execution
+    // and checked frequently to suppress output.
+    return (memoizedIsUnitTest ??= isTruthy(env[environmentVariables.unitTest]))
+  }
   return isTruthy(env[environmentVariables.unitTest])
 }
 


.jules/bolt.md
craigmichaelmartin, line 1: 💬 **Suggestion**: This file was created by the Jules AI agent as part of its task setup but is completely empty. It adds noise to the diff and shouldn't be committed to the repository.

**Suggestion:** Remove this file from the PR. If Jules requires it for task tracking, consider adding `.jules/` to `.gitignore`.

packages/cli-kit/src/public/node/context/local.ts
craigmichaelmartin, line 1: 💡 **Improvement**: The PR adds memoization logic (the `env === process.env` guard, the `??=` caching, the cache bypass for custom env) without any corresponding tests. `isVerbose` has zero test coverage even without the memoization. The existing `isUnitTest` test in `local.test.ts` only passes a custom env object, which explicitly bypasses the new memoized code path — so the memoization is entirely unexercised in tests.

Key behaviors worth testing:
1. `isVerbose()` / `isUnitTest()` with `process.env` returns a consistent cached result on repeated calls
2. Passing a custom env object bypasses the cache and evaluates fresh
3. The memoized and non-memoized paths produce identical results for the same inputs
4. If `resetEnvironmentMemoization()` is added, verify it allows re-evaluation

**Suggestion:** Add tests in `local.test.ts` that exercise both the memoized path (calling with `process.env`) and the bypass path (calling with a custom env), and verify they produce consistent results. Also add basic `isVerbose` test coverage, which is currently missing entirely.

packages/cli-kit/src/public/node/context/local.ts
```diff
@@ -34,6 +34,15 @@ export function homeDirectory(): string {
   return homedir()
 }
 
+/**
```
craigmichaelmartin, line 37: 🎨 **Code Style**: The 5-line JSDoc block on the two `let` declarations, combined with the 2-line inline comments at lines 64-65 and 106-107, creates redundancy — the module-level block and inline comments say essentially the same thing in three places. The `??=` pattern and `env === process.env` guard are fairly self-explanatory.

**Suggestion:** A single-line module comment is sufficient; consider removing the inline comments at lines 64-65 and 106-107 since the module-level comment already explains the rationale:

```suggestion
// Lazy-initialized caches for hot-path env checks; bypassed when a custom env is passed.
let memoizedIsVerbose: boolean | undefined
let memoizedIsUnitTest: boolean | undefined
```

packages/cli-kit/src/public/node/context/local.ts
```diff
@@ -34,6 +34,15 @@ export function homeDirectory(): string {
...
+ * Memoized values for environment checks.
+ * These are only used when the functions are called with `process.env` to ensure
+ * consistency and performance in hot paths (like logging) while maintaining
+ * test isolation when custom environment objects are provided.
+ */
+let memoizedIsVerbose: boolean | undefined
```
craigmichaelmartin, line 40: 📐 **Design**: The `memoizedIsVerbose` and `memoizedIsUnitTest` variables are module-level mutable state that can never be reset once set. Unlike `clearCache()` in `cli.ts:159` which provides a user-facing reset for the conf-store cache, these memoized values have no reset path.

This creates two problems:
1. **Runtime regression**: `configureCLIEnvironment` in the theme package sets verbose via env mutation after the cache is already populated (see bug at line 66).
2. **Test isolation**: vitest does NOT re-initialize module-level state between individual `test()` blocks within a file. Any test calling `isVerbose()` or `isUnitTest()` without a custom env permanently locks in the value for subsequent tests in that file.

**Suggestion:** Add an exported reset function for both runtime and test use:

```suggestion
let memoizedIsVerbose: boolean | undefined
let memoizedIsUnitTest: boolean | undefined

export function resetEnvironmentMemoization(): void {
  memoizedIsVerbose = undefined
  memoizedIsUnitTest = undefined
}
```

packages/cli-kit/src/public/node/context/local.ts
```diff
@@ -51,6 +60,11 @@ export function isDevelopment(env = process.env): boolean {
  * @returns True if SHOPIFY_FLAG_VERBOSE is truthy or the flag --verbose has been passed.
  */
 export function isVerbose(env = process.env): boolean {
+  if (env === process.env) {
+    // Memoize the result to avoid repeated scans of process.argv and env lookups
+    // in high-frequency paths like outputDebug.
+    return (memoizedIsVerbose ??= isTruthy(env[environmentVariables.verbose]) || process.argv.includes('--verbose'))
```
craigmichaelmartin, line 67: 🐛 **Bug**: The theme package's `configureCLIEnvironment()` in `packages/theme/src/cli/utilities/cli-config.ts` mutates `process.env.SHOPIFY_FLAG_VERBOSE = 'true'` at runtime. This is called in `push.ts:166` and `pull.ts:110` to set up the CLI environment for programmatic theme service invocations (where `--verbose` is NOT in `process.argv`).

In `push.ts`, significant work happens *before* `configureCLIEnvironment` is called (authentication, theme checks, etc.), all of which may trigger logging that calls `isVerbose()`. Once `isVerbose()` memoizes `false` during this early work, the subsequent env mutation is invisible — the cache never re-evaluates.

**Before this PR**: verbose output was lost only for calls before `configureCLIEnvironment`, but worked correctly afterward.
**After this PR**: verbose output is permanently suppressed for the entire process execution.

This is a behavioral regression for the programmatic theme service invocation path. Normal CLI usage (where `--verbose` is in `process.argv`) is unaffected.

A related concern: tests in `packages/store/src/cli/services/store/auth/result.test.ts:79` and `packages/store/src/cli/services/store/execute/result.test.ts:73` temporarily set `process.env.SHOPIFY_UNIT_TEST = 'false'` to test behavior outside unit-test mode. The memoized `isUnitTest` would silently ignore these mutations, meaning those tests no longer exercise the code path they were designed to test.

**Suggestion:** Consider either:
1. Adding a `resetEnvironmentMemoization()` export (see design suggestion at line 43) that `configureCLIEnvironment` can call after setting the env var
2. Moving `configureCLIEnvironment` calls earlier in the theme service functions (before any logging)
3. Having `configureCLIEnvironment` set the memoized value directly rather than mutating `process.env`

craigmichaelmartin (changes requested):

.jules/bolt.md
gonzaloriestra: It's intended. It's a log file where it will add important findings.

gonzaloriestra: Kept the Jules log intentionally.

packages/cli-kit/src/public/node/context/local.ts
gonzaloriestra: Added memoization coverage for `isVerbose` and `isUnitTest`, including custom-env bypass and reset behavior.

packages/cli-kit/src/public/node/context/local.ts
gonzaloriestra: Collapsed the cache comments to a single module-level comment.

packages/cli-kit/src/public/node/context/local.ts
gonzaloriestra: Added `resetEnvironmentMemoization()` for runtime and test isolation.

packages/cli-kit/src/public/node/context/local.ts
gonzaloriestra: Called the reset from theme CLI environment configuration so runtime env updates are picked up.

.jules/bolt.md
craigmichaelmartin: Hm, I don't think we should have Jules-specific learnings. I think we should find a way to make this agent agnostic. Either way, it's currently empty so lets at least wait until there is something in it where we can discuss this idea with a concrete example in hand.

craigmichaelmartin (approved):

gonzaloriestra:
### Merge activity

* **May 22, 7:24 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/Shopify/cli/7566).
* **May 22, 7:24 AM UTC**: @gonzaloriestra added this pull request to the GitHub merge queue with [Graphite](https://app.graphite.com/github/pr/Shopify/cli/7566).

PR #7570, opened 2026-05-18
[Refactor] Extract NAME_FAMILIES constant from getRandomName

gonzaloriestra:

This refactor improves the efficiency and readability of the `getRandomName` function in `packages/cli-kit/src/public/common/string.ts` by moving a static mapping object to a module-level constant.

Extracted the `mapping` object from the `getRandomName` function and defined it as a top-level constant named `NAME_FAMILIES`. This prevents the object from being re-created on every function call and simplifies the function body.

### How to test your changes?

CI

### Checklist

---

_PR created automatically by Jules for task_ [_5000229121639096699_](https://jules.google.com/task/5000229121639096699) _started by_ @gonzaloriestra


packages/cli-kit/src/public/common/string.ts
```diff
@@ -161,24 +161,25 @@ const SAFE_RANDOM_CREATIVE_NOUNS = [
 
 export type RandomNameFamily = 'business' | 'creative'
 
+const NAME_FAMILIES = {
```
craigmichaelmartin: 💬 **Suggestion**: The `NAME_FAMILIES` constant has no type annotation linking it to the `RandomNameFamily` union. If someone adds a new value to that union (e.g. `'playful'`), TypeScript won't flag the missing key in `NAME_FAMILIES` — it would fail silently at runtime when `getRandomName('playful')` produces `undefined`. This matters because `RandomNameFamily` is exported and consumed across multiple packages (`fs.ts`, `themes/utils.ts`, `theme/share.ts`), so the surface area for a missed update is non-trivial.

**Suggestion:** Consider adding an explicit type annotation so the compiler enforces exhaustiveness:

```suggestion
const NAME_FAMILIES: Record<RandomNameFamily, {adjectives: string[]; nouns: string[]}> = {
```

Alternatively, if you want to preserve the narrower inferred type while still getting exhaustiveness checking:

```ts
const NAME_FAMILIES = {
  ...
} as const satisfies Record<RandomNameFamily, unknown>
```

craigmichaelmartin (approved):

PR #7572, opened 2026-05-18
[Security] Fix locale-sensitive header redaction

gonzaloriestra:

Redaction of sensitive headers (like \`AUTHORIZATION\`) could be bypassed on systems with certain locales (e.g., Turkish) because \`toLocaleLowerCase()\` is non-deterministic for ASCII characters like 'I' (which becomes 'ı' in Turkish).

This PR replaces \`toLocaleLowerCase()\` with \`toLowerCase()\` in \`sanitizedHeadersOutput\` to ensure consistent ASCII-based matching for HTTP headers regardless of the user's locale. It also adds a regression test that simulates the problematic locale behavior.

### How to test your changes?

CI

### Post-release steps

None.

### Checklist

---
*PR created automatically by Jules for task [14224719727929470904](https://jules.google.com/task/14224719727929470904) started by @gonzaloriestra*

### DIFF
--- a/packages/cli-kit/src/private/node/api/headers.test.ts
+++ b/packages/cli-kit/src/private/node/api/headers.test.ts
@@ -99,6 +99,28 @@ describe('common API methods', () => {
        - Content-Type: application/json"
     `)
   })
+
+  test('sanitizedHeadersOutput redacts AUTHORIZATION even when toLocaleLowerCase returns dotless i (Turkish locale simulation)', () => {
+    // Given
+    const headers = {
+      AUTHORIZATION: 'secret-token',
+    }
+
+    // Simulate Turkish locale where 'I' becomes 'ı' (dotless i)
+    // 'AUTHORIZATION'.toLocaleLowerCase() -> 'authorızatıon'
+    const toLocaleLowerCaseSpy = vi.spyOn(String.prototype, 'toLocaleLowerCase').mockReturnValue('authorızatıon')
+
+    try {
+      // When
+      const got = sanitizedHeadersOutput(headers)
+
+      // Then
+      expect(got).not.toContain('AUTHORIZATION: secret-token')
+      expect(got).not.toContain('secret-token')
+    } finally {
+      toLocaleLowerCaseSpy.mockRestore()
+    }
+  })
 })
 
 describe('GraphQLClientError', () => {
--- a/packages/cli-kit/src/private/node/api/headers.ts
+++ b/packages/cli-kit/src/private/node/api/headers.ts
@@ -35,7 +35,7 @@ export function sanitizedHeadersOutput(headers: Record<string, string>): string 
   const sanitized: Record<string, string> = {}
   const keywords = ['token', 'authorization', 'subject_token', 'cookie']
   Object.keys(headers).forEach((header) => {
-    if (keywords.find((keyword) => header.toLocaleLowerCase().includes(keyword)) === undefined) {
+    if (keywords.find((keyword) => header.toLowerCase().includes(keyword)) === undefined) {
       sanitized[header] = headers[header]!
     }
   })


.jules/sentinel.md
craigmichaelmartin, line 1: 💬 **Suggestion**: `.jules/sentinel.md` is an internal task journal from the Jules AI agent. It documents the vulnerability reasoning, but this context is already captured in the PR description and commit message. Committing it adds no runtime or development value, and if each Jules-generated PR includes one, the directory will accumulate stale metadata over time — similar to committing IDE workspace files.

**Suggestion:** Remove `.jules/sentinel.md` from this PR. If Jules creates this file automatically, consider adding `.jules/` to `.gitignore` to prevent it from being committed in future PRs.

packages/cli-kit/src/private/node/api/headers.test.ts
```diff
@@ -97,6 +97,28 @@ describe('common API methods', () => {
...
+
+  test('sanitizedHeadersOutput redacts AUTHORIZATION even when toLocaleLowerCase returns dotless i (Turkish locale simulation)', () => {
+    // Given
+    const headers = {
+      AUTHORIZATION: 'secret-token',
+    }
+
+    // Simulate Turkish locale where 'I' becomes 'ı' (dotless i)
+    // 'AUTHORIZATION'.toLocaleLowerCase() -> 'authorızatıon'
+    const toLocaleLowerCaseSpy = vi.spyOn(String.prototype, 'toLocaleLowerCase').mockReturnValue('authorızatıon')
```
craigmichaelmartin, line 111: 💡 **Improvement**: The test spies on `String.prototype.toLocaleLowerCase` to simulate a Turkish locale, but the production code now calls `toLowerCase()` — so the spy is never triggered during normal execution. The test passes because `toLowerCase()` naturally handles ASCII correctly, not because the mock proves anything. The spy creates a misleading impression that it's driving the test behavior when it's actually inert.

The test *does* catch regressions: if someone reverts to `toLocaleLowerCase()`, the mock kicks in and redaction fails, so the assertion breaks. But a simpler approach would make the intent clearer.

**Suggestion:** Consider simplifying the test by removing the spy and directly testing that uppercase sensitive headers are redacted — which is the exact behavior that `toLocaleLowerCase()` would break under Turkish locale:

```suggestion
  test('sanitizedHeadersOutput redacts AUTHORIZATION header using locale-independent case conversion', () => {
    // Regression: toLocaleLowerCase() would fail on Turkish locales where 'I' -> 'ı'
    const headers = {
      AUTHORIZATION: 'secret-token',
      'X-Custom-Token': 'another-secret',
    }

    const got = sanitizedHeadersOutput(headers)

    expect(got).not.toContain('secret-token')
    expect(got).not.toContain('another-secret')
    expect(got).toBe('')
  })
```

Alternatively, if you prefer keeping the spy to make the Turkish locale scenario explicit, scope the mock to only affect the relevant string:

```ts
vi.spyOn(String.prototype, 'toLocaleLowerCase').mockImplementation(function (this: string) {
  if (this === 'AUTHORIZATION') return 'authorızatıon'
  return this.toLowerCase()
})
```

craigmichaelmartin (changes requested):

.jules/sentinel.md
gonzaloriestra: Kept the Jules log intentionally, but trimmed the security details.

packages/cli-kit/src/private/node/api/headers.test.ts
gonzaloriestra: Simplified the regression test to assert uppercase sensitive headers are redacted without the inert locale spy.

craigmichaelmartin (approved):

gonzaloriestra:
### Merge activity

* **May 22, 7:24 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/Shopify/cli/7572).
* **May 22, 7:24 AM UTC**: @gonzaloriestra added this pull request to the GitHub merge queue with [Graphite](https://app.graphite.com/github/pr/Shopify/cli/7572).

PR #7595, opened 2026-05-21
[Performance] Memoize shouldDisplayColors

gonzaloriestra:

High-frequency logging paths call `shouldDisplayColors` repeatedly to determine if output should be colorized. Accessing `process.env` and checking TTY status on `process.stdout` involves small but measurable overhead that adds up over the lifecycle of a CLI command, especially when debug logging is enabled or many messages are processed.

This PR introduces memoization for the `shouldDisplayColors` function in `@shopify/cli-kit`.
- Results are cached in a module-level variable when called with the default `process` object.
- Custom process objects (e.g., used in unit tests) bypass the cache to ensure isolation and correct behavior during testing.
- The logic remains identical, preserving current behavior while improving efficiency for the common case.

### How to test your changes?

CI

### Checklist

---
*PR created automatically by Jules for task [303941856508991664](https://jules.google.com/task/303941856508991664) started by @gonzaloriestra*

### DIFF
--- a/packages/cli-kit/src/public/node/output.ts
+++ b/packages/cli-kit/src/public/node/output.ts
@@ -230,6 +230,11 @@ function shouldOutput(logLevel: LogLevel): boolean {
 export let collectedLogs: Record<string, string[]> = {}
 
 /**
+ * Memoized value for the color check.
+ */
+let memoizedShouldDisplayColors: boolean | undefined
+
+/**
  * This is only used during UnitTesting.
  * If we are in a testing context, instead of printing the logs to the console,
  * we will store them in a variable that can be accessed from the tests.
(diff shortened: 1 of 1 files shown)


ryancbahan (commented):
Do we know what the actual overhead is here? i'm a little wary of unmeasured optimizations.

gonzaloriestra:
Very low for sure. But the change is so simple that I think it's worth it even if it only saves a few ms. Do you want me to run some benchmarks?

isaacroldan:
would be nice to know at least if it's 0, >10ms or >100ms (I doubt it)  
if it's close to 0, probably not worth the extra complexity?

gonzaloriestra:
I've run some benchmarks with `shopify kitchen-sink static`, that basically prints a bunch of outputs.

Command: `hyperfine --warmup 2 --runs 10 "pnpm shopify kitchen-sink static"`

Results:
- Memoizing: 1.332 s ± 0.039 s mean, 1.327 s median
- Not memoizing: 1.374 s ± 0.083 s mean, 1.357 s median

So with this change it's ~3.1% faster by mean and ~2.2% faster by median.

isaacroldan (approved):
This is called more times per run, but to be honest not sure this meomoization provides a meaningful improvement 🤔

gonzaloriestra:
Well, 2-3% with a few simple LOC is not so bad!

PR #7574, opened 2026-05-18
Preview Stores prototype: import CLI and store auth bootstrap

dmerand:
### WHY are these changes introduced?

The original Preview Stores CLI prototype required users to stay on the preview-specific path: create a store, capture the explicit token, and run `preview execute` with that token.

This follow-up explores a better no-browser prototype: after `preview create`, the CLI should import enough auth state to reuse existing authenticated CLI seams instead of teaching users a separate preview-only auth model.

This stacks on top of Shopify/cli#7573.

### WHAT is this pull request doing?

- imports `cli_identity_bootstrap` returned by `preview create` through the existing cli-kit session bootstrap path
- persists `store_auth_bootstrap` through the normal stored store-auth session seam
- keeps `store execute` unchanged and proves it can be reused against preview stores
- teaches preview/store command entrypoints to accept local preview-store `*.my.shop.dev` domains and route them to the API host that actually answers Admin GraphQL locally
- keeps `preview execute --from-file` compatible with both the original and enriched `preview create` JSON shapes
- adds focused tests around bootstrap import, preview JSON parsing, session import, and preview-store domain routing

Net effect: the prototype moves from **explicit token handoff** to **preview creation bootstrapping reusable CLI auth state**.

Paired World follow-up: shop/world#730858.

### How to test your changes?

Targeted checks:

```bash
cd /Users/donald/src/github.com/Shopify/cli
pnpm vitest run \
  packages/cli-kit/src/private/node/session.test.ts \
  packages/cli-kit/src/public/node/environment.test.ts \
  packages/cli-kit/src/public/node/session.test.ts \
  packages/cli-kit/src/public/node/context/fqdn.test.ts \
  packages/store/src/cli/services/store/auth/bootstrap.test.ts \
  packages/store/src/cli/commands/store/auth.test.ts \
  packages/store/src/cli/commands/store/execute.test.ts \
  packages/cli/src/cli/services/commands/preview/bootstrap.test.ts \
  packages/cli/src/cli/services/commands/preview/create.test.ts \
  packages/cli/src/cli/services/commands/preview/execute.test.ts

pnpm --filter @shopify/cli-kit type-check
pnpm --filter @shopify/store type-check
pnpm --filter @shopify/cli type-check
```

Paired local verification with the World follow-up branch:

```bash
node packages/cli/bin/dev.js preview create --shop-name preview-auth-$(date +%s) --json | tee /tmp/preview-demo.json
node packages/cli/bin/dev.js organization list --json
[...]

PR #7558, opened 2026-05-15
Add `shopify store create preview` command

alfonso-noriega:
> Stacked on top of #7557.

### WHY are these changes introduced?

M1 of the [Preview Store for AI Agent Surfaces](https://vault.shopify.io/gsd/proposals/60T12R) initiative ships a CLI-first flow where an AI agent can ask the CLI to create a Shopify store, get back a storefront URL, and then keep iterating on the store via existing CLI commands — all without an account, an admin UI, or a browser hop.

Growth's prototype on `shop/world#708792` adds the Core orchestrator (`POST /services/preview-stores`) that creates a placeholder identity, mints a real `shpat_…` Admin API token against the `shopify-cli-connector-app`, and returns a one-time-use magic link to admin. River put up a parallel CLI prototype (branch `Shopify/cli@preview-store/prototype`) that called this endpoint via a new top-level `preview` command group and required users to either pipe `--json` output into a file or pass `--domain`/`--token` flags into a custom `preview execute` clone of `store execute`.

This PR ports the prototype onto the production CLI's `store` namespace and wires it through the existing stored-session machinery so we can delete the parallel transport entirely:

- `shopify preview create` → `shopify store create preview` (matches the language used in the DevTools kickoff and slack threads).
- `shopify preview execute` → simply use `shopify store execute --store <permanent-domain> …` because the new command writes a `kind: 'preview'` session into the same LocalStorage namespace that `store execute` reads from.
- `shopify preview claim` is M2 work and is intentionally not included.

The schema work that lets `kind: 'preview'` round-trip safely is in #7557; this PR introduces the only producer of those sessions.

### WHAT is this pull request doing?

**New command — `shopify store create preview`** (`packages/store/src/cli/commands/store/create/preview.ts`)

Flags:

| Flag | Env var | Purpose |
|---|---|---|
| `-n, --shop-name` | `SHOPIFY_FLAG_PREVIEW_STORE_SHOP_NAME` | Subdomain prefix; auto-generated when omitted. |
| `--email` | `SHOPIFY_FLAG_PREVIEW_STORE_EMAIL` | Email for the placeholder identity; defaults to a Core-generated `@previewstore.invalid` address. |
| `--country` | `SHOPIFY_FLAG_PREVIEW_STORE_COUNTRY` | ISO country code for the new store. Defaults to `"US"`. |
[...]

PR #7557, opened 2026-05-15
Add preview-store discriminator to stored store auth sessions

alfonso-noriega:
### WHY are these changes introduced?

Internal scaffolding for the upcoming `shopify store create preview` command (part of the [Preview Store for AI Agent Surfaces](https://vault.shopify.io/gsd/proposals/60T12R) M1 milestone, Foundations/DevTools epic).

Preview stores are backed by a placeholder identity rather than a real Shopify account. They are minted by Core's `/services/preview-stores` orchestrator, hand back a shop-scoped Admin API token, and have no relationship with the OAuth token endpoint — so the existing PKCE refresh and re-authentication paths in `@shopify/store` don't apply to them.

Before we can wire up `store create preview` and start persisting these sessions, the stored-session schema needs a discriminator and the recovery surfaces in `session-lifecycle` and `admin-transport` need to know not to send users back through `shopify store auth` for a placeholder-owned store.

This PR is intentionally scoped to that scaffolding. There is no new command, no new way to mint a preview session, and no user-visible behavior change. The follow-up PR introduces `shopify store create preview` and starts writing `kind: 'preview'` sessions through the existing `setStoredStoreAppSession`.

### WHAT is this pull request doing?

**Schema additions in `session-store.ts`:**
- New discriminator field `kind?: 'standard' | 'preview'` on `StoredStoreAppSession`. Optional in storage so sessions written before this field existed read back as `'standard'`.
- New optional `preview?: { placeholderAccountUuid, coreUrl, magicLinkUrl?, magicLinkExpiresAt? }` sub-object, present iff `kind === 'preview'`.
- New helpers `sessionKind(session)` and the type guard `isPreviewStoreSession(session)`.

**Sanitizer updates in `session-store.ts`:**
- Missing or unknown `kind` values are coerced to `'standard'` and the field is omitted on write — legacy buckets stay byte-identical to today.
- A `kind: 'preview'` session with missing or malformed `preview` metadata is rejected outright (downstream code requires `placeholderAccountUuid` and `coreUrl` to act on it).

**Recovery dispatcher in `recovery.ts`:**
- New `throwReauthenticatePreviewStoreError(message, store)` surfaces a preview-specific error that does not point users at `shopify store auth` (which would be meaningless for a placeholder-owned store).
- New `throwReauthenticateForSession(message, session)` dispatches to the right helper based on `isPreviewStoreSession(session)`.
- The existing `throwReauthenticateStoreAuthError` is unchanged and retained as the standard-session helper.

[...]

PR #7609, opened 2026-05-22
feat(store): add shopify store list command

alfonso-noriega:
### WHY are these changes introduced?

Fixes #0000 

### Checklist

- [ ] I've considered possible cross-platform impacts (Mac, Linux, Windows)
- [ ] I've considered possible [documentation](https://shopify.dev) changes
- [ ] I've considered analytics changes to measure impact
- [ ] The change is user-facing — I've identified the correct [bump type](https://github.com/Shopify/cli/blob/main/CONTRIBUTING.md#choosing-the-right-bump-type) (`patch` for bug fixes · `minor` for new features · `major` for [breaking changes](https://github.com/Shopify/cli/blob/main/CONTRIBUTING.md#what-counts-as-a-breaking-change)) and added a changeset with `pnpm changeset add`

PR #7610, opened 2026-05-22
Preview Stores prototype: import cli_identity_bootstrap on `store create preview`

alfonso-noriega:

This stacks on top of `store-list-command` and continues the Preview Stores prototype work led by Donald (shop/world#730858, Shopify/cli#7574). The original `store create preview` PoC handed users an explicit admin API token and a magic link, leaving every downstream CLI interaction (Identity-backed reads, `whoami`, future cross-org commands) on a separate auth path.

Donald's stacked World PR extends the orchestrator response with two new bootstrap blocks:

```
"cli_identity_bootstrap": { access_token, refresh_token, expires_in, user_id }
"store_auth_bootstrap":   { access_token, scopes, api_key, shop_domain }
```

…and Donald's CLI prototype (cli#7574) consumes them through `importIdentitySession` + `importStoreAuthBootstrap` shims that live alongside the existing PoC commands.

This PR brings the equivalent **into the `store create preview` command itself**, so the existing PoC stack (#7557 → #7558 → `store-list-command`) becomes the canonical place where preview-store creation also leaves the CLI fully authenticated. No new preview-only commands, no parallel auth paths — `store create preview` is now the only entry point and it lands the user in a real (placeholder-backed) Identity session.

**`@shopify/cli-kit` — new `importIdentitySession` primitive**

- `public/node/session.ts`: adds `IdentitySessionBootstrap` + `importIdentitySession({accessToken, refreshToken, expiresAt, userId, adminApiStoreFqdns?})`. The function writes the bootstrap tokens directly into `Sessions[identityFqdn][userId]` storage and marks the row as current. To satisfy `validateSession` without triggering device-auth re-prompts or the multi-audience token exchange, the import pre-seeds:
  - `identity.scopes` with `allDefaultScopes()` (plus `employee` when `firstPartyDev()` is true, because `validateScopes` requires the firstPartyDev flag and the `employee` scope to track each other);
  - `applications` with the bootstrap accessToken aliased under every non-store-scoped appId (partners / business-platform / storefront-renderer / app-management) AND under the store-prefixed Admin key (`${storeFqdn}-${adminAppId}`) for each entry in `adminApiStoreFqdns`. `tokensFor` reads the bare appId for non-store-scoped APIs and the store-prefixed key for Admin; aliasing the bootstrap under both lets `ensureAuthenticatedBusinessPlatform` / `ensureAuthenticatedAdmin(<shop>)` read it straight out of the cache without re-running the exchange. Without the store-prefixed seed, `theme push --store <preview>`, `store execute --store <preview>`, and any other `ensureAuthenticatedAdmin`-gated command would miss the cache and attempt an Identity token exchange that fails for placeholders (they aren't standard UserAccount resource owners that Identity can mint shop-scoped Admin tokens for).
- `public/node/environment.ts`: `getIdentityTokenInformation()` now accepts an `env` arg and surfaces `expiresAt` + an explicit `userId` when `SHOPIFY_CLI_IDENTITY_TOKEN_EXPIRES_AT` / `SHOPIFY_CLI_IDENTITY_USER_ID` are present.
- `private/node/constants.ts`: new env-var names (`SHOPIFY_CLI_IDENTITY_USER_ID`, `SHOPIFY_CLI_IDENTITY_TOKEN_EXPIRES_AT`).
- `private/node/session.ts`: `executeCompleteFlow` / `ensureAuthenticated` accept an explicit `env` arg so the bootstrap path can opt into `noPrompt: true` without prompting the user mid-import. `throwOnNoPrompt` no longer calls `logout()` on failure — wiping the entire Sessions store on every `noPrompt` rejection is destructive when the caller is in a 401-retry cascade against a backend-issued session that can't be reconstructed without re-creating a shop. Users who explicitly want to clear sessions can run `shopify auth logout`.
- `private/node/session/exchange.ts`: `exchangeAccessForApplicationTokens` is now tolerant of per-audience failures via `Promise.allSettled`. When at least one exchange succeeds the result is the union of successes; only when every exchange fails do we surface the first identity-level error (matching the previous "all-or-nothing" behavior). Motivation: backend-issued sessions (preview-store placeholders, app-automation tokens) may have a narrower audience grant than a real PKCE user; the old all-or-nothing semantics would surface that as a single confusing `invalid_request` mid-`Promise.all` rather than as a scoped "no `<api>` token found" at the actual call site.

**`@shopify/store` — `store create preview` consumes the new bootstrap shape**

- `services/store/create/preview/client.ts`: parses `cli_identity_bootstrap` and `store_auth_bootstrap` from the orchestrator JSON. Defensive narrowing drops partial/malformed payloads silently so the create flow can't be stranded by a half-formed bootstrap (analytics still report the store as created on Core; the user just doesn't get the import side effects).
- `services/store/create/preview/index.ts`: persists the store-auth bucket **first** (so a failed Identity import never orphans the just-created shop on the CLI side), then best-effort imports the Identity session via `importIdentitySession` with `adminApiStoreFqdns: [resolvedShopDomain]` so the store-prefixed Admin entry lands alongside the bare-audience entries. When `store_auth_bootstrap` is present, the bucket is keyed under `bootstrap.shop_domain` because that's the host that actually answers Admin API requests for this shop in the local rig (top-level `shop_permanent_domain` is canonical/display-only and doesn't route to a Spin instance on local). That same `shop_domain` is also what the result struct surfaces as `shopPermanentDomain`, so users can pipe `--json` output straight into `store execute --store ...` and the Admin URL the CLI builds resolves.
- `services/store/create/preview/result.ts`: JSON output now includes `identityImported: boolean` so agents can tell whether the bootstrap was actually consumed; text output gains a "CLI identity" section explaining the outcome.

#### Automated

```bash
pnpm --filter @shopify/cli-kit vitest run src/private/node/session src/public/node/session
pnpm --filter @shopify/store vitest run preview
pnpm --filter @shopify/cli-kit --filter @shopify/store type-check
```

Expected: 126 cli-kit session tests pass, 22 preview tests pass (4 new for bootstrap behavior + the previously-existing baseline), and the only type-check error is the pre-existing `client.test.ts` `Response`-vs-`node-fetch.Response` mismatch that's unrelated to this PR.

#### Manual against the local rig (with Donald's paired World branch checked out)

```bash
export SHOPIFY_SERVICE_ENV=local

# 1) Wipe any stale session row so we're starting clean
rm ~/Library/Preferences/shopify-cli-kit-nodejs/config.json
rm ~/Library/Preferences/shopify-cli-store-nodejs/config.json

# 2) Create a preview store. Expect both bootstrap blocks in the orchestrator
#    response, `identityImported: true` in the CLI output, and a placeholder
#    Identity session persisted on disk.
node packages/cli/bin/dev.js store create preview --json \
  | tee /tmp/preview.json \
  | jq '{shopPermanentDomain, identityImported, userId}'

# 3) Confirm the placeholder Identity session is keyed correctly:
jq '.sessions | fromjson | to_entries[] | .value | keys' \
  ~/Library/Preferences/shopify-cli-kit-nodejs/config.json

# 4) Use the surfaced shopPermanentDomain to round-trip through store execute.
#    No `--token` flag needed — the store-auth bucket is hit via the standard seam.
DOMAIN=$(jq -r .shopPermanentDomain /tmp/preview.json)
node packages/cli/bin/dev.js store execute --store "$DOMAIN" \
  --query 'query { shop { name myshopifyDomain } }' --json
```

### Post-release steps

None. This is part of the Preview Stores prototype stack and is paired to Donald's local-only `shop/world#730858`; not intended to ship as-is.

### Checklist

- [ ] The change is user-facing — no changeset added because this is a prototype stack paired to local-only World endpoints and is not intended to ship as-is

### DIFF
--- a/packages/cli-kit/src/private/node/constants.ts
+++ b/packages/cli-kit/src/private/node/constants.ts
@@ -42,6 +42,8 @@ export const environmentVariables = {
   spinAppHost: 'SPIN_APP_HOST',
   organization: 'SHOPIFY_CLI_ORGANIZATION',
   identityToken: 'SHOPIFY_CLI_IDENTITY_TOKEN',
+  identityTokenUserId: 'SHOPIFY_CLI_IDENTITY_USER_ID',
+  identityTokenExpiresAt: 'SHOPIFY_CLI_IDENTITY_TOKEN_EXPIRES_AT',
   refreshToken: 'SHOPIFY_CLI_REFRESH_TOKEN',
   otelURL: 'SHOPIFY_CLI_OTEL_EXPORTER_OTLP_ENDPOINT',
   themeKitAccessDomain: 'SHOPIFY_CLI_THEME_KIT_ACCESS_DOMAIN',
--- a/packages/cli-kit/src/private/node/session.test.ts
+++ b/packages/cli-kit/src/private/node/session.test.ts
@@ -177,7 +177,7 @@ describe('ensureAuthenticated when previous session is invalid', () => {
     expect(fetchSessions).toHaveBeenCalledOnce()
   })
 
-  test('throws an error and logs out if there is no session and prompting is disabled,', async () => {
+  test('throws an error when there is no session and prompting is disabled, without wiping the Sessions store', async () => {
     // Given
     vi.mocked(validateSession).mockResolvedValueOnce('needs_full_auth')
     vi.mocked(fetchSessions).mockResolvedValue(undefined)
@@ -189,13 +189,12 @@ describe('ensureAuthenticated when previous session is invalid', () => {
 
 The CLI is currently unable to prompt for reauthentication.`,
     )
-    expect(secureRemove).toHaveBeenCalled()
-
-    // Then
-    await expect(getLastSeenAuthMethod()).resolves.toEqual('none')
-
-    // If there never was an auth event, the userId is 'unknown'
-    await expect(getLastSeenUserIdAfterAuth()).resolves.toBe('unknown')
+    // `throwOnNoPrompt` intentionally does NOT call `logout()` anymore: wiping the
+    // full Sessions store on every noPrompt failure is destructive for callers that
+    // rely on a backend-issued session (e.g. preview-store placeholders) that can't
+    // be reconstructed without re-creating a shop. Users who explicitly want to
+    // clear sessions can run `shopify auth logout`.
+    expect(secureRemove).not.toHaveBeenCalled()
   })
 
   test('executes complete auth flow if session is for a different fqdn', async () => {
--- a/packages/cli-kit/src/private/node/session.ts
+++ b/packages/cli-kit/src/private/node/session.ts
@@ -228,22 +228,39 @@ ${outputToken.json(applications)}
   const validationResult = await validateSession(scopes, applications, currentSession)
 
   let newSession = {}
+  const canAuthenticateWithoutPrompt = canAuthenticateWithoutPromptFromEnvironment(_env)
 
   if (validationResult === 'needs_full_auth') {
-    await throwOnNoPrompt(noPrompt)
+    if (!canAuthenticateWithoutPrompt) {
+      await throwOnNoPrompt(noPrompt)
+    }
     outputDebug(outputContent`Initiating the full authentication flow...`)
-    newSession = await executeCompleteFlow(applications, currentSession?.identity.alias)
+    newSession = await executeCompleteFlow(applications, _env, currentSession?.identity.alias)
   } else if (validationResult === 'needs_refresh' || forceRefresh) {
     outputDebug(outputContent`The current session is valid but needs refresh. Refreshing...`)
     try {
       newSession = await refreshTokens(currentSession!, applications)
     } catch (error) {
       if (error instanceof InvalidGrantError) {
-        await throwOnNoPrompt(noPrompt)
-        newSession = await executeCompleteFlow(applications, currentSession?.identity.alias)
+        if (!canAuthenticateWithoutPrompt) {
+          await throwOnNoPrompt(noPrompt)
+        }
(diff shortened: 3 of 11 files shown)


alfonso-noriega:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/Shopify/cli/7610?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#7612** <a href="https://app.graphite.com/github/pr/Shopify/cli/7612?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#7611** <a href="https://app.graphite.com/github/pr/Shopify/cli/7611?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#7610** <a href="https://app.graphite.com/github/pr/Shopify/cli/7610?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Shopify/cli/7610?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#7609** <a href="https://app.graphite.com/github/pr/Shopify/cli/7609?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#7667](https://github.com/Shopify/cli/pull/7667) <a href="https://app.graphite.com/github/pr/Shopify/cli/7667?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#7558** <a href="https://app.graphite.com/github/pr/Shopify/cli/7558?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#7557** <a href="https://app.graphite.com/github/pr/Shopify/cli/7557?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

PR #7611, opened 2026-05-22
Add `shopify auth whoami` diagnostic command

alfonso-noriega:
### WHY are these changes introduced?

This stacks on top of the Preview Stores prototype work. Once `store create preview` started importing real Identity sessions (the parent PR in this stack), we needed a reliable way to inspect "what's actually on disk right now" without leaking tokens. Until now the only ways to answer that were `cat ~/Library/Preferences/shopify-cli-kit-nodejs/config.json | jq` (which dumps raw tokens) or guessing from `shopify auth login`/`logout` behavior.

A first-class `shopify auth whoami` solves the visibility gap and gives us a single, scriptable probe for downstream work — including the BP-backed `store list` work that depends on knowing whether the current session is a placeholder.

The command is intentionally scoped as a **diagnostic**: it inspects local state only, doesn't refresh tokens, doesn't make network calls, and doesn't return account info derived from a remote API. That makes it cheap, idempotent, and safe to run mid-debug.

### WHAT is this pull request doing?

**`@shopify/cli-kit` — `getCurrentSessionInfo` snapshot helper**

- `public/node/session.ts`: adds `getCurrentSessionInfo({raw?: boolean})` that reads `sessionStore.fetch()` + `getCurrentSessionId()` and projects the active row into a typed `CurrentSessionInfo` snapshot. By default token values are **masked** to first-8-chars + length; `{raw: true}` includes both the unredacted current session row and the full Sessions blob from disk (every fqdn × every userId) for cases where you need to debug "why isn't the session I think is active actually active." Adds a heuristic flag `looksLikePlaceholder: boolean` based on the session shape (UUID-shaped userId + no alias), which downstream commands can use to choose between BP-backed and local-backed code paths.

**`@shopify/cli` — `shopify auth whoami` command**

- `commands/auth/whoami.ts`: new command, registered as `auth:whoami` alongside the existing `auth:login` and `auth:logout`. Renders a compact human-readable summary by default (userId / fqdn / alias / masked-token previews / scope count / cached application audiences) and emits JSON with `--json`. The `--raw` flag implies `--json` and adds the unredacted session row + full Sessions blob; it's documented as containing live access tokens with an explicit warning not to paste output into shared channels.
- `index.ts`: registers the command in the static `COMMANDS` map alongside `auth:login` and `auth:logout`.
- `oclif.manifest.json`: regenerated to include the new command.

[...]

PR #7612, opened 2026-05-22
Preview Stores prototype: default `store list` to Business Platform with local fallback

alfonso-noriega:

The original `store list` command (introduced in the `store-list-command` branch) reads the local `shopify-cli-store-nodejs` LocalStorage cache and lists every shop the CLI has previously authed against on this machine. That's the right answer when you want "what's on this disk" — useful for debugging, for freshly-created preview stores, and offline use — but it's the wrong default for the common case of "what stores does my Shopify account actually have access to."

After the parent PR in this stack, `store create preview` leaves the CLI authenticated as a real (placeholder-backed) Identity session. With that piece in place, the logical default for `store list` becomes Business Platform: list what the currently-logged-in account can see in the platform itself.

But the placeholder session is a wrinkle:

- Donald's stacked World PR (shop/world#730858) deliberately doesn't make `PlaceholderAccount` a `UserAccount` in BP. The "Path B keep the change surface tiny" framing means BP's `currentUserAccount.organizationsWithAccessToDestination` returns `null` for placeholders, so a BP listing for a placeholder session is always empty.
- Donald's own CLI verification (Shopify/cli#7574) calls `organization list --json` for the placeholder session, but his description explicitly says "succeeds" — i.e. doesn't crash — without asserting that it returns useful data. His prototype is silent on the broader UX gap.

This PR takes the smallest deviation from Donald's pattern that still gives users a sensible default: keep BP as the default for real users, fall back to the local cache for placeholder sessions, and let the user override with `--source` either way.

**Two sources, dispatched by the user's choice (or sensible default)**

- `services/store/list/bp-source.ts` (new): `listBusinessPlatformStores()` does a two-phase fetch — destinations BP for the orgs the current user has CLI access to (`organizationsWithAccessToDestination(destination: APPS_CLI)`), then per-org organizations BP for `accessibleShops` paginated to completion (50/page, capped at 200 pages as a runaway guard). No `STORE_TYPE` filter — `store list` should show production / development / preview / etc., not just app-dev sandboxes, so the type appears in its own column instead. Per-org fetches run in parallel because CLI users typically have one or two orgs and the politeness cost is dwarfed by the wall-clock gain.
- `services/store/list/index.ts`: refactored from a synchronous list-and-project into an async dispatch that picks the source from `options.source ∈ {'bp', 'local'}`. The function now returns a `ListStoredStoresResult` wrapper with `{entries, source, notice?, currentUserEmail?}` instead of bare `StoreListEntry[]`, so the renderer can distinguish "I queried BP and it returned 0 because you're a placeholder" from "I queried local and there's nothing on disk."
- `commands/store/list.ts`: adds `--source` (default `bp`, with `local` as the explicit opt-out), `--search` (forwarded to BP only — local-source ignores it; callers can post-filter), keeps the existing `--kind` flag but documents that it only applies to `--source local` because BP entries don't carry the local-cache `standard`/`preview` discriminator.
- `services/store/list/result.ts`: two different table layouts based on source — BP shows `Store / Name / Type / Organization`, local keeps the original `Store / Kind / User`. The empty-state message for BP suggests `--source local` as the natural fallback for users who just created a preview store and want to see it.

**Notice carries the placeholder-aware messaging**

When BP returns `currentUserAccount: null`, the result includes a notice:

```
Business Platform could not resolve the current session as a user account
(this is expected for placeholder / preview-store sessions).
Re-run with `--source local` to list stores from the local cache instead.
```

This is the deliberate UX choice: rather than auto-falling-back to local (which would mask the fact that the user is on a placeholder session), we surface the explanation and let the user opt in.

#### Automated

```bash
pnpm --filter @shopify/store vitest run list
pnpm --filter @shopify/store type-check
```

Expected: 20 tests pass (5 local-source + 4 BP-source dispatch + 6 result-renderer × 2 modes). The only type-check error is the pre-existing `client.test.ts` `Response`-vs-`node-fetch.Response` mismatch unrelated to this PR.

#### Manual smoke tests

```bash
export SHOPIFY_SERVICE_ENV=local

# Setup a placeholder session via the parent PR's flow
shopify store create preview --json > /tmp/preview.json

# Default (--source bp) for a placeholder session
shopify store list
# → Notice about "could not resolve the current session as a user account"
# → Empty BP table + suggestion to use --source local

# Local source — should show the freshly-created preview store
shopify store list --source local
# → 1 store, kind=preview, user=—

# JSON output of each
shopify store list --json
shopify store list --source local --json

# Filter by kind on local source only
shopify store list --source local --kind preview
```

For the "real user post auth login" path, the BP query should return their accessible stores across every org with CLI access. That path will be exercisable once `auth login` is wired through the rig.

### Post-release steps

None. This is part of the Preview Stores prototype stack.

### Checklist

- [x] I've considered possible documentation changes (descriptions updated to reflect the dual-source model)
- [ ] The change is user-facing — no changeset added because this is a prototype stack paired to local-only World endpoints and is not intended to ship as-is

### DIFF
--- a/packages/store/src/cli/commands/store/list.ts
+++ b/packages/store/src/cli/commands/store/list.ts
@@ -1,31 +1,49 @@
-import {listStoredStores, type StoreListEntryKind} from '../../services/store/list/index.js'
+import {listStoredStores, type StoreListEntryKind, type StoreListSource} from '../../services/store/list/index.js'
 import {writeStoreListResult} from '../../services/store/list/result.js'
 import StoreCommand from '../../utilities/store-command.js'
 import {globalFlags, jsonFlag} from '@shopify/cli-kit/node/cli'
 import {Flags} from '@oclif/core'
 
 const STORE_LIST_KINDS: StoreListEntryKind[] = ['standard', 'preview']
+const STORE_LIST_SOURCES: StoreListSource[] = ['bp', 'local']
 
 export default class StoreList extends StoreCommand {
-  static summary = 'List stored store-auth sessions.'
+  static summary = 'List stores accessible to the current Shopify CLI session.'
 
-  static descriptionWithMarkdown = `Lists every store that has a locally stored auth session, including both standard PKCE-authenticated stores (via \`shopify store auth\`) and preview stores (via \`shopify store create preview\`).
+  static descriptionWithMarkdown = `Lists every shop the currently-authenticated Shopify account has access to.
 
-Use \`--kind\` to filter by session type, or \`--json\` to emit a machine-readable list for agent consumption.`
+By default (\`--source bp\`) the command queries Business Platform across every organization the logged-in user belongs to, mirroring the stores you'd see in the Shopify admin. Use \`--source local\` to enumerate only the locally-cached store-auth sessions (including preview stores created by \`shopify store create preview\` that haven't propagated to BP yet).
+
+The \`--kind\` filter only applies to \`--source local\` because BP-sourced entries don't carry the local-cache \`standard\` / \`preview\` discriminator.`
 
   static description = this.descriptionWithoutMarkdown()
 
   static examples = [
     '<%= config.bin %> <%= command.id %>',
-    '<%= config.bin %> <%= command.id %> --kind preview',
+    '<%= config.bin %> <%= command.id %> --source local',
+    '<%= config.bin %> <%= command.id %> --source local --kind preview',
+    '<%= config.bin %> <%= command.id %> --search "preview-"',
     '<%= config.bin %> <%= command.id %> --json',
   ]
 
   static flags = {
     ...globalFlags,
     ...jsonFlag,
+    source: Flags.string({
+      description:
+        'Data source for the listing. `bp` (default) queries Business Platform for the logged-in account; `local` enumerates the on-disk store-auth cache.',
+      env: 'SHOPIFY_FLAG_STORE_LIST_SOURCE',
+      options: STORE_LIST_SOURCES,
+      default: 'bp',
+      required: false,
+    }),
+    search: Flags.string({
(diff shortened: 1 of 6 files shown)


alfonso-noriega:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/Shopify/cli/7612?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#7612** <a href="https://app.graphite.com/github/pr/Shopify/cli/7612?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Shopify/cli/7612?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#7611** <a href="https://app.graphite.com/github/pr/Shopify/cli/7611?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#7610** <a href="https://app.graphite.com/github/pr/Shopify/cli/7610?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#7609** <a href="https://app.graphite.com/github/pr/Shopify/cli/7609?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#7667](https://github.com/Shopify/cli/pull/7667) <a href="https://app.graphite.com/github/pr/Shopify/cli/7667?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#7558** <a href="https://app.graphite.com/github/pr/Shopify/cli/7558?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#7557** <a href="https://app.graphite.com/github/pr/Shopify/cli/7557?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

PR #7615, opened 2026-05-22
Send analytics in background

gonzaloriestra:

HackDays project: https://vault.shopify.io/hackdays/154/projects/24279-Shopify-CLI-UX-improvements

Analytics delivery currently blocks command completion, adding noticeable latency to every CLI invocation.

Passes the completed analytics payload over stdin to a detached internal command, allowing the original command to exit immediately.

This works for Mac and Linux, but on Windows we still have to wait for the process because of a known issue. We also wait on CI to avoid issues with the job teardown.

Benchmark for `shopify version`:

| Version | Average time (10 runs) |
| --- | ---:|
| `4.5.0` | 1\.352s |
| `snapshot` | 0\.248s |

https://github.com/user-attachments/assets/cea88f10-ad76-4f42-82f9-b79a6ac7fe76

### How to test your changes?

- `pnpm i -g @shopify/cli@0.0.0-snapshot-20260811121607`
- `shopify version --verbose`
- `shopify deploy --wrong-flag-analytics-test`
- Check that the payload was processed:
```
SELECT *
FROM `sdp-ingest.monorail.monorail_app_cli3_command_1`
WHERE event_timestamp >= TIMESTAMP_SUB(CURRENT_TIMESTAMP(), INTERVAL 1 HOUR) AND CONTAINS_SUBSTR(
    payload.error_message,
    'Nonexistent flag: --wrong-flag-analytics-test'
  )
```

### Checklist

### DIFF
--- a/packages/cli-kit/src/public/node/analytics.ts
+++ b/packages/cli-kit/src/public/node/analytics.ts
@@ -1,6 +1,6 @@
-import {alwaysLogAnalytics, alwaysLogMetrics, analyticsDisabled, isShopify} from './context/local.js'
+import {alwaysLogAnalytics, alwaysLogMetrics, analyticsDisabled, ciPlatform, isShopify} from './context/local.js'
 import * as metadata from './metadata.js'
-import {publishMonorailEvent, MONORAIL_COMMAND_TOPIC} from './monorail.js'
+import {publishMonorailEvent, MONORAIL_COMMAND_TOPIC, type Schemas} from './monorail.js'
 import {fanoutHooks} from './plugins.js'
 import {sendErrorToBugsnag} from './error-handler.js'
 import {outputContent, outputDebug, outputToken} from './output.js'
@@ -36,6 +36,70 @@ interface ReportAnalyticsEventOptions {
   exitMode: CommandExitMode
 }
 
+type MonorailCommandPayload = Schemas[typeof MONORAIL_COMMAND_TOPIC]
+interface AnalyticsPayload {
+  public: MonorailCommandPayload['public'] & {cmd_all_exit: CommandExitMode}
+  sensitive: MonorailCommandPayload['sensitive']
+}
+
+async function sendAnalyticsEvent(
+  payload: AnalyticsPayload,
+  skipMonorailAnalytics: boolean,
+  skipMetricAnalytics: boolean,
+): Promise<void> {
+  const doMonorail = async () => {
+    if (skipMonorailAnalytics) return
+    const response = await publishMonorailEvent(MONORAIL_COMMAND_TOPIC, payload.public, payload.sensitive)
+    if (response.type === 'error') {
+      outputDebug(response.message)
+    }
+  }
+
+  const doOpenTelemetry = async () => {
+    const active = payload.public.cmd_all_timing_active_ms ?? 0
+    const network = payload.public.cmd_all_timing_network_ms ?? 0
+    const prompt = payload.public.cmd_all_timing_prompts_ms ?? 0
+
+    return recordMetrics(
+      {
+        skipMetricAnalytics,
+        cliVersion: payload.public.cli_version,
+        owningPlugin: payload.public.cmd_all_plugin ?? '@shopify/cli',
+        command: payload.public.command,
+        exitMode: payload.public.cmd_all_exit,
+      },
+      {
+        active,
+        network,
+        prompt,
+      },
+    )
+  }
+
+  await Promise.all([doMonorail(), doOpenTelemetry()])
+}
+
+export async function sendAnalyticsEventFromStdin(): Promise<void> {
+  try {
+    const {readStdinString} = await import('./system.js')
+    const payloadStr = await readStdinString()
+    if (payloadStr === undefined) throw new Error('No analytics payload received from stdin')
+
+    const {payload, skipMonorailAnalytics, skipMetricAnalytics} = JSON.parse(payloadStr) as {
+      payload: AnalyticsPayload
+      skipMonorailAnalytics: boolean
+      skipMetricAnalytics: boolean
+    }
+
+    await sendAnalyticsEvent(payload, skipMonorailAnalytics, skipMetricAnalytics)
+    // eslint-disable-next-line no-catch-all/no-catch-all
+  } catch (error) {
+    const message = error instanceof Error ? error.message : String(error)
+    outputDebug(`Failed to send analytics in background: ${message}`)
+    await sendErrorToBugsnag(error, 'expected_error')
+  }
+}
+
 /**
  * Report an analytics event, sending it off to Monorail -- Shopify's internal analytics service.
  *
@@ -44,11 +108,11 @@ interface ReportAnalyticsEventOptions {
  */
 export async function reportAnalyticsEvent(options: ReportAnalyticsEventOptions): Promise<void> {
   try {
+    const commandStartOptions = metadata.getAllSensitiveMetadata().commandStartOptions
+    if (commandStartOptions?.startCommand === 'send-analytics') return
+
     const payload = await buildPayload(options)
-    if (payload === undefined) {
-      // Nothing to log
-      return
-    }
+    if (payload === undefined) return
 
     let withinRateLimit = false
     await runWithRateLimit({
@@ -65,40 +129,42 @@ export async function reportAnalyticsEvent(options: ReportAnalyticsEventOptions)
 
     const skipMonorailAnalytics = !alwaysLogAnalytics() && analyticsDisabled()
     const skipMetricAnalytics = !alwaysLogMetrics() && analyticsDisabled()
-    if (skipMonorailAnalytics || skipMetricAnalytics) {
+    if (skipMonorailAnalytics && skipMetricAnalytics) {
       outputDebug(outputContent`Skipping command analytics, payload: ${outputToken.json(payload)}`)
+      return
     }
 
-    const doMonorail = async () => {
-      if (skipMonorailAnalytics) {
-        return
-      }
-      const response = await publishMonorailEvent(MONORAIL_COMMAND_TOPIC, payload.public, payload.sensitive)
-      if (response.type === 'error') {
-        outputDebug(response.message)
-      }
-    }
-    const doOpenTelemetry = async () => {
-      const active = payload.public.cmd_all_timing_active_ms ?? 0
-      const network = payload.public.cmd_all_timing_network_ms ?? 0
-      const prompt = payload.public.cmd_all_timing_prompts_ms ?? 0
+    const [{platformAndArch}, {isInsideContainer}] = await Promise.all([import('./os.js'), import('./system.js')])
+    const sendInBackground =
+      !commandStartOptions?.requiresSyncAnalytics &&
+      platformAndArch().platform !== 'windows' &&
+      !ciPlatform().isCI &&
+      !isInsideContainer()
+    const deliveryDescription = sendInBackground ? ' in background' : ''
+    outputDebug(outputContent`Sending command analytics${deliveryDescription}, payload: ${outputToken.json(payload)}`)
 
-      return recordMetrics(
-        {
-          skipMetricAnalytics,
-          cliVersion: payload.public.cli_version,
-          owningPlugin: payload.public.cmd_all_plugin ?? '@shopify/cli',
-          command: payload.public.command,
-          exitMode: options.exitMode,
-        },
-        {
-          active,
-          network,
-          prompt,
-        },
-      )
+    if (!sendInBackground) {
+      await sendAnalyticsEvent(payload, skipMonorailAnalytics, skipMetricAnalytics)
+      return
     }
-    await Promise.all([doMonorail(), doOpenTelemetry()])
+
+    const {exec} = await import('./system.js')
+    const argv = process.argv
+    if (!argv[1]) return
+    const nodeBinary = process.execPath
+    const shopifyBinary = argv[1]
+    const args = [shopifyBinary, 'send-analytics']
+
+    const analyticsProcess = exec(nodeBinary, args, {
+      background: true,
+      env: {...process.env, SHOPIFY_CLI_NO_ANALYTICS: '1'},
+      input: JSON.stringify({payload, skipMonorailAnalytics, skipMetricAnalytics}),
+      externalErrorHandler: async (error: unknown) => {
+        outputDebug(`Failed to send analytics in background: ${(error as Error).message}`)
+      },
+    })
+    // eslint-disable-next-line no-void
+    void analyticsProcess
 
     // eslint-disable-next-line no-catch-all/no-catch-all
   } catch (error) {
@@ -111,7 +177,11 @@ export async function reportAnalyticsEvent(options: ReportAnalyticsEventOptions)
   }
 }
 
-async function buildPayload({config, errorMessage, exitMode}: ReportAnalyticsEventOptions) {
+async function buildPayload({
+  config,
+  errorMessage,
+  exitMode,
+}: ReportAnalyticsEventOptions): Promise<AnalyticsPayload | undefined> {
   const {commandStartOptions, environmentFlags, ...sensitiveMetadata} = metadata.getAllSensitiveMetadata()
   if (commandStartOptions === undefined) {
     outputDebug('Unable to log analytics event - no information on executed command')
--- a/packages/cli-kit/src/public/node/system.ts
+++ b/packages/cli-kit/src/public/node/system.ts
@@ -278,11 +278,12 @@ function buildExec(
   }
   const executionCwd = options?.cwd ?? cwd()
   checkCommandSafety(command, {cwd: executionCwd})
+  const backgroundStdio = options?.input === undefined ? 'ignore' : (['pipe', 'ignore', 'ignore'] as const)
   const commandProcess = execa(command, args, {
     env,
     cwd: executionCwd,
     input: options?.input,
-    stdio: options?.background ? 'ignore' : options?.stdio,
+    stdio: options?.background ? backgroundStdio : options?.stdio,
     stdin: options?.stdin,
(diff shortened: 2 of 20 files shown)


gonzaloriestra:
* **#8137** <a href="https://app.graphite.com/github/pr/Shopify/cli/8137?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#8131** <a href="https://app.graphite.com/github/pr/Shopify/cli/8131?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#8136** <a href="https://app.graphite.com/github/pr/Shopify/cli/8136?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#7615** <a href="https://app.graphite.com/github/pr/Shopify/cli/7615?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Shopify/cli/7615?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/cli-kit/src/public/node/analytics.ts
```diff
@@ -36,6 +36,58 @@ interface ReportAnalyticsEventOptions {
   exitMode: CommandExitMode
 }
 
+export async function sendAnalyticsEventFromFile(payloadFile: string): Promise<void> {
```
dmerand: Is it possible to type the shape of `payloadFile`? I don't love that we lose some type safety by passing a file in. If not that, perhaps we should validate the JSON schema against our expected monorail output?

If we're not already doing this, I'm OK with it being a follow-up.

packages/cli-kit/src/public/node/analytics.ts
```diff
@@ -65,40 +116,55 @@ export async function reportAnalyticsEvent(options: ReportAnalyticsEventOptions)
...
-      )
+    await writeFile(payloadPath, JSON.stringify(fullPayload), {encoding: 'utf8', mode: 0o600, flag: 'wx'})
+
+    const {exec} = await import('./system.js')
+    const argv = process.argv
+    if (!argv[0] || !argv[1]) {
+      await removeFile(payloadPath)
+      return
+    }
+    const nodeBinary = argv[0]
```
dmerand: This raised my hackles so I asked my robot pals, who tell me that `process.execPath` is the canonical path to the current Node executable, and that `argv[0]` can be overridden.

Suggested approach was this, but I'm not sure from experience whether it's better:

 ```ts
   import {spawn} from 'node:child_process'
   import {fileURLToPath} from 'node:url'

   const workerPath = fileURLToPath(
     new URL('./analytics-worker.js', import.meta.url),
   )

   const child = spawn(process.execPath, [workerPath, payloadPath], {
     detached: process.platform !== 'win32',
     stdio: 'ignore',
     env: {...process.env, SHOPIFY_CLI_NO_ANALYTICS: '1'},
   })

   child.unref()
 ```

gonzaloriestra: The notification system have been working exactly that way for a while, so I guess it's fine. But it looks more robust with execPath, so I've update both places.

packages/cli-kit/src/public/node/analytics.ts
gonzaloriestra: Good idea, done!

packages/cli/src/index.ts
```diff
@@ -149,6 +150,7 @@ export const COMMANDS: any = {
   search: Search,
   upgrade: Upgrade,
   version: VersionCommand,
+  'send-analytics': SendAnalytics,
```
isaacroldan, line 153: This command is only available in the main CLI, BUT, we also have `create-app` which is like a "small CLI" with just "app init".

With this change, `create-app` will try to trigger the analytics in background and will always fail.

Can we force analytics from this CLI to be sync?

packages/cli-kit/src/public/node/analytics.ts
```diff
@@ -36,6 +36,85 @@ interface ReportAnalyticsEventOptions {
...
+    }
+
+    await Promise.all([doMonorail(), doOpenTelemetry()])
+  } catch (error) {
+    if (error instanceof Error) {
+      outputDebug(`Failed to send analytics in background: ${error.message}`)
+    } else {
+      throw error
+    }
+  } finally {
```
isaacroldan: If this starts failing, we have no visibility, we could lose analytics and not realize until days later...
Should this report to bugsnag?

packages/cli-kit/src/public/node/analytics.ts
```diff
@@ -36,6 +36,85 @@ interface ReportAnalyticsEventOptions {
...
+            cmd_all_timing_active_ms: zod.number().optional(),
+            cmd_all_timing_network_ms: zod.number().optional(),
+            cmd_all_timing_prompts_ms: zod.number().optional(),
+          })
+          .passthrough(),
+        sensitive: zod.object({args: zod.string()}).passthrough(),
+        skipMonorailAnalytics: zod.boolean(),
+        skipMetricAnalytics: zod.boolean(),
+      })
+      .parse(JSON.parse(payloadStr))
```
isaacroldan: is this payload schema "duplicated" with the schema we use to actually build the payload? I think there is some drift risk here... can we use the same Typescript types we have when building the payload?

packages/cli-kit/src/public/node/analytics.ts
```diff
@@ -65,40 +143,55 @@ export async function reportAnalyticsEvent(options: ReportAnalyticsEventOptions)
...
-          command: payload.public.command,
-          exitMode: options.exitMode,
-        },
-        {
-          active,
-          network,
-          prompt,
-        },
-      )
+    await writeFile(payloadPath, JSON.stringify(fullPayload), {encoding: 'utf8', mode: 0o600, flag: 'wx'})
```
isaacroldan: this is writing sensitive data to a file... there is a small posibility that the clenaup fails, or the send-analytics invocation fails... something fails and this file is never deleted.
What can we do to prevent that? 🤔

gonzaloriestra: In theory we delete even on failure paths.  
  
But if that fails, we are reducing the risk by:  
\- Creating the file in the tmp folder, that should eventually be removed  
\- Setting the file permissions so that only the user running the CLI can access it (0600)   
\- Using a random filename  
  
Do you think it's enough? Alternatively, we could try to encrypt or pass the data in a different way.

packages/cli/src/index.ts
gonzaloriestra: Good catch! Fixed

packages/cli-kit/src/public/node/analytics.ts
gonzaloriestra: Good idea, I added it.

packages/cli-kit/src/public/node/analytics.ts
gonzaloriestra: Agree, I was already working on that. Fixed now.

packages/cli-kit/src/public/node/analytics.ts
gonzaloriestra: Another option is to clean all the CLI payloads after every run (or old ones at least)

isaacroldan: um, cleaning all files wouldn't be compatible with multiple CLIs running in different terminals (one might delete another's file).  
  
maybe ask Claude/Codex if there is reliable way to encrypt? if not I guess there isn't much we can do

gonzaloriestra: Yeah, that's why I suggested to only remove the old ones.  
  
Encrypting looks complex because we need to keep a key somehow. But let me try by passing the data through stdin instead.

gonzaloriestra: I'm finally passing it via stdin and it's working :v:

packages/cli-kit/src/public/node/analytics.ts
```diff
@@ -45,8 +109,7 @@ interface ReportAnalyticsEventOptions {
 export async function reportAnalyticsEvent(options: ReportAnalyticsEventOptions): Promise<void> {
   try {
     const payload = await buildPayload(options)
```
isaacroldan, line 114: have you measured this part? because I think this could be sometimes more time expensive that the call to monorail :thinking_face: And this is still being run sync right?

gonzaloriestra: Not that part especifically, but in the description there are some benchmarks of the full command. The slowest parts are clearly the requests.  
  
I didn't check the difference after the file/stdin change, but it looks similar.

packages/cli-kit/src/public/node/analytics.ts
```diff
@@ -65,40 +128,43 @@ export async function reportAnalyticsEvent(options: ReportAnalyticsEventOptions)
...
-      )
+
+    const {exec} = await import('./system.js')
+    const argv = process.argv
+    if (!argv[1]) return
+    const nodeBinary = process.execPath
+    const shopifyBinary = argv[1]
+    const args = [shopifyBinary, 'send-analytics']
+
+    const analyticsProcess = exec(nodeBinary, args, {
```
isaacroldan, line 158: I think this breaks analytics for `app dev` (and any command using `waitForPostRunHookAndExit`):

1. On exit, the postrun hook spawns the detached `send-analytics` child and sets `postRunHookCompleted = true`
2. ~100ms later, `waitForPostRunHookAndExit` runs `treeKill(process.pid, 'SIGINT')`
3. `detached: true` changes the process group but not the PPID, so tree-kill still finds the child (`pgrep -lfP`) and kills it while it's still booting, before it can send anything

So dev session events would be silently lost on Mac/Linux.

Simplest fix: make `app dev` (or any command that tree-kills on exit) send analytics synchronously in-process instead of spawning the child — the sync path already exists for `create-app`.

packages/cli-kit/src/public/node/analytics.ts
```diff
@@ -65,40 +128,43 @@ export async function reportAnalyticsEvent(options: ReportAnalyticsEventOptions)
...
+      input: JSON.stringify({payload, skipMonorailAnalytics, skipMetricAnalytics}),
+      externalErrorHandler: async (error: unknown) => {
+        outputDebug(`Failed to send analytics in background: ${(error as Error).message}`)
+      },
+    })
+    if (sendInBackground) {
+      // eslint-disable-next-line no-void
+      void analyticsProcess
+    } else {
+      await analyticsProcess
```
isaacroldan: On Windows and CI we now spawn the child and `await` it with no timeout: a full second CLI boot (node startup + plugin load + hooks) plus the send, where the old code only awaited the in-process Monorail POST. That makes every command on those platforms slower than `main`, and a wedged child (e.g. blackholed network) hangs the command forever — the old `fetch` path used the `'non-blocking'` behavior so it couldn't do that.

Since we have to wait on these platforms anyway, spawning a process buys us nothing: I'd route the non-background case through the existing in-process `sendAnalyticsEvent` (same as `create-app`), which keeps the old non-blocking fetch semantics for free.

packages/cli-kit/src/public/node/analytics.ts
```diff
@@ -45,8 +109,7 @@ interface ReportAnalyticsEventOptions {
 export async function reportAnalyticsEvent(options: ReportAnalyticsEventOptions): Promise<void> {
   try {
     const payload = await buildPayload(options)
-    if (payload === undefined) {
-      // Nothing to log
+    if (payload === undefined || payload.public.command === 'send-analytics') {
```
isaacroldan: Minor: in the child, the `command === 'send-analytics'` guard runs *after* `buildPayload`, so the child builds a full payload and then throws it away. `buildPayload` is actually expensive (~150–200ms, mostly loading the app metadata hooks), so checking the command name before building would skip real work.

packages/cli-kit/src/public/node/analytics.ts
gonzaloriestra: Good catch! App dev is now sync.

packages/cli-kit/src/public/node/analytics.ts
gonzaloriestra: You are right, it doesn't make sense to run the CLI command in that case. It's now using `sendAnalyticsEvent` directly.

packages/cli-kit/src/public/node/analytics.ts
gonzaloriestra: Fixed, thanks!

isaacroldan:
I think we should replace these command-name and launcher exceptions with command metadata.

For example, add `static requiresSyncAnalytics = false` to `BaseCommand`, then set it to `true` on `app init` and `app dev`. We can capture that value in `startAnalytics`, where the command class is available, and store it with the internal command metadata so that `reportAnalyticsEvent` can select synchronous delivery.

This also covers `app dev stop`: it is derived from the `app dev` command through `analyticsStopCommand()`, so it can keep the same delivery requirement.

This keeps command lifecycle knowledge on the command and removes the `create-app`, `app dev`, and `app dev stop` checks from the analytics transport.

gonzaloriestra:
@isaacroldan Good point. That's something we need to improve for other things like notifications, but it's better to do it now. Fixed!

isaacroldan (commented):
One stale change remains from the earlier file-based implementation.

---
🤖 *Generated by the review-pr skill*

packages/cli-kit/src/public/node/fs.ts
```diff
@@ -211,6 +211,8 @@ export function appendFileSync(path: string, data: string): void {
 
 export interface WriteOptions {
   encoding: BufferEncoding
+  mode?: number
+  flag?: string
```
isaacroldan: These fields appear to be left over from the earlier temporary-file transport. The current implementation passes the analytics payload through stdin and does not use them. Can we remove them from this PR?

---
🤖 *Generated by the review-pr skill*

packages/cli-kit/src/public/node/analytics.ts
```diff
@@ -65,40 +129,39 @@ export async function reportAnalyticsEvent(options: ReportAnalyticsEventOptions)
...
-        },
-        {
-          active,
-          network,
-          prompt,
-        },
-      )
+    const {platformAndArch} = await import('./os.js')
+    const sendInBackground =
+      !commandStartOptions?.requiresSyncAnalytics && platformAndArch().platform !== 'windows' && !ciPlatform().isCI
```
isaacroldan: I think we missed two more commands that need synchronous analytics:

- `theme dev` awaits `reportAnalyticsEvent()` and immediately calls `process.exit(0)`, so it can exit before the child has consumed all of stdin.
- `app function replay` runs `treeKill` two seconds after abort. The detached child still has the CLI as its PPID, so it can be found and killed before it finishes.

Can we set `requiresSyncAnalytics` on both command classes?

---
🤖 *Generated by the review-pr skill*

isaacroldan (commented):
Two more command shutdown paths need synchronous analytics.

---
🤖 *Generated by the review-pr skill*

packages/cli-kit/src/public/node/fs.ts
gonzaloriestra: Right, removed!

packages/cli-kit/src/public/node/analytics.ts
gonzaloriestra: Fixed!

isaacroldan: Can we keep analytics synchronous when running inside a container too? An ephemeral container exits with its entrypoint and kills the detached process before it can finish sending the event. We already have isInsideContainer() in system.ts, so we can reuse it in this condition and add a test for the synchronous container path.

---
🤖 *Generated by the review-pr skill*

isaacroldan (commented):
One remaining issue with background delivery.

---
🤖 *Generated by the review-pr skill*

packages/cli-kit/src/public/node/system.ts
```diff
@@ -278,11 +278,12 @@ function buildExec(
...
+  const backgroundStdio = options?.input === undefined ? 'ignore' : (['pipe', 'ignore', 'ignore'] as const)
   const commandProcess = execa(command, args, {
     env,
     cwd: executionCwd,
     input: options?.input,
-    stdio: options?.background ? 'ignore' : options?.stdio,
+    stdio: options?.background ? backgroundStdio : options?.stdio,
     stdin: options?.stdin,
     stdout: options?.stdout === 'inherit' ? 'inherit' : undefined,
     stderr: options?.stderr === 'inherit' ? 'inherit' : undefined,
```
isaacroldan, line 289: Not exactly introduced here, but we shouldn't be using both `stdio` and `stdin/stdout/stderr`, the first one basically covers all other 3. And accidentally passing values to stdio and any other property will fail.

gonzaloriestra: Well, those are the options provided by Execa and we should avoid changing our public interface. If we explicitly pass both `stdio` and `stdout`, Execa will fail immediately, so we should detect it at the moment, no?

packages/cli-kit/src/public/node/analytics.ts
gonzaloriestra: Makes sense, fixed!

packages/cli-kit/src/public/node/system.ts
isaacroldan: Yeah, so far no code path passes both properties, so it's safe, but it's weird that we allow both and `background` only affects `stdio` and not the others. So someone passing ` {background: true, stdin: 'inherit'}` will crash.   
  
We could find a way to just use one or the other, but it's ok for now

isaacroldan (approved):
Approving! great work, wasn't an easy change. We should keep an eye on analytics after the next release

EvilGenius13 (approved):

PR #7618, opened 2026-05-23
[Performance] Memoize platformAndArch and use startsWith

gonzaloriestra:
Optimizes `platformAndArch` by adding memoization for the default case and using a more efficient string check for Windows detection. This reduces overhead in frequent operations like analytics reporting.

---
*PR created automatically by Jules for task [15461649029740018611](https://jules.google.com/task/15461649029740018611) started by @gonzaloriestra*


isaacroldan (changes requested):
We are adding complexity for no real benefit here. We save a string match but add memoization and other string checks. The complexity is not worth it, we are literally saving 0ms: `platformAndArch` is called onlye once or maybe maybe twice in a normal run.

gonzaloriestra:
I wanted to check how many times this was used before sharing it, but I forgot about it... You are right that it's only 1 or 2, so it's not worth it. Thanks!

PR #7623, opened 2026-05-24
[Security] Harden random selection in takeRandomFromArray

gonzaloriestra:
### Why is this change needed?

The `takeRandomFromArray` utility was using `Math.random()`, which is not cryptographically secure. While its current primary use is for generating random names for new app/theme directories, using a CSPRNG is a better default for any random selection utility in the CLI.

### How to test your changes?

CI

### Post-release steps

None.

---
*PR created automatically by Jules for task [8018340101932167753](https://jules.google.com/task/8018340101932167753) started by @gonzaloriestra*


isaacroldan (changes requested):
I think this is not necessary, we don't need this to be "cryptographically secure". Adds an extra import for no real benefit

gonzaloriestra:
Yeah, for the current usage is not needed... The change is small, but you are right that it's not worth it.

PR #7626, opened 2026-05-25
[Performance] Optimize unstyled utility

gonzaloriestra:

The `unstyled` utility is used frequently in CLI operations, especially during logging and terminal layout calculations. The underlying `stripAnsi` function uses a complex regular expression that can be computationally expensive when called repeatedly on plain strings.

Optimizes the `unstyled` function in `packages/cli-kit/src/public/node/output.ts` by adding an early return: `if (!message.includes('\u001b')) return message`. This avoids the regex overhead for any string that doesn't contain ANSI escape codes.

### How to test your changes?

CI

### Post-release steps

None

### Checklist

---
*PR created automatically by Jules for task [6608015497183229439](https://jules.google.com/task/6608015497183229439) started by @gonzaloriestra*

### DIFF
--- a/packages/cli-kit/src/public/node/output.ts
+++ b/packages/cli-kit/src/public/node/output.ts
@@ -402,6 +402,10 @@ export function outputWhereAppropriate(logLevel: LogLevel, logger: Logger, messa
  * @returns The message without styles.
  */
 export function unstyled(message: string): string {
+  // Optimization: skip regex execution for strings that don't have ANSI escape codes.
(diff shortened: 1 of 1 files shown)


isaacroldan (approved):
I'm pretty sure that we don't use this in a way that would have a meaningful performance impact, but ok

gonzaloriestra:
Right, the improvement will be minimal. It will probably save a few ms only for outputs in non-TTY (no color). But the change is very simple.

PR #7633, opened 2026-05-25
Allow theme profile with Theme Access tokens

gonzaloriestra:
### WHY are these changes introduced?

The interactivity audit found that theme profile required browser OAuth even when a Theme Access token was available for non-interactive runs.

### WHAT is this pull request doing?

- Allows theme profile to use Theme Access tokens.
- Keeps the existing rejection for raw Admin API tokens with clearer guidance.

### How to test your changes?

- pnpm vitest run packages/theme/src/cli/services/profile.test.ts

### Checklist

- [ ] I've considered possible cross-platform impacts (Mac, Linux, Windows)
- [ ] I've considered possible documentation changes
- [ ] I've considered analytics changes to measure impact
- [ ] The change is user-facing, so I've added a changelog entry with pnpm changeset add

PR #7632, opened 2026-05-25
Add non-interactive app dev prompt bypasses

gonzaloriestra:
### WHY are these changes introduced?

Related to https://github.com/shop/issues-develop/issues/22868

The interactivity audit found several `app dev` paths that still required prompts or manual recovery in automation:

- URL update confirmation for linked dev-session apps.
- mkcert installation when running with localhost certificates.
- Storefront password entry for theme app extension preview on password-protected storefronts.

The transfer-disabled store conversion prompt is removed since the CLI only allows using dev stores now that should already be transfer-disabled.

### WHAT is this pull request doing?

- Adds explicit `app dev` flags for non-interactive flows:
    - `--install-mkcert` / `SHOPIFY_FLAG_INSTALL_MKCERT` controls localhost certificate installation.
    - `--store-password` / `SHOPIFY_FLAG_STORE_PASSWORD` supplies the storefront password needed by theme app extensions.
- Defaults linked dev-session apps to `automatically_update_urls_on_dev = true`, avoiding the app dev URL prompt for that flow.
- Removes the transfer-disabled store conversion prompt from `app dev`; transfer-enabled dev stores now fail with an actionable error instead of prompting to mutate the store.
- Redacts `--store-password` and `SHOPIFY_FLAG_STORE_PASSWORD` values from analytics payloads.

### How to test your changes?

- `shopify app dev --use-localhost --install-mkcert`
- Add a theme app extension and run: `shopify app dev --store-password yourpassword`

### Checklist

- [x] I've considered possible cross-platform impacts (Mac, Linux, Windows)
- [x] I've considered possible documentation changes
- [x] I've considered analytics changes to measure impact
- [x] The change is user-facing, so I've added a changelog entry with `pnpm changeset add`

PR #7631, opened 2026-05-25
Allow app config validate to target config by client ID

gonzaloriestra:
### WHY are these changes introduced?

The interactivity audit found that `shopify app config validate --client-id <id>` can still prompt while resolving the active local config, especially when the cached config points at a file that no longer exists.

### WHAT is this pull request doing?

- Skips active-config recovery prompts during the validation preflight when `--client-id` is provided.
- Preserves the existing interactive config recovery behavior when `--client-id` is not provided.

### How to test your changes?

Remove the current config and pass a client from another one. It should validate without asking for the config.

`shopify app config validate --client-id`

### Checklist

- [x] I've considered possible cross-platform impacts (Mac, Linux, Windows)
- [x] I've considered possible documentation changes
- [x] I've considered analytics changes to measure impact
- [ ] The change is user-facing, so I've added a changelog entry with `pnpm changeset add`

PR #7630, opened 2026-05-25
Add flags to avoid prompts in app config link

gonzaloriestra:
### WHY are these changes introduced?

Related to https://github.com/shop/issues-develop/issues/22869

The interactivity audit found that `shopify app config link --client-id <id>` can still prompt for a local configuration file name when the target file cannot be inferred.

### WHAT is this pull request doing?

- Adds `--file-name` to choose the config file that `app config link` should create.
- Adds `--force` to overwrite an existing config file without prompting.

### How to test your changes?

With a new client-id (not used in any toml file) and an existing shopify.app.toml:

- `shopify app config link --client-id XXX` => should ask for the file
- `shopify app config link --client-id XXX --file-name staging` => should create shopify.app.staging.toml
- `shopify app config link --client-id XXX --file-name shopify.app.toml` => already exists error
- `shopify app config link --client-id XXX --file-name shopify.app.toml --force` => works

### Checklist

- [x] I've considered possible cross-platform impacts (Mac, Linux, Windows)
- [x] I've considered possible documentation changes
- [x] I've considered analytics changes to measure impact
- [x] The change is user-facing, so I've added a changelog entry with `pnpm changeset add`

PR #7634, opened 2026-05-25
Add theme dev reconciliation strategy flag

gonzaloriestra:

Fixes https://github.com/shop/issues-develop/issues/22873

The interactivity audit found that theme dev with --theme-editor-sync could still prompt when JSON files needed reconciliation.

- Adds --reconciliation-strategy with keep-local, keep-remote, and abort options.
- Threads the strategy through theme editor sync reconciliation and refreshes generated command docs.

### How to test your changes?

Make a change in a theme file (like templates/404.json) locally and remotely and run:

- shopify theme dev --theme-editor-sync
- shopify theme dev --theme-editor-sync --reconciliation-strategy keep-local
- shopify theme dev --theme-editor-sync --reconciliation-strategy keep-remote
- shopify theme dev --theme-editor-sync --reconciliation-strategy abort

### Checklist

- [x] The change is user-facing, so I've added a changelog entry with pnpm changeset add


gonzaloriestra:
* **#7687** <a href="https://app.graphite.com/github/pr/Shopify/cli/7687?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#7634** <a href="https://app.graphite.com/github/pr/Shopify/cli/7634?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Shopify/cli/7634?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#7632** <a href="https://app.graphite.com/github/pr/Shopify/cli/7632?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#7633](https://github.com/Shopify/cli/pull/7633) <a href="https://app.graphite.com/github/pr/Shopify/cli/7633?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#7631** <a href="https://app.graphite.com/github/pr/Shopify/cli/7631?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#7630** <a href="https://app.graphite.com/github/pr/Shopify/cli/7630?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

packages/theme/src/cli/commands/theme/dev.ts
```diff
@@ -82,6 +82,12 @@ You can run this command only in a directory that matches the [default Shopify t
       env: 'SHOPIFY_FLAG_STANDARD_EVENTS_INSPECTOR',
       default: false,
     }),
+    'reconciliation-strategy': Flags.string({
```
lucyxiang, line 85: `theme dev --reconciliation-strategy keep-local` on its own does nothing and emits no warning, shall we add dependency on `theme-editor-sync`?

**Suggestion:** Consider declaring the dependency so oclif surfaces an error instead of silently ignoring the flag:
```suggestion
    'reconciliation-strategy': Flags.string({
      description:
        'How to resolve JSON conflicts when --theme-editor-sync is enabled. Use keep-local to keep local files, keep-remote to keep remote files, or abort to fail instead of prompting.',
      options: ['keep-local', 'keep-remote', 'abort'],
      env: 'SHOPIFY_FLAG_RECONCILIATION_STRATEGY',
      dependsOn: ['theme-editor-sync'],
    }),
```

packages/theme/src/cli/utilities/theme-environment/theme-reconciliation.test.ts
```diff
@@ -104,6 +104,26 @@ describe('reconcileJsonFiles', () => {
   })
 
   describe('files only present on remote developer theme', () => {
+    test('should download assets from remote theme without prompting when keep-remote strategy is selected', async () => {
```
lucyxiang, line 107: Can we check that local files do get deleted

packages/theme/src/cli/utilities/theme-environment/theme-reconciliation.test.ts
```diff
@@ -220,6 +240,46 @@ describe('reconcileJsonFiles', () => {
   })
 
   describe('files with conflicting checksums', () => {
+    test('should keep local files without prompting when keep-local strategy is selected', async () => {
```
lucyxiang, line 282: Similarly check than remote files are deleted

lucyxiang (commented):
---

packages/theme/src/cli/utilities/theme-environment/theme-reconciliation.ts
```diff
@@ -158,6 +167,34 @@ async function promptFileReconciliationStrategy(
...
+      options.remote.callback(files)
+      break
+    case 'keep-local':
+      options.local.callback(files)
+      break
+    case 'abort':
+      throw new AbortError(
+        'Theme JSON files need reconciliation.',
+        `${title}\n${files.map((file) => `- ${file.key}`).join('\n')}`,
+      )
```
lucyxiang: How about collecting all files and throwing one AbortError that lists every file needing reconciliation, rather than throwing inside the per-category callback on the first hit.

.changeset/theme-dev-reconciliation-strategy.md
```diff
@@ -0,0 +1,5 @@
+---
+'@shopify/theme': patch
```
karreiro: Given we're updating flag interfaces, I'd consider this a `minor`

karreiro (commented):
Thank you for this PR, @gonzaloriestra!

What if we repurpose the `--theme-editor-sync` flag, since we're defining the sync strategy? We could have something like this:

```
shopify theme dev --theme-editor-sync ask [default]
shopify theme dev --theme-editor-sync keep-local
shopify theme dev --theme-editor-sync keep-remote
shopify theme dev --theme-editor-sync abort
```

This would be a backward-compatible change, since `ask` would be the default value. It also reduces the blast radius of the change and keeps the `dev` command simpler.

Thanks again for the PR!

gonzaloriestra:
@karreiro I like the idea, but Oclif doesn't support default values like that. A default value can be used when the flag is omitted, but if you just add `--theme-editor-sync`, it would fail because it expects an additional string (it's not a boolean).

Maybe there's a hacky way to do it, but I think I prefer to just add an additional flag. Wdyt?

packages/theme/src/cli/commands/theme/dev.ts
gonzaloriestra: Done. `--reconciliation-strategy` now declares `dependsOn: ["theme-editor-sync"]`, so oclif surfaces an error when the strategy flag is used without theme editor sync.

packages/theme/src/cli/utilities/theme-environment/theme-reconciliation.test.ts
gonzaloriestra: Done. Added coverage for the `keep-remote` path with local-only files, including an assertion that the local file is deleted.

packages/theme/src/cli/utilities/theme-environment/theme-reconciliation.test.ts
gonzaloriestra: Done. `keep-local` now queues conflicting remote files for deletion, and the tests cover both the flag-driven and prompt-selected local paths.

packages/theme/src/cli/utilities/theme-environment/theme-reconciliation.ts
gonzaloriestra: Done. Abort now collects local-only, remote-only, and conflicting files into a single `AbortError`, and the test asserts all file names are included.

karreiro (approved):
Thank you for this PR, @gonzaloriestra!

I looked into it, and the alternatives for implementing the single-flag approach are indeed hacky. It also reveals that having union types (`boolean | string`) in CLI flags might not be the ideal pattern. So, from my side, +1 for proceeding with this approach.

Thanks again for this PR!

isaacroldan (approved):

PR #7639, opened 2026-05-26
Handle missing Git in optional repo checks

charlespwd:
## Why

Some theme workflows use Git only as an optional safety check before continuing. When Git is not installed or unavailable on PATH, those checks should not surface raw command failures to users.

## What changed

- Treat missing Git as “not inside a Git directory” for optional repository detection.
- Keep a friendly, actionable error for flows that explicitly require Git.
- Added unit coverage for both behaviours.

## Testing

- `pnpm vitest run packages/cli-kit/src/public/node/git.test.ts --config packages/cli-kit/vite.config.ts`
- `pnpm vitest run packages/theme/src/cli/services/pull.test.ts --config packages/theme/vite.config.ts`

Fixes shop/issues#47568

### DIFF
--- a/.changeset/old-poems-speak.md
+++ b/.changeset/old-poems-speak.md
@@ -1,0 +1,5 @@
+---
+'@shopify/cli-kit': patch
+---
+
+Fix error reporting of cli-kit methods that try to run git when git is unavailable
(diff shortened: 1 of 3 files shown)


karreiro (approved):
Thank you, @charlespwd!

gonzaloriestra (approved):

PR #7642, opened 2026-05-26
fix(theme): preserve dev preview theme across SFR redirects

EvilGenius13:

The storefront renderer fetch call was using fetch's default redirect behavior ('follow'), which silently dropped the intermediate Set-Cookie headers that bind _shopify_essential to the dev theme.

Set redirect: 'manual' on storefront-renderer fetches and patch response headers (Set-Cookie capture + Location rewrite) on 3xx so the browser can follow the redirect cleanly while we keep the dev-bound session cookie.

on 3xx so the browser can follow the redirect cleanly while we keep the dev-bound session cookie.

### Checklist

### DIFF
--- a/.changeset/tidy-bats-prove.md
+++ b/.changeset/tidy-bats-prove.md
@@ -1,0 +1,5 @@
+---
+'@shopify/theme': patch
+---
+
+Fix `theme dev` preview occasionally rendering the live theme by preserving the Shopify `_shopify_essential` cookie in redirects
--- a/packages/theme/src/cli/utilities/theme-environment/proxy.ts
+++ b/packages/theme/src/cli/utilities/theme-environment/proxy.ts
@@ -194,12 +194,12 @@ export async function patchRenderingResponse(
   rawResponse: Response,
   patchCallback?: (html: string) => string | undefined,
 ): Promise<Response> {
-  // 3xx responses should be returned
-  if (rawResponse.status >= 300 && rawResponse.status < 400) {
-    return rawResponse
-  }
-
   const response = patchProxiedResponseHeaders(ctx, rawResponse)
+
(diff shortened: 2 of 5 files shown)


packages/theme/src/cli/utilities/theme-environment/proxy.ts
```diff
@@ -194,13 +194,13 @@ export async function patchRenderingResponse(
   rawResponse: Response,
   patchCallback?: (html: string) => string | undefined,
 ): Promise<Response> {
+  const response = patchProxiedResponseHeaders(ctx, rawResponse)
```
EvilGenius13, line 197: Flipping this to happen earlier makes sure the cookie is available to our session for any following requests.

packages/theme/src/cli/utilities/theme-environment/storefront-renderer.ts
```diff
@@ -23,6 +23,7 @@ export async function render(session: DevServerSession, context: DevServerRender
     response = await fetch(url, {
       method: 'POST',
       body: bodyParams,
+      redirect: 'manual',
```
EvilGenius13, line 26: This makes the cookie visible by not having node drop it on a redirect.

karreiro (approved):
Thank you, @EvilGenius13!

.changeset/tidy-bats-prove.md
```diff
@@ -0,0 +1,5 @@
+---
+'@shopify/theme': patch
+---
+
+Fix `theme dev` preview occasionally rendering the live theme by preserving shopify cookie in redirects
```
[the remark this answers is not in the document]
EvilGenius13: Fixed

PR #7652, opened 2026-05-27
Improve missing theme scope error message

charlespwd:

No public issue.

When theme commands use an account or token that cannot access Online Store themes, the Admin GraphQL API returns an `ACCESS_DENIED` error for the `themes` field. The CLI currently surfaces the raw GraphQL `ClientError` payload, which makes the actionable fix hard to find.

The high-volume observed case is a user-supplied theme password/token path, not a CLI-generated token path. Theme commands accept this value through `--password`, `SHOPIFY_CLI_THEME_TOKEN`, or a `password` entry in `shopify.theme.toml`. If that value is a custom app Admin API access token, the custom app owner/admin controls its scopes in Shopify Admin. When the custom app is missing `read_themes`, Shopify correctly denies the `themes` field.

If no password/token is supplied, the CLI uses the normal Shopify account OAuth flow and requests theme access. In that path, this error generally means the authenticated staff or collaborator account cannot access Online Store themes, so the fix is account permission/reauth rather than a missing custom app scope.

So the underlying failure is usually user/admin configuration. The CLI bug is that we present it like an unexpected GraphQL failure instead of an expected, actionable permission error.

`fetchThemes` now detects Admin GraphQL `ACCESS_DENIED` errors that include a required access scope and rethrows them as a friendly `AbortError`.

Before, users saw a raw response/request JSON blob like:

```text
Access denied for themes field. Required access: `read_themes` access scope.: {"response":{...},"request":{...}}
```

After, users see a direct explanation and next step:

```text
The authenticated account or access token is missing `read_themes` access scope.

If you authenticated with a custom app Admin API access token, open the custom app in your Shopify admin, add the required theme access scopes, reinstall the app, and use the new access token. For theme pull, theme list, and theme info, add `read_themes`. For theme push and theme dev, add both `read_themes` and `write_themes`. If you authenticated with your Shopify account, make sure your staff or collaborator account can access Online Store themes, then run `shopify auth logout` and try again. See https://shopify.dev/api/usage/access-scopes.
```

### How to test your changes?

- [x] `pnpm --filter @shopify/cli-kit vitest run src/public/node/themes/api.test.ts`

### Post-release steps

No post-release steps.

### Checklist

---

Fixes shop/issues#32995

### DIFF
--- a/packages/cli-kit/src/public/node/themes/api.ts
+++ b/packages/cli-kit/src/public/node/themes/api.ts
@@ -25,11 +25,13 @@ import {GetTheme} from '../../../cli/api/graphql/admin/generated/get_theme.js'
 import {FindDevelopmentThemeByName} from '../../../cli/api/graphql/admin/generated/find_development_theme_by_name.js'
 import {OnlineStorePasswordProtection} from '../../../cli/api/graphql/admin/generated/online_store_password_protection.js'
 import {RequestModeInput} from '../http.js'
-import {adminRequestDoc} from '../api/admin.js'
+import {adminRequestDoc, type AdminRequestOptions} from '../api/admin.js'
 import {AdminSession} from '../session.js'
 import {AbortError} from '../error.js'
 import {outputDebug} from '../output.js'
 import {recordTiming, recordEvent, recordError} from '../analytics.js'
+import {ClientError, type Variables} from 'graphql-request'
+import type {InlineToken, TokenItem} from '../ui.js'
 
 export type ThemeParams = Partial<Pick<Theme, 'name' | 'role' | 'processing' | 'src'>>
 export type AssetParams = Pick<ThemeAsset, 'key'> & Partial<Pick<ThemeAsset, 'value' | 'attachment'>>
@@ -40,6 +42,7 @@ const THEME_API_NETWORK_BEHAVIOUR: RequestModeInput = {
   maxRetryTimeMs: 90 * 1000,
   recordCommandRetries: true,
 }
+const DEFAULT_THEME_ACCESS_REQUIREMENT = 'the required theme access scope'
 
 export async function fetchTheme(id: number, session: AdminSession): Promise<Theme | undefined> {
   const gid = composeThemeGid(id)
@@ -65,6 +68,7 @@ export async function fetchTheme(id: number, session: AdminSession): Promise<The
 
     // eslint-disable-next-line no-catch-all/no-catch-all
   } catch (error) {
+    abortIfMissingThemeAccessScope(error)
     /**
      * Consumers of this and other theme APIs in this file expect either a theme
      * or `undefined`.
@@ -84,13 +88,14 @@ export async function fetchThemes(session: AdminSession): Promise<Theme[]> {
 
   while (true) {
     // eslint-disable-next-line no-await-in-loop
-    const response = await adminRequestDoc({
+    const response = await requestThemeAdminDoc({
       query: GetThemes,
       session,
       variables: {after},
       responseOptions: {handleErrors: false},
       preferredBehaviour: THEME_API_NETWORK_BEHAVIOUR,
     })
+
     if (!response.themes) {
       unexpectedGraphQLError('Failed to fetch themes')
     }
@@ -117,7 +122,7 @@ export async function fetchThemes(session: AdminSession): Promise<Theme[]> {
 export async function findDevelopmentThemeByName(name: string, session: AdminSession): Promise<Theme | undefined> {
   recordEvent('theme-api:find-development-theme-by-name')
 
-  const {themes} = await adminRequestDoc({
+  const {themes} = await requestThemeAdminDoc({
     query: FindDevelopmentThemeByName,
     session,
     variables: {name},
@@ -188,7 +193,7 @@ export async function fetchThemeAssets(id: number, filenames: Key[], session: Ad
 
   while (true) {
     // eslint-disable-next-line no-await-in-loop
-    const response = await adminRequestDoc({
+    const response = await requestThemeAdminDoc({
       query: GetThemeFileBodies,
       session,
       variables: {id: themeGid(id), filenames, after},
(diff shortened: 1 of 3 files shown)


packages/cli-kit/src/public/node/themes/api.ts
```diff
@@ -606,6 +611,37 @@ function unexpectedGraphQLError(message: string): never {
...
 
+function abortIfMissingThemeAccessScope(error: unknown): void {
+  if (!(error instanceof ClientError)) return
+
+  const requiredAccess = getRequiredAccessForAccessDeniedError(error)
+  if (!requiredAccess) return
+
+  const tryMessage = [
+    'If you authenticated with a custom app Admin API access token, open the custom app in your Shopify admin,',
+    'add the required theme access scopes, reinstall the app, and use the new access token.',
```
graygilmore: With the announcement about [custom apps](https://changelog.shopify.com/posts/legacy-custom-apps-can-t-be-created-after-january-1-2026) we might want to change the language here.

packages/cli-kit/src/public/node/themes/api.ts
graygilmore, line 1: There are quite a few conditionals in this new logic that aren't covered by the test. We don't need to add all (e.g. don't add a test asserting that the message doesn't appear for other exceptions) but some extra coverage would be good.

packages/cli-kit/src/public/node/themes/api.ts
```diff
@@ -90,7 +91,11 @@ export async function fetchThemes(session: AdminSession): Promise<Theme[]> {
       variables: {after},
       responseOptions: {handleErrors: false},
       preferredBehaviour: THEME_API_NETWORK_BEHAVIOUR,
+    }).catch((error: unknown) => {
```
graygilmore: Should we apply this same logic to `findDevelopmentThemeByName`? Are there any other theme fetches that we could drop this into?

packages/cli-kit/src/public/node/themes/api.ts
```diff
@@ -606,6 +611,37 @@ function unexpectedGraphQLError(message: string): never {
...
+
+function getRequiredAccessForAccessDeniedError(error: ClientError): string | undefined {
+  const graphQLErrors = error.response.errors
+  if (!Array.isArray(graphQLErrors)) return undefined
+
+  const accessDeniedError = graphQLErrors.find((graphQLError) => graphQLError.extensions?.code === 'ACCESS_DENIED')
+  const requiredAccess = accessDeniedError?.extensions?.requiredAccess
+  if (typeof requiredAccess !== 'string') return undefined
+
+  return requiredAccess.replace(/\.$/, '')
```
graygilmore: I worry that there's some fragility/assumptions here where we're relying on the shape of the translation from the API but I don't have a good alternative suggestion.

packages/cli-kit/src/public/node/themes/api.ts
graygilmore: (BOT COMMENT)

🎨 **Code Style**: Every other `adminRequestDoc` call in this file uses a bare `await`; the `.catch()` chained directly onto the awaited promise is the only instance of that pattern. A try/catch block reads more naturally for the intent (intercept an error, optionally translate it, otherwise rethrow) and doesn't rely on the reader verifying that `abortIfMissingThemeAccessScope` either throws or returns void.

**Suggestion:** Consider restructuring as:

```ts
let response
try {
  response = await adminRequestDoc({
    query: GetThemes,
    session,
    variables: {after},
    responseOptions: {handleErrors: false},
    preferredBehaviour: THEME_API_NETWORK_BEHAVIOUR,
  })
} catch (error) {
  abortIfMissingThemeAccessScope(error)
  throw error
}
```

graygilmore (commented):
Love this! Couple of small things.

packages/cli-kit/src/public/node/themes/api.ts
```diff
@@ -606,6 +610,50 @@ function unexpectedGraphQLError(message: string): never {
...
+    'If you authenticated with an Admin API access token, update the app or integration that issued the token to include the required theme access scopes, then reauthorize it or generate a new token.',
+    'For theme pull, theme list, and theme info, add `read_themes`.',
+    'For theme push and theme dev, add both `read_themes` and `write_themes`.',
+    'If you authenticated with your Shopify account, make sure your staff or collaborator account can access Online Store themes, then run `shopify auth logout` and try again.',
+    'See https://shopify.dev/api/usage/access-scopes.',
+  ].join(' ')
+
+  throw recordError(
+    new AbortError(`The authenticated account or access token is missing ${requiredAccess}.`, tryMessage),
+  )
```
aswamy:
```suggestion
  const nextSteps: TokenItem<InlineToken>[] = [
    [
      'If you authenticated with an Admin API access token, update the app or integration that issued the token to include the required theme access scopes, then reauthorize it or generate a new token.',
    ],
    [
      'For',
      {command: 'theme pull'},
      {char: ','},
      {command: 'theme list'},
      {char: ','},
      'and',
      {command: 'theme info'},
      {char: ','},
      'add the',
      {command: 'read_themes'},
      'scope',
      {char: '.'},
    ],
    [
      'For',
      {command: 'theme push'},
      'and',
      {command: 'theme dev'},
      {char: ','},
      'add both the',
      {command: 'read_themes'},
      'and',
      {command: 'write_themes'},
      'scopes',
      {char: '.'},
    ],
    [
      'If you authenticated with your Shopify account, make sure your staff or collaborator account can access Online Store themes, then run',
      {command: 'shopify auth logout'},
      'and try again',
      {char: '.'},
    ],
    ['See', {link: {label: 'Shopify access scopes', url: 'https://shopify.dev/api/usage/access-scopes'}}, {char: '.'}],
  ]

  throw recordError(
    new AbortError(`The authenticated account or access token is missing ${requiredAccess}.`, undefined, nextSteps),
  )
```

Can we make this visually a little easier to read with bullet points 😅 

### Before

Image

### After

Image

aswamy (approved):
Looks good 👍 

Tophatted and had a tiny nit, but not a blocker.

PR #7010, opened 2026-03-13
Faster startup: Lazy command loading

gonzaloriestra:
### WHY are these changes introduced?

The CLI requires too much time to start working, it feels slow

### WHAT is this pull request doing?

Splits the monolithic `index.ts` (which eagerly imports all 106 commands and their dependency trees) into a lightweight `bootstrap.ts` entry point and a lazy `command-registry.ts`. Commands are loaded on-demand via a `LazyCommandLoader` passed to `ShopifyConfig.runCommand()`.

- **Separate bootstrap from index.ts**: `bootstrap.ts` sets up global proxy, signal handlers, and calls `runCLI` — it does NOT import any command modules
- **Lazy command registry**: `command-registry.ts` maps command IDs to dynamic `import()` expressions, so only the invoked command's module is loaded
- **Custom** **`runCommand`** **override in** **`ShopifyConfig`**: when a `LazyCommandLoader` is set, `runCommand` bypasses oclif's default `cmd.load()` (which would pull in the entire `index.ts`) and directly runs the lazily-loaded command class
- **Individual hook files**: hooks are moved to separate files (`hooks/app-init.ts`, `hooks/did-you-mean.ts`, etc.) so oclif loads them independently instead of through `index.ts`
- **Package exports for hooks**: `@shopify/app` exports hooks via subpath entries (`./hooks/public-metadata`, `./hooks/sensitive-metadata`) so they can be imported individually
- **Deferred error-handler import**: `errorHandler` is imported lazily in the catch block of `launchCLI`, removing it from the critical startup path

### How to test your changes?

- `shopify version` / `shopify help` return correct output
- `shopify app dev` and other commands work normally

### Measuring impact

- [x] n/a - this doesn't need measurement, e.g. a linting rule or a bug-fix

### Checklist

- [x] I've considered possible cross-platform impacts (Mac, Linux, Windows)
- [x] I've considered possible [documentation](https://shopify.dev) changes

PR #7662, opened 2026-05-28
Use validated storefront password for theme apps

graygilmore:
- [Community post](https://community.shopify.dev/t/shopify-cli-4-1-0-keeps-asking-for-store-password/34735/2)

`ensureValidPassword` now treats an empty string as a provided password (https://github.com/Shopify/cli/pull/7010), so passing `''` from the `app dev` flow skips the normal prompt/local storage fallback.

Pass `undefined` instead so password-protected stores use the same validation path as `theme dev`, and keep the returned password on the theme app extension process options.

### How to test your changes?

1. Run `shopify app dev` against a shop that has a storefront password
2. Enter the password
3. Ensure that the server runs normally
4. Stop the server and start it again
5. Ensure that you are not prompted for the password again (ie it should not say "Incorrect store password")

### DIFF
--- a/.changeset/strict-toys-try.md
+++ b/.changeset/strict-toys-try.md
@@ -1,0 +1,5 @@
+---
+'@shopify/app': patch
+---
+
+Fix `app dev` always prompting for storefront password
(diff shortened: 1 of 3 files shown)


EvilGenius13 (approved):
🎩 on current CLI vs the snapshot, works great!

dmerand (approved):

PR #7668, opened 2026-05-29
Open auth browser without keypress prompt

gonzaloriestra:

Fixes https://github.com/shop/issues-develop/issues/22867

The login flow currently pauses in interactive terminals and asks users to press a key before Shopify CLI opens the browser. We want authentication to start directly by opening the browser, making it easier for agents.

- Removes the `keypress()` wait and "Press any key" message from the interactive device auth flow.
- Keeps the existing manual-link fallback when the browser can't be opened.

### How to test your changes?

`shopify auth login`

### Checklist

### DIFF
--- a/packages/cli-kit/src/private/node/session/device-authorization.ts
+++ b/packages/cli-kit/src/private/node/session/device-authorization.ts
@@ -5,9 +5,7 @@ import {identityFqdn} from '../../../public/node/context/fqdn.js'
 import {shopifyFetch} from '../../../public/node/http.js'
 import {outputContent, outputDebug, outputInfo, outputToken} from '../../../public/node/output.js'
 import {AbortError, BugError} from '../../../public/node/error.js'
-import {isCloudEnvironment} from '../../../public/node/context/local.js'
 import {isCI, openURL} from '../../../public/node/system.js'
-import {isTTY, keypress} from '../../../public/node/ui.js'
 
 import {Response} from 'node-fetch'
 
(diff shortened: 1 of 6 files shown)


packages/cli-kit/src/private/node/session/device-authorization.ts
```diff
@@ -81,21 +79,11 @@ export async function requestDeviceAuthorization(scopes: string[]): Promise<Devi
   outputInfo(outputContent`User verification code: ${jsonResult.user_code}`)
   const linkToken = outputToken.link(jsonResult.verification_uri_complete)
 
-  const cloudMessage = () => {
-    outputInfo(outputContent`👉 Open this link to start the auth process: ${linkToken}`)
-  }
-
-  if (isCloudEnvironment() || !isTTY()) {
-    cloudMessage()
+  const opened = await openURL(jsonResult.verification_uri_complete)
```
dmerand, line 82: Should we harden `openURL` to more reliably return `false` in cloud environments? I guess non-TTY isn't a good enough signal (agentic flows are non-TTY) but I'm wondering about situations where `open` may just hang.

gonzaloriestra: Apparently it won't hang, but it may return `true` even when there is no browser. In that case, it will also print the URL, but saying that it was opened, so it's not a big deal.

Anyway, I added a check [here](https://github.com/Shopify/cli/pull/7668/changes#diff-7271ed460c5ae25bff3018dbf798d5fab57407b99dbb894f044b66c69cf23b8eR55) to return `false` in cloud environments.

alfonso-noriega (approved):

gonzaloriestra:
### Merge activity

* **Jun 12, 9:27 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/Shopify/cli/7668).
* **Jun 12, 9:27 AM UTC**: @gonzaloriestra added this pull request to the GitHub merge queue with [Graphite](https://app.graphite.com/github/pr/Shopify/cli/7668).

PR #6833, opened 2026-02-09
Fix theme editor sync not reloading on schema changes

kdaviduik:
### WHY are these changes introduced?

Closes https://github.com/Shopify/developer-tools-team/issues/1033.

When using `shopify theme dev --theme-editor-sync`, making schema changes to section files (e.g., adding/modifying settings in `{% schema %}` blocks) did not trigger the theme customizer to reload. Users had to manually refresh the customizer to see schema updates.

This is a regression since version 3.88.1.

### WHAT is this pull request doing?

Bumps `@shopify/theme-hot-reload` from `^0.0.18` to `^0.0.22`.

**Root cause:** The hot-reload client was only checking `javascriptTag` changes to trigger full page reloads, but ignored `schemaTag` changes that the server was already sending.

The fix existed upstream in v0.0.20+ (commit 4785747 in the theme-hot-reload package), which adds:
```javascript
if (event.payload?.updatedFileParts?.javascriptTag ||
    event.payload?.updatedFileParts?.schemaTag) {
    return fullPageReload(event.key);
}
```

### How to test your changes?

_Note: I'm currently having 1Password issues so I'm locked out of the admin and haven't been able to manually test these myself_

1. Run `shopify theme dev --theme-editor-sync` in a theme directory
2. Open the theme customizer in your browser
3. Edit a section's `{% schema %}` block (e.g., add a new setting)
4. Save the file
5. **Expected:** The customizer reloads automatically and shows the new schema setting
6. **Before fix:** No reload occurred; manual refresh was required

### Measuring impact

- [x] n/a - this doesn't need measurement, e.g. a linting rule or a bug-fix

### Checklist

- [x] I've considered possible cross-platform impacts (Mac, Linux, Windows)
- [x] I've considered possible [documentation](https://shopify.dev) changes

Issue #1033, opened 2023-01-06
[Bug]: WSL2 Ubuntu CLI 3 Gem error nokogiri


hoektoe:
### Please confirm that you have:

- [X] Searched [existing issues](.) to see if your issue is a duplicate. (If you’ve found a duplicate issue, feel free to add additional information in a comment on it.)
- [X] Reproduced the issue in [the latest CLI version](https://www.npmjs.com/package/@shopify/cli).

### In which of these areas are you experiencing a problem?

Theme

### Expected behavior

shopify theme dev should have served my theme successfully and create localhost + editor linkgs

### Actual behavior

Got error for nokogiri failure of gem file.

### Stack trace

```shell
An error occurred while installing nokogiri (1.13.10), and Bundler cannot
continue.

In Gemfile:
  shopify-cli was resolved to 2.33.0, which depends on
    theme-check was resolved to 1.12.1, which depends on
      nokogiri
Fetching gem metadata from https://rubygems.org/.............
Resolving dependencies...
Using ast 2.4.2
Using concurrent-ruby 1.1.10
Using bundler 2.4.2
Using ffi 1.15.5
Using liquid 5.4.0
Using rb-fsevent 0.11.2
Using mini_portile2 2.8.1
Using racc 1.6.2
Using parser 3.2.0.0
Using bugsnag 6.25.1
Using rb-inotify 0.10.1
Using listen 3.7.1
Installing nokogiri 1.13.10 with native extensions
    at makeError (file:///home/linuxbrew/.linuxbrew/lib/node_modules/@shopify/theme/node_modules/execa/lib/error.js:59:11)
    at handlePromise (file:///home/linuxbrew/.linuxbrew/lib/node_modules/@shopify/theme/node_modules/execa/index.js:119:26)
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
    at async Module.exec (file:///home/linuxbrew/.linuxbrew/lib/node_modules/@shopify/theme/node_modules/@shopify/cli-kit/dist/system.js:41:9)
    at async bundleInstallShopifyCLI (file:///home/linuxbrew/.linuxbrew/lib/node_modules/@shopify/theme/node_modules/@shopify/cli-kit/dist/public/node/ruby.js:197:5)
    at async Task.task (file:///home/linuxbrew/.linuxbrew/lib/node_modules/@shopify/theme/node_modules/@shopify/cli-kit/dist/public/node/ruby.js:131:21)
[...]

PR #7673, opened 2026-05-29
Fix hot module reloading for theme development

graygilmore:

Fixes Shopify/developer-tools-team#1033.

Hot reload stopped working in the theme editor (Online Store Editor) a while back. With `shopify theme dev` running, reloading the customizer just logs `[HotReload] Connection closed by the server, attempting to reconnect...` and never reconnects. It's been reported a few times (community forum threads, dev-tools issue #1033), and an earlier attempt to fix it (#6833) stalled.

There are two separate problems:

1. The hot-reload client was a few versions behind and didn't trigger a full reload on `{% schema %}` changes.
2. The bigger one: the dev server's CORS allowlist was locked down to `localhost` and the store domain in 226b49e740 (to close a hole where any website could read from the localhost server). The theme editor renders its preview in an iframe served from `https://online-store-web.shopifyapps.com`, which opens the hot-reload `EventSource` back to the local dev server. That origin isn't on the allowlist, so the browser blocks the connection — which is the reconnect loop above.

The dependency bump on its own doesn't fix this — I confirmed the CORS errors on #6833 back in March. The server has to allow the editor's origin too.

- Bumps `@shopify/theme-hot-reload` from `0.0.18` to `0.0.22`. On the client side this triggers a full reload on schema changes, restricts the EventSource to known origins, and uses the original storefront domain when calling the Section Rendering API.
- Adds `https://online-store-web.shopifyapps.com` to the dev server's CORS allowlist so the theme editor's preview iframe can open the hot-reload connection.
- Adds tests for the allowlist: editor origin allowed, storefront origin allowed, unknown origins rejected, and no CORS headers on direct navigation (no `Origin` header).

This intentionally keeps the allowlist approach from 226b49e740 instead of going back to wildcard CORS, so the original vulnerability stays closed. It also only allowlists the production editor origin.

### How to test your changes?

A snapshot has been created for easy testing:

```sh
pnpm i -g --@shopify:registry=https://registry.npmjs.org @shopify/cli@0.0.0-snapshot-20260529205834
```

1. Install the snapshot (or pull this branch down).
2. Run `shopify theme dev` against a store and open the theme editor (use `e` shortcut).
3. Edit code locally
4. Expected: the preview hot reloads and the console logs `[HotReload] Connected`. Before this change you'd get `[HotReload] Connection closed by the server, attempting to reconnect...` instead.

### DIFF
--- a/.changeset/fix-theme-editor-hot-reload-cors.md
+++ b/.changeset/fix-theme-editor-hot-reload-cors.md
@@ -1,0 +1,5 @@
+---
+'@shopify/theme': patch
+---
+
+Fix `theme dev` hot reload not working in the theme editor by allowing the Online Store Editor origin through the dev server's CORS policy
--- a/.changeset/perky-melons-arrive.md
+++ b/.changeset/perky-melons-arrive.md
@@ -1,0 +1,5 @@
+---
+'@shopify/theme': patch
+---
+
+Bump @shopify/theme-hot-reload to address HMR issues
--- a/packages/theme/package.json
+++ b/packages/theme/package.json
@@ -49,7 +49,7 @@
     "yaml": "2.8.3"
   },
   "devDependencies": {
-    "@shopify/theme-hot-reload": "^0.0.18",
+    "@shopify/theme-hot-reload": "^0.0.22",
     "@vitest/coverage-istanbul": "^3.1.4",
     "node-stream-zip": "^1.15.0"
   },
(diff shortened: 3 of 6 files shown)


frandiox (approved):
Thanks for checking this!

>The hot-reload client was a few versions behind and didn't trigger a full reload on {% schema %} changes.

Actually, the CLI only uses the NPM package to import types. The actual code is injected via CDN, which is always using the latest version.

karreiro (approved):
Thank you, @graygilmore!

EvilGenius13 (approved):

isaacroldan (approved):

Issue #4567, opened 2024-10-02
[Bug]: Theme App Extension - Asset CDN link random UUID


AlvaApps:
### Please confirm that you have:

- [x] Searched [existing issues](.) to see if your issue is a duplicate. (If you’ve found a duplicate issue, feel free to add additional information in a comment on it.)
- [x] Reproduced the issue in [the latest CLI version](https://www.npmjs.com/package/@shopify/cli).

### In which of these areas are you experiencing a problem?

Extension

### Expected behavior

Running npm app deploy should release a new version of the theme app extension. If asset files are present they should be uploaded to the following url pattern "https://cdn.shopify.com/extensions/$EXTENSION_ID/$EXTENSION_VERSION/assets/$FILE_NAME" 
For example "https://cdn.shopify.com/extensions/1234_5678_9123_4567/0.0.1/assets/extension.js" 

### Actual behavior

The CDN link is of pattern "http://cdn.shopify.com/extensions/$RANDOM_UUID/0.0.1/assets/extension.js" 

### Verbose output

n/a

### Reproduction steps

1. Create a theme app extension and include an empty JS file ( extension.js) in the assets folders
2. In the liquid file add <script src="{{ 'extension.js' | asset_url }}" defer></script>
3. npm run deploy
4. Install the theme app extension on a page and view source. Locate the the CDN link

### Operating System

OSX

### Shopify CLI version (check your project's `package.json` if you're not sure)

3.66.0

### Shell

bhas

### Node version (run `node -v` if you're not sure)

22.6.0

### What language and version are you using in your application?

React/Remix

aswamy:
> Running npm app deploy should release a new version of the theme app extension. If asset files are present they should be uploaded to the following url pattern "https://cdn.shopify.com/extensions/$EXTENSION_ID/$EXTENSION_VERSION/assets/$FILE_NAME"

Hey @AlvaApps, do you know where it was said that the URL should conform to the standard above? The uuid chosen **does** seem to be random characters chosen during the creation of the app, but not sure if this is a request or a bug.

Issue #4549, opened 2024-10-01
[Bug]: shopify app dev is not working when a theme extension exists


RaphaelLima123:
### Please confirm that you have:

- [x] Searched [existing issues](.) to see if your issue is a duplicate. (If you’ve found a duplicate issue, feel free to add additional information in a comment on it.)
- [x] Reproduced the issue in [the latest CLI version](https://www.npmjs.com/package/@shopify/cli).

### In which of these areas are you experiencing a problem?

App, Extension

### Expected behavior

The shopify app dev command should successfully start the local development environment for my Shopify app without encountering any errors or permission issues. The app should load correctly in my default browser, allowing me to view and interact with my local development version of the app.

The error just happen when the app has a theme extension.

### Actual behavior

shopify app dev
╭─ info ─────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│                                                                                                                    │
│  Using shopify.app.test-app.toml for default values:                                                               │
│                                                                                                                    │
│    • Org:             Raphael Lima                                                                                 │
│    • App:             test app                                                                                     │
│    • Dev store:       raphael-lima.myshopify.com                                                                   │
│    • Update URLs:     Yes                                                                                          │
│                                                                                                                    │
│   You can pass  `--reset`  to your command to reset your app configuration.                                        │
│                                                                                                                    │
╰────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯

╭─ error ────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│                                                                                                                    │
│  Maximum call stack size exceeded                                                                                  │
│                                                                                                                    │
│  To investigate the issue, examine this stack trace:                                                               │
│    at openSync (node:fs:562)                                                                                       │
│    at attempt                                                                                                      │
│    (opt/homebrew/Cellar/shopify-cli/3.67.3/libexec/lib/node_modules/@shopify/cli/dist/chunk-NV2K2JNI.js:9332)      │
│    at attempt                                                                                                      │
│    (opt/homebrew/Cellar/shopify-cli/3.67.3/libexec/lib/node_modules/@shopify/cli/dist/chunk-NV2K2JNI.js:9337)      │
│    at attempt                                                                                                      │
│    (opt/homebrew/Cellar/shopify-cli/3.67.3/libexec/lib/node_modules/@shopify/cli/dist/chunk-NV2K2JNI.js:9337)      │
│    at attempt                                                                                                      │
[...]

Issue #3488, opened 2024-02-27
[Bug]: Incorrect JavaScript Asset URLs for Theme App Extensions in Some Stores


daviareias:
### In which of these areas are you experiencing a problem?

Extension

### Expected behavior

The expected behavior is that all stores serve the correct JavaScript asset link for theme app extensions, which should be:
https://cdn.shopify.com/extensions/df1598d5-ff9e-4a84-8767-9f1d0731c6cc/2.107.0/assets/oneator.js

### Actual behavior

In some stores, an incorrect JavaScript asset URL is being served, leading to a 404 error as the asset does not exist. The incorrect URL being served is:
https://cdn.shopify.com/extensions/df1598d5-ff9e-4a84-8767-9f1d0731c6cc/scala-parcelamentos-0/assets/oneator.js

**The version of the app being served is the same.**

### Verbose output

Since this is related to theme app extensions rather than a CLI command execution, the verbose output might not be applicable. However, any logs from Shopify's side regarding asset delivery failures would be helpful.

### Reproduction steps

To reproduce the error:

    1. Find a theme extension that has not been updated since may last year
    2. Inspect the page to find the script tag sourcing the JavaScript asset.
    3. Note the incorrect URL being served for the asset.
    4. Compare it to the correct url being served in most stores

### Operating System

not applicable

### Shopify CLI version (check your project's `package.json` if you're not sure)

not applicable

sagar-ranglani:
@daviareias What changes did you make with the app?

daviareias:
> @daviareias in that case the end users would need to re-enable the app embed block, right? For larger scale apps that's a real problem.

They don't need to renable the app, the problem is that Shopify is serving the wrong link. I'm going to reopen this, because I found out is actually affecting EVERY embedded app

avocadoslab:
Lots of merchants/partners facing this issue and intensity is increasing day by day

https://community.shopify.com/c/shopify-apps/app-embed-extension-blocks-url-throwing-404/m-p/2469135

agrohs:
What was fixed/what change does a developer need to make to have this work (and what are the correctly generate URLs supposed to look like format/pattern wise) - We seem to still be experiencing the issue w/ some of our apps.

PR #7674, opened 2026-05-29
Fix theme dev serving theme asset for /cdn/extensions/ requests

wes-shaw:
### Summary

`shopify theme dev` was serving a **theme** asset in response to an app extension's asset request (`/cdn/extensions/<uuid>/<app>/assets/<name>`) whenever the theme had a same-named file. The colliding script then loaded twice (once as the theme asset, once as the extension asset that was silently swapped), causing duplicated execution — e.g. double-bound event handlers. Reported in [this community thread](https://community.shopify.dev/t/cli-is-serving-incorrect-cdn-assets/34726) (Klaviyo Email Marketing's `app.js` vs a theme `app.js`).

The live storefront serves both URLs correctly; this was local-dev only. Reproduces on `main` and on `@shopify/cli` 4.1.0.

### Root cause

In `packages/theme/src/cli/utilities/theme-environment/local-assets.ts`, `findLocalFile` tried the theme matcher first via `??`:

```ts
tryGetFile(/^(?:\/cdn\/.*?)?\/assets\/([^?]+)/, ctx.localThemeFileSystem) ??
tryGetFile(/^(?:\/ext\/cdn\/extensions\/.*?)?\/assets\/([^?]+)/, ctx.localThemeExtensionFileSystem) ??
```

The theme matcher's optional `(?:\/cdn\/.*?)?` prefix is greedy enough to swallow the entire `/cdn/extensions/<uuid>/<app>` segment, capturing only the basename after `/assets/`. `tryGetFile` derives the lookup key from the capture alone (`joinPath('assets', matchedFileName)`), discarding the URL path, so `/cdn/extensions/.../assets/app.js` resolves to `assets/app.js` in `localThemeFileSystem` and is served.

The extension matcher only accepts `/ext/cdn/extensions/` (note `/ext/`), so it never gets a chance for a bare `/cdn/extensions/` request, and the `??` short-circuits before `getProxyHandler` (which already forwards extension CDN traffic to `cdn.shopify.com`) is ever reached.

### Fix

Tighten the theme regex with a negative lookahead so its `/cdn/` prefix cannot be followed by `extensions/`, and broaden the extension regex to also accept `/cdn/extensions/...` in addition to `/ext/cdn/extensions/...`.

```ts
const THEME_ASSET_PATTERN = /^(?:\/cdn\/(?!extensions\/).*?)?\/assets\/([^?]+)/
const THEME_EXTENSION_ASSET_PATTERN = /^(?:\/(?:ext\/)?cdn\/extensions\/.*?)?\/assets\/([^?]+)/
```

When the requested extension asset is *not* in `localThemeExtensionFileSystem`, `findLocalFile` returns `{fileKey: undefined}` and `getAssetsHandler` returns early — the request falls through to `getProxyHandler` which forwards to `cdn.shopify.com` and the real installed-app asset is returned.

### Behaviour after fix

- `/cdn/extensions/<uuid>/<app>/assets/<name>` → proxied to CDN, or served from `localThemeExtensionFileSystem` when present (covers theme-app-extension dev against the vanity CDN form).
- `/cdn/shop/t/<id>/assets/<name>` → still served from `localThemeFileSystem`.
- `/ext/cdn/extensions/...` → still served from `localThemeExtensionFileSystem`.
- Bare `/assets/<name>` → unchanged.

### Test coverage

New `packages/theme/src/cli/utilities/theme-environment/local-assets.test.ts` covers six cases for `findLocalFile`, including the regression (a theme `assets/app.js` must **not** be returned for a `/cdn/extensions/.../assets/app.js` request) and the locally-developed-extension precedence case. Confirmed locally: with the regex change reverted, exactly the 2 collision tests fail; with the fix applied, all 6 new tests pass and the full `packages/theme` test suite (163 tests) still passes, along with `nx type-check theme` and `nx lint theme`.

### Tophatting

In a project that uses `shopify theme dev`:

1. Install an app whose extension bundles an asset, e.g. `assets/app.js`.
2. Add a theme asset with the same filename: `assets/app.js`, with distinct contents (e.g. `console.log('THEME')` vs the app's content).
3. Run `shopify theme dev` and load a page that renders both:
   - theme tag: `<script src="/cdn/shop/t/<id>/assets/app.js?v=...">`
   - extension tag (from `content_for_header`): `<script src="/cdn/extensions/<uuid>/<app>/assets/app.js">`
4. `curl http://localhost:9292/cdn/extensions/<uuid>/<app>/assets/app.js` — before the fix this returns the theme's `app.js`; after the fix it returns the real app's `app.js` (proxied from `cdn.shopify.com`).
5. Confirm `curl http://localhost:9292/cdn/shop/t/<id>/assets/app.js` still returns the local theme's `app.js`.

### Notes

- File/line references are from `main`; 4.1.0 line numbers differ slightly but the logic is identical.
- No existing Shopify/cli issue or PR tracks this. Related but distinct: #4549, #4567, #3488.
- Per `.github/CODEOWNERS`, `packages/theme/**` is owned by `@shopify/developer-platforms` and `@shopify/dev_experience`.

Requested by Wes Shaw <wes.shaw@shopify.com>

### DIFF
--- a/.changeset/river-theme-dev-cdn-extensions-collision.md
+++ b/.changeset/river-theme-dev-cdn-extensions-collision.md
@@ -1,0 +1,5 @@
+---
+'@shopify/theme': patch
+---
+
+Fix `theme dev` serving a same-named theme asset in response to a `/cdn/extensions/...` request. When a theme and an installed app extension shared an asset filename (e.g. `app.js`), the local dev server's theme matcher swallowed the extension URL prefix and returned the theme file. Extension asset requests now fall through to the CDN proxy (or to a locally-developed extension's filesystem) as intended.
--- a/packages/theme/src/cli/utilities/theme-environment/local-assets.test.ts
+++ b/packages/theme/src/cli/utilities/theme-environment/local-assets.test.ts
@@ -1,0 +1,99 @@
+import {findLocalFile} from './local-assets.js'
+import {emptyThemeFileSystem, emptyThemeExtFileSystem} from '../theme-fs-empty.js'
+import {describe, expect, test} from 'vitest'
+import {createEvent} from 'h3'
+import {IncomingMessage, ServerResponse} from 'node:http'
+import {Socket} from 'node:net'
+import type {DevServerContext} from './types.js'
+
+function createH3Event(path: string) {
+  const req = new IncomingMessage(new Socket())
+  const res = new ServerResponse(req)
+  req.method = 'GET'
+  req.url = path
+  req.headers = {}
+  return createEvent(req, res)
+}
+
+function buildCtx({
+  themeFiles = [] as [string, unknown][],
+  extFiles = [] as [string, unknown][],
+}): DevServerContext {
+  const themeFs = emptyThemeFileSystem()
+  for (const [key, value] of themeFiles) themeFs.files.set(key, value as never)
+  const extFs = emptyThemeExtFileSystem()
+  for (const [key, value] of extFiles) extFs.files.set(key, value as never)
+  return {localThemeFileSystem: themeFs, localThemeExtensionFileSystem: extFs} as unknown as DevServerContext
+}
+
+describe('findLocalFile', () => {
+  const themeAsset = {checksum: 't', key: 'assets/app.js', value: '// theme app.js'}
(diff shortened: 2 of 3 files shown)


graygilmore (approved):
Changes make sense to me. Was able to do a bit of 🎩.

PR #7682, opened 2026-05-31
[Security] Upgrade hashing to SHA-256

gonzaloriestra:

SHA-1 is considered a weak cryptographic hash and should be replaced with a stronger algorithm like SHA-256, especially when handling sensitive data or generating identifiers.

- Upgrades `hashString` to use SHA-256.
- Upgrades `nonRandomUUID` to use SHA-256 while maintaining the 36-character UUID format by truncating the hash.
- Updates relevant unit tests in `cli-kit` that were affected by the change in hash algorithm and generated values.

### How to test your changes?

CI

### Checklist

---
*PR created automatically by Jules for task [811138014799838999](https://jules.google.com/task/811138014799838999) started by @gonzaloriestra*

### DIFF
--- a/packages/app/src/cli/services/generate/extension.test.ts
+++ b/packages/app/src/cli/services/generate/extension.test.ts
@@ -266,7 +266,7 @@ describe('initialize a extension', async () => {
           name,
           handle: slugify(name),
           flavor,
-          uid: 'ba7c20a9-578d-6fee-8cd2-044af992dabd92d8bbfe',
+          uid: '37e88e17-2d38-e2a3-4753-3e3066cf549c',
         })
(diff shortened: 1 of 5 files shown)


isaacroldan:
Not sure if this is safe to do, it changes the length of identifiers no? 🤔

gonzaloriestra:
@isaacroldan the length is the same, the UIDs you see from tests were wrong previously.  
  
But I'm now thinking that it could be problematic if we compare UIDs from static extensions, because if I recall correctly they have deterministic IDs based on the handle, no? So the matching would probably fail... Let's better forget about this :sweat_smile:

PR #7708, opened 2026-06-03
Add store list for organization stores

dmerand:
## Summary

Add `shopify store list`, listing stores from one Shopify organization available to the current CLI account.

Builds `store list` out from the prototype in the [`shopify store create preview` stack](https://app.graphite.com/github/pr/Shopify/cli/7558/Add-%60shopify-store-create-preview%60-command#file-packages/store/src/cli/services/store/create/preview/index.ts).

Closes https://github.com/shop/issues-develop/issues/22723

## Scope

- add `shopify store list` as an organization-backed store-management list
- resolve one organization before listing stores:
  - `--organization-id` selects that organization and uses the shared integer organization flag
  - a single accessible organization is selected automatically
  - multiple accessible organizations prompt in interactive terminals
  - non-interactive multiple-organization runs require `--organization-id`
- move the reusable "select from an already-fetched organization list" flow into `@shopify/organizations.selectOrganizationFromList()` and use it from both app context and store list
- fetch only active stores from the selected organization: `accessibleShops(first: 250, sort: SHOP_CREATED_AT_DESC, filters: STORE_STATUS = active)`
- render `Subdomain | Name | Type | Created`, sorted newest first
- cap output at 250 stores; on truncation, emit a stderr warning and `truncated: true` in JSON
- include selected organization context in JSON as `organization: {id, name}`
- model organization selection errors distinctly from unresolved-session results
- harden the CLI session read against malformed persisted JSON

## Review notes

- user-facing output and the command contract avoid the internal "Business Platform" naming; the command is framed around Shopify organizations
- this PR is organization-source only; directly authenticated store sessions are intentionally handled by a separate auth-listing command upstack
- no `--from` flag and no local store-auth fallback; Slack feedback on 2026-06-14/15 made the two jobs clearer as separate commands
- the command intentionally does not fan out across all organizations anymore
- store list keeps command-specific policy local (zero-org result, non-interactive error), while ID matching and prompt selection reuse `@shopify/organizations`

## Not changed / deferred

- Did not add `--from`, `auto`, or store-auth fallback behavior. Those made one command responsible for two different jobs and were explicitly removed after Slack review.
- Did not make `store list` list every organization. It still resolves one organization first to avoid broad fan-out, ambiguous output, and partial multi-org policy in this first command contract.
- Did not switch `--organization-id` back to a string after follow-through review. There is a theoretical JS safe-integer concern, but the current reviewer direction is to use the shared integer `organization-id` flag. Reopen only if Shopify organization IDs can exceed JS safe integer range.
- Did not make `store list` show whether each store is already authenticated for `store execute`. That belongs to the direct store-auth listing command upstack, not the organization store-management list.


fahadhataba (approved):

dmerand:
* **#7864** <a href="https://app.graphite.com/github/pr/Shopify/cli/7864?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#7709** <a href="https://app.graphite.com/github/pr/Shopify/cli/7709?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#7708** <a href="https://app.graphite.com/github/pr/Shopify/cli/7708?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Shopify/cli/7708?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

amcaplan:
I believe you should be able to gitignore packages/store/src/cli/api/graphql/business-platform-organizations/organizations_schema.graphql

.changeset/store-list-bp-auto.md
```diff
@@ -0,0 +1,8 @@
+---
+'@shopify/cli': minor
+'@shopify/store': minor
+'@shopify/organizations': minor
+'@shopify/cli-kit': patch
```
alfonso-noriega: Should this also be minor?

.changeset/store-list-bp-auto.md
```diff
@@ -0,0 +1,8 @@
...
+---
+
+Add `shopify store list` to list stores from Business Platform for the current Shopify CLI account. When some organization-scoped Business Platform lookups fail, the command now returns successful organizations with a warning and failure details for the skipped organizations.
```
amcaplan: Changeset shouldn't mention BP

packages/store/src/cli/commands/store/list.test.ts
```diff
@@ -0,0 +1,42 @@
...
+    await StoreList.run(['--json'])
+
+    expect(listStores).toHaveBeenCalledWith({organizationId: undefined})
+    expect(writeStoreListResult).toHaveBeenCalledWith({entries: []}, 'json')
+  })
+
+  test('defines the expected flags', () => {
+    expect(StoreList.flags.json).toBeDefined()
+    expect(StoreList.flags['organization-id']).toBeDefined()
+    expect('source' in StoreList.flags).toBe(false)
```
amcaplan: This line looks like a slop artifact

packages/store/src/cli/commands/store/list.ts
```diff
@@ -0,0 +1,38 @@
+import {listStores} from '../../services/store/list/index.js'
+import {writeStoreListResult} from '../../services/store/list/result.js'
+import StoreCommand from '../../utilities/store-command.js'
+import {globalFlags, jsonFlag} from '@shopify/cli-kit/node/cli'
+import {Flags} from '@oclif/core'
+
+export default class StoreList extends StoreCommand {
+  static summary = 'List stores in your Shopify organization.'
```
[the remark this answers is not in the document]
amcaplan: Agreed with Copilot

packages/store/src/cli/services/store/list/bp-source.ts
```diff
@@ -0,0 +1,195 @@
...
+      ...(organizationsResult.currentUserEmail ? {currentUserEmail: organizationsResult.currentUserEmail} : {}),
+    }
+  }
+
+  const organizationResults = await Promise.allSettled(
+    organizations.map(async (organization) => ({
+      organization,
+      ...(await fetchOrganizationStores(token, organization.id, organization.businessName)),
+    })),
+  )
```
amcaplan: This is going to create big problems for users with many organizations - they will get rate limited! We should try to do this in batches instead, and stop searching when we've hit 250 results.

This change alone should impact much of the code below, so I'll hold off on commenting further because much of it is likely to change.

packages/store/src/cli/services/store/list/result.ts
```diff
@@ -0,0 +1,97 @@
...
+  if (format === 'json') {
+    outputResult(JSON.stringify(result, null, 2))
+    return
+  }
+
+  renderTextResult(result)
+}
+
+function truncationWarning(): string {
+  return `Showing the ${STORE_LIST_LIMIT} most recent stores. More stores exist — narrow the results with \`--organization-id\`, or filter the JSON output.`
```
amcaplan: "filter the JSON output" won't actually help in this case

packages/store/src/cli/services/store/list/result.ts
```diff
@@ -0,0 +1,97 @@
...
+
+function summaryLine(result: ListStoresResult): string {
+  const noun = result.entries.length === 1 ? 'store' : 'stores'
+  const currentUser = result.currentUserEmail ? ` (logged in as ${result.currentUserEmail})` : ''
+
+  return `${result.entries.length} ${noun}${currentUser}`
+}
+
+function subdomainFor(store: string): string {
+  return store.replace(/\.myshopify\.com$/i, '')
```
amcaplan: We should use the generic helper from CLI kit. Local dev doesn't use .myshopify.com domains.

packages/store/src/cli/services/store/list/result.ts
```diff
@@ -0,0 +1,97 @@
...
+}
+
+function storeTypeLabel(type?: string): string {
+  if (!type) return ''
+  return STORE_TYPE_LABELS[type] ?? titleCaseFromEnum(type)
+}
+
+function titleCaseFromEnum(value: string): string {
```
amcaplan: We should have a util for this in CLI kit

packages/store/src/cli/services/store/list/types.ts
```diff
@@ -0,0 +1,25 @@
+export const STORE_LIST_LIMIT = 250
```
amcaplan: Repo convention: Usually we put constants in a constants file, not a types file.

packages/store/src/cli/services/store/list/result.ts
```diff
@@ -0,0 +1,97 @@
...
+  return value
+    .toLowerCase()
+    .split('_')
+    .map((word) => (word ? word.charAt(0).toUpperCase() + word.slice(1) : word))
+    .join(' ')
+}
+
+function formatListDate(value: string): string {
```
amcaplan: We shouldn't have date formatting in a specific file - we should strive for consistency and pull out a helper. As a bonus, this should help us be consistent about dates throughout the CLI.

packages/store/src/cli/api/graphql/business-platform-organizations/queries/list_accessible_shops.graphql
```diff
@@ -0,0 +1,22 @@
+query ListAccessibleShops($first: Int!) {
+  organization {
+    id
+    name
+    accessibleShops(first: $first, sort: SHOP_CREATED_AT_DESC) {
```
alfonso-noriega: We should be filtering by STORE_STATUS: ACTIVE to avoid deleted stores here

alfonso-noriega (dismissed):

packages/store/src/cli/commands/store/list.ts
```diff
@@ -0,0 +1,42 @@
...
+import {globalFlags, jsonFlag} from '@shopify/cli-kit/node/cli'
+import {Flags} from '@oclif/core'
+
+export default class StoreList extends StoreCommand {
+  static summary = 'List stores in a Shopify organization.'
+
+  static descriptionWithMarkdown = `Lists stores in a Shopify organization available to the current CLI account.
+
+When more than one organization is available, the command prompts you to pick one unless you provide \`--organization-id\`.
+In non-interactive environments, provide \`--organization-id\`.
```
amcaplan: Something about this language feels a bit forced. Maybe something like

```suggestion
In non-interactive environments, \`--organization-id\` is required.
```

packages/store/src/cli/commands/store/list.ts
```diff
@@ -0,0 +1,42 @@
...
+
+Run \`shopify organization list\` to find organization IDs.`
```
amcaplan: Can we use <%= config.bin %> here?

packages/store/src/cli/commands/store/list.ts
```diff
@@ -0,0 +1,42 @@
...
+  static examples = [
+    '<%= config.bin %> <%= command.id %>',
+    '<%= config.bin %> <%= command.id %> --organization-id 1234567',
+    '<%= config.bin %> <%= command.id %> --json',
+  ]
+
+  static flags = {
+    ...globalFlags,
+    ...jsonFlag,
+    'organization-id': Flags.string({
```
amcaplan: Why isn't this an integer? Would we expect gid format sometimes?

packages/store/src/cli/services/store/list.ts
```diff
@@ -0,0 +1,70 @@
...
+import {AbortError} from '@shopify/cli-kit/node/error'
+import {ensureAuthenticatedBusinessPlatform} from '@shopify/cli-kit/node/session'
+import {isTTY} from '@shopify/cli-kit/node/ui'
+import {fetchOrganizationsWithAccessInfo, selectOrganizationFromList, type Organization} from '@shopify/organizations'
+
+interface ListStoresOptions {
+  organizationId?: string
+}
+
+export async function listStores(options: ListStoresOptions = {}): Promise<ListStoresResult> {
```
amcaplan, line 13: I believe making `--organization-id` an integer flag would make some of this unnecessary.

packages/store/src/cli/services/store/list/index.ts
```diff
@@ -0,0 +1,70 @@
+import {listBusinessPlatformStores} from './bp-source.js'
```
amcaplan: I just noticed that we have a few barrel files like this one in this package. Unfortunately I introduced at least one, so mea culpa there... generally the practice in this repo is not to use barrel files except on the whole-package level (for package exports). We should probably build this into our agent guidelines.

In this case, the expectation would be that this file would live at `packages/store/src/cli/services/store/list.ts` with no difference in the import string in other files, but of course the import strings in this file will need to be updated.

packages/organizations/src/cli/services/select.ts
```diff
@@ -5,7 +5,13 @@ import {AbortError} from '@shopify/cli-kit/node/error'
 
 export async function selectOrg(orgIdFromFlag?: string): Promise<Organization> {
   const organizations = await fetchOrganizations()
+  return selectOrganizationFromList(organizations, orgIdFromFlag)
+}
 
+export async function selectOrganizationFromList<T extends Organization>(
```
amcaplan: I don't really understand why this function was broken out. It's only used in 1 file, where it's immediately preceded by a call to `fetchOrganizations()` - exactly what's done in `selectOrg` - so why wouldn't we just align with what's being done here? We could drop many LoC from this PR if we don't make this change.

.changeset/store-list-bp-auto.md
```diff
@@ -0,0 +1,8 @@
+---
+'@shopify/cli': minor
+'@shopify/store': minor
+'@shopify/organizations': minor
+'@shopify/cli-kit': minor
+---
+
+Add `shopify store list` to list the stores in the Shopify organizations available to the current CLI account. When some organization lookups fail, the command returns the successful organizations with a warning and failure details for the skipped ones.
```
isaacroldan: This is later used for the changelog, I think just `Add shopify store list to list the stores in the Shopify organizations available to the current CLI account` is enough, no need for the second part.

docs-shopify.dev/generated/generated_docs_data_v2.json
```diff
@@ -4339,6 +4339,53 @@
       "value": "export interface storeinfo {\n  /**\n   * Output the result as JSON. Automatically disables color output.\n   * @environment SHOPIFY_FLAG_JSON\n   */\n  '-j, --json'?: ''\n\n  /**\n   * Disable color output.\n   * @environment SHOPIFY_FLAG_NO_COLOR\n   */\n  '--no-color'?: ''\n\n  /**\n   * The myshopify.com domain of the store.\n   * @environment SHOPIFY_FLAG_STORE\n   */\n  '-s, --store <value>': string\n\n  /**\n   * Increase the verbosity of the output.\n   * @environment SHOPIFY_FLAG_VERBOSE\n   */\n  '--verbose'?: ''\n}"
     }
   },
+  "storelist": {
```
isaacroldan: remember that you still need to create the command manually in `shopify-dev` for it to appear in the docs.

packages/store/src/cli/commands/store/list.ts
isaacroldan: I don't think there is any benefit on using `<%= config.bin %>`, it will always translate to `shopify` no?  
No need to change it, just curious why you prefer it

packages/store/src/cli/services/store/display.ts
```diff
@@ -0,0 +1,19 @@
...
+
+export function formatShortDate(value: Date | number | string): string {
+  const date = value instanceof Date ? value : new Date(value)
+  if (Number.isNaN(date.getTime())) return ''
+
+  const month = SHORT_DATE_MONTHS[date.getUTCMonth()]
+  if (!month) return ''
+
+  return `${month} ${String(date.getUTCDate()).padStart(2, '0')}, ${date.getUTCFullYear()}`
+}
```
isaacroldan, line 15: I think we should use a Date Formatter instead of manually formatting here:  


```javascript
const shortDateFormatter = new Intl.DateTimeFormat('en-US', {
  timeZone: 'UTC',
  month: 'short',
  day: '2-digit',
  year: 'numeric',
})

export function formatShortDate(value: Date | number | string): string {
  const date = value instanceof Date ? value : new Date(value)
  if (Number.isNaN(date.getTime())) return ''

  return shortDateFormatter.format(date)
}
```



The shortDateFormatter can even be in a new `date-formatter.ts` file in `cli-kit` to be reused by others if you want to.

packages/store/src/cli/services/store/list/bp-source.ts
```diff
@@ -0,0 +1,93 @@
...
+
+interface ListBusinessPlatformStoresOptions {
+  token: string
+  organization: Organization
+}
+
+interface BusinessPlatformStoreListResult {
+  entries: StoreListEntry[]
+  // True when the selected organization had more stores than we fetched (server-side limited).
+  hasMore?: boolean
```
isaacroldan: why make it optional? is weird that when it's false, we return undefined

packages/store/src/cli/services/store/list/bp-source.ts
```diff
@@ -0,0 +1,93 @@
...
+
+// Fetches one server-sorted page of the selected organization's newest stores. The page size is the
+// maximum number of stores we can display, and hasMore reflects whether more stores exist beyond it.
+async function fetchOrganizationStores(
+  token: string,
+  organization: Organization,
+): Promise<{entries: StoreListEntry[]; hasMore: boolean}> {
+  const unauthorizedHandler: UnauthorizedHandler = {
+    type: 'token_refresh',
+    handler: async () => ({token: await ensureAuthenticatedBusinessPlatform()}),
```
isaacroldan: the `unauthorizedHandler` we have in the app package is more complex (handles inProgress refrehes), do we need that or is ok as it is?

isaacroldan: Also, I see that in the store package we have multiple instances of `unauthorizedHandler`, should we unify this? maybe have a `business-platform-client.ts` with all the queries/mutations?

packages/store/src/cli/services/store/business-platform.ts
```diff
@@ -0,0 +1,20 @@
...
+): UnauthorizedHandler {
+  return {
+    type: 'token_refresh',
+    handler: async () => ({token: await refreshBusinessPlatformToken(options)}),
+  }
+}
+
+async function refreshBusinessPlatformToken(options: BusinessPlatformTokenRefreshHandlerOptions): Promise<string> {
+  if (options.noPrompt === undefined) return ensureAuthenticatedBusinessPlatform()
+  return ensureAuthenticatedBusinessPlatform([], {noPrompt: options.noPrompt})
```
isaacroldan, line 18: why the special case when noPrompt is undefined? if it's undefined the default value is false

isaacroldan (approved):
remove `refreshBusinessPlatformToken` before merging

PR #7709, opened 2026-06-03
Add local store auth summary API

dmerand:
## Summary

Introduce a local store-auth summary API without changing command behavior yet.

## Scope

- add `StoredStoreAuthSummary`
- add `listStoredStoreAuthSummaries()` under `services/store/auth`
- keep token-bearing session buckets as the current persistence source of truth
- keep `conf` traversal and bucket cleanup in `session-store.ts`
- re-export the summary API from `services/store/auth/index.ts`
- sort summaries newest-auth-first for the “return to the store I was working on” use case

## Why this split exists

The next PR needs one narrow question answered first:

- what directly authenticated store auth exists on this machine?

This PR answers that without taking on product behavior:

- no auth-listing command yet
- no persistence migration
- no access-token exposure

That keeps the behavior PR focused on command policy and output, not on storage traversal.

## Boundary this PR is trying to establish

- persistence details stay in `session-store.ts`
- callers see `StoredStoreAuthSummary[]`
- callers do not need token-bearing session buckets or `conf` layout knowledge

## Not changed / deferred

- Did not add the `shopify auth list` command in this PR. This PR is only the storage projection; command behavior lands upstack.
- Did not expose access tokens or refresh tokens. The summary API is deliberately safe for display/automation consumers.
- Did not migrate the persisted store-auth schema or add an index key for known stores. This keeps existing auth sessions compatible and avoids making a storage migration part of this stack.
- Did not add a public `LocalStorage.entries()` / enumeration API in cli-kit. Follow-through review correctly noted that `session-store.ts` still relies on the underlying `conf` layout to discover existing buckets; fixing that should be a focused storage follow-up before adding more local-auth inventory features.

### DIFF
--- a/packages/store/src/cli/services/store/auth/session-store.ts
+++ b/packages/store/src/cli/services/store/auth/session-store.ts
@@ -1,5 +1,6 @@
 import {STORE_AUTH_APP_CLIENT_ID, storeAuthSessionKey} from './config.js'
 import {LocalStorage} from '@shopify/cli-kit/node/local-storage'
+import {type JsonMapType} from '@shopify/cli-kit/node/toml'
 
 /**
  * Discriminator for a stored store auth session.
@@ -61,6 +62,8 @@ interface StoredStoreAppSessionBucket {
 interface StoreSessionSchema {
   [key: string]: StoredStoreAppSessionBucket
 }
+
+type RawStoreSessionStorage = JsonMapType
 
 let _storeSessionStorage: LocalStorage<StoreSessionSchema> | undefined
 
@@ -205,8 +208,8 @@ function readStoredStoreAppSessionBucket(
 // `conf` persists dotted keys as nested objects. Store-auth callers should not
 // learn that layout directly; this helper keeps the current traversal private to
 // the persistence seam while higher-level code projects summaries instead.
-function readRawStoreSessionStorage(storage: LocalStorage<StoreSessionSchema>): Record<string, unknown> {
-  return (storage as unknown as {config: {store: Record<string, unknown>}}).config.store ?? {}
+function readRawStoreSessionStorage(storage: LocalStorage<StoreSessionSchema>): RawStoreSessionStorage {
+  return (storage as unknown as {config?: {store?: RawStoreSessionStorage}}).config?.store ?? {}
 }
 
 function collectCurrentStoredStoreAppSessions(
@@ -224,7 +227,7 @@ function collectCurrentStoredStoreAppSessions(
     return
   }
 
-  for (const [childKey, childValue] of Object.entries(value as Record<string, unknown>)) {
+  for (const [childKey, childValue] of Object.entries(value as RawStoreSessionStorage)) {
     collectCurrentStoredStoreAppSessions(storage, `${store}.${childKey}`, childValue, sessions)
   }
 }
--- a/packages/store/src/cli/services/store/auth/session-store.ts
+++ b/packages/store/src/cli/services/store/auth/session-store.ts
@@ -1,4 +1,4 @@
-import {storeAuthSessionKey} from './config.js'
+import {STORE_AUTH_APP_CLIENT_ID, storeAuthSessionKey} from './config.js'
 import {LocalStorage} from '@shopify/cli-kit/node/local-storage'
 
 /**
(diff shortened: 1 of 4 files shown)


dmerand:
* **#7864** <a href="https://app.graphite.com/github/pr/Shopify/cli/7864?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#7709** <a href="https://app.graphite.com/github/pr/Shopify/cli/7709?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Shopify/cli/7709?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#7708** <a href="https://app.graphite.com/github/pr/Shopify/cli/7708?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

packages/store/src/cli/services/store/auth/session-store.ts
```diff
@@ -81,21 +83,22 @@ function sanitizeStoredStoreAppSession(value: unknown): StoredStoreAppSession |
     ...(isString(session.refreshToken) ? {refreshToken: session.refreshToken} : {}),
     ...(isString(session.expiresAt) ? {expiresAt: session.expiresAt} : {}),
     ...(isString(session.refreshTokenExpiresAt) ? {refreshTokenExpiresAt: session.refreshTokenExpiresAt} : {}),
-    ...(sanitizeAssociatedUser(session.associatedUser)
-      ? {associatedUser: sanitizeAssociatedUser(session.associatedUser)}
-      : {}),
+    ...(associatedUser ? {associatedUser} : {}),
```
amcaplan, line 148: Nice drive-by refactor 👍

packages/store/src/cli/services/store/auth/session-store.ts
```diff
@@ -131,6 +134,58 @@ function readStoredStoreAppSessionBucket(
...
+  store: string,
+  storage: LocalStorage<StoreSessionSchema>,
+): StoredStoreAppSessionBucket | undefined {
+  return sanitizeStoredStoreAppSessionBucket(store, storage.get(storeAuthSessionKey(store)), storage)
+}
+
+// `conf` persists dotted keys as nested objects. Store-auth callers should not
+// learn that layout directly; this helper keeps the current traversal private to
+// the persistence seam while higher-level code projects summaries instead.
+function readRawStoreSessionStorage(storage: LocalStorage<StoreSessionSchema>): Record<string, unknown> {
```
amcaplan: Do these need to be unknowns? Would a union type be possible?

packages/store/src/cli/services/store/auth/session-store.ts
```diff
@@ -131,6 +134,58 @@ function readStoredStoreAppSessionBucket(
...
+  return (storage as unknown as {config: {store: Record<string, unknown>}}).config.store ?? {}
+}
+
+function collectCurrentStoredStoreAppSessions(
```
amcaplan, line 215: This function is kind of an abomination. I would strongly prefer that we have a background task that fixes the storage once (e.g. `foo.myshopify.com` => `foo-myshopify-com`) and then we can avoid this complexity going forward.

By the way, we don't really test this crawling logic anywhere if I am understanding correctly.

packages/store/src/cli/services/store/auth/session-store.ts
```diff
@@ -62,6 +62,15 @@ interface StoreSessionSchema {
...
 }
 
+type RawStoreSessionNode =
+  | StoredStoreAppSessionBucket
+  | {[key: string]: RawStoreSessionNode}
+  | string
+  | number
+  | boolean
+  | null
+type RawStoreSessionStorage = Record<string, RawStoreSessionNode>
```
isaacroldan: Better to use `JsonMapType` if we want it generic, or a explicit schema with all the fields.

packages/store/src/cli/services/store/auth/session-store.ts
isaacroldan: Let's migrate to the new format and ignore old sessions. Worst case, people will need to re-auth their stores. But it will be better long time

isaacroldan (approved):
Approving because `collectCurrentStoredStoreAppSessions`is being cleaned up in a following PR

PR #7726, opened 2026-06-05
Remove semver from cli-kit/app

amcaplan:
## Remove `semver` from `@shopify/cli-kit`

**Why:** `semver` is a high-churn dependency (**56 Dependabot bumps / 24 months**). Part of an initiative to cut low-value dependency churn in the CLI monorepo.

**Replacement:** Swapped to `compare-versions` (already in the tree).
- `checkForNewVersion` / `checkForCachedNewVersion`: `new SemVer(a).compare(b)` → `compareVersions(a, b)`.
- `versionSatisfies`: `semver.satisfies` → `compare-versions`' `satisfies`, guarded by its `validate` so invalid input returns `false` instead of throwing (matching semver's forgiving behavior).
- `version.ts` `isMajorVersionChange`: `.major` extraction via a small local helper backed by `validateStrict`.

**Prerelease / snapshot handling:** `compare-versions` compares prereleases by numeric precedence and has no semver-style "exclude prereleases from normal ranges" rule. Rather than reproduce that quirk inside `versionSatisfies`, the policy for the CLI's `0.0.0-*` snapshot builds now lives explicitly at the two call sites that have a product opinion about it:
- **notifications:** snapshot builds don't match version-pinned notifications.
- **app templates (`allowedTemplates`):** snapshot builds get all non-deprecated templates (its existing `isPreReleaseVersion` path is now authoritative).

This keeps `versionSatisfies` a plain, ecosystem-standard range check and is **behavior-preserving** for all real release versions and for snapshot builds at every call site. The live `notifications.json` feed only uses full semver versions, so no partial-range handling is needed.

**Sites:** `version.ts`, `node-package-manager.ts`, `notifications-system.ts` (cli-kit); `app-management-client.ts` (app).

**Validation:** `lint` ✅, `type-check` ✅, `vitest` ✅ (node-package-manager 82, app-management-client 65, binaries 34; notifications + version suites green). Parity/edge-case tests added for `versionSatisfies`, `isMajorVersionChange`, and snapshot notification filtering.

🤖 AI-generated — needs human review before merge.


packages/cli-kit/package.json
```diff
@@ -121,6 +121,7 @@
     "chalk": "5.4.1",
     "change-case": "4.1.2",
     "color-json": "3.0.5",
+    "compare-versions": "6.1.1",
```
gonzaloriestra, line 124: This is not removing a dependency, but replacing it with another, no? The description says that `compare-versions` already existed, but it's not true.

gonzaloriestra (commented):
I see there were 9 semver releases in the last 2 years, where is `56 Dependabot bumps / 24 months` coming from?

PR #7728, opened 2026-06-05
Remove dotenv from cli-kit; use native util.parseEnv

amcaplan:
## Remove `dotenv` from `@shopify/cli-kit`

**Why:** Part of an initiative to cut low-value dependency churn (**10 Dependabot bumps / 24 months**).

**Replacement:** Node's built-in `util.parseEnv()` (cli-kit already requires `engines.node ≥ 22.12`).
- Sole site: `packages/cli-kit/src/public/node/dot-env.ts` — `dotenv.parse()` → `util.parseEnv()`.

**`@types/node` bump (`18.19.70` → `22.19.17`):** The pinned types predated `util.parseEnv`, so a naked import wouldn't type-check. Rather than a local type shim, the types are bumped to match the `engines.node ≥ 22.12` requirement (root + `packages/e2e`). `parseEnv` types values as `string | undefined`, so the result is narrowed to `Record<string, string>` at the single call site with a short documented cast. The bump also widened `process.exit`'s signature, so two theme test mocks are updated to match.

**ESLint rule fix (`no-inline-graphql`), separate commit:** The `@types/node` bump deterministically flips how pnpm installs `@shopify/eslint-plugin-cli` (`file:` injection → `link:` symlink), which shifts the rule's `__dirname` and broke its fixed-offset repo-root calculation — flagging every grandfathered file. The repo-root lookup is now install-layout-independent: it asks git (`git rev-parse --show-toplevel`) instead of deriving the root from `__dirname`. Kept as its own commit so it lands cleanly under rebase-merge.

**Validation:** CI green — `type-check`, `lint`, `vitest`, `e2e`.

**Out of scope:** The e2e package keeps its own `dotenv` dependency (only its `@types/node` was bumped here).

🤖 AI-generated draft — needs human review before merge.

### DIFF
--- a/packages/e2e/package.json
+++ b/packages/e2e/package.json
@@ -33,7 +33,6 @@
     "@playwright/test": "^1.50.0",
     "@shopify/toml-patch": "0.3.0",
     "@types/node": "22.19.17",
-    "dotenv": "16.4.7",
     "execa": "^7.2.0",
     "node-pty": "^1.0.0",
     "strip-ansi": "^7.1.0",
--- a/packages/e2e/package.json
+++ b/packages/e2e/package.json
@@ -32,7 +32,7 @@
     "@iarna/toml": "2.2.5",
     "@playwright/test": "^1.50.0",
     "@shopify/toml-patch": "0.3.0",
-    "@types/node": "18.19.70",
+    "@types/node": "22.19.17",
     "dotenv": "16.4.7",
     "execa": "^7.2.0",
     "node-pty": "^1.0.0",
--- a/packages/eslint-plugin-cli/rules/no-inline-graphql.js
+++ b/packages/eslint-plugin-cli/rules/no-inline-graphql.js
@@ -2,6 +2,7 @@
 const path = require('path')
 const fs = require('fs')
 const crypto = require('crypto')
+const {execFileSync} = require('child_process')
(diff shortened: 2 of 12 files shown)


packages/eslint-plugin-cli/rules/no-inline-graphql.js
```diff
@@ -31,9 +32,28 @@ function hashFileSync(filePath, algorithm = 'sha256') {
...
+let repoRoot
+function findRepoRoot(filePath) {
+  if (repoRoot === undefined) {
+    repoRoot = execFileSync('git', ['rev-parse', '--show-toplevel'], {
+      cwd: path.dirname(filePath),
+      encoding: 'utf8',
+    }).trim()
+  }
+  return repoRoot
+}
```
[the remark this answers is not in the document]
amcaplan: I'm not really concerned, because it's only going to be run in the context of git repos.

packages/e2e/package.json
```diff
@@ -32,7 +32,7 @@
     "@iarna/toml": "2.2.5",
     "@playwright/test": "^1.50.0",
     "@shopify/toml-patch": "0.3.0",
-    "@types/node": "18.19.70",
+    "@types/node": "22.19.17",
     "dotenv": "16.4.7",
```
gonzaloriestra: We should remove this one as well

amcaplan: Good call, will do!

PR #7731, opened 2026-06-05
Swap jsdom for happy-dom test env (ui-extensions)

amcaplan:
## Swap `jsdom` → `happy-dom` test environment in ui-extensions packages

**Why:** Part of an initiative to cut low-value dependency churn (jsdom: ~2 Dependabot bumps / 24 months). `happy-dom` is a lighter, faster vitest browser environment with no feature gap for these packages' test suites.

**What changed (devDep-only — 4 files):**
- `packages/ui-extensions-dev-console/vite.config.mts` — vitest `environment: 'jsdom'` → `'happy-dom'`
- `packages/ui-extensions-server-kit/vite.config.mts` — vitest `environment: 'jsdom'` → `'happy-dom'`
- both `package.json` — removed `"jsdom": "^25.0.0"`, added `"happy-dom": "^20.10.1"` to devDependencies
- patch changeset for `@shopify/ui-extensions-server-kit` (dev-console is `"private": true`)

**Validation:** All tests pass under happy-dom **with no test-code changes** — the swap is transparent.
| Package | Tests | Lint |
|---|---|---|
| ui-extensions-dev-console | 6 files / 28 tests ✅ | ✅ |
| ui-extensions-server-kit | 7 files / 61 tests ✅ | ✅ |

(server-kit type-check ✅. dev-console has *pre-existing* ReactNode type errors in `.tsx` files untouched by this change — a test-environment/devDep swap cannot affect those.)

**Surface impact:** devDependencies only — the published artifact of `@shopify/ui-extensions-server-kit` is unaffected.

🤖 AI-generated draft — needs human review before merge.

### DIFF
--- a/.changeset/swap-jsdom-happy-dom-ui-extensions.md
+++ b/.changeset/swap-jsdom-happy-dom-ui-extensions.md
@@ -1,5 +1,0 @@
----
-'@shopify/ui-extensions-server-kit': patch
----
-
-Replace `jsdom` devDependency with `happy-dom` as the vitest test environment. This is an internal test-tooling change with no impact on the published package surface.
--- a/.changeset/swap-jsdom-happy-dom-ui-extensions.md
+++ b/.changeset/swap-jsdom-happy-dom-ui-extensions.md
@@ -1,0 +1,5 @@
+---
+'@shopify/ui-extensions-server-kit': patch
+---
+
+Replace `jsdom` devDependency with `happy-dom` as the vitest test environment. This is an internal test-tooling change with no impact on the published package surface.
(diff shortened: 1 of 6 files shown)


gonzaloriestra (commented):
The new dependency is 2x heavier (1.17MB vs 0.55MB) and has 7x more releases (185 vs 24 in the last 2 years).

What's the point?

amcaplan:
@gonzaloriestra you have to consider transitive dependencies.

gonzaloriestra:
Right, but:
- This is a dev dependency, so it won't affect the bundle size
- Jsdom is still present in the lockfile (used by eslint-plugin-cli and vitest)
- We are adding a new dependency

So I'm still not sure about the purpose of this 😅 

What about just deduplicating jsdom instead? We have 25.0.1 and 28.1.0 now

gonzaloriestra (approved):
Ok, now that jsdom was completely removed, I guess it makes sense to switch in order to reduce the number of transitive deps 👍

PR #7732, opened 2026-06-05
Single-source codegen freshness checks via pnpm codegen scripts

dmerand:
### About this stack

This stack came out of reviewing recent contributor and agent sessions in `Shopify/cli`, where the same avoidable friction kept recurring: failures found only after pushing, generated files regenerated by hand, local checks drifting from what CI actually enforces, and setup conventions relearned every time. The goal is to make the local dev loop mirror CI and keep the two single-sourced so they can't drift — then document it so people and agents stop rediscovering it.

---

Two of our PR checks — `Check graphql-codegen has been run` and `Check OCLIF manifests & readme & docs` — spell out their regenerate-and-verify steps right inside `tests-pr.yml`. There's no local command that does the same thing, so the usual way you learn a generated file is stale is by watching CI go red. Then you re-run `refresh-manifests`, `graphql-codegen`, and `build-dev-docs` one at a time and push again. In effect the "what to regenerate" list lives in two heads — the workflow's and yours — and they drift.

Making `package.json` the one place that defines codegen freshness, for people and CI alike:

- `pnpm codegen` regenerates everything.
- `pnpm codegen:check:graphql` / `:oclif` regenerate, then check the tree came back clean.

`bin/check-codegen-clean.js` owns each check's paths and the message you see when it fails (now pointing you at `pnpm codegen`). The two gate jobs simply call these scripts instead of repeating the steps inline.

It's deliberately a no-op in behavior — same generators, same order, same messages — and the job names, runners, and the oclif build step are untouched, so required checks and the merge queue don't notice the difference. Dev tooling only, nothing user-facing.

### How to test your changes?

1. On a clean checkout, `pnpm codegen` should produce no diff (it's idempotent).
2. Add a flag without refreshing, then run `pnpm codegen:check:oclif` — it fails and points you at `pnpm codegen`; running that puts you back to clean.
3. The two gate jobs pass and fail exactly as they did before.

### Notes

- There's intentionally no combined `codegen:check`: the whole-repo `git status` checks step on each other in a single working tree, which is why CI keeps them as separate, fresh-checkout jobs.
- Heads-up: `pnpm codegen` isn't quite idempotent on `main` today — it drops some orphaned `config:autoupgrade` docs that the OCLIF gate doesn't watch (it only checks `generated/*.json`). That's pre-existing and worth a separate fix.

### Post-release steps

None.


dmerand:
* **#7734** <a href="https://app.graphite.com/github/pr/Shopify/cli/7734?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#7751** <a href="https://app.graphite.com/github/pr/Shopify/cli/7751?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#7733** <a href="https://app.graphite.com/github/pr/Shopify/cli/7733?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#7732** <a href="https://app.graphite.com/github/pr/Shopify/cli/7732?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Shopify/cli/7732?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #7733, opened 2026-06-05
Add pnpm pre-ci mirroring PR CI gates, with manifest drift guard

dmerand:
### About this stack

This stack came out of reviewing recent contributor and agent sessions in `Shopify/cli`, where the same avoidable friction kept recurring: failures found only after pushing, generated files regenerated by hand, local checks drifting from what CI actually enforces, and setup conventions relearned every time. The goal is to make the local dev loop mirror CI and keep the two single-sourced so they can't drift — then document it so people and agents stop rediscovering it.

---

Building on the codegen PR below it. Even with that in place, there's still no single command that runs what the PR pipeline actually checks, so lint, type-check, knip, codegen, and test failures tend to surface only after you've pushed. And any helper we write to "run the checks locally" will quietly fall out of step with `tests-pr.yml` as gates get added or the pinned tool versions move.

- `pnpm pre-ci` (also `dev pre-ci`) runs the local gates in the same order CI does — type-check, lint, build, knip, the two codegen checks, unit tests — then prints a summary and tells you which CI-only gates it deliberately skips and why. It mirrors CI's full targets, so green here means green there; it's slower than the affected-only `dev check`, on purpose.
- `bin/ci-gates.js` is the one place that maps every `tests-pr.yml` job (by id) to either a local command or a reason it's CI-only.
- `pnpm check-ci-gates` is the guard: it fails if a workflow job isn't classified, if the manifest names a job that's gone, or if the Node/pnpm versions in `dev.yml` and `tests-pr.yml` disagree. It parses the files with a small, hardened, dependency-free reader (kept lightweight so the CI job needs nothing installed) and is covered by `bin/check-ci-gates.test.js`.
- A small `CI gate manifest` job runs that test and the guard on every PR, so the manifest can't silently drift from the workflow.

Dev tooling only, nothing user-facing.

### How to test your changes?

1. `pnpm check-ci-gates` passes today.
2. Add an unclassified job to `tests-pr.yml`, or bump `DEFAULT_NODE_VERSION` — it fails with a specific message. (Both are covered by the tests.)
3. `pnpm pre-ci` (or `dev pre-ci`) runs the set and prints the summary.

### Notes

- The guard keys on job ids rather than display names, since the matrix jobs interpolate `${{ … }}` into theirs.
- `pre-ci` regenerates files as it checks, so it reminds you to glance at `git status` afterward.
- It does add one new (fast, no-install) required check by design — classify any future gate in `bin/ci-gates.js` to keep it green.

### Post-release steps

None.


dmerand:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/Shopify/cli/7733?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#7734** <a href="https://app.graphite.com/github/pr/Shopify/cli/7734?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#7751** <a href="https://app.graphite.com/github/pr/Shopify/cli/7751?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#7733** <a href="https://app.graphite.com/github/pr/Shopify/cli/7733?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Shopify/cli/7733?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#7732** <a href="https://app.graphite.com/github/pr/Shopify/cli/7732?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

PR #7747, opened 2026-06-08
Improve performance for extension types generation

gonzaloriestra:

Re: https://community.shopify.dev/t/cli-type-generation-for-ui-extensions-reads-too-many-files-restrict-scanner-to-tsconfig-include-or-toml-entry/34957

UI extension type generation can recursively scan too many files, including external imports and files excluded by a project's TypeScript config. This can make `shopify app build` and app loading noticeably slower.

- Limits recursive import scanning to files inside the extension directory.
- Skips type-only imports/exports and `.d.ts` files.
- Respects explicit `tsconfig.json` `files` / `include` lists when present.
- Reuses an import cache across repeated scans for UI extension targets.

### How to test your changes?

`shopify app build`

### Checklist

### DIFF
--- a/packages/app/src/cli/models/extensions/specifications/type-generation.test.ts
+++ b/packages/app/src/cli/models/extensions/specifications/type-generation.test.ts
@@ -25,6 +25,72 @@ describe('getGeneratedTypesHelperImportPath', () => {
 })
 
 describe('findAllImportedFiles', () => {
+  test('ignores commented-out imports', async () => {
+    await inTemporaryDirectory(async (tmpDir) => {
+      const entryPath = joinPath(tmpDir, 'index.ts')
+      const commentedPath = joinPath(tmpDir, 'commented.ts')
+
+      await writeFile(
+        entryPath,
+        `
+          // import './commented.ts'
+          /*
+            import './commented.ts'
+          */
+        `,
+      )
+      await writeFile(commentedPath, `export const commented = true`)
+
+      const importedFiles = (await findAllImportedFiles(entryPath)).map((file) => normalizePath(file))
+
+      expect(importedFiles).not.toContain(normalizePath(commentedPath))
+    })
+  })
+
+  test('does not follow type-only imports and exports', async () => {
+    await inTemporaryDirectory(async (tmpDir) => {
+      const entryPath = joinPath(tmpDir, 'index.ts')
+      const valuePath = joinPath(tmpDir, 'value.ts')
+      const valueNestedPath = joinPath(tmpDir, 'value-nested.ts')
+      const mixedValuePath = joinPath(tmpDir, 'mixed-value.ts')
+      const mixedExportPath = joinPath(tmpDir, 'mixed-export.ts')
+      const typePath = joinPath(tmpDir, 'types.ts')
+      const typeNestedPath = joinPath(tmpDir, 'type-nested.ts')
+      const exportedTypePath = joinPath(tmpDir, 'exported-types.ts')
+      const exportedTypeNestedPath = joinPath(tmpDir, 'exported-type-nested.ts')
+
+      await writeFile(
+        entryPath,
+        `
+          import './value.ts'
+          import MixedValue, { type MixedType } from './mixed-value.ts'
+          import type { TypeOnly } from './types.ts'
+          export { mixedValue, type MixedExportType } from './mixed-export.ts'
+          export type { ExportedTypeOnly } from './exported-types.ts'
+        `,
+      )
+      await writeFile(valuePath, `import './value-nested.ts'`)
+      await writeFile(valueNestedPath, `export const valueNested = true`)
+      await writeFile(mixedValuePath, `export default true; export type MixedType = string`)
+      await writeFile(mixedExportPath, `export const mixedValue = true; export type MixedExportType = string`)
+      await writeFile(typePath, `import './type-nested.ts'; export type TypeOnly = string`)
+      await writeFile(typeNestedPath, `export const typeNested = true`)
+      await writeFile(exportedTypePath, `import './exported-type-nested.ts'; export type ExportedTypeOnly = string`)
+      await writeFile(exportedTypeNestedPath, `export const exportedTypeNested = true`)
+
+      const importedFiles = (await findAllImportedFiles(entryPath)).map((file) => normalizePath(file))
+
+      expect(importedFiles).toContain(normalizePath(valuePath))
+      expect(importedFiles).toContain(normalizePath(valueNestedPath))
+      expect(importedFiles).toContain(normalizePath(mixedValuePath))
+      expect(importedFiles).toContain(normalizePath(mixedExportPath))
+      expect(importedFiles).not.toContain(normalizePath(typePath))
+      expect(importedFiles).not.toContain(normalizePath(typeNestedPath))
+      expect(importedFiles).not.toContain(normalizePath(exportedTypePath))
+      expect(importedFiles).not.toContain(normalizePath(exportedTypeNestedPath))
+    })
+  })
+
   test('stops recursive import scanning at the boundary directory', async () => {
     await inTemporaryDirectory(async (tmpDir) => {
       const extensionDir = joinPath(tmpDir, 'extensions', 'extension')
--- a/packages/app/src/cli/models/extensions/specifications/type-generation.test.ts
+++ b/packages/app/src/cli/models/extensions/specifications/type-generation.test.ts
@@ -1,9 +1,12 @@
 import {
   createIntentsTypeDefinition,
   createToolsTypeDefinition,
+  findAllImportedFiles,
   getGeneratedTypesHelperImportPath,
 } from './type-generation.js'
 import {AbortError} from '@shopify/cli-kit/node/error'
+import {inTemporaryDirectory, mkdir, writeFile} from '@shopify/cli-kit/node/fs'
+import {joinPath, normalizePath} from '@shopify/cli-kit/node/path'
 import {describe, expect, test} from 'vitest'
 
 const adminGeneratedTypesHelperImportPath = '@shopify/ui-extensions/admin'
@@ -18,6 +21,47 @@ describe('getGeneratedTypesHelperImportPath', () => {
     expect(getGeneratedTypesHelperImportPath(['customer-account.order-status.block.render'])).toBe(
       '@shopify/ui-extensions/customer-account',
     )
+  })
+})
+
+describe('findAllImportedFiles', () => {
+  test('stops recursive import scanning at the boundary directory', async () => {
+    await inTemporaryDirectory(async (tmpDir) => {
+      const extensionDir = joinPath(tmpDir, 'extensions', 'extension')
+      const srcDir = joinPath(extensionDir, 'src')
+      const sharedDir = joinPath(tmpDir, 'shared')
+
+      await mkdir(extensionDir)
+      await mkdir(srcDir)
+      await mkdir(sharedDir)
+
+      const entryPath = joinPath(srcDir, 'index.ts')
+      const localPath = joinPath(srcDir, 'local.ts')
+      const nestedPath = joinPath(srcDir, 'nested.ts')
+      const externalPath = joinPath(sharedDir, 'utils.ts')
+      const externalNestedPath = joinPath(sharedDir, 'secret.ts')
+
+      await writeFile(
+        entryPath,
+        `
+          import './local.ts'
+          import '../../../shared/utils.ts'
+        `,
+      )
+      await writeFile(localPath, `import './nested.ts'`)
+      await writeFile(nestedPath, `export const nested = true`)
+      await writeFile(externalPath, `import './secret.ts'`)
+      await writeFile(externalNestedPath, `export const secret = true`)
+
+      const importedFiles = (await findAllImportedFiles(entryPath, {boundaryDirectory: extensionDir})).map((file) =>
+        normalizePath(file),
+      )
+
+      expect(importedFiles).toContain(normalizePath(localPath))
+      expect(importedFiles).toContain(normalizePath(nestedPath))
+      expect(importedFiles).not.toContain(normalizePath(externalPath))
+      expect(importedFiles).not.toContain(normalizePath(externalNestedPath))
+    })
   })
 })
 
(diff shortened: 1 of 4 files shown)


packages/app/src/cli/models/extensions/specifications/type-generation.test.ts
isaacroldan, line 1: This change is worth a changeset

packages/app/src/cli/models/extensions/specifications/type-generation.ts
```diff
@@ -46,23 +46,27 @@ export function parseApiVersion(apiVersion: string): {year: number; month: numbe
...
   }
 
   const configFile = ts.readConfigFile(configPath, ts.sys.readFile.bind(ts.sys))
   if (configFile.error) {
-    return {compilerOptions: {}, configPath}
+    return {compilerOptions: {}, configPath, hasExplicitFiles: false}
   }
 
   const parsedConfig = ts.parseJsonConfigFileContent(configFile.config, ts.sys, dirname(configPath))
+  const hasExplicitFiles = Boolean(configFile.config.files ?? configFile.config.include)
```
isaacroldan: 🐛 **Bug**: `hasExplicitFiles` is derived from `configFile.config` (the raw JSON from `ts.readConfigFile`), which does NOT resolve `extends`. A tsconfig that gets its `include`/`files` from a base config (e.g. `extends: '../../tsconfig.base.json'`) reports `hasExplicitFiles: false`, so `findExplicitTsConfigFiles` returns `undefined`, no allowlist is applied, and the scanner falls back to boundary-only scanning. Meanwhile `parsedConfig.fileNames` (which the allowlist would use) DOES honor `extends`, so the gate and the data are inconsistent. This degrades gracefully rather than producing wrong output, but it silently defeats the tsconfig-restriction optimization for the common monorepo pattern of extensions extending a shared base tsconfig — precisely the perf win this PR targets.

**Suggestion:** Consider deriving hasExplicitFiles from the extends-resolved config rather than the raw one — e.g. read the resolved include/files from parsedConfig.raw, or gate on parsedConfig.fileNames being non-empty, so the allowlist path fires for extensions that inherit their file list.

packages/app/src/cli/models/extensions/specifications/type-generation.ts
```diff
@@ -119,13 +164,38 @@ async function parseAndResolveImports(filePath: string): Promise<string[]> {
...
       if (ts.isImportDeclaration(node) && node.moduleSpecifier && ts.isStringLiteral(node.moduleSpecifier)) {
+        if (node.importClause?.isTypeOnly) {
+          return
+        }
+
+        if (
+          !node.importClause?.name &&
+          node.importClause?.namedBindings &&
+          ts.isNamedImports(node.importClause.namedBindings) &&
+          node.importClause.namedBindings.elements.every((element) => element.isTypeOnly)
```
isaacroldan: 🐛 **Bug**: `node.importClause.namedBindings.elements.every((element) => element.isTypeOnly)` returns `true` for an empty `elements` array, so `import {} from './x'` is skipped as if it were type-only; the same applies to the export branch (`export {} from './x'`). Both forms still trigger module evaluation for side effects per the ES spec, so the target file would be incorrectly excluded from scanning. This is a real edge case but extremely rare in practice — side-effect imports are almost always written as `import './x'` — so practical impact is low.

**Suggestion:** Guard the every() checks with a non-empty length so empty braces aren't treated as type-only, e.g. `node.importClause.namedBindings.elements.length > 0 && node.importClause.namedBindings.elements.every(...)` (and likewise for the export branch).

packages/app/src/cli/models/extensions/specifications/type-generation.ts
```diff
@@ -170,26 +239,48 @@ async function parseAndResolveImports(filePath: string): Promise<string[]> {
...
   }
 
   return uniq(allFiles)
 }
 
+export async function findExplicitTsConfigFiles(
+  fromFile: string,
+  extensionDirectory: string,
+): Promise<Set<string> | undefined> {
+  const {configPath, fileNames, hasExplicitFiles} = await loadTsConfig(fromFile)
```
isaacroldan: ⚡ **Performance**: This is explicitly a performance change (it introduces `importCache` to avoid re-reading files), yet `loadTsConfig` remains uncached even as this PR adds new call sites to it. `ts.parseJsonConfigFileContent` walks the filesystem to expand the tsconfig `include`/`files`/`exclude` globs into `fileNames`. It is now invoked both per-scanned-file inside `parseAndResolveImports` (on every cache miss, where only `compilerOptions` is consumed and the glob expansion is wasted) and once per extension point / should_render via the new `findExplicitTsConfigFiles` (ui_extension.ts:259, 282). Since the extension points of one extension typically resolve to the same tsconfig, that becomes repeated full tsconfig glob expansions per build. The new `importCache` short-circuits the file-read path but not this tsconfig-parse path.

**Suggestion:** Consider memoizing loadTsConfig results (keyed by the resolved configPath) and threading that cache alongside importCache through the options, so parseJsonConfigFileContent runs once per tsconfig rather than once per file and once per entry point. This would be a natural, consistent extension of the caching approach this PR introduces.

packages/app/src/cli/models/extensions/specifications/ui_extension.ts
```diff
@@ -254,7 +256,13 @@ const uiExtensionSpec = createExtensionSpecification({
       if (!exists) continue
 
       // Find all imported files recursively
-      const importedFiles = await findAllImportedFiles(fullPath)
+      const allowedFiles = await findExplicitTsConfigFiles(fullPath, extension.directory)
+      const importedFiles = await findAllImportedFiles(fullPath, {
+        boundaryDirectory: extension.directory,
+        allowedFiles,
+        alwaysAllowedFiles: new Set([fullPath]),
```
isaacroldan: 🎨 **Code Style**: `isAllowedFile` resolves the lookup key (`resolvePath(filePath)`) and checks both `allowedFiles` and `alwaysAllowedFiles` against that resolved form, and `findExplicitTsConfigFiles` also normalizes its entries via `resolvePath`. So the established convention is that these sets contain resolved paths. Here (and at line 286) the caller seeds `alwaysAllowedFiles` with the raw `fullPath`/`shouldRenderPath` instead. In practice `fullPath = joinPath(extension.directory, extensionPoint.module)` is normally already absolute and normalized, so the lookup matches and there is no runtime impact today — but the mismatch is fragile: if `extension.directory` were ever relative or not fully normalized, the membership check would silently miss.

**Suggestion:** 
```suggestion
        alwaysAllowedFiles: new Set([resolvePath(fullPath)]),
```

isaacroldan (commented):
---

packages/app/src/cli/models/extensions/specifications/type-generation.ts
gonzaloriestra: Fixed in 96a37f89b4. Added a shared `tsConfigCache`, keyed by resolved config path, and threaded it through explicit tsconfig lookup plus recursive import parsing for both main and `should_render` scans.

packages/app/src/cli/models/extensions/specifications/type-generation.test.ts
gonzaloriestra: Added in 96a37f89b4 via `.changeset/skip-external-type-generation-imports.md`.

packages/app/src/cli/models/extensions/specifications/ui_extension.ts
gonzaloriestra: Fixed in 96a37f89b4. `alwaysAllowedFiles` now stores resolved paths for both the main module and `should_render`, matching the lookup convention in `isAllowedFile`.

packages/app/src/cli/models/extensions/specifications/type-generation.ts
gonzaloriestra: Fixed in 96a37f89b4. The type-only named import/export checks now require a non-empty element list, so `import {} from ...` and `export {} from ...` are still followed. Added coverage for both forms.

packages/app/src/cli/models/extensions/specifications/type-generation.ts
gonzaloriestra: Fixed in 96a37f89b4. `hasExplicitFiles` now uses `parsedConfig.raw`, which includes inherited `include`/`files` from extended tsconfigs, and I added regression coverage for an extension tsconfig extending a shared base config.

packages/app/src/cli/models/extensions/specifications/type-generation.ts
```diff
@@ -170,26 +260,49 @@ async function parseAndResolveImports(filePath: string): Promise<string[]> {
...
 
+export async function findExplicitTsConfigFiles(
+  fromFile: string,
+  extensionDirectory: string,
+  options: {tsConfigCache?: TsConfigCache} = {},
+): Promise<Set<string> | undefined> {
+  const {configPath, fileNames, hasExplicitFiles} = await loadTsConfig(fromFile, options.tsConfigCache)
+  if (!configPath || !hasExplicitFiles) return
+
+  if (!isWithinBoundary(configPath, extensionDirectory)) return
```
vividviolet: Can we remove this? Many apps share utils imported outside of the immediate extension folder and we'd be breaking that

packages/app/src/cli/models/extensions/specifications/type-generation.ts
```diff
@@ -119,13 +183,40 @@ async function parseAndResolveImports(filePath: string): Promise<string[]> {
...
         if (firstArg && ts.isStringLiteral(firstArg)) {
           importPaths.push(firstArg.text)
         }
       } else if (ts.isExportDeclaration(node) && node.moduleSpecifier && ts.isStringLiteral(node.moduleSpecifier)) {
+        if (node.isTypeOnly) {
+          return
+        }
+
+        if (
+          node.exportClause &&
```
vividviolet: Can you make the type only check a reusable util?

packages/app/src/cli/models/extensions/specifications/type-generation.ts
```diff
@@ -147,9 +238,7 @@ async function parseAndResolveImports(filePath: string): Promise<string[]> {
       if (resolvedModule.resolvedModule?.resolvedFileName) {
         const resolvedPath = resolvedModule.resolvedModule.resolvedFileName
 
-        if (!resolvedPath.includes('node_modules')) {
-          resolvedPaths.push(resolvedPath)
-        }
+        resolvedPaths.push(resolvedPath)
```
vividviolet: Why was this changed?

gonzaloriestra: Good call. This had moved to the `isScannableFile` filter, but that made the lower-level helper broader than necessary. Restored the local `node_modules` guard at the resolution site in 0d42833959 and kept `isScannableFile` as the recursive safety net.

packages/app/src/cli/models/extensions/specifications/type-generation.ts
gonzaloriestra: Removed. A shared/root tsconfig can now provide the explicit `files` / `include` set; we still filter the resolved `fileNames` to the extension directory so type generation stays bounded while apps can continue importing shared utils outside the immediate extension folder.

packages/app/src/cli/models/extensions/specifications/type-generation.ts
gonzaloriestra: Done. I extracted the checks into `isTypeOnlyImport` / `isTypeOnlyExport` with a shared `hasOnlyTypeOnlyElements` helper, keeping the non-empty element guard so `import {} from ...` and `export {} from ...` are still followed.

packages/app/src/cli/models/extensions/specifications/type-generation.ts
```diff
@@ -95,14 +116,79 @@ async function fallbackResolve(importPath: string, baseDir: string): Promise<str
...
+
+  const resolvedPath = resolvePath(filePath)
+  return options.allowedFiles.has(resolvedPath) || Boolean(options.alwaysAllowedFiles?.has(resolvedPath))
+}
+
+function isScannableFile(filePath: string, options: FindAllImportedFilesOptions): boolean {
+  return (
+    !filePath.includes('node_modules') &&
+    !filePath.endsWith('.d.ts') &&
+    isWithinBoundary(filePath, options.boundaryDirectory) &&
```
vividviolet: This is still restricting the imports to be within the extension directory which makes shared utils not work. I just realized though that the [pr](https://github.com/Shopify/cli/pull/6957) to do shared util generation haven't shipped yet. @justinhenricks can you revive the PR? We should rebase it on top of this

gonzaloriestra: Fixed in 213c3a4399. The scanner now separates including an imported file from recursively scanning through it, so shared imports outside the extension directory are still returned. We only recurse outside the boundary when the file is explicitly allowed by tsconfig, which keeps the perf guard in place and should make the shared-util generation PR easier to rebase on top.

packages/app/src/cli/models/extensions/specifications/type-generation.test.ts
```diff
@@ -21,6 +26,331 @@ describe('getGeneratedTypesHelperImportPath', () => {
...
+      expect(importedFiles).toContain(normalizePath(emptyExportPath))
+      expect(importedFiles).toContain(normalizePath(emptyExportNestedPath))
+      expect(importedFiles).not.toContain(normalizePath(typePath))
+      expect(importedFiles).not.toContain(normalizePath(typeNestedPath))
+      expect(importedFiles).not.toContain(normalizePath(exportedTypePath))
+      expect(importedFiles).not.toContain(normalizePath(exportedTypeNestedPath))
+    })
+  })
+
+  test('includes imports outside the boundary directory without recursively scanning them', async () => {
```
vividviolet, line 110: Why wouldn't we scan these files? What if the have their own imports?

vividviolet (approved):
Tophat'd - it's working well. Once we support generating types for shared files we might have to make some changes around the extension boundary but that is out of scope for this P

packages/app/src/cli/models/extensions/specifications/type-generation.test.ts
gonzaloriestra: To improve performance in big projects. It's true that we could miss some external files now, but we need to find a balance. After some tests, the user confirmed that they can run dev again with this branch... We can revisit this later with the other work.

isaacroldan (approved):

PR #7757, opened 2026-06-08
chore: use temp runner for the shopify-dev preview automation

majd-shopify:
We need to temporarily use shopify-docs-ubuntu-latest because ubuntu-latest has been blocked from communicating with shopify-dev. Until we find an alternative.


majd-shopify:
No longer relevant

PR #7758, opened 2026-06-08
Add is_first_party label to CLI command metrics

SimonSiju:

closes: https://github.com/shop/issues-develop/issues/22933

Part of the 1P org migration observability work (GSD #50881). "1P CI/CD pipelines function without breakage after cutover" is a named guardrail on the **[Dev Platform] 1P Orgs Migration** dashboard (Observe `si2kjvt`), but today it can't be measured: `cli_commands_total` (and the companion duration/elapsed metrics) carry only `cli_version`, `exit`, and `job` labels — no org / 1P dimension — so the CLI guardrail panels blend 1P + 3P + all external developers.

Adds an `is_first_party` label to the CLI command metrics emitted from `cli-kit`:

- `cli_commands_total`
- `cli_commands_duration_ms`
- `cli_commands_wall_clock_elapsed_ms`

The label is keyed off the 1P dev path via `firstPartyDev()` (the `SHOPIFY_CLI_1P_DEV` env var), which is the same signal that sets the `X-Shopify-Cli-Employee` header. Values are the strings `"true"` / `"false"`, keeping cardinality low.

This lets the dashboard's CLI guardrail panels filter to 1P, so we get a 1P-scoped success/failure signal for commands like `app dev`, `app dev clean`, and `generate extension` after cutover.

**Files changed:**

- `packages/cli-kit/src/private/node/otel-metrics.ts` — added `is_first_party` to the `Labels` type and populated it from `firstPartyDev()` in `recordMetrics`; imported `firstPartyDev` from `context/local`.
- `packages/cli-kit/src/private/node/otel-metrics.test.ts` — added a test asserting all recorded metrics carry `is_first_party: "true"` when `SHOPIFY_CLI_1P_DEV` is set; added `afterEach` env cleanup.
- `packages/cli-kit/src/private/node/__snapshots__/otel-metrics.test.ts.snap` — regenerated to include the new label (`"false"` when the 1P dev path is off).
- `.changeset/first-party-cli-metrics-label.md` — `@shopify/cli-kit` patch changeset.

### How to test your changes?

1. From the repo root, run the unit tests for the metrics module:
   ```
   pnpm --filter @shopify/cli-kit vitest run src/private/node/otel-metrics.test.ts
   ```
   All three tests should pass, including `labels metrics as first-party when the 1P dev path is enabled`.
2. Manually verify the label is emitted. With metrics debug output enabled, run a CLI command as a third-party developer and confirm `cli_commands_total` records `is_first_party: "false"`.
3. Re-run the same command with the 1P dev path enabled (`SHOPIFY_CLI_1P_DEV=1 shopify app dev`) and confirm the metric records `is_first_party: "true"`.

### Checklist

### DIFF
--- a/.changeset/first-party-cli-metrics-label.md
+++ b/.changeset/first-party-cli-metrics-label.md
@@ -1,5 +1,0 @@
----
-'@shopify/cli-kit': patch
----
-
-Add an `is_first_party` label to the CLI command metrics (`cli_commands_total`, `cli_commands_duration_ms`, `cli_commands_wall_clock_elapsed_ms`) so first-party (1P) Shopify developer usage can be distinguished from third-party usage in observability dashboards. The label is keyed off the 1P dev path (`SHOPIFY_CLI_1P_DEV`), the same signal that sets the `X-Shopify-Cli-Employee` header.
--- a/.changeset/first-party-cli-metrics-label.md
+++ b/.changeset/first-party-cli-metrics-label.md
@@ -1,0 +1,5 @@
+---
+'@shopify/cli-kit': patch
+---
+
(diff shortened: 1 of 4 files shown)


murray-andrew-r (changes requested):
I don't think CLI is the right place to inject these metrics. We don't have guarantees that CLI will get updated in time, and I'm generally leery of trusting the client in general.

Let's look for another solution.

PR #7764, opened 2026-06-09
Add production preview store create command

alfonso-noriega:

This productionizes `shopify store create preview` on top of the shipped preview-store backend endpoint so an agent can create a preview store and immediately use the returned Admin API token through existing store command auth plumbing.

The command targets the production endpoint contract from shop/world:

- unauthenticated `POST /services/preview-stores`
- request body with optional top-level `name`
- optional country sent as `variables.storeCreatePayload.country`
- CLI attribution headers for rollout/rate-limit/tracking
- response body with `shop` details, `placeholder_account_uuid`, `admin_api_token`, and `access_url`

- Adds `shopify store create preview` with flags:
    - `--name`
    - `--country` with two-letter code validation
    - `--json`
    - existing global `--no-color` / `--verbose`
- Calls `https://<app-management-fqdn>/services/preview-stores` without Basic auth or Identity auth.
- Sends:
    - `X-Shopify-CLI-Instance` from a stable locally persisted install id
    - `X-Shopify-CLI-Version`
    - `User-Agent`
    - JSON accept/content headers
    - `variables.storeCreatePayload.country` when `--country` is provided
- Parses the production response shape:
    - `shop.id`
    - `shop.name`
    - `shop.domain`
    - `placeholder_account_uuid` when present
    - `admin_api_token`
    - `access_url`
- Persists the returned Admin API token into the existing local store-auth session cache, tagged as `kind: 'preview'`, so the store is immediately usable by `shopify store execute --store <domain>` without `shopify store auth`.
- Persists preview-store metadata, including the returned access URL, in the store-auth session cache.
- Prints the backend-returned access URL in text output and includes it in JSON output.
- Redacts both `admin_api_token` and tokenized `access_url` from malformed-response diagnostics.
- Adds generated command README, OCLIF manifest, shopify.dev command interface docs, and a changeset.

### Notes / open questions

- Preview-store token expiry is intentionally not managed in this PR.
- The CLI handles both the requested `422 preview_store_create_failed` path and the currently observed backend `500 preview_store_create_failed` path defensively.
- Store-auth persistence is performed before best-effort store FQDN metadata recording so a metadata failure does not orphan a successfully created preview store without local credentials.

### How to test your changes?

Ideally, the endpoint will be ready in prod to test this (protected behind a flag) and we can test as follows:

- Enable [flag](https://experiments.shopify.io/flags/31810) for your CLI instance UUID
- Run from this branch `pnpm shopify store create preview --country US`
- A new preview store should be created
- The returned access URL should open the preview store
- `pnpm shopify store execute --store <returned-domain>` should use the cached preview-store Admin API token

Local validation run:

- `pnpm --filter @shopify/store exec vitest run src/cli/commands/store/create/preview.test.ts src/cli/services/store/create/preview/client.test.ts src/cli/services/store/create/preview/index.test.ts src/cli/services/store/create/preview/result.test.ts src/cli/services/store/auth/session-store.test.ts`
- `pnpm nx run store:lint --skip-nx-cache --output-style=stream`
- `pnpm --filter @shopify/store run type-check`
- `/usr/bin/git diff --check`

### Post-release steps

None.

### Checklist

- [x] I've considered possible cross-platform impacts (Mac, Linux, Windows) — local persistence uses existing `LocalStorage`; command logic is platform-neutral.
- [x] I've considered possible [documentation](https://shopify.dev) changes — command README, OCLIF manifest, generated docs interface, and generated docs data are updated.
- [x] I've considered analytics changes to measure impact — request sends CLI instance/version/user-agent headers read by the backend tracking path; store FQDN metadata is recorded after persistence.
- [x] The change is user-facing — minor changeset added for `@shopify/cli` and `@shopify/store`.

### DIFF
--- a/packages/store/src/cli/commands/store/create/preview.ts
+++ b/packages/store/src/cli/commands/store/create/preview.ts
@@ -1,0 +1,50 @@
+import {isCountryCode, previewStoreFlags} from '../../../flags.js'
+import {type CreatePreviewStoreResult, createPreviewStoreCommand} from '../../../services/store/create/preview/index.js'
+import {writeCreatePreviewStoreResult} from '../../../services/store/create/preview/result.js'
+import StoreCommand from '../../../utilities/store-command.js'
+import {globalFlags, jsonFlag} from '@shopify/cli-kit/node/cli'
+import {outputContent} from '@shopify/cli-kit/node/output'
+import {renderSingleTask} from '@shopify/cli-kit/node/ui'
+import {Flags} from '@oclif/core'
+
+export default class StoreCreatePreview extends StoreCommand {
+  static hidden = true
+
+  static summary = 'Create a preview Shopify store.'
+
+  static descriptionWithMarkdown = `Creates a new Shopify store, with no need for an existing account.`
+
+  static description = this.descriptionWithoutMarkdown()
+
+  static examples = [
+    '<%= config.bin %> <%= command.id %> --name "Lavender Candles"',
+    '<%= config.bin %> <%= command.id %> --name "Lavender Candles" --country US',
+    '<%= config.bin %> <%= command.id %> --name "Lavender Candles" --json',
+  ]
+
+  static flags = {
+    ...globalFlags,
+    ...jsonFlag,
+    name: Flags.string({
+      description: 'The name of the store.',
+      env: 'SHOPIFY_FLAG_PREVIEW_STORE_NAME',
+      required: false,
+    }),
+    country: previewStoreFlags.country,
+  }
+
+  public async run(): Promise<void> {
+    const {flags} = await this.parse(StoreCreatePreview)
+
+    if (flags.country !== undefined && !isCountryCode(flags.country)) {
+      this.error('Country must be a two-letter country code, for example: US.')
+    }
+
+    const result = await renderSingleTask<CreatePreviewStoreResult>({
+      title: outputContent`Creating store…`,
+      task: async () => createPreviewStoreCommand({name: flags.name, country: flags.country}),
+    })
+
+    writeCreatePreviewStoreResult(result, flags.json ? 'json' : 'text')
+  }
+}
--- a/packages/store/src/cli/services/store/create/preview/client.ts
+++ b/packages/store/src/cli/services/store/create/preview/client.ts
@@ -1,0 +1,224 @@
+import {appManagementFqdn, normalizeStoreFqdn} from '@shopify/cli-kit/node/context/fqdn'
+import {randomUUID} from '@shopify/cli-kit/node/crypto'
+import {AbortError} from '@shopify/cli-kit/node/error'
+import {shopifyFetch} from '@shopify/cli-kit/node/http'
+import {LocalStorage} from '@shopify/cli-kit/node/local-storage'
+import {CLI_KIT_VERSION} from '@shopify/cli-kit/common/version'
+
+export const CLI_INSTANCE_HEADER = 'X-Shopify-CLI-Instance'
+export const CLI_VERSION_HEADER = 'X-Shopify-CLI-Version'
+
+interface PreviewStoreClientStorageSchema {
+  cliInstanceId?: string
+}
+
+let _clientStorage: LocalStorage<PreviewStoreClientStorageSchema> | undefined
+
+function clientStorage() {
+  _clientStorage ??= new LocalStorage<PreviewStoreClientStorageSchema>({projectName: 'shopify-cli-store'})
+  return _clientStorage
+}
+
+export interface PreviewStoreClientOptions {
+  storage?: LocalStorage<PreviewStoreClientStorageSchema>
+}
+
+interface PreviewStoreCreateRequest {
+  name?: string
+  country?: string
+}
+
+interface PreviewStoreResponseShop {
+  id: string
+  name: string
+  domain: string
+}
+
+export interface PreviewStoreCreateResponse {
+  shop: PreviewStoreResponseShop
+  placeholderAccountUuid?: string
+  adminApiToken: string
+  accessUrl: string
+}
+
+interface RawPreviewStoreResponseShop {
+  id?: unknown
+  name?: unknown
+  domain?: unknown
+}
+
+interface RawPreviewStoreCreateResponse {
+  shop?: RawPreviewStoreResponseShop
+  placeholder_account_uuid?: unknown
+  admin_api_token?: unknown
+  access_url?: unknown
+}
+
+interface RawPreviewStoreErrorResponse {
+  error_code?: string
+  message?: string
+}
+
+export function getOrCreateCliInstanceId(
+  storage: LocalStorage<PreviewStoreClientStorageSchema> = clientStorage(),
+): string {
+  const existing = storage.get('cliInstanceId')
+  if (typeof existing === 'string' && existing.length > 0) return existing
+
+  const next = randomUUID()
+  storage.set('cliInstanceId', next)
+  return next
+}
+
+export function previewStoreCreateHeaders(cliInstanceId: string): Record<string, string> {
+  return {
+    Accept: 'application/json',
+    'Content-Type': 'application/json',
+    'User-Agent': `Shopify CLI; v=${CLI_KIT_VERSION}`,
+    [CLI_INSTANCE_HEADER]: cliInstanceId,
+    [CLI_VERSION_HEADER]: CLI_KIT_VERSION,
+  }
+}
+
+export async function createPreviewStore(
+  request: PreviewStoreCreateRequest,
+  options: PreviewStoreClientOptions = {},
+): Promise<PreviewStoreCreateResponse> {
+  const fqdn = await appManagementFqdn()
+  const url = `https://${fqdn}/services/preview-stores`
+  const body = JSON.stringify({
+    ...(request.name ? {name: request.name} : {}),
+    ...(request.country
+      ? {
+          variables: {
+            storeCreatePayload: {
+              country: request.country,
+            },
+          },
+        }
+      : {}),
+  })
+
+  const response = await shopifyFetch(url, {
+    method: 'POST',
+    headers: previewStoreCreateHeaders(getOrCreateCliInstanceId(options.storage)),
+    body,
+  })
+
+  const rawText = await response.text()
+  if (!response.ok) {
+    const error = previewStoreError(response.status, rawText)
+    throw new AbortError(error.message, error.tryMessage)
+  }
+
+  let parsed: RawPreviewStoreCreateResponse
+  try {
+    parsed = JSON.parse(rawText) as RawPreviewStoreCreateResponse
+  } catch (error) {
+    const message = error instanceof Error ? error.message : 'Unknown error'
+    throw new AbortError(
+      'Preview store creation returned a non-JSON response.',
+      `Parse error: ${message}. Body (truncated): ${redactPreviewStoreRawText(rawText).slice(0, 500)}`,
+    )
+  }
+
+  return narrowResponse(parsed)
+}
+
+function previewStoreError(status: number, rawText: string): {message: string; tryMessage?: string} {
+  const parsed = parseErrorBody(rawText)
+  const errorCode = parsed.error_code
+  const message = parsed.message
+
+  if (errorCode === 'not_in_rollout') {
+    return {
+      message: 'Preview store creation is not enabled yet.',
+      tryMessage: 'Try again later.',
+    }
+  }
+
+  if (errorCode === 'service_unavailable') {
+    return {message: 'Preview store creation is temporarily unavailable.', tryMessage: 'Try again later.'}
+  }
+
+  if (errorCode === 'rate_limited') {
+    return {message: 'Too many preview store creation requests.', tryMessage: 'Try again later.'}
+  }
+
+  if (errorCode === 'preview_store_create_failed') {
+    return {message: 'Preview store creation failed.', tryMessage: 'Try again later.'}
+  }
+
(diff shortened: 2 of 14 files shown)


alfonso-noriega:
* **#7788** <a href="https://app.graphite.com/github/pr/Shopify/cli/7788?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#7783](https://github.com/Shopify/cli/pull/7783) <a href="https://app.graphite.com/github/pr/Shopify/cli/7783?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#7831](https://github.com/Shopify/cli/pull/7831) <a href="https://app.graphite.com/github/pr/Shopify/cli/7831?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#7764** <a href="https://app.graphite.com/github/pr/Shopify/cli/7764?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Shopify/cli/7764?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/store/src/cli/services/store/create/preview/index.ts
```diff
@@ -0,0 +1,106 @@
...
+
+async function persistPreviewStoreSession(
+  response: PreviewStoreCreateResponse,
+  country: string | undefined,
+  dependencies: CreatePreviewStoreDependencies,
+): Promise<CreatePreviewStoreResult> {
+  const acquiredAt = dependencies.now().toISOString()
+  const userId = previewUserId(response)
+
+  await dependencies.recordStoreFqdnMetadata(response.shop.domain, false)
```
tizmagik: 🐛 **Bug**: Two concerns at this call site, both pointing at the same fix:

1. **Ordering risk**: Both `recordStoreFqdnMetadata` calls precede `setStoredStoreAppSession`. The test `does not persist a store session when recording store metadata fails` locks this in, but at that point the backend has already created the preview store and issued an admin token — if metadata throws (e.g., 'bubble' contexts), the token is dropped and the merchant is left with an orphaned store and no local credential. Metadata is best-effort observability; it should not gate token persistence.

2. **Redundant pair of calls**: In `auth/index.ts` the `validated:false` → `validated:true` transition straddles the OAuth handshake, so both states are meaningful. Here the two calls fire back-to-back with no intervening validation step, and `recordStoreFqdnMetadata` `Object.assign`s into a shared bag — so the `false` write is immediately overwritten by `true` and adds no telemetry signal.

**Suggestion:** Persist the session first, then record metadata once with `true`:

```suggestion
  dependencies.setStoredStoreAppSession({
    store: response.shop.domain,
    clientId: STORE_AUTH_APP_CLIENT_ID,
    userId,
    accessToken: response.adminApiToken,
    scopes: [],
    acquiredAt,
    kind: 'preview',
    preview: {
      shopId: response.shop.id,
      name: response.shop.name,
      createdAt: acquiredAt,
      ...(response.placeholderAccountUuid ? {placeholderAccountUuid: response.placeholderAccountUuid} : {}),
      ...(country ? {country} : {}),
    },
  })
  dependencies.setLastSeenUserId(userId)
  await dependencies.recordStoreFqdnMetadata(response.shop.domain, true)
```

tizmagik (approved):
One minor thing inline, otherwise lgtm

---

packages/store/src/cli/services/store/create/preview/client.ts
```diff
@@ -0,0 +1,211 @@
...
+export const CLI_VERSION_HEADER = 'X-Shopify-CLI-Version'
+
+interface PreviewStoreClientStorageSchema {
+  cliInstanceId?: string
+}
+
+let _clientStorage: LocalStorage<PreviewStoreClientStorageSchema> | undefined
+
+function clientStorage() {
+  _clientStorage ??= new LocalStorage<PreviewStoreClientStorageSchema>({projectName: 'shopify-cli-store'})
```
dmerand, line 18: Nit: Should we extract this into a shared constant?

packages/store/src/cli/services/store/create/preview/result.test.ts
```diff
@@ -0,0 +1,97 @@
...
+  status: 'success' as const,
+  message: 'Your preview store is ready.',
+  store: {
+    id: '123',
+    name: 'Lavender Candles',
+    subdomain: 'x12y45z.myshopify.com',
+    accessUrl: 'https://app.shopify.com/auth/preview-store?token=access-token',
+    requestedCountry: 'US',
+  },
+  nextSteps: [
```
dmerand: These next steps might drift from the non-JSON result. Any way to consolidate?

packages/store/src/cli/services/store/create/preview/client.ts
```diff
@@ -0,0 +1,211 @@
...
+  }
+
+  let parsed: RawPreviewStoreCreateResponse
+  try {
+    parsed = JSON.parse(rawText) as RawPreviewStoreCreateResponse
+  } catch (error) {
+    const message = error instanceof Error ? error.message : 'Unknown error'
+    throw new AbortError(
+      'Preview store creation returned a non-JSON response.',
+      `Parse error: ${message}. Body (truncated): ${rawText.slice(0, 500)}`,
```
dmerand: 🔒 **Security**: Consider reviewing the JSON parse failure path because it includes the raw response body in the user-visible AbortError diagnostic. The preview-store endpoint can return an Admin API token and tokenized access URL, and the current redaction helper only runs after JSON parsing succeeds. A malformed or truncated credential-bearing response could therefore expose secrets in CLI output or copied bug reports.

**Suggestion:** Redact raw response text before adding it to any `AbortError` try message, or avoid including the body for this endpoint. The redaction should cover at least `admin_api_token`, `access_url`, and token-shaped preview-store access URLs. It would also be worth applying the same protection to any other fallback diagnostics that include raw response text for this endpoint.

packages/store/src/cli/services/store/create/preview/index.ts
```diff
@@ -0,0 +1,109 @@
...
+      shopId: response.shop.id,
+      name: response.shop.name,
+      createdAt: acquiredAt,
+      ...(response.placeholderAccountUuid ? {placeholderAccountUuid: response.placeholderAccountUuid} : {}),
+      ...(country ? {country} : {}),
+      accessUrl: response.accessUrl,
+    },
+  })
+  dependencies.setLastSeenUserId(userId)
+  await dependencies.recordStoreFqdnMetadata(response.shop.domain, true)
```
dmerand: 🐛 **Bug**: Worth reviewing this await because it can turn a successful remote store creation into a failed command after credentials have already been persisted locally. If `recordStoreFqdnMetadata` rejects, the user may see a failure and never receive the access URL, even though the preview store exists and the Admin API token is cached. That conflicts with the intended best-effort nature of this metadata write and can encourage duplicate retries.

**Suggestion:** Treat `recordStoreFqdnMetadata` as non-blocking for command success. Catch failures after persisting the session and last-seen user, optionally debug-log them, and still return the success result so the user receives the access URL.

```suggestion
  try {
    await dependencies.recordStoreFqdnMetadata(response.shop.domain, true)
  } catch {
    // Store metadata is best-effort; credentials and access URL are already persisted.
  }
```

dmerand (commented):
Agent pal comments

packages/store/src/cli/commands/store/create/preview.ts
```diff
@@ -0,0 +1,56 @@
+import {type CreatePreviewStoreResult, createPreviewStoreCommand} from '../../../services/store/create/preview/index.js'
+import {writeCreatePreviewStoreResult} from '../../../services/store/create/preview/result.js'
+import StoreCommand from '../../../utilities/store-command.js'
+import {globalFlags, jsonFlag} from '@shopify/cli-kit/node/cli'
+import {outputContent} from '@shopify/cli-kit/node/output'
+import {renderSingleTask} from '@shopify/cli-kit/node/ui'
+import {Flags} from '@oclif/core'
+
+export default class StoreCreatePreview extends StoreCommand {
```
amcaplan, line 10: Shouldn't we ship this as hidden for now? It won't be usable until we flip on the flag.

packages/store/src/cli/commands/store/create/preview.ts
```diff
@@ -0,0 +1,56 @@
...
+  static summary = 'Create a preview Shopify store.'
+
+  static descriptionWithMarkdown = `Creates a new preview Shopify store for a merchant who wants to try Shopify without needing to immediately create an account.`
```
amcaplan: Something seems off about this text - the merchant is described as a 3rd party, but the CLI descriptions are usually more about what the command will do for you, the CLI user. I'd expect something like


```suggestion
  static descriptionWithMarkdown = `Creates a new Shopify store, with no need for an existing account.`
```

packages/store/src/cli/commands/store/create/preview.ts
```diff
@@ -0,0 +1,56 @@
...
+  static flags = {
+    ...globalFlags,
+    ...jsonFlag,
+    name: Flags.string({
+      description: 'The name of the store.',
+      env: 'SHOPIFY_FLAG_PREVIEW_STORE_NAME',
+      required: false,
+    }),
+    country: Flags.string({
+      description: 'Two-letter ISO 3166-1 alpha-2 country code for the store, such as US, CA, or GB.',
```
amcaplan: Do people generally know what `ISO 3166-1 alpha-2` is? I only heard of it now... Maybe we can leave out that part? I don't think there are other popular 2-char country code systems.

packages/store/src/cli/services/store/create/preview/client.ts
amcaplan: I don't think so, this is just a memoization strategy we use often throughout the CLI.

packages/store/src/cli/services/store/create/preview/client.ts
```diff
@@ -0,0 +1,211 @@
...
+  admin_api_token?: unknown
+  access_url?: unknown
+}
+
+interface RawPreviewStoreErrorResponse {
+  error_code?: string
+  message?: string
+}
+
+export function getOrCreateCliInstanceId(
```
amcaplan, line 62: This feels like something that belongs in CLI-kit - it should have utility beyond the current need. Especially for analytics.

packages/store/src/cli/services/store/create/preview/client.ts
```diff
@@ -0,0 +1,211 @@
...
+
+  return narrowResponse(parsed)
+}
+
+function previewStoreError(status: number, rawText: string): {message: string; tryMessage?: string} {
+  const parsed = parseErrorBody(rawText)
+  const errorCode = parsed.error_code
+  const message = parsed.message
+
+  if (errorCode === 'service_unavailable' || errorCode === 'not_in_rollout') {
```
amcaplan: Just checking, are these error codes part of a confirmed contract with the API?

Also, `service_unavailable` sounds more like the service _should_ be available but is currently down - like a 500.

packages/store/src/cli/services/store/create/preview/index.ts
```diff
@@ -0,0 +1,109 @@
+import {PreviewStoreClientOptions, PreviewStoreCreateResponse, createPreviewStore} from './client.js'
```
amcaplan, line 1: I don't think we generally do barrel files like this

packages/store/src/cli/services/store/create/preview/result.ts
```diff
@@ -0,0 +1,68 @@
+import {type CreatePreviewStoreResult} from './index.js'
```
amcaplan, line 1: I love everything about this file. So nice to see UI kit being used.

packages/store/src/cli/commands/store/create/preview.ts
```diff
@@ -0,0 +1,56 @@
...
+
+  static flags = {
+    ...globalFlags,
+    ...jsonFlag,
+    name: Flags.string({
+      description: 'The name of the store.',
+      env: 'SHOPIFY_FLAG_PREVIEW_STORE_NAME',
+      required: false,
+    }),
+    country: Flags.string({
```
amcaplan: Can this be moved to `flags.ts`? The command to create a dev store will require the same flag and validation logic, so it would be best if we can reuse.

Can also be a followup PR, just suggesting in case it's convenient to do here.

packages/store/src/cli/services/store/create/preview/client.ts
amcaplan: Still think it should probably be bumped to CLI-kit if it stays, but also we might delete this all when the feature gets shipped. So I guess it can stay here for now.

packages/store/src/cli/services/store/create/preview/index.ts
amcaplan: Spoke with @alfonso-noriega - we will need to update the entire package later to promote the barrel files, but we'll leave for now to avoid unnecessary churn in an area of active development.

amcaplan (approved):
Approving contingent on:

1. An answer to https://github.com/Shopify/cli/pull/7764/changes#r3409778219
2. Copy is updated to match the spec

packages/store/src/cli/services/store/create/preview/client.ts
alfonso-noriega: yes, they are all real error codes. But thanks for pointing this out, there are a few improvements to do like splitting this into 2 separate ones and  a missing rate_limit specific code

packages/store/src/cli/services/store/create/preview/result.ts
alfonso-noriega: all the credit to Pi 😅

PR #7766, opened 2026-06-09
Add `shopify doc fetch` command to download MD doc from shopify.dev

nelsonwittwer:
## What

Adds **`shopify doc fetch --url <url>`**, which downloads a complete document from shopify.dev and prints it to stdout (or writes it to a file with `--output`).

This is part of the new **`doc`** namespace for documentation tooling — the "full document" half. The "discovery" half is `doc search` (#7770).

## Why

Every page on shopify.dev has a Markdown representation. `doc fetch` requests that representation, giving agents (and scripts) a clean, verbatim way to pull instructional content from the centralized docs — for example, a set of instructions an agent follows like a centrally-served skill.

## Details

- `shopify doc fetch --url <url>` — requests the Markdown version of the page.
- `--output <path>` (`-o`) — write to a file instead of stdout.
- Only `shopify.dev` (and subdomains) URLs are allowed; malformed/disallowed URLs and non-ok responses throw before any output.

```
shopify doc fetch --url https://shopify.dev/docs/api/shopify-cli
shopify doc fetch --url https://shopify.dev/docs/api/shopify-cli --output docs/shopify-cli.md
```

## 🎩 Testing instructions

`doc fetch` hits a **public** endpoint on shopify.dev — no auth, no app, no login required. It just needs network access.

```bash
# 1. Check out this branch and build (dev.js runs the prebuilt bundle, so a build is required)
cd <your cli checkout>
git checkout add-fetch-doc-command && git pull
pnpm nx build cli
```

Run commands from the repo root via `node packages/cli/bin/dev.js …`:

**1. Fetch a real page → prints Markdown to stdout**
```bash
node packages/cli/bin/dev.js doc fetch --url https://shopify.dev/docs/api/shopify-cli | head -20
#   → Markdown content of the page
```

**2. `--output` writes to a file (creating missing parent dirs)**
```bash
node packages/cli/bin/dev.js doc fetch --url https://shopify.dev/docs/api/shopify-cli --output /tmp/doc-fetch/cli.md
head -5 /tmp/doc-fetch/cli.md
#   → "Saved …" message, and the file exists with the Markdown content
```

**3. Disallowed host → rejected before any network call**
```bash
node packages/cli/bin/dev.js doc fetch --url https://example.com/whatever
#   → "Only documents from the following hosts can be fetched: shopify.dev."
```

**4. Malformed URL → rejected**
```bash
node packages/cli/bin/dev.js doc fetch --url not-a-url
#   → "Invalid URL: not-a-url"
```

[...]

PR #7770, opened 2026-06-09
Add `doc search` command (JSON documentation search)

nelsonwittwer:
## What

Adds **`shopify doc search --query <query>`**, which queries the shopify.dev vector store and prints the most relevant documentation chunks as **JSON** to stdout.

This is the "discovery" half of the new **`doc`** namespace; the "full document" half is `doc fetch` (#7766). It no longer repurposes the top-level `search` command — that stays as the human, browser-based command (and is deprecated separately in #7778).

## Why

Gives agents (and scripts) a way to discover relevant documentation across shopify.dev as structured JSON, without a browser. Paired with `doc fetch`, an agent can find the right docs and pull them in full, entirely through the CLI.

## Details

- `shopify doc search --query "<query>"` — required query; prints the raw JSON response.
- `--api-name` (e.g. `admin`, `storefront`, `hydrogen`, `functions`; unknown values ignored).
- `--api-version` (e.g. `2025-10`, `latest`, `current`).
- Non-ok responses surface the server's `{error}` message (e.g. an invalid version lists the valid ones).

```
shopify doc search --query "subscribe to webhooks"
shopify doc search --query "create a product" --api-name admin --api-version latest
```

## 🎩 Testing instructions

`doc search` hits a **public** endpoint on shopify.dev (`/assistant/search`) — no auth, no app, no login required. It just needs network access and prints JSON to stdout.

```bash
# 1. Check out this branch and build (dev.js runs the prebuilt bundle, so a build is required)
cd <your cli checkout>
git checkout search-vector-store-json && git pull
pnpm nx build cli
```

Run commands from the repo root via `node packages/cli/bin/dev.js …`:

**1. Basic query → well-formed JSON**
```bash
node packages/cli/bin/dev.js doc search --query "subscribe to webhooks" | jq 'length'
#   → a number (count of chunks), with no jq parse error
```

**2. API/version filters are forwarded**
```bash
node packages/cli/bin/dev.js doc search --query "create a product" --api-name admin --api-version latest | jq '.[0]'
#   → the first result object
```

**3. Invalid version (with an api-name) surfaces the helpful error**
```bash
node packages/cli/bin/dev.js doc search --query "create a product" --api-name admin --api-version 1999-01
#   → "Search failed: Invalid api_version '1999-01' for api_name 'admin'. Available versions: …"  (exit non-zero)
```
Note: the endpoint only validates the version in the context of an `--api-name`. A bad `--api-version` on its own is silently ignored and still returns results.

**4. Required arg + help**
```bash
node packages/cli/bin/dev.js doc search          # → "Missing required flag query"
node packages/cli/bin/dev.js doc search --help   # → usage, description, examples, --api-name / --api-version flags
```

### Inspecting the Monorail analytics payload

This branch is rebased on the app-context telemetry PR (#7771), so every command emits a command-analytics event. To see exactly what *would* be sent to Monorail — without sending it — set two env vars:

- `SHOPIFY_CLI_NO_ANALYTICS=1` — dump the payload via debug logging instead of POSTing it.
- `SHOPIFY_FLAG_VERBOSE=1` — raise the log level so the debug line prints.

```bash
SHOPIFY_CLI_NO_ANALYTICS=1 SHOPIFY_FLAG_VERBOSE=1 \
  node packages/cli/bin/dev.js doc search --query "webhooks" 2>&1 \
  | grep "Skipping command analytics"
#   → a "Skipping command analytics, payload: {…}" line with "command": "doc search"
```

To confirm the inherited app-context hook, run the same command **from inside an app directory while logged in** — the payload then also carries app-level identifiers (`api_key`, `project_type`, `app_*`). Run it from a non-app directory and those fields are absent.

Screenshot 2026-06-22 at 2 35 22 PM

I saw the APP metrics broadcast via monorail when executing in the path of an app.

When *not* in the path of an app (in this case in the path of the CLI that I pulled down) - I did not broadcast app context.
Screenshot 2026-06-22 at 2 36 53 PM

## Related

- #7766 — `doc fetch` (full-document retrieval).
- #7778 — deprecates the old browser `search`, pointing users to `doc search`.


docs-shopify.dev/commands/interfaces/search.interface.ts
```diff
@@ -4,5 +4,27 @@
  * @publicDocs
  */
 export interface search {
+  /**
+   * Limit results to a specific API (for example: admin, storefront, hydrogen, functions). Unrecognized values are ignored.
```
dmerand: ❤️

packages/cli/README.md
```diff
@@ -2083,22 +2083,33 @@ DESCRIPTION
...
+  --api-version=<value>  [env: SHOPIFY_FLAG_API_VERSION] Limit results to a specific API version (for example: 2025-10,
+                         latest, current).
+  --no-color             [env: SHOPIFY_FLAG_NO_COLOR] Disable color output.
+  --verbose              [env: SHOPIFY_FLAG_VERBOSE] Increase the verbosity of the output.
+
 DESCRIPTION
-  Starts a search on shopify.dev.
+  Query the shopify.dev vector store and print the most relevant documentation chunks as JSON. Best for programmatic
+  discovery — surfacing the relevant pieces of documentation for a topic, rather than retrieving a whole document. To
+  download a full document verbatim, use `fetch-doc`.
```
dmerand: Should discuss this contract with @nickwesselman . I'd assume that `shopify search` could just have a `--fetch-doc` param rather than a whole separate command for that case.

nickwesselman:
Having second thoughts about using `shopify search` for this.

I'm sure use of `shopify search` is near-zero, but should we stick to our SemVer committment, and create a new command? We could fix the existing command or just deprecate it.

This would also give us an opportunity to apply our CLI design guidelines to a new command. e.g. The `shopify search` use of positional arguments is against those guidelines.

packages/cli/src/cli/commands/doc/search.ts
```diff
@@ -0,0 +1,46 @@
+import {docSearchService} from '../../services/commands/doc/search.js'
+import Command from '@shopify/cli-kit/node/base-command'
+import {globalFlags} from '@shopify/cli-kit/node/cli'
+import {Args, Flags} from '@oclif/core'
+
+export default class DocSearch extends Command {
+  static description =
+    'Query the shopify.dev vector store and print the most relevant documentation chunks as JSON. Best for programmatic discovery — surfacing the relevant pieces of documentation for a topic, rather than retrieving a whole document. To download a full document verbatim, use `doc fetch`.'
+
+  static usage = `doc search [query]`
```
durga256: nit: doc search <query> since query is required

packages/cli/src/cli/services/commands/doc/search.ts
```diff
@@ -0,0 +1,39 @@
...
+// originating from the CLI can be attributed as such.
+const SURFACE_HEADER = 'X-Shopify-Surface'
+const SURFACE = 'cli'
+
+export async function docSearchService(query: string, apiName?: string, apiVersion?: string) {
+  const params = new URLSearchParams({query})
+  if (apiName) params.append('api_name', apiName)
+  if (apiVersion) params.append('api_version', apiVersion)
+
+  const response = await fetch(`${SEARCH_URL}?${params.toString()}`, {
```
durga256: use shopifyFetch?

durga256 (approved):

packages/cli/src/cli/services/commands/doc/search.ts
durga256: also need to handle network errors gracefully

kaylarobinson077:
Looking through the existing cli monorail schema, nothing stands out to me as a pre-existing fit for capturing doc search results. My suggestion would be that we add a new monorail field `cmd_doc_search_result` where we put the json result, then when we extend to validate and other commands, add in related fields like `cmd_validate_passes`.

Additionally, currently the `args` monorail field is flagged as PII, so anything we are passing in there (e.g. the search query) is redacted after 30 days. The pattern followed here for cases where we want to retain the flags for longer duration is to additionally log them in a separate field, so we might consider also adding a field like `cmd_doc_search_query` to explicitly log the search query and not flag it as PII to retain. Not directly related to this PR, but this also is impacting any vars we are saving through the agent env vars, as they're subject to the same PII policy.

So, if this sounds reasonable, I can update the monorail schema to include `cmd_doc_search_query` and `cmd_doc_search_result`. For doc fetch, maybe I'll also add in `cmd_doc_fetch_query` (but skip result there, since it's just the docs content for the requested link?)

nelsonwittwer:
@kaylarobinson077 - I think the cleaner solution for tracking search would be on the search handler. That one hanbdler is where all search traffic goes already

kaylarobinson077:
> @kaylarobinson077 - I think the cleaner solution for tracking search would be on the search handler. That one hanbdler is where all search traffic goes already

​Is the search handler something that lives within this command, or is it something outside of CLI? My home is that we can track the query and result in the regular CLI monorail event, so everything lives together in one place in terms of where the data lands.

isaacroldan (approved):

PR #7771, opened 2026-06-09
Capture app context in analytics for every command run inside an app

nelsonwittwer:
## What

Makes command analytics capture app context for **every** CLI command run inside an app — not just commands that explicitly load an app.

## Why

App-level telemetry is already flowing, but only as a side effect of commands that authenticate and load an app.

Commands that don't load an app (`search`, `fetch-doc`, `version`, `upgrade`, …) emit nothing, so we can't tell what app context they were run in. We need this signal everywhere so we can judge the effectiveness of commands, especially when run by agents in different harnesses/models.

## How

The `public_command_metadata` hook runs on the postrun of every command. It now calls a best-effort helper, `logAppContextMetadataIfAuthenticated(cwd())`, which — **only** when the user is already authenticated and is inside an app project — loads the app from the current directory and broadcasts the same app-context block a `shopify app *` command would (`api_key`, `project_type`, the `app_*` fields, `cmd_app_linked_config_*`, and the sensitive `app_name`).

We do **not** deduce an app for commands run outside an app directory. The behavior is simply: if the command is run within the path of an app, it broadcasts that app's context; otherwise it broadcasts nothing app-related.

It is strictly best-effort and safe to run everywhere:
- **Gated on an existing local session** (`sessionExists()`) — never triggers a login, never makes a network request. Anonymous/CI usage stays unenriched.
- **Short-circuits when `api_key` is already set** — app commands already populated it, so there's no redundant work.
- **Never prompts** — `skipPrompts` is threaded through `localAppContext` → `getAppConfigurationContext`.
- **Bounded by a 3s timeout** — it runs before process exit, so it can't hang the CLI.
- **Swallows all errors** — not an app dir, invalid config, etc. → silently does nothing.

## 🎩 testing instructions

```bash
# 1. Build the branch (run from your CLI repo root)
pnpm nx build app

# 2. Run a non-app command INSIDE an app directory.
#    Replace the cd path with one of your own apps; replace the dev.js path with your checkout.
cd ~/code/luxury-duplication-app
SHOPIFY_CLI_NO_ANALYTICS=1 SHOPIFY_FLAG_VERBOSE=1 node /Users/nelson/src/github.com/Shopify/cli/packages/cli/bin/dev.js version 2>&1 | grep -E '"command"|"api_key"|"project_type"|"app_scopes"|"cmd_app_linked_config_name"'
#   → api_key, project_type, app_scopes, cmd_app_linked_config_name, … all present
#     (the full app-context block, same as a `shopify app *` command)

# 3. Run the SAME command from a directory that is NOT an app (/tmp works on any machine):
cd /tmp
SHOPIFY_CLI_NO_ANALYTICS=1 SHOPIFY_FLAG_VERBOSE=1 node /Users/nelson/src/github.com/Shopify/cli/packages/cli/bin/dev.js version 2>&1 | grep -E '"command"|"api_key"|"project_type"|"app_scopes"'
#   → only "command" — no api_key / project_type / app_* (no app context is broadcast)
```


packages/app/src/cli/hooks/public_metadata.ts
```diff
@@ -1,7 +1,13 @@
 import metadata from '../metadata.js'
+import {logAppContextMetadataIfAuthenticated} from '../services/app-context.js'
 import {FanoutHookFunction} from '@shopify/cli-kit/node/plugins'
+import {cwd} from '@shopify/cli-kit/node/path'
 
 const gatherPublicMetadata: FanoutHookFunction<'public_command_metadata', '@shopify/app'> = async () => {
+  // Runs for every CLI command. Best-effort: if the user is logged in and is inside an
```
nelsonwittwer: remove comment

packages/app/src/cli/services/app-context.ts
```diff
@@ -61,10 +63,12 @@ interface LoadedAppContextOptions {
  *
  * @param directory - The directory containing the app.
  * @param userProvidedConfigName - The name of an existing config file in the app, if not provided, the cached/default one will be used.
+ * @param skipPrompts - When true, never prompts the user (e.g. to re-select a config). Required for non-interactive callers such as telemetry.
  */
 interface LocalAppContextOptions {
   directory: string
   userProvidedConfigName?: string
+  skipPrompts?: boolean
```
nelsonwittwer, line 69: I don't see why this is needed?

packages/app/src/cli/services/app-context.ts
```diff
@@ -204,3 +209,56 @@ export async function localAppContext({
...
+ *     the app — no need to redo the work),
+ *   - never prompts, never makes a network request,
+ *   - is bounded by a short timeout so it can't delay command exit, and
+ *   - swallows all errors (a directory that isn't an app, an invalid config, etc.).
+ *
+ * @param directory - The working directory to inspect for an app.
+ */
+export async function logAppContextMetadataIfAuthenticated(directory: string): Promise<void> {
+  try {
+    // A command that loads the app (e.g. `app dev`) already populated identity.
```
nelsonwittwer: redundant comment

packages/app/src/cli/services/app-context.ts
```diff
@@ -204,3 +209,56 @@ export async function localAppContext({
...
+    if (metadata.getAllPublicMetadata().api_key !== undefined) return
+    // Only enrich for users who are already authenticated; never trigger a login.
```
nelsonwittwer: redundant comment

packages/app/src/cli/services/app-context.ts
nelsonwittwer: `localAppContext` calls `getAppConfigurationContext` → `selectActiveConfig`, which prompts the user (via `app config use`) to pick a new config when a previously-cached config file is missing from disk. This helper runs in the postrun `public_command_metadata` hook of *every* command, so an interactive prompt there would be a bad surprise. `skipPrompts: true` guarantees the enrichment stays fully non-interactive — if a prompt would be needed, it just bails out instead. It's load-bearing, not decorative.

packages/app/src/cli/services/app-context.ts
```diff
@@ -204,3 +208,56 @@ export async function localAppContext({
...
+      // block. It does not emit api_key (that comes from the remote app), so add it here.
+      const {app} = await localAppContext({directory, skipPrompts: true})
+      const clientId = app.configuration.client_id
+      if (typeof clientId === 'string' && clientId.length > 0) {
+        await metadata.addPublicMetadata(() => ({api_key: clientId}))
+      }
+    })()
+
+    try {
+      await Promise.race([load, deadline])
```
durga256: are we cancelling/stopping/cleaning up load if deadline wins? It could reject later?

packages/app/src/cli/services/app-context.ts
```diff
@@ -204,3 +208,56 @@ export async function localAppContext({
...
+
+    let timer: ReturnType<typeof setTimeout> | undefined
+    const deadline = new Promise<void>((resolve) => {
+      timer = setTimeout(resolve, APP_CONTEXT_METADATA_TIMEOUT_MS)
+    })
+
+    const load = (async () => {
+      // localAppContext drives the app loader, which populates project_type and the app_*
+      // block. It does not emit api_key (that comes from the remote app), so add it here.
+      const {app} = await localAppContext({directory, skipPrompts: true})
```
durga256: this calls loadAppFromContext which can potentially now mutate project files. mitigate or flag risk?

.changeset/app-context-telemetry-everywhere.md
isaacroldan, line 1: Why do we need the user to be authenticated? we can find the api-key in the toml regardless of that.

Also, changeset descriptions should be short, i'll update that

isaacroldan (approved):

PR #7782, opened 2026-06-11
Improve JSON schema error messages

Barefoot0:

Fixes #[22951](https://github.com/shop/issues-develop/issues/22951#issue-4626249810)

TOML app module validation could report confusing JSON Schema errors such as `Expected object, received object` when a TOML array of tables was used where an object was expected.

- Updates JSON Schema validation error formatting to report arrays as `array`instead of relying on JavaScript `typeof` output.
- Adds a TOML-specific hint for app configuration errors when an object was expected but an array was received.
- Adds regression tests for array/null received-type formatting and the TOML validation hint.

### How to test your changes?

- `pnpm --filter @shopify/cli-kit vitest src/public/node/json-schema.test.ts`
- `pnpm --filter @shopify/app vitest src/cli/services/validate.test.ts`
- Can run Shopify CLI locally with a local app module and verify validation. I followed [this document ](https://docs.google.com/document/d/1_fs9oHyxt4xqKnLk3WpMVIAU2gSaPUBHIVsFzcUMSlk/edit?tab=t.0)to set up a local test app and used this command in your cli repo: `SHOPIFY_CLI_NEVER_USE_PARTNERS_API=1 SHOPIFY_SERVICE_ENV=local pnpm shopify app dev --path [path to test app]`

### Post-release steps

None.

  
The change is user-facing. I've identified the correct [bump type](https://github.com/Shopify/cli/blob/main/CONTRIBUTING.md#choosing-the-right-bump-type) (`patch` for bug fixes · `minor` for new features · `major` for [breaking changes](https://github.com/Shopify/cli/blob/main/CONTRIBUTING.md#what-counts-as-a-breaking-change)) and added a changeset with `pnpm changeset add`

### DIFF
--- a/.changeset/mighty-hotels-tickle.md
+++ b/.changeset/mighty-hotels-tickle.md
@@ -1,5 +1,6 @@
 ---
 '@shopify/cli-kit': patch
+'@shopify/app': patch
 ---
 
-Improve JSON Schema validation error messages for arrays.
+Improve JSON Schema validation error messages for arrays and add a TOML table hint for object/array mismatches.
--- a/.changeset/mighty-hotels-tickle.md
+++ b/.changeset/mighty-hotels-tickle.md
@@ -1,0 +1,5 @@
+---
+'@shopify/cli-kit': patch
+---
+
+Improve JSON Schema validation error messages for arrays.
--- a/packages/app/src/cli/models/app/loader.ts
+++ b/packages/app/src/cli/models/app/loader.ts
@@ -73,11 +73,21 @@ export interface ConfigurationError {
   code?: string
 }
 
+function tomlObjectArrayHint(error: ConfigurationError): string | undefined {
+  if (!error.file.endsWith('.toml')) return undefined
+  if (error.message !== 'Expected object, received array') return undefined
+
+  return 'Use a single TOML table instead of an array of tables. [table] defines one table; [[table]] defines an array of tables.'
+}
+
 export function formatConfigurationError(error: ConfigurationError): string {
+  const hint = tomlObjectArrayHint(error)
+  const message = hint ? `${error.message}. ${hint}` : error.message
+
   if (error.path?.length) {
-    return `[${error.path.join('.')}]: ${error.message}`
+    return `[${error.path.join('.')}]: ${message}`
   }
-  return error.message
+  return message
 }
 
 type ConfigurationResult<T> = {data: T; errors?: never} | {data?: never; errors: ConfigurationError[]}
--- a/packages/app/src/cli/models/app/loader.ts
+++ b/packages/app/src/cli/models/app/loader.ts
@@ -77,7 +77,7 @@ function tomlObjectArrayHint(error: ConfigurationError): string | undefined {
   if (!error.file.endsWith('.toml')) return undefined
   if (error.message !== 'Expected object, received array') return undefined
 
-  return 'Use a single TOML table instead of an array of tables. [table] defines one table; [[table]] defines an array of tables.'
+  return 'Use a TOML table instead of an array. [table] defines a single table; [[table]] defines an array of tables.'
 }
 
 export function formatConfigurationError(error: ConfigurationError): string {
--- a/packages/app/src/cli/services/validate.test.ts
+++ b/packages/app/src/cli/services/validate.test.ts
@@ -6,6 +6,7 @@ import {describe, expect, test, vi} from 'vitest'
 import {outputResult} from '@shopify/cli-kit/node/output'
 import {renderError, renderSuccess} from '@shopify/cli-kit/node/ui'
 import {AbortSilentError} from '@shopify/cli-kit/node/error'
+import {jsonSchemaValidate} from '@shopify/cli-kit/node/json-schema'
 
 vi.mock('../metadata.js', () => ({default: {addPublicMetadata: vi.fn()}}))
 vi.mock('@shopify/cli-kit/node/output', async (importOriginal) => {
@@ -46,7 +47,45 @@ describe('formatConfigurationError', () => {
         message: 'Expected object, received array',
       }),
     ).toBe(
-      '[events.1.metrics]: Expected object, received array. Use a single TOML table instead of an array of tables. [table] defines one table; [[table]] defines an array of tables.',
+      '[events.1.metrics]: Expected object, received array. Use a TOML table instead of an array. [table] defines a single table; [[table]] defines an array of tables.',
+    )
+  })
+
+  test('adds a TOML table hint to JSON Schema object array type mismatches', () => {
+    const schemaParsed = jsonSchemaValidate(
+      {events: [{metrics: []}]},
+      {
+        type: 'object',
+        properties: {
+          events: {
+            type: 'array',
+            items: {
+              type: 'object',
+              properties: {
+                metrics: {type: 'object'},
+              },
+            },
+          },
+        },
+      },
+      'strip',
+    )
+
+    expect(schemaParsed.state).toBe('error')
+    expect(schemaParsed.errors).toEqual([
+      {
+        path: ['events', '0', 'metrics'],
+        message: 'Expected object, received array',
+      },
+    ])
+
+    expect(
+      formatConfigurationError({
+        file: 'shopify.app.toml',
+        ...schemaParsed.errors![0],
+      }),
+    ).toBe(
+      '[events.0.metrics]: Expected object, received array. Use a TOML table instead of an array. [table] defines a single table; [[table]] defines an array of tables.',
     )
   })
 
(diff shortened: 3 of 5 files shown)


Barefoot0:
* **#7782** <a href="https://app.graphite.com/github/pr/Shopify/cli/7782?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Shopify/cli/7782?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

.changeset/mighty-hotels-tickle.md
```diff
@@ -0,0 +1,5 @@
+---
+'@shopify/cli-kit': patch
```
isaacroldan, line 2: 🐛 **Bug**: The changeset only patches `@shopify/cli-kit`, but the new `tomlObjectArrayHint` in `packages/app/src/cli/models/app/loader.ts` produces user-visible TOML validation output that ships from `@shopify/app`. Even though all packages bump together (fixed group per `.changeset/config.json`), the changeset body lands in per-package release notes — so anyone reading `@shopify/app`'s changelog will see no mention of the new hint. Sibling changesets (e.g. `fix-app-env-pull-absolute-path.md`, `app-dev-port-validation.md`) follow the convention of listing the package whose user-facing behaviour changed, and the PR description itself acknowledges this is a user-facing change.

**Suggestion:** 
```suggestion
'@shopify/cli-kit': patch
'@shopify/app': patch
```

packages/app/src/cli/models/app/loader.ts
isaacroldan, line 1: 📐 **Design**: `formatConfigurationError` is the common formatter for all configuration errors (web, app, extension) and has 5 call sites across `loader.ts`, `app-context.ts`, and `validate.ts`. Embedding a TOML-format-specific hint here folds a format concern into what was previously a presentation-only function. With one hint this is fine, but if more accumulate (a JSON-file hint, an `Expected array, received object` hint going the other way), this function risks becoming a hint registry.

**Suggestion:** If more hints are likely (the linked issue suggests this is one of several confusing messages), consider extracting `tomlObjectArrayHint` and any siblings into a small `loader-hints.ts` module exporting a single `getFormatHint(error): string | undefined`, and call that from `formatConfigurationError`. If this is anticipated to be the only such hint, leaving it inline is reasonable.

packages/app/src/cli/models/app/loader.ts
```diff
@@ -73,11 +73,21 @@ export interface ConfigurationError {
   code?: string
 }
 
+function tomlObjectArrayHint(error: ConfigurationError): string | undefined {
+  if (!error.file.endsWith('.toml')) return undefined
+  if (error.message !== 'Expected object, received array') return undefined
```
isaacroldan, line 78: 📐 **Design**: `tomlObjectArrayHint` fires only when `error.message === 'Expected object, received array'`. That exact literal is built in a separate package (`packages/cli-kit/src/public/node/json-schema.ts:148`) via a template literal, with no shared constant or type-level link between the two sites. Any future tweak in cli-kit — an added period, a capitalisation change, expanded union rendering, extra context — would silently disable this hint. The existing tests can't catch the drift either: `validate.test.ts` hand-constructs the input message, and `json-schema.test.ts` never exercises `formatConfigurationError`, so the cross-package contract is currently unenforced.

**Suggestion:** Consider one of: (1) export a constant (e.g. `EXPECTED_OBJECT_RECEIVED_ARRAY_MESSAGE`) from `json-schema.ts` and import it here so a rename is a compile-time break; (2) populate `ConfigurationError.code` from the JSON-schema path (the zod path already does this on line 73) and key the hint off `code === 'invalid_type'` plus expected/received metadata; or (3) keep the string match but add an integration test that drives `jsonSchemaValidate` → `formatConfigurationError` end-to-end so future drift fails loudly.

packages/app/src/cli/models/app/loader.ts
```diff
@@ -73,11 +73,21 @@ export interface ConfigurationError {
...
+
+  return 'Use a single TOML table instead of an array of tables. [table] defines one table; [[table]] defines an array of tables.'
```
isaacroldan: 💬 **Suggestion**: The hint text specifically explains TOML's `[table]` vs `[[table]]` syntax, but `Expected object, received array` also fires when the user writes an inline array like `metrics = []` or `metrics = ["foo"]` where the schema expects an inline table. In that scenario the suggested fix (switch from `[[table]]` to `[table]`) doesn't map to what the user actually did. Issue #22951 is about the `[[table]]`-vs-`[table]` confusion, so the linked report is well served, but other users hitting the same message may be steered wrong.

**Suggestion:** Either broaden the wording (e.g. `Use a TOML table (e.g., [section] or { key = value }) instead of an array. [[name]] defines an array of tables; [name] defines a single table.`) or scope the hint to the array-of-tables case only by also checking that the path's last segment isn't a numeric index.

packages/app/src/cli/services/validate.test.ts
```diff
@@ -37,6 +37,28 @@ describe('formatConfigurationError', () => {
...
+
+  test('does not add a TOML table hint for non-TOML files', () => {
+    expect(
+      formatConfigurationError({
+        file: 'config.json',
+        path: ['events', '1', 'metrics'],
+        message: 'Expected object, received array',
+      }),
+    ).toBe('[events.1.metrics]: Expected object, received array')
+  })
```
isaacroldan, line 104: 💡 **Improvement**: The new cases cover (a) hint fires for the exact message on a `.toml` file and (b) hint does not fire on a non-TOML file. There's no case for a `.toml` file with a *different* type mismatch (e.g. `Expected object, received string`). Given the implementation gate is strict equality on the message, a refactor that loosens the check (e.g. to `message.startsWith('Expected object, received')`) wouldn't be caught — both for the message-shape axis and for the accidental-broadening risk on inline-array vs `[[table]]` cases.

**Suggestion:** Add a third case asserting no hint is appended for, e.g., `{file: 'shopify.app.toml', message: 'Expected object, received string'}` so the message gate is pinned alongside the existing file-extension gate.

packages/cli-kit/src/public/node/json-schema.ts
isaacroldan, line 1: 💡 **Improvement**: Three call sites in `convertJsonSchemaErrors` were migrated from `getPathValue(subject, path.join('.'))` to the new `getJsonSchemaErrorValue(subject, path)` helper, but `simplifyUnionErrors` (around line 256, and the subsequent call on `subjectValue` near 271) still uses the old style. The result is that top-level unions still suffer the empty-path-returns-undefined issue this PR fixes elsewhere, and the file now carries two competing conventions for the same operation.

**Suggestion:** Consider migrating these call sites to use `getJsonSchemaErrorValue` as well so the file converges on one convention. If left out of scope intentionally, a short comment noting why would help future readers avoid the same drift.

isaacroldan (commented):
---

.changeset/mighty-hotels-tickle.md
Barefoot0: Roger that, I added it here.

packages/app/src/cli/models/app/loader.ts
Barefoot0: I kept this inline for now because this PR is only adding one narrow TOML hint for this specific object/array mismatch, and I wanted to avoid introducing a broader hint registry before we have more cases.

If we add more format-specific hints later, I will change it to a `loader-hints.ts`
module!

packages/app/src/cli/models/app/loader.ts
Barefoot0: I broadened the wording so it no longer assumes the user specifically used an array of
tables.

The hint now says to use a TOML table instead of an array, while still explaining the `[table]` vs
`[[table]]` distinction.

packages/app/src/cli/models/app/loader.ts
Barefoot0: I kept the string match for now to avoid making this JSON Schema message a shared API or introducing new structured error metadata in this PR, but added an integration-style test for the current contract.

The new test drives `jsonSchemaValidate` and then passes the generated error into `formatConfigurationError`, so if the cli-kit message changes in a way that stops the TOML hint from firing, the app test should fail loudly.

packages/app/src/cli/services/validate.test.ts
Barefoot0: I added a negative `.toml` case for a different type mismatch (`Expected object, received string`) to pin the message gate.

That should catch accidental broadening of the hint logic while keeping the current behaviour narrowly scoped to the object/array mismatch.

packages/cli-kit/src/public/node/json-schema.ts
Barefoot0: Nice catch! I migrated the remaining relevant lookups in `simplifyUnionErrors` to use `getJsonSchemaErrorValue` as well.

isaacroldan (approved):

PR #7783, opened 2026-06-11
Use store auth sessions for theme commands

alfonso-noriega:

Fixes https://github.com/shop/issues-develop/issues/22916.

Preview stores created by `shopify store create preview` persist an Admin API token in the store-auth cache. `theme pull` and `theme push` should be able to use that cached token when the user does not provide a Theme Access/Admin token via `--password`.

- Moves the store-auth session storage schema, sanitization, keying, listing, reading, writing, and clearing helpers into `@shopify/cli-kit/node/store-auth-session`.
- Deletes `packages/store`'s old `auth/session-store` module; store and theme now use the cli-kit boundary directly.
- Updates store auth, store info, store execute, and preview store creation code/tests to import store-auth session helpers from cli-kit.
- Removes the direct `@shopify/theme` → `@shopify/store` package dependency.
- Updates theme command session creation so explicit `--password` still wins, but missing `--password` can fall back to a matching cached store-auth Admin API token before normal theme authentication.
- Allows multi-environment theme configs to omit `password` when the matching store-auth cache session satisfies auth.
- Adds/expands focused tests for the shared cli-kit storage boundary, explicit password precedence, store-auth fallback, missing-cache fallback, and multi-environment validation.

### How to test your changes?

```bash
pnpm vitest run packages/cli-kit/src/public/node/store-auth-session.test.ts packages/store/src/cli/services/store/auth/session-store.test.ts packages/store/src/cli/services/store/auth/session-lifecycle.test.ts packages/store/src/cli/services/store/auth/stored-auth.test.ts packages/store/src/cli/services/store/auth/index.test.ts packages/store/src/cli/services/store/auth/existing-scopes.test.ts packages/store/src/cli/services/store/info/index.test.ts packages/store/src/cli/services/store/execute/admin-transport.test.ts packages/store/src/cli/services/store/create/preview/index.test.ts packages/theme/src/cli/utilities/theme-command.test.ts
pnpm --filter @shopify/cli-kit lint
pnpm --filter @shopify/store lint
pnpm --filter @shopify/theme lint
pnpm --filter @shopify/cli-kit build
pnpm --filter @shopify/store build
pnpm --filter @shopify/theme build
```

### Post-release steps

None.

### Checklist

- [x] The change is user-facing — I've identified the correct bump type and added a changeset


alfonso-noriega:
* **#7783** <a href="https://app.graphite.com/github/pr/Shopify/cli/7783?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Shopify/cli/7783?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#7788** <a href="https://app.graphite.com/github/pr/Shopify/cli/7788?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#7831](https://github.com/Shopify/cli/pull/7831) <a href="https://app.graphite.com/github/pr/Shopify/cli/7831?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#7764** <a href="https://app.graphite.com/github/pr/Shopify/cli/7764?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

packages/theme/src/cli/utilities/theme-command.ts
```diff
@@ -323,14 +334,34 @@ export default abstract class ThemeCommand extends Command {
...
+  private async storeAuthSessionForTheme(flags: FlagValues): Promise<AdminSession | undefined> {
+    const store = typeof flags.store === 'string' ? flags.store : undefined
+    const password = flags.password
+    if (!store || password) return undefined
+
+    try {
+      const {adminSession} = await loadAdminSessionFromStoreAuth(store)
+      return adminSession
+    } catch (error) {
+      if (error instanceof AbortError && error.message.startsWith('No stored app authentication found for ')) {
```
dmerand: 🐛 **Bug**: The store-auth lookup appears to be an opportunistic fallback before normal theme authentication, but it only treats the exact missing-cache message as optional. Other recoverable store-auth cache states, such as expired sessions without refresh tokens or failed refreshes, can abort `theme pull` or `theme push` before `ensureAuthenticatedThemes` has a chance to succeed. That changes behavior for users with stale store-auth cache entries even when normal theme auth remains valid.

**Suggestion:** Treat recoverable store-auth `AbortError`s as an unavailable cache in this optional fallback path, while still rethrowing unexpected or programming errors. One option is to add a store-auth helper that returns `undefined` for unusable cached sessions and preserves hard failures.

dmerand (commented):
'bot found one concern worth looking into.

packages/theme/package.json
```diff
@@ -42,6 +42,7 @@
   "dependencies": {
     "@oclif/core": "4.8.3",
     "@shopify/cli-kit": "4.2.0",
+    "@shopify/store": "4.2.0",
```
amcaplan: Generally speaking, we should try and avoid command packages pulling from each other. I am thinking of what would be a better way to organize this code, but I don't have a good answer yet. Something to think about down the road.

packages/store/src/cli/services/store/auth/admin-session.ts
```diff
@@ -0,0 +1,46 @@
...
+  try {
+    const {adminSession} = await loadAdminSessionFromStoreAuth(store)
+    return adminSession
+  } catch (error) {
+    if (isRecoverableStoreAuthCacheError(error)) return undefined
+    throw error
+  }
+}
+
+function isRecoverableStoreAuthCacheError(error: unknown): boolean {
```
amcaplan: This is an incredibly brittle function. Ideally we'd go and update these various errors to use particular types, have a specific attribute (`isRetryable` perhaps), something like that. What's the blast radius if these strings fall out of date?

amcaplan (commented):
I think we should limit this explicitly to preview stores. We know those tokens have theme permissions. But we don't know that for regular store auth tokens, which means we're likely to hit all kinds of weird errors.

packages/theme/package.json
alfonso-noriega: Agreed. I don't love adding a command-package dependency here either, but keeping the store-auth cache/session interpretation in the package that owns that lifecycle avoids duplicating it in theme. I'd leave this as-is for this stack and revisit extracting shared auth/session utilities if more command packages need this pattern.

.changeset/theme-preview-store-auth.md
```diff
@@ -0,0 +1,6 @@
+---
+'@shopify/theme': patch
+'@shopify/store': patch
+---
+
+Use matching store-auth cache sessions for theme pull and push when no theme password is provided.
```
isaacroldan: Is this a user-facing change? if it is, it probably needs a better description.

packages/store/src/cli/services/store/auth/admin-session.ts
```diff
@@ -0,0 +1,38 @@
...
+  return error instanceof AbortError
+}
```
isaacroldan: how being `instanceof` AbortError makes it recoverable? :thinking_face:

packages/theme/package.json
isaacroldan: This feels like something that could be in cli-kit? Is just session storage, we have generic session storage in cli-kit. In theory as long as both packages use the same key, they both should be able to access the same session? Or is this not correct?

alfonso-noriega:
> I think we should limit this explicitly to preview stores. We know those tokens have theme permissions. But we don't know that for regular store auth tokens, which means we're likely to hit all kinds of weird errors.

​That is a good product question. In my opinion we shouldn't try to have specific preview store behavior if we can avoid it. The same problem is already live if you run store execute with a query that you do not have scopes for.

packages/store/src/cli/services/store/auth/admin-session.ts
```diff
@@ -0,0 +1,38 @@
...
+  return {
+    adminSession: {
+      token: session.accessToken,
+      storeFqdn: session.store,
+    },
+    session,
+  }
+}
+
+export async function tryLoadAdminSessionFromStoreAuth(store: string): Promise<AdminSession | undefined> {
```
isaacroldan: Curious why we want to expose two functions and one of them being just a try/catch.  
We could export only `loadAdminSessionFromStoreAuth` and let the consumers add the try/catch? Or the other way: just export the try/catch version?

packages/theme/package.json
alfonso-noriega: Moved the shared key/client-id and optional cached Admin session lookup into `cli-kit`, and updated theme to import from `@shopify/cli-kit/node/store-auth-session` instead of depending on `@shopify/store`. Store auth config now re-exports the same cli-kit key helpers, so both packages use the same storage key. The store package still keeps its full lifecycle/refresh helper for store-owned commands; the cli-kit helper is intentionally a read-only optional cache lookup for theme fallback.

amcaplan:
> ​That is a good product question. In my opinion we shouldn't try to have specific preview store behavior if we can avoid it. The same problem is already live if you run store execute with a query that you do not have scopes for.

@alfonso-noriega I don't think this is the same thing. `store execute` will fail with an obvious error that triggers you to increase the permissions on the token. But for theme commands, for non-preview stores, usually the right thing to do would be to log in via organizations. Of course they can also request the `write_themes` permission, but that's not documented in these commands at the moment.

@nickwesselman what do you think?

nickwesselman:
> > ​That is a good product question. In my opinion we shouldn't try to have specific preview store behavior if we can avoid it. The same problem is already live if you run store execute with a query that you do not have scopes for.
>
> @alfonso-noriega I don't think this is the same thing. `store execute` will fail with an obvious error that triggers you to increase the permissions on the token. 

​What is the error message and why wouldn't it trigger the agent to `store auth` with the additional required scopes?

> But for theme commands, for non-preview stores, usually the right thing to do would be to log in via organizations. Of course they can also request the `write_themes` permission, but that's not documented in these commands at the moment.

The goal is to have a solution here that works for `store auth` generally, not just for preview stores. We don't want this functionality to stop working after they claim/save the store, right?

packages/cli-kit/src/public/node/store-auth-session.test.ts
```diff
@@ -56,15 +135,6 @@ describe('store session storage', () => {
     expect(getCurrentStoredStoreAppSession('shop.myshopify.com', storage as any)).toEqual(secondSession)
   })
 
-  test('clears all stored sessions for a store', () => {
```
amcaplan, line 59: A few tests were deleted here, I'm not clear as to why

packages/cli-kit/src/public/node/store-auth-session.ts
```diff
@@ -1,6 +1,12 @@
...
-import {LocalStorage} from '@shopify/cli-kit/node/local-storage'
-import {type JsonMapType} from '@shopify/cli-kit/node/toml'
+import {normalizeStoreFqdn} from './context/fqdn.js'
+import {LocalStorage} from './local-storage.js'
+import {setLastSeenUserId} from './session.js'
+import type {JsonMapType} from './toml/codec.js'
+import type {AdminSession} from './session.js'
+
+export const STORE_AUTH_APP_CLIENT_ID = '7e9cb568cfd431c538f36d1ad3f2b4f6'
+const STORE_AUTH_PROJECT_NAME = 'shopify-cli-store'
```
amcaplan: These should probably go in the package-level constants file

packages/theme/src/cli/utilities/theme-command.ts
```diff
@@ -323,14 +338,94 @@ export default abstract class ThemeCommand extends Command {
...
+  }
+
+  private scopesFromStoreAuthToken(token: string): string[] | undefined {
+    const payload = this.decodeJwtPayload(token)
+    if (!payload) return undefined
+
+    return this.readScopesFromTokenPayload(payload)
+  }
+
+  private decodeJwtPayload(token: string): Record<string, unknown> | undefined {
```
amcaplan: This function feels like it's going too deep for a command prototype class. I think we should extract some of these functions to a helper utility. Some of this stuff actually feels like it would go really well in CLI-kit, everything about parsing a JWT feels really generic.

packages/theme/src/cli/utilities/theme-command.ts
```diff
@@ -323,14 +338,94 @@ export default abstract class ThemeCommand extends Command {
...
+      const payload: unknown = JSON.parse(payloadJson)
+      if (!payload || typeof payload !== 'object' || Array.isArray(payload)) return undefined
+      return payload as Record<string, unknown>
+    } catch (error) {
+      if (error instanceof SyntaxError) return undefined
+      throw error
+    }
+  }
+
+  private readScopesFromTokenPayload(payload: Record<string, unknown>): string[] | undefined {
```
amcaplan: This function is super defensive, is that necessary?

packages/theme/src/cli/utilities/theme-command.ts
```diff
@@ -323,14 +338,94 @@ export default abstract class ThemeCommand extends Command {
...
+  private async storeAuthSessionForTheme(flags: FlagValues): Promise<AdminSession | undefined> {
+    const store = typeof flags.store === 'string' ? flags.store : undefined
+    const password = flags.password
+    if (!store || password) return undefined
+
+    const storedSession = getCurrentStoredStoreAppSession(store)
+    if (!storedSession) return undefined
+
+    const requiredScopes = this.storeAuthScopes()
+    const scopes = requiredScopes.length > 0 ? this.scopesFromStoreAuthToken(storedSession.accessToken) : []
```
amcaplan: Are we finding, parsing, and evaluating the token's scopes for every single store in the user's environments? That seems inefficient, unless I'm really misunderstanding.

pnpm-lock.yaml
```diff
@@ -920,48 +920,56 @@ packages:
     engines: {node: '>= 10'}
     cpu: [arm64]
     os: [linux]
+    libc: [glibc]
```
amcaplan: Should probably revert these changes - they aren't part of the PR content

packages/theme/src/cli/utilities/theme-command.ts
alfonso-noriega: No, `StoredSession` is the auth session for the store we are currently trying to run the command for. It is retrived in line 358:

```javascript
const storedSession = getCurrentStoredStoreAppSession(store)
```

packages/cli-kit/src/public/node/store-auth-session.ts
```diff
@@ -1,6 +1,14 @@
...
-import {type JsonMapType} from '@shopify/cli-kit/node/toml'
+import {normalizeStoreFqdn} from './context/fqdn.js'
+import {LocalStorage} from './local-storage.js'
+import {setLastSeenUserId} from './session.js'
+import {storeAuthAppClientId} from '../../private/node/constants.js'
+import * as jose from 'jose'
+import type {JsonMapType} from './toml/codec.js'
+import type {AdminSession} from './session.js'
+
+export const STORE_AUTH_APP_CLIENT_ID = storeAuthAppClientId
```
amcaplan: No need to re-export, we should just export directly from the constants file if it should be exported

packages/theme/src/cli/utilities/theme-command.ts
amcaplan: OK. But that still means a file read for every store in the environment. This is fairly expensive/slow if there are many environments.

packages/store/src/cli/services/store/execute/admin-context.test.ts
```diff
@@ -40,10 +39,7 @@ describe('prepareAdminStoreGraphQLContext', () => {
   test('returns the stored admin session, version, and full auth session', async () => {
     const result = await prepareAdminStoreGraphQLContext({store})
 
-    expect(loadStoredStoreSession).toHaveBeenCalledWith(store)
-    expect(recordStoreFqdnMetadata).toHaveBeenCalledOnce()
-    expect(recordStoreFqdnMetadata).toHaveBeenCalledWith(store, true)
-    expect(setLastSeenUserId).toHaveBeenCalledWith('42')
```
amcaplan, line 46: Centralizing analytics definitely has a great payoff in this PR. Nice work!

amcaplan (approved):

PR #7822, opened 2026-06-16
Add `--standard-events-inspector` flag to `theme preview`.

frandiox:

We are releasing [standard storefront events and actions](https://shopify.dev/docs/storefronts/themes/best-practices/standard-events-and-actions).

Add a new flag to inject UI during development that helps debug events and actions.

### How to test your changes?

Run `shopify theme dev --standard-events-inspector` in Horizon v4. You should see a floating window like the following:

image

### Checklist


packages/theme/src/cli/utilities/theme-environment/html.ts
```diff
@@ -180,6 +182,11 @@ async function tryProxyRequest(event: H3Event, ctx: DevServerContext, response:
 
   if (proxyResponse.status < 400) {
     outputDebug(`Proxy status: ${proxyResponse.status}. Returning proxy response.`)
+
+    if ((proxyResponse.headers.get('content-type') ?? '').includes('text/html')) {
```
lucyxiang: ?

bashu-shopify:
just checking that you're aware: it looks like this PR also adds `standardEventsDevBundle`, which is separate from the Inspector. I'm fine with it personally (devs should have the standard events dev bundle when working on a theme), but the PR description doesn't reflect that.

lucyxiang:
:Tophat: with flag shows inspector

Screenshot 2026-06-16 at 11 00 51

Without flag doesn't
Screenshot 2026-06-16 at 11 01 33

packages/theme/src/cli/utilities/theme-environment/html.ts
bashu-shopify: Yeah, this seems too broad. Every text/html content type response would be affected here?

bashu-shopify (changes requested):
apart from the if-condition in `tryProxyRequest`, this looks good.

isaacroldan (approved):
Ok from the dev-experience side

packages/theme/src/cli/services/dev.ts
```diff
@@ -132,6 +133,8 @@ export async function dev(options: DevOptions) {
       port,
       open: options.open,
       liveReload: options['live-reload'],
+      standardEventsDevBundle: options['standard-events-inspector'],
```
lucyxiang: Non blocking but if DevBundle always follow flag behaviour, does it need to be an option here?

packages/theme/src/cli/services/dev.ts
```diff
@@ -153,7 +156,7 @@ export async function dev(options: DevOptions) {
     backgroundJobPromise,
     renderDevSetupProgress()
       .then(serverStart)
-      .then(() => {
+      .then(async () => {
```
lucyxiang: ?

lucyxiang: I think it's leftover from a refactor, removing

packages/theme/src/cli/services/dev.ts
bashu-shopify: i think we can leave it bc in the future I would want to turn it on independently

lucyxiang (commented):
~Looks good, @frandiox  one change I made was made `standardEventsDevBundle` true only when `standardEventsInspector` flag is passed. In the original PR `standardEventsDevBundle` was always true~
Nvm intention decision from https://shopify.slack.com/archives/C099734SGN7/p1776188523891689?thread_ts=1776186844.346429&cid=C099734SGN7

bashu-shopify (approved):
lgtm

PR #7838, opened 2026-06-16
Support store-auth backed store info

amcaplan:
## What changed
- Preserve the existing enriched `store info` path as the preferred source of store metadata.
- When a stored store-auth session exists, attempt the enriched lookup without prompting for login.
- Fall back to the stored store-auth/Admin API path only when the enriched lookup cannot proceed without prompting, or when the store is not found there.
- Share Admin API stored-auth error handling with `store execute`.
- Update command help and generated CLI docs to describe the bottom-line behavior without exposing internal auth details.

## Store-auth fallback field coverage
The fallback path can return:
- ID
- Display name
- Subdomain
- Store owner name/email
- Plan public display name
- Admin URL
- `Type: Dev` when the Admin plan reports a partner development store

The fallback path does not include fields that currently only come from the enriched organization-backed lookup:
- Organization name / organization ID
- Feature preview handle
- Full store type mapping such as production, collaborator, client transfer, or development superset

When those fields are unavailable, `store info` omits them from text and JSON output instead of blocking the command. Stores that are available through the enriched lookup still get the fuller output, even if store auth also exists.

## Testing
- `pnpm vitest packages/store/src/cli/services/store/info/index.test.ts packages/store/src/cli/services/store/info/destinations.test.ts packages/store/src/cli/services/store/info/organization-shop.test.ts packages/store/src/cli/services/store/info/plan.test.ts packages/store/src/cli/services/store/info/result.test.ts packages/store/src/cli/commands/store/info.test.ts packages/store/src/cli/services/store/execute/admin-transport.test.ts packages/cli-kit/src/public/node/session.test.ts`
- `pnpm nx run store:type-check`

### DIFF
--- a/packages/store/src/cli/services/store/admin-errors.ts
+++ b/packages/store/src/cli/services/store/admin-errors.ts
@@ -1,0 +1,68 @@
+import {throwReauthenticateStoreAuthError} from './auth/recovery.js'
+import {clearStoredStoreAppSession} from './auth/session-store.js'
+import {AbortError} from '@shopify/cli-kit/node/error'
+import type {StoredStoreAppSession} from './auth/session-store.js'
+
+interface GraphQLClientErrorLike {
+  response: {status?: number; errors?: unknown}
+}
+
+export function isGraphQLClientErrorLike(error: unknown): error is GraphQLClientErrorLike {
+  if (!error || typeof error !== 'object' || !('response' in error)) return false
+  const response = (error as {response?: unknown}).response
+  return Boolean(response) && typeof response === 'object'
+}
+
+function graphQLClientErrorStatus(error: unknown): number | undefined {
+  if (!isGraphQLClientErrorLike(error)) return undefined
+  const status = error.response.status
+  return typeof status === 'number' ? status : undefined
+}
+
+// Lower-cased substrings that Node's fetch/undici implementations use to signal an
+// aborted request. cli-kit's lower-level transport (`isTransientNetworkError` in
+// packages/cli-kit/src/private/node/api.ts) already recognizes 'the operation was
+// aborted'; we cover that shape and add 'the user aborted a request' (the literal
+// node-fetch surfaces) so a fetch that bubbles past cli-kit's retry layer is still
+// classified correctly. Exported so tests reference the same source of truth as
+// production.
+export const ABORTED_FETCH_MESSAGE_FRAGMENTS = ['the user aborted a request', 'the operation was aborted'] as const
+
+function isUserAbortedFetchError(error: unknown): boolean {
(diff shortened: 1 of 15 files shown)


packages/store/src/cli/services/store/info/index.ts
```diff
@@ -1,17 +1,51 @@
-import {fetchDestinationsContext} from './destinations.js'
-import {fetchOrganizationShop} from './organization-shop.js'
-import {mapPlanToPublicHandle} from './plan.js'
+import {classifyAdminApiError, throwIfStoredStoreAuthIsInvalid} from '../admin-errors.js'
+import {loadStoredStoreSession} from '../auth/session-lifecycle.js'
```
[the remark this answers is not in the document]
amcaplan: This is no longer relevant

packages/store/src/cli/services/store/admin-errors.ts
```diff
@@ -0,0 +1,68 @@
+import {throwReauthenticateStoreAuthError} from './auth/recovery.js'
+import {clearStoredStoreAppSession} from './auth/session-store.js'
+import {AbortError} from '@shopify/cli-kit/node/error'
+import type {StoredStoreAppSession} from './auth/session-store.js'
+
+interface GraphQLClientErrorLike {
```
alfonso-noriega, line 6: I love this refactor!

packages/store/src/cli/services/store/info/index.ts
```diff
@@ -21,15 +59,70 @@ export async function getStoreInfo(options: GetStoreInfoOptions): Promise<StoreI
...
+    if (!hasStoredStoreAuth || !isBusinessPlatformFallbackError(error)) {
+      throw error
+    }
+
+    outputDebug(`BP store info lookup failed; falling back to stored store auth: ${errorMessage(error)}`)
+    return getAdminStoreInfo(store)
+  }
+}
+
+async function getAdminStoreInfo(store: string): Promise<StoreInfoResult> {
```
alfonso-noriega, line 76: Are we only calling this on preview stores? in which other cases there won't be BP store info? if so, the name is a little confusing but it does represent what the function is doing. So no strong opinion on changing this, just trying to clarify when do we hit this branch

alfonso-noriega (approved):

packages/store/src/cli/services/store/info/index.ts
amcaplan: It's on any store-auth store if we can't auth to BP or BP doesn't get info about the store

alfonso-noriega (commented):
Actually I am not sure we can do an admin graphql request on preview stores to get its data, I think that won't be allowed.

PR #7847, opened 2026-06-17
Only treat .myshopify.io as a normalized store FQDN suffix

PhilippeCollin:
The `shopify.io` suffix check matched any domain ending in `shopify.io` (e.g. `notshopify.io`); tighten it to `.myshopify.io`.

`*.myshopify.io` is the hostname the legacy local dev server assigns to stores (see [`dev-server-2016.ts`](https://github.com/Shopify/cli/blob/0991f169bc613b2378ecb2e5c51af4b28268fa28/packages/cli-kit/src/public/node/vendor/dev_server/dev-server-2016.ts#L25)), which is why it's treated as already-normalized.

### DIFF
--- a/.changeset/normalize-store-fqdn-suffix.md
+++ b/.changeset/normalize-store-fqdn-suffix.md
@@ -1,0 +1,5 @@
+---
+'@shopify/cli-kit': patch
+---
+
+Only treat `.myshopify.io` (not any `shopify.io` suffix) as an already-normalized store FQDN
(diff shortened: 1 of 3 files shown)


graygilmore (approved):
Makes sense. I can't think of any other `*shopify.io` that we'd use through here. Tbh we might even just be able to drop `myshopify.io` altogether but we can poke at that later.

graygilmore (approved):

EvilGenius13 (approved):

alfonso-noriega (approved):
Great catch!

PR #7714, opened 2026-06-04
Add store auth list for local store auth sessions

dmerand:
## Summary

Add `shopify store auth list` for stores authenticated directly with `shopify store auth`.

Closes https://github.com/shop/issues-develop/issues/22919

## Scope

- add `shopify store auth list`
- list only local store-auth sessions; no organization auth probe and no organization prompt
- render text output as stores authenticated directly with `shopify store auth`
- emit deterministic JSON as `{sessions: [...]}` with `kind: "store"` rows so agents can resume without reading cache internals
- keep access tokens and refresh tokens out of the command output
- add cross-command hints:
  - `store auth list` points users to `shopify store list` for organization stores
  - `store list` points users to `shopify store auth list` for direct store-auth sessions
- keep `shopify store list` organization-backed only; no `--from`, no `auto`, no fallback behavior

## Command split

This stack intentionally keeps the two jobs separate:

- `shopify store list` lists organization stores using the active CLI account, prompting for organization selection unless `--organization-id` is provided.
- `shopify store auth list` lists stores already present in local store-auth storage.

## Not changed / deferred

- Did not keep `store list --from store-auth` or `--from auto`. Those were the source of the confusing mixed-command contract this PR replaces.
- Did not attempt organization auth or show organization stores from `store auth list`. That stays in `shopify store list`.
- Did not expose access tokens or refresh tokens in `store auth list --json`. Agents get enough metadata to choose a stored session without reading cache internals; token return would require a separate product/security decision.
- Did not add the LocalStorage enumeration API suggested in follow-through review. That belongs in a focused cli-kit/storage follow-up, not in the command behavior PR.
- Did not add E2E evals for “returning to my preview store.” That is tracked separately in https://github.com/shop/issues-develop/issues/22970.

## Smoke checks

- `shopify store auth list --json` returns local store-auth sessions and does not prompt for organization auth
- `shopify store list --help` has no `--from` flag
- `shopify auth list --help` is not registered

PR #7864, opened 2026-06-19
Escape store auth session keys

alfonso-noriega:

Follow-up to discussion on #7709 about avoiding recursive crawling of `conf` storage for store-auth sessions.

Escapes dots in store-auth session keys before passing them to `conf`, so a store domain is persisted as one top-level key instead of a nested dotted path. The listing code now scans only top-level store-auth buckets and ignores legacy nested entries.

This intentionally means stores authenticated with the old nested key format may need to be re-authenticated.

### How to test your changes?

```bash
pnpm --filter @shopify/store vitest src/cli/services/store/auth/session-store.test.ts src/cli/services/store/auth/stored-auth.test.ts
pnpm --filter @shopify/store lint
pnpm --filter @shopify/store type-check
```

### Post-release steps

None.

### Checklist

### DIFF
--- a/packages/store/src/cli/services/store/auth/config.ts
+++ b/packages/store/src/cli/services/store/auth/config.ts
@@ -7,7 +7,11 @@ export function storeAuthRedirectUri(port: number): string {
 }
 
 export function storeAuthSessionKey(store: string): string {
-  return `${STORE_AUTH_APP_CLIENT_ID}::${store}`
+  return `${STORE_AUTH_APP_CLIENT_ID}::${escapeStoreAuthSessionKeySegment(store)}`
+}
+
+function escapeStoreAuthSessionKeySegment(value: string): string {
+  return value.replace(/\./g, '\\.')
 }
 
 export function maskToken(token: string): string {
--- a/packages/store/src/cli/services/store/auth/session-store.ts
+++ b/packages/store/src/cli/services/store/auth/session-store.ts
@@ -212,26 +212,6 @@ function readRawStoreSessionStorage(storage: LocalStorage<StoreSessionSchema>): 
   return (storage as unknown as {config?: {store?: RawStoreSessionStorage}}).config?.store ?? {}
 }
 
-function collectCurrentStoredStoreAppSessions(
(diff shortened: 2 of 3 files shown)


alfonso-noriega:
* **#7864** <a href="https://app.graphite.com/github/pr/Shopify/cli/7864?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Shopify/cli/7864?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#7709** <a href="https://app.graphite.com/github/pr/Shopify/cli/7709?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#7708** <a href="https://app.graphite.com/github/pr/Shopify/cli/7708?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

packages/store/src/cli/services/store/auth/config.ts
```diff
@@ -7,7 +7,11 @@ export function storeAuthRedirectUri(port: number): string {
 }
 
 export function storeAuthSessionKey(store: string): string {
-  return `${STORE_AUTH_APP_CLIENT_ID}::${store}`
+  return `${STORE_AUTH_APP_CLIENT_ID}::${escapeStoreAuthSessionKeySegment(store)}`
+}
+
+function escapeStoreAuthSessionKeySegment(value: string): string {
+  return value.replace(/\./g, '\\.')
```
isaacroldan, line 14: is it better to escape or to replace them? maybe is safer to use `-`?

alfonso-noriega: I think escaping is safer here than replacing with `-`: `-` is valid in store domains, so replacement can create collisions between distinct domains. `dot-prop` explicitly supports escaped dots for this exact case, and the persisted JSON key remains the original domain string rather than an encoded/replaced value. If we want to avoid relying on dot-prop escaping semantics entirely, I think `encodeURIComponent(store)` would be the safer alternative over `-` replacement.

isaacroldan (approved):

PR #7870, opened 2026-06-19
fix(theme): ignore atomic-write tmp files in the theme dev watcher

andershagbard:
## Summary

This is an error which happens because Claude Code generates these tmp files. Not sure if other tools use the same method or naming.

- Editor tools (including AI coding assistants) create temporary files like `blocks/foo.liquid.tmp.93809.4dbd82e0b95a` during atomic save operations and delete them within milliseconds.
- Because the per-event debounce fires independently for `add` and `unlink` events, the file watcher sees the `add`, tries to upload the invalid asset key to Shopify, and crashes `shopify theme dev` — requiring a manual restart.
- Adds `**/*.tmp.*` to `DEFAULT_IGNORE_PATTERNS` so chokidar never emits events for these files.

## Root cause

`queueFsEvent` debounces per `${fileKey}:${eventName}`, meaning `add` and `unlink` are queued separately. By the time the 250ms debounce fires for `add`, the file is already gone — but Shopify still gets the upload attempt with an invalid key.

## Test plan

- [x] New unit test asserting `**/*.tmp.*` is present in chokidar's `ignored` option
- [x] All 44 existing `theme-fs` tests pass
- [x] Type-check passes
- [x] Lint passes

### DIFF
--- a/packages/theme/src/cli/constants.ts
+++ b/packages/theme/src/cli/constants.ts
@@ -32,7 +32,7 @@ export const DEFAULT_IGNORE_PATTERNS = [
   '**/node_modules/',
   '.prettierrc.json',
   // Atomic-write temp files (e.g. blocks/foo.liquid.tmp.bar.baz)
-  '**/{blocks,layouts,snippets}/**/*.liquid.tmp.*',
+  '**/{blocks,layout,snippets}/**/*.liquid.tmp.*',
   '**/{config,locales}/**/*.json.tmp.*',
   '**/{sections,templates}/**/*.{liquid,json}.tmp.*',
 ]
--- a/packages/theme/src/cli/constants.ts
+++ b/packages/theme/src/cli/constants.ts
@@ -31,6 +31,10 @@ export const DEFAULT_IGNORE_PATTERNS = [
   '**/config.yml',
   '**/node_modules/',
   '.prettierrc.json',
+  // Atomic-write temp files (e.g. blocks/foo.liquid.tmp.93809.4dbd82e0b95a): created and
+  // deleted within milliseconds, but Shopify's watcher sees the add before the delete and
+  // tries to upload an invalid key, crashing the dev server.
+  '**/*.tmp.*',
 ]
 
 export const MAX_GRAPHQL_THEME_FILES = 50
--- a/packages/theme/src/cli/constants.ts
+++ b/packages/theme/src/cli/constants.ts
@@ -31,10 +31,7 @@ export const DEFAULT_IGNORE_PATTERNS = [
   '**/config.yml',
   '**/node_modules/',
   '.prettierrc.json',
-  // Atomic-write temp files (e.g. blocks/foo.liquid.tmp.93809.4dbd82e0b95a): created and
-  // deleted within milliseconds, but Shopify's watcher sees the add before the delete and
-  // tries to upload an invalid key, crashing the dev server.
-  '**/*.tmp.*',
+  '**/*.tmp.*', // Atomic-write temp files (e.g. blocks/foo.liquid.tmp.93809.4dbd82e0b95a)
 ]
 
 export const MAX_GRAPHQL_THEME_FILES = 50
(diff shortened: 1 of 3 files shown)


packages/theme/src/cli/constants.ts
```diff
@@ -31,6 +31,7 @@ export const DEFAULT_IGNORE_PATTERNS = [
   '**/config.yml',
   '**/node_modules/',
   '.prettierrc.json',
+  '**/*.tmp.*', // Atomic-write temp files (e.g. blocks/foo.liquid.tmp.93809.4dbd82e0b95a)
```
graygilmore: Thanks for this @andershagbard. We're just having a discussion internally here because there are going to be themes that already contain files with `.tmp.` in the name that this would break.

The purpose behind the PR makes sense but we need to think of how to make sure our fix doesn't break existing themes.

andershagbard: I added this because the devserver will crash because the API doesn't accept that file name. I assume it's blocked serverside, and therefore these files would not exist?

andershagbard: Correct me if I am wrong, but I think this can be enforced in all folders except assets.

I am getting these errors in the theme editor:
> FileSaveError: Must have a .liquid file extension

> FileSaveError: "sections/test.json.tmp.1231231" contains illegal characters

graygilmore: You’re right that files like `sections/test.json.tmp.1231231` are not valid theme files and won’t be accepted server-side.

The piece I’m worried about is the breadth of `**/*.tmp.*`. `DEFAULT_IGNORE_PATTERNS` is used for the initial file glob too, so this can hide real files (e.g. `snippets/test.tmp.liquid`).

Maybe we can narrow this to non-asset theme folders and only ignore files where the temp suffix is appended after a valid extension?

```ts
[
  '**/blocks/**/*.liquid.tmp.*',
  '**/layout/**/*.liquid.tmp.*',
  '**/snippets/**/*.liquid.tmp.*',
  '**/config/**/*.json.tmp.*',
  '**/locales/**/*.json.tmp.*',
  '**/sections/**/*.{liquid,json}.tmp.*',
  '**/templates/**/*.{liquid,json}.tmp.*',
]
```

That should catch `sections/test.json.tmp.1231231` and `blocks/foo.liquid.tmp.93809.4dbd82e0b95a`, but not ignore valid files like `snippets/test.tmp.liquid` or any legitimate `assets/*` files.

As a personal workaround for now you could also add the following to your `.shopifyignore`:

```regex
/\.tmp\.[0-9]+(\.[0-9a-fA-F]+)?$/
```

packages/theme/src/cli/constants.ts
```diff
@@ -31,6 +31,10 @@ export const DEFAULT_IGNORE_PATTERNS = [
   '**/config.yml',
   '**/node_modules/',
   '.prettierrc.json',
+  // Atomic-write temp files (e.g. blocks/foo.liquid.tmp.bar.baz)
+  '**/{blocks,layouts,snippets}/**/*.liquid.tmp.*',
```
graygilmore:
```suggestion
  '**/{blocks,layout,snippets}/**/*.liquid.tmp.*',
```

packages/theme/src/cli/utilities/theme-fs.test.ts
```diff
@@ -158,6 +158,23 @@ describe('theme-fs', () => {
       expect(watchedPaths.some((path) => path.includes('listings'))).toBe(false)
     })
 
+    test('ignores editor atomic-write tmp files in the watcher', async () => {
```
graygilmore: This is essentially just a duplicate of passing the options in the first place so I'd be happy seeing this test removed.

graygilmore (dismissed):

graygilmore (approved):

isaacroldan (approved):

PR #7892, opened 2026-06-22
Add `shopify store bulk` commands backed by a shared cli-kit engine

teddyhwang:
## What

Adds `shopify store bulk execute`, `shopify store bulk status`, and `shopify store bulk cancel`, mirroring the existing `shopify app bulk` commands but running against a store using **stored auth** (`shopify store auth`) — so you no longer need an app to run Admin API bulk operations.

## Why

`shopify app bulk *` requires a linked app + org/store selection. For folks who just want to run bulk operations against a store they've already authenticated with via `shopify store auth`, that's unnecessary friction. These commands reuse the stored store session directly.

## How — shared engine, no copy/paste

Rather than duplicating the app implementation into `store`, the auth-agnostic bulk-operation primitives now live in **`@shopify/cli-kit/node/api/bulk-operations`** and are consumed by both packages (`store` can't depend on `app`, and cli-kit already owns the Admin API client + had a `bulkOperations` codegen project).

Shared in cli-kit:
- `runBulkOperationQuery`, `runBulkOperationMutation`, `stageFile`
- `watchBulkOperation`, `shortBulkOperationPoll`, `downloadBulkOperationResults`
- `fetchBulkOperationById`, `fetchRecentBulkOperations`, `cancelBulkOperationRequest`
- `formatBulkOperationStatus`, `renderBulkOperationUserErrors`, `formatBulkOperationCancellationResult(op, statusCommand)`
- `resolveApiVersion`, `isMutation`, `validateSingleOperation`, id helpers, `BULK_OPERATIONS_MIN_API_VERSION`
- The bulk-operations GraphQL docs (moved from `app`, codegen repointed)

Per-package (kept thin):
- **app** — `createAdminSessionAsApp` auth, org/app/store display, dev-store mutation check, `shopify app bulk status` help text.
- **store** — `prepareBulkAdminContext` (loads the stored session into an `AdminSession`), store-only display, `shopify store bulk status` help text, and a `--allow-mutations` opt-in gate (consistent with `store execute`) instead of app's dev-store check.

## Notes / decisions

- `app` was refactored to consume the shared engine too, so there's a single source of truth (no app↔cli-kit duplication). Git tracked the primitive + GraphQL files as renames, preserving history.
- **Mutation safety differs by design:** `app bulk` knows the store type and blocks mutations on non-dev stores; `store bulk` only has a stored session, so it uses the `--allow-mutations` opt-in. Happy to adjust if reviewers prefer a different gate.

## Testing

- Type-check ✅ (cli-kit, app, store)
- Lint ✅
- Tests ✅ — cli-kit 68, app 87, store 24 (all bulk-operation suites)


teddyhwang:
Splitting this into two smaller, stacked PRs for easier review:
1. Extract the shared bulk-operation engine into `cli-kit` (+ refactor `app` to use it)
2. Add the `shopify store bulk` commands (stacked on #1)

Closing this combined PR in favor of those.

PR #7894, opened 2026-06-22
Split CLI error analytics grouping

### LINKED ISSUE #7891: split observe analytics group for cleaner resiliency tracking
## Summary
A single error-analytics group (Observe grouping hash `16864652937831232783`) is acting as a catch-all for essentially **every GraphQL / Admin-API client error in the CLI**, across four unrelated product areas. Over the last 10 days it held ~1,170 events spanning `app`, `theme`, `store`, and `hydrogen`. This makes the group un-routable, un-actionable, and a permanent source of false P1 escalations. We should make the CLI emit a meaningful `groupingHash` so distinct failures land in distinct buckets.

Context: this surfaced from resiliency issue [Vault 31608](https://vault.shopify.io/issues/31608) / shop/issues#32995 ("Access denied for themes field… `read_themes`"). That specific error was fixed in CLI 4.2.0 (#7652) and is confirmed gone on 4.2.0 — but the resiliency item stays hot because ~94% of the bucket is unrelated errors sharing the same hash.

## What's actually in the bucket (last 10 days, ~1,170 events)
By slice: `app` 618 · `theme` 385 · `store` 105 · `hydrogen` 40 · unknown 18 · `cli` 4.
Handled split: 708 unhandled / 462 handled.

Distinct families lumped together:
- **App Management 403 "Unauthorized"** (~316; slice `app`, `app deploy`) — caller lacks permission/membership. Owner: App Management.
- **401 authentication failures** (~390; slice `store`/`theme`, e.g. `store info`, `store execute`) — invalid/expired session token; 163 are literally `Service is not valid for authentication`. Owner: CLI auth/identity.
- **Missing access scope ACCESS_DENIED** (166 = `read_themes` 75 + `write_themes` 91; slice `theme`) — custom-app token missing scope. Already fixed in 4.2.0 (#7652); now a clean `AbortError`, aging out as adoption rolls.
- **5xx server errors** (~150; Admin + App-Management HTTP 500) — server-side API reliability, not a CLI bug. Owner: the respective API teams.
- **THROTTLED rate limiting** (46; slice `theme`) — expected/transient; should be retried or suppressed, not crash-reported.
- **Hydrogen 403/404** (~40; `hydrogen deploy`/`link`) — separate surface again.

## Root cause
In `packages/cli-kit/src/public/node/error-handler.ts` (`sendErrorToBugsnag`):
1. Every error is rebuilt as `reportableError = new Error(error.message)`, so the error **class is always the generic `Error`**.
2. **No `event.groupingHash` is ever set**, so the backend falls back to stack-trace grouping (cf. the `stack_frame_grouping_hash` column).
3. Stack frame paths are aggressively normalized (`cleanStackFrameFilePath`), and there are really two message shapes thrown from the same request site (`graphql-request`'s `GraphQL Error (Code: NNN): {…}` for theme/store, and `The Admin/App Management GraphQL API responded unsuccessfully with the HTTP status NNN …` for app/hydrogen).

Net: same class + same normalized stack + ignored message ⇒ one bucket for all of them.

## Why we should track these separately
- **Different owners.** This group is assigned to one team, but it contains errors owned by App Management, Hydrogen, Storefront, CLI auth, and the server-side API teams. It cannot be routed as a single item.
- **Different root causes and fixes.** Scope→UX messaging, 401→token refresh, 403→app permissions, THROTTLED→backoff/suppress, 5xx→server reliability. One issue can't carry five fixes, one assignee, or one fix-due date.
- **Expected vs. real regressions get mixed.** Most of the volume is expected user-config / transient throttles; a genuine regression in any single family (e.g. a 401 or 5xx spike from a real bug) is invisible inside the aggregate.
- **It manufactures false P1s.** Severity escalates on aggregate volume, so the group oscillates P1↔P3 independent of whether any underlying problem is bad — and a shipped fix (like #7652) can never turn it green.

## Proposed fix
Set a meaningful grouping key in the Bugsnag `eventHandler`, reusing the existing analytics taxonomy (`categorizeError` + `formatErrorMessage` in `packages/cli-kit/src/private/node/analytics/error-categorizer.ts`, already used by `storage.ts` to emit `error:${category}:${signature}` events):

```ts
import {categorizeError, formatErrorMessage} from '../../private/node/analytics/error-categorizer.js'

const category = categorizeError(error)
event.groupingHash = `${sliceName}:${category.toLowerCase()}:${formatErrorMessage(error, category)}`
```

Notes:
- Include `slice_name` so `app` / `theme` / `store` / `hydrogen` split immediately.
- Trim the message at the first `: {` before categorizing — the raw GraphQL `ClientError` dumps the full request/response JSON, and the literal `"request"` substring currently mis-routes errors into the `network` category. Trimming makes categories semantically correct (scope→Permission, 401→Authentication, THROTTLED→RateLimit) and keeps signatures stable.
- Complementary, higher-leverage: stop reporting known transient/user errors (THROTTLED, and arguably the 401 "Service is not valid for authentication") as *unexpected* — same pattern as #7652 — so they leave crash reporting entirely rather than just getting relabeled.

## Considerations
- Setting `groupingHash` universally re-buckets **all** CLI errors (a one-time grouping migration across every error dashboard), not just these. That's an improvement (grouping finally matches the analytics taxonomy) but should be a conscious, coordinated change.
- Update `packages/cli-kit/src/public/node/error-handler.test.ts`.

## References
- Resiliency: [Vault 31608](https://vault.shopify.io/issues/31608) / shop/issues#32995
- Prior fix: #7652 (shipped in CLI 4.2.0, 2026-06-16)
- Observe grouping hash: `16864652937831232783`
- Reporting code: `packages/cli-kit/src/public/node/error-handler.ts`

stephanie-shopify:
## Summary
- set Bugsnag groupingHash for CLI error reports using command slice + analytics error taxonomy
- normalize GraphQL ClientError payload dumps so access/auth/throttle errors don’t collapse into network buckets
- add regression coverage for app/theme/store/cli grouping behavior

closes  #7891  which stems  from this [resiliency issue](https://vault.shopify.io/teams/13725-Developer-Platforms/issues/31608-error-access-denied-for-themes-field-required-access-read_themes-access-scope-response-errors-message-access-denied-for-themes-field-required-access-read_themes-access-scope-locations-line-2-column-3-extensions)  which is currently an aggregate of a bunch of different issues so we can never properly address it

## Testing
- `pnpm vitest run packages/cli-kit/src/public/node/error-handler.test.ts` (passed in primary checkout)
- `pnpm nx type-check cli-kit` (passed in primary checkout)
- `pnpm nx lint cli-kit` (passed in primary checkout; existing warnings only)

## context
https://shopify.slack.com/archives/C0AFPGQJENQ/p1782147409267669

### DIFF
--- a/packages/cli-kit/src/public/node/error-handler.ts
+++ b/packages/cli-kit/src/public/node/error-handler.ts
@@ -217,7 +217,7 @@ export async function sendErrorToBugsnag(
         }
         const errorHandler = (error: unknown) => {
           if (error) {
-            reject(error)
+            reject(error instanceof Error ? error : new Error(String(error)))
           } else {
             resolve(reportableError)
           }
--- a/packages/cli-kit/src/public/node/error-handler.ts
+++ b/packages/cli-kit/src/public/node/error-handler.ts
@@ -12,6 +12,7 @@ import {
 } from './error.js'
 import {outputDebug, outputInfo} from './output.js'
 import {getEnvironmentData} from '../../private/node/analytics.js'
+import {categorizeError, formatErrorMessage} from '../../private/node/analytics/error-categorizer.js'
 import {isLocalEnvironment} from '../../private/node/context/service.js'
 import {bugsnagApiKey, reportingRateLimit} from '../../private/node/constants.js'
 import {CLI_KIT_VERSION} from '../common/version.js'
@@ -27,6 +28,72 @@ import {realpath} from 'fs/promises'
 // Allowed slice names for error analytics grouping.
 // Hardcoded list per product slices to keep analytics consistent.
 const ALLOWED_SLICE_NAMES = new Set<string>(['app', 'theme', 'hydrogen', 'store'])
+
+function sliceNameFromStartCommand(startCommand: string | undefined): string {
+  if (!startCommand) return 'cli'
+
+  const firstWord = startCommand.trim().split(/\s+/)[0] ?? 'cli'
+  return ALLOWED_SLICE_NAMES.has(firstWord) ? firstWord : 'cli'
+}
+
+interface GraphQLErrorPayload {
+  response?: {
+    status?: number
+    errors?: {
+      message?: string
+      extensions?: {
+        code?: string
+      }
+    }[]
+  }
+}
+
+function errorForGrouping(error: Error): Error {
+  const graphQLError = graphQLErrorForGrouping(error.message)
+  if (graphQLError) return graphQLError
+
+  const messageWithoutJsonDump = error.message.split(': {')[0] ?? error.message
+  return new Error(messageWithoutJsonDump)
+}
+
+function graphQLErrorForGrouping(message: string): Error | undefined {
+  const jsonStart = message.indexOf('{')
+  if (jsonStart === -1) return undefined
+
+  let payload: GraphQLErrorPayload
+  try {
+    payload = JSON.parse(message.slice(jsonStart)) as GraphQLErrorPayload
+    // eslint-disable-next-line no-catch-all/no-catch-all
+  } catch {
+    return undefined
+  }
+
+  const {response} = payload
+  if (!response) return undefined
+
+  const firstError = response.errors?.[0]
+  const code = firstError?.extensions?.code
+  const firstErrorMessage = firstError?.message?.replace(/too many requests/gi, 'rate limit exceeded')
+
+  const parts: string[] = []
+  if (response.status === 401 && !firstErrorMessage?.match(/auth|credential|token|unauthorized/i)) {
+    parts.push('Unauthorized')
+  }
+  if (response.status === 403 && !firstErrorMessage?.match(/access|denied|forbidden|permission|unauthorized/i)) {
+    parts.push('Forbidden')
+  }
+  if (response.status) parts.push(`HTTP ${response.status}`)
+  if (code) parts.push(code)
+  if (firstErrorMessage) parts.push(firstErrorMessage)
+
+  return new Error(parts.length > 0 ? parts.join(' ') : (message.split(': {')[0] ?? message))
+}
+
+function bugsnagGroupingHash(error: Error, sliceName: string): string {
+  const groupingError = errorForGrouping(error)
+  const category = categorizeError(groupingError)
+  return `${sliceName}:${category.toLowerCase()}:${formatErrorMessage(groupingError, category)}`
+}
 
 export async function errorHandler(
   error: Error & {exitCode?: number | undefined},
@@ -73,11 +140,6 @@ export async function sendErrorToBugsnag(
   try {
     if (isLocalEnvironment() || settings.debug) {
       outputDebug(`Skipping Bugsnag report`)
-      return {reported: false, error, unhandled: undefined}
-    }
-
-    if (exitMode === 'expected_error') {
-      outputDebug(`Skipping Bugsnag report for expected error`)
       return {reported: false, error, unhandled: undefined}
     }
 
@@ -147,15 +209,15 @@ export async function sendErrorToBugsnag(
           // Attach command metadata so we know which CLI command triggered the error
           const {commandStartOptions} = metadata.getAllSensitiveMetadata()
           const {startCommand} = commandStartOptions ?? {}
+          const sliceName = sliceNameFromStartCommand(startCommand)
+          event.groupingHash = bugsnagGroupingHash(reportableError, sliceName)
           if (startCommand) {
-            const firstWord = startCommand.trim().split(/\s+/)[0] ?? 'cli'
-            const sliceName = ALLOWED_SLICE_NAMES.has(firstWord) ? firstWord : 'cli'
             event.addMetadata('custom', {slice_name: sliceName})
           }
         }
         const errorHandler = (error: unknown) => {
           if (error) {
-            reject(error instanceof Error ? error : new Error(String(error)))
+            reject(error)
           } else {
             resolve(reportableError)
           }
@@ -192,7 +254,7 @@ export function cleanStackFrameFilePath({
     ? currentFilePath
     : path.joinPath(projectRoot, currentFilePath)
 
-  const matchingPluginPath = pluginLocations.find(({pluginPath}) => fullLocation.startsWith(pluginPath))
+  const matchingPluginPath = pluginLocations.filter(({pluginPath}) => fullLocation.startsWith(pluginPath))[0]
 
   if (matchingPluginPath !== undefined) {
     // the plugin name (e.g. @shopify/cli-kit), plus the relative path of the error line from within the plugin's code (e.g. dist/something.js )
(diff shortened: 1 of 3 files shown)


packages/cli-kit/src/public/node/error-handler.ts
```diff
@@ -76,11 +143,6 @@ export async function sendErrorToBugsnag(
       return {reported: false, error, unhandled: undefined}
     }
 
-    if (exitMode === 'expected_error') {
-      outputDebug(`Skipping Bugsnag report for expected error`)
-      return {reported: false, error, unhandled: undefined}
-    }
-
     // If the error was unexpected, we flag it as "unhandled" in Bugsnag. This is a helpful distinction.
     const unhandled = exitMode === 'unexpected_error'
```
stephanie-shopify, line 147: 🚨 **Critical / not in the summary:** the diff removes the `if (exitMode === 'expected_error') { return {reported:false…} }` guard that used to sit right above this line, and flips the test from `'when error is expected'` (not reported) to `'processes AbortErrors as handled'` (reported: true).

Net effect: **every AbortError now flows to Bugsnag as a handled event** — these were deliberately suppressed before. This is unrelated to grouping, increases Bugsnag volume/cost by an unquantified amount, and runs *opposite* to the linked issue's own recommendation ("stop reporting known transient/user errors … so they leave crash reporting entirely").

Please either (a) revert this removal and keep the PR purely about grouping, or (b) split it out, justify it with projected event volume, and document it in **both** the summary and the changeset (which currently only mentions grouping). As written it's the riskiest line in the diff and the least visible.

packages/cli-kit/src/public/node/error-handler.ts
```diff
@@ -147,9 +209,9 @@ export async function sendErrorToBugsnag(
           // Attach command metadata so we know which CLI command triggered the error
           const {commandStartOptions} = metadata.getAllSensitiveMetadata()
           const {startCommand} = commandStartOptions ?? {}
+          const sliceName = sliceNameFromStartCommand(startCommand)
+          event.groupingHash = bugsnagGroupingHash(reportableError, sliceName)
```
stephanie-shopify, line 213: Setting `groupingHash` unconditionally re-buckets **all** CLI errors (known, acknowledged). The sharp edge is the `Unknown` category: `formatGenericError` keys on the first 50 slugified chars, so two genuinely distinct crashes in the same slice that share an opening message fragment will now **merge** into one bucket, discarding the stack-trace grouping that keeps them apart today.

Consider skipping the override when `category === Unknown` (let Bugsnag's stack grouping stand), or folding a short stack fingerprint into the hash for that case. That keeps the win for categorized buckets without coarsening the truly-unknown ones.

packages/cli-kit/src/public/node/error-handler.ts
```diff
@@ -28,6 +29,72 @@ import {realpath} from 'fs/promises'
...
+
+  const firstError = response.errors?.[0]
+  const code = firstError?.extensions?.code
+  const firstErrorMessage = firstError?.message?.replace(/too many requests/gi, 'rate limit exceeded')
+
+  const parts: string[] = []
+  if (response.status === 401 && !firstErrorMessage?.match(/auth|credential|token|unauthorized/i)) {
+    parts.push('Unauthorized')
+  }
+  if (response.status === 403 && !firstErrorMessage?.match(/access|denied|forbidden|permission|unauthorized/i)) {
```
stephanie-shopify, line 82: ⚠️ The injected `'Forbidden'` is misleading given the taxonomy: `'forbidden'` is an **Authentication** term, not a Permission term (`['unauthorized','forbidden','auth','token','credential']` vs `['permission','denied','access','insufficient']`). So a 403 with a generic message lands in `:authentication:`, even though 403 = permission semantically — and the issue explicitly frames App-Management 403 as a *permission* problem distinct from 401 auth.

Decide explicitly: if 403 should be `permission`, push `'access denied'` instead of `'Forbidden'`. If everything 4xx→authentication is fine, drop the misleading word and add a comment. Either way there's **no test** for the 403-with-JSON-payload path, so current behavior is unverified.

packages/cli-kit/src/public/node/error-handler.ts
```diff
@@ -28,6 +29,72 @@ import {realpath} from 'fs/promises'
...
+  } catch {
+    return undefined
+  }
+
+  const {response} = payload
+  if (!response) return undefined
+
+  const firstError = response.errors?.[0]
+  const code = firstError?.extensions?.code
+  const firstErrorMessage = firstError?.message?.replace(/too many requests/gi, 'rate limit exceeded')
```
stephanie-shopify, line 76: 💡 `'too many requests'` is already a `RateLimit` term, so this rewrite doesn't change categorization — both forms categorize as `rate_limit`. If the goal is just a tidier signature, a one-line comment would help; otherwise it's removable.

packages/cli-kit/src/public/node/error-handler.ts
```diff
@@ -28,6 +29,72 @@ import {realpath} from 'fs/promises'
...
+      }
+    }[]
+  }
+}
+
+function errorForGrouping(error: Error): Error {
+  const graphQLError = graphQLErrorForGrouping(error.message)
+  if (graphQLError) return graphQLError
+
+  const messageWithoutJsonDump = error.message.split(': {')[0] ?? error.message
```
stephanie-shopify, line 55: 💡 `String.prototype.split` always returns a non-empty array, so `[0]` is never `undefined` and the `?? error.message` fallback is unreachable. Same dead fallback on the `return new Error(... ?? message)` line below. Harmless, just drop for clarity.

packages/cli-kit/src/public/node/error-handler.ts
```diff
@@ -192,7 +254,7 @@ export function cleanStackFrameFilePath({
     ? currentFilePath
     : path.joinPath(projectRoot, currentFilePath)
 
-  const matchingPluginPath = pluginLocations.find(({pluginPath}) => fullLocation.startsWith(pluginPath))
+  const matchingPluginPath = pluginLocations.filter(({pluginPath}) => fullLocation.startsWith(pluginPath))[0]
```
stephanie-shopify, line 257: 💡 Unrelated change, and strictly worse: `.find(pred)` short-circuits on first match; `.filter(pred)[0]` walks the whole array and allocates a throwaway array for an identical result. Recommend reverting to `.find(...)` unless there's a reason I'm missing.

stephanie-shopify (commented):
Review focused on one critical behavior change plus a few correctness/cleanup items. The grouping mechanism itself is well-designed and the GraphQL normalization targets the right root cause — main blocker is the silent removal of expected-error suppression (see inline). Posting as comments since this is my own PR.

packages/cli-kit/src/public/node/error-handler.ts
stephanie-shopify: 403 can be permission instead of forbidden

stephanie-shopify:
generally dont like this approach. feels brittle to have the hard coded categories and there are some weird unnecessary changes. going back to drawing board

PR #7896, opened 2026-06-22
Extract shared bulk-operation engine into cli-kit

teddyhwang:
## What

Extracts the auth-agnostic Admin API bulk-operation engine into a shared **`@shopify/cli-kit/node/api/bulk-operations`** module, and refactors the existing `shopify app bulk {execute,status,cancel}` commands to consume it. **No behavior change** to the app commands.

> **Stacked PR 1 of 2.** PR 2 (`shopify store bulk` commands) builds on this: #7897.

## Why

I'm adding `shopify store bulk` commands (PR 2) that run bulk operations using stored store auth instead of an app. Since `store` can't depend on `app`, the shared logic needs a common home — and `cli-kit` already owns the Admin API client and even had a `bulkOperations` codegen project. Extracting first keeps both command sets on a single source of truth instead of copy/pasting.

## What moved into cli-kit

- Primitives (with their tests): `runBulkOperationQuery`, `runBulkOperationMutation`, `stageFile`, `watchBulkOperation`, `shortBulkOperationPoll`, `downloadBulkOperationResults`
- New thin wrappers: `fetchBulkOperationById`, `fetchRecentBulkOperations`, `cancelBulkOperationRequest`
- Helpers: `resolveApiVersion`, `isMutation`, `validateSingleOperation`, `normalizeBulkOperationId`, `extractBulkOperationId`, `BULK_OPERATIONS_MIN_API_VERSION`
- Formatting: `formatBulkOperationStatus`, `renderBulkOperationUserErrors`, `formatBulkOperationCancellationResult(op, statusCommand)` — the status command string is now a parameter instead of hardcoded
- The bulk-operations GraphQL documents + codegen project (moved from `app`; `graphql.config.ts` and both `project.json`s repointed)

Files were moved with `git mv` so history is preserved.

## What stayed in app

The `app bulk` orchestration — auth via `createAdminSessionAsApp`, org/app/store display, the dev-store mutation gate, and the `shopify app bulk status` help text — now calling the shared engine.

## Testing

- Type-check ✅ (cli-kit, app)
- Lint ✅
- Tests ✅ — cli-kit 68, app 87 (all bulk-operation suites)


teddyhwang:
* **#7897** <a href="https://app.graphite.com/github/pr/Shopify/cli/7897?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#7896** <a href="https://app.graphite.com/github/pr/Shopify/cli/7896?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Shopify/cli/7896?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

.changeset/extract-bulk-operations-engine.md
```diff
@@ -0,0 +1,6 @@
+---
+'@shopify/cli-kit': minor
+'@shopify/app': patch
+---
+
+Extract the shared Admin API bulk-operation engine into `@shopify/cli-kit/node/api/bulk-operations` (run query/mutation, stage file, watch/poll, download, fetch, cancel, status/cancellation formatting, version resolution, and id helpers). The `shopify app bulk` commands now consume this shared engine instead of their own copies, with no change to their behavior.
```
isaacroldan: no need for a changeset, this is purely an internal change and changesets are meant for user-facing changes.

teddyhwang: Good call — removed the changeset since this is a purely internal refactor.

packages/cli-kit/src/public/node/api/bulk-operations/helpers.ts
```diff
@@ -0,0 +1,123 @@
...
+ *
+ * @param graphqlOperation - The GraphQL query or mutation string to check.
+ * @returns True if the operation is a mutation, false otherwise.
+ */
+export function isMutation(graphqlOperation: string): boolean {
+  const document = parse(graphqlOperation)
+  const operationDefinition = document.definitions.find((def) => def.kind === 'OperationDefinition')
+
+  return operationDefinition?.operation === 'mutation'
+}
```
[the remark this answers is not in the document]
teddyhwang: Fixed — `isMutation` now wraps `parse()` in the same try/catch as `validateSingleOperation`, throwing an `AbortError` with a friendly "Invalid GraphQL syntax" message instead of leaking a raw `GraphQLError`. Added a test covering invalid syntax.

packages/app/src/cli/services/bulk-operations/execute-bulk-operation.ts
```diff
@@ -13,6 +6,17 @@ import {
   isMutation,
 } from '../graphql/common.js'
```
jordanverasamy, line 7: i'm confused. it looks to me like we now have these methods defined twice: once in `common.js` and once in `helpers.ts`. is that intentional?

packages/cli-kit/src/public/node/api/bulk-operations/format-bulk-operation-status.ts
```diff
@@ -68,7 +88,7 @@ export function formatBulkOperationCancellationResult(
         headline: 'Bulk operation is being cancelled.',
         body: [
           'This may take a few moments. Check the status with:\n',
-          {command: `shopify app bulk status --id=${extractBulkOperationId(operation.id)}`},
+          {command: `${statusCommand} --id=${extractBulkOperationId(operation.id)}`},
```
jordanverasamy: [non-blocking] passing the command string in works, but now the engine has to know about commands, and PR 2 will pass the same string again for `store bulk`.

would it be simpler to let the formatter just return the result, and let each command write its own status line?

packages/cli-kit/src/public/node/api/bulk-operations/helpers.test.ts
```diff
@@ -0,0 +1,103 @@
...
+    vi.mocked(fetchApiVersions).mockResolvedValue([
+      {handle: '2025-07', supported: true},
+      {handle: '2025-10', supported: true},
+      {handle: '2099-01', supported: false},
+    ])
+    const version = await resolveApiVersion({adminSession})
+    expect(version).toBe('2025-10')
+  })
+
+  test('falls back to the minimum default version when newer than supported', async () => {
```
jordanverasamy, line 87: [nit] i think we're missing a case here. we test when all the versions are older than the default, and when the default is newer than the supported one, but not when there's a default *and* a newer supported version. that last one is the case where the sort matters, right? maybe add supported `['2025-10']`, default `'2024-01'`, expect `'2025-10'`.

jordanverasamy (approved):
looks right correctness-wise, no problems. some style/design nitpicks but looks good

packages/app/src/cli/services/bulk-operations/execute-bulk-operation.ts
teddyhwang: Good catch — that was unintentional duplication. `common.ts` now re-exports `isMutation`/`validateSingleOperation`/`resolveApiVersion` from the cli-kit engine instead of redefining them, so there's a single implementation and existing app call sites keep the same import surface.

packages/cli-kit/src/public/node/api/bulk-operations/format-bulk-operation-status.ts
teddyhwang: Agreed — done. The formatter is now command-agnostic: it returns just the classification (headline/renderType/sections), and each command writes its own 'check status' hint. No more passing a command string into the engine, and PR 2 no longer repeats it.

packages/cli-kit/src/public/node/api/bulk-operations/helpers.test.ts
teddyhwang: Added — new case with supported `['2025-10']` + default `'2024-01'` expecting `'2025-10'`, covering the path where the sort decides the winner.

packages/cli-kit/src/public/node/api/bulk-operations/download-bulk-operation-results.ts
```diff
@@ -0,0 +1,41 @@
...
+ * @returns True if any result line reports user errors.
+ */
+export function resultsContainUserErrors(results: string): boolean {
+  const lines = results
+    .trim()
+    .split('\n')
+    .filter((line) => line.trim().length > 0)
+
+  return lines.some((line) => {
+    const parsed = JSON.parse(line)
```
isaacroldan: 🐛 **Bug**: `resultsContainUserErrors` correctly filters blank lines now, but a single malformed JSONL line — truncated download, partial flush, API change — throws a raw `SyntaxError` from `JSON.parse(line)` with no context about which line failed or which operation triggered it. Since this helper now lives in cli-kit, every command importing it inherits that failure mode.

**Suggestion:** Wrap the parse in try/catch. Skipping is the safest default since the caller is just checking for user errors; one corrupt line shouldn't dictate the full result:

```suggestion
  return lines.some((line) => {
    let parsed
    try {
      parsed = JSON.parse(line)
    } catch {
      return false
    }
    if (!parsed.data) return false
    const result = Object.values(parsed.data)[0] as {userErrors?: unknown[]} | undefined
    return result?.userErrors !== undefined && result.userErrors.length > 0
  })
```

packages/cli-kit/src/public/node/api/bulk-operations/helpers.ts
```diff
@@ -0,0 +1,133 @@
...
+  const availableVersions = await fetchApiVersions(adminSession)
+
+  if (!userSpecifiedVersion) {
+    // Return the most recent supported version, or minimumDefaultVersion if specified, whichever is newer.
+    const supportedVersions = availableVersions.filter((version) => version.supported).map((version) => version.handle)
+    if (minimumDefaultVersion) {
+      supportedVersions.push(minimumDefaultVersion)
+    }
+
+    return supportedVersions.sort().reverse()[0]!
```
isaacroldan: 🐛 **Bug**: `return supportedVersions.sort().reverse()[0]!` declares a string return but lies to TypeScript when `supportedVersions` is empty (degraded Admin API response, no `minimumDefaultVersion` passed). Downstream `adminRequestDoc` then receives `version: undefined`, producing a hard-to-debug failure that masquerades as a generic request error.

**Suggestion:** Convert the non-null assertion into an explicit guard:

```suggestion
    const latest = supportedVersions.sort().reverse()[0]
    if (!latest) {
      throw new AbortError('No supported API versions available for this store.')
    }
    return latest
```

packages/cli-kit/src/public/node/api/bulk-operations/watch-bulk-operation.ts
isaacroldan, line 1: (Ref Line 153) 🐛 **Bug**: `new Promise((resolve) => abortSignal.addEventListener('abort', resolve))` attaches a fresh listener to the AbortSignal on every poll but never removes it when `sleep` wins the race. With a default 5s poll interval and no upper bound on duration, a long-running watch accumulates listeners proportional to poll count. After 11 polls the AbortSignal (extends EventTarget) emits Node's `MaxListenersExceededWarning` on stderr, and each leaked closure pins its `resolve` until the operation finishes.

**Suggestion:** Attach the listener with `{once: true}` and remove it after the race resolves so the common case where `sleep` wins doesn't leak:

```suggestion
    if (abortSignal) {
      let removeAbortListener: (() => void) | undefined
      const abortPromise = new Promise<void>((resolve) => {
        const handler = () => resolve()
        abortSignal.addEventListener('abort', handler, {once: true})
        removeAbortListener = () => abortSignal.removeEventListener('abort', handler)
      })
      // eslint-disable-next-line no-await-in-loop
      await Promise.race([sleep(pollInterval), abortPromise])
      removeAbortListener?.()
    } else {
```

isaacroldan (commented):
---

packages/app/src/cli/services/bulk-operations/bulk-operation-status.ts
```diff
@@ -1,45 +1,22 @@
...
-  // Otherwise return as-is (let API handle any errors)
-  return id
-}
-
-export function extractBulkOperationId(gid: string): string {
-  // Extract the numeric ID from a GID like "gid://shopify/BulkOperation/123"
-  const match = gid.match(/^gid:\/\/shopify\/BulkOperation\/(\d+)$/)
-  return match?.[1] ?? gid
-}
+export {normalizeBulkOperationId, extractBulkOperationId}
```
isaacroldan: why do you need to re-export here? you can import directly from `'@shopify/cli-kit/node/api/bulk-operations'` from anywhere in `app`

packages/cli-kit/src/public/node/api/bulk-operations/download-bulk-operation-results.ts
teddyhwang: Fixed — wrapped the per-line `JSON.parse` in try/catch. A malformed line (truncated download, partial flush) is now skipped; non-SyntaxError is rethrown. Added a test for the malformed-line case.

packages/cli-kit/src/public/node/api/bulk-operations/helpers.ts
teddyhwang: Fixed — replaced the non-null assertion with an explicit guard that throws `AbortError('No supported API versions available for this store.')` when the list is empty. Added a test.

packages/cli-kit/src/public/node/api/bulk-operations/watch-bulk-operation.ts
teddyhwang: Fixed — the abort listener is now registered with `{once: true}` and removed after each poll race, so the common case (sleep wins) no longer leaks a listener per poll.

packages/app/src/cli/services/bulk-operations/bulk-operation-status.ts
teddyhwang: Good point — removed the re-export. The app commands now import `normalizeBulkOperationId` directly from `@shopify/cli-kit/node/api/bulk-operations`.

isaacroldan (approved):

PR #7897, opened 2026-06-22
Add shopify store bulk commands

teddyhwang:
## What

Adds `shopify store bulk execute`, `shopify store bulk status`, and `shopify store bulk cancel` — the same bulk-operation workflow as `shopify app bulk`, but running against a store using **stored auth** (`shopify store auth`) instead of requiring an app.

> **Stacked PR 2 of 2**, on top of #7896 (extract shared engine into cli-kit). Review/merge #7896 first.

## How

Consumes the shared `@shopify/cli-kit/node/api/bulk-operations` engine introduced in #7896. The store package adds only its own orchestration:

- `prepareBulkAdminContext` — loads the stored store session into an `AdminSession`
- Store-only info display + `shopify store bulk status` help text
- A `--allow-mutations` opt-in gate (consistent with `shopify store execute`) instead of app's dev-store check — `store bulk` only has a stored session, so it can't infer the store type

New: `commands/store/bulk/{execute,status,cancel}.ts`, `services/store/bulk/*`, `bulkOperationFlags` in `flags.ts`, and command registration in `src/index.ts`.

## Testing

- Type-check ✅ (store)
- Lint ✅
- Tests ✅ — store 24 (services + commands)


teddyhwang:
* **#7897** <a href="https://app.graphite.com/github/pr/Shopify/cli/7897?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Shopify/cli/7897?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#7896** <a href="https://app.graphite.com/github/pr/Shopify/cli/7896?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

packages/store/src/cli/services/store/bulk/execute-bulk-operation.ts
```diff
@@ -0,0 +1,257 @@
...
+  } else {
+    renderWarning({
+      headline: 'Bulk operation not created successfully.',
+      body: 'This is an unexpected error. Please try again later.',
+    })
+    throw new BugError('Bulk operation response returned null with no error message.')
+  }
+}
+
+async function renderBulkOperationResult(operation: BulkOperation, outputFile?: string): Promise<void> {
```
jordanverasamy, line 157: [question] i don't think any test gets here. they all stop at a `CREATED` op, so none of the result rendering runs. worth running a `COMPLETED` (and `FAILED`) op through here?

packages/store/src/cli/services/store/bulk/execute-bulk-operation.ts
```diff
@@ -0,0 +1,257 @@
...
+      renderError({headline, customSections})
+      break
+  }
+}
+
+function resultsContainUserErrors(results: string): boolean {
+  const lines = results.trim().split('\n')
+
+  return lines.some((line) => {
+    const parsed = JSON.parse(line)
```
jordanverasamy: this breaks on an empty result file, i think? i think a completed op that matched nothing can do that

packages/store/src/cli/services/store/bulk/bulk-admin-context.ts
```diff
@@ -0,0 +1,28 @@
...
+ * Loads previously stored store authentication and builds an Admin API session for it.
+ *
+ * Run `shopify store auth` first to create stored auth for the store.
+ *
+ * @param store - The store domain to load stored auth for.
+ * @returns The Admin session and the full stored auth session.
+ */
+export async function prepareBulkAdminContext(
+  store: string,
+): Promise<{adminSession: AdminSession; session: StoredStoreAppSession}> {
```
jordanverasamy: [non-blocking] every caller only uses `adminSession` from this. is `session` used anywhere? if not, could it just return the `AdminSession`?

packages/store/src/cli/services/store/bulk/execute-bulk-operation.ts
teddyhwang: Good call — added a `--watch` result-rendering describe block that runs completed (with results) and failed operations through `renderBulkOperationResult`, asserting the success/error banners and that results get downloaded/output.

packages/store/src/cli/services/store/bulk/execute-bulk-operation.ts
teddyhwang: You're right, an empty result file crashed `JSON.parse`. Fixed by skipping blank lines, and since the same helper was duplicated in app I pulled it into the shared cli-kit engine (`resultsContainUserErrors`) so both are fixed. Added a test for the empty-file case (and a unit test in cli-kit).

packages/store/src/cli/services/store/bulk/bulk-admin-context.ts
teddyhwang: Right — no caller used `session`. `prepareBulkAdminContext` now returns the `AdminSession` directly; updated the callers and tests.

jordanverasamy (approved):

packages/store/src/cli/commands/store/bulk/cancel.ts
```diff
@@ -0,0 +1,37 @@
...
+  static examples = ['<%= config.bin %> <%= command.id %> --store shop.myshopify.com --id 123456789']
+
+  static flags = {
+    ...globalFlags,
+    store: storeFlags.store,
+    id: Flags.string({
+      description: 'The bulk operation ID to cancel (numeric ID or full GID).',
+      env: 'SHOPIFY_FLAG_ID',
+      required: true,
+    }),
```
isaacroldan: This flag is going to be used in more commands (status), let's extract it to `storeFlags` and add a `parse` property to automatically normalize it (like the store flag does)  


```
parse: async (input) => normalizeBulkOperationId(input),
```

packages/store/src/cli/commands/store/bulk/cancel.test.ts
```diff
@@ -0,0 +1,20 @@
...
+  beforeEach(() => {
+    vi.mocked(cancelBulkOperation).mockResolvedValue()
+  })
+
+  test('cancels the operation, normalizing the ID', async () => {
+    await StoreBulkCancel.run(['--store', 'shop.myshopify.com', '--id', '123'])
+
+    expect(cancelBulkOperation).toHaveBeenCalledWith({
+      store: 'shop.myshopify.com',
+      operationId: 'gid://shopify/BulkOperation/123',
```
isaacroldan, line 18: Once you move the logic to normalize to the flag definition itself (see my other comment), i'm not sure if this test has enough value to keep it

packages/store/src/cli/commands/store/bulk/execute.ts
```diff
@@ -0,0 +1,84 @@
...
+          queryFile,
+        )} is empty. Please provide a valid GraphQL query or mutation.`,
+      )
+    }
+    return fileContents
+  }
+
+  throw new BugError(
+    'Query should have been provided via --query or --query-file flags due to exactlyOne constraint. This indicates the oclif flag validation failed.',
+  )
```
isaacroldan: this function is duplicated from `prepareExecuteContext` in `execute-command-helpers.ts`  
Can we extract this to cli-kit as some "bulk query validation"?

packages/store/src/cli/services/store/bulk/bulk-operation-status.ts
```diff
@@ -0,0 +1,159 @@
...
+  normalizeBulkOperationId,
+  extractBulkOperationId,
+  type BulkOperation,
+} from '@shopify/cli-kit/node/api/bulk-operations'
+import {renderInfo, renderSuccess, renderError, renderTable} from '@shopify/cli-kit/node/ui'
+import {outputContent, outputToken, outputNewline} from '@shopify/cli-kit/node/output'
+import {timeAgo, formatDate} from '@shopify/cli-kit/common/string'
+import colors from '@shopify/cli-kit/node/colors'
+
+export {normalizeBulkOperationId}
```
isaacroldan: Why this re-export?

packages/store/src/cli/commands/store/bulk/cancel.ts
teddyhwang: Done — extracted a shared `--id` flag into store `flags.ts` with a `parse` that normalizes to a GID. `status` uses the optional variant, `cancel` the required one, so neither command calls `normalizeBulkOperationId` manually anymore.

(Heads up: the `command-flags-with-env` lint rule crashes on a spread inside a command's `Flags.string({...})`, so the commands reference the fully-defined flags from `flags.ts` rather than spreading shared props inline.)

packages/store/src/cli/commands/store/bulk/cancel.test.ts
teddyhwang: Kept a slimmed-down version: since the flag's `parse` now does the normalization, the test now reads as a wiring check (`--id 123` → service receives the normalized GID), which still guards against the parse being dropped. Happy to remove it entirely if you'd prefer.

packages/store/src/cli/commands/store/bulk/execute.ts
teddyhwang: Done — extracted `resolveBulkOperationQuery` into the cli-kit bulk-operations engine (reads inline/file, validates non-empty + single operation) and used it here. Also refactored app's `prepareExecuteContext` to use it, removing that duplication too.

packages/store/src/cli/services/store/bulk/bulk-operation-status.ts
teddyhwang: Removed — the commands now get `normalizeBulkOperationId` via the flag's `parse`, so the service no longer needs to re-export it.

isaacroldan (approved):

PR #7901, opened 2026-06-23
Group CLI crash reports on structured error signals

### LINKED ISSUE #7891: split observe analytics group for cleaner resiliency tracking

stephanie-shopify:

Fixes #7891. which comes from resiliency issue 

[A single Bugsnag bucket ](https://observe.shopify.io/a/observe/errors/16864652937831232783)(`groupingHash 16864652937831232783`) had become a catch-all for ~1,170 unrelated GraphQL/API errors spanning `app`/`theme`/`store`/`hydrogen`. That makes the bucket un-routable (five owning teams in one group), mixes expected/transient noise with genuine regressions, and manufactures false P1s.

The earlier fix attempt (#7894) addressed the symptom by **regex-reparsing the stringified error message** to reconstruct type information, and removed expected-error suppression entirely (reporting every `AbortError`). This PR is the redesign: it groups on the structured signals the error already carries.

The key insight: in `sendErrorToBugsnag`, only the *reportable* copy is flattened to `new Error(error.message)` — the **original typed error is still in scope** and already carries `statusCode` + `errors[].extensions.code`. So we group on facts, not on a string we just serialized.

- **New `error-grouping.ts`** — derives `{httpStatus, code, errorClass}` from the original error (`GraphQLClientError` and raw graphql-request `ClientError`) and maps them via an explicit decision table:
  - `THROTTLED` / `429` → `rate_limit`
  - `401` → `authentication`
  - `403` / `ACCESS_DENIED` → `permission`
  - `5xx` → `server`
  - else → shared keyword categorizer (fallback only)
  - unknown → **leave `groupingHash` unset** so Bugsnag's stack-trace grouping applies and distinct unknown bugs aren't merged.
- **`error-handler.ts`** — sets `event.groupingHash` only when a real category resolves, and emits `error_grouping` metadata (`slice_name`, `http_status`, `error_code`, `error_class`) so backend dashboards/routing work **regardless of CLI version** (the hash is client-side; old versions keep the old hash until adoption rolls).
- **`error.ts`** — raw `ClientError` with `401`/`429`/`THROTTLED` is now treated as **expected** and kept out of crash reporting.
- The shared `error-categorizer.ts` / `storage.ts` (Monorail analytics taxonomy) are **untouched** — their tests stay green, proving the `error:${category}:${signature}` events are unaffected.
- A short `error-grouping.DESIGN.md` records the problem, constraints, and the three decisions.

### How to test your changes?

```bash
pnpm vitest run packages/cli-kit/src/private/node/analytics/error-grouping.test.ts
pnpm vitest run packages/cli-kit/src/public/node/error-handler.test.ts
pnpm vitest run packages/cli-kit/src/public/node/error.test.ts
pnpm vitest run packages/cli-kit/src/private/node/analytics/   # categorizer + storage stay green
```

Decision-table matrix (asserted in tests): 403 `ACCESS_DENIED` → `theme:permission:http-403-access-denied`; 401 → `*:authentication:http-401`; `THROTTLED` → `*:rate_limit:*` **and** not reported; 5xx → `server`; raw `ClientError`; generic `Error` → `groupingHash` unset (stack grouping).

### Rollout note

Setting `groupingHash` re-buckets **all** CLI errors once — a one-time grouping migration across CLI error dashboards. This is intended and an improvement, but should be flagged to whoever owns the Bugsnag/Observe dashboards before merge. The structured **metadata tags** give backend routing that works across all CLI versions immediately, which is the mitigation while client adoption rolls.

### Follow-ups (out of scope)

- At-source `category`/`code` on the `FatalError` hierarchy + API wrappers (the "proper" version of this).
- Folding status/code grouping into `error-categorizer.ts` (needs a coordinated `storage.ts` update + pinned tests).
- Alert on `*:unknown:*` and `cli:*` bucket growth so the catch-all surfaces itself.
- Tie Bugsnag severity to SLO/error-budget burn rather than aggregate volume.

### Checklist

- [x] I've considered possible cross-platform impacts (Mac, Linux, Windows) — pure TS, no platform-specific paths.
- [x] I've considered possible [documentation](https://shopify.dev) changes — none needed (internal crash-reporting behavior).
- [x] I've considered analytics changes to measure impact — the shared Monorail taxonomy is deliberately untouched; new Bugsnag `error_grouping` metadata is additive.
- [x] The change is user-facing — `patch` bump added via changeset.


packages/cli-kit/src/private/node/analytics/error-grouping.DESIGN.md
mbarak, line 1: should this be included? it has language that is specific to the current issue and not necessarily for long term use.

packages/cli-kit/src/private/node/analytics/error-grouping.ts
```diff
@@ -0,0 +1,179 @@
...
+ *
+ * @param error - The original error (any type).
+ * @param sliceName - The product slice (`app`, `theme`, `store`, `hydrogen`, or `cli`).
+ * @returns The grouping hash, or `undefined` to fall back to stack-trace grouping.
+ */
+export function errorGroupingHash(error: unknown, sliceName: string): string | undefined {
+  const message = errorMessage(error)
+  // Object signals are authoritative; message-derived signals fill gaps (e.g. 5xx errors that the
+  // API layer flattens into an AbortError, dropping the structured status field).
+  const signals: ErrorGroupingSignals = {...signalsFromMessage(message), ...errorGroupingSignals(error)}
```
mbarak: above in `errorGroupingSignals` only assign if there's a value i.e. never assign null or undefined.

---
🐛 **Bug**: The comment says message-derived signals fill gaps, but object spread copies keys even when their value is explicitly `undefined`. `errorGroupingSignals` can assign `httpStatus` or `code` as `undefined`; spreading those values over `signalsFromMessage(message)` can erase a valid status or code recovered from the message fallback. That makes the fallback less reliable for flattened API errors.

**Suggestion:** Consider merging field-by-field with nullish coalescing, or only assigning properties to `signals` when the extracted value is not `undefined`. For example, derive `fromMessage` and `fromError` separately, then build `httpStatus`, `code`, and `errorClass` as `fromError.value ?? fromMessage.value`.

packages/cli-kit/src/private/node/analytics/error-grouping.ts
```diff
@@ -0,0 +1,179 @@
...
+/**
+ * Resolves a semantic category from structured signals using an explicit decision table.
+ *
+ * 403 and ACCESS_DENIED map to `permission` (a forbidden request is a permission problem, not an
+ * authentication one). Returns `undefined` when no structured signal is present.
+ */
+function categoryFromSignals(signals: ErrorGroupingSignals): string | undefined {
+  const {httpStatus, code} = signals
+
+  if (code === 'THROTTLED' || httpStatus === 429) return 'rate_limit'
```
mbarak: 🐛 **Bug**: The structured decision table handles HTTP 429 and `extensions.code === 'THROTTLED'`, but existing retry logic also treats `extensions.code === '429'` as a GraphQL rate-limit signal, even when the HTTP status is 200. A raw `ClientError` with that established shape can currently miss the `rate_limit` bucket.

**Suggestion:** Consider adding `code === '429'` to the rate-limit branch and covering the HTTP-200/code-429 shape with a regression test.

packages/cli-kit/src/private/node/analytics/error-grouping.ts
```diff
@@ -0,0 +1,179 @@
...
+  if (httpStatus === 401) return 'authentication'
```
mbarak, line 144: 💬 **Suggestion**: The documented decision table treats `ACCESS_DENIED` as a permission signal, but the implementation checks `httpStatus === 401` before `code === 'ACCESS_DENIED'`. If an error carries both HTTP 401 and `ACCESS_DENIED`, it will be grouped as `authentication` rather than `permission`.

**Suggestion:** Consider moving the `httpStatus === 403 || code === 'ACCESS_DENIED'` branch above the 401 branch if access-denied should win. If 401 is intentionally higher precedence, add a focused test or documentation note so that routing choice is explicit.

packages/cli-kit/src/private/node/analytics/error-grouping.ts
```diff
@@ -0,0 +1,179 @@
...
+  }
+}
+
+/**
+ * Returns the GraphQL `extensions.code` of the first error, when the errors value is an array.
+ * The API can return `errors` as a string, hence the array guard.
+ */
+function firstExtensionCode(errors: unknown): string | undefined {
+  if (!Array.isArray(errors)) return undefined
+  const code = (errors[0] as {extensions?: {code?: unknown}} | undefined)?.extensions?.code
```
mbarak: 🐛 **Bug**: `firstExtensionCode` only reads `errors[0].extensions.code`, but GraphQL responses can include multiple errors and existing code in this package treats those arrays as multi-entry structures. A later error may carry `THROTTLED`, `429`, or `ACCESS_DENIED`, and App Management errors can carry nested `extensions.app_errors.errors[].category === 'access_denied'`. Missing those shapes can route known API conditions to a fallback or default bucket instead of the intended structured grouping.

**Suggestion:** Consider replacing the first-error lookup with a small normalization helper that scans every GraphQL error, recognizes both `THROTTLED` and `429` as rate-limit signals, recognizes `ACCESS_DENIED` and nested App Management `access_denied` as permission signals, and is shared with the reporting-suppression logic where possible.

packages/cli-kit/src/public/node/error-handler.ts
```diff
@@ -147,11 +156,27 @@ export async function sendErrorToBugsnag(
...
+          // Group on structured signals from the *original* error (not the flattened
+          // `reportableError`), and emit those signals as metadata so backend dashboards and
+          // routing work without depending on the grouping hash or on CLI version adoption.
+          // When no meaningful category resolves we leave `groupingHash` unset so Bugsnag's
+          // stack-trace grouping applies and distinct unknown bugs are not merged.
+          const groupingHash = errorGroupingHash(error, sliceName)
+          if (groupingHash) {
+            event.groupingHash = groupingHash
+          }
+          const {httpStatus, code, errorClass} = errorGroupingSignals(error)
```
mbarak: 🐛 **Bug**: `event.groupingHash` is derived through `errorGroupingHash(error, sliceName)`, which can combine typed error signals, message-derived HTTP/code signals, and fallback keyword categorization. The `error_grouping` metadata immediately below is built from `errorGroupingSignals(error)` alone, so it can omit facts that the hash used, such as a message-derived `http_status` for flattened errors or a fallback category. It also omits the semantic `category` field that the design document says will be emitted, which weakens backend routing that is supposed to depend on stable metadata tags rather than just the hash.

**Suggestion:** Consider exposing a single helper from `error-grouping.ts` that returns the fully resolved grouping result, for example `{hash, category, signals}`, and use that result for both `event.groupingHash` and `event.addMetadata('error_grouping', ...)`. That would also avoid recalculating `errorGroupingSignals(error)` separately for the hash and the metadata.

packages/cli-kit/src/public/node/error.ts
```diff
@@ -204,6 +211,33 @@ export function shouldReportErrorAsUnexpected(error: unknown): boolean {
...
+ *
+ * @param error - Error to be checked.
+ * @returns A boolean indicating if the error is a known-transient API error.
+ */
+function isTransientApiError(error: Error): boolean {
+  if (!(error instanceof ClientError)) {
+    return false
+  }
+  const status = error.response?.status
+  if (status === 401 || status === 429) {
```
mbarak, line 236: why is 401 transient?

packages/cli-kit/src/public/node/error.ts
```diff
@@ -204,6 +211,33 @@ export function shouldReportErrorAsUnexpected(error: unknown): boolean {
...
+    return true
+  }
+  const errors = error.response?.errors
+  if (Array.isArray(errors)) {
+    const code = (errors[0] as {extensions?: {code?: unknown}} | undefined)?.extensions?.code
+    return code === 'THROTTLED'
```
mbarak: 🐛 **Bug**: `isTransientApiError` suppresses HTTP 429 and first-error `THROTTLED`, but it misses the established GraphQL `extensions.code === '429'` rate-limit shape and any transient code that appears after the first error entry. After retries are exhausted, those raw `ClientError`s could still be treated as unexpected and sent to Bugsnag even though the retry layer classifies them as transient rate limiting.

**Suggestion:** Consider scanning all entries in `error.response.errors` and treating both `THROTTLED` and `429` as transient. A regression test like `shouldReportErrorAsUnexpected(clientError(200, '429')) === false` would keep the suppression logic aligned with the retry-layer fixture shape.

mbarak (approved):

packages/cli-kit/src/private/node/analytics/error-grouping.DESIGN.md
stephanie-shopify: will remove

stephanie-shopify: Removed in 8e0dc34 — the design record lives in the PR description and plan, not in the shipped package.

packages/cli-kit/src/private/node/analytics/error-grouping.ts
stephanie-shopify: Real bug, fixed. `errorGroupingSignals` no longer assigns a field at all when the value is absent, and the merge in `resolveErrorGrouping` is now field-by-field with `??` (`fromError.x ?? fromMessage.x`) so an absent object signal can never clobber one recovered from the message. Pinned with a test: a `GraphQLClientError` whose code is only present in the embedded message JSON now keeps that code in the signals.

packages/cli-kit/src/private/node/analytics/error-grouping.ts
stephanie-shopify: Confirmed against `errorsIncludeStatus429` in `private/node/api.ts:266` — `extensions.code === '429'` is the established shape. Added `'429'` to the rate-limit set and a regression test (`clientError(200, '429') -> theme:rate_limit:http-200-429`).

packages/cli-kit/src/private/node/analytics/error-grouping.ts
stephanie-shopify: Deliberate — keeping 401 → `authentication`. HTTP 401 is the authoritative transport signal: a request the server rejected as unauthenticated is an authentication problem even if a GraphQL code also reports access-denied (which in practice pairs with 403, handled by the very next branch). Made the precedence explicit in a `categoryFromSignals` comment and pinned it with a test: a 401 + `ACCESS_DENIED` groups as `theme:authentication:http-401-access-denied` (the code is still recorded in the signature).

packages/cli-kit/src/private/node/analytics/error-grouping.ts
stephanie-shopify: Agreed. Replaced the first-error lookup with a shared scanner that walks every error and recognizes `THROTTLED`/`429` (rate-limit) and `ACCESS_DENIED` plus the nested App Management `extensions.app_errors.errors[].category === 'access_denied'` (permission). It lives in a new dependency-free `graphql-error-codes.ts` so it can be shared with the suppression logic in `error.ts` — `error.ts` can't import `error-grouping.ts` directly without creating an `error.ts → headers.ts → error.ts` cycle. Grouping picks a *routable* code (rate-limit, then permission) so a benign leading error can't mask one further down the array.

packages/cli-kit/src/public/node/error-handler.ts
stephanie-shopify: Good catch — emitting `category` was in the design and I dropped it in the wiring. Refactored to a single `resolveErrorGrouping(error, slice)` that returns `{hash, category, signals}`; the reporter now drives both `event.groupingHash` and the `error_grouping` metadata from that one result, so the metadata reflects message-derived signals and includes `category`. Added a test asserting a 5xx flattened into an `AbortError` surfaces `http_status: 503` and `category: 'server'` in the metadata.

packages/cli-kit/src/public/node/error.ts
stephanie-shopify: Fair — 401 isn't transient in the retry sense. It means the user's session token is expired or invalid: a credential/environment condition we never want as a crash report (called out in #7891), distinct from rate limiting. `transient` was a poor umbrella. Renamed to `isExpectedApiError` and split the doc comment so 401 (credentials) and 429/THROTTLED (rate limit) are explained separately.

packages/cli-kit/src/public/node/error.ts
stephanie-shopify: Fixed via the same shared scanner as the grouping comment — `isExpectedApiError` now calls `hasRateLimitCode(error.response?.errors)`, which scans every error for `THROTTLED`/`429`. Added regression tests: `shouldReportErrorAsUnexpected(clientError(200, '429')) === false`, and a rate-limit code on a non-first error entry is also suppressed.

EvilGenius13 (approved):

packages/cli-kit/src/private/node/analytics/graphql-error-codes.ts
dmerand, line 1: Nit: This helper is not really analytics-specific. It encodes GraphQL/API error semantics used for crash-report suppression and routing.

Recommendation: not required for this PR, but if this grows, move it closer to the API layer, e.g.

```text
  packages/cli-kit/src/private/node/api/graphql-error-codes.ts
```

or

```text
  packages/cli-kit/src/private/node/api/error-codes.ts
```

That would make the dependency direction easier to understand: API error shape → reporting decisions, rather than analytics → public error handling.

packages/cli-kit/src/public/node/error-handler.ts
```diff
@@ -147,11 +156,29 @@ export async function sendErrorToBugsnag(
...
+          // routing work without depending on the grouping hash or on CLI version adoption.
+          // A single resolution drives both the hash and the metadata so they can never disagree:
+          // the metadata carries the resolved category and the same (message-merged) signals.
+          // When no meaningful category resolves we leave `groupingHash` unset so Bugsnag's
+          // stack-trace grouping applies and distinct unknown bugs are not merged.
+          const {hash: groupingHash, category, signals} = resolveErrorGrouping(error, sliceName)
+          if (groupingHash) {
+            event.groupingHash = groupingHash
+          }
+          event.addMetadata('error_grouping', {
```
dmerand, line 175: Nit:  This shape being constructed directly in error-handler is fine for now, but this PR explicitly says backend dashboards/routing should rely on these tags. Once external dashboards depend on them, the metadata shape becomes a contract.

Recommendation: follow-up only. Consider a small typed formatter near `resolveErrorGrouping`, e.g. `errorGroupingMetadata(resolution, sliceName)`, so future changes to hash/category/signal behavior update metadata in one place.

dmerand (approved):

.changeset/structured-error-grouping.md
```diff
@@ -0,0 +1,7 @@
+---
+'@shopify/cli-kit': patch
+---
+
+Group CLI crash reports on structured error signals instead of one catch-all bucket
+
+Bugsnag error grouping now derives a stable `slice:category:signature` hash from the original typed error's HTTP status and GraphQL `extensions.code` (falling back to the shared keyword categorizer, and to Bugsnag's stack-trace grouping for genuinely unknown errors). The same signals are emitted as `error_grouping` metadata so backend routing works regardless of CLI version. Known-transient API errors (HTTP 401/429 and `THROTTLED`) are now treated as expected and kept out of crash reporting.
```
isaacroldan: This is not an user-facing change, please remove the changeset 🙏

packages/cli-kit/src/private/node/analytics/error-grouping.ts
```diff
@@ -0,0 +1,237 @@
...
+    if (code !== undefined) signals.code = code
+    return signals
+  }
+
+  // Raw graphql-request ClientError (handleErrors: false) exposes the full response.
+  if (error instanceof ClientError) {
+    const status = error.response?.status
+    if (status !== undefined) signals.httpStatus = status
+    const code = routableCode(error.response?.errors)
+    if (code !== undefined) signals.code = code
```
isaacroldan, line 61: you don't really need to check for not-undefined here no? assigning an undefined value should be safe

packages/cli-kit/src/private/node/analytics/error-grouping.ts
```diff
@@ -0,0 +1,237 @@
...
+function stripJsonDump(message: string): string {
+  return message.split(': {')[0] ?? message
+}
+
+function errorMessage(error: unknown): string {
+  if (error instanceof Error) return error.message
+  return typeof error === 'string' ? error : ''
+}
+
+function slugify(value: string): string {
```
isaacroldan: an slugify function already exists at `'@shopify/cli-kit/common/string'`, let's reuse that.

packages/cli-kit/src/public/node/error-handler.ts
```diff
@@ -28,6 +29,14 @@ import {realpath} from 'fs/promises'
 // Hardcoded list per product slices to keep analytics consistent.
 const ALLOWED_SLICE_NAMES = new Set<string>(['app', 'theme', 'hydrogen', 'store'])
```
isaacroldan: Please add `organization` to this list, that's a new slice of commands added recently

stephanie-shopify: addressed

packages/cli-kit/src/private/node/analytics/error-grouping.ts
stephanie-shopify: addressed

.changeset/structured-error-grouping.md
stephanie-shopify: addressed

packages/cli-kit/src/private/node/analytics/error-grouping.ts
stephanie-shopify: It would be safe with the current field-by-field ?? merge, but I’m keeping the guard to preserve this helper’s invariant: it only emits fields that are actually present. That avoids reintroducing “present but undefined hammers fallback” class of bug if this object is later spread/merged elsewhere.

PR #7907, opened 2026-06-23
Use crawler signatures

graygilmore:
## Summary

Adds support for Shopify-managed storefront request headers during local theme development in an effort to eliminate legitimate CLI users being flagged as a bot. 

- fetches reusable headers during `theme dev` startup when available
- sends them with storefront render, proxy, session-cookie, and password-check requests
- wires the same behavior into theme app extension previews
- keeps this best-effort so local development continues if the headers cannot be fetched

> [!NOTE]
> Does not work in combination with the `--password` flag.


.changeset/theme-dev-crawler-signatures.md
```diff
@@ -0,0 +1,7 @@
+---
+'@shopify/app': patch
+'@shopify/cli-kit': patch
+'@shopify/theme': patch
+---
+
+Add crawler signature support to theme development storefront requests.
```
isaacroldan: if it's a new feature, it's worth a minor changeset

packages/theme/src/cli/utilities/theme-environment/crawler-signature-types.ts
graygilmore, line 1: We need to define these types ourselves because these fields are hidden and only available to our client so they don't get automatically generated like our other GraphQL queries/mutations do.

packages/theme/src/cli/services/dev.ts
```diff
@@ -81,9 +82,13 @@ export async function dev(options: DevOptions) {
     await ensureListingExists(options.directory, options.listing)
   }
 
-  const storefrontPasswordPromise = await isStorefrontPasswordProtected(options.adminSession).then((needsPassword) =>
-    needsPassword ? ensureValidPassword(options.storePassword, options.adminSession.storeFqdn) : undefined,
-  )
+  const [crawlerSignatureHeaders, isPasswordProtected] = await Promise.all([
```
graygilmore, line 85: Run these both at the same time since one doesn't block the other.

packages/theme/src/cli/utilities/theme-environment/proxy.ts
```diff
@@ -341,6 +341,7 @@ export function proxyStorefrontRequest(event: H3Event, ctx: DevServerContext): P
     headers = cleanHeader({
       ...headers,
       ...defaultHeaders(),
+      ...(host === ctx.session.storeFqdn ? ctx.session.crawlerSignatureHeaders : {}),
```
graygilmore, line 344: This proxy handles a few kinds of requests so we only want to add the headers when the hosts match.

EvilGenius13 (approved):
Ran the latest snap build confirmed everything is still working and can find the necessary info in our logs. 🚀 Thanks for the PR @graygilmore

isaacroldan (approved):

PR #7914, opened 2026-06-24
Render store info details as a data table with link footnotes

amcaplan:

The `store info` command rendered the Store details section as a plain
list inside a bordered `renderInfo` banner. The URL fields (Admin URL,
Access URL, Save URL) were plain strings, so long URLs wrapped inside the
box and interleaved with the border characters (`│`), making the output
unreadable.

Two related changes to `packages/store/src/cli/services/store/info/result.ts`:

1. **URLs as link tokens** — Admin/Access/Save URLs now render as link
   tokens. Because the banner provides a `LinksContext`, each URL shows a
   compact `Link [n]` reference inside the box and the full URL prints as a
   footnote outside the box, where it can wrap freely without breaking the
   border.
2. **Two-column data table** — The Store details section now renders via the
   `TabularData` component as a two-column `[label, value]` table
   (`firstColumnSubdued`) instead of a flat list.

Tests updated in `result.test.ts` to assert the new row-based structure.

### How to test your changes?

1. Run `shopify store info` against a store (or the relevant dev flow).
2. Confirm Store details render as a two-column table.
3. Confirm the URL rows show `Link [n]` inside the box and the full URLs
   appear as footnotes below the box — no wrapping into the border.

### Checklist

- [x] The change is user-facing — added a `patch` changeset for `@shopify/store`

### DIFF
--- a/.changeset/store-info-data-table.md
+++ b/.changeset/store-info-data-table.md
@@ -1,0 +1,5 @@
+---
+'@shopify/store': patch
+---
+
+Render `store info` details as a two-column data table and display store URLs as link footnotes so long URLs no longer wrap inside the bordered box.
--- a/.changeset/store-info-data-table.md
+++ b/.changeset/store-info-data-table.md
@@ -2,4 +2,4 @@
 '@shopify/store': patch
 ---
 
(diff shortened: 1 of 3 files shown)


alfonso-noriega (approved):
Great update. Looking forward to see this live!

tizmagik (approved):

PR #7915, opened 2026-06-24
Fix long URL wrapping in banner/alert footnotes

amcaplan:

When a terminal does **not** support hyperlinks (OSC 8), the `Banner`/`Alert` footnote system falls back to rendering the raw URL in a plain Ink `<Text>`. Ink hard-wraps that text to the terminal width, **inserting real newline characters in the middle of the URL**. This makes long links impossible to copy-paste and breaks clickability.

Example before:

```
[1] https://shopify.com/some/lon
g/url/maybe-with-a-query?string=
avalue
```

After:

```
[1] https://shopify.com/some/long/url/maybe-with-a-query?string=avalue
```

### What is this PR doing?

Renders each footnote line inside a `Box` whose width equals the line length (`[id] url`). Ink then emits the full line on a single row instead of hard-wrapping it; the terminal soft-wraps the overflow at display time **without** mutating the underlying text, so the URL stays a single contiguous, copy-pasteable string.

- `packages/cli-kit/src/private/node/ui/components/Banner.tsx` — size each footnote line's `Box` to its content length.
- `packages/cli-kit/src/private/node/ui/components/Alert.test.tsx` — strengthen the existing footnote regression test to assert the URL appears as a single contiguous `[1] <url>` line (it previously tolerated the wrapped version).

The hyperlink-supported path (`ansiEscapes.link`) is unchanged.

### How to test your changes?

In a terminal **without** hyperlink support, run a command that emits a long URL footnote (e.g. `store info`). The footnote URL should appear on one logical line, soft-wrapped by the terminal, and be copy-pasteable as a single string.

Automated: `Alert.test.tsx` + `Banner.test.tsx` (9 tests passing). lint + type-check clean.

### DIFF
--- a/.changeset/fix-footnote-long-url-wrapping.md
+++ b/.changeset/fix-footnote-long-url-wrapping.md
@@ -1,0 +1,5 @@
+---
+'@shopify/cli-kit': patch
+---
+
+Render banner/alert footnote URLs on a single contiguous line so long links are no longer broken by hard newlines and remain copy-pasteable when terminal hyperlinks are unsupported
--- a/packages/cli-kit/src/private/node/ui/components/Alert.test.tsx
+++ b/packages/cli-kit/src/private/node/ui/components/Alert.test.tsx
@@ -165,14 +165,14 @@ describe('Alert', async () => {
     })
 
     // The footnote block (rendered after the closing `╰`) must list the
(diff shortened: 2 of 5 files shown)


alfonso-noriega (approved):
LGTM, nice catch

PR #7936, opened 2026-06-26
Fix e2e cleanup

isaacroldan:
### Why

The e2e tests create temporary Shopify apps in the Dev Dashboard. Each app name includes the GitHub Actions run token, for example `E2E-scaf-rcyjrwc3a1-mqu8cgqh`, so cleanup can find everything created by one run.

A previous run left apps behind even though the cleanup step ran. The per-test delete flow failed on a Dev Dashboard error page, then the fallback cleanup opened the apps list and saw `0` app cards. Because the script treated that as a real empty result, it exited successfully instead of retrying or failing.

### What

This makes the fallback cleanup more reliable:

- It opens the Dev Dashboard apps page with `search_term=<run token>`, so the page only returns apps from the current e2e run instead of scanning every app in the org.
- It still checks the returned cards locally before deleting anything.
- It accepts the dashboard's real empty-search state (`No apps matched your search.`) as a valid zero-result cleanup.
- It rejects broken dashboard states, such as error pages or pages with no app cards and no empty-search message, and retries before failing.

### Testing

- `pnpm --filter e2e exec tsx scripts/cleanup-apps.ts --list --pattern rcyjrwc3a1`
- `pnpm --filter e2e exec tsx scripts/cleanup-apps.ts --list --pattern rcyjrwc3a1ss`
- `pnpm --filter e2e type-check`
- `git diff --check`

### DIFF
--- a/packages/cli-kit/src/cli/api/graphql/admin/generated/types.d.ts
+++ b/packages/cli-kit/src/cli/api/graphql/admin/generated/types.d.ts
@@ -232,6 +232,8 @@ export type OnlineStoreThemeFilesUserErrorsCode =
   | 'LESS_THAN_OR_EQUAL_TO'
   /** The record with the ID used as the input value couldn't be found. */
   | 'NOT_FOUND'
+  /** Theme contextualization and condition types are not compatible with each other. */
+  | 'THEME_CONTEXTUALIZATION_NOT_COMPATIBLE_WITH_CONDITION_TYPES'
   /** There are theme files with conflicts. */
   | 'THEME_FILES_CONFLICT'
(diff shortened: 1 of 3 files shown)


isaacroldan:
* **#7936** <a href="https://app.graphite.com/github/pr/Shopify/cli/7936?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Shopify/cli/7936?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

nelsonwittwer (approved):

PR #7963, opened 2026-06-30
Validate GCS bundle upload status in uploadToGCS

amcaplan:
## What

Hardens `uploadToGCS` in `packages/app/src/cli/services/bundle.ts` so a failed GCS bundle upload can no longer be silently treated as a success.

Previously, `uploadToGCS` called `fetch` with the signed PUT URL but **ignored the response**. Since `fetch` does not throw on a non-2xx HTTP response, an expired signature, malformed request, or transient GCS 5xx left the function returning normally — and the caller then handed `devSessionCreate`/deploy an `assetsUrl` for an object that was never stored.

## Why

This is the client-side root cause of the recurring `RuntimeError: Uploaded file not found` (shop/issues#809).

The server-side existence check (`assets.exist?(path)`) resolves to a GCS **GET-by-name** (`bucket.file(name)`, `skip_lookup: false`), which has **strong** read-after-write consistency. A `false` therefore means the object genuinely isn't there — not list/eventual-consistency lag. The earlier server-side retry fix (#233268) couldn't help because the object was never durably stored in the first place: the CLI swallowed the failed upload.

## Changes

- `uploadToGCS` now checks the PUT response status:
  - `2xx` → returns.
  - Retryable transient status (408/429/500/502/503/504) → retries up to 3 attempts with exponential backoff.
  - Otherwise → throws an `AbortError` with the HTTP status and GCS's response body.
- Dropped the misleading `formData`/`my_upload` multipart header construction. The body was always the raw buffer, and a signed PUT only signs the `host` header — the multipart envelope was never honored.
- Added tests: non-2xx throw, transient-retry-then-succeed, give-up-after-retries; updated the existing success test to return `{ok, status}`.
- Changeset (`@shopify/app`, patch).

## Testing

- `pnpm vitest run packages/app/src/cli/services/bundle.test.ts` → 13/13 pass.
- ESLint clean on changed files.

🤖 Diagnosis by River; implementation reviewed and landed by Ariel.

Closes shop/issues#809

### DIFF
--- a/packages/app/src/cli/services/bundle.ts
+++ b/packages/app/src/cli/services/bundle.ts
@@ -4,14 +4,20 @@ import {AssetUrlSchema, DeveloperPlatformClient} from '../utilities/developer-pl
 import {MinimalAppIdentifiers} from '../models/organization.js'
 import {joinPath} from '@shopify/cli-kit/node/path'
 import {brotliCompress, zip} from '@shopify/cli-kit/node/archiver'
-import {formData, fetch} from '@shopify/cli-kit/node/http'
+import {fetch, Response} from '@shopify/cli-kit/node/http'
 import {fileSize, readFileSync} from '@shopify/cli-kit/node/fs'
 import {AbortError} from '@shopify/cli-kit/node/error'
+import {sleep} from '@shopify/cli-kit/node/system'
 import {writeFile} from 'fs/promises'
 
 const MEGABYTE = 1024 * 1024
 const MAX_BUNDLE_SIZE_MB = 100
 const MAX_BUNDLE_SIZE_BYTES = MAX_BUNDLE_SIZE_MB * MEGABYTE
+
+// A signed PUT upload is idempotent (same object key, same bytes), so it is safe
+// to retry. We only retry statuses that are plausibly transient.
+const UPLOAD_MAX_ATTEMPTS = 3
+const RETRYABLE_UPLOAD_STATUS_CODES = new Set([408, 429, 500, 502, 503, 504])
 
 export async function writeManifestToBundle(appManifest: AppManifest, bundlePath: string) {
   const manifestPath = joinPath(bundlePath, 'manifest.json')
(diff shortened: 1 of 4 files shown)


packages/app/src/cli/services/bundle.ts
```diff
@@ -44,10 +59,28 @@ export async function uploadToGCS(signedURL: string, filePath: string) {
...
+  let response: Response | undefined
+  for (let attempt = 1; attempt <= UPLOAD_MAX_ATTEMPTS; attempt++) {
+    // The signed URL only signs the `host` header, so no extra headers are
+    // required; node-fetch derives Content-Length from the buffer body.
+    // eslint-disable-next-line no-await-in-loop
+    response = await fetch(signedURL, {method: 'put', body: buffer}, 'slow-request')
+    if (response.ok) return
+    if (!RETRYABLE_UPLOAD_STATUS_CODES.has(response.status) || attempt === UPLOAD_MAX_ATTEMPTS) break
+    // eslint-disable-next-line no-await-in-loop
+    await sleep(2 ** attempt * 0.1)
```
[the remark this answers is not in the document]
amcaplan: Good catch — addressed in da3d2e5389. The retry path now drains the response body with `await response.text()` before sleeping/retrying, so node-fetch can release/reuse the socket. On the final (non-retried) attempt the body is read once for the error message.

packages/app/src/cli/services/bundle.ts
```diff
@@ -44,10 +59,38 @@ export async function uploadToGCS(signedURL: string, filePath: string) {
...
+    const lastAttempt = attempt === UPLOAD_MAX_ATTEMPTS
+    const retryable = RETRYABLE_UPLOAD_STATUS_CODES.has(response.status)
+    // node-fetch keeps the socket open until the body is consumed. On the final
+    // attempt we read it below for the error message; otherwise drain it here so
+    // the connection can be reused or released before the next attempt.
+    if (retryable && !lastAttempt) {
+      // eslint-disable-next-line no-await-in-loop
+      await response.text().catch(() => {})
+      // eslint-disable-next-line no-await-in-loop
+      await sleep(2 ** attempt * 0.1)
```
gonzaloriestra, line 81: Will this small sleep help when there is a 429 (too many requests)? Maybe is better to just fail in that case?

gonzaloriestra (approved):

packages/app/src/cli/services/bundle.ts
amcaplan: Time will be a total of 2.4 sec + 3 request roundtrips. I'd rather not have special case handling for 429s, let's wait and see if it's necessary.

PR #7986, opened 2026-07-02
Recover on `store info` after a preview store has been claimed

amcaplan:
## Problem

From [this Slack thread](https://shopify.slack.com/archives/C0B242KDQPN/p1783012223457189): running `shopify store info` (and, it turns out, `shopify store execute`) against a preview store *after* it has been claimed through the browser claim flow throws an unhelpful, non-actionable error instead of pointing the user at `shopify store auth`.

## Root cause

The CLI has no local signal for when a preview store gets claimed — the stored session keeps reporting `kind: 'preview'` forever. `store info` unconditionally takes the preview-session branch for such stores and calls the preview-stores REST service with the (now stale) preview admin token. Once claimed, that service starts rejecting the request, but `getPreviewStore`'s error handling only produced a generic message string with no HTTP status exposed and no recovery guidance.

Verified end-to-end against real stores: created preview stores, claimed them through the actual browser flow, and confirmed the preview-stores service really does return `401 {"error_code":"unauthorized"}` afterward for the cached token — and that the real Admin API (used by `store execute` and `store info`'s Admin fallback) does the same for the same stale token.

## Fix

**Detecting the claim (`store info`)**
- `getPreviewStore` now throws a typed `PreviewStoreRequestError` extending `AbortError` (so an uncaught instance still surfaces as a clean user-facing abort, not an unexpected-bug report) and carrying the HTTP status, so callers can classify the failure instead of only seeing a pre-rendered message string.
- `getStoreInfo`'s preview-session path treats a 401/404 from the preview-stores service as a signal the store has likely been claimed.

**Three distinct, consistent recovery messages** (`recovery.ts`), used everywhere a stored session is found invalid — `store info`'s preview branch, `store info`'s Admin fallback, and `store execute` (both the query/mutation path and the API-version-discovery path):
1. **No stored auth at all** → `"No stored app authentication found for {store}."` — Next steps: `Run ... to authenticate`.
2. **Stored auth exists but is invalid** (standard session) → `"Stored app authentication for {store} is no longer valid."` — Next steps show the **real, previously-granted scopes** and say `to re-authenticate`.
3. **Stored auth exists but is invalid, for a preview store** → `"The preview store {store} has likely been claimed, so its stored authentication is no longer valid."` — Next steps show a `<comma-separated-scopes>` placeholder instead of the real scopes, and say `to re-authenticate`.

Case 3's placeholder matters because preview stores are preapproved for a large, fixed scope catalog (~35 scopes) that nobody deliberately chose — suggesting the user re-request all of them against what's now a live, claimed store encourages over-scoping. Case 2 still shows the real scopes since those *were* a deliberate choice (`store auth --scopes ...`) and echoing them back is a genuine convenience.

Originally case 3 was only reachable via `store info`'s own preview-branch check; `store execute` and `store info`'s Admin fallback shared the same underlying invalid-session detection (`throwIfStoredStoreAuthIsInvalid`) but always fell through to case 2's generic message and full scope list, even for a preview session. Consolidated the message/scope selection into `throwStoredAuthInvalidError(session)` / `reauthScopesFor(session)` in `recovery.ts` so every call site is automatically consistent.

**Deliberately does not clear a lingering preview session on 401/404** (in `throwIfStoredStoreAuthIsInvalid` and `store info`'s own check). I initially had this clearing the session (mirroring how standard-session invalidation is handled), but testing against a real claimed store — and later against a real store hit by `store execute` — showed that's the wrong call for preview sessions:
- `store auth` overwrites the session bucket's `currentUserId` when it stores a fresh session, so nothing depends on the old preview entry being removed first.
- Preview sessions have no refresh-token cycle, so the "avoid repeatedly retrying with a dead refresh token" justification for clearing standard sessions doesn't apply.
- Clearing it caused a real regression twice: a second `store info` run, or a `store execute` run followed by `store info`, would silently fall through to a full interactive Business Platform login instead of repeating the actionable `store auth` guidance. Leaving the (already server-invalidated) session in place keeps every retry consistent. Standard sessions are still cleared as before — this only changes preview-session behavior.

**Cleanup**
- Removed a redundant "Next steps" rendering bug in the shared recovery errors (a separate `tryMessage` like "To re-authenticate, run:" duplicated with the "Next steps" heading); folded into a single bullet: `Run \`shopify store auth ...\` to re-authenticate`.
- Dropped now-dead parameters/exports surfaced along the way (`storeAuthCommandNextSteps`'s unused `scopes` param, an unused `UNKNOWN_SCOPES_PLACEHOLDER` export flagged by `knip`).

## Testing

- Added `PreviewStoreRequestError` coverage in `client.test.ts` (401 and 404, including that it's an `AbortError`).
- Added a dedicated `recovery.test.ts` directly covering `recovery.ts`'s branching logic (scope placeholder selection, message selection, "to authenticate" vs "to re-authenticate"), which had previously only been exercised indirectly through other modules' tests.
- Added `getStoreInfo` coverage for the claimed-preview-store scenario, both via the dedicated preview-stores lookup and via a lingering preview session hitting the Admin fallback.
- Added `store execute` coverage (`admin-transport.test.ts`, both `runAdminStoreGraphQLOperation` and `fetchPublicApiVersions`) for the same lingering-preview-session scenario, with paired tests showing the contrast against a standard invalid session (real scopes shown) right next to the preview case (placeholder shown).
- Added coverage for the "no stored auth" case saying "to authenticate" rather than "to re-authenticate".
- Manually verified live against several real preview stores (created via `shopify store create preview`, claimed via the real browser claim flow and via a corrupted-token shortcut for faster iteration):
  - `store info` and `store execute` now produce identical, consistent messaging for the same claimed store.
  - Running either command repeatedly (or one after the other) no longer degrades into a full interactive login prompt.
  - An unseen store domain correctly says "to authenticate"; a claimed preview store correctly says "to re-authenticate" and omits the full scope list.
- Full `packages/store` unit test suite (358 tests), type-check, lint, and `knip` all pass clean.

## Manual QA / tophatting

1. **No stored auth** — run against a store domain you've never used with this CLI install:
   ```
   shopify store execute --store <unseen-store>.myshopify.com --query "query { shop { name } }"
   ```
   Expect: `No stored app authentication found for ...` and Next steps saying `to authenticate` (not "re-authenticate").

2. **Invalid standard auth** — run `shopify store auth --store <store> --scopes read_products,write_orders` against a real store, then corrupt the stored `accessToken` in `~/Library/Preferences/shopify-cli-store-nodejs/config.json` (or just wait for real expiry), then run `store execute`/`store info` again.
   Expect: `Stored app authentication for ... is no longer valid.`, Next steps show the real scopes (`--scopes read_products,write_orders`), session is cleared afterward.

3. **Claimed preview store**:
   ```
   shopify store create preview
   shopify store info --store <domain> --json   # copy saveUrl, open it, complete the claim in browser
   shopify store info --store <domain>
   shopify store execute --store <domain> --query "query { shop { name } }"
   ```
   Expect both commands: `The preview store ... has likely been claimed, so its stored authentication is no longer valid.`, Next steps show `--scopes <comma-separated-scopes>` (not the full preview grant list). Run either command again — the message should stay identical, not fall back to a browser login prompt. `shopify store auth list` should still show the store until you actually re-authenticate.

## Changeset

Added — this is a user-facing bug fix (`store info` and `store execute` error behavior).

### DIFF
--- a/packages/store/src/cli/services/store/auth/recovery.ts
+++ b/packages/store/src/cli/services/store/auth/recovery.ts
@@ -8,16 +8,24 @@ function storeAuthCommandNextSteps(store: string, scopes: string) {
   return [[storeAuthCommand(store, scopes)]]
 }
 
+// Folds the instruction into the "Next steps" bullet itself (`Run ... to <purpose>`) instead of
+// repeating it in a separate `tryMessage` line. Passing both a `tryMessage` like "To re-authenticate,
+// run:" and a `nextSteps` list renders as a stuttering "To re-authenticate, run:\nNext steps\n  •
+// shopify store auth ..." — the "Next steps" heading already says what the list is for.
+function storeAuthCommandNextStepsWithPurpose(store: string, scopes: string, purpose: string) {
+  return [['Run', storeAuthCommand(store, scopes), purpose]]
+}
+
 export function throwStoredStoreAuthError(store: string): never {
   throw new AbortError(
     `No stored app authentication found for ${store}.`,
-    'To create stored auth for this store, run:',
-    storeAuthCommandNextSteps(store, '<comma-separated-scopes>'),
+    undefined,
+    storeAuthCommandNextStepsWithPurpose(store, '<comma-separated-scopes>', 'to create stored auth for this store'),
   )
 }
 
 export function throwReauthenticateStoreAuthError(message: string, store: string, scopes: string): never {
-  throw new AbortError(message, 'To re-authenticate, run:', storeAuthCommandNextSteps(store, scopes))
+  throw new AbortError(message, undefined, storeAuthCommandNextStepsWithPurpose(store, scopes, 'to re-authenticate'))
 }
 
 export function retryStoreAuthWithPermanentDomainError(returnedStore: string): AbortError {
--- a/packages/store/src/cli/services/store/auth/recovery.ts
+++ b/packages/store/src/cli/services/store/auth/recovery.ts
@@ -38,6 +38,20 @@ export function throwReauthenticateStoreAuthError(message: string, session: Stor
   )
 }
 
+// A preview store's local session has no way to know it was claimed through the browser claim
+// flow; a 401/404 the first time the stale session is used again is the only signal. Surfacing
+// that possibility is more useful than the generic "no longer valid" message a standard session
+// gets, so every call site that detects an invalid stored session (regardless of which API it
+// hit) should go through here instead of writing its own message.
+export function throwStoredAuthInvalidError(session: StoredStoreAppSession): never {
+  const message =
+    session.kind === 'preview'
+      ? `The preview store ${session.store} has likely been claimed, so its stored authentication is no longer valid.`
+      : `Stored app authentication for ${session.store} is no longer valid.`
+
+  throwReauthenticateStoreAuthError(message, session)
+}
+
 export function retryStoreAuthWithPermanentDomainError(returnedStore: string): AbortError {
   // eslint-disable-next-line @shopify/cli/no-error-factory-functions
   return new AbortError(
--- a/packages/store/src/cli/services/store/auth/recovery.ts
+++ b/packages/store/src/cli/services/store/auth/recovery.ts
@@ -8,24 +8,20 @@ function storeAuthCommandNextSteps(store: string, scopes: string) {
   return [[storeAuthCommand(store, scopes)]]
 }
 
-// Folds the instruction into the "Next steps" bullet itself (`Run ... to <purpose>`) instead of
-// repeating it in a separate `tryMessage` line. Passing both a `tryMessage` like "To re-authenticate,
-// run:" and a `nextSteps` list renders as a stuttering "To re-authenticate, run:\nNext steps\n  •
-// shopify store auth ..." — the "Next steps" heading already says what the list is for.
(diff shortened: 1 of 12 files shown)


packages/store/src/cli/services/store/create/preview/client.ts
```diff
@@ -82,6 +82,23 @@ interface RawPreviewStoreErrorResponse {
...
+  public readonly status: number
+  public readonly tryMessage?: string
+
+  constructor(status: number, message: string, tryMessage?: string) {
+    super(message)
+    this.name = 'PreviewStoreRequestError'
+    this.status = status
+    this.tryMessage = tryMessage
+  }
+}
```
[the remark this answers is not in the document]
amcaplan: Resolved in a later commit, it now extends `AbortError`

packages/store/src/cli/services/store/auth/recovery.ts
isaacroldan, line 1: this file doesn't have tests... should it? it has more logic now

isaacroldan (approved):
Left a comment about tests, code looks good

packages/store/src/cli/services/store/auth/recovery.ts
amcaplan: Good call, I missed this! Filled 'em in.

dmerand (approved):

PR #7993, opened 2026-07-04
[Refactor] Modernize archiver utilities with async/await

gonzaloriestra:
This PR refactors the archiving utilities in `@shopify/cli-kit` to follow modern asynchronous patterns.

### Why:
The previous implementation used manual `Promise` wrappers and `.then()` chains which were more verbose and harder to reason about. Moving to `async/await` makes the control flow clearer, especially when coordinating multiple asynchronous operations like globbing, file reading, and stream finalization.

### Changes:
- **`zip`**: Now uses `async/await` for globbing and file addition. It waits for `archive.finalize()` and the output stream's `close` event before resolving.
- **`brotliCompress`**: Similarly modernized to use `async/await` for the internal tar archive creation step.
- **Documentation**: Added `@returns` to `zip` JSDoc.

### How to test your changes?
CI. (Integration tests in `packages/cli-kit/src/public/node/archiver.integration.test.ts` cover these changes).

---
*PR created automatically by Jules for task [2020363541973881611](https://jules.google.com/task/2020363541973881611) started by @gonzaloriestra*

### DIFF
--- a/packages/cli-kit/src/public/node/archiver.ts
+++ b/packages/cli-kit/src/public/node/archiver.ts
@@ -31,6 +31,7 @@ interface ZipOptions {
  * Windows.
  *
  * @param options - ZipOptions.
+ * @returns A promise that resolves when the directory has been zipped.
  */
 export async function zip(options: ZipOptions): Promise<void> {
   const {inputDirectory, outputZipPath, matchFilePattern = '**/*'} = options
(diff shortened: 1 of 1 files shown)


alexanderMontague (commented):
`archivePromise` is not observed until after `archive.finalize()` resolves. An archive error can reject both promises; when the `finalize()` await throws first, the separately rejected `archivePromise` is left unobserved and can surface as an unhandled rejection. The same ordering appears in `brotliCompress`. Please attach rejection handlers to both promises before waiting—for example, await `Promise.all([archive.finalize(), archivePromise])`—and add an archive-error regression test.

---
_Generated with Pi. Vetted by Me._

PR #6839, opened 2026-02-10
Fix 401 Unauthorized on cart AJAX endpoints during `shopify theme dev`

kdaviduik:
### WHY are these changes introduced?

Closes https://github.com/Shopify/developer-tools-team/issues/1034.

Cart AJAX endpoints (`/cart/add.js`, `/cart/update.js`, `/cart/change.js`, `/cart.js`, `/cart.json`) return **401 Unauthorized** when running `shopify theme dev` on CLI 3.89.0, but work correctly on 3.88.0. 5+ developers affected with confirmed reproduction.

### WHAT is this pull request doing?

Conditionally omits the `Authorization: Bearer <storefrontToken>` header on proxied requests to endpoints that use session-cookie auth (cart, checkout, account). CDN/asset paths continue receiving the Bearer token for theme preview rendering.

**Changes:**

| File | Change |
|------|--------|
| `proxy.ts` | Added `CART_SESSION_PATTERN` constant, `needsBearerToken()` helper, updated Authorization conditional |
| `proxy.test.ts` | Added 15 regression tests for Bearer token header behavior |
| `.changeset/fix-cart-ajax-401.md` | Patch bump for `@shopify/theme` |

**Key design decisions:**

1. **Denylist over allowlist**: The set of cookie-auth endpoints is small and stable (cart, checkout, account). The set of Bearer-needing endpoints is large and open-ended (CDN assets, vanity CDN rewrites, `.js.liquid` files, etc.). An allowlist would risk silent rendering failures; a denylist's failure mode is a loud 401 — easy to detect.

2. **Separate `CART_SESSION_PATTERN` from `CART_PATTERN`**: `CART_PATTERN` (`/^\/cart\//`) is for *routing* — it deliberately excludes `GET /cart` (the cart HTML page, which should render locally). `CART_SESSION_PATTERN` (`/^\/cart(\/|\.js|\.json|$)/`) is for *auth* — broader because ALL cart paths use session-cookie auth. Collapsing these would contaminate routing with auth concerns.

3. **Reuse `CHECKOUT_PATTERN` and `ACCOUNT_PATTERN`**: Their routing and auth scopes currently align. The comment in `needsBearerToken()` warns future editors that changes to those patterns should consider auth implications.

4. **Query string stripping**: `needsBearerToken()` strips query strings before regex matching to ensure `ACCOUNT_PATTERN`'s `$` anchor matches correctly (e.g., `/account?foo=bar`).

### How to test your changes?

**Automated:**
```bash
[...]

PR #8027, opened 2026-07-07
fix(theme): don't send Bearer token to cart/checkout/account proxy en…

### LINKED ISSUE #7568: POST /cart/add returns 401 → 403 with Clear-Site-Data header via local dev proxy
## Bug Description

When running `shopify theme dev`, POST requests to `/cart/add` fail with:
1. **401 Unauthorized** (empty body)
2. **403 Forbidden** (HTML response: `<!DOCTYPE ...`)

The response to `/cart/add` includes an unusual `Clear-Site-Data` header, which appears to be terminating the session and causing the 401 → 403 cascade.

GET requests work normally. The cart works correctly when accessing the store via the preview URL (`?preview_theme_id=`).

## Environment
- CLI version: 3.94.3
- Store: emaus-7331.myshopify.com
- Development theme ID: 181626339439
- OS: macOS

## Steps to Reproduce
1. Run `shopify theme dev --store emaus-7331.myshopify.com`
2. Open the local preview at `127.0.0.1`
3. Try to add a product to the cart
4. Observe POST /cart/add returning 401 (empty body) then 403 (HTML)

## Already Tried
- Updated CLI from 3.92.1 to 3.94.3
- `shopify auth logout` + `shopify auth login` + restarted dev server
- Cleared all CLI cache at `~/Library/Preferences/shopify-cli-*`
- Disabled store password protection
- Problem persists in all cases

## Expected Behavior
POST /cart/add should return a JSON response with cart data.

## Actual Behavior
Returns 401 (empty body) then 403 (HTML with `<!DOCTYPE`) with `Clear-Site-Data` header on the response.

gotesgan:

Fixes #7568.

When running `shopify theme dev`, Cart AJAX endpoints (such as `/cart/add.js`, `/cart.js`, `/cart.json`, `/cart/change.js`, and `/cart/update.js`) return `401` or `403` responses.

The proxy currently attaches an `Authorization: Bearer <storefrontToken>` header to every request. However, Cart AJAX endpoints authenticate using the storefront session cookie rather than a Storefront API access token. Because the Bearer token is present, SFR attempts token-based authentication instead of cookie-based authentication. The CLI's storefront token does not include the required cart scopes, causing these requests to fail.

This behavior was previously fixed in commit `d0e135b8f`, but was later reverted in `11240b342`. This PR restores the same behavior while adapting the implementation to the current codebase.

* Adds a `CART_SESSION_PATTERN` regex to match cart endpoints (`/cart`, `/cart.js`, `/cart.json`, and `/cart/*`).
* Introduces a `needsBearerToken()` helper that determines whether a request should include the `Authorization` header.
* Skips sending the Bearer token for:

  * Cart endpoints
  * Checkout endpoints
  * Customer account endpoints
* Continues sending the Bearer token for all other requests that require Storefront API authentication.
* Adds 9 new unit tests covering authentication behavior for:

  * Cart endpoints
  * Checkout endpoints
  * Customer account endpoints
  * CDN and asset requests
  * Theme extension requests

### How to test your changes?

Run the proxy test suite:

```bash
pnpm vitest run packages/theme/src/cli/utilities/theme-environment/proxy.test.ts
```

Expected result:

* All 47 tests pass.
* Cart AJAX requests no longer return `401`/`403` while running `shopify theme dev`.
* Non-cart requests continue to include the Bearer token and behave as before.

### Post-release steps

None.

### Checklist

* [x] I've considered possible cross-platform impacts (Mac, Linux, Windows)
* [x] I've considered possible documentation changes
* [x] I've considered analytics changes to measure impact
* [x] The change is user-facing — this is a bug fix, so a **patch** changeset has been added.

### DIFF
--- a/packages/theme/src/cli/utilities/theme-environment/proxy.test.ts
+++ b/packages/theme/src/cli/utilities/theme-environment/proxy.test.ts
@@ -630,16 +630,32 @@ describe('dev proxy', () => {
       expect(headers.Authorization).toBeUndefined()
     })
 
-    test('sends Bearer token for theme-extension type', async () => {
+    test('does NOT send Bearer token for theme-extension type', async () => {
       const extCtx = {
         ...tokenCtx,
-        type: 'theme-extensions',
+        type: 'theme-extension',
       } as unknown as DevServerContext
       const event = createH3Event('GET', '/assets/style.css')
       await proxyStorefrontRequest(event, extCtx)
       const [, init] = fetchMock.mock.calls[0] as [URL, RequestInit]
       const headers = init.headers as Record<string, string>
       expect(headers.Authorization).toBeUndefined()
+    })
+
+    test('strips query string before auth check for /cart.js?sections=header', async () => {
+      const event = createH3Event('GET', '/cart.js?sections=header')
+      await proxyStorefrontRequest(event, tokenCtx)
+      const [, init] = fetchMock.mock.calls[0] as [URL, RequestInit]
+      const headers = init.headers as Record<string, string>
+      expect(headers.Authorization).toBeUndefined()
+    })
+
+    test('sends Bearer token for non-CDN paths that are not cart/checkout/account', async () => {
+      const event = createH3Event('GET', '/products/some-product')
+      await proxyStorefrontRequest(event, tokenCtx)
+      const [, init] = fetchMock.mock.calls[0] as [URL, RequestInit]
+      const headers = init.headers as Record<string, string>
+      expect(headers.Authorization).toBe('Bearer sfr-devtools-token')
     })
   })
 
--- a/packages/theme/src/cli/utilities/theme-environment/proxy.test.ts
+++ b/packages/theme/src/cli/utilities/theme-environment/proxy.test.ts
@@ -553,6 +553,96 @@ describe('dev proxy', () => {
     })
   })
 
+  describe('proxyStorefrontRequest — Bearer token auth scoping', () => {
+    let fetchMock: ReturnType<typeof vi.fn>
+    const tokenCtx = {
+      ...ctx,
+      type: 'theme',
+      session: {
+        storeFqdn: 'my-store.myshopify.com',
+        sessionCookies: {},
+        storefrontToken: 'sfr-devtools-token',
+      },
+    } as unknown as DevServerContext
+
+    beforeEach(() => {
+      fetchMock = vi.fn().mockResolvedValue(new Response('OK'))
+      vi.stubGlobal('fetch', fetchMock)
+    })
+
+    afterEach(() => {
+      vi.unstubAllGlobals()
+    })
+
+    test('sends Bearer token for CDN asset requests', async () => {
+      const event = createH3Event('GET', '/cdn/shop/files/style.css')
+      await proxyStorefrontRequest(event, tokenCtx)
+      const [, init] = fetchMock.mock.calls[0] as [URL, RequestInit]
+      const headers = init.headers as Record<string, string>
+      expect(headers.Authorization).toBe('Bearer sfr-devtools-token')
+    })
+
+    test('does NOT send Bearer token for /cart/add.js', async () => {
+      const event = createH3Event('POST', '/cart/add.js')
+      await proxyStorefrontRequest(event, tokenCtx)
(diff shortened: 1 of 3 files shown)


gotesgan:
I have signed the CLA

packages/theme/src/cli/utilities/theme-environment/proxy.test.ts
```diff
@@ -553,6 +553,96 @@ describe('dev proxy', () => {
...
+
+    test('does NOT send Bearer token for account endpoints', async () => {
+      const event = createH3Event('GET', '/account/logout')
+      await proxyStorefrontRequest(event, tokenCtx)
+      const [, init] = fetchMock.mock.calls[0] as [URL, RequestInit]
+      const headers = init.headers as Record<string, string>
+      expect(headers.Authorization).toBeUndefined()
+    })
+
+    test('sends Bearer token for theme-extension type', async () => {
```
EvilGenius13: This test says it's sending a bearer token for theme-extension types but the expectation is `Undefined`. The test needs to be renamed.

packages/theme/src/cli/utilities/theme-environment/proxy.test.ts
```diff
@@ -553,6 +553,96 @@ describe('dev proxy', () => {
...
+      const extCtx = {
+        ...tokenCtx,
+        type: 'theme-extensions',
```
EvilGenius13: The type is `theme-extension` for clarity.

packages/theme/src/cli/utilities/theme-environment/proxy.ts
```diff
@@ -116,6 +119,18 @@ export function canProxyRequest(event: H3Event) {
   return Boolean(extension) || acceptsType !== '*/*'
 }
 
+// Cart, checkout, and account endpoints use session-cookie auth, not Bearer tokens.
+// CHECKOUT_PATTERN and ACCOUNT_PATTERN are reused from routing because their
+// routing and auth scopes currently align; changes to those patterns should
+// consider the auth implications here.
+function needsBearerToken(path: string): boolean {
+  const [pathname] = path.split('?') as [string]
```
EvilGenius13, line 127: Can we add a test that shows `path.split` working such as `/cart.js?sections=header`

packages/theme/src/cli/utilities/theme-environment/proxy.test.ts
```diff
@@ -553,6 +553,96 @@ describe('dev proxy', () => {
     })
   })
 
+  describe('proxyStorefrontRequest — Bearer token auth scoping', () => {
```
EvilGenius13, line 556: Can we get a test for a non CDN pattern that shows normal paths still authenticate?

EvilGenius13 (commented):
Thank you for the PR @gotesgan! I have left a few comments.

The core logic makes sense as it's a 1:1 recreation of the previous [PR ](https://github.com/Shopify/cli/pull/6839/changes)introducing these changes. That one was reverted because dropping the bearer token caused legitimate traffic to be flagged as a bot. This is now handled by the new headers shipped in `4.4.0` so removing the token off cart/checkout/account should be safe again.

Could you please confirm that you've tested this by running `theme dev` against a live store?

gotesgan:
Hello I will add test and resubmit everything once again thanks for the comments

gotesgan:
I've tested this live against a store (`shantanu-test-retnzy.myshopify.com`):
- `/cart.js` → returns **200** with product data
- `POST /cart/add.js` with a real variant → returns **200** with cart item data (previously was 401/403)

The fix is working as expected. Cart endpoints now use cookie-based auth instead of being sent a Bearer token they can't use.

EvilGenius13:
@gotesgan thank you for the updates! Your PR is almost ready. There is a failing piece related to linting [here](https://github.com/Shopify/cli/actions/runs/28971823652/job/86159341524?pr=8027). If you can fix that and push up the changes I believe this PR should be fine to merge.

gotesgan:
Thanks @EvilGenius13! I've fixed the prettier formatting issue and pushed it up. All checks should pass now:
Appreciate the review!

EvilGenius13 (approved):
Thank you for your contribution @gotesgan!

gotesgan:
Thank you!

PR #8035, opened 2026-07-08
Add store delete command

alfonso-noriega:

Fixes https://github.com/shop/issues-develop/issues/22936

Shopify CLI needs a CLI-side `shopify store delete` command for app development stores. Business Platform now accepts app development store deletion requests by calling Core `billingCloseShop(immediate: true)`, but Core cancellation completes asynchronously and BP no longer eagerly removes the BP Shop row. Because mutation success means “close request accepted”, the CLI must wait for the authoritative cancellation signal from Organizations API instead of treating row disappearance as success.

- Adds a hidden `shopify store delete` command in `@shopify/store` for deleting development stores.
- Supports the command flag shape: global flags, `--json`, `--store`, and `--organization-id`.
- Adds the `deleteAppDevelopmentStore` Business Platform mutation document and typed document.
- Resolves the target store’s `shopifyShopId` using the existing Organizations API accessible shops lookup before polling.
- After a successful delete mutation, polls `organization.accessibleShop(id:)` until `planName` is exactly `cancelled`.
- Keeps delete mutation user error handling unchanged and does not poll when the mutation returns user errors.
- Uses bounded polling consistent with development store creation polling, and reports deletion as requested but not confirmed if cancellation is still asynchronous at timeout.
- Treats a null `accessibleShop` response as inconclusive rather than confirmed cancellation.
- Registers the command in the store plugin and CLI oclif manifest.
- Adds tests for successful cancellation confirmation, mutation user errors, timeout behavior, and null `accessibleShop` polling responses.

### How to test your changes?

Tophat:

1. Enable the backend store-create/delete CLI flag ([`f_cli_store_create_enabled`](https://experiments.shopify.io/flags/f_cli_store_create_enabled)) for your organization.
2. From this branch, create a new development store through the local CLI:

   ```bash
   pnpm shopify store create dev --name <store-name> --organization-id <organization-id> --plan plus
   ```

3. Confirm the store appears in Dev Dashboard: `https://dev.shopify.com/dashboard/<organization-id>/stores`.
4. From this branch, run the new delete command through the local CLI:

   ```bash
   pnpm shopify store delete --store <store-subdomain>.myshopify.com
   ```

5. Confirm the command reports the deletion request/confirmation and that the development store is no longer active in Dev Dashboard.

Automated checks:

- `pnpm --filter @shopify/store vitest run src/cli/commands/store/delete.test.ts src/cli/services/store/delete/dev.test.ts src/cli/utilities/store-lookup/organization-shop.test.ts`
- `pnpm --filter @shopify/store lint`
- `pnpm --filter @shopify/store type-check`

### Post-release steps

None.

### Checklist


alfonso-noriega:
* **#8035** <a href="https://app.graphite.com/github/pr/Shopify/cli/8035?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Shopify/cli/8035?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

alexCevi:
This has been NEEDED!

packages/store/src/cli/commands/store/delete.ts
```diff
@@ -0,0 +1,70 @@
...
+import {AbortError} from '@shopify/cli-kit/node/error'
+import {outputResult} from '@shopify/cli-kit/node/output'
+import {Flags} from '@oclif/core'
+
+export default class StoreDelete extends Command {
+  static hidden = true
+
+  static summary = 'Delete a development store.'
+
+  static descriptionWithMarkdown = 'Deletes an app development store from your organization.'
```
isaacroldan: why do we specify "app" development store here? a development store is not only for apps, you can use it for themes too no?

packages/store/src/cli/commands/store/delete.ts
```diff
@@ -0,0 +1,70 @@
...
+    ...globalFlags,
+    ...jsonFlag,
+    store: Flags.string({
+      char: 's',
+      description: 'The myshopify.com domain of the development store to delete.',
+      env: 'SHOPIFY_FLAG_STORE',
+      parse: async (input) => normalizeStoreFqdn(input),
+      required: true,
+      aliases: ['name'],
+    }),
```
isaacroldan: There is a shared `storeFlags.store`, let's reuse that

isaacroldan (approved):

.changeset/confirm-store-cancellation.md
```diff
@@ -0,0 +1,5 @@
+---
+'@shopify/store': patch
+---
+
+Wait for development store cancellation confirmation after deletion is requested.
```
isaacroldan: this is not a user visible change no? so no need for a changeset then

packages/store/src/cli/services/store/info/organization-shop.ts
```diff
@@ -18,6 +18,21 @@ interface FetchOrganizationShopOptions {
 }
 
 export async function fetchOrganizationShop(options: FetchOrganizationShopOptions): Promise<OrganizationShopFields> {
+  const shop = await fetchOptionalOrganizationShop(options)
+
+  if (!shop) {
+    throw new BugError(
```
isaacroldan: is this a BugError (will be reported, might trigger an issue) or an AbortError (not reported)

packages/store/src/cli/services/store/delete/dev.ts
```diff
@@ -0,0 +1,173 @@
...
+import {type UnauthorizedHandler} from '@shopify/cli-kit/node/api/graphql'
+import {ensureAuthenticatedBusinessPlatform} from '@shopify/cli-kit/node/session'
+import {renderSingleTask, renderSuccess, renderWarning} from '@shopify/cli-kit/node/ui'
+import {outputContent, outputResult} from '@shopify/cli-kit/node/output'
+import {AbortError} from '@shopify/cli-kit/node/error'
+import {sleep} from '@shopify/cli-kit/node/system'
+
+const CANCELLED_PLAN_NAME = 'cancelled'
+const POLL_INTERVAL_SECONDS = 2
+const POLL_TIMEOUT_MS = 5 * 60 * 1000
```
isaacroldan, line 20: how long does it usually take? polling every 2seconds for 5min seems like a lot :thinking_face:  
Can we poll every 5s instead?

packages/store/src/cli/services/store/delete/dev.ts
```diff
@@ -0,0 +1,173 @@
...
+        >({
+          query: OrganizationAccessibleShop,
+          token: options.token,
+          organizationId: options.organizationId,
+          variables: {id: options.shopifyShopId},
+          unauthorizedHandler: options.unauthorizedHandler,
+        })
+
+        const accessibleShop = pollResult.organization?.accessibleShop
+        if (accessibleShop?.planName === CANCELLED_PLAN_NAME) {
```
isaacroldan: 🐛 **Bug**: The cancellation gate hinges on a strict, case-sensitive check `accessibleShop?.planName === CANCELLED_PLAN_NAME` where `CANCELLED_PLAN_NAME = 'cancelled'`. This is the linchpin of the delete flow, yet the rest of this package deliberately treats BP `Shop.planName` casing as unreliable: `mapPlanToPublicHandle` in `info/plan.ts:9` lowercases before matching, `PLAN_HANDLES_BY_NAME` in `services/store/constants.ts` documents that "the exact form BP returns isn't pinned down by the schema," and the generated type keeps `planName` as an opaque `string | null`. If the Organizations API ever returns `Cancelled`/`CANCELLED`, this exact match never becomes true, so every successful cancellation polls for the full 5-minute timeout and then renders the misleading "cancellation was not confirmed" warning. The passing test only proves the mock uses lowercase, not that BP does.

**Suggestion:** Consider normalizing before comparing, mirroring the established convention in `info/plan.ts:9`:
```suggestion
        if (accessibleShop?.planName?.toLowerCase() === CANCELLED_PLAN_NAME) {
```

isaacroldan (commented):
---

.changeset/confirm-store-cancellation.md
alfonso-noriega: Done — removed the changeset since this is still a hidden command / internal flow change and should not appear in public release notes.

packages/store/src/cli/services/store/info/organization-shop.ts
alfonso-noriega: Done — changed this path to throw AbortError instead of BugError, and updated the organization-shop tests accordingly. This is an expected lookup/access failure mode rather than something we should report as a CLI bug.

packages/store/src/cli/services/store/delete/dev.ts
alfonso-noriega: Done — changed the polling interval from 2s to 5s and updated the delete polling tests to assert the new interval.

packages/store/src/cli/services/store/delete/dev.ts
alfonso-noriega: Addressed intentionally by keeping the strict exact match. The backend cancellation confirmation contract for this flow is planName == "cancelled" specifically, and we should not treat other values/casing as confirmed cancellation. I added/kept test coverage that continues polling through "deleted" and "Cancelled" and only succeeds on exact "cancelled".

alfonso-noriega: Follow-up: changed this after discussion to normalize both sides with lowercasing before comparing, matching the existing BP plan-name handling pattern. Added/updated test coverage so polling continues through unrelated values like "deleted" but confirms cancellation when the returned plan normalizes to "cancelled".

dmerand (approved):

packages/store/src/cli/commands/store/delete.ts
```diff
@@ -0,0 +1,84 @@
...
+
+  static examples = [
+    '<%= config.bin %> <%= command.id %> --store shop.myshopify.com --organization-id 1234567',
+    '<%= config.bin %> <%= command.id %> --store shop.myshopify.com --organization-id 1234567 --json',
+  ]
+
+  static flags = {
+    ...globalFlags,
+    ...jsonFlag,
+    store: storeFlags.store,
```
gonzaloriestra, line 26: I see this flag is required, but I think we could prompt when the flag is missing. The issue is common to all the store commands, so no need to fix it now, but I think we should improve it.

packages/store/src/cli/commands/store/delete.ts
```diff
@@ -0,0 +1,84 @@
...
+    'organization-id': storeFlags['organization-id'],
+  }
+
+  async run(): Promise<void> {
+    const {flags} = await this.parse(StoreDelete)
+    this.failMissingNonTTYFlags(flags, ['store'])
```
gonzaloriestra: The flag is required, so this is not needed, no?

packages/store/src/cli/commands/store/delete.ts
```diff
@@ -0,0 +1,84 @@
...
+
+    const organization = await organizationForStoreDeletion(flags.store, flags['organization-id']?.toString())
```
gonzaloriestra: If there's an error here, it will show a regular abort when the `--json` flag is provided. Not a big deal because we are not handling as expected yet json errors everywhere. But I think we could solve it by moving it inside the try-catch.

gonzaloriestra (approved):
I guess it's a backend issue, but the confirmation timed out:

```
❯ shopify store delete --store dev-store-cli.myshopify.com

╭─ warning ─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│  Development store "dev-store-cli.myshopify.com" deletion was requested, but not confirmed.                                                               │
│  Shopify accepted the deletion request, but deletion was not confirmed before the CLI stopped waiting. The store may still finish deleting                │
│  asynchronously.                                                                                                                                          │
```

I got an email about the store being closed and when I open admin it says `You can’t access this store because it’s frozen`. However, it's still listed in Dev Dashboard. I'll check later if it's still there.

packages/store/src/cli/commands/store/delete.test.ts
```diff
@@ -0,0 +1,149 @@
...
+
+  test('infers the organization from the store when --organization-id is omitted', async () => {
+    await StoreDelete.run(['--store', 'my-store.myshopify.com'])
+
+    expect(findStoreOwningOrganization).toHaveBeenCalledWith({store: 'my-store.myshopify.com', noPrompt: false})
+    expect(selectOrg).not.toHaveBeenCalled()
+    expect(deleteDevStore).toHaveBeenCalledWith(expect.objectContaining({organization: inferredOrg}))
+  })
+
+  test('falls back to prompting for the organization when store ownership cannot be inferred interactively', async () => {
```
amcaplan: Shouldn't this be non-interactively? Maybe this was meant to be "when running interactively"?

packages/store/src/cli/commands/store/delete.test.ts
```diff
@@ -0,0 +1,149 @@
...
+      message: 'Something went wrong',
+      nextSteps: [],
+      exitCode: 1,
+    })
+    expect(mockExit).toHaveBeenCalledWith(1)
+
+    mockExit.mockRestore()
+  })
+
+  test('does not output JSON for non-AbortError even when --json is active', async () => {
```
amcaplan, line 111: Is this actually a good thing? I guess the idea is to default to global error handling?

packages/store/src/cli/commands/store/delete.ts
amcaplan: Agreed here, I think we might need to hook a bit deeper into oclif to make that work though. Note that per [oclif docs](https://oclif.io/docs/flags/) we have have a flag with a `default` that's an async function, so this is very doable - and presumably that function fails in non-TTY environments.

packages/store/src/cli/commands/store/delete.ts
```diff
@@ -0,0 +1,84 @@
...
+          ),
+        )
+        process.exit(1)
+      }
+      throw error
+    }
+  }
+}
+
+async function organizationForStoreDeletion(store: string, organizationId?: string): Promise<Organization> {
```
amcaplan: Nothing about this function really has to do with deletion specifically. I'd recommend moving to a shared space - I could imagine other commands wanting to do this as well.

packages/store/src/cli/services/store/create/dev.ts
```diff
@@ -32,11 +32,11 @@ type StoreCreationStatus = NonNullable<
 function friendlyStatus(status: StoreCreationStatus): string {
   switch (status) {
     case 'CALLING_CORE':
-      return 'Initiating store creation...'
+      return 'Initiating store creation'
```
amcaplan, line 35: Thanks for fixing these up.

packages/store/src/cli/services/store/delete/dev.ts
```diff
@@ -0,0 +1,195 @@
...
+        unauthorizedHandler,
+      })
+    : false
+
+  if (options.json) {
+    outputResult(deletionResultJson({store, organization: org, deletionConfirmed}))
+  } else if (deletionConfirmed) {
+    renderSuccess({
+      headline: `Development store "${store}" deleted successfully.`,
+      body: ['Shopify accepted the deletion request and confirmed the store was deleted.'],
```
amcaplan: This could probably be more concise:
```suggestion
      body: ['The store was deleted.'],
```

packages/store/src/cli/services/store/delete/dev.ts
```diff
@@ -0,0 +1,195 @@
...
+async function fetchShopifyShopId(options: {
+  store: string
+  organizationId: string
+  token: string
+}): Promise<string | undefined> {
+  const shop = await fetchOptionalOrganizationShop(options)
+  return shop?.shopifyShopId
+}
+
+export function toOrganizationsShopifyShopId(shopifyShopId: string | number): string {
```
amcaplan, line 101: This feels a bit generic, but it's also too early to extract. So just a mental note to keep an eye out for maybe extracting this later.

packages/store/src/cli/services/store/delete/dev.ts
```diff
@@ -0,0 +1,195 @@
...
+  // The backend delete operation currently reuses the store-create CLI gate and localized
+  // message. Keep the CLI output delete-specific so users understand which operation failed.
+  if (isCliStoreCreateGateMessage(message)) {
+    return 'Deleting a development store from the Shopify CLI is not yet enabled for your organization.'
+  }
+
+  return message
+}
+
+function isCliStoreCreateGateMessage(message: string): boolean {
```
amcaplan: This doesn't seem right, why are we only checking 2 languages?

packages/store/src/cli/commands/store/delete.ts
alfonso-noriega: Good catch — since `storeFlags.store` is `required: true`, oclif fails parsing before this line could ever run, so it was dead code. Removed in b65bffb. If we later make the flag optional with a prompt fallback (per the other thread), we can reintroduce it as part of that work.

packages/store/src/cli/commands/store/delete.ts
alfonso-noriega: Done in b65bffb — organization resolution now happens inside the try/catch, so an `AbortError` thrown while resolving the organization is also rendered as structured JSON when `--json` is active. Added a test covering that path.

packages/store/src/cli/commands/store/delete.ts
alfonso-noriega: Done in b65bffb — moved it to `utilities/store-lookup/organization.ts` as `resolveOrganizationForStore(store, organizationId?)`, next to `findStoreOwningOrganization`, and moved the org-resolution tests alongside it so future store commands can reuse both.

packages/store/src/cli/services/store/delete/dev.ts
alfonso-noriega: Done in b65bffb — took the suggestion. The headline already communicates the outcome, so the body is now just "The store was deleted."

packages/store/src/cli/commands/store/delete.test.ts
alfonso-noriega: Yes, it's deliberate: only expected failures (`AbortError`) are rendered as structured JSON. Unexpected errors rethrow to the global cli-kit error handler so they keep their stack traces and get reported as CLI bugs instead of being swallowed into JSON output. Added a code comment in `delete.ts` documenting this (b65bffb).

amcaplan: Makes sense. Perhaps the bigger win would be to detect `--json` in the global error handler and output an appropriate error as JSON. But that's for another PR.

packages/store/src/cli/commands/store/delete.ts
amcaplan: Actually, I may have been wrong to consider this a small ask. Because in general, `--store` can refer to a store accessed via org or via store auth. So it's actually somewhat confusing how one would even select the store (org => store? list of authed stores?), and it might differ from command to command.

packages/store/src/cli/services/store/create/dev.ts
alfonso-noriega: 🙏 resolving.

packages/store/src/cli/services/store/delete/dev.ts
alfonso-noriega: Agreed — it's generic, but with a single caller it's too early to extract. Leaving it here until a second caller appears; resolving as no action needed for this PR.

packages/store/src/cli/commands/store/delete.test.ts
alfonso-noriega: Good catch — the wording was ambiguous. The test moved to `utilities/store-lookup/organization.test.ts` during the org-resolution extraction, and it's now named `'falls back to prompting for the organization when running interactively and ownership cannot be inferred'`.

packages/store/src/cli/services/store/delete/dev.ts
alfonso-noriega: Agreed, string-matching localized messages is brittle — this is a stopgap. What I found digging into it:

- The BP `UserError` type has an optional `code` field, and our mutation document **already selects it** — but the backend gate (`ShopManagement::Shop::DeleteAppDevelopmentStore#check_cli_gate`) only populates `field` + `message` (`I18n.t("operations.errors.cli_store_create_not_enabled")`), so `code` is always null today.

alfonso-noriega: Done — went with the `code` approach:

- **Backend:** shop/world#958405 populates `UserError.code` with `cli_store_create_not_enabled` on the CLI gate user error for both `createAppDevelopmentStore` and `deleteAppDevelopmentStore` (the gate error is built in BP's `check_cli_gate`, so it's a one-key addition; the `UserError` type already exposed `code`).
- **CLI (this PR):** `deleteDevelopmentStoreUserErrorMessage` now branches on `error.code === 'cli_store_create_not_enabled'` instead of matching localized strings — `isCliStoreCreateGateMessage` is gone. Test updated to prove the rewrite works regardless of message locale (Spanish message + code).

Until the BP PR deploys, `code` is null and users see the raw (localized) backend message — acceptable since the command is hidden and gated.

alfonso-noriega: Update — we went one step further and made the backend gate error operation-agnostic instead (shop/world#958405, updated): both mutations now return *"Store management from the Shopify CLI is not yet enabled for your organization."* with `code: cli_store_management_not_enabled`, since a single flag gates all CLI store management.

That let the CLI drop the special-casing entirely: `deleteDevelopmentStoreUserErrorMessage` is gone and the delete flow is now a plain `userErrors.map((error) => error.message)` passthrough, matching the convention everywhere else in the codebase (including `create/dev.ts`).

PR #8057, opened 2026-07-10
[Prototype] POS UI extension intercept declaration (capabilities.intercepts)

henryStelle:
## What this adds

Schema support for a UI extension to manually declare the intercept events planned for each target in `shopify.extension.toml`:

```toml
[[extensions.targeting]]
target = "pos.app.ready.data"
module = "./src/index.ts"

[extensions.targeting.capabilities]
intercepts = ["beforecheckout"]
```

`intercepts` is target-level configuration, not extension-level configuration. Event names are simply the same names exposed by the approved [`shopify.intercept()` runtime API](https://github.com/Shopify/ui-api-design/pull/1557). The CLI carries the explicit declaration through the deploy flow; Core/server validates at deploy time which events are allowed for each target and API version.

This configuration proposal and the approved runtime APIs have separate responsibilities:

- `intercepts` declares the events a target plans to register.
- `shopify.intercept()` registers the actual runtime validators.
- This prototype uses the existing, approved [`shopify.capabilities` API](https://github.com/Shopify/ui-api-design/blob/be97e2ca7089b05db762a00941400c0e4dd3df94/libraries/javascript/ui-api-design/types/extensions/configuration/capabilities.md?plain=1) to expose the merchant-approved validation severity for each declared event. The per-event severity capability names are proposed by this configuration contract.

For a declared `beforecheckout` event, the proposed capabilities are `beforecheckout.error`, `beforecheckout.warning`, and `beforecheckout.info`. `beforecheckout.error` permits `ERROR`, `WARNING`, and `INFO`; `beforecheckout.warning` permits `WARNING` and `INFO`; and `beforecheckout.info` permits `INFO`. Lower severities are explicitly cumulative in the returned array: error approval returns all three strings, warning approval returns `beforecheckout.warning` plus `beforecheckout.info`, and info approval returns only `beforecheckout.info`.

## Implementation

- Adds `InterceptEventsSchema` in `packages/app/src/cli/models/extensions/schemas.ts` for the structural array-of-strings and uniqueness rules.
- Adds optional `intercepts` to the existing `TargetCapabilitiesSchema` used by each `[[extensions.targeting]]` entry. The CLI deliberately defines no event-name regex or separate grammar; Core validates the exact Intercept API event name against the target and API version.
- Uses the existing UI extension transform and deploy flow, which carries each target's capabilities into the literal extension-point config.
- Does not change extension-level `capabilities` or checkout's `block_progress`.
- Adds focused schema tests for free-form string forwarding, non-string values, and duplicate declarations.

## Why explicit target-level declarations are recommended

- Developers manually and explicitly declare every intercept event their target plans to register. The declaration is reviewable and remains reliable when registration is indirect or dynamic.
- Target-level granularity is intentional. Separate targets can register different event sets, and this leaves room for different target-specific configurations in the future.
- Checkout's `block_progress` is not the right model. It is one boolean for a single buyer-journey blocking concept, while POS can intercept distinct workflow boundaries. Merchants need to know exactly where participation can occur and what severity they approved per event.
- Core/server can validate target/event/API-version support at deploy time.
- POS can compare declared events with validators actually registered through `shopify.intercept()` and detect a missing registration as an extension failure. This is particularly important for compliance validators, where failure to run can have legal implications.

## Alternative explored

[#8058](https://github.com/Shopify/cli/pull/8058) explores automatically deriving intercept events from extension source. That follows the CLI's preference for derived defaults, but it is not the recommended approach: source analysis is complex, incomplete, and cannot handle every valid registration pattern. Explicit declarations support deploy-time target/event/API-version validation and provide a complete runtime contract for registration verification and compliance failure detection, while also supporting merchant visibility.

## Validation

- `pnpm test packages/app/src/cli/models/extensions/schemas.test.ts` — 26 tests passed
- `pnpm nx run app:lint`
- `pnpm nx run app:type-check`
- `git diff --check`


henryStelle:
Superseded by https://github.com/Shopify/cli/pull/8345

PR #8071, opened 2026-07-13
Update from stable/4.5

amcaplan:
Carrying over the changes we made on the stable branch


docs-shopify.dev/commands/interfaces/store-auth-list.interface.ts
isaacroldan, line 1: In main I recently merged a change to stop tracking all the `*.interface.ts` files, we should remove these.

alfonso-noriega (approved):

PR #8138, opened 2026-07-21
fix(theme): preserve preview_theme_id on password-protected theme dev sessions

EvilGenius13:
## What's happening

When you run `theme dev` on a password-protected store, sometimes it shows the live theme instead of the theme you're actually previewing. We have made some fixes to it but are still seeing problems in the dev [forums](https://community.shopify.dev/t/theme-id-mismatch/33427/29)

To set up the session we make two requests, and each one hands back a `_shopify_essential` cookie:
- the preview request → gives a cookie that knows which theme to preview
- the password request → gives a cookie that proves you passed the store password

The problem: we were doing preview first, password second. So the password cookie came back last and overwrote the preview cookie which meant we lost the preview theme and the store just showed the live theme. I believe it only happens sometimes, because the password step doesn't always hand back a new cookie.

## The fix

Just flip the order for password stores:

1. do the password request first, so we get the cookie that proves we're in
2. do the preview request last, passing that cookie along, so the theme id gets
   added on top of it

Now the final cookie has both things it needs, so it always previews the right theme. Same number of requests as before.

## Testing

-Run `theme dev` against a password-protected store confirm it renders the theme.

### DIFF
--- a/packages/theme/src/cli/utilities/theme-environment/storefront-session.test.ts
+++ b/packages/theme/src/cli/utilities/theme-environment/storefront-session.test.ts
@@ -57,14 +57,9 @@ describe('Storefront API', () => {
     })
 
     test('retrieves _shopify_essential and storefront_digest cookies when a password is provided', async () => {
-      // Given
+      // Given: POST /password runs first (cold, returns the digest), then the
+      // preview HEAD runs last and owns the final _shopify_essential.
       vi.mocked(shopifyFetch)
-        .mockResolvedValueOnce(
-          response({
-            status: 200,
-            headers: {'set-cookie': '_shopify_essential=:AABBCCDDEEFFGGHH==123:; path=/; HttpOnly'},
-          }),
-        )
         .mockResolvedValueOnce(
           response({
             status: 302,
@@ -72,6 +67,12 @@ describe('Storefront API', () => {
               'set-cookie': 'storefront_digest=digest-value; path=/; HttpOnly',
               location: 'https://example-store.myshopify.com/',
             },
+          }),
+        )
+        .mockResolvedValueOnce(
+          response({
+            status: 200,
+            headers: {'set-cookie': '_shopify_essential=:AABBCCDDEEFFGGHH==123:; path=/; HttpOnly'},
           }),
         )
 
@@ -91,17 +92,17 @@ describe('Storefront API', () => {
       vi.mocked(shopifyFetch)
         .mockResolvedValueOnce(
           response({
-            status: 200,
-            headers: {'set-cookie': '_shopify_essential=:AABBCCDDEEFFGGHH==123:; path=/; HttpOnly'},
-          }),
-        )
-        .mockResolvedValueOnce(
-          response({
             status: 302,
             headers: {
               'set-cookie': 'storefront_digest=digest-value; path=/; HttpOnly',
               location: 'https://example-store.myshopify.com/',
             },
+          }),
+        )
+        .mockResolvedValueOnce(
+          response({
+            status: 200,
+            headers: {'set-cookie': '_shopify_essential=:AABBCCDDEEFFGGHH==123:; path=/; HttpOnly'},
           }),
         )
 
@@ -117,9 +118,10 @@ describe('Storefront API', () => {
         },
       )
 
+      // Password POST runs first, then the preview HEAD.
       expect(shopifyFetch).toHaveBeenNthCalledWith(
         1,
-        'https://example-store.myshopify.com?preview_theme_id=123456&_fd=0&pb=0',
+        'https://example-store.myshopify.com/password',
         expect.objectContaining({
           headers: expect.objectContaining({
             Signature: 'signature-value',
@@ -130,7 +132,7 @@ describe('Storefront API', () => {
       )
       expect(shopifyFetch).toHaveBeenNthCalledWith(
         2,
-        'https://example-store.myshopify.com/password',
+        'https://example-store.myshopify.com?preview_theme_id=123456&_fd=0&pb=0',
         expect.objectContaining({
           headers: expect.objectContaining({
             Signature: 'signature-value',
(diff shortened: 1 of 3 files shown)


packages/theme/src/cli/utilities/theme-environment/storefront-session.test.ts
graygilmore: Note from me: I didn't validate this but I thought it was worth flagging if we've changed the signature of `getStorefrontSessionCookies` in previous commits but didn't change this here.

---

🐛 **Bug**: This call passes three positional arguments to a five-parameter function `(storeUrl, storeFqdn, themeId, password?, headers?)`, so `storeFqdn` binds to `'123456'`, `themeId` to `'password'`, and `password` is **undefined**. The test therefore takes the early-return branch at `storefront-session.ts:71-75` and only ever drives the cold retry loop. It passes because the four mocked no-cookie responses exactly exhaust the retry budget (`retries > 3`), which is why the defect has stayed invisible. The stale fourth mock returning `storefront_digest` (line 221) is unreachable in the passwordless branch — good evidence the test was written with the password flow in mind. Every other call in this describe block passes the full argument list.

That was tolerable before this PR, because the preview HEAD ran first and identically in both branches. It isn't now: the newly added retry forwarding at `storefront-session.ts:153` threads `incomingEssential` through every attempt, and dropping it would lose the password digest — the same class of failure this PR fixes. That line, and the `ShopifyEssentialError` path for password-protected stores generally (including its ripple into `abortOnMissingRequiredFile` at `dev-server-session.ts:128-130`), currently has no coverage at all.

**Suggestion:** 
```suggestion
      const cookies = getStorefrontSessionCookies(
        'https://example-store.myshopify.com',
        'example-store.myshopify.com',
        '123456',
        'password',
      )
```

With the password supplied, you'll want to prepend a 302 `/password` mock so the four no-cookie responses land on the preview HEAD retries, replacing the now-redundant `storefront_digest` mock at lines 218-224. A sibling test asserting that every retry attempt carries `Cookie: _shopify_essential=<digestEssential>` would pin line 153's forwarding behavior directly.

packages/theme/src/cli/utilities/theme-environment/storefront-session.ts
graygilmore, line 145: Note from me: this might be the same thing that Copilot is flagging above

---

🐛 **Bug**: This retry-then-throw block is unchanged by the diff, but its precondition just changed, so the interaction is easy to miss. `sessionEssentialCookie` treats an absent `_shopify_essential` in the response as a hard failure: retry up to 4 times (sleeping 1+2+3s), then throw `ShopifyEssentialError`. That contract was safe while the function only ever issued a *cold* request — a client with no session forces SFR to mint one and Set-Cookie it.

After this change, the preview HEAD on the password path carries an existing `_shopify_essential` (line 118). Servers commonly omit `Set-Cookie` when they mutate a session in place rather than rotating its identifier, and the PR description makes exactly that observation about SFR ("the password step doesn't always hand back a new cookie") — for the same request shape that the preview HEAD now has. If SFR stamps `preview_theme_id` onto the forwarded session without rotating the value, `getCookie` returns undefined even though the session is valid and already sitting in `incomingEssential`. The retry cannot recover, since every attempt resends the identical cookie and gets the identical no-op response — so the user pays 4 HEAD requests and ~6s of sleeps, then `ShopifyEssentialError` propagates to `dev-server-session.ts:128` and triggers `abortOnMissingRequiredFile`, aborting with a misleading "missing required files" diagnosis.

This is the mirror image of the bug being fixed: previously a missing Set-Cookie on the second request was harmless because the earlier value was retained; now the second request owns the cookie, so a missing Set-Cookie is fatal. Worth confirming against real SFR before merge, since whether SFR rotates on a warm preview HEAD is server-side behavior that can't be settled from this repo.

**Suggestion:** Consider falling back to the forwarded essential before entering the retry loop — on a request that already carried a valid session, an absent `Set-Cookie` means "keep the one you sent", not "transient failure":

```suggestion
  if (!shopifyEssential && incomingEssential) {
    /**
     * SFR omits Set-Cookie when it updates the session in place instead of
     * rotating it, so the essential we forwarded is still the current one.
     */
    return incomingEssential
  }

  if (!shopifyEssential) {
```

A test for it would be cheap: POST /password returns a digest essential, the preview HEAD returns no `set-cookie`, and `getStorefrontSessionCookies` resolves to the digest essential after exactly 2 fetches.

graygilmore (commented):
Couple of bot comments

packages/theme/src/cli/utilities/theme-environment/storefront-session.test.ts
frandiox: Looks right. This call uses the stale argument shape, so it never enters the password-protected branch. I think we should fix it before merging and add an assertion that every preview retry forwards the `_shopify_essential` returned by `POST /password`.

packages/theme/src/cli/utilities/theme-environment/storefront-session.ts
frandiox: I think the current retry/fail behavior is safer than the proposed fallback.

I verified this against the backend behavior. The incoming `_shopify_essential` comes from the password request, so it can represent an authenticated session without containing the preview theme. The preview request is expected to return an updated cookie with that preview state.
If no updated cookie is returned, reusing the incoming one could omit `preview_theme_id` and recreate the original theme mismatch.

packages/theme/src/cli/utilities/theme-environment/storefront-session.ts
```diff
@@ -102,11 +109,15 @@ async function sessionEssentialCookie(storeUrl: string, themeId: string, headers
...
     ...defaultHeaders(),
   }
 
+  if (incomingEssential) {
+    requestHeaders.Cookie = serializeCookies({_shopify_essential: incomingEssential})
+  }
+
   const response = await shopifyFetch(url, {
     method: 'HEAD',
     redirect: 'manual',
```
frandiox, line 124: Non-blocking: the backend may redirect this request before applying `preview_theme_id`. Because redirects are handled manually, the CLI could stop before receiving the updated `_shopify_essential`.

Note: simply using `redirect: 'follow'` would not fully solve this because node-fetch does not preserve cookies set during redirects. We should handle this separately with redirect-aware cookie handling. This behavior predates the PR, so it should not block this fix.

frandiox (commented):
Thanks! I verified the request ordering against the backend behavior with agents. Running the password request first and the preview request last seems like the correct approach.

karreiro (commented):
Thank you for this PR, @EvilGenius13!

While the root cause looks relatively deterministic, I wasn't able to reproduce it locally to confirm our hypothesis and verify the fix

I've sent you a message on Slack so we can zoom in into this together :)

Thanks again for catching this and proposing a fix!

packages/theme/src/cli/utilities/theme-environment/storefront-session.ts
aswamy, line 129: Can we add a log after this, but NOT the actual cookie value so verbose logs can help debug it?

```ts
outputDebug(
  `Storefront preview session: status=${response.status}, request_id=${
    response.headers.get('x-request-id') ?? 'unknown'
  }, attempt=${retries}, forwarded_essential=${Boolean(
    incomingEssential,
  )}, returned_essential=${Boolean(shopifyEssential)}, essential_rotated=${Boolean(
    incomingEssential && shopifyEssential && incomingEssential !== shopifyEssential,
  )}`,
)
```

This log (and the other one i mention) should help debug it. If we see the following, we know the password cookie was forwarded and the preview response replaced it:
```
Storefront password session: status=302, request_id=abc, returned_essential=true, returned_digest=false
Storefront preview session: status=200, request_id=def, forwarded_essential=true, returned_essential=true, essential_rotated=true
```

aswamy, line 216: Same here

```ts
outputDebug(
  `Storefront password session: status=${response.status}, request_id=${
    response.headers.get('x-request-id') ?? 'unknown'
  }, returned_essential=${Boolean(newShopifyEssential)}, returned_digest=${Boolean(storefrontDigest)}`,
)
```

PR #8147, opened 2026-07-22
Make payments extension capability fields optional

Shuhala:
## What
Makes `supports_installments`, `supports_deferred_payments`, and `supports_3ds` optional in payments app extension TOML config schemas, so partners can omit them.

`supports_3ds` stays required for `credit_card` and `custom_credit_card` extensions.

## Why
Companion to the core change in shop/world#940396 that relaxed the same properties on the server side. Without this CLI-side change, `shopify app deploy` rejects configs that omit these fields even though core now accepts them.

### DIFF
--- a/packages/app/src/cli/models/extensions/specifications/payments_app_extension_schemas/base_payments_app_extension_schema.ts
+++ b/packages/app/src/cli/models/extensions/specifications/payments_app_extension_schemas/base_payments_app_extension_schema.ts
@@ -49,8 +49,8 @@ export const BuyerLabelSchema = zod.object({
 })
 
 export const DeferredPaymentsSchema = zod.object({
-  supports_installments: zod.boolean(),
-  supports_deferred_payments: zod.boolean(),
+  supports_installments: zod.boolean().optional(),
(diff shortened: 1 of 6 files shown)


packages/app/src/cli/models/extensions/specifications/payments_app_extension_schemas/base_payments_app_extension_schema.ts
dmerand, line 1: Could you confirm whether `supports_installments` and `supports_deferred_payments` may be omitted independently? The schema still rejects configs where only one is missing; the new tests cover only both being omitted. If the pair must remain synchronized, please add a test documenting that contract.

dmerand (approved):
LGTM overall, just one question about test coverage.

packages/app/src/cli/models/extensions/specifications/payments_app_extension_schemas/base_payments_app_extension_schema.ts
Shuhala: Yes they can be optional independently. Neither of these fields is actually used.

PR #8159, opened 2026-07-22
Fix cli-kit SelectInput/MultiSelectInput viewport and state management

amcaplan:
## Summary

Fixes two critical bugs in cli-kit `SelectInput` and `MultiSelectInput` when using the description panel:

1. **R1 (Vertical Overflow)**: Grouped lists could overflow the viewport in stacked layout when the available height budget is tight (availableLines < 7), reintroducing visual ghosting.

2. **R2 (Selection Reset)**: Width-only terminal resizes that cross the description-panel layout breakpoint would silently reset the highlighted item back to the first option.

### Core State Management Fix
**`use-select-state.ts`**:
- Added `AdjustVisibleWindowAction` type to handle visibleOptionCount changes without resetting the option set
- New reducer case that preserves the current selection (`value`) and only recomputes the visible scroll window
- Updated the effect that detects visibleOptionCount changes to dispatch the new action instead of resetting to default state

### Viewport Layout Fixes
**`SelectInput.tsx` and `MultiSelectInput.tsx`**:
- Apply a hard-ceiling clamp to `sectionHeight` **only** in stacked description layout (gated by `descriptionsEnabled && !showDescriptionBeside`)
- Clamp is applied **after** the `minHeight` floor to ensure `sectionHeight + STACKED_HINT_RESERVE <= availableLines`
- Preserves byte-for-byte invariance: no-description and beside-layout paths are unchanged

### Test Coverage
**`use-select-state.test.tsx`** (new file):
- Unit tests for the state hook with two scenarios:
  - Verifies selection is preserved when visibleOptionCount changes (R2 path)
  - Verifies reset path still works when the option set itself changes

**`SelectInput.description.test.tsx` and `MultiSelectInput.description.test.tsx`**:
- R1 regression test: "keeps a grouped stacked list within the vertical budget (availableLines=3|6)"
  - Tests both tight budgets to catch the overflow at edge cases
  - Validates the invariant `listHeight + STACKED_HINT_RESERVE <= availableLines` throughout navigation
- R2 regression test: "preserves the highlighted item across a width-only resize (beside↔stacked)"
  - Navigates to an item in narrow/stacked layout, then simulates a resize to wide/beside layout
  - Verifies the highlight stays on the same item (not reset to item 0)
  - Verifies onChange is not called with a different value (no silent selection jump)

## Testing

- All 249 ui tests pass (33 test files)
- New regression tests exercise both bugs explicitly
- No type errors or lint issues

### Why the hard-ceiling clamp placement matters
The `minHeight` floor (5 for grouped lists, 2 otherwise) was originally applied unconditionally, forcing the list height to a minimum even when the available budget couldn't accommodate it. In stacked layout, this meant the list + gap + preview could exceed the available height. The hard-ceiling clamp comes after the floor, ensuring we never exceed the budget.

### Why the state action preserves selection
When only the visible-row budget changes (e.g., a width resize), the option set is unchanged, so the user's current selection is still valid. Resetting to the first option would silently move the highlight without user input, violating the principle of least surprise.


amcaplan:
Closing: this PR was opened in error by an automated tool and was never intended to exist. The branch stays on origin, but this is prototype/hackday work (the `store report` command depends on an employee-only internal LLM proxy) and is not a merge candidate. No review needed.

PR #8177, opened 2026-07-27
Add additional detail to country flag description

amcaplan:
Touchup, see https://github.com/Shopify/cli/pull/7947#discussion_r3656862827


gonzaloriestra (approved):
Thanks!

PR #7254, opened 2026-04-13
Default to pnpm/pnpx internally

gonzaloriestra:
### WHY are these changes introduced?

We should align internal tooling with the repo’s default package manager (pnpm), so contributors and CI use the same commands.

### WHAT is this pull request doing?

- **Root** **`create-app`** **script**: Passes `--package-manager pnpm` into the create-app dev entrypoint.
- **E2E**: Default `packageManager` in `packages/e2e/setup/app.ts` is `pnpm` when `E2E_PACKAGE_MANAGER` is unset; Playwright steps in `tests-pr.yml` use `pnpm exec playwright` so CI uses the workspace’s pinned Playwright (avoids `pnpx` pulling a different version). E2E helper script comments / create-app flows updated for pnpm where relevant.
- **Misc**: Small script and workflow/doc tweaks (e.g. `dev.yml`, `build-dev-docs.sh`, `create-notification-pr.js`, `release.yml`) for consistency with pnpm-first usage.

### How to test your changes?

CI

### Checklist

- [x] I've considered possible cross-platform impacts (Mac, Linux, Windows)
- [x] I've considered possible [documentation](https://shopify.dev) changes
- [x] I've considered analytics changes to measure impact
- [ ] The change is user-facing, so I've added a changelog entry with `pnpm changeset add`

PR #8178, opened 2026-07-27
Fix duplicated package manager issue with create-test-app script

lusarz:

`./bin/create-test-app.js` fails with `Flag --package-manager can only be specified once` on every local run.

The root `create-app` script in `package.json` was changed in https://github.com/Shopify/cli/pull/7254 to hardcode `--package-manager pnpm`:

```
"create-app": "nx build create-app && node packages/create-app/bin/dev.js --package-manager pnpm"
```

`create-test-app.js` also appended its own `--package-manager=${nodePackageManager}` flag when building the local install args, so the underlying `app:init` command ended up receiving `--package-manager` twice, which oclif rejects.

Removes the redundant `--package-manager` flag that `create-test-app.js` was appending for local installs, since the root `create-app` script already supplies it.

### How to test your changes?

```bash
./bin/create-test-app.js
```

Should complete without the `Flag --package-manager can only be specified once` error.

### Checklist

### DIFF
--- a/bin/create-test-app.js
+++ b/bin/create-test-app.js
@@ -191,7 +191,6 @@ program
 
         log(`Creating new app in '${appPath}'...`);
         initArgs.push("--local");
-        initArgs.push(`--package-manager=${nodePackageManager}`);
         await nodeExec(["create-app"], initArgs, defaultOpts);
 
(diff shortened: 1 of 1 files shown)


lusarz:
I believe it's still important

PR #8182, opened 2026-07-28
Emit a JSON error document when --json is active

amcaplan:

`--json` was never considered by the global error renderer. Because the human error banner is written to **stderr**, a failing `shopify <command> --json | jq` previously received *nothing* on stdout — not malformed output, but no output at all. Scripts had no parseable failure signal.

This routes fatal errors through a JSON serializer when JSON output is active, emitting exactly one document on stdout.

**Shape**

```json
{"error":{"type":"abort","message":"Couldn't find an app","tryMessage":"Run shopify app init","nextSteps":["Visit the docs"],"customSections":[{"title":"Extensions","body":[["name","status"],["my-ext","failed"]]}]}}
```

- **`{"error": ...}` envelope.** Successful `--json` payloads in this CLI are bare values — `theme list` and `app versions list` emit arrays, `theme push` and `store list` emit command-shaped objects. A top-level `error` key is therefore unambiguous, and no existing success payload uses one. It also matches oclif's and npm's convention.
- **`type` is a closed set**: `abort`, `bug`, `external`. Machine-stable by design (see below).
- **Absent fields are omitted, never `null`** — consumers use presence checks, consistent with existing `JSON.stringify`/conditional-spread practice in the repo.
- **`stack` only appears for `bug`**, the one type that asks the user to file a report.
- **No `exitCode` field.** `errorMapper` constructs a fresh `AbortError` that drops `oclif.exit`, so any code in the payload could contradict the real status. The **process exit status** is the contract.
- Every nested `TokenItem` is flattened to a plain string with ANSI stripped; no Ink/React props reach the payload.

**Why the discriminator is a string literal**

Not `constructor.name` and not `error.name`. `packages/cli/bin/bundle.js` builds the published npm bundle with `minifyIdentifiers: true` and no `keepNames`, so `class AbortError` becomes `e=class extends s{}` — any reflective name is garbage in production. (`error.name` is worse: no error class in `error.ts` assigns it, and ES subclasses don't inherit it, so it is `"Error"` for every type.) Each class carries an explicit literal, backed by a `satisfies Record<FatalErrorType, JsonErrorType>` map so adding a `FatalErrorType` member fails the build. A test transforms `error.ts` with `minifyIdentifiers: true` and asserts every literal survives.

**Known approximations**

- **Detection is argv sniffing, not oclif parsing.** `sniffForJson` inspects `process.argv` before oclif parses, because the error path must work even when parsing itself failed. It respects the `--` passthrough boundary and handles clustered short flags. It cannot model per-flag arity, so `--path --json` (where `--json` is consumed as another flag's *value*) is a false positive. Fixing that means reimplementing oclif's parser in the error path; not worth the fragility.
- **`theme preview -j` is a permanent false positive** — that command declares `-j` for an unrelated flag. Blast radius is limited to printing a JSON error instead of a banner on a failing `theme preview`.
- **`renderConcurrent` writes Ink frames to stdout.** No command pairs it with `--json` today (`app build`/`app deploy` declare no json flag); only `theme-command.ts`'s multi-environment runner remains, where output is already interleaved across environments. Gating the progress renderers on JSON mode is a sensible hardening follow-up, not a fix this change needs. `renderTasks` already defaults to stderr.

**Review provenance**

Three independent reviews on three different models audited this before submission (control flow, public contract, test integrity). They produced one blocker and several fixes, all addressed: the explicit discriminator is now runtime-validated so an unadvertised value from a foreign cli-kit copy cannot reach the wire; `FatalErrorType`'s numeric values are pinned as append-only since they act as a cross-copy wire format; `abortSilent` can no longer appear in an emitted document's type; and two false-green tests were rewritten to be load-bearing and mutation-verified.

**Follow-ups (deliberately out of scope)**

- Both `uncaughtException` handlers (`bootstrap.ts`, `index.ts`) still render human output.
- Render-and-continue sites in theme utilities and `plugin-did-you-mean` are ungated.
- Gate progress renderers (`renderConcurrent`) on JSON mode.
- `launchCLI` double-prints errors (missing `skipOclifErrorHandling`).
- `error-grouping.ts` keys Bugsnag's `groupingHash` on `constructor.name`, which minification mangles.
- `theme preview` declares `json` without `char: 'j'`, inconsistent with the global flag.

### How to test your changes?

**Testing note — an honest limitation**

Stream identity is **not** directly assertable in this harness: `shouldOutput()` returns `false` under `isUnitTest()`, so nothing reaches a real stream, and `collectLog` keys captured output by log *level* rather than by stream — so `outputResult` (stdout) and `outputInfo` (stderr) both land under `info` and the captured log cannot tell them apart.

What the suite asserts instead is the **logger reference**: `outputResult` delegates to the private `output()` with `consoleLog` (`process.stdout.write`), while `outputInfo` uses `consoleWarn` and never calls `output()` at all. Asserting that call is what catches a switch between the two. That is the closest a unit test here can get to observing stream identity; verifying real file-descriptor separation needs an integration-level test.

This distinction is not academic. The original version of that test asserted only that captured output was non-empty — and mutating `outputResult` to `outputInfo` left **27 of 28 tests in the file green**. Only the rewritten test catches it.

**Testing**

- Four affected test files: **117 passed (117)**, exit 0
- Full `cli-kit` suite: **146 files, 1735 passed | 1 skipped**, exit 0
- Type-check: clean (verified with `--skip-nx-cache`, since a cached pass would prove nothing)
- Lint: `pnpm eslint src` clean, 0 errors 0 warnings
- The four test-integrity fixes were **mutation-verified**: each mutation was applied, observed to fail with a recorded message, and reverted. Six mutations total, all six targets confirmed back in their final state.

### Checklist

### DIFF
--- a/.changeset/json-error-output.md
+++ b/.changeset/json-error-output.md
@@ -1,0 +1,5 @@
+---
+'@shopify/cli-kit': minor
+---
+
+Emit a parseable JSON error document on stdout instead of an error banner when `--json` is active
--- a/docs/cli/error_handling.md
+++ b/docs/cli/error_handling.md
@@ -142,6 +142,8 @@ However, until then, we get a lot of leverage in the CLI from `FatalError`, so i
 
 When a command runs with JSON output active (`--json` or `-j`) and fails with a `FatalError`, the CLI writes a machine-readable error document instead of the human-readable banner. This is a public contract: scripts parse it, so treat changes to it as breaking.
 
+The shape of the document depends on `type`. Here is a typical `abort`, produced when the user can fix the problem themselves:
+
 ```json
 {
   "error": {
@@ -149,13 +151,37 @@ When a command runs with JSON output active (`--json` or `-j`) and fails with a 
     "message": "Couldn't find the app's configuration file",
     "tryMessage": "Run shopify app config link to create one",
     "nextSteps": ["Check that you're in the right directory"],
-    "customSections": [{"title": "Extensions", "body": [["name", "status"], ["my-ext", "failed"]]}],
-    "command": "npm",
-    "args": ["install"],
+    "customSections": [{"title": "Extensions", "body": [["name", "status"], ["my-ext", "failed"]]}]
+  }
+}
+```
+
+A `bug` is the only type that carries `stack`, since it is the only type where users are asked to file a report:
+
+```json
+{
+  "error": {
+    "type": "bug",
+    "message": "Unexpected error while parsing the theme manifest",
     "stack": "Error: ...\n    at ..."
   }
 }
 ```
+
+An `external` document additionally carries `command` and `args`, naming the subprocess that failed:
+
+```json
+{
+  "error": {
+    "type": "external",
+    "message": "npm install exited with a non-zero status",
+    "command": "npm",
+    "args": ["install"]
+  }
+}
+```
+
+`tryMessage`, `nextSteps`, and `customSections` are optional on any of the three types; they are simply omitted above where a real document wouldn't include them.
 
 ### The envelope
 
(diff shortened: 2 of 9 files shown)


amcaplan:
Closing in favor of #8405

PR #8190, opened 2026-07-28
Recover on theme commands after a preview store has been claimed

amcaplan:

PR #7986 made `shopify store info` and `store execute` fail actionably when run against a preview store that has since been claimed through the browser claim flow — the stored session stays `kind: 'preview'` forever, the CLI has no local signal of the claim, and the stale Admin token starts returning 401. Theme commands never got that treatment: they reported an opaque failure, or worse, "Development theme #X could not be found" for what was actually an authentication problem.

Ports the same recovery to `theme` commands, and fixes two things that stopped the recovery from working at all.

- *The advice #7986 prints could never be followed.* `throwIfPreviewStore` aborts `store auth` for any store that still has a stored preview session, and the recovery deliberately retained that session — so the command every one of these errors points at refused to run, with no reset flag to escape. `store info` was the last path still broken, because its preview branch returns before any Admin call and never reached the shared helper. Invalid preview sessions are now cleared, which is what lets the suggested command start. This reverses a deliberate choice from #7986; the trade-off is that preview credentials are write-once, so clearing is irreversible — but a preview token never expires, so a 401 on one means it stopped being accepted outright and nothing usable is discarded.
- *Theme commands can't safely condemn a standard session.* `store` reaches a 401 only after `loadStoredStoreSession` has already refreshed an expired token, so there the status is definitive. Theme uses the stored token as-is, so an ordinary expired-but-refreshable session 401s for a reason re-authenticating wouldn't fix — and clearing it would destroy a usable refresh token. Theme therefore classifies preview sessions only and never clears.

Also included: the Admin transport stops destroying the HTTP status (`AdminApiRequestError`), `fetchTheme` stops reporting a 401 as a missing theme, and multi-environment failures stop dropping the `nextSteps` that make these errors actionable.

**Known follow-up (not in this PR):** theme commands still read stored tokens with no expiry check or refresh, unlike `store`'s `loadStoredStoreSession`. That gap is why theme classifies preview-only. Exposing a refresh-aware lifecycle at the cli-kit layer would let theme handle standard sessions properly too.

### How to test your changes?

`packages/store/src/cli/services/store/auth/preview-claim-recovery.test.ts` is the key one — it composes both halves against real `LocalStorage` in a temp directory: `store auth` refuses to start, the 401 is classified and the session cleared, then `store auth` gets past the guard.

Full suite green: 266 files / 2764 passed / 2 skipped, plus type-check, lint and knip.

### Post-release steps

None.

### Checklist


dmerand:
Closing, superceded by https://github.com/Shopify/cli/pull/8390

amcaplan:
Closing as this PR has fallen out of date, and other PRs have taken up the mantle.

PR #8192, opened 2026-07-28
Improve theme list ordering: sort newest to oldest within each role group

krjo:
### WHY are these changes introduced?

Theme lists shown by the CLI (`theme list` and the interactive theme picker used
by `theme push`, `theme pull`, `theme open`, etc.) display themes in the order
returned by the Admin API, which is creation order — oldest first. On stores with
many themes (e.g. 100+), the newest themes — the ones developers actually need to
see or push to — are at the very bottom, requiring a lot of scrolling and wasted time.

### WHAT is this pull request doing?

Sorts theme lists newest to oldest, while keeping the existing role grouping
(live → unpublished → development).

All theme list displays funnel through `fetchStoreThemes` in
`packages/theme/src/cli/utilities/theme-selector/fetch.ts`, which previously
sorted only by role and let the API's oldest-first order leak through within each
group. The comparator (`byRole` → `byRoleThenNewest`) now adds an ID-descending
tiebreaker — theme IDs increase monotonically, so a higher ID means a newer
theme. No API or schema changes needed.

This changes ordering in one place for both the `theme list` output and every
interactive theme picker (`push`, `pull`, `open`, `publish`, `rename`, `delete`,
`duplicate`, `info`).

### How to test your changes?

1. Run `pnpm shopify theme list --store <store>` against a store with several
   themes — themes should appear newest first within each role group, with the
   live theme at the top.
2. Run `pnpm shopify theme open --store <store>` without `--theme` — the picker
   should list themes newest first within each role group.
3. Unit tests: `pnpm vitest run src/cli/utilities/theme-selector/fetch.test.ts`
   from `packages/theme` — includes a new test asserting newest-first order
   within role groups.

### Checklist

- [x] I've considered possible cross-platform impacts (Mac, Linux, Windows)
- [x] I've considered possible [documentation](https://shopify.dev) changes
- [x] I've considered analytics changes to measure impact
- [x] The change is user-facing — I've identified the correct [bump type](https://github.com/Shopify/cli/blob/main/CONTRIBUTING.md#choosing-the-right-bump-type) (`patch` for bug fixes · `minor` for new features · `major` for [breaking changes](https://github.com/Shopify/cli/blob/main/CONTRIBUTING.md#what-counts-as-a-breaking-change)) and added a changeset with `pnpm changeset add`

PR #8194, opened 2026-07-29
Add theme ID search to interactive theme pickers

zacariec:

Interactive theme pickers already support type-to-search by theme name and role, but theme IDs aren't shown or searchable. Developers who copy an ID from `theme list` therefore can't use it to find a theme in the pickers used by `theme push`, `theme pull`, `theme open`, and related commands.

This complements #8192: that PR improves the default ordering for large theme libraries, while this PR makes specific themes directly searchable by ID.

- Renders the existing autocomplete choice `helperText` beside labels in a dimmed style.
- Highlights matching text inside both labels and helper text, so ID matches are visually explained.
- Extends the default autocomplete search to match `helperText` while preserving existing name and group matching.
- Adds each theme's `#`-prefixed ID as helper text, matching the format shown by `theme list`.
- Preserves the existing instant, in-memory search behavior and result ordering.

A theme row now reads like `Dawn #123456789`, with the ID visually subdued unless it contains the active search match.

### How to test your changes?

1. Run `pnpm shopify theme open --store <store>` against a store with more than five themes.
2. Confirm each picker row shows a dimmed `#<theme-id>` after the theme name.
3. Type part of a theme name and confirm name filtering and highlighting still work.
4. Type or paste a theme ID, with and without its `#` prefix, and confirm the matching theme remains and the matching ID text is highlighted.
5. Run the focused automated checks:

```sh
cd packages/cli-kit
pnpm vitest run src/private/node/ui/components/SelectInput.test.tsx src/private/node/ui/components/AutocompletePromptSearch.test.ts src/private/node/ui/components/AutocompletePrompt.test.tsx
pnpm lint
pnpm type-check
pnpm build

cd ../theme
pnpm vitest run src/cli/utilities/theme-selector.test.ts
pnpm lint
pnpm type-check
pnpm build
```

### Checklist

- [x] The change is user-facing — I've identified the correct [bump type](https://github.com/Shopify/cli/blob/main/CONTRIBUTING.md#choosing-the-right-bump-type) (`minor`) and added a changeset

### DIFF
--- a/.changeset/search-themes-by-id.md
+++ b/.changeset/search-themes-by-id.md
@@ -1,0 +1,6 @@
+---
+'@shopify/cli-kit': minor
+'@shopify/theme': minor
+---
+
+Render searchable autocomplete helper text and show theme IDs in interactive theme pickers.
--- a/packages/cli-kit/src/private/node/ui/components/SelectInput.tsx
+++ b/packages/cli-kit/src/private/node/ui/components/SelectInput.tsx
@@ -87,6 +87,7 @@ function Item<T>({
   index,
 }: ItemProps<T>): React.ReactElement {
   const label = highlightedLabel(item.label, highlightedTerm)
+  const helperText = item.helperText ? highlightedLabel(item.helperText, highlightedTerm) : undefined
   let title: string | undefined
   let labelColor
 
@@ -120,6 +121,7 @@ function Item<T>({
         <Text wrap="end" color={labelColor}>
           {showKey ? `(${item.key}) ${label}` : label}
         </Text>
+        {helperText ? <Text dimColor>{` ${helperText}`}</Text> : null}
       </Box>
     </Box>
   )
--- a/packages/theme/src/cli/utilities/theme-selector.ts
+++ b/packages/theme/src/cli/utilities/theme-selector.ts
@@ -1,7 +1,7 @@
 import {Filter, FilterProps, filterThemes} from './theme-selector/filter.js'
 import {fetchStoreThemes} from './theme-selector/fetch.js'
 import {getDevelopmentTheme} from '../services/local-storage.js'
-import {renderAutocompletePrompt} from '@shopify/cli-kit/node/ui'
+import {renderAutocompletePrompt, type RenderAutocompleteOptions} from '@shopify/cli-kit/node/ui'
 import {AdminSession} from '@shopify/cli-kit/node/session'
 import {capitalize} from '@shopify/cli-kit/common/string'
 import {themeCreate} from '@shopify/cli-kit/node/themes/api'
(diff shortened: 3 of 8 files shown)


packages/theme/src/cli/utilities/theme-selector.ts
graygilmore, line 1: I like this idea! But I don't like that the feature is invisible to users. You can search by ID, but the picker never shows IDs, so the only people who'll use this are people who already know it exists. It also means there's no feedback when a match happens (you paste an ID, one row survives, and nothing on screen explains why).

My suggestion here would be to show the ID in each row, dimmed so it reads as metadata instead of competing with the theme name:

| Name matching | ID matching |
| -- | -- |
| Image | Image |

`Item<T>` already has a `helperText?: string` field that we can leverage by including 

```js
helperText: `#${theme.id}`,
```

in the object that `findOrSelectTheme` returns. Then in `SelectInput` it's a matter of rendering that text and showing highlighting on it for any matches.

graygilmore (changes requested):

packages/theme/src/cli/utilities/theme-selector.ts
zacariec: Addressed in 03bb150fd2. Theme choices now set the existing `helperText` field to the `#<theme-id>` value. `SelectInput` renders helper text dimmed, and the active search term is highlighted in it. The default autocomplete search now matches helper text, so both raw and `#`-prefixed IDs work. I also added focused rendering/highlighting and filtering tests.

packages/cli-kit/src/private/node/ui/components/SelectInput.tsx
```diff
@@ -120,6 +121,7 @@ function Item<T>({
         <Text wrap="end" color={labelColor}>
           {showKey ? `(${item.key}) ${label}` : label}
         </Text>
+        {helperText ? <Text dimColor>{` ${helperText}`}</Text> : null}
```
graygilmore, line 124: This isn't the most common case but since the label has `wrap="end"` if we have the helper text outside of that the text could overlap. Can fix this by putting the helper text inside of that `Text` component:

```tsx
<Text wrap="end" color={labelColor}>
  {showKey ? `(${item.key}) ${label}` : label}
  {helperText ? <Text dimColor>{` ${helperText}`}</Text> : null}
</Text>
```

.changeset/search-themes-by-id.md
```diff
@@ -0,0 +1,6 @@
+---
+'@shopify/cli-kit': minor
+'@shopify/theme': minor
+---
+
+Render searchable autocomplete helper text and show theme IDs in interactive theme pickers.
```
isaacroldan, line 6: let's rewrite this changeset in a way that's meaningful for our users :)

isaacroldan (approved):

packages/cli-kit/src/private/node/ui/components/SelectInput.tsx
```diff
@@ -87,6 +87,7 @@ function Item<T>({
   index,
 }: ItemProps<T>): React.ReactElement {
   const label = highlightedLabel(item.label, highlightedTerm)
+  const helperText = item.helperText ? highlightedLabel(item.helperText, highlightedTerm) : undefined
```
graygilmore, line 90: I'm not sure if this is working quite right:

Image

Note how the highlight on the ID for the `Big files` theme stops having the dimmed color after the bold?

PR #8198, opened 2026-07-29
Limit SHOPIFY_ environment variables and credential flags in analytics

isaacroldan:

The analytics payload is printed by `outputDebug`, including under `--verbose` and when analytics are disabled, so credential values shouldn't be in it. Two gaps let them through, plus a bug in the redaction itself:

- `sanitizePayload()` only knew about `shptka_` tokens and `--store-password`, so flags like `--client-secret` passed through, and so did the environment variables the CLI reads its own tokens from.
- It rewrites the already-serialised payload and parses it back, and its patterns didn't account for JSON escaping. `--store-password ""` unbalanced the string, the closing `JSON.parse` threw, and the whole event was dropped. Credential keys nested inside `metadata` were missed for the same reason.

Widening what the sanitizer redacts. Nothing is dropped from the payload that `main` reports today, and `getShopifyEnvironmentVariables` is untouched.

Two passes over the payload, and one extra over the copies that get printed.

`sanitizePayload` matches any flag or key whose name contains `password`, `token`, `secret` or `credential` — covering `args`, a command line quoted into `error_message`, and keys at any nesting depth inside `metadata`, `cmd_all_environment_flags` and `env_shopify_variables`. This replaces the two `--store-password` rules, which it is a superset of. Values are matched escape-aware so a replacement can't cut a `\"` in half and leave the payload unparseable — `--store-password ""` used to do exactly that, and the event was dropped.

`auth` is excluded from the name matching on purpose, so `env_auth_method` keeps reporting.

`key` is handled separately, in `private/node/analytics/redact-output.ts`. It marks a credential in the environment — `SHOPIFY_PROXY_KEY` holds a signed token and `SHOPIFY_FLAG_GRAPHIQL_KEY` is derived from the app secret — but `api_key` is an app's public client ID that Monorail is meant to receive. So the name is redacted only in the copies `outputDebug` prints, which is what `--verbose` turns on. Telemetry keeps the values; a terminal, and whatever scrapes it, does not see them.

Still best effort. A credential whose name gives no hint is not redacted, and neither is a non-string value like `{"password": 12345}`, since the key rule expects a quoted value.

#8205 stacks on this and notes in the `--verbose` help that the output may include sensitive data.

### How to test your changes?

`pnpm test packages/cli-kit/src/public/node/analytics.test.ts`

```bash
SHOPIFY_FLAG_VERBOSE=1 SHOPIFY_CLI_PARTNERS_TOKEN=atkn_fake_value \
  node packages/cli/bin/run.js version --verbose --client-secret cs_fake_value
```

`env_shopify_variables` shows `SHOPIFY_CLI_PARTNERS_TOKEN` with its value as `*****`, and `args` shows `--client-secret *****`.

### Checklist

- [x] The change is not user-facing — internal hardening of what the analytics payload reports, so no changeset

### DIFF
--- a/.changeset/analytics-redact-shopify-env.md
+++ b/.changeset/analytics-redact-shopify-env.md
@@ -2,4 +2,4 @@
 '@shopify/cli-kit': patch
 ---
 
-Exclude credential-shaped environment variables and redact credential flag values in CLI analytics
+Redact credential environment variables and flag values in CLI analytics
--- a/.changeset/analytics-redact-shopify-env.md
+++ b/.changeset/analytics-redact-shopify-env.md
@@ -2,4 +2,4 @@
 '@shopify/cli-kit': patch
 ---
 
-Only include allowlisted SHOPIFY_ environment variables and redact credential flag values in CLI analytics
+Exclude credential-shaped environment variables and redact credential flag values in CLI analytics
--- a/.changeset/analytics-redact-shopify-env.md
+++ b/.changeset/analytics-redact-shopify-env.md
@@ -1,5 +1,0 @@
----
-'@shopify/cli-kit': patch
----
-
-Redact credential environment variables and flag values in CLI analytics
--- a/.changeset/analytics-redact-shopify-env.md
+++ b/.changeset/analytics-redact-shopify-env.md
@@ -1,0 +1,5 @@
+---
+'@shopify/cli-kit': patch
+---
+
+Only include allowlisted SHOPIFY_ environment variables and redact credential flag values in CLI analytics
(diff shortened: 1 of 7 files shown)


graygilmore (approved):
Makes sense!

isaacroldan:
* **#8198** <a href="https://app.graphite.com/github/pr/Shopify/cli/8198?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Shopify/cli/8198?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #8199, opened 2026-07-29
Halve CLI startup time by not loading the app for non-app commands

isaacroldan:
## What

`shopify version` goes from **369 ms → 180 ms (−51%)**. The win applies to every command that isn't an `app` command.

| | before | after | Δ |
|---|---|---|---|
| `version` total | 369 ms | **180 ms** | −189 ms (−51%) |
| `version` tail after output | 199 ms | 8 ms | −191 ms |
| `version` time to first output | 222 ms | 213 ms | −9 ms |
| `app --help` time to first output | 182 ms | 172 ms | −10 ms |
| modules loaded / bytes read | 112 / 8.5 MB | 90 / 6.5 MB | −22 / −2.0 MB |

Measured on `bin/run.js` (the production entry point — `bin/dev.js` sets `SHOPIFY_CLI_ENV=development`, which disables analytics and takes a different path), with `HOME` pointed at a throwaway dir so the 300/day analytics rate limit couldn't skew repeats, 15 runs, load average under 4.

## Why

The dominant cost was the `public_command_metadata` hook. oclif fans it out on **every** command via `analytics.ts`, and `@shopify/app`'s implementation statically imported `localAppContext` — dragging in the whole app graph, including `@shopify/theme-check-common`, whose **ohm-js Liquid grammar is compiled from source at module-evaluation time**. That was ~190 ms of the 199 ms that ran *after* the command had already printed its output.

Making the import dynamic is **not** sufficient: the hook always calls `localAppContext()`, so the import happens anyway. It now gates on `getCurrentCommandId()`, which cli-kit's `BaseCommand` already sets to the canonical oclif id — so no hook signature or fanout plumbing had to change.

## Also included

- **analytics** — check the `analyticsDisabled`/`alwaysLog*` gates *before* `buildPayload()`, so nothing is built when neither destination can receive it. Guarded with `isVerbose()` so `--verbose` still logs the would-be payload.
- **base-command** — import `session.js` on demand, only when `--auth-alias` is passed (`setCurrentSessionAlias(undefined)` returns immediately). `commandSessionId` moves into a new dependency-free `private/node/session/command-session-id.ts` so clearing the selection stays correct without loading the session/identity/API graph. This is the ~10 ms pre-output gain.
- **otel** — `unref()` the throttle timer. It doesn't fire today (`onForceFlush` is called once per process), but a second call within the window would pin the process for up to 5 s.
- **notifications** — skip the background refresh, which spawns a whole extra CLI process, when the cached payload is under an hour old. Previously it spawned on every non-CI command with no freshness check.
- **app-init** — skip the fsync'd atomic write when the store is already empty.

## Trade-off to review

**Non-app commands no longer report `app_*` analytics fields.** They couldn't produce meaningful ones without an app in the working directory, but a non-app command run *inside* an app directory now loses them too. Worth a look from whoever owns those dashboards.

## Testing

- 3790 unit tests pass across `cli-kit`, `app`, `cli`
- `type-check` and `lint` pass across all 13 projects (CI-equivalent, `--skip-nx-cache`)
- `knip` clean; `refresh-code-documentation` produces no churn
- Bundle builds; smoke-tested `version`, `--version`, `app/theme/store/auth/config --help`, `app build --help`, `--verbose`, and `--auth-alias <bad>` (still errors correctly through the lazy import)

Two test files timed out on full-suite runs — different files on different runs, both pass in isolation, both slow timing-based suites (16 s and 26 s of test time against a 5 s per-test limit). Pre-existing flakes.

## Unrelated bug found while measuring

`packages/cli/project.json` declares `build` with `outputs: ["{workspaceRoot}/dist"]` (should be `{projectRoot}/dist`) and `bundle` with no `inputs`/`outputs` at all. nx consequently serves stale bundles: with these changes fully stashed, `dist` still contained the compiled new code. Every number above comes from `rm -rf dist && nx bundle cli --skip-nx-cache`. Not fixed here — it deserves its own PR since touching nx caching could affect CI.

## Dead end worth recording

Stubbing out 1.5 MB of `graphql` + OpenTelemetry + Bugsnag chunks saved **−1.4 ms**. Bytes loaded are nearly free — V8 lazy-parses and the CJS bodies never execute. Only module bodies that actually *run* cost anything, which is why one ohm-js grammar compile outweighed 2 MB of dependencies. Same for the 3.5 MB TypeScript compiler chunk (pulled in by `@oclif/core`'s guarded `require('typescript')`): never executed, ~2 ms. The remaining ~213 ms of pre-output latency lives in `base-command`'s import graph, not in bundle size.

### DIFF
--- a/packages/app/src/cli/hooks/public_metadata.ts
+++ b/packages/app/src/cli/hooks/public_metadata.ts
@@ -1,14 +1,29 @@
 import metadata from '../metadata.js'
-import {localAppContext} from '../services/app-context.js'
 import {FanoutHookFunction} from '@shopify/cli-kit/node/plugins'
 import {cwd} from '@shopify/cli-kit/node/path'
+import {getCurrentCommandId} from '@shopify/cli-kit/node/global-context'
 
 const APP_CONTEXT_METADATA_TIMEOUT_MS = 3000
+
+/**
+ * Loading an app to gather `app_*` analytics only makes sense for `app` commands. Every other
+ * command (`version`, `theme *`, `store *`, ...) would load the whole app graph — which reaches
+ * theme-check and its ohm-js Liquid grammar — to report metadata it can't produce anyway.
+ *
+ * The command id is the canonical oclif id (`app:dev`), set by cli-kit's BaseCommand. It is empty
+ * for commands that don't extend BaseCommand, which are never app commands.
+ */
+function isAppCommand(commandId: string): boolean {
+  return commandId === 'app' || commandId.startsWith('app:')
+}
 
 async function logAppContextMetadata(directory: string): Promise<void> {
   let timer: ReturnType<typeof setTimeout> | undefined
   try {
     if (metadata.getAllPublicMetadata().api_key !== undefined) return
+
+    // Imported lazily so that non-app commands never pay for the app graph.
+    const {localAppContext} = await import('../services/app-context.js')
 
     await Promise.race([
       localAppContext({directory, skipPrompts: true}),
@@ -25,7 +40,9 @@ async function logAppContextMetadata(directory: string): Promise<void> {
 }
 
 const gatherPublicMetadata: FanoutHookFunction<'public_command_metadata', '@shopify/app'> = async () => {
-  await logAppContextMetadata(cwd())
+  if (isAppCommand(getCurrentCommandId())) {
+    await logAppContextMetadata(cwd())
+  }
   return metadata.getAllPublicMetadata()
 }
 
--- a/.changeset/faster-cli-startup.md
+++ b/.changeset/faster-cli-startup.md
@@ -1,0 +1,7 @@
+---
+'@shopify/cli-kit': patch
+'@shopify/app': patch
+'@shopify/cli': patch
+---
+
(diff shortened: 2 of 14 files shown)


packages/app/src/cli/hooks/public_metadata.ts
```diff
@@ -1,15 +1,30 @@
...
+  return commandId === 'app' || commandId.startsWith('app:')
+}
+
 async function logAppContextMetadata(directory: string): Promise<void> {
   let timer: ReturnType<typeof setTimeout> | undefined
   try {
     if (metadata.getAllPublicMetadata().api_key !== undefined) return
 
+    // Imported lazily so that non-app commands never pay for the app graph.
+    const {localAppContext} = await import('../services/app-context.js')
```
graygilmore, line 26: If non-app commands don't call `logAppContextMetadata` at all is this part necessary?

isaacroldan: I was just experimenting on ways to reduce the CLI startup time, but you can ignore this PR for now 🙏

PR #8216, opened 2026-07-30
Log the reason a token exchange fails in verbose output

wes-shaw:

When an app automation token exchange fails, the CLI prints `The custom token provided can't be used for the App Management API.` with nothing about why. Identity's response carries an OAuth `error` code and an `error_description` that provides more context but the CLI discards both. Anyone debugging a CI failure has to guess, or find someone with access to Identity's logs.

All of these cases resolve to the same user action (create a fresh token), so the default error output does not need to change. The detail belongs in `--verbose` logs, where app developers and support can read it without Identity-side access. That is the full scope of this PR: default output stays byte-identical to `main`.

Adds two debug lines, both visible only with `--verbose`:

1. When Identity returns an OAuth failure body, `tokenRequest` logs it: `Token request to Identity failed with status 400: invalid_grant - <description>`. The description is flattened to a single line and capped at 200 characters.
2. When the app automation token exchange throws with no OAuth body at all (for example, an unreachable Identity host), the catch block logs the caught error's message, so `--verbose` always carries a reason.

It also fixes two pre-existing bugs on `main` that surfaced while testing this path. Their fixes are interleaved with the logging commits, so they are included here (with a note in each changeset) rather than split into a separate PR:

- Device authorization polling looped forever when Identity returned an error code outside the known set; it now fails with an `AbortError`.
- An OAuth failure body with a missing or non-string `error` field produced `AbortError(undefined)`; it now normalizes to `unknown_error`.

One suggestion from the internal discussion was a `Re-run with --verbose` pointer in the default error text. It is deliberately left out to keep the default output unchanged; happy to add it in a follow-up if reviewers want it.

### How to test your changes?

- Unit tests: from `packages/cli-kit`, run `pnpm exec vitest run src/private/node/session/exchange.test.ts src/private/node/session/device-authorization.test.ts` (46 tests cover the failure paths above).
- Manual: set `SHOPIFY_APP_AUTOMATION_TOKEN` to an expired or revoked token and run `shopify app deploy --verbose`. The debug output includes the `Token request to Identity failed with status 400: ...` line, and the final error banner is unchanged from `main`.

### Checklist

### DIFF
--- a/.changeset/device-auth-unknown-error-hang.md
+++ b/.changeset/device-auth-unknown-error-hang.md
@@ -1,5 +1,0 @@
----
-'@shopify/cli-kit': patch
----
-
-Fail with an error instead of hanging when device login polling receives an unrecognized error code
--- a/.changeset/device-auth-unknown-error-hang.md
+++ b/.changeset/device-auth-unknown-error-hang.md
@@ -1,0 +1,5 @@
+---
+'@shopify/cli-kit': patch
+---
+
+Fail with an error instead of hanging when device login polling receives an unrecognized error code
--- a/packages/cli-kit/src/private/node/session/device-authorization.ts
+++ b/packages/cli-kit/src/private/node/session/device-authorization.ts
@@ -136,8 +136,6 @@ export async function pollForDeviceAuthorization(code: string, interval = 5): Pr
         case 'expired_token':
           reject(new AbortError(`Device authorization failed: Token expired. Please try again.`))
           return
-        // The default guards against codes outside IdentityDeviceError reaching this switch: without
-        // it the poll would neither settle nor reschedule, and the login would hang with no output.
         case 'unknown_failure':
         default:
           reject(new Error(`Device authorization failed: ${error}`))
--- a/packages/cli-kit/src/private/node/session/device-authorization.ts
+++ b/packages/cli-kit/src/private/node/session/device-authorization.ts
@@ -136,9 +136,11 @@ export async function pollForDeviceAuthorization(code: string, interval = 5): Pr
         case 'expired_token':
           reject(new AbortError(`Device authorization failed: Token expired. Please try again.`))
           return
-        case 'unknown_failure': {
+        // The default guards against codes outside IdentityDeviceError reaching this switch: without
+        // it the poll would neither settle nor reschedule, and the login would hang with no output.
+        case 'unknown_failure':
+        default:
           reject(new Error(`Device authorization failed: ${error}`))
-        }
       }
     }
 
--- a/packages/cli-kit/src/private/node/session/exchange.ts
+++ b/packages/cli-kit/src/private/node/session/exchange.ts
@@ -93,10 +93,8 @@ async function exchangeAppAutomationTokenForAccessToken(
     return {accessToken, userId}
   } catch (error) {
     const prettyName = apiName.replace(/-/g, ' ').replace(/\b\w/g, (char) => char.toUpperCase())
-    const reason = error instanceof Error ? error.message.slice(0, 200) : ''
-    if (reason !== '') {
-      outputDebug(`Token exchange for the ${prettyName} API failed: ${reason}`)
-    }
+    const reason = error instanceof Error ? error.message.slice(0, 200) : 'unknown error'
+    outputDebug(`Token exchange for the ${prettyName} API failed: ${reason}`)
     throw new AbortError(
       `The custom token provided can't be used for the ${prettyName} API.`,
       'Ensure the token is correct and not expired.',
--- a/packages/cli-kit/src/private/node/session/exchange.ts
+++ b/packages/cli-kit/src/private/node/session/exchange.ts
@@ -258,16 +258,24 @@ async function tokenRequest(params: Record<string, string>): Promise<Result<Toke
     const payload = JSON.parse(responseText)
     if (res.ok) return ok(payload)
 
+    // The responder isn't always Identity: proxies and gateways can answer with arbitrary JSON, so neither
+    // field is guaranteed to be present or well-formed. Normalize both here, before they reach the debug
+    // log or any user-facing message, so downstream code can rely on their shape.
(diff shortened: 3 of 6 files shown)


.changeset/device-auth-unknown-error-hang.md
```diff
@@ -0,0 +1,5 @@
+---
+'@shopify/cli-kit': patch
+---
+
+Fail with an error instead of hanging when device login polling receives an unrecognized error code
```
isaacroldan: This change doesn't need a changeset, is barely user facing. You can delete both changesets

packages/cli-kit/src/private/node/session/device-authorization.ts
```diff
@@ -136,9 +136,11 @@ export async function pollForDeviceAuthorization(code: string, interval = 5): Pr
         case 'expired_token':
           reject(new AbortError(`Device authorization failed: Token expired. Please try again.`))
           return
-        case 'unknown_failure': {
+        // The default guards against codes outside IdentityDeviceError reaching this switch: without
+        // it the poll would neither settle nor reschedule, and the login would hang with no output.
```
isaacroldan: No need for the comment, the behavior of "default" is obvious here :)

packages/cli-kit/src/private/node/session/exchange.ts
```diff
@@ -92,6 +93,13 @@ async function exchangeAppAutomationTokenForAccessToken(
     return {accessToken, userId}
   } catch (error) {
     const prettyName = apiName.replace(/-/g, ' ').replace(/\b\w/g, (char) => char.toUpperCase())
+    // When the request itself fails (e.g. the authentication service is unreachable) there is no OAuth
+    // body, so the debug line in tokenRequest never runs. Log the caught error here so --verbose always
+    // carries the reason for the failure.
+    const reason = error instanceof Error ? flattenAndTruncate(error.message) : ''
+    if (reason !== '') {
+      outputDebug(`Token exchange for the ${prettyName} API failed: ${reason}`)
+    }
```
isaacroldan: Better to always print something
```suggestion
   const reason = error instanceof Error ? flattenAndTruncate(error.message) : 'unknown reason'
   outputDebug(`Token exchange for the ${prettyName} API failed: ${reason}`)
```

I don't think we need a new `flattenAndTruncate` function, just do `error.message.slice(0, 200)` and that should be enough.

also, no need for the comment here either

packages/cli-kit/src/private/node/session/exchange.ts
```diff
@@ -131,7 +139,20 @@ export async function exchangeAppAutomationTokenForBusinessPlatformAccessToken(
   return exchangeAppAutomationTokenForAccessToken('business-platform', token, tokenExchangeScopes('business-platform'))
 }
 
-type IdentityDeviceError = 'authorization_pending' | 'access_denied' | 'expired_token' | 'slow_down' | 'unknown_failure'
+export type IdentityDeviceError =
```
isaacroldan: This is exported just for a test right? Can we avoid that?

packages/cli-kit/src/private/node/session/exchange.ts
```diff
@@ -131,7 +139,20 @@ export async function exchangeAppAutomationTokenForBusinessPlatformAccessToken(
...
+  | 'slow_down'
+  | 'unknown_failure'
+
+const identityDeviceErrors: IdentityDeviceError[] = [
+  'authorization_pending',
+  'access_denied',
+  'expired_token',
+  'slow_down',
+  'unknown_failure',
+]
```
isaacroldan: This feels like duplication, we can either derive the types from the const:

```ts
const identityDeviceErrors = [
  'authorization_pending',
  'access_denied',
  'expired_token',
  'slow_down',
  'unknown_failure',
] as const

export type IdentityDeviceError = (typeof identityDeviceErrors)[number]
```

Or just have a function to convert an error to IdentityError:

```ts
function toIdentityDeviceError(error: string): IdentityDeviceError {
  switch (error) {
    case 'authorization_pending':
    case 'access_denied':
    case 'expired_token':
    case 'slow_down':
    case 'unknown_failure':
      return error
    default:
      return 'unknown_failure'
  }
}
```
This way you don't need the extra const.

packages/cli-kit/src/private/node/session/exchange.ts
```diff
@@ -245,7 +281,23 @@ async function tokenRequest(
...
+    // field is guaranteed to be present or well-formed. Normalize both here, before they reach the debug
+    // log, so downstream code can rely on their shape.
+    const oauthError = typeof payload.error === 'string' && payload.error !== '' ? payload.error : 'unknown_error'
+    const rawDescription = typeof payload.error_description === 'string' ? payload.error_description : ''
+    const oauthDescription = flattenAndTruncate(rawDescription)
+
+    // Only logged for failed responses: a successful body contains the access token. An OAuth error body
+    // carries just `error` and `error_description`, and both are logged rather than the raw body so an
+    // unexpected payload can't leak into the debug output.
+    outputDebug(
```
isaacroldan, line 268: too many long comments in this file, they can be removed or if any of them you feel is really relevant, simplified

packages/cli-kit/src/private/node/session/exchange.ts
```diff
@@ -92,6 +93,10 @@ async function exchangeAppAutomationTokenForAccessToken(
     return {accessToken, userId}
   } catch (error) {
     const prettyName = apiName.replace(/-/g, ' ').replace(/\b\w/g, (char) => char.toUpperCase())
+    const reason = error instanceof Error ? error.message.slice(0, 200) : ''
+    if (reason !== '') {
+      outputDebug(`Token exchange for the ${prettyName} API failed: ${reason}`)
+    }
```
isaacroldan: Still think this should print always:

```typescript
const reason = error instanceof Error ? error.message.slice(0, 200) : 'unknown error'
outputDebug(`Token exchange for the ${prettyName} API failed: ${reason}`)
```

isaacroldan (approved):

Issue #6069, opened 2025-07-04
[Bug]: shopify app deploy fails with "Version couldn't be created" with theme and web pixel extension.


themazim:
### Please confirm that you have:

- [x] Searched [existing issues](.) to see if your issue is a duplicate. (If you’ve found a duplicate issue, feel free to add additional information in a comment on it.)
- [x] Reproduced the issue in [the latest CLI version](https://www.npmjs.com/package/@shopify/cli).

### In which of these areas are you experiencing a problem?

Extension, Other

### Expected behavior

When an app has 2 extensions (theme & web pixel) "shopify app deploy" shall be able to deploy both extensions without issue.

### Actual behavior

When both extensions are present, the deployment process fails and no extension is deployed with:

```
 error │                                                                                                                                                                                                                                                  │
│  Version couldn't be created.                                                                                                                                                                                                                    │
│                                                                                                                                                                                                                                                  │
│  There was an error creating an app version.
```

No additional information is given why it failed. 

### Verbose output

<details>
  <summary>shopify app deploy --verbose</summary>

  ```
$ shopify app deploy --verbose
2025-07-04T11:44:36.981Z: Running command app deploy
2025-07-04T11:44:36.990Z: Running system process in background:
  · Command: /usr/bin/node /usr/bin/shopify notifications list --ignore-errors
  · Working directory: /home/myuser/bug-test

2025-07-04T11:44:37.007Z: Notifications to show: 0
2025-07-04T11:44:37.028Z: Reading cached app information for directory /home/myuser/bug-test...
2025-07-04T11:44:37.030Z: Reading the content of file at shopify.app.toml...
2025-07-04T11:44:37.035Z: Reading the content of file at shopify.app.toml...
2025-07-04T11:44:37.040Z: Reading cached app information for directory /home/myuser/bug-test...
2025-07-04T11:44:37.041Z: Reading the content of file at shopify.app.toml...
2025-07-04T11:44:37.043Z: Reading the content of file at shopify.app.toml...
2025-07-04T11:44:37.046Z: Ensuring that the user is authenticated with the Partners API with the following scopes:
[]

2025-07-04T11:44:37.046Z: Getting session store...
2025-07-04T11:44:37.048Z: Validating existing session against the scopes:
[
  "openid",
  "https://api.shopify.com/auth/shop.admin.graphql",
  "https://api.shopify.com/auth/shop.admin.themes",
  "https://api.shopify.com/auth/partners.collaborator-relationships.readonly",
[...]

PR #8236, opened 2026-08-03
Fix app deploy including stale dist files in the extension bundle

alfonso-noriega:

Related to #6069.

Since 3.94.0 (0986095dfb), `executeBundleUIStep` builds the UI extension into the extension's local `dist/` directory and then copies the **whole directory** into the deploy bundle. Any stale file sitting in `dist/` (a script built under an old handle, a dev's own bundler output, old sourcemaps/metafiles) ships inside the module.

For web pixel extensions this is fatal: core's `FetchScriptContent` rejects modules whose uploaded files contain more than one `.js`, so `appVersionCreate` fails with a generic "Version couldn't be created" even though every prior step (build, GCS upload) succeeds. `dist/` is gitignored, so reverting source changes doesn't clear it — which is why "the same source deployed fine before and now fails".

Root cause analysis in [this Slack thread](https://shopify.slack.com/archives/C07UJ7UNMTK/p1785155330271909). Server-side error message improvement in https://github.com/shop/world/pull/967972.

- `bundle-ui-step.ts`: instead of copying the entire local output directory, copy only the artifacts produced by the build — `<output-name>.*` for the main entry and each extra asset (which also covers their sourcemaps and `.metafile.json`). Stale files next to the built ones no longer ship.
- Adds two tests: a stale `.js` in the local `dist/` is excluded from the bundle; extra built assets (e.g. `-conditions.js`) are still copied.

### How to test your changes?

1. Create an app with a web pixel extension and deploy it once.
2. Rename the extension handle in `shopify.extension.toml` (leaving the old `extensions/<pixel>/dist/<old-handle>.js` in place) and run `shopify app deploy`.
3. Without this fix, `appVersionCreate` fails ("Version couldn't be created"); with it, only the freshly built script is included and the deploy succeeds.

### Checklist

### DIFF
--- a/.changeset/fix-stale-dist-files-in-deploy-bundle.md
+++ b/.changeset/fix-stale-dist-files-in-deploy-bundle.md
@@ -1,0 +1,5 @@
+---
+'@shopify/app': patch
+---
+
+Fix `app deploy` including stale files from the extension's dist directory in the uploaded bundle
--- a/packages/app/src/cli/services/build/steps/bundle-ui-step.test.ts
+++ b/packages/app/src/cli/services/build/steps/bundle-ui-step.test.ts
@@ -3,6 +3,7 @@ import * as generateManifest from './include-assets/generate-manifest.js'
 import * as buildExtension from '../extension.js'
 import {BundleUIStep, BuildContext} from '../client-steps.js'
 import {ExtensionInstance} from '../../../models/extensions/extension-instance.js'
+import {AssetIdentifier} from '../../../models/extensions/specification.js'
 import {describe, expect, test, vi, beforeEach} from 'vitest'
 import {inTemporaryDirectory, mkdir, writeFile, fileExists} from '@shopify/cli-kit/node/fs'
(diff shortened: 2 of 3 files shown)


isaacroldan:
The fix restores the pre-3.94 contract (only build artifacts reach the bundle), but two risks with the glob approach:

1. **The glob can drift from what esbuild actually emits.** Today `outfile` with no splitting/file loaders means every output is named `<entry-name>.*`, so the enumeration is complete. If we ever enable code splitting, a `file` loader, or CSS extraction, those outputs silently disappear from the bundle and the extension breaks at runtime. The tests mock `buildUIExtension`, so nothing would catch it.

2. **The glob is prefix-loose.** `handle.*` also matches stale files like `handle.old.js`, which would still ship and still fail a web pixel deploy.

**Recommendation:** copy the exact file list from esbuild's build result instead of globbing — `bundleExtension` already generates a metafile in production, and `result.metafile.outputs` lists every file written. Same cost, and both risks go away.

PR #8238, opened 2026-08-03
Keep the terminal UI runtime out of the CLI startup path

isaacroldan:
### Why

Every CLI invocation loads common output code. That code imported a small text-formatting helper from a React and Ink component, so even `shopify --version` loaded the terminal UI runtime. The error path also loaded `graphql-request` for one type check.

This work made startup slower before a command could run.

### What

- Move the token types and pure formatting helpers into a dependency-free module.
- Keep React and Ink imports in rendering modules.
- Replace the eager `ClientError` import with a structural check that preserves the existing expected-error cases.

No output changes are intended.

### Impact

On clean bundled builds with the Nx cache disabled:

- `shopify --version`: 174 ms to 99 ms, with 52% fewer instructions.
- `shopify --help`: 50% fewer instructions.
- JavaScript loaded at startup: 5.31 MB to 4.16 MB.

### Testing

- Type-check, lint, Knip, bundle, unit, and E2E checks pass.
- Output is unchanged for `--version`, `--help`, `app --help`, and `commands`.
- Existing tests use real `ClientError` instances and pass with the structural check.

Context: https://shopify.slack.com/archives/C0AG0L37Q4C/p1785681072538939


gonzaloriestra (approved):
Nice! Simple change, good improvement :ok_hand:

PR #8239, opened 2026-08-03
Fix Nx caching configuration

isaacroldan:
## Why

Nx caching was not effective or safe to rely on:

- normal local commands and CI jobs passed `--skip-nx-cache`, so unchanged tasks ran again
- the global cache input included the current Git SHA, which invalidated every task after each commit
- package builds declared `{workspaceRoot}/dist`, but they write to `{projectRoot}/dist`; Nx therefore could not restore the correct build output
- shared configuration, tool versions, custom ESLint rules, and some GraphQL generators were missing from the cache input model

## What improves

- repeated build, bundle, lint, type-check, and E2E prerequisite tasks can restore unchanged work from the Nx cache
- a cached CLI rebuild measured about 0.5 seconds instead of 3 seconds locally
- cache entries remain reusable across commits when their real inputs do not change
- cached package builds restore the correct package-local `dist` output
- changes to shared configuration, dependencies, Node, pnpm, custom lint rules, and GraphQL inputs invalidate the applicable tasks
- GraphQL generation and the UI extensions server build are cached as complete units

## What changed

- replace the legacy task-runner cache configuration with explicit cached targets and shared inputs
- correct package output declarations
- stop bypassing the cache in normal scripts and CI
- remove commit-based cache invalidation
- complete the GraphQL generator inputs and simplify ownership of the UI extensions server build output

## Validation

- `pnpm nx sync:check`
- repeated `pnpm build`: 12/12 tasks restored from cache
- repeated `pnpm nx run-many --target=bundle --all`: 12/12 tasks restored from cache
- repeated schema refresh and `pnpm graphql-codegen`: restored from cache
- `pnpm lint`
- `pnpm type-check`
- `pnpm test`
- JSON and workflow YAML parsing
- `git diff --check`


amcaplan (approved):

karreiro (approved):
Thank you, @isaacroldan!

PR #8240, opened 2026-08-03
Allowlist Shopify environment variables in analytics

isaacroldan:
## Why

`getShopifyEnvironmentVariables()` collected every `SHOPIFY_*` variable. This could send arbitrary credentials to Monorail and show them in verbose output.

The collector was originally added to track `shopify-function-test-helpers` through `SHOPIFY_INVOKED_BY`.

## What

- Collect `SHOPIFY_INVOKED_BY` and the five `SHOPIFY_CLI_AGENT*` telemetry variables.
- Exclude all other `SHOPIFY_*` variables.
- Keep the existing sensitive analytics payload in verbose output for debugging.

## Testing

- `pnpm vitest run packages/cli-kit/src/public/node/analytics.test.ts`
- `pnpm eslint packages/cli-kit/src/private/node/analytics.ts packages/cli-kit/src/public/node/analytics.test.ts`
- `pnpm nx bundle cli --skip-nx-cache`
- Verified that verbose output includes sensitive arguments, `SHOPIFY_INVOKED_BY`, and the allowlisted agent variable, but excludes `SHOPIFY_SOMETHING_KEY=123`.
- `git diff --check`

### DIFF
--- a/packages/cli-kit/src/private/node/analytics.ts
+++ b/packages/cli-kit/src/private/node/analytics.ts
@@ -72,6 +72,15 @@ interface EnvironmentData {
   env_auto_upgrade_enabled: boolean | null
 }
 
+const allowedShopifyEnvironmentVariableNames = new Set([
+  'SHOPIFY_INVOKED_BY',
+  'SHOPIFY_CLI_AGENT',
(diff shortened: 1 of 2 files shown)


dmerand (approved):
Code change LGTM!

PR #8256, opened 2026-08-06
theme: overlap checksum fetch with file system setup

Arjunmehta312:
## Summary

Start mounting the local theme file system before fetching remote checksums during `theme push`, then wait for both operations together.

This matches the existing pattern used by `theme pull` and allows the two independent operations to overlap while preserving the existing behavior.

## Testing

- Updated unit tests (only if needed)
- Relevant tests pass
- Lint passes
- Typecheck passes

### DIFF
--- a/packages/theme/src/cli/services/push.ts
+++ b/packages/theme/src/cli/services/push.ts
@@ -223,11 +223,11 @@ async function executePush(
   context?: {stdout?: Writable; stderr?: Writable},
 ) {
   recordTiming('theme-service:push:file-system')
-  const themeChecksums = await fetchChecksums(theme.id, session)
   const themeFileSystem = mountThemeFileSystem(options.path, {
     filters: options,
     listing: options.listing,
   })
+  const [themeChecksums] = await Promise.all([fetchChecksums(theme.id, session), themeFileSystem.ready()])
   recordTiming('theme-service:push:file-system')
 
   const {uploadResults, renderThemeSyncProgress} = uploadTheme(
--- a/packages/theme/src/cli/services/push.test.ts
+++ b/packages/theme/src/cli/services/push.test.ts
@@ -2,13 +2,14 @@ import {setDevelopmentTheme} from './local-storage.js'
 import {PullFlags} from './pull.js'
 import {createOrSelectTheme, push, PushFlags} from './push.js'
 import {mountThemeFileSystem} from '../utilities/theme-fs.js'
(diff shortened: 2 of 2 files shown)


Arjunmehta312:
I have signed the CLA!

packages/theme/src/cli/services/push.ts
```diff
@@ -223,11 +223,11 @@ async function executePush(
   context?: {stdout?: Writable; stderr?: Writable},
 ) {
   recordTiming('theme-service:push:file-system')
-  const themeChecksums = await fetchChecksums(theme.id, session)
   const themeFileSystem = mountThemeFileSystem(options.path, {
     filters: options,
     listing: options.listing,
   })
+  const [themeChecksums] = await Promise.all([fetchChecksums(theme.id, session), themeFileSystem.ready()])
```
EvilGenius13, line 230: Is there any reason we couldn't just literally flip the order of the values and keep this as `const themeChecksums = await fetchChecksums(theme.id, session)`?

EvilGenius13 (commented):
Hi @Arjunmehta312, thank you for opening this PR!

Could you please confirm that you tested these changes locally? A screenshot would be an easy way to show everything is still working! Thanks 😄

Arjunmehta312:
Hi @EvilGenius13 
I tested this locally by running the theme package unit tests, lint, and typecheck after the change. I don't have a practical way to capture a meaningful screenshot in my current environment, but everything completed successfully before I opened the PR.

packages/theme/src/cli/services/push.ts
Arjunmehta312: Good question. I considered that approach as well. The reason I used Promise.all is that mountThemeFileSystem() starts its asynchronous setup immediately, and ready() exposes that promise. Waiting on fetchChecksums() first leaves the mount promise without a rejection handler until later. Using Promise.all attaches handlers to both operations immediately while still allowing them to run concurrently. It also matches the existing pattern used in theme pull.

If I'm mistaken about that behavior, I'm happy to change it.

Arjunmehta312:
Hi @EvilGenius13 , I’ve addressed both comments and requested a re-review. Thanks for the feedback and for taking the time to review!

EvilGenius13:
Hi @Arjunmehta312 thank you for the changes. We prefer to have the code tested outside of the test suite so if you can test this against a store that would be very helpful.

PR #8258, opened 2026-08-06
Add --unsafe-validation to app dev

dmerand:
## What

Add `--unsafe-validation` (`SHOPIFY_FLAG_UNSAFE_VALIDATION`, default `false`) to `shopify app dev`. Pass the option to dev-session create and update as `isUnsafe`, and display warnings returned by updates.

## Why

Let developers opt into the validation exceptions that the server permits for dev previews. The server decides which checks can be relaxed.

## How

Forward the option through dev-session setup and both mutation paths. Warnings do not hide errors, session takeover still aborts, and `shopify app deploy` stays unchanged.

Regeneration also includes three upstream schema-comment corrections required by the codegen clean-tree check.

## Release gate

Keep this PR as a draft until the compatible server API and unsafe-validation policy are deployed. The mutation documents name `isUnsafe` even when its value is false, so older servers reject all dev sessions that use these documents.

## How to test

Once the server policy is deployed, run `shopify app dev --unsafe-validation` with an app that triggers an eligible validation warning. Confirm the preview starts and shows the warning. Run without the flag and confirm that the same check remains blocking.

### DIFF
--- a/.changeset/unsafe-app-dev.md
+++ b/.changeset/unsafe-app-dev.md
@@ -1,0 +1,5 @@
+---
+'@shopify/app': patch
+---
+
+Add an `--unsafe` flag to `shopify app dev`.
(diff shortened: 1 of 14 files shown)


amcaplan (commented):
The code itself looks fine.

I do find that the feature isn't really self-documenting; `Allow unsafe app version validation during dev previews.` doesn't really explain to me what that means. Maybe we need a different term?

"Unsafe" tells me something has potentially negative consequences; I'm not sure what could potentially happen just because I run this command. It feels like basically nothing - the preview might not work, but that's not going to break things irreparably.

The terminology might make sense in the API itself, I just think we need something clearer for the CLI flag.

PR #8262, opened 2026-08-07
[Refactor] Refactor resolveFramework to use declarative .some() and .every()

gonzaloriestra:
Refactored resolveFramework in packages/cli-kit/src/public/node/framework.ts to use standard .some() and .every() array helper methods instead of .reduce(). This fixes a bug in the old reduce implementation where the accumulator was not correctly passed on the `some` detectors and enables early exit short-circuiting. Added unit tests for these behaviors in framework.test.ts.

---
*PR created automatically by Jules for task [23259390411241093](https://jules.google.com/task/23259390411241093) started by @gonzaloriestra*

### DIFF
--- a/packages/cli-kit/src/public/node/framework.ts
+++ b/packages/cli-kit/src/public/node/framework.ts
@@ -151,23 +151,21 @@ const frameworks: Framework[] = [
 export async function resolveFramework(rootDirectory: string): Promise<string> {
   const fwConfigFiles: Record<string, string | undefined> = {}
 
-  const matchedFramework = frameworks.find(
-    (framework) =>
-      (!framework.detectors.some ||
-        framework.detectors.some.reduce(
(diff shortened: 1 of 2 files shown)


packages/cli-kit/src/public/node/framework.ts
```diff
@@ -151,23 +151,21 @@ const frameworks: Framework[] = [
...
-          (previousDetectorsMatch: boolean, detector) =>
-            previousDetectorsMatch
-              ? matchDetector(detector, loadFwConfigFile(rootDirectory, detector.path, fwConfigFiles))
-              : false,
-          true,
-        )),
-  )
+  const matchedFramework = frameworks.find((framework) => {
+    const matchesSome =
+      !framework.detectors.some ||
```
isaacroldan, line 156: Looking at the frameworks definition, I think detectors.some is always undefined, should we simplify this? :thinking_face:

PR #8269, opened 2026-08-08
Redact sensitive device authorization debug output

dmerand:

The device authorization debug log serialized the full server response. In verbose mode this wrote `device_code` — a live credential while the flow is pending — and `verification_uri_complete` to the terminal and to anything that captures CLI output.

- Log only non-sensitive metadata after device authorization: HTTP status, `interval`, `expires_in`. Omitted optional fields render as `not provided`.
- Keep user-facing output unchanged: the user code and verification link still appear in info output, and the returned values are untouched.
- Add regression tests that assert the secrets are absent from debug output, the retained diagnostics are present (so a later cleanup cannot silently empty the support channel), and the fallback rendering works.

### How to test your changes?

Run `shopify auth logout`, then any authenticated command with `--verbose` (for example `shopify theme list --verbose`). Confirm the "Received device authorization response" debug line shows only status/interval/expiry, while the login prompt still shows the user code and link.

### Checklist


byrichardpowell (approved):

packages/cli-kit/src/private/node/session/device-authorization.test.ts
```diff
@@ -68,6 +68,48 @@ describe('requestDeviceAuthorization', () => {
     expect(got).toEqual(dataExpected)
   })
 
+  test('does not include authorization credentials in debug output', async () => {
```
amcaplan, line 106: This feels like a test that "We don't do what we used to do" - but it doesn't actually add a whole lot of value in the current state.

packages/cli-kit/src/private/node/session/device-authorization.test.ts
```diff
@@ -68,6 +68,48 @@ describe('requestDeviceAuthorization', () => {
...
+    vi.mocked(identityFqdn).mockResolvedValue('fqdn.com')
+    vi.mocked(clientId).mockReturnValue('clientId')
+
+    // When
+    await requestDeviceAuthorization(['scope1', 'scope2'])
+
+    // Then
+    const debugOutput = JSON.stringify(outputDebug.mock.calls)
+    expect(debugOutput).toContain('interval=not provided')
+    expect(debugOutput).not.toContain('interval=undefined')
```
amcaplan: This last line feels unnecessary

amcaplan (approved):
Appreciate the change. A couple of comments on making sure our tests add value.

packages/cli-kit/src/private/node/session/device-authorization.test.ts
dmerand: Kept this one deliberately: the `not.toContain` lines are the security regression guard — leaking `device_code` is the exact behavior we never want back, so "we don't do what we used to do" is this test's whole job. The positive assertions protect the retained diagnostics from a future over-zealous cleanup. Happy to revisit if you feel strongly.

packages/cli-kit/src/private/node/session/device-authorization.test.ts
dmerand: Agreed — removed in 31aa4a307d.

packages/cli-kit/src/private/node/session/device-authorization.test.ts
dmerand: Agent getting a little feisty there, but yeah, we want a regression test here IMO.

isaacroldan (approved):

PR #8270, opened 2026-08-08
Encode device authorization request parameters

dmerand:

The device authorization request built its `application/x-www-form-urlencoded` body by string concatenation without encoding. Scopes are joined with spaces, so the body carried literal spaces, and any `&`, `=`, or `%` in a value could change the request structure. It works today only because the server is lenient. The token endpoint in `exchange.ts` already uses `URLSearchParams` for the same reason.

- Build the device authorization body with `URLSearchParams`, matching the existing token-request precedent.
- Preserve the current behavior of omitting empty values (an empty scope string stays out of the body).
- Pin the request in tests: exact URL, method, `Content-Type`, and a hand-written encoded body (spaces as `+`, `%26`/`%3D`/`%25` for reserved characters), so the expectation does not depend on `URLSearchParams` itself.

One pre-existing assertion changes from the unencoded body to the encoded form — that wire change is the fix. No change to prompts, returned data, or requested scopes.

### How to test your changes?

Run `shopify auth logout`, then any authenticated command (for example `shopify theme list`). Device login completes as before; the request body is now correctly encoded.

### Checklist

### DIFF
--- a/packages/cli-kit/src/private/node/session/device-authorization.test.ts
+++ b/packages/cli-kit/src/private/node/session/device-authorization.test.ts
@@ -74,17 +74,15 @@ describe('requestDeviceAuthorization', () => {
     vi.mocked(shopifyFetch).mockResolvedValue(response)
     vi.mocked(identityFqdn).mockResolvedValue('fqdn.com')
     vi.mocked(clientId).mockReturnValue('clientId')
-    const scopes = ['scope&name', 'scope=value', 'scope%value']
 
     // When
-    await requestDeviceAuthorization(scopes)
+    await requestDeviceAuthorization(['scope&=%'])
 
     // Then
-    expect(shopifyFetch).toHaveBeenCalledWith('https://fqdn.com/oauth/device_authorization', {
-      method: 'POST',
-      headers: {'Content-type': 'application/x-www-form-urlencoded'},
-      body: 'client_id=clientId&scope=scope%26name+scope%3Dvalue+scope%25value',
-    })
+    expect(shopifyFetch).toHaveBeenCalledWith(
+      expect.any(String),
+      expect.objectContaining({body: 'client_id=clientId&scope=scope%26%3D%25'}),
+    )
   })
 
   test('omits empty authorization scope values', async () => {
--- a/packages/cli-kit/src/private/node/session/device-authorization.test.ts
+++ b/packages/cli-kit/src/private/node/session/device-authorization.test.ts
@@ -63,9 +63,46 @@ describe('requestDeviceAuthorization', () => {
     expect(shopifyFetch).toBeCalledWith('https://fqdn.com/oauth/device_authorization', {
       method: 'POST',
       headers: {'Content-type': 'application/x-www-form-urlencoded'},
-      body: 'client_id=clientId&scope=scope1 scope2',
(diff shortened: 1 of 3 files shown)


packages/cli-kit/src/private/node/session/device-authorization.test.ts
```diff
@@ -63,11 +63,48 @@ describe('requestDeviceAuthorization', () => {
     expect(shopifyFetch).toBeCalledWith('https://fqdn.com/oauth/device_authorization', {
       method: 'POST',
       headers: {'Content-type': 'application/x-www-form-urlencoded'},
-      body: 'client_id=clientId&scope=scope1 scope2',
+      body: new URLSearchParams({client_id: 'clientId', scope: 'scope1 scope2'}).toString(),
     })
```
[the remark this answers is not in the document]
dmerand: Fixed in 250f635e74 — the basic scope test now asserts the literal encoded body (`client_id=clientId&scope=scope1+scope2`), matching the special-characters and empty-scope tests.

byrichardpowell (approved):

packages/cli-kit/src/private/node/session/device-authorization.test.ts
```diff
@@ -63,11 +63,48 @@ describe('requestDeviceAuthorization', () => {
     expect(shopifyFetch).toBeCalledWith('https://fqdn.com/oauth/device_authorization', {
       method: 'POST',
       headers: {'Content-type': 'application/x-www-form-urlencoded'},
-      body: 'client_id=clientId&scope=scope1 scope2',
+      body: 'client_id=clientId&scope=scope1+scope2',
     })
     expect(got).toEqual(dataExpected)
   })
 
+  test('encodes special characters in authorization scopes', async () => {
```
amcaplan, line 71: Are we testing the behavior of URLSearchParams? I feel like we could get away with much simpler testing here...

packages/cli-kit/src/private/node/session/device-authorization.ts
```diff
@@ -152,10 +152,8 @@ export async function pollForDeviceAuthorization(code: string, interval = 5): Pr
 }
 
 function convertRequestToParams(queryParams: {client_id: string; scope: string}): string {
-  return Object.entries(queryParams)
-    .map(([key, value]) => value && `${key}=${value}`)
-    .filter((hasValue) => Boolean(hasValue))
-    .join('&')
+  // URLSearchParams applies form encoding so scope values cannot change the request structure.
```
amcaplan: I don't think we need a lot of justification for the use of URLSearchParams over a hand-rolled conversion function 😆

packages/cli-kit/src/private/node/session/device-authorization.test.ts
dmerand: Fair point — simplified in 3471d72f99: one representative reserved-character case asserting only the body. The basic test above already pins the full request shape, so the char matrix was mostly exercising Node stdlib.

packages/cli-kit/src/private/node/session/device-authorization.ts
dmerand: Removed in 3471d72f99 😄

PR #8279, opened 2026-08-10
Remove obsolete polyfill dependencies from cli-kit

isaacroldan:

Reduces cli-kit's external dependency count. These four dependencies are polyfills for constraints that no longer exist, or are misplaced test-only packages.

- **Remove `node-abort-controller`**: it polyfills AbortController for Node < 15; the CLI requires Node >= 22.12. `abort.ts` now re-exports the native globals. Call sites doing `new AbortSignal()` (which throws "Illegal constructor" on the native class) were changed to `new AbortController().signal`. Bonus: signals are now real native AbortSignals instead of polyfill objects, so native `fetch`/APIs treat them first-class.
- **Remove `ts-error` and the `ExtendableError` export entirely**: `ExtendableError` only matters when TypeScript targets ES5; we target ES2020 where `class extends Error` works natively. `TunnelError`, `InvalidGrantError`, and `InvalidRequestError` now extend `Error` directly, the one direct `throw new ExtendableError(...)` becomes `throw new Error(...)`, and the redundant `instanceof ExtendableError` check in `Result.valueOrAbort` is gone. A GitHub-wide code search found no consumer of the export outside this repo; a changeset records the removed public API.
- **Move `@types/archiver` to devDependencies**: a types package was shipping in runtime deps.
- **Move `is-executable` to devDependencies**: only used in `github.test.ts`.

No behavior change intended, other than the removed `ExtendableError` export and thrown errors from those classes reporting `name: 'Error'`. Nothing branches on that name.

## How to test your changes?

- `pnpm vitest run` for cli-kit error/abort/result/tunnel/session/github tests and the app test files touched — all green.
- Type-check passes for cli-kit and app.

## Measuring impact
How do we know this change was effective? Existing test suite.

## Checklist

- [x] I've added a changelog entry (removed public export)
- [x] I've considered when to gather analytics


gonzaloriestra (approved):
Nice!

PR #8280, opened 2026-08-10
Replace macaddress, is-wsl and network-interfaces with os built-ins

isaacroldan:

Part of an effort to reduce external dependencies in the CLI. `macaddress`, `is-wsl` and `network-interfaces` can all be replaced with small implementations on top of Node's `os.networkInterfaces()` and `os.release()`.

- Removes `macaddress`: `macAddress()` in `context/local.ts` now reads from `os.networkInterfaces()`, preferring external interfaces with a non-empty MAC (same behavior as the library).
- Removes `is-wsl`: `isWsl()` in `system.ts` checks `os.release()` and `/proc/version` for `microsoft` on Linux, excluding Docker/Podman containers — the same detection logic the library uses.
- Removes `network-interfaces` (and its vendored type declaration): the dev-server internal-IP check validates the resolved backend IP against internal IPv4 interfaces from `os.networkInterfaces()` directly.

No changeset: internal refactor with no user-visible behavior change.

### How to test your changes?

- `pnpm vitest run src/public/node/context/local.test.ts src/public/node/context/local.integration.test.ts src/public/node/system.test.ts` in `packages/cli-kit` (60 tests pass)
- `pnpm type-check` in `packages/cli-kit`

### Measuring impact

- [x] n/a — internal dependency cleanup

### Checklist


packages/cli-kit/src/public/node/vendor/dev_server/dev-server-2024.ts
gonzaloriestra, line 1: To avoid conflicts the next time we want to update this code, what about opening a PR in https://github.com/Shopify/dev_server as well?

isaacroldan: So apparently the original code doesn't depend on this... I think it was added when vendoring it...

gonzaloriestra (commented):
Can we add tests for the new code? The old libraries probably had it.

packages/cli-kit/src/public/node/vendor/dev_server/dev-server-2024.ts
isaacroldan: Confirmed — upstream resolves a hostname (`getBackendHost`) and has no IP validation at all; the whole `getBackendIp` path was added when vendoring, so there's nothing to change in dev_server. I've updated the comment in the file to make that clear for future syncs.

isaacroldan:
@gonzaloriestra Added tests :v:

gonzaloriestra (approved):
Great!

PR #8282, opened 2026-08-10
Remove color-json in favor of a chalk-based colorizer

isaacroldan:

Part of an effort to reduce external dependencies in the CLI. `color-json` is used in a single place (`JsonContentToken` in cli-kit's `content-tokens.ts`), and its behavior is a ~20-line token colorizer we can implement with the chalk instance cli-kit already ships.

Replaces `color-json` with a local `colorJson` helper. The implementation is identical to the one we are replacing: same token regex and same color mapping (yellow structural characters, white keys, green strings, magenta numbers, cyan booleans, red `null`), verified character by character against `color-json`'s output. String input is still parsed and re-stringified with 2-space indentation, and the existing plain `JSON.stringify` fallback for non-serializable values is unchanged.

No changeset: internal refactor with no user-visible behavior change.

### How to test your changes?

- `pnpm vitest run src/public/node/output.test.ts` in `packages/cli-kit`
- `pnpm vitest run src/cli/services/dev/extension/websocket/handlers.test.ts` in `packages/app` (uses `token.json`)
- `pnpm type-check` in `packages/cli-kit`

### Measuring impact

- [x] n/a — internal dependency cleanup

### Checklist

### DIFF
--- a/packages/cli-kit/src/private/node/content-tokens.ts
+++ b/packages/cli-kit/src/private/node/content-tokens.ts
@@ -48,6 +48,8 @@ export class CommandContentToken extends ContentToken<OutputMessage> {
   }
 }
 
+// Token-matching regex taken from the color-json package this implementation replaces:
+// https://www.npmjs.com/package/color-json
 const jsonTokenRegex =
   /("(\\u[a-zA-Z0-9]{4}|\\[^u]|[^\\"])*"(\s*:)?|\b(true|false|null)\b|-?\d+(?:\.\d*)?(?:[eE][+-]?\d+)?)/g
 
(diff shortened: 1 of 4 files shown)


packages/cli-kit/src/private/node/content-tokens.ts
gonzaloriestra, line 1: Can we add tests?  
  
Also, is this code taken from color-json? If so, I think we should mention it in a comment at least.

isaacroldan: I'll add test yes, good idea.  
  
About the code: No, we only took the regex, the rest of the code is different (simpler)

isaacroldan: Tests added

gonzaloriestra (approved):

PR #8290, opened 2026-08-10
Use CLI commands for E2E store lifecycle

isaacroldan:
### Why

The E2E suite provisions and deletes development stores through Playwright even though the CLI now supports these operations.

### What

- Create per-test development stores with `shopify store create dev --json`.
- Delete stores with `shopify store delete --force --json` after app uninstall verification.
- Validate command results and allow enough time for provisioning and deletion polling.

### Testing

- `pnpm --filter @shopify/e2e type-check`
- `pnpm --filter @shopify/e2e lint`
- `pnpm --filter @shopify/e2e exec playwright test tests/app-dev-server.spec.ts --list`

A live local Playwright run is blocked before test setup by the existing local `node-pty` error `posix_spawnp failed`. CI rebuilds `node-pty` before the E2E job.

### DIFF
--- a/packages/e2e/scripts/cleanup-stores.ts
+++ b/packages/e2e/scripts/cleanup-stores.ts
@@ -25,8 +25,9 @@ import * as path from 'path'
 import * as fs from 'fs'
 import {fileURLToPath} from 'url'
 import {chromium} from '@playwright/test'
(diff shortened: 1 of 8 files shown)


gonzaloriestra (approved):
Simpler and more robust 👌

PR #8291, opened 2026-08-10
Clean up Node <22 compatibility leftovers

isaacroldan:

The CLI requires Node >=22.12.0 (`engines` in every package), but several pieces of the codebase still targeted older Node and ES versions.

- **Remove the `Promise.withResolvers` polyfill** (`packages/theme/src/cli/polyfills/promiseWithResolvers.ts`). Its own comment said to remove it once the minimum Node version is 22. The one usage in `theme-environment.ts` now calls `Promise.withResolvers` directly.
- **Bump the TypeScript `target`/`lib` from ES2020 to ES2024** in `configurations/tsconfig.json`. Node 22 supports ES2024 natively, and the `es2024` lib provides the `Promise.withResolvers` types.
- **Bump `@types/node` from 18 to 22** (root and `packages/e2e`). Two test mocks of `process.exit` needed their signatures widened to match the Node 22 types.
- **Update the runtime minimum-Node check from 18 to 22** in `cli-kit`'s `runCLI`. Users on Node 18/20 now get the upgrade error instead of undefined behavior.
- **Remove redundant field re-declarations in `AbortError`**. Surfaced by the target bump: under ES2022+ class-field semantics, re-declaring `nextSteps`/`customSections` without initializers would reset the values the `FatalError` constructor assigns. The base class already declares both fields.

No changeset: internal cleanup, except the version-check bump which only affects already-unsupported Node versions.

### How to test your changes?

- `pnpm type-check`, `pnpm lint:affected`, and knip pass.
- Full `cli-kit` and `theme` test suites pass (218 files, 2548 tests).
- `pnpm refresh-code-documentation` produces no churn.
- The existing old-Node test now exercises Node 20 to prove the new threshold.

### Measuring impact

- [x] n/a


packages/theme/src/cli/polyfills/promiseWithResolvers.ts
```diff
@@ -1,28 +0,0 @@
...
-
-declare global {
-  interface PromiseConstructor {
-    withResolvers<T>(): PromiseWithResolvers<T>
-  }
-}
-
-// Polyfill for Promise.withResolvers
-// Can remove once our minimum supported Node version is 22
-export function promiseWithResolvers<T>(): PromiseWithResolvers<T> {
```
isaacroldan, line 15: @EvilGenius13 you added this polyfill (9c35fb110c4) with a note to remove it once our minimum Node version is 22. That minimum is now in place (`engines` requires >=22.12.0), so this PR removes it and calls `Promise.withResolvers` directly. Can you confirm this is safe to remove?

graygilmore (approved):

gonzaloriestra (approved):
:clap:

PR #8317, opened 2026-08-12
Retry GraphQL requests rejected as Throttled without a 429 code

isaacroldan:

`retryAwareRequest` only retries rate-limited requests on HTTP 429 or a GraphQL error with `extensions.code === '429'`. But Shopify GraphQL APIs (App Management among them) throttle with a 200 response and `extensions.code: "THROTTLED"` — those requests fail instantly on first rejection.

The CLI already treats `THROTTLED` as a rate-limit signal in crash-report suppression and analytics grouping. The retry path was the only place that didn't.

This is currently the top E2E failure mode (32 of 79 failed shards last week), and real `app dev` / `app deploy` users hit the same instant failure.

The retryability check now reuses the shared `hasRateLimitCode` helper, so `extensions.code === 'THROTTLED'` is retried exactly like `'429'`. The code is server-set (see `complexity_throttle.rb` and `app_error_handling.rb` in shop/world), so unlike message matching, user-controlled strings can't trigger retries. Retry mechanics are unchanged: up to 10 retries, `Retry-After` when present, 1s default backoff.

### How to test your changes?

`api.test.ts` adds two cases: a 200 + `THROTTLED` error is retried and succeeds; an error whose message merely says "Throttled" without the code is not retried.


alfonso-noriega (approved):

packages/cli-kit/src/private/node/api.ts
```diff
@@ -263,7 +263,14 @@ function errorsIncludeStatus429(error: ClientError): boolean {
   if (typeof error.response.errors === 'string') {
     return false
   }
-  return error.response.errors?.some((error) => error.extensions?.code === '429') ?? false
+  // Some Shopify APIs (e.g. App Management) throttle with a 200 response whose
+  // GraphQL error message is "Throttled", with no 429 status or code — match
+  // the message so those are retried too.
+  return (
+    error.response.errors?.some(
+      (graphqlError) => graphqlError.extensions?.code === '429' || /^throttled/i.test(graphqlError.message ?? ''),
```
amcaplan: Regex tests always raise some flags... Is there nothing like a canonical error code we can use? Or at least some tighter enforcement around the message text? So someone can't name an app "throttled" and then retry on user errors...

isaacroldan: yeah, you are right, this was a very naive solution. I'll look more into this, apparently the AI understood that the  API is returning a 200 code with a throttled message, but that's not true

isaacroldan: Good call — there is a canonical code: throttle errors carry a server-set `extensions.code: "THROTTLED"`. I assumed there was none because the CLI's rendered error only prints the message.

Reworked to match the code exactly, no message matching left. Added a negative test: a message that just says "Throttled" without the code is not retried.

PR #8319, opened 2026-08-12
Show App Events extension status in dev sessions

Barefoot0:
# TL;DR

Show `analytics_app_events  Extension loaded` after Shopify CLI successfully creates or updates a dev session containing an Analytics App Events extension.

# Context

`analytics_app_events` is currently a remote-only extension specification. Shopify CLI includes its TOML configuration in the dev-session flow, but the extension has neither client build steps nor a `getDevSessionUpdateMessages` hook, so it is silent after `Ready`.

# Changes

- Register a thin local, contract-based `analytics_app_events` specification that returns `Extension loaded` through the existing dev-session message hook.
- Let contract-based specifications opt into that existing hook without changing the default behavior of any other extension type.
- Retain the remote platform JSON schema as the source of truth for App Events configuration validation.
- Add coverage for the exact handle-prefixed message, remote-schema merging, and silence for a representative existing contract-based extension.

### DIFF
--- a/.changeset/quiet-app-events-status.md
+++ b/.changeset/quiet-app-events-status.md
@@ -1,0 +1,5 @@
+---
+'@shopify/app': minor
+---
+
+Show a status message when an Analytics App Events extension loads during `shopify app dev`.
--- a/packages/app/src/cli/models/extensions/extension-instance.ts
+++ b/packages/app/src/cli/models/extensions/extension-instance.ts
@@ -394,9 +394,9 @@ export class ExtensionInstance<TConfiguration extends BaseConfigType = BaseConfi
     }
   }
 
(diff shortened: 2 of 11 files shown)


Barefoot0:
Screenshot 2026-08-17 at 2 57 29 PM
Image of my local testing. 
It logs a message when it is loaded, but doesn't repeat that message every time it is changed.

amcaplan:
This code passes through a very specific value (session status) for a very specific purpose, which I don't love as it emphasizes special casing in shared code. ‏ ‏I can see an argument for passing through the entire dev session result, though, as it opens up more possibilities for various extensions in the future.

Barefoot0:
Closed by https://github.com/Shopify/cli/pull/8350 (Dev Platform decided to make it a general implementation, not specific to one extension).

PR #8323, opened 2026-08-13
Create dev stores from app dev

dmerand:

Interactive `app dev` on App Management dead-ends when an organization has no usable development store: the CLI can only point to the Developer Dashboard. Developers should be able to create a dev store without leaving the `app dev` flow.

Tracks [shop/issues-develop#23596](https://github.com/shop/issues-develop/issues/23596).

Extracts the existing `store create dev` create/poll service and the Business Platform token-refresh helper from `@shopify/store` into `@shopify/organizations` so `app dev` can reuse them. The only extensions to the service are: `createDevStore` now returns the created shop domain, and a `summary: false` option suppresses the standalone success output when the service is embedded. `@shopify/store` keeps compatibility re-exports, and standalone `store create dev` behavior and output are unchanged.

On top of that shared service:

- Store selection gains an explicit `storeCreationMode` capability (`'disabled' | 'when-empty'`). Every `storeContext` command defaults to `disabled`; only `app dev` opts in with `when-empty`.
- Interactive `app dev` (App Management) with zero usable stores goes straight to inline creation: it prompts for store name and plan, waits for creation, fetches the new store, and continues `dev` with it.
- Existing-store selection behavior is unchanged: one store still auto-selects, and the many-store picker and its search show only stores. An initial store remains selectable after the search input is cleared. Stacked follow-up [#8397](https://github.com/Shopify/cli/pull/8397) adds a create option to the picker on top of this PR.
- Noninteractive zero-store `app dev` never prompts or creates. It directs the developer to run `shopify store create dev` with the required organization, name, and plan flags, then run `shopify app dev --store <store-domain>`. Existing-store selection, explicit `--store`, and cached `dev_store_url` behavior is unchanged.
- Partners keeps the existing Developer Dashboard link and reload flow. Commands with creation disabled keep their previous selection behavior.
- Cap handling: only the zero-store `app dev` path queries `Organization.devStoreCapReached`. The flow checks the cap before noninteractive guidance and again before it asks for the store name or plan. A capped organization must first make a development-store slot available in Dev Dashboard, then use the finite `shopify store create dev` and `shopify app dev --store` sequence. Query errors or a missing field fail open, so the creation mutation remains the final authority.
- This PR does not define JSON output, NDJSON, an event stream, or a prompt protocol for `app dev`. The [CLI JSON outputs design](https://docs.google.com/document/d/1M2YZHKJFFgZo0O8TvBG0NMJGDaiDu98xWOeUxW2QMFM/edit?tab=t.0#heading=h.ixauxxv1jr68) defers long-lived commands such as `app dev` and their prompt protocol.

`Organization.devStoreCapReached` comes from [shop/world PR #944713](https://github.com/shop/world/pull/944713), which is merged. Until that deploy is fully rolled out, the CLI fails open and keeps creation available; cap enforcement needs a tophat against a deployed environment.

### How to test your changes?

Manual tophat on `4e85e16c` verified the shared create-and-continue flow and standalone compatibility:

1. Ran `shopify app dev` without `--store` and created a store inline with a name and Basic plan. The store was created with one success banner before `app dev` reached preview-ready. At that revision the flow was entered from the store picker; this PR narrows entry to organizations with zero usable stores, and that entry point has not been manually retested.
2. Re-ran `shopify app dev` with an explicit `--store`, then without `--store` using the cached selection. Both bypassed store creation and reached preview-ready on the selected domain.
3. Confirmed the inline-created store appeared as type `dev` in `shopify store list --json`.
4. Ran standalone `shopify store create dev --json` to verify its structured output after the shared-service move. Deletion requests were confirmed for both stores created during the tophat.

The zero-store entry point, noninteractive missing-store behavior, cap enforcement, and the Partners dashboard fallback remain manual gaps; these paths have focused test coverage.

### Checklist


packages/app/src/cli/services/dev/cap.ts
```diff
@@ -0,0 +1,18 @@
...
+  organizationId: string,
+  developerPlatformClient: DeveloperPlatformClient,
+): Promise<boolean> {
+  const capChecker = developerPlatformClient.devStoreCapReached
+  if (developerPlatformClient.clientName !== ClientName.AppManagement || !capChecker) {
+    return false
+  }
+
+  try {
+    return await capChecker(organizationId)
```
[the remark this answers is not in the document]
dmerand: Thanks — fixed in `cf825a30`. The helper now invokes `developerPlatformClient.devStoreCapReached(organizationId)` through its owning client, and the regression test uses a `this`-dependent implementation so a detached call fails the test.

packages/store/src/cli/services/store/business-platform.ts
```diff
@@ -1,19 +1 @@
...
-  return {
-    type: 'token_refresh',
-    handler: async () => ({token: await refreshBusinessPlatformToken(options)}),
-  }
-}
-
-async function refreshBusinessPlatformToken(options: BusinessPlatformTokenRefreshHandlerOptions): Promise<string> {
-  return ensureAuthenticatedBusinessPlatform([], {noPrompt: options.noPrompt})
-}
+export {businessPlatformTokenRefreshHandler} from '@shopify/organizations'
```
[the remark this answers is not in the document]
dmerand: Thanks — I could not reproduce this. `@shopify/organizations` exposes only its root entrypoint, so the compatibility re-export uses the supported public boundary. The module graph is acyclic, the cited delete test initializes and runs, and the full `@shopify/store` suite passed 429/429 tests. Adding a leaf export would expand the package API without fixing an observed failure.

packages/store/src/cli/services/store/create/dev.ts
```diff
@@ -1,171 +1 @@
...
-    })
-  }
-}
-
-function pushRow(rows: InlineToken[][], label: string, value: InlineToken | undefined): void {
-  if (value !== undefined && value !== null && value !== '') {
-    rows.push([label, value])
-  }
-}
+export {createDevStore} from '@shopify/organizations'
```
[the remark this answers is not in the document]
dmerand: Thanks — I could not reproduce this. `@shopify/organizations` exposes only its root entrypoint, and the Store command test retains the real root exports while overriding the required symbols. The cited service and command tests initialize and pass, and the full `@shopify/store` suite passed 429/429 tests. A new leaf export is not needed.

packages/app/src/cli/services/dev/select-store.ts
```diff
@@ -1,35 +1,63 @@
...
   // If no stores, guide the developer through creating one.
   // Then, with a store selected, make sure it's transfer-disabled.
   let store = await selectStorePrompt({
     onSearchForStoresByName,
     ...storesSearch,
     showDomainOnPrompt,
+    ...(onCreateStoreWhenEmpty ? {onCreateStoreWhenEmpty} : {}),
   })
   if (!store) {
+    if (storeCreationEnabled) {
```
craigmichaelmartin, line 57: Since this also fires when the org *has* stores and the picker returns nothing, and an empty `CancelExecution()` prints no message at all, should the check be `if (onCreateStoreWhenEmpty)` instead?

packages/app/src/cli/services/dev/select-store.ts
```diff
@@ -1,35 +1,63 @@
...
+  const storeCreationEnabled =
+    storeCreationMode === 'when-empty' && developerPlatformClient.clientName === ClientName.AppManagement
+
+  let onCreateStoreWhenEmpty: (() => Promise<OrganizationStore>) | undefined
+  if (storeCreationEnabled && storesSearch.stores.length === 0) {
+    // Inline store creation needs an interactive terminal.
+    if (isTTY() === false) {
+      throw new AbortError(
+        'No development store was specified.',
+        'Run `app dev --store <store-domain>` to select a development store.',
```
craigmichaelmartin, line 32: We only reach this when the org has zero stores, and `--store` searches that same empty list — should this point at creating a store instead?

packages/app/src/cli/services/dev/select-store.ts
```diff
@@ -85,6 +149,9 @@ async function waitForCreatedStore(
   return data
 }
 
+const devStoreCapReachedMessage = 'Your organization has reached its development store limit.'
+const devStoreCapReachedTryMessage = 'Run `app dev --store <store-domain>` to select an existing development store.'
```
craigmichaelmartin, line 153: Same question here — should this point at the dashboard rather than at picking a store that isn't there?

packages/app/src/cli/services/dev/select-store.ts
```diff
@@ -41,21 +69,57 @@ export async function selectStore(
...
+        for (let i = 0; i < retries; i++) {
+          try {
+            // eslint-disable-next-line no-await-in-loop
+            const fetchedStore = await fetchStore(org, shopDomain, developerPlatformClient)
+            if (fetchedStore) {
+              store = fetchedStore
+              return
+            }
+          } catch (error) {
+            if (!(error instanceof AbortError)) throw error
```
craigmichaelmartin, line 101: `storeByDomain` also throws `AbortError` for a missing org or an expired session — should this catch only the not-found case, so real errors aren't reported as "not available yet"?

craigmichaelmartin (approved):

packages/app/src/cli/services/dev/select-store.ts
dmerand: Fixed in f6d42d48d6. `fetchStore` now throws `StoreNotFoundError` only when the domain is absent. Created-store polling retries only that subtype and propagates all other errors immediately. Tests cover one not-found retry and immediate propagation of an ordinary `AbortError`.

packages/app/src/cli/services/dev/select-store.ts
dmerand: Fixed in f6d42d48d6. The immediate silent cancellation now checks `onCreateStoreWhenEmpty`. If an existing-store picker returns no selection, the flow shows the dashboard link and reload prompt. The regression test uses a two-store picker so cancellation is reachable.

packages/app/src/cli/services/dev/select-store.ts
dmerand: Fixed in f6d42d48d6. A non-interactive zero-store run now says: `Create a development store in Dev Dashboard, then run app dev again.` It no longer recommends `--store` when no store exists.

packages/app/src/cli/services/dev/select-store.ts
dmerand: Fixed in f6d42d48d6. The capped zero-store case now gives distinct guidance to manage the development-store limit in Dev Dashboard, then run `app dev` again. It does not recommend `--store` or invite another creation.

packages/app/src/cli/services/dev/select-store.ts
dmerand: What is this agent garbage? Fixing.

packages/app/src/cli/services/dev/select-store.ts
dmerand: More garbage, also fixing.

dmerand: Correction: final head 6458525fa0 uses a capacity-first sequence. A capped organization must make a development-store slot available in Dev Dashboard, then run the finite `shopify store create dev` command and continue with `shopify app dev --store <store-domain>`. The flow does not suggest selecting a nonexistent store or imply creation can succeed while capped.

packages/app/src/cli/services/dev/select-store.ts
dmerand: Correction: my previous reply described the wrong path. Final head 6458525fa0 directs a non-interactive zero-store run to the finite command `shopify store create dev --organization-id <organization-id> --name <store-name> --plan <plan>`, then to `shopify app dev --store <store-domain>`. The cap is checked before this guidance.

packages/app/src/cli/services/dev/select-store.ts
```diff
@@ -41,21 +70,57 @@ export async function selectStore(
...
+        }
+      },
+    },
+  ]
+  await renderTasks(tasks)
+
+  if (!store) {
+    throw new AbortError(
+      `The newly created development store (${shopDomain}) is not available yet.`,
+      'Run `shopify app dev --store <store-domain>` to select it when it is ready.',
```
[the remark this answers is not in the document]
dmerand: Fixed in 67ba0e9f25. The exhausted-retry guidance now interpolates the known created domain, for example `shopify app dev --store created-store.myshopify.com`, and the 10-retry exhaustion test asserts that exact command. The parent focused run passes 96 tests; App type-check, lint, and diff checks pass.

PR #8329, opened 2026-08-14
Stop reporting gateway errors as CLI crashes

isaacroldan:
Raw `graphql-request` `ClientError` objects can reach the top-level reporter when a call site uses `handleErrors: false`. For HTTP 502, 503, and 504, that records an infrastructure failure as an unexpected CLI bug.

Classify these gateway statuses as expected in `shouldReportErrorAsUnexpected`, consistent with the handled path that maps 5xx responses to `AbortError`. HTTP 500 keeps its existing classification.

### Testing

- `pnpm --filter @shopify/cli-kit vitest run src/public/node/error.test.ts`
- `pnpm --filter @shopify/cli-kit type-check`
- ESLint on the two changed files

### DIFF
--- a/packages/cli-kit/src/public/node/error.test.ts
+++ b/packages/cli-kit/src/public/node/error.test.ts
@@ -114,6 +114,10 @@ describe('shouldReportErrorAsUnexpected helper', () => {
     expect(shouldReportErrorAsUnexpected(error)).toBe(false)
   })
 
+  test.each([502, 503, 504])('returns false for a raw ClientError from the gateway (HTTP %i)', (status) => {
+    expect(shouldReportErrorAsUnexpected(clientError(status))).toBe(false)
+  })
+
   test('returns true for a raw ClientError that is a genuine failure (HTTP 500)', () => {
     expect(shouldReportErrorAsUnexpected(clientError(500))).toBe(true)
   })
--- a/packages/cli-kit/src/public/node/error.ts
+++ b/packages/cli-kit/src/public/node/error.ts
@@ -209,12 +209,13 @@ export function shouldReportErrorAsUnexpected(error: unknown): boolean {
 /**
  * Detects raw graphql-request `ClientError`s that are expected environmental conditions rather than
  * CLI bugs. These reach the reporter as plain `Error`s (not `FatalError`s) and would otherwise be
- * reported as unexpected. Two distinct cases, both kept out of crash reporting:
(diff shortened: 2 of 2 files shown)


craigmichaelmartin:
Reviewed in the context of resolving Vault 68247 / shop/issues#73647, which this PR targets. **No merge-blocking issues — this looks correct and I'd land it.**

Two things I verified rather than assumed:

1. **The new branch actually fires for the reported signature.** `isExpectedApiError` only proceeds when the error carries both `response` and `request`, and then reads `response.status`. The real Observe payload for #73647 is `{"response":{"error":"<html>…502 Bad Gateway…nginx…","status":502,"headers":{}},"request":{"query":"query getThemes…"}}` — so `response.status` is present even though the body is HTML, and the structural match holds. Worth stating because an HTML-bodied 502 has **no** `response.errors`, so the pre-existing `hasRateLimitCode` fallback could never have caught it; the explicit status check is doing the work.

2. **Keeping 500 reportable is the right line**, and the existing "genuine failure" assertion stays untouched. 502/503/504 come from the proxy in front of the API and say nothing about the CLI; a 500 means the API itself answered and failed.

Non-blocking notes:

- **This fixes the reporting half only.** The issue is typed `Resiliency` and was auto-reopened on event volume, so after this lands the Observe signal goes quiet while the user still loses their command to a transient blip and still sees the raw nginx HTML. I've opened #8450 as a companion PR, restoring the retry half from this branch's own first push (commit 2edc7d0, before it was narrowed) — 502/503/504 become retryable, gated on `requestIsIdempotent` so a mutation is never resent. The two together actually close 68247, and they don't overlap: yours touches `public/node/error.ts`, mine touches the request layer.
- **The three statuses are inlined here.** #8450 adds back your dependency-free `private/node/api/gateway-status.ts` with `isGatewayErrorStatus()`. Whichever of us lands second, it'd be worth collapsing this condition onto that helper so the gateway status list lives in one place rather than two.
- Tiny doc-comment nit, not worth a re-push on its own: dropping "Two distinct cases" leaves 401 and rate-limiting still reading as a pair, with the gateway sentence appended to the rate-limiting paragraph.

For context on why it's worth unblocking: the error has been re-firing since this PR went up on Aug 14 — 37 events / 35 users in the last 7-day window on the theme path alone.

craigmichaelmartin (approved):

PR #8344, opened 2026-08-17
[WIP] Add bundle size exception request flow

igor10k:
### Why?

Remote-DOM UI extension bundles (API version `2025-10` or later) are limited to 64 KB (compressed), enforced at deploy time. Some apps can't reduce their bundles below the limit even after optimization, and Shopify can grant a per-app bundle size exception.

Today there's no way to request one from the CLI, and the deploy error doesn't tell developers what to do next. This PR adds that flow.

### What?

Three commits, reviewable independently:

**1. Keep esbuild metafiles in the app deploy archive**
- Production builds already emit `*.metafile.json` next to each extension bundle; stop excluding them from the deploy archive so bundle composition can be analyzed when an exception request is reviewed. Sourcemaps remain excluded.
- Note: affects the archive contents for every deploy, not just apps requesting exceptions.

**2. Add hidden `shopify app bundle-size-exception request` command**
- Builds each Remote-DOM UI extension for production in a temporary directory and measures raw + compressed sizes.
- Short-circuits when a request is already pending or when all bundles fit the app's current limit.
- Prompts for a justification and confirmation; `--reason` makes it non-interactive.
- Submits via the new `appBundleSizeExceptionRequest` App Management mutation. The CLI sends measurements and a justification, and Shopify reviews every request. Server-side `userErrors` are surfaced directly.
- `DeveloperPlatformClient` gains `bundleSizeException` / `bundleSizeExceptionRequest` (App Management client only).

**3. Point at the request flow from bundle-size deploy errors**
- When a deploy fails with a structured `{type: bundle_size_limit}` entry in the error's `on` payload, the validation-error output gains a "Bundle size exception" section telling the developer to run the request command. Structured detection only — no message parsing.

### Testing

Requires an App Management API that exposes the new `appBundleSizeExceptionRequest` mutation and `app.bundleSizeException` field (rolling out server-side), and an app with a UI extension on `api_version = "2025-10"`.

1. **Request flow**: `pnpm shopify app bundle-size-exception request --path <app>` (`--path` is optional)
   - Small bundle → "No exception needed" with measured per-extension sizes.
   - Bloat the extension (e.g. inline a large dependency) → justification prompt → confirmation → success message.
   - Re-run → "already pending review" info, no resubmission.
   - `--reason "text"` → runs with no prompts.
2. **Deploy CTA**: with an oversized bundle, run `pnpm shopify app deploy` → the bundle-size validation error shows the "Bundle size exception" section.
3. **Metafiles**: covered by unit tests in `bundle.test.ts` (`*.metafile.json` kept, `*.js.map` excluded).

### DIFF
--- a/.changeset/bundle-size-exception-cta.md
+++ b/.changeset/bundle-size-exception-cta.md
@@ -1,0 +1,6 @@
+---
+'@shopify/app': minor
+'@shopify/cli': minor
+---
+
+Point at the bundle size exception request flow from bundle-size deploy errors
--- a/.changeset/deploy-archive-metafiles.md
+++ b/.changeset/deploy-archive-metafiles.md
@@ -1,0 +1,6 @@
+---
+'@shopify/app': minor
+'@shopify/cli': minor
+---
+
+Include esbuild metafiles in the app deploy archive to enable bundle-composition analysis
--- a/packages/app/src/cli/api/graphql/app-management/generated/bundle-size-exception-request.ts
+++ b/packages/app/src/cli/api/graphql/app-management/generated/bundle-size-exception-request.ts
@@ -1,0 +1,120 @@
+/* eslint-disable @typescript-eslint/consistent-type-definitions */
+import * as Types from './types.js'
+import {JsonMapType} from '@shopify/cli-kit/node/toml'
+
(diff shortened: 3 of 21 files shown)


igor10k:
We decided to choose a different approach.

PR #8345, opened 2026-08-17
[Feature] Add intercepts target capability support

henryStelle:
### Why

UI extension target configurations did not recognize `intercepts`, so the capability could be removed while parsing extension configuration. The UI extensions server kit also omitted target capabilities from its extension point type.

### What

- Adds an `intercepts` target capability schema with a string `event`, boolean `blocking`, and optional string-array `applies_to`.
- Preserves the existing `allow_direct_linking` target capability validation.
- Exposes opaque target capabilities on the server kit's `ExtensionPoint`, separately from extension-level capabilities.
- Adds parsing, transformation, and schema validation coverage.

### Testing

- `pnpm vitest run packages/app/src/cli/models/extensions/schemas.test.ts packages/app/src/cli/models/extensions/specifications/ui_extension.test.ts`
- `pnpm nx lint app`
- `pnpm nx type-check app`
- `pnpm --filter @shopify/ui-extensions-server-kit vitest run`
- `pnpm nx lint ui-extensions-server-kit`
- `pnpm nx build ui-extensions-server-kit`
- `pnpm exec prettier --check packages/app/src/cli/models/extensions/schemas.ts packages/app/src/cli/models/extensions/schemas.test.ts packages/app/src/cli/models/extensions/specifications/ui_extension.test.ts`
- `git diff --check`

### DIFF
--- a/packages/app/src/cli/models/extensions/schemas.test.ts
+++ b/packages/app/src/cli/models/extensions/schemas.test.ts
@@ -1,4 +1,4 @@
-import {BaseSchema, MAX_UID_LENGTH} from './schemas.js'
+import {BaseSchema, MAX_UID_LENGTH, NewExtensionPointsSchema} from './schemas.js'
 import {describe, expect, test} from 'vitest'
 
 const validUIDTestCases = [
@@ -53,5 +53,23 @@ describe('UIDSchema', () => {
         expect(result.error.issues[0]?.message).toBe(expectedError)
       }
     })
+  })
+})
+
+describe('NewExtensionPointsSchema', () => {
+  test.each([
+    ['event', {event: 1, blocking: true}],
+    ['blocking', {event: 'cartvalidations', blocking: 'true'}],
+    ['applies_to', {event: 'paymentvalidations', blocking: true, applies_to: [1]}],
+  ])('rejects an intercept with an invalid %s value', (_field, intercept) => {
+    const result = NewExtensionPointsSchema.safeParse([
+      {
+        target: 'pos.app.ready.data',
+        module: './src/app.ts',
+        capabilities: {intercepts: [intercept]},
+      },
+    ])
+
+    expect(result.success).toBe(false)
   })
 })
--- a/packages/app/src/cli/models/extensions/schemas.test.ts
+++ b/packages/app/src/cli/models/extensions/schemas.test.ts
@@ -58,18 +58,26 @@ describe('UIDSchema', () => {
 
 describe('NewExtensionPointsSchema', () => {
   test.each([
-    ['event', {event: 1, blocking: true}],
-    ['blocking', {event: 'cartvalidations', blocking: 'true'}],
-    ['applies_to', {event: 'paymentvalidations', blocking: true, applies_to: [1]}],
-  ])('rejects an intercept with an invalid %s value', (_field, intercept) => {
+    [
+      'valid intercepts',
+      [
+        {event: 'cartvalidations', blocking: true},
+        {event: 'paymentvalidations', blocking: true, applies_to: ['cash']},
+      ],
+      true,
+    ],
+    ['an invalid event', [{event: 1, blocking: true}], false],
+    ['an invalid blocking value', [{event: 'cartvalidations', blocking: 'true'}], false],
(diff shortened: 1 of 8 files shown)


vctrchu (approved):

packages/app/src/cli/models/extensions/specifications/ui_extension.test.ts
```diff
@@ -241,15 +241,23 @@ describe('ui_extension', async () => {
...
-    test('targeting object accepts allow_direct_linking for target capabilities', async () => {
+    test('targeting object accepts typed and arbitrary JSON target capabilities', async () => {
       const allSpecs = await loadLocalExtensionsSpecifications()
       const specification = allSpecs.find((spec) => spec.identifier === 'ui_extension')!
+      const intercepts = [
+        {
+          event: 'paymentvalidations',
+          blocking: true,
+          applies_to: ['cash'],
+          future_options: {enabled: true, values: ['one', 2, null]},
```
vctrchu: whats `future_options`?

packages/app/src/cli/models/extensions/schemas.ts
```diff
@@ -39,6 +39,7 @@ export const ExtensionsArraySchema = zod.object({
 
 const TargetCapabilitiesSchema = zod.object({
   allow_direct_linking: zod.boolean().optional(),
+  intercepts: zod.any().optional(),
 })
```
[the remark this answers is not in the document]
vctrchu: Response seems decent. Remind me why we had to do a any() here?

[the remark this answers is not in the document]
vctrchu: Ah I see we are keeping the type loose and will tighten up later https://github.com/shop/issues-retail/issues/33985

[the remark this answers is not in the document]
henryStelle: We don't have to use any or unknown. the stronger option is to define a zod object for the expected shape of the intercepts, eg zod.arrray(zod.object({ event, blocking, applies_to ... }}

but that could force us to update the CLI if we change the shape and force devs to update or they would get errors. If we are confident the shape won't change, I can update this to check that shape here instead of zod.unknown()

vctrchu (approved):

packages/app/src/cli/models/extensions/specifications/ui_extension.test.ts
henryStelle: just that, any future field. Proving the shape isn't being validated against a strict schema

vctrchu: Gotcha, if we change the types in the future this unit test should fail

packages/app/src/cli/models/extensions/schemas.ts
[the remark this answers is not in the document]
vctrchu: It might actually be better to just implement the proposed shape now because once we add tighten it for GA, they'll need to update the CLI anyways.

vctrchu:
Looks good thanks!

packages/app/src/cli/models/extensions/schemas.test.ts
```diff
@@ -55,3 +55,21 @@ describe('UIDSchema', () => {
     })
   })
 })
+
+describe('NewExtensionPointsSchema', () => {
+  test.each([
```
vividviolet, line 60: Can we have tests for the success case too?

packages/ui-extensions-server-kit/src/types.ts
```diff
@@ -93,6 +93,7 @@ export interface ExtensionPoint {
   name: string
   description?: string
   assets?: Record<string, Asset>
+  capabilities?: Record<string, unknown>
```
vividviolet: Why is this unknown - can we get the right types?

henryStelle: yes we can. to keep the types in sync I updated the tsconfg to pull the declared types from @shopify/ui-extensions-server-kit but didn't declare it a devDependancy of the app package as there was president and the types were not being re-exported. However, I recognize this might be considered not worth it or maybe it should be a formal dependancy?

aaronschubert0 (approved):

vividviolet (approved):

durga256 (approved):

PR #8350, opened 2026-08-19
Add a default dev-session message for configuration-only extensions

amcaplan:

Config-only extension modules — the ones whose configuration *is* the whole extension, with nothing for the CLI to build, bundle or serve — say nothing at all during `shopify app dev`. `logExtensionEvents` only runs inside `processEvents`, so at startup these modules are invisible: you run `dev`, your module is in the manifest and live on the platform, and the terminal never mentions it.

#8319 fixes this for one extension type (`analytics_app_events`) by registering a local specification whose only purpose is to carry a single string. That works, but it means every future config-only, remote-only extension type needs its own local spec file for the same one line — which is exactly the coupling that remote-only specs exist to avoid.

This PR solves the class instead of the instance: **a default `getDevSessionUpdateMessages` for config-only modules, overridable per specification.** `analytics_app_events` gets its message with no local spec file at all, and so does the next one.

A default implementation on `ExtensionInstance.getDevSessionUpdateMessages`, gated on a capability predicate and on the dev-session lifecycle:

- **A per-spec hook still wins.** If a specification defines `getDevSessionUpdateMessages`, it is used unchanged. The three existing implementers (`app_config_app_home`, `app_config_app_access`, `app_config_app_proxy`) keep their current behaviour.
- **The predicate is a capability check, not an allowlist**: no `appModuleFeatures`, no `clientSteps` deploy group, no build output, and not an app-config module. Read off existing lazy getters, so it stays correct when the platform response rewrites `experience`/`uidStrategy` after the factory has run.
- **It fires once**, on the first successful dev session create. The event type can't signal "first time" — `app-event-watcher.ts` stamps every initial extension as `EventType.Updated` — so the signal is `DevSessionResult.status === 'created'`, which by construction happens exactly once per `dev` run.

The default lives on the instance rather than in `createExtensionSpecification` on purpose. `createConfigExtensionSpecification` and `createContractBasedModuleSpecification` both pass `getDevSessionUpdateMessages` through unconditionally, so an explicit `undefined` reaches `{...defaults, ...spec}` and clobbers any factory-level default — a default written there is a silent no-op for most specs. There is a test covering the override through all three factories.

The hook also gains a second argument, `getDevSessionUpdateMessages(config, context)`, where `context` is a `DevSessionUpdateContext` carrying a `DevSessionUpdateStatus` of `'created' | 'updated'`. `DevSessionResult` reuses that union whole for its success branch, so the two can't drift: renaming a status now breaks the service at compile time.

Responsibility splits across three layers. `handleDevSessionResult` passes its `DevSessionResult` straight to the logger — no reconstructed status literal, so there's no second copy that could disagree with the branch it was written in. `DevSessionLogger.logExtensionUpdateMessages` decides which outcomes are worth speaking about, returning early on the two error branches, which sits next to the `EventType.Deleted` skip it already owned. Errors already have a per-extension voice — `logUserErrors` maps each `UserError` to its owning extension by `uid` and prefixes the line with that handle — so letting specs speak there too would be a second extension-prefixed channel on one event. Only the narrow `DevSessionUpdateContext` reaches the specs, so a status added later to the service's state machine doesn't silently widen what every spec has to handle — and `specification.ts` stays free of any import back into the dev-session service, which would close a cycle through `app-event-watcher.ts` and `extension-instance.ts`.

**One drive-by removal:** the `'aborted'` status is gone. Nothing has constructed it since 4bc22d2c8c ("Only one app-preview update", May 2025) removed the `newBundleController.signal.aborted` check, so the `handleDevSessionResult` branch that debug-logged it was unreachable — the statuses actually produced are `created`, `updated`, `remote-error` and `unknown-error`. Deleting it lets `DevSessionResult` reuse `DevSessionUpdateStatus` whole rather than grafting a third literal onto it, and leaves the logger's guard as an exact match on the two error branches. Behaviour is unchanged because the branch could not run. Happy to split this out into its own PR if you'd rather review it separately.

The three existing implementations ignore the context and are unchanged — the only reason `app_config_app_access.test.ts` and `app_config_app_proxy.test.ts` appear in this diff is that they call the hook directly and now pass the new argument.

#### Which specs this affects

| | |
|---|---|
| **In** (local) | `editor_extension_collection`, `flow_action`, `flow_trigger`, `payments_extension` |
| **In** (remote-only) | every contract-based spec with no features — including `analytics_app_events` |
| **Out** | `channel_config`, `order_attribution_config`, `tax_calculation`, `admin_link` (deploy steps and/or build output), and all app-config modules (already summarised once as `App config updated`) |

### Open questions

- **The copy is provisional and wants a product call.** `Configuration accepted` says only that the platform took the configuration. It deliberately claims nothing about local files, because the predicate admits modules where such a claim would be false: `flow_action` and `flow_trigger` both read a local schema file (`loadSchemaFromPath`, e.g. `flow_action.ts:83`). Anything more descriptive has to stay true for every module the predicate admits, now and as the set grows. Owners of the four affected local specs should also confirm they want the line at all.
- **A deliberate asymmetry worth a second opinion:** a `flow_action` prints `Configuration accepted` while a `checkout_ui_extension` in the same app prints nothing. The rule is "nothing is *built* for this module", not "this module is inert" — buildable modules are already visible through their build output and preview URL, whereas these had no representation in `dev` output at all. That's the intended reading, but it is the boundary most likely to be contested, so it should be an explicit decision rather than a side effect.
- **A localization-only remote-only spec won't get the message**, because `features === ['localization']`. That's a deliberate consequence of the predicate rather than an oversight; widening it would start printing for modules that do have local output.
- **Adjacent bug, deliberately not fixed here:** `createContractBasedModuleSpecification` passes `uidStrategy: spec.uidStrategy` unconditionally, so omitting it clobbers the computed default at `{...defaults, ...spec}` with `undefined`, `buildHandle`'s switch falls through, and `constantize(undefined)` throws. No production caller hits it — `createRemoteOnlySpecification` always passes a strategy — so it's dormant, and fixing it means touching the merge line this PR is specifically designed to avoid. Worth its own PR. It's also a live demonstration, on a second property, of why the default here lives on the instance rather than in the factory.
- **Follow-up, not here:** platform specifications already carry `meta_data` with `description` and `documentation_links` (23 registrations in `shop/world` have it). Sourcing dev-session copy from the platform instead of hardcoding it in the CLI is the better long-term shape, and would let each extension team own its own string. Out of scope here, which is CLI-side copy only.

### How to test your changes?

1. `shopify app dev` on an app with a config-only module (e.g. a Flow action, or an `analytics_app_events` module).
2. On startup, after `✅ Ready, watching for changes`, the module prints `Configuration accepted` under its own handle.
3. Edit and save the module's TOML — it prints `Extension changed` / `✅ Updated dev preview on …` and does *not* repeat `Configuration accepted`.

### Post-release steps

None.

### Checklist

- [x] I've considered possible cross-platform impacts (Mac, Linux, Windows) — terminal output only, no path or process handling
- [x] I've considered possible [documentation](https://shopify.dev) changes — none; no command, flag or schema surface changes
- [x] I've considered analytics changes to measure impact — none
- [x] The change is user-facing — `minor`, changeset added

### DIFF
--- a/packages/app/src/cli/models/extensions/extension-instance.ts
+++ b/packages/app/src/cli/models/extensions/extension-instance.ts
@@ -1,6 +1,11 @@
 import {BaseConfigType, MAX_EXTENSION_HANDLE_LENGTH, MAX_UID_LENGTH} from './schemas.js'
 import {FunctionConfigType} from './specifications/function.js'
-import {DevSessionWatchConfig, ExtensionFeature, ExtensionSpecification} from './specification.js'
+import {
+  DevSessionUpdateContext,
+  DevSessionWatchConfig,
+  ExtensionFeature,
+  ExtensionSpecification,
+} from './specification.js'
 import {SingleWebhookSubscriptionType} from './specifications/app_config_webhook_schemas/webhooks_schema.js'
 import {ExtensionBuildOptions} from '../../services/build/extension.js'
 import {ExtensionUuidsByLocalIdentifier} from '../app/identifiers.js'
@@ -38,6 +43,8 @@ const DEFAULT_WATCH_IGNORE = [
   '**/generated/**',
   '**/.gitignore',
 ]
+
+export const DEFAULT_DEV_SESSION_UPDATE_MESSAGE = 'Configuration accepted'
 
 /**
  * Class that represents an instance of a local extension
@@ -141,6 +148,10 @@ export class ExtensionInstance<TConfiguration extends BaseConfigType = BaseConfi
 
   get outputRelativePath() {
     return this.specification.getOutputRelativePath?.(this) ?? ''
+  }
+
+  get hasNoLocalDevOutput(): boolean {
+    return this.features.length === 0 && !this.hasDeploySteps && this.outputRelativePath === ''
   }
 
   constructor(options: {
@@ -394,9 +405,16 @@ export class ExtensionInstance<TConfiguration extends BaseConfigType = BaseConfi
     }
   }
 
-  async getDevSessionUpdateMessages(): Promise<string[] | undefined> {
-    if (!this.specification.getDevSessionUpdateMessages) return undefined
-    return this.specification.getDevSessionUpdateMessages(this.configuration)
+  async getDevSessionUpdateMessages(context: DevSessionUpdateContext): Promise<string[] | undefined> {
+    if (this.specification.getDevSessionUpdateMessages) {
+      return this.specification.getDevSessionUpdateMessages(this.configuration, context)
+    }
+
+    if (context.status !== 'created') return undefined
+    if (this.isAppConfigExtension) return undefined
+    if (!this.hasNoLocalDevOutput) return undefined
+
+    return [DEFAULT_DEV_SESSION_UPDATE_MESSAGE]
   }
 
   /**
--- a/packages/app/src/cli/services/dev/processes/dev-session/dev-session-logger.test.ts
+++ b/packages/app/src/cli/services/dev/processes/dev-session/dev-session-logger.test.ts
@@ -1,10 +1,39 @@
 import {DevSessionLogger} from './dev-session-logger.js'
 import {UserError} from './dev-session.js'
 import {AppEvent, EventType} from '../../app-events/app-event-watcher.js'
-import {ExtensionInstance} from '../../../../models/extensions/extension-instance.js'
+import {
+  DEFAULT_DEV_SESSION_UPDATE_MESSAGE,
+  ExtensionInstance,
+} from '../../../../models/extensions/extension-instance.js'
+import {createExtensionSpecification} from '../../../../models/extensions/specification.js'
+import {BaseConfigType, BaseSchema} from '../../../../models/extensions/schemas.js'
 import {describe, expect, test, vi, beforeEach} from 'vitest'
 import {JsonMapType} from '@shopify/cli-kit/node/toml'
 import {Writable} from 'stream'
+
+function extensionWithNoLocalDevOutput(handle: string): ExtensionInstance {
+  const specification = createExtensionSpecification({
+    identifier: handle.replace(/-/g, '_'),
+    schema: BaseSchema,
+    appModuleFeatures: () => [],
+    uidStrategy: 'single',
+  })
+  return new ExtensionInstance({
+    configuration: {name: handle, type: specification.identifier, handle} as BaseConfigType,
+    configurationPath: '',
+    directory: '/tmp/test-extension',
+    specification,
+  })
+}
+
+function eventForExtensions(extensions: ExtensionInstance[], type = EventType.Updated): AppEvent {
+  return {
+    app: {configuration: {}} as AppEvent['app'],
+    extensionEvents: extensions.map((extension) => ({type, extension})),
+    path: '',
+    startTime: [0, 0],
+  }
+}
 
 describe('DevSessionLogger', () => {
   let output: string[]
@@ -166,8 +195,36 @@ describe('DevSessionLogger', () => {
 
   describe('logExtensionUpdateMessages', () => {
     test('does nothing when no event is provided', async () => {
-      await logger.logExtensionUpdateMessages()
+      await logger.logExtensionUpdateMessages(undefined, {status: 'created'})
       expect(output).toMatchInlineSnapshot(`[]`)
+    })
+
+    test('does not consult extensions when the dev session errored', async () => {
+      const mockExtension = {
+        getDevSessionUpdateMessages: vi.fn().mockResolvedValue(['This would be logged if the session had succeeded.']),
+        entrySourceFilePath: '',
+        devUUID: '',
+        localIdentifier: '',
+        idEnvironmentVariableName: '',
+        handle: 'test-extension',
+      } as unknown as ExtensionInstance
+
+      const event: AppEvent = {
+        app: {configuration: {}} as any,
+        extensionEvents: [
+          {
+            type: 'updated' as EventType,
+            extension: mockExtension,
+          },
+        ],
+        path: '',
+        startTime: [0, 0],
+      }
+
+      await logger.logExtensionUpdateMessages(event, {status: 'remote-error', error: []})
+
+      expect(output).toMatchInlineSnapshot(`[]`)
+      expect(mockExtension.getDevSessionUpdateMessages).not.toHaveBeenCalled()
     })
 
     test('logs messages', async () => {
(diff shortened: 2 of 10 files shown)


packages/app/src/cli/models/extensions/extension-instance.ts
```diff
@@ -394,9 +415,23 @@ export class ExtensionInstance<TConfiguration extends BaseConfigType = BaseConfi
...
+    // Only acknowledge the module once, on the first successful dev session; later updates stay quiet.
+    if (context.status !== 'created') return undefined
+
+    // App config modules are already summarised together as `App config updated`, so a per-module
+    // line would just be noise.
+    if (this.isAppConfigExtension) return undefined
+
+    // Anything with local dev output is already visible through its build output or preview URL.
+    // Modules with none of that are never mentioned by `dev` otherwise, so acknowledge them here.
+    if (!this.hasNoLocalDevOutput) return undefined
```
amcaplan: Are these comments helpful, or just a distraction? I'm honestly not sure.

packages/app/src/cli/services/dev/processes/dev-session/dev-session-logger.test.ts
```diff
@@ -1,11 +1,44 @@
...
+import {createExtensionSpecification} from '../../../../models/extensions/specification.js'
+import {BaseConfigType, BaseSchema} from '../../../../models/extensions/schemas.js'
 import {describe, expect, test, vi, beforeEach} from 'vitest'
 import {JsonMapType} from '@shopify/cli-kit/node/toml'
 import {Writable} from 'stream'
 
+/**
+ * A real extension instance for a module that builds nothing locally, so that the default
+ * `getDevSessionUpdateMessages` implementation is genuinely exercised rather than mocked away.
+ */
```
amcaplan: This comment is not helpful. It's actually pretty obvious. Looking at you, Claude.

packages/app/src/cli/services/dev/processes/dev-session/dev-session-logger.ts
```diff
@@ -1,6 +1,7 @@
-import {UserError} from './dev-session.js'
+import {DevSessionResult, UserError} from './dev-session.js'
 import {AppEvent, EventType} from '../../app-events/app-event-watcher.js'
 import {ExtensionInstance} from '../../../../models/extensions/extension-instance.js'
+import {DevSessionUpdateContext} from '../../../../models/extensions/specification.js'
```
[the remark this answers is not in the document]
amcaplan: Checked this one carefully, and I don't think there's a cycle to fix here.

**The import edge isn't new.** `main`'s line 1 is already `import {UserError} from './dev-session.js'` — this PR added a second name to an import statement that was already there.

**And nothing is emitted at runtime, before or after.** `UserError` is an `interface`, `DevSessionResult` a `type`, and `DevSessionUpdateContext` an `interface` — all three are type-only. `packages/app` builds with `pnpm tsc -b`, `isolatedModules: false`, and `verbatimModuleSyntax` doesn't appear anywhere in the repo, so TypeScript elides the whole statement. Rebuilt to confirm — `dist/cli/services/dev/processes/dev-session/dev-session-logger.js` imports exactly this:

```js
import { EventType } from '../../app-events/app-event-watcher.js';
import { outputToken, outputContent, outputDebug } from '@shopify/cli-kit/node/output';
import { useConcurrentOutputContext } from '@shopify/cli-kit/node/ui/components';
```

No `./dev-session.js`, no `specification.js`. The only runtime edge is one-directional: `dev-session.js` → `dev-session-logger.js`, importing the `DevSessionLogger` class.

Bundling can't reintroduce it either — `packages/cli`'s `bundle` target has `dependsOn: ["app:build"]`, so esbuild consumes the tsc-emitted `dist` where the import is already gone. Worth spelling out, because esbuild *can't* tell a type import from a value one; it just never sees the TypeScript source here.

**That said, there's a real point underneath this.** Elision is a property of our compiler settings, not of the code. Flip on `verbatimModuleSyntax` and this import starts being emitted, and then it genuinely would be a cycle. But the fix for that is enforcing type-only imports repo-wide — `@typescript-eslint/consistent-type-imports`, which we don't currently have — rather than hand-annotating one file. Right now 32 of 533 files in `packages/app` use `import type` (~6%), so plain imports for types are the prevailing style, including on the line flagged here. Changing just this file buys nothing today and quietly implies a guarantee the other 501 files don't have.

Leaving it as-is. Happy to open a separate PR for the lint rule if people want the guarantee made real.

---
<sub>🤖 Investigation and drafting by Claude Opus 5 (Claude Code), reviewed and posted by @amcaplan.</sub>

packages/app/src/cli/services/dev/processes/dev-session/dev-session.ts
```diff
@@ -11,6 +11,7 @@ import {
 } from '../../../bundle.js'
 import {DevSessionCreateOptions, DevSessionUpdateOptions} from '../../../../utilities/developer-platform-client.js'
 import {AppManifest} from '../../../../models/app/app.js'
+import {DevSessionUpdateStatus} from '../../../../models/extensions/specification.js'
```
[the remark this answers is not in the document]
amcaplan: I think this is covered by [the other comment](https://github.com/Shopify/cli/pull/8350#discussion_r3838660121)

packages/app/src/cli/models/extensions/extension-instance.ts
dmerand: Not helpful IMO

packages/app/src/cli/models/extensions/extension-instance.ts
```diff
@@ -39,6 +44,13 @@ const DEFAULT_WATCH_IGNORE = [
   '**/.gitignore',
 ]
 
+/**
```
dmerand: Nit: comment (most of these comments) are not needed.

dmerand (approved):

packages/app/src/cli/models/extensions/extension-instance.ts
```diff
@@ -39,6 +44,8 @@ const DEFAULT_WATCH_IGNORE = [
   '**/.gitignore',
 ]
 
+export const DEFAULT_DEV_SESSION_UPDATE_MESSAGE = 'Configuration accepted'
```
gonzaloriestra, line 47: I find `accepted` a bit weird... What about `ready` or `active`?

gonzaloriestra (approved):
This is very useful, thanks!

The PR description is too long, please remember to ask for a concise one next time! 🙏

packages/app/src/cli/models/extensions/extension-instance.ts
amcaplan: I think the message makes sense, it means you have a valid configuration and we didn't hit any errors pushing it up

amcaplan:
Screenshot 2026-08-25 at 12 04 12

PR #8360, opened 2026-08-21
Add `shopify app import channel-config` for partner channel spec import

suzkiee:

Legacy channel apps need a migration path onto partner-owned `channel_config` extension specs. Shopify already has app-level FrozenRecord fallback specs for some legacy channels. Partners need a safe way to receive a deployable, partner-owned copy of the **public parts** of that spec.

The CLI team reviewed the proposal ([thread](https://shopify.slack.com/archives/C07UJ7UNMTK/p1789407252436359?thread_ts=1788973623.008519&cid=C07UJ7UNMTK)) and approved this as a new `import` namespace under `app`:

> start implementing this as `import channel-config` in Shopify CLI and we will work on a PR to stack on that moves the existing commands.

This PR adds `app import channel-config` and introduces the `app import` topic. A follow-up PR from the CLI team will migrate the existing `import-extensions` / `import-custom-data-definitions` commands into the same namespace.

Adds a new command:

```sh
shopify app import channel-config [--stdout] [--overwrite]
```

**Flow:**

1. Resolves the app from local context via `linkedAppContext` (standard `--path` / `--client-id` / `--config` / `--reset` flags).
2. Calls a thin transport (`services/import-channel-config/fetch.ts`) that delegates to the canonical Channels-owned exporter (`Channels::PartnerChannelSpecExport`). The CLI **never** transforms or validates spec fields — the backend response *is* the deployable artifact.
3. Writes `extensions/channel-config/specifications/<handle>.toml` (creates directories as needed).

**Behavior details:**

- Refuses to overwrite an existing file unless `--overwrite` is passed.
- `--stdout` prints only the TOML to stdout (pipeable); warnings go to stderr.
- Backend warnings (`{code, message}`, e.g. `automatic_product_feed_management`) render out-of-band — never written into the TOML.
- Failure reason codes map to partner-facing guidance: `no_exportable_frozen_record`, `multiple_exportable_records`, `not_allowlisted`, `contains_no_public_fields`, `invalid_public_schema`.
- Never deploys; success output points at `shopify app deploy` as the next step.

**Files:**

| File | Purpose |
|---|---|
| `packages/app/src/cli/commands/app/import/channel-config.ts` | oclif command (`app:import:channel-config`) |
| `packages/app/src/cli/services/import-channel-config/import.ts` | file/stdout UX + failure mapping |
| `packages/app/src/cli/services/import-channel-config/fetch.ts` | transport + response parsing |
| `packages/app/src/cli/services/import-channel-config/import.test.ts` | 11 service tests (real temp dirs) |
| `packages/app/src/cli/services/import-channel-config/fetch.test.ts` | 10 transport tests (GID URL, status handling, malformed bodies) |
| `packages/app/src/cli/index.ts` | command registration |
| `packages/cli/package.json` | new `app:import` topic description |
| Generated files | oclif manifest, README, dev-docs data |

## Transport / backend contract

The command calls `GET app_management/unstable/organizations/:orgId/apps/:id/channel_spec_export.json`:

- Success: `{handle, filename, toml, warnings}`
- Well-formed failure: `422 {error: "not_exportable_yet", reason}`
- `404` is reserved for "endpoint unavailable / app not found" and surfaces as an abort.

The corresponding App Management endpoint is being added in shop/world (rework of shop/world#2051080).

## Design constraints

- Export is **app-level and shop-independent** — app-owner/collaborator auth only; no shop/admin/runtime auth contexts.
- Internal-only fields are omitted entirely server-side via a public-field allowlist. Generated TOML must pass the public `channel_config` validation contract; the exporter fails closed otherwise.
- Single-spec legacy apps only for MVP (multi-spec apps out of scope). A `--specification-handle <handle>` flag is reserved for future multi-spec support and intentionally **not** included here.
- No comments, provenance, or warnings inside the TOML file.

### DIFF
--- a/packages/cli/README.md
+++ b/packages/cli/README.md
@@ -4,6 +4,7 @@
 * [`shopify app bulk cancel`](#shopify-app-bulk-cancel)
 * [`shopify app bulk execute`](#shopify-app-bulk-execute)
 * [`shopify app bulk status`](#shopify-app-bulk-status)
+* [`shopify app channel-spec generate`](#shopify-app-channel-spec-generate)
 * [`shopify app config link`](#shopify-app-config-link)
 * [`shopify app config pull`](#shopify-app-config-pull)
 * [`shopify app config use [config] [flags]`](#shopify-app-config-use-config-flags)
@@ -363,6 +364,62 @@ DESCRIPTION
   (https://shopify.dev/docs/api/usage/bulk-operations/imports).
 
   Use "`bulk execute`" (https://shopify.dev/docs/api/shopify-cli/app/app-bulk-execute) to start a new bulk operation.
+```
+
+## `shopify app channel-spec generate`
+
+Generate a channel spec TOML file from the Shopify-authored default.
+
+```
+USAGE
+  $ shopify app channel-spec generate [--auth-alias <value>] [--client-id <value> | -c <value>] [--no-color] [--overwrite]
+    [--path <value>] [--reset | ] [--stdout] [--verbose]
+
+FLAGS
+  -c, --config=<value>
+      The name of the app configuration.
+      [env: SHOPIFY_FLAG_APP_CONFIG]
+
+  --auth-alias=<value>
+      Alias of the Shopify account to use for authentication.
+      [env: SHOPIFY_FLAG_AUTH_ALIAS]
+
+  --client-id=<value>
+      The Client ID of your app.
+      [env: SHOPIFY_FLAG_CLIENT_ID]
+
+  --no-color
+      Disable color output.
+      [env: SHOPIFY_FLAG_NO_COLOR]
+
+  --overwrite
+      Overwrite the existing channel spec file if one already exists.
+      [env: SHOPIFY_FLAG_OVERWRITE]
+
+  --path=<value>
+      The path to your app directory.
+      [env: SHOPIFY_FLAG_PATH]
+
+  --reset
+      Reset all your settings.
+      [env: SHOPIFY_FLAG_RESET]
+
+  --stdout
+      Print the generated TOML to stdout instead of writing it to a file.
+      [env: SHOPIFY_FLAG_STDOUT]
+
+  --verbose
+      Increase the verbosity of the output. May include sensitive data.
+      [env: SHOPIFY_FLAG_VERBOSE]
+
+DESCRIPTION
+  Generate a channel spec TOML file from the Shopify-authored default.
+
+  Generates a deployable `channel_config` extension spec from the Shopify-authored default channel specification for
+  your app.
+
+  The generated TOML file contains only public `channel_config` fields. Review it, commit it to your app, then deploy it
+  with `shopify app deploy`. This command never deploys the spec itself.
 ```
 
 ## `shopify app config link`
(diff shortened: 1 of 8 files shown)


packages/app/src/cli/services/import-channel-config/fetch.ts
```diff
@@ -0,0 +1,125 @@
...
+  const appId = numericIdFromGid(remoteApp.id) ?? remoteApp.id
+  const url = `https://${fqdn}/app_management/unstable/organizations/${encodeURIComponent(
+    remoteApp.organizationId,
+  )}/apps/${encodeURIComponent(appId)}/channel_spec_export.json`
+  const token = (await developerPlatformClient.session()).token
+
+  const response = await shopifyFetch(url, {
+    method: 'GET',
+    headers: appManagementHeaders(token),
+  })
```
suzkiee, line 55: Should this mirror other App Management only operations and have the HTTP call be moved in a DeveloperPlatformClient method?

packages/cli/README.md
```diff
@@ -1512,6 +1513,129 @@ EXAMPLES
   $ shopify app graphiql --store shop.myshopify.com --port 9123
 ```
 
+## `shopify app import channel-config`
```
tauthomas01: Small question: should the command be visible to every CLI user? Since we're targeting a small group of channel partners, I am not sure if we want to expose this to the public.

packages/app/src/cli/services/import-channel-config/fetch.ts
tauthomas01, line 1: From my [other comment in the other PR](https://meteorite.shopify.io/repos/shop/world/pulls/2051080/files?file=areas%2Fcore%2Fshopify%2Fcomponents%2Fapps%2Fapp%2Fcontrollers%2Fapp_management%2Fchannel_spec_exports_controller.rb&comment=2150713262147), should we add the 426 error code where someone has an outdated CLI version?

packages/app/src/cli/services/import-channel-config/import.test.ts
```diff
@@ -0,0 +1,236 @@
...
+      await importChannelConfig(testOptions(app))
+
+      // Then
+      const extensionConfigPath = joinPath(tmpDir, CHANNEL_SPEC_EXTENSION_DIRECTORY, 'shopify.extension.toml')
+      await expect(readFile(extensionConfigPath)).resolves.toContain('type = "channel_config"')
+      expect(outputMock.info()).toContain('shopify.extension.toml')
+    })
+  })
+
+  test('does not replace an existing shopify.extension.toml', async () => {
```
tauthomas01, line 216: If we have an existing `extensions/channel-config/shopify.extension.toml` containing another type than "channel_config":

```toml
name = "Some other extension"
type = "ui_extension"
handle = "something-else"
```

That means we are preserving that file and we write the channel specification under `extensions/channel-config/specifications/example.toml`. The `specifications/` directory is not treated as `channel-config` content so it would likely not be included as a channel-config extension in the deployment bundle. 

Should we address this case?

tauthomas01 (commented):
Looks good! I left some minor comments regarding the channel context. I am not the best person to review the CLI modules/architecture so it would be best to have someone from that team to review.

packages/app/src/cli/services/import-channel-config/fetch.ts
suzkiee: Yes!

packages/app/src/cli/services/import-channel-config/import.test.ts
suzkiee: `ensureExtensionConfig` now parses an existing `shopify.extension.toml` and aborts with guidance unless `type = "channel_config"` ("…already defines an extension of type \"ui_extension\", so the imported channel spec would not be deployed. Move that extension to another directory (or change its type…)"). Unparseable configs abort too. I also moved the check ahead of the spec write so an abort doesn't leave a stray `specifications/<handle>.toml` behind. Tests added for both branches.

packages/cli/README.md
suzkiee: No it shouldn't be visible to everyone

packages/app/src/cli/services/import-channel-config/fetch.ts
suzkiee: Done in 07a4651: 426 now surfaces the server's `reason` with "Upgrade the Shopify CLI to the latest version and re-run the command" instead of falling through to the generic "likely temporary" message. Test added.

---
_Generated with Pi._

tauthomas01 (approved):
Thank you for addressing my comments. Looks good from Channels side.

packages/app/src/cli/services/import-channel-config/fetch.ts
suzkiee: Done in 6ab0fc5 — `channelSpecExport(app)` now lives on `DeveloperPlatformClient` / `AppManagementClient` (URL builder in cli-kit's app-management module next to `appManagementAppLogsUrl`), and `fetch.ts` only applies the status contract to the returned `{status, ok, body}`. Tests mock the client instead of HTTP.

---
_Generated with Pi._

PR #8387, opened 2026-08-24
Fix apps failing validation when a configuration module contract requires its own section

### LINKED ISSUE #8386: shopify app deploy blocked by phantom [events]: Required — Partner Dashboard shows no Events feature enabled
**CLI version:** reproduced identically on 4.7.0, 4.6.1, 4.0.0, and 3.94.3

**Problem**

`shopify app deploy` (and `shopify app config validate`) fails with:

```
[events]: Required
```

even though `shopify.app.toml` has no `[events]` section and never has. I checked the Partner Dashboard for this app — there's no mention of "Events" / "Next-Gen Events" anywhere, no toggle, nothing to enable or configure.

**Steps to reproduce**

1. Have an existing app with a `shopify.app.toml` that has never used `[events]` (just a plain `[webhooks]` section with `api_version` set, no subscriptions).
2. Run `shopify app config validate` (or `shopify app deploy`).
3. Observe:
   ```
   │  Validation errors found.                                                    │
   │  • [events]: Required                                                        │
   ```

**What I tried**

1. Confirmed it's not a local-CLI-version issue — reproduced the identical error on `@shopify/cli` **4.7.0, 4.6.1, 4.0.0, and 3.94.3** (full major-version spread) via `npx @shopify/cli@<version> app config validate`. Since the required-field set differs across those releases, this points to the check coming from the Partners backend, not code bundled in the CLI.
2. Added a minimal `[events]` block to `shopify.app.toml`:
   ```toml
   [events]
   api_version = "2026-07"
   ```
   This flips to two contradictory errors:
   ```
   [events.subscription]: Required
   Unsupported section(s) in app configuration: events
   ```
   So the backend simultaneously demands the section *and* rejects it as unsupported for this app — no toml content can satisfy both at once.

**Expected behavior**

`shopify app deploy` should succeed without requiring a section for a feature that isn't enabled for the app/org, or the error message should point at what actually needs enabling instead of a self-contradictory pair of validation errors.

**Context**

- App has an existing, working `[webhooks]` section (`api_version` set, no subscriptions configured yet).
- Partner Dashboard shows no "Events" (developer preview) toggle or notice for this app.

kutaysahindev:

Fixes #8386

Since 2026-08-24, `shopify app deploy`, `shopify app dev`, and `shopify app config validate` fail with `[events]: Required` for apps whose `shopify.app.toml` has never had an `[events]` section. Adding the section only produces a contradictory pair of errors (`[events.subscription]: Required` + `Unsupported section(s) in app configuration: events`), so no toml content can pass. The failure reproduces identically on CLI 3.94.3 through 4.7.0, which points at the remote specification contracts rather than any single release.

Root cause: `unifiedConfigurationParserFactory` validates the **whole app config object** against every configuration-experience spec's remote JSON-schema contract, even when that spec's section is absent from the toml. The platform's `events` spec contract now ships a root-level `"required": ["events"]`, so every app without an `[events]` section fails validation before the loader's existing "empty config → no module for this spec" short-circuit (`createConfigExtensionInstances`) can run. Any configuration spec shipping a root-level `required` for its own section would break all apps the same way.

- `packages/app/src/cli/utilities/json-schema.ts`: for configuration-experience specs, drop the contract's **root-level** `required` before compiling the AJV validator. Whether a config module exists is decided by the presence of its section in the toml (the loader's model), so the contract must not force the section into existence. Nested `required` properties (e.g. `events.api_version`, `events.subscription`) are still fully enforced once the section is present, and extension-experience specs — where the toml *is* the module config — are untouched.
- `packages/app/src/cli/models/extensions/specifications/transform/app_config_events.ts`: the events reverse transform returned `{events: {}}` when there is no events config, instead of omitting the section like every other config transform (e.g. `transformToWebhookConfig`). That empty object leaked into linked configurations and deploy diffs. It now returns `{}`.
- Tests for both changes, including an end-to-end one through `fetchSpecifications` using the existing `remote_only_extension_schema_config_style` fixture (configuration experience + root-level `required`).
- Changeset (`@shopify/cli`, patch).

### How to test your changes?

1. `pnpm vitest run packages/app/src/cli/utilities/json-schema.test.ts packages/app/src/cli/models/extensions/specifications/transform/app_config_events.test.ts packages/app/src/cli/services/generate/fetch-extension-specifications.test.ts`
2. To reproduce the incident end-to-end: with a development store app whose `shopify.app.toml` has no `[events]` section, run `shopify app config validate` on `main` (fails with `[events]: Required` while the platform serves the current events contract) and again with this branch (passes).
3. Regression check: an app **with** an `[events]` section still gets nested contract validation (`[events.api_version]: Required` etc.).

### Checklist

### DIFF
--- a/.changeset/config-contract-root-required.md
+++ b/.changeset/config-contract-root-required.md
@@ -1,0 +1,5 @@
+---
+'@shopify/cli': patch
+---
+
+Fix "[events]: Required" validation errors for apps whose configuration has no [events] section
--- a/packages/app/src/cli/models/extensions/specifications/transform/app_config_events.test.ts
+++ b/packages/app/src/cli/models/extensions/specifications/transform/app_config_events.test.ts
@@ -192,4 +192,9 @@ describe('transformToEventsConfig', () => {
       },
     })
   })
+
+  test('omits the events section when there is no events config', () => {
+    expect(transformToEventsConfig({})).toEqual({})
+    expect(transformToEventsConfig({events: {}})).toEqual({})
+  })
 })
--- a/packages/app/src/cli/models/extensions/specifications/transform/app_config_events.ts
+++ b/packages/app/src/cli/models/extensions/specifications/transform/app_config_events.ts
@@ -54,8 +54,11 @@ export function transformToEventsConfig(content: object) {
     return rest
   })
 
-  const events =
-    (apiVersion ?? cleanedSubscriptions) ? {api_version: apiVersion, subscription: cleanedSubscriptions} : {}
+  // Omit the section entirely when there is no events config, so an empty `events` object
+  // isn't injected into linked configurations or deploy diffs
+  if (apiVersion === undefined && cleanedSubscriptions === undefined) {
+    return {}
+  }
 
-  return {events}
+  return {events: {api_version: apiVersion, subscription: cleanedSubscriptions}}
 }
--- a/packages/app/src/cli/services/generate/fetch-extension-specifications.test.ts
+++ b/packages/app/src/cli/services/generate/fetch-extension-specifications.test.ts
@@ -106,4 +106,33 @@ describe('fetchExtensionSpecifications', () => {
     expect(withoutLocalization?.appModuleFeatures()).toEqual([])
     expect(withLocalization?.appModuleFeatures()).toEqual(['localization'])
   })
+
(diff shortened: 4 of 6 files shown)


kutaysahindev:
I have signed the CLA!

PR #8397, opened 2026-08-25
Add a create store option to the app dev store picker

dmerand:
> [!IMPORTANT]
> **Stacked PR.** Base is `dlm-app-dev-inline-store-creation` (#8323), not `main`. This PR only shows the store-picker option on top of the inline creation flow that #8323 adds. Do not merge before #8323.

Closes shop/issues-develop#23597

#8323 lets `app dev` create a dev store inline, but only when the organization has zero stores. Developers with existing stores still have to leave the CLI to create a new one. Per the [command-scope decision](https://github.com/shop/issues-develop/issues/23597#issuecomment-5413800138), the store picker in `app dev` should also offer a "create store" option.

Adds a `Create a new dev store` choice to the `app dev` store picker for App Management organizations that already have stores. Selecting it runs the same name/plan/create/poll flow that #8323 added for the zero-store case.

- New `StoreCreationMode` value `'selection-option'`; only `app dev` opts in. All other callers keep the default `'disabled'`, and `'when-empty'` behavior is unchanged.
- With one or more returned stores, the picker shows those stores plus the create choice. Search finds other stores. One store no longer auto-selects when the create choice is available. The choice persists through search, and an initial store remains selectable after the search input is cleared.
- Zero stores still create directly (no picker), as in #8323.
- Cap handling: with existing stores, a capped organization keeps normal store selection and cancellation; the create choice is hidden. A partial one-store page opens a searchable picker. The flow checks the cap again before it asks for the store name or plan. With zero usable stores, a capped organization must first make a slot available in Dev Dashboard, then use the finite creation sequence from #8323. The cap query still fails open and never runs for Partners or disabled callers.
- Non-interactive runs never prompt or create. With zero stores, the CLI directs the developer to run `shopify store create dev` with the required flags, then `shopify app dev --store <store-domain>`. Multiple returned stores or a partial page uses explicit `shopify app dev --store <store-domain>` guidance. One confirmed store still auto-selects, and explicit or cached selection still bypasses the picker.
- Partners keeps the dashboard link/reload flow.
- This PR does not define JSON output, NDJSON, an event stream, or a prompt protocol for `app dev`. The [CLI JSON outputs design](https://docs.google.com/document/d/1M2YZHKJFFgZo0O8TvBG0NMJGDaiDu98xWOeUxW2QMFM/edit?tab=t.0#heading=h.ixauxxv1jr68) defers long-lived commands such as `app dev` and their prompt protocol.

### How to test your changes?

1. Link an app in an App Management organization that has at least one dev store.
2. Run `shopify app dev --reset`. The store picker lists your stores plus `Create a new dev store`.
3. Pick an existing store: dev starts as before.
4. Run `shopify app dev --reset` again and pick `Create a new dev store`: enter a name, pick a plan, and dev continues on the new store after the success banner.

Manual tophat: this picker-entry behavior was tophatted before the stack split, on the pre-split equivalent SHA `4e85e16c` (multi-store list with create choice, name/basic plan, single success banner, preview ready, explicit `--store` and cached-store bypass, store list stays dev-type, standalone `store create dev --json` unaffected, cleanup). The current child SHA has not been live-tophatted yet.

tophat screenshot

### Checklist

### DIFF
--- a/packages/app/src/cli/prompts/dev.ts
+++ b/packages/app/src/cli/prompts/dev.ts
@@ -70,6 +70,7 @@ interface SelectStorePromptOptions {
   hasMorePages?: boolean
   showDomainOnPrompt: boolean
   onCreateStoreWhenEmpty?: () => Promise<OrganizationStore | undefined>
+  onCreateStore?: () => Promise<OrganizationStore | undefined>
 }
 
 interface ExtraAutoCompletePropsForStoreSelect {
@@ -82,9 +83,10 @@ export async function selectStorePrompt({
   onSearchForStoresByName,
   showDomainOnPrompt = true,
   onCreateStoreWhenEmpty,
+  onCreateStore,
 }: SelectStorePromptOptions): Promise<OrganizationStore | undefined> {
   if (stores.length === 0) return onCreateStoreWhenEmpty?.()
-  if (stores.length === 1) {
+  if (stores.length === 1 && !onCreateStore) {
     outputCompleted(`Using your default dev store, ${stores[0]!.shopName}, to preview your project.`)
     return stores[0]
   }
@@ -99,6 +101,11 @@ export async function selectStorePrompt({
 
   let currentStores = stores
   const storesById = new Map(stores.map((store) => [store.shopId, store]))
+  const createStoreChoice = '__create_new_dev_store__'
+  const choices = () => [
+    ...currentStores.map(storeToChoice),
+    ...(onCreateStore ? [{label: 'Create a new dev store', value: createStoreChoice}] : []),
+  ]
 
   const extraAutocompletePromptProps: ExtraAutoCompletePropsForStoreSelect = {}
   if (onSearchForStoresByName) {
@@ -110,7 +117,7 @@ export async function selectStorePrompt({
       }
 
       return {
-        data: currentStores.map(storeToChoice),
+        data: choices(),
         meta: {
           hasNextPage: result.hasMorePages,
         },
@@ -120,10 +127,11 @@ export async function selectStorePrompt({
 
   const id = await renderAutocompletePrompt({
     message: 'Which store would you like to use to view your project?',
-    choices: currentStores.map(storeToChoice),
+    choices: choices(),
     hasMorePages,
     ...extraAutocompletePromptProps,
   })
+  if (id === createStoreChoice) return onCreateStore?.()
   return storesById.get(id)
 }
 
--- a/packages/app/src/cli/prompts/dev.ts
+++ b/packages/app/src/cli/prompts/dev.ts
@@ -86,7 +86,7 @@ export async function selectStorePrompt({
   onCreateStore,
 }: SelectStorePromptOptions): Promise<OrganizationStore | undefined> {
   if (stores.length === 0) return onCreateStoreWhenEmpty?.()
-  if (stores.length === 1 && !onCreateStore) {
+  if (stores.length === 1 && !hasMorePages && !onCreateStore) {
     outputCompleted(`Using your default dev store, ${stores[0]!.shopName}, to preview your project.`)
     return stores[0]
   }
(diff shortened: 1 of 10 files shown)


packages/app/src/cli/prompts/dev.ts
```diff
@@ -116,10 +121,11 @@ export async function selectStorePrompt({
 
   const id = await renderAutocompletePrompt({
     message: 'Which store would you like to use to view your project?',
-    choices: currentStores.map(storeToChoice),
+    choices: choices(),
     hasMorePages,
     ...extraAutocompletePromptProps,
   })
+  if (id === createStoreChoice) return onCreateStore?.()
   return currentStores.find((store) => store.shopId === id)
```
craigmichaelmartin, line 129: Clearing the search box restores the initial `choices` but not `currentStores` — should the id be resolved against every store seen, since a miss becomes a silent `CancelExecution` at `select-store.ts:68`?

packages/app/src/cli/services/dev/select-store.ts
```diff
@@ -21,27 +21,37 @@ export async function selectStore(
...
-      const name = await devStoreNamePrompt()
-      const plan = await devStorePlanPrompt()
-      const domain = await createDevStore({name, plan, organization: org, json: false, summary: false})
-      const createdStore = await waitForCreatedStoreByDomain(org, domain, developerPlatformClient)
-      renderSuccess({headline: `Development store "${createdStore.shopName}" created successfully.`})
-      return createdStore
+    onCreateStore = createStoreInline
+  } else if (storeCreationEnabled && storeCreationMode === 'selection-option' && isTTY()) {
+    // A capped organization keeps normal store selection without the create choice.
+    if (!(await devStoreCapReached(org.id, developerPlatformClient))) {
```
craigmichaelmartin, line 53: `devStoreCapReached` fails open — should this move inside `createStoreInline` so a cap error aborts before the name/plan prompts?

craigmichaelmartin (approved):

packages/app/src/cli/prompts/dev.ts
dmerand: Fixed in parent f6d42d48d6 and retained in restacked head 8626ddcd82. The prompt now keeps an ID map initialized from the first page and updated by each search. A regression test searches, then returns an initial-store ID without another search call, which models clearing the input.

packages/app/src/cli/services/dev/select-store.ts
dmerand: Fixed in restacked head 8626ddcd82. I kept the pre-picker cap check so capped organizations do not see an unusable create choice, and added a second check at the start of `createStoreInline`, before name, plan, or mutation work. The race test returns false before the picker and true in the callback, then verifies no prompts or creation run.

packages/app/src/cli/services/dev/select-store.test.ts
```diff
@@ -115,7 +115,7 @@ describe('selectStore', async () => {
     ).resolves.toEqual(STORE1)
     expect(devStoreCapReached).not.toHaveBeenCalled()
     expect(createDevStore).not.toHaveBeenCalled()
-    expect(vi.mocked(selectStorePrompt).mock.calls[0]?.[0]).not.toHaveProperty('onCreateStoreWhenEmpty')
+    expect(vi.mocked(selectStorePrompt).mock.calls[0]?.[0]).not.toHaveProperty('onCreateStore')
```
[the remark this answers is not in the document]
dmerand: Fixed in 24981cb15b. The four disabled/non-applicable creation cases now assert that both `onCreateStoreWhenEmpty` and `onCreateStore` are absent. This keeps the zero-store handler out of non-interactive single-store, Partners, existing-store `when-empty`, and disabled-mode paths. The focused select-store suite passes 33 tests, and App lint passes.

PR #8399, opened 2026-08-25
Add app subscription migration commands

tyler-eon:
## Why

Partners migrating legacy manual-billing subscriptions to Shopify-managed app pricing need a safe, scriptable CLI workflow.

## What changed

Adds four core Shopify CLI commands:

- `shopify app subscription-migrations schedule`
- `shopify app subscription-migrations unschedule`
- `shopify app subscription-migrations status`
- `shopify app subscription-migrations cancel`

The commands are hidden from normal CLI help, command listings, generated README, Shopify.dev command docs, and release notes while the feature remains in early access. They remain directly invokable by users who know the command path; backend rollout and authorization gates remain the security boundary.

The commands:

- use the standard linked-app context, with an optional `--client-id` override;
- accept migration CSVs through `--input`, stdin, or `--input -`;
- validate the complete CSV before making a network mutation;
- normalize and sort shop GIDs, then create deterministic batches of at most 250 shops;
- generate unique internal per-invocation batch keys without exposing idempotency controls;
- preserve accepted operation GIDs and structured failures in one typed result;
- support repeated `--id` flags for multi-batch status and cancellation;
- show visible polling progress for watched submissions and status checks;
- keep JSON stdout to one machine-readable document while polling progress is rendered on stderr.

`cancel` stops an operation from processing additional shops. It does not undo shops that have already been scheduled or migrated; `unschedule` is the separate reversal workflow.

## Top-hatting

> [!CAUTION]
> These commands change real app-subscription migration state. Use only a controlled test app and shop, and confirm the intended plan and shop before submitting.

### Prerequisites

1. Use a controlled app that is initially configured for manual billing.
2. Install the app on a controlled test shop and create/approve a recurring manual-billing subscription.
3. Configure the app's intended Shopify App Pricing target plan and record its plan handle.
4. Switch the app to Shopify App Pricing so the legacy subscription is eligible for migration.
5. Confirm the Partners CLI migration schema and required backend rollout gates are available for the app/organization.
6. Check out this branch and install/build Shopify CLI with a supported Node version.

```bash
pnpm install
pnpm --filter @shopify/app build
pnpm --filter @shopify/cli build
```

Run commands from the app project. To select another accessible app explicitly, add `--client-id <client-id>`.

### Schedule one subscription

Create `/tmp/subscription-migrations.csv`:

```csv
shop_id,target_plan_handle,price_behavior,notification
gid://shopify/Shop/123456789,target-plan-handle,HONOR_BILLING_PRICE,WHEN_REQUIRED
```

Submit and watch the operation:

```bash
pnpm shopify app subscription-migrations schedule \
  --input /tmp/subscription-migrations.csv \
  --force \
  --watch
```

Preserve every operation GID printed by the command. `COMPLETED` means processing is terminal; inspect the per-shop result code before treating the migration as successful.

### Check status

```bash
pnpm shopify app subscription-migrations status \
  --id 'gid://shopify/AppSubscriptionMigrationOperation/123' \
  --watch
```

Repeat `--id` for every operation returned by a multi-batch submission. Add `--json` to verify that stdout contains one structured document while progress remains on stderr.

### Cancel unprocessed work

```bash
pnpm shopify app subscription-migrations cancel \
  --id 'gid://shopify/AppSubscriptionMigrationOperation/123'
```

Cancellation prevents additional shops from being processed. It does not reverse shops that were already scheduled or migrated.

### Unschedule a reversible migration

You can reuse `/tmp/subscription-migrations.csv` from earlier. As long as the input has a `shop_id` column, it will correctly create an unscheduling job. Or, you can create a separate file `/tmp/subscription-migrations-unschedule.csv`:

```csv
shop_id
gid://shopify/Shop/123456789
```

Then run:

```bash
pnpm shopify app subscription-migrations unschedule \
  --input /tmp/subscription-migrations-unschedule.csv \
  --force \
  --watch
```

Only run this while the migration remains reversible.

## Backend dependency

Uses the Partners CLI GraphQL fields introduced by Shopify/partners#63097:

- `appSubscriptionMigrationOperationCreate`
- `appSubscriptionMigrationOperation`
- `appSubscriptionMigrationOperationCancel`

The app Client ID is a selector rather than authorization. Partners derives organization membership from the authenticated CLI session and enforces migration permissions and app scoping.

## Testing

- `pnpm vitest run packages/app/src/cli` — 2,258 tests passed
- focused subscription migration suite — 168 tests passed
- `pnpm --filter @shopify/app lint`
- `pnpm --filter @shopify/app build`
- `pnpm knip`
- CLI command discovery and help for all four commands
- command-tree snapshot verification
- malformed file/stdin CSV validation before authentication or network access
- generated README, oclif manifest, and Shopify.dev command documentation verification


packages/app/src/cli/services/subscription-migrations/submit-migration-plan.ts
```diff
@@ -0,0 +1,109 @@
...
+export interface MigrationSubmission {
+  clientId: string
+  action: MigrationPlan['action']
+  rootIdempotencyKey: string
+  inputDigest: string
+  total: number
+  operations: SubmittedMigrationOperation[]
+}
+
+export class MigrationSubmissionError extends Error {
```
dmerand: Partners `userErrors` are expected API rejections, but this ordinary `Error` will use the unexpected/unhandled failure path. Could we separate the two cases this type currently represents? Use an `AbortError` subclass, while preserving the partial-submission fields, when `userErrors` is non-empty. A response with neither an operation nor user errors is an invariant failure and should use `BugError`. Please also assert both classifications in the tests. This follows `docs/cli/error_handling.md`.

packages/app/src/cli/utilities/developer-platform-client/partners-client.ts
```diff
@@ -22,10 +22,8 @@ import {ensureAuthenticatedPartners, Session} from '@shopify/cli-kit/node/sessio
 import {RequestModeInput} from '@shopify/cli-kit/node/http'
 
 /**
- * The Partners-backed client now exists solely to run the legacy extension migrations
- * (see {@link MigrationDeveloperPlatformClient}). Every other developer platform operation is
- * served by the App Management client, so only the migration mutations and the session/request
- * infrastructure they depend on remain here.
+ * The Partners-backed client serves migration workflows, including legacy extension migrations
```
dmerand: The Partners-only subscription-migration path has been vetted. Could we preserve why this is an intentional exception rather than only list the callers? For example:

```ts
/**
 * This client is limited to migration operations that are not available through the App Management API:
 * legacy extension migrations (see {@link MigrationDeveloperPlatformClient}) and app subscription migrations.
 * New developer platform operations should use {@link AppManagementClient} unless they are Partners-only.
 */
```

.changeset/quiet-badgers-poll.md
```diff
@@ -0,0 +1,5 @@
+---
+'@shopify/app': patch
```
dmerand: Nit: could we fold this into `schedule-subscription-migrations.md`? Polling progress is part of the same unreleased feature as the new commands, so a separate patch changeset creates an artificial second changelog entry.

packages/app/src/cli/services/subscription-migrations/submit-migration-plan.ts
dmerand: This class currently combines two materially different cases:

1. Partners completed the request and returned `userErrors`, potentially after accepting one or more operations.
2. Partners returned neither an operation nor user errors, which indicates an incomplete or invalid response.

The first case contains useful domain data; the second is a protocol/invariant failure. Could we represent them separately, either as a discriminated result or separate domain failure types, while letting the command layer preserve today’s output and exit behavior?

This matters for the planned typed JSON output work: a partial submission must become one result containing both the accepted operations and failure details. The current print-partial-result-then-rethrow flow would otherwise become two JSON documents after common JSON error handling is added.

packages/app/src/cli/services/subscription-migrations/cancel-operations.ts
```diff
@@ -0,0 +1,26 @@
...
+export async function cancelMigrationOperations({
+  clientId,
+  operationIds,
+  cancelOperation = cancelMigrationOperation,
+}: CancelMigrationOperationsOptions): Promise<MigrationOperation[]> {
+  return Promise.all(
+    operationIds.map(async (operationId) => {
+      const payload = await cancelOperation({clientId, operationId})
+      if (payload.userErrors.length > 0) {
+        throw new AbortError(payload.userErrors.map(({message}) => message).join('\n'))
```
dmerand: Each requested cancellation completes independently, but `Promise.all` rejects the whole call when one ID returns `userErrors`. The caller then receives none of the successful cancellation results. Could this return a per-ID outcome—such as the operation ID, optional operation, and optional structured failure—and let the command layer decide how to render it and set the exit status?

This prevents data loss today. It also lets the planned typed JSON adapter emit successful and failed cancellations together as one result instead of choosing between discarding successes and throwing a second error. The no-operation/no-user-errors case should remain a distinct protocol failure.

packages/app/src/cli/utilities/developer-platform-client/partners-client.ts
dmerand: The Partners-only subscription-migration path has been vetted. Could we preserve why this is an intentional exception rather than only list the callers? For example:

```ts
/**
 * This client is limited to migration operations that are not available through the App Management API:
 * legacy extension migrations (see {@link MigrationDeveloperPlatformClient}) and app subscription migrations.
 * New developer platform operations should use {@link AppManagementClient} unless they are Partners-only.
 */
```

.changeset/quiet-badgers-poll.md
dmerand: Non-blocking: could we fold this into `schedule-subscription-migrations.md`? Polling progress is part of the same unreleased feature as the new commands, so a separate patch changeset creates an artificial second changelog entry.

dmerand (commented):
I think there are a couple of minor changes we can make to structure here to better-align with repo guidance + also prepare these commands for incoming changes.

Context: Shopify CLI is moving finite commands toward [typed result contracts](https://docs.google.com/document/d/1M2YZHKJFFgZo0O8TvBG0NMJGDaiDu98xWOeUxW2QMFM/edit?tab=t.0#heading=h.ixauxxv1jr68). In that model, command execution returns structured domain data, while a CLI adapter decides whether to render terminal text or encode JSON. This lets the same execution support other adapters later and ensures `--json` emits exactly one document, including when useful data exists alongside an unsuccessful outcome.

tyler-eon:
@dmerand 

Submission and cancellation execution now return typed domain results, while command-local codecs and presenters own JSON/text rendering and exit status. Expected data-bearing failures produce one structured result and one JSON document; protocol/invariant failures remain exceptions. I believe this work should now align better with the document you referenced.

.changeset/quiet-badgers-poll.md
tyler-eon: Moved.

packages/app/src/cli/utilities/developer-platform-client/partners-client.ts
tyler-eon: Comment reworked as suggested.

packages/app/src/cli/services/subscription-migrations/cancel-operations.ts
tyler-eon: This should be resolved now as part of the structured response data change, where now we capture user errors and place those into a domain-specific result structure to ensure all values are "valid" and don't cause the promise to fail if the operations are only partially successful.

packages/app/src/cli/services/subscription-migrations/submit-migration-plan.ts
tyler-eon: This should be resolved now as part of the structured response data change.

packages/app/src/cli/commands/app/subscription-migrations/schedule.ts
```diff
@@ -0,0 +1,60 @@
+import {submissionFlags} from './flags.js'
+import {presentAcceptedMigrationSubmission, presentMigrationSubmissionResult} from './result-presenter.js'
+import {resolveSubscriptionMigrationClientId} from '../../../services/subscription-migrations/resolve-client-id.js'
+import {runSubmissionCommand} from '../../../services/subscription-migrations/run-submission-command.js'
+import BaseCommand from '@shopify/cli-kit/node/base-command'
+
+export default class Schedule extends BaseCommand {
```
gonzaloriestra, line 7: Could we extend from `AppLinkedCommand` and use `linkedAppContext()` as other app commands? App loading and client-ID resolution would be more consistent

packages/app/src/cli/services/subscription-migrations/run-submission-command.ts
```diff
@@ -0,0 +1,76 @@
...
+  if (submissionResult.status === 'failed' || !options.watch) return submissionResult
+
+  options.onSubmissionAccepted?.(submissionResult.submission)
+  const terminalOperations = await watchMigrationOperations({
+    clientId: options.clientId,
+    operationIds: submissionResult.submission.operations.map(({operation}) => operation.id),
+  })
+
+  return {
+    status: 'success',
```
gonzaloriestra, line 48: When `--watch` finishes, should this return a failure if any operation has a `FAILED` status?

gonzaloriestra (commented):
Could you share tophatting instructions? A demo would be useful as well.  
  
Also, the PR is too large, next time please split in a stack :pray:

packages/app/src/cli/commands/app/subscription-migrations/schedule.ts
tyler-eon: All four commands now extend AppLinkedCommand, use canonical appFlags, resolve through linkedAppContext(), pass remoteApp.apiKey to the migration services, and return {app}.

I am wondering about the automation flow here. I don't think it's an issue, but this does mean there's now a hard dependency on having *some* app context present whereas previously you could get away without one if you're on like a CI env. But that doesn't seem like a big deal.

packages/app/src/cli/services/subscription-migrations/run-submission-command.ts
tyler-eon: After `--watch` completes, any terminal operation with status `FAILED` now produces a typed operation failure containing the final submission and failed operation IDs and sets the status code to a non-zero value.

packages/app/src/cli/commands/app/subscription-migrations/schedule.ts
gonzaloriestra: Yeah, that's a small tradeoff, but it's more consistent now. Thanks!

packages/app/src/cli/commands/app/subscription-migrations/result-presenter.ts
```diff
@@ -0,0 +1,140 @@
...
+        ...result.submission.operations.map(({operation}) => formatMigrationOperationStatus(operation)),
+        'Save the root idempotency key and every operation ID. You will need them to inspect or retry this submission.',
+      ],
+    })
+    return
+  }
+
+  const operationIds = result.submission.operations.map(({operation}) => operation.id)
+  renderWarning({
+    headline: 'Some subscription migration operations were accepted before submission failed.',
```
gonzaloriestra: I'm getting this message even when the app is not found and 0 operations were accepted:

Image

I think it should fail in that case.

gonzaloriestra: Also, I double-checked that the app is right, so `App is not found` doesn't look correct either. It may be a permission issue.

gonzaloriestra: The first issue is solved now, thanks!

image

The second one was a permission issue and apparently that message is expected.

gonzaloriestra (approved):
👏

PR #8315, opened 2026-08-12
Prototype JSON event streaming for app dev

gonzaloriestra:
## What

Add a prototype `--json` mode to `shopify app dev` that emits newline-delimited JSON (NDJSON) events for:

- dev session startup metadata
- structured dev session status changes
- stdout and stderr chunks from each concurrent dev process

The existing text and interactive UI paths remain unchanged.

## Why

`app dev` never produces a single final result, so it cannot return one conventional JSON document. This prototype uses one complete JSON object per line, allowing consumers to process events as they arrive.

It is stacked on #8241 and uses its synchronous diagnostic channel to keep event production separate from the JSON output adapter.

## Prototype limitations

- Setup and prompting before the dev processes launch still use the existing output paths.
- Process logs are wrapped as chunks; they are not yet domain-specific structured events.
- Errors still use the existing command error handling rather than a universal JSON error envelope.

## How to test

`pnpm shopify app dev --json`

Output example:

```
{"type":"started","level":"info","message":"Dev session started","preview_url":"https://admin.shopify.com/store/gonzaloriestra-dev/apps/2f4820d3b2edbe40905e302a191e1f03?dev-console=show","graphiql_url":"http://localhost:59562/graphiql?key=d86ce21c80cfb56a0ac7029a29978cae0a790c055d01f10570e2e3783d6759be"}
{"type":"log","level":"info","message":"Running pre-dev command: \"pnpm exec prisma generate\"","source":"React Router","stream":"stdout"}
{"type":"status","level":"info","message":"Ready, watching for changes in your app","status":"success","is_ready":true,"preview_url":"https://admin.shopify.com/store/gonzaloriestra-dev/apps/2f4820d3b2edbe40905e302a191e1f03?dev-console=show","graphiql_url":"http://localhost:59562/graphiql?key=d86ce21c80cfb56a0ac7029a29978cae0a790c055d01f10570e2e3783d6759be"}
```

PR #8241, opened 2026-08-03
Prototype store info diagnostic event sink

dmerand:
## What

Add a private, dependency-free `@shopify/diagnostics` leaf and a typed synchronous diagnostic side channel to `store:info`. The CLI adapter connects those events to the existing `debugMessageSink`, so current output and result behavior remain unchanged.

## Why

Shopify CLI is moving toward data-only finite execution contracts that can serve CLI, SDK, MCP, and other adapters without depending on terminal rendering. This proof of concept demonstrates the compatibility seam between renderer-first execution and future explicit result/event adapters without claiming a universal event framework.

## How

- Keep diagnostics data-only and synchronous; observers run in registration order and observer failures do not affect command results.
- Keep domain events owned by `store:info`, with typed variants and safe error projections.
- Preserve the existing JSON wire shape, stdout/stderr separation, debug filtering, timestamps, color handling, test log collection, fallback/result/error behavior, and non-interactive behavior through the existing CLI output adapter.

Async delivery, stream/progress protocols, universal result/error/event envelopes, launcher or `BaseCommand` changes, infrastructure logging, and broader command migration are out of scope.

## Verification and limitations

- 55 focused tests pass.
- `@shopify/diagnostics` build, type-check, and lint pass.
- cli-kit type-check, changed-file ESLint, and `git diff --check` pass.
- Full `store` type-check remains blocked by the pre-existing missing `@shopify/organizations` declarations; no errors were reported in the changed diagnostics or store-info code.

Post-release steps: None.

PR #8403, opened 2026-08-26
Add command side event infrastructure

gonzaloriestra:

Closes shop/issues-develop#23661

Finite commands need structured diagnostics and progress that remain separate from their final result.

Based on prototypes:
- #8241
- #8315

Architecture: https://arch-diagrams.quick.shopify.io/d/ea355799-a9ad-47d6-b315-3eae713c6434

- Adds typed diagnostic and progress events with timestamps.
- Adds an AsyncLocalStorage-backed event context for each command execution.
- Wraps the full CLI lifecycle so prerun hooks, command services, and postrun hooks share the same context, with BaseCommand covering direct command runs.
- Automatically converts standard info, warning, debug, success, and completion output into diagnostic events in JSON mode.
- Makes renderSingleTask emit start, status, and completion progress events automatically.
- Keeps outputResult on stdout and suppresses standalone newlines in JSON mode.
- Keeps custom loggers, text output, and concurrent command executions isolated.

### How to test your changes?

See https://github.com/Shopify/cli/pull/8415

Also: `pnpm shopify app info --json --verbose`

### Checklist

- [x] This internal infrastructure change does not require a changeset.


gonzaloriestra:
* **#8413** <a href="https://app.graphite.com/github/pr/Shopify/cli/8413?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#8405** <a href="https://app.graphite.com/github/pr/Shopify/cli/8405?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#8403** <a href="https://app.graphite.com/github/pr/Shopify/cli/8403?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Shopify/cli/8403?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/cli-kit/src/public/common/command-events.ts
```diff
@@ -0,0 +1,93 @@
...
+  .strict()
+
+/** Schema for a progress update emitted while a command executes. */
+export const commandProgressEventSchema = z
+  .object({
+    type: z.literal('progress'),
+    timestamp: z.string().datetime({offset: true}),
+    message: z.string(),
+    current: z.number().nonnegative().optional(),
+    total: z.number().nonnegative().optional(),
```
dmerand, line 23: We originally specified having a `status` and `operation` field here, did that get omitted intentionally?

dmerand: LLM explanation:

`renderSingleTask()` shows why it matters:

 ```text
   {message: "Creating store"}                     // start
   {message: "Saving session"}                     // update
   {message: "Saving session", current: 1, total: 1} // completion
 ```

Without operation, a consumer cannot reliably tell that "Creating store" and "Saving session" describe the same operation.

packages/cli-kit/src/public/common/command-events.ts
```diff
@@ -0,0 +1,93 @@
+import {z} from 'zod'
+
+/** Schema for a diagnostic emitted while a command executes. */
+export const commandDiagnosticEventSchema = z
+  .object({
+    type: z.literal('diagnostic'),
+    timestamp: z.string().datetime({offset: true}),
+    level: z.enum(['debug', 'info', 'warning']),
```
dmerand: The design also includes an `error` diagnostic level. Was that intentionally removed because all errors should be fatal? Do we still need it here for error diagnostics?

packages/cli-kit/src/public/common/command-events.ts
```diff
@@ -0,0 +1,93 @@
...
+ */
+export function createCommandEventChannel<TEvent extends CommandEvent = CommandEvent>(
+  options: CommandEventChannelOptions<TEvent> = {},
+): CommandEventChannel<TEvent> {
+  const sink = options.sink ?? (() => {})
+  const clock = options.clock ?? (() => new Date())
+
+  return {
+    emit(event, emissionOptions) {
+      const timestampedEvent = {...event, timestamp: clock().toISOString()} as TEvent
```
dmerand: Is the lack of runtime validation intentional? The channel casts the timestamped value to `TEvent`, and the JSON sink serializes it without parsing `commandEventSchema`. Should the JSON adapter validate an event before writing it?

packages/cli-kit/src/public/node/ui.tsx
```diff
@@ -529,11 +530,30 @@ export async function renderSingleTask<T>({
...
+    })
+    emitCommandEvent(
+      {type: 'progress', message: unstyled(currentStatus.value), current: 1, total: 1},
+      {alreadyRendered: true},
+    )
+    return result
+  }
+
+  if (commandEventOutputMode() === 'json') {
+    return taskWithProgressEvents(() => {})
```
dmerand: The LLM noticed this: The JSON branch skips SingleTask, which also handles Ctrl+C and calls `onAbort`. How can we preserve that callback in JSON mode?

dmerand (commented):
This is a really clever approach. I have a few questions to iron out before we ship.

packages/cli-kit/src/public/common/command-events.ts
gonzaloriestra: Good catch! I missed that. Updated.

packages/cli-kit/src/public/common/command-events.ts
gonzaloriestra: Hmm you are right, I found examples where we print an error without aborting. Added.

packages/cli-kit/src/public/common/command-events.ts
gonzaloriestra: Good point. I'm now parsing it before and added tests.

packages/cli-kit/src/public/node/ui.tsx
gonzaloriestra: Fixed by registering a one-time SIGINT listener when a JSON task provides onAbort, and removing it in finally. Without a callback, default SIGINT handling remains. Added tests.

packages/cli-kit/src/public/node/command-events.test.ts
```diff
@@ -0,0 +1,253 @@
...
+import type {CommandEvent} from '../common/command-events.js'
+
+const outputMock = mockAndCaptureOutput()
+
+beforeEach(() => {
+  outputMock.clear()
+})
+
+describe('command event context', () => {
+  test('shares the context across separately loaded cli-kit module instances', async () => {
```
amcaplan: Nit: Isn't this basically testing `command-event-context` behavior? If so, doesn't it belong in a `command-event-context.test.ts` file?

packages/cli-kit/src/public/node/command-events.ts
```diff
@@ -0,0 +1,115 @@
...
+}
+
+/**
+ * Runs a command execution with an event channel available to all nested asynchronous work.
+ *
+ * @param options - The event sink, clock, and output mode used by the channel.
+ * @param execute - The command execution to run with the channel.
+ * @returns The result of the command execution.
+ */
+export function runWithCommandEvents<TResult>(options: RunWithCommandEventsOptions, execute: () => TResult): TResult {
```
amcaplan: This seems like just a passthrough, what do we gain through an explicit call instead of just re-exporting?

packages/cli-kit/src/private/node/command-event-context.ts
```diff
@@ -0,0 +1,70 @@
...
+  channel: CommandEventChannel
+  outputMode: CommandEventOutputMode
+}
+
+interface RunWithCommandEventsOptions extends CommandEventChannelOptions<CommandEvent> {
+  outputMode?: CommandEventOutputMode
+}
+
+const commandEventStorageKey = Symbol.for('@shopify/cli-kit/command-event-storage')
+const existingCommandEventStorage = Reflect.get(globalThis, commandEventStorageKey) as
```
amcaplan: This pattern is unusual for our codebase. We have other patterns for lazy-creating globals; usually we do it in a function so loading a file has no side effects; this also has the benefit of keeping types simple because the function can more easily guarantee a return type.

gonzaloriestra: Good catch, fixed!

packages/cli-kit/src/public/node/command-events.test.ts
gonzaloriestra: Moved

packages/cli-kit/src/public/node/command-events.ts
gonzaloriestra: Agreed, re-exporting now

dmerand (approved):
Thanks for the changes!

PR #8404, opened 2026-08-26
Add JSON result schema infrastructure

gonzaloriestra:
### WHY are these changes introduced?

Closes shop/issues-develop#23660

Finite commands need one typed contract for validating JSON results and documenting them in command help.

Based on these prototypes:

- #8283 
- #8242
- #8243
- #8288

### WHAT is this pull request doing?

- Adds typed JSON result schemas with validation and encoding.
- Exposes schemas through `BaseCommand` and generated command help.
- Rejects unsupported or unnamed nested schema constructs.

### How to test your changes?

See https://github.com/Shopify/cli/pull/8415

### Checklist

- [x] I've considered possible cross-platform impacts (Mac, Linux, Windows)
- [x] I've considered possible [documentation](https://shopify.dev) changes
- [x] I've considered analytics changes to measure impact
- [ ] The change is user-facing — I've identified the correct [bump type](https://github.com/Shopify/cli/blob/main/CONTRIBUTING.md#choosing-the-right-bump-type) and added a changeset

PR #8405, opened 2026-08-26
Emit fatal errors as JSON

gonzaloriestra:

Closes shop/issues-develop#23662

Commands using JSON output need one machine-readable fatal error on stdout.

Based on [#8182](https://github.com/Shopify/cli/pull/8182).

- Emits JSON fatal errors with the same meaningful content as regular errors.
- Keeps diagnostics and progress off stdout.
- Handles uncaught errors consistently and preserves nonzero exit codes.
- Keeps silent aborts silent and ignores `--json` after the `--` boundary.

### How to test your changes?

- `pnpm shopify app info --wrong-flag`
- `pnpm shopify app info --wrong-flag --json`

Add `if (1 === 1) throw new Error('boom')` at the beginning of commands/app/info.ts and run:
- `pnpm shopify app info`
- `pnpm shopify app info --json`

### Checklist

- [x] The change is user-facing — I've identified the correct [bump type](https://github.com/Shopify/cli/blob/main/CONTRIBUTING.md#choosing-the-right-bump-type) and added a changeset


packages/cli-kit/src/public/node/error/index.ts
```diff
@@ -0,0 +1,296 @@
...
+      const {renderFatalErrorAsJson} = await import('../../../private/node/json-error.js')
+      renderFatalErrorAsJson(fatal)
+      return Promise.resolve(error)
+      // eslint-disable-next-line no-catch-all/no-catch-all
+    } catch (serializationError) {
+      outputDebug(`Failed to render the error as JSON: ${serializationError}`)
+    }
+  }
+
+  const {renderFatalError} = await import('../ui.js')
```
gonzaloriestra, line 157: The content of this file was moved from ../error.ts. These are the only changes here.

gonzaloriestra:
* **#8413** <a href="https://app.graphite.com/github/pr/Shopify/cli/8413?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#8405** <a href="https://app.graphite.com/github/pr/Shopify/cli/8405?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Shopify/cli/8405?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#8403** <a href="https://app.graphite.com/github/pr/Shopify/cli/8403?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

tizmagik:
This closes the empty-stdout problem, which is what actually blocks scripting against `--json`. Good to see it moving. One follow-up while the shape is still open.

For `store execute`, the interesting part of a failure is the GraphQL `errors` array, and after this change it arrives as a serialized JSON string inside `tryMessage`. That comes from [the throw site in `admin-transport.ts`](https://github.com/Shopify/cli/blob/3bb4bc7b5076ea9303ee63c0248496e98dadb369/packages/store/src/cli/services/store/execute/admin-transport.ts#L94):

```ts
throw new AbortError('GraphQL operation failed.', JSON.stringify({errors: error.response.errors}, null, 2))
```

`json-error.ts` passes a plain-string `tryMessage` through verbatim, so the data survives intact. A consumer can `JSON.parse` the document, then `JSON.parse` the `tryMessage`, and recover the error codes exactly. That works, and it beats parsing the rendered banner.

What I would push back on is that it works by accident. `tryMessage` is typed `zod.string()` and named for display. The only reason it carries structured data is that this one call site serializes into it. The token handling in `json-error.ts` invites turning a message like that into token items or a `customSections` table, and the moment someone does, every consumer's second `JSON.parse` fails. No test would catch that. On `main`, [the only test asserting on `tryMessage` in that file](https://github.com/Shopify/cli/blob/3bb4bc7b5076ea9303ee63c0248496e98dadb369/packages/store/src/cli/services/store/execute/admin-transport.test.ts#L309) covers the 402 case and checks for a human-readable phrase, so the GraphQL branch's payload is unpinned.

Is there room in this design for structured error detail that is not a display string? Two shapes would do it.

1. An optional `details` field on `JsonError`, carrying unknown-shaped data from the raising site. `admin-transport.ts` then passes the `errors` array as data and keeps `tryMessage` for humans. This is the bigger change, since it touches the schema the rest of your stack validates against.
2. Leave `JsonError` alone and let `store execute` handle its own failure, returning the GraphQL error payload through `writeOrOutputStoreExecuteResult`, which already serializes cleanly in `json` mode.

I have a working branch for the second one, scoped to `store execute`. It adds an `AbortError` subclass carrying `readonly response`, and the command emits `error.response` as JSON when `--json` is set. Identical banner for humans, unchanged success shape, tests included. Happy to open it against your branch, hand it over, or drop it if the first shape is where you would rather go. Your call, this is your stack.

Separate and smaller, probably its own issue rather than folded in here. `store execute` drops `extensions` on both the success and failure paths, so a client cannot read `cost.throttleStatus` and honor the leaky bucket. It passes `handleErrors: false` with no `onResponse`, and `autoRateLimitRestore` defaults to false.

packages/cli-kit/src/public/node/error/schema.ts
dmerand, line 1: Building on [this comment](https://github.com/Shopify/cli/pull/8405#issuecomment-5482690680): The renderer already limits output to known fields, which we should keep. But there is no field for extra structured error data.

Could a command pass selected error data to the renderer in an optional `details` field? The command would choose what to include. The renderer would continue to ignore other properties on the original error.

This would let callers read the data without parsing display strings such as `tryMessage`. I think it could fit into the existing error-output path without another error-handling layer.

packages/cli-kit/src/public/node/path.test.ts
```diff
@@ -131,6 +131,14 @@ describe('sniffForJson', () => {
   test('returns false if neither is present', () => {
     expect(sniffForJson(['node', 'script.js', '--other-flag'])).toBe(false)
   })
+
+  test.each(['--json', '-j'])('returns false if %s is a passthrough argument', (jsonFlag) => {
```
dmerand, line 135: Ugh passthrough arguments are one of my least favorite things about node-based CLIs. This is such confusing syntax, and easy to forget!

packages/cli-kit/src/public/node/error/index.ts
```diff
@@ -0,0 +1,296 @@
...
+    if (maybeError?.stack) {
+      fatal.stack = maybeError?.stack
+    }
+  }
+
+  const {jsonOutputEnabled} = await import('../environment.js')
+  if (jsonOutputEnabled()) {
+    try {
+      const {renderFatalErrorAsJson} = await import('../../../private/node/json-error.js')
+      renderFatalErrorAsJson(fatal)
```
dmerand: Could we flush stdout on this path before Oclif exits, and cover the normal exit path with a subprocess test? The uncaught-error path flushes stdout, but this path returns to `Errors.handle()`, which calls `process.exit()`.

In an isolated subprocess check using this PR's error handler, serializer, stdout writer, and Oclif 4.8.3, a 1 MiB error stopped at 65,536 bytes in three runs. `JSON.parse` failed each time. Adding the PR's `flushStdout()` before `Errors.handle()` delivered the full document and kept exit code 1. Unrelated environment/UI dependencies were stubbed; this was not a full CLI invocation.

The existing fixtures miss this case: one sets `process.exitCode`, which permits normal draining; the large-output test explicitly calls `flushStdout()` rather than using the normal error path.

packages/cli-kit/src/public/node/error/index.ts
```diff
@@ -0,0 +1,296 @@
...
+      return Promise.resolve(error)
+      // eslint-disable-next-line no-catch-all/no-catch-all
+    } catch (serializationError) {
```
dmerand: If JSON serialization fails, could we emit a minimal JSON error instead of falling back to a text banner? When stdout is still writable, callers using `--json` should still receive a parseable error document. A broken stdout pipe is a separate case.

Could we add a test for this serialization-failure fallback?

packages/cli-kit/src/public/node/error/types.ts
```diff
@@ -0,0 +1,34 @@
+export type JsonErrorType = 'abort' | 'bug' | 'external'
+
+export interface JsonErrorCustomSection {
+  title?: string
+  body: string | string[][]
+}
+
+interface JsonErrorBase {
```
dmerand: Can we infer these error types from the Zod schemas instead of maintaining separate interfaces?

The `satisfies ZodType<...>` checks do not require the two declarations to match exactly. Optional fields can still differ. Inference would keep the TypeScript type, runtime validation, and help output tied to one declaration, as intended in #8404.

dmerand (commented):
Mix of human/LLM feedback on this one. Again, off to a great start.

packages/cli-kit/src/public/node/path.test.ts
gonzaloriestra: Yeah, it's horrible!

packages/cli-kit/src/public/node/error/index.ts
gonzaloriestra: Fixed by flushing in the shared handler. Also added a test to check that a full 1 MiB document survives `Errors.handle()` with exit code 1.

packages/cli-kit/src/public/node/error/index.ts
gonzaloriestra: Good idea, I added the fallback.

packages/cli-kit/src/public/node/error/schema.ts
gonzaloriestra: Totally agree, I added the optional details field and it's used by `store execute` now.

packages/cli-kit/src/public/node/error/types.ts
gonzaloriestra: Done, the schemas are now the single source of truth

gonzaloriestra:
@tizmagik nice point! I added the optional details field, and it's used by `store execute` now.

Regarding the second issue, you are right. The new details field preserves per-error extensions, but response-level extensions.cost.throttleStatus is still dropped. I agree this deserves a separate fix covering both success and failure. I can have a look later.

dmerand (approved):

PR #8406, opened 2026-08-26
Require JSON support for new commands

gonzaloriestra:
### WHY are these changes introduced?

Closes shop/issues-develop#23663

New finite commands should include typed JSON output by default.

### WHAT is this pull request doing?

- Requires new finite commands to declare a JSON schema and `jsonFlag`.
- Enforces this through ESLint while preserving an exact legacy baseline.
- Allows an explicit exemption for streaming commands.

### How to test your changes?

`pnpm lint`

If you add a new command or remove it from the exemption list, it should show something like this:

```
~/src/github.com/Shopify/cli/packages/app/src/cli/commands/app/build.ts
  9:16  error  New finite commands must declare a static jsonOutputSchema. See docs/cli/json-output.md          @shopify/cli/command-json-output
  9:16  error  New finite commands must include ...jsonFlag in their static flags. See docs/cli/json-output.md  @shopify/cli/command-json-output

✖ 2 problems (2 errors, 0 warnings)
```

### Checklist

- [x] I've considered possible cross-platform impacts (Mac, Linux, Windows)
- [x] I've considered possible [documentation](https://shopify.dev) changes
- [x] I've considered analytics changes to measure impact
- [ ] The change is user-facing — I've identified the correct [bump type](https://github.com/Shopify/cli/blob/main/CONTRIBUTING.md#choosing-the-right-bump-type) and added a changeset

PR #8413, opened 2026-08-27
Add JSON schema flag to commands

gonzaloriestra:

Part of shop/issues-develop#23660.

Command help can contain long JSON output schemas, that can be crop by agents. Agents and developers need a way to request the schemas without running the command. Also, the current help command only includes the result schema, but the events/error schemas are useful as well.

- Adds `--json-schema` to every command.
- Prints the declared result schema and the shared error and event schemas, then exits before normal command work.
- Shows an error when a command has no JSON result schema.
- Explains the flag immediately before inline schemas in command help.
- Regenerates manifests, README content, and Shopify.dev command data.

### How to test your changes?

See https://github.com/Shopify/cli/pull/8415

### Checklist

- [x] I have considered possible cross-platform impacts (Mac, Linux, Windows)
- [x] I have considered possible documentation changes
- [x] I have considered analytics changes to measure impact
- [x] The change is user-facing and includes a minor changeset


gonzaloriestra:
* **#8413** <a href="https://app.graphite.com/github/pr/Shopify/cli/8413?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Shopify/cli/8413?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#8405** <a href="https://app.graphite.com/github/pr/Shopify/cli/8405?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#8403** <a href="https://app.graphite.com/github/pr/Shopify/cli/8403?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

packages/cli-kit/src/public/node/base-command.ts
```diff
@@ -124,6 +132,21 @@ abstract class BaseCommand extends Command {
     }
   }
 
+  protected exitWithJsonSchemaWhenRequested(): void {
+    if (!this.argv.includes('--json-schema') && !isTruthy(process.env.SHOPIFY_FLAG_JSON_SCHEMA)) return
```
dmerand: Could we check only arguments before `--`, as we now do for `--json`? This check also matches `--json-schema` after the passthrough boundary.

packages/cli-kit/src/public/node/base-command.ts
```diff
@@ -124,6 +132,21 @@ abstract class BaseCommand extends Command {
...
+    const command = this.constructor as typeof BaseCommand
+    const outputSchema = command.jsonOutputSchema
+    if (!outputSchema) {
+      throw new AbortError('This command does not define a JSON output schema.')
+    }
+
+    outputResult(
+      [outputSchema.typescript, jsonErrorOutputSchema.typescript, commandEventOutputSchema.typescript].join('\n\n'),
+    )
+    process.exit(0)
```
dmerand, line 161: Downstack, we've flushed STDOUT prior to exiting in a similar scenario; do we need to do that here as well?

packages/cli-kit/src/public/node/base-command.ts
gonzaloriestra: Good catch, fixed!

packages/cli-kit/src/public/node/base-command.ts
gonzaloriestra: Yes, thanks. Updated.

packages/cli-kit/package.json
```diff
@@ -153,7 +153,8 @@
     "strip-ansi": "7.2.0",
     "supports-hyperlinks": "3.2.0",
     "which": "4.0.0",
-    "zod": "3.25.76"
+    "zod": "3.25.76",
+    "zod-to-json-schema": "3.25.2"
```
gonzaloriestra, line 157: Zod 4 includes a `toJSONSchema()`, so once we upgrade it, we can get rid of this dependency.

I've created an issue: https://github.com/shop/issues-develop/issues/23880

dmerand (approved):

packages/cli-kit/src/public/node/base-command.ts
```diff
@@ -81,6 +87,7 @@ abstract class BaseCommand extends Command {
   }
 
   protected async init(): Promise<unknown> {
+    await this.exitWithJsonSchemaWhenRequested()
```
fredericoo: non-blocking: this runs after oclif `init` hooks, so plugin hooks get a say before the schema is printed. Two things I hit with hydrogen:

- `shopify hydrogen build --json-schema` outside a Hydrogen project prints hydrogen's "Looks like you're trying to run a Hydrogen command outside of a Hydrogen project" warning and exits 1, never reaching this
- `--json-schema --verbose` shows the `notifications list` child process being spawned before the abort.

IIRC `--help` and `--version` are handled before hooks run. Would it make sense to treat `--json-schema` at the same layer? Happy to leave it if that's a bigger change than it's worth for now (`hydrogen build` is not on the green path anyway), but then it might be worth a line in the docs that plugin hooks run first.

packages/cli-kit/src/public/node/base-command.ts
```diff
@@ -33,7 +35,13 @@ interface EnvironmentFlags {
 }
 
 abstract class BaseCommand extends Command {
-  static baseFlags: FlagInput<{}> = {}
+  static baseFlags: FlagInput<{}> = {
+    'json-schema': Flags.boolean({
+      description: "Print the command's JSON schemas.",
```
fredericoo: nit: no production command defines a `jsonOutputSchema` in this stack yet, so this flag shows up in every command's help (all 26 hydrogen commands included) but errors on all of them for now. Fine as infra-first, but the description reads as a promise. Let's qualify it a bit?

```suggestion
      description: "Print the command's JSON schemas, when the command supports `--json`.",
```

docs/cli/json-output.md
```diff
@@ -115,5 +115,9 @@ Tests should verify:
 - errors and exit behavior; and
 - prompt behavior independently from `--json` and `--no-input`.
 
-Command help includes the generated TypeScript contract automatically through `jsonOutputSchema`. Run the manifest,
+Command help includes the result's JSON Schema automatically through `jsonOutputSchema`. `--json-schema` prints one
+JSON Schema (draft-07) accepting a result, a fatal error document, or a side event. The `Result`, `Error`, and `Event`
+definitions describe these separately; results and fatal errors go to stdout, and side events go to stderr.
```
fredericoo, line 136: non-blocking: a few gaps that surfaced testing through hydrogen (a plugin that hasn't adopted any of this yet). Worth a short "plugin authors" note in here, or at least confirming which are intended:

- Errors raised from oclif `init` hooks bypass the JSON fatal error path entirely. Hydrogen's hook does `renderWarning` + `process.exit(1)`, so `SHOPIFY_FLAG_JSON=1 shopify hydrogen build --path /nope` gives an empty stdout and exit 1, no error document.
- `render*` banners aren't converted to events in JSON mode while `output*` helpers are. `SHOPIFY_FLAG_JSON=1 shopify hydrogen check routes` prints the raw success banner to stderr, whereas `hydrogen build`'s `outputInfo` became a `diagnostic` event. Is that a deliberate split or a TODO?
- Third-party loggers land straight on stdout. `SHOPIFY_FLAG_JSON=1 shopify hydrogen build` gives 112 non-JSON lines on stdout (Vite + React Router) and 0 JSON. That's on us to silence via `jsonOutputEnabled()`, but the doc could say so explicitly since "results go to stdout" only holds if the command owns everything it writes there.

None of these are blockers for this PR, they're the work hydrogen would need to do to actually be usable with `--json`.

packages/cli/oclif.manifest.json
```diff
@@ -5632,6 +6066,13 @@
           "name": "inspector-port",
           "type": "option"
         },
+        "json-schema": {
```
fredericoo, line 6069: non-blocking: heads up for when the lint exceptions list grows to cover plugins: `hydrogen dev`, `hydrogen preview` and `hydrogen debug cpu` are dev-server wrappers, so we'd want them in the streaming section rather than offering a `--json` mode. 

those commands are being deprecated in the next Hydrogen anyway. The finite ones we'd actually want schemas for are `list`, `env list`, `link`. Nothing to change here, just flagging so it doesn't surprise anyone.

fredericoo (approved):
Tested this from the Hydrogen side against a fresh `hydrogen init` project (mock-shop, hydrogen 13.0.4 as pinned here): `--help`, `init`, `build`, `check routes`, `g route`, `dev`, `list`, `env list` all still behave. `CliKitDedupPlugin` means every `hydrogen:*` command inherits `--json-schema` and the JSON error path for free, which is nice.

No regressions for us, so from the hydrogen perspective this LGTM. A few non-blocking notes inline from poking at the JSON paths through a plugin that hasn't adopted schemas yet.

packages/cli-kit/src/public/node/base-command.ts
gonzaloriestra: Good catch! Moved schema inspection before Oclif’s run(), so it skips init and prerun hooks. Hydrogen now reaches the missing-schema error outside a project, without the project warning or notification subprocess.

docs/cli/json-output.md
gonzaloriestra: Added a plugin authors section covering those points, thanks!

packages/cli/oclif.manifest.json
gonzaloriestra: Thanks for the heads up. We had already tagged those as streaming: https://docs.google.com/spreadsheets/d/1j3jr9sLbWwpsW50yJB-5wlgNofRMOSlCalOf8HKZTeM/edit?gid=0#gid=0

PR #8425, opened 2026-08-28
Support single-subscription events modules when reading remote configuration

rezaansyed:

Shopify Core is adding a single-subscription shape for `events` app modules: the `:events` module config will accept either the existing subscription list (`subscription: [...]`, one module with uid `"events"`) or a new single-subscription object (`subscription: {...}`, one module per subscription, uid = subscription handle). This enables shop-scoped (imperative) event subscriptions and a migration from one list module to N single-subscription modules.

The CLI's events transforms are currently hard-typed to the list shape: `transformToEventsConfig` calls `.map` on `subscription` and crashes on an object, and `config link` would need N single-subscription modules to merge back into one `[[events.subscription]]` list in the local TOML.

- `transformToEventsConfig` (remote → local): accepts a single-subscription object **or** the legacy array. Single objects are normalized to a one-element array (after stripping the server-managed `identifier`), so merging N single-subscription modules via `deepMergeObjects` implodes them into one subscription list in the TOML.
- `transformFromEventsConfig` (local → remote): resolves relative subscription URIs for both shapes, preserving the input shape.

The local TOML format is unchanged: `[[events.subscription]]` stays a list. This is read-side tolerance only.

### How to test your changes?

- `pnpm vitest run src/cli/models/extensions/specifications/transform/app_config_events.test.ts` (14 tests, 5 new covering single-object stripping, multi-module merge, mixed-shape merge, and relative-URI resolution for the object shape).

### Post-release steps

None.

### Measuring impact

- [x] n/a: covered by existing events module metrics in Core

### Checklist

### DIFF
--- a/packages/app/src/cli/models/extensions/specifications/transform/app_config_events.ts
+++ b/packages/app/src/cli/models/extensions/specifications/transform/app_config_events.ts
@@ -2,10 +2,15 @@ import {prependApplicationUrl} from '../validation/url_prepender.js'
 import {CurrentAppConfiguration} from '../../../app/app.js'
 import {getPathValue} from '@shopify/cli-kit/common/object'
 
+interface EventSubscription {
+  uri: string
+  [key: string]: unknown
+}
+
 interface EventsConfig {
   events?: {
     api_version?: string
-    subscription?: {uri: string; [key: string]: unknown}[]
+    subscription?: EventSubscription | EventSubscription[]
   }
 }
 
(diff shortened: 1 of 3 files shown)


packages/app/src/cli/models/extensions/specifications/transform/app_config_events.ts
```diff
@@ -27,35 +32,61 @@ export function transformFromEventsConfig(content: object, appConfiguration?: ob
...
+  // The server always includes identifier, and materializes the events default
+  // api_version onto every subscription. Both are derived, so they are stripped
+  // for the local TOML; an api_version that differs from the default is a real
+  // override and is kept. Single-subscription modules are normalized to a
+  // one-element array so that merging multiple modules accumulates a single
+  // subscription list.
+  const cleanedSubscriptions =
+    subscription === undefined
+      ? undefined
+      : wrapSubscriptions(subscription).map((sub) => {
```
cdarne: It looks like this would throw an error if subscription is `null`. The check to `undefined` above is not enough. Before we were checking `subscription?.map(...)` .

packages/app/src/cli/models/extensions/specifications/transform/app_config_events.ts
```diff
@@ -27,35 +32,61 @@ export function transformFromEventsConfig(content: object, appConfiguration?: ob
     appUrl = (appConfiguration as CurrentAppConfiguration)?.application_url
   }
 
+  const subscription = eventsConfig.events.subscription
+  const resolved = wrapSubscriptions(subscription).map((sub) => ({
+    ...sub,
+    uri: prependApplicationUrl(sub.uri, appUrl),
```
cdarne: Suggestion: we could guard against a `null` => `sub.uri`. I know it's not new logic, but a `null` value is plausible (an error/typo in the toml?) and it'd blow in `prependApplicationUrl` instead of giving a nice error. Something like this might work `sub.uri ? { ...sub, uri: prependApplicationUrl(sub.uri, appUrl) } : sub` (no the prettiest, I'm not a JS/TS expert)

PR #8426, opened 2026-08-28
Fan out one events module per subscription behind an organization flag or environment opt-in

rezaansyed:

Stacked on #8425.

Shopify Core is migrating `events` app modules from one list-shaped module (`subscription: [...]`, uid `"events"`) to one single-subscription module per subscription (`subscription: {...}`, uid = subscription handle). The Core contract, uid derivation, and validation already support both shapes; this PR makes the CLI deploy the new shape.

- **Loader fan-out** (`loader.ts`): when enabled, each `[[events.subscription]]` entry in the TOML becomes its own `events` module instance with `{events: {api_version, subscription: <object>}}` — the single shape Core recognizes. The fan-out runs inside `createConfigExtensionInstances` so the `events` key is still claimed (no "Unsupported section(s)" false positive). The tuple shape changed from `[instance, keys]` to `[instances[], keys]`.
- **Identity** (`extension-instance.ts`): for a single-shape events config, `buildHandle`/`buildUIDFromStrategy` return the subscription handle — shape-driven and strategy-agnostic, mirroring Core's uid derivation (uid = subscription handle). Duplicate subscription handles are rejected by the loader's existing global duplicate-handle check.
- **Gating**: fan-out is enabled by either
  - the org-scoped exp flag `f_single_subscription_events_modules_cli`, fetched in `appFromIdentifiers` via the existing `organizationExpFlags` Business Platform query and surfaced through the previously-empty `Flag` enum / `remoteFlags` pipe (`Flag.SingleSubscriptionEventsModules`), best-effort with a catch-all so a failed lookup never breaks a command; or
  - the `SHOPIFY_CLI_EVENTS_SUBSCRIPTION_FANOUT` env var as a local opt-in/escape hatch.

Core's per-app Verdict flag remains the authoritative server-side gate: a fanned-out deploy for an app that isn't allowed the single shape fails loudly at Core validation.

### How to test your changes?

- `pnpm vitest run src/cli/models/app/loader.test.ts src/cli/utilities/developer-platform-client/app-management-client.test.ts src/cli/models/extensions/extension-instance.test.ts` (new tests: fan-out shapes/handles/uids via env var, fan-out via remote flag, disabled path unchanged, duplicate-handle rejection, exp-flag enabled/disabled/failure in `appFromIdentifiers`).
- Manually: set `SHOPIFY_CLI_EVENTS_SUBSCRIPTION_FANOUT=1`, add two `[[events.subscription]]` entries with handles, run `app deploy` against an app allowlisted in Core — the version should contain two `events` modules whose uids are the subscription handles.

### Post-release steps

- Create the org-scoped exp flag `f_single_subscription_events_modules_cli` before ramping (until then the env var is the only way to enable fan-out).

### Measuring impact

- [x] n/a: measured server-side via Core's `next_gen_events.list_shape_module_validated` metric decline

### Checklist

### DIFF
--- a/.changeset/events-subscription-fanout.md
+++ b/.changeset/events-subscription-fanout.md
@@ -1,0 +1,5 @@
+---
+'@shopify/app': patch
+---
+
+Fan out one events module per subscription, gated by an organization flag or SHOPIFY_CLI_EVENTS_SUBSCRIPTION_FANOUT
--- a/packages/app/src/cli/models/app/loader.test.ts
+++ b/packages/app/src/cli/models/app/loader.test.ts
@@ -2174,6 +2174,217 @@ describe('load', () => {
     ])
   })
 
+  test('fans out one events module per subscription when the fan-out is enabled', async () => {
+    // Given
+    const appConfigurationWithEvents = `
+    name = "for-testing-events"
+    client_id = "1234567890"
+    application_url = "https://example.com/lala"
+    embedded = true
+
+    [build]
+    include_config_on_deploy = true
+
+    [webhooks]
+    api_version = "2024-01"
+
+    [auth]
+    redirect_urls = [ "https://example.com/api/auth" ]
+
+    [events]
+    api_version = "2024-01"
+
+    [[events.subscription]]
+      topic = "orders/create"
+      actions = ["create"]
+      handle = "order-notifier"
+      uri = "https://example.com/events/orders"
+
+    [[events.subscription]]
+      topic = "products/update"
+      actions = ["update"]
+      handle = "product-sync"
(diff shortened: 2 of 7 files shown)


cdarne (commented):
Overall the logic is sound, the observation I have is that we might want to move some code: the events modules logic in `loader.ts` feels like it should rather be in a file more specific to the events logic (e.g. something like `specifications/app_config_events.ts`).

We could move the logic that splits the events config and the one finding the module handle in a specific events/helper module and have `loader.ts`, `extension-instance.ts`, etc use those helpers and only deal with arrays of subscriptions (keeping the hard logic in the events module).

PR #8427, opened 2026-08-28
Avoid printing sensitive store auth URLs

dengjeffrey:
## Summary

Prevent Shopify CLI from printing manual store authorization URLs that the caller has marked as sensitive.

## Context

The Stripe-backed `store:stripe-auth` flow passes the signup JWT through the OAuth authorization URL as the `signup` query parameter.

The normal path opens that URL in the browser and does not print it. If the browser cannot be opened automatically, the fallback path previously printed the full manual URL to the terminal. That could expose the signup JWT in shell history, logs, screenshots, or CI output.

## Implementation

- Adds a `sensitive` option to the manual auth URL presenter API.
- Marks the manual auth URL as sensitive at the call site when the store auth input includes `signup`.
- Suppresses the manual URL output when `sensitive` is true and shows safer guidance instead.

This keeps the sensitivity decision at the point where we know why the URL is sensitive, rather than having the presenter parse and infer sensitive query parameters after the fact.

## Testing

```sh
pnpm vitest packages/store/src/cli/services/store/auth/result.test.ts packages/store/src/cli/services/store/auth/index.test.ts packages/store/src/cli/commands/store/stripe-auth.test.ts --run
```

### DIFF
--- a/packages/store/src/cli/services/store/auth/result.ts
+++ b/packages/store/src/cli/services/store/auth/result.ts
@@ -19,9 +19,13 @@ export interface StoreAuthResult {
 
 type StoreAuthOutputFormat = 'text' | 'json'
 
+interface ManualAuthUrlOptions {
+  sensitive?: boolean
+}
+
 export interface StoreAuthPresenter {
   openingBrowser: () => void
-  manualAuthUrl: (authorizationUrl: string) => void
+  manualAuthUrl: (authorizationUrl: string, options?: ManualAuthUrlOptions) => void
   success: (result: StoreAuthResult) => void
 }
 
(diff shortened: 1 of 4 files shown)


packages/store/src/cli/services/store/auth/result.ts
```diff
@@ -47,7 +51,16 @@ function displayStoreAuthOpeningBrowser(): void {
   outputInfo('')
 }
 
-function displayStoreAuthManualAuthUrl(authorizationUrl: string): void {
+function displayStoreAuthManualAuthUrl(authorizationUrl: string, options: ManualAuthUrlOptions = {}): void {
+  if (options.sensitive) {
+    outputInfo(
+      'Browser did not open automatically. The manual authorization URL contains sensitive credentials and was not printed.',
```
tizmagik, line 57: If we don't print it, does this affect AI toolkit behavior?

robthornton (approved):

packages/store/src/cli/services/store/auth/result.ts
dengjeffrey: No as the AI tooklit doesn't pass in the url as sensitive when going through the natural Oauth flow without the signup token

tizmagik (approved):

craigmichaelmartin (approved):

craigmichaelmartin:
Opened https://github.com/Shopify/cli/pull/8446 with a follow-up.

PR #8438, opened 2026-08-31
Prevent GitHub credentials from being logged locally

gonzaloriestra:

GitHub helper scripts printed authentication tokens and signed download URLs to command output. This could expose credentials in local terminal history or agent transcripts. CI should be safe because it uses secrets instead of `dev` commands.

- Captures `dev github print-auth` output without echoing it.
- Logs credential sources without logging credential values.
- Stops printing signed LFS download URLs.
- Fails clearly when `dev` cannot provide a token.
- Adds regression tests for environment and `dev` token paths.

### How to test your changes?

`pnpm graphql-codegen:get-graphql-schemas` and check that it doesn't include any token

### Checklist

- [x] I have considered possible cross-platform impacts.
- [x] I have considered possible documentation changes; none are required.
- [x] I have considered analytics changes; none are required.
- [x] This changes internal tooling only; no changeset is required.

### DIFF
--- a/bin/get-graphql-schemas.js
+++ b/bin/get-graphql-schemas.js
@@ -121,7 +121,7 @@ async function fetchFileForSchema(schema, octokit) {
           path: schema.pathToFile,
           ref: branch,
         })
-        console.log(`LFS download via ${download_url}...`)
+        console.log('Downloading LFS schema content...')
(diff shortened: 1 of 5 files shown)


dmerand (approved):
Thank you!

PR #8445, opened 2026-08-31
Review fixes for the store signup JWT handoff

craigmichaelmartin:

Review fixes for #8428, stacked on `signup-stdin-loopback` so its branch keeps its own history. Take these as commits to cherry-pick or squash in, whichever suits.

Rebased onto current `main` now that #8427 has merged. Two changes since the first version of this PR:

- **Dropped the manifest commit.** It now lives on #8428 itself as `e4458fdae5`, regenerated properly with `pnpm refresh-manifests` on current `main` rather than transcribed from a CI log. That was #8428's only red required check.
- **Added `d12057e7a` — restoring the fail-fast abort**, with an explanation below, because I was the one who removed it.

Five commits. Each source change was verified to have failing tests when reverted in isolation.

**1. `Harden the store auth handoff endpoint`** — `callback.ts`

- Every rejection answers `404`. Previously a wrong nonce got `403` and a spent handoff `410`, which tells a local prober that a store auth flow is live on this port and already used.
- Restricted to `GET`.
- **A browser's speculative fetch no longer spends the single-use handoff.** The manual path prints `http://127.0.0.1:13387/auth/handoff?nonce=…` for a human to paste; if the browser prefetches or prerenders it from the omnibox, the nonce is spent and the real navigation finds a dead handoff — an intermittent auth failure in exactly the headless case the handoff exists to serve. Requests carrying `Sec-Purpose`, `Purpose` or `X-moz` with `prefetch`/`prerender` are declined without consuming it. Tested by prefetching, then navigating, and asserting the navigation still gets its 302.

**2. `Keep the signup credential out of buildStoreAuthUrl`** — `pkce.ts`

The finding's literal first step, which the stack skipped: `buildStoreAuthUrl` still took `signup?: string` and still ran `params.set('signup', …)`. Now it neither accepts nor serializes the credential — the URL carrying it is derived only for the loopback redirect. Also removes `signup?: string` from `StoreAuthorizationContext`, which had no readers but handed the raw credential back to every caller of `createPkceBootstrap`.

**3. `Withhold manual auth URLs that carry a signup credential`** — `result.ts`

The refusal was keyed on a caller-supplied boolean. After the handoff no caller passes `true`, so the guard goes dormant and a future caller that forgets leaks the URL. The presenter now also refuses any URL that actually carries a `signup` parameter, and fails closed on one it cannot parse.

**4. `Don't wait on an interactive stdin for the signup JWT`** — `stripe-auth.ts`

- `readSignupJwtFromStdin` read `process.stdin` unconditionally, so with `--signup` optional, `shopify store stripe-auth --store … --scopes …` on a terminal **hung forever** instead of reporting the missing credential. Guarded with cli-kit's existing `isStdinPiped()` (`system.ts:413`).
- Reads capped at 8 KiB.
- `flags.signup ?? …` meant `--signup ''` ran a signup-less authorization; a blank flag is now treated as not supplied. Implemented by normalising the value rather than `||`, since `@typescript-eslint/prefer-nullish-coalescing` is an error here.

**5. `Abort when no authorization URL reached the browser`** — `result.ts`, `index.ts`

This one needs context, because it undoes a mistake of mine.

The fail-fast abort merged with #8427 was keyed on `Boolean(input.signup)`. When I restacked #8428, that condition became permanently false — the handoff URL carries no credential — so the abort was unreachable and I resolved the conflict by dropping it. That was defensible for #8428 alone, but wrong in the round: **commit 3 of this PR reintroduces a withhold path**, via the content check and the fail-closed unparseable branch. With the abort gone, a withheld URL means the CLI waits out the full `timeoutMs = 5 * 60 * 1000` for a callback that cannot arrive.

`Boolean(input.signup)` was only ever a proxy. The real invariant is *don't wait when the browser was given no URL*, so `manualAuthUrl` now returns whether it surfaced a usable one and the caller aborts when it did not. That holds whatever the reason — caller flag, credential in the URL, or unparseable input — and can't drift the way a duplicated predicate would.

This changes `StoreAuthPresenter.manualAuthUrl` from `void` to `boolean`. It's an internal interface with one caller and two implementations, and `options.sensitive` is kept as the extension point for credentials the presenter can't recognise. If you'd rather keep the signature and duplicate the predicate in `index.ts`, say so and I'll flip it.

### How to test your changes?

```sh
pnpm vitest run packages/store/src/cli/services/store/auth packages/store/src/cli/commands/store/stripe-auth.test.ts
```

Full `packages/store`: 445 passed / 60 files. `nx run store:type-check` 0 errors. `eslint`, `prettier --check` and `pnpm knip` clean. Reverting each source file in isolation produces failing tests for that change specifically.

### Post-release steps

None.

### Checklist

- [x] I've considered possible cross-platform impacts (Mac, Linux, Windows) — `isStdinPiped()` is existing cli-kit code; the handoff endpoint is loopback HTTP with no platform-specific behaviour
- [x] I've considered possible [documentation](https://shopify.dev) changes — the manifest regeneration for the flag change is on #8428; the command is hidden so README and shopify.dev docs are unaffected
- [x] I've considered analytics changes to measure impact — none here. #8444 separately stops the same credential reaching Monorail and is independent of this stack.
- [x] The change is user-facing — no changeset, consistent with #8427 and #8428 for this hidden command

### DIFF
--- a/packages/store/src/cli/commands/store/stripe-auth.ts
+++ b/packages/store/src/cli/commands/store/stripe-auth.ts
@@ -4,6 +4,7 @@ import StoreCommand from '../../utilities/store-command.js'
 import {storeFlags} from '../../flags.js'
 import {globalFlags, jsonFlag} from '@shopify/cli-kit/node/cli'
 import {AbortError} from '@shopify/cli-kit/node/error'
+import {isStdinPiped} from '@shopify/cli-kit/node/system'
 import {Flags} from '@oclif/core'
 
 export default class StoreStripeAuth extends StoreCommand {
@@ -39,7 +40,7 @@ export default class StoreStripeAuth extends StoreCommand {
 
   public async run(): Promise<void> {
     const {flags} = await this.parse(StoreStripeAuth)
-    const signup = flags.signup ?? (await readSignupJwtFromStdin())
+    const signup = signupFlagValue(flags.signup) ?? (await readSignupJwtFromStdin())
 
     await authenticateStoreWithApp(
       {
@@ -54,21 +55,39 @@ export default class StoreStripeAuth extends StoreCommand {
   }
 }
 
+const MAX_SIGNUP_JWT_BYTES = 8 * 1024
+const MISSING_SIGNUP_JWT = 'Missing signup JWT.'
+const MISSING_SIGNUP_JWT_GUIDANCE = 'Pass --signup <jwt>, set SHOPIFY_FLAG_SIGNUP, or pipe the JWT to stdin.'
+
+// A blank --signup is a credential that was never supplied rather than an empty one, so it falls
+// through to stdin instead of starting an authorization without it.
+function signupFlagValue(signup: string | undefined): string | undefined {
+  const trimmed = signup?.trim()
+  return trimmed === '' ? undefined : trimmed
+}
+
 export async function readSignupJwtFromStdin(
   stdin: NodeJS.ReadableStream & AsyncIterable<Buffer | string> = process.stdin,
 ): Promise<string> {
+  // An interactive stdin never ends, so reading it would hang the command instead of reporting the
+  // credential that was never supplied.
+  if (stdin === process.stdin && !isStdinPiped()) {
+    throw new AbortError(MISSING_SIGNUP_JWT, MISSING_SIGNUP_JWT_GUIDANCE)
+  }
+
   const chunks: Buffer[] = []
+  let byteLength = 0
   for await (const chunk of stdin) {
-    chunks.push(Buffer.isBuffer(chunk) ? chunk : Buffer.from(chunk))
+    const buffer = Buffer.isBuffer(chunk) ? chunk : Buffer.from(chunk)
+    byteLength += buffer.length
+    if (byteLength > MAX_SIGNUP_JWT_BYTES) {
+      throw new AbortError('The signup JWT piped to stdin is too large.', 'Pipe only the JWT.')
+    }
+    chunks.push(buffer)
   }
 
   const signup = Buffer.concat(chunks).toString('utf8').trim()
-  if (!signup) {
-    throw new AbortError(
-      'Missing signup JWT.',
-      'Pass --signup <jwt>, set SHOPIFY_FLAG_SIGNUP, or pipe the JWT to stdin.',
-    )
-  }
+  if (!signup) throw new AbortError(MISSING_SIGNUP_JWT, MISSING_SIGNUP_JWT_GUIDANCE)
 
   return signup
 }
--- a/packages/store/src/cli/services/store/auth/callback.test.ts
+++ b/packages/store/src/cli/services/store/auth/callback.test.ts
@@ -92,6 +92,100 @@ describe('store auth callback server', () => {
     ).resolves.toBe('abc123')
   })
 
+  test('waitForStoreAuthCode answers 404 for a wrong nonce, a replay, and a non-GET handoff request', async () => {
+    const port = await getAvailablePort()
+    const params = callbackParams()
+    const authorizationUrl = 'https://shop.myshopify.com/admin/oauth/authorize?signup=signed.signup.jwt'
+    const handoffUrl = `http://127.0.0.1:${port}/auth/handoff?nonce=nonce-123`
+    const statuses: Record<string, number> = {}
+    const bodies: string[] = []
+
+    const onListening = async () => {
+      const wrongNonce = await globalThis.fetch(`http://127.0.0.1:${port}/auth/handoff?nonce=wrong`, {
+        redirect: 'manual',
+      })
+      statuses.wrongNonce = wrongNonce.status
+      bodies.push(await wrongNonce.text())
+
+      const missingNonce = await globalThis.fetch(`http://127.0.0.1:${port}/auth/handoff`, {redirect: 'manual'})
+      statuses.missingNonce = missingNonce.status
+      bodies.push(await missingNonce.text())
+
+      const notGet = await globalThis.fetch(handoffUrl, {method: 'POST', redirect: 'manual'})
+      statuses.notGet = notGet.status
+      bodies.push(await notGet.text())
+
+      const served = await globalThis.fetch(handoffUrl, {redirect: 'manual'})
+      statuses.served = served.status
+      await served.text()
+
+      const replay = await globalThis.fetch(handoffUrl, {redirect: 'manual'})
+      statuses.replay = replay.status
+      bodies.push(await replay.text())
+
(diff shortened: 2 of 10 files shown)


dengjeffrey (approved):

packages/store/src/cli/commands/store/stripe-auth.ts
```diff
@@ -54,21 +55,39 @@ export default class StoreStripeAuth extends StoreCommand {
   }
 }
 
+const MAX_SIGNUP_JWT_BYTES = 8 * 1024
+const MISSING_SIGNUP_JWT = 'Missing signup JWT.'
+const MISSING_SIGNUP_JWT_GUIDANCE = 'Pass --signup <jwt>, set SHOPIFY_FLAG_SIGNUP, or pipe the JWT to stdin.'
+
+// A blank --signup is a credential that was never supplied rather than an empty one, so it falls
+// through to stdin instead of starting an authorization without it.
```
amcaplan: Nit: I think this comment could be more legible, but also it belongs above before line 43 rather than here.

packages/store/src/cli/commands/store/stripe-auth.ts
```diff
@@ -54,21 +55,39 @@ export default class StoreStripeAuth extends StoreCommand {
...
+function signupFlagValue(signup: string | undefined): string | undefined {
+  const trimmed = signup?.trim()
+  return trimmed === '' ? undefined : trimmed
+}
+
 export async function readSignupJwtFromStdin(
   stdin: NodeJS.ReadableStream & AsyncIterable<Buffer | string> = process.stdin,
 ): Promise<string> {
+  // An interactive stdin never ends, so reading it would hang the command instead of reporting the
+  // credential that was never supplied.
```
amcaplan: I don't think this comment is necessary

packages/store/src/cli/commands/store/stripe-auth.ts
```diff
@@ -54,21 +55,39 @@ export default class StoreStripeAuth extends StoreCommand {
...
+  }
+
   const chunks: Buffer[] = []
+  let byteLength = 0
   for await (const chunk of stdin) {
-    chunks.push(Buffer.isBuffer(chunk) ? chunk : Buffer.from(chunk))
+    const buffer = Buffer.isBuffer(chunk) ? chunk : Buffer.from(chunk)
+    byteLength += buffer.length
+    if (byteLength > MAX_SIGNUP_JWT_BYTES) {
+      throw new AbortError('The signup JWT piped to stdin is too large.', 'Pipe only the JWT.')
```
amcaplan: Nit: This assumes the signup JWT is in fact what's being piped, as opposed to the user accidentally piping in some other nonsense (which is perhaps more likely). I would suggest an error message that makes fewer assumptions but still directs the user to pipe only the signup JWT.

packages/store/src/cli/services/store/auth/callback.test.ts
```diff
@@ -92,6 +92,100 @@ describe('store auth callback server', () => {
     ).resolves.toBe('abc123')
   })
 
+  test('waitForStoreAuthCode answers 404 for a wrong nonce, a replay, and a non-GET handoff request', async () => {
```
amcaplan: This test structure is a bit odd. It feels like there should be several smaller tests, maybe with some shared setup steps extracted to functions. Is the concern one of test performance? If so, can we quantify the cost of whatever we think is the expensive part?

packages/store/src/cli/services/store/auth/callback.test.ts
```diff
@@ -92,6 +92,100 @@ describe('store auth callback server', () => {
...
+    const bodies: string[] = []
+
+    const onListening = async () => {
+      const wrongNonce = await globalThis.fetch(`http://127.0.0.1:${port}/auth/handoff?nonce=wrong`, {
+        redirect: 'manual',
+      })
+      statuses.wrongNonce = wrongNonce.status
+      bodies.push(await wrongNonce.text())
+
+      const missingNonce = await globalThis.fetch(`http://127.0.0.1:${port}/auth/handoff`, {redirect: 'manual'})
```
amcaplan: This case isn't reflected in the test description.

packages/store/src/cli/services/store/auth/callback.ts
```diff
@@ -119,21 +129,26 @@ export async function waitForStoreAuthCode({
...
+      }
 
-        if (authorizationRedirectUsed) {
-          res.statusCode = 410
-          res.setHeader('Cache-Control', 'no-store')
-          res.setHeader('Connection', 'close')
-          res.end('Authorization handoff already used')
+      if (requestUrl.pathname === STORE_AUTH_HANDOFF_PATH) {
+        const returnedNonce = requestUrl.searchParams.get('nonce')
+        // Every rejection answers 404 so a local prober cannot tell a wrong nonce from a spent
```
amcaplan, line 141: Just checking my understanding - the point is to drop the 410 case to make it harder for a local prober to know whether the nonce is wrong or spent?

packages/store/src/cli/commands/store/stripe-auth.ts
craigmichaelmartin: good idea

packages/store/src/cli/services/store/auth/callback.test.ts
craigmichaelmartin: changed it up

packages/store/src/cli/services/store/auth/callback.ts
craigmichaelmartin: yeah - having a uniform 404 also hides whether an auth is in flight at all. every rejection looks like any other unknown path. the real browser flow doesn't notice since it only follows the one valid link.

packages/store/src/cli/commands/store/stripe-auth.ts
```diff
@@ -54,21 +56,35 @@ export default class StoreStripeAuth extends StoreCommand {
...
+  }
+
   const chunks: Buffer[] = []
+  let byteLength = 0
   for await (const chunk of stdin) {
-    chunks.push(Buffer.isBuffer(chunk) ? chunk : Buffer.from(chunk))
+    const buffer = Buffer.isBuffer(chunk) ? chunk : Buffer.from(chunk)
+    byteLength += buffer.length
+    if (byteLength > MAX_SIGNUP_JWT_BYTES) {
+      throw new AbortError('The input piped to stdin is too large to be a signup JWT.', 'Pipe only the signup JWT.')
```
amcaplan, line 81: Definitely an improvement!

amcaplan (approved):
Looks like everything has been addressed. Great work!

amcaplan:
A rebase off main should fix the graphql-codegen issue.

PR #8459, opened 2026-09-02
Ask about demo data when creating a dev store interactively

nickwesselman:

`shopify store create dev` can already populate a new store with demo data through
`--with-demo-data`, but nothing ever asks about it. A developer creating a store
interactively — including through the inline creation flow that #8323 and #8397 just
added to `app dev` — gets an empty store and no indication that demo data was an
option.

Adds a demo data question to interactive dev store creation, asked after the name and
plan prompts.

- New shared `devStoreDemoDataPrompt` in `@shopify/organizations`, next to the existing
  name and plan prompts, re-exported as `storeDemoDataPrompt` (store) and
  `devStoreDemoDataPrompt` (app) through the same thin delegates those two already use.
  "Yes, add demo data" is preselected.
- `store create dev` gives the flag the standard tri-state behavior. `--with-demo-data` is
  renamed to `--demo-data` and gains `allowNo`, so:

  ```ts
  // An unspecified flag means "ask" where we can prompt and "no" where we can't.
  const withDemoData = flags['demo-data'] ?? (terminalSupportsPrompting() ? await storeDemoDataPrompt() : false)
  ```

  `--demo-data` and `--no-demo-data` answer the question outright and skip the prompt. An
  unspecified flag prompts in an interactive terminal and defaults to `false` in a
  non-interactive one, using the same `terminalSupportsPrompting()` check that
  `requiredIfNonInteractive` enforcement uses. `SHOPIFY_FLAG_STORE_DEMO_DATA` counts as
  specified, so it also skips the prompt.
- `app dev` has no demo data flag and rejects non-TTY runs upstream, so it always asks,
  on both the zero-store and store-picker creation paths.

One note for reviewers: `--with-demo-data` is renamed rather than aliased. The command is
`hidden`, and the flag only reached `main` alongside the unreleased inline creation work, so
no documented workflow should depend on the old spelling. Say the word if you would rather
keep `--with-demo-data` as a hidden alias.

### How to test your changes?

1. `shopify store create dev --organization-id <id> --name test-store --plan basic` in a
   terminal — asks "Populate the store with demo data?" with Yes preselected.
2. Same command with `--demo-data` or `--no-demo-data` — no question, and the created store
   reflects the flag.
3. Same command piped from another process (or with `CI=1`) — no question, demo data off.
4. `shopify app dev --reset` in an App Management organization, then either accept the
   zero-store creation flow or pick "Create a new dev store" from the picker — name, plan,
   then the demo data question, and the created store reflects the answer.

Tophat: the `app dev` path was exercised locally end to end (name, plan, demo data
question, store created, dev started). The `store create dev` flag-gating paths are
covered by unit tests but were not live-tophatted.

### Changelog

No changeset. This builds on the inline dev store creation from #8323 and #8397, which is
not released yet, so the whole flow will be covered by a single changelog entry rather than
one per PR in this stack.

### Checklist

- [ ] The change is user-facing — I've identified the correct [bump type](https://github.com/Shopify/cli/blob/main/CONTRIBUTING.md#choosing-the-right-bump-type) (`patch` for bug fixes · `minor` for new features · `major` for [breaking changes](https://github.com/Shopify/cli/blob/main/CONTRIBUTING.md#what-counts-as-a-breaking-change)) and added a changeset with `pnpm changeset add` — intentionally omitted, see Changelog above


packages/organizations/src/cli/prompts/dev.ts
```diff
@@ -19,3 +19,12 @@ export function devStorePlanPrompt(): Promise<DevStorePlan> {
     choices: devStorePlanHandles.map((handle) => ({label: PLAN_LABELS[handle], value: handle})),
   })
 }
+
+export function devStoreDemoDataPrompt(): Promise<boolean> {
+  return ui.renderConfirmationPrompt({
+    message: 'Populate the store with demo data?',
+    confirmationMessage: 'Yes, add demo data',
+    cancellationMessage: 'No, start with an empty store',
+    defaultValue: true,
```
amcaplan, line 28: The CLI defaults to no demo data; it's an opt-in. Here it's phrased as an opt-out. Do we want to align the two experiences?

packages/store/src/cli/commands/store/create/dev.ts
```diff
@@ -57,17 +56,23 @@ export default class StoreCreateDev extends Command {
...
 
+    // Keep a fully specified invocation prompt-free, so scripts running in an interactive
+    // terminal never start blocking on the demo data question.
+    const allRequiredFlagsProvided =
+      flags.name !== undefined && flags.plan !== undefined && flags['organization-id'] !== undefined
+
     const organization = await selectOrg(flags['organization-id']?.toString())
     const name = flags.name ?? (await storeNamePrompt())
     const plan = (flags.plan as DevStorePlan | undefined) ?? (await storePlanPrompt())
+    const withDemoData = flags['with-demo-data'] ?? (allRequiredFlagsProvided ? false : await storeDemoDataPrompt())
```
amcaplan: This just seems like a messy and non-intuitive choice to me. I don't expect that:

* If I specify name and plan and org I then I won't be prompted...
* But if I leave out plan then I will be prompted

packages/organizations/src/cli/prompts/dev.ts
nickwesselman: I think we should encourage new users to use it.

packages/store/src/cli/commands/store/create/dev.ts
nickwesselman: Yeah I was struggling with the boolean flag, updated to use the `--no-` pattern 👍🏻

amcaplan (approved):
I'm ambivalent about the whole idea, but I can live with it.

PR #8460, opened 2026-09-02
Explain the empty-organization store creation prompts in app dev

nickwesselman:
> [!IMPORTANT]
> **Stacked PR.** Base is `nick/dev-store-demo-data-prompt` (#8459), not `main`. Do not
> merge before #8459.

Since #8323, an App Management organization with no dev stores drops straight into
"Name for the new development store" when you run `app dev`. The prompt arrives without
context: the developer asked to start a dev session, not to create a store, and the old
flow that explained the situation ("Looks like you don't have any dev stores associated
with …") was replaced by inline creation.

Renders a short notice before the creation prompts on the zero-store path:

> You don't have any dev stores associated with `<organization>`'s Dev Dashboard. Let's create one.

```ts
// The developer never asked to create a store here, so explain why they are being
// prompted. The picker's create choice is self-explanatory and stays quiet.
onCreateStoreWhenEmpty = async () => {
  renderInfo({body: emptyOrgNoticeBody(org)})
  return createStoreInline()
}
```

Scoped deliberately to the zero-store path. When a developer picks "Create a new dev
store" from the store picker, they have already expressed the intent, so a notice
explaining why they are being asked would be noise. That is why the message lives in the
`onCreateStoreWhenEmpty` wrapper rather than inside `createStoreInline`, which both paths
share. Wrapping the callback rather than placing `renderInfo` beside the cap and TTY
guards also means the notice cannot print if the prompts never run. Tests assert it both
ways — present on the zero-store path, `expect(renderInfo).not.toHaveBeenCalled()` on the
picker path.

One naming note for reviewers: the message hardcodes "Dev Dashboard", matching the
adjacent copy in `AppManagementClient.getCreateDevStoreLink`. The client also carries a
`webUiName` property whose value is `'Developer Dashboard'`, so the codebase already
disagrees with itself about the product name. Using `webUiName` here would render
"Developer Dashboard" instead. Reconciling the two feels like its own cleanup rather than
part of this change, but say the word if you would rather this PR use `webUiName`.

### How to test your changes?

1. Link an app in an App Management organization that has **no** dev stores.
2. Run `shopify app dev --reset`. The notice appears, naming the organization, before the
   store name prompt.
3. Run `shopify app dev --reset` in an organization that **does** have dev stores and pick
   "Create a new dev store" from the picker. No notice — straight to the name prompt.

### Changelog

No changeset. This builds on the inline dev store creation from #8323 and #8397, which is
not released yet, so the whole flow will be covered by a single changelog entry rather than
one per PR in this stack.

### Checklist

- [ ] The change is user-facing — I've identified the correct [bump type](https://github.com/Shopify/cli/blob/main/CONTRIBUTING.md#choosing-the-right-bump-type) (`patch` for bug fixes · `minor` for new features · `major` for [breaking changes](https://github.com/Shopify/cli/blob/main/CONTRIBUTING.md#what-counts-as-a-breaking-change)) and added a changeset with `pnpm changeset add` — intentionally omitted, see Changelog above

### DIFF
--- a/packages/app/src/cli/services/dev/select-store.test.ts
+++ b/packages/app/src/cli/services/dev/select-store.test.ts
@@ -12,7 +12,7 @@ import {
 import {testDeveloperPlatformClient} from '../../models/app/app.test-data.js'
 import {ClientName} from '../../utilities/developer-platform-client.js'
 import {sleep} from '@shopify/cli-kit/node/system'
-import {isTTY, renderSuccess, renderTasks, Task} from '@shopify/cli-kit/node/ui'
+import {isTTY, renderInfo, renderSuccess, renderTasks, Task} from '@shopify/cli-kit/node/ui'
 import {AbortError, CancelExecution} from '@shopify/cli-kit/node/error'
 import {createDevStore} from '@shopify/organizations'
 import {beforeEach, describe, expect, vi, test} from 'vitest'
@@ -404,6 +404,8 @@ describe('selectStore', async () => {
       json: false,
       summary: false,
     })
+    // The developer picked the create choice, so no explanatory notice is needed.
+    expect(renderInfo).not.toHaveBeenCalled()
     expect(renderSuccess).toHaveBeenCalledWith({headline: 'Development store "store1" created successfully.'})
   })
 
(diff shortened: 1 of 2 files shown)


dmerand (approved):
I could have sworn the original implementation had this... I definitely *thought* I'd done it but maybe it didn't make its way into the final PR. Thanks for updating!

amcaplan (approved):
The change makes sense. Great catch!

PR #8462, opened 2026-09-02
Add hidden App Doctor submit command

jek:
Adds `shopify app doctor submit`, which reads the local App Doctor trace, builds a fail-closed allowlisted submission, and uploads it to App Management. The submission carries only check outcomes, finding fingerprints, redacted titles and justifications, engine and CLI versions, and a trace digest. No source code, file paths, snippets, evidence,
 messages, or commit identifiers are sent.

The command extends BaseCommand rather than AppLinkedCommand so trace validation runs before the app is linked. It always writes `.shopify/app-doctor/submission.json` for inspection, then either stops (`--dry-run`), asks for confirmation, or uploads (`--force`). `--json` emits a single tagged result and links non-interactively. In non-interactive terminals `--force` is required unless `--dry-run` is set.

Plumbing: `appRequestScanUploadUrl` and `appScanCreate` mutations with generated types, `generateScanUploadUrl`/`createAppScan` on the developer platform client, an `artifactName` label for `uploadToGCS`, and a `skipPrompts` option on `linkedAppContext`.

The command is hidden, so there is no changeset. The App Management schema additions live in the gitignored local schema; codegen and the `graphql-schema` CI job will fail until Core ships the contract.

Co-authored-by: AI <noreply@pi.dev>

Assisted-By: devx/f000c74b-02bd-420a-abaf-487fbbbd8b3e

Implements "submit-to-shopify" flow for App Doctor, sending a redacted copy of scan results to Shopify for analysis.

To be used after `shopify app doctor` has run and produced results.

image

image

(will update this after core submission is wired)
image

image

image

### How to test your changes?

$ shopify app doctor 
$ shopify app doctor submit

### Checklist


packages/app/src/cli/commands/app/doctor/submit.ts
```diff
@@ -0,0 +1,72 @@
...
+      char: 'f',
+      description: 'Skip confirmation. Required if non interactive.',
+      env: 'SHOPIFY_FLAG_FORCE',
+      default: false,
+    }),
+    'dry-run': Flags.boolean({
+      description: 'Write the submission payload without uploading it.',
+      env: 'SHOPIFY_FLAG_APP_DOCTOR_DRY_RUN',
+      default: false,
+    }),
```
dmerand: It'd be good to have a `--json` option. This is also helpful in general to encourage decoupling command results from rendering.

packages/app/src/cli/services/app-doctor-engine/submission/index.ts
```diff
@@ -0,0 +1,228 @@
...
+  }
+  attestation: {trace_digest: string}
+}
+
+function submissionFinding(finding: TraceFinding): SubmissionFinding {
+  const common = {
+    fingerprint: finding.fingerprint,
+    source: finding.source,
+    severity: finding.severity,
+    title: redactText(finding.title),
```
dmerand: The command says that it does not send source code, file paths, or code snippets. This line adds the complete finding title to the upload. `redactText()` removes only text that matches a known secret pattern. It does not remove a path such as `/Users/alice/private-app/server.ts`. Line 210 also adds the complete suppression reason. Remove these free-text fields, or add validation and tests that prevent source code, file paths, and code snippets in these fields.

packages/app/src/cli/services/app-doctor-submit-api.ts
```diff
@@ -0,0 +1,57 @@
...
+  const byteSize = await dependencies.fileSize(options.submissionPath)
+  const uploadResult = await options.developerPlatformClient.generateSourceScanUploadUrl({
+    appId: options.app.id,
+    byteSize,
+  })
+  if (!uploadResult.sourceScanUploadUrl || uploadResult.userErrors.length > 0) {
+    throw new AbortError(userErrorMessage(uploadResult.userErrors, 'Shopify did not return a source scan upload URL.'))
+  }
+
+  await dependencies.upload(uploadResult.sourceScanUploadUrl, options.submissionPath, {
```
dmerand: The confirmation shows data from the `submission` object. After confirmation, this call reads `submission.json` again. Another process can change the file before this call. The upload can then contain data that the user did not confirm. Create the JSON bytes one time. Use those same bytes to write the file, calculate the byte size, show the confirmation, and upload the report.

packages/app/src/cli/services/doctor-submit.ts
```diff
@@ -0,0 +1,228 @@
...
+  const {remoteApp, developerPlatformClient} = await dependencies.linkApp({
+    directory: appRoot,
+    clientId: options.clientId,
+    forceRelink: false,
+    userProvidedConfigName: options.configName,
+    skipPrompts: options.json,
+  })
+
+  if (!options.force) {
+    if (options.json || !dependencies.canPrompt()) {
```
dmerand: Move the `--json` and `--force` check before `linkApp()`. With the current order, the command can authenticate the user and link an app before it rejects the flags. Invalid JSON-mode input must not cause network access, authentication, or changes to the project.

packages/app/src/cli/commands/app/doctor/submit.ts
```diff
@@ -0,0 +1,72 @@
...
+  static flags = {
+    ...globalFlags,
+    path: appFlags.path,
+    config: appFlags.config,
+    'client-id': appFlags['client-id'],
+    ...jsonFlag,
+    version: Flags.string({
+      hidden: false,
+      description:
+        'Optional version tag that will be associated with this app version. If not provided, an auto-generated identifier will be generated for this app version.',
```
dmerand: This command does not create an app version. It also does not generate a version identifier when the user omits this flag. The flag only sets `report.metadata.version_tag`. The `--source-control-url` flag also sets report metadata. Change the help text for both flags to say that they add optional metadata to the App Doctor report.

packages/app/src/cli/commands/app/doctor/submit.ts
```diff
@@ -0,0 +1,72 @@
...
+import {globalFlags, jsonFlag} from '@shopify/cli-kit/node/cli'
+
+export default class DoctorSubmit extends BaseCommand {
+  static hidden = true
+
+  static summary = 'Submit App Doctor results to Shopify.'
+
+  static descriptionWithMarkdown = `Reads the most recent App Doctor trace, writes a \`.shopify/app-doctor/submission.json\` file for inspection, asks for confirmation, and uploads the result to Shopify.
+
+Generated report fields exclude source code, file paths, code snippets, evidence, finding messages, and commit identifiers. Optional feedback is included without redaction. Optional \`--version\` and \`--source-control-url\` metadata is included only when supplied. Use \`--dry-run\` to write and inspect the exact payload without uploading it.`
```
dmerand: `app-doctor-engine/INSTRUCTIONS.md:96` says that `shopify app doctor submit` is for a future workflow. This PR adds that command. The CLI copies these instructions to the generated App Doctor artifacts. As a result, the instructions tell agents not to use the new command. Update the instructions to describe the new submit workflow.

packages/app/src/cli/commands/app/doctor/submit.ts
dmerand: I'm wondering if this flag should even be allowed? Do we want to indicate a version anything other than what is currently deployed? Would we validate this version?

packages/app/src/cli/services/app-context.ts
```diff
@@ -53,6 +54,7 @@ interface LoadedAppContextOptions {
   forceRelink: boolean
   clientId: string | undefined
   userProvidedConfigName: string | undefined
+  skipPrompts?: boolean
```
dmerand: `--json` must only select the output format. It must not change app-resolution behavior. A programmatic caller must supply each required value through a flag. Use `--config` to select the local configuration, `--client-id` to select the remote app, and `--force` to skip confirmation.

Please remove `skipPrompts` from this shared service. Resolve the submit target from the flags or from an already linked configuration. If a required value is missing, stop before authentication and name the required flag. The submit command only uses `remoteApp` and `developerPlatformClient`. It does not need the full linked-app load or a new link workflow. Also, `link(options, false)` only disables the success message. It does not disable prompts.

packages/app/src/cli/services/app-doctor-submit-api.ts
```diff
@@ -0,0 +1,57 @@
...
+  options: SubmitAppDoctorScanOptions,
+  dependencies: SubmitAppDoctorScanDependencies = defaultDependencies,
+): Promise<void> {
+  const byteSize = await dependencies.fileSize(options.submissionPath)
+  const uploadResult = await options.developerPlatformClient.generateSourceScanUploadUrl({
+    appId: options.app.id,
+    byteSize,
+  })
+  if (!uploadResult.sourceScanUploadUrl || uploadResult.userErrors.length > 0) {
+    throw new AbortError(userErrorMessage(uploadResult.userErrors, 'Shopify did not return a source scan upload URL.'))
```
dmerand: `submitAppDoctorScan()` converts structured API responses to `AbortError`. `uploadToGCS()` can also throw `AbortError`. `BaseCommand.catch()` always sends these errors to the terminal error renderer. As a result, an API failure in `--json` mode produces a terminal error banner instead of JSON.

Return a typed result for expected upload and create failures. Keep the API user errors and the accepted state as data. Let the command boundary select JSON or human output. Add a command test for an API failure with `--json`.

packages/app/src/cli/services/doctor-submit.ts
```diff
@@ -0,0 +1,228 @@
...
+  confirm: renderDoctorSubmitConfirmation,
+  submitScan: submitAppDoctorScan,
+  renderDryRun: renderDoctorSubmitDryRun,
+  renderSuccess: renderDoctorSubmitSuccess,
+  output: outputResult,
+  now: () => new Date().toISOString(),
+  cliVersion: CLI_KIT_VERSION,
+}
+
+function dryRunJsonResult(submissionPath: string, submission: AppDoctorSubmission): DoctorSubmitJsonResult {
```
dmerand: Nit: do these `Result` types belong in `doctor-submit-output`?

packages/app/src/cli/services/doctor-submit.ts
```diff
@@ -0,0 +1,228 @@
+import {linkedAppContext} from './app-context.js'
+import {appDoctorArtifactPaths, readTrace, writeSubmission} from './app-doctor-artifacts.js'
+import {findAppRoot} from './app-doctor-engine/scanners/discover.js'
```
dmerand: Nit: `findAppRoot()` throws `AppRootDiscoveryError` when it cannot find an app. The original App Doctor flow calls `resolveAppDoctorRoot()` so the command treats this as an expected user error. This direct call bypasses that boundary. As a result, the global handler can classify an invalid `--path` as a CLI defect and send it to crash reporting. Map this error at the command boundary. Add a command test for a missing app root in JSON and human output modes.

packages/app/src/cli/commands/app/doctor/submit.test.ts
```diff
@@ -0,0 +1,123 @@
+import DoctorSubmit from './submit.js'
+import {appFlags} from '../../../flags.js'
+import doctorSubmit from '../../../services/doctor-submit.js'
+import AppLinkedCommand from '../../../utilities/app-linked-command.js'
+import BaseCommand from '@shopify/cli-kit/node/base-command'
+import {cwd, resolvePath} from '@shopify/cli-kit/node/path'
+import {terminalSupportsPrompting} from '@shopify/cli-kit/node/system'
+import {beforeEach, describe, expect, test, vi} from 'vitest'
+
+vi.mock('../../../services/doctor-submit.js')
```
dmerand, line 12: Nit: The command test mocks `doctorSubmit()`. The service tests also replace app resolution, API submission, rendering, and output. No test covers the complete command boundary. Add a test that covers flag parsing, root and trace input, JSON output, and exit status. Include a successful dry run and an API failure. Use the real result conversion and error handler in this test.

packages/app/src/cli/services/doctor-submit.ts
```diff
@@ -0,0 +1,228 @@
...
+import {buildSubmission, SUBMISSION_SCHEMA_VERSION} from './app-doctor-engine/submission/index.js'
```
dmerand: Nit: `doctor-submit.ts` and other outer services import `submission/index.ts`, `scanners/discover.ts`, and `types.ts` directly. This expands the engine implementation boundary. Export the submission use case and its public types from `app-doctor-engine/index.ts`. Import them through that entry point.

packages/app/src/cli/services/app-doctor-submit-api.ts
```diff
@@ -0,0 +1,57 @@
...
+
+function userErrorMessage(userErrors: {message: string}[], fallback: string): string {
+  return userErrors.map(({message}) => message).join(', ') || fallback
+}
+
+export async function submitAppDoctorScan(
+  options: SubmitAppDoctorScanOptions,
+  dependencies: SubmitAppDoctorScanDependencies = defaultDependencies,
+): Promise<void> {
+  const byteSize = await dependencies.fileSize(options.submissionPath)
```
dmerand: Core accepts a source-scan report that is no larger than 1 MiB. The shared upload helper permits 100 MiB. A dry run can therefore accept a report that the remote API must reject. Apply the 1 MiB source-scan limit to the exact serialized bytes before the command requests an upload URL. Add tests at the limit and above the limit.

dmerand (changes requested):
I left pretty verbose review comments here, hopefully that's not too painful. I've tried to be clear about what's a "nit" (fix at your discretion) and what requires at a minimum some evaluation before shipping.

packages/app/src/cli/services/app-doctor-submit-api.ts
jek: This behavior makes sense as a goal.  For EA, core's 1 MiB isn't fixed and we may adjust it.  I favor deferring this until we have a value that's documentable as a limit & safe to inline into cli releases.

packages/app/src/cli/commands/app/doctor/submit.ts
jek: Fixed to clarify this is a reference to the app version corresponding to the local file state, and not create an app version.

jplhomer (approved):
My review agent only came back with a couple nits: 

- `feedback` being unsanitized (meh, user is providing this at the end of the day) \
- `title` is redundant with agent check, could "leak" private package names... the user has an opportunity to review this so also falls in the "meh" category for me.

packages/app/src/cli/services/app-doctor-engine/submission/index.ts
```diff
@@ -0,0 +1,226 @@
...
+}
+
+export function buildSubmission(trace: TraceV2, options: BuildSubmissionOptions): AppDoctorSubmission {
+  return {
+    schemaVersion: SUBMISSION_SCHEMA_VERSION,
+    report: {
+      trace_schema_version: trace.schema_version,
+      engine: {
+        name: redactText(trace.engine.name),
+        version: redactText(trace.engine.version),
```
dmerand, line 179: **Block: do not upload arbitrary text from the unsigned trace**

`redactText()` only removes known secret formats. It does not remove source text or file paths. The trace digest does not establish a trusted producer because it is unsigned and a process that changes the trace can recompute it.

I changed `trace.engine.version` to `/Users/alice/private-app/server.ts: eval(req.body.code)`, recomputed `attestation.digest`, and ran the current validator. `validateTrace()` returned `{valid: true}`. `buildSubmission()` then copied that text into `report.engine.version`. The confirmation still says that file paths and code snippets are excluded.

Other copied strings, such as language names and check IDs, use the same trust boundary. Derive transmitted text from trusted local definitions, validate it against closed values, or omit it. Do not use `redactText()` as a source or path filter. Feedback is different because the command explicitly discloses that it is sent without redaction.

packages/app/src/cli/services/doctor-submit.ts
```diff
@@ -0,0 +1,180 @@
...
+        submittedAt,
+        versionTag: options.versionTag,
+        feedback,
+      }),
+    )
+
+  const feedback = await resolveExplicitFeedback(options, dependencies)
+  let payload = prepareFeedback(feedback)
+
+  if (options.dryRun) {
```
dmerand, line 137: **Non-blocking: validate the target during a dry run**

This return occurs before `resolveClientId()`. Therefore, `--dry-run --config missing` succeeds. The instructions tell callers to use the same target selection for the dry run and the upload.

Please resolve the client ID from the selected local configuration during the dry run. This change will detect invalid flags and configuration before the upload. It will not require authentication or a network request.

packages/app/src/cli/services/doctor-submit-json.ts
```diff
@@ -0,0 +1,50 @@
...
+        ...(result.error.nextSteps === undefined
+          ? {}
+          : {next_steps: result.error.nextSteps.map((step) => unstyled(itemToString(step)))}),
+      },
+    }
+  }
+
+  const payload = {path: result.payload.path, schema_version: result.payload.schemaVersion}
+  if (result.status === 'dry-run') return {operation: 'submit', dry_run: true, payload}
+  return {operation: 'submit', dry_run: false, payload, submitted_at: result.submittedAt}
```
dmerand: **Non-blocking: include the target identity in JSON success output**

`doctorSubmit()` resolves the remote app and gives its title to the human output. This mapper removes all target data. If a caller uses the default or cached configuration, its logs cannot show which app received the report.

Please include a stable target identifier, such as `client_id` or the remote app ID, in the structured result and the JSON success fixtures.

dmerand (approved):
Agent review found a few things worth looking at. I can't smoke-test this in production so tophat is limited. Otherwise, this is OK to ship as a hidden command for early access.

packages/app/src/cli/services/doctor-submit-json.ts
jek: Updated

packages/app/src/cli/services/doctor-submit.ts
jek: Updated to resolve from local

packages/app/src/cli/services/app-doctor-engine/submission/index.ts
jek: The contract here is preventing unintended disclosure from normal App Doctor output, not authenticating the producer of an arbitrarily rewritten trace. Agreed that the digest is not a signature and that redactText() is not a
source-code or path filter.

In the normal scan/compile path, engine version comes from the CLI version, language names come from a fixed mapping, and built-in check IDs come from the check registry.  If this path places source code or paths into transmitted fields, I agree it'd would be blocking.  But I don’t consider manipulating the trace a blocker under the intended contract.

PR #8463, opened 2026-09-02
Skip hidden commands in breaking change detection

nickwesselman:

The breaking change check treats every command in `packages/cli/oclif.manifest.json` as
public surface. `extractManifestSurface` never reads a command's `hidden` field, so
renaming a flag on a hidden command fails `Breaking change detection` even though nothing
users can discover has changed.

`CONTRIBUTING.md` scopes the stable **Command surface** to what the CLI exposes; a command
marked `static hidden = true` isn't part of it. The check is currently stricter than the
written policy.

This surfaced on #8459, which renames `--with-demo-data` to `--demo-data` on
`store store create dev` — a `hidden` command whose flag only reached `main` alongside
unreleased work. The check reported a removed flag and a removed env var and failed the PR.

Skips hidden commands when building the manifest surface:

```js
if (cmd.hidden) continue
```

Because `removedCommands`, `removedFlags` and `removedEnvVars` all derive from that
surface, one guard covers all three scans.

Two deliberate details:

- **`hidden` is read from each manifest independently**, not unioned across baseline and
  current. A command that was visible in the baseline and is hidden now still reports as a
  removed command. That keeps the obvious loophole closed — a PR can't hide a command and
  strip its flags in the same change without being flagged — and it errs the way the rest
  of this file already errs, "widening rather than silently skipping potential removals."
- **An env var shared with a visible command stays tracked** through that command, so
  removing it there is still reported.

`extractManifestSurface` is now exported so it can be tested directly.

### How to test your changes?

`node --test workspace/src/major-change-check.test.js` — 20 tests pass, including four new
ones covering hidden-command omission, env vars used only by hidden commands, env vars
shared with a visible command, and the per-manifest `hidden` reading.

I also ran the extractor against the real manifests to confirm both directions:

- `main` vs #8459's branch (the hidden-command flag rename) → `removedCommands: []`,
  `removedFlags: []`, `removedEnvVars: []`, so the check passes.
- `main` with a flag deleted from `app build` (a visible command) → still reports
  `app:build --auth-alias`, so genuine breaking changes are unaffected.

### Checklist

- [ ] The change is user-facing — I've identified the correct [bump type](https://github.com/Shopify/cli/blob/main/CONTRIBUTING.md#choosing-the-right-bump-type) (`patch` for bug fixes · `minor` for new features · `major` for [breaking changes](https://github.com/Shopify/cli/blob/main/CONTRIBUTING.md#what-counts-as-a-breaking-change)) and added a changeset with `pnpm changeset add` — not user-facing, CI tooling only

### DIFF
--- a/workspace/src/major-change-check.js
+++ b/workspace/src/major-change-check.js
@@ -163,11 +163,24 @@ async function parseManifest(directory) {
   }
 }
 
-function extractManifestSurface(manifest) {
+export function extractManifestSurface(manifest) {
   const surface = {commands: {}, envVars: {}}
   if (!manifest?.commands) return surface
 
   for (const [cmdName, cmd] of Object.entries(manifest.commands)) {
+    // Hidden commands are not part of the stable command surface that
+    // CONTRIBUTING.md defines, so renaming or removing their flags is not a
+    // breaking change. Dropping them here excludes them from the removed
+    // command, flag and env var scans alike, since all three derive from this
+    // surface. An env var shared with a visible command stays tracked through
+    // that command.
+    //
+    // Hidden-ness is read from each manifest independently, on purpose: a
+    // command that was visible in the baseline and is hidden now still reports
+    // as a removed command, so a PR cannot hide a command and strip its flags
+    // in one move without being flagged.
+    if (cmd.hidden) continue
+
     const flags = {}
     if (cmd.flags) {
       for (const [flagName, flag] of Object.entries(cmd.flags)) {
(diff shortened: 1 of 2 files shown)


amcaplan (approved):

amcaplan:
I really appreciate the dedication to fixing broken windows in the experience of developing the CLI!

nickwesselman:
I'm going to abandon this as the check is non-blocking and there is some nuance. We might hide a command to deprecate it, in which case removing after would be the breaking change.

PR #8471, opened 2026-09-03
Support target-level input query variables on functions

saga-dasgupta:

Function input queries have always been **per-target** — each `[[targeting]]` entry gets its own `input_query`. Their variables, however, were **per-function**: a single extension-level `[input.variables]` applied to every target.

[shop/world#1015868](https://github.com/shop/world/pull/1015868) and [shop/world#1016656](https://github.com/shop/world/pull/1016656) close that asymmetry in Core by adding `input_query_variables` to `Execution::FunctionTarget`. Both are merged. This is the CLI half.

In [`function.ts`](packages/app/src/cli/models/extensions/specifications/function.ts):

1. **Accepts `input_variables` on a `[[targeting]]` entry.** Same `{namespace, key}` shape as the extension-level field, now extracted into a shared `InputVariablesSchema` so the two can't drift.

2. **Forwards it to the deploy payload** as `input_query_variables: {single_json_metafield: ...}` on each target — the shape `ShopifyVM::FunctionExtension::FunctionTarget` accepts, identical to the extension-level field.

3. **Rejects mixing the two levels** via `superRefine`, mirroring `InputQueryValidator#variables_placement_errors`.

```toml
# now valid
[[targeting]]
target = "cart.lines.discounts.generate.run"
input_query = "run.graphql"
input_variables = { namespace = "my-namespace", key = "my-key" }

# rejected: variables declared at both levels
[input.variables]
namespace = "extension-namespace"
key = "extension-key"

[[targeting]]
target = "cart.lines.discounts.generate.run"
input_variables = { namespace = "target-namespace", key = "target-key" }
```

The error is reported once against `[input.variables]` — the field Core reports on, and the one the message asks you to delete — however many targets declare variables of their own:

> Input variables must be defined either at the extension level or on a target, not both. Remove `input.variables` from your extension configuration. Declare `input_variables` on each target that needs them instead.

## How to test your changes?

```
pnpm i -g --@shopify:registry=https://registry.npmjs.org @shopify/cli@0.0.0-snapshot-20260914192509
```
Verified end to end on a real app

Both target and extension level input query variable placements were exercised against a dev store using a snapshot of this branch:

- **Target level** — a Cart & Checkout Validation function declaring `input_variables` on its single target.
- **Extension level** — a Discount function declaring `[input.variables]` once and fanning out to two targets (`cart.lines.discounts.generate.run` and `cart.delivery-options.discounts.generate.run`), where each target's input query declares only the variable it needs and ignores the other top-level key in the same metafield.

Both pushed cleanly via `shopify app dev` and ran as expected at checkout, applying the functions based on the provided variable values. 

## Measuring impact

- [x] n/a - this doesn't need measurement, e.g. a linter rule or a bug-fix

## Checklist

- [x] I've added a changeset if this PR contains user-facing or noteworthy changes
- [x] I've added tests to cover my changes
- [ ] I've updated the documentation if applicable


adampetro (approved):

gonzaloriestra (approved):
Tested with real storefront API carts and function logs:
- Configured product and delivery targets with separate variable metafields. Enabled each independently, then both and neither, and checked the cart discounts.
- Moved variables to the extension level and checked that both targets used the shared metafield.
- Tested one target with variables and another without, then both without variables.
- Declared variables at both levels and checked that validation reported one error.

PR #8473, opened 2026-09-03
Fix false success message when CLI self-upgrade installs the wrong version

craigmichaelmartin:

A user behind a private npm registry ran the CLI self-update on 4.7.0: the registry (401ing against upstream) resolved a stale `latest` tag, so `npm install -g @shopify/cli@latest` **downgraded** them to 3.94.3 with a zero exit code — while the CLI printed "Shopify CLI upgraded. You're now on version 4.7.1."

Root cause: `runCLIUpgrade` built the success banner from the cached version check (which queries the public npm registry) and trusted the install's exit code, but the install resolves through whatever registry the user's npm config points at. The two can disagree, and the exit code alone doesn't prove the right version landed.

In `packages/cli-kit/src/public/node/upgrade.ts`, after the package-manager install completes, verify what was actually installed using the existing `globalCLIVersion()` helper (runs the installed `shopify` binary and parses its version):

- If the installed version is older than expected (the cached newer version, or the current version when none was cached), throw an `AbortError` naming both versions, with a hint that the package manager may be resolving `@shopify/cli` from a registry with outdated versions.
- If the version can't be determined, throw an `AbortError` suggesting `shopify version`.
- On success, the banner reports the **verified** installed version instead of the cached prediction.

The auto-upgrade postrun hook already catches errors from `runCLIUpgrade`, so a failed background upgrade now logs, shows the upgrade reminder, records `env_auto_upgrade_success: false`, and reports to Bugsnag instead of silently claiming success.

Minor wording change: `shopify upgrade` when already up to date now says "You're now on version X." (the verified version) instead of "You're now on the latest version."

### How to test your changes?

1. Unit tests: `pnpm vitest run src/public/node/upgrade.test.ts` in `packages/cli-kit` (new cases cover the downgrade, same-version, unverifiable, and verified-success paths).
2. Manually: with a global install, run `shopify upgrade` normally — the success banner should show the actual installed version. To simulate the failure, point npm at a registry serving an older `@shopify/cli` (`npm config set registry ...`) and run `shopify upgrade` — it should abort with "Failed to upgrade Shopify CLI. Expected to be on version X, but version Y is now installed." instead of a success banner.

### Checklist

### DIFF
--- a/packages/cli-kit/src/public/node/upgrade.test.ts
+++ b/packages/cli-kit/src/public/node/upgrade.test.ts
@@ -1,4 +1,4 @@
-import {isDevelopment} from './context/local.js'
+import {isDevelopment, isUnitTest} from './context/local.js'
 import {currentProcessIsGlobal, inferPackageManagerForGlobalCLI} from './is-global.js'
 import {checkForCachedNewVersion, packageManagerFromUserAgent, PackageManager} from './node-package-manager.js'
 import {exec, isCI} from './system.js'
@@ -10,8 +10,10 @@ import {
   versionToAutoUpgrade,
 } from './upgrade.js'
 import {Notification, fetchNotifications} from './notifications-system.js'
-import {isPreReleaseVersion} from './version.js'
+import {globalCLIVersion, isPreReleaseVersion} from './version.js'
+import {mockAndCaptureOutput} from './testing/output.js'
 import {getAutoUpgradeEnabled} from '../../private/node/conf-store.js'
+import {CLI_KIT_VERSION} from '../common/version.js'
 import {vi, describe, test, expect, beforeEach} from 'vitest'
 
 vi.mock('./notifications-system.js', async (importOriginal) => {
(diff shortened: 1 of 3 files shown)


packages/cli-kit/src/public/node/upgrade.test.ts
```diff
@@ -219,6 +228,63 @@ describe('runCLIUpgrade', () => {
...
+    expect(outputMock.info()).toContain("You're now on version 4.7.1")
+  })
+
+  test('throws when the install lands an older version than expected (e.g. a stale private registry)', async () => {
+    // Given
+    vi.mocked(currentProcessIsGlobal).mockReturnValue(true)
+    vi.mocked(inferPackageManagerForGlobalCLI).mockReturnValue('npm')
+    vi.mocked(exec).mockResolvedValue()
+    vi.mocked(checkForCachedNewVersion).mockReturnValue('4.7.1')
+    vi.mocked(globalCLIVersion).mockResolvedValue('3.94.3')
```
amcaplan, line 256: Might be slightly better to have a more subtly reduced version, like 4.7.0. But it's not a big deal either way - ultimately semver underlies this all anyway.

amcaplan (approved):
Makes sense to me!

PR #8477, opened 2026-09-04
Fall back to local validation when a remote contract cannot be compiled

alfonso-noriega:

A server-provided module validation schema that AJV cannot compile crashes **every `shopify app *` command at load time** — including for apps that don't use the affected module at all. `createConfigExtensionInstances` parses the app config against every remote app-config spec, so one bad contract takes down the whole CLI for that environment.

Live repro: any app command against a `shop/world` local rig whose custom-data spec renders `"enum": []` for standard metaobject templates (empty taxonomy registry — empty `enum` is invalid JSON Schema):

```
schema is invalid: data/definitions/MetaobjectNamespaces/properties/standard_metaobjects/items/enum
must NOT have fewer than 1 items, …
    at map
    at createConfigExtensionInstances
```

The server-side instance of that bug is fixed in shop/world#1016890, but the CLI shouldn't hard-fail on data it doesn't control: a malformed contract for one module should degrade, not brick `app info`.

`unifiedConfigurationParserFactory` now degrades per module when a contract is unusable:

- **Normalisation failure** (broken `$ref`): fall back to the merged spec's local zod parser for that module
- **Compilation failure** (first `jsonSchemaValidate` call, e.g. empty enum): return the zod parse result for that call
- Either way: `outputWarn` once per module identifier per process, naming the module and the underlying reason — and noting the server still validates on deploy, so nothing ships unvalidated

### How to test your changes?

- `pnpm vitest run packages/app/src/cli/utilities/json-schema.test.ts` — 3 new tests: empty-enum contract → zod result + single warning (deduped on repeat calls); zod errors still gate when the contract is broken; broken `$ref` → factory returns the local parser
- Manual: point the CLI at a rig serving the invalid custom-data contract — `shopify app info` loads with a warning instead of crashing

### Measuring impact

- [x] n/a — resilience fix; the warning line is the observable change

### DIFF
--- a/.changeset/soft-schema-compilation-failures.md
+++ b/.changeset/soft-schema-compilation-failures.md
@@ -1,0 +1,5 @@
+---
+'@shopify/app': patch
+---
+
+Fall back to local validation instead of crashing when a server-provided module validation schema cannot be compiled
--- a/packages/app/src/cli/utilities/json-schema.test.ts
+++ b/packages/app/src/cli/utilities/json-schema.test.ts
@@ -1,6 +1,7 @@
 import {unifiedConfigurationParserFactory} from './json-schema.js'
 import {describe, test, expect} from 'vitest'
 import {randomUUID} from '@shopify/cli-kit/node/crypto'
+import {mockAndCaptureOutput} from '@shopify/cli-kit/node/testing/output'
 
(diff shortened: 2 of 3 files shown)


alfonso-noriega:
Closing — the triggering issue is rig-local; withdrawing for now.

PR #8478, opened 2026-09-04
Add subscription migration list command

tyler-eon:
## What this changes

Adds the subscription migration command:

```sh
shopify app subscription-migrations list
```

The command:

- fetches every page from the new `migratableAppSubscriptions` Partners CLI GraphQL query;
- optionally filters by `UNSCHEDULED`, `SCHEDULED`, or `MIGRATED`;
- streams a comprehensive CSV to stdout by default;
- streams one versioned JSON document to stdout with `--json`;
- supports saving either format through normal shell redirection;

The read model can preserve historical `NONE` notification values returned by the API without re-enabling `NONE` as a schedule-input option.

## Examples

```sh
shopify app subscription-migrations list
shopify app subscription-migrations list --status UNSCHEDULED
shopify app subscription-migrations list --json
shopify app subscription-migrations list > subscriptions.csv
shopify app subscription-migrations list --json > subscriptions.json
```

## Verification

- 22 focused test files: 251 passed, 1 platform-specific skip
- App and CLI Kit type-checks
- App, CLI, and CLI Kit builds
- ESLint
- Knip
- Oclif manifest generation
- Command snapshot validation
- Direct hidden-command help and normal-discovery checks
- CSV escaping, JSON schema, stdout purity, nullable API response, and cursor pagination coverage


packages/app/src/cli/commands/app/subscription-migrations/flags.ts
```diff
@@ -26,6 +28,26 @@ const statusWatchFlag = {
   }),
 }
 
+export const listFlags = {
+  ...sharedFlags,
+  output: Flags.string({
+    description: 'Path to write the subscription export.',
+    env: 'SHOPIFY_FLAG_OUTPUT',
```
dmerand: UNIXey nit: Do we need `--output` / `--force`? Shouldn't folks just be able to run `subscription migrations list > file.csv`?

packages/app/src/cli/services/subscription-migrations/list-migratable-subscriptions.ts
```diff
@@ -0,0 +1,49 @@
...
+}: ListMigratableSubscriptionsOptions): Promise<MigratableSubscription[]> {
+  const subscriptions: MigratableSubscription[] = []
+  const seenCursors = new Set<string>()
+  let after: string | undefined
+
+  while (true) {
+    // Pages must be requested sequentially because each request depends on the previous opaque cursor.
+    // eslint-disable-next-line no-await-in-loop
+    const page = await getPage({clientId, first: PAGE_SIZE, after, status})
+    if (page === null) throw new AbortError('App not found')
```
dmerand: Could this throw a domain-specific error and let `List.run()` translate it to `AbortError`? Throwing `AbortError` makes three CLI-specific choices: how to show the error, which exit status to use, and how to encode the JSON error. Ideally those choices belong at the command boundary rather than in the internal execution.

dmerand (approved):
This LGTM from a code standpoint. My comments about output format are non-blocking, though my opinion is that we don't need so much code to handle output formatting when that's already handled well-enough by the OS.

packages/app/src/cli/commands/app/subscription-migrations/flags.ts
tyler-eon: I'm going to remove `--output` and `--force`, but for CSV output specifically I'm also going to make it stream each page of results. If we're writing CSV data to stdout then we can stream each page without risking a malformed csv output, but with `--json` we can't safely do that. So CSV output can stream, JSON output won't for now just to avoid a failed invocation leaving behind a malformed JSON file.

packages/app/src/cli/services/subscription-migrations/list-migratable-subscriptions.ts
tyler-eon: Done. Throws a custom error now that `List.run` catches and translates into `AbortError` at the command boundary.

PR #8491, opened 2026-09-07
Recover gracefully when pnpm blocks build scripts during extension generation

amcaplan:

Fixes https://github.com/shop/issues-develop/issues/23837

Recent pnpm versions refuse to run the build scripts of newly installed dependencies until the user approves them, and they fail the install outright (`ERR_PNPM_IGNORED_BUILDS`) when they can't prompt for approval. Since `shopify app generate extension` installs dependencies from within the generation task UI (non-interactive stdio), the prompt can never be shown, so the install always fails on affected setups.

Verified against pnpm 12.3.4: the install completes resolution and linking (creating `node_modules` inside the new extension directory) and then exits 1. Two problems with how we handled that:

1. The raw `pnpm install` failure was surfaced as-is, with no guidance on how to recover.
2. The cleanup of the partially generated extension (`removeFile(directory)` in `extensionInit`'s catch) was unguarded — if the removal failed for any reason, its error **replaced** the original install error and the partially removed extension directory (e.g. just a `node_modules` folder) was left behind, matching the state reported in the issue.

In `packages/app/src/cli/services/generate/extension.ts` (plus a small `@shopify/cli-kit` addition):

- Detects the pnpm blocked-builds failure (`ERR_PNPM_IGNORED_BUILDS`) and raises an `AbortError` with next steps: run `pnpm approve-builds` in the app directory, then generate the extension again. This works because the failed install records the ignored builds in the app root's `node_modules/.modules.yaml`, so `pnpm approve-builds` lists them even after the extension directory is removed (verified locally).
- Makes the cleanup resilient and best-effort: cli-kit's `removeFile` now accepts optional `maxRetries`/`retryDelay` options passed straight through to Node's `fs.rm`, which retries transient errors (EBUSY/ENOTEMPTY/EPERM — e.g. antivirus or indexer locks on the freshly written `node_modules`) with linear backoff, retrying the specific failing entry rather than restarting the walk. The cleanup uses `{maxRetries: 10, retryDelay: 100}` (~5.5s worst case). If removal still fails, a warning names the leftover directory and tells the user to delete it, and the error that actually interrupted the generation is still thrown.

All generation flows (UI, function, theme; workspace and non-workspace installs) go through this catch, so they're all covered.

### How to test your changes?

1. Create an app that uses pnpm workspaces, with pnpm ≥ 12 (or pnpm ≥ 10 with `strictDepBuilds: true`).
2. Ensure at least one extension-template dependency has a build script that isn't approved (no `onlyBuiltDependencies` entry).
3. Run `shopify app generate extension` and pick a JS-flavored template.
4. Before: raw `Command failed with exit code 1: pnpm install` dump. After: an error banner explaining pnpm blocked the build scripts, with `pnpm approve-builds` + regenerate as next steps, and no leftover `extensions/<name>` directory.

### Checklist

### DIFF
--- a/packages/cli-kit/src/public/node/fs.ts
+++ b/packages/cli-kit/src/public/node/fs.ts
@@ -7,7 +7,6 @@ import {
   copy as fsCopy,
   ensureFile as fsEnsureFile,
   ensureFileSync as fsEnsureFileSync,
-  remove as fsRemove,
   removeSync as fsRemoveSync,
   move as fsMove,
   // eslint-disable-next-line @typescript-eslint/ban-ts-comment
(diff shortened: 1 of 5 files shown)


packages/cli-kit/src/public/node/fs.ts
```diff
@@ -260,14 +259,29 @@ export function mkdirSync(path: string): void {
...
- * @param path - Path to the file to be removed.
+ * @param path - Path to the file or directory to be removed.
+ * @param options - Retry behavior, passed through to Node's `fs.rm`. Useful when the removal can
+ * race with transient locks, such as an antivirus scanning freshly written files.
  */
-export async function removeFile(path: string): Promise<void> {
+export async function removeFile(path: string, options: RemoveFileOptions = {}): Promise<void> {
   outputDebug(outputContent`Removing file at ${outputToken.path(path)}...`)
-  await fsRemove(path)
+  await fsRm(path, {recursive: true, force: true, ...options})
```
dmerand, line 284: TIL we have two options for removing files in this file alone. I'm presuming `fsrm` is coming from Node based on the PR description, and is therefore the better long-term dependency.

dmerand (approved):
Thanks for this change!

PR #8494, opened 2026-09-08
Dasherize the client transfer store type handle

amcaplan:

`store list --type client_transfer` was the only underscored enumerated flag value in the entire CLI manifest. Every other multi-word one is dashed:

| Command | Flag | Values |
|---|---|---|
| `app generate extension` | `--flavor` | `vanilla-js`, `typescript-react` |
| `app webhook trigger` | `--delivery-method` | `google-pub-sub`, `event-bridge` |
| `theme dev` | `--live-reload` | `hot-reload`, `full-page` |
| `theme dev` | `--reconciliation-strategy` | `keep-local`, `keep-remote` |

The underscore was inherited from Business Platform's `STORE_TYPE` casing, but `STORE_TYPE_HANDLES` was never a passthrough — three BP enum members already collapse onto `dev`, and `dev` already maps back out to BP's `development_superset`. So there was no consistency argument for keeping BP's spelling on the public side, only on the wire.

Renames the public handle to `client-transfer`. `STORE_TYPE_FILTERS` spells it back out as BP's `client_transfer` on the way to the API — the same shape the `dev` entry already had.

Two things that travel with the handle:

- `storeNounPhrase` in `list/result.ts` built the "client transfer stores" wording by splitting the handle on `_`; it now splits on `-`.
- `storeTypeLabel` needed no change — `capitalizeWords` is `capitalCase`, which already splits on either separator, so the Type column still reads `Client Transfer`.

`--type` hasn't been released yet (its changeset is still pending), so the flag side of this needs no changeset of its own. The changeset here covers `store info`'s `type` field, which has reported `client_transfer` since 4.2.0 and changes with it, because the two commands deliberately share one handle vocabulary.

Regenerated `oclif.manifest.json` and the CLI README; dev docs produced no diff.

### DIFF
--- a/.changeset/dasherize-client-transfer-store-type.md
+++ b/.changeset/dasherize-client-transfer-store-type.md
@@ -1,0 +1,5 @@
+---
+'@shopify/store': patch
+---
+
+Report the client transfer store type as `client-transfer` in `store info`
--- a/packages/cli/README.md
+++ b/packages/cli/README.md
@@ -3983,7 +3983,7 @@ List stores in a Shopify organization.
 ```
 USAGE
   $ shopify store list [-j] [--no-color] [--organization-id <value>] [--type
-    dev|production|client_transfer|collaborator] [--verbose]
+    dev|production|client-transfer|collaborator] [--verbose]
 
 FLAGS
   -j, --json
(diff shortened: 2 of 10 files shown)


dmerand (approved):

amcaplan:
Merged despite the lock, to avoid an unnecessary breaking change later.

PR #8496, opened 2026-09-08
Fix startup crash when the OS can't look up the current user's shell

### LINKED ISSUE #8497: plugins topic still crashes when the OS can't resolve the current user (residual of #8496)
### Summary

On machines where the OS cannot resolve the current user's passwd entry, the `plugins` topic still crashes with `A system error occurred: uv_os_get_passwd returned ENOMEM (not enough memory)`.

#8496 guards oclif's shell detection at the root `Config` (`ShopifyConfig._shell`), which fixes startup for every other command. This path is not reachable from that guard.

### Cause

`@oclif/plugin-plugins` resolves to `@oclif/core` **4.11.14**, while the CLI pins **4.8.3**. In 4.11 the shell lookup moved out of `Config.prototype._shell` into a module-level `getShell()` (`lib/util/os.js:37`), still calling `os.userInfo()` unguarded, and `Config.load()` calls it at `config.js:307`.

4.11's `Command.run()` unconditionally does `await Config.load(opts)`. `ShopifyConfig.runCommand` passes `this` to `commandClass.run(argv, this)`, `isConfig(this)` is true, so 4.11 rebuilds and re-loads its own `Config` from our 4.8.3 instance. A subclass cannot override a module-level function, so the guard in #8496 has no seam here.

### Affected commands

The whole (hidden) `plugins` topic: `plugins`, `plugins install`, `plugins link`, `plugins uninstall`, `plugins update`, `plugins reset`, `plugins inspect`. Everything else works after #8496.

### Reproduction

```bash
cat > /tmp/patch-os.cjs <<'JS'
const os = require('node:os')
os.userInfo = () => { throw new Error('A system error occurred: uv_os_get_passwd returned ENOMEM (not enough memory)') }
JS
env -u SHELL node --require /tmp/patch-os.cjs packages/cli/bin/run.js plugins
```

Stack: `getShell (@oclif/core@4.11.14/lib/util/os.js:37)` ← `Config.load (config.js:307)` ← `Command.run (command.js:161)` ← `ShopifyConfig.runCommand`.

### Possible fixes

- Set `process.env.SHELL` once at startup, but only when the guarded detection already failed, so 4.11's `process.env.SHELL ?? userInfo()` never reaches the syscall. Healthy machines keep their real `SHELL` for spawned subprocesses. Deliberately not done in #8496 to keep that fix scoped to the reported startup crash.
- Fix upstream in `@oclif/core` (no version guards this today), or align the `@oclif/plugin-plugins` resolution onto a guarded core.

### Priority

Low. The `plugins` topic is hidden (`packages/cli/package.json`, and `packages/cli/src/index.ts:94`), and the affected population is the intersection of "OS passwd lookup is broken" and "uses an undocumented topic". No user has reported it — found while reviewing #8496.

Related: internal Observe report shop/issues#75079.

craigmichaelmartin:

Fixes the top unhandled CLI startup crash in Observe: [shop/issues#75079](https://github.com/shop/issues/issues/75079) — `A system error occurred: uv_os_get_passwd returned ENOMEM (not enough memory)`. P2, 113 events / 53 affected users over 7 days. For these users the CLI does not start at all — no command runs, and no analytics row is emitted.

Also fixes #8497.

While loading, oclif resolves the active shell from `process.env.SHELL ?? os.userInfo().shell`. Node's `os.userInfo()` throws a `SystemError` whenever the OS can't resolve the current user, and oclif's call is unguarded — so the process dies inside the `config.load()` we invoke ourselves at `packages/cli-kit/src/public/node/cli-launcher.ts:31`.

Reported instances are Windows (`%APPDATA%\npm` global installs), where `SHELL` is normally unset so the syscall runs on every invocation. Worth noting libuv hardcodes the shell field to `NULL` on Windows, so that call can never return anything useful there — it can only succeed uselessly or crash.

This sets `SHELL` before handing off to oclif, but only when the lookup actually fails:

```ts
function setShellVariableWhenUserLookupFails(): void {
  if (process.env.SHELL !== undefined) return

  try {
    os.userInfo()
  } catch {
    process.env.SHELL = 'unknown'
  }
}
```

**Why the environment variable rather than a guard.** The obvious fix is to override oclif's `_shell()` on our `ShopifyConfig` subclass and catch the throw. That works on the pinned 4.8.3 and nothing else: 4.11 drops `_shell` from `Config` and moves the logic into a module-level `getShell()` (`lib/util/os.js:37`) that a subclass cannot reach. Both versions read the same `process.env.SHELL ?? os.userInfo().shell`, so seeding the variable is the one lever that covers both — and it does not depend on an undocumented protected member.

That also fixes #8497: the `plugins` topic resolves to `@oclif/core` **4.11.14**, whose `Command.run()` re-loads its own `Config`. A subclass override could never reach it; the environment variable does.

Behaviour on healthy machines is unchanged — the probe only runs when `SHELL` is unset, and only a throwing lookup writes anything. On a failing machine, `config.shell` becomes `'unknown'`, which is what oclif's own fallback branch emits, and feeds one consumer: `env_shell` in `packages/cli-kit/src/private/node/analytics.ts:101`. Those users currently send no analytics at all, so this is strictly more data, not less.

### Notes for the reviewer

**The tests drive the real crash path.** They spy on `node:os`'s `userInfo` — the default import is the same `module.exports` object oclif's compiled CJS does a property lookup on — and unset `SHELL`, so `config.load()` runs oclif's actual detection. Mutation-verified three ways: dropping the call from `load()`, having the helper swallow the failure without setting `SHELL`, and setting `SHELL` unconditionally without probing. Each fails exactly the expected test.

**Removal condition.** Drop this once `@oclif/core` guards the call itself. No released version does — it is unguarded in 4.8.3, 4.11.14 and on main. The note lives on the helper.

**Not added to `environmentIssueMessages`.** Leaving this class of report unhandled in Observe is what will tell us if the workaround ever stops working or the failure moves to a different frame.

### How to test your changes?

```bash
pnpm vitest run packages/cli-kit/src/public/node/custom-oclif-loader.test.ts
```

End to end, with the syscall forced to fail against a built CLI:

```bash
cat > /tmp/patch-os.cjs <<'JS'
const os = require('node:os')
os.userInfo = () => { throw new Error('A system error occurred: uv_os_get_passwd returned ENOMEM (not enough memory)') }
JS
env -u SHELL node --require /tmp/patch-os.cjs packages/cli/bin/run.js version
env -u SHELL node --require /tmp/patch-os.cjs packages/cli/bin/run.js plugins
```

Both crash on `main` and pass here.

Verified locally: `tsc --noEmit -p packages/cli-kit`, `nx run cli-kit:build`, `nx run cli:build` and `bin/run-knip-ci.js` all exit 0; eslint clean; full `packages/cli-kit` suite matches a stashed `main` baseline exactly (3 pre-existing failures in `deprecations.test.ts` and `node-package-manager.test.ts`). Two port-binding tests (`tcp.test.ts`, `graphiql/server.test.ts`) flake intermittently under load on my machine — confirmed unrelated by re-running this branch with the two new tests skipped, where they still flake at the same rate.

### Measuring impact

- [x] Error monitoring (Observe)

### DIFF
--- a/packages/cli-kit/src/public/node/custom-oclif-loader.test.ts
+++ b/packages/cli-kit/src/public/node/custom-oclif-loader.test.ts
@@ -128,7 +128,6 @@ describe('ShopifyConfig', () => {
   })
 
   test('loads successfully when the OS user lookup fails during shell detection', async () => {
-    // oclif and the guard both only reach os.userInfo() when SHELL is unset.
     vi.stubEnv('SHELL', undefined)
     vi.spyOn(os, 'userInfo').mockImplementation(() => {
       throw osUserLookupError()
--- a/packages/cli-kit/src/public/node/custom-oclif-loader.test.ts
+++ b/packages/cli-kit/src/public/node/custom-oclif-loader.test.ts
@@ -1,6 +1,8 @@
 import {ShopifyConfig} from './custom-oclif-loader.js'
 import {Config} from '@oclif/core'
 import {describe, expect, test, vi} from 'vitest'
+import os from 'node:os'
+import {fileURLToPath} from 'node:url'
 
 describe('ShopifyConfig', () => {
   test('delegates to super.runCommand when no lazy command loader is configured', async () => {
@@ -120,4 +122,46 @@ describe('ShopifyConfig', () => {
     expect(config.findCommand).not.toHaveBeenCalled()
     expect(lazyCommandLoader).toHaveBeenCalledWith('test-command')
   })
+
+  test('loads successfully when the OS user lookup fails during shell detection', async () => {
+    // SHELL short-circuits oclif's `process.env.SHELL ?? userInfo()`, so it has to be unset for the
+    // OS lookup — the call that crashes in production — to run at all.
+    vi.stubEnv('SHELL', undefined)
+    vi.spyOn(os, 'userInfo').mockImplementation(() => {
+      throw osUserLookupError()
+    })
+    const config = new ShopifyConfig({root: fileURLToPath(import.meta.url)})
+
+    await config.load()
+
+    expect(config.shell).toBe('unknown')
+  })
+
+  test('reports the shell oclif detected when the OS user lookup succeeds', async () => {
+    vi.stubEnv('SHELL', undefined)
+    vi.spyOn(os, 'userInfo').mockReturnValue({
+      username: 'test-user',
+      uid: 1000,
+      gid: 1000,
+      homedir: '/home/test-user',
+      shell: '/bin/fish',
+    })
+    const config = new ShopifyConfig({root: fileURLToPath(import.meta.url)})
+
+    await config.load()
+
+    expect(config.shell).toBe('fish')
+  })
 })
+
+/**
+ * The SystemError Node throws from os.userInfo() when the OS can't resolve the current user.
+ *
+ * @returns An error shaped like the real failure reported by affected users.
+ */
+function osUserLookupError(): Error {
+  return Object.assign(new Error('A system error occurred: uv_os_get_passwd returned ENOMEM (not enough memory)'), {
+    code: 'ERR_SYSTEM_ERROR',
+    info: {code: 'ENOMEM', errno: -4057, syscall: 'uv_os_get_passwd'},
+  })
+}
--- a/packages/cli-kit/src/public/node/custom-oclif-loader.test.ts
+++ b/packages/cli-kit/src/public/node/custom-oclif-loader.test.ts
@@ -1,10 +1,14 @@
 import {ShopifyConfig} from './custom-oclif-loader.js'
 import {Config} from '@oclif/core'
-import {describe, expect, test, vi} from 'vitest'
+import {afterEach, describe, expect, test, vi} from 'vitest'
 import os from 'node:os'
 import {fileURLToPath} from 'node:url'
 
 describe('ShopifyConfig', () => {
+  afterEach(() => {
+    vi.unstubAllEnvs()
+  })
+
   test('delegates to super.runCommand when no lazy command loader is configured', async () => {
     const config = new ShopifyConfig({root: import.meta.url})
     const superRunCommandSpy = vi.spyOn(Config.prototype, 'runCommand').mockResolvedValue('super-result')
@@ -124,8 +128,7 @@ describe('ShopifyConfig', () => {
   })
 
   test('loads successfully when the OS user lookup fails during shell detection', async () => {
-    // SHELL short-circuits oclif's `process.env.SHELL ?? userInfo()`, so it has to be unset for the
-    // OS lookup — the call that crashes in production — to run at all.
+    // oclif and the guard both only reach os.userInfo() when SHELL is unset.
     vi.stubEnv('SHELL', undefined)
     vi.spyOn(os, 'userInfo').mockImplementation(() => {
       throw osUserLookupError()
@@ -154,11 +157,6 @@ describe('ShopifyConfig', () => {
   })
 })
 
-/**
- * The SystemError Node throws from os.userInfo() when the OS can't resolve the current user.
- *
- * @returns An error shaped like the real failure reported by affected users.
- */
 function osUserLookupError(): Error {
   return Object.assign(new Error('A system error occurred: uv_os_get_passwd returned ENOMEM (not enough memory)'), {
     code: 'ERR_SYSTEM_ERROR',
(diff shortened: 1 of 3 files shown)


packages/cli-kit/src/public/node/custom-oclif-loader.test.ts
```diff
@@ -120,4 +122,46 @@ describe('ShopifyConfig', () => {
     expect(config.findCommand).not.toHaveBeenCalled()
     expect(lazyCommandLoader).toHaveBeenCalledWith('test-command')
   })
+
+  test('loads successfully when the OS user lookup fails during shell detection', async () => {
+    // SHELL short-circuits oclif's `process.env.SHELL ?? userInfo()`, so it has to be unset for the
+    // OS lookup — the call that crashes in production — to run at all.
```
amcaplan: This comment seems immediately dated - the call doesn't crash in production anymore, right?

packages/cli-kit/src/public/node/custom-oclif-loader.test.ts
```diff
@@ -120,4 +122,46 @@ describe('ShopifyConfig', () => {
...
+    await config.load()
+
+    expect(config.shell).toBe('fish')
+  })
 })
+
+/**
+ * The SystemError Node throws from os.userInfo() when the OS can't resolve the current user.
+ *
+ * @returns An error shaped like the real failure reported by affected users.
```
amcaplan: Again, something that should be a PR description/comment is leaking into the code itself.

packages/cli-kit/src/public/node/custom-oclif-loader.ts
```diff
@@ -61,4 +61,27 @@ export class ShopifyConfig extends Config {
...
+   * Guard oclif's shell detection, which otherwise crashes the CLI when the OS can't resolve the current user.
+   *
+   * Oclif populates `config.shell` during `Config.load()`. When the SHELL environment variable is unset it
+   * falls back to Node's `userInfo()`, which throws a SystemError whenever the OS passwd lookup fails —
+   * ENOMEM on Windows profiles where the lookup is broken, ENOENT for a container UID with no passwd entry.
+   * The call is unguarded, so it kills the CLI before it can run any command.
+   *
+   * Note when bumping the oclif dependency: 4.11 drops `_shell` from `Config` and moves the logic into a
+   * module-level `getShell()`, so this override stops compiling. The call is still unguarded upstream, so
+   * reinstate the guard at whatever seam replaces it rather than dropping it.
```
amcaplan: This seems like a problem waiting to happen, I don't think a guard does it. Can we research what 4.11 does instead and compensate for both, with a note to remove this temporary fix when upgrading?

amcaplan (changes requested):

packages/cli-kit/src/public/node/custom-oclif-loader.ts
craigmichaelmartin: Good call — researched it, and you were right that a guard doesn't do it.

4.11 drops `_shell` from `Config` and moves the logic to a module-level `getShell()` in `lib/util/os.js`, which a subclass can't reach. But both versions read the same expression — `process.env.SHELL ?? os.userInfo().shell` — so setting `SHELL` short-circuits the failing syscall on either one. Switched to that and dropped the override entirely.

The probe only runs when `SHELL` is unset, and only a throwing lookup writes the variable, so healthy machines are untouched. Verified against both copies in the tree:

```
BEFORE   4.11.14 getShell() -> THREW: uv_os_get_passwd returned ENOMEM
AFTER    4.8.3 root config.load() -> OK, shell=unknown
         SHELL after load: "unknown"
         4.11.14 getShell() -> unknown
HEALTHY  SHELL=/bin/zsh -> getShell() = zsh, process.env.SHELL untouched
```

The removal note is now on the helper: drop it once oclif guards the call itself. There is no released version that does.

Unplanned upside: this also fixes the `plugins` topic, which builds its own 4.11 `Config` and was still crashing under the override. That was #8497 — now closed by this PR.

packages/cli-kit/src/public/node/custom-oclif-loader.test.ts
craigmichaelmartin: Right, it was stale the moment it merged. Cut down to what the next reader actually needs to know — why `SHELL` has to be unset for the test to reach the code under test:

```ts
// oclif and the guard both only reach os.userInfo() when SHELL is unset.
```

packages/cli-kit/src/public/node/custom-oclif-loader.test.ts
craigmichaelmartin: Fair — dropped that docblock. The function name carries it, and the error string in the body is self-evident. Took the same pass over the source comments; the only prose left is the two things that aren't recoverable from the code: why we set `SHELL` at all, and when it can be deleted.

amcaplan (approved):
Thanks for the updates! Approving as it makes sense, but would upgrading oclif to 4.11 (we're already on 4.8.3) solve this already?

craigmichaelmartin:
Thanks! No unfortunately there is the same issue in all oclif 4.x and 5.x versions.

PR #8499, opened 2026-09-08
Fix App Doctor deterministic rule gaps

lopez-mar:

The deterministic App Doctor rules had two still-reproducible blind spots and one metadata mismatch:
- `REQUEST_CONTROLLED_ADMIN_CONTEXT` missed the canonical one-hop `readParams(request) -> params.shop -> unauthenticated.admin(...)` shape.
- `APP_PROXY_LIQUID_INJECTION` used a 500-character proximity check, so request-derived values in response headers could look like body flow and escaped body values could still be reported.
- `INSECURE_WEBHOOK_URL` already inspects OAuth redirect URLs, but its title/catalog/prompt still described only webhook URLs.

- Extends `REQUEST_CONTROLLED_ADMIN_CONTEXT` with one-hop helper-return tracking for helpers that visibly return a request-derived `shop` field.
- Reworks `APP_PROXY_LIQUID_INJECTION` to inspect explicit `new Response(...)` body expressions with same-call active content-type evidence instead of source-level proximity.
- Suppresses `text/html` findings only when the request-derived value is wrapped in an imported `escape-html` encoder.
- Aligns `INSECURE_WEBHOOK_URL` catalog/rule/prompt metadata with its existing behavior of checking both webhooks and OAuth redirect URLs.
- Adds regressions for helper-return taint, trusted session-shop negatives, header-only proxy values, same-file HTML/JSON separation, imported HTML escaping, and redirect-only config scanning.

`res.send/end/write(...)` support is intentionally not included here: the old implementation only identified those sinks through the same file-wide content-type heuristic this PR removes. A correct Express-style implementation needs response-state tracking for `res.type(...)`, `res.set(...)`, or `res.setHeader(...)` and should land separately.

### Verification

- `pnpm --filter @shopify/app generate:app-doctor-checks`
- `pnpm exec vitest run packages/app/src/cli/services/app-doctor-engine/tests/rule-analysis.test.ts packages/app/src/cli/services/app-doctor-engine/tests/scan-contract.test.ts packages/app/src/cli/services/app-doctor-engine/tests/checks.test.ts packages/app/src/cli/services/app-doctor-engine/tests/deterministic-rules.test.ts packages/app/src/cli/services/app-doctor-engine/tests/registry.test.ts`
- `pnpm nx run workspace:lint --skip-nx-cache`
- `pnpm nx run app:lint --skip-nx-cache`


jplhomer (commented):
Did we lose support for Node/Express-style `res.send/end` in App Proxy injection sinks? 

From my agent:

>The PR says the proxy rule inspects new Response(...), res.send/end/write(...), and literal return bodies, and that it accepts escapeHtml / escapeHTML / htmlEscape. The code only looks at new Response(...), and only trusts a default import from 'escape-html'. 

Otherwise LGTM!

lopez-mar:
Good catch. I intentionally narrowed this PR to explicit `new Response(...)` sinks with same-call active content-type evidence. The prior `res.send/end/write(...)` coverage depended on the same file-wide content-type heuristic this PR removes, so keeping it would have preserved the false-positive class instead of fixing it. Bare `return` bodies were only part of the old prefilter and were never an actual `directFlow` sink. I updated the PR body to match the implemented scope. Proper Express-style response-state tracking (`res.type`, `res.set`, `res.setHeader`) should be a separate follow-up.

jplhomer (approved):

packages/app/src/cli/services/app-doctor-engine/rules/proxy-rules.ts
```diff
@@ -43,8 +49,251 @@ export function scanAppProxyLiquidInjection(files: SourceFile[]): Issue[] {
...
+}
+
+function hasRequestControlledBodyFlow(
+  candidate: ResponseBodyCandidate,
+  requestBindings: Set<string>,
+  trustedHtmlEscapers: Set<string>,
+): boolean {
+  const body =
+    candidate.responseType === 'html'
+      ? maskTrustedHtmlEscapes(candidate.expression, trustedHtmlEscapers)
```
dmerand: Do not treat `escape-html` as safe in every HTML context.

`maskTrustedHtmlEscapes()` removes the full call before this check reads the output context. HTML escaping does not make a value safe in JavaScript or URL contexts. For example, this scan returns no finding:

```ts
const action = request.query.action
return new Response(`<button onclick="${escapeHtml(action)}">Run</button>`, {
  headers: {'Content-Type': 'text/html'},
})
```

`escape-html` leaves an input such as `alert(1)` executable. It also does not reject `javascript:` URLs. Suppress the flow only when the scanner proves a safe HTML context. Otherwise, keep the finding or mark the check unresolved.

packages/app/src/cli/services/app-doctor-engine/rules/proxy-rules.ts
```diff
@@ -43,8 +49,251 @@ export function scanAppProxyLiquidInjection(files: SourceFile[]): Issue[] {
...
+  return escapers
+}
+
+function hasLocalDefinition(source: string, name: string): boolean {
+  return new RegExp(`\\b(?:function|const|let|var)\\s+${escapeRegExp(name)}\\b`).test(source)
+}
+
+function responseBodyCandidates(source: string): ResponseBodyCandidate[] {
+  const candidates: ResponseBodyCandidate[] = []
+  const responsePattern = /\bnew\s+Response\s*\(/g
```
dmerand, line 109: Use token-aware boundaries when this rule finds and parses `Response` calls.

`responsePattern` reads source that still contains literal text. As a result, this code reports a finding:

```ts
const example = "new Response(request.query.shop, {headers: {'Content-Type': 'text/html'}})"
```

`callArguments()` also treats `)` in a regular expression as a call boundary. It treats a nested backtick as the end of the outer template. These valid response bodies return no finding:

```ts
new Response(request.query.shop.replace(/\)/g, ''), {headers: {'Content-Type': 'text/html'}})
new Response(`${request.query.shop ? `<div>${request.query.shop}</div>` : ''}`, {headers: {'Content-Type': 'text/html'}})
```

Use a JavaScript tokenizer or an offset-preserving token mask. Add tests for quoted code, regular expressions, and nested templates.

packages/app/src/cli/services/app-doctor-engine/rules/catalog.ts
```diff
@@ -73,13 +73,12 @@ export const RULE_CATALOG: RuleCatalogEntry[] = [
   },
   {
     id: 'INSECURE_WEBHOOK_URL',
-    title: 'Webhook URL is not HTTPS',
+    title: 'Configured callback URL is not HTTPS',
     severity: 'high',
     points: -12,
-    description: 'Detects webhook subscriptions configured with non-HTTPS absolute URLs.',
-    fix: 'Change the webhook URI to use HTTPS.',
+    description: 'Detects webhook subscriptions and OAuth redirect URLs configured with unsafe absolute URLs.',
```
dmerand, line 79: Bump the check versions when this PR changes their behavior.

The current definitions still emit these versions:

- `REQUEST_CONTROLLED_ADMIN_CONTEXT`: `2`
- `APP_PROXY_LIQUID_INJECTION`: `1`
- `INSECURE_WEBHOOK_URL`: `1`

`rule_version` goes into the trace and the submitted report. The old behavior and the new behavior now have the same provenance. Set these versions to `3`, `2`, and `2`. Update the matching Markdown frontmatter and regenerate `embedded.ts`. Let `configRule()` accept a version instead of always using `1`.

dmerand (approved):
This LGTM from a CLI command perspective. My agents found some questions that may be worth considering before merging.

PR #8500, opened 2026-09-08
Add static frame-ancestors check to App Doctor

lopez-mar:

`STATIC_FRAME_ANCESTORS` already exists as an App Doctor product ID in the agentic pack, but the deterministic engine had no narrow implementation for the literal cases it can prove safely.

- Adds a deterministic `STATIC_FRAME_ANCESTORS` runner for embedded admin apps only.
- Introduces a distinct `embedded_app` capability derived from root `embedded = true` in `shopify.app.toml`, separate from theme app-embed blocks.
- Flags only literal wildcard or clearly permissive `frame-ancestors` tokens inside `Content-Security-Policy` text.
- Keeps missing-CSP detection and nuanced policy construction in the agentic tier.
- Adds negative fixtures for fixed allowlists, computed policies, non-embedded apps, and theme-only app-embed surfaces.
- Updates JSON fixtures and a patch changeset for the new capability field.

### Verification

- `pnpm exec vitest run packages/app/src/cli/services/app-doctor-engine/tests/deterministic-rules.test.ts packages/app/src/cli/services/app-doctor-engine/tests/scan-contract.test.ts packages/app/src/cli/services/app-doctor-engine/tests/trace.test.ts packages/app/src/cli/services/app-doctor-json.test.ts packages/app/src/cli/services/doctor-output.test.ts packages/app/src/cli/services/doctor.test.ts`
- `pnpm --filter @shopify/app lint`

### DIFF
--- a/packages/app/src/cli/services/app-doctor-engine/rules/csp-rules.ts
+++ b/packages/app/src/cli/services/app-doctor-engine/rules/csp-rules.ts
@@ -4,7 +4,7 @@ import type {SourceFile} from './types.js'
 const JAVASCRIPT_EXTENSIONS = new Set(['.js', '.jsx', '.mjs', '.cjs', '.ts', '.tsx', '.mts', '.cts'])
 const CSP_FRAME_ANCESTORS_DIRECTIVE =
   /\bContent-Security-Policy\b[^\r\n]{0,300}?\bframe-ancestors\b([^;\r\n]*)/gi
-const CLEARLY_PERMISSIVE_SOURCE = /(^|\s)\*(?=\s|["'`;,}]|$)|https?:\/\/\*\.|\*\.myshopify\.com/i
+const CLEARLY_PERMISSIVE_SOURCE = /(^|\s)\*(?=\s|["'`;,}]|$)|https?:\/\/\*\.myshopify\.com/i
 
 export function scanStaticFrameAncestors(files: SourceFile[]): Issue[] {
   const issues: Issue[] = []
--- a/packages/app/src/cli/services/app-doctor-engine/rules/csp-rules.ts
+++ b/packages/app/src/cli/services/app-doctor-engine/rules/csp-rules.ts
@@ -2,8 +2,7 @@ import type {Issue} from '../types.js'
 import type {SourceFile} from './types.js'
 
 const JAVASCRIPT_EXTENSIONS = new Set(['.js', '.jsx', '.mjs', '.cjs', '.ts', '.tsx', '.mts', '.cts'])
-const CSP_FRAME_ANCESTORS_DIRECTIVE =
-  /\bContent-Security-Policy\b[^\r\n]{0,300}?\bframe-ancestors\b([^;\r\n]*)/gi
+const CSP_FRAME_ANCESTORS_DIRECTIVE = /\bContent-Security-Policy\b[^\r\n]{0,300}?\bframe-ancestors\b([^;\r\n]*)/gi
 const CLEARLY_PERMISSIVE_SOURCE = /(^|\s)\*(?=\s|["'`;,}]|$)|https?:\/\/\*\.myshopify\.com/i
 
 export function scanStaticFrameAncestors(files: SourceFile[]): Issue[] {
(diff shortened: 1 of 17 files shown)


packages/app/src/cli/services/app-doctor-engine/scanners/index.ts
```diff
@@ -358,7 +375,7 @@ function selectedFiles(definition: DeterministicCheckDefinition, context: ScanCo
   if (definition.target === 'manifest') return context.manifests.map((manifest) => manifest.path)
   if (definition.target === 'secrets')
     return context.sensitiveFiles.filter((file) => file.content !== undefined).map((file) => file.path)
-  let files = reactRouterFiles(context)
+  let files = definition.target === 'app_source' ? appSourceFiles(context) : reactRouterFiles(context)
```
jplhomer, line 393: `appSourceFiles` and `reactRouterFiles` are effectively the same function output - do they need to be split here?

packages/app/src/cli/services/app-doctor-engine/tests/deterministic-rules.test.ts
```diff
@@ -156,6 +158,96 @@ describe('JavaScript regex mode', () => {
   })
 })
 
+describe('STATIC_FRAME_ANCESTORS regex mode', () => {
```
jplhomer, line 167: Suggested test additions by my agent:

 - headers.set / res.setHeader / response.headers.append positive cases. The only setter test is a negative (headers.set(..., policy)). If HEADER_SETTER_PREFIX is wrong, CI will not notice.                                      
 - 'frame-ancestors ' + '*' — evaluateStaticStringExpression implements this, untested.                                   
 - backtick literals with no interpolation: `frame-ancestors *`                                                           
 - mixed sources: frame-ancestors 'self' *

jplhomer (commented):
Couple small questions!

lopez-mar:
Good questions. `appSourceFiles()` is intentionally the framework-agnostic app-source boundary for `STATIC_FRAME_ANCESTORS`; `reactRouterFiles()` currently delegates to it, but keeping the names separate makes the execution contract explicit because existing `source` checks remain React Router-gated while the new `app_source` target is not. I also added regression coverage for `headers.set`, `response.headers.append`, `res.setHeader`, static concatenation, static template literals, and mixed `frame-ancestors 'self' *` source lists.

jplhomer (approved):

packages/app/src/cli/services/app-doctor-engine/rules/csp-rules.ts
```diff
@@ -0,0 +1,257 @@
...
+}
+
+function hasClearlyPermissiveFrameAncestors(value: string): boolean {
+  return value.split(';').some((directive) => {
+    const [name, ...sources] = directive.trim().split(/\s+/)
+    return name?.toLowerCase() === 'frame-ancestors' && sources.some(isClearlyPermissiveSource)
+  })
+}
+
+function isClearlyPermissiveSource(source: string): boolean {
```
dmerand, line 85: Treat scheme-only sources as clearly permissive.

`frame-ancestors https:` is a valid CSP scheme source. It lets any HTTPS origin frame the app. This static policy returns no finding:

```ts
const headers = {'Content-Security-Policy': 'frame-ancestors https:'}
```

Shopify requires the exact authenticated shop origin and Shopify Admin origin. Recognize at least `http:` and `https:` as permissive sources. Add tests for a scheme-only source and for a mixed list such as `frame-ancestors 'self' https:`.

packages/app/src/cli/services/app-doctor-engine/rules/csp-rules.ts
```diff
@@ -0,0 +1,257 @@
...
+    if (source[index] === open) depth++
+    else if (source[index] === close) {
+      depth--
+      if (depth === 0) return index
+    }
+  }
+  return undefined
+}
+
+function parseStringLiteral(source: string, start: number): ParsedStringLiteral | undefined {
```
dmerand, line 199: Consume an unterminated string literal only once.

`parseStringLiteral()` scans to the end of the file and returns `undefined` when it cannot find a closing quote. `staticCspHeaderValues()` then advances one character. Each escaped quote can start another scan of the same remainder. This makes malformed input take quadratic time.

A focused reproduction took approximately 0.72 seconds for 64 KB and 2.84 seconds for 128 KB. App Doctor accepts source files up to 500 KB. Return an end-of-input, non-static token for an unterminated literal, or use a single-pass tokenizer. Add a malformed-input regression test.

dmerand (approved):
Again, approved from the overall lens of CLI command membership. My agents had a couple of notes about the implementation that may be worth considering.

PR #8511, opened 2026-09-09
Add workflow to create PRs from the Gardener backlog

gonzaloriestra:

Turn prioritized Gardener backlog issues into reviewable CLI PRs.

Adds a workflow for Claude to pick the highest-priority eligible CLI issue from the Gardener board, create a draft PR, and share its linked title in `#devtools-gardener-backlog`. Runs Monday–Friday at 00:00 UTC or manually.

### How to test

Merge and trigger the workflow

### Checklist

### DIFF
--- a/.github/workflows/gardener-backlog-prs.yml
+++ b/.github/workflows/gardener-backlog-prs.yml
@@ -1,0 +1,140 @@
+name: gardener-backlog-prs
+
+on:
+  workflow_dispatch:
+  schedule:
+    # Monday–Friday at 00:00 UTC
+    - cron: '0 0 * * 1-5'
+
+permissions:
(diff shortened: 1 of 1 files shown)


amcaplan (approved):
The flow makes sense. I suspect 5 PRs weekly is a bit too aggressive (we are already overwhelmed with Jules and I don't think anyone is looking at all of those), but given that it shouldn't post multiple PRs per issue, we can run with this and see how it goes.

gonzaloriestra:
Worst thing it can happen is that it creates PRs for all the tasks in the backlog and they remain there unreviewed for a while... But they are "free", so I think we can try, no?

Thanks for the reviews!

PR #8512, opened 2026-09-09
Fail the build when App Doctor embedded checks are stale

jplhomer:

Developers edit App Doctor checks and instructions in Markdown, then run a script that copies that text into `embedded.ts`. If they skip the script, the released CLI keeps serving the old content.

Fail the app Nx build when `embedded.ts` does not match the Markdown sources, so local builds and CI catch a missed regenerate.

```
pnpm nx run app:check-app-doctor-checks
```

That should pass on this branch. Then change any file under `packages/app/src/cli/services/app-doctor-engine/checks/*.md` or `INSTRUCTIONS.md` without regenerating, and run:

```
pnpm nx run app:build
```

The build should fail and tell you to run `pnpm --filter @shopify/app generate:app-doctor-checks`.

### Checklist

### DIFF
--- a/packages/app/src/cli/services/app-doctor-engine/embed-checks.mjs
+++ b/packages/app/src/cli/services/app-doctor-engine/embed-checks.mjs
@@ -3,6 +3,9 @@
 /**
  * Embeds the versioned semantic-check markdown so the bundled CLI does not
  * depend on runtime assets. The markdown files remain the source of truth.
+ *
+ *   node embed-checks.mjs           Write checks/embedded.ts
+ *   node embed-checks.mjs --check   Exit 1 if embedded.ts is stale
  */
 import {readFileSync, readdirSync, writeFileSync} from 'node:fs'
 import {dirname, join} from 'node:path'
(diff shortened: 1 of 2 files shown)


packages/app/src/cli/services/app-doctor-engine/embed-checks.mjs
```diff
@@ -29,7 +32,32 @@ export const EMBEDDED_CHECK_SOURCES: ReadonlyArray<string> = ${JSON.stringify(so
 export const EMBEDDED_APP_DOCTOR_INSTRUCTIONS = ${JSON.stringify(instructions)};
 `
 
-writeFileSync(outputPath, output)
-console.log(
-  `Embedded ${files.length} semantic checks and App Doctor instructions in app-doctor-engine/checks/embedded.ts`,
-)
+if (process.argv.includes('--check')) {
```
dmerand: Nit: Since this is a Node app, you may only want to check arguments before `--` if it exists. Anything after that would be a command against the Node binary itself. I dunno if it includes a `--check` but that would be a conflict.

packages/app/src/cli/services/app-doctor-engine/embed-checks.mjs
```diff
@@ -29,7 +32,32 @@ export const EMBEDDED_CHECK_SOURCES: ReadonlyArray<string> = ${JSON.stringify(so
...
+
+Markdown checks and instructions are the source of truth, but
+packages/app/src/cli/services/app-doctor-engine/checks/embedded.ts
+does not match what embed-checks.mjs would generate.
+
+Run:
+  pnpm --filter @shopify/app generate:app-doctor-checks
+
+Then commit the updated embedded.ts.`)
+    process.exit(1)
```
dmerand: Nit: you could set `process.exitCode = 1` here + leave Node to flush stderr.

packages/app/src/cli/services/app-doctor-engine/embed-checks.mjs
dmerand: Do we need a catch for a situation where somebody sends garbled args e.g. `--chek` which would currently act like `do not check`?

packages/app/src/cli/services/app-doctor-engine/embed-checks.mjs
```diff
@@ -29,7 +32,32 @@ export const EMBEDDED_CHECK_SOURCES: ReadonlyArray<string> = ${JSON.stringify(so
...
-writeFileSync(outputPath, output)
-console.log(
-  `Embedded ${files.length} semantic checks and App Doctor instructions in app-doctor-engine/checks/embedded.ts`,
-)
+if (process.argv.includes('--check')) {
+  let actual = ''
+  try {
+    actual = readFileSync(outputPath, 'utf8')
+  } catch {
+    // Missing generated file is a mismatch.
```
dmerand, line 51: Nit: This also catches file read errors as mismatches.

dmerand (approved):

packages/app/src/cli/services/app-doctor-engine/embed-checks.mjs
jplhomer: Done — now sets `process.exitCode = 1` and lets the process exit normally so stderr can flush.

packages/app/src/cli/services/app-doctor-engine/embed-checks.mjs
jplhomer: Done — unknown args (including typos like `--chek`) now fail with usage instead of generating.

packages/app/src/cli/services/app-doctor-engine/embed-checks.mjs
jplhomer: Done — only `ENOENT` is treated as a mismatch; other read errors are rethrown.

PR #8513, opened 2026-09-09
Keep App Doctor engine internals behind scan, compile, and submit operations

jplhomer:
### WHY are these changes introduced?

App Doctor CLI services imported scanners, check registries, merge helpers, trace validators, and redaction functions from the engine. Those internals made engine changes ripple into command-layer files.

### WHAT is this pull request doing?

Expose only the operations the CLI needs: locate an app, scan, parse and compile findings, parse a stored trace, build a submission, and load agent instructions. Move compile/merge/findings-parse into the engine. CLI output, artifacts, and submit now use that surface; engine tests still import internal modules directly.

Display redaction stays an engine contract. The report layer renders already-redacted scan results instead of calling engine redact helpers.

No user-facing behavior change.

### How to manually test your changes?

From an app directory:

```
shopify app doctor
shopify app doctor --findings .shopify/app-doctor/findings.json
shopify app doctor submit --dry-run
```

Scan, compile, and dry-run submit should behave as they do on main.

### Checklist

- [ ] I've considered possible cross-platform impacts (Mac, Linux, Windows)
- [ ] I've considered possible [documentation](https://shopify.dev) changes
- [ ] I've considered analytics changes to measure impact
- [ ] The change is user-facing — I've identified the correct [bump type](https://github.com/Shopify/cli/blob/main/CONTRIBUTING.md#choosing-the-right-bump-type) (`patch` for bug fixes · `minor` for new features · `major` for [breaking changes](https://github.com/Shopify/cli/blob/main/CONTRIBUTING.md#what-counts-as-a-breaking-change)) and added a changeset with `pnpm changeset add`

PR #8514, opened 2026-09-09
Reuse Shopify CLI app identity and config schemas in App Doctor

jplhomer:

App Doctor walked up for `shopify.app*.toml` and parsed access, auth, and webhooks with its own schemas and fallbacks. That can disagree with Shopify CLI as those rules change.

Stacked on #8513.

Share CLI app-config identity (`isValidFormatAppConfigurationFileName` / `APP_CONFIG_FILE_GLOB`) in Project discovery, config-file prompts, and App Doctor. Parse access, auth, and webhook subscriptions with the CLI schemas. Keep webhook URIs as strings so `INSECURE_WEBHOOK_URL` can still report http.

This is not `Project.load()`. Bounded reads, coverage gaps, no-cwd-fallback, and scanning unconfigured extensions stay Doctor-specific.

No user-facing CLI command change.

From an app directory:

```
shopify app doctor
shopify app doctor --path ./shopify.app.toml
```

A directory that only has `shopify.application.toml` should not be treated as an app root.

### Checklist

### DIFF
--- a/packages/app/src/cli/prompts/config.ts
+++ b/packages/app/src/cli/prompts/config.ts
@@ -1,5 +1,5 @@
 /* eslint-disable no-await-in-loop */
-import {AppConfigurationFileName} from '../models/app/loader.js'
+import {AppConfigurationFileName, isValidFormatAppConfigurationFileName} from '../models/app/loader.js'
 import {
   RenderTextPromptOptions,
   renderConfirmationPrompt,
@@ -43,7 +43,8 @@ function filenameFromName(name: string, highlight = false): AppConfigurationFile
 }
 
 export async function findConfigFiles(directory: string): Promise<string[]> {
-  return glob(joinPath(directory, 'shopify.app*.toml'))
+  const files = await glob(joinPath(directory, 'shopify.app*.toml'))
+  return files.filter((path) => isValidFormatAppConfigurationFileName(basename(path)))
 }
 
 export async function selectConfigFile(directory: string): Promise<Result<string, string>> {
--- a/packages/app/src/cli/prompts/config.ts
+++ b/packages/app/src/cli/prompts/config.ts
@@ -1,4 +1,5 @@
 /* eslint-disable no-await-in-loop */
+import {APP_CONFIG_FILE_GLOB} from '../models/app/config-file-naming.js'
 import {AppConfigurationFileName, isValidFormatAppConfigurationFileName} from '../models/app/loader.js'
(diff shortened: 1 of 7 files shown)


packages/app/src/cli/services/app-doctor-engine/scanners/discover.ts
```diff
@@ -207,22 +205,12 @@ function projectWebhooks(
...
-        topics: [...(strict.data.topics ?? []), ...(strict.data.compliance_topics ?? [])],
-        uri: strict.data.uri,
-      },
-    ]
-  }
-
-  const evidence = EvidenceWebhookSubscriptionSchema.safeParse(value)
-  if (!evidence.success) return []
+  const parsed = SecurityWebhookSubscriptionSchema.safeParse(value)
+  if (!parsed.success) return []
```
dmerand: My agent sez: Keep insecure URI evidence when another subscription field is invalid.

`SecurityWebhookSubscriptionSchema` validates the complete object. For example, this subscription has an insecure URI and an invalid `filter`:

```toml
[[webhooks.subscriptions]]
topics = ["orders/create"]
uri = "http://insecure.example/webhooks"
filter = 42
```

The schema rejects `filter`, so this function returns `[]`. It does not record a coverage gap. `INSECURE_WEBHOOK_URL` then receives no URI, and the scan can report complete coverage without this evidence.

Keep an independently valid URI for the security check and record the schema failure as a coverage gap. At minimum, do not discard the subscription without a coverage gap.

packages/app/src/cli/prompts/config.ts
```diff
@@ -43,7 +43,8 @@ function filenameFromName(name: string, highlight = false): AppConfigurationFile
 }
 
 export async function findConfigFiles(directory: string): Promise<string[]> {
-  return glob(joinPath(directory, 'shopify.app*.toml'))
+  const files = await glob(joinPath(directory, 'shopify.app*.toml'))
```
dmerand: Can we also use `APP_CONFIG_FILE_GLOB` for this call?

jplhomer: Done!

packages/app/src/cli/services/app-doctor-engine/scanners/discover.ts
jplhomer: Done!

dmerand (approved):

PR #8524, opened 2026-09-10
Add JSON output schema for app versions list

dmerand:

`shopify app versions list --json` is a public result but has no declared machine-readable contract. Users cannot inspect its type through the standard `--json-schema` interface.

Related to shop/issues-develop#23670

- Add and expose the `AppVersionsListResult` schema through `--json-schema`.
- Preserve the existing JSON array contract, including `versionId`, and text output.
- Move JSON/text selection to the command result boundary.
- Detect a missing app beside its API response and report a factual error.

Normal output remains a table:

```text
VERSION       STATUS    MESSAGE  DATE CREATED         CREATED BY
────────────  ────────  ───────  ───────────────────  ──────────
versionTag    ★ active  message  2021-01-01 00:00:00  createdBy
```

The matching JSON output is:

```json
[
  {
    "createdAt": "2021-01-01 00:00:00",
    "createdBy": "createdBy",
    "versionTag": "versionTag",
    "status": "active",
    "versionId": "gid://shopify/Version/1",
    "message": "message"
  }
]
```

```sh
pnpm shopify app versions list
pnpm shopify app versions list --json
pnpm shopify app versions list --json-schema
```

Confirm that `--json-schema` includes `AppVersionsListResult` and `AppVersion`. An empty JSON result is `[]`; the empty text result keeps its existing guidance on stderr.

### Checklist


packages/app/src/cli/commands/app/versions/list.ts
```diff
@@ -27,13 +33,19 @@ export default class VersionsList extends AppLinkedCommand {
...
 
-    await versionList({
-      app,
-      remoteApp,
-      organization,
-      developerPlatformClient,
-      json: flags.json,
-    })
+    const result = await getAppVersions(developerPlatformClient, remoteApp)
+    if (!result) throw new AbortError(`Invalid API Key: ${remoteApp.apiKey}`)
```
gonzaloriestra: Is this the only possible error when the result is empty? 🤔 

I'd keep this error closer to the implementation. There could be a new path where it returns undefined, and this error would be misleading.

packages/app/src/cli/services/versions-list.ts
```diff
@@ -1,128 +1,48 @@
...
-type AppVersionLine = {
-  createdAt: string
-  createdBy?: string
-  message?: string
-  versionTag?: string | null
-  status: string
+import {defineJsonOutputSchema, type InferJsonOutputSchema} from '@shopify/cli-kit/node/json-output-schema'
+import {zod} from '@shopify/cli-kit/node/schema'
+
+const appVersionJsonOutputSchema = zod.object({
```
gonzaloriestra: What about moving this to a `types.ts` file to be consistent with other PRs?

packages/app/src/cli/commands/app/versions/list.test.ts
```diff
@@ -0,0 +1,111 @@
+import {testAppLinked, testDeveloperPlatformClient, testOrganizationApp} from '../../../models/app/app.test-data.js'
+import {Organization, OrganizationSource} from '../../../models/organization.js'
+import {afterEach, describe, expect, test, vi} from 'vitest'
+
+vi.mock('../../../services/app-context.js')
+vi.mock('../../../services/versions-list/result.js')
```
gonzaloriestra: We could add populated and empty-result tests using the real presenter and writer. That would let us verify stdout and stderr separately.

packages/app/src/cli/commands/app/versions/list.ts
dmerand: Good point, I'll make it more precise.

dmerand: Fixed in `66edfe5c`. `getAppVersions` now handles the exact null-app response and throws `Shopify did not return app information for API key <apiKey>.` The ambiguous command guard is removed, and an empty versions array remains successful.

packages/app/src/cli/services/versions-list.ts
dmerand: Fixed in `66edfe5c`. `packages/app/src/cli/services/versions-list/types.ts` now owns the public schema and inferred result type. The service keeps only its internal execution result type.

packages/app/src/cli/commands/app/versions/list.test.ts
dmerand: Added in `66edfe5c`. The command tests now exercise the real presenter, encoder, and writer for populated JSON, empty JSON, and empty text, with exact stdout and stderr assertions.

gonzaloriestra (approved):

packages/app/src/cli/services/versions-list.ts
```diff
@@ -1,128 +1,33 @@
...
-    // Update the message column to fit the max length
-    appVersions.forEach((appVersion) => {
-      if (appVersion.message.length > maxMessageLength) {
-        appVersion.message = `${appVersion.message.slice(0, maxMessageLength - 3)}...`
-      }
-    })
+): Promise<AppVersionsList> {
+  const response: AppVersionsQuerySchema = await developerPlatformClient.appVersions(app)
+  if (!response.app) {
+    throw new AbortError(`Shopify did not return app information for API key ${app.apiKey}.`)
```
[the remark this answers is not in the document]
dmerand: Fixed in `6b0046f`. `AppManagementClient.appVersions` now returns `{app: null}` before dereferencing the app, so `getAppVersions` reaches the factual `AbortError`. A client-boundary regression test covers the production path.

PR #8525, opened 2026-09-11
Add JSON schema for theme info

dmerand:

`theme info --json` has two established result shapes, but neither is discoverable as a command contract.

Related to shop/issues-develop#23691

- Add a schema for the existing theme and environment JSON results.
- Preserve both JSON shapes, text output, and per-environment emission.
- Move JSON/text selection to the command result boundary.
- Preserve analytics timing: JSON records the opening timing; text records opening and closing.

Normal output remains human-readable sections:

```text
Theme Configuration
Store                 my-shop.myshopify.com
Development Theme ID  Not set

Tooling and System
Shopify CLI  3.91.0
OS           darwin-arm64
Shell        /bin/zsh
Node version v24.15.0
```

The matching JSON output remains:

```json
{
  "store": "my-shop.myshopify.com",
  "development_theme_id": null,
  "cli_version": "3.91.0",
  "os": "darwin-arm64",
  "shell": "/bin/zsh",
  "node_version": "v24.15.0"
}
```

A multi-environment invocation continues to emit one existing JSON document per environment. This PR does not add a wrapper or change that public behavior.

```sh
shopify theme info
shopify theme info --json
shopify theme info --json-schema
shopify theme info --environment development --environment staging --json
```

### Checklist


packages/theme/src/cli/commands/theme/info.ts
```diff
@@ -42,19 +49,10 @@ export default class Info extends ThemeCommand {
         throw new AbortError('Theme not found!')
       }
 
-      if (flags.json) {
-        return outputResult(JSON.stringify(output, null, 2))
-      }
-
-      const formattedInfo = await formatThemeInfo(output, flags)
-      renderInfo(formattedInfo)
+      renderThemeInfoResult(output, flags.json ? 'json' : 'text', flags)
```
gonzaloriestra, line 52: This command allows multiple environments and it's printing multiple JSON objects in that case. 

Could we collect the results across environments and write a single JSON object once all environments finish, with each result identified by its environment? We’d also need to update the schema and add a multi-environment test, while preserving the existing single-environment output shape.

packages/theme/src/cli/commands/theme/info.ts
```diff
@@ -42,19 +49,10 @@ export default class Info extends ThemeCommand {
...
-      if (flags.json) {
-        return outputResult(JSON.stringify(themeEnvironmentInfoJSON({cliVersion: this.config.version}), null, 2))
-      }
-      const infoMessage = await fetchDevInfo({cliVersion: this.config.version})
-
-      renderInfo({customSections: infoMessage})
+      const {result, developmentTheme} = getThemeEnvironmentInfo({cliVersion: this.config.version})
+      renderThemeInfoResult(result, flags.json ? 'json' : 'text', {developmentTheme})
     }
     recordTiming('theme-command:info')
```
gonzaloriestra: This now runs in JSON mode. I guess it's intentional, and it makes sense. Just noting that the analytics behavior is not preserved in this case.

packages/theme/src/cli/commands/theme/info.ts
dmerand: Implemented in `85d7d6b`. Multi-environment `theme info --json` now writes one ordered `{environments:[...]}` document after all environments finish. Entries identify their environment, include successful results only, and preserve the existing single-environment shapes. Failures keep their environment-prefixed stderr output and are omitted; all-failed output is `{"environments":[]}` with existing exit behavior. The wrapper and entry are included in `themeInfoJsonOutputSchema` and covered by multi-environment tests.

packages/theme/src/cli/commands/theme/info.ts
dmerand: Fixed in `85d7d6b`. JSON mode again returns after the result path and records only the opening `theme-command:info` timing. Text mode keeps the opening and closing timings. Focused tests assert one call for JSON and two for text.

packages/theme/src/cli/commands/theme/info.ts
gonzaloriestra: I just realized that the json flag was already present here... So this would be a breaking change, no? Maybe it's better to keep the current shape of existing JSON outputs, as we had decided, and let this change for another moment?

Sorry for the back and forth! 🙏

dmerand: Ah yes, good point! Yes, we should maintain the schema as it currently exists, and perhaps just flag in our tracker that there may be some follow-up. I'll make that change back.

packages/theme/src/cli/commands/theme/info.ts
```diff
@@ -1,20 +1,27 @@
...
 export default class Info extends ThemeCommand {
-  static description =
-    'Displays information about your theme environment, including your current store. Can also retrieve information about a specific theme.'
+  static get jsonOutputSchema() {
+    return themeInfoJsonOutputSchema
+  }
+
+  static descriptionWithMarkdown = `Displays information about your theme environment, including your current store. Can also retrieve information about a specific theme.
+
+Use \`--json\` for machine-readable output.`
```
gonzaloriestra: I think we don't need this with the help and new instructions from the skill, I'll remove it.

PR #8528, opened 2026-09-11
Fix theme pull, push, and package to cover all theme files

karreiro:

The checksums API returns config/styles.css and AGENTS.md, but Shopify CLI does not handle them.

Recognize config/styles.css and AGENTS.md during theme synchronization, and include both files in theme packages.

Run the following commands against a theme containing both files:

```
shopify theme pull
shopify theme push
shopify theme package
```

Notice that config/styles.css and AGENTS.md are pulled, pushed, and included in the generated package.

### Checklist

### DIFF
--- a/packages/theme/src/cli/services/package.test.ts
+++ b/packages/theme/src/cli/services/package.test.ts
@@ -15,9 +15,11 @@ describe('packageTheme', () => {
       const inputDirectory = joinPath(tmpDir, 'theme')
       await mkdir(inputDirectory)
       const themeRelativePaths = [
+        'AGENTS.md',
         'assets/base.css',
         'layout/theme.liquid',
         'config/settings_schema.json',
+        'config/styles.css',
         'listings/canine-gourmand/sections/header-group.json',
         'listings/dawn/templates/index.json',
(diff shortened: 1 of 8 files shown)


packages/theme/src/cli/services/package.test.ts
```diff
@@ -15,9 +15,11 @@ describe('packageTheme', () => {
       const inputDirectory = joinPath(tmpDir, 'theme')
       await mkdir(inputDirectory)
       const themeRelativePaths = [
+        'AGENTS.md',
```
EvilGenius13, line 18: When we run `theme init` we have the option between 3 different LLM files (or all of them). Should we be including those as well?

Image

`AGENTS.MD`, `CLAUDE.md`, `copilot-instructions.md`.

karreiro: Currently, the platform won't accept provider specific files, so we're leaning only on agents.md for now

EvilGenius13 (approved):

karreiro:
(the changelog was applied in the patch counter part https://github.com/Shopify/cli/pull/8529)

PR #8530, opened 2026-09-11
Fix `theme dev` ignoring `SHOPIFY_HTTP(S)_PROXY`

### LINKED ISSUE #5890: Support SHOPIFY_HTTP(S)_PROXY usage in theme commands

- [x] Searched [existing issues](.) to see if your issue is a duplicate. (If you’ve found a duplicate issue, feel free to add additional information in a comment on it.)
- [x] Reproduced the issue in [the latest CLI version](https://www.npmjs.com/package/@shopify/cli).

Theme

### Expected behavior

The CLI should successfully render the storefront through the configured HTTP proxy.

### Actual behavior

The CLI fails to render the storefront with a 502 Bad Gateway error, despite proper proxy configuration.

### Verbose output

<details>
  <summary>Verbose output</summary>

  ```
Failed to render storefront with status 502 (Bad Gateway).
URL: [REDACTED]

TypeError: fetch failed
    at node:internal/deps/undici/undici:13510:13
    at process.processTicksAndRejections (node:internal/process/task_queues:105:5)
    at async render2 [...]
    at async Object.handler [...]
    at async Server.<anonymous> [...]
  ```

</details>

### Reproduction steps

1.Set HTTP proxy environment variables:
```powershell
$env:SHOPIFY_HTTP_PROXY = "http://127.0.0.1:PORT"
$env:SHOPIFY_HTTPS_PROXY = "http://127.0.0.1:PORT"
2.Run shopify theme dev -e development
3.CLI starts successfully and provides preview URLs
4.Immediately encounters 502 Bad Gateway error

### Operating System

Windows 11

### Shopify CLI version (`shopify --version`)

@shopify/cli/0.0.0-snapshot-20250313171017

### Shell

powershell

v22.15.1

_No response_

ujjawal-yadav:

Fixes #5890.

`shopify theme dev` ignores `SHOPIFY_HTTP(S)_PROXY`, so behind a corporate proxy every page render fails with the 502 / `TypeError: fetch failed` from the issue.

`bootstrap.ts` sets the proxy up with `global-agent`, which patches Node's `http`/`https` agents. The storefront path never reaches them: both branches of `render()` in `storefront-renderer.ts` and `proxyStorefrontRequest()` in `proxy.ts` call the built-in `fetch`, which is undici and ignores those agents. Each carries an `eslint-disable-next-line no-restricted-globals`, and the rule's own message is the diagnosis: "Built-in fetch does not support HTTP proxies." Confirmed against a local proxy: the built-in `fetch` goes direct, `node-fetch` goes through it.

Routes those three call sites through `fetch` from `@shopify/cli-kit/node/http`, as `storefront-session.ts` in the same directory already does, and drops the suppressions.

Two details are why this isn't a one-line swap:

- **H3 rejects a `node-fetch` `Response`.** Both functions are returned from H3 handlers, and returning the client's response as-is crashes the dev server (`ERR_INVALID_ARG_TYPE ... Received an instance of Response`). Both already re-wrapped with `new Response(response.body, response)`, so a `toWebResponse()` helper converts at that same spot. `render()`'s signature and its consumers are unchanged.
- **`node-fetch` always exposes a body stream, even on `304`.** The built-in fetch reports `null` there, and `new Response(stream, {status: 304})` throws, so every revalidated asset would answer 500. `toWebResponse()` also reads headers through `raw()`, since iterating `node-fetch` headers joins repeated `set-cookie` values and would merge the storefront's session cookies.

Requests use `slow-request` to keep the built-in fetch's semantics here: no automatic cancellation, so a slow theme still renders, and no retries, since this path carries cart and checkout writes. The request body goes through `Readable.fromWeb()`; a 4 MB POST arrives intact. Proxied assets now also pass through cli-kit's `runWithTimer`, which retains two `performance.measure` entries per request, as every other cli-kit request already does.

`getProxyStorefrontHeaders()` no longer forwards the browser's `accept-encoding`, so the storefront only picks a coding the client decodes. `content-encoding` is dropped only when the client actually decompressed the body, so anything it passes through, such as a chained `br, gzip`, still reaches the browser labelled correctly.

Not touched: `notifier.ts`, since `--notify` targets a user-chosen and usually local URL, and `parseThemeFileContent()` in `cli-kit/node/themes/api.ts`, which has the same bug for `theme pull` downloads but sits in another package. Happy to follow up there.

This also sits alongside #8287 rather than against it: the change routes theme through cli-kit's client instead of around it, which holds whichever implementation cli-kit uses. `toWebResponse()` gets simpler there, since h3 accepts an undici `Response` directly.

1. Start a local HTTP proxy, for example `mitmdump -p 8080`.
2. `SHOPIFY_HTTP_PROXY=http://127.0.0.1:8080 SHOPIFY_HTTPS_PROXY=http://127.0.0.1:8080 shopify theme dev --store <store>`
3. Open the preview URL. On `main` it fails with 502. With this change the render and asset requests appear in the proxy log and the page loads. Reload once more to exercise the `304` path.

### Checklist

### DIFF
--- a/.changeset/theme-dev-http-proxy.md
+++ b/.changeset/theme-dev-http-proxy.md
@@ -1,0 +1,5 @@
+---
+'@shopify/theme': patch
+---
+
+Respect `SHOPIFY_HTTP(S)_PROXY` in `theme dev` when rendering and proxying the storefront
--- a/packages/theme/src/cli/utilities/theme-environment/proxy.test.ts
+++ b/packages/theme/src/cli/utilities/theme-environment/proxy.test.ts
@@ -11,13 +11,21 @@ import {
   standardEventsRuntimeDevUrl,
   standardEventsRuntimeUrl,
 } from './standard-events.js'
-import {describe, test, expect, vi, beforeEach, afterEach} from 'vitest'
+import {describe, test, expect, vi, beforeEach, afterEach, type Mock} from 'vitest'
 import {createEvent} from 'h3'
+import {fetch, Response as HttpResponse} from '@shopify/cli-kit/node/http'
 import {IncomingMessage, ServerResponse} from 'node:http'
 
 import {Socket} from 'node:net'
+import {Readable} from 'stream'
+import {text} from 'stream/consumers'
 
 import type {DevServerContext} from './types.js'
+
+vi.mock('@shopify/cli-kit/node/http', async (importOriginal) => ({
+  ...(await importOriginal<typeof import('@shopify/cli-kit/node/http')>()),
+  fetch: vi.fn(),
+}))
 
 function createH3Event(method = 'GET', path = '/', headers = {}) {
   const req = new IncomingMessage(new Socket())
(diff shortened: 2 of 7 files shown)


ujjawal-yadav:
"I have signed the CLA!"

PR #8541, opened 2026-09-14
Add JSON Schema for subscription migration cancellation

dmerand:
# Add JSON Schema for subscription migration cancellation

## What

Add a typed JSON Schema contract for `app subscription-migrations cancel`.

## Why

Allow scripts to discover and validate the existing cancellation result without changing its JSON or terminal behavior.

Related to shop/issues-develop#23886


.changeset/clever-cancellations.md
gonzaloriestra, line 1: The command is still hidden, so we should remove this

packages/app/src/cli/services/subscription-migrations/cancel-operations.ts
```diff
@@ -1,18 +1,68 @@
...
-    }
-
-export interface MigrationCancellationResult {
-  outcomes: MigrationCancellationOutcome[]
-}
+import {cancelMigrationOperation} from './partners-api.js'
+import {defineJsonOutputSchema, type InferJsonOutputSchema} from '@shopify/cli-kit/node/json-output-schema'
+import {zod} from '@shopify/cli-kit/node/schema'
+
+const MigrationOperationResultNodeSchema = zod.object({
```
gonzaloriestra: Can we move this to a `types.ts` for consistency?

.changeset/clever-cancellations.md
dmerand: Removed `.changeset/clever-cancellations.md` in `af133cd` because the command remains hidden.

packages/app/src/cli/services/subscription-migrations/cancel-operations.ts
dmerand: Moved the public schemas and inferred cancellation result types to `services/subscription-migrations/types.ts` in `af133cd`. `cancel-operations.ts` now keeps only execution and protocol-error logic.

PR #8542, opened 2026-09-14
Add JSON output to version command

dmerand:
## What

Add typed JSON object output to `shopify version` while preserving raw version text in normal mode.

Normal output:

```text
$ shopify version
4.8.0
```

JSON output:

```json
$ shopify version --json
{
  "version": "4.8.0"
}
```

## Why

Let scripts read the installed Shopify CLI version without parsing text output.

## How to manually test your changes?

Run `shopify version`, `shopify version --json`, and `shopify version --json-schema`. Confirm that the schema requires an object with a string `version` property.

Related to shop/issues-develop#23675


packages/cli/src/cli/services/commands/version/types.ts
```diff
@@ -1,6 +1,14 @@
 import {CLI_KIT_VERSION} from '@shopify/cli-kit/common/version'
-import {outputResult} from '@shopify/cli-kit/node/output'
+import {defineJsonOutputSchema, type InferJsonOutputSchema} from '@shopify/cli-kit/node/json-output-schema'
+import {zod} from '@shopify/cli-kit/node/schema'
 
-export async function versionService(): Promise<void> {
-  outputResult(CLI_KIT_VERSION)
+export const versionJsonOutputSchema = defineJsonOutputSchema({
```
gonzaloriestra, line 4: Can we move this to a `types.ts` file for consistency?

packages/cli/src/cli/services/commands/version.ts
```diff
@@ -1,6 +1,14 @@
 import {CLI_KIT_VERSION} from '@shopify/cli-kit/common/version'
-import {outputResult} from '@shopify/cli-kit/node/output'
+import {defineJsonOutputSchema, type InferJsonOutputSchema} from '@shopify/cli-kit/node/json-output-schema'
+import {zod} from '@shopify/cli-kit/node/schema'
 
-export async function versionService(): Promise<void> {
-  outputResult(CLI_KIT_VERSION)
+export const versionJsonOutputSchema = defineJsonOutputSchema({
+  name: 'VersionResult',
+  schema: zod.string(),
```
gonzaloriestra: I think this should be an object. See [this comment](https://github.com/shop/issues-develop/issues/23675#issuecomment-5679968242).

packages/cli/src/cli/services/commands/version/types.ts
dmerand: I had wondered about this because this is such a small change, but I agree that for consistency we can just standardize on `types.ts`.

dmerand: Fixed in `6d8cd8a3`. The schema and inferred type now live in `version/types.ts`, with data-only execution in `index.ts` and presentation in `result.ts`.

packages/cli/src/cli/services/commands/version.ts
dmerand: Fixed in `6d8cd8a3`. The result is now `{version: string}`. Normal mode remains raw text, and `--json` emits one object with the required string `version` property.

gonzaloriestra (approved):

PR #8544, opened 2026-09-14
[4.8] Fix theme pull, push, and package to cover all theme files

karreiro:
Back-port https://github.com/Shopify/cli/pull/8543


EvilGenius13 (approved):

karreiro:
The "Check graphql-codegen has been run" check was picking newer schema values against the older stable/4.8 base

PR #8183, opened 2026-07-28
Make store auth session reuse opt-in per theme command

alfonso-noriega:
### WHY are these changes introduced?

Regression from #7783 (`4.4.0`), reported by Enterprise Support ([Slack thread](https://shopify.slack.com/archives/C07UJ7UNMTK/p1783604773690349)) and in the [dev community forum](https://community.shopify.dev/t/theme-dev-silently-adopts-a-stored-store-auth-session-without-checking-theme-scopes-breaking-theme-commands-for-oauth-users/36141).

Store-auth session reuse was added to the shared `ThemeCommand.createSession`, so every theme command with a `password` flag inherited it — while #7783 only intended it for `theme pull` and `theme push`. The base `storeAuthScopes()` returned `[]`, and `[].every(...)` is always `true`, so commands like `theme dev` adopted **any** cached `store auth` session, including ones without theme scopes. The resulting `401 Invalid API key or access token` / `missing read_themes access scope` errors were terminal, and `shopify auth logout` couldn't fix them (store-auth sessions live in a separate `shopify-cli-store` storage bucket).

### WHAT is this pull request doing?

- Changes the `storeAuthScopes()` contract to `string[] | undefined`: returning `undefined` (the default) opts the command **out** of store-auth session reuse entirely; commands opt in by declaring the scopes they require.
- Only `theme pull` (`read_themes`) and `theme push` (`read_themes`, `write_themes`) opt in — matching #7783's stated intent. `theme dev` and all other theme commands authenticate normally again.
- Threads the required scopes explicitly through the session-lookup helpers instead of re-reading them via `this`.
- Adds tests pinning that unscoped commands never read the store-auth cache (single- and multi-environment), and reworks existing tests to use a scoped test command where adoption is the behavior under test.

### How to test your changes?

1. `shopify store auth --store <store> --scopes read_products` (a session without theme scopes)
2. `shopify theme dev --store <store>` — before: fails with `missing read_themes access scope` or 401; after: authenticates through the normal theme flow.
3. `shopify theme pull --store <store>` after `shopify store auth --store <store> --scopes read_themes,write_themes` — still reuses the stored session (no login prompt).

### Measuring impact

- [x] Existing analytics will cater for this addition

PR #8546, opened 2026-09-14
Reuse preview store sessions in every theme command

dmerand:

A preview store from `shopify store create preview` stores an Admin API token in the store-auth cache. A preview store cannot complete a device login and has no password, so that token is its only credential.

#8183 made store auth session reuse opt-in per theme command, to stop commands from adopting standard store auth sessions with insufficient scopes. Its opt-out default also stopped every theme command except `theme pull` and `theme push` from reusing the cached preview token. Commands such as `theme list` fell through to device authentication, which the user cannot complete. The command is then a dead end.

Regression from 4.6.0. Reported in [Slack](https://shopify.slack.com/archives/C0B242KDQPN/p1789411766995049).

- Reuse a stored `kind: 'preview'` session in every theme command, whatever `storeAuthScopes()` returns.
- Keep the declared-scope requirement for standard sessions, so #8183's fix still holds.
- Make multi-environment configs work on preview stores. `validateEnvironments` dropped them for the same reason.

`theme pull` and `theme push` keep their declared scopes, so they also reuse standard sessions.

Seed a session with `kind: 'preview'` and `scopes: []`, then run a command that does not opt in to store auth reuse:

```sh
shopify theme list --store <preview-store>.myshopify.com --json --verbose
```

Result with this branch:

```text
Using stored store auth session for <preview-store>.myshopify.com (scopes: )
```

Result on `main`: `Ensuring that the user is authenticated with the Theme API` followed by `You are not authorized to use the CLI to develop in the provided store`.

`theme pull` and `theme push` against a preview store still reuse the stored session.

### Checklist

- [ ] The change is user-facing — I've identified the correct bump type (`patch`) and added a changeset

Follow-up: #8547 adds the skipped-session reason to authentication errors.


tizmagik (commented):
Went through this against the history. The preview-session bypass is the right shape, and better than opting each command in one at a time, which is what I'd suggested in the Slack thread. #7783 says the store-auth cache was built for preview stores in the first place, and #8183's gate was aimed at standard sessions. Splitting the two paths restores that intent, and it won't drift when someone adds a new theme command.

Checked the back-compat case too: `kind: 'preview'` has been written since 4.3.0, the first release with `store create preview`, so no stored session reads back as `standard` by accident.

Two inline notes below. Neither blocks the fix landing if you'd rather file them as follow-ups.

packages/theme/src/cli/utilities/theme-command.ts
```diff
@@ -430,13 +429,23 @@ export default abstract class ThemeCommand extends Command {
...
     }
 
-    if (!this.hasRequiredStoreAuthScopes(storedSession.scopes, requiredScopes)) {
-      outputDebug(
-        `Ignoring stored store auth session for ${storeFqdn}: it is missing required scopes (has: ${storedSession.scopes.join(
-          ', ',
-        )}; needs: ${requiredScopes.join(', ')}).`,
-      )
-      return undefined
+    const isPreviewSession = storedSession.kind === 'preview'
```
tizmagik, line 432: Worth thinking about what this does to a *stale* preview session, the one from a store that's since been claimed. Before this PR it broke `theme pull` and `theme push`. After it, every theme command.

There's no way for a user to clear it. `shopify auth logout` is just `sessionStore.remove()` (`packages/cli-kit/src/public/node/session.ts:327`) and never touches the `shopify-cli-store` storage, and there's no `store auth logout`. The only clear that exists today is the side effect in `shopify store info` (`packages/store/src/cli/services/store/info/index.ts:166`), which fires on a 401/404 from the preview-store API. Nobody is going to find that.

That's the failure mode #8183's commit message called out, 401s that can't be fixed with `auth logout`. #8390 is one answer, but the concern raised in the thread about clear-on-401 being unrecoverable for an unclaimed store still stands. A `shopify store auth logout --store <store>` would cover it without that risk.

packages/theme/src/cli/utilities/theme-command.ts
```diff
@@ -430,13 +429,23 @@ export default abstract class ThemeCommand extends Command {
...
+    if (!isPreviewSession) {
+      if (!requiredScopes) {
+        outputDebug(
+          `Ignoring stored store auth session for ${storeFqdn}: it is a standard session and this command only reuses preview store sessions.`,
+        )
+        return undefined
+      }
+
+      if (!this.hasRequiredStoreAuthScopes(storedSession.scopes, requiredScopes)) {
```
tizmagik, line 441: This gate is now standard-sessions-only, which turns the first line of `hasRequiredStoreAuthScopes` into a hole:

```ts
if (sessionScopes.length === 0) return true
```

That came from #7783, to let empty-scope preview sessions through. Preview sessions have their own bypass now, so all it still does is let a standard session with no recorded scopes satisfy `theme pull` and `theme push`, which is what #8183 set out to close. Looks like a one-line delete.

packages/theme/src/cli/utilities/theme-command.ts
dmerand: Deferring this from #8546. #8390 already clears the stored preview session on a 401 and points to `store auth`, so it covers the main path. Two notes: (1) #8390 is still a draft, so landing order matters — if #8546 lands first there is a window with no manual clear; (2) the residual concern stands: clear-on-401 is unrecoverable for an unclaimed store, and `store auth logout --store` would cover it without that risk. Keeping both as follow-ups.

packages/theme/src/cli/utilities/theme-command.ts
dmerand: Checked this against current `main`: `if (sessionScopes.length === 0) return true` is already gone. It was removed by `ed332dd4ef` (2026-07-28) and is absent from `main`, #8546, and #8547. No hole to close here — flagging in case the note was based on an older revision.

tizmagik: You're right, and I was wrong. I pulled that line out of the `ba47dd464c` diff without checking whether anything later removed it, and `ed332dd4ef` removed it the same day. It's not on `main`, `#8546`, or `#8547`. Nothing to close. Sorry for the noise.

gonzaloriestra (approved):
Works as expected 👌

PR #8547, opened 2026-09-14
Name the skipped store auth session in theme auth errors

dmerand:

Stacked on #8546.

When a theme command skips a stored store auth session and falls back to device authentication, it reports the skip reason only through `outputDebug`. Without `--verbose`, the user cannot see that a session sat in the cache. A device login cannot replace a preview store session, so the silent fall-through costs the user all recovery options.

- Record why each stored session was skipped: expired, standard session, or missing scopes.
- Attach the reason and a recovery step to the next steps of the authentication error.
- Pass errors through unchanged when no session was skipped, or when the error is not an `AbortError`.

The recovery step matches the reason. A standard session points to `--password`. Missing scopes point to `shopify store auth --scopes`.

Seed a standard session with `scopes: ['read_products']`, which `theme pull` rejects, then run the command.

```sh
CI=true shopify theme pull --store <store>.myshopify.com --development --verbose
```

Result with this branch:

```text
Authorization is required to continue, but the current environment does not support interactive prompts.
...
Next steps
  • The CLI found a stored store auth session for <store>.myshopify.com, but did not use it: it is missing required scopes (has: read_products; needs: read_themes).
  • Run `shopify store auth --store <store>.myshopify.com --scopes read_themes` to grant the required scopes.
```

Result on `main`: the same error without the two next steps.

### Checklist

- [ ] The change is user-facing — I've identified the correct bump type (`patch`) and added a changeset


tizmagik (commented):
Direction is right. Four inline notes below, on the advice strings and the plumbing.

packages/theme/src/cli/utilities/theme-command.ts
```diff
@@ -56,9 +56,16 @@ type EnvironmentName = string
...
 
+interface SkippedStoreAuthSession {
+  reason: string
+  advice: string
+}
+
 export default abstract class ThemeCommand extends Command {
   static baseFlags = authAliasFlag
 
+  private readonly skippedStoreAuthSessions = new Map<string, SkippedStoreAuthSession>()
```
tizmagik: This is mutable instance state used as a side channel: a private helper writes it, `createSession` reads it back by store key. AGENTS.md asks for pure functions without side effects.

Returning `{session}` or `{skipped}` from `adminSessionFromStoreAuthSession` carries the same information through the call graph, where a reader can follow it without knowing the map exists.

packages/theme/src/cli/utilities/theme-command.ts
```diff
@@ -362,13 +369,23 @@ export default abstract class ThemeCommand extends Command {
...
+    const session = storeAuthSession ?? (await this.storeAuthSessionForTheme({store}))
+    if (session) return session
+
+    try {
+      return await ensureAuthenticatedThemes(store, password)
+    } catch (error) {
+      const skipped = this.skippedStoreAuthSessions.get(normalizeStoreFqdn(store))
+      if (!(error instanceof AbortError) || !skipped) throw error
+
+      throw new AbortError(error.message, error.tryMessage, [
```
tizmagik: This flattens an `AbortError` subclass down to a plain `AbortError` and drops any extra fields it carried. Appending to `error.nextSteps` and re-throwing the original keeps identity.

packages/theme/src/cli/utilities/theme-command.ts
```diff
@@ -423,27 +440,37 @@ export default abstract class ThemeCommand extends Command {
...
   ): AdminSession | undefined {
     if (isSessionExpired(storedSession)) {
-      outputDebug(
-        `Ignoring stored store auth session for ${storeFqdn}: it expired at ${storedSession.expiresAt ?? 'unknown'}.`,
-      )
+      const reason = `it expired at ${storedSession.expiresAt ?? 'an unknown time'}.`
+      outputDebug(`Ignoring stored store auth session for ${storeFqdn}: ${reason}`)
+      this.skippedStoreAuthSessions.set(storeFqdn, {
+        reason,
+        advice: `Run \`shopify store auth --store ${storeFqdn}\` to store a fresh session.`,
```
tizmagik: This advice can't work on a preview store. `store auth` needs an account to log in as, and `recovery.ts` says it outright: preview stores aren't a logged-in experience, so scopes can't be granted later.

The expiry check sits above the `isPreviewSession` branch, so a preview session does reach it. Unreachable today because preview sessions carry no `expiresAt`, but it'll be wrong the moment one does.

packages/theme/src/cli/utilities/theme-command.ts
```diff
@@ -423,27 +440,37 @@ export default abstract class ThemeCommand extends Command {
...
       if (!requiredScopes) {
-        outputDebug(
-          `Ignoring stored store auth session for ${storeFqdn}: it is a standard session and this command only reuses preview store sessions.`,
-        )
+        const reason = 'it is a standard session and this command only reuses preview store sessions.'
+        outputDebug(`Ignoring stored store auth session for ${storeFqdn}: ${reason}`)
+        this.skippedStoreAuthSessions.set(storeFqdn, {
+          reason,
+          advice:
+            'Pass a Theme Access password with `--password`, or run `theme pull` or `theme push`, which reuse standard store auth sessions.',
```
tizmagik: Someone who ran `theme list` can't act on "run `theme pull` or `theme push`". Those are different commands, not a recovery step.

It also puts the asymmetry from #8546 into a user-facing string: `theme list` accepts a preview session with zero recorded scopes, but turns down a standard session that carries `read_themes`. Two ways out. Declare `storeAuthScopes()` on the read-only commands so the standard path works and this advice stops being needed, or trim the string down to the `--password` half.

packages/theme/src/cli/utilities/theme-command.ts
dmerand: Done. Removed the map and returned a `StoreAuthSessionResult` (`{session?, skipped?}`) from `adminSessionFromStoreAuthSession` / `storeAuthSessionForTheme`; `createSession` reads the skip reason from the returned value. Also threaded the result through the multi-env pre-pass, so a skipped store is not re-read. `c51b565c5e`.

packages/theme/src/cli/utilities/theme-command.ts
dmerand: Done. `createSession` now appends to `error.nextSteps` and re-throws the original error, so the `AbortError` subclass, `tryMessage`, and `customSections` survive. Added a test asserting identity. `c51b565c5e`.

packages/theme/src/cli/utilities/theme-command.ts
dmerand: Done. `isPreviewSession` is computed before the expiry check and an expired preview session carries no `store auth` advice. `throwIfPreviewStore` (`packages/store/src/cli/services/store/auth/index.ts`) refuses `store auth` while a preview session is present, so the advice could not work. Added a test. Preview sessions carry no `expiresAt` today, so this is a forward-looking guard. `c51b565c5e`.

packages/theme/src/cli/utilities/theme-command.ts
dmerand: Done — trimmed the standard-session advice to `Pass a Theme Access password with --password.`, dropping the pull/push half. Left `storeAuthScopes()` off the read-only commands for now; that is a behavior change worth a separate decision. `c51b565c5e`.

dmerand:
I've converted this one to a draft while I revisit the scope of this change. The downstack PR is the more important unblocker.

PR #8554, opened 2026-09-15
Record fetched URL path in `doc fetch` analytics

brandonreid:

`shopify doc fetch` is what the `shopify-app-store-review` skill uses to pull the App Store AI self-review requirements page. Since #7771 the command event already carries `api_key`/`project_type` when run inside an app, but nothing in the public payload says *which* document was fetched — the URL only exists in `args`, which is `privacy: TEXT` and redacted in the warehouse. So today we can see "this app ran `doc fetch`" but not "this app ran a self-review", which blocks correlating skill usage with review outcomes.

- Adds `cmd_doc_fetch_url_path` to the `app_cli3_command` public schema and bumps `MONORAIL_COMMAND_TOPIC` to `1.29`.
- `docFetchService` records `URL.pathname` only — no query string or fragment — after the host allowlist check passes, so the value is always a public shopify.dev docs path.
- Extends `CmdFieldsFromMonorail` with the `cmd_doc_` prefix so the field type-checks through `addPublicMetadata`.

Depends on the schema landing first: Shopify/monorail#24489.

### How to test your changes?

```
pnpm vitest run packages/cli/src/cli/services/commands/doc/fetch.test.ts
```

Or with `SHOPIFY_CLI_ANALYTICS_DEBUG`-style verbose output, run `shopify doc fetch --url https://shopify.dev/docs/apps/launch/app-store-review/app-store-ai-self-review-requirements --verbose` and confirm the analytics payload contains `cmd_doc_fetch_url_path`.

### Measuring impact

- [x] Existing analytics will cater for this addition — `shopify-dw.base.base__monorail_app_cli3_command` picks up new public fields on the 1.x topic.

### Checklist


brandonreid:
Opting to close. Will rely on the private `args` URL for now instead.

PR #8555, opened 2026-09-15
Remove the dependency CVE audit check from app doctor

jplhomer:
### WHY are these changes introduced?

https://github.com/shop/issues-develop/issues/23910

CVE auditing is tricky. Previous support was pinned to just NPM dependencies, and even within those, false positives were detected for dependencies used only in development or tests. `npm audit` also calls out to the public internet which makes it a little iffy as far as exposing private package names to third parties.

This PR removes App Doctor's package-manager CVE audit while a replacement is developed in a future PR.

### WHAT is this pull request doing?

Removes the CVE audit runner, check registration, fallback prompt, and related tests. Cleans up command help and generated metadata, preserves submission coverage using an existing check, and adds a removal changeset. No replacement or absence tests are added.

### How to manually test your changes?

```sh
pnpm shopify app doctor --path /path/to/shopify-app --skip-instructions
pnpm shopify app doctor --path /path/to/shopify-app --json
```

### Checklist

- [ ] I've considered possible cross-platform impacts (Mac, Linux, Windows)
- [ ] I've considered possible [documentation](https://shopify.dev) changes
- [ ] I've considered analytics changes to measure impact
- [x] The change is user-facing — I've identified the correct [bump type](https://github.com/Shopify/cli/blob/main/CONTRIBUTING.md#choosing-the-right-bump-type) (`patch` for bug fixes · `minor` for new features · `major` for [breaking changes](https://github.com/Shopify/cli/blob/main/CONTRIBUTING.md#what-counts-as-a-breaking-change)) and added a changeset with `pnpm changeset add`

PR #8556, opened 2026-09-15
Check for dependency automation configuration files in App Doctor

jplhomer:

Fixes https://github.com/shop/issues-develop/issues/23910

Follow-up to #8555. After removing package-manager CVE audits, App Doctor should flag missing dependency-automation configuration without trying to interpret CI workflows or shell commands.

Adds `MISSING_DEPENDENCY_SECURITY_AUTOMATION`, a low-severity check worth -5 points. The presence of a standard Dependabot or Renovate configuration file satisfies the check; contents are not parsed or validated. This is an adoption signal, not proof that automation is enabled, covers the app, or fixes vulnerabilities.

Keeps bounded reads, path/symlink containment, input hashing, and normal scoring, blocking, trace, and submission handling. Unsafe or unreadable inputs and nested repository boundaries remain unresolved. Hosted integrations, centrally managed configuration, CI workflows, and Renovate configuration embedded in `package.json` are not detected.

No new dependencies. Replaces the previous workflow analyzer and supersedes the GitLab/CircleCI extensions in #8558 and #8559.

```sh
pnpm shopify app doctor --path /path/to/app --skip-instructions
pnpm shopify app doctor --path /path/to/app --skip-instructions --json --blocking low
```

### Checklist

- [x] The change is user-facing — I've identified the correct bump type and added a changeset


packages/app/src/cli/services/app-doctor-engine/scanners/discover.ts
```diff
@@ -599,6 +649,146 @@ export function findSensitiveFiles(appRoot: string): SourceFile[] {
...
+  return error instanceof Error && 'code' in error && (error.code === 'ENOENT' || error.code === 'ENOTDIR')
+}
+
+/**
+ * Resolve an allowlisted path while preserving the difference between absence
+ * and an unsafe or unreadable entry. `realpathSync` follows every intermediate
+ * directory link, allowing contained links but rejecting escapes and dangling
+ * links explicitly.
+ */
+function inspectAllowedPath(appRoot: string, relative: string): AllowedPathResult {
```
dmerand: Could we consolidate this with `containedRepositoryPath` instead of maintaining a second path-containment implementation? After this succeeds, `readRepositoryFile` repeats the `realpath`, containment, and file-type checks. Some custom logic is necessary if we must distinguish an absent optional file from a dangling intermediate symlink, but one shared `missing | file | unresolved` result would make that contract explicit and avoid double validation. Also, cli-kit’s `normalizePath` is `pathe`-backed, so the manual separator replacement below should not be necessary.

packages/app/src/cli/services/app-doctor-engine/scanners/discover.ts
```diff
@@ -599,6 +649,146 @@ export function findSensitiveFiles(appRoot: string): SourceFile[] {
...
+    return {files: [], unresolvedReason: filesystemError(appRoot, error)}
+  }
+
+  const repositoryReason = nestedRepositoryReason(canonicalRoot)
+  if (repositoryReason) return {files: [], unresolvedReason: repositoryReason}
+
+  const files: SourceFile[] = []
+  for (const relative of DEPENDENCY_AUTOMATION_CONFIG_PATHS) {
+    const inspected = inspectAllowedPath(canonicalRoot, relative)
+    if (inspected.unresolvedReason) return {files, unresolvedReason: inspected.unresolvedReason}
```
dmerand: Could we save this unresolved reason and continue checking the remaining allowlisted paths? A safe later configuration should satisfy this presence-only check. For example, an external `.github` symlink currently prevents a safe root `renovate.json` from being detected because the loop returns before reaching it. We should return the stored reason only if no safe configuration is found.

packages/app/src/cli/services/app-doctor-engine/scanners/discover.ts
```diff
@@ -599,6 +649,146 @@ export function findSensitiveFiles(appRoot: string): SourceFile[] {
...
+    if (!isMissingFilesystemEntry(error)) return filesystemError(rootMarker, error)
+  }
+
+  let ancestor = dirname(appRoot)
+  while (true) {
+    const marker = joinPath(ancestor, '.git')
+    try {
+      const stats = lstatSync(marker)
+      if (stats.isDirectory() || stats.isFile()) {
+        return `App root is nested below repository root ${ancestor}`
```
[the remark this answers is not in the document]
jplhomer: Addressed in 19dafdeb2a

packages/app/src/cli/services/app-doctor-engine/scanners/discover.ts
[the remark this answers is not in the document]
jplhomer: Addressed in 19dafdeb2a

packages/app/src/cli/services/app-doctor-engine/scanners/discover.ts
jplhomer: Addressed in 19dafdeb2a

packages/app/src/cli/services/app-doctor-engine/scanners/discover.ts
jplhomer: Addressed in 19dafdeb2a

packages/app/src/cli/services/app-doctor-engine/scanners/discover.ts
```diff
@@ -412,13 +431,122 @@ function readBoundedFile(path: string): RepositoryReadResult {
...
+  const segments = normalizeCliPath(relativePath(absoluteRoot, absolutePath))
+    .split('/')
+    .filter((segment) => segment.length > 0 && segment !== '.')
+  if (segments.includes('..')) return {status: 'unresolved', reason: `${display} escapes the app root`}
+
+  let currentPath = canonicalRoot
+  for (const [index, segment] of segments.entries()) {
+    currentPath = joinPath(currentPath, segment)
+    try {
+      lstatSync(currentPath)
```
dmerand, line 514: This is a performance nit from the LLM review: 

`readRepositoryFile` is used for every source and sensitive file, but this now performs `lstatSync`, `realpathSync`, and another `lstatSync` for every path segment. The previous containment path resolved the complete file once, so this can multiply filesystem calls by path depth across thousands of files. 

One option would be to keep a fast path that resolves and validates the complete file, then walk prefixes only after `ENOENT` or `ENOTDIR` when we need to distinguish an absent optional path from a dangling intermediate symlink? That would preserve the new result model and allowlist behavior without moving the expensive path walk into every repository read.

packages/app/src/cli/services/app-doctor-engine/rules/dependency-automation-rules.ts
```diff
@@ -0,0 +1,54 @@
...
+  const manifests = context.manifests.filter(
+    (manifest) =>
+      Object.keys(manifest.dependencies).length > 0 || Object.keys(manifest.devDependencies ?? {}).length > 0,
+  )
+  if (manifests.length === 0) return {issues: []}
+  const {files, unresolvedReason} = context.dependencyAutomation
+  if (unresolvedReason) return {issues: [], unresolvedReason}
+  if (files.length > 0) return {issues: []}
+
+  const manifest = [...manifests].sort((left, right) => left.path.localeCompare(right.path))[0]!
```
dmerand: I tested this manually and got:

Image

This output can imply that dependency automation should be configured beside `extensions/app-home/package.json`, but Dependabot and Renovate configuration is repository-level. 

Robot sez: It looks like the `sort` also selects `extensions/...` before a root `package.json`. Could we prefer the root manifest as the finding location when it declares dependencies, and describe the missing file as repository-level configuration? We should still inspect nested manifests to determine applicability, but I do not think we should accept configuration files inside extension folders.

dmerand (commented):
Sorry for the review churn. I found a couple more things on manual testing that are worth consideration. The overall approach looks OK, though I still kind of wonder if there are FS primitives we could be sharing with all of the FS reads happening in the app doctor space.

jek (approved):
Tested locally.  Good capability to add.

packages/app/src/cli/services/app-doctor-engine/scanners/discover.ts
jplhomer: Addressed in c8553626e6. Existing files take a single `realpath` again; we only walk prefixes after `ENOENT`/`ENOTDIR` so optional allowlist paths can still distinguish ordinary absence from a dangling or escaping intermediate.

packages/app/src/cli/services/app-doctor-engine/rules/dependency-automation-rules.ts
jplhomer: Addressed in c8553626e6. The finding now prefers the root `package.json` over nested extension manifests, and the copy describes the missing file as repository-level Dependabot or Renovate configuration.

dmerand (approved):
Thanks for the changes! (I got tired of waiting for /snapit and just tested against the branch)

PR #8573, opened 2026-09-17
Retry gateway failures during device authorization

gonzaloriestra:

Related to https://shopify.slack.com/archives/C0ARV62K59C/p1789515820672379

Temporary gateway failures during sign-in immediately fail device authorization, and non-JSON error responses are reported as CLI bugs.

Retry device authorization requests that return HTTP 502, 503, or 504 twice. Respect valid `Retry-After` headers (seconds or HTTP dates), falling back to delays of 200 ms and 400 ms when the header is missing or invalid. If a non-JSON gateway response persists, report an `AbortError` with the existing service-error context. JSON error responses retain their existing handling.

Share the gateway-status helper with the existing GraphQL error classification and record retry analytics. Include regression coverage and a patch changeset.

```sh
pnpm shopify auth login
```

### Checklist


packages/cli-kit/src/private/node/session/device-authorization.ts
```diff
@@ -153,6 +146,42 @@ export async function pollForDeviceAuthorization(code: string, interval = 5): Pr
...
+      'Check your network connection and try again.',
+    )
+  }
+
+  if (!isGatewayErrorStatus(response.status) || gatewayRetriesUsed >= GATEWAY_ERROR_RETRY_LIMIT) {
+    return {response, responseText}
+  }
+
+  const retryNumber = gatewayRetriesUsed + 1
+  const retryDelaySeconds = GATEWAY_ERROR_INITIAL_RETRY_DELAY_SECONDS * 2 ** gatewayRetriesUsed
```
amcaplan: The server may return a `Retry-After` header. In that case, we should respect the header.

amcaplan (changes requested):
The code looks good. One small change I'd like to see, and then I think this PR is ready to go

packages/cli-kit/src/private/node/session/device-authorization.ts
gonzaloriestra: Good point, fixed!

packages/cli-kit/src/private/node/api/status-checks.ts
dmerand, line 1: Small nit that this might be more generically useful/aimed like the other files in this utilty folder e.g. `status-checks`.

dmerand (approved):

packages/cli-kit/src/private/node/api/status-checks.ts
gonzaloriestra: Renamed, thanks!

PR #8575, opened 2026-09-17
Add typed JSON output to app info

gonzaloriestra:

Related to https://github.com/shop/issues-develop/issues/23670.

`app info` lacks a declared JSON contract and omits public data already loaded by the command.

Return typed app information and move terminal formatting into a presenter. Preserve existing JSON fields and `--web-env`, while adding remote app details, cached account identity, discovered configuration files, extension metadata, resolved dev store and system information without additional API requests. Record the field inventory and exclusions beside the schema.

Expose the schema in help and refresh generated documentation. Share `mockAndCaptureStandardStreams()` in CLI Kit for command output tests. Text output, extension filtering and invalid-app exit behavior are preserved.

Validation: 34 focused app-info tests, app type checking, focused lint, Knip, CLI build and generated help/docs checks pass. Tests cover legacy field compatibility, added metadata, credential exclusions, stdout/stderr and exit behavior.

```sh
pnpm shopify app info --path /path/to/app
pnpm shopify app info --path /path/to/app --json
pnpm shopify app info --path /path/to/app --web-env
pnpm shopify app info --path /path/to/app --web-env --json
pnpm shopify app info --json-schema
```

### Checklist


packages/app/src/cli/services/info/types.ts
```diff
@@ -0,0 +1,169 @@
...
+    externalIdentifier: zod.string(),
+    externalName: zod.string(),
+    additionalIdentifiers: zod.array(zod.string()),
+    partnersWebIdentifier: zod.string(),
+    surface: zod.string(),
+    registrationLimit: zod.number(),
+    experience: zod.enum(['extension', 'configuration']),
+    uidStrategy: zod.enum(['single', 'dynamic', 'uuid']),
+  })
+  .passthrough()
```
dmerand: Could we make the dependency on `ExtensionSpecification` explicit instead of coupling the public contract through `.passthrough()` and `JSON.stringify`? This schema documents only a subset, but the snapshot shows passthrough fields such as `clientSteps` and `dependency`; future enumerable internal fields could therefore enter the public JSON without appearing in `--json-schema`.

I'm wondering if we should be fully-explicit with what we support, rather than using passthrough?

packages/eslint-plugin-cli/rules/json-output-command-exceptions.js
dmerand, line 1: Nit (not on this file) -- missing changeset?

packages/app/src/cli/services/info/types.ts
```diff
@@ -0,0 +1,169 @@
...
+    system: systemSchema,
+    devStoreUrl: zod.string().optional(),
+    name: zod.string(),
+    idEnvironmentVariableName: zod.literal('SHOPIFY_API_KEY'),
+    directory: zod.string(),
+    configPath: zod.string(),
+    configuration: configurationSchema,
+    webs: zod.array(webSchema),
+    dotenv: zod
+      .object({path: zod.string(), variables: zod.record(zod.string())})
```
dmerand, line 194: I wonder if we need to redact sensitive fields before making this a public contract? 

Agent lingo about it: The legacy object spread already emits `dotenv.variables` and `_hiddenConfig`, but that does not make them safe to formalize. `.env` values routinely contain credentials, and the newly added raw config-file contents need the same review. `remoteAppInfoSchema` deliberately excludes `apiSecretKeys`; The broader result needs an equivalent allow-list, with tests proving secrets are not emitted, even if that requires an explicit compatibility decision.

packages/app/src/cli/services/info.ts
```diff
@@ -1,98 +1,60 @@
...
-      null,
-      2,
-    )}`
-  } else {
-    const appInfo = new AppInfo(app, remoteApp, organization, project, options)
-    return appInfo.output()
-  }
+    ),
+  )
+  return appInfoJsonOutputSchema.validate(result)
```
dmerand, line 58: Nit from the LLM: Do we want normal text execution to depend on constructing, JSON-stringifying, and validating the entire public JSON payload? The text branch in `renderAppInfoResult` still renders from the original `app`, `remoteApp`, `organization`, and `project` context, using only `result.account`. A JSON-only serialization or schema problem can therefore make `shopify app info` fail even without `--json`. I think either the text presenter should consume `AppInfoResult` as its source of truth, completing the boundary, or JSON-only materialization and validation should remain on the JSON path.

packages/eslint-plugin-cli/rules/json-output-command-exceptions.js
gonzaloriestra: I'm skipping them intentionally to avoid having a trillion changeset entries. Instead, I'd like to add them per areas/groups, like I did here for the [store commands](https://github.com/Shopify/cli/blob/main/.changeset/bulk-store-json-results.md).  
  
So my plan is to add a changeset later (before the release) for all the app commands, or subgroups if they are not ready. Makes sense?

packages/app/src/cli/services/info/types.ts
gonzaloriestra: Good catch! Added the explicit dependency.

packages/app/src/cli/services/info.ts
gonzaloriestra: Agreed. Moved JSON construction and validation off the normal text path. Text renders directly from the loaded context

packages/app/src/cli/services/info/types.ts
gonzaloriestra: Hmm but the purpose of this command is to print app details, why not dotenv vars as well?

Anyway, I think we should preserve the existing fields to avoid a breaking change. We can address redaction or removal separately later. Wdyt?

packages/eslint-plugin-cli/rules/json-output-command-exceptions.js
dmerand: Yep, that makes sense!

packages/app/src/cli/services/info/types.ts
dmerand: Okay, let's keep it for now, and we can re-evaluate once we've got the fuller set of commands in. Perhaps a principle will emerge as we look at more cases, perhaps omitting the ENV values would be to defensive at this stage.

dmerand (approved):

PR #8585, opened 2026-09-18
Add typed JSON output for store delete

gonzaloriestra:

The store deletion command supports JSON but does not expose or validate its output contract.

Adds a discoverable schema for confirmed deletion, pending deletion, and existing JSON errors. Reuses shared store, organization, and error schemas while preserving confirmation and polling behavior. Includes the single release note for the store JSON stack.

### Example output

Illustrative sample data. Text excerpts omit terminal decoration and progress. JSON blocks show the final stdout result; progress and diagnostics go to stderr.

#### `shopify store delete --store disposable-dev-store.myshopify.com --organization-id 1234 --force`

**Text**

```text
Dev store "disposable-dev-store.myshopify.com" deleted successfully.

The store was deleted.
```

**JSON** (`--json`)

```json
{
  "store": {
    "domain": "disposable-dev-store.myshopify.com",
    "deletionRequested": true,
    "deletionConfirmed": true
  },
  "organization": {
    "id": "1234",
    "name": "Acme"
  }
}
```

#### Deletion requested, confirmation still pending

**Text**

```text
Dev store "disposable-dev-store.myshopify.com" deletion was requested, but not confirmed.

Shopify accepted the deletion request, but deletion was not confirmed before the CLI stopped waiting.
The store may still finish deleting asynchronously.
```

**JSON** (`--json`)

```json
{
  "store": {
    "domain": "disposable-dev-store.myshopify.com",
    "deletionRequested": true,
    "deletionConfirmed": false
  },
  "organization": {
    "id": "1234",
    "name": "Acme"
  },
  "message": "Deletion was requested, but has not been confirmed yet. The store may still finish deleting asynchronously."
}
```

**JSON error** (non-interactive invocation without `--force`; exit code 1)

```json
{
  "error": true,
  "message": "Deleting the dev store disposable-dev-store.myshopify.com requires confirmation.",
  "nextSteps": [
    "Use the `--force` flag to skip confirmation when running non-interactively."
  ],
  "exitCode": 1
}
```

```sh
shopify store delete --json-schema
shopify store delete --store disposable-dev-store.myshopify.com --organization-id 1234 --force --json
```

### Checklist

- [x] The change is user-facing — I've identified the correct bump type and added a changeset.


packages/store/src/cli/services/store/delete/types.ts
```diff
@@ -0,0 +1,26 @@
+import {StoreSchema, StoreOrganizationSchema, StoreCommandErrorSchema} from '../types.js'
+import {defineJsonOutputSchema} from '@shopify/cli-kit/node/json-output-schema'
+import {zod} from '@shopify/cli-kit/node/schema'
+
+const DeletedStoreSchema = zod.object({
+  domain: StoreSchema.shape.subdomain,
+  deletionRequested: zod.literal(true),
```
dmerand, line 7: This is kind of funny: we report that we requested the thing the command is supposed to request, and we will never report anything else.

I don't really see harm in it 🤷🏼

dmerand (approved):

packages/store/src/cli/services/store/delete/types.ts
gonzaloriestra: Yeah, it's a bit nonsense, but it's already there and we want to keep the shape... If something fails, it will return the error schema. So there's no situation where we could return false here

PR #8592, opened 2026-09-18
Fix App Doctor secret scanning false positives

jplhomer:

Fixes https://github.com/shop/issues-develop/issues/23957

Early-access partners were flooded with `COMMITTED_SECRET` findings for placeholders, `.env.example` templates, public client IDs, and Stripe publishable keys. A name match plus a non-empty value was enough to score -50 high, which fails the static-tier bar of provable facts.

Require two pieces of evidence before scoring an env/secret file: a known credential format, or a secret-named assignment whose value is not a placeholder. Template env files (`.env.example`, `.sample`, `.template`, `.dist`) only fire on a known token format. Drop `SHOPIFY_API_KEY` / `api_key` and Stripe `pk_` from secret detection. Keep fail-closed for real secrets when git status is unknown, but stop rendering that case as confirmed-tracked.

From an app with a committed `.env.example` of placeholders and a tracked `.env` that contains a real `shpss_` / `shpat_` token:

```
pnpm shopify app doctor --path /path/to/app --skip-instructions
```

Confirm the template is not reported and the real token still is.

### Checklist


packages/app/src/cli/services/app-doctor-engine/rules/secret-rules.ts
```diff
@@ -127,35 +127,92 @@ export function redactText(text: string): string {
 
 const ENV_FILE_PATTERN = /(^|\/)\.env(?:\.[^/]+)?$/
 const NAMED_SECRET_FILE_PATTERN = /(^|\/)(?:\.env\.(?:secrets|keys)|(?:secrets|credentials)\.json)$/
+const TEMPLATE_ENV_FILE_PATTERN = /(^|\/)\.env(?:\.[^/]+)*\.(?:example|sample|template|dist)$/i
+// `SHOPIFY_API_KEY` / `api_key` are client IDs and public; they are not secret names.
 const SECRET_ASSIGNMENT_PATTERN =
-  /(?:api[_-]?key|api[_-]?secret|access[_-]?token|secret[_-]?key|private[_-]?key|password|SHOPIFY_API_KEY|SHOPIFY_API_SECRET)\s*[:=]\s*(?:"[^"\r\n]+"|'[^'\r\n]+'|[^\s#'"\r\n][^#\r\n]*)/i
+  /(?:api[_-]?secret|access[_-]?token|secret[_-]?key|private[_-]?key|password|SHOPIFY_API_SECRET)\s*[:=]\s*(?:"([^"\r\n]*)"|'([^'\r\n]*)'|([^\s#'"\r\n][^#\r\n]*))/gi
```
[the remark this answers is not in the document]
jplhomer: Assignment names now allow optional quotes, so `{ "password": "…" }` in `secrets.json` is extracted. Tracked JSON with a real quoted value scores; a placeholder like `changeme` does not.

packages/app/src/cli/services/app-doctor-engine/checks/COMMITTED_SECRET.md
```diff
@@ -1,9 +1,11 @@
...
-version: 1
+version: 2
 severity: high
 ---
 
 # Committed Secret
 
 Inspect files skipped by deterministic secret scanning for committed credentials. Never quote or reproduce a secret; cite only the file and redacted credential kind, and recommend rotation.
+
+Do not report placeholders, template env files (`.env.example`, `.sample`, `.template`, `.dist`), public client identifiers (`SHOPIFY_API_KEY`, Stripe `pk_`), or files git confirms are untracked and ignored.
```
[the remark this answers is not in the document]
jplhomer: Updated the prompt: template env files are findings only when they contain a known credential format. Placeholders still are not reported — that was the original partner false positive, not “ignore this filename.”

packages/app/src/cli/services/app-doctor-engine/checks/embedded.ts
```diff
@@ -7,7 +7,7 @@ export const EMBEDDED_CHECK_SOURCES: ReadonlyArray<string> = [
...
-  "---\nid: COMMITTED_SECRET\nversion: 1\nseverity: high\n---\n\n# Committed Secret\n\nInspect files skipped by deterministic secret scanning for committed credentials. Never quote or reproduce a secret; cite only the file and redacted credential kind, and recommend rotation.\n",
+  "---\nid: COMMITTED_SECRET\nversion: 2\nseverity: high\n---\n\n# Committed Secret\n\nInspect files skipped by deterministic secret scanning for committed credentials. Never quote or reproduce a secret; cite only the file and redacted credential kind, and recommend rotation.\n\nDo not report placeholders, template env files (`.env.example`, `.sample`, `.template`, `.dist`), public client identifiers (`SHOPIFY_API_KEY`, Stripe `pk_`), or files git confirms are untracked and ignored.\n",
```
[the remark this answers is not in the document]
jplhomer: Regenerated `embedded.ts` from the updated markdown. The agent prompt now matches the scanner: templates are excluded for placeholders, not for known token formats.

packages/app/src/cli/services/app-doctor-engine/rules/secret-rules.ts
```diff
@@ -127,35 +127,92 @@ export function redactText(text: string): string {
 
 const ENV_FILE_PATTERN = /(^|\/)\.env(?:\.[^/]+)?$/
```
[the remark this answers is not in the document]
jplhomer: Aligned `ENV_FILE_PATTERN` with the multi-suffix form (`\.env(?:\.[^/]+)*$`) so `.env.local.example` is classified as an env/template file. Added a regression test that a known token in that path still scores high.

packages/app/src/cli/services/app-doctor-engine/rules/secret-rules.ts
```diff
@@ -125,37 +125,95 @@
   return redacted
 }
 
-const ENV_FILE_PATTERN = /(^|\/)\.env(?:\.[^/]+)?$/
+const ENV_FILE_PATTERN = /(^|\/)\.env(?:\.[^/]+)*$/
 const NAMED_SECRET_FILE_PATTERN = /(^|\/)(?:\.env\.(?:secrets|keys)|(?:secrets|credentials)\.json)$/
+const TEMPLATE_ENV_FILE_PATTERN = /(^|\/)\.env(?:\.[^/]+)*\.(?:example|sample|template|dist)$/i
```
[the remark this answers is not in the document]
jplhomer: Same fix as the sibling alert: env/template classification is now basename-based, so this repeating-dot regex is gone.

packages/app/src/cli/services/app-doctor-engine/rules/secret-rules.ts
```diff
@@ -125,37 +125,95 @@
   return redacted
 }
 
-const ENV_FILE_PATTERN = /(^|\/)\.env(?:\.[^/]+)?$/
+const ENV_FILE_PATTERN = /(^|\/)\.env(?:\.[^/]+)*$/
```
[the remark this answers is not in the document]
jplhomer: Replaced the nested `(?:\.[^/]+)*` env matchers with basename checks (`.env` / `.env.*`, last suffix for templates). Same multi-suffix coverage without overlapping quantifiers.

packages/app/src/cli/services/app-doctor-engine/rules/secret-rules.ts
```diff
@@ -125,37 +125,106 @@ export function redactText(text: string): string {
...
 }
 
-const ENV_FILE_PATTERN = /(^|\/)\.env(?:\.[^/]+)?$/
 const NAMED_SECRET_FILE_PATTERN = /(^|\/)(?:\.env\.(?:secrets|keys)|(?:secrets|credentials)\.json)$/
+const TEMPLATE_ENV_SUFFIXES = new Set(['example', 'sample', 'template', 'dist'])
+// `SHOPIFY_API_KEY` / `api_key` are client IDs and public; they are not secret names.
+// Optional quotes cover JSON/YAML keys (`"password": "…"`) as well as env assignments.
 const SECRET_ASSIGNMENT_PATTERN =
-  /(?:api[_-]?key|api[_-]?secret|access[_-]?token|secret[_-]?key|private[_-]?key|password|SHOPIFY_API_KEY|SHOPIFY_API_SECRET)\s*[:=]\s*(?:"[^"\r\n]+"|'[^'\r\n]+'|[^\s#'"\r\n][^#\r\n]*)/i
+  /["']?(?:api[_-]?secret|access[_-]?token|secret[_-]?key|private[_-]?key|password|SHOPIFY_API_SECRET)["']?\s*[:=]\s*(?:"([^"\r\n]*)"|'([^'\r\n]*)'|([^\s#'"\r\n][^#\r\n]*))/gi
```
lopez-mar: i worry that we're being a bit too broad with this exemption, i know for our apps that the api key is public, but for other services that they might have keys for, they could very much be confidential and offer broad access (e.g., mailchimp).

i know it's a tricky balance here, but can we maybe balance this by looking at the _value_ assigned to a given key? as in, we know what our prefixes are for different shopify keys

fact check me on this, but i think:
- `shpat_` -> is the public prefix i think?
- `shpca_`, `shppa_`, `shpss_`, `shpua_`

lopez-mar: when it comes to leaked credentials, do we care about shopify credentials specifically? or any credentials? because i think we can look for the token prefix patterns maybe? (like shpat, shpss, etc.)

question: ae we aiming to detect leaked Shopify credentials specifically, or credentials more broadly that could compromise the app and merchant data?

because i'm wondering if the deterministic check should focus on recognized secret formats, like how we use prefixes for our secrets (e.g., shpat_, shpss sort of thing). that wouldn't account for legacy secrets that don't have prefixes admittedly, though i'm not sure how common that is. wrt non-shopify secrets, we could also search for things with well known formats like stripe secret keys (e.g., sk_live_ but not sk_test_), github tokens (e.g., ghp_, github_pat_) , etc.

i think this change reduces the false positives which is good, but i think it's worth being explicit about whether we care about non-shopify api keys being detected. i'm not sure but maybe there are things we could flag for agentic review to be verified? just spitballing

packages/app/src/cli/services/app-doctor-engine/rules/secret-rules.ts
```diff
@@ -125,37 +125,106 @@ export function redactText(text: string): string {
...
+  for (const match of content.matchAll(pattern)) {
+    values.push((match[1] ?? match[2] ?? match[3] ?? '').trim())
+  }
+  return values
+}
+
+function containsCommittedSecretEvidence(content: string, templateFile: boolean): boolean {
+  if (containsKnownSecretFormat(content)) return true
+  if (templateFile) return false
+  return secretAssignmentValues(content).some((value) => !isPlaceholderSecretValue(value))
```
lopez-mar, line 195: from some poking i got my agent to do, it looks like `SHOPIFY_API_SECRET=<32 hex characters>` doesn't produce a finding, but enclosing the same value in quotes does still produce a finding

packages/app/src/cli/services/app-doctor-engine/rules/secret-rules.ts
lopez-mar: it seems that blank assignments consume the following line in this regex? because `\s*` crosses the newline and then it thinks whatever is on the next line is the secret value. this was a pre-existing issue and i'm not sure if it's realistic to see these empty assignments, but might be something worth tightening up. if we're parsing assignments we should limit it to horizontal whitespace consumption, not vertical. 

example
``` dotenv
SHOPIFY_API_SECRET=
PORT=3000
```

if we can parse complete declarations that might also help us disambiguate comment only examples and flags like `HAS_PASSWORD=true` (flagged by my agent as something that would currently trigger as well)

jplhomer: >when it comes to leaked credentials, do we care about shopify credentials specifically? or any credentials?

I _think_ our product aim is to try to catch common versioned secrets, even non-Shopify. I'll double-check this assumption!

>wrt non-shopify secrets, we could also search for things with well known formats like stripe secret keys (e.g., sk_live_ but not sk_test_), github tokens (e.g., ghp_, github_pat_) , etc.

We support some of these common ones (Google, Stripe, AWS, Slack) today - see test cases in `packages/app/src/cli/services/app-doctor-engine/tests/secret-safety.test.ts`.

packages/app/src/cli/services/app-doctor-engine/rules/secret-rules.ts
jplhomer: Confirmed — the Shopify API secret format required quotes, so dotenv-style `SHOPIFY_API_SECRET=<32 hex>` missed in source while the quoted form hit. Quotes are optional now; added a regression test for both quoted and unquoted.

packages/app/src/cli/services/app-doctor-engine/rules/secret-rules.ts
jplhomer: Confirmed both. `\s*` after `=` was eating the newline so `SHOPIFY_API_SECRET=` stole `PORT=3000`, and `password` matched inside `HAS_PASSWORD`. Assignments now use horizontal whitespace only and require a name boundary. Tests cover both cases.

packages/app/src/cli/services/app-doctor-engine/rules/secret-rules.ts
```diff
@@ -125,37 +128,106 @@ export function redactText(text: string): string {
...
+  if (!basename) return false
+  return TEMPLATE_ENV_SUFFIXES.has(basename.slice(basename.lastIndexOf('.') + 1).toLowerCase())
+}
+
+function containsKnownSecretFormat(content: string): boolean {
+  return SECRET_PATTERNS.some((pattern) => pattern.regex.test(content))
+}
+
+/**
+ * Name matches are not evidence. Partners commit `.env.example` with
```
dmerand: Nit: wordy/unintelligible comment

packages/app/src/cli/services/app-doctor-engine/rules/secret-rules.ts
```diff
@@ -125,37 +128,106 @@ export function redactText(text: string): string {
...
+ * `SHOPIFY_API_SECRET=your-secret-here`; treating that as a leak is how this
+ * check missed its own static-tier bar (provable facts, near-zero FPs).
+ */
+function isPlaceholderSecretValue(raw: string): boolean {
+  const value = raw.trim()
+  if (value.length === 0) return true
+  if (/^<[^>\n]+>$/.test(value)) return true
```
dmerand: Nit: it'd be good to name these regexes to explain their intent.

dmerand (approved):

packages/app/src/cli/services/app-doctor-engine/rules/secret-rules.ts
lopez-mar: discussed on slack https://shopify.slack.com/archives/C0BT7EGQZ32/p1789748720104359

i've updated the pr to focus on key prefixes, which should hopefully reduce false positives 🤞

packages/app/src/cli/services/app-doctor-engine/rules/secret-rules.ts
lopez-mar: Addressed in df53398 — the placeholder-classifier docblock is gone with the classifier itself, and the surviving comment above SECRET_PATTERNS has been rewritten to state the rule plainly: prefixed value = evidence, secret-sounding name = not evidence.

packages/app/src/cli/services/app-doctor-engine/rules/secret-rules.ts
lopez-mar: No longer applies — the unnamed placeholder regexes this referred to were deleted when the check moved to prefix-based value detection. The remaining regexes are all named: NAMED_SECRET_FILE_PATTERN and the named entries in SECRET_PATTERNS.

PR #8594, opened 2026-09-18
Prevent App Doctor rescans from dropping compiled findings

jek:

Fixes #23971

Warn agents not to rerun the initial scan after compiling findings, because a new deterministic scan replaces the compiled trace.

Test coverage of dry-run and live submission to ensure neither modifies a compiled trace containing agent findings.

### Checklist

Assisted-By: devx/a43d5d42-941f-47ca-ba7d-47133d5a8cc6


packages/app/src/cli/services/security-check.ts
```diff
@@ -56,8 +76,11 @@ export const appSecurityInstructionsPrompt: RenderSelectPromptOptions<AppSecurit
 }
 
 const defaultDependencies: SecurityDependencies = {
-  execute: async ({directory, findingsPath}) => {
-    const appRoot = resolveAppSecurityRoot(directory)
+  resolveRoot: resolveAppSecurityRoot,
```
dmerand, line 81: I think what the LLM is saying here is that integration tests would be helpful 👍🏼 

Please add command-boundary tests for the real dependency wiring. `check.test.ts` mocks `securityCheck`, and the service tests supply dependencies. Use one compiled-trace case and one default-findings case from a nested `--path`. Each case must prove that the command fails and leaves the protected artifacts byte-identical.

packages/app/src/cli/services/security-check.ts
```diff
@@ -101,16 +124,39 @@ function securityReportInput(
...
+): Promise<void> {
+  const traceResult = await dependencies.readTrace(paths.tracePath)
+  if (traceResult.status === 'ok' && hasRecordedAgentReview(traceResult.trace)) {
+    throw new AbortError(
+      'App Security did not start a new scan.',
+      `The existing trace contains agent review results:\n  ${paths.tracePath}\n\nUse the existing trace, or discard the current review and start over:\n  ${formatAppSecurityCommand(commands.clean)}`,
+    )
+  }
+
+  if (await dependencies.findingsFileExists(paths.findingsPath)) {
```
dmerand, line 146: If we pass a custom findings path, this check won't see it. A plain scan can then replace the active review artifacts. Please require the default path for this workflow, or record review state at a fixed path.

packages/app/src/cli/services/app-security-artifacts.ts
```diff
@@ -42,13 +51,27 @@ export async function writeAppSecurityArtifacts(execution: AppSecurityExecution)
   }
 
   await writeAtomicArtifact(paths.reviewPath, `${JSON.stringify(execution.reviewPack, null, 2)}\n`)
+  if (options.clean) {
```
dmerand, line 61: `clean` is silently ignored when `execution.operation` is `compile`, because the function returns before this block. Please reject that combination before the trace write, or use a type that cannot represent it.

packages/app/src/cli/commands/app/security/check.ts
```diff
@@ -27,6 +29,13 @@ Pass \`--findings\` after completing the review pack to validate agent findings
       description: 'Validate agent findings from a JSON file and compile them into the trace.',
       parse: async (input) => resolvePath(input),
       env: 'SHOPIFY_FLAG_APP_SECURITY_FINDINGS',
+      exclusive: ['clean'],
+    }),
+    clean: Flags.boolean({
+      description: 'Discard the current local review and start a new scan.',
+      default: false,
+      exclusive: ['findings'],
+      env: 'SHOPIFY_FLAG_APP_SECURITY_CLEAN',
```
dmerand: Please remove environment support from this destructive flag. With oclif 4.8.3, `SHOPIFY_FLAG_APP_SECURITY_CLEAN=true` bypasses the guard for every inherited scan. Any non-empty value, including `false`, also conflicts with `--findings`.

dmerand (commented):
This works for me + LGTM overall. I have a few agentic findings worth looking at/resolving before we ship.

packages/app/src/cli/services/security-check.ts
jek: A plain scan won’t overwrite a custom findings file. Once those findings are compiled, the guard checks the trace itself, regardless of where the findings came from.

I don’t think we need to restrict findings paths or add separate review state. I’ll add an integration test that compiles findings from a custom path, then verifies that a plain scan refuses to run and leaves the artifacts unchanged. There is a unit test for this already I believe.

packages/app/src/cli/services/security-check.ts
jek: Added

packages/app/src/cli/services/app-security-artifacts.ts
jek: Added as a check.  Type looks subject to churn when we re-do how findings are attached.

packages/app/src/cli/commands/app/security/check.ts
jek: Removed

dmerand (approved):

jplhomer (approved):

PR #8596, opened 2026-09-18
Move App Doctor route authentication review to the agent

lopez-mar:

The deterministic route-auth check treats an unrecognized authentication spelling as a vulnerability. Valid load-context, dependency-injection, and middleware patterns can therefore produce false positives, while matching authentication text does not prove it protects the operation.

Move `UNAUTHENTICATED_ENDPOINT` to agentic-only review and strengthen its prompt to trace the real verification path before protected operations. Preserve the check's identity and unresolved agent execution until a review is reported; deterministic-only scans no longer claim this check ran.

Remove the obsolete matcher and its tests, add a scan-to-review handoff regression, and update the catalog, embedded prompts, and deterministic-only guidance. Authentication review still requires a coding agent; this does not depend on the new agent loop in #8591.

```sh
pnpm shopify app doctor --path /path/to/react-router-app --json --skip-instructions
pnpm shopify app doctor instructions --path /path/to/react-router-app
```

### Checklist

- [x] The change is user-facing and includes a patch changeset.


packages/app/src/cli/services/app-security-engine/rules/js-rules.ts
```diff
@@ -7,8 +7,14 @@ const SHOP_FIELD = 'shop(?:Domain)?'
 const CREDENTIAL =
   /\b(?:accessToken|access_token|sessionToken|session_token|apiSecret|api_secret|clientSecret|client_secret|SHOPIFY_API_SECRET|SHOPIFY_ACCESS_TOKEN)\b/i
 
-export function scanUnauthenticatedEndpoints(files: SourceFile[]): Issue[] {
+/** Retain the template heuristic; context-based authentication needs semantic review. */
```
dmerand: Tiny nit: These comments read like internal monologue, not long-term useful.

packages/app/src/cli/services/app-security-engine/rules/js-rules.ts
```diff
@@ -7,8 +7,14 @@ const SHOP_FIELD = 'shop(?:Domain)?'
...
+export function scanUnauthenticatedEndpoints(files: SourceFile[]): {
+  issues: Issue[]
```
dmerand: Tiny nit: this might be better as a shared type rather than a one-off type, if possible

dmerand (approved):

